Issue #3242, opened 2026-07-21
Design review: cacheInvalidation subscription broadcasts row content (RecordData) to all clients with no permission filter (B50)


AN-BC:
**Needs a design decision — this touches the cross-server cache-invalidation architecture, so it's not a drop-in fix. Tagging @rkihm-bc to lead.**

## The issue (B50, found by the `subscription-isolation` integration work)

`CacheInvalidationResolver.cacheInvalidation` (`packages/MJServer/src/generic/CacheInvalidationResolver.ts`) is a `@Subscription` with **no filter at all** — the code comment even states *"No session filter — every browser connected via WebSocket receives every event."* The publisher (`ResolverBase.PublishCacheInvalidation`) attaches the full row on saves:

```ts
recordData: action === 'save' ? JSON.stringify(entityObject.GetAll()) : undefined,
```

So every authenticated browser receives the **full content of every row anyone saves**, with **zero permission filtering**.

### Blast radius (three dimensions)
- **Cross-tenant** (multi-tenant `MJTenantFilterMiddleware` deployments): tenant X's row content reaches tenant Y's browser. Highest severity.
- **Row-level**: within a tenant, a client receives rows it could not `RunView` (RLS is bypassed on this path).
- **Column/field-level**: `GetAll()` serializes *every* column, so even a client that legitimately can see the row may receive fields it has no field-level permission to read.

Precondition is only "be a connected client" — every authenticated browser already subscribes — so reach is broad. It's "content of rows mutated while connected," not a bulk export, but the cross-tenant leak in particular is serious.

## Why it's not a trivial fix — RecordData is load-bearing

`RecordData` is **used**, not dead weight: the client (`GraphQLDataProvider.SubscribeToCacheInvalidation`) forwards it into a `remote-invalidate` event, and `BaseEngine.applyRemoteRecordData` applies the row **in-place** to update engine caches **without a refetch**. This is the cross-server cache-invalidation performance design. Removing it changes that architecture, which is why this is a design review, not a patch.

Mitigating fact for the options below: the client **already has a graceful fallback** — when `recordData` is absent, `BaseEngine` falls through to a permissioned server refetch (`LoadSingleConfig`), which runs today for every delete and every save-without-recordData.

## Options for the team

[...]

PR #3244, opened 2026-07-21
fix(server): bind statusUpdates push delivery to authenticated connection identity (B49 session-hijack)

AN-BC:
## The bug (B49, SEC-HIGH)

The `statusUpdates` GraphQL subscription (`PushStatusResolver`) filtered **only** on the client-supplied `sessionId`:
```ts
filter: ({ payload, args }) => payload.sessionId === args.sessionId
```
`args.sessionId` is client-chosen and was never checked against the subscriber's authenticated identity (the `context` was available but unused). Since `sessionId` is a per-tab `uuidv4()` persisted in browser local storage, a subscriber who **obtained** another user's `sessionId` (e.g. from a compromised node) would receive that user's pushes: agent-run progress, MCP/test/remote-browser progress, and completion notifications. Not remotely predictable (it's a real UUID), but discoverable — and there was no server-side gate.

Found while authoring the `subscription-isolation` integration bundle (SI2 currently documents this as a known omission).

## The fix

Every push now carries the authenticated **owner's user ID**, and the filter additionally requires it to match the subscriber **connection's** server-authenticated identity:
```ts
payload.sessionId === args.sessionId
  && UUIDsEqual(payload.ownerUserId, context.userPayload.userRecord.ID)
```
The connection identity is established once at WS `onConnect` (validated JWT → `ctx.extra.userPayload`, refreshed on token-expiry reconnect), so it's a stable per-connection trust anchor. `sessionId` still routes a push to the right browser tab, but knowing it is no longer sufficient. **Fails closed** — a missing owner or connection identity never matches. `ownerUserId` is server-side only; it is deliberately **not** exposed on the client-facing `PushStatusNotification` (no user IDs leak to subscribers).

## Centralized so it can't regress

All **~22 publish sites** across 9 files now route through a single vehicle:
```ts
publishStatusUpdate(pubSub, { sessionId, ownerUserId, message })  // ownerUserId REQUIRED
```
- Resolvers use the new `ResolverBase.PublishStatusUpdate(pubSub, sessionId, message, userPayload)` wrapper (sibling of the existing `PublishCacheInvalidation`), which stamps `userPayload.userRecord.ID`.
- Non-resolvers (`FireAndForgetHeartbeat`, `TaskOrchestrator`) call the shared `publishStatusUpdate()` directly.
- Because `ownerUserId` is a **required** field on the payload/params types, the compiler rejects any publish that omits identity — a future publisher physically cannot forget it. After this change, `grep "publish(PUSH_STATUS_UPDATES_TOPIC"` returns **zero** hits outside the one sanctioned publisher.

Also **normalized 3 previously-malformed** publishes (in `RunAIAgentResolver` ×2 and `TaskOrchestrator`) that sent `{ userPayload: JSON.stringify(...), message }` with **no `sessionId`** at all — now proper `{ sessionId, ownerUserId, message }`. Flagged inline with `// NOTE (B49)` comments.

## Tests

- **New**: `pushStatusResolver.filter.test.ts` unit-tests the extracted `statusUpdatesFilter` in isolation — the core hijack rejection (matching sessionId + **different** connection identity → denied), fail-closed on missing owner/identity, and case-insensitive UUID match (SS-upper vs PG-lower). 6/6.
- MJServer builds clean (`tsc` exit 0); **719** existing unit tests pass.
- The end-to-end two-identity assertion (user B truly cannot receive user A's live pushes) needs a multi-connection WebSocket harness the integration suite doesn't have yet — planned as a next-phase item, at which point `subscription-isolation.SI2` flips from "documented" to "tested." This PR's unit test is the direct proof of the filter's security property in the interim.

## Review asks (@rkihm-bc @cadam11)

1. **`ownerUserId` = the operation's initiating user.** For any **system-initiated** push targeting a user's session (if such a flow exists), `userRecord.ID` would be the system actor and the real user's subscription would (correctly, per this model) not match — fail-closed. Please confirm there's no legitimate system-push-to-user flow that this would silently drop; if there is, we'd special-case it.
2. **The 3 normalized malformed sites** changed delivery semantics (they had no `sessionId`, so their pushes likely weren't reaching `statusUpdates` subscribers before). Please sanity-check those notification flows.
3. Scoped intentionally to `statusUpdates`. The **sibling** `cacheInvalidation` unfiltered-broadcast issue (B50) is a separate design review — #3242.


AN-BC:
@cadam11 and @rkihm-BC CC our new integration test suite identified this security vulnerability. I wanted to get both of your eyes on this before we shipped it so this will not make it into `5.49.0` but the next build

rkihm-BC (approved):
**Approve — merge as-is.** Reviewed the fix in depth and verified the security property, the centralization, and (critically) review-ask #1 against real deployments.

## The fix is correct
- The subscriber trust anchor is real and non-spoofable: `context.userPayload` is set once at WS `onConnect` from a validated JWT (`index.ts` `useServer` → `ctx.extra.userPayload`, refreshed on token-expiry reconnect), not per-message client input. So `UUIDsEqual(payload.ownerUserId, context.userPayload.userRecord.ID)` gates delivery on server-authenticated identity — knowing a victim’s client-chosen `sessionId` is no longer sufficient.
- Fails closed on missing owner / missing connection identity / undefined context, and matches UUIDs case-insensitively (SS-upper vs PG-lower) — the unit test covers all of these, including the money case (matching sessionId + different connection identity → denied).
- Centralization is genuinely regression-proof: confirmed on-branch that `grep "publish(PUSH_STATUS_UPDATES_TOPIC"` returns exactly one hit (inside `publishStatusUpdate`), and because `ownerUserId` is a **required** field, a future publisher physically cannot omit identity.
- No identity leak: the client-facing `PushStatusNotification` ObjectType exposes only `message`/`date`/`sessionId`; `ownerUserId` is server-side only.
- `userRecord.ID` is reliably populated across all auth paths (JWT, magic-link, user API key, system API key), so legitimate same-user subscribers won’t be spuriously dropped.

## Review-ask #1 (system-initiated push-to-user) — answered: no such flow needs special-casing
The only system-identity agent path is `RunAIAgentSystemUser` (`@RequireSystemUser`), which publishes progress under the **system** identity. Its sole caller is MJ’s `graphQLSystemUserClient` as a **server-to-server `query` that awaits the full result** — it does not rely on the push subscription for a UI, so nothing is dropped.

I also checked two real MJ deployments for any legitimate cross-identity push-to-user flow:
- **Skip (open-app):** zero custom publishes to the topic, zero `statusUpdates` subscriptions, zero custom `@Subscription` resolvers, and no `RunAIAgentSystemUser` calls. Skip’s service account holds only a scoped inbound callback API key. Agent progress flows through the standard runner `onProgress` callback and is published by `RunAIAgentResolver` under the **conversation-initiating (real) user’s** identity → `ownerUserId` == subscriber. Not affected.
- **CDP:** zero custom publishes/subscriptions in TS (the only `statusUpdates` hits are generated type defs). Custom agent-progress UI is **polling-based** (`RunView` on `MJ: AI Agent Runs`/`Run Steps`), and scheduled jobs run under system identity but write to DB entities and push nothing to a user session. Not affected.

**Conclusion:** no known system-push-to-user flow is broken by the fail-closed behavior. Residual caveat is the honest one — this covers Skip + CDP, not every downstream deployment — but since the entire surface has exactly one system-identity publisher and it’s server-to-server await-the-result, the risk of a hidden real-user-progress dependency is very low.

## Non-blocking nits (optional follow-up)
1. Two now-dead imports of `PUSH_STATUS_UPDATES_TOPIC`: `RemoteBrowserActionResolver.ts` (unused) and `RealtimeClientSessionResolver.ts` (only referenced in a `{@link}` doc comment).
2. Minor inconsistency: `ResolverBase.PublishStatusUpdate` reads identity defensively (`userPayload?.userRecord?.ID ?? ''`), but the direct callers (`startLivenessPulse` in RunAIAgent/RunTest, `TaskOrchestrator`) use unguarded `.userRecord.ID`. Safe in known paths; mirroring the optional-chaining everywhere would be more robust.

## Ask #2 (3 normalized malformed sites)
Agree with the analysis — those pushes had no `sessionId` and could never match a subscriber before, so this is a fix, not a regression. Only thing to eyeball is whether the in-app notification UX might now show a duplicate if the same notification also arrives via another path.

Nice work — the required-field design is what makes this durable rather than a one-time patch.

AN-BC:
> **Approve — merge as-is.** Reviewed the fix in depth and verified the security property, the centralization, and (critically) review-ask #1 against real deployments.
> 
> ## The fix is correct
> * The subscriber trust anchor is real and non-spoofable: `context.userPayload` is set once at WS `onConnect` from a validated JWT (`index.ts` `useServer` → `ctx.extra.userPayload`, refreshed on token-expiry reconnect), not per-message client input. So `UUIDsEqual(payload.ownerUserId, context.userPayload.userRecord.ID)` gates delivery on server-authenticated identity — knowing a victim’s client-chosen `sessionId` is no longer sufficient.
> * Fails closed on missing owner / missing connection identity / undefined context, and matches UUIDs case-insensitively (SS-upper vs PG-lower) — the unit test covers all of these, including the money case (matching sessionId + different connection identity → denied).
> * Centralization is genuinely regression-proof: confirmed on-branch that `grep "publish(PUSH_STATUS_UPDATES_TOPIC"` returns exactly one hit (inside `publishStatusUpdate`), and because `ownerUserId` is a **required** field, a future publisher physically cannot omit identity.
> * No identity leak: the client-facing `PushStatusNotification` ObjectType exposes only `message`/`date`/`sessionId`; `ownerUserId` is server-side only.
> * `userRecord.ID` is reliably populated across all auth paths (JWT, magic-link, user API key, system API key), so legitimate same-user subscribers won’t be spuriously dropped.
> 
> ## Review-ask #1 (system-initiated push-to-user) — answered: no such flow needs special-casing
> The only system-identity agent path is `RunAIAgentSystemUser` (`@RequireSystemUser`), which publishes progress under the **system** identity. Its sole caller is MJ’s `graphQLSystemUserClient` as a **server-to-server `query` that awaits the full result** — it does not rely on the push subscription for a UI, so nothing is dropped.
> 
> I also checked two real MJ deployments for any legitimate cross-identity push-to-user flow:
> 
> * **Skip (open-app):** zero custom publishes to the topic, zero `statusUpdates` subscriptions, zero custom `@Subscription` resolvers, and no `RunAIAgentSystemUser` calls. Skip’s service account holds only a scoped inbound callback API key. Agent progress flows through the standard runner `onProgress` callback and is published by `RunAIAgentResolver` under the **conversation-initiating (real) user’s** identity → `ownerUserId` == subscriber. Not affected.
> * **CDP:** zero custom publishes/subscriptions in TS (the only `statusUpdates` hits are generated type defs). Custom agent-progress UI is **polling-based** (`RunView` on `MJ: AI Agent Runs`/`Run Steps`), and scheduled jobs run under system identity but write to DB entities and push nothing to a user session. Not affected.
> 
> **Conclusion:** no known system-push-to-user flow is broken by the fail-closed behavior. Residual caveat is the honest one — this covers Skip + CDP, not every downstream deployment — but since the entire surface has exactly one system-identity publisher and it’s server-to-server await-the-result, the risk of a hidden real-user-progress dependency is very low.
> 
> ## Non-blocking nits (optional follow-up)
> 1. Two now-dead imports of `PUSH_STATUS_UPDATES_TOPIC`: `RemoteBrowserActionResolver.ts` (unused) and `RealtimeClientSessionResolver.ts` (only referenced in a `{@link}` doc comment).
> 2. Minor inconsistency: `ResolverBase.PublishStatusUpdate` reads identity defensively (`userPayload?.userRecord?.ID ?? ''`), but the direct callers (`startLivenessPulse` in RunAIAgent/RunTest, `TaskOrchestrator`) use unguarded `.userRecord.ID`. Safe in known paths; mirroring the optional-chaining everywhere would be more robust.
> 
> ## Ask #2 (3 normalized malformed sites)
> Agree with the analysis — those pushes had no `sessionId` and could never match a subscriber before, so this is a fix, not a regression. Only thing to eyeball is whether the in-app notification UX might now show a duplicate if the same notification also arrives via another path.
> 
> Nice work — the required-field design is what makes this durable rather than a one-time patch.

Thanks @rkihm-BC after @cadam11 has a chance to share his comments here I'll have CC take care of those minor bits and anything Craig suggests.

AN-BC:
@cadam11 merging this now, but please do let me know if you have any input per the above. Thanks.

PR #3249, opened 2026-07-22
Persist vector-DB record identifiers on Content Items (+ new ContentItemChunk entity)

topherHunnewell:
---
Persist vector-DB record identifiers on Content Items (+ new ContentItemChunk entity)

Summary

Content autotagging vectorizes each Content Item and upserts the embedding(s) into the configured vector database, but MemberJunction never persisted which vector record an item maps to. Two gaps:

1. No back-reference to the stored vector. After embedding + upsert, the vector-DB record identifier was discarded, so there was no way to trace a Content Item to its stored vector.
2. No chunk → item mapping. When an item's text is too large for one embedding, AutotagBaseEngine.VectorizeContentItems splits it into multiple chunks and creates several vectors that all point back to the same Content Item — with no record of which portion of the item each vector represents.

This PR closes both gaps.

Changes

Schema (migration V202607221540__v5.50.x__ContentItem_VectorRecordID_And_ContentItemChunk.sql)
- ContentItem.VectorRecordID NVARCHAR(100) NULL — the vector-DB record id for items embedded as a single vector.
- New ContentItemChunk entity — ContentItemID (FK → ContentItem), Sequence (int), Text (nvarchar(max)), VectorRecordID (nvarchar(100)), with UQ_ContentItemChunk_ContentItemID_Sequence so a parent can't have two chunks at the same ordinal.

Engine (AutotagBaseEngine.VectorizeContentItems)
- After a successful upsert, persist the record ids the engine already generates (sha1('content-item_<id>'), _chunkN suffix per chunk):
  - Single-chunk item → set ContentItem.VectorRecordID.
  - Multi-chunk item → write one ContentItemChunk row per chunk (ordered), deleting any prior chunk rows first so re-vectorization is idempotent and never collides with the unique constraint.

Design notes

- The persisted VectorRecordID is MJ's own deterministic key (the id used to upsert the vector), not a value returned by the vector DB.
- For multi-chunk items, ContentItem.VectorRecordID is intentionally left NULL — the ContentItemChunk table is the source of truth for per-chunk provenance.
- Migration is hand-written DDL with mj codegen output appended per the migrations/CLAUDE.md separator convention (entity/field metadata, vwContentItemChunks, spCreate/spUpdate/spDelete, permissions).

Testing

✅ Beyond unit tests — live end-to-end verification against SQL Server

A self-cleaning integration script (packages/ContentAutotagging/scripts/verify-vector-persistence.mts) exercises the real stack against the live database — real SQLServerDataProvider, the CodeGen-generated entity classes, and the generated spCreate/spUpdate/spDelete procedures. The embedding provider and vector-DB upsert are stubbed (no external calls / no cost); this is valid because the persisted id is MJ's deterministic key, independent of the embedding vectors or anything the vector DB returns. Everything below that stub line runs for r

The script seeds a minimal content fixture (source-source + a short item → 1 chunk and a long item →multiple chunks), runs VectorizeContentItems, and asserts against the actual persisted rows. All 12 checks pass:

- Single-chunk item — ContentItem.VectorRecordID populated, equals the deterministic id, and no ContentItemChunk rows created.
- Multi-chunk item — VectorRecordID left NULL; one ordered ContentItemChunk row per chunk with correct sequences (0..n-1), per-chunk ids (<base>, <base>_chunk1, …), and the parent ContentItemID.
- Idempotent re-vectorization — a second run keeps the rows (all-new IDs), and produces no duplicatesequences (the unique constraint holds).

Run from repo root:
  npx tsx packages/ContentAutotagging/scripts/verify-vector-persistence.mts
Exit 0 on all-pass, 1 on any failure; fixture removed on exit.

✅ Migration validated from scratch (fresh DB)

The single combined migration file (hand DDL + appended CodeGen output) was applied to a freshly-created database — drop → create
→ mj migrate (12 migrations applied, success) — theUD validation passed (381 entities checked).
Confirmed via direct SQL that the migration alone pD column + EntityField, the ContentItemChunk table +
entity metadata (BaseView = vwContentItemChunks, En all three stored procedures. Verified the migration
does not carry any unrelated CodeGen churn.

✅ Unit tests

ContentAutotagging: 171 passed (7 files), including 3 new tests for the persistence logic (single-chunk column, multi-chunk ordered rows + null column, delete-first idempotency).

✅ Build

npx turbo build for the affected packages and theirssful.

Notes for reviewers

- The migration's v5.50.x label and timestamp are set for the post-5.49.0 cycle (orders after the current tip of next).
- verify-vector-persistence.mts is included as a decan be promoted into the@memberjunction/testing-integration suite as a dete.


.changeset/content-item-vector-record-tracking.md
```diff
@@ -0,0 +1,14 @@
+---
+"@memberjunction/content-autotagging": patch
```
AN-BC: migration == minor

packages/ContentAutotagging/src/Engine/generic/AutotagBaseEngine.ts
```diff
@@ -1616,6 +1619,99 @@ export class AutotagBaseEngine extends BaseEngine<AutotagBaseEngine> {
...
+        // Remove any prior chunk rows for this item.
+        const existing = await new RunView().RunView<MJContentItemChunkEntity>({
+            EntityName: 'MJ: Content Item Chunks',
+            ExtraFilter: `ContentItemID='${contentItemID}'`,
+            ResultType: 'entity_object'
+        }, contextUser);
+        if (existing.Success) {
+            for (const row of existing.Results) {
+                const deleted = await row.Delete();
+                if (!deleted) {
```
AN-BC: 2 thing:
1. If any part of the operation fails, I would argue the whole thing should roll back? Or do we not care? If we do, need transactions wrapping all of this
2. You are unnecessarily awaiting every single db change, instead you could fire off all of the deletes and creates in parallel and push the promise into an array and `await Promise.all()` at the end and check status there and then commit if they're all good or rollback otherwise. More efficient by far and also can be done in transaction if desired.

AN-BC: if doing transaction use the `DatabaseProviderBase.BeginTransaction/CommitTransaction` **not** TransactionGroups, those are used for bundling client-side requests to send them as a single trans to the server in one network round trip, here you want to use true server side transactions.

packages/ContentAutotagging/src/Engine/generic/AutotagBaseEngine.ts
```diff
@@ -1616,6 +1619,99 @@ export class AutotagBaseEngine extends BaseEngine<AutotagBaseEngine> {
...
+     */
+    private async replaceContentItemChunks(
+        contentItemID: string,
+        chunks: { chunkIndex: number; text: string; vectorRecordID: string }[],
+        contextUser: UserInfo
+    ): Promise<void> {
+        const md = this.ProviderToUse;
+
+        // Remove any prior chunk rows for this item.
+        const existing = await new RunView().RunView<MJContentItemChunkEntity>({
```
AN-BC: Invalid code - you need to use the `providerToUse` paradigm where you thread in the provider you want. So for RunView and also for md. In the case of `const md = this.ProviderToUse` you are good.

The RunView provider is simply a cast of the same object so you would do `const rv = this.ProviderToUse as unknown as IRunViewProvider` and then you're good. If you don't do this `new RunView()` uses the **global** database connection/provider, which is very bad in transaction management or in general for multi-user scenarios. Always use provider to use on server as that is the connection allocated for your specific request chain. @topherHunnewell @dray-mc @rkihm-BC

AN-BC (changes requested):
LGTM overall, some changes to make though.

migrations/v5/V202607230133__v5.50.x__ContentItem_ParentID_DisplayLink_And_Chunk_Status_Fields.sql
```diff
@@ -0,0 +1,2305 @@
+/*
+    ContentItem + ContentItemChunk field additions
+
```
AN-BC: no need to have two migrations, that is expensive and this woudl also break when applied in another env. Why? because you are altering the same table sequentially without running R scripts. Solution is, don't do this inefficiently as shown here, instead combine this with your original DDL, get a clean DB, reapply, run CodeGen again and then test.

AN-BC (commented):
only issue is need for unified migrations. @dray-mc  should look at the actual chunking algo for a few minutes to ensure he is on board with it.

migrations/v5/V202607240220__v5.50.x__ContentItem_VectorRecordID_And_ContentItemChunk.sql
```diff
@@ -0,0 +1,3064 @@
...
+    CreatedAt/UpdatedAt timestamp columns + FK auto-indexes are all generated by `mj codegen`
+    after this migration runs. This file contains hand-written DDL only.
+*/
+
+-------------------------------------------------------------------------------
+-- ContentItem — new columns (single ALTER) + self-FK
+-------------------------------------------------------------------------------
+ALTER TABLE ${flyway:defaultSchema}.ContentItem ADD
+    VectorRecordID NVARCHAR(100) NULL,
+    ParentID UNIQUEIDENTIFIER NULL,
```
AN-BC, line 30: missing an sp_addextendedproperty for `ParentID` I think

AN-BC (approved):
LGTM, but given the missed documentation extended property for the `ParentID` col, you need to update the migration to include that and run the Codegen cycle again as that documentation pushes to the DB and then downstream from there into other spots.

Other than that this LGTM, so do that and then merge please @topherHunnewell 

@dray-mc for visibility.

PR #3250, opened 2026-07-22
feat(communication): push-notification subscription abstraction + MS Graph implementation

bc-izygmunt:
## Summary

Adds a **provider-agnostic push-notification subscription abstraction** to the communication framework, and implements it across **every provider that supports inbound push** — MS Graph, Gmail, Twilio, and SendGrid. Consumers that ingest inbound messages can subscribe to "new message" push instead of polling `GetMessages` on a timer — dropping thousands of mostly-empty sweeps per deployment per day.

**Packages touched:**
- `@memberjunction/communication-types` (`packages/Communication/base-types`)
- `@memberjunction/communication-ms-graph`, `-gmail`, `-twilio`, `-sendgrid`

## Why

Every consumer of the MJ communication framework polls `GetMessages()` on a timer, the overwhelming majority of sweeps finding nothing — wasted remote-API quota, DB work, and CPU. Every major messaging service offers push, but MJ's providers exposed none of it. This PR adds the abstraction and implements it for all four providers whose services support inbound push, through the same provider interface consumers already use.

### base-types (`@memberjunction/communication-types`)
- **Five push methods** on `BaseCommunicationProvider` — `CreateSubscription`, `RenewSubscription`, `DeleteSubscription`, `ParseNotification`, `getSubscriptionCapabilities` — all with **not-supported defaults**. Fully backward compatible: existing providers compile and behave identically with zero changes.
- **`SupportsPush` gate** — a `public get SupportsPush(): boolean` that **derives** from `getSubscriptionCapabilities()` (`!== undefined`). Defaults `false`; a provider opts in simply by returning capabilities — no hand-maintained flag to drift out of sync. Lets callers cleanly short-circuit: `if (provider.SupportsPush) { ... }`. *(Per review request from @AN-BC.)*
- **HINT vs. INLINE delivery** — `NormalizedNotification` gains an optional inline `Message: GetMessageMessage`. HINT-mode providers (Graph, Gmail) carry `MessageIDs` and the consumer re-fetches; INLINE-mode providers (Twilio, SendGrid) carry the parsed message directly (no re-fetch — SendGrid in particular has no inbound-retrieval API). This models the same inline-or-pointer duality already established across MJ (`FileOutputRef` `fileData`/`fileId`, `ArtifactVersion.ContentMode` Text/File, `MediaOutput` `data`/`url`).
- **Generalized capability invariant** — `SubscriptionCapabilities` gains `SupportsSubscriptionManagement` and `DeliversPayloadInline`. The old strict "capabilities IFF all four ops" invariant is relaxed into two honest push shapes: **subscription-managed** (create/renew/delete + parse) and **inbound-parse** (parse, with optional management and no expiry). The common denominator is `ParseNotification`; `SupportsPush ⇔ capabilities defined`.

### Providers
- **MS Graph** — full subscription-managed push (change-notification subscriptions; HINT mode). Now also reports the two new capability fields; `SupportsPush` derives `true`.
- **Gmail** — full subscription-managed push: `users.watch` (create) / re-watch (renew, ~7-day expiry) / `users.stop` (delete), Pub/Sub push parse (HINT mode). Requires a Pub/Sub topic granting publish rights to `gmail-api-push@system.gserviceaccount.com` (documented, not provisioned).
- **Twilio** — inbound-parse push: `CreateSubscription` points a phone number's inbound `SmsUrl` at the consumer URL (resolving E.164 → SID), `DeleteSubscription` clears it (idempotent); no expiry → no renew. `ParseNotification` verifies `X-Twilio-Signature` (HMAC-SHA1 over URL + params) and delivers the SMS **inline**.
- **SendGrid** — inbound-parse push (was send-only): `CreateSubscription`/`DeleteSubscription` manage the Inbound Parse hostname→URL mapping via `/v3/user/webhooks/parse/settings`; `ParseNotification` parses the `multipart/form-data` POST and delivers the full email **inline**. Inbound Parse is **unsigned** — documented endpoint-protection caveat (consumer owns a URL secret / network controls).

## Design principles (per PRD)
- **Providers stay stateless** — the consumer persists subscription IDs/expirations/secrets and schedules renewals.
- **Notifications default to hints** — HINT-mode consumers re-fetch through authenticated pull, so a forged notification is harmless. INLINE-mode providers (whose transport delivers the payload) additionally authenticate (Twilio HMAC) or document the out-of-band protection requirement (SendGrid).
- **No HTTP/routing code in the framework** — the consumer owns the endpoint and passes a transport-neutral `WebhookNotificationInput`.

## Review resolution (was: out of scope)

The initial PR shipped the abstraction + MS Graph only, deliberately deferring the other providers. Per @AN-BC's review, this revision:
1. Adds the `SupportsPush` gate (derived, drift-free).
2. Implements push for **all** providers whose services support it — Gmail, Twilio, SendGrid.
3. Confirms **Expo Push** is outbound-only (no inbound to subscribe to) and stays unsupported.

SendGrid initially looked like a poor fit (send-only, unsigned, payload-carrying); rather than force it, the `NormalizedNotification` inline-`Message` generalization makes it — and Twilio — first-class, following MJ's existing inline-or-pointer precedent.

## Testing

**Unit** — **223 tests pass** across the five packages (base-types 83, MS Graph 32, Gmail 41, Twilio 32, SendGrid 35), including hostile-input `ParseNotification` fuzz tests, the `SupportsPush`/capability invariant assertions, and per-provider CRUD + parse coverage with mocked clients. All five packages **build clean**.

**Live smoke** (real credentials, throwaway harnesses, not committed):
- **MS Graph** — validated against LIVE Microsoft Graph:
  - *Smoke tier* (no public endpoint): app-only auth, `Mail.Read` permission, request shaping, verbatim error surfacing, real 404 → idempotent delete, real 404 → debuggable renew message — **7/7**.
  - *Full loop* (behind a tunnel): **create → real validation handshake → renew → real inbound push notification received & normalized → clientState verified constant-time → delete**. Every path exercised end-to-end.
  - Two behavioral findings for downstream consumers (no code change — handled correctly): Graph **re-validates the endpoint on renew** (renewal must run where the listener is reachable), and the notification `Identifier` arrives as the AAD object GUID rather than the email (which is why the design treats `Identifier` as a convenience and routes by `SubscriptionID`/`ClientState`).
- **Twilio** — full round trip: `CreateSubscription` (E.164 → real SID, `SmsUrl` set) → `DeleteSubscription` (cleared) → `ParseNotification` with a **real HMAC signature** (valid accepted, tampered rejected, inline `Message` + `MessageSid` populated). ✅
- **Gmail** — live auth + `users.stop` (`DeleteSubscription`) succeeded; `users.watch` correctly fail-fasts without a Pub/Sub topic (guard + unit-tested). ✅
- **SendGrid** — `ParseNotification` multipart parser validated live (inline `Message` + envelope-derived identifier). ✅

## Backward compatibility
Additive. Providers not implementing push report the new ops as unsupported and `SupportsPush === false`. No new runtime dependencies (Gmail/Twilio reuse their existing SDKs; SendGrid CRUD uses global `fetch`).


AN-BC (commented):
Great improvement but can you have CC add the feature for the other Comms providers that support it please and also in the `BaseProvider` also add a new `public get SupportsPush` or similar property that defaults to false and then in the subclasses that support it override to return true, that way the logic is simple to check that if true we can do stuff otherwise it is a clean short circuit?

bc-izygmunt:
> Great improvement but can you have CC add the feature for the other Comms providers that support it please and also in the `BaseProvider` also add a new `public get SupportsPush` or similar property that defaults to false and then in the subclasses that support it override to return true, that way the logic is simple to check that if true we can do stuff otherwise it is a clean short circuit?

Added the SupportsPush gate and implemented for Gmail, Twilio, and SendGrid as well.

packages/Communication/base-types/src/BaseProvider.ts
```diff
@@ -981,6 +1287,137 @@ export abstract class BaseCommunicationProvider {
...
+    /**
+     * Returns this provider's subscription capabilities, or `undefined` when subscriptions
+     * are not supported. Consumers use this to schedule renewals and validate change types
+     * generically. Override in subclasses that support subscriptions.
+     *
+     * INVARIANT: returning a defined value here REQUIRES the four subscription operations
+     * to appear in {@link getSupportedOperations}.
+     * @returns SubscriptionCapabilities when supported, otherwise undefined
+     */
+    public getSubscriptionCapabilities(): SubscriptionCapabilities | undefined {
```
AN-BC: convention violation, public methods must be `PascalCase` please.

AN-BC (approved):
LGTM, one minor tweak, outside of this, please merge when ready. Great work @bc-izygmunt  very excited about this optimization!

Issue #3257, opened 2026-07-22
test(integration): suite cannot run against PostgreSQL — SS/PG runtime parity is intended but unimplemented and untracked


SDesai-BC:
## Summary

MemberJunction's integration test suite is intended to run **twice per build — once against SQL Server, once against PostgreSQL** — so that backend parity is actually verified rather than assumed. In the words of the engineer who built the suite:

> "intent is to do the integration test suite twice for each build, once for SQL and once for PG, so we're testing it with both back ends — that's a key part of the idea to ensure SQL and PG parity."

`DEPLOYMENT.md` Step 8 captures this near-verbatim:

> **"The stated intent is to run the integration suite twice per build — once per backend — for SS/PG parity. That is a roadmap item, not yet a release step."**

This capability is **not implemented**, and until this issue it was **not tracked anywhere** — it existed only as prose inside `DEPLOYMENT.md`. A search of open issues for `postgres parity`, `pg-parity`, `postgres integration`, and `platform matrix` returned nothing relevant. This issue exists to give the gap an owner and a place in planning.

## Current coverage

From `DEPLOYMENT.md` Step 8's own table:

| Parity | Covered? | By what |
|---|---|---|
| Migrations apply on PG | ✅ Yes | Step 8 + `pg-migrations.yml` (a `postgres:17` service running `mj migrate` with `DB_PLATFORM: postgresql`) |
| CodeGen / External-Data-Source PG paths | ✅ Partly | `pg-migrations.yml`, `eds-integration.yml` |
| **Integration suite on PG** (runtime behaviour: RunView/RunQuery SQL generation, provider code paths, UUID casing, identifier quoting) | ❌ **No** | Nothing. Zero PG bundles exist |

Every release currently ships PostgreSQL with **migration parity verified and runtime parity unverified**. `.github/workflows/integration.yml` provisions SQL Server 2022 only (`mcr.microsoft.com/mssql/server:2022-latest`); there is no platform matrix.

## Why it can't be done today

This is not achievable by configuration or by provisioning a database. All three blockers are in code.

### 1. The testing CLI is SQL-Server-hardcoded and has no PG driver to switch to

`packages/TestingFramework/CLI/src/lib/mj-provider.ts` imports `mssql` and `setupSQLServerClient` and builds an mssql pool with no platform branch:

```ts
import { setupSQLServerClient, SQLServerProviderConfigData, UserCache } from '@memberjunction/sqlserver-dataprovider';
import sql from 'mssql';
[...]

Issue #3254, opened 2026-07-22
check-pg-migration-parity.mjs only checks file existence, so empty/stub PG counterparts pass silently


SDesai-BC:
## Summary

`scripts/check-pg-migration-parity.mjs` — the L1 verification layer for PostgreSQL migration parity in the release process — only tests whether a counterpart `.pg.sql` file **exists**. It never inspects the file's content. An empty or stub counterpart therefore passes the gate silently.

Discovered during the MemberJunction v5.49.0 release build on 2026-07-22.

## Evidence

The script's success path, verbatim:

```js
if (missing.length === 0) {
    console.log(`PG parity OK — all ${tsqlFiles.length} T-SQL V-migrations have PG counterparts (or are documented exclusions).`);
}
```

Existence is the entire test. Git history shows a single commit for this file — `20a8c68735`, 2026-05-30, bc-izygmunt — so it has never checked content since inception.

## Consequence

The two failure modes are asymmetric, and the dangerous one is the invisible one:

| Failure mode | Parity check | Outcome |
|---|---|---|
| **Missing** `.pg.sql` | FAILS loudly | gets fixed before release |
| **Empty / stub** `.pg.sql` | PASSES silently | **ships** |

This is why v5.45's 126-byte `Metadata_Sync.pg.sql` marker shipped, and why three of the v5.49 stub outputs would have shipped had they not been caught by manually diffing output line counts against source line counts.

## Suggested fix (options — deliberately not implemented)

Any one of these would have caught v5.45 **and** all three of the v5.49 stubs automatically:

1. Fail when a `.pg.sql` body past the standard header block is empty.
2. Fail when a `.pg.sql` is below some small fraction of its SQL Server source's size, absent a documented exclusion.
3. Fail when the file's only statements are bare `;`.

This is the single cheapest change available that converts *"we got lucky nine releases running"* into an actual gate. It requires no changes to the converter itself and no schema knowledge — just reading the bytes that are already on disk.

---

**Related:** #3252 (the converter bug that produces the stubs this gate fails to catch), #3253 (the shipped production instance this gate would have caught)

SDesai-BC:
## Update — confirmed blind in practice, and a concrete design suggestion from the v5.49.0 build

## Confirmed blind, empirically

[...]

Issue #3253, opened 2026-07-22
v5.45 Metadata_Sync.pg.sql shipped as a 126-byte marker — PG deployments through v5.45 are missing that release's curated metadata


SDesai-BC:
## Summary

`migrations-pg/v5/V202607071019__v5.45.x__Metadata_Sync.pg.sql` shipped as a **126-byte marker file** instead of the release's metadata DML. Any PostgreSQL deployment that migrated through v5.45 received **none** of that release's curated metadata from migrations.

This is a **confirmed instance of the silent-drop converter bug reaching production.**

Discovered during the MemberJunction v5.49.0 release build on 2026-07-22 while auditing the committed PG ledger.

## Evidence

The committed file is **2 lines / 126 bytes**, in full:

```sql
-- V202607071019__v5.45.x__Metadata_Sync.sql — no DDL to translate.
-- Metadata is re-seeded via `mj sync push` against PG.
```

Its SQL Server source, `migrations/v5/V202607071019__v5.45.x__Metadata_Sync.sql`, is **12,041 lines**.

Compared against neighbouring releases' PG metadata-sync counterparts:

| Release | PG bytes | `DO $mj$` blocks |
|---|---|---|
| v5.44 | 733,519 | 295 |
| **v5.45** | **126** | **0** |
| v5.47 | 26,578 | 12 |
| v5.48 | 20,336 | 9 |

v5.44 before it and v5.47 / v5.48 after it are healthy self-seeding DML of the form `DO $mj$ ... PERFORM __mj."spCreateAIModel"(...) ... END $mj$`. v5.45 is the single release where it collapsed to a marker.

## Impact

A PostgreSQL deployment that migrated through v5.45 is **silently dependent** on someone having run `mj sync push` afterwards to obtain that release's curated rows. The release runbook names this exact trap: *"leaves a migrate-only database missing curated rows."*

There is no signal in the migration history that anything is missing — the marker file applies cleanly.

## Root cause

Documented in `.claude/commands/pg-migrate-experimental.md`: metadata-sync migrations **must** be converted with the LEGACY converter — `mj migrate convert` **without** `--split`.

The `--split` v2 AST dialect cannot transpile the mj-sync grammar:

```
DECLARE @x_<hash>; SET @x = ...; EXEC __mj.spCreate/Update... @x
```

Every `DECLARE` block lands in `unhandled`, and the converter emits a marker file instead of failing. v5.45 was evidently converted with `--split`.

## Asks

[...]

Issue #3252, opened 2026-07-22
mj migrate convert silently drops statements and emits empty PG migrations while reporting success


SDesai-BC:
## Summary

`mj migrate convert` can drop source statements entirely and emit an effectively empty PostgreSQL migration **while reporting success**. No `.needs-hand` file is produced, the converter summary prints `with unhandled stmts: 0`, and the exit code is zero. The result is a `.pg.sql` file that applies cleanly to a fresh PostgreSQL database — because it does almost nothing — and therefore passes every gate in the release process.

Discovered during the **MemberJunction v5.49.0 release build** on 2026-07-22, while running DEPLOYMENT.md Step 8 (PostgreSQL migration conversion, via the `/pg-migrate-experimental` runbook) inside the `claude-dev` Docker workbench. All evidence below was gathered first-hand and verified.

## What happened

Six of the seven in-scope v5.49 migrations were converted with:

```
npx mj migrate convert --split --bake-codegen --file <name> --verbose
```

against a live PG working database seeded to the prior state. **All six reported success.** The actual output:

| Migration | SS lines | PG lines | Reality |
|---|---|---|---|
| `V202607191254__v5.49.x__Backfill_Missing_FK_Auto_Indexes` | 111 | 23 | **BROKEN** — 7 `CREATE INDEX` in source; PG output contains **six bare `;` statements** and zero index DDL |
| `V202607201104__v5.49.x__Agent_Conversation_Compaction` | 8963 | — | **FAILED** — `Error: Conversion failed ... syntax error at or near "CONSTRAINT"`; no file produced |
| `V202607202000__v5.49.x__SS_Materialize_Catalog_Views_spUpdateExistingEntityFields` | 186 | 13 | empty stub — **but legitimate, see "Ruled out" below** |
| `V202607202100__v5.49.x__SoftPK_Guard_Materialized_spUpdateExistingEntityFieldsFromSchema` | 206 | 13 | empty stub — **but legitimate, see "Ruled out" below** |
| `V202607202110__v5.49.x__Fix_ConversationDetail_Sequence_Deadlock` | 97 | 13 | **GAP** — source contains `CREATE OR ALTER TRIGGER trgConversationDetail_AssignSequence`; output says "no DDL to translate" |
| `V202607202322__v5.49.x__Add_Theme_Entity` | 1789 | 1026 | substantive, unvalidated |

### Verbatim evidence

[...]

PR #3258, opened 2026-07-22
docs(deployment): gate PG migration content + fix the runbook gaps v5.49.0 exposed

SDesai-BC:
**A PostgreSQL migration can be silently emptied by the converter and pass every check we have, including the one the runbook calls "the real gate."** It has already shipped: `migrations-pg/v5/V202607071019__v5.45.x__Metadata_Sync.pg.sql` is **126 bytes** against a **12,041-line** T-SQL source. PostgreSQL deployments that migrated through v5.45 silently received none of that release's curated metadata, and nobody noticed for two months.

The same thing happened three more times during the v5.49.0 build. This PR adds the gate that catches it, plus the runbook corrections for everything else that cost time in that release.

## The gate

`mj migrate convert` can write a header-only stub while printing `unhandled stmts: 0` and exiting `0`. Real output from this release, in full:

```sql
-- ============================================================================
-- MemberJunction PostgreSQL Migration — ...Fix_ConversationDetail_Sequence_Deadlock.sql
-- ============================================================================

-- V202607202110__..._Fix_ConversationDetail_Sequence_Deadlock.sql — no DDL to translate.
```

The source contains a `CREATE OR ALTER TRIGGER`. Another file emitted **six bare `;`** where six `CREATE INDEX` statements belonged.

Every automated check passed, and not by accident — **an empty migration applies perfectly cleanly, so the fresh-database apply cannot fail on it.** `check-pg-migration-parity.mjs` only asserts the counterpart *exists*. It was caught by hand-diffing line counts.

`scripts/check-pg-migration-content.mjs` closes that, wired into `pg-migrations.yml` as Step 3b — **before** the apply (content check at line 121, apply at 152), so a failure is attributable to conversion rather than to SQL.

### Two design choices worth defending

**It counts statements, not lines.** The bare-`;` failure was 23 lines of nothing. A line-count heuristic scores that 23 and waves it through; a statement count scores it 0. Boilerplate (`CREATE EXTENSION IF NOT EXISTS`, `CREATE SCHEMA IF NOT EXISTS`, `SET search_path`, `SET standard_conforming_strings`) is excluded, since every converted file carries it and it says nothing about whether content survived.

**The rule is "empty AND undeclared", not "empty".** Some counterparts are legitimately empty — the SS migration alters `spUpdateExistingEntityFieldsFromSchema`, which PostgreSQL maintains in `packages/CodeGenLib/src/Database/providers/postgresql/metadataSupportObjects.ts` rather than in a migration. A blunt rule would produce false positives and get switched off inside two releases. An intentional no-op must instead declare itself:

```sql
-- PG-EMPTY-BY-DESIGN: <why, and what carries the change instead>
```

The judgement stays with the human; the machine only enforces that it was recorded. Illustrating the point, `Fix_ConversationDetail_Sequence_Deadlock.pg.sql` already carries a *prose* explanation written during the release, and the detector still flags it — prose is not machine-checkable.

### Grandfathering, and why it is not an amnesty

Five existing counterparts are effectively empty and **cannot** be retrofitted with the token: committed `migrations-pg/v5/*.pg.sql` are byte-for-byte immutable because Flyway checksums them, so editing one breaks `mj migrate` on any deployment that already applied it. That includes the three that shipped in v5.49.0 hours before this branch was cut.

They sit in a `GRANDFATHERED` map with a written reason each, following the `ai-providers.AIP1` precedent ("known-dead allowlist, fails on any NEW unresolvable driver"). **The v5.45 entry says plainly that it is not correct** and points at #3253 rather than pretending it is fine. The map's own comment reads: *"Do not add entries to silence a new failure — that is the exact reflex this script exists to prevent."*

One entry changed the story. #3252 originally listed `v5.38 Fix_AllowUpdateAPI_On_Virtual_Transition` as an unverified second suspected escape. Building this forced the investigation: it alters the same proc PostgreSQL keeps in TypeScript, so it is **correctly** empty. **Confirmed shipped escapes drop from two to one.**

## Evidence

Detector self-test — fixtures are real production shapes, not invented:

| | |
|---|---|
| `--self-test` (14 fixtures after the review follow-up below) | **14/14, exit 0** |
| Repo as-is | **exit 0** — 193 counterparts checked, 5 grandfathered, 0 suspect |
| Sabotage `Add_Theme_Entity.pg.sql` (1026 lines → 11-line stub) | **exit 1** — `source has 84 content statement(s); PG counterpart has 0.` |
| Same file + `PG-EMPTY-BY-DESIGN` line | **exit 0**, counted as documented no-op |
| Restore | `git status migrations-pg/` clean |

CI-shape and edge cases:

| | |
|---|---|
| Wrong working directory | exit **2** (errors, does not falsely pass) |
| Unknown argument | exit 2 |
| `--help` | exit 0 without checking |
| SS migration with no counterpart at all | exit 0 (that is parity's job) |
| No migrations at all (tooling-only PR) | exit 0 |

Meta-gate — if someone breaks the detector, does CI notice?

| | |
|---|---|
| Neuter `classify()` to always return `ok` | self-test **exit 1** ✓ (repo check still exits 0 — which is exactly why the self-test step exists) |
| Break the `PG-EMPTY-BY-DESIGN` regex only | self-test **exit 1** ✓, naming the fixture |

The CLI dispatch is guarded by an `import.meta.url` check so the module can be imported. Without it, `classify` and `countContentStatements` were exported but unreachable — importing ran the whole repo check and called `process.exit`, which would have silently killed any unit-test process before its first assertion.

## Runbook corrections

Each one is a thing that cost time in v5.49.0, not a hypothetical.

- **Step 0 (new)** — preflight. The gitignored `.env.database`; live credential probes (an expired Gemini key produced 11 red agent tests triaged as a `BaseAgent` regression before the real cause surfaced); `MJ_API_KEY` placement; Docker headroom.
- **Agent checklist** — a release spans hours and several CI waits, so steps get skipped. **Step 5 was nearly missed** because it sits between two long-running steps. Names the sub-steps that are separately skippable, and flags that Steps 5 and 11 have no automated backstop.
- **Per-engineer local values** — every port, container name and database name in the guide is one machine's example. Adds the two commands to resolve your own, and lists the known-variable ones (including that `.env` says `GRAPHQL_PORT=4000` while root `CLAUDE.md` still claims 4001).
- **Step 3 runs in parallel with Step 1** — Step 1's CI takes ~15 minutes and Step 3 reads nothing from it.
- **Step 4.3** — `npm run start:api` goes through turbo, which **strips env vars it does not declare**; `DB_DATABASE` vanished and surfaced as `Error parsing config file … "path": ["dbDatabase"]`, which reads as a config-file problem. Run from `packages/MJAPI`.
- **Step 4.6 (#3251)** — the "re-run the single bundle" triage advice is **wrong** for IT53–IT62. `mj test run` takes a different transport path than the same bundle in-suite (`agent-loop-live` is not in the `CLIENT_BUNDLES` set), so standalone it runs the agent in-process with no server `contextUser` and fails 7/7 while passing in-suite. Re-run the whole live suite instead.
- **Step 7** — was "commit the regenerated manifests"; CI commits them post-release (`9cc0a42fa5`). Now a decision table keyed on whether the registration *count* changed.
- **Step 8** — the content-verification section, and check-git-history-before-hand-authoring. Two counterparts this release already existed, authored and reviewed in feature PRs then deleted by policy; recovered from `f0ceec06fb^`, including a `BEFORE INSERT FOR EACH ROW` trigger using `pg_advisory_xact_lock`.
- **Step 9** — which CI absences are expected (`build.yml` is path-filtered to `packages/**`; `dependency-check.yml` is PR-into-`next` only), so absence stops reading as failure.
- **Post-merge** — GitHub's "Cancel workflow" has **no confirmation dialog**, and a cancel mid-`npm publish` splits the fixed-version group across two versions on npm with no rollback. This happened in v5.49.0. Documents the `gh run rerun --failed` recovery and explicit post-recovery verification.
- **10c** — what `docs.yml` publishes, and that it failed silently from v5.45.1 through v5.48.0 while each release looked green.
- **Doc-site download step — retired.** The old Step 11 edited a per-version download link that no longer exists (the `Distributions/` zip is gone; `mj install` fetches from the tagged source). Removed as a numbered step, replaced with a deprecation note; the release now runs 0–11.
- **Step 11 (Changelog, was Step 12)** — said only *when*, never *where*. The release PR body is the changelog entry; adds the fetch command and notes that **no GitHub Release exists** (verified — no workflow creates one).

## Workbench

`docker/workbench/.env.database` is gitignored, so `docker compose up` fails on a clean checkout with `env file … not found`, and `README.md` described the file as "committed". Ships a tracked `.env.database.example` and corrects the README.

The `.gitignore` negation is deliberate: `docker/workbench/.env.example` is tracked only by **grandfathering** (committed before the `.env.*` rule existed, and git ignores untracked files only). A new `.example` file gets no such grace, so it needs an explicit negation rather than a silent `git add -f` that leaves no record of the decision.

## Risk

Low, and mostly self-limiting. The only behavioural change outside documentation is a new CI step on `pg-migrations.yml`, which is path-filtered to migration and converter changes. It passes on `next` today (verified above), and **this PR exercises it against itself** since it touches `.github/workflows/pg-migrations.yml` — so if the wiring is wrong, this PR is where it surfaces.

The grandfathered list is load-bearing: the gate is green today because of those five entries. That is the intended ratchet, but it means a future engineer tempted to silence a failure by appending to the map is doing the precise thing the check exists to prevent.

## Related

Nothing is closed here. #3252 (converter still writes empty files while reporting success) and #3254 (parity script checks existence only) are both **compensated, not fixed** — the tool defect is untouched. #3253 (v5.45's shipped marker) is recorded in the grandfathered map. #3251 is the Step 4.6 correction. #3257 (integration suite cannot run on PostgreSQL) is unrelated to this diff.

## Review response (`5711961a36`)

Follow-ups from review, TDD'd (7 red → green):

- **The `SOURCE_STATEMENT_FLOOR` coverage gap is closed — by narrowing where the floor applies, not lowering it.** An empirical sweep of all 193 committed pairs showed every real escape scored **pg=0** and every legitimately-thin counterpart scored **pg=1** (e.g. the v5.39 pair where an `IF EXISTS` guard + two `sp_*extendedproperty` calls genuinely fuse into one `COMMENT ON TABLE`). `classify()` is now two-band: **zero content statements is suspect at any source size** (fusion never reaches zero); the floor governs only the pg=1 band. This catches the reviewed 5-statement case *and* the 1-statement case that lowering the floor to 1 would still miss — with zero new suspects on the committed corpus.
- **Grandfather staleness guard**: `staleGrandfatherWarnings()` warns (never fails; `::warning` in CI) when an entry no longer matches a checked pair or no longer classifies as suspect. Warn is correct because a stale entry can't mask a *new* escape — an unmatched suspect still fails loudly.
- **The promised unit suite now exists**: `.github/scripts/__tests__/check-pg-migration-content.test.mjs`, 23 tests in the existing guard-suite convention, locking the exported API, both threshold constants, and the CLI contract. Full guard config: 73/73.
- **Parser heuristic edges documented** in `countContentStatements` (string-literal comment mis-stripping miscounts but never zeroes; `;`-termination assumed, `GO`-batch T-SQL under-counts).
- **CI wiring gap found and fixed**: neither workflow path-filtered on the detector itself, so a detector-only PR would have run *neither* suite. Both now do.
- `#3252` / `#3253` remain open — this still compensates for the converter defect, it does not fix it.

Smoke-tested end-to-end with real fixture files through `runCheck`: emptied 5-stmt and 1-stmt counterparts fail the gate naming the file, a `PG-EMPTY-BY-DESIGN` declaration passes as documented, a genuine conversion passes; sabotaging the rule back to the old short-circuit is caught by 4 unit tests + the self-test, and a bogus grandfather entry triggers the staleness warning.

### DIFF
--- a/.changeset/pg-migration-content-gate.md
+++ b/.changeset/pg-migration-content-gate.md
@@ -1,0 +1,43 @@
+---
+"@memberjunction/sql-converter": patch
+---
+
+Add a content gate for converted PostgreSQL migrations. `mj migrate convert` can emit an
+empty `.pg.sql` while printing `unhandled stmts: 0` and exiting `0`, and nothing caught it —
+**an empty migration applies perfectly cleanly, so the fresh-database apply that the release
+process treats as authoritative is structurally incapable of failing on it.** The parity check
+only asserts the counterpart file exists.
+
+This already shipped once. `V202607071019__v5.45.x__Metadata_Sync.pg.sql` is 126 bytes against
+a 12,041-line T-SQL source, so PostgreSQL deployments migrating through v5.45 silently received
+none of that release's curated metadata (#3253). The same behavior recurred three times during
+the v5.49.0 build — header-only stubs, and one file with six bare `;` where six `CREATE INDEX`
+statements belonged. It was found by hand-diffing line counts against sources.
+
+`scripts/check-pg-migration-content.mjs` now runs in `pg-migrations.yml` before the apply step,
+so a failure points at conversion rather than at SQL. Two design points:
+
+- **It counts statements, not lines.** The bare-`;` output was 23 lines carrying zero statements;
+  a line-count heuristic scores that 23 and passes it. Header boilerplate is excluded, since every
+  converted file has it regardless of whether content survived.
+- **The rule is "empty AND undeclared".** Some counterparts are correctly empty — the SQL Server
+  migration may alter a routine PostgreSQL maintains in TypeScript (`metadataSupportObjects.ts`).
+  A blunt "empty fails" rule would produce false positives and get disabled, so an intentional
+  no-op declares itself with `-- PG-EMPTY-BY-DESIGN: <reason>`. The judgement stays with the
+  author; the check only enforces that it was recorded.
+
+Five pre-existing empty counterparts are grandfathered with written reasons, because committed
+`.pg.sql` files are Flyway-checksummed and immutable — editing one breaks `mj migrate` on any
+deployment that already applied it. It is a ratchet rather than an amnesty: the v5.45 entry
+records that it is **not** correct. Investigating another entry narrowed the problem — `v5.38
+Fix_AllowUpdateAPI_On_Virtual_Transition`, previously suspected as a second escape, is correctly
+empty, so confirmed shipped escapes drop from two to one.
+
+The detector self-tests against the real failure shapes, and that step is wired into CI too: a
+neutered comparison and a broken declaration-token regex were both verified to fail it. Without
+it a broken detector would pass everything silently.
+
+This compensates for the converter defect rather than fixing it (#3252, #3254) — `mj migrate
+convert` still writes empty files while reporting success. The `/pg-migrate-experimental` runbook
+and `DEPLOYMENT.md` Step 8 gain the same content check plus a recipe for recovering counterparts
+that feature PRs authored and later deleted by policy.
--- a/.claude/commands/pg-migrate-experimental.md
+++ b/.claude/commands/pg-migrate-experimental.md
@@ -465,7 +465,17 @@ Each of these cost real time. Heed them.
 - For pg_dump, exclude `__mj.flyway_schema_history`, strip `\restrict` and `\unrestrict`, and make `CREATE SCHEMA` idempotent. The data section goes between pre-data and post-data.
 - For the browser test, use the Owner user, build the servers once and early and in parallel, and never restart to fix permissions. Kill a delegated agent that loops on a permission workaround.
 - Run the L3 and L4 scripts from the host, because they use `docker exec`. L1 needs no database. L2 needs the SQL Server compare database, so build it in parallel.
-- The real gate is a clean `mj migrate` on a fresh database. A "0 gaps" result from convert is structural only.
+- The real gate is a clean `mj migrate` on a fresh database. A "0 gaps" result from convert is structural only. **But that gate is blind to an EMPTIED migration — empty SQL applies cleanly.** The converter can write a header-only stub (or bare `;` statements) while printing `unhandled stmts: 0` and exiting 0. This shipped in v5.45 (`Metadata_Sync.pg.sql` = 126 bytes against 12,041 lines) and recurred 3× in v5.49. Run the content check before copy-back — it is also a CI gate in `pg-migrations.yml`:
+  ```bash
+  node scripts/check-pg-migration-content.mjs        # exits 1 on a silently-emptied counterpart
+  ```
+  A counterpart that is *correctly* empty (PG maintains the routine in `metadataSupportObjects.ts`, or never had the defect) must declare itself so it is distinguishable from the bug: `-- PG-EMPTY-BY-DESIGN: <reason>`.
+- **Before hand-authoring a `.needs-hand`, check git history.** Feature PRs sometimes author a counterpart that is later deleted under the "build engineer creates PG migrations" policy — that work is reviewed and recoverable. In v5.49 this recovered a 7,002-line file including a hand-written `BEFORE INSERT FOR EACH ROW` trigger using `pg_advisory_xact_lock`, ~30 minutes before it would have been rewritten from scratch.
+  ```bash
+  git log --all --oneline --diff-filter=A -- '*<MigrationName>.pg.sql'
+  git show <deleting-commit>^:migrations-pg/v5/<file>.pg.sql > /tmp/recovered.pg.sql
+  git log --oneline <deleting-commit>..HEAD -- migrations/v5/<MigrationName>.sql   # blank = SS unchanged, safe to reuse
+  ```
 - Committed `.pg.sql` and `.pg-only.sql` files are immutable. Only ever produce PG counterparts for new SQL Server migrations. Never reconvert or hand-patch a committed one.
 - For `mj codegen`, always use `scripts/pg-codegen-await.mjs`. The bare CLI can fire-and-forget and exit 0 as a silent no-op.
 - The converter's own "EntityField Sequence deduplication" pass edits committed `.pg.sql` files — check `git status migrations-pg/` after every convert. They never ship (Phase-5 copy-back is selective); restoring them in-container needs a `git diff` confirmation + explicit approval first (`git checkout --`/`git restore` are Rule-#3 destructive commands, never run blind). The gate proves whether the dedup mattered (so far: it did not).
--- a/.github/scripts/__tests__/check-pg-migration-content.test.mjs
+++ b/.github/scripts/__tests__/check-pg-migration-content.test.mjs
@@ -1,0 +1,162 @@
+import { describe, it, expect } from 'vitest';
+import { join, dirname } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { execFile } from 'node:child_process';
+import { promisify } from 'node:util';
+import {
+    countContentStatements,
+    classify,
+    staleGrandfatherWarnings,
+    SOURCE_STATEMENT_FLOOR,
+    PG_EMPTY_CEILING,
+} from '../../../scripts/check-pg-migration-content.mjs';
+
+const execFileAsync = promisify(execFile);
+const SCRIPT = join(dirname(fileURLToPath(import.meta.url)), '..', '..', '..', 'scripts', 'check-pg-migration-content.mjs');
+
+// The converted-file header every fixture shares — all boilerplate, must count as 0.
+const HEADER = `-- ============================================================================
+-- MemberJunction PostgreSQL Migration — X.sql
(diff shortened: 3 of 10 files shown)


rkihm-BC (commented):
## Review — thorough, with the detector run locally to validate the claims

Strong, well‑reasoned PR. The core is a small, correct, self‑exercising CI gate that closes a genuine hole that has *already* shipped a broken release. I verified the substantive claims rather than taking them on faith — recommend merge with a couple of non‑blocking follow‑ups.

### What I verified (not just read)
- **Self‑test: 10/10, exit 0** — all five "must flag" and five "must not flag" fixtures classify correctly.
- **Repo check against real committed migrations: exit 0, 0 suspects across 186 counterparts** — confirms the low false‑positive rate isn't just asserted.
- **`.gitignore` negation works** — simulated the PR's exact `.gitignore`: `.env.database.example` → tracked, `.env.database` → ignored, `.env.example` → ignored‑by‑pattern (confirms the "tracked only by grandfathering" claim).
- **Step renumbering is internally consistent** in the final `DEPLOYMENT.md` (0–11).

### What's good
- **Real problem, real fix.** The 126‑byte `Metadata_Sync.pg.sql` vs. 12,041‑line source (#3253) is a true silent‑emptying escape, and the reasoning that *a clean `mj migrate` structurally cannot catch it* (empty SQL applies fine) is exactly right.
- **Statement‑counting, not line‑counting** — the bare‑`;` case (23 lines, 0 statements) is the proof this matters; boilerplate exclusion is correct.
- **"empty AND undeclared" with `PG-EMPTY-BY-DESIGN`** is mature judgment — a blunt "empty = fail" rule *would* get disabled within two releases. Keeping judgment with the human and only enforcing that it was recorded is the right call.
- **Meta‑gate**: self‑test runs in CI *before* the real check, so a neutered detector fails loudly instead of passing everything silently — the part most people skip.
- **`import.meta.url` guard** is correct and necessary.
- **Grandfather map is honest** — a ratchet with written reasons; the v5.45 entry says plainly it is *not* correct and points at #3253, and the map comment warns against appending to silence new failures.
- **Low blast radius**: one path‑filtered CI step + docs, and it exercises itself (the PR touches `pg-migrations.yml`). Changeset correctly scoped to `@memberjunction/sql-converter` patch.

### What's concerning
1. **Coverage floor leaves the exact failure class partly uncovered.** `SOURCE_STATEMENT_FLOOR = 5` means any source with ≤5 content statements is never checked. The v5.49 `Backfill_Missing_FK_Auto_Indexes` had 6 indexes → caught, but a **5‑index** backfill emptied to zero would slip through silently. Small index/FK backfills are exactly this shape. Worth lowering the floor (the `PG_EMPTY_CEILING=1` + grandfather/token already guard against false positives on legitimately thin migrations) or documenting the residual gap in the script header. **This is my one substantive ask before merge.**
2. **Compensates, doesn't fix.** Disclosed, but worth stating plainly: `mj migrate convert` (#3252) still writes empty files and exits 0 — the real fix is the converter failing loudly on empty output. And #3253's shipped PG data gap is **not** remediated here; PG installs through v5.45 still need `mj sync push`. Keep #3252/#3253 open.
3. **No staleness detection on the grandfather map.** If a grandfathered stem later gains real content, or is renamed/removed, the entry becomes dead weight with nothing flagging it. The map is load‑bearing (the gate is green partly because of it), so a cheap guard — warn if a `GRANDFATHERED` stem no longer resolves to a suspect — would be worth it.
4. **Parser heuristic edges** (all low‑risk, worth a one‑line comment): `--` and `/* */` inside string literals get mis‑stripped, and the SS side is coarser than the PG side — a T‑SQL file using `GO` batches without `;` terminators would under‑count and, if under 5, skip the check. Not how MJ migrations are written today, but it's an undocumented assumption in a safety‑critical script.

### What's missing
- **A committed unit test.** The `import.meta.url` guard was added specifically so `classify`/`countContentStatements` are unit‑testable, but no `.test.ts` ships with the PR. MJ's Vitest convention (one test file per source) would lock the exported API and the threshold constants. `--self-test` covers the logic behaviorally, so this is a gap, not a hole — but the guard's stated rationale ("exported for unit tests") is currently unfulfilled.
- **Cosmetic drift**: the PR *body* still says "Steps 5, 11 **and 12**" while the final renumbered doc correctly says "Steps 5 and 11." The doc is right; the description is stale from an earlier commit.

### Bottom line
The engineering judgment here is notably good — it fails safe, self‑tests, refuses to become false‑positive noise, and is honest about what it does and doesn't fix. The only thing I'd want addressed before merge is the **`SOURCE_STATEMENT_FLOOR = 5`** coverage question, since that's the one place the gate doesn't quite reach the failure class it was built for. Everything else is follow‑up material.

PR #3259, opened 2026-07-22
perf(MJServer): fix O(N²) GraphQL schema-build scaling at boot (−66% cold boot) (6.x)

MS-BC:
## Summary

`buildSchemaSync` — called once at MJAPI boot in `packages/MJServer/src/index.ts` — is **super-linear in entity count**, and is the dominant cost of MJAPI cold-boot at scale (it was the headline finding of the scale-runtime study). This PR root-causes that to **two O(N²) hotspots inside `type-graphql`** and fixes them with pure memoization/indexing (applied via `patch-package`), plus flag-gated boot-profiling instrumentation.

## Measured effect (1,380-entity environment)

| | `buildSchemaSync` | schema phase | total cold boot |
|---|---:|---:|---:|
| stock type-graphql | 66,261 ms | 23.3 s | 95.6 s |
| **this PR** | **1,230 ms** | **2.9 s** | **32.1 s** |

→ **54× faster `buildSchemaSync` (−98.1%)**, **total cold boot −66%**. The win grows with entity count (the removed cost is super-linear). It is a **one-time boot cost** — per-request serving latency is unaffected (schema is built once).

## What changed

**1. `patch-package` patch of `type-graphql@2.0.0-beta.3`** (`packages/MJServer/patches/type-graphql+2.0.0-beta.3.patch`, both `esm` + `cjs`):
- `schema/schema-generator.js` — the per-field config thunk recomputed a loop-invariant (`fieldResolvers.filter(...)` + a `.find()` scan) once per field × per object type → memoized into a `Set` + `Map<target, Map<method, meta>>` (O(1) lookup).
- `metadata/metadata-storage.js` — `buildClassMetadata` / `buildResolversMetadata` / `buildFieldResolverMetadata` / `findFieldRoles` each did per-field/per-def `.filter()` scans over the global `fields`/`params`/`middlewares`/`fieldDirectives`/`authorizedFields` arrays → a one-time `_mjBuildIndexes()` (called in `build()`) pre-groups each into `target → (name → list)` maps.

**2. `patch-package` wiring** — `postinstall` in `packages/MJServer/package.json` + a root-level `postinstall` delegation (so it runs on a top-level `npm install`), and `patch-package` added to devDependencies. See `packages/MJServer/patches/README.md`.

**3. Flag-gated boot-profiling instrumentation** in `src/index.ts` — `MJ_SCHEMA_PROFILE=1` (per-sub-step ms + metadata cardinality) and `MJ_SCHEMA_CPUPROF=1` (a `.cpuprofile` of just `buildSchemaSync`). Both **no-op unless set**; this is the tool that produced the numbers above and re-verifies future boot work.

### CI — green
All five required checks pass on the current head: `Check migrations`, `Detect missing and unused dependencies`, `Integration (SQL Server, deterministic)`, `Run unit tests`, `packages/** must not peer-depend on @memberjunction/*`. This also closes the two questions that were previously open: a clean-install compile of `@memberjunction/server`, and that `patch-package` applies during install in the nested-workspace layout (`type-graphql` resolves only to `packages/MJServer/node_modules`, not hoisted — install log shows `type-graphql@2.0.0-beta.3 ✔`, and all four target files carry the patch markers while the un-patched main repo carries none).

### Live A/B against a real database and a real MJAPI boot

Method: one build of this branch, booted twice against the **same** SQL Server database — first with the patch applied, then with `npx patch-package --reverse` (0 patch markers in all four files, verified) and nothing else changed. So the only variable is the patch itself. Test database is an isolated `COPY_ONLY` backup/restore clone at v5.50, **380 entities**; authenticated via a throwaway system API key so real resolvers ran, not just introspection.

| Check | Stock (patch reversed) | Patched | Result |
|---|---|---|---|
| Metadata cardinality | resolvers=2279 objectTypes=1019 fields=19894 typeMap=1899 | identical | no metadata dropped or reordered out of existence |
| `buildSchemaSync` | 3628 ms | **241 ms** | **15.1×** at 380 entities |
| Schema phase / total startup | 4.1 s / 13.0 s | 0.7 s / 8.8 s | — |
| Emitted SDL (`schema.graphql`) | sha256 `a47e9301…` | sha256 `a47e9301…` | **byte-for-byte identical** — 39,382 lines, 1,289,162 bytes |
| 7 authenticated GraphQL queries | — | — | **all 7 responses byte-identical** |
| MJServer unit suite | — | 720 passed / 56 skipped / 0 failed | green |

The query battery deliberately covers what SDL cannot prove — SDL captures types, fields, args and nullability, but **not which resolver function is bound to which field**, which is exactly what the patched `buildFieldResolverMetadata` decides. So: `q3`/`q6` resolve `MJEntityFields_EntityIDArray`, a real `@FieldResolver` — 5,517 related rows across 380 entities, 912 KB of response, identical on both sides. Also identical: `q1` `GetAllEntities` with nested `Fields` (646 KB), `q2` a `RunView` dynamic-view resolver with input args, `q4` a two-level nested field resolver, `q5` a list resolver, `q7` (deterministic introspection-disabled error, same both sides).

### Order- and aliasing-equivalence, checked in the source rather than assumed
- **`findFieldRoles` changed from `.find()` to `list[0]`.** `_mjGroup2` builds each list by pushing in source-array iteration order, so `list[0]` is the same element `.find()` returned. Equivalent by construction — auth roles cannot shift.
- **`field.params` / `def.params` now receive the shared grouped array** instead of a fresh `.filter()` result. Every consumer in type-graphql reads it via `.reduce()`; the only mutation anywhere is `this.params.push()` in the pre-build collector, which runs before `_mjBuildIndexes()`. No aliasing hazard.
- Same reasoning for `middlewares` / `fieldDirectives`, both of which are additionally re-materialized through `.map()` before assignment.

### Stated limits
- The live A/B ran at **380 entities**, not the 1,380 behind the headline 54× figure, so the *measured* speedup there is 15.1× — the ratio grows with N because the removed work is super-linear. This bounds the performance claim at that scale; it does not affect the equivalence result.
- Queries ran as the system user, so role-gated **denial** was not exercised against a restricted user. That path is covered by the `findFieldRoles` order-equivalence argument above rather than by runtime.

### Failure mode
Inert. If the patch does not apply, MJAPI runs stock type-graphql — a slower boot, never incorrect behavior.

## Upstream
This is a generic type-graphql scaling bug, not MJ-specific. A companion PR to `MichalLytek/type-graphql` ports both fixes to their TypeScript source; once merged + released, this patch can be dropped for a version bump.


AN-BC (commented):
@MS-BC I need to ask @cadam11  to review and approve this as this is outside of my knowledge area. Also, I think this is probably a `6.0.0` as it is an optimization not a bug fix, unless @cadam11 and you see it differently.

Always great to see performance improvement opportunities!

MS-BC:
> @MS-BC I need to ask @cadam11 to review and approve this as this is outside of my knowledge area. Also, I think this is probably a `6.0.0` as it is an optimization not a bug fix, unless @cadam11 and you see it differently.
> 
> Always great to see performance improvement opportunities!

Yes, this is intended for 6.x, sorry I should've made that clear

MS-BC:
Re-verified at the current head (`db27bd4`, after the merge from `next`) — the merge does not touch the patch or the profiling hooks:

- **All five required checks pass** at this head, including `Integration (SQL Server, deterministic)` (which boots a real server) and `Run unit tests`.
- **The patch still applies cleanly to a clean install** of `type-graphql@2.0.0-beta.3`: `patch-package 8.0.1 → type-graphql@2.0.0-beta.3 ✔`, with the index builder present in **both** `build/cjs/metadata/metadata-storage.js` and `build/esm/metadata/metadata-storage.js`. This was the one thing a merge could silently break, since a failed patch is inert (stock, slower boot) rather than loud.
- `MJ_SCHEMA_PROFILE` / `MJ_SCHEMA_CPUPROF` hooks survive the merge in `packages/MJServer/src/index.ts`.

One equivalence detail the description argued only informally, now checked against the type definitions rather than assumed — `_mjGroup2` keys each entry on `it.fieldName !== undefined ? it.fieldName : it.methodName`, so it must pick the *same* key the original `.filter()` compared against:

| index | original predicate | key `_mjGroup2` picks | correct? |
|---|---|---|---|
| `params` | `param.methodName === field.name` / `=== def.methodName` | `methodName` — `ParamMetadata` is `BasicParamMetadata`-derived and has **no** `fieldName` on any of its seven variants | ✓ |
| `middlewares` | `middleware.fieldName === …` | `fieldName` | ✓ |
| `fieldDirectives` | `it.fieldName === …` | `fieldName` | ✓ |
| `authorizedFields` | `authField.fieldName === …` | `fieldName` | ✓ |

The only way this could mis-key is a `params` entry carrying a `fieldName`, which the `ParamMetadata` union makes impossible. And an entry with `fieldName` explicitly `undefined` degrades to an `undefined` key that no real field name can look up — the same "never matches" the original filter produced.

The full live A/B (real database, real MJAPI boot, byte-identical SDL, 7 authenticated queries) in the description stands; nothing in the merge changes the code it exercised.

AN-BC:
LGTM - @cadam11 when you can bless this and @MS-BC you have resolved merge conflicts and any Craig input, please merge after that, I don't need to review again.

MS-BC:
Base synced to current `next`. Conflict was in root `package.json` scripts — kept this branch's `"postinstall": "npm run postinstall --workspace=@memberjunction/server --if-present"` alongside next's added `prebuild` line; `package-lock.json` regenerated against the merged tree.

Manual test note: MJServer's suite could not complete in my local worktree (`Failed to resolve entry for package "@memberjunction/auth-providers"` — unbuilt workspace `dist/`, an environment gap, not a code failure). It passes on a fully-built tree: **773 passed / 56 skipped**. Deferring the authoritative run to CI.

AN-BC:
ok @MS-BC merge away

cadam11:
We're moving to `pnpm` so `patch-package` isn't the best approach for this

cadam11 (changes requested):
The find/fix here is great, but the mechanism should use our target state, pnpm

MS-BC:
## Rebased onto `next`, and the pnpm mechanism verified end-to-end

@cadam11 — the mechanism change you asked for was already in (`f54da2aba4`), but the branch
had drifted **462 commits behind `next`**, so nothing here had been proven against current
code. That is now fixed, and the patch has been verified working through pnpm's native
mechanism rather than argued for.

### Base updated

`origin/next` merged in. Two conflicts, both in the expected place:

- **`package.json`** — `next` added five `@codemirror/*` overrides plus `jsdom` alongside
  this branch's `patchedDependencies` block. Resolved by keeping **both**.
- **`pnpm-lock.yaml`** — taken from `next` wholesale and regenerated by `pnpm install`, so
  the patch entry is re-derived from the resolved `package.json` rather than hand-merged.

Branch is now **0 behind / 17 ahead** of `next`.

### The pnpm mechanism genuinely applies the patch

Not inferred from config — observed in the installed tree:

- pnpm's store path itself encodes the patch:
  `type-graphql@2.0.0-beta.3_patch_hash=4d3544a1e64ad80b22832bb91021718c73db40ae6c6ac140b8…`
- **both** the `esm` and `cjs` builds carry the markers — 3 in `schema/schema-generator.js`,
  2 `_mjBuildIndexes` in `metadata/metadata-storage.js`
- `packages/MJServer/node_modules/type-graphql` resolves to that patched copy
- the lockfile records the patch hash + path
- zero `patch-package` remnants anywhere in the tree

### A/B on one build against one database — only the patch varying

Method: boot MJAPI, then remove `pnpm.patchedDependencies`, `pnpm install` (verified
MJServer then resolves an **unpatched** copy — 0 markers), boot again. Same dist, same
database, nothing else touched.

| | patched | stock | |
|---|---|---:|---|
| `buildSchemaSync` | **248 ms** | 3,692 ms | **14.9×** |
| schema phase | 0.7 s | 4.1 s | |
| **total cold boot** | **5.2 s** | 12.8 s | **−59%** |
| metadata cardinality | `resolvers=2281 objectTypes=1019 fields=19959 typeMap=1899` | identical | nothing dropped |
| emitted SDL sha256 | `a544a498f5dc…` | `a544a498f5dc…` | **byte-for-byte identical** |
| `GetAllEntities` | 377 entities | 377 entities | same data |

MJAPI **boots and serves** — that query runs through a real resolver and returns all 377
entities.

### It reproduces the PR's own numbers

This PR reports `buildSchemaSync` at **241 ms** for 380 entities on SQL Server. Measured
here: **248 ms at 377 entities on PostgreSQL** — a different backend, same result.

The stock side corroborates independently too. An unrelated MJAPI boot earlier the same
evening, on a different branch and different worktree at 377 entities, logged
`schema 4.1s` — the identical stock schema phase measured here.

### Environment

PostgreSQL 17, 377 entities, `MJ_SCHEMA_PROFILE=1`. Measured at this scale the win is
**14.9×**, not the 54× headline — the removed work is super-linear, so the ratio grows with
entity count. This bounds the performance claim at this scale; it does not affect the
equivalence result, which is exact.

cadam11 (approved):

PR #3260, opened 2026-07-23
fix(testing): thread contextUser through live-agent invokers + fail-loud transport guard

### LINKED ISSUE #3251: test(integration): single-bundle "mj test run" gives false failures for live-agent bundles (bypasses client transport, runs agent in-process)
## Summary

Running a single integration bundle with `mj test run "IT## - <name>"` is **not** equivalent to running that same bundle inside `mj test suite`. For the live-agent bundles the difference is decisive: in-suite the agent executes **server-side over the GraphQL wire** (as the checks are written to do), standalone it silently degrades to **in-process execution inside the test-runner process**, where there is no server `contextUser`, `ActionEngineServer.Config` fails, and nothing the oracles look for gets persisted. The bundle then reports a full sweep of failures that look exactly like a product regression in `BaseAgent` / the agent-persistence layer, but are entirely an artifact of the harness.

The mechanism is a hardcoded allow-list. `IntegrationTestDriver.bundleTransport()` decides a bundle's transport from a fixed set:

```ts
// packages/TestingFramework/testing-integration/src/IntegrationTestDriver.ts:55
const CLIENT_BUNDLES = new Set(['client-cache', 'rls-isolation-client', 'remote-op-wire-progress']);
```

`agent-loop-live` is not in that set, so it takes the `'server'` branch of `buildCheckContext()` and gets `ctx.Provider = activeBootstrap?.Provider ?? Metadata.Provider`. The checks then hand that provider straight to an in-process runner:

```ts
// packages/TestingFramework/integration-test-suite/src/checks/agent-live-shared.ts:48
export function makeAIClient(provider: IMetadataProvider): AgentInvoker {
    return { RunAIAgent: (params) => new AgentRunner(provider).RunAgent({ ... provider }) };
}
```

So the transport is decided **entirely by which provider happens to be bound**, and nothing in the bundle asserts it got the right one. Inside a suite run, one of the three `CLIENT_BUNDLES` members has already executed and rebound the process's global provider to the GraphQL client provider — so `Metadata.Provider` resolves to a client provider and the agent bundles go over the wire *by accident of suite ordering*. Standalone, no client bundle ran, the global provider is still the SQL provider, and the same code path runs the agent locally.

DEPLOYMENT.md §4.6 already carries a related warning — *"Do not reorder suite membership (client-transport members are deliberately sequenced last — the client bootstrap rebinds the process's global provider)"* — which is the same mechanism. The gap is that nothing states the corollary: **single-bundle runs are therefore not equivalent, and the live-agent bundles' correct transport is load-bearing on ordering rather than declared.**

## Impact

DEPLOYMENT.md §4.6 prescribes exactly the broken command as the standard triage step for any red live test:

> **Anything else red** — a real product defect. Re-run the single bundle before re-running the tier:
> ```bash
> MJ_INTEGRATION_TEST=1 npx mj test run "IT## - <name>"
> ```

For live-agent bundles that instruction manufactures a clean, reproducible-looking failure. During the **v5.49.0 release build (2026-07-22)** it cost real time: the single-bundle re-run was performed twice, reproduced identically both times, and led to a phantom diagnosis of a product defect in `BaseAgent` / the agent persistence layer. A regression issue against `base-agent.ts` was nearly filed. The next person following §4.6 will be misled the same way.

Secondary risk: because the live-agent bundles get their correct transport only as a side effect of suite ordering, a future reordering (or dropping one of the three `CLIENT_BUNDLES` members from the suite) would silently flip every agent bundle to in-process and turn the live tier red for no product reason.

## Evidence

Environment: fresh scratch DB `MJ_Release_v549` (SQL Server, Docker, `localhost:1455`), MJAPI on `:4000` from a current build, integration metadata seeded (242 records via `mj sync push --dir ./metadata-optional/integration-test --ci`).

**1. MJAPI hit count — 22 vs 0.** Same bundle, two invocations, MJAPI log grepped for `RunAIAgent` / `agentRunId` / `prompt_execution`:

| Invocation | MJAPI agent-execution hits |
|---|---|
| `IT53` inside `mj test suite "Integration Tests — Live Model"` | **22** |
| `IT53` via `mj test run "IT53 - Agent Loop Foundation (live)"` | **0** |

Zero server-side hits standalone: the agent never reached MJAPI.

**2. Stack trace from the standalone run** — the in-process path, terminating in the test runner's own check file:

```
ActionEngineBase.Load          (packages/MJCore/dist/generic/baseEngine.js:310)
  ← ActionEngineBase.Config    (packages/Actions/Base/dist/ActionEngine-Base.js:108)
  ← ActionEngineServer.Config  (packages/Actions/Engine/dist/generic/ActionEngine.js:51)
  ← BaseAgent.initializeEngines(packages/AI/Agents/dist/base-agent.js:2445)
  ← BaseAgent.Execute          (packages/AI/Agents/dist/base-agent.js:1184)
  ← AgentRunner.RunAgent       (packages/AI/Agents/dist/AgentRunner.js:84)
  ← agent-loop-live.checks.js:94
```

**3. `contextUser` CRITICALs, logged only by the test-runner process:**

```
[CRITICAL] [AgentExecution] For server-side use of all engine classes, you must provide the contextUser parameter
```

**21** occurrences per standalone bundle run; **216** across a full 15-bundle live-tier run. MJAPI logged **zero** of these. That asymmetry is itself the proof of where execution happened.

**4. `dist/` freshness ruled out as a cause.** `agent-loop-live.checks.js` (09:10:14) is newer than `agent-loop-live.checks.ts` (08:24:57), and `dist` contains `runAgentOverWire` 8 times. The compiled bundle under test was current.

**5. Run-level outcome.** Live tier run 1: 4/15. After an unrelated local credential fix + MJAPI restart, run 4: 8/15 — IT17, IT19, IT58, IT59 recovered; IT53–IT57, IT60, IT61 still failed. IT53 standalone: all 7 oracles (AL1–AL7) failed, twice, identically.

For reference the deterministic tier was clean throughout: `RUN_MUTATION_TESTS=1 npm run test:integration` → **52/52 passed**, 0 environment gaps, DB confirmed `Passed=52` with no Error/Timeout rows. This issue is scoped to the live tier only.

## Secondary: the `fire-and-forget write never landed` message is factually wrong

```ts
// packages/TestingFramework/testing-integration/src/ai-verify.ts:40
Assert(false, `${entity} ${id} not found after bounded poll (fire-and-forget write never landed)`);
```

Reported as e.g. `MJ: AI Agent Runs <guid> not found after bounded poll (fire-and-forget write never landed)`, alongside `AL4: child prompt runs recorded tokens (Σ=0 > 0)`.

**The writes did land.** Querying the database directly over the same window (2026-07-22 14:17:27 → 15:50:11 UTC) returned **42 rows** in `__mj.AIAgentRun` and **186 rows** in `__mj.AIPromptRun`. Nothing was lost — the rows simply commit after the poll window closes.

The message asserts data loss as fact, which sends the reader hunting for a persistence bug that does not exist. It should state the bound it actually waited and name the knob, e.g.:

> `MJ: AI Agent Runs <guid> not found within 12000ms — the write may still be in flight; consider raising AGENT_LIVE_SETTLE_MS`

Relevant bounds today:
- `fetchById` in `ai-verify.ts` polls 24 × 500 ms ≈ **12 s**, hardcoded, not env-configurable.
- `AGENT_LIVE_SETTLE_MS` (`agent-live-shared.ts:26`) defaults to **5000 ms**, `process.env.AGENT_LIVE_SETTLE_MS`-overridable, and is slept *before* the poll in `landRun`.

On a loaded box — MJAPI, the test runner, and SQL Server all on one machine, which is exactly the documented release-build setup — exceeding that window is routine rather than exceptional.

## Ruled out — please do not re-investigate

These were each chased down during the v5.49.0 build and are dead ends:

1. **A product defect in `BaseAgent.initializeEngines` dropping `contextUser`** — not a product bug. It is a downstream artifact of the in-process path above. `initializeEngines` is unchanged since 2025-06-14, and MJAPI logged zero `contextUser` CRITICALs; only the test-runner process did.
2. **Agent-run persistence is broken / data loss** — disproved by the 42 + 186 row counts above.
3. **A regression in the 21 commits to `base-agent.ts` since v5.48.0** — investigated, not supported by any evidence.
4. **Stale `dist/` running an old check bundle** — ruled out by the timestamp comparison in Evidence #4.
5. **`model-noncompliance:` failures** (20 oracle failures across IT56 `agent-payload-guards` PG1–PG9 and IT57 `agent-artifact-tools` AT1–AT7) — not bugs. This is the suite working as designed, honestly surfacing real-model variance after 3 bounded attempts. Explicitly out of scope here.
6. **An expired `AI_VENDOR_API_KEY__GeminiLLM`** — real, but purely local credential rot, already fixed by swapping the key, not a repo issue. Google is the primary binding in every IT prompt's Google → OpenAI → Groq ladder, and it returned `400 API_KEY_INVALID` from `generativelanguage.googleapis.com`; that is why run 1 scored 4/15 vs run 4's 8/15. Noted only so the run-to-run delta above is not mistaken for signal.
7. **`[COST] $0.0000` despite real billed calls** — real, but already known and documented in DEPLOYMENT.md.

## Suggested fix directions

Options, deliberately not prescriptive — the right shape is the maintainers' call.

- **Preferred: make single-bundle runs genuinely equivalent.** Have `mj test run` perform the same client bootstrap `mj test suite` effectively ends up performing (`packages/TestingFramework/CLI/src/commands/run.ts`, which today calls `initializeMJProvider()` and nothing else). Then §4.6's triage instruction becomes true again.
- **Make transport declared, not inherited.** The deeper problem is that `CLIENT_BUNDLES` is a hardcoded set in the driver while the actual transport is whatever provider happens to be bound. Letting a bundle declare its required transport (and having `buildCheckContext` honour that declaration) removes the ordering dependency for every bundle at once, not just the agent ones.
- **Fail loudly instead of degrading silently.** Whatever else changes, a bundle written to run over the wire should assert it actually got a wire-capable provider and abort with a clear harness error, rather than executing in-process and reporting product-shaped failures. This alone would have prevented the phantom diagnosis.
- **Cheap mitigation if none of the above lands soon:** document in DEPLOYMENT.md §4.6 and `packages/TestingFramework/integration-test-suite/docs/build-engineering-runbook.md` that single-bundle triage is invalid for client-transport / live-agent bundles, and say what to do instead.
- **Separately, for the secondary item:** correct the `ai-verify.ts:40` message to state the elapsed bound and name `AGENT_LIVE_SETTLE_MS` rather than asserting the write never landed. Making the 12 s poll adaptive (or at least env-configurable, as the settle already is) would be a reasonable follow-on.

## Repro

Environment: freshly-provisioned scratch DB (SQL Server in Docker), MJAPI running on `:4000` from a current build, integration metadata seeded.

```bash
# 0. Seed integration metadata against the scratch DB
npx mj sync push --dir ./metadata-optional/integration-test --ci

# 1. Baseline — the bundle inside its suite. Watch MJAPI's log: agent execution hits it.
MJ_INTEGRATION_TEST=1 npx mj test suite "Integration Tests — Live Model"

# 2. The same bundle standalone — the step DEPLOYMENT.md §4.6 tells you to run.
#    MJAPI's log shows ZERO agent-execution hits; all 7 oracles (AL1–AL7) fail.
MJ_INTEGRATION_TEST=1 npx mj test run "IT53 - Agent Loop Foundation (live)"
```

To see the divergence directly, grep the MJAPI log for `RunAIAgent` / `agentRunId` / `prompt_execution` across each invocation (22 vs 0), and grep the test-runner output for `For server-side use of all engine classes` (0 vs 21).

Discovered during the MemberJunction **v5.49.0** release build, DEPLOYMENT.md Step 4, on 2026-07-22.

SDesai-BC:
Closes #3251.

## The finding

The live-agent integration bundles (`IT53`–`IT57`) **never actually ran the agent** — every run entered `BaseAgent` with a null `contextUser` and died in `BaseEngine.Load` before doing any work, so `IT53` reported **0 of 7 oracles passing** with a wall of `For server-side use of all engine classes, you must provide the contextUser parameter` CRITICALs. This happened **identically** in `mj test run` and `mj test suite`.

The issue diagnosed this as a *transport divergence* (single-bundle runs supposedly degrade to in-process while suite runs go over the wire). That mechanism is not what happens — verified by code-trace and by the suite run's own logs (216 `contextUser` CRITICALs logged by the **test-runner** process, zero by MJAPI). These bundles run the agent **server-in-process in both cases**, by design (the Q8 ruling, commit `799bb1d0b8`); the wire path is a separate bundle (`IT63`). The real cause is a dropped `contextUser`.

## Root cause

The Q8 conversion made the agents run in-process via `AgentRunner.RunAgent` but kept a client-era fallback for the context user, and no call site ever threaded `ctx.User`:

```ts
// agent-live-shared.ts (before)
export function makeAIClient(provider: IMetadataProvider): AgentInvoker {
    return { RunAIAgent: (params) => new AgentRunner(provider).RunAgent({
        ...params,
        contextUser: params.contextUser ?? provider.CurrentUser,   // <- null on the SQL provider
        provider,
    })};
}
```

`provider.CurrentUser` is populated only on a **client** (GraphQL) provider — the server tells the client who it is. On the CLI's `SQLServerDataProvider` nothing ever assigns it (`GetCurrentUser()` circularly returns `this.CurrentUser`), so it is null. `BaseEngine.Load` throws when `ProviderType === Database && !contextUser`, and the run dies in `initializeEngines`. The bundles that thread `contextUser: ctx.User` explicitly (`IT17`, `IT58`, `IT59`, `IT62`) are exactly the ones that passed.

## What this changes

- **`makeAIClient` / `resolveClient` now take a required `user: UserInfo`** and thread it; all **11** call sites pass `ctx.User`. A genuinely unresolvable user throws a **harness-attributed** error instead of handing a null user to `BaseAgent`. Making the parameter required means a missed call site is a compile error, not a runtime surprise. `resolveClient` also drops its dead `| undefined` return and the two unreachable `if (!client)` skip branches (post-Q8 it always returns an invoker).
- **`IntegrationTestDriver` fails loudly** if a `server`-transport bundle resolves a non-Database provider (a client-transport bundle rebound the process-global provider earlier in the process). This turns the previously prose-only "client members are sequenced last" suite-ordering rule into an enforced invariant. The guard is a no-op when `ProviderType` is absent (unset/facade provider in unit tests), so it only fires on a genuinely rebound client provider. Verified inert on both shipping suites: every server-transport member is sequenced before every client-transport member.
- **`runWithCompliance` reclassifies a no-run** (`scenario()` → `undefined`) as `agent-run-failed:` immediately, instead of retrying it 3× and laundering it into `model-noncompliance:`. This is exactly how the `contextUser` defect hid behind `IT56`/`IT57`. The sibling `withBoundedRetry` seam already hard-fails a no-run, so it is left as-is.
- **`ai-verify.ts`'s bounded poll** no longer asserts `fire-and-forget write never landed` (which claimed data loss that did not occur — in the release build the rows *did* land, the poll just closed first). It now states the actual bound it waited and adds a `MJ_IT_FETCH_POLL_MS` env knob (default `12000`ms) so loaded boxes can widen the window.
- **Docs**: rewrote the stale `CLIENT-TRANSPORT` / `over the GraphQL wire` docblocks to say server-in-process, and corrected `DEPLOYMENT.md` §4.6 + the build-engineering runbook (the single-bundle-equivalence contract and the `agent-run-failed:` vs `model-noncompliance:` distinction).

## Evidence

Diagnosis adversarially re-verified before any code was written (6 independent verifiers over the load-bearing claims: 20 confirmed, 0 refuted).

| Check | Result |
|---|---|
| `@memberjunction/integration-test-suite` unit | **131/131** pass |
| `@memberjunction/testing-integration` unit | **69/69** pass |
| Both packages build (`tsc`) | clean |
| Live smoke — `MJ_INTEGRATION_TEST=1 mj test run "IT53 - Agent Loop Foundation (live)"`, run twice against the release scratch DB with live model keys | **6/7 oracles** (was 0/7), **zero** `contextUser` CRITICALs (was 21), fixtures self-cleaned |

The one remaining live failure (`IT53` `AL6`) is a **separate, pre-existing** finding, not this defect and not a regression: with all model bindings deactivated the run finalizes non-Completed/non-Running (those asserts pass) but `AIAgentRun.ErrorMessage` is empty on the header. It became *visible* only because this fix lets the run reach its failure path at all. Filed separately.

## Risk

Test-framework only — no migration, metadata, or scheduled job, so no deploy-order hazard. The one behavior change a consumer could notice: the new driver guard converts a previously-silent misconfiguration (a suite interleaving client and server transports) into a loud, actionable `Error`. That configuration was already broken (server bundles were secretly running over the wire); failing loudly is the intended trade.

Full post-mortem and per-work-item detail: `plans/integration-test-expansion/fix-3251-live-agent-contextuser-and-transport-guard.md`.


rkihm-BC (commented):
## Review — thorough, with the full new test set run locally

Excellent PR. Correct root-cause diagnosis (which *corrects the issue's own hypothesis*), a minimal and type-safe fix, a loud-fail guard for the real misconfiguration, and unusually good test coverage. Recommend merge — only nitpicks below, none blocking.

### What I verified (not just read)
- **All 11 call sites thread `ctx.User`**, and grep confirms **no stragglers** — the "required param = compile error" safety net holds.
- **`integration-test-suite` units: 9/9** (invoker 6 + run-with-compliance 3); **`testing-integration`: 22/22** (driver 19 + ai-verify 3) — ran in a worktree, all green.
- **`ProviderType` plumbing is real**: enum is `{Database, Network}`; both `SQLServerDataProvider` and `PostgreSQLDataProvider` return `Database`, GraphQL returns `Network` — so the guard's `!== ProviderType.Database` is sound.
- **Guard placement**: inside the `transport === 'server'` branch only (after the client branch returns), so it can't false-fire on legitimately-Network client bundles. The `provider?.ProviderType &&` prefix keeps it inert on the facade/unset provider in unit tests — confirmed by the passing default-path tests.

### What's good
- **Intellectually honest diagnosis.** The PR disproves the issue's "transport divergence" theory with a code-trace *and* the suite's own logs (216 `contextUser` CRITICALs in the test-runner, zero in MJAPI), then names the real cause. Rare and correct.
- **Minimal, type-enforced fix.** Making `user: UserInfo` required turns a dropped call site into a compile error rather than a runtime null. Also removes genuinely dead code (`resolveClient`'s `| undefined` return + two unreachable skip branches).
- **`runWithCompliance` three-way outcome** — a no-run now fails immediately as `agent-run-failed:` instead of retrying 3× and laundering into `model-noncompliance:`. That laundering is exactly how the bug hid behind IT56/IT57; fixing it is the right structural change.
- **`ai-verify` truthfulness fix** — drops the false "fire-and-forget write never landed" (asserted data loss that didn't occur), states the *actual* waited bound, adds a tunable `MJ_IT_FETCH_POLL_MS`. A test even pins the non-multiple-of-interval rounding (700ms → reports 500ms).
- **Tests pin behavior, not lines** — explicit-param-wins, no-user-throws-*and-never-runs*, guard-fires-on-Network, guard-inert-on-Database.
- **Correct risk read + changeset scope.** Test-framework-only (no migration/metadata/job → no deploy-order hazard), and the changeset lists only `@memberjunction/testing-integration` because `integration-test-suite` is `private: true` (unpublished). Right call.

### What's concerning (all minor)
1. **The `?? provider.CurrentUser` fallback is now effectively dead** for typed callers — since `user` is required and non-null, `params.contextUser ?? user` always resolves first; `provider.CurrentUser` is reachable only when a caller passes `null as any` (only the test does). Harmless belt-and-suspenders, but it slightly muddies the "user is required" contract — `params.contextUser ?? user` alone (plus the throw guard) reads cleaner.
2. **Two near-duplicate invokers.** `makeAIClient` (agent-live-shared) and `resolveClient` (_it-live-agent-harness) now carry essentially identical contextUser-resolve-or-throw logic (only makeAIClient's `conversationId`/`data` handling differs). The PR extends the duplication rather than consolidating it — a shared `resolveContextUserOrThrow()` would keep the error message and fallback policy in lockstep. Both are tested, so low risk; a follow-up.
3. **The invoker throws synchronously** (the arrow fn isn't `async`), so it surfaces as a rejection only because every call site sits inside an async wrapper. A future `client.RunAIAgent(p).catch(...)` caller wouldn't catch it. Making the invoker `async` would make it always reject rather than sometimes throw. Very minor given the current call pattern.

### What's missing
- Essentially nothing. The one thing a unit suite can't cover — the actual "0/7 oracles + `contextUser` CRITICALs" regression — is inherently live-model-gated, so the invoker-contract unit tests are the right proxy, backed by the live smoke (6/7, was 0/7). Worth confirming the **separately-filed AL6 finding** (empty `AIAgentRun.ErrorMessage` when all bindings are deactivated) actually got an issue number, since this fix is what made it visible.
- Cosmetic: because the headline `contextUser` fix lives in the private package, the published `testing-integration` changelog won't mention it — technically correct per changeset semantics, just slightly under-informative.

### Bottom line
Strong, honest, well-tested. The diagnosis-correction alone makes it worth merging, and the loud-fail guard converts a class of silent misconfiguration into actionable errors. My only *substantive* suggestion is the invoker de-duplication (#2) so the two copies can't drift — a follow-up, not a merge blocker.

PR #3262, opened 2026-07-23
fix(migrate-convert): stop silently dropping statements in convert (#3252)

### LINKED ISSUE #3252: mj migrate convert silently drops statements and emits empty PG migrations while reporting success

SDesai-BC:
`mj migrate convert --split` could emit an empty or broken `.pg.sql` migration while printing `unhandled stmts: 0` and exiting 0 — an empty migration applies cleanly to a fresh PostgreSQL database, so it passed every gate and shipped. Three independent root causes, confirmed by deterministic repro against the real v5.49 migrations and fixed here at the dialect, classifier, and bake-path layers. Closes #3252.

## RC1 — block-less `IF` guards emitted a bare `;`

`V202607191254__…__Backfill_Missing_FK_Auto_Indexes.sql` uses the T-SQL block-less form (no `BEGIN`/`END`):

```sql
IF NOT EXISTS (SELECT 1 FROM sys.indexes WHERE name = '…' AND object_id = OBJECT_ID('…'))
    CREATE INDEX … ON … ([Col]);
```

The `IF-EXISTS` envelope in `mj_postgres.py` required the word `BEGIN`, so a block-less guard fell through to sqlglot, parsed as `exp.IfBlock`, and sqlglot's PG generator returned `""` (a *warning*, not an error) — six guards became six bare `;` with `unhandled: []`. Three dialect fixes: the envelope now captures a block-less guard's single governed statement (so `sys.indexes`/`columns`/`tables` guards translate to the same `DO $$ … pg_indexes … END IF $$` as the `BEGIN…END` form); an `exp.If`/`exp.IfBlock` guard plus an **EMPTY-EMISSION postcondition** report any node that renders to nothing rather than emitting it — closing the whole class, not just this instance; and an inline named `DEFAULT` constraint (`CONSTRAINT [DF_x] DEFAULT (75)`, invalid PG — this was the statement that crashed the `--bake-codegen` apply) has its name stripped.

## RC2 — a hand-written trigger was classified as CodeGen output and dropped

`MigrationStatementSplitter.ts` matched CodeGen objects by name convention:

```ts
const CODEGEN_NAME = /^(spCreate|spUpdate|spDelete|spRecompile|vw|fn|trgUpdate|trgCreate|trgDelete|trg)/i;
```

The bare `trg` alternative matched **every** trigger, so the hand-written `trgConversationDetail_AssignSequence` was classified `codegen-object`, which tripped the unbannered-snapshot heuristic, flipped the whole file into statement-mode, and dropped the trigger — the file reported a clean `converted` with empty T-SQL. Fixed by removing the bare `trg` alternative (a ledger sweep confirms it is safe: of 381 distinct triggers, 379 are `trgUpdate*`, zero are `trgCreate*`/`trgDelete*`, and the only two non-`trgUpdate` triggers are hand-written and now correctly routed) and by requiring a `vw*`/`sp*`/`fn*` object before an unbannered file flips into statement-mode, so a lone trigger/index can no longer route a hand-authored file into the drop path.

## RC3 — `--bake-codegen` applied gappy SQL then crashed with zero artifacts

Re-bake mode gated on conversion gaps before touching the working DB; forward mode did not — it applied the transpiled hand body even when it carried unhandled statements (including the RC1 named-default), so `db.apply` threw and the CLI's per-file `catch` called `this.error`, force-exiting with no `.needs-hand` file and no gap report. `IncrementalBaker` now gates forward-mode baking on gaps and returns `mode: 'gap-no-bake'` **without** touching the working DB (placed strictly after the re-bake and baseline branches, which are legitimately exempt). The CLI (`convert.ts`) halts at the first bake-mode gap with a guaranteed non-zero exit *before* the unconditional `process.exit(0)`, forces any gap to a `.needs-hand` artifact, writes an artifact on any throw instead of a bare `this.error`, and rejects `--allow-gaps` together with `--bake-codegen`.

## Reconciliation (issue's fix direction #2)

Adds statement-accounting as defense-in-depth. The dialect self-checks `parsed == emitted + unhandled + dropped` per call, surfacing an `ACCOUNTING-LEAK` gap on mismatch but **never raising** — a raise would propagate to the CLI catch and re-create the RC3 zero-artifacts failure on any routine-bearing file. Each conversion also carries a `reconciliation` block whose `suspiciousEmptyOutput` flag fires when substantive T-SQL produces no output and no gap. This was deliberately kept **coarse and soft** rather than an exact per-statement byte-balance: the dialect emits multi-statement `DO $$…$$` expansions and GO-batching changes chunk counts, so an exact balance is fragile and false-positive-prone — the same reason the external `check-pg-migration-content.mjs` detector (#3254) counts coarsely.

## A note on the two commits and the plan doc

The plan at `plans/pg-migration-architecture/ISSUE_3252_SILENT_DROP_FIX_PLAN.md` is included as the design + verification record. The original plan's Phases 3–4 were revised after an adversarial review found the exact-balance reconciliation would false-block good conversions and the hard assert would re-introduce RC3; the amendments (soft reconciliation, forward-only gate placement, intersection typing) are tracked inline with `[V]` markers. A separate `BakeApplyError` class from the plan was dropped as unnecessary — once the forward gate stops applying gappy SQL, `db.apply` only runs for clean conversions and the generic catch already covers any apply/capture throw.

## Evidence

Full-ledger sweep — every one of the **201** v5 source migrations run through the real `MJPostgresTranspiler`:

| Check | Result |
|---|---|
| Pipeline crashes | 0 |
| Accounting leaks (false `ACCOUNTING-LEAK`) | 0 |
| Bare-`;`-only bodies (the RC1 signature) | 0 |
| Reconciliation false-positives (`suspiciousEmptyOutput`) | 0 |
| Classification change vs. pre-fix code | exactly 1 file — `Fix_ConversationDetail_Sequence_Deadlock`, from an empty `converted` stub to `needs-hand-authoring` with the trigger body preserved (the intended RC2 fix) |

Target files after the fix: `Backfill_Missing_FK_Auto_Indexes` → `converted`, emits the `pg_indexes` `DO` block, 0 unhandled; both triggers → `needs-hand-authoring`; no named-default statements in any output.

Unit tests (all green): `@memberjunction/sqlglot-ts` 69 Python dialect + 69 TS; `@memberjunction/sql-converter` 1087; `@memberjunction/cli` 432. All three packages build clean.

## Not covered here

These packages are release-time build tooling — this PR adds no migration, no metadata, and no scheduled job, so there is no deploy-order hazard. The `--bake-codegen` path is validated at every unit seam (the `FakeDB` stub asserts the working DB is never touched on a gap), but its live end-to-end run against a real PostgreSQL working DB is a release-build step and was **not** exercised in this environment. The deterministic integration tier (`npm run test:integration`, which needs a live dev DB + MJAPI) was likewise not run here.


rkihm-BC (commented):
## Review — with the transpiler run locally against the pinned dependency

The TS-side fixes (RC2, RC3, reconciliation, CLI gating) are strong and well-tested — I ran them, 54/54 green. But I found a **critical version-skew bug in the RC1 Python fix that crashes the transpiler on the pinned sqlglot**, reproduced deterministically below. I'd treat it as merge-blocking; the fix is ~2 lines.

### 🔴 Blocker — `exp.IfBlock` does not exist in the pinned sqlglot (`~=27.18.0`)

`packages/SQLGlotTS/src/python/mj_postgres.py:1210`:
```python
if isinstance(stmt, (exp.If, exp.IfBlock)):
```
`exp.IfBlock` was introduced in **sqlglot 29.0.0**, but `requirements.txt` pins **`sqlglot~=27.18.0`** (27.18.x — no `IfBlock`). Python evaluates `exp.IfBlock` as an attribute access on *every* call, so this raises `AttributeError` on the first ordinary statement that reaches the plain-SQL path. Reproduced against the pinned version in a clean venv:

```
$ pip install -r requirements.txt        # → sqlglot 27.18.0
$ python -c "from mj_postgres import mj_transpile; \
    print(mj_transpile('ALTER TABLE [\${flyway:defaultSchema}].[Foo] ADD Bar INT NULL;')['sql'])"
  File ".../mj_postgres.py", line 1210, in _transpile_plain
    if isinstance(stmt, (exp.If, exp.IfBlock)):
AttributeError: module 'sqlglot.expressions' has no attribute 'IfBlock'
```
The identical input succeeds on sqlglot 29.0.0+. I bisected it: `hasattr(exp,'IfBlock')` is `False` for 27.18.0/27.19.0/28.0.0 and `True` from 29.0.0 onward.

Consequences:
- On **any environment that honors `requirements.txt`** (the reproducible-install contract that file exists to provide — its own header touts "reproducible conversion results"), `mj migrate convert --split` crashes on essentially every migration reaching the plain-SQL path — i.e. the release-build path this PR is about.
- That `AttributeError` propagates to the CLI's per-file `catch` → **the exact RC3 "throw → zero artifacts" pathology this PR set out to eliminate**, now reintroduced upstream of the RC3 gate.
- The `--self-test`/unit files can't even be *collected* on 27.18.0 (they import `mj_transpile`), so the PR's "69 Python + 69 TS green" evidence cannot have been produced against the pin — it must have run against a locally-installed sqlglot ≥29.0.0. No CI workflow installs sqlglot or runs these tests (`grep -r sqlglot .github/workflows` → nothing), so nothing catches the mismatch.

The irony: the comment 16 lines below (line 1226) already states *"sqlglot's node type … varies across versions (exp.If in some, exp.IfBlock in others)"* and deliberately detects that case version-tolerantly — but line 1210 hardcodes both names.

**Fix (surgical, keeps the pin, matches the author's own cross-version pattern):**
```python
_IF_NODE_TYPES = tuple(t for t in (getattr(exp, 'If', None), getattr(exp, 'IfBlock', None)) if t)
...
if isinstance(stmt, _IF_NODE_TYPES):
```
Alternative: bump the pin to `>=29.0.0` and re-run the full v5 regression per requirements.txt's upgrade procedure — heavier, since the pin exists precisely to freeze conversion output. Either way, please also add a CI step that installs `-r requirements.txt` and runs the Python dialect tests, so the pin and the code can't drift again.

### What's good (the rest is genuinely strong)
- **RC2** — removing the bare `trg` alternative from `CODEGEN_NAME` and gating the unbannered statement-mode flip on real snapshot evidence (`VIEW`/`PROCEDURE`/`FUNCTION`, not any `codegen-object` — which also covers triggers/auto-FK-indexes) is well-reasoned and backed by a ledger sweep (381 triggers, 379 `trgUpdate*`, only 2 hand-written).
- **RC3** — the forward-mode `gap-no-bake` gate that never touches the working DB on a gap; the CLI halt-at-first-gap placed *before* the unconditional `process.exit(0)`; writing a `.needs-hand` artifact on any throw instead of a bare `this.error`; and rejecting `--allow-gaps` with `--bake-codegen` — a clean, layered fix with the baseline/re-bake exemptions correctly preserved.
- **Reconciliation** — soft, never-raise accounting (`ACCOUNTING-LEAK` + `suspiciousEmptyOutput`), with the never-raise rationale correctly tied to not re-introducing RC3, and the honest note that it's coarse-by-design (DO-block/GO-batch expansions make exact balance false-positive-prone).
- **EMPTY-EMISSION postcondition** closes the whole "generator warned and returned nothing" class rather than just the IfBlock instance — good instinct (and it would be the backstop for the block-less case *if the isinstance above didn't crash first*).
- Tests pin real behavior — halt semantics, MAJOR-2 (`converted`+`gap-no-bake` still forced to `.needs-hand`+halt), baseline exemption (MAJOR-3), forward-gate no-apply. **48 SQLConverter + 6 MJCLI decision tests pass locally.**

### Minor
- `decideConvertWrite` extracted as a pure, unit-testable function — nice.
- `MigrationConversionResult.reconciliation` is now required; compiles because builds are green, and the baker/CLI use `Pick`/intersection so they're unaffected — worth a quick confirm no external constructor was missed.

### Bottom line
Excellent TS-side work; the RC2/RC3/reconciliation design is exactly right. But the RC1 change ships a hard crash against the committed dependency pin, and the green-test evidence was produced against a different sqlglot than the one `requirements.txt` specifies. Fix the `exp.IfBlock` reference (and pin the CI to the same version it tests), and this is a merge.

SDesai-BC:
@rkihm-BC good catch on the `exp.IfBlock` crash. Fixed with a getattr-based tuple (`_IF_NODE_TYPES`) so it degrades safely on the pin, and added a CI step that installs `requirements.txt` and runs the dialect tests against 27.18.0 so it can't drift back in. Ran your repro against the pinned venv just now — clean.

rkihm-BC (approved):
## Approve — with one small ask (2-char typing fix)

I verified the fixes rather than trusting the description: ran the pinned dialect suite in a clean `sqlglot~=27.18.0` venv (**all dialect tests pass, exit 0**, incl. the new `check_exp_attribute_references()` guard — 60 `exp.*` names checked), reproduced the earlier `AttributeError` repro → now emits valid PG, and confirmed the new **"Python dialect tests (pinned sqlglot)"** CI step actually ran and passed on this PR (`all dialect tests passed` / `60 names checked`). All 5 checks green. Two independent deep reviews of the TS and Python layers confirmed RC2/RC3/reconciliation and the RC4/RC5 dialect gates are implemented correctly. Deploy safety holds — no migrations, metadata, or scheduled jobs; this is release-time build tooling.

The `exp.IfBlock` version-skew blocker is genuinely fixed and now can't drift back silently. 

### One ask before merge (non-blocking-ish, 2-char fix)

**`decideConvertWrite` widens two value-list unions to `string`** — [`packages/MJCLI/src/commands/migrate/convert.ts:51`](https://github.com/MemberJunction/MJ/blob/fix/3252-migrate-convert-drops-statements/packages/MJCLI/src/commands/migrate/convert.ts#L51):

```ts
result: { status: string; unhandled: MigrationConversionResult['unhandled']; mode?: string },
```

`unhandled` is already correctly derived from the result type right beside them, so the pattern was known — `status` and `mode?` should be derived too rather than widened to `string`:

```ts
result: { status: MigrationConversionResult['status']; unhandled: MigrationConversionResult['unhandled']; mode?: BakedMigrationResult['mode'] },
```

(use whatever the baker's actual `mode` union type is named). Soft CLAUDE.md rule 2c drift — as written, a future typo like `'gap-no-bake '` or a renamed mode value would be silently accepted by the `=== 'gap-no-bake'` / `=== 'needs-hand-authoring'` checks. Trivial to fix.

### Minor, optional (not asking for changes)
- **Double-`RAISERROR` in one guard** drops the second unaccounted (`IF EXISTS(...) BEGIN RAISERROR('a'...); RAISERROR('b'...) END` → single `RAISE EXCEPTION 'a'`). Defensible since PG's `RAISE EXCEPTION` aborts and real guards never do this — flagging only against the strict "never silently drop" contract.
- Sub-statement discards (e.g. an ISJSON `CHECK` filtered from a multi-constraint `ADD`) aren't in the `dropped` ledger. Accounting is statement-granular by design, so the invariant still holds.

### Worth restating (the PR is honest about it)
The `--bake-codegen` live end-to-end run against a real PostgreSQL working DB wasn't exercised — only the `FakeDB`-stubbed unit seams. Logic is sound and every unit boundary asserts the working DB is untouched on a gap, but the first real release-build bake is the true integration test; don't treat the green checks as proof that path works live.

Great work — the design is exactly right and the blocker can't regress silently. Approving; please land the typing fix (#1) since CLAUDE.md is strict about derived union types.

PR #3264, opened 2026-07-23
fix(core): bound ArtifactMetadataEngine boot load + guard oversized cache entries

bc-izygmunt:
## Summary

Fixes an unbounded application-boot payload (and a secondary full-cache wipe) caused by `ArtifactMetadataEngine` eager-loading every artifact **version — including the full `Content` column** — at startup. On deployments with real artifact volume this made every cold boot download the entire artifact-version corpus and, because the result dwarfs the client cache budget, evict the **entire** `LocalCacheManager` cache on each load.

Addresses the Blue Cypress / Izzy PRD (~12s boots traced to ~244 MB of `ArtifactVersion.Content` loaded at boot).

### 1. `ArtifactMetadataEngine` — bounded boot, on-demand content (`@memberjunction/core-entities`)
- `Config()` now loads **only** `MJ: Artifact Types` (a small, fixed registry) at boot. Artifacts and versions are **no longer bulk-loaded**.
- New on-demand APIs, memoized in per-ID maps:
  - `LoadVersionsForArtifact(artifactId, contextUser?, provider?, forceRefresh?)` — loads one artifact's versions (payload bounded to a single artifact).
  - `GetVersionContent(versionId, ...)` — loads a single version incl. `Content`.

Boot cost is now **independent of total artifact content** in the database.

### 2. `LocalCacheManager` — oversized-entry write guard (`@memberjunction/core`)
- New `maxEntryPercentOfCache` config (**default 25%** of the cache budget) + an `exceedsMaxEntrySize()` gate at the top of `SetRunViewResult` and `SetRunQueryResult`.
- An entry estimated larger than the cap is **skipped** (logged; data still returned to the caller uncached) **before** eviction runs. Previously an oversized write forced `evictIfNeeded` to try to free more than the whole budget — evicting every entry and then failing to persist anyway (a full cache wipe on every store). This is a standalone hardening that helps every engine, not just artifacts.

### 3. Sync accessors renamed to `Cached*` — intentional compile break (per review)

Following the review discussion, the artifact/version **synchronous accessors** were **renamed** rather than left under their original names. Their old names implied a full boot-time load that no longer happens, so a stale external consumer (Skip, Betty, etc.) would have silently received empty/null results — a hidden runtime regression. The rename turns that into a **hard compile error** instead:

| Old (implied full corpus) | New (session-scoped on-demand cache) |
|---|---|
| `Artifacts` | `CachedArtifacts` |
| `ArtifactVersions` | `CachedArtifactVersions` |
| `FindArtifactByID()` | `FindCachedArtifactByID()` |
| `FindArtifactVersionByID()` | `FindCachedArtifactVersionByID()` |
| `GetVersionsForArtifact()` | `GetCachedVersionsForArtifact()` |

The `Cached*` prefix makes it unmistakable that these return only what has been fetched on demand this session (via `LoadVersionsForArtifact` / `GetVersionContent`) — never the whole system's artifacts. The **type-registry accessors are unchanged** (`ArtifactTypes`, `FindArtifactType`, `FindArtifactTypeByID`, `GetArtifactTypeByMimeType`) — they are still boot-loaded, so their names were never misleading.

### 4. Consumer migration
- `data-artifact-viewer` now calls `LoadVersionsForArtifact()` before its sync reads (it parses `Content` across an artifact's versions), reading via `GetCachedVersionsForArtifact()`.
- `lazy-artifact-info` uses `FindCachedArtifactByID()` / `FindCachedArtifactVersionByID()` behind its `engine.Loaded` guard, falling back to a direct DB load (the normal hot path now).
- Refreshed stale comments in `lazy-artifact-info` and `SharedService.preWarmEngines` ("lightweight (just artifact types)") to match the new reality.

## Adversarial consumer audit

All 36 files referencing `ArtifactMetadataEngine` were audited (see the automated review comment on this PR). Every in-repo consumer is one of: type-registry-only (still boot-loaded), direct entity loading (bypasses engine caches), on-demand with a guarded `engine.Loaded` fallback, or explicitly migrated here. **No internal consumer assumed full-corpus loading.** The rename in §3 is the guard for *external* consumers.

## Testing

- **Unit:** MJCore **1573 passing** (incl. **24** cache-gate tests in `localCacheManager.maxEntrySizeGate.test.ts`); MJCoreEntities **537 passing**; ng-artifacts **61 passing**; ng-conversations **881 passing**.
- **Integration:** deterministic tier **52/52 passing (100%)** against live MJAPI over the GraphQL wire — including the client-first cache, RunView, RunQuery, and artifact bundles that exercise these changes (and the renamed accessors) end-to-end.
- **Builds clean:** `@memberjunction/core`, `@memberjunction/core-entities`, `@memberjunction/ng-artifacts`, `@memberjunction/ng-conversations`, Explorer `shared`, and `@memberjunction/server`.

## Notes for reviewers

- No schema/migration changes; no new resolvers, GraphQL types, or entities.
- ⚠️ **Intentional API break** (per review): the five sync accessors in §3 were renamed. This forces a compile error for any consumer that assumed full-corpus loading — deliberate, and does not warrant a semver major (the misleading names were the risk). The new on-demand APIs (`LoadVersionsForArtifact`, `GetVersionContent`) and the type-registry accessors are the supported surface going forward.
- A **changeset** is present (managed by the maintainer). Affected packages: `@memberjunction/core` and `@memberjunction/core-entities` (behavioral); the Angular packages had consumer + comment changes.


AN-BC (commented):
@bc-izygmunt this looks good overall, the main thing I want to ensure is that we ask @claude  to run a full adverserial review of the repo to ensure that there are no consumers of this engine that assume that artifacts and versions are fully loaded. Since the properties for accessing artifacts are the same, by signature, this is a regression risk, internal to the MJ repo and also for other consumers, possibly Skip (@CaelebB-BC @jordanfanapour ) or Betty (@dray-mc @topherHunnewell @cmbrock-BC ).

Another approach would be to drop those properties from the engine and put in place alternative property names that are more correct for what we are doing, this is a bit of a break in compatibility but I don't think justifies a semver major anyway and would ensure that there is a hard compile error with any external consumers that assume this has all artifacts/versions. I prefer this approach as it is forces the issue rather than a possible hidden error.

cmbrock-BC:
> @bc-izygmunt this looks good overall, the main thing I want to ensure is that we ask @claude to run a full adverserial review of the repo to ensure that there are no consumers of this engine that assume that artifacts and versions are fully loaded. Since the properties for accessing artifacts are the same, by signature, this is a regression risk, internal to the MJ repo and also for other consumers, possibly Skip (@CaelebB-BC @jordanfanapour ) or Betty (@dray-mc @topherHunnewell @cmbrock-BC ).
> 
> Another approach would be to drop those properties from the engine and put in place alternative property names that are more correct for what we are doing, this is a bit of a break in compatibility but I don't think justifies a semver major anyway and would ensure that there is a hard compile error with any external consumers that assume this has all artifacts/versions. I prefer this approach as it is forces the issue rather than a possible hidden error.

For Betty's current usage of ArtifactMetadataEngine this would have no impact.

bc-izygmunt:
> @bc-izygmunt this looks good overall, the main thing I want to ensure is that we ask @claude to run a full adverserial review of the repo to ensure that there are no consumers of this engine that assume that artifacts and versions are fully loaded. Since the properties for accessing artifacts are the same, by signature, this is a regression risk, internal to the MJ repo and also for other consumers, possibly Skip (@CaelebB-BC @jordanfanapour ) or Betty (@dray-mc @topherHunnewell @cmbrock-BC ).
> 
> Another approach would be to drop those properties from the engine and put in place alternative property names that are more correct for what we are doing, this is a bit of a break in compatibility but I don't think justifies a semver major anyway and would ensure that there is a hard compile error with any external consumers that assume this has all artifacts/versions. I prefer this approach as it is forces the issue rather than a possible hidden error.

renamed the sync accessors so external consumers get a hard compile error rather than a silent empty-result regression

AN-BC (approved):

AN-BC:
> so

@bc-izygmunt LGTM merge away

PR #3266, opened 2026-07-24
fix: Collection sharing — server share gate always rejecting, masked errors, Sharing Center navigation, and shared-indicator UI

CaelebB-BC:
## Summary

Fixes collection sharing end-to-end: shares failed server-side for **every** user with an unreadable generic error, shared collections couldn't be opened from the Sharing Center, and the shared indicator in the Collections UI was visually broken.

### 1. Server-side share-authorization gate always rejected (the core bug)

`MJCollectionPermissionEntityExtended.callerMayShareCollection()` (and the artifact sibling) ran its owner/grant checks through `new RunView()` with **no contextUser**. On MJAPI the global provider has no `CurrentUser`, so `InternalRunView` threw `User not found in metadata and no contextUser provided to RunView()`, both gate queries returned `Success: false`, and **every** share attempt was blocked — even for the collection owner.

**Fix:** the gate queries now run through the entity's own provider (`RunView.FromMetadataProvider(this.ProviderToUse)`) with `this.ContextCurrentUser` passed as contextUser, per the repo's server-side provider/contextUser rules. Applied to `callerMayShareCollection()`, `callerMayShareArtifact()`, and `fetchArtifactName()`.

### 2. Real error message swallowed → "Unknown error creating record"

`failSave()` in `BaseShareEntityExtended` recorded the block reason on `entity.LatestResult` — but the gate fires **before** `super.Save()`, and `LatestResult` is `null` until a save attempt pushes onto `ResultHistory`. The reason was silently dropped and the resolver fell back to the generic `Unknown error creating record`.

**Fix:** when no result history exists yet, `failSave()` pushes a `BaseEntityResult(false, reason, 'create')` so the actual reason reaches `LatestResult.CompleteMessage` and therefore the client.

### 3. Sharing Center couldn't open shared collections

`SharingCenterDialogHostComponent.openResourceForExplorer()` had no `'Collection Permissions'` case — clicks on shared-collection rows fell into `default: return false` and did nothing.

**Fix:** new `openCollection()` handler. Collections have no `NavigationService.Open*` helper (they live behind the Chat app's *Collections* nav item), so the handler finds whichever app exposes the Collections nav item and calls `SwitchToApp(app.ID, navItem.Label, { collectionId })` — the established pattern that works for both fresh tabs (initial param read) and cached ones (`OnQueryParamsChanged`).

### 4. Shared indicator styling in the Collections UI

- **Grid badge:** the `fa-users` badge icon was inheriting `font-size: 36px` from the higher-specificity `.grid-icon.folder-icon i` rule, rendering as a giant glyph overflowing the badge. That rule is now direct-child scoped (`> i`), and the badge is restyled as a clean surface bubble (token border/shadow, brand-primary icon) that adapts to dark mode.
- **List view:** the bare shared icon (which was also losing its styling to `.list-name-cell i`) is now a proper "Shared" chip — brand-tinted pill via `color-mix` on design tokens, matching the existing badge idiom.
- **Owner line:** long unbreakable values (emails) overflowed the card; now single-line ellipsis truncation.
- **Owner display name:** cards/list now show a friendly owner name — one batched narrow-field query resolves `FirstName LastName` from `MJ: Users` (keys normalized via `NormalizeUUID`), falling back to the denormalized `Owner` value.

- The client no longer offers Share on legacy null-`OwnerID` collections — `canShare()` and `validatePermission('share')` require a real owner match or an explicit `CanShare` grant, closing the permit-then-deny mismatch with the server gate. Edit/delete keep the null-owner backwards-compat rule. The full ownership-claim design for legacy collections remains a follow-up.
- New `ShareEntityCreateGate.test.ts` — 9 tests pinning the gate's provider/contextUser threading (the exact original bug), the `failSave` `LatestResult` recording (owner allows / non-owner blocks with readable reason / grant-holder allows / gate-query-failure still records reason / update and client-side short-circuits), mirrored for the Artifact entity.
- `openCollection()` matches the Collections nav item by `DriverClass === 'ChatCollectionsResource'` (stable identity) instead of the English label, skips non-`Active` items, and passes the resolved `Label` to `SwitchToApp`.
- `loadOwnerNames()` clears `ownerNameMap` up front so a failed reload falls back to denormalized names instead of serving stale entries.

## Testing

- `@memberjunction/core-entities`: build clean, **546/546** unit tests pass (24 files — includes the 9 new gate regression tests)
- `@memberjunction/ng-conversations`: build clean, **881/881** unit tests pass
- `@memberjunction/ng-explorer-core`: build clean, **76/76** unit tests pass
- Verified visually in MJExplorer (dark mode): badge renders as the small corner bubble, owner line truncates, share flow exercises the corrected server gate


AN-BC (approved):
LGTM

PR #3269, opened 2026-07-24
docs(claude-md): cut root instruction file from 2,256 to 195 lines

SDesai-BC:
Root `CLAUDE.md` had reached **2,256 lines / 153 KB — roughly 38,100 tokens loaded into every session before a single line of code is read**. Anthropic's guidance is ~200 lines per `CLAUDE.md`, and the reason is not only cost: their docs state that bloated files *cause Claude to ignore your instructions*. Past a point, every rule added makes the rules already there less likely to be followed — the opposite of what each addition intended.

The decay was already measurable. The guide index in root referenced 26 of the 40 guides on disk; `guides/README.md` indexed 19; neither was complete. Roughly 5,300 tokens duplicated four nested `CLAUDE.md` files, giving two sources of truth for the same rules. Three nested files were absent from the index entirely. Two `plans/` references pointed at files that no longer exist.

No issue — this came out of a direct request to shrink the file without losing guidance.

## What this does

Content is routed by two questions: *is violating it unrecoverable?* (those stay in root, because lazily-loaded content is dropped by `/compact` while root is re-injected) and *what is its scope?*

| Destination | Loads when | Carries |
|---|---|---|
| **root** (195 lines) | always; survives `/compact` | no-commit / no-destructive-git / branch-tracking, definition of done, always-true facts, routing table |
| **`.claude/rules/`** (new, 4 files) | a matching file is opened | `data-access` + `typescript-style` (`**/*.ts`), `design-tokens` (`**/*.{scss,css}`), `testing` (`**/*.test.ts`) |
| **nested `CLAUDE.md`** (16 total) | a file in that tree is read | 3 new (`MJCoreEntities`, `MJAPI`, `CodeGenLib`); Angular / Actions / migrations / metadata absorbed their root duplicates |
| **skills** | on invocation | `debug-build-failures` (new); MJ dev-server loop folded into `playwright-cli` |
| **`guides/README.md`** | on read | the 7,267-token index, backfilled 19 → **all 40** guides |

### `@import` was deliberately not used

The intuitive fix — split the file up and `@import` the pieces back — saves nothing. The documentation is explicit that imported files *"still load and enter the context window at launch."* Imports are an organization tool with zero token benefit. Only nested `CLAUDE.md` and path-scoped rules actually defer loading. This is recorded here so it isn't rediscovered.

### Why root isn't emptied

Lazily-loaded content does not survive `/compact`; root is re-injected. So a rule whose violation is **irreversible** must stay in root — "never commit without approval", "never `git reset --hard`", branch-tracking. A rule caught in code review can be lazy. That distinction, not token count, decided every move.

## Deletions

Three sections were removed, each with a reason recorded in the manifest:

- **Claude Code Fast Mode** — described Opus 4.6 and a `settings.json` workaround; superseded, and stale instructions about which model is running are worse than none.
- **Monorepo Structure** — three bullets stating packages live under `/packages`, each has its own `tsconfig`, and `turbo.json` drives build deps. All directly observable from the tree. The one non-obvious part (`npm install` at repo root only) is **kept** in root.
- **`Active Technologies` / `Recent Changes`** — auto-generated speckit residue describing an unrelated `601-mcp-oauth` feature.

Everything else was relocated, not dropped. `.claude/claude-md-manifest.json` maps all **46** original sections to destinations, and the gate fails if any section is unaccounted for or any deletion lacks a written reason.

## Second commit: what smoke testing found

Validating the result against the skill-creator validator and a rule-by-rule preservation sweep surfaced four defects, all fixed in `e99341852a`.

The worst was a pointer that had become actively misleading. `packages/Angular/CLAUDE.md` said:

    See the root [CLAUDE.md](../../CLAUDE.md) rule #4 for the full policy. Summary:

That rule was removed by this change — and root's rule 4 is now *"start work on a feature branch"*, so it resolved to a real but unrelated rule. The full standalone-vs-NgModule policy is restored inline, including bullets the 6-line summary had dropped ("never mix within a single component", the `imports`-array requirement).

Alongside it: **11 references into root were orphaned** by relocating its content (in `MJServer` and `AuthProviders` READMEs, `docker/regression/ARCHITECTURE.md`, three guides), one a dead heading anchor. A contradiction was introduced — root summarized decomposition as *"a smell to investigate"* while `typescript-style.md` carries the project's verbatim *"MAXIMUM ~30-40 lines … refactor immediately"*; contradictory instructions get resolved arbitrarily, so the paraphrase is gone and the rule owns it. And `metadata/components/claude.md` was **lowercase**, so 359 lines never loaded as a memory file on any case-sensitive filesystem — renamed, content unchanged.

The checker missed three of those four because it validated that linked *files* exist, not that the sections they point at do, and its corpus excluded guides. It now validates heading anchors and scans 65 files instead of 24.

## Evidence

Both gates re-run on the pushed tree:

| Check | Result |
|---|---|
| `npm run check:claude-md` | ✅ 46 sections accounted for · 195/260 lines · 594 links + 7 anchors across 65 instruction files · **134 inbound references across all 1,727 markdown files** · 16 nested files routed · 4 rules validated · 40 guides indexed |
| `npx vitest --config .github/scripts/vitest.config.mts` | ✅ 104 passed (4 files) — includes 31 new tests for the checker |

### Measured token cost

Re-derived from the committed files, with HTML comments excluded since they are stripped before injection:

| Session type | Tokens | vs 38,122 |
|---|---:|---:|
| docs / SQL / JSON / metadata only | 3,083 | **92% less** |
| any `.ts` file touched | 16,258 | **57% less** |
| all rules fire (worst case) | 18,855 | **51% less** |

**57% is the figure for typical work, not 92%.** The `**/*.ts` rules match most sessions in a TypeScript monorepo. Narrowing those two globs is the obvious next lever.

### A/B compliance test

Because the real risk is behavioural, not numeric, the two instruction sets were tested head to head: same four tasks, same model, blind grading against 26 assertions drawn from MJ's own documented rules, with the arm→X/Y mapping alternated per task to prevent position bias.

| Task | File type | Old monolith | New routed |
|---|---|---:|---:|
| `ts-data-access` | `.ts` | 8/8 | 8/8 |
| `scss-design-tokens` | `.scss` | 7/7 | 7/7 |
| `ts-entity-naming` | `.ts` | 5/5 | 5/5 |
| `sql-migration` | `.sql` | 6/6 | 6/6 |
| **Total** | | **26/26 (100%)** | **26/26 (100%)** |

The migration case was designed to fail: the new arm got **no** migration rules pre-loaded, because they live in `migrations/CLAUDE.md`, which loads only when a file in that subtree is read — and the model was *creating* a new file there. It had to notice the routing table and go read them. It did, and beat the monolith on the one real gap the grader found: it used `v5.51.x` correctly where the old arm reused `v5.50.x`. The increment rule exists **only** in `migrations/CLAUDE.md` and was never in the 2,256-line root, so the monolith could not get it right.

Three limits worth stating. Both arms hit 100%, so the instrument saturated — this is solid evidence of **no regression** and weak evidence of improvement. Old-arm agents still had the new root auto-loaded by their own harness and were only *instructed* to ignore it, which biases toward a tie. And it is four tasks with one sample each, no variance estimate. An earlier run of this test was discarded entirely: a fixture path resolved to `undefined`, both arms silently fell back on the same instructions, and it produced clean-looking numbers that meant nothing.

## Review follow-up (`537b9b5052`)

Review caught a reference this refactor left pointing at a relocated section, from a file the checker could not see. Verifying it turned up a defect in the checker itself.

**The checker only guarded one direction.** Anchor validation ran over its 65-file instruction corpus, which covers links going *out* of instruction files. The hazard is the other direction: moving a section out of root orphans every reference *into* it, and those live in `plans/` and package docs. A new check now scans **all 1,727 markdown files** but validates only links whose *target* is an instruction file (a `CLAUDE.md`, or anything under `.claude/rules/`). Scoping by target is what makes it usable — widening the corpus to `plans/**` wholesale, the obvious move, pulls in **111 pre-existing broken links** unrelated to instruction files, and grandfathering that many would bury the signal the check exists to produce.

**The slugifier was wrong.** It trimmed after stripping punctuation, dropping the leading/trailing hyphen GitHub keeps where an emoji was removed — `## 🚨 Button Styling 🚨` slugs to `-button-styling-`, not `button-styling`. With **34 emoji-prefixed headings** across this repo's `CLAUDE.md` files that was a live false-positive source, and it had already produced three bogus `knownBrokenReferences` entries, now removed because those anchors were never broken.

Repairs: `plans/complete/repair-conversations-ui-performance.md` pointed at root's RunView section, which moved to `.claude/rules/data-access.md` — and was doubly wrong, since `../CLAUDE.md` from `plans/complete/` resolves to `plans/CLAUDE.md`, which has never existed at any revision. `packages/Actions/GENERATED-ACTIONS.md` used root's heading against the Actions file (invalid at `next` too). Ten links in `plans/` named an instruction file with the wrong `../` depth — fixed rather than grandfathered, since each had one unambiguous existing target.

One is grandfathered with the reason recorded: `plans/MEMORY_LEAK_AUDIT.md -> ../../../CLAUDE.md#memory-leaks` escapes the repo root **and** `#memory-leaks` was never a heading in root at any revision. No instruction file documents listener-leak antipatterns, so there is no correct target — repointing would yield a link that resolves but does not contain what the sentence claims.

## Hazards

**The `claude.md` → `CLAUDE.md` rename is a behaviour change.** 359 lines of component-authoring guidance that previously loaded for nobody on Linux or CI will now load for anyone working under `metadata/components/`. Content is byte-identical (`R100`); if that guidance has gone stale, this is when it starts being followed.

**`check:claude-md` is not a CI gate yet.** The npm script runs locally; nothing in `.github/workflows/` invokes it, so it will not block a PR. Its 25 tests *do* run in CI through the existing `.github/scripts/vitest.config.mts` job (`test.yml:185`). Wiring the gate itself is a deliberate follow-up rather than something to slip in here.

**Seven pre-existing broken references are grandfathered** in the manifest as a ratchet, not an amnesty: listed entries report as notes, anything new fails. Tests cover both halves. The list started at nine; three were removed once the slugifier stopped falsely reporting them (see below) and one was added for a reference with no correct target.

No migrations, no `metadata/**` changes requiring `mj sync push`, no scheduled jobs. The changeset is intentionally empty (`releases: []`) — no published package's code or behaviour changes, and naming any package would bump all of them, since `config.json` declares `@memberjunction/*` as `fixed`.

Six files are modified in my working tree and deliberately **not** in this PR: `.vscode/settings.json`, `package-lock.json`, and four generated `mj-class-registrations.ts` manifests. I did not author them and they appeared mid-session.


AN-BC (commented):
LGTM, but would like @rkihm-BC to put eyes on this before you merge.

rkihm-BC:
Reviewed in depth — the mechanism checks out (`.claude/rules/` path-scoped loading is a native, documented Claude Code feature; the compaction/routing reasoning is correct), the gate passes cleanly on the pushed tree, content preservation holds up under a semantic spot-check, and all four second-commit repairs are verified. Nice work, and the manifest-as-evidence + ratchet approach is excellent.

One thing to address before merge:

**A reference this PR freshly orphaned sits outside the checker's corpus.** [`plans/complete/repair-conversations-ui-performance.md:60`](../blob/worktree-refactor-claude-md/plans/complete/repair-conversations-ui-performance.md#L60) links to `../CLAUDE.md#runview-resulttype-and-fields-optimization` — a section this change relocated to `.claude/rules/data-access.md`. The file-level link still resolves, so nothing flags it, but the anchor is now dead. The root cause is that the checker's 65-file corpus covers `CLAUDE.md` + `.claude/rules` + `.claude/skills` + `guides/`, but **not** `plans/**` or non-README package docs (e.g. `packages/Actions/GENERATED-ACTIONS.md`), so the anchor-validation gate can't see it.

Recommendation:
- Repoint that link to `.claude/rules/data-access.md` (or grandfather it in the manifest), **and**
- Consider widening the checker corpus to include `plans/**` so this class of orphan can't recur silently.

Note: the only other out-of-corpus hit — `plans/MEMORY_LEAK_AUDIT.md#memory-leaks` — is pre-existing (no "Memory Leaks" section ever existed in the old root), so it's not caused here; fold it into `knownBrokenReferences` or leave it, your call.

Everything else I'd file as advisory only. Approving once the above is dispositioned.

Issue #2975, opened 2026-06-29
open-app-engine: `mj app install` corrupts mj.config.cjs (missing comma before `dynamicPackages`), breaking all later app commands


MS-BC:
## Summary
`mj app install <connector>` (open-app-engine `InstallApp`) appends a `dynamicPackages` block to the repo's `mj.config.cjs` **without the separating comma** after the preceding top-level property (the `openApps` block's closing `}`). The result is invalid JavaScript, so the file can no longer be `require()`-d.

## Impact
- The corrupted `mj.config.cjs` is **committed** by the install.
- Every subsequent `mj app *` command (and anything else that loads `mj.config.cjs`) fails with:
  ```
  SyntaxError: Unexpected identifier 'dynamicPackages'
  ```
- Net effect: the **first** connector/app installs fine; the **second** install fails. The instance can't add more connectors until the file is hand-repaired. The failure is silent at install time and only surfaces on the *next* command.

## Repro
1. Repo seeded from a template whose `mj.config.cjs` top-level object ends with an `openApps: { ... }` block.
2. `mj app install https://github.com/MemberJunction/Integrations/Marketing/PheedLoop` → succeeds, appends `dynamicPackages`.
3. `mj app install <any second connector>` → fails parsing `mj.config.cjs`.

## Actual generated output (excerpt)
```js
  openApps: {
    serverPackagePath: 'apps/MJAPI',
    clientPackagePath: 'apps/MJExplorer',
    github: { token: process.env.GITHUB_TOKEN }
  }            // <-- missing comma here

  dynamicPackages: {
    server: [ { PackageName: '@memberjunction/connector-pheedloop', StartupExport: 'registerConnector', AppName: 'connector-pheedloop', Enabled: true } ],
    client: []
  },
};
```

## Expected
The appended `dynamicPackages` key must be comma-separated from the preceding property, i.e. close the prior block with `},` before inserting `dynamicPackages:`, so `mj.config.cjs` stays valid JS.

## Version / env
- `@memberjunction/cli` **5.43.0** (open-app-engine `InstallApp`)
- Node 24

## Workaround
Insert the missing comma manually; or validate+repair after install — `node -e "require('./mj.config.cjs')"` catches it, and a targeted fix (regex: a `}` directly followed by `dynamicPackages:` → insert the comma) repairs it idempotently.

MS-BC:
Fixed in #2988 (branch `fix/openapp-config-comma-2975`, against `next`).

### Root cause
[...]

PR #3272, opened 2026-07-24
fix(open-app): detect workspace layout and write config to every consumer target (#3270, #3271)

### LINKED ISSUE #3270: mj app install assumes the monorepo packages/MJAPI layout; fails on the apps/ layout created by mj install
## Summary

`mj app install` resolves the server/client workspaces at the monorepo defaults (`packages/MJAPI`, `packages/MJExplorer`), but `mj install` — MJ's own distribution installer — scaffolds them at `apps/MJAPI` and `apps/MJExplorer`. Installing an Open App onto a host created by `mj install` therefore fails partway through, **after** the database work has already committed.

## Reproduction

```bash
# 1. official distribution host
mj install -t v5.49.0 -y -c ./install.config.json     # creates apps/MJAPI + apps/MJExplorer

# 2. install any Open App
mj app install https://github.com/MemberJunction/bizapps-sonar --version v0.4.0 --non-interactive
```

Observed:

```
- [Schema]    Creating schema '__mj_BizAppsSonar'...
- [Migration] Running 7 migration(s)...
✔ [Record]    App 'mj-bizapps-sonar' recorded as installed
- [Config]    Updating configuration files...
- [Packages]  Adding packages...
✖ Install failed: Could not read package.json at <root>/packages/MJAPI/package.json:
  ENOENT: no such file or directory
```

Because the failure lands after `[Record]`, the app is left in `__mj.OpenApp` with `Status='Error'` and no package/config wiring.

## Second failure on the same host

Supplying the paths gets past `[Packages]`, then `[Config]` fails:

```
✖ Install failed: Failed to update config: Could not find a config object in mj.config.cjs
  to insert into. Expected either `module.exports = { ... }` or `module.exports = <variable>;`
```

The distribution ships `apps/MJAPI/mj.config.cjs` as a re-export:

```js
module.exports = require('../../mj.config.cjs');
```

which the config writer's parser (`open-app-engine/dist/install/config-manager.js`, `InsertBeforeModuleExportsClose`) doesn't handle — it only matches an inline object literal or a variable declared as one.

## Workarounds

1. Add to the **root** `mj.config.cjs` (the keys must sit inside the `openApps` section; top-level is ignored):

```js
openApps: {
  serverPackagePath: 'apps/MJAPI',
  clientPackagePath: 'apps/MJExplorer',
},
```

2. Rewrite `apps/MJAPI/mj.config.cjs` as an object literal so the writer can parse it:

```js
module.exports = { ...require('../../mj.config.cjs') };
```

With both in place the install completes cleanly and the app activates.

## Suggested fixes

- Detect the layout (probe `apps/MJAPI` when `packages/MJAPI` is absent), or have `mj install` write the `openApps.serverPackagePath` / `clientPackagePath` keys into the config it generates so the two commands agree out of the box.
- Support `module.exports = require(...)` / spread re-exports in the config writer, or fall back to the root config when the per-app file isn't an object literal.
- Consider validating the workspace paths in a preflight step so the install fails **before** committing schema + migrations rather than after.

## Environment

- `@memberjunction/cli` 5.49.0, MJ distribution v5.49.0, Node 24.16.0, macOS
- Reproduced on a clean database + clean `mj install` host (not an upgrade)

Related: #2975 (different `dynamicPackages` writer bug, closed).

### LINKED ISSUE #3271: mj app install writes dynamicPackages to the server workspace config, not the repo root — client bootstrap and container deploys silently never load the app
## Summary

`mj app install` writes the `dynamicPackages` block into `<serverPackagePath>/mj.config.cjs` (e.g. `apps/MJAPI/mj.config.cjs`) and never into the repo-root `mj.config.cjs`. But two of the three consumers of that block resolve config from the **repo root**, so a successful install can silently produce wiring that never takes effect.

| Consumer | Resolves config from | Sees the installer's write? |
|---|---|---|
| MJAPI started locally via `npm run --workspace=apps/MJAPI` | `apps/MJAPI/` | ✅ |
| Explorer's `mj codegen manifest --open-app-client-bootstrap` | repo root | ❌ |
| MJAPI in a container / Azure App Service deploy that ships only the root config | repo root | ❌ |

Note the installer puts the **client** array inside the **server** package's config file, which is the crux of the problem: the client build can't reach it.

## Symptom 1 — client classes never register

With the block only in `apps/MJAPI/mj.config.cjs`, the Explorer prebuild reports:

```
[class-manifest] Open App client bootstrap: 0 client packages wired
```

so no `import '<app-client-package>'` side-effect lines are emitted, no `@RegisterClass` decorators fire, and the app's UI/entity subclasses are absent from the bundle. Moving the identical block to the root config changes this to `2 client packages wired` and the imports appear. (This matters more than it looks: entity classes are typically referenced only in *type* positions, so TypeScript elides those imports — the side-effect import from the manifest is the only thing that registers them.)

## Symptom 2 — server package silently not loaded in a deployed environment

Worse in production. We hit this on an Azure App Service deployment where the pipeline ships the **root** `mj.config.cjs` only. The root config had the app's `client` entries (hand-added earlier to fix Symptom 1) but not its `server` entry, so:

- `__mj.OpenApp` reported `Status='Active'`, `Version='0.4.0'` — the install had genuinely completed
- Explorer rendered the app's UI (its client packages were bundled)
- the API loaded **no** Open App server package, so its GraphQL schema contained none of the app's types

Every read and write failed with validation errors that look like a code bug but aren't:

```
Cannot query field "mjBizAppsSonarScoreBandSet" on type "Query".
Unknown type "CreatemjBizAppsSonarScoreModelInput". Did you mean "CreateMJAIAgentModelInput"...?
```

and downstream, on the client:

```
Entity <Name> could not be instantiated via MJGlobal Class Factory.
Make sure you have registered the class reference ... ALSO make sure you call
LoadGeneratedEntities() ... tree-shaking sometimes removes subclasses
```

That last message points at tree-shaking, which sends you down the wrong path — the real cause is that the server package was never loaded.

## Why this was expensive to diagnose

Every signal a reasonable person checks reports healthy:

- the `OpenApp` row says Active with the right version
- Explorer displays that row (served by core resolvers) so the app "looks installed"
- the app's UI renders
- restarting the API changes nothing, because the deployed artifact is what's missing the wiring

The only truthful signal is one line in the API boot log (`Loaded Open App server package: …`), and its **absence** is what you need to notice.

## Suggested fixes

1. **Write `dynamicPackages` to the repo-root `mj.config.cjs`** — the one location all three consumers resolve. (Or write both, or split so each array lands where its consumer looks.)
2. **Reconcile claimed vs loaded at boot.** At startup, compare `__mj.OpenApp` rows marked Active against what the Open App loader actually loaded, and warn loudly on a mismatch:

```
⚠ OpenApp 'mj-bizapps-sonar' is Active in the database but was not loaded
  (package missing from node_modules, or no dynamicPackages entry in the config this process read)
```

That single check would have turned a multi-hour investigation into one log line, and it guards every future deployment where the DB and the shipped artifact drift apart.

## Environment

- `@memberjunction/cli` 5.49.0, MJ 5.49.0, Node 24.16.0
- Symptom 1 reproduced locally on a clean `mj install` host; Symptom 2 observed on Azure App Service

Related: #3270 (workspace-path assumption in the same `[Packages]`/`[Config]` flow), #2975 (closed, different `dynamicPackages` writer bug).

BarnattW-BC:
Closes #3270. Closes #3271.

Both issues came out of installing an Open App onto a host created by `mj install`. Together they mean `mj app install` can fail late, or report success, while leaving an app that never actually loads — and in the success case the symptoms point everywhere except the cause.

## #3270 — workspace layout

The installer resolved its workspaces at the monorepo defaults (`packages/MJAPI`, `packages/MJExplorer`), but `mj install` scaffolds a distribution under `apps/`. So installing onto a host built by MJ's own installer died at `[Packages]`:

```
✔ [Record]   App 'mj-bizapps-sonar' recorded as installed
- [Packages] Adding packages...
✖ Install failed: Could not read package.json at <root>/packages/MJAPI/package.json: ENOENT
```

That's *after* schema creation and migrations commit, so the app is left recorded with `Status='Error'` and no wiring.

Both layouts are now probed — `packages/…` first, then `apps/…` — through a new shared `install/workspace-paths.ts` used by both the package manager and the config manager. An explicit `openApps.serverPackagePath` / `clientPackagePath` still wins, so hosts with unusual layouts are unaffected.

## #3271 — config write target

Config edits went to exactly one file: the server workspace's `mj.config.cjs` when present, else the repo root. But the consumers of that config don't all resolve from the same place:

| Consumer | Resolves from |
|---|---|
| MJAPI started from its workspace (`npm run start --workspace=apps/MJAPI`) | server workspace |
| `mj codegen manifest --open-app-client-bootstrap` (MJExplorer prebuild) | repo root |
| MJAPI in a container / App Service that ships only the root config | repo root |

Note the installer writes the **client** array into the **server** workspace's config — a file the client build never reads. Both failure modes are silent:

- the client bootstrap reports `Open App client bootstrap: 0 client packages wired`, so no side-effect imports are emitted and the app's `@RegisterClass` decorators never fire (its entity subclasses aren't registered, which surfaces as a misleading "tree-shaking removed subclasses" class-factory error);
- a deployed API loads no server package at all, so its GraphQL schema contains none of the app's types — `Cannot query field …` / `Unknown type …Input` on every read and write — while `__mj.OpenApp` still cheerfully reports the app `Active`.

That last combination is genuinely expensive to diagnose: the DB says installed, the UI renders, and restarting the API changes nothing because the deployed artifact is what's missing the wiring.

All config writes — `dynamicPackages`, `entityPackageName`, `excludeSchemas` — now target **every** `mj.config.cjs` a consumer may load. Entry insertion was already idempotent, so a config that already has the entry is left byte-identical, and writing the same resolved path twice is deduped.

## Partial success, deliberately

A config file can legitimately be un-editable by this module's string surgery — most commonly the distribution's re-export shape:

```js
module.exports = require('../../mj.config.cjs');
```

There's no object literal to insert into, which is the second failure in #3270 (`Could not find a config object in mj.config.cjs to insert into`). That file needs no edit — it re-exports the root config, which *is* in the target list and *does* get written. So a per-file failure is now collected as a warning rather than failing the install; the install fails only when **no** config could be updated. `ConfigOperationResult` gains an optional `Warnings` field and the orchestrator surfaces them through `OnWarn`, so a skipped file is visible rather than silent.

## End-to-end validation (real distribution host)

Run against a live `mj install` v5.49.0 distribution host — the #3270 repro — with **no** `openApps` config, `apps/MJAPI/mj.config.cjs` left as the delegating re-export the distribution actually ships, Sonar deps stripped from both workspaces, and a fresh database:

- `mj app install … v0.4.0` → **succeeded** (previously died at `[Packages]` with `ENOENT: <root>/packages/MJAPI/package.json`)
- root `mj.config.cjs` received `dynamicPackages` incl. `@mj-biz-apps/sonar-server` + `LoadSonarServer`
- `apps/MJAPI/mj.config.cjs` correctly left untouched
- both workspace `package.json`s and `apps/MJExplorer/angular.json` updated
- **zero** "skipped a config file" warnings
- `mj codegen` → 394 entities; `__mj.OpenApp` = `Active`, 14 entities / 15 tables in `__mj_BizAppsSonar`
- API booted: `Loaded Open App server package: @mj-biz-apps/sonar-server (ran LoadSonarServer) (+1 resolver path)` … `Ready http://localhost:4995/`
- emitted schema carries **196** `mjBizAppsSonar*` types, including the two #3271 reported missing: `input CreatemjBizAppsSonarScoreModelInput` and `mjBizAppsSonarScoreBandSet(ID: String!)`

## Tests

5 new cases in `config-manager-dynamic.test.ts`:

- writes the entry to both the server workspace config and the repo root
- doesn't write the same file twice when the server path *is* the repo root
- succeeds with a warning when one config is a bare re-export, writing only the editable one
- fails only when no config could be updated (and leaves every file untouched)
- falls back to the `apps/` layout when `packages/MJAPI` is absent

Full suite: **298 passing**. The one pre-existing assertion that changed is `writes exactly once when the resulting config is valid` → `writes exactly once per resolvable config`, which is the intended contract change (its mock makes every probe exist, so both targets resolve).

## Review follow-ups (applied)

- **Root-config write failure is now fatal.** `ApplyToConfigs` previously treated all targets as fungible, so a root failure could degrade to a warning while the server write succeeded — reporting success while silently re-creating #3271, since root is the only config the client bootstrap and container deploys ever read. Root failures now abort; non-root failures remain warnings.
- **The delegating re-export shape is a silent no-op.** `packages/MJAPI/mj.config.cjs` is `module.exports = require('../../mj.config.cjs')` in the monorepo *and* in a distribution, so detection finds it on every install. It needs no edit — it re-exports root, which is written — and treating it as a "skipped config" emitted an alarming warning on every monorepo install. `IsDelegatingConfig` now recognises the shape and skips it quietly; a genuinely unsupported non-root config still warns.
- **`AngularConfigManager` uses the shared layout detection.** It hardcoded `packages/MJExplorer`, so on an unconfigured `apps/` host `Load()` returned false and the prebundle-exclude step was silently skipped — the same class of bug this PR fixes, one function over. Fixed at the source so both call sites benefit.
- **Per-operation error labels restored.** Funnelling six writers through one helper had collapsed their specific prefixes into a generic `Failed to update config`. `ApplyToConfigs` now takes an operation label, so failures read `Failed to update excludeSchemas in <path>: …`.

## Known remaining hazard (not fixed here)

#3270 also suggested a **preflight** that validates workspace paths before any DDL runs. This PR removes the common cause of the late failure but does not change the ordering: a genuinely custom layout that detection cannot resolve will still fail at `[Packages]` *after* schema creation and migrations have committed, leaving the app at `Status='Error'`. Making path validation a preflight step is a separate, slightly invasive change to the orchestrator's phase order and is deliberately out of scope — worth its own issue so the commits-then-fails hazard stays tracked.

## Not addressed here

The diagnostic gap that made #3271 so costly is worth its own change: nothing reconciles what `__mj.OpenApp` claims is Active against what the loader actually loaded at boot. A startup warning along the lines of

```
⚠ OpenApp 'x' is Active in the database but was not loaded
  (package missing from node_modules, or no dynamicPackages entry in the config this process read)
```

would turn this class of problem into one log line. Happy to follow up if you'd like it in the same release.


rkihm-BC:
## Review — approve with minor changes

Thorough pass over the diff plus both issues and the code paths the fix depends on. I verified the central claims against the codebase: the client bootstrap (`mj codegen manifest --open-app-client-bootstrap`) uses `cosmiconfigSync('mj', { searchStrategy: 'global' })` from the client workspace's cwd, and MJExplorer intentionally ships **no** own `mj.config.cjs`, so cosmiconfig walks up to root — confirming that writing to root does reach that consumer. The re-export shape (`module.exports = require('../../mj.config.cjs')`) is real and is exactly what `packages/MJAPI/mj.config.cjs` is even in the monorepo.

### What's good
- **Excellent diagnosis** of two genuinely expensive silent-failure bugs (DB says Active, UI renders, restart doesn't help).
- **Right architecture** — one shared layout-detection module consumed by both package-manager and config-manager.
- **Idempotency is real, not asserted** — `AddEntryToDynamicArray` guards on `PackageName`+`AppName`, and `WriteConfigChecked` still `compileFunction`-validates every edit before writing, so the multi-write path can't ship a broken config (the #2975 safety net survives).
- **Explicit config still wins** (`configured ?? detect`), so custom-layout hosts are unaffected.
- **Tests are well-targeted** — both-target write, root/server dedupe, partial-success-with-warning, total failure, and `apps/` fallback; the changed `writes exactly once` → `once per resolvable config` assertion is a legitimate, explained contract change.

### Asks before merge

1. **Don't let a root-config write failure degrade to a warning (moderate).** `ApplyToConfigs` treats all targets as fungible — any per-file failure is a warning as long as one file wrote. That's correct for the re-export case (server file fails, root succeeds). But it's asymmetric: if the **root** write were the one to fail while the server workspace succeeded, the install would report success-with-warning while re-introducing the exact silent #3271 failure (the client bootstrap and container deploys read root and would get nothing). In practice root is a plain object literal and edits fine, so this is largely theoretical — but the abstraction has no notion that root is load-bearing for two of the three consumers. Please treat a root-config write failure as fatal (or at minimum comment the asymmetry).

2. **Suppress the new benign warning on every monorepo install (minor, user-visible).** Previously, with `serverPackagePath` unconfigured, only root was touched. Now detection finds `packages/MJAPI` unconditionally, and since its config is a re-export, every monorepo `mj app install` will emit `Skipped a config file: …/packages/MJAPI/mj.config.cjs: Could not find a config object …`. It's harmless but reads like something is wrong. Consider recognizing the expected re-export shape as a correct no-op rather than surfacing it as a skipped file.

### Non-blocking (fix or explicitly acknowledge)

3. **Layout detection isn't applied uniformly.** `HandleAngularPrebundleExcludes` still builds `new AngularConfigManager(context.RepoRoot, context.ClientPackagePath)` with `ClientPackagePath = undefined` on an unconfigured `apps/` host, and `AngularConfigManager` falls back to the hardcoded `packages/MJExplorer`. On a distribution host `manager.Load()` won't find `packages/MJExplorer/angular.json`, so the prebundle-exclude step is silently skipped. Per the existing comment this is non-fatal (dev-server singleton bug only; prod builds fine), so it doesn't block — but it's the same class of bug the PR fixes, left one function over. A `ResolveClientPackagePath(...)` here would make it consistent. (I confirmed `RegenerateClientBootstrap`, the other hardcoded-path spot, is not invoked in the live install flow, so it's dormant.)

4. **#3270's preflight suggestion is unaddressed and unmentioned.** #3270 explicitly asked to validate workspace paths in a preflight so the install fails *before* committing schema + migrations. This PR fixes the common case so it won't fail — but a genuinely broken/custom layout that detection can't resolve will still fail late at `[Packages]`, leaving `Status='Error'` after migrations commit. The "commits-then-fails" ordering hazard remains. The "Not addressed here" section only defers the #3271 boot reconciliation, not this — worth a sentence acknowledging it.

5. **Minor:** the per-operation error prefixes (`Failed to update excludeSchemas config`, `Failed to remove entityPackageName mapping`, etc.) all collapse into the generic `Failed to update config: …` from `ApplyToConfigs` — a small loss of diagnostic signal given how much this PR otherwise cares about diagnosability.

### Question

Was this validated end-to-end against a real `mj install` distribution host (the repro in #3270), or only via the unit tests? The unit tests mock `fs`, so they prove the path-resolution logic but not that a live `apps/` install now completes and the app actually loads and serves its GraphQL types — which is the whole point of the fix. A quick confirmation of a real install → app-loads run would close the loop.

BarnattW-BC:
Thanks — this was a genuinely useful review, and both "asks" were real rather than theoretical. All four are addressed in `0bfc2a6` (now merged with `next` at `6d67a5b`), and I closed the end-to-end gap you asked about.

## Ask 1 — root-config write failure is now fatal

You were right that the abstraction had no notion of root being load-bearing. `ApplyToConfigs` now aborts on a root write failure even when another target succeeded, since root is the only config the client bootstrap manifest and container deploys ever read — degrading it to a warning would report success while silently re-creating #3271. Non-root failures remain warnings. New test: `treats a ROOT-config write failure as FATAL even when the server config succeeded`.

## Ask 2 — the benign warning is gone (and you were right that it fires on every monorepo install)

Confirmed in the monorepo: `packages/MJAPI/mj.config.cjs` is literally `module.exports = require('../../mj.config.cjs');`, so detection finds it unconditionally and every install would have emitted that skip. Added `IsDelegatingConfig`, which recognises the re-export shape as a correct no-op and stays silent. A genuinely unsupported non-root config still warns — split into two tests so the distinction is pinned:

- `silently no-ops a delegating config — no warning (#3270 re-export shape)` (asserts `Warnings` is `undefined`)
- `warns (but succeeds) when a NON-root config is unsupported for another reason`

Worth noting the spread form `module.exports = { ...require(...) }` deliberately does **not** match — it has an object literal, so inserting into it is correct.

## Non-blocking 1 — `AngularConfigManager` now uses the shared detection

Fixed at the source (`angular-config-manager.ts`) rather than at the two call sites, so both — and any future one — benefit. Agreed it was the same bug one function over. I left `RegenerateClientBootstrap` alone given your confirmation that it's dormant in the live install flow.

## Non-blocking 3 — per-operation error labels restored

`ApplyToConfigs` takes an operation label, so failures read `Failed to update excludeSchemas in <path>: …` instead of the generic `Failed to update config`. Test asserts the operation-specific message.

## Non-blocking 2 — preflight hazard acknowledged, not fixed

Added a "Known remaining hazard" section to the PR description. This PR removes the common *cause* of the late failure but doesn't change the ordering: a custom layout detection can't resolve will still fail at `[Packages]` after migrations commit, leaving `Status='Error'`. Reordering path validation into a preflight is a more invasive change to the orchestrator's phase sequence and I'd rather it be its own issue than smuggle it in here. Happy to file it.

## Your question — yes, validated live now

Fair challenge; it had only been unit-tested. Re-ran on a real `mj install` v5.49.0 distribution host with **no** `openApps` config, `apps/MJAPI/mj.config.cjs` restored to the delegating re-export, Sonar deps stripped from both workspaces, and a fresh database — i.e. the #3270 repro, not a host I'd already hand-patched:

```
✔ [Install] Successfully installed mj-bizapps-sonar v0.4.0        (was: ENOENT packages/MJAPI/package.json)
Loaded Open App server package: @mj-biz-apps/sonar-server (ran LoadSonarServer) (+1 resolver path)
Ready     http://localhost:4995/
```

Root config received `dynamicPackages` + `LoadSonarServer`; `apps/MJAPI/mj.config.cjs` untouched; both workspace `package.json`s and `angular.json` updated; **zero** skip warnings; `__mj.OpenApp` = `Active` with 14 entities / 15 tables. The emitted schema carries 196 `mjBizAppsSonar*` types — including `CreatemjBizAppsSonarScoreModelInput` and `mjBizAppsSonarScoreBandSet(ID: String!)`, the exact two #3271 reported missing. Full details in the PR description.

Tests: **300 passing** against merged `next`.

## On #3290

Agreed it should merge first — it carries the RCE fix and gates the bizapps manifest PRs. The overlap is narrower than the warning suggests: #3290 rewrites the low-level string builders to `JSON.stringify` manifest values, while this PR restructures the high-level exported functions and adds `resolveConfigPaths`/`ApplyToConfigs`. Different regions, and complementary — my multi-target write calls those same helpers, so the escaping applies to every config file written. I'll rebase once it lands and re-run both the suite and the live install.

PR #3219, opened 2026-07-20
Feat/explicit vector ID strategy + minimal-metadata mode for entity vectorization

dray-mc:
## Summary
Metadata is particularly expensive (storage cost and performance) in the context of vector databases and we want to be able to control what gets set very carefully to make sure there's no unintended overhead. The defaults adds some metadata fields that make sense for the base MJ use cases (the app search) but do not for Betty use cases and others - ie. including an Entity name when our primary use is all ContentItem records. Didn't want to break anything already in place, so didn't adjust what the 'include' config strategy does, but added 'explicit'. This also supports fields that were previously filtered out by a higher level validation (uniqueidentifiers specifically were needed). Also for our specific case, since we're only using one table and it always has a uniqueidentifer as its ID, it is convenient to just use that ID as the vector's ID *and* allows us to eliminate another piece of metadata so included that as an option while leaving the existing hash strategy as the default.

- Add `vectorIdStrategy` to `EntityDocumentConfiguration`: `'hash'` (default, unchanged SHA-1 behavior) or `'recordId'`, which uses the source record's primary key directly as the vector database ID — UUIDs normalized to lowercase for SQL Server/PostgreSQL portability, composite PKs joined with `||`, hard failure (not a silent fallback) on empty keys or IDs over the 512-byte provider limit.
- Add `fieldStrategy: 'explicit'`: vector metadata contains *exactly* the configured fields — no system-injected `RecordID`/`Entity`/`TemplateID`, and `includeEntityIcon`/`includeUpdatedAt` flip to opt-in. All existing strategies (`'all'`/`'include'`/`'exclude'`) are byte-for-byte unchanged.
- Fix: explicitly-included fields (`included: true`) under `'include'`/`'explicit'` are now honored even when they're a uniqueidentifier, PK, or `__mj_*` field — previously a type-eligibility filter silently dropped them before the strategy ever saw them. Only genuinely unstorable binary types are refused, now with a logged warning instead of a silent drop. Uniqueidentifier metadata values are normalized to lowercase for cross-platform filter consistency.
- `VectorSearchProvider` (SearchEngine): when a match's metadata omits `Entity` (e.g. an `'explicit'`-mode index), resolve a fallback entity name from the index's Entity Documents instead of labeling results `Unknown`. The lookup is skipped entirely when metadata already has `Entity` (the case for every pre-existing setup) and is cached per index for the life of the `SearchEngine` singleton.
- No schema/migration changes — both packages ship as `patch`.

## Why
[...]

PR #3175, opened 2026-07-16
Feat/optional embedding dimensions + Pinecone namespace routing

dray-mc:
## Summary

Several changes needed to allow embeddings to be created, stored, and queried the way Betty relies on. Dimensions parameter existed but wasn't respected, Betty team has seen success in controlling vector size while using a larger, more advanced model by throwing text-embedding-3-large from OpenAI in but limiting vector size to 1024. Namespaces are regularly used in Pinecone not just to segment data for security reasons but to drastically reduce costs and improve performance, generally aligned with Organization in this use case but framework needed to be flexible enough to support other choices.

All changes here were made without new schema requirements so no migration required.

Will be some more changes related to this process coming soon!

- Add optional `dimensions` parameter to embedding types (`EmbedTextParams` / `EmbedTextsParams` / `EmbedContentParams`) and wire it through OpenAI embeddings, the EntityVectorSyncer pipeline, and `VectorSearchProvider` — sourced from `MJ: Vector Indexes.Dimensions`. Models that don't support it ignore the field; models that do (e.g. `text-embedding-3-*`) use it to produce reduced-dimension vectors.

- Add Pinecone namespace routing for both ingestion and search. Ingestion: `VectorDBBase.buildProviderDirectives()` virtual method lets providers extract per-record routing values from the source row; Pinecone overrides it to read `VectorIndex.ProviderConfig.namespaceField`, look up that field's value on the record, and return `{ namespace: '<value>' }` so `CreateRecords` groups upserts by namespace. Search: `VectorSearchProvider` passes the scope's rendered `ExternalIndexConfig` as `providerConfig` to `QueryIndex`; Pinecone reads `providerConfig.namespace` and calls `index.namespace(ns)` before querying. Includes a fix to `SearchEngine.buildScopeConstraints` to Nunjucks-render `ExternalIndexConfig` before parsing (matching how `MetadataFilter` was already handled), enabling dynamic values like `{"namespace": "{{ context.PrimaryScopeRecordID }}"}`.

- Add `providerConfig` to `QueryParamsBase` and `providerDirectives` to `VectorRecord` as typed extension points for future providers.

## Changed packages

| Package | Change |
|---|---|
| `@memberjunction/ai` | `dimensions?` on all embed param types |
| `@memberjunction/ai-openai` | Pass `dimensions` to OpenAI embeddings API |
[...]

PR #3275, opened 2026-07-25
feat(content-autotag): metadata-driven vector config, chunk lifecycle…

topherHunnewell:
Summary

Brings the ContentSource / autotag vectorization pipeline (AutotagBaseEngine) to parity with the EntityDocument pipeline (from #3175 / #3219), adds the missing chunk lifecycle operations, and ships a conforming integration check bundle that exercises the whole thing end-to-end.

Everything here is additive and opt-in — existing content sources behave identically unless they opt into the new config — and there is no schema migration: the configuration rides the existing Configuration JSONType on ContentSource / ContentType.

The headline for our app: a scoped-search hit on chunked content now returns the matched ContentItemChunk id (not just the parent ContentItem id), so external hydration can fetch the exact passage.

What changed

- Metadata-driven vector config on the Configuration JSONType of both ContentSource and ContentType (ContentSource overrides ContentType, then a hardcoded default):
  - VectorIDStrategy ('hash' | 'recordId', default 'recordId') — 'recordId' uses each chunk's own id as its vector-DB id (purge-safe); 'hash' is 5.49 EntityDocument parity and unsafe with re-chunk + purge (documented on the field).
  - ChunkTextStorage ('mixed' | 'alwaysChunk', defaysChunk' writes a ContentItemChunk row for every item and leaves ContentItem.VectorRecordID null; 'mixed' keeps single-chunk items' text/vector on the ContentItem.
  - VectorMetadata — full structural parity with the entity pipeline's metadata control: FieldStrategy: 'all' | 'include' | 'exclude' | 'explicit' (unset ⇒ the curated content default), per-field Fields overrides (Included / TruncationLimit / StoreAs),
DefaultTruncationLimit, and IncludeEntityIcon / Incs / IncludeText toggles. The runner mirrors theentity side's decomposition (system / icon / updatedAt / display-field helpers, StoreAs coercion, UUID normalization, truncation). Content-specific deviations: Entity is always kept under 'explicit' (results stay labeled — record id recovers from the vector id under recordId), and Tags is a toggle rather than a discovered field.
- Chunk-Identity Contract — chunk vectors carry their own identity: Entity='MJ: Content Item Chunks', RecordID=<ContentItemChunk.ID>, ContentItemID, Sequence. The chunk row PK is minted up front and used as its identity (and, under recordId, its vector id), so scoped search returns the matched chunk id with no search-side changes. Item-level (mixed single-chunk) vectors keep MJ: Content Items identity.
- AutotagBaseEngine.EmbedPendingChunks(user, { maxItems }) — (re)embeds persisted ContentItemChunk rows whose EmbeddingStatus='Pending', for migration backfill and error recovery. Bounded + rate-limited; best-effort per chunk.
- AutotagBaseEngine.PurgeDeletedChunks is now triggerable — the Autotag/Vectorize action gains optional Purge (Phase 4) and
EmbedPendingChunks (Phase 3) params, both independeded by MaxItems, both best-effort.
- Embedding dimensions — the resolved infrastructures.Dimensions and threads it into the embedding call(new optional Dimensions on AIModelRunner's EmbeddingRunParams, forwarded to EmbedTexts), so reduced-dimension indexes work in the
autotag path and the dedup-check query embeds at th
- Provider routing — the resolved infrastructure carries the parsed VectorIndex.ProviderConfig; per-record providerTemporaryDirectives are built via VectorDBBase.BuildProviderDirectives (e.g. Pinecone namespace from a configured source field) and providerConfig is passed to CreateRecords. Only invoked when the index has a ProviderConfig.
- Integration coverage — a new conforming content-vectorization check bundle (IT67, CV1–CV6) in
@memberjunction/integration-test-suite exercises the live DB, stubbing only the embedding + vector-DBseams the deterministic tier must not call.

Changed packages

┌─────────────────────────────────────────┬──────────────┬────────────────────────────────────────────────────────────────────┐
│                 Package                 │     Bump     │                               Change                               │
├─────────────────────────────────────────┼──────────────┼────────────────────────────────────────────────────────────────────┤
│                                         │              │ Config resolution, chunk-identity metadata, EmbedPendingChunks,    │
│ @memberjunction/content-autotagging     │ minor        │ dimensions + provider-routing threading, decomposed metadata       │
│                                         │              │ runner                                                             │
├─────────────────────────────────────────┼──────────────┼────────────────────────────────────────────────────────────────────┤
│ @memberjunction/actions-content-autotag │ minor        │ Purge + EmbedPendingChunks action params (Phases 3/4)              │
├─────────────────────────────────────────┼──────────────┼────────────────────────────────────────────────────────────────────┤
│ @memberjunction/core-entities           │ minor        │ Generated Configuration JSONType types for the new config          │
├─────────────────────────────────────────┼──────────────┼────────────────────────────────────────────────────────────────────┤
│ @memberjunction/ai-prompts              │ minor        │ Optional Dimensions on AIModelRunner.EmbeddingRunParams            │
├─────────────────────────────────────────┼──────────────┼────────────────────────────────────────────────────────────────────┤
│ @memberjunction/integration-test-suite  │ private (no  │ New content-vectorization bundle (IT67, CV1–CV6) + metadata        │
│                                         │ bump)  in                                               │
├─────────────────────────────────────────┼──────────────┼────────────────────────────────────────────────────────────────────┤
│ @memberjunction/search-engine           │ test-only    │ Regression test pinning the chunk-id read path in                  │
│                                         │ (no bumer                                               │
└─────────────────────────────────────────┴──────────────┴────────────────────────────────────────────────────────────────────┘

Behavior note

The default ChunkTextStorage='alwaysChunk' + VectorIDStrategy='recordId' means newly-embedded single-chunk items now get a ContentItemChunk row with a unique vector id instead of an item-level hash id. Already-embedded items are not reprocessed, so
existing data is untouched until re-embedded; set Cr source to retain the item-level single-chunk
behavior. No schema/migration changes.

Testing

- [x] Unit tests — 1,281 across the touched package
- [x] Live end-to-end (dev DB, embedding + Pinecone stubbed & captured) — a standalone rig created + inspected real ContentItemChunk rows across 7 scenarios (default identity, multi-chunk, re-vectorize soft-delete, purge, embed-pending, explicit metadata, dimensions + namespace) — 34/34 checks passed.
- [x] Integration tier (mj test) — the new IT67 con(CV1–CV6) dispatches live through
IntegrationTestDriver and passes 6/6; conforms to t check-registry gates green).

To Verify

- The default behavior change (single-chunk items ndId) — confirm it's acceptable, or that deploymentswill set ChunkTextStorage='mixed'. Already-embedded rows are not reprocessed.
- 'hash' caveat — intentionally unsafe with re-chunk + purge; only recordId (default) is purge-safe.
- Namespace routing needs app schema — providerConfig.namespaceField reads a real field off the source row; this PR wires the
framework mechanism, the field itself is downstream
- Generated entity_subclasses.ts — the diff is the  only; the auto-merge with next's CodeGen output wasverified by a clean core-entities build.

🤖 Generated with Claude Code (https://claude.com/c


packages/ContentAutotagging/src/Engine/generic/AutotagBaseEngine.ts
```diff
@@ -43,6 +44,71 @@ interface ResolvedVectorInfrastructure {
...
+ * Vector-storage behavior resolved for a single content item, derived from its ContentSource
+ * configuration (falling back to its ContentType defaults, then hardcoded defaults). Union types
+ * are derived from the generated JSONType interface so they track the metadata definition — see
+ * {@link resolveItemVectorStorageConfig}.
+ */
+type VectorIDStrategy = NonNullable<MJContentSourceEntity_IContentSourceConfiguration['VectorIDStrategy']>;
+type ChunkTextStorage = NonNullable<MJContentSourceEntity_IContentSourceConfiguration['ChunkTextStorage']>;
+type VectorMetadataConfig = NonNullable<MJContentSourceEntity_IContentSourceConfiguration['VectorMetadata']>;
+type VectorMetadataFieldConfig = NonNullable<VectorMetadataConfig['Fields']>[string];
+interface ResolvedVectorStorageConfig {
```
AN-BC, line 71: would this interface and ones below be good to export from the package for other future consumers potentially? And should have good JSDoc on them

packages/ContentAutotagging/src/Engine/generic/AutotagBaseEngine.ts
```diff
@@ -1613,53 +1694,106 @@ export class AutotagBaseEngine extends BaseEngine<AutotagBaseEngine> {
...
-                ? chunk.chunkId
-                : this.contentItemVectorId(chunk.item.ID),
-            values: vectors[idx],
-            metadata: this.buildVectorMetadata(chunk.item, tagMap.get(chunk.item.ID))
-        }));
+        // The ContentItem entity metadata drives config-selected display fields (types, MaxLength,
+        // eligibility). Resolved once per batch — every chunk shares the same source entity.
+        const contentItemEntity = this.ProviderToUse.EntityByName('MJ: Content Items');
+        return allChunks.map((chunk, idx) => {
+            const count = countByItem.get(chunk.item.ID) ?? 1;
```
AN-BC, line 1731: might be nice to break this out into its own method and make this method and the one we call from inside the `map` as `protected` with solid JSDoc so sub-classes could override if they want behavior mods?

packages/ContentAutotagging/src/Engine/generic/AutotagBaseEngine.ts
```diff
@@ -2209,6 +2566,46 @@ export class AutotagBaseEngine extends BaseEngine<AutotagBaseEngine> {
...
+        return crypto.createHash('sha1').update(`content-item-chunk_${contentItemId}_${chunkIndex}`).digest('hex');
+    }
+
+    /**
+     * Resolve the vector-storage behavior for a content item using the cascade:
+     * ContentSource override -> ContentType default -> hardcoded default. Reads the strongly-typed
+     * ConfigurationObject accessors emitted by CodeGen for each entity's Configuration JSONType.
+     */
+    private resolveItemVectorStorageConfig(item: MJContentItemEntity): ResolvedVectorStorageConfig {
+        const source = this.khEngine.ContentSources.find(s => UUIDsEqual(s.ID, item.ContentSourceID));
```
AN-BC, line 2585: while this might not be a huge deal relative to the vector db upsert or the embedding model, I would think that we could do a better job of building `O(1)` lookup by building a cache/hashmap of the contentsources by ID (lazy load first time we try to get a content source) -- **inside** the khEngine, so that we have a nice `FindContentSourceByID` method that is `O(1)` after the first hit? And other similar lookups. If this is called thousands of times no need to waste CPU cycles, eh?

packages/MJCoreEntities/src/generated/entity_subclasses.ts
```diff
@@ -67538,6 +67621,62 @@ export interface MJContentTypeEntity_IContentTypeConfiguration {
...
+     * metadata is kept.
+     */
+    VectorMetadata?: MJContentTypeEntity_IContentTypeVectorMetadataConfig;
+}
+
+/**
+ * Content-type-level default for vector metadata shape. Structurally identical to
+ * {@link IContentSourceConfiguration.VectorMetadata}; a source's own setting overrides it.
+ */
+export interface MJContentTypeEntity_IContentTypeVectorMetadataConfig {
```
AN-BC, line 67649: very nice

AN-BC (commented):
LGTM overall @topherHunnewell , very minor comments in here that are more perf optimization in one area and about future flexibility in another. After you make those changes and run tests again to ensure no breakgae, you can merge

AN-BC:
Cross-linking: **#3286** adds `@memberjunction/ai-segmentation`, which is the strategy layer that will eventually feed this pipeline's chunking. Flagging it here so we sequence cleanly rather than collide.

**#3286 deliberately does not touch `AutotagBaseEngine.ts`** — this PR rewrites it (+653/−67), and a competing edit in the same functions would be a painful merge for no benefit. It ships the framework standalone: `BaseSegmenter` + four strategies (`StructuralText`, `SemanticText`, `Transcript`, `FixedWindow`), resolved through the class factory the same way `BaseEmbeddings`/`VectorDBBase` providers are.

Checking the overlap concretely against this PR's diff:

| Function | In #3275's diff? | Integration risk |
|---|---|---|
| `chunkExtractedText` (tagging path) | **0 occurrences** | None — safe to wire independently |
| `buildEmbeddingChunks` | touched | Wait for this PR to land |
| `buildChunksForBatch` | touched | Wait for this PR to land |

So the tagging path can be wired now without touching anything here, and the embedding path waits for you.

**One thing worth confirming:** this PR adds no migration (config rides the existing `Configuration` JSONType — makes sense). That means the chunk columns needed for multimodal — `Modality`, `StartMs`/`EndMs`, `Description`, `Transcript`, `ParentChunkID`, and relaxing `Text` to nullable — are a **separate migration that doesn't exist yet**. We'd author it on top of this PR rather than expecting it here.

Two findings from #3286 that touch your area, both worth a look:

1. **`TextChunker` offsets were wrong for repeated text** — `indexOf` from position 0 meant any recurring sentence made later chunks report the *first* occurrence (a chunk truly spanning 61–86 reported 0–86). Fixed in #3286. Relevant because chunk offsets are provenance, and `Content Item Chunks` is where they'd be persisted.
2. **`DetectVectorDuplicates` is a second `EmbedTexts` caller** (`AutotagBaseEngine.ts:2474`) — easy to miss when changing how items are embedded.

No action needed on your side; just flagging the sequencing. Happy to rebase #3286's follow-up work onto this whenever it lands.

---

AN-BC:
@topherHunnewell — plan is to get this into `next` now so downstream work can build on it, rather than holding the whole PR for the two minor items.

Could you take the [review feedback](https://github.com/MemberJunction/MJ/pull/3275#pullrequestreview-4778136335) (the perf optimization and the future-flexibility point) into a **small follow-up PR** instead of another round here? Nothing in it is blocking, and #3286 plus the chunk-schema migration are queued behind this landing.

For context on what's queued: #3286 adds `@memberjunction/ai-segmentation` (pluggable chunking strategies) and deliberately doesn't touch `AutotagBaseEngine.ts` to stay out of your way. Once this merges, the sequence is chunk-schema migration → wire `buildEmbeddingChunks` to the segmenter → `EmbedContent` for multimodal.

---

PR #3276, opened 2026-07-25
fix(pg-migrations): reseed v5.45 metadata and enforce delete parity

### LINKED ISSUE #3253: v5.45 Metadata_Sync.pg.sql shipped as a 126-byte marker — PG deployments through v5.45 are missing that release's curated metadata

SDesai-BC:
`V202607071019__v5.45.x__Metadata_Sync.pg.sql` shipped as a **126-byte marker** in place of that release's 12,041 lines of metadata DML. Every PostgreSQL database that migrated through v5.45 is missing that release's curated metadata, and so is every fresh install — the v5.46 PG baseline was dumped from a database that had itself migrated through the marker, so it inherited the hole (**0 of 161** v5.45-created rows present, against 261/261 for v5.44). No check caught it, because **empty SQL applies cleanly**.

Investigating the fix surfaced a second, wider defect: the legacy converter — the path the release runbook *mandates* for `*_Metadata_Sync.sql` — silently dropped every mj-sync record deletion. Not one of **196 deletions across 10 releases (v5.9 through v5.45)** ever reached a committed PG counterpart. v5.45 is not where that started; it is where it became visible.

Closes #3253.

## The converter defect

mj-sync emits a deletion as a bare sp call with no `DECLARE` block:

```sql
EXEC [${flyway:defaultSchema}].[spDeleteComponentRegistry] @ID = '…';
```

`StatementClassifier` fell through to the generic bare-`EXEC` skip, so the statement vanished into an anonymous "skipped" count while the run exited 0. Two changes fix it. `StatementClassifier` now routes mj-sync's delete signature — a single `@ID = '<uuid literal>'` argument — to `EXEC_BLOCK`; the signature requirement is what keeps CodeGen's maintenance procs (`spDeleteUnneededEntityFields`, `spUpdateEntityFieldRelatedEntityNameFieldMap`) skipping, since the name prefix alone cannot discriminate them.

That alone recovered only **16 of 196**. The rest are emitted by an older mj-sync that wrote an entire session as one `GO`-less batch, where the delete trailed a save and was swallowed by the preceding block — again with no `SKIPPED` trace, 0 skipped, 0 errors. `ExecBlockRule.splitIntoBlocks` now starts a new block at a line-leading `EXEC` once the current block already holds a complete one. **196 of 196**, zero errors.

The `ExecBlockRule` guard also changes shape. On `next` it is:

```ts
if (!exec || declareVars.length === 0) {
```

A DECLARE-less block was unconvertible by definition, which is exactly what dropped the deletes. It now converts when the block needs no variables, and still skips **visibly** when it would otherwise assign to variables it never declared — a block whose `DECLARE` the parser cannot read (a dotted user type, say) yields zero declared vars, and emitting its `SET`s anyway produces PL/pgSQL that dies at apply time with `"p_x" is not a known variable`. A visible skip is the honest output.

**This is additive, and that is checked rather than asserted.** Both converter builds were run over all 49 metadata syncs and every non-delete `DO $mj$` block body compared element-wise: **byte-identical**. The only difference in the corpus is 196 deletes appearing.

## The forward fix

`V202607241200__v5.50.x__Reseed_v545_Metadata.pg-only.sql` — 7,153 lines, 175 `DO $mj$` blocks. Committed `.pg.sql` files are Flyway-checksummed and immutable, so the marker and the baseline are healed **forward**, never rewritten. Rationale and the three rejected alternatives: `plans/adr/0001-forward-dated-reseed-for-ledger-gaps.md`.

It is idempotent because it runs on every database, gapped or whole: **161 creates** each guarded `IF EXISTS (… WHERE "ID" = …) THEN RETURN`, **13 updates** replayed unconditionally (idempotent by value), **1 delete** guarded by `IF EXISTS`.

The non-obvious part is the **20 updates deliberately absent**. The reseed is timestamped after v5.49's sync, so replaying a v5.45 update whose row a later release re-updated full-row would *revert* newer state. Those 20 are dropped, each with an asserted safety condition: the later sync's field set must be a superset of v5.45's, or generation fails.

The 7,153 lines are generated, not hand-written — `scripts/generate-v545-metadata-reseed.mjs` legacy-converts the SQL Server source and post-processes it, and **reproduces the committed file byte-for-byte**. Review the script, not the SQL.

## Enforcement — issue ask (3)

Ask (3) was for tooling rather than "remembered by whoever runs the release," so `check-pg-migration-content.mjs` now counts mj-sync deletions on both sides of every counterpart pair and fails on disagreement. It is deliberately separate from the existing `classify` check: that one is a magnitude detector, and a sync with 194 saves that loses its 1 deletion still scores hundreds of statements on both sides. That is precisely how 196 deletions went unnoticed — a missing statement moves the file size by ~90 bytes.

The 10 historical gaps are grandfathered, with a warning if any ever stops being a gap. **A new release must have parity**; an entry added to silence a failure means the converter dropped a statement.

The generator gained the assertion its update path always had. PostgreSQL's generated `spUpdateX` silently no-ops on a missing row (`GET DIAGNOSTICS ROW_COUNT = 0 → RETURN`), so on a gapped database every v5.46–v5.49 update aimed at a v5.45-created row already did nothing, without a trace. Creating that row now from v5.45's values would make the loss permanent and invisible. Currently **zero of the 161** creates are affected — which is why it needed asserting rather than assuming, since nothing else would notice if it changed.

## Evidence

| Check | Result |
|---|---|
| SQLConverter suite | **1099 passed**, 3 skipped, 0 failed |
| Gate-script suite (`.github/scripts`) | **111 passed** |
| Converter over all 49 metadata syncs | 196 source deletions → **196 converted**, 0 errors |
| Non-delete `DO` block bodies vs `next` | **byte-identical** across all 49 files |
| Reseed regenerated from the committed script | **byte-for-byte identical** to the committed file |
| `check-pg-migration-content.mjs` | 193 counterparts, 0 suspect; delete parity 10 pairs, 0 mismatched |
| Delete-parity gate vs a synthetic new gap | **exits 1**, names the file |
| Create-supersession assertion vs an injected later sync | **exits 1**, names the row and superseding release |

Applied to a live `postgres:16` covering both gapped populations — migrate-through from the v5.38 baseline, and a fresh install from the v5.46 baseline. Pre-reseed all 161 rows **absent**; post-reseed all 161 **present**; the superseded `TemplateParam` retained its **v5.49** value rather than reverting; the `ComponentRegistry` row deleted. **Double-apply**: second run completes with zero errors and zero state changes.

## Risk and ordering

**Do not renumber the reseed earlier.** `V202607241200` is deliberately after v5.49's sync (`V202607221340`) — the supersession analysis assumes the newer syncs have already run. An earlier timestamp inverts that and reverts v5.49 state on every database.

**`.pg-only.sql` is intentional.** There is no SQL Server counterpart and none should be authored; SQL Server never had the gap.

**The new gate can fail future PRs.** That is the point, but a reviewer seeing it fail should read it as "the converter dropped a statement," not as a flaky check.

**Not from this PR:** `check-pg-migration-parity.mjs` reports 3 v5.50 SS migrations without PG counterparts (`Lists_UI_Role_Permissions`, `ContentItem_VectorRecordID_And_ContentItemChunk`, `Break_CodeGen_Cycle_Remove_PromptRun_AgentRunID`). All three are pre-existing on `next` and are build-engineer release tasks.

**No `mj sync push` is required.** `mj migrate` alone converges every population — which is what makes ask (1) moot rather than merely documented.

## Also included

`DEPLOYMENT.md` §10c described the docs pipeline that #3268 replaced — it pointed the build engineer at `memberjunction.github.io/MJ/` (now a redirect), called `docs.yml` a TypeDoc-only deploy, and listed `publish.yml` completion as its only trigger. It also now fires on doc pushes to `main` and on manual dispatch, so the Post-Merge Checklist item could be ticked against a green run belonging to an unrelated docs deploy. Corrected against the workflow, the CNAME, and `typedoc.json`.


plans/adr/0001-forward-dated-reseed-for-ledger-gaps.md
```diff
@@ -0,0 +1,34 @@
+# Heal PG ledger gaps with forward-dated idempotent reseed migrations
```
jordanfanapour: Remove this plan document before merging.

CONTEXT.md
```diff
@@ -0,0 +1,55 @@
+# MemberJunction Release Engineering — PG Migration Ledger
```
jordanfanapour: Nit: useful glossary for the PG migration domain, but a new root-level file adds clutter (root already has CLAUDE.md, DEPLOYMENT.md, CONTRIBUTING.md, UPGRADE-v5.0.md…). Would `guides/PG_MIGRATION_GLOSSARY.md` (indexed in `guides/README.md`) be a better home? This reads more like a reference guide than a top-level project doc.

.github/scripts/__tests__/generate-v545-metadata-reseed.test.mjs
```diff
@@ -0,0 +1,39 @@
+import { describe, it, expect } from 'vitest';
```
jordanfanapour, line 1: The `findSupersededCreates` tests are good, but the generator has several other safety-critical functions that aren't tested here:

- `assertSupersessionSafe` — field-superset validation (what happens when a later sync's fields are NOT a superset?)
- `assertDeleteParity` — converter/source delete count agreement
- `assertShape` — the 161/13/20/1 shape pin
- `guardCreate` / `synthesizeDeletes` — output format correctness

The corpus-level "byte-identical regeneration" check is strong evidence these work _today_, but won't protect against future regressions. Worth adding at least the `assertSupersessionSafe` and `assertShape` edge cases since those are the guards that prevent silent data loss.

packages/SQLConverter/src/__tests__/StatementClassifier.test.ts
```diff
@@ -722,4 +722,21 @@ END`;
       expect(classifyBatch(sql)).toBe('SKIP_SQLSERVER');
     });
   });
```
jordanfanapour, line 724: Good that the `spDelete` → `EXEC_BLOCK` and the `spDeleteUnneededEntityFields` → `SKIP_SQLSERVER` near-miss are both pinned. Consider also adding:

- `DECLARE + UPDATE` → `DECLARE_DML_BLOCK`
- `DECLARE + INSERT` → `DECLARE_DML_BLOCK`
- `DECLARE` with no `EXEC`/DML → `SKIP_SQLSERVER`

These classification paths existed before this PR, but they're closely related to the new routing and would protect against regressions if the classifier's match ordering changes.

jordanfanapour (approved):
You can proceed with merging this PR after addressing the four comments that I left.

PR #3279, opened 2026-07-25
fix(codegen): exclude external entity schemas from generated GraphQL server code

MS-BC:
## The bug

`RunCodeGenBase.generateCode()` passed the **unfiltered** `nonCoreEntities` list to `generateGraphQLServerCode`, while entity-subclass generation (and Angular generation) used the external-schema-filtered `localNonCoreEntities`:

```ts
// GraphQL server code — UNFILTERED
graphQLGenerator.generateGraphQLServerCode(nonCoreEntities, graphqlOutputDir, entityPackageName, false)

// ...computed AFTER, and used only for entity subclass / Angular generation
const externalSchemas = getExternalEntitySchemas().map(s => s.toLowerCase());
const localNonCoreEntities = externalSchemas.length > 0
  ? nonCoreEntities.filter(e => !externalSchemas.includes(e.SchemaName.toLowerCase()))
  : nonCoreEntities;
```

`getExternalEntitySchemas()` returns the schemas owned by *other* packages (the keys of the `entityPackageName` map). So a package codegen'd in a workspace where sibling packages' entities are also present in metadata emitted `@ObjectType` classes for those foreign entities — e.g. a `forms-server` package shipping `mjBizAppsCommon*` and `mjBizAppsTasks*` ObjectTypes that `common-server` and `tasks-server` also define.

## Why it detonates at boot

Those stowaway types are inert until two packages defining the same GraphQL type name are loaded together. `type-graphql`'s `buildSchemaSync` then hands graphql-js duplicate named types, and graphql-js enforces name uniqueness as a hard invariant:

```
Schema must contain uniquely named types but contains multiple types named "mjBizAppsTasksTaskActivity_"
```

The throw happens during module init, so the process exits before any HTTP/resolver surface comes up — the API crash-loops under pm2. This stays latent until a deployment enables all the affected `-server` packages in `dynamicPackages.server` simultaneously, at which point it is a hard outage with no partial degradation.

## The fix

Move the `externalSchemas` / `localNonCoreEntities` computation above the GraphQL block and pass the filtered list — making GraphQL generation consistent with every other generator in the function. Confirmed on a live instance: after removing the duplicate ObjectTypes, schema build succeeded and the API came up.

Behavior is unchanged when no external schemas are configured (`localNonCoreEntities === nonCoreEntities`).

### The second half: what the filter alone would have broken

Withholding an entity also withholds its `@ObjectType` **class declaration**. A local entity with a one-to-many to an externally-owned entity still emitted `@Field(() => [<external>_])`, a matching `@FieldResolver`, and an `import` of the related table — all naming a class no longer declared anywhere in the file. That converts the boot crash into a TypeScript compile error, which is better but still broken.

A single predicate, `isRelatedTypeOutOfScope`, now gates the three emission sites (relationship field, field resolver, header import). MJ-core related entities are deliberately **not** out of scope: they are absent from the generated set too, but they resolve through the existing `mj_core_schema_server_object_types` namespace import rather than a local declaration.

---

## Testing

Unit: `packages/CodeGenLib` builds clean; `npm run test` → **622 passed, 60 skipped, 0 failed** (28 files, 5 skipped). New `graphql-external-schema-filter.test.ts` drives the real generator with a schema→package map configured and asserts the withheld type is never named, the skip comment is emitted, the local entity still generates, and MJ-core relationships are unaffected.

The manual check compiles what the generator actually writes. The same local entity — schema `host`, one-to-many to an entity in `appschema`, with `entityPackageName` set to `{ appschema: '@acme/app-entities' }` — was emitted by **both** builds, and each emitted body was fed to the real `tsc`:

```
[PRE-FIX ] references to the withheld ObjectType: 5 -> appschemaAppMembers_, AppMembers_
[PRE-FIX ] skip comment emitted: false
[PRE-FIX ] tsc "TS2304: Cannot find name 'appschemaAppMembers_'" occurrences: 3

[POST-FIX] references to the withheld ObjectType: 0
[POST-FIX] skip comment emitted: true
[POST-FIX] tsc "TS2304: Cannot find name 'appschemaAppMembers_'" occurrences: 0
```

Three real `tsc` errors, not a lint rule or a string assertion — the pre-fix file does not compile once `runCodeGen` withholds the entity, which is precisely the state the filter creates.

The original failure — the one that crash-loops the API — was confirmed on a live instance: with both packages loaded, `buildSchemaSync` threw `Schema must contain uniquely named types but contains multiple types named "mjBizAppsTasksTaskActivity_"`; after the duplicate ObjectTypes were removed the schema built and the API came up.

Not covered: a full `mj codegen` over a real multi-package OpenApp workspace. The generator was driven directly so the pre-fix and post-fix emissions could be compared byte for byte on identical input; a full run would exercise the same code path with more entities and no different decision.


AN-BC (commented):
@MS-BC this looks good, however, doesn't excluding them from GraphQL mean that we cannot run views on these entities? Or is that through a generic mechanism? How about loading single records on generated entity forms for external entities?

MS-BC:
Good question, and the answer isn't obvious from the diff — I traced both paths.

**Neither is generic at the operation level.** `GraphQLDataProvider.RunView` builds the query name from the entity: `Run${graphQLTypeName}DynamicView` / `ViewByID` / `ViewByName` ([graphQLDataProvider.ts:846-863](https://github.com/MemberJunction/MJ/blob/next/packages/GraphQLDataProvider/src/graphQLDataProvider.ts#L846-L863)). Single-record load does the same — `Single${graphQLTypeName}` ([:1967-1969](https://github.com/MemberJunction/MJ/blob/next/packages/GraphQLDataProvider/src/graphQLDataProvider.ts#L1967-L1969)). So those operations must exist in the schema, per entity.

They're generic in the *body* only: each generated resolver is a typed wrapper that delegates straight to `ResolverBase` ([graphql_server_codegen.ts:372-391](https://github.com/MemberJunction/MJ/blob/next/packages/CodeGenLib/src/Misc/graphql_server_codegen.ts#L372-L391)):

```ts
@Query(() => Run${typeNameBase}ViewResult)
async Run${typeNameBase}DynamicView(...) {
    input.EntityName = '${entity.Name}';
    return super.RunDynamicViewGeneric(input, provider, userPayload, pubSub);
}
```

**What the PR removes is the duplicate, not the operation.** The filter only fires when `entityPackageName` is a schema→package map — i.e. when another package is declared as owning that schema. That package's CodeGen emits the entity's ObjectType and its resolver, and its resolvers are loaded into the *same* `buildSchemaSync` call: `mj app install` writes `dynamicPackages.server[]`, server-bootstrap imports each package and collects its exported `RESOLVER_PATHS` ([ServerBootstrap/src/index.ts:161-171](https://github.com/MemberJunction/MJ/blob/next/packages/ServerBootstrap/src/index.ts#L161-L171)), those are appended to `serve()`'s globs ([:257-264](https://github.com/MemberJunction/MJ/blob/next/packages/ServerBootstrap/src/index.ts#L257-L264)) and end up in `resolvers: allResolvers` ([MJServer/src/index.ts:828-829](https://github.com/MemberJunction/MJ/blob/next/packages/MJServer/src/index.ts#L828-L829)).

So the count goes 2 → 1, not 1 → 0. And 2 was not a working state: that's exactly what threw `Schema must contain uniquely named types but contains multiple types named "mjBizAppsTasksTaskActivity_"` and crash-looped the API — at which point `RunView` worked for *no* entity, external or not.

With a single-package config `getExternalEntitySchemas()` returns `[]` and nothing is filtered, so output is byte-identical to today.

**On forms specifically.** `Load()` only requests nested relationship fields when the caller passes `EntityRelationshipsToLoad` ([:1965](https://github.com/MemberJunction/MJ/blob/next/packages/GraphQLDataProvider/src/graphQLDataProvider.ts#L1965)); with the default `null` it emits a plain `Single<Type>` over scalar fields. Nothing in MJ passes it — grepping the repo, the only hits are the generator that emits the optional parameter, the provider that implements it, and the generated signatures. Related-entity sections on a form issue their own `RunView` against the related entity, which routes to that entity's own `Run<Type>DynamicView` in its owning package. So the relationship-skip half of this PR only affects the nested `@Field(() => [external_])` on the local type, reachable solely through `Single<Type>Full`.

Worth adding: this list isn't new. `localNonCoreEntities` already drove entity-subclass and Angular generation — GraphQL was the one artifact still getting the unfiltered list. The diff moves the computation above the GraphQL call and passes it. So after this, an entity's `BaseEntity` subclass and its resolver come from the same package; before, they were split.

**The one real hazard**, stated plainly: if a schema is listed in the map but its owning package isn't actually installed, those operations won't exist and RunView on that entity fails at the GraphQL layer. That's pre-existing — the same config already withholds the entity's `BaseEntity` subclass — but this PR extends it to the resolver, so the failure mode moves from "type missing at compile time" to "operation missing at query time."

**Separate concept, since "external" is overloaded here:** entities with `ExternalDataSourceID` set (a non-MJ database) are gated independently and untouched by this PR — the single-record resolver routes through `LoadExternalRecordByKey`, `All<Entity>` is skipped since there's no MJ base view, and `graphql-server-codegen-external.test.ts` covers it. This PR is about schema *ownership across packages*, not external data sources.

AN-BC:
@MS-BC ok sounds like a good fix, and we'll need to sort through more of this external data source stuff with @rkihm-BC and @JosueGarcia-BC when Robert is back next week. I'll let you guys sort that out on the one remaining gap. Thanks.

PR #3280, opened 2026-07-26
fix(MJCore): surface parent errors when an IS-A parent save or delete fails

AN-BC:
@MarceloT-BC @AN-BC for review.

## The bug

When a Table-Per-Type (IS-A) child's **parent** operation fails, `BaseEntity` rolls back and returns `false` **without recording anything on the child's `ResultHistory`**. Every result was written to the parent object, which callers have no reference to — `_parentEntity` is private. So the caller sees:

```
Save()/Delete() -> false
LatestResult    -> null
ResultHistory   -> []
```

A failure with **no diagnostic anywhere reachable**. This existed on **both** the parent-save path and the symmetric parent-delete path.

## How it surfaced

**Save:** Saving an `AccountingCompanyProfile` (IS-A `__mj.Company`) in bizapps-accounting. `__mj.Company` has both `Name` and `Description` `NOT NULL`; the child never set them, so the parent failed validation with `Name cannot be null` / `Description cannot be null`.

Neither message reached the caller — `Save()` just returned `false`. `ValidateAsync()` on the child passed (its own fields were fine), and calling the generated sproc directly worked, which made it look like an app-level defect in accounting. It wasn't. Diagnosing it required stepping into `BaseEntity` internals to find that the parent held the errors.

**Delete:** The same black hole existed on delete. In an IS-A delete the child's own row is removed first and the delete then cascades up the parent chain; if a **parent** delete fails (e.g. an FK constraint referencing the parent row from elsewhere), `Delete()` rolled back and returned `false` with nothing on the child — identical invisible failure.

## The fix

Record a result on the child before returning on **both** the parent-save and parent-delete failure paths, mirroring the two pre-existing failure paths in each method — the transaction-group-failure handler and the `catch` block — including the same `currentResultCount` guard so a result recorded elsewhere is never double-reported.

Two details worth noting:

- **Parent field errors are copied onto the child's result**, so callers can act on them programmatically rather than parsing a string.
- **The message falls back to the joined error text.** A failed parent commonly reports its detail *only* in `Errors` — validation failures leave `Message` empty — so using `Message` alone would still hand the caller a message that says nothing. Caught this because the first cut of the save fix produced exactly that.

Note on the delete path: the child's **own** row has already been deleted successfully at the point the parent-chain delete fails; it's the parent delete that failed and rolled the transaction back. The recorded result is typed `'delete'` and names the parent entity that failed.

## Verification

**Save** — reproduced and confirmed at runtime against the compiled build, saving an IS-A child whose parent has unset `NOT NULL` columns:

| | before | after |
|---|---|---|
| `ResultHistory.length` | `0` | `1` |
| `LatestResult` | `null` | populated |
| `LatestResult.Errors` | — | `['Name cannot be null', 'Description cannot be null']` |

**Delete** — covered by a new regression test file that fails against the pre-fix source (all four cases: `LatestResult` null / empty `ResultHistory`) and passes with the fix.

`MJCore` typechecks clean. Full unit suite: **1581 passing** (102 files), including the 4 parent-save and 4 parent-delete regression tests.

## Scope

Behaviour on the success path is unchanged; this only adds diagnostics to two paths that previously returned silently. Any caller currently relying on `ResultHistory` being empty after an IS-A parent failure would see one entry now — that seems strictly desirable, but flagging it since it is technically observable.


AN-BC:
@rkihm-BC  please take a quick peak at this PR and merge if it looks ok. We'll need to do a `5.50.0` build soon to include this as it is an important defect to patch.

rkihm-BC:
@AN-BC , code review identified the same problem with Delete so I had CC implement the same fix on Delete.

AN-BC:
> @AN-BC , code review identified the same problem with Delete so I had CC implement the same fix on Delete.

Good addition. Merging now.

PR #3282, opened 2026-07-26
fix(cli): keep Open App client packages from being tree-shaken out of production builds

MS-BC:
## The bug

`mj codegen manifest --open-app-client-bootstrap` emitted a **bare side-effect import** for each Open App client package:

```ts
// BEGIN Open App client bootstrap
import '@mj-biz-apps/committees-ng';
// END Open App client bootstrap
```

Bundlers are permitted to drop a bare side-effect import when the imported package declares `"sideEffects": false` in its `package.json` — and **ng-packagr emits exactly that by default for every Angular library**, so it applies to essentially all `-ng` Open App packages.

When the import is dropped, the package's module-scope `@RegisterClass(...)` calls never run. `MJGlobal.ClassFactory` then has no registration for the package's resource/view components (e.g. `MemberHomeComponent`), and they are simply absent from the production build.

There is **no error** — no missing-module failure, no console warning. The manifest looks correct in source, the build succeeds, and the app renders nothing for that resource. This is the same class of failure the manifest system exists to prevent; the manifest just wasn't immune to it itself.

## The fix

Emit **referenced** namespace imports collected into an exported array, anchored by an observable side effect:

```ts
// BEGIN Open App client bootstrap
import * as __openAppClient0 from '@mj-biz-apps/committees-ng';

/* eslint-disable @typescript-eslint/no-explicit-any */
export const OPEN_APP_CLIENT_MODULES: any[] = [__openAppClient0];
(globalThis as any).__mjOpenAppClientModules = OPEN_APP_CLIENT_MODULES;
// END Open App client bootstrap
```

A namespace import that is actually referenced forces the bundler to evaluate the module. The `globalThis` assignment guarantees the reference survives even aggressive dead-code elimination, independent of whether or how the host app imports `OPEN_APP_CLIENT_MODULES` — the same anchoring strategy `CLASS_REGISTRATIONS` already relies on.

Unchanged: disabled entries still become comments (so a re-enable restores the import without losing the record), the block is still idempotent, a changed entry set still fully replaces the block, and zero entries still removes the block entirely.

## Note

This is one of two independent findings from the same investigation. The other — CodeGen passing the unfiltered entity list to GraphQL server generation — is already up as #3279. They are separate fixes with no ordering dependency, but both are worth carrying in the same release.

---

## Testing

Unit: `packages/MJCLI` builds clean; `manifest-client-bootstrap.test.ts` passes and now asserts both that the referenced form is emitted **and** that the bare form is not. Full-suite numbers come from CI's `Run unit tests` — a local run in a scratch worktree with symlinked `node_modules` shows 431 passed with 5 failures confined to `claude-pack/claude-commands.test.ts`, which passes 15/15 in a normal worktree and is untouched by this change.

The claim under test is a **bundler** claim, so the manual check is a real production bundle, not an assertion about emitted text. Both forms were bundled with the same esbuild the Angular `application` builder uses (`--bundle --minify --format=esm --platform=browser`) against a package built the way ng-packagr builds every Angular library — `"sideEffects": false` — whose module scope has one observable side effect standing in for `@RegisterClass(...)`:

```js
globalThis.__registered = (globalThis.__registered ?? 0) + 1;
```

**Bare side-effect import** (what the manifest emitted before):
```js
import '@mj-biz-apps/committees-ng';
```
```
bundle: 69 bytes
console.log("registrations at runtime:",globalThis.__registered??0);
registrations at runtime: 0
```

The bundle contains **zero bytes of the package** — not a stripped call, the entire module is gone (`committees-ng` appears 0 times in the output). No error, no warning, exit code 0.

**Referenced namespace import** (what it emits now):
```js
import * as __openAppClient0 from '@mj-biz-apps/committees-ng';
export const OPEN_APP_CLIENT_MODULES = [__openAppClient0];
globalThis.__mjOpenAppClientModules = OPEN_APP_CLIENT_MODULES;
```
```
bundle: 350 bytes
...globalThis.__registered=(globalThis.__registered??0)+1;var l="committees-ng";var _=[e];globalThis.__mjOpenAppClientModules=_;...
registrations at runtime: 1
```

The module survives minification and its module-scope side effect runs. This is the whole failure mode end to end: with the bare form the class registration never happens, `MJGlobal.ClassFactory` has no entry, and the resource renders nothing — with no build-time signal that anything was dropped.

Not covered: a full MJExplorer production build with a real Open App client package installed. The mechanism above is bundler behaviour driven by `sideEffects: false`, which is identical in that build; what a full build would add is confirmation that no *other* Explorer-specific config already worked around it, and it did not.


jordanfanapour (dismissed):
## Review: fix(cli): keep Open App client packages from being tree-shaken out of production builds

The problem is real and well-diagnosed — bare side-effect imports are legally dropped when `sideEffects: false` is set, and ng-packagr sets that by default. But the fix is a workaround for a gap in the existing manifest system, not a proper solution.

### The existing manifest system already solves this problem

`GenerateClassRegistrationsManifest.ts` walks the dependency tree from `package.json`, scans for `@RegisterClass` decorators, and generates **named imports** collected into an exported `CLASS_REGISTRATIONS` array. That array reference makes the imports immune to tree-shaking — it's why every `@memberjunction/*` registered class survives production builds today.

Open App client packages are added to MJExplorer's `package.json` as direct dependencies by `mj app install` (via `AddAppPackages()` in `packages/OpenApp/Engine/src/install/package-manager.ts`). So they **are** in the dependency tree. They should be getting discovered, scanned, and added to `CLASS_REGISTRATIONS` automatically.

### Why they're not: `findScannableFiles()` silently skips them

The actual root cause is in `GenerateClassRegistrationsManifest.ts` lines 345-361. Open App packages are **npm-distributed** — they ship compiled JS in `dist/`, not TypeScript in `src/`. The scanner checks for `src/` first, and only falls back to `dist/` if the `--scan-dist` flag is explicitly passed:

```typescript
async function findScannableFiles(packageDir, excludePatterns, scanDist) {
    const srcDir = path.join(packageDir, 'src');
    if (fs.existsSync(srcDir)) {
        return { files, isCompiledJS: false };
    }
    if (scanDist) {  // ← FALSE BY DEFAULT
        const distDir = path.join(packageDir, 'dist');
        // ...
    }
    return { files: [], isCompiledJS: false };  // ← empty → package silently skipped
}
```

MJExplorer's `prestart` script doesn't pass `--scan-dist`. So every Open App package returns zero files and gets skipped. The scanner already knows how to extract `@RegisterClass` from compiled `__decorate()` calls (lines 414-478) — the infrastructure is there, it's just gated behind a flag that's never set for the client manifest.

### What this PR does instead

Rather than fixing the manifest generator to discover these classes, this PR builds a **parallel anti-tree-shaking system**:

- Discovery via `mj.config.cjs` `dynamicPackages.client` instead of the `package.json` → AST scan pipeline
- Namespace imports (`import * as __openAppClient0`) instead of named class imports
- A separate `OPEN_APP_CLIENT_MODULES` array alongside the existing `CLASS_REGISTRATIONS`
- A `globalThis` anchor that wouldn't be needed if the classes were in `CLASS_REGISTRATIONS`
- `any[]` with `eslint-disable` — compounding the existing `any` in the manifest (which itself should eventually be fixed)

This works, but it means Open App packages use a fundamentally different registration path than every other package in the system. Two arrays, two discovery mechanisms, two anti-tree-shaking strategies.

### Suggested approach

Fix `findScannableFiles()` to auto-enable `dist/` scanning for packages that lack `src/`:

```typescript
async function findScannableFiles(packageDir, excludePatterns, scanDist) {
    const srcDir = path.join(packageDir, 'src');
    if (fs.existsSync(srcDir)) {
        const files = await findSourceFiles(srcDir, excludePatterns);
        return { files, isCompiledJS: false };
    }
    // Auto-scan dist/ for npm-distributed packages (no src/ directory).
    // Workspace-linked packages always have src/; published packages never do.
    const distDir = path.join(packageDir, 'dist');
    if (fs.existsSync(distDir)) {
        const files = await findDistFiles(distDir, excludePatterns);
        return { files, isCompiledJS: true };
    }
    return { files: [], isCompiledJS: false };
}
```

With this change:
- Open App classes land in `CLASS_REGISTRATIONS` via the same pipeline as everything else
- Named imports with array references prevent tree-shaking (proven pattern)
- No `globalThis` anchors, no namespace imports, no second array
- `dynamicPackages.client` can still exist for the `Enabled` flag (`mj app disable`), but it's not responsible for import generation
- The `--scan-dist` flag becomes opt-in for other use cases but is no longer required for the common case

### Minor notes

- The `--scan-dist` flag could be kept as an explicit override (force dist-only scanning even when `src/` exists), but the default behavior should be "scan what's available"
- The `Enabled` flag from `dynamicPackages.client` would need to be checked during manifest generation to skip disabled apps — a small addition to the existing filtering logic

MS-BC:
Thanks — the diagnosis is right and it's the most useful part of this review. `findScannableFiles()` does gate the `dist/` fallback behind `--scan-dist` (`GenerateClassRegistrationsManifest.ts:346`), and MJExplorer's `prestart` doesn't pass it:

```
mj codegen manifest --exclude-packages @memberjunction --output ./src/app/generated/class-registrations-manifest.ts --open-app-client-bootstrap
```

So Open App packages are scanned as zero files and never reach `CLASS_REGISTRATIONS`. That's a real gap and it should be fixed.

Where I don't follow you is the conclusion that fixing it makes this PR unnecessary — there's a load-bearing step in the proposed path that hasn't been checked.

**`filterToExportedClasses` silently drops classes that aren't in the package's public `.d.ts`.** `GenerateClassRegistrationsManifest.ts:745`:

```typescript
if (packageExports.has(cls.className)) {
    exported.push(cls);
} else {
    skipped.push(cls);   // logged as "Skipped N classes not exported from their package's public API"
}
```

It has to work this way — the manifest emits a *named* import, so it can only reference something the entry point actually exports. But Open App resource/view components are registered via `@RegisterClass` precisely so that nothing has to import them by name; they're instantiated through `ClassFactory` by key. They are the classes most likely to be absent from `public-api.ts`, because nothing in the codebase has a reason to put them there.

If `MemberHomeComponent` isn't exported from `@mj-biz-apps/committees-ng`'s entry point, turning on dist scanning discovers it, then drops it one step later — and the app is missing the component again, with the same silent failure mode, just relocated. That's the specific thing I'd want verified before treating the named-import path as a replacement rather than a complement.

Two smaller things in the same area:

- **Auto-enabling `dist/` scanning changes discovery for every dist-only package, not just Open Apps.** In a distributed install nothing has `src/` — that's what the pre-built bootstrap manifests in `server-bootstrap` / `ng-bootstrap` exist to cover. Making dist the default fallback means globbing the `dist/` of every transitive dep on the consumer's install path. That may well be the right call, but it's a scope change to the install path, not a local fix.
- **ng-packagr emits both `fesm2022/*.mjs` and `esm2022/**/*.mjs`**, and `findDistFiles` globs `**/*.{js,mjs,cjs}` across the whole `dist/`. The same class gets discovered from two files, so the dedup path would need checking too.

**On the "parallel system" framing** — agreed that two discovery mechanisms is not where this should end up. But the reason to keep this fix for 5.50 is that it works *independently of export status*: a referenced namespace import forces module evaluation, which is all `@RegisterClass` needs. It doesn't care whether the class is in the public API. The `CLASS_REGISTRATIONS` path does.

So I'd take these as sequential rather than either/or:

1. **Now (5.50):** this PR. 54 lines in one file, no change to how any other package is discovered, and it fixes a bug where the app silently ships without its components.
2. **6.x:** your `findScannableFiles()` change, gated on confirming Open App `-ng` packages export their registered components from the entry point (and if they don't, that's a separate fix on the Open App build side — arguably the *real* root cause, since `filterToExportedClasses` is doing the right thing given what it's handed). Once that holds, this block collapses into `CLASS_REGISTRATIONS` and `dynamicPackages.client` goes back to owning only the `Enabled` flag. Happy to open the issue for it.

**Taken directly:** the `any[]` — you're right, and it wasn't defensible. Pushed in 7b7145ad: `unknown[]` with the `globalThis` cast as `Record<string, unknown>`. The elements are module namespace objects that exist only to be referenced; nothing reads a member off them, so there's no member access to type. That drops the `eslint-disable` entirely, and the test now asserts the generated block contains neither `any` nor an eslint escape hatch.

jordanfanapour:
> Thanks — the diagnosis is right and it's the most useful part of this review. `findScannableFiles()` does gate the `dist/` fallback behind `--scan-dist` (`GenerateClassRegistrationsManifest.ts:346`), and MJExplorer's `prestart` doesn't pass it:
> 
> ```
> mj codegen manifest --exclude-packages @memberjunction --output ./src/app/generated/class-registrations-manifest.ts --open-app-client-bootstrap
> ```
> 
> So Open App packages are scanned as zero files and never reach `CLASS_REGISTRATIONS`. That's a real gap and it should be fixed.
> 
> Where I don't follow you is the conclusion that fixing it makes this PR unnecessary — there's a load-bearing step in the proposed path that hasn't been checked.
> 
> **`filterToExportedClasses` silently drops classes that aren't in the package's public `.d.ts`.** `GenerateClassRegistrationsManifest.ts:745`:
> 
> ```ts
> if (packageExports.has(cls.className)) {
>     exported.push(cls);
> } else {
>     skipped.push(cls);   // logged as "Skipped N classes not exported from their package's public API"
> }
> ```
> 
> It has to work this way — the manifest emits a _named_ import, so it can only reference something the entry point actually exports. But Open App resource/view components are registered via `@RegisterClass` precisely so that nothing has to import them by name; they're instantiated through `ClassFactory` by key. They are the classes most likely to be absent from `public-api.ts`, because nothing in the codebase has a reason to put them there.
> 
> If `MemberHomeComponent` isn't exported from `@mj-biz-apps/committees-ng`'s entry point, turning on dist scanning discovers it, then drops it one step later — and the app is missing the component again, with the same silent failure mode, just relocated. That's the specific thing I'd want verified before treating the named-import path as a replacement rather than a complement.
> 
> Two smaller things in the same area:
> 
> * **Auto-enabling `dist/` scanning changes discovery for every dist-only package, not just Open Apps.** In a distributed install nothing has `src/` — that's what the pre-built bootstrap manifests in `server-bootstrap` / `ng-bootstrap` exist to cover. Making dist the default fallback means globbing the `dist/` of every transitive dep on the consumer's install path. That may well be the right call, but it's a scope change to the install path, not a local fix.
> * **ng-packagr emits both `fesm2022/*.mjs` and `esm2022/**/*.mjs`**, and `findDistFiles` globs `**/*.{js,mjs,cjs}` across the whole `dist/`. The same class gets discovered from two files, so the dedup path would need checking too.
> 
> **On the "parallel system" framing** — agreed that two discovery mechanisms is not where this should end up. But the reason to keep this fix for 5.50 is that it works _independently of export status_: a referenced namespace import forces module evaluation, which is all `@RegisterClass` needs. It doesn't care whether the class is in the public API. The `CLASS_REGISTRATIONS` path does.
> 
> So I'd take these as sequential rather than either/or:
> 
> 1. **Now (5.50):** this PR. 54 lines in one file, no change to how any other package is discovered, and it fixes a bug where the app silently ships without its components.
> 2. **6.x:** your `findScannableFiles()` change, gated on confirming Open App `-ng` packages export their registered components from the entry point (and if they don't, that's a separate fix on the Open App build side — arguably the _real_ root cause, since `filterToExportedClasses` is doing the right thing given what it's handed). Once that holds, this block collapses into `CLASS_REGISTRATIONS` and `dynamicPackages.client` goes back to owning only the `Enabled` flag. Happy to open the issue for it.
> 
> **Taken directly:** the `any[]` — you're right, and it wasn't defensible. Pushed in [7b7145a](https://github.com/MemberJunction/MJ/commit/7b7145ad061fa9e380de74728f1999c7d4d8c514): `unknown[]` with the `globalThis` cast as `Record<string, unknown>`. The elements are module namespace objects that exist only to be referenced; nothing reads a member off them, so there's no member access to type. That drops the `eslint-disable` entirely, and the test now asserts the generated block contains neither `any` nor an eslint escape hatch.

Tagging @AN-BC and @rkihm-BC to this discussion.

Thanks for the detailed response — the filterToExportedClasses concern is the right thing to check. But I investigated and every @memberjunction/* package consistently exports every @RegisterClass-decorated class from its entry point — ai-anthropic, ai-openai, ai-gemini, ng-bootstrap, etc. Zero exceptions across both Angular and non-Angular packages. This is by design: the manifest emits named imports, so the class must be publicly exported.

I propose that we require every open app to export its set of @RegisterClass-decorated class from its entry point, making it part of the Open App contract. If Open App -ng packages don't follow this same convention, that's a bug in the Open App package build — not a reason to route around the manifest system. A @RegisterClass component that isn't exported from its entry point is already broken by MJ convention, and the fix should be making it exported, not building a parallel import mechanism to compensate.

So before we accept this as a 5.50 workaround: do any of the shipped Open App -ng packages actually omit their registered classes from public-api.ts? If they do, I'd rather fix that at the source. If they don't, then the findScannableFiles() fix + confirming the export convention holds is the complete solution — filterToExportedClasses wouldn't drop anything.

On the scope concerns — auto-enabling dist/ scanning for packages without src/ and ng-packagr's duplicate fesm2022/esm2022 output — those are fair points worth handling, but they're implementation details of the manifest generator fix, not reasons to avoid the fix. The dedup path already exists for workspace packages that appear multiple times in the dep tree; extending it to handle ng-packagr's dual output is straightforward.

Appreciate the any[] → unknown[] cleanup.

MS-BC:
> do any of the shipped Open App `-ng` packages actually omit their registered classes from `public-api.ts`?

Yes — and so does MJ. I measured both.

**`@mj-biz-apps/committees-ng@1.0.0`** (straight off npm, not a local build): 32 `@RegisterClass` classes in `dist/`, **19 not exported** from the entry point. Seventeen are per-entity `*FormComponent`s; the other two are `ActionItemTrackerComponent` and `ManagementDashboardComponent`.

**`@memberjunction/ng-core-entity-forms`** on `next`: 414 `@RegisterClass` classes, **393 not exported** from `public-api.ts`. 382 are generated forms; 11 are hand-written:

```
AISkillSharingPanel, AgentRealtimePanel, ModelPredictionPanel,
TagPipelineConfigurationPanel, WebsiteCrawlerSettingsPanel,
MJAIAgentSessionBridgeFormComponentExtended, MJAIBridgeAgentIdentityFormComponentExtended,
MJAIBridgeProviderFormComponentExtended, MJSearchScopeFormComponentExtended,
MJSearchScopeProviderFormComponentExtended, RecordProcessFormComponentExtended
```

`public-api.ts` does `export * from './lib/generated/generated-forms.module'`, and that module *imports* 410 components to declare them but re-exports only the `GeneratedForms_SubModule_N` NgModules. So the components are reachable for evaluation and invisible to `collectExportedClassNames`.

Which means those 393 aren't in `CLASS_REGISTRATIONS` today either. They survive production builds because MJExplorer imports `CoreEntityFormsModule` by name and that evaluates the whole graph behind it — the same mechanism this PR's namespace import uses. `CLASS_REGISTRATIONS` isn't carrying them; module evaluation is.

So it isn't an Open App build bug to fix in one repo. It's what `core-entity-forms` emits, for MJ and for every Open App with entities alike.

Concretely for the proposed path on committees-ng: `resolveTypesEntryPoint` falls back to `main` → `dist/public-api.d.ts`, which exists, so `exportsMap` gets populated and `filterToExportedClasses` takes the `else` branch on all 19 — discovered by dist scanning, then skipped one line later. Same missing components, same silent failure, one stage further along. (19 is the optimistic count; I resolved `export *` recursively.)

Module evaluation reaches 29 of the 32. The 3 it doesn't — `ActionItemTrackerComponent`, `ManagementDashboardComponent`, `mjBizAppsCommitteesActionItemFormComponent` — aren't reachable from the entry by *any* mechanism, named or namespace. That's a genuine app-side bug and neither approach here fixes it.

**On the contract.** I'm for it, with one amendment: it can't be met by app authors today, because the bulk of the classes are files they don't write. It needs `core-entity-forms`' generator to emit the re-exports — for MJ's own package as much as for Open Apps. That's the fix that makes `CLASS_REGISTRATIONS` the single mechanism, which is where you want this and where I want it.

But it's a CodeGen change plus a republish of every Open App, and it does nothing for an app already installed at 1.0.0. This PR is 54 lines in one file, changes nothing about how any other package is discovered, and works against the packages as they exist on npm right now.

Sequence:

1. **5.50** — this PR.
2. **6.x** — generator emits the re-exports; `findScannableFiles()` auto-enables `dist/` (with the `fesm2022`/`esm2022` dedup you flagged); Open Apps republish. Once a version floor makes that safe, this block deletes and `dynamicPackages.client` goes back to owning only `Enabled`.

I'll open the 6.x issue with these numbers attached so step 2 starts from evidence rather than from this thread.

Reproduce:

```bash
npm pack @mj-biz-apps/committees-ng@1.0.0 && tar xzf mj-biz-apps-committees-ng-1.0.0.tgz
# count `X = __decorate([...RegisterClass(...)...], X)` in package/dist/**/*.js          -> 32
# resolve export*/export{} recursively from package/dist/public-api.d.ts                 -> 19 absent
# same walk over packages/Angular/Explorer/core-entity-forms/src/public-api.ts on next   -> 393 of 414 absent
```

AN-BC:
@MS-BC @jordanfanapour I'm going to ask @cadam11 to step in for me as he is taking on other infra stuff related to OpenApp and I want him to weigh in here on the best approach.

cadam11:
IMO, merge this for 5.50 as a short-term quick fix, and point the 6.x issue somewhere else entirely.

You're both right. The scanner gap is real, and Madhav's numbers show the export-contract path doesn't work against today's packages (our own core-entity-forms has 393 unexported registered classes, so this isn't an Open App bug to fix in one repo).

But both fixes are the same workaround in different hats. The root problem is registration by import-time side effect: the bundler correctly sees dead code, and we keep inventing tricks to make it evaluate the modules anyway. We already have machinery that skips the fight entirely: `ClassFactory.RegisterLazyLoader` plus `LazyModuleRegistry`, mapping `Base::Key` to `() => import(...)` chunk loaders. A dynamic import is a real reference so nothing gets shaken out, evaluating the chunk runs every `@RegisterClass` regardless of what `public-api.ts` exports, and each app becomes its own lazy chunk instead of another tenant in our already-large main bundle.

The missing piece is the key-to-package map, and the publisher should ship it: `mj app publish` has the app's source right there, so scan it then and emit a small manifest. No dist scanning at install time, no export contract.

So: merge this now. Madhav, aim the 6.x issue you offered to open at the lazy path rather than dist scanning plus an export contract. I.e. stop forcing Open App client packages into the main bundle at all, and instead have each app declare its registrations at publish time so the client can lazy-load the app as its own chunk on first use, through the lazy-loading machinery Explorer already has.

Two caveats worth capturing in it: sync `CreateInstance` call sites can't trigger a lazy load, and an app overriding an existing key never misses, so those keys need eager registration.

MS-BC:
Follow-ups from this thread are up:

- **#3337** — the review items that belong in this repo: alias numbering pinned to the array position (so the declared aliases and `OPEN_APP_CLIENT_MODULES` can't drift), regression tests for mixed enabled/disabled and all-disabled entry sets, and two docs corrections — that the `globalThis` anchor is a deliberate variant of, not the same mechanism as, the `CLASS_REGISTRATIONS` array-spread anchor (my PR description was wrong to call it precedent), and that this block is defense-in-depth against a package declaring `"sideEffects": false` while self-registering classes. `any[]` → `unknown[]` already landed here before merge.
- **#3338** — @cadam11's 6.x direction, aimed at the lazy path as asked: publish-time registration manifest → `LAZY_FEATURE_CONFIG` → per-app chunk via the `RegisterLazyLoader` / `LazyModuleRegistry` machinery Explorer already has, rather than dist scanning plus an export contract. Both caveats are captured with numbers — sync `CreateInstance` sites can't trigger a lazy load (~9 sync vs ~12 async in `packages/Angular` + `packages/MJExplorer`), and override keys need eager registration because a lazy miss silently returns the base implementation.

The scaffold change (`"sideEffects": true` on generated `-ng` packages) is noted in both but lives in the app-authoring repo, not this one.

PR #3283, opened 2026-07-26
mj migrate: print the batch context Skyway already collects

AN-BC:
## The problem

When a migration fails, `mj migrate` prints only `Error.message`:

```
Error: Failed at batch 283/954 (lines 5428-5430): Transaction (Process ID 77) was deadlocked
on lock resources with another process and has been chosen as the deadlock victim.
```

Skyway had **already built** a `MigrationExecutionError` carrying the batch number, its line range, how many batches succeeded, the batch SQL, and the lines within it matching identifiers named in the error — and we discarded all of it. For a 60,000-line generated migration, that leaves you a line range with no way to see what's in it.

## After

```
Error: Failed at batch 283/954 (lines 5428-5430): Msg 1205, Severity 13, State 55,
in trg_OrderLine_RollupTotals line 22: Transaction (Process ID 59) was deadlocked ...
Batch: 283 of 954 (lines 5428-5430)
Batches applied before the failure: 282
Failing batch:
  /* SQL text to add special date field __mj_CreatedAt to entity ...OrderLine */
  UPDATE [__mj_BizAppsOrders].[OrderLine] SET [__mj_CreatedAt] = GETUTCDATE() WHERE [__mj_CreatedAt] IS NULL;
```

The `Msg 1205 ... in trg_OrderLine_RollupTotals line 22` portion comes from a companion PR on the Skyway side — MemberJunction/skyway#22 — which stops `err.message` swallowing the procedure/line the driver reported. **This PR is independently useful without it**; together they turn an unattributed deadlock into a named object and a visible statement.

## Notes

- Both failure paths (`result.Details` and the `OnProgress` callback fallback) route through one `printMigrationError`, so they no longer report different amounts of detail.
- `BatchInfo` is read defensively — it's optional on the error type, and a reporting path must never throw while reporting a failure.
- Prefers Skyway's matched context lines (they carry **file** line numbers, so they paste straight into an editor) and falls back to a 10-line batch preview when nothing matched.

## How this came up

Debugging a genuine migration failure in a BizApps app. It took hours largely because the report named neither the object nor the statement — the failure read like server instability, so I chased connection pooling, `MAXDOP`, and stale sessions before running the same SQL through `sqlcmd`, which named the trigger immediately. The real bug was ordering in *my* migration; the CLI just never said so.

Verified end-to-end by reintroducing that failure and running it through the locally-built CLI — the before/after above is the actual output.

### DIFF
--- a/.changeset/migrate-error-detail.md
+++ b/.changeset/migrate-error-detail.md
@@ -1,0 +1,5 @@
+---
+'@memberjunction/cli': patch
+---
+
+`mj migrate` now prints the batch context Skyway already collects when a migration fails: the batch number and its line range in the script, how many batches applied before it, and either the lines matched to identifiers named in the error or a preview of the failing batch SQL. Previously only `Error.message` was shown, so a failure in a large generated migration gave a line range with no way to see what was in it.
--- a/packages/MJCLI/src/commands/migrate/index.ts
+++ b/packages/MJCLI/src/commands/migrate/index.ts
@@ -221,7 +221,7 @@ export default class Migrate extends Command {
           this.logToStderr(`    Version: ${detail.Migration.Version ?? '(repeatable)'}`);
           this.logToStderr(`    Description: ${detail.Migration.Description}`);
           if (detail.Error) {
-            this.logToStderr(`    Error: ${detail.Error.message}`);
+            this.printMigrationError(detail.Error);
           }
         }
       } else {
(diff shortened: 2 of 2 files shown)


AN-BC:
Rebased onto `next` — this PR now contains **only** the `mj migrate` change (1 commit, 2 files: `packages/MJCLI/src/commands/migrate/index.ts` + the changeset).

I'd branched off `an-dev-35` rather than `next`, which swept in two unrelated commits of @AN-BC's. Those are now #3284.

Verified the rebased branch builds: the 3 `convert.ts` type errors present are pre-existing on `next` (confirmed by building the same tree with my change reverted) and unrelated to this.

PR #2834, opened 2026-06-12
feat: Add multimodal embeddings: EmbedContent on BaseEmbeddings + Gemini Embedding 2 & Cohere providers

BarnattW-BC:
## Summary
Adds multimodal embedding support to the AI framework. Introduces `EmbedContent` on
`BaseEmbeddings` (text and/or interleaved media fused into one vector) and two providers:
**Gemini Embedding 2** (`gemini-embedding-2`) and **Cohere Embed v4** (`embed-v4.0`).
`EmbedText`/`EmbedTexts` are unchanged — this is purely additive.

## What's included
- **`BaseEmbeddings`**: new `EmbedContent(params)` (non-abstract — text-only content delegates to
  `EmbedText`; multimodal providers override) + `GetFileCapabilities()` capability declaration and
  `ValidateContentSupported()` mime gating. New `EmbedContentParams`/`EmbedContentResult` types
  (reuse `ChatMessageContent`).
- **Gemini** `GeminiEmbedding2` — multimodal (image/video/audio/PDF), via `@google/genai` 2.x.
- **Cohere** `CohereEmbedding` — text + images via the v2 embed API.
- **`AIEngine.EmbedContent`** — additive engine entry point (text routes through cached `EmbedText`).
- **`@google/genai` bumped 1.40 → 2.x** (Gemini + Vertex; required for multimodal embed support).
- **Metadata**: registered `Gemini Embedding 2` (Cohere Embed v4 already present).
- **Docs**: updated Core README (`BaseEmbeddings`, `FileCapabilities`), AI capability matrix, and
  the Gemini/Cohere package READMEs.

## Design notes (flagging for review)
- **Error handling**: `EmbedContent` currently **swallows** errors to an empty vector to match
  `EmbedText`/`EmbedTexts`. The root cause is `EmbedResult` not extending `BaseResult` (no success/
  error fields like `ChatResult`). Aligning that is **deferred** — worth a decision.
- **Deferred**: batch multimodal (`EmbedContents`), GraphQL transport for `EmbedContent`, and
  additional providers (Voyage/Jina). No current consumer needs them yet.
- Cohere PDF: embed-v4 markets PDF support, but the v2 `image_url` path only accepts PNG/JPEG/WebP/
  GIF, so PDF is intentionally not declared (would pass validation but fail at the API).

## Testing
- Live-verified both providers against the real APIs: text, batch, text+image (fused vector),
  format validation, and PDF behavior.
[...]

PR #3286, opened 2026-07-26
feat(ai-segmentation): pluggable content segmentation + chunk schema, wired into ContentAutotagging

AN-BC:
> ## 🚧 Blocked on one task — @topherHunnewell, CodeGen re-run
>
> The generated section in the migration is **stale**: it predates the four `ContentSource`/`ContentType` strategy columns, so it covers `MJ: Content Item Chunks` only. Step-by-step runbook in [this comment](https://github.com/MemberJunction/MJ/pull/3286#issuecomment-5104198667). Targeting **5.50** — @sohamdesai-BlueCypress please hold the build until that lands. Everything else is green.

---

## Segmentation decides *what* gets embedded. The embedding model only decides *how*.

Retrieval quality is capped by segmentation quality, and no embedding-model upgrade recovers a bad split. Until now, chunking in MJ was a private helper inside whichever pipeline needed it — so every new strategy would have been another bespoke branch in someone else's engine.

This PR makes **cleaning** and **segmentation** registered, swappable strategies resolved through the MJ class factory — the same pattern `BaseEmbeddings` and `VectorDBBase` use for providers — wires ContentAutotagging onto them behaviour-unchanged, adds the chunk schema for multimodal segments, and teaches the external search providers to return media chunks.

## The two failure modes this addresses

| | What happens | Why it's invisible |
|---|---|---|
| **Straddling** | An arbitrary token boundary puts the tail of one topic and the head of the next in one chunk. Its vector is the average of two meanings. | It ranks mediocre for *both* queries — never first, so it never shows up as a failure. |
| **Mush** | A 60-minute recording embedded in one call. Multimodal embedders sample a bounded number of frames regardless of clip length. | The effective window is ~30s; the other 59 minutes are silently dropped. |

## Architecture

```mermaid
graph TD
    TC["<b>TextChunker</b><br/><i>ai-vectors</i><br/>'split this string to fit a budget'"]
    CL["<b>BaseContentCleaner</b><br/><i>ai-segmentation</i><br/>'which text is actually content'"]
    BS["<b>BaseSegmenter</b><br/><i>ai-segmentation</i><br/>'where does that content divide'"]
    CA["<b>content-autotagging</b>"]
    SE["<b>search-engine</b><br/>ExternalHitMapper"]

    CL --> BS
    BS -->|uses to enforce token ceiling| TC
    CA --> CL
    CA --> BS
    SE -.->|returns chunk provenance| CA
```

`ai-segmentation` sits **above** `ai-prompts` deliberately: `ai-prompts → templates → ai-provider-bundle → ai-vectors-pinecone → ai-vectors`, so `ai-vectors` importing `ai-prompts` is a genuine cycle. One layer up lets `SemanticText` use `AIPromptRunner` directly — versioned prompt metadata, real cost attribution — and keeps segmentation reusable by `ai-vector-sync`.

## Cleaning — the stage before segmentation

Segmenting dirty content produces well-bounded garbage. Navigation, sidebars, cookie banners and ads usually outweigh the article, and because that chrome repeats across every page of a site it yields many near-identical chunks that crowd out real answers.

| Key | Best for |
|---|---|
| `Html` | Markup. CSS-selector extraction with a sensible default exclusion list. |
| `PlainText` | Everything else — whitespace normalization, preserving paragraph breaks. |

**`IncludeSelectors` is the high-leverage knob**: naming the element that holds the content (`.article-body`, `main`) discards everything else without enumerating what to drop. Per-source, because the right selector is a property of the site's template. An invalid selector is skipped rather than failing the clean, and if cleaning would remove *everything* the original is returned with a warning — over-including beats silently dropping a document.

## Choosing a segmentation strategy

```mermaid
graph TD
    A{Timed transcript?} -->|yes| T["<b>Transcript</b><br/>chapters + StartMs/EndMs"]
    A -->|no| P{Paginated source?}
    P -->|yes| G["<b>PagedContent</b><br/>one segment per page"]
    P -->|no| B{Is there text?}
    B -->|no| F["<b>FixedWindow</b><br/>duration windows / single image"]
    B -->|yes| C{Headings present?}
    C -->|yes| S["<b>StructuralText</b><br/>sections + hierarchy"]
    C -->|no| D{Worth an LLM call?}
    D -->|yes| M["<b>SemanticText</b><br/>LLM topic boundaries"]
    D -->|no| AD["<b>AdaptiveBoundary</b><br/>target size, nearest natural break"]
```

| Key | Model calls | Best for |
|---|---|---|
| `StructuralText` | none | Markdown/HTML headings → sections with real hierarchy |
| `AdaptiveBoundary` | none | Prose without headings — target size, escalating boundary preference |
| `SemanticText` | 1/doc | Prose where topics shift with no structural signal |
| `Transcript` | none | Audio/video with cues → chapters, optional per-speaker sub-chapters |
| `PagedContent` | none | PDFs and slide decks, preserving `PageNumber` |
| `FixedWindow` | none | Universal fallback |

### Size chunks to your queries, not your model

The context window is an **upper bound, not a target**. A chunk should be about as much content as a good answer, so a matching chunk is mostly signal. `AdaptiveBoundary` makes this explicit: `TargetTokens` is the goal, `MaxSegmentTokens` only the ceiling. It escalates **paragraph → sentence → word → hard cut**, so sizes vary on purpose — a slightly short segment ending at a paragraph beats an exactly-sized one ending mid-clause. It also declines to split when the whole text is only modestly over target, avoiding one full chunk plus a context-free runt.

## Adding a strategy costs one method

```typescript
@RegisterClass(BaseSegmenter, 'SlideDeck')
export class SlideDeckSegmenter extends BaseSegmenter {
    public get Key(): string { return 'SlideDeck'; }
    public get SupportedModalities(): ContentModality[] { return ['text', 'image']; }

    protected async SegmentCore(params: SegmentationParams): Promise<RawSegment[]> {
        return findSlides(params).map((slide, i) => ({
            Modality: 'text', Title: slide.Title, Text: slide.Notes, PageNumber: i + 1,
        }));
    }
}
```

`BaseSegmenter` is a template method, so every strategy inherits validation, the token ceiling (splitting oversized segments while **preserving titles and rebasing offsets**), undersized merging, sequence numbering, `ParentIndex` → `ParentSequence` remapping *after* splits, cycle-safe depth, and provenance stamping.

## Multimodal: one vector, dual representation

```mermaid
graph LR
    SEG["<b>Media segment</b><br/>video 14:22–15:05"]
    SEG --> V["<b>Native vector</b><br/>→ multimodal index<br/><i>VectorRecordID stays 1:1</i>"]
    SEG --> D["<b>Description + Transcript</b><br/>→ chunk row columns"]
    D -.->|summary + structured fields| MD["vector metadata"]
    D -.->|lexical + external indexes| SR["search providers"]
```

**Which text is embedded, and which isn't** — a precise rule, because the columns look similar:

| Field | Embedded? | Purpose |
|---|---|---|
| `Text` | **Yes — it *is* the payload** | Fused with media into one vector for multimodal segments (image + caption → one embedding) |
| `Description` | No | Readable representation for agents, reranking, lexical search |
| `Transcript` | No | Verbatim record; lexically searchable |

That's what keeps `VectorRecordID` 1:1 with the chunk row, preserving the Chunk-Identity Contract that makes soft-delete and purge safe.

**Decided here:** ship native similarity **+ lexical** indexing of `Description`/`Transcript`; a description *vector* is a per-corpus opt-in added only after measurement shows paraphrase misses — and when warranted, as a **sibling text chunk row**, not a second vector on the same row.

> **Worth pricing first:** for speech-dominant archives the transcript carries nearly all retrievable meaning. Transcript-derived text chapters alone give full text→text semantics, hybrid FTS, and reranking at **zero multimodal embedding spend**.

## External search — media chunks become findable

All four providers (Azure AI Search, Elasticsearch, OpenSearch, Typesense) now share one `ExternalHitMapper` instead of four inline copies. It resolves snippets from `description`/`transcript` as well as `content`/`body`/`text` — so a media chunk returns readable text rather than an empty snippet — and recovers chunk provenance (`chunkId`, `modality`, `startMs`/`endMs`, `pageNumber`) into `RawMetadata`, which is what lets a hit deep-link to a moment in a recording or a page in a PDF. Names match across camel/Pascal/snake case and numeric strings are coerced, since those indexes are populated outside MJ.

## ContentAutotagging integration — behaviour-preserving

Both chunk sites route through one seam (`resolveSegmenterKey()` / `segmentTextForChunking()`), defaulting to `FixedWindow` with **each site's historical budget**. Nothing chunks differently today; the strategy is now swappable.

**The two sites keep separate budgets by design** — embedding chunks are sized to the embedding model (7500 tokens) and persisted; tagging chunks to the LLM context window (`InputTokenLimit / 1.5`) and transient. ⚠️ `chunkExtractedText` is now **async** (was sync); the only callers in-repo are its own tests.

## `TextChunker` bug fixes (no API change)

**Offsets were wrong for repeated text.** `buildChunkFromUnits` resolved each chunk's start with `indexOf` from position 0, so any recurring sentence made later chunks report the *first* occurrence — on an 86-char document, a chunk truly spanning **61–86 reported 0–86**. Offsets are provenance; this silently corrupted the link from a search hit back to its source passage. Now a single forward cursor (also O(n) instead of O(n²)).

**`chunkByFixed` could never terminate** when `OverlapTokens >= MaxChunkTokens`. Overlap is now capped at half the window with guaranteed forward progress.

## Testing

| Package | Result |
|---|---|
| `ai-segmentation` | **107 pass** |
| `search-engine` | **276 pass** |
| `content-autotagging` | **194 pass** |
| `ai-vectors` | **63 pass** |
| Build | **56/56 tasks** |
| `check:esm` / `check:claude-md` | clean |

Four bugs were caught by these tests during development: the non-terminating fixed-window loop, the same overlap-explosion class in the AV time-window path, the `ClassFactory` mock gap, and `BaseSegmenter` rejecting pages-only params.

## Documentation

New **[Content Segmentation Guide](https://github.com/MemberJunction/MJ/blob/claude/multimodal-embeddings-rag-o5hx8g/guides/CONTENT_SEGMENTATION_GUIDE.md)**, indexed in `guides/README.md` and cross-linked from `ai-vectors`, `ai-segmentation`, `TEXT_PROCESSING_GUIDE`, and `ContentAutotagging`.

## Follow-ups

- **#3287** — three deferred #3275 review items (@topherHunnewell)
- `EmbedContent` ingestion wiring (still no caller since #2834); un-skip empty-`.Text` items in `VectorizeContentItems`
- AI-suggested CSS selectors (@dray-mc's idea) — the config field it would write to now exists


AN-BC:
@dray-mc and @topherHunnewell working on this a bit more to ensure that we have a deeper/richer story for chunking for multi-modal functionality than we have now. This has dual pathway(getting LLM descr for each non-text chunk) and also smarter plug-in ideas for chunking built into the framework and we can choose chunking strategy by content source. Please share any input. 

This isn't 100% done, current version is docs only but will have implementation added shortly.

AN-BC:
@dray-mc and @topherHunnewell this is now ready for review and testing

AN-BC:
@cmbrock-BC meant to cc you

AN-BC:
## Sequencing with #3275

Cross-linked from #3275. This PR ships the segmentation framework standalone and **deliberately leaves `AutotagBaseEngine.ts` untouched** — #3275 rewrites that file (+653/−67), so a competing edit would collide in its most delicate functions.

Checked the overlap against #3275's actual diff:

| Function | In #3275's diff? | When it can be wired |
|---|---|---|
| `chunkExtractedText` (tagging path) | **0 occurrences** — untouched | **Now.** No conflict, no schema needed (tagging chunks are transient, never persisted) |
| `buildEmbeddingChunks` | touched | After #3275 lands |
| `buildChunksForBatch` | touched | After #3275 lands |

### Correction to an earlier assumption

I'd been describing the chunk schema as landing "with #3275." **It doesn't.** #3275 adds no migration — its config rides the existing `Configuration` JSONType. So the columns multimodal needs — `Modality`, `StartMs`/`EndMs`, `Description`, `Transcript`, `ParentChunkID`, and relaxing `Text` to nullable — are a **separate migration that doesn't exist yet**, to be authored on top of #3275.

### Remaining path to multimodal

1. ✅ **This PR** — segmentation framework + `TextChunker` fixes (no schema, no migration)
2. **Tagging-path integration** — wire `chunkExtractedText` to `ResolveSegmenter`; independent of #3275, no schema. Better tag quality from coherent sections rather than arbitrary windows
3. **#3275 lands**
4. **Chunk-schema migration** + CodeGen — the columns above
5. **Embedding-path integration** — `buildEmbeddingChunks` → segmenter, persisting modality/time/description
6. **`EmbedContent` wiring** — the multimodal embed call itself (still has no ingestion caller since #2834)

Steps 1 and 2 are unblocked today; 4–6 depend on 3.

---

AN-BC:
@topherHunnewell @dray-mc this is ready for review.

plans/multimodal-rag-semantic-chunking.md
```diff
@@ -0,0 +1,236 @@
...
+the schema, on top of #3275 rather than against it.
+
+---
+
+## 4. Remaining roadmap
+
+**4A — Chunk schema** (fold into #3275, one migration): `Modality`, `StartMs`/`EndMs`,
+`StartOffset`/`EndOffset`, `PageNumber`, `Description`, `Transcript`, `SegmentTitle`,
+`ParentChunkID`, and **relax `Text` to nullable**. `VectorRecordID` stays a single column —
+one chunk, one vector (§2.1) — with `Description`/`Transcript` as text hydrated at retrieval and
```
topherHunnewell: description/transcript seems valuable for the sake of text based retrieval, but seems like a lot to include in metadata, does it make sense to add this tot he ContentItemChunks for supporting retrieval?

topherHunnewell (approved):
huge fan of the generalized segmentation approach. on the multimodal side, I do think there is value in a text based searchable column living with the contentItemChunk whether it is a description or transcript, does not need its own vector. not a blocker for this work

dray-mc:
This is fantastic! Covers almost all of what we've done plus other options and multimodal... Definitely a few thoughts to share, wandered all over the place:

Our segmentation process is currently where we also do some cleanup. Things like removing scripts and tags from HTML content or focusing on specific selectors, dropping extra whitespace, and... that might actually be all we do universally. Think that fits here, too, but 'cleaning' rules might be something worth breaking out further if I didn't just miss the right place for it. LLMs reading the content and spitting out clean versions works great, but still takes some liberties when using cheaper models so we haven't adopted it at full scale, and instead we've had to try to use other methods to minimize cost while still getting the right content from messy sources. Allowing inputs to specify HTML selectors to include or exclude has been a solid win (focusing on the body is pretty obvious, but sometimes within that we might want to ignore advertising blocks or maybe we just want one element rather than all the navigation, sidebar, or other blocks of text that might be included). ...it's a bit of a pain for people to remember to set up properly, though, so before we make this customer-facing we'll have to build in having AI check a few pages and make suggestions.

-----------------------------

Our version of the fixed window attempts a few different strategies that I think are pretty generically useful. We have a target size - not based on the embedding model support, based on where we think we get the best retrieval, which means the size should be somewhat related to what we search on. Since our topic refinement tends to generate a paragraph of 2-5 sentences (could do longer now that models are faster and cheaper, but this immediately and directly hits performance) that means our chunks aim for around 1,200 words.
Based on that target size, as we get close - can't remember our thresholds, but let's say within 20% - we start looking for paragraph endings to segment on. So, we could have chunks as small as 20% less than our target size.
If we haven't found a paragraph ending by the time we hit our target, we start looking for sentence endings.
If we haven't found a sentence ending by the time we hit 20% over our target, we look for words.
If we haven't found words, we're also checking token counts and have an upper limit on how large we'll let a chunk be.

Then the overlap is added.

So, our chunks vary in size by aiming for a good point to separate. And if the full text is only a certain range larger (20-40% maybe) than our ideal chunk, we don't separate at all.

Part of the reason for all this is the data coming in is sometimes fairly wild. Spaces, new lines, sentence breaks, all seem a bit chaotic and our regex just hasn't always done what I expect it to.

The use of word count is unnecessary and should have been tokens all the way, but the escalating rules worked nicely.

-----------------------------

I believe the multimodal embedding models will often allow you to *include* text with your image/video, so there's technically an option for the text on a chunk to actually be a part of the single vector if I understand that correctly. Thinking there's value in like... an image and its caption to all be one embedding, maybe. I like having a separate description that explicitly is *not* part of the vector, but if you allow text to be embedding with the media that could be useful.

-----------------------------

I want to make sure the way we intend to populate ContentItems and their text is valid in this model. Think it is but... as we've discussed, our tools split the process of training on content into a few steps. We get a list of all the content we want to read, we read all the content, then we trigger the chunking/embedding. That second step populates the Text fields on ContentItem records with the full text content we consume, and we assume the chunking process will just work with that Text. I believe what this model allows us to do is, in the cases where we want to use the source file, we can store the reference but no text and the chunking process will work with the source file instead of the text we've extracted. Images and videos are the obvious cases, but this easily applies to PDFs where we want smarter chunking (not the thing we formally refer to as Smart Chunking, just maybe OCR or leveraging something like the Google embedding model's capabilities to embed a PDF directly, page by page) and maybe more down the line.

-----------------------------

I noticed the Segment Title field being added and just wanted to note we intentionally use the parent ContentItem title when displaying anything to users. I don't think there's a problem here and it could make some things we do a bit easier to have the option of a segment title in specific use cases, but when we find 10 chunks from the same document we aren't likely to show the 10 different segment titles. When we currently want to do that, we've intentionally broken them into the equivalent of different ContentItems (our Smart Chunking does that), but this will open the door to a cleaner approach, i think. Will also be interested in how we could leverage the chunk information to link users to specific points in a document or timestamps in a video, so pretty excited about that additional, structured data.

dray-mc:
> huge fan of the generalized segmentation approach. on the multimodal side, I do think there is value in a text based searchable column living with the contentItemChunk whether it is a description or transcript, does not need its own vector. not a blocker for this work

Searchable text means the other, non-semantic searches could find these items, too. ie. Azure search indexes in our case.

AN-BC:
@dray-mc @topherHunnewell — thank you both, this was genuinely useful feedback and **all of it is now implemented** rather than deferred. Point-by-point:

---

### @topherHunnewell — Description/Transcript as columns

You were right, and you were reading a doc line that undersold it. **They are `NVARCHAR(MAX)` columns on `ContentItemChunk`** — that's their home, in the migration on this branch. Only a *short summary + structured fields* (`SegmentTitle`, `StartMs`/`EndMs`, `Speaker`, `Modality`, `PageNumber`) go into vector metadata, which is size-capped and is filter/payload rather than ranked full-text. The doc line led with the metadata mirroring, which is what made it read as "a lot to put in metadata" — reordered to lead with the columns.

---

### @dray-mc

**1. Cleaning — you found a real gap, now a first-class plug-in.** We had no place to express "for this source, only `.article-body`" — `parseHTML` hardcoded `script, style, nav, footer, header, .hidden`. Now there's `BaseContentCleaner` alongside `BaseSegmenter`, with `Html` (CSS-selector driven) and `PlainText`. `IncludeSelectors` is the high-leverage knob exactly as you described — naming the element that holds the content discards nav/sidebar/ads in one stroke, rather than enumerating what to drop; `ExcludeSelectors` handles what survives inside it. Both are per-source config, because the right selector is a property of the site's template.

Two safety behaviours from your "regex hasn't always done what I expect" note: an invalid selector is **skipped rather than failing the clean**, and if cleaning would remove *everything*, the original content is returned with a warning — over-including beats silently dropping a document.

Kept it a separate stage from segmentation deliberately: they change for different reasons (a new CMS template needs new selectors, not a new chunking strategy), and cleaning once benefits embedding chunks, tagging chunks, and full-text indexing alike. Your "AI checks a few pages and suggests selectors" idea layers cleanly on top of this — the config field it would write to now exists.

**2. Your escalating fixed-window is better than mine — implemented as `AdaptiveBoundary`.** Target size → within tolerance look for **paragraph** → past target **sentence** → past overshoot **word** → hard ceiling → then overlap, plus your "don't split if the whole text is only 20–40% over" rule. Sizes vary on purpose, which was the point.

The principle I hadn't stated and have now written into the guide is yours: **size chunks to your queries, not to the model's context window.** Your ~1,200 words follows from 2–5 sentence topic-refinement queries; deriving it from `MAX_EMBEDDING_TOKENS` is how retrieval quietly goes wrong. And per your own note — it's tokens throughout, not words.

**3. Text fused with media — supported, and now stated explicitly.** You're right that multimodal models take text alongside media. A segment can carry both `Text` and `Media`, and `Modality: 'multimodal'` means exactly that. The rule is now spelled out because the columns look similar: **`Text` IS the embedded payload** (fused with media when present — image + caption as one vector); **`Description`/`Transcript` are never embedded.** That's what keeps `VectorRecordID` 1:1 with the chunk row.

**4. Your ingestion flow — valid, plus a page-level win.** Store a reference with no `Text` and chunking works from the source file: `SegmentationParams` takes `Media` + `MimeType`, and `Text` is now nullable. One thing to flag: `VectorizeContentItems` currently **skips empty-`.Text` items**, so that filter has to change or media items get silently dropped — tracked as part of the `EmbedContent` wiring.

Your PDF case got its own segmenter: **`PagedContent`** emits one segment per page with `PageNumber`, and a page can carry text, a rendered-page image, or both — so a page can be embedded *as an image* (preserving tables and charts that text extraction flattens) with its text riding along for lexical search. Pages carrying media are never merged, so provenance stays true.

**5. `SegmentTitle` — no behaviour change for you.** It's optional, used mainly to prefix embedded text so a chunk's vector carries its topic; display can stay on the parent ContentItem title. And your deep-linking interest is exactly what `StartMs`/`EndMs`/`PageNumber`/offsets are for.

**Your Azure Search point drove a fifth change.** Since `Description`/`Transcript` are columns, they're indexable by everything non-semantic. All four external providers (Azure AI Search, Elasticsearch, OpenSearch, Typesense) now share one `ExternalHitMapper` instead of four inline copies: it resolves snippets from `description`/`transcript` as well as `content`/`body`/`text` — so a media chunk returns readable text rather than an empty snippet — and recovers chunk provenance (`chunkId`, `modality`, `startMs`/`endMs`, `pageNumber`) into `RawMetadata`, which is what lets a hit deep-link to a moment in a recording. Names match across camel/Pascal/snake case and numeric strings are coerced, since those indexes are populated outside MJ.

---

**State of the PR:** ai-segmentation 107 tests, search-engine 276, content-autotagging 194, ai-vectors 63 — all passing; 56 build tasks green; `check:esm` and `check:claude-md` clean.

⚠️ **Still blocked on one thing:** the migration carries hand-written DDL only — the CodeGen output is **not yet appended**, and the file has a banner saying so. It must not be merged or applied until someone runs it against a v5.49.0 database and appends `mj codegen` output, since otherwise the schema moves ahead of MJ metadata.

---

AN-BC:
## 🔧 @topherHunnewell — CodeGen re-run needed (one task, detailed below)

cc @sohamdesai-BlueCypress @dray-mc

### Heads up: the generated section in the migration is now **stale**

@AN-BC ran CodeGen locally on Jul 26 (`7439860e0`) and appended the output. **After that**, this PR added four more columns to the hand-written DDL — so the generated section no longer matches the DDL above it.

Verified by grepping below the separator block:

| Column | In generated section? |
|---|---|
| `ContentItemChunk.*` (Modality, StartMs/EndMs, Description, Transcript, …) | ✅ covered |
| `ContentSource.SegmenterKey` | ❌ **missing** |
| `ContentSource.CleanerKey` | ❌ **missing** |
| `ContentType.SegmenterKey` | ❌ **missing** |
| `ContentType.CleanerKey` | ❌ **missing** |

Merging as-is would move the schema ahead of MJ metadata for those four: no `EntityField` rows, un-regenerated `vwContentSources`/`vwContentTypes`, and `spCreate`/`spUpdate` procs that don't know the columns exist. The file carries a banner saying exactly this.

**This is a replace, not an append** — the existing generated section must be swapped out wholesale, because a fresh CodeGen run re-emits the ContentItemChunk work too.

### Runbook

**1. Clean DB at the 5.50 candidate baseline**

Create an empty database and migrate it to the current head *excluding* this migration, so CodeGen sees only the pre-existing state:

```bash
mj migrate -t v5.49.0     # or the 5.50 candidate baseline you normally cut from
```

**2. Pull the branch and confirm ordering**

```bash
git fetch origin claude/multimodal-embeddings-rag-o5hx8g
git checkout claude/multimodal-embeddings-rag-o5hx8g
ls migrations/v5 | sort | tail -3
```

The file is `V202607281231__v5.50.x__ContentItemChunk_Multimodal_Segmentation.sql` — retimestamped so it sorts after `V202607241645`, the current head.

**3. Strip the stale generated section before applying**

Delete everything from the separator block (`==== BELOW THIS LINE: MEMBERJUNCTION CODEGEN OUTPUT ====`, ~line 310) to EOF, plus the ⚠️ banner at the top. You want the DB to see **hand-written DDL only** — otherwise the old generated inserts run against a schema they were built for and you'll get duplicate/conflicting metadata rows.

**4. Apply + CodeGen**

```bash
mj migrate                # applies the hand-DDL-only migration
mj codegen
```

**5. Re-append**

Take the produced `CodeGen_Run_*.sql`, append it beneath **50 blank lines** and the generated-output header block, then delete the standalone file. Layout reference: `V202607240220__v5.50.x__ContentItem_VectorRecordID_And_ContentItemChunk.sql`.

**6. Sanity check before pushing**

```bash
grep -c "'CleanerKey'" migrations/v5/V202607281231__*.sql   # must be > 0 now
```

Also confirm `SegmenterKey` appears for **three** entities (`MJ: Content Item Chunks`, `MJ: Content Sources`, `MJ: Content Types`) — it currently appears for one.

**7. Commit the CodeGen-emitted TS/HTML too**

CodeGen regenerates `packages/MJCoreEntities/src/generated/entity_subclasses.ts` and the Angular form components. Those belong in the same commit.

### What's in the DDL

`ContentItemChunk` gains `Modality`, `StartOffset`/`EndOffset`, `StartMs`/`EndMs`, `PageNumber`, `SegmentTitle`, `Description`, `Transcript`, `SegmenterKey`, `ParentChunkID` (self-FK), and **`Text` is relaxed to NULL** — the single blocker preventing a pure-media segment from persisting. `VectorRecordID` deliberately stays one column: one vector per chunk.

`ContentSource`/`ContentType` gain `SegmenterKey` + `CleanerKey`, mirroring the `EmbeddingModelID`/`VectorIndexID` resolution order (source overrides type, then a built-in default). Their *options* — CSS selectors, target sizes — ride the existing `Configuration` JSONType, whose interfaces are updated in `metadata/entities/JSONType-interfaces/`, so CodeGen will pick those up in the same pass.

---

### @sohamdesai-BlueCypress — please hold the 5.50 build

Targeting 5.50, but **don't cut it until Topher's CodeGen commit is in**. Right now `main`-bound state would ship those four columns with no MJ metadata behind them, which breaks the entity layer for `Content Sources`/`Content Types` rather than failing loudly. Once his commit lands, this is ready from a schema standpoint.

### @dray-mc

Everything from your review is implemented — details in the [comment above](https://github.com/MemberJunction/MJ/pull/3286#issuecomment-5099343473). The config surface your selector work needs (`CleanerKey` + `CleaningOptions` with `IncludeSelectors`/`ExcludeSelectors`) is part of what's in this CodeGen pass, so it'll be usable from metadata once Topher's run lands.

---

**Everything else is green:** ai-segmentation 107 tests, search-engine 276, content-autotagging 194, ai-vectors 63; 56 build tasks; `check:esm` + `check:claude-md` clean. The `hasPayload` bug is fixed and covered by the `PagedContent` suite.

---

PR #3290, opened 2026-07-27
fix(open-app): install-engine hardening — config injection, PG schema casing, shared deps, honest upgrade status

MS-BC:
Four independent install-engine fixes (one commit each), sequenced so the security fix is cherry-pickable on its own. This PR gates the bizapps manifest PRs — the PG canonicalization must land before any app reinstall/remove runs against a Postgres tenant.

## 1. RCE: manifest values were spliced into mj.config.cjs as raw strings (`f2809ed`)
`mj.config.cjs` is **executed** (`require`'d) by every `mj migrate`/`codegen`/`build`. The config writer interpolated manifest-sourced values (package name, startupExport, app name, schema name) between raw single quotes — a name containing `'` and `;` escapes the literal and appends arbitrary statements that run at the next `mj` command.
- `config-manager`: every manifest-sourced value is now `JSON.stringify`-escaped; removal/detection regexes accept both quote styles so legacy single-quoted entries stay removable.
- `manifest-schema` (defence in depth): package names + `schema.entityPackage` must be valid npm names (npm's charset, ≤214); `startupExport` must be a single JS identifier.
- New regression suite (`config-manager-injection.test.ts`) executes the written config the way `require` would and asserts hostile values survive only as inert string data.

## 2. PG schema-name canonicalization at read (`863a23e`)
PostgreSQL folds unquoted DDL identifiers to lowercase, so `__mj_BizAppsCommon` exists on PG as `__mj_bizappscommon` while stored metadata can carry either casing. Fix heals existing installs by comparing case-insensitively at **read** — never by rewriting stored rows:
- `HandleSchemaCreation` checks existence of the **canonical** name (the form `CreateAppSchema` actually creates) — raw-casing checks sent PG reinstalls into an "already exists" dead end.
- Legacy `RemoveAppEntityMetadata` path: `Schema Info` delete filter is now `LOWER()` both sides (matching the entity query + FK-graph predicate already on `next`).
- `CheckSchemaSharedByOtherApps` is now case-insensitive — previously two apps storing different casings of the same schema missed the co-tenant and a remove could CASCADE-drop a schema another app still lives in.

## 3. Shared npm deps survive app removal (`de5002b`)
`RemoveApp` already computed `otherManifests` for shared prebundle excludes but never told `RemoveAppPackages` — removing one app stripped package.json deps a surviving app still declares. New optional `RetainPackages` on `PackageManagerOptions`; the orchestrator derives the retained set from every other installed app's server/client/shared lists.

## 4. Upgrade finalizes Disabled when npm install fails (`358ccef`)
The upgrade path wrote `Status: 'Active'` unconditionally even with `npmInstallWarning` set. It now mirrors the install path: finalize `Disabled` + flip the app's `dynamicPackages` entries off so server loader/client bootstrap skip it until `npm install` + `mj app enable`.

⚠️ Note: #3272 touches the same files for a different fix (workspace layout) — whichever merges second should rebase.

---

## Testing

Unit: `packages/OpenApp/Engine` builds clean, **301/301 pass** (25 files), including the new `config-manager-injection.test.ts`, which `require`s the written config the way `mj migrate` does and asserts hostile manifest values survive only as inert string data.

Everything below is a **manual run of the real `mj app install / upgrade / remove` CLI** against real databases — no mocks, no unit-test harness. Three fixture Open Apps were published as GitHub repos and installed into the same tenant so the multi-app interactions (shared deps, co-tenant schemas) are real:

| fixture | schema | server packages | purpose |
|---|---|---|---|
| `mj-oa-alpha` | `oa_alpha` | `is-odd`, `is-even`, `left-pad` (bootstrap), **`@mj-oa-fixture/definitely-not-published`** | forces an npm-install failure on upgrade |
| `mj-oa-beta` | `oa_beta` | `is-odd`, `is-number` (bootstrap) | shares `is-odd` with the others |
| `mj-oa-gamma` | **`OA_Gamma`** (mixed case) | `is-odd` | exercises schema-name casing |

### SQL Server — `localhost:1477 / MJ_PR1`

Recorded in `__mj.OpenAppInstallHistory` (all `Success=true`): alpha Install 1.0.0 → beta Install → beta Remove → beta Install → beta Remove → **alpha Upgrade 1.0.0→1.0.1** → gamma Install → gamma Remove. Migrations really ran: `oa_alpha.flyway_schema_history` holds both scripts and `oa_alpha.Widget` gained its `Notes` column at 1.0.1 (`ID, Name, Notes`).

**Fix 3 — shared deps survive removal (the check that matters).** With `mj-oa-gamma` installed (declares `is-odd`), installing `mj-oa-beta` (declares `is-odd` + `is-number`) then removing it:

```
before beta install   is-odd=ABSENT    is-number=ABSENT
after  gamma install  is-odd=^1.0.0    is-number=ABSENT
after  beta install   is-odd=^1.0.0    is-number=^1.0.0
after  beta remove    is-odd=^1.0.0    is-number=ABSENT     <-- retained / removed correctly
```

`is-number` (beta-only) is gone; `is-odd` survives because a still-installed app declares it. Pre-fix, `RemoveAppPackages` was never told about the other manifests and stripped `is-odd` too, leaving the survivor unable to boot. `RetainPackages` is derived from every app with `Status != 'Removed'`, so a **Disabled** app's deps are retained as well — removing gamma afterwards correctly left `is-odd` in place for the Disabled alpha.

**Fix 4 — upgrade finalizes honestly.** Alpha's manifest carries an unpublishable package, so `npm install` warns during the 1.0.1 upgrade. The history row is `Action=Upgrade, PreviousVersion=1.0.0, Version=1.0.1, Success=true` and the app's terminal state is **`Status=Disabled`** — not the unconditional `Active` the old code wrote. The second half is observable too: while alpha sits installed-but-Disabled, none of its three packages (`is-odd`, `is-even`, `left-pad`) are wired into MJAPI's `package.json` — `is-odd` only reappeared when *gamma* was installed.

### PostgreSQL 17 — `localhost:5499 / MJ_OA_PG`

Fix 2 is a PG-specific bug, so it was exercised on a real PG server through the real `PostgreSQLDialect` and the real `SchemaExists` / `CreateAppSchema` / `DropAppSchema`, with a manifest declaring `schema.name = "OA_MixedCase"`:

```
Dialect.CanonicalSchemaName() : oa_mixedcase
FIRST INSTALL          -> {"Success":true,"Created":true,"Branch":"create"}
information_schema says-> schema_name = ["oa_mixedcase"]          <-- PG folded the DDL identifier

SchemaExists('OA_MixedCase')  [PRE-fix, raw manifest casing] -> false
SchemaExists('oa_mixedcase')  [POST-fix, canonical]          -> true

REINSTALL  PRE-fix  -> {"Success":false,"ErrorMessage":"Schema 'OA_MixedCase' already exists","Branch":"create"}
REINSTALL  POST-fix -> {"Success":true,"Created":false,"Branch":"reuse-existing"}
```

That is the dead end exactly: the raw-casing existence check says "absent", so `HandleSchemaCreation` takes the create branch, and PG then rejects the create because the schema is in fact there. The reinstall cannot complete. Checking the canonical name reuses the existing schema instead.

The removal-path filter change was verified on the same server with a real mixed-case schema present:

```
"SchemaName = 'OA_MixedCase'"               -> 0 row(s)
"LOWER(SchemaName) = LOWER('OA_MixedCase')" -> 1 row(s)
```

PG's `=` on text is case-sensitive, so the strict form silently matched nothing and left the `Schema Info` row orphaned after removal. `DropAppSchema` called with the manifest's casing then dropped the schema and `SchemaExists` returned false.

### Not covered

Fix 1 (config injection) is proven by the executing regression suite rather than a live tenant — the hostile value never has to reach a database to demonstrate the escape. The bizapps-common install/remove/reinstall on a real Postgres tenant named in the release checklist is still worth running during certification; these fixtures reproduce the mechanism, not that tenant's data.


jordanfanapour (approved):
Changes look good to me and I'm assuming that you tested app installation and removal with the updated changes. Please proceed with merging to next if all manual testing looks good.

PR #3291, opened 2026-07-27
feat(integration): durable sync runs — lease/fence ownership, DB-backed cancel + progress, opt-in worker mode

MS-BC:
Implements **tasks.md PR 1 — durable sync runs**. A `CompanyIntegrationRun` is now an owned, leased, fenced unit of work that survives a process death, can be cancelled and observed from any process, and can be executed by a worker rather than the request that asked for it.

Two commits: `9cdf67f` is the schema half (columns, index, `Queued` status, claim/renew/release sprocs, `RSUPendingWork` table + CodeGen output from a fresh DB). `83fb9e0` is the runtime half.

## What changed

**Ownership — `RunOwnershipService` (new)**
- **Claim** is a single atomic `UPDATE … WHERE (unowned OR lease expired)` that bumps `FenceToken`. Never select-then-update. Zero rows back means another process owns the run, and we do not proceed.
- **Renewal** runs on a **timer at ~lease/3**, not in the batch loop, so a long batch never looks dead. The lease is `max(default, MaxRuntimeMinutes)` — a configured long run only ever *extends* protection, never shrinks it.
- **The fence precedes writes.** Every batch boundary re-checks owner token *and* fence before anything is written, so a run whose lease was reclaimed mid-batch writes nothing further.
- Terminal release and progress writes are ownership-guarded, so a stale holder can never clobber the new owner's row.

**Cancel and progress moved into the database**
- The in-memory cancel map is gone. `CancelSync` stamps `CancelRequestedAt`; the heartbeat's renewal result doubles as the cross-process cancel poll, so a cancel issued on one node stops a run on another.
- Progress is persisted to `ProgressJSON` (throttled to 5s) so a reader in *any* process sees it. The snapshot still only ratchets upward.

**Per-run context**
- Provider, abort controller and progress snapshot live in an `AsyncLocalStorage` scope per run. The engine's shared `_provider` field is gone — see the scoping note under "What this does and doesn't fix" for exactly what that buys.

**RSU pending work is durable**
- `.rsu_pending` JSON files that the consumer deleted as it read them are replaced by rows in `MJ: RSU Pending Works`. A row is registered only for migrations that succeeded, before the restart, and is marked `Completed` only **after** the consumer's work actually succeeds. Failures record why and stay visible; anything older than 30 minutes is named at boot.

**Worker mode (opt-in, disabled by default)**
- `EnqueueSync` writes `Status='Queued'` carrying trigger type + options on `ConfigData`, returning a real RunID immediately — which also removes the recency-poll dance callers previously needed to learn their own RunID (new `IntegrationEnqueueSync` mutation).
- `IntegrationSyncWorkerService` polls for claimable rows with bounded concurrency and a reentrancy guard; `ExecuteQueuedRun` claims, promotes to `In Progress`, executes, releases. Losing a claim race is ordinary traffic, not an error. Stopping the loop leaves in-flight runs alone — their leases are still being renewed.
- Config block `integrationSyncWorker` defaults to `enabled: false`, so existing deployments are unaffected until they opt in.

## Latent bugs this surfaced

Typing the pending-work payload (it previously went straight through `JSON.stringify`, so nothing checked it) caught two unvalidated strings in the resolver: `SyncScope` and `SyncDirection` were being passed through as `string` where the contract is a literal union.

## Live verification — SQL Server

Run against a real SQL Server (`MJ_PR1`, the migration applied from a fresh DB), driving the real sprocs and the real `RunOwnershipService` through a real `SQLServerDataProvider`. **57 live assertions, 0 failures.**

**1. Sproc semantics, raw SQL — 10/10.** Fence starts at 0. First claim wins and sets fence 1. A second claim against a live lease returns zero rows and leaves the owner untouched. Renew persists `ProgressJSON`; a renew with `NULL` progress does *not* wipe it (the `COALESCE` holds). A cancel stamped by a different connection is visible on the next renew. An expired lease is reclaimed and the fence goes to 2. The old holder's renew is rejected at both the old fence and the new one, and writes nothing. The old holder's release is a no-op that leaves status and owner intact. The real owner's release returns a row, clears owner + lease, and stamps `EndedAt`.

**2. Concurrency — 12 parallel connections claiming one row, 15 trials, 0 failures.** Every trial: exactly one winner, fence advanced by exactly 1. This is the property the whole design rests on, and it is the one an in-process test cannot show.

**3. `RunOwnershipService` through the real provider — 28/28.** This is what proves the `@RunID=@p0` placeholder convention actually binds against a live driver, which raw `EXEC` cannot. Includes the `kill -9` case: expire the lease, a second service adopts the run, fence advances exactly once (18 → 19), the original owner's `CheckBoundary()` flips to `Owned: false`, its `Renew()` is rejected and writes nothing, and its `Release()` returns false leaving status untouched. Also covers `max(default, MaxRuntimeMinutes)` in both directions (45 → 45, 2 → 10).

**4. Timer-driven renewal — 4/4, measured over 220 s.** Claimed a run with a 10-minute lease (heartbeat interval = 200 s), started the heartbeat, then idled with **zero loop activity** for interval + 20 s. `HeartbeatAt` advanced, the lease was pushed past its original expiry, the run was still owned, `onLost` never fired, and the progress payload was piggybacked onto the renewal. This is the "long batch doesn't look dead" claim, tested the only way that means anything — by doing nothing for longer than a heartbeat.

**5. RSU durable pending work — 15/15.** A row is registered `Pending`; a read returns it *without* consuming it (the behaviour the `.rsu_pending` files got wrong); a second read after a simulated crash mid-consumption still returns it; the stale sweep names it; `CompletePendingWork` stamps `ProcessedAt` and removes it from the pending set; `FailPendingWork` records the reason and stops it being retried forever.

## What this does and doesn't fix

The `AsyncLocalStorage` change is worth stating precisely, because the earlier draft of this description overstated it.

**Fixed:** the engine is a singleton and previously kept a mutable `_provider` field that each run assigned on entry. Two concurrent runs meant last-writer-wins — run B's provider silently became run A's provider mid-run. That field is gone; each run now resolves its provider from its own ALS scope, so runs can no longer clobber each other.

**Not yet true:** "each of N concurrent syncs on its own connection." The per-run provider is whatever the caller passes, and all three production callers (`index.ts:1660`, `IntegrationDiscoveryResolver.ts:3750` and `:4000`) plus the worker (`IntegrationSyncWorkerService.ts:122`) pass nothing, so today every run still resolves `Metadata.Provider`. This PR makes per-run connections *possible* and removes the cross-run clobbering; actually handing each run its own connection is a follow-up at the call sites.

## Tests

| Package | Result |
|---|---|
| `@memberjunction/integration-engine` | 641 passed (37 files) |
| `@memberjunction/schema-engine` | 94 passed (4 files) |
| `@memberjunction/server` | 735 passed, 56 skipped |

The one MJServer failure is `mjapi-bootstrap.test.ts`, a vite package-entry resolution artifact for `mj_generatedentities` in a local worktree. That file is not touched by this PR and imports nothing from the sync path; CI is green on the branch.

The four engine suites that mock `Metadata.Provider` now share an ownership provider surface so their mocks answer the claim/heartbeat/fence statements.

## Scope

**PostgreSQL is deliberately not in this PR.** T-SQL lands here; the PG conversion is a separate downstream pass, as with every other migration on this line. Nothing above claims PG behaviour, and none of it was tested there.

---

## Live verification on a real instance (2026-08-04)

Everything above this section was unit-level. Worker mode has now been run end-to-end against a **real MJAPI process** (v5.50.0, port 4110) talking to a **real SQL Server database** (`MJ_PR1`), with `integrationSyncWorker.enabled = true`, `pollingIntervalMs: 3000`, `maxConcurrentRuns: 2`. The `CompanyIntegration` under test is `B0000000-…-0000000C1000` (File Feed). Mutations were issued over GraphQL with `x-mj-api-key`; every assertion below reads the resulting row straight out of `__mj.CompanyIntegrationRun` with `mssql`, not through the API that wrote it.

**Result: 13/13.**

| What was proven | Observed |
|---|---|
| `IntegrationEnqueueSync` returns a real RunID without executing anything | RunID `36BEE8AE-…` returned in **291 ms** |
| the row is persisted as queued work, not a started run | `Status='Queued'`, `ConfigData={"triggerType":"Manual","options":null}`, `OwnerToken=null`, `FenceToken=0` |
| a **different** process picks it up with no further request | worker claimed it ~6 s later, out of band |
| the claim fences the run | `FenceToken` advanced `0 → 1`, `Status` promoted off `Queued` |
| the run terminates and the lease is given back | `EndedAt` stamped; `OwnerToken` and `LeaseExpiresAt` both cleared |
| cancel is a database fact, visible to any process | `IntegrationCancelSync` stamped `CancelRequestedAt` on the live row |

### A defect this found, and the fix

The live run exposed a real hole. `CancelSync` stamps `CancelRequestedAt` on rows in **both** `'In Progress'` and `'Queued'` — but the only consumer of that stamp was the running loop's batch-boundary / lease-renewal check. A queued run has no loop yet, so nothing looked at the stamp before starting:

```
run E3F51F9A-132D-41FC-924C-A336DB4EF413
  CancelRequestedAt  15:50:46.643   ← cancelled while still Queued
  EndedAt            15:50:48.646
  Status             Success        ← ran anyway, 2s later
```

The cancellation was silently ignored, and the run history shows a clean success.

Fixed in `1f135cc`: `ExecuteQueuedRun` now checks `CancelRequestedAt` after loading the row and refuses to execute, finalizing it exactly the way `FinalizeRun` finalizes an aborted in-flight run — `Status='Failed'` with an explicit `ErrorLog` (`CompanyIntegrationRun` has no `'Cancelled'` status value), `EndedAt` stamped so the row is terminal and the poll cannot hand it back to a worker.

Re-verified live after the fix — **9/9**:

```
enqueued C710CC63-5197-45CF-B3DD-5D702F296E4C
cancel   → "Sync cancellation signalled — will stop after current batch completes"
row at cancel time      Status='Queued'   CancelRequestedAt=15:55:10.170
row ~3s later           Status='Failed'   EndedAt=15:55:13.122
                        ErrorLog='Sync cancelled by user before it started'
                        OwnerToken=null   TotalRecords=0
CompanyIntegrationRunDetail rows for this run: 0   ← the sync body never executed
```

The zero run-details is the load-bearing assertion: the run didn't merely get relabelled, it never ran.

A unit pin accompanies the fix (`ExecuteQueuedRun` refuses a cancelled queued row and finalizes it). Engine suite is now **642 passed / 37 files**.

**Known gap, stated plainly:** the connector under test (File Feed) completes with zero batches, so this exercised the *claim-time* cancel gate, not the batch-boundary cancel path inside a long-running sync. That path remains unit-covered only.


MS-BC:
Base synced to current `next` and conflicts resolved manually.

- `packages/SchemaEngine/package.json` — conflict was the 5.50.0 → 6.0.0 version bump landing on `next` against this branch's added dep. Resolved to next's 6.0.0 for every `@memberjunction/*` dep while keeping the branch-added `@memberjunction/queue` (also at 6.0.0).
- Migration renamed for the 6.0.0 gate: `V202607300300__v5.51.x__Durable_Sync_Runs.sql` → `V202608041910__v6.1.x__Durable_Sync_Runs.sql` (content unchanged).

Manually re-ran SchemaEngine locally post-merge: **94 passed**.

PR #6, opened 2024-01-31
Feature: Populate MJAPI/config.json with sample data

JS-BC:
Since we cannot add comments to json files, providing sample data should help developers get a better understanding on how to populate certain fields when making changes to the config file. This change comes from a result of some users of MJ having difficulties in automatically adding new users to the DB due to some configuration related errors

PR #3293, opened 2026-07-27
fix(integration): apply-phase correctness + hot-path cost

MS-BC:
Six defects in the integration apply phase. All six are silent — they produce wrong data or wasted work rather than an error — so each ships with tests that pin the behaviour rather than the implementation.

> **Rebased 2026-07-27.** The `RunView` speculative-`COUNT(*)` change that used to be item 6 here has been split out to #3298 and is **not** in this PR any more. It changes what `TotalRowCount` means product-wide across ~125 unaudited call sites and should not go into 5.50 (the first LTS candidate). What remains is correctness-only, scoped to `@memberjunction/integration-engine` + `@memberjunction/schema-engine`.

### Apply correctness (`@memberjunction/integration-engine`)

**1 — Content-unchanged records skipped their tombstone repair.** A record marked deleted in MJ that is alive again externally hashes identically on every subsequent sync, so it stayed tombstoned forever. The unchanged path now runs the sync-state repair before returning.

**2 — Record-map loads stopped at the entity row cap.** `RunView` without `MaxRows` is not unbounded; it falls back to `UserViewMaxRows` (default 1000). Two callers read the result as "every mapping that exists":
- the orphan sweep silently stopped cleaning past row 1000;
- the full-push path reads a missing mapping as *never pushed*, so it re-creates records that already exist in the customer's external system.

`LoadAllRecordMaps` now walks the map with `AfterKey` keyset paging (`OrderBy: 'ID ASC'`, `IgnoreMaxRows`, `BypassCache`, `MAX_PAGES` backstop) and returns `Complete`. A caller that did not get the whole map refuses to act rather than acting on a partial one — the sweep logs `ORPHAN_SWEEP_SKIPPED`, the push throws. Keyset rather than `StartRow` because OFFSET paging skips rows on PostgreSQL, where `gen_random_uuid()` keys are random rather than monotonic and a concurrent insert can land before the cursor.

**3 — Identity was decided by two competing rules.** The entity primary key — soft PKs included — is now the single definition of record identity for both matching and saving. `IsKeyField` stays a *fallback lookup* for records whose mapped data does not carry the PK; it is deliberately not AND-ed in when a complete PK is present, which is what made two records with the same PK but differing key fields resolve as distinct.

**4 — Record-map writes were a statement pair per record on the hot path.** New `RecordMapBatch` chunks them at 500 with a dialect-portable row source (`WITH src AS (SELECT … UNION ALL …)` — `VALUES (…) AS v(col)` is not portable between the two dialects, and the record-map table has no physical PK/UNIQUE, so neither `MERGE … ON` nor `ON CONFLICT` is available).

The hard requirement is that batching must not cost per-row error attribution. It doesn't: each chunk is read back and every missing or value-mismatched mapping is attributed to its own external ID; a chunk that throws replays row by row; a read-back that itself fails reports *every* row unverified rather than claiming success. Mappings queued by a batch that rolled back are discarded, never written. `Queue()` deliberately does not auto-flush on a full chunk — it is called from inside the apply pass's batch transaction, where a flush would write rows `Discard()` could no longer take back on rollback and that the read-back would "verify" while reading uncommitted state through the same connection. The apply loop flushes once per batch, after commit.

**5 — Identity lookups were one query per record.** Now prefetched per field-set group. The correctness constraint is that a batched lookup must never turn a match into a miss (that creates a duplicate), and local key comparison can legitimately disagree with the database's. So the batch indexes external IDs both exactly and collation-folded (lowercased, trailing blanks stripped) — the single-query path let SQL Server compare, which is case- and trailing-blank-insensitive under a default collation, while pairing rows in JS with `===` would read `AB-100 ` against a stored `ab-100` as unmapped. A folded key that more than one distinct mapping collapses onto is marked ambiguous and deferred to the per-record query rather than guessed at, as is an ID that was never in the batch's `IN (…)` list. Absence is concluded **only** from a completely empty group read.

**6 — `verifyChunk` hand-rolled its own escaping** (`replace(/'/g, "''")`) while the write path used `Dialect.QuoteStringLiteral`. Both use the dialect now.

Batch sizes are configurable and clamped: `MJ_INTEGRATION_RECORD_MAP_CHUNK_SIZE` (default 500, ceiling 5,000 — the upsert inlines one `SELECT` per mapping, so generated SQL grows linearly with the chunk) and `MJ_INTEGRATION_RECORD_MAP_PAGE_SIZE` (default 10,000, ceiling 50,000). Out-of-range values are clamped with a logged warning, not accepted.

### Retry safety (`@memberjunction/schema-engine`)

**7 —** `RunPipelineWithRetry` replayed the whole pipeline on any step failure. When the failure was downstream of `ExecuteMigration`, the retry re-executed DDL that had already committed — the retry breaks the schema, not the original error. Retry is now gated on both (a) the failed step being one of the four genuinely replayable ones and (b) no migration having committed in the run. The batch path gets the same guard keyed on `SuccessCount > 0`; `SchemaBuilder` and the `IntegrationApplyAll`/`…Batch` resolvers route through the guarded entry points.

## Tests

| package | result |
|---|---|
| `@memberjunction/integration-engine` | **579 pass** / 35 files |
| `@memberjunction/schema-engine` | **80 pass** / 3 files |

New: `RecordMapBatch.test.ts`, `IntegrationEngine.recordmap-paging.test.ts`, `IntegrationEngine.run-retention.test.ts`, plus batched-lookup, tombstone-repair and retry-safety cases.

`MatchEngine.test.ts` was rewritten off call-index mocking onto entity-routed, filter-aware mocks. The old style asserted "the 3rd RunView call had this filter", which broke the moment prefetching changed the call order — the tests now assert what was asked of which entity, which is the actual contract.

## What to interrogate in review

Items **4** and **5** are performance rewrites of the write path, not correctness fixes. They are the only real regression surface here, and both are deliberately conservative (per-row attribution with row-by-row replay; absence only ever concluded from a read that can prove it). They are kept together with item 3 because item 5 shares the `MatchEngine` identity changes.

Items 2, 4 and 5 are about behaviour at volume against a real database; the unit tests pin the logic but not the SQL. Worth exercising on a **Postgres** tenant: a >1000-mapping entity through both the orphan sweep and a full push, and a batch write where one row's external ID is pathological (quotes, unicode) to confirm attribution names that row.

---

## Testing

Run against a **real SQL Server** — `localhost:1477 / MJ_PR1`, 383 entities, CodeGen-generated `spCreate*`/`spUpdate*`, driven through the real `SQLServerDataProvider`. Not a unit test: every read and write below is a round trip to the database, and the record-map table is the real one. **32 checks, 0 failures.**

Seed: one entity with **1,500** record-map rows (deliberately past the 1,000 `UserViewMaxRows` cap), plus a set of pathological external IDs.

**A — the truncation is real, and silent.** An unpaged `RunView` over the map returned `Success=true, rows=1000, TotalRowCount=1500` — no error, no warning, a caller cannot tell. `LoadAllRecordMaps` returned `Complete=true, rows=1500`. Forcing an 11-page walk (`pageSize=137`) returned the same 1,500 rows with 0 duplicates, so keyset paging neither skips nor repeats.

**B — the consequence.** Resolving the 1,500 records against the truncated map produced `Create=500, Update=1000` — 500 records that already exist externally would have been **re-created in the customer's system**. With the complete map: `Create=0, Update=1500`.

**C — batched writes against real sprocs.**
| check | result |
|---|---|
| first flush of 400 new mappings | `spCreate=400 spUpdate=0 failures=0`, rows 1500→1900, 2502 ms |
| re-flush, nothing changed | `spCreate=0 spUpdate=0 rowsTouched=0`, 9 ms |
| one mapping changed | `spUpdate=1`, `touchedID=NEW-000007`, other 399 untouched |
| 4 pathological external IDs (quotes, unicode, trailing blanks) | `spCreate=4 readBack=4` |
| re-flush of those 4 | `spCreate=0` |
| `Discard()` after a simulated rollback | `spCreate=0`, `rowsInDB=0` — queued rows never written |

**C5b — why the dialect escape matters, measured on the server.** For `ünïcödé-Ω-日本語`, a bare `'…'` literal **matches 0 rows** and reads back as `ünïcödé-O-???` (varchar under `SQL_Latin1_General_CP1_CI_AS`); `N'…'` matches 1. **C5c** — the per-record fallback resolved the ambiguous case to exactly one row.

**D — batched identity lookup vs. the database's own comparison.** `EXT-000001 / 000500 / 001499` each resolved to Update against the right MJ record; `EXT-NOPE-*` to Create. The three cases where JS `===` and SQL Server disagree — `"ext-000001"` (case), `"EXT-000002  "` (trailing blanks), `"ünïcödé-Ω-日本語"` — all resolved to **Update**, i.e. the batch did not turn a match into a duplicate. A 600-record batch issued **1** record-map read and produced `Update=600`.

**E — agreement + sweep.** One-at-a-time resolution agreed with the batched result on every record. The orphan sweep saw **100** orphans past row 1000 with the complete map and **0** with the truncated one — the pre-fix sweep silently under-cleans rather than failing.

Not covered here: PostgreSQL. The keyset choice is specifically for PG's random `gen_random_uuid()` keys, and the >1000-mapping push + sweep should be repeated on a PG tenant during certification.


jordanfanapour (commented):
## Review: fix(integration): apply-phase correctness + hot-path cost

This is a substantial, well-structured PR. The seven defects are clearly diagnosed, the fixes are logically sound, and the test coverage is strong — tests assert behavioral contracts rather than implementation details, which is the right approach for this kind of work. A few findings below, roughly ordered by importance.

---

### 1. Hand-rolled escaping in `buildRecordMapViewParams` contradicts fix #6

`MatchEngine.buildRecordMapViewParams` builds the `IN` list with:
```typescript
const inList = externalIDs.map(id => `'${id.replace(/'/g, "''")}'`).join(',');
```

Fix #6's stated goal was to stop hand-rolling escaping and use `Dialect.QuoteStringLiteral` everywhere — `verifyChunk` was corrected, but this new code reintroduces the same pattern. I understand `MatchEngine` operates through `RunView` (not `ExecuteSQL`) and doesn't have dialect access, but the inconsistency should at minimum be documented with a comment explaining why the dialect can't be used here.

`CriteriaToSQL` has the same pattern with field values (`criteria.Values[i].replace(/'/g, "''")`), plus it doesn't escape double-quotes in field names (theoretical — MJ column names never contain `"`, but worth noting).

---

### 2. `normalizeExternalID` strips more whitespace than SQL Server does

```typescript
function normalizeExternalID(id: string): string {
    return id.replace(/\s+$/, '').toLowerCase();
}
```

`\s+$` strips trailing tabs, newlines, carriage returns, etc. SQL Server's trailing-blank insensitivity under `SQL_Latin1_General_CP1_CI_AS` only applies to spaces (ASCII 0x20). A stored external ID like `abc\t` (trailing tab) would normalize differently in JS vs. SQL Server, creating a false match in the batch index that the per-record query wouldn't confirm.

Very unlikely in practice for external IDs, but `/ +$/` would be more precise and wouldn't cost anything.

---

### 3. `CriteriaKey` has a theoretical collision risk

The key is built by concatenating `${fieldName} ${value}` pairs with no inter-pair separator:
```typescript
.map(([f, v]) => `${f} ${v}`)
.join('')
```

Fields `A` with value `BC` and fields `AB` with value `C` would both produce `A BC` / `AB C` depending on concatenation order, but a pathological case like field `A B` with value `C` and field `A` with value `B C` would collide: both produce `A B C`. A `\0` separator between pairs would eliminate this class entirely.

---

### 4. `verifyChunk` throwing loses failure attribution

If `verifyChunk` throws (rather than returning `Success: false`), the exception propagates through `flushChunk` → `Flush` → `FlushRecordMaps`. The wrapper catches it, but `TakeFailures()` returns nothing for that chunk because failures were never recorded. The rows in that chunk would be silently unattributed.

`RunView` returning `Success: false` is handled correctly — it's only an actual throw (network error, connection drop) that would hit this. Consider wrapping `verifyChunk` in a try/catch inside `flushChunk` that attributes all chunk rows as unverified on throw.

---

### 5. `needsSyncStateRepair` triggers on first-ever sync

```typescript
if (has('__mj_integration_SyncStatus') && entity.Get('__mj_integration_SyncStatus') !== 'Active') return true;
```

When `SyncStatus` is `null` (never synced), `null !== 'Active'` is `true`, so every record's first sync triggers a "repair" write. Likely intentional (the first sync should set the status), but it means the unchanged-content shortcut is defeated for every record on a fresh entity's initial sync. Worth confirming this is the desired behavior.

---

### 6. `Queue()` deduplication is O(n) per call

`Queue()` uses `findIndex` to check for duplicates. At `ChunkSize = 500` this is fine. At the configurable ceiling of 5,000, it's up to 12.5M comparisons across a full chunk. A `Map` keyed on `EntityID|ExternalID` would be O(1) per call.

---

### 7. Test gaps (minor)

All test files are well-structured and test behavioral contracts. A few boundary cases aren't pinned:

- **No chunk-boundary test** in `RecordMapBatch.test.ts` — a queue of `ChunkSize + 1` items should produce two `flushChunk` calls. Currently only tests queues at exactly `ChunkSize`.
- **No exactly-PAGE_SIZE boundary test** in paging tests — a result set that's an exact multiple of `PAGE_SIZE` should issue one more empty-results page and still return `Complete: true`.
- **No `failedStep === undefined` test** in retry safety — the code handles it (`!failedStep` returns false), but no test pins that behavior.
- **No unicode/pathological string test** in `RecordMapBatch.test.ts` — the `o'brien` test covers single quotes but not emoji, CJK, or null bytes. The PR description itself flags this as worth exercising against a real database, so this is fine for unit tests.

---

### 8. Batch retry `failedStep` selection

`RunPipelineBatchWithRetry` takes `result.Results.find(r => !r.Success)?.ErrorStep` — the first failed result in array order. In a batch with failures at different steps, the retry decision is based on whichever happens to come first, not the most critical. The `SuccessCount > 0` guard catches the truly dangerous case (committed DDL), so this is safe but slightly arbitrary.

---

### Things done well

- **`upsertChunk` SQL** uses `Dialect.QuoteStringLiteral` and `Dialect.QuoteIdentifier` consistently — the most sensitive injection surface is handled correctly
- **Queue/Discard/Flush lifecycle** correctly prevents writing uncommitted state — the decision not to auto-flush is well-reasoned and documented
- **Keyset paging** in `LoadAllRecordMaps` is correct — proper termination, `MAX_PAGES` backstop, `BypassCache`, partial-failure returns `Complete: false`
- **Identity rule unification** is clean — complete PK present → match on PK alone, partial/absent → key fields with dedup
- **Ambiguity detection** in collation folding is correctly conservative — ambiguous folds defer to per-record queries rather than guessing
- **Error attribution** (verify + replay) ensures no failure is silently swallowed, and all-unverified on read-back failure is the safe direction
- **MatchEngine.test.ts rewrite** from call-index mocking to entity-routed mocks is a clear improvement — tests now assert *what was asked of which entity*, not *the 3rd RunView call*
- **Retry guard** in `RuntimeSchemaManager` is correct — gates on both replayable step AND no committed migration
- **`.Get()` usage** in `needsSyncStateRepair` is properly justified (runtime-created integration tables have no generated entity type)

MS-BC:
Thanks — this is a careful read and six of the eight are now in, pushed in 088c517c.

**1. Hand-rolled escaping.** Taken, and not just documented — `MatchEngine` does have dialect access, I just wasn't using it. `ProviderToUse` is an `IMetadataProvider`, but `DatabaseProviderBase` is an abstract class, so a server-side provider narrows with a plain `instanceof`. Added `quoteLiteral()`:

```typescript
private quoteLiteral(value: string): string {
    const provider = this.ProviderToUse;
    return provider instanceof DatabaseProviderBase
        ? provider.Dialect.QuoteStringLiteral(value)
        : `'${value.replace(/'/g, "''")}'`;
}
```

The fallback exists because `IMetadataProvider` in general carries no dialect — a client-side provider has none — and ANSI quote-doubling is what both supported platforms do anyway. All three sites route through it now: the `IN` list, `CriteriaToSQL`, and the per-record record-map query. That last one was the important part of your point: the read path and `RecordMapBatch`'s write path were escaping the same values by two different rules, which is the thing that eventually disagrees.

The unescaped double-quotes in field names I left alone — `QuoteIdentifier` would be the fix, but `CriteriaToSQL` emits into a `RunView` `ExtraFilter` where the identifier quoting is the provider's, not mine, so changing it there is a wider change than this PR should carry.

**2. `normalizeExternalID`.** Taken — `/ +$/`. You're right that it's ASCII 0x20 only, and the direction of the error matters: folding a trailing tab makes the batch index equate two IDs the database does not, so the index would hand back a match the per-record query would never have made. A missed match costs a query; a wrong one writes to the wrong row.

**3. `CriteriaKey` collision.** Taken — `\0` both between and within pairs. `A B`/`C` vs `A`/`B C` is exactly the case, and since this key decides which read-back row answers which record, a collision is a mis-attribution rather than a miss.

**4. `verifyChunk` throwing.** Taken. Worse than losing attribution for that chunk, actually — the throw unwound through `Flush`'s chunk loop, so every *later* chunk was skipped and reported as written too. Now:

```typescript
try {
    await this.verifyChunk(entityID, chunk);
} catch (err) {
    this.markChunkUnverified(entityID, chunk, err instanceof Error ? err.message : String(err));
}
```

`markChunkUnverified` is extracted from the existing `Success: false` branch, so both paths report identically.

**6. `Queue()` O(n).** Taken — `pending` is a `Map` keyed `EntityID|ExternalID`, which is the upsert's own key, so it dedups and preserves order (JS Maps iterate insertion-order) with no scan.

**7. Test gaps.** Three added; skipping the unicode one for the reason you gave.

- `ChunkSize + 1` produces two statement pairs, and asserts the overflow row is in the *second* — a split that silently dropped it would otherwise still show two pairs.
- An exact `PAGE_SIZE` multiple pays for one empty read and still returns `Complete: true`. Worth pinning in both directions: stopping on the full page is the truncation the loader exists to prevent, and treating the empty page as a failure is a false partial.
- `failedStep === undefined` on both the single and batch paths.

---

**5. `needsSyncStateRepair` on first sync — intentional, and self-limiting.** It's what stamps sync state onto records that pre-date the integration. A `null` `SyncStatus` means the row has never been through this path, so there is nothing to shortcut — the shortcut's premise is "we wrote this row before and the content hasn't changed since". When it returns true the record falls through to `SetStandardIntegrationFields` + `Save()`, which sets `SyncStatus = 'Active'`, so the record is written exactly once and every later sync takes the shortcut. It's one extra write per record on the first sync of an entity, not per sync.

**8. `failedStep` selection.** Agreed it's arbitrary, and I'd rather leave it arbitrary than make it look principled. `RETRYABLE_STEPS` is a small allowlist of steps with no committed side effects, so the only way the choice changes the outcome is a batch where one failed result is on the list and another isn't — and there the `SuccessCount > 0` guard has already aborted, because a batch with mixed failures has committed something. Picking "the most critical" would mean ranking steps by severity, which is a second ordering to keep in sync with the allowlist for no behavioral gain.

PR #3294, opened 2026-07-27
fix(sql-converter): repair 25 tables of drift in the SS→PG boolean-column catalog

MS-BC:
## The bug

`CORE_METADATA_BOOLEAN_COLUMNS` in `packages/SQLConverter/src/rules/CoreMetadataBooleanColumns.ts` tells the SS→PG converter which baseline-table columns are PostgreSQL `BOOLEAN`, so SQL Server `BIT` literals (`0`/`1`) get rewritten to `FALSE`/`TRUE`. The converter learns column types from `CREATE TABLE` statements inside the migration being converted; for long-lived MJ tables created back in the baseline, this catalog is the only source.

**A missing entry does not fail conversion.** The literal passes through unchanged and the migration fails later, at *apply* time:

```
column "SupportsCreate" is of type boolean but expression is of type integer
```

The catalog had drifted from the v5.46 baseline by **25 tables**:

- **21 absent entirely** — `RemoteOperation`, `RecordProcess`, `ProcessRun`, `AISkillPermission`, `SignatureProvider`, `SignatureAccount`, `SignatureRequestLog`, `ExternalDataSourceType`, `ViewType`, `UserRoutine`, `UserRoutineRun`, `MLAlgorithm`, `ClusterAnalysisCluster`, `CollectionArtifact`, `EntityDocument`, `ConversationWidgetInstance`, `AIAgentChannel`, `AIAgentCoAgent`, `AIAgentSessionBridgeParticipant`, `AIBridgeAgentIdentity`, `AIBridgeProviderChannel`
- **4 with missing columns** — `AIAgent` (`AllowMemoryWrite`, `SupportsPlanMode`, `RequirePlanMode`), `AIAgentRun` (`PlanMode`), `ScheduledJob` (`RunImmediatelyIfNeverRun`), `IntegrationObject` (`SupportsCreate`, `SupportsUpdate`, `SupportsDelete`, `ContentHashApplicable`)

## Why it went unnoticed

The root cause is in the file's own regeneration recipe. It matched the type name as `/BOOLEAN/` — **uppercase only**. pg_dump-style baselines spell it lowercase:

```sql
"SupportsCreate" boolean DEFAULT false NOT NULL,
```

So re-running the documented recipe against a current baseline produced an **empty** result, and the catalog quietly stopped tracking the schema. Every table added since then is missing.

The failure mode is especially confusing because coverage is *per column*, not per table. In this statement, `SupportsWrite` converts correctly and `SupportsCreate` does not:

```sql
UPDATE [__mj].IntegrationObject
SET SupportsWrite = 0, SupportsCreate = 0
WHERE ID = '...';
```

That is exactly how this was found — converting a connector migration in `MemberJunction/Integrations` that moves a create operation between two `IntegrationObject` rows.

## The fix

1. **Catalog regenerated** from `migrations-pg/v5/B202607091514__v5.46.x__Baseline.pg.sql`. Purely additive — the diff was checked both ways and no shipped entry was absent from the baseline, so nothing was removed.
2. **Recipe corrected** to match the type name case-insensitively, with a comment explaining the empty-output trap so the next person doesn't re-introduce it.
3. **New test** `CoreMetadataBooleanColumns.test.ts` re-derives the catalog from the newest `B*__Baseline.pg.sql` and fails on any gap, listing the missing tables/columns. It no-ops outside the monorepo (a published install has no `migrations-pg/`), and also asserts the seed never overrides a table the migration itself created.

No schema change, no migration, no dependency change — a data-correctness fix inside one converter file plus a test.

---

## Testing

Unit: `cd packages/SQLConverter && npm run test` → **1107 passed, 3 skipped, 0 failed** (33 files, 2 skipped). Package builds clean. The new `CoreMetadataBooleanColumns.test.ts` re-derives the catalog from the newest `B*__Baseline.pg.sql` and fails on any gap.

The manual check is the part the unit test can't do: the bug only bites when converted SQL is **executed by PostgreSQL**. So the exact statement class that surfaced it — a connector migration moving a create operation between two `IntegrationObject` rows — was converted with the real converter build and then **applied to a real PostgreSQL 17 server** (`localhost:5499 / MJ_OA_PG`) against a table with the same `boolean` columns as `__mj.IntegrationObject`. Run twice: once with `origin/next`'s converter, once with this branch's.

Input (T-SQL):
```sql
UPDATE [__mj].[IntegrationObject]
SET SupportsWrite = 0, SupportsCreate = 0, SupportsUpdate = 1, SupportsDelete = 0, ContentHashApplicable = 1
WHERE ID = '...';
```

```
[PRE-FIX ] catalog IntegrationObject boolean cols (4): iscustom, supportsincrementalsync, supportspagination, supportswrite
[PRE-FIX ] SET -> SET "SupportsWrite" = FALSE, "SupportsCreate" = 0, "SupportsUpdate" = 1, "SupportsDelete" = 0, "ContentHashApplicable" = 1
[PRE-FIX ] APPLY FAILED -> column "SupportsCreate" is of type boolean but expression is of type integer

[POST-FIX] catalog IntegrationObject boolean cols (8): contenthashapplicable, iscustom, supportscreate, supportsdelete,
                                                      supportsincrementalsync, supportspagination, supportsupdate, supportswrite
[POST-FIX] SET -> SET "SupportsWrite" = FALSE, "SupportsCreate" = FALSE, "SupportsUpdate" = TRUE, "SupportsDelete" = FALSE, "ContentHashApplicable" = TRUE
[POST-FIX] APPLY OK -> {"SupportsWrite":false,"SupportsCreate":false,"SupportsUpdate":true,"SupportsDelete":false,"ContentHashApplicable":true}
```

Three things this pins down that a unit test wouldn't:

- **The failure is per column, not per statement.** In the pre-fix output `SupportsWrite` converts to `FALSE` and `SupportsCreate` stays `0` — in the same `SET` clause. Conversion reports success; PostgreSQL rejects it later, at apply time, with the exact message quoted at the top of this PR.
- **The catalog drift is measurable**: 4 known boolean columns on `IntegrationObject` before, 8 after.
- **Post-fix values are correct, not just accepted** — the read-back confirms `0`→`false` and `1`→`true` per column rather than a blanket cast.

Not covered: the other 24 repaired tables were not each round-tripped through a live server. Their entries come from the same regenerated baseline in one pass, and the new test fails if any of them drifts again.


jordanfanapour (approved):
LGTM

PR #3195, opened 2026-07-19
fix(core): maintain RunView cache totalRowCount on subset slots after mutations

AN-BC:
## What this fixes
The reported **`count_only` vs paginated `TotalRowCount` discrepancy** on high-churn entities (AI Agent Runs/Steps, Scheduled Job Runs) — AND, underneath it, a bug in the **shared RunView local cache** (`LocalCacheManager`, used by BOTH the server providers, where `TrustLocalCacheCompletely=true`, and the client providers): the cached `totalRowCount` for a **paginated / MaxRows-limited (subset) slot** was not maintained across mutations, so after a save/delete event a cached read returned the wrong total. The manifestation you observe over the wire comes through the **server-side** (MJAPI) cache. Two paths in `packages/MJCore/src/generic/localCacheManager.ts`:

1. `ApplyDifferentialUpdate` set `mergedTotalRowCount = mergedResults.length` and **ignored** `_serverRowCount` (the server's fresh authoritative COUNT).
2. `storeCachedResults` (in-place upsert/remove on entity events) rebuilt the entry **without** `totalRowCount`, so reads fell back to `results.length`.

Full-dataset slots hide it (`results.length == totalRowCount`); only **subset** slots expose it, after the first mutation. `count_only` is cache-ineligible so it always ran a fresh COUNT (correct); the cached paginated read returned the collapsed subset total.

### Verified root cause
On a **cold** MJAPI cache, `count_only` and the paginated read **agree** (e.g. 156 == 156); the discrepancy only appears once the server cache has been warmed + churned by mutations. So it's this cache-maintenance bug in the shared `LocalCacheManager`.

### Fix
- `ApplyDifferentialUpdate` honors the server's authoritative count as `totalRowCount`.
- `storeCachedResults` maintains `totalRowCount` across the net row delta. Full-dataset behavior provably unchanged.
- No view/SQL change.

### Verification
- Build clean; full MJCore suite green (**1488 passed / 95 files**), including **4 new regression tests** that fail against the unfixed code and pass with the fix.
- Live: cold-cache reads over the GraphQL wire agree across the affected entities (see PR comments).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3298, opened 2026-07-27
perf(core): stop paying for a speculative COUNT(*) on reads nobody counts - 6.x

MS-BC:
> **🚦 Confirmed not for 5.50.0 — targeting 6.x (opens Aug 3).** Split out of #3293 and now removed from
> that branch, so #3293 can go into the LTS candidate without this semantic change.

Split out of #3293 so it can be reviewed on its own.

## What it does

`RunView` decided whether to run its `COUNT(*)` in two places. Up front — when the caller paginated or asked for `ResultType: 'count_only'` — the count was issued in the same parallel batch as the data query, so one round trip. After the fact, if neither applied, it fired a **second, serial** `COUNT(*)` over the whole filtered view whenever the returned rows happened to fill the page:

```ts
const maxRowsUsed = params.MaxRows || entityInfo.UserViewMaxRows;
} else if (countSQL && maxRowsUsed && retData.length === maxRowsUsed) { … }
```

`maxRowsUsed` falls back to `UserViewMaxRows` (1000 by default), so that condition is met by every batch of any bulk read — and the resulting number was then discarded, because the callers that actually read `TotalRowCount` are pagination-aware and already took the first branch. A full extra scan, after the data had already returned, paid for on roughly every full page product-wide.

That fallback is removed. The count now runs only for OFFSET pagination, `ResultType: 'count_only'`, or the explicit `RunViewParams.ReturnTotalRowCount` — which routes into the up-front branch and so is issued in parallel with the data query rather than after it.

## The semantic change

| rows matching | MaxRows | before | after |
|---|---|---|---|
| 800 | 1000 | 800 | 800 |
| 5000 | 1000 | **5000** | **1000** |

`TotalRowCount ?? Results.length` returned the right answer in both rows before. It is now wrong in the second — and only in the second, which is the case you don't hit in dev against small data. Nothing throws, and no existing test fails. That is what made the call-site audit the real work of this PR.

## Live verification — real SQL Server + real MJAPI

Run against `MJ_3298` (a clone of a real database: 200,091 `Record Changes`, 3,755 `State Provinces`) with MJAPI booted on this branch and driven over the actual GraphQL wire. Measured, before the fixes below:

| query | TotalRowCount | rows returned |
|---|---|---|
| `MaxRows:100`, no StartRow | **100** | 100 |
| `MaxRows:1000`, no StartRow | **1000** | 1000 |
| `MaxRows:1`, no StartRow | **1** | 1 |
| no MaxRows | 200091 | 200091 |
| `MaxRows:100, StartRow:0` | 200091 | 100 |
| `MaxRows:100, StartRow:200` | 200091 | 100 |

This confirmed the intended narrowing exactly — and surfaced two things the original commit got wrong.

### 1. The escape hatch was unreachable from any client

`ReturnTotalRowCount` existed on `RunViewParams` but on **no GraphQL input type**. Every capped, non-paginated read from the browser therefore changed meaning with no way to ask for the old behaviour — the "widening the wire contract is a separable decision" note in the original description was wrong, because without it the opt-in only exists for code that runs inside the server process.

Now plumbed end to end: all four input types (`RunViewByIDInput`, `RunViewByNameInput`, `RunDynamicViewInput`, `RunViewGenericInput`), `ResolverBase` on both the single and batch paths, `RunViewGenericParams`, and the GraphQL data provider on both paths.

### 2. Two cache-key collisions

`TotalRowCount` is now a property of the **request**, but `GenerateRunViewFingerprint` did not encode that — so two requests carrying different meanings shared one cache slot, and whichever ran first decided what the other saw:

- **`{MaxRows:N}` vs `{MaxRows:N, StartRow:0}`.** The startRow segment is `params.StartRow ?? 0`, so these key byte-identically; but pagination requires `StartRow !== undefined`, so only the second gets the true count.
- **`{ReturnTotalRowCount:true}` vs the same params without it.** The opt-in was absent from the key entirely.

Both were harmless before this PR — every variant returned the same true count, so sharing a slot was invisible. Proven live: with a cache-busting filter, running the paginated variant first made *both* report 200091; running the plain variant first made *both* report 100.

Fixed with a `trc:` segment appended **only** when the request is paginated or opted in, so the overwhelmingly common plain read keeps its byte-identical pre-existing fingerprint and no existing cache entries are invalidated. Eight new tests in `localCacheManager.totalRowCountFingerprint.test.ts` cover both collisions in both orderings, plus the boundary cases (`ReturnTotalRowCount: false`, `MaxRows: 0`, keyset reads).

Re-measured after the fix — all four semantic lanes correct, and both collisions order-independent in both directions:

```
capped 100, no StartRow                 Total 100      rows 100
capped 100 + ReturnTotalRowCount        Total 200091   rows 100
paginated StartRow:0                    Total 200091   rows 100
count_only                              Total 200091   rows 0
collision 1, non-paginated first        100  /  200091   ✅
collision 1, paginated first            200091  /  100   ✅
collision 2, opt-in first               200091  /  100   ✅
collision 2, plain first                100  /  200091   ✅
```

## The call-site audit — done

The original description flagged ~125 unaudited `TotalRowCount` references as the reason this PR exists separately. That audit is complete. Every read was classified by whether the `RunView` that produced it is OFFSET-paginated, `count_only`, or opted in:

| classification | count |
|---|---|
| already safe — paginated | 34 |
| already safe — `count_only` | 48 |
| already safe — opted in | 1 |
| writes into a `TotalRowCount` field, not a read | 10 |
| no `RunView` in scope (telemetry passthrough, test assertions, cache plumbing) | 33 |
| **at risk — hand-verified** | **25** |

Of the 25 at-risk, **11 are genuine and are fixed here**. The rest are informational reads with a `?? Results.length` fallback (the React runtime's capture layer), internal cache plumbing where both sides share the same params, or false positives from the scan (the infinite-scroll datasource assigns `runViewParams.StartRow = startRow` rather than using a literal, so it *is* paginated).

The 11:

- **autotagging inbox + health badge counts** → `count_only` (they never touch the rows)
- **ComponentRegistry `/health` `componentCount`** → `count_only` (was `MaxRows: 1`, so it would have reported a component count of 1)
- **vector-management** ×2 — wants the single latest row *and* the true count → explicit opt-in
- **tags dashboard** ×2 — the KPI counts whose own comment says "TotalRowCount reflects the full DB count" → explicit opt-in
- **AI agent form execution history** — first page of a paged list that displays the total → explicit opt-in
- **entity-data-grid non-infinite path** — the grid's row-count display → explicit opt-in
- **`get-records` action** — the "showing N of M total" message → opt-in gated on the caller's own `includeCount`
- **`get-entity-details` action** — deliberately samples rows but reports the entity's true count → explicit opt-in

Two callers were already fixed in the original commit and remain so: `IntegrationEngine.pruneOldRunHistory` (a capped read made its guard read `keep <= keep`, so run history grew forever while looking bounded) and the Integration dashboard's destination record count.

## Scope

- `packages/GenericDatabaseProvider/src/GenericDatabaseProvider.ts` — the narrowing
- `packages/MJCore/src/views/runView.ts` — `ReturnTotalRowCount` param + params equality
- `packages/MJCore/src/generic/localCacheManager.ts` — the `trc:` fingerprint segment
- `packages/MJServer/src/generic/RunViewResolver.ts`, `ResolverBase.ts`, `types.ts` — the wire contract
- `packages/GraphQLDataProvider/src/graphQLDataProvider.ts` — client forwarding, single + batch
- `packages/Integration/engine/src/IntegrationEngine.ts` — pruner opt-in
- 8 call sites across `ng-dashboards`, `ng-core-entity-forms`, `ng-entity-viewer`, `core-actions`, `component-registry-server`
- `runViewTotalRowCount.test.ts`, `localCacheManager.totalRowCountFingerprint.test.ts` — coverage

## Tests

| Package | Result |
|---|---|
| `@memberjunction/core` | 1589 passed (103 files) |
| `@memberjunction/generic-database-provider` | 846 passed, 5 skipped |
| `@memberjunction/server` | 720 passed, 56 skipped |
| `@memberjunction/core-actions` | 268 passed |
| `@memberjunction/integration-engine` | 545 passed |
| `ng-dashboards`, `ng-core-entity-forms`, `ng-entity-viewer` | build clean |

## The open question, answered

The original description asked whether the audit was worth doing at all, or whether the safer resolution was to keep the old behaviour and have the sync hot path pass `MaxRows` explicitly. The audit is now done and it found 11 real breakages, so that question resolves in favour of doing it — but note what the audit actually shows: the failure mode is silent, only appears above the row cap, and would not have been caught by any test or by dev-scale data. The narrowing is right, and it is only safe *because* the opt-in is now reachable from every caller rather than only from inside the server process.

## Batch transport — live-verified

The description above measured the single-view path only. The batch path is now measured too, on the same instance (MJAPI on this branch, `MJ_3298`, driven over the wire):

| query | TotalRowCount | rows |
|---|---:|---:|
| `RunDynamicViewSystemUser` `MaxRows:100` | 100 | 100 |
| `RunDynamicViewSystemUser` `MaxRows:100, ReturnTotalRowCount:true` | **200093** | 100 |
| `RunDynamicViewSystemUser` `MaxRows:100, StartRow:0` | 200093 | 100 |
| `RunViewsSystemUser` `[{MaxRows:100}]` | 100 | 100 |
| `RunViewsSystemUser` `[{MaxRows:100, ReturnTotalRowCount:true}]` | **200093** | 100 |
| `RunViews` (non-system, API key) `[{MaxRows:100}]` | 100 | 100 |

And the case that actually exercises the fingerprint fix on the batch transport — **one batch carrying both variants of the same entity and the same `MaxRows`**, differing only in the opt-in:

```
RunViewsSystemUser(input: [
  { EntityName: "MJ: Record Changes", MaxRows: 100, ReturnTotalRowCount: true },
  { EntityName: "MJ: Record Changes", MaxRows: 100 }
])
→ [ { TotalRowCount: 200093, RowCount: 100 },
    { TotalRowCount: 100,    RowCount: 100 } ]
```

Both meanings survive in a single round trip. Without the `trc:` fingerprint segment these two entries key identically and whichever executed first would decide both.

## Correction to the earlier "unrelated observation"

An earlier revision of this description reported that batch `RunViews` returns null under API-key auth with `Entity <name> not found in metadata`. **That was my error, not a server bug.** The entity is named `MJ: Record Changes` in this database; the probe passed `Record Changes`. With the correct name every batch form above succeeds under API-key auth, including the non-system `RunViews`. There is no batch/system-user defect and nothing here is blocked on one.

What the mistake did expose is a pre-existing **diagnostics** weakness, unrelated to this PR and not changed by it: `RunViewsGeneric` catches, logs to the server console, and returns `null` into a non-nullable GraphQL field, so the client sees `Cannot return null for non-nullable field Query.RunViews` with the real reason visible only in the server log. The lookup behind it is also `md.Entities.find(e => e.Name === …)` — a case- and whitespace-sensitive scan where the codebase standard is `EntityByName`. Both are worth fixing on their own branch; flagging, not touching, here.

---

## Verification round 2 — adversarial review, full build, live MJAPI smoke test

Re-verified after merging `next` (the branch was 63 behind; both prior CI failures were staleness, not defects — the integration tier was failing on `No valid API credentials/keys are configured`, an environment condition).

### Full build

`turbo build --filter="@memberjunction*" --force` — **299/299 tasks, 0 cached, exit 0**. Forced rather than incremental: an ordinary run reported 267/271 restored from cache, which proves nothing about whether this code compiles.

Unit tests on the packages that own the change: `@memberjunction/generic-database-provider` **902 passed / 5 skipped**, `@memberjunction/core` **1854 passed**.

### Live MJAPI smoke test

A real server — not a mock — built from this branch: `packages/MJAPI` on port 4310 against SQL Server, confirmed to be *this* worktree's process by matching the listening PID's cwd rather than trusting that something answered the port. Queries go over HTTP/GraphQL as `RunMJEntityFieldDynamicView`, so they traverse the entire remote path: GraphQL input type → `ResolverBase` → `RunViewParams` → `GenericDatabaseProvider`.

Ground truth for the entity came from `ResultType: 'count_only'` — **5,510 rows** — the one path whose meaning this PR does not alter.

| # | Assertion | Result |
|---|---|---|
| 1 | Capped, non-paginated read reports the page, not a phantom total | `TotalRowCount=10`, `RowCount=10` |
| 2 | `ReturnTotalRowCount: true` crosses GraphQL and restores the true total | `TotalRowCount=5510`, `RowCount=10` |
| 3 | OFFSET-paginated read still gets the true total (unchanged) | `TotalRowCount=5510` |
| 4 | `StartRow: 0` counts as pagination, not as absent | `TotalRowCount=5510` |
| 5 | A cached no-count result cannot satisfy an opted-in request, or vice versa | warm `10` / warm `5510` |
| 6 | The opt-in changes only the count, never the rows | `10` rows either way |

**6/6 passed.**

Check 1 is the substantive proof, not merely a semantic one. Under the old code that exact query satisfied `retData.length === maxRowsUsed` (10 === 10) and therefore fired the extra `COUNT(*)`, returning 5,510. It now returns 10 — the phantom scan is demonstrably no longer issued.

Check 4 matters because `usingPagination` tests `StartRow !== undefined && StartRow >= 0` rather than truthiness. A truthiness check would silently drop `StartRow: 0`, and a real caller (`classify-item-grid`) passes exactly that.

Check 5 exercises the new `trc:` fingerprint segment against the real cache in both directions, warm.

### Adversarial review

Every `TotalRowCount` reader in `packages/` was enumerated and classified, not sampled. All are OFFSET-paginated, `count_only`, or explicitly opted in. Three findings, all fixed here:

1. **The runtime entity data grid was missed.** `metadata/components/code/generic/entity-data-grid.js` renders a "total records" figure from a capped read using `Skip`, which is not `StartRow` and so never counted as pagination, via the exact `TotalRowCount ?? RowCount ?? length` idiom this PR invalidates. Its Angular twin `EntityDataGridComponent` was fixed; this one was not. TypeScript cannot catch it — the file is runtime JS shipped as component metadata — so its displayed total would have silently pinned to the row cap.
2. **The changeset covered 4 of the 10 changed packages.** `@memberjunction/server`, `graphql-dataprovider`, `core-actions`, `ng-core-entity-forms`, `ng-entity-viewer` and `component-registry-server` all ship changed code and would have been released without a version bump.
3. **The changeset's closing paragraph contradicted the code**, stating `ReturnTotalRowCount` was "deliberately not plumbed through GraphQL" when the PR plumbs it through all four `RunView*` input types, both client paths and the resolvers. Smoke check 2 is the live disproof.

No `any`, `as any`, or `.Get()`/`.Set()` on any added line.


MS-BC:
Base synced to current `next` so CI re-runs against the current base. No conflicts.

Manually re-ran the affected suites locally post-merge; the RunView count-speculation path is exercised by the driver integration matrix on this PR (SQL Server / Postgres / MySQL / MongoDB / Oracle), all of which are green.

MS-BC:
## A 12th breakage the audit missed — fixed here

Auditing this PR's central risk (the narrowing is silent, only bites above the row cap, and
no test would catch it), I went looking for `TotalRowCount` consumers this branch does not
touch. There are ~165 read sites tree-wide, so most had to be cleared by inspection — but
one is a genuine miss, and it announces itself in a comment:

`packages/Actions/CoreActions/src/custom/geo/scheduled-geocoding.action.ts`

```ts
// TotalRowCount reports the full count matching the filter regardless of the
// keyset seek predicate / page offset, so the first page gives us a stable denominator.
if (candidateTotal === 0) candidateTotal = pageResult.TotalRowCount;
```

That assumption is exactly what this PR invalidates. The action has two loaders and they
diverge on precisely the property that now matters:

| loader | `MaxRows` | `StartRow` | after this PR |
|---|---|---|---|
| `loadEntityPageOffset` | 500 | `offset + 1` | ✅ correct — paginated, takes the up-front branch |
| `loadEntityPageKeyset` | 500 | **none** (seeks via `AfterKey`) | ❌ returns **500**, not the filtered total |

The keyset path is the **preferred** one — `canUseKeyset ? keyset : offset`, chosen per its
own comment because it "keeps each page O(log N) regardless of depth — a critical win when
the entity has millions of rows". So the break lands specifically on the large tables where
a 500-cap denominator is most wrong, while the fallback path stays correct. That asymmetry
is what makes it easy to miss.

**Blast radius is reporting, not data.** `candidateTotal` feeds `logEntityHeader`, the
`Total:` field in the batch progress metadata, and `logEntityComplete`. Records still
geocode correctly; a scheduled run over a million-row table would just report
"500 records with address data" indefinitely.

**Fix:** `ReturnTotalRowCount: true` on the keyset loader only.

### Checked and *not* changed

- The retry loop reads `pageResult.TotalRowCount` into `retryTotal` with the same
  "from page-1 TotalRowCount" comment — but its `RunView` is inline and passes
  `StartRow: pageOffset + 1`, so it is paginated and already correct. Left alone.
- `loadEntityPageOffset` — paginated, correct. Left alone.

### Why this had to land here rather than as its own PR

`ReturnTotalRowCount` does not exist on `next` — this PR introduces it. I started the fix
on a branch cut from `next` and it could not compile, which is itself a useful data point:
**any caller broken by this change can only be repaired on this branch**, because the escape
hatch ships with the breakage. Worth keeping in mind if more sites surface after merge.

rkihm-BC (commented):
Reviewed the full diff plus both follow-up comments, and independently re-audited every
`TotalRowCount` reader on current `next` (~165 sites) rather than sampling. Also traced the
provider → cache → transport seams and checked the integration tier. Notes below.

Heads up on scope: this head is **204 commits behind `next`**, so every audit in the description
was run against a tree that's now ~2 weeks old.

---

## What's good

- The diagnosis is right and the win is real. `maxRowsUsed = params.MaxRows || entityInfo.UserViewMaxRows`
  meant the fallback fired on essentially every full page, product-wide, serially, after the data had
  already returned, for a number the caller usually discarded.
- The predicates mirror provider internals exactly. `usingKeyset` really is just `!!params.AfterKey`
  (`GenericDatabaseProvider.ts:1546`), so `isOffsetPaginated` in `localCacheManager` is a faithful
  mirror of `usingPagination`. That's the kind of duplicated predicate that usually drifts.
- Both self-found cache collisions are genuine and well-closed. Noticing that `startRow = params.StartRow ?? 0`
  makes `{MaxRows:N}` and `{MaxRows:N, StartRow:0}` byte-identical while only the second is "paginated"
  is a subtle catch, and appending `trc:` only for non-plain requests keeps the additive discipline of
  the existing `imr:` / `ak:` / `rls:` / `vw:` segments — no existing entries invalidated.
- `AreParamsEqual` was updated. The dedup path is the easy one to forget.
- GraphQL plumbing is complete and I verified it in the diff, not just in the description: all four
  input types, both `ResolverBase` paths, `RunViewGenericParams`, both client paths including batch.
- Two real bugs fixed along the way — `pruneOldRunHistory`'s `keep <= keep` guard, and
  ComponentRegistry `/health` reporting `componentCount: 1`.
- Per-call-site comments explaining *why* each needs the opt-in, and the tombstone comment in
  `GenericDatabaseProvider`, are exactly right.
- Correctly declines to fix the `RunViewsGeneric` null-into-non-nullable / `Entities.find` diagnostics
  weakness here. Agreed, separate branch.

---

## Blockers

### 1. `runview-matrix` RVM2 fails in the default deterministic tier

`runview-matrix.checks.ts:283` `probeFullWidth`:

```ts
const r = await new RunView().RunView({ EntityName: entity.Name, ResultType: 'simple', MaxRows: maxRows }, ctx.User);
…
if (expected !== undefined && r.TotalRowCount !== expected)
    t.Failures.push(`${entity.Name}: count_only=${expected} but full-width TotalRowCount=${r.TotalRowCount}`);
```

`sweepMaxRows` defaults to **5** and the sweep runs over **every** entity. `.IT31-runview-matrix.json`
is `{"tier": "deterministic", "transport": "client"}`, sequence 41, and no RVM check carries
`RequiresMutation` — so this runs unconditionally on `pnpm run test:integration` and fails for every
entity with more than 5 rows.

The description reports unit-test counts and a hand-rolled MJAPI smoke test, but never the integration
tier. That's the tier CLAUDE.md points at for exactly this class of cross-package seam.

### 2. `cache-gauntlet` CG5 — and the design conflict it exposes

`subsetParams(1)` is `{EntityName, Fields, ResultType:'simple', MaxRows: 1}`. CG5 asserts:

```ts
Assert(before > warm.Results.length, `TotalRowCount (${before}) must exceed the truncated row count — this is the #3195 collapse`);
AssertEqual(after.TotalRowCount ?? -1, before + 1, …);
```

Both are now false (`1 > 1`). To be precise about severity: CG5 is `RequiresMutation: true`, so it's
skipped on a default run and fails only on a mutation-enabled one — not a default-tier blocker.

The reason I'm still listing it here is its name: *"a subset slot keeps a DB-accurate TotalRowCount
across a save (#3195 and #3199 compose)"*. The cache layer was **deliberately built** to preserve the
DB-accurate total on a `MaxRows` slot — `localCacheManager.ts:2296` (`mergedTotalRowCount = max(serverRowCount, mergedResults.length)`,
whose comment names "the exact defect behind the RunView TotalRowCount discrepancy") and
`localCacheManager.ts:2476` (`storeCachedResults` carries `prior.totalRowCount` forward across saves).

After this PR the provider stops producing that total while the cache layer keeps maintaining it. A
`MaxRows:1` slot that has seen three saves now reports `TotalRowCount: 4` while returning 1 row —
neither the page size nor the true total. Either #3195's premise was wrong or this narrowing is, and
the PR doesn't engage with the conflict. That's the discussion I'd most like to see in the description.

### 3. The fix in the Aug 10 comment reintroduces the cost, per page

`loadEntityPageKeyset` is called once per iteration of `while (totalProcessed < maxRows)`
(`scheduled-geocoding.action.ts:213`), with `BypassCache: true` and an **unconditional**
`ReturnTotalRowCount: true` — while the caller consumes only page 1's value:

```ts
if (candidateTotal === 0) candidateTotal = pageResult.TotalRowCount;
```

`PAGE_SIZE` is 500, and the loader's own comment justifies keyset because the entity may have
"millions of rows." So this buys a correct denominator by paying an unbounded `COUNT(*)` every 500
rows — the exact cost this PR exists to delete, reintroduced per page.

```ts
ReturnTotalRowCount: lastSeenKey === undefined   // page 1 only
```

---

## Consumers the audit missed

All confirmed against current `next`.

**a. `BaseEntity.GetDataObject` — emits an affirmatively false statement.**
`baseEntity.ts:2738`, via `GetRelatedEntityDataExt` → `EntityInfo.BuildRelationshipViewParams`, which
sets only `MaxRows` and never `StartRow`:

```ts
if (pre.maxRecords < reData.TotalRowCount)
    msg = `Only the first ${pre.maxRecords} records are included…There are ${TotalRowCount} total`;
else
    msg = `All ${reData.TotalRowCount} records are included in this response.`;
```

`TotalRowCount` now equals `maxRecords`, so `N < N` is false and the note flips to **"All N records
are included in this response."** — in exactly the truncation case the note exists for. This payload
feeds agents/LLMs and API consumers, so it's a lie rather than a wrong number. In MJCore, a package
this PR already touches.

Fix belongs in `BuildRelationshipViewParams`, **gated on `maxRecords > 0`** to match the existing
`if (maxRecords && maxRecords > 0) params.MaxRows = maxRecords;`. Unconditional would put a `COUNT(*)`
back on every related-entity grid load in every generated form — `base-form-component` calls it with
two args and no `maxRecords`.

**b. `EntityDocumentTemplateParser.ts:52`** — same `GetRelatedEntityDataExt(re, null, maxRows)` path.

**c. RecordSetProcessor — the same keyset/offset asymmetry, twice, and persisted.**
`RecordSetProcessor.ts:125` seeds the run's `Total` from batch 1's `TotalRowCount`, and that `Total`
is written into the persisted `ProcessRunSummary`.

| source | pagination | after this PR |
|---|---|---|
| `viewSource.ts` | `StartRow` | correct |
| `listSource.ts` | `AfterKey` (keyset default; offset only for legacy resume) | `Total = batchSize` |
| `sourceUtil.ts` `pageEntity` | `preferKeyset ? AfterKey : StartRow` | wrong on the preferred path |

Same first-batch gate as the geocoding fix — `ReturnTotalRowCount: cursor == null` — since `NextBatch`
runs per batch with `BypassCache: true` and the value is consumed once.

**d. `RuntimeActionBridge` — the escape hatch is unreachable, again.**
`RuntimeActionBridge.ts:313` declares `RunViewArgs` as an explicit allowlist
(`EntityName, ExtraFilter, OrderBy, Fields, MaxRows, UserSearchString, ResultType`) and rebuilds the
params object field by field. `ReturnTotalRowCount` cannot cross it. This is the same defect the
description correctly identifies for GraphQL — "the escape hatch was unreachable from any client" —
reproduced on the runtime-action transport, which the description names as the thing that makes the
narrowing safe.

Four shipped runtime actions read `TotalRowCount` as a true total through that bridge:

- `metadata/actions/runtime/code/entity-data-quality-report.js:54` — the comment says *"TotalRowCount
  comes back from RunView even when we capped MaxRows"*. `TotalRowCount > cappedAt` becomes false, so
  the **"Sampled N of M rows — results are an estimate" warning silently disappears**, and
  `totalRowsInEntity` reports the sample size.
- `summarize-entity-records.js` (MaxRows 50), `weekly-entity-digest.js` (MaxRows 100),
  `find-similar-records.js` (MaxRows 5000) — each reports its cap as the total.

### Checked and genuinely clear

So the scope of the above is honest: `entity-viewer.FetchAllRowsForExport` (passes `StartRow: 0`);
every `count_only` KPI site (Communication ×2, list-management, entity-record-detail-panel,
entity-form, list-form, Lists dashboards ×2, single-list-detail, PredictiveStudio `seams.ts`,
memory-manager-agent); ComponentStudio artifact dialogs, conversation-feedback, classify-item-grid,
data-artifact-viewer (all paginated). Also worth recording: `count_only` **cannot** collide on the
fingerprint despite `ResultType` being excluded from it, because `runViewCacheEligible`
(`providerBase.ts:1274`) bars it from both cache read and write.

---

## Worth verifying before merge

The smart-cache status query is an unbounded `SELECT COUNT(*) AS TotalRows`
(`GenericDatabaseProvider.ts:2467`). `runDifferentialQueryAndReturn` skips the differential for
"subset shapes", but the predicate is `MaxRows != null && > 0 || StartRow != null && > 0` — so a read
with **no explicit `MaxRows`**, capped by `UserViewMaxRows`, is not a subset shape and returns
`rowCount: serverStatus.rowCount` (the true total). With `TotalRowCount: cached.totalRowCount ?? results.length`
on the read side, a `CacheLocal` read of an oversized entity may return the true total when warm and
the page size when cold.

That's a code-reading inference across several hops, not something I ran — but non-determinism keyed
on cache warmth would be worse than either semantic alone, so it's worth a live probe.

---

## Smaller items

- **Changeset level.** `.claude/rules/changesets.md` and `check-changeset-bump.mjs:54`
  (`{ label: 'metadata', pattern: /^metadata\/.+/ }`) make anything under `metadata/` a `minor`. This
  PR edits `metadata/components/code/generic/entity-data-grid.js` and every entry is `patch`; base
  `next` is not a certified line, so `npm run check:changeset` should fail. Under-bumping is the
  direction the rule calls out as the worse one.
- **React linter whitelist.** `runview-call-validation.ts:30` `VALID_RUNVIEW_PROPS` doesn't include
  `ReturnTotalRowCount`, and neither do the two author-facing suggestion strings. Not a CI break —
  nothing lints `metadata/components/code/**`, and `Skip: 0` already sits there unflagged — but that
  whitelist is the contract surface component authors and the component-generating model read, so
  today it would tell them the correct property is invalid.
- **Tests pin the mechanism, not the contract.** The two new files pin *when the COUNT is issued* and
  *what the fingerprint looks like*. Nothing pins the thing that broke. Fixing RVM2/CG5 to state the
  new rule, plus a `ReturnTotalRowCount` lane in `runview-features`, is the durable pin.
- **Consider a machine-readable signal.** The description's own framing is that the failure is silent,
  only appears above the cap, and no test catches it. Every miss above is a consumer that had no way
  to tell. Something like `RunViewResult.TotalRowCountIsExact` would let consumers detect the
  difference instead of having to know the request shape.
- Rebase and re-run the call-site audit as the last step before merge. The Aug 10 comment makes the
  argument itself: anything broken by this can only be repaired on this branch, because the escape
  hatch ships with the breakage.

---

## Summary

Right idea, real win, and unusually rigorous self-review — the audit is more thorough than most PRs
get. I don't think it's mergeable as-is: RVM2 fails the default deterministic tier; CG5 exposes a
direct conflict with what #3195/#3199 were written to establish; five consumers are still unfixed,
one of which produces a false statement to LLM consumers and one of which reproduces the exact
"unreachable escape hatch" defect the description identifies as the safety precondition; and the
most recent fix pays an unbounded `COUNT(*)` per page on the tables it was written to protect.

Happy to pair on the `#3195` question specifically — that one seems like a genuine design decision
rather than an oversight, and it'd be good to settle it in the description before the rest gets
patched around it.

MS-BC:
Closing this.

@rkihm-BC — thank you for the review; it was the thing that settled this. Re-auditing all ~165 `TotalRowCount` readers rather than sampling is what turned "the audit looks thorough" into a countable answer: five consumers still broken, one of them emitting an affirmatively false statement to LLM consumers, and one reproducing the same unreachable-escape-hatch defect the description had named as the safety precondition.

The reason for closing rather than patching those five: the defect isn't the misses, it's the predicate. This PR decides whether to compute the count from the **request shape** (is it paginated? did it opt in?), while whether the count is *needed* is a property of the **consumer** (does it read `TotalRowCount` as a true total?). Those two sets don't coincide, so the misses aren't oversights to be swept up — they're what that mismatch produces, and it leaked five times under an unusually careful audit. Patching the five known ones leaves the mechanism intact for the next reader who caps `MaxRows` and trusts the number.

Two further things found while working the review, worth recording since they outlive the branch:

1. **The `#3195` conflict resolves in an unexpected direction.** `ApplyDifferentialUpdate` declines to merge into a subset fingerprint and invalidates instead (`isSubsetFingerprint`), so the `max(serverRowCount, merged.length)` branch added for paginated slots can only ever run for full-dataset slots — where the merged length already *is* the true total. The subset case it was written for cannot reach it. That also answers the cache-warmth non-determinism you flagged as a code-reading inference: a warm subset read can't adopt a server `COUNT(*)` there, because it isn't served from that path at all. No live probe needed.
2. **The three CG-adjacent unit tests were passing for the wrong reason.** The block is named *"TotalRowCount preservation (paginated / MaxRows-limited slots)"*, but its fixture is `{EntityName, ExtraFilter}` — no `MaxRows`, no `StartRow`, no opt-in. It's a plain unbounded read, with the subset simulated by handing `SetRunViewResult` a total directly. Your "tests pin the mechanism, not the contract" note was more accurate than it looked.

The replacement will be a much smaller change built on a rule that can't change any consumer's semantics: **derive the total when it's free** — when a result comes back short of its cap, the total is `StartRow + rows.length` by arithmetic, so no query is needed. That also wins the last page of every grid, which this PR doesn't. Paginated and `count_only` reads keep their parallel count exactly as today. The bulk-read win that motivated this branch (the sync engine's batches, which fill their cap every time and never read the number) needs an explicit opt-out rather than an inference, and belongs in its own PR with measurements attached.

The two incidental bug fixes here are worth keeping and I'll carry them over: `pruneOldRunHistory`'s `keep <= keep` guard, and ComponentRegistry `/health` reporting `componentCount: 1`.

PR #3300, opened 2026-07-27
[6.0]: docs(plans): agent context optimization proposal

AN-BC:
## Summary

Adds `plans/agent-context-optimization.md` — a proposal for reducing MJ agent prompt context, prompted by Anthropic's [*The new rules of context engineering for Claude 5 generation models*](https://claude.com/blog/the-new-rules-of-context-engineering-for-claude-5-generation-models).

**Docs only — no code or metadata changes.** This is a proposal for review, not an implementation.

Complements the existing [`plans/agent-token-optimization.md`](https://github.com/MemberJunction/MJ/blob/next/plans/agent-token-optimization.md), which covers *runtime* token reduction. This one covers *authoring-time* reduction — what we write into prompts in the first place. No overlapping work items.

## Why

MJ ships ~250k tokens of hand-authored prompt text across 103 templates, a quarter of it inside fenced example blocks that frontier models no longer need — and that measurably narrow their exploration space. Separately, the model that will run a prompt is currently chosen *after* the prompt is rendered, so prompts can't adapt to the model executing them.

The proposal does two complementary things: **subtract** redundant content, and **condition** what remains on the measured capability of the model actually running.

## Measured baseline

| Measure | Value |
|---|---|
| Prompt templates | 103 files, 927 KB ≈ **250k tokens** |
| Share inside fenced blocks | **25% — 228 KB ≈ 62k tokens** |
| Loop system prompt | 33.4 KB ≈ 9k tokens, 37% fenced, 16 JSON example blocks |
| Agents trimming via `AgentTypePromptParams` | **3 of 26** |
| Guardrail markers across templates | **861** |
| Agent Eval tests | **2** |
| `MJ: Test Rubrics` seeded | **0** |

## Phases

| Phase | Content | Risk |
|---|---|---|
| **0** | Eval baseline, token reporting | gating |
| **0.5** | PowerRank rebase — absolute, anchored, locked | gates Phase 4 |
| **1** | `AgentTypePromptParams` on the 23 untrimmed agents | low |
| **2** | Loop-prompt subtraction; relocate tool instructions into Action descriptions | medium |
| **3** | Generator coverage, authoring gate, externalize large inline examples | low-med |
| **4** | Resolve model before render; capability-conditional template content | med-high |
| **5** | Two-tier Action catalog (progressive disclosure) | highest |

Phases 1–3 and 4–5 are independent and can run in parallel.

## Notable findings

- **Four of the post's six shifts are already partly or wholly shipped** — `generate-prompt-types.mjs` already compiles TS interfaces into prompt docs, the Skills catalog is textbook progressive disclosure, auto-memory is done, and the prompt-cache prefix discipline is sound. This is a trim, not a rewrite.
- **`_MODEL_ID` / `_VENDOR_ID` resolve to empty string on every non-override run** (`prompt.system-placeholders.ts:243-257`) because rendering precedes selection. Phase 4a fixes this as a side effect.
- **`PowerRank` failover currently retries a lower-power model using a prompt written for the original.** A live mismatch today, independent of anything else here; per-model rendering fixes it.
- **The rebase window is now.** 0 prompts set `MinPowerRank`, and `PowerPreference` is rank-order relative and invariant under rescaling — so blast radius is effectively nil, and stops being nil once anyone pins an absolute threshold.
- **Power, speed, and cost are near-independent** across 136 LLMs (Power↔Speed `r = -0.220`, Power↔Cost `r = +0.137`), and four models are simultaneously fast and frontier-class. Thresholds must key on power only.

## Design decisions worth reviewing

- **PowerRank becomes absolute and locked, not percentile.** Capability is a property of the model, not its standing among peers. A percentile would make the persisted `AIPromptRun.ModelPowerRank` column time-dependent, would let adding a model silently re-tier existing ones, and would demote still-capable models into verbose prompts they don't need. Scale is 1–1000 per model type with current frontier around the 500s.
- **No tier abstraction.** Templates compare raw `PowerRank` against absolute thresholds — `{% if _MODEL_POWER_RANK < 100 %}`. A named-tier layer (`frontier`/`standard`/`light`) was designed and rejected: those labels are definitionally market-relative and would reintroduce the same time-dependence the rebase removes, one level up where it's harder to see. A threshold is a capability claim that doesn't move; more models simply qualify over time.
- **MJ ranks are authoritative.** Independent evaluations are inputs for placing new models on our scale — recorded with source, version, and as-of date, derived once at entry, never re-derived when a provider rebases methodology.
- **Selection logic is untouched.** Rendering adapts to selection, never the reverse; multi-model paths re-render per model, which the plan shows is near-free.
- **Safety guardrails are explicitly retained.** Judgment replaces stylistic rules, not unrecoverable ones.

## Test plan

None — documentation only. The plan's own Phase 0 defines the eval baseline that would gate the implementation work.

https://claude.ai/code/session_01X3AMTFw6aPe5s3zu2iwZBE


AN-BC:
@MarceloT-BC @SDesai-BC noticed some activity above from you two, this is **not** for action yet, we will look at this for the `6.0.0` release. Please do share any feedback you may have

Also, adding: @jordanfanapour @rkihm-BC @cadam11 @EL-BC @bc-izygmunt @arie-glazier-bcc @CaelebB-BC @MS-BC and others, please review and comment on this, I think there is a lot of upside here but this will require a **lot** of testing as it has a wide effect.

MarceloT-BC:
I added Soham here because I thought this was something else he had done. Was only intending to share the PR nothing more. Will share the link in the future.

AN-BC:
Converting this PR to an issue to deprioritize it. The branch `claude/mj-agent-context-optimization-eptdee` is preserved and can be picked back up later. Tracking issue: #3947.

---

PR #3304, opened 2026-07-27
fix(codegen-lib): gate reverse-relationship emission on type availability; add opt-in includeSchemas

MarceloT-BC:
## Short Summary
Reverse-relationship GraphQL fields are now emitted only when the related type is resolvable in the generated file — its class is emitted inline in the run, or it is a core (__mj) type reached via the mj_core_schema_server_object_types.* namespace import. This fixes the TS2304 build break that occurred when a base Open App was generated alongside a dependent app that foreign-keys into it (the base emitted fields typed with the dependent's classes, which its package cannot import). The gate emits exactly the compilable subset of prior output, so any run that compiled before produces byte-identical output.

Also adds an opt-in includeSchemas positive scope, resolved into excludeSchemas at a single point before the metadata and file-generation phases — pure sugar over the existing exclude mechanism (in-scope iff listed and not excluded; no implicit includes). Absent/unused leaves classic exclude-only behavior unchanged.

# Fix CodeGen for linked Open Apps (+ opt-in `includeSchemas` scope)

## Why this matters

You **cannot develop the Open App suite without linking the apps together** — a hard build-time dependency, not a convenience. `bizapps-orders` npm-depends on `@mj-biz-apps/accounting-*`, which depends on `bizapps-tasks` + `bizapps-common`; those packages are unpublished, so they resolve only when co-linked in one instance. The chain **common → tasks → accounting → orders must be co-linked just to compile.** But co-linking then **breaks the base app's CodeGen**: generating `bizapps-common` with a dependent present injects reverse-relationship GraphQL fields typed with the *dependent's* classes — which the base package cannot import — yielding `TS2304`. Measured: `bizapps-common` gains **20 uncompilable references** and fails to build.

## Root cause & fix

A reverse-relationship field references the related entity's GraphQL type by **bare class name**, which resolves only if that class is emitted inline in the file. The old non-core pass emitted these for **every** related entity regardless of whether its class was present. The new gate emits one **iff the related type is actually resolvable where it's referenced**:

- its class is emitted **inline** in this run (`generatedEntityNames`), or
- it's a **core (`__mj`) entity in a non-core file**, reached via the `mj_core_schema_server_object_types.*` namespace import (this mirrors, 1:1, how `generateServerRelationship` builds the reference).

## Behavioral impact — reviewed per scenario

**The governing fact (provable from the gate):** the new gate emits **exactly the compilable subset** of what the old code emitted. An un-emitted class yields a bare reference that cannot compile; the new gate emits iff the reference *is* resolvable; the resolvable set is a subset of "always." **Therefore, for any input that produced a compiling result under the old code, the new output is byte-identical — the only emissions removed are ones that would not have compiled.** With that established:

- **Regular case (typical single instance — core + app schemas, normal excludes like `sys`/`staging` that nothing FKs into):** **Identical output.** Every entity a reverse-relationship points at is generated in the run, so it's in the set and still emitted — the gate is a no-op. It differs *only* if that instance excluded a schema something FKs into, in which case the old code emitted an uncompilable field (that instance was already failing to build) and the new code drops it — a fix, never the removal of a working field.

- **Monolith install (single repo, string `entityPackageName`, all schemas generated into one `mj_generatedentities` file):** **Identical output — and this is the case worth being explicit about.** A monolith's defining property is *one file*: every schema's class is emitted together, so every cross-schema reverse-relationship (e.g. `sales`↔`crm`) is "in the set" and is emitted exactly as before. Nothing a monolith references is ever out-of-set, so set-membership is a no-op there. A monolith does **not** lose any cross-schema navigation. (It would differ only for a reference into an *excluded* schema — which never compiled.)

- **Production / Open-App install (published packages, map-form `entityPackageName`):** **Identical where the old code compiled; fixes the break where it didn't.** Two sub-cases: (a) if a host co-generates several installed apps' schemas into one file (none excluded), their classes are all inline → cross-app reverse-relationships compiled under the old code and are **still emitted** under the new one (set-membership keeps them). (b) if an app's schema is **excluded** from the run — the base-app-with-dependent case — the old code emitted uncompilable refs (`TS2304`); the new code drops them and the build succeeds. Also verified: the **committed generated code of shipping apps is unchanged** — those files were generated in isolation (nothing excluded that they reference), and re-running CodeGen with this change produces an **empty diff**.

**The one honest semantic boundary:** cross-app reverse GraphQL traversal *from a base type into a dependent app that is not co-generated with it* is no longer emitted. It never compiled in that configuration; if such navigation is ever wanted, it belongs in the dependent app's package (which imports the base), not the base's.

## Also included: opt-in `includeSchemas`

Today an Open App must **exclude every other app's schema** to codegen cleanly — an O(N²) hand-maintained list that can't anticipate schemas it doesn't know about (a deployed app's CodeGen would even traverse a client's own schemas). `includeSchemas` lets an app name **only what it owns**. Pure sugar: resolved into `excludeSchemas` at a single point before both phases, so every downstream consumer is unchanged (in-scope ⇔ listed ∧ not excluded; no implicit includes — core must be listed explicitly). Absent/unused ⇒ classic exclude-only behavior, unchanged.

## Validation

| | Before (branch base) | After |
|---|---|---|
| `codegen bizapps-common` (co-linked w/ tasks) | 20 cross-package refs | **0** |
| generated file vs committed | +65 lines | **clean diff (idempotent)** |
| `build bizapps-common` | **FAIL** — TS2304 ×11 | **PASS** (5 packages) |
| `includeSchemas` scoping | — | validated live, correct scoped output |
| CodeGenLib unit suite | — | **597 passed / 0 failed** |

## Scope / follow-up

Confined to `@memberjunction/codegen-lib` (`graphql_server_codegen.ts`, `runCodeGen.ts`, `config.ts`, new `schema-scope.ts` + tests). A separate pre-existing hazard — destructive CodeGen paths that assume full-database visibility (Angular orphan-dir deletion, whole-file GeneratedEntities rewrite, the CRUD validator) — is documented in `MJ-UPSTREAM.md` and intentionally **not** touched here; it should be hardened before CodeGen scope is narrowed further.

### DIFF
--- a/.changeset/fix-codegen-linked-open-apps.md
+++ b/.changeset/fix-codegen-linked-open-apps.md
@@ -1,0 +1,9 @@
+---
+"@memberjunction/codegen-lib": patch
+---
+
+Fix CodeGen for linked/co-generated Open Apps and add an opt-in `includeSchemas` scope.
+
+- **Reverse-relationship emission is now gated on type availability, not schema/package heuristics.** A reverse-relationship (child-array) GraphQL field references the related entity's type by bare class name, so it is now emitted only when that type is actually resolvable in the generated file — i.e. the related entity's class is emitted inline in this run (`generatedEntityNames`), or it is a core (`__mj`) entity in a non-core file (reached via the `mj_core_schema_server_object_types.*` namespace import). This fixes the `TS2304: Cannot find name` build break that occurred when a base Open App was generated alongside a dependent app that foreign-keys into it (the base emitted fields typed with the dependent's classes, which its package cannot import), and it does so without dropping legitimate cross-schema relationships in a monolith, a multi-schema single app, or a single run that co-generates several apps into one file.
+
+- **New opt-in `includeSchemas` config (positive scope).** When set, CodeGen processes only the listed schemas; it is resolved into `excludeSchemas` at a single point before the metadata and file-generation phases, so it is pure sugar over the existing exclude mechanism (in-scope ⇔ named in `includeSchemas` and not in `excludeSchemas`; no implicit includes — the core schema must be listed explicitly). Lets an Open App scope its CodeGen to its own schema without hand-maintaining an exclude list of every other installed app.
--- a/packages/CodeGenLib/src/Config/config.ts
+++ b/packages/CodeGenLib/src/Config/config.ts
@@ -459,9 +459,21 @@ const configInfoSchema = z.object({
     { name: 'auto_index_foreign_keys', value: true },
   ]),
   excludeSchemas: z.string().array().default(['sys', 'staging']),
+  /**
+   * Opt-in POSITIVE scope list. When set (non-empty), CodeGen processes ONLY these schemas — every other
+   * schema present in the database is treated as excluded. It is pure sugar over {@link excludeSchemas}:
+   * it is resolved into `excludeSchemas` at ONE point (ManageMetadataBase.manageMetadata) by appending
+   * every not-included schema, so nothing downstream changes. In-scope ⇔ named in `includeSchemas` AND not
+   * in `excludeSchemas` (include shrinks the addressable space; exclude overlays on top). No hidden
+   * auto-includes — a schema (including the MJ core schema) is in scope ONLY if listed explicitly. Leave
+   * undefined/empty to use the classic exclude-only behavior (unchanged). Primary use: scope an Open App's
+   * CodeGen to just its own schema without hand-maintaining an exclude list of every other app. See PR
+   * "CodeGen includeSchemas for Open Apps".
+   */
+  includeSchemas: z.string().array().optional(),
   excludeTables: tableInfoSchema.array().default([
     { schema: '%', table: 'sys%' },
-    { schema: '%', table: 'flyway_schema_history' }
+    { schema: '%', table: 'flyway_schema_history' },
   ]),
   customSQLScripts: customSQLScriptSchema.array().default([
     // AS OF 5.3.0 we are NOT including this as it wipes out standard views and procs
--- a/packages/CodeGenLib/src/Database/schema-scope.ts
+++ b/packages/CodeGenLib/src/Database/schema-scope.ts
@@ -1,0 +1,39 @@
+/**
+ * Schema-scope resolution helpers for CodeGen.
+ *
+ * The `includeSchemas` config option is an opt-in POSITIVE scope list. Rather than add a second filtering
+ * path throughout CodeGen, it is resolved INTO the existing `excludeSchemas` list at a single point
+ * (`ManageMetadataBase.manageMetadata`). This module holds the pure resolution logic so it can be
+ * unit-tested without a database or config globals.
+ */
+
+/**
+ * Resolve an opt-in `includeSchemas` positive scope into the list of schema names to ADD to
+ * `excludeSchemas`. A schema stays IN SCOPE iff it is named in `includeSchemas` AND not already in
+ * `existingExcludeSchemas`; every other schema in `allSchemas` (the schema universe present in the
+ * database) is returned for exclusion. Matching is case-insensitive + trimmed, and the result is
+ * de-duplicated. No hidden auto-includes — the MJ core schema is excluded unless it is listed explicitly
+ * (include shrinks the addressable space; `excludeSchemas` still overlays on top).
+ *
+ * @param allSchemas             every schema name present in the database
+ * @param includeSchemas         the opt-in positive scope (assumed non-empty by the caller)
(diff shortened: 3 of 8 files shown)


MarceloT-BC:
@AN-BC  @rkihm-BC 
This is a meaningful codegen change that has the potential to effect production deployment. I have instructed claude to give a detailed PR message, and I need your review on this. I am not fully aware of hour our production deployments work. This case turned out to be more complicated than I originally thought. This was not as simple ad flipping a boolean vairable. There are some cases that work and some that do not for cross app FK traversal. Claude code identified a solution that I think works, but it needs your review.

AN-BC:
Closing this in favor of #3350, which rebuilds the same work cleanly on current `next`. Not a rejection — the core insight here is the better one and is carried forward intact. Details, since the reasoning matters more than the outcome:

**What happened underneath this branch.** This PR branched from `next` at `2ded59e1` (Jul 27, 10:57). Later the same day `f96804b9c` — "fix(codegen): omit relationship members whose GraphQL type was withheld" (18:30) — landed on `next` and fixed the same TS2304 break from a different angle, via `isRelatedTypeOutOfScope`. That's why this went un-mergeable. Nobody did anything wrong; two fixes for one bug crossed in flight.

**Your mechanism is the more correct of the two, and that's what #3350 keeps.** `next`'s predicate infers "will this class be emitted?" from the `entityPackageName` schema→package map. But `runCodeGen` narrows the generated entity list by *both* that map and the `excludeSchemas`/inclusion filters (`runCodeGen.ts:499-511`), so a predicate modeling only the package map still emits dangling references for anything dropped by the other filter. Your `generatedEntityNames` set is the actual array handed to the generator, so it cannot drift from what was emitted. #3350 keeps `next`'s method name, signature shape and drop-comment output, and swaps in your set-membership body.

Your `!isInternal` refinement was also right and `next` is missing it: in the core pass there's no `mj_core_schema_server_object_types` namespace import to resolve through (that file *is* the module), so a core child must satisfy set membership like anything else. That's carried over too.

**Two things #3350 does differently, both worth explaining.**

1. **The signature change is additive rather than a replacement.** This PR removes `excludeRelatedEntitiesExternalToSchema` from `public generateGraphQLServerCode` / `generateServerEntityString`. `GraphQLServerGeneratorBase` is documented as a `RegisterClass` extension point, so a customer subclass overriding those with the old signature would get a `Set<string>` where it expects a `boolean` — a silent break shipped under a `patch` changeset. #3350 adds `availability` as an optional trailing parameter instead; omitting it preserves the old heuristic exactly, and there's a test pinning that fallback.

2. **`includeSchemas` resolves against the database, not `Metadata.Entities`.** This is the one substantive defect, and it hits the feature's headline use case. `createNewEntities()` discovers tables with no `EntityID` straight from the DB and filters them only through `excludeSchemas`. Resolving the include list from `md.Entities` gives a universe containing, by definition, only schemas that *already* have entities — so a schema MJ has never seen (exactly the "a deployed app's CodeGen would traverse a client's own schemas" case in your description) is never added to the exclude list. It gets adopted into metadata on the first run, and then, because it now *does* have entities, correctly excluded on every run after — leaving those entity records permanently orphaned. #3350 runs the authoritative resolution inside `manageMetadata` against `SELECT DISTINCT SchemaName FROM vwSQLTablesAndEntities`, before `createNewEntities()`, with a second idempotent pass in `runCodeGen` so `--skipdb` still gets scoped. Worth noting your `config.ts` docstring already said the resolution belonged in `ManageMetadataBase.manageMetadata` — the doc described the right design, the code just landed elsewhere.

Also bumped the changeset to `minor`, since `includeSchemas` is new config surface.

**On the default-off question** that prompted the review: we swept the repo for consumers of the generated reverse-relationship arrays. There are five, all on core `__mj` types — `MJUserRoles_UserIDArray` in the `CurrentUser` query (with a hand-written override in `UserResolver`), and the four `MJQuery*Array` fields in `graphQLSystemUserClient` / `QuerySystemUserResolver`. The generic mechanism (`BaseEntity.InnerLoad(key, EntityRelationshipsToLoad)` → `getRelatedEntityString`) is plumbed throughout but has **zero** call sites passing a non-empty array. So turning these off for non-core entities would break nothing in-repo — but that's a public GraphQL surface and we can't see customer clients from here, so we're not flipping the default. Open Apps get the same practical outcome by opting into `includeSchemas`.

Your PR description's per-scenario analysis was genuinely useful in reviewing this — the monolith and co-generated cases are called out in #3350 because you'd already worked out why they're safe. The `MJ-UPSTREAM.md` note about destructive CodeGen paths assuming full-database visibility is untouched and still stands as the right follow-up before CodeGen scope narrows further.

---

PR #3305, opened 2026-07-27
feat(conversations): white-label theming tokens + chrome toggles for the conversation list

cmbrock-BC:
## Summary

Lets an embedding host restyle and pare down the conversation **list panel** through the component contract instead of `::ng-deep` on internal class names. Nine new design tokens plus four chrome toggles. Additive: with no tokens set and all toggles at their defaults, rendering is unchanged — one disclosed exception below.

## Changes

- **Nine public tokens** — `--mj-chat-list-{bg,ink,hover-bg,active-bg,active-ink,active-hover-bg,accent,accent-ink,accent-hover}`. ~40 rules re-pointed; hover, border, divider, preview, badge, and placeholder tints derive from `ink` (or `active-ink` on the active row) via `color-mix`, so they follow a remap automatically. They resolve in the component's `:host` onto private `--conv-list-*` names — a custom property can't name itself in its own fallback without forming a cycle, so the two-name indirection is required, not stylistic.
- **Four chrome toggles**, all default `true`: `showSearch`, `showNewConversationButton`, `showHeaderMenu`, `showSectionHeaders`. The header strip is removed entirely when nothing would occupy it. `showSectionHeaders=false` renders flat and fully expanded — the folder tree's root drop-zone and New Folder action live in the section header, so a headerless tree would let a user nest a folder with no way to drag it back out.
- **Sidebar** passes all four through and now re-emits the list's `(conversationDeleted)` (payload = the deleted ID) and `(refreshRequested)`, which previously died at that boundary.
- Docs: package README (toggle + token tables) and `guides/CONVERSATIONS_UX_STACK_GUIDE.md` §9.

## One intentional default-state change

The `renameHighlight` keyframes were hardcoded `rgba()` values that ignored theming entirely — inline `styles:` blocks in `.ts` files escape `check:ui-tokens`, which only scans `.css`/`.scss`. They now run on brand/status tokens at the same alpha ramp: same animation, theme-following hue. A node-side guard test greps the inline styles so the violation class can't return.

## Question for review — @MattC-BC

Two theming decisions worth your call before this merges:

1. **Namespace.** I put these under `--mj-chat-*` to sit with the existing 12 rather than opening a `--mj-conversations-*` family — there was no `--mj-conversations-*` precedent in the package.
2. **Declaration site.** These resolve in the component's `:host`, not the runtime-injected `:root` block in `conversations-runtime-bootstrap.service.ts` where the other 12 live. The indirection itself is forced (the var() cycle), but *where* the family is declared isn't — and Theme Studio may want to own it.

Both are cheap to change now and expensive after this publishes.

## Tests

73 files / 904 passing. DOM specs cover each toggle, the empty-header-strip case, the flat-list forcing, and that active-row rules derive from `active-ink` rather than panel ink. `npm run build`, `check:ui`, `check:ui-tokens` all clean.

## Live testing

Ran this in the Betty application and changed theming colors to confirm everything changes as expected. Also checked the Folders and Ungrouped section headers still render and collapse normally with the defaults on.


AN-BC:
@cmbrock-BC I'm waiting for Matt's review, don't tag me on a PR or for review until it has passed others please or I get pinged too early (now) and again when it is actually ready for my review.

MattC-BC:
> @cmbrock-BC I'm waiting for Matt's review, don't tag me on a PR or for review until it has passed others please or I get pinged too early (now) and again when it is actually ready for my review.

@AN-BC @cmbrock-BC - this looks good to me - thanks Colin!

PR #3176, opened 2026-07-17
feat(theme): brand theming system — ThemeEngine, Theme Studio, and live theme bridging into React

CaelebB-BC:
## Overview

A full brand-theming system for MemberJunction. Brands are defined by a small set of **seeds** (primary/accent hues, neutral chroma, radius, fonts, …) and the entire `--mj-*` design-token contract — OKLCH ramps, semantic tokens, state families, dark re-point, chart palette, shadows, radii — is **derived** at load. Themes are authored in a new **Theme Studio** dashboard, stored via a new **Theme** entity, and flow through to Angular chrome and React interactive components (including dark mode and state families) instead of frozen defaults.

### `@memberjunction/theme-engine` (new package)
Framework-agnostic, zero-runtime-dependency derivation of the full `--mj-*` token contract from ~8 brand seeds. The same code runs in the Angular theme builder (live preview) and is ready for a future server-side overlay endpoint.
- `derive()` — seeds → resolved light/dark semantic tokens (OKLCH ramps → hex)
- Per-mode + hover WCAG contrast report with clamp suggestions
- `emitOverlayCss()` — CSS scoped to `[data-theme-overlay="…"]`
- Logo derivation and color/contrast utilities, with unit tests

### Theme Studio + Theme Manager dashboards
Explorer dashboards to create, edit, preview, and manage brand themes.
- Live token preview, custom-CSS authoring with autocomplete over derived `--mj-*` tokens and curated MJ chrome selectors
- A protected, seeded built-in MJ theme that always exists as a safe fallback (edit/delete guarded; duplication allowed)
- Registered as a Custom-resource application via metadata

### Theme entity + migration
- New migration adding the Theme entity that stores brand **seeds** (not tokens)

### Live theme bridging into React interactive components
- `BuildStylesFromTheme()` derives `ComponentStyles` from the host's live `--mj-*` tokens
- Auto-themes antd via `wrapWithLibraryThemeProviders` (light/dark)
- `MJReactComponent` memoizes bridged styles and re-renders on theme change
- Adds `--mj-viz-*` categorical palette tokens; resolves Venn/chart colors from them
- Extends `ComponentStyles` with `chartPalette`, `link`, and `primaryActive`

### Supporting Angular/shared changes
- `theme.service.ts` runtime wiring for applying/overlaying derived themes
- New design tokens in `_tokens.scss`; typography, shell, and profile-dialog adjustments
- Minor consistency touch-ups across generic components (file-storage, find-record, resource-permissions, templates grid)

[...]

PR #3306, opened 2026-07-27
feat(conversations): brandable conversation export + additive headerActions slot

cmbrock-BC:
## Summary

White-label branding for exported conversations across all four formats, plus a seventh chat slot for host header buttons. An unbranded export is unchanged from today.

## Changes

- **`ExportBranding`** (`brandTokens` / `logoUrl` / `title` / `trademark`) via `exportBranding` on `mj-conversation-chat-area`. HTML gets a sanitized `:root{}` theme block, an inlined data-URI logo, the title, and a styled trademark footer; markdown/text carry title + trademark (markdown also references the logo); JSON emits a `branding` block.
- **HTML export stylesheet converted to `var(--mj-token, <legacyHex>)`**, every fallback being the exact prior literal — so with no `:root` block emitted the file renders identically to previous releases. JSON/markdown/text are byte-identical unbranded.
- **`headerActions` slot** (7th) — additive host buttons inside the *default* header's action strip. A projected `header` slot replaces the whole header and therefore suppresses it.
- `exportButtonLabel` / `exportButtonIcon`; public download-free `ExportService.BuildExportContent()` and `SnapshotBrandTokens()` seams.
- Modal: one "Include branding" switch in the general options — it affects every format, so it doesn't belong under HTML Options. `includeCSS` stays HTML-only and is resolved at export time rather than mutating the user's choice.

## Security

Host CSS values baked into the exported document pass a **safe-function allowlist** (`var`/`calc`/`min`/`max`/`clamp`/`rgb`/`rgba`/`hsl`/`hsla`/`hwb`/`lab`/`lch`/`oklab`/`oklch`/`color`/`color-mix`) plus rejection of `<>{};@\`. Blocking only `url()` would have let `image-set()`, `-webkit-image-set()`, `cross-fade()`, `image()`, and `src()` reach the network from an exported local file without ever writing the literal `url(`. Keys must match `^--[A-Za-z0-9_-]+$`; a rejected entry is dropped whole, not stripped. Logo inlining requires `content-type: image/*`.

Disclosed in the docstrings: markdown branding values aren't escaped (host config, not end-user input), and `inlineLogo` has no size/timeout guard and falls back to the raw URL on failure.

@MattC-BC — smaller Theme Studio surface here if you want it: the export stylesheet now reads `--mj-*` tokens, and `DEFAULT_EXPORT_THEME_TOKENS` defines the 7-token snapshot baked into a themed export. Worth a glance at whether those seven are the right set.

## Tests

72 files / 911 passing, including full-palette parity assertions and injection cases for every bypass vector above. Build + both UI gates clean.

## Live testing

Tested on the local Betty version. Exported with branding on and off, and checked the branding toggle shows up in the main options and applies to markdown too, not just HTML. The unbranded export looks the same as it always has.


MattC-BC:
Hey @cmbrock-BC - took a look, this is nice work. It plays perfectly with the Theme Studio work in #3176 with zero extra wiring. Your snapshot reads the same --mj-* tokens the brand overlay writes, and your sanitizer already allows the color-mix()/rgba() values our derived themes resolve to. A themed session exports themed files automatically.

One real bug: if a user exports while in dark mode, the snapshot bakes dark text colors into a white page and the file comes out unreadable. Rather than silently picking a mode, let's let the user decide - add a Light/Dark option to the export dialog (default Light) and snapshot the chosen mode's values instead of whatever the app happens to be in. To make a dark export actually coherent you'll need two small additions anyway: --mj-bg-page in the token set and a body { background: var(--mj-bg-page, #fff); color: var(--mj-text-primary, #333); } rule. For grabbing the non-active mode's values: toggle data-theme on documentElement, read getComputedStyle, restore - it's synchronous so nothing paints in between.

On your "are these 7 right" question - close. I'd add --mj-bg-page (needed above) and --mj-font-family with a body font rule, so themed exports carry the brand's type too (font stacks degrade fine on machines without the font). Take or leave: --mj-radius-md for the message card corners, and switching the user-message tint from --mj-status-info-bg to var(--mj-brand-accent-subtle, #e3f2fd) - status colors never re-theme, so info-blue bubbles would stay MJ-blue in every brand.

cmbrock-BC:
Got it Ill add those changes you mentioned, thanks for taking a look

PR #3307, opened 2026-07-27
feat(conversations): assistant identity inputs for the message feed

cmbrock-BC:
## Summary

Two inputs let a white-label host brand the AI side of the message feed through the component contract instead of `::ng-deep` on `.message-sender` / `.avatar-circle`. Both default `null` = today's engine-resolved identity.

## Changes

- `assistantDisplayName` and `assistantAvatarUrl` on `mj-conversation-chat-area`, threaded chat-area → message-list → message-item.
- **Display-only by design.** `isConversationManager` and the run-details header/record link keep comparing the *engine-resolved* name, so relabeling a persona can never change routing or behavior. `aiAgentInfo` applies the override on top of a new `engineAgentInfo`.
- The AI avatar branch mirrors the existing user-avatar pattern; `.avatar-has-image` drops the brand fill and clips to the tile radius so nothing peeks out behind a transparent image. Whitespace-only or failing URLs degrade to the agent icon.
- Identity changes **restamp already-rendered messages** — message items are created dynamically with these stamped as static config, so a branding config resolving after first render still lands. Implemented with input setters, not `ngOnChanges`, per `packages/Angular/CLAUDE.md`; that also covers hosts setting the property imperatively.
- Docs: README + `guides/CONVERSATIONS_UX_STACK_GUIDE.md` §8, including what the override deliberately does *not* cover (run-details header, role tooltip, realtime session cards).

## Tests

73 files / 899 passing. Covers override precedence, whitespace handling, the `(error)` fallback, that the conversation-manager check is unaffected, and the restamp path (change / no-op / clear-to-null / pre-render / immediately-post-construction).

## Live testing

Both inputs are bound to a per-tenant persona name and avatar, and I deleted the two `::ng-deep` pierces we had been using for this the message feed renders the same as before, just through the inputs now. Tested within the Betty app and everything worked as expected, both with custom and with default (tested with Betty and Sage).


.changeset/conversations-assistant-identity.md
```diff
@@ -0,0 +1,5 @@
+---
+"@memberjunction/ng-conversations": minor
```
AN-BC, line 2: @cmbrock-BC letting this through as it doesn't negatively impact 5.50 but this is a code only change with no metadata or migration so it should be patch

AN-BC (approved):

PR #3309, opened 2026-07-27
feat(conversations): per-type mention gating on the chat area

cmbrock-BC:
## Summary

`allowAgentMentions`, `allowEntityMentions`, and `allowSkillCommands` (all default `true`, all under the pre-existing `allowMentions` master) forward host-level caps to `mj-ai-composer`'s existing `Enable*` toggles.

## Changes

- Forwarded through every composer consumer: the empty state and both message inputs.
- **No composer or trigger-provider logic changed.** The composer already treated `@` agents, `#` entities, and `/` skills as three independent plugins — only the forwarding was missing.
- Docs: README toggle table, and `guides/CONVERSATIONS_UX_STACK_GUIDE.md` §5 now notes that an `@` outranks the whole default-agent resolution chain.

## Why

A surface pinned to one agent via `[defaultAgentId]` hasn't really pinned it while `@` is available — an `@` mention outranks every step of the resolution chain. This lets a host keep `/` skill commands while dropping `@`. Disabling a type **removes its trigger provider** rather than hiding UI, and `allowMentions=false` still disables all three regardless of the per-type flags (enforced one layer down in `mention-editor`, whose `activeProviders` returns `[]`).

## Tests

72 files / 889 passing. A behavioral spec on the composer seam proves each toggle includes/excludes exactly its provider (skills-only = 1 provider), plus a contract spec pinning the forwarding at every hop.

## Live testing
Ran in the Betty app `/` brings up the skill picker and `@` does nothing, which is what was needed an `@` was overriding the agent pinned. Allowed Betty to always be the agent that responds but gives access to skills, with no threat of calling another agent.


.changeset/conversations-mention-type-gating.md
```diff
@@ -0,0 +1,7 @@
+---
+"@memberjunction/ng-conversations": minor
```
AN-BC, line 2: same comment here @cmbrock-BC  this is `patch`

AN-BC (approved):

PR #3150, opened 2026-07-14
fix(eds): resolve external object names per-driver; dedupe external entity import

JosueGarcia-BC:
## Summary

Fixes two **pre-existing** External Data Sources bugs, independent of the Microsoft Fabric work (#3110, already merged to `next`). Both are framework-level and affect any SQL-backed external entity — not just Fabric.

## Fix 1 — non-default-schema objects were read against the wrong schema

**Symptom:** an external entity whose remote object lives in a non-default schema (e.g. medallion `bronze`/`silver`/`gold`, or any multi-schema source) failed at read time with `Invalid object name`.

**Root cause:** the read router passed a bare object name to the driver, and the SQL driver's `qualifyObject` qualifies a bare name with the data source's `DefaultSchema`. So a `bronze.` object was read as `dbo.` and missed.

**Fix (driver-aware):** object-name resolution now lives on the driver via a polymorphic `ResolveObjectName(entity)`:
- `BaseExternalDataSourceDriver` returns the name verbatim (`ExternalObjectName ?? BaseTable ?? Name`) — correct for stores where the name is a **literal** identifier (e.g. a MongoDB collection).
- `BaseSqlExternalDataSourceDriver` **overrides** it to schema-qualify a bare name with the entity's `SchemaName` (reusing the base fallback chain via `super()`), so non-default schemas resolve correctly across all five SQL drivers.
- `ExternalDataSourceReadRouterImpl` **delegates** to `driver.ResolveObjectName(entity)` on both read paths (`RunViewExternal` + `LoadExternalRecord`) and never munges the name itself.

**Why it's correct and safe:**
- **Symmetric with introspection:** CodeGen introspects the remote object using the same `entity.SchemaName` (`manage-metadata.ts`), so the schema that located the object at CodeGen time is the schema that qualifies it at read time. A mismatched `SchemaName` never syncs fields, so it can't be a working entity that this regresses.
- Composes with `qualifyObject`, which splits a dotted name and quotes each part (`bronze.sales` → `[bronze].[sales]`); no double-qualification.
- **Non-SQL drivers are unaffected:** MongoDB inherits the verbatim default, so a collection read is never schema-mangled. (`Entity.SchemaName` is `NOT NULL DEFAULT 'dbo'`, so a router-side qualification would otherwise have silently targeted a non-existent `dbo.<collection>` — this design prevents that class of bug entirely.)

## Fix 2 — CodeGen emitted a duplicate base-class import with 2+ external entities

[...]

PR #2770, opened 2026-06-05
docs: Query & Entity Materialization design plan

AN-BC:
## Summary

Adds `plans/query-entity-materialization.md` — a unified **design / decision record** for a materialization capability in MJ: persisting *proven-hot* stored-Query results and entity base views into MJ-managed physical tables (wrapped in views, refreshed on a schedule) to buy back execution speed on hot analytical paths **without** abandoning the normalized base or reintroducing ETL.

This is a **planning doc only** — no code changes. It supersedes a standalone draft and folds in a detailed review against how MJ actually works (stored Queries + the extraction pipeline, CodeGen's cross-engine view/SP generation, RLS, the scheduler, and the existing query cache).

## Key decisions captured

- **CodeGen-driven, not runtime.** Base-view materialization is declared via `additionalSchemaInfo`; query materialization is handled by a new CodeGen section that scans `Query.IsMaterialized`. Keeps materialization on MJ's disciplined schema-management rails.
- **Materialized results are real entities.** A materialized query becomes a read-only `VirtualEntity`, so it gets `RunView`/`RunViews`, RLS, server-side caching, drill-down, and GraphQL **for free** — no parallel read path. Base-view materialization reuses the existing entity.
- **RLS = standard entity RLS on the new shape.** Source-entity RLS does *not* carry over (you can't project per-row entitlement onto an aggregate); it's authored fresh on the new entity. Includes a **mandatory promotion-time downgrade gate** (using `QueryEntity` provenance) so materializing over an RLS-protected source can't silently become a data-leak vector.
- **`DataSource: 'Live' | 'Materialized'`** read-path param for the base-view case (explicit door, no silent routing to stale data).
- **Keys:** combined-key-set surrogate hashing for fan-out joins and aggregates, with a cross-engine determinism requirement.
- **Refresh:** reuse `ScheduledJobEngine` (a `MaterializationRefreshDriver`), **not** MJ Audience (which is list-delta + messaging, not a sync engine). Shadow-table atomic swap via cheap view repoint on both engines.
- **`MJ: Materialized Results`** entity as the auditable metadata edge + machine-readable selection contract for Skip/agents.
- **Positioned distinctly from the existing query cache** (`CacheEnabled`/`CacheValidationSQL`); the smart-cache fingerprint is reused as the incremental-refresh watermark.
- **Migrations own table/index DDL**, with a CodeGen minimal-table fallback (single source of truth, no parallel index-apply path).

## Parameterization & verification

[...]

PR #3311, opened 2026-07-27
feat: Query & Entity Materialization (cross-engine)

JosueGarcia-BC:
# Query & Entity Materialization

Materializes a stored Query's result (or an entity's base view) into a **physical table that is its own read-only entity** — refreshed on a schedule via a shadow-table build + atomic wrapper-view swap. Cross-engine (SQL Server + PostgreSQL), with one PG caveat below. Closes the EDS cross-source-join gap: mirror external data locally, then join it with internal data.

Design doc: [`plans/query-entity-materialization.md`](plans/query-entity-materialization.md). Feature issue: #2770.

## What's in it

**New package `@memberjunction/materialization`** — the runtime refresh engine:
- Full-rebuild refresh (shadow table → atomic `CREATE OR ALTER/REPLACE VIEW` swap), cross-engine.
- **Keyed aggregations**: combined-key `SHA2_256` surrogate hashing (within-engine deterministic; §17.1 canonical serialization), with two incremental strategies — `DirtyGroupRecompute` (delete+reinsert changed groups) and MERGE-upsert `Incremental` (in-place, for additive aggregations).
- `MaterializationFreshness` — advisory mixed-freshness inspector for agents composing multi-entity reads.

**CodeGen** (`@memberjunction/codegen-lib`):
- Materializes flagged stored Queries + entity base views: cross-engine minimal-table DDL, wrapper view, read-only Virtual Entity minting, migration-reuse detection.
- Parameterization (Bucket 1): row-filter params → materialize **broad** + read-time predicate (render-and-diff verification).
- Aggregation-key auto-detection → `KeyColumns`; strategy assignment (additive → `Incremental`, else `DirtyGroupRecompute`, else `FullRebuild`).
- **RLS-downgrade gate** (§6.2) — refuses to materialize a query over RLS-protected sources.
- **Drift detection** → `DriftHold` flag-and-hold (§13/§17.2): a materialization whose source shape drifts is held for review, not silently rebuilt.

**Read path**: `RunViewParams.DataSource: 'Live' | 'Materialized'` (`@memberjunction/core`), routed by `GenericDatabaseProvider.GetEffectiveBaseView`, plumbed through GraphQL (`@memberjunction/server`, `@memberjunction/graphql-dataprovider`).

**Scheduling** (`@memberjunction/scheduling-engine`): `MaterializationRefreshScheduledJobDriver` sweeps due materializations (skips `Disabled` / `DriftHold`).

**Metadata**: `MJ: Materialized Results` entity + `Query.IsMaterialized` / `MaterializedResultID` (generated types + forms).

## Phasing (all shipped in this PR)
- **1 — Foundation**: entity/flags, CodeGen DDL + minting, scheduled-job full-rebuild refresh, read-only enforcement, `DataSource` read path, RLS gate.
- **1.5 — EDS composition**: materialize external entities/queries (mirror external → join locally).
- **2 — Parameterization**: row-filter params (materialize broad + read-time predicate).
- **3 — Aggregations & keys**: hashing, key auto-detection, `DirtyGroupRecompute`.
- **4 — Incremental & drift**: MERGE-upsert `Incremental` refresh, `DriftHold` drift automation, mixed-freshness flagging.

Deferred by design (§17): Bucket-2 per-value cache, AI-suggested promotion.

## Testing
- **Full build**: 267/267 (on current `next` + this work).
- **Clean-DB migration**: fresh SQL Server DB → `mj migrate` applies all migrations in strict order, the 4 materialization migrations landing after `next`'s latest (0 failures); `MaterializedResult` table + `Query` flags verified present.
- **Unit tests**: all materialization suites green (analysis, drift, broad-render, param classify/verify/qualify, refresher, freshness, GenericDatabaseProvider, PG/SS CodeGen materialization).
- **Live-validated on both engines** (SQL Server + PostgreSQL): hash determinism (incl. PG session-TZ / lc_monetary invariance), `DirtyGroupRecompute`, `Incremental` MERGE / `ON CONFLICT`.
- **Full E2E pipeline** (SQL Server): real `mj codegen` → mint (`Incremental`, `KeyColumns`) → `RefreshOne` baseline + incremental → read-back via `DataSource:'Materialized'`.
- **DriftHold E2E**: drifted materialization → `mj codegen` → flagged `DriftHold`; intact one stayed `Active`.
- **GraphQL**: MJAPI boots with the materialization resolvers integrated into the schema (`buildSchemaSync` accepts them).
- **High-effort adversarial code review** (multi-agent, independent verify): 10 findings triaged; **9 fixed + unit-tested + live-validated on both engines**, 1 documented as a known v1 limitation. Fixes:
  1. RLS-downgrade gate now **fails closed** when a query's source-entity provenance is empty (can't prove RLS-safety → refuse) — closes a silent privilege-escalation path.
  2. NULL bit/boolean key column no longer hashes identically to `false` (explicit `IS NULL` branch → null sentinel).
  3. Full rebuild now **restores the surrogate UNIQUE index** post-swap (the `MERGE`/`ON CONFLICT` incremental upsert depends on it; `SELECT…INTO`/`CREATE TABLE AS` had dropped it).
  4. Keyed surrogate is typed to match the hash it holds (`varchar(64)` SS / `text` PG), not `int IDENTITY` — no PK-type divergence from the minted entity.
  5/6. `DataSource:'Materialized'` now routes the **aggregate** query and the **saved-view run-logging** read to the snapshot (both previously fell back to the live base view).
  7. Full-rebuild watermark is captured **before** the source read (was after → a concurrent update could be excluded from every future incremental pass).
  8. SQL Server swap is now a **single transaction** (view only ever on the canonical name) — closes the "Invalid object name …__shadow" window a concurrent reader could hit mid-swap.
  9. External-source rebuild is now **parameterized** (`@pN` / `$N` bind params, `NULL` as literal, batched by the engine's bind-parameter ceiling) instead of inlining rows as SQL literals — the SQL text stays small/constant regardless of row width, so a large external mirror no longer risks Node-heap/parser/packet blowups, and values are injection-safe by construction.
  - Documented limitation (one): incremental **delete-detection** is net-count-based — a delete balanced by an insert lags until the next full rebuild; delete-heavy sources should use `FullRebuild` (a persisted forced-rebuild cadence is the follow-up, needs a new column). *(The external inline-`VALUES` scale limit noted earlier is now fixed by #9 above.)*
  - Re-validated live on **SQL Server** and **PostgreSQL** (transactional swaps; NULL/false distinct; unique index; MERGE/`ON CONFLICT` upsert; parameterized external rebuild — quote/SQL-ish value bound verbatim).
- **A second adversarial review pass** (on the reassembled branch, weighted to the newly-written fix code) found **8 more issues, all fixed + validated on both engines**: (1) external-query **column-name identifier escaping** (`]`→`]]` / `"`→`""`) — a hostile column name can't break out of its quoting; (2) external-query **PG column typing** — SS-style types are mapped once to native PG types (no double-conversion to `text`); (3) **`DataSource` now participates in the RunView cache fingerprint** — Live and `Materialized` reads no longer collide on one cache slot; (4) **`DirtyGroupRecompute` DELETE+INSERT runs in one transaction** — no partial-data window for concurrent readers; (5) **`SET XACT_ABORT ON`** on the SQL Server transactional swaps — a mid-swap error rolls back instead of orphaning the transaction; (6) external rebuild **restores the surrogate unique index** (parity with the full-rebuild path); (7) source Query loaded **once** per refresh (was twice for local queries); (8) migration header banners corrected to `v5.50.x`.
- **A third adversarial review pass** found **10 more, all fixed + validated**: (1) `mapSqlTypeToPostgres` now passes through PG-native type names — an external entity on a PG deployment no longer has its columns coerced to `text`; (2) **`DataSource:'Materialized'` reads bypass the result cache** — the snapshot is refreshed out-of-band, so a cached materialized result would go stale; (3) the scheduled-job driver now **checks `RunView.Success`** (a failed sweep fails loud instead of reporting green while refreshing nothing); (4) a **malformed refresh cron** now skips + reports the row instead of rebuilding it every sweep forever; (5) external-query rebuild **refuses a column that collides with the surrogate name**; (6) RLS gate **fails closed on any *unresolvable* source link** (not just zero links); (7/8) `inferSqlType` + `SourceRowCount` **widen to bigint** (int32-overflow guard); (9) resolve paths use `EntityByID` (O(1)); (10) drift check collapses an N+1 cascade to one set-based query.
- **Further review rounds (4–10) + manual audit.** Additional high-effort rounds fixed, at the root: the materialized-snapshot cache-staleness class in every transport (a shared `IsMaterializedDataSource` now drives every cache/read decision so Live vs Materialized can never cross-serve or cross-cache); watermark advanced **only on success** on both the full-rebuild and incremental paths (never persisted by a failed refresh); a unified failure-backoff path; collision-safe per-part surrogate hashing; NaN-safe freshness math; case-insensitive `DataSource`; query output-shape drift detection; a short per-table surrogate index name (avoids the 128-char `sysname` overflow); `stripTopLevelOrderBy` for the rebuild's derived-table wrap (strips only a *bare* ORDER BY — kept when `TOP`/`OFFSET` is present, since it then selects *which* rows); external queries forced un-keyed at mint; identifier escaping in the keyed builders; and a bigint-overflow guard in external type inference. Several of these corrected edges introduced by earlier fixes — each round's fixes were re-reviewed the next round.
  - Final unit-suite state: **Materialization 75, CodeGenLib 730, MJCore 1570, GenericDatabaseProvider 846**; full monorepo build **267/267**.
  - **Honest note on review status:** the later rounds trended to zero confirmed-correctness findings, but a formal *two consecutive fully-clean max rounds* was **not** reached — the org review budget was exhausted mid-run. A comprehensive **manual audit** then verified every prior fix is present and self-consistent in the committed diff (no lost/half-applied fixes, no stale symbols). **Recommended before merge:** one more adversarial `max` review pass + a fresh SQL Server + PostgreSQL end-to-end run on the final code.

## Known limitations & deferred (v1)

**Correctness items intentionally out of this PR's scope** (documented, not silently shipped):
- **PostgreSQL query-materialization *mint* is SQL-Server-only** — see the ⚠️ reviewer note below. Base-view materialization and the whole refresh engine are cross-engine; only the query-mat mint's `spCreateVirtualEntity` step is SS-only.
- **`bigint` count columns → GraphQL `Int`.** `MaterializedResult.RowCount` / `SourceRowCount` are `bigint`, but MJ CodeGen maps every `bigint` column to GraphQL `Int` (`graphql_server_codegen.ts`) — a **repo-wide framework mapping**, not a materialization bug. Reading a `MaterializedResult` row whose count exceeds 2³¹ would error at the GraphQL layer. Fixing it (bigint → Float/BigInt scalar) is a framework change affecting all entities → flagged as a **follow-up**, not done here.
- **SQL Server source `COUNT(*)`** (fingerprint/row-count probes) returns `int` and would overflow above 2³¹ rows on the *source* — should be `COUNT_BIG`. Same multi-billion-row regime as above; cheap follow-up (2 call sites) left out to keep scope tight.

**By-design v1 boundaries:**
- **Incremental delete-detection is net-count-based** — a delete *balanced* by an insert in the same window isn't localized until the next full rebuild; delete-heavy sources should use `RefreshStrategy='FullRebuild'`.
- **`MaterializationFreshness` (mixed-freshness advisory) is built but not yet wired** into a read path/resolver — inert until a consumer calls it (Phase-4 groundwork).
- **Bucket-2 per-value cache** (§17.1) and **AI-suggested promotion** (§17.4) are deferred phases.

**Cleanup / DRY follow-ups** (no functional impact, deliberately not refactored without an adversarial-review safety net): SS atomic-swap SQL is duplicated across two builders; a few dedup opportunities (`soleStatement`, the surrogate-column constant, `getTableColumns`); minor per-sweep query/parse repetition in CodeGen drift detection.

## Reviewer notes
- **Branch shape**: fresh single commit off current `origin/next` (this PR is authored clean, not a continuation of an earlier branch). The materialization work is orthogonal to recent `next` changes — it touches no files the last batch of `next` commits changed except two `package.json` dep additions.
- **One Windows-only test failure, unrelated to this PR**: `integration-engine` fails locally because a test does `mkdir` on a log-dir path containing a colon (`…\new-MJ: Company Integration Runs-id`), which is illegal on Windows — it passes on Linux CI. The branch does not modify `integration-engine`. (Full monorepo build is 267/267.)
- **Depends on #3150** (EDS driver-aware `ResolveObjectName`) for Phase 1.5 external reads. **Merge #3150 first**; the EDS files are intentionally NOT in this PR (they come from `next` once #3150 lands).
- **⚠️ PostgreSQL caveat — QUERY-materialization MINT is SQL-Server-only for now (raise at review).** A full PG codegen-mint E2E confirmed the runtime refresh engine (all strategies + hashing) and base-view materialization work on PG, and the query-mat table/wrapper-view + `text` surrogate mint correctly — BUT the query-mat mint's final step calls `spCreateVirtualEntity` to create the read-only VE, and that helper proc exists **only on SQL Server** (no PG counterpart in any baseline). So on PG the `MaterializedResult` row + back-link aren't written and query-mat mint can't complete. This is a **pre-existing MJ↔PG gap** (the same helper backs config-based virtual entities, equally SS-only on PG), not introduced here. **Follow-up options to discuss at review:** (a) port `spCreateVirtualEntity` to PG, or (b) replace the sproc call in the mint with cross-engine entity-layer VE creation. Base-view materialization + the refresh engine remain fully cross-engine.
- **PostgreSQL migrations** are generated by the `/pg-migrate` toolchain (`mj migrate convert` + `pg-migrations.yml`) — not hand-authored here.
- **Class-registration manifests** regenerate in the release build's `prebuild` (`mj codegen manifest`).
- **Migrations (`v5.50.x`, `202607231400–1403`) will need re-stamping before merge.** They were validated in strict order on a fresh SQL Server DB, but `next` has since merged later-dated migrations (`202607240220`, `202607241645`), so these four are now *earlier* than `next`'s latest → Flyway strict-in-order would reject them. Re-date them to after `next`'s newest migration when bringing the branch current (mechanical rename; done each time `next` outpaces them).
- Changeset: **minor** (new additive feature; `@memberjunction/*` fixed-version group).

Version bump: minor.


rkihm-BC (commented):
## Code review — Query & Entity Materialization (Phase 1)

_This is a detailed review performed with **Claude Code** — an independent multi-agent pass, with every finding re-verified by hand against the branch before posting._

Overall this is strong, unusually well-documented work; the fail-closed posture mostly holds and your own adversarial passes caught most of what matters. Findings below, with the specific lines flagged inline.

### 🔴 Merge blocker
**Migration re-stamp + version label.** `next` now carries `V202607311852__v5.51.x__Metadata_Sync`, which is later than this PR's `V202607291000–1005`. Flyway strict (in-order) validation will reject them. Re-stamp all 12 files (SS + PG) to a timestamp **after `202607311852`**, and — since the next release is **6.1.0** — relabel `v5.51.x → v6.1.x` in both the filename and the header banner. (The changeset is currently `minor`; confirm that's intended for a 6.1.0 cut.)

### 🟡 Security / correctness (inline)
- Minted-entity read permissions are scoped only on the first mint — a later revocation of source read access isn't re-applied.
- Mint-time table DDL doesn't escape identifiers, while the refresh path does (`quoteIdent`/`escId`).
- Incremental eligibility admits a view as the single source table (watermark can miss changed groups).

### 🔵 Lower risk (inline)
- SS incremental `MERGE` lacks `WITH (HOLDLOCK)` (PG `ON CONFLICT` is race-safe).
- The RLS-drift branch sets `DriftHold` before revoking read access — reverse so the readable window closes first.
- `RunViewParams.Equals` omits `DataSource` (latent; a Live↔Materialized toggle wouldn't trigger a reload).

### ✅ Verified correct
Watermark captured-before-read and advanced-only-on-success on both paths; DirtyGroupRecompute delete+insert transactional; NULL-vs-false hash distinction; surrogate index restored post-swap; the DataSource read-path/cache separation (single `IsMaterializedDataSource` helper) threaded end-to-end; the CodeName collision guard and `UUIDsEqual` usage; SS↔PG migration parity and generated-ORM alignment. I also checked the RLS mint-gate's use of `vwQueryEntities` (INNER JOIN) — the `FK_QueryEntity_Entity` foreign key guarantees it returns the same source-entity set as the drift path, so that is **not** a leak. The SQL-Server full-rebuild swap is genuinely atomic.

A few of these are judgment calls on your threat model (the DDL-escaping and MERGE items especially). If any land differently than I've described, I'd rather talk it through than have you take them on faith — please reach out and we can pair on any of it.

migrations/v5/V202607291000__v5.51.x__Query_Entity_Materialization_Foundation.sql
```diff
@@ -0,0 +1,143 @@
+/* ============================================================================
+   Query & Entity Materialization — Phase 1 Foundation (Metadata)
+   v5.51.x
```
rkihm-BC: **[Merge blocker] Re-stamp + version label.** `next` now has `V202607311852` (later than this file's `202607291000`), so Flyway strict-in-order validation will reject it. Re-date all 12 migrations (SS + PG) to after `202607311852`, and since the next release is **6.1.0**, change `v5.51.x → v6.1.x` in the filename and this header banner.

packages/CodeGenLib/src/Database/manage-metadata.ts
```diff
@@ -1268,6 +1311,559 @@ export class ManageMetadataBase {
...
+                  BaseView: viewName,
+                  SchemaName: coreSchema,
+                  PrimaryKeyFieldName: analysis.surrogateColumnName,
+                  Description: `Materialized result of query "${queryName}".`,
+               });
+               generatedEntityId = createRes.recordset?.[0]?.[''] || createRes.recordset?.[0]?.ID || createRes.recordset?.[0]?.Column0;
+               if (generatedEntityId) {
+                  mintedCount++;
+                  await this.addEntityToApplicationForSchema(pool, generatedEntityId, queryName, coreSchema, currentUser);
+                  await this.addMaterializedQueryEntityPermissions(pool, generatedEntityId, queryName, resolvedSourceEntities);
```
rkihm-BC, line 1711: **[Medium · security] Minted read permissions are scoped only on the first mint.** `addMaterializedQueryEntityPermissions` runs only inside `if (!generatedEntityId)`, so on later CodeGen runs (entity already exists) the grant is never recomputed. The RLS-drift re-check reacts to a source *gaining* RLS but not to a plain `CanRead` revocation on a source entity — so if a role loses read on a source after mint, the minted snapshot keeps granting it read. Consider recomputing the intersection grant every run (dropping rows no longer justified), or extending the drift check to diff current source read-access against the minted grants.

packages/CodeGenLib/src/Database/providers/sqlserver/SQLServerCodeGenProvider.ts
```diff
@@ -147,6 +148,61 @@ FROM
...
+     * is absent) so a migration-provided `materialized_<Name>` table with bespoke indexing
+     * is detected and reused rather than clobbered (plan §12). Emits the single-column
+     * surrogate PRIMARY KEY, which is itself the required unique index.
+     *
+     * Returns a single GO-free batch: it is executed directly via `LogSQLAndExecute`
+     * (`ds.query`, which doesn't split on `GO`), and the caller passes
+     * `includeBatchSeparator: true` so the migration *file* still gets a `GO` after it.
+     */
+    generateMaterializedTableSQL(schema: string, tableName: string, columns: MaterializedColumnSpec[]): string {
+        const colLines = columns.map((c) => `        [${c.Name}] ${c.SQLType} ${c.Nullable ? 'NULL' : 'NOT NULL'}`);
```
rkihm-BC: **[Medium · security] Mint-time table DDL doesn't escape identifier delimiters, but the refresh path does.** `[${c.Name}]` here (and `pgDialect.QuoteIdentifier(c.Name)` in the PG generator, which is a bare `"${name}"`) emit column names unescaped, whereas `MaterializationRefresher.quoteIdent`/`escId` correctly double `]` / `"`. Column names flow from `QueryField.Name`, which for **external** queries originate from untrusted remote-schema introspection — a name containing `]` / `"` breaks out of the quoted identifier during a CodeGen-privileged `CREATE TABLE`. Suggest mirroring `quoteIdent` here for parity with the sibling path you already hardened.

packages/Materialization/src/MaterializationRefresher.ts
```diff
@@ -0,0 +1,1274 @@
...
+        const updatedAtColumn = '__mj_UpdatedAt';
+        const cols = await this.getTableColumns(src.schema, src.table, exec);
+        if (!cols.some((c) => c.toLowerCase() === updatedAtColumn.toLowerCase())) return null;
+        return await this.probeSourceFingerprint(src.schema, src.table, updatedAtColumn, exec, isPostgres);
+    }
+
+    /** Extracts the single source table of an aggregation SELECT, or null if it isn't exactly one table. */
+    private resolveSingleSourceTable(sql: string, platformKey: string | undefined): { schema: string; table: string } | null {
+        const refs = SQLParser.ExtractTableRefs(sql, GetDialect(platformKey ?? 'sqlserver'));
+        if (!refs || refs.length !== 1 || !refs[0].TableName) return null;
```
rkihm-BC, line 634: **[Medium] Incremental eligibility admits a view as the single source table.** `resolveSingleSourceTable` accepts any single table-ref, and `getTableColumns` (INFORMATION_SCHEMA) resolves columns for views too. If the aggregation's `FROM` is a view wrapping a join, `MAX(__mj_UpdatedAt)` over that view can yield a watermark that doesn't correspond to the rows actually read, so the strict `> watermark` filter can skip genuinely-changed groups (silent staleness, bounded by the forced-rebuild cadence). Consider restricting to `TABLE_TYPE = 'BASE TABLE'`. If this is the deferred base-view-staleness refinement you noted, disregard.

packages/Materialization/src/MaterializationRefresher.ts
```diff
@@ -0,0 +1,1274 @@
...
+        const q = (n: string) => MaterializationRefresher.quoteIdent(n, false);
+        const changedSince = `s.${q(opts.updatedAtColumn)} > ${opts.watermarkSql}`;
+        const match = MaterializationRefresher.buildKeyMatchPredicate('agg', 's', opts.keyColumns, false);
+        const hashExpr = MaterializationRefresher.buildHashKeyExpression(opts.keyColumns, false);
+        const insertCols = [opts.surrogateColumn, ...opts.dataColumns].map(q).join(', ');
+        const insertVals = [`src.${q(opts.surrogateColumn)}`, ...opts.dataColumns.map((c) => `src.${q(c)}`)].join(', ');
+        const setList = opts.dataColumns.map((c) => `t.${q(c)} = src.${q(c)}`).join(', ');
+        const selectList = [`${hashExpr} AS ${q(opts.surrogateColumn)}`, ...opts.dataColumns.map((c) => `agg.${q(c)}`)].join(', ');
+        return [
+            `MERGE INTO ${matTable} AS t ` +
```
rkihm-BC: **[Low · concurrency parity] SS `MERGE` has no `WITH (HOLDLOCK)`.** Under READ COMMITTED the NOT-MATCHED probe takes no range lock, so a MERGE racing another writer on the same surrogate can double-insert and hit the unique index. PG's `INSERT … ON CONFLICT` is race-safe by construction, so the two engines diverge. Serial per-materialization sweeps make this unlikely today; adding `WITH (HOLDLOCK)` on the target is the standard defensive fix.

packages/CodeGenLib/src/Database/manage-metadata.ts
```diff
@@ -1268,6 +1311,559 @@ export class ManageMetadataBase {
...
+            const qe = await this.runQueryWithParams(pool, `SELECT EntityID FROM ${this.qs(coreSchema, 'QueryEntity')} WHERE QueryID = @Q`, { Q: r.SourceQueryID as string });
+            const rlsVerdict = this.assessQuerySourceRLSSafety(md, qe.recordset.map((x: CodeGenQueryRow) => x.EntityID as string));
+            if (!rlsVerdict.safe) {
+               await this.LogSQLAndExecute(
+                  pool,
+                  `UPDATE ${this.qs(coreSchema, 'MaterializedResult')} SET ${this.qi('Status')}='DriftHold' WHERE ID='${r.ID}'`,
+                  `Flag materialization "${r.TableName}" as DriftHold (source RLS drift)`,
+               );
+               if (r.GeneratedEntityID) {
+                  await this.revokeMaterializedEntityReadAccess(pool, r.GeneratedEntityID as string, r.TableName as string, rlsVerdict.reason ?? 'source RLS drift');
```
rkihm-BC: **[Low] Revoke runs after the `DriftHold` update.** On RLS drift you set `Status='DriftHold'` first, then revoke the minted entity's read access. With no surrounding transaction, if the revoke throws the row is held-but-still-readable — the wrong direction for a leak response. Revoke first (close the readable window), or wrap the pair in a transaction. Also worth surfacing a drift-pass failure into the run's success flag.

packages/CodeGenLib/src/Database/providers/postgresql/PostgreSQLCodeGenProvider.ts
```diff
@@ -429,6 +430,70 @@ END $vw_regen$;
...
+     * `GENERATED ALWAYS AS IDENTITY` clause (see {@link getMaterializedSurrogateColumnType}).
+     */
+    override generateMaterializedTableSQL(schema: string, tableName: string, columns: MaterializedColumnSpec[]): string {
+        // MJ entity-field metadata stores CANONICAL (SQL Server-flavored) type names — e.g. 'uniqueidentifier',
+        // 'nvarchar(50)' — regardless of the physical DB dialect. Every other PG DDL path maps these via
+        // mapSQLType (see spCreate/spUpdate generation); the materialized-table DDL must do the same, or a
+        // base-view/query materialization emits `... uniqueidentifier ...` and PG fails the CREATE TABLE with
+        // `type "uniqueidentifier" does not exist`, aborting the whole codegen run.
+        const colLines = columns.map(
+            (c) => `    ${pgDialect.QuoteIdentifier(c.Name)} ${this.mapSQLType(c.SQLType)} ${c.Nullable ? 'NULL' : 'NOT NULL'}`,
```
rkihm-BC, line 452: **[Blocker — the one remaining item] PG mint DDL still doesn't escape identifier delimiters.** This routes column names through `pgDialect.QuoteIdentifier(c.Name)`, but that method is a bare `` `"${name}"` `` (`postgresqlDialect.ts:170-172`) with no `"`-doubling — and `QuoteSchema` (`:174-176`) leaves the schema part entirely unquoted. Meanwhile the PG **refresh** path *does* escape (`MaterializationRefresher.ts:1079`, `escId = n.replace(/"/g, '""')`). So the exact mint↔refresh asymmetry I flagged on Aug 3 is fixed on SQL Server (`SQLServerCodeGenProvider.ts:167`) but still open on PostgreSQL: a column name containing `"` — which for **external** queries originates from untrusted remote-schema introspection — breaks out of quoting during a CodeGen-privileged `CREATE TABLE`.

Because `QuoteIdentifier`/`QuoteSchema` back the *whole* PG CodeGen provider, the clean fix is in `postgresqlDialect.ts` (double `"` in `QuoteIdentifier`; quote the schema in `QuoteSchema`) — that hardens all PG codegen, not just materialization. Low reach for internal entities; matters for the EDS/external-query path this feature explicitly targets. Please close this (or, if you'd rather scope it as a documented v1 limitation like the other honest PG gaps, let's discuss).

rkihm-BC (changes requested):
## Re-review — Query & Entity Materialization (Phase 1)

Thanks for the thorough turnaround. I re-verified every item from my Aug 3 review against the current head (63a7302) — actual code, not commit messages — and scanned the new commits for regressions.

**Resolved and verified (6 of 7):**
- 🔴 **Migration re-stamp** — closed. All 12 migrations moved to `v6/` + `migrations-pg/v6/`, stamped `V20260807xxxx` (strictly after next's `V202608061930`), relabeled `v6.2.x`, banners consistent, changeset `minor`.
- 🟡 **Post-mint permission staleness** — `detectMaterializationDrift` now runs every CodeGen pass and `reconcileMaterializedQueryEntityReadGrants` revokes grants where a role can no longer read *all* sources (`manage-metadata.ts:6172-6199`). Closes the plain-`CanRead`-revocation gap, not just the RLS-gain case.
- 🟡 **Incremental admits a VIEW as source** — `sourceIsBaseTable()` gates on `TABLE_TYPE='BASE TABLE'` and fails safe before trusting the watermark (`MaterializationRefresher.ts:624,653`).
- 🔵 **MERGE HOLDLOCK** — added (`:1415`).
- 🔵 **DriftHold-before-revoke** — reversed; revoke now closes the readable window first (`manage-metadata.ts:1808-1814`).
- 🔵 **`RunViewParams.Equals` ignores DataSource** — fixed via the canonical helper (`runView.ts:446`).

I also reviewed the new commits (cross-schema fail-closed `e72b66fc3`, under-linking/CTE `855157282`, cache-status probe gating `6adfa87af`, DriftHold-honor `899562a42`, concurrency/watermark/PG-atomicity `efec637d2`) — all correct, no regressions.

**One residual to close before merge — flagged inline.** The SQL Server mint now escapes `]`, but the PostgreSQL mint still doesn't double embedded `"`, so the mint↔refresh escaping asymmetry I raised is only half-fixed on PG. Details inline.

Minor (non-blocking): the "batched" server cache-status probe now issues a per-item `resolveEffectiveBaseView` (a `RunView` per base-view-materialized entity) inside the batch loop — correct, but partially defeats the batching. Worth a follow-up, not a blocker.

Happy to pair on the PG escaping fix if useful — if you read it differently, let's talk it through rather than have you take it on faith.

rkihm-BC (approved):
## Re-review — the requested change is resolved ✅

Thanks for the fast, clean turnaround. Verified against head `ab50b5b`.

### The PG escaping blocker — fixed, and fixed the *right* way
You took it to the dialect level (`370660ec3` → `e2f8fb216` → `9cdcb37b9`) rather than patching materialization locally, which hardens **all** PostgreSQL codegen:
- `QuoteIdentifier` now doubles embedded `"` (`postgresqlDialect.ts:170-174`).
- `QuoteSchema` quotes **both** parts (schema was previously interpolated bare) (`:177-181`) — also fixes case-folding for non-lowercase schemas, behavior-neutral for `__mj`.
- `QuoteColumnAlias` escapes too (`:193-195`); the provider-local `quoteMintIdentifier` was reverted so the mint path uses the hardened dialect methods; SQL Server mint escaping retained (`SQLServerCodeGenProvider.ts:167`).
- Directly tested: `postgresqlDialect.test.ts` asserts `ev"il` → `"ev""il"` and `QuoteSchema('sch"ema','obj"ect')` → both parts doubled. The mint↔refresh asymmetry I flagged on both prior passes is now closed on PG.

### New drift-hold change (`5cecd034d`) — good, security-positive
Revoking read on **any** hold of a query mat is the correct fail-closed call: a query VE's reads aren't status-gated, so a benign shape-drift hold that later gains source RLS (or loses a role's source read) would otherwise keep serving the unscoped snapshot. Base-view exemption is right (their reads fall back to the RLS-enforcing live base view via `resolveEffectiveBaseView`). Two more real improvements ride along: per-row `try/catch` isolation so one failing row no longer aborts reconciliation for the rest (that was a latent fail-open), with the pass now reporting *partial* reconciliation via `success`; and the external RLS base-view guard now **empties** the mirror rather than just DriftHold-ing it (DriftHold alone left already-populated RLS-refused rows queryable over the wrapper view). All five branches covered in `manage-metadata-drift-hold-revoke.test.ts`.

### One remaining pre-merge step — the migration re-stamp treadmill (mechanical, not code)
`next` has since advanced to `V202608072030` (task-graph work), which is **later** than this branch's materialization migrations `V202608070001–0006`. On a next-tracking DB those would come in out-of-order and Flyway strict validation would reject them. When you bring `next` current for the merge, re-stamp the 6 SS + 6 PG files (and the #3365 ReadFilterSpec) to **after** next's newest at that moment. Purely mechanical — flagging so it isn't missed, not a code concern.

### Deferred follow-ups you disclosed — I concur they're safe to defer
Concurrent-refresh watermark ordering (freshness, bounded by forced rebuild, serial scheduler); PG `DROP … CASCADE` transitive-dependent drop (only bites hand-built objects on a wrapper view); and the P1 under-linking guard failing **open** on unparseable SQL. The last is the only fail-open — I agree it's narrow (the primary linked-source RLS check still runs every pass) and that failing closed there would false-refuse legitimate hard-to-parse queries, so it's a defensible judgment call. Worth a tracked issue so it's a conscious standing decision.

Approving on the code. Standing recommendation before the stack lands: the migration re-stamp above, plus a fresh SQL Server + PostgreSQL E2E on the reassembled code given how much has layered on. Nice work on this one.

PR #3336, opened 2026-07-28
fix(explorer): don't let one resource's load error hang the whole Explorer loading screen

MarceloT-BC:
# fix(explorer): don't let one resource's load error hang the whole Explorer loading screen

## The bug

The Explorer shell's loading screen blocks on the first resource's `NotifyLoadComplete()`
signal. `BaseDashboard` called it **after** `await this.loadData()` with no `try/finally`:

```typescript
async ngOnInit() {
  super.ngOnInit();
  this.initDashboard();
  await this.loadData();        // if this throws…
  this.NotifyLoadComplete();    // …this never runs → shell hangs forever
}
```

So if `initDashboard()` or `loadData()` threw — e.g. an in-flight query rejecting while MJAPI
is restarting, or missing data — the signal never fired and the **entire** Explorer hung on the
loading screen. This reproduced on any full reload of a deep resource URL (a dev-server
live-reload after an edit, or a browser refresh) while the API was momentarily down. A fresh
magic link worked because it lands at the root, which has no resource-load dependency.

## The fix (two parts)

**1. `BaseDashboard` — guarantee the signal.** `ngOnInit` and `Refresh` run their load through a
`runGuardedLoad()` helper that wraps the work in `try/catch/finally`. On error it logs via
`LogError` and emits the existing `Error` output (so the dashboard/container can render its own
error state); `NotifyLoadComplete()` runs in `finally`, so the loading screen always clears. Both
`initDashboard()` **and** `loadData()` are inside the guard — `initDashboard()` runs first and can
throw too.

**2. `BaseResourceComponent` — fail-open watchdog.** The existing 15s load-complete watchdog now
**forces** `NotifyLoadComplete()` on timeout (still logging a warning that names the culprit)
instead of only warning. This covers every `BaseResourceComponent` subclass — including ones whose
own `ngOnInit` bypasses `BaseDashboard`'s guarded lifecycle, or whose load genuinely hangs — so no
single resource can brick the shell.

## Why a helper (`runGuardedLoad`) instead of inlining?

It's applied at **both** load entry points — `ngOnInit` and `Refresh` — and the extracted part
(the `catch`: normalize → `LogError` → `Error.emit`; the `finally`: `NotifyLoadComplete`) is
identical in both; only the wrapped work differs. Since the whole purpose of this change is that
the loading-screen release is *guaranteed in every path*, sharing one helper means the two paths
can't drift out of sync later (a future tweak to the error handling applied to only one copy would
silently reintroduce the hang). It's a small call — 2 sites — so if reviewers prefer the two blocks
inlined for readability, that's an easy change with no behavioral difference.

## Validation

- `turbo build` green (`ng-shared` + deps).
- New unit spec (6/6) against the **real** base classes: happy path · `loadData` throw ·
  `initDashboard` throw · `Refresh` throw · watchdog forces release · watchdog no-ops when
  signalled in time.
- **Not** browser-verified — the deep-URL-reload-while-API-down repro is a full-stack scenario;
  the guarantee is proven at the control-flow level by the unit spec.
- Pre-existing `shared.test.ts` failures (12/23, `SharedService` singleton) are unrelated —
  confirmed present on `origin/next` with these edits stashed.

Fixes the "Explorer hangs on the loading screen after an agent edits code / rebuilds the API"
report.


MattC-BC:
@MarceloT-BC - Took this for a spin today and I'm good with merging after Soham is done with the release. Broke the API on purpose a few ways and the loading screen always let go instead of hanging, which is exactly what we wanted.

Two things came out of the testing:

1. I managed to get the app stuck on the loading screen through a different path that this PR doesn't touch. Filed it with all the details in Issue [#3440](https://github.com/MemberJunction/MJ/issues/3340).

2. When a page fails to load now, it quietly looks like an empty page instead of a failed one. I got a "No content sources yet" with a green checkmark while the API was completely down. I'll be speccing a proper error state with a retry button as a follow-up.

Nice work on this one!

MattC-BC (approved):

MarceloT-BC:
@MattC-BC  Glad this works. We need to wait for build, but are we good to merge after build here?

MattC-BC:
> @MattC-BC Glad this works. We need to wait for build, but are we good to merge after build here?

Yep!

MarceloT-BC:
We can merge this. Build is pushed to tomorrow at 10am. are you good to merge?

AN-BC (commented):
@MarceloT-BC please review this with @rkihm-BC and @MattC-BC next week, I'm aware of this issue and thought we plugged most of it, but if this further strengthens it, I have no objection but Matt and Robert need to be approval on UX stuff. Thanks.

MarceloT-BC:
@rkihm-BC Pinging for review here.

rkihm-BC (commented):
## Code review — don't let one resource's load error hang the Explorer loading screen

**Verdict: a correct, well-tested, tightly-scoped fix for a real infinite-hang. Safe to merge on correctness. The one thing worth a UX call before final approval: the "container renders its own error state" benefit isn't actually wired in the primary dashboard host — so on a load failure the spinner clears to a *blank* dashboard with a console-only error, not a visible one. Still strictly better than the hang it replaces, but not the full UX the description implies.**

Checked out the branch and ran the new spec: **6/6 green**.

### What's good

- **Correct diagnosis and fix.** `ngOnInit` isn't awaited by Angular, so a throw in `initDashboard()`/`loadData()` became an unhandled rejection and `NotifyLoadComplete()` never fired → the shell's loading screen (which blocks on that signal) hung forever. Wrapping both phases in `try/catch/finally` with the signal in `finally` is exactly right.
- **The two-layer defense is well-shaped.** The `BaseDashboard` guard fixes the common case immediately; the `BaseResourceComponent` fail-open watchdog is a genuine net for subclasses whose own `ngOnInit` bypasses `BaseDashboard`'s lifecycle or that truly hang. One dashboard can no longer brick the whole shell.
- **`runGuardedLoad` is justified.** Two call sites (`ngOnInit`, `Refresh`) with identical catch/finally; extracting it prevents the paths drifting and silently reintroducing the hang. Wrapping `initDashboard()` (runs first, can throw) is correct.
- **Idempotency/lifecycle reasoning holds — verified.** `NotifyLoadComplete()` clears the watchdog and is already called repeatedly on the normal path (every `Refresh`), so the finally + watchdog + normal signal can't collide destructively. Watchdog is started in `ngOnInit`, cleared in both `ngOnDestroy` and `NotifyLoadComplete` — no fire-after-destroy, no spurious fire after a normal load. The "signals in time → watchdog no-ops" test confirms it.
- **Tests are real, not tautological.** They assert on `LoadCompleteEvent` (precisely what the shell waits on) plus the error side-channel, instantiate the **real** base classes rather than mocking the logic under test, and cover each failure mode (loadData throw, initDashboard throw, Refresh throw, watchdog force, watchdog no-op). Fake timers used correctly for the 15s window.
- **Cleanly scoped and conventional:** patch changeset, purely defensive, no API changes, `err instanceof Error` normalization, `LogError` per MJ conventions.

### Concerning

1. **The "container can render its own error state" claim isn't wired for the primary host.** `dashboard-resource.component.ts` → `loadCodeBasedDashboard` subscribes to `LoadCompleteEvent`, `OpenEntityRecord`, and `UserStateChanged` on the dynamically-created dashboard — but **not** `instance.Error`. So when a code-based dashboard's `loadData()` throws: `runGuardedLoad` logs to console, emits `Error` **to no listener**, then fires `NotifyLoadComplete()` → the host clears the spinner → the user sees the empty/half-rendered dashboard. The host already has an error card (`setError()`), but it's only triggered by errors the *host* catches, never by the dashboard's internal throw. **High-value one-liner follow-up:** `instance.Error.subscribe(err => this.setError('The dashboard could not be loaded.', err))` in the host. Ship the hang-fix now and treat the visible-error state as a fast follow, or fold it in here — it's a few lines.
   - *Corollary:* the existing `this.Error.emit(...)` calls in the Testing/Communication/Credentials dashboards are also unheard by this host today — the `Error` output is largely a vestigial channel. The PR leans on it as its error-surfacing story, but nothing consumes it.

2. **Fail-open terminal UX for a genuine hang is 15s → blank.** For a resource that truly hangs, the user waits the full 15 seconds, then the shell releases to a half-state with only a console warning — neither the wait nor the failure is communicated. Better than infinite, but worth a follow-up: a user-visible "this resource failed to load" state on fail-open, and a second look at whether 15s is the right perceived-responsiveness threshold.

3. **`Refresh()` no longer rejects on load failure** — it now swallows → logs → emits `Error` → resolves. No current caller awaits/catches it (the host calls `instance.Refresh()` fire-and-forget), so this is safe today, but it's a quiet contract change for any future caller that wants to know a refresh failed.

### Minor / verified

- **Pre-existing `shared.test.ts` failures confirmed unrelated** — `TitleService`/`EventCodes` failing on an Angular JIT-compiler bootstrap issue, in a file this PR doesn't touch; the new spec passes 6/6 independently. One correction to the PR text: it's **14/23** now, not 12/23 — drift on `next`, still not caused here.
- A genuinely **slow-but-healthy** load exceeding 15s trips the reworded "FAILING OPEN … most likely errored or hung" warning and releases early, then re-notifies harmlessly when it finishes. Cosmetic, but the wording slightly over-accuses a merely-slow load.
- **Not browser-verified** (the PR says so). The control-flow guarantee is well-covered by unit tests, but the "deep-URL reload while API down" repro isn't exercised end-to-end. For a shell-wide loading path, one manual repro before release would be reassuring.

### What's missing

Only the host-side `instance.Error` subscription — the single change that would make the "renders its own error state" claim actually true — and, optionally, a user-visible fail-open state. Neither blocks the hang fix; both are what separate "the shell no longer hangs" from "the user understands what happened."

### Bottom line

The core fix is correct, minimal, and well-tested, and it eliminates a genuinely bad hang with a sound two-layer guarantee. The remaining gap is UX-completeness, not correctness: the failure path currently degrades to a silent blank rather than a shown error, because the primary host never listens to the `Error` output the fix emits. Recommend approving the hang-fix and filing the host `Error` subscription (+ a fail-open user message) as an immediate follow-up — or folding the host subscription into this PR, since it's only a few lines in `dashboard-resource.component.ts`.

*Reviewed with Claude Code.*

MarceloT-BC:
Thanks @rkihm-BC — genuinely useful review. All three "Concerning" items are addressed; here's what changed, what I left, and one correction to my own PR description.

## 1. The host never listened to `Error` — folded into this PR

You were right that this made the PR's error-surfacing story only half true. `DashboardResource` now subscribes to `instance.Error` and routes it into the existing `setError()` card, so a failed load clears the spinner to a *visible* "Unable to Load Dashboard" instead of a blank pane.

Two things worth flagging from wiring it:

- **Ordering is load-bearing.** The subscription has to go immediately after `createComponent`, *before* the `await this.loadDashboardUserState(...)` — Angular can run the dashboard's own `ngOnInit` (and therefore its guarded load, and therefore the `Error` emit) while the host sits on that await. Subscribing next to the existing `OpenEntityRecord`/`UserStateChanged` wiring further down would have missed exactly the failure we care about. The new spec pins this: it holds the user-state promise open and emits `Error` inside that window.
- **Two other hosts had the same gap** — the Data Explorer path in the same file, and `BaseAdminContainerComponent` (which embeds code dashboards in the four Admin shells, surfaced through its existing `LoadError`). That closes your "vestigial channel" corollary: `Error` is now consumed everywhere a `BaseDashboard` is hosted.

**One I deliberately left alone:** the config-based path's `instance.error` (the `DashboardViewerComponent` output) is still console-only. That emitter is overloaded — it fires for "Failed to load dashboard" *and* "Unknown panel type" *and* "Failed to save dashboard". Routing it to the full-page card would blank a working dashboard on a failed save. It needs the fatal/non-fatal split @MattC-BC's error-state spec will bring.

**A risk this subscription introduces, which is worth your eyes:** `Error` used to have no listeners, and now it blanks the pane into a full-page error card. For a *load* failure that's the intent. But `Refresh()` also emits `Error`, so a dashboard that calls its own `Refresh()` post-mount (a refresh button) and fails would replace an already-rendered dashboard with the error card — and `clearError()` only runs when the resource's record changes, so it would stick. I checked the current callers: only a handful call `this.Refresh()` after mount, and the existing `Error.emit` sites (Testing/Communication/Credentials) are all `initDashboard` failures, so nothing regresses today. But it's a live edge for future dashboards, and the clean fix is scoping the error card to the initial load — happy to add that here if you'd rather not leave it to the follow-up.

## 2. Fail-open terminal UX (15s → blank)

Agreed, and left for the follow-up. The watchdog lives in `BaseResourceComponent`, which has no error surface of its own — a user-visible fail-open state means introducing one, which is exactly what Matt is speccing. I did reword the warning; it no longer over-accuses a merely-slow load and notes that a slow-but-healthy load re-signals harmlessly. I left 15s alone — worth revisiting alongside the retry UI rather than in isolation.

## 3. `Refresh()` no longer rejecting

Kept, documented explicitly. Making it reject would produce an unhandled rejection at the fire-and-forget call sites — the exact bug class this PR exists to fix — so the JSDoc now states that the promise resolves on failure and points future callers at the `Error` output.

## The manual repro you asked for — and a correction to my PR description

I ran it live (real MJAPI + Explorer, headless Chrome). Results, including the part that didn't go my way:

- ✅ **The loading screen released in every scenario I could produce.** No hang.
- ✅ **The error card renders** on a failed dashboard load — the user now sees "Unable to Load Dashboard" with technical details where Matt got a silent empty page.
- ❌ **I could not reproduce, in a browser, a dashboard whose own `loadData()` throws** — the precise path the new `Error` subscription serves. Every code dashboard I could drive either loads from cached metadata (ERD, Data Explorer, Component Studio all render fine with GraphQL dead) or fails *earlier* at class registration, which the host's pre-existing `catch` already handled. So that subscription is proven by unit test, not end-to-end. Calling that out rather than implying more.
- 🆕 **A cold reload of a deep URL with the API fully down never reaches resource mounting at all** — the shell hangs in bootstrap at *"Getting things ready… Taking longer than expected"*. **This PR cannot help there**, which means my PR description's "reproduced on any full reload of a deep resource URL while the API was momentarily down" is overstated: the real trigger is the API dying *while the shell is already up*. The cold-reload variant looks like exactly the path @MattC-BC filed as #3340.

## On the `shared.test.ts` count — and something worse underneath it

On this branch head it's **12/23**, not 14; `origin/next` is 117 commits ahead of this branch now, so I think you measured against a newer next.

More importantly, chasing it turned up that **none of those tests were running in CI at all** — `@memberjunction/ng-shared`'s `package.json` still had `"test": "echo \"No tests configured yet\""` despite having a `vitest.config.ts`, and CI runs `turbo run test`. So this PR's own 6/6 spec was invisible to CI. Fixed here: 9 failures were `document is not defined` (the file needs jsdom; the package default environment is node), 2 were a missing `GetGlobalObjectStore` export on the `@memberjunction/global` mock, and the 12th was a genuinely wrong assertion — the test expected `mapResourceTypeNameToRouteSegment('MJ: User Views')` → `'view'`, but resource-type *row* names are unprefixed (`N'User Views'` in the baseline seed; the `MJ: ` prefix rule applies to entity names), and the same file's reverse-mapping test already expected `'user views'`. Corrected, not relaxed. `ng-shared` is now 71/71 and the script is `vitest run`, so the package is genuinely covered.

## Tests

- New: host `Error` subscription spec (`dashboard-resource.component.test.ts`) and Admin-container spec (`base-admin-container-error.test.ts`). Both verified as real oracles — each fails when its subscription is removed.
- `ng-explorer-core` 114/114 · `ng-dashboards` 1692/1692 · `ng-shared` 71/71 · `turbo build` 267/267 green.

Changeset now covers `ng-explorer-core` and `ng-dashboards` alongside `ng-shared`.

MarceloT-BC:
Pushed — the review fixes above are now on the branch.

Heads-up on one wrinkle while catching up to `next`: `shared.test.ts` and `ng-shared`'s `package.json` conflicted, because the same three repairs had **already been made independently on this branch** by the ui-layers work — the `ng-shared-generic` mock, the `RecordNavigationAdapter` mock, and the `'MJ: User Views'` → `'User Views'` assertion fix — plus the same `"test": "vitest run"` enablement. Convergent fixes, so I took theirs on both (their `RecordNavigationAdapter` mock is a superset — `{ Register, Instance: { Register } }` — and their `fakeDocument` for `TitleService` is lighter than the jsdom switch I'd made). `package.json` kept both sides: `test` + their new `test:watch`.

Revalidated on the merged tree (which now includes #3444's record-open / mobile UX):

- `turbo build --force` on explorer-core + shared + dashboards — **105/105**, nothing cached
- `ng-shared` **81/81** · `ng-explorer-core` **151/151** · `ng-dashboards` **1694/1694**
- `npm run check:standards` (ui-layers) — **✓ pass**

PR diff is 8 files, +461/−12 — no merge bloat.

@rkihm-BC the open question from my previous comment still stands if you have a view: whether to scope the new error card to the initial load, so a failed post-mount `Refresh()` can't replace an already-rendered dashboard with a sticky error state. Happy to fold it in here or leave it to the follow-up.

packages/Angular/Explorer/explorer-core/src/lib/resource-wrappers/dashboard-resource.component.ts
```diff
@@ -728,6 +728,13 @@ export class DashboardResource extends BaseResourceComponent {
                 this.NotifyLoadComplete();
             };
 
+            // Same as the code-based path: surface the dashboard's own load failures instead of
+            // clearing the spinner to a silent blank page (see loadCodeBasedDashboard).
+            instance.Error.subscribe((err: Error) => {
+                this.setError('The Data Explorer could not be loaded.', err);
+                this.cdr.markForCheck();
+            });
```
rkihm-BC:
```suggestion
            // Surface the dashboard's own load failures — but only for the INITIAL load. Wrap the
            // completion hook to learn when the first load has settled, so a later Refresh() that
            // fails does NOT replace an already-rendered Data Explorer with a sticky error card.
            let initialLoadSettled = false;
            const onComplete = instance.LoadCompleteEvent;
            instance.LoadCompleteEvent = () => { initialLoadSettled = true; onComplete?.(); };
            instance.Error.subscribe((err: Error) => {
                if (initialLoadSettled) return; // post-mount refresh failure — already logged by BaseDashboard.runGuardedLoad
                this.setError('The Data Explorer could not be loaded.', err);
                this.cdr.markForCheck();
            });
```

Scopes the error card to the initial load so a failed post-mount `Refresh()` doesn't blank a rendered Data Explorer.

packages/Angular/Explorer/explorer-core/src/lib/resource-wrappers/dashboard-resource.component.ts
```diff
@@ -774,6 +781,17 @@ export class DashboardResource extends BaseResourceComponent {
...
+            // Surface the dashboard's own load failures in the host's error card. BaseDashboard
+            // guarantees the loading screen is released even when initDashboard()/loadData() throws
+            // (it emits Error, then NotifyLoadComplete in a finally) — without this subscription the
+            // spinner would clear to a blank/half-rendered dashboard with a console-only error.
+            // Subscribed immediately after creation, BEFORE the first await below, so an Error
+            // emitted during the instance's own ngOnInit isn't missed.
+            instance.Error.subscribe((err: Error) => {
+                this.setError(`The dashboard "${dashboard.Name}" could not be loaded.`, err);
+                this.cdr.markForCheck();
+            });
```
rkihm-BC:
```suggestion
            // Surface the dashboard's own load failures in the host's error card — but only for the
            // INITIAL load. BaseDashboard guarantees the loading screen is released even when
            // initDashboard()/loadData() throws (it emits Error, then NotifyLoadComplete in a
            // finally). Wrap the completion hook to learn when the first load has settled, so a later
            // Refresh() failure keeps the rendered dashboard instead of blanking it to a sticky error
            // card. Wired BEFORE the first await below, so an Error during the instance's own
            // ngOnInit isn't missed.
            let initialLoadSettled = false;
            const onComplete = instance.LoadCompleteEvent;
            instance.LoadCompleteEvent = () => { initialLoadSettled = true; onComplete?.(); };
            instance.Error.subscribe((err: Error) => {
                if (initialLoadSettled) return; // post-mount refresh failure — already logged by BaseDashboard.runGuardedLoad
                this.setError(`The dashboard "${dashboard.Name}" could not be loaded.`, err);
                this.cdr.markForCheck();
            });
```

Same initial-load scoping for the code-based dashboard path.

rkihm-BC (commented):
## Re-review — the three concerns are addressed; one residual edge

The two new commits resolve all three items from my prior review, each backed by a real oracle test:

- **#1 (host never listened to `instance.Error` → blank pane): fixed.** `instance.Error` is now surfaced in all three hosting paths — code-based dashboards and the Data Explorer path in `dashboard-resource.component.ts` (→ `setError()` card + `cdr.markForCheck()`), and embedded admin dashboards in `base-admin-container.component.ts` (→ `LoadError`). The subscription is wired **before** the `await loadDashboardUserState(...)`, so an `Error` emitted during the instance's own `ngOnInit` isn't missed — and the new spec pins exactly that by holding the user-state promise open. The instance-`Error` route and the host's own `try/catch` are disjoint, so there's no double error-card.
- **#2 (fail-open 15s → blank): addressed as agreed** — watchdog forces `NotifyLoadComplete()` and the warning no longer over-accuses a slow-but-healthy load; the user-visible fail-open message is correctly deferred to the error-state spec.
- **#3 (`Refresh()` silently stopped rejecting): documented** — explicit JSDoc that it resolves on failure and directs callers to subscribe to `Error`. Right call.

### One residual edge — worth folding in here

Wiring `Error` to a full-page card introduces the trap the PR description itself flagged: **`Refresh()` also emits `Error`, but the host's `clearError()` only runs on record-change / full reload — never on an instance-initiated `Refresh()`.** So a dashboard whose refresh button fails **post-mount** would replace an already-rendered view with the error card, and it would stick until the user navigates records. Low-likelihood today (current `Error.emit` sites are all initial-load failures), but a latent trap for any future dashboard.

The fix is to scope the card to the **initial** load. It's self-contained within each existing `Error` block — wrap the completion hook to learn when the first load has settled, then ignore a later `Error`. Drafted as two click-to-apply suggestions below (lint-clean; `onComplete?.()` preserves the existing `LoadCompleteEvent`; on an initial-load failure the `Error` still precedes the `finally`'s completion signal, so the card shows exactly when it should).

Optional follow-on: a test asserting a post-mount `Refresh()` failure does **not** blank an already-rendered dashboard — happy to draft it.

Everything else is approvable on correctness. Nice turnaround.

packages/Angular/Explorer/explorer-core/src/lib/resource-wrappers/dashboard-resource.component.test.ts
```diff
@@ -147,3 +153,73 @@ describe('DashboardResource config dashboard loading', () => {
...
+    dashboardInstance.Error.emit(new Error('loadData blew up'));
+
+    expect(resource.errorMessage).toContain('Broken Dashboard');
+    expect(resource.errorDetails).toContain('loadData blew up');
+    expect(markForCheck).toHaveBeenCalled();
+
+    releaseUserState();
+    await loadPromise;
+  });
+});
```
rkihm-BC, line 288:
```suggestion
  });

  /**
   * Once the initial load has SETTLED (the dashboard signalled LoadCompleteEvent), a later Error —
   * e.g. a refresh button inside the dashboard whose Refresh() fails — must NOT replace the
   * already-rendered dashboard with a sticky error card. The host scopes the error card to the
   * initial load; this pins that so a future dashboard with a post-mount refresh can't regress it.
   * (Fails without the initialLoadSettled scoping in loadCodeBasedDashboard/loadDataExplorer.)
   */
  it('does NOT blank an already-rendered dashboard when a post-mount Refresh() emits Error', async () => {
    const { DashboardResource } = await import('./dashboard-resource.component');

    const dashboardInstance = {
      Error: new MockEventEmitter<Error>(),
      OpenEntityRecord: new MockEventEmitter(),
      UserStateChanged: new MockEventEmitter(),
      LoadCompleteEvent: null as (() => void) | null,
      Config: null as unknown,
      Refresh: vi.fn(),
    };

    const viewContainer = {
      createComponent: vi.fn(() => ({
        instance: dashboardInstance,
        hostView: { rootNodes: [{ style: {} }] },
      })),
    };

    const markForCheck = vi.fn();
    const resource = new DashboardResource(
      viewContainer as any,
      { detectChanges: vi.fn(), markForCheck } as any,
    ) as any;

    resource.containerElement = {
      nativeElement: { innerHTML: 'old', appendChild: vi.fn() },
    };
    resource.navigationService = { OpenEntityRecord: vi.fn() };
    resource.loadDashboardUserState = vi.fn(async () => ({ UserState: null }));

    // Drive the initial load to completion, then let the dashboard signal it is ready.
    await resource.loadCodeBasedDashboard({
      ID: 'dash-1',
      Name: 'Rendered Dashboard',
      DriverClass: 'RenderedDashboard',
    });
    expect(dashboardInstance.LoadCompleteEvent).toBeTypeOf('function');
    dashboardInstance.LoadCompleteEvent!(); // initial load has now SETTLED
    expect(resource.errorMessage).toBeNull();

    markForCheck.mockClear();

    // A post-mount Refresh() inside the dashboard fails and emits Error.
    dashboardInstance.Error.emit(new Error('refresh blew up'));

    // The rendered dashboard is preserved — no error card, no repaint into one.
    expect(resource.errorMessage).toBeNull();
    expect(resource.errorDetails).toBeNull();
    expect(markForCheck).not.toHaveBeenCalled();
  });
});
```

Regression guard for the two scoping suggestions — asserts a failed post-mount `Refresh()` doesn't blank a rendered dashboard. It **fails on purpose** without the `initialLoadSettled` scoping, so land it together with the two suggestions.

rkihm-BC (commented):
Regression test to accompany the two scoping suggestions above. It settles the initial load (fires `LoadCompleteEvent`), then emits a post-mount `Error` and asserts the rendered dashboard is preserved (no error card, no repaint). It's a genuine oracle: **it fails without the `initialLoadSettled` scoping**, so the three should land as one change. The existing "surfaces a dashboard Error … while awaiting user state" test still passes with the scoping, since that Error is emitted before the load settles.

MarceloT-BC:
Thanks @rkihm-BC — all three of your suggestions are in, as one commit so the scoping and its guard land together.

**Both scoping suggestions applied as written.** Your reasoning on the ordering checks out: `runGuardedLoad` emits `Error` *before* the `finally`'s `NotifyLoadComplete`, so a genuine initial-load failure still reaches the card, and the wrapper forwards to the original hook so the shell's release signal is unaffected.

**Extended it to a third host you didn't flag.** `BaseAdminContainerComponent` has a worse version of the same trap: its sections are **cached and kept alive** across switches, and `LoadError` only resets on the next `selectSection` — so a post-mount `Refresh()` failure could pin a stale banner *while a different section is on screen*. Same pattern applied there, plus a mirror regression test.

**Took your test.** Yours is better than the one I'd written — the `markForCheck` no-repaint assertion is a sharper check than what I had. I kept one thing from mine: an assertion that the wrapped `LoadCompleteEvent` still forwards to `NotifyLoadComplete`, since wrapping that hook is the one way this change could silently break the loading-screen release it exists to protect. Confirmed as a real oracle — fails with the `initialLoadSettled` guard removed.

### Two edges the scoping doesn't cover

Neither is a regression (both land on today's behaviour — blank pane + console error), but they bound what the fix buys, so you should know:

1. **The watchdog can pre-empt the flag.** A load that runs past 15s and *then* throws: the watchdog has already forced `NotifyLoadComplete()`, so `initialLoadSettled` is true and the genuine initial-load error is suppressed.
2. **Two loads race at mount.** The host calls `instance.Refresh()` while Angular separately runs the instance's own `ngOnInit` guarded load. Whichever settles first flips the flag, so a failure in the other can be swallowed. A precise fix means treating "the host's mount sequence finished" as the boundary — i.e. awaiting `Refresh()` — but that would block section render, which seems a bad trade for the coverage it adds. Left as-is unless you disagree.

**Validation:** `ng-explorer-core` 152/152 · `ng-dashboards` 1695/1695 · `turbo build --force` on both packages 105/105, nothing cached.

rkihm-BC (approved):
## Approve — scoping fix applied cleanly, plus a well-reasoned extension

The new commit (`21b658957c`) applies all three of my re-review suggestions and extends them correctly. Verified against the code, not just the description.

**All three applied:**
- Both `DashboardResource` paths (Data Explorer + code-based) now wrap the completion hook and guard the `Error` subscription with `if (initialLoadSettled) return;`.
- The regression test is applied and **strengthened** — the added `vi.spyOn(NotifyLoadComplete)` + `toHaveBeenCalledTimes(1)` proves the wrapper *forwards* the release signal, closing a gap my version left.

**The extension is a genuinely good catch.** Scoping `BaseAdminContainerComponent` wasn't in my suggestions, and the reasoning is right: the admin container caches sections and keeps them alive across switches, and `LoadError` only resets on the next `selectSection` — so an unscoped post-mount `Error` could pin a stale banner *while a different section is on screen*, a worse variant of the trap I flagged.

**Verified no bug in the extension:** `LoadCompleteEvent` is a plain nullable callback (`_loadCompleteEvent: any = null`, invoked as `if (this._loadCompleteEvent) this._loadCompleteEvent()`). The admin container never sets it before wrapping, so `onComplete` captures `null` and `onComplete?.()` safely no-ops — no crash. In all three sites, an initial-load failure still shows the error surface (the `Error` precedes the `finally`'s completion signal) and a later refresh failure is ignored. Both new tests are real oracles (fail with the guard removed).

**Housekeeping clean:** changeset covers all three affected packages (`ng-shared`, `ng-explorer-core`, `ng-dashboards` — the last is where the admin-container change lives). Reported green: ng-explorer-core 152/152 · ng-dashboards 1695/1695 · turbo build 105/105 (I reviewed the code; didn't re-run).

**One trivial, non-blocking nit:** `const onComplete = instance.LoadCompleteEvent` infers `any` from the getter — inherited weak typing, not newly introduced. Typing the local as `(() => void) | null` would make it airtight, but it's optional.

The full arc — infinite-hang fix → surface the error instead of a blank pane → scope the card to the initial load → guarded by real regression tests across all three hosting surfaces — is complete and internally consistent. The deferred items (fail-open user-visible message, config-path fatal/non-fatal split, 15s threshold) are correctly left to the error-state spec, not regressions here.

Nice work on this one.

*Reviewed with Claude Code.*

MarceloT-BC:
Timing note, @rkihm-BC: the push landed **after** my comment above, not with it. If you opened the diff when that comment arrived, the scoping changes and the new tests weren't on the branch yet — they are now (`21b658957c`). Worth a re-pull before you re-review so you're looking at the code the comment describes.

PR #3337, opened 2026-07-28
fix(cli): pin Open App bootstrap alias numbering to the array position

MS-BC:
Follow-up to #3282, addressing the review points raised on it.

## 1. Alias numbering is now single-sourced (review: "add a test for mixed enabled/disabled entries")

The merged form emits indexed aliases (`__openAppClient0/1/…`) where the bare-import form used none. The concern: if the index basis and the array contents ever disagree, the block emits an array element naming a variable that was never declared — a **TS2304 in the host app's build**, not a CLI test failure, so nothing here would catch it.

They agreed already (both derived from the same `forEach` index), but nothing pinned that. Alias numbering now comes from `refs.length` — the position the alias is about to occupy in `OPEN_APP_CLIENT_MODULES` — so the declaration and the reference share one counter by construction, and numbering stays contiguous when disabled entries are interleaved:

```ts
// [disabled, enabled]
// '@acme/disabled-ng' disabled by `mj app disable`
import * as __openAppClient0 from '@acme/enabled-ng';

export const OPEN_APP_CLIENT_MODULES: unknown[] = [__openAppClient0];
```

Three new tests: `[disabled, enabled]`, several interleaved disabled entries, and the all-disabled case (empty anchor, no dangling reference). The mixed cases assert the exact aliases **and**, structurally, that every alias the array references is declared by an import above it. Verified with teeth — switching the numbering basis back to the entry index fails both new tests and neither of the pre-existing ones.

## 2. Docs corrections, from the same review

- **The `globalThis` anchor is not the `CLASS_REGISTRATIONS` mechanism.** That array is anchored by being spread into `combinedClasses` in the host's `app.module.ts` — a real consumer reference, no `globalThis` in that path. #3282's description called it "the same anchoring strategy", which isn't right. It's now recorded in the emitter's docs as a deliberate variant whose tradeoff is anchoring the block without requiring an `app.module.ts` edit per installed app, at the cost of a module-scope global write.
- **This block is defense-in-depth, not the durable fix.** A package that self-registers classes at module scope while declaring `"sideEffects": false` is making a false declaration. MJ's own such packages already set `"sideEffects": true` (e.g. `@memberjunction/ng-core-entity-forms`), and the Open App scaffold should do the same for generated `-ng` packages — that change lives in the app-authoring repo, not here. Recorded in the docs so the next reader doesn't mistake the workaround for the design.

## Not in this PR

- **`any[]` → `unknown[]`** — already landed in #3282 before merge; the emitted block has no `any` and no eslint escape hatch, and the existing test asserts both.
- **The scaffold `sideEffects: true` change** — `mj app publish` and the `-ng` scaffold are not in this repo. Captured above and in the 6.x issue.
- **@cadam11's 6.x direction** (stop forcing Open App client packages into the main bundle; have each app declare its registrations at publish time and lazy-load it as its own chunk through `ClassFactory.RegisterLazyLoader` + `LazyModuleRegistry`) — filed separately as an issue, per his comment. This PR does not attempt it.

## Testing

`packages/MJCLI`: build clean, **24 test files / 440 tests pass**.

## Live verification — real install, real database, real production build

The failure mode this PR guards against is a break in the **host app's** build, which no CLI test can see. So it was exercised for real, in a dedicated worktree against a live SQL Server database (`MJ_3298`):

1. `mj app install https://github.com/MemberJunction/mj-sample-open-app` — a real install against the live DB (migrations + metadata + config). It left the app disabled because its packages are unpublished, which wrote three genuine `Enabled: false` client entries.
2. Added the real npm package `@mj-biz-apps/common-ng@5.32.0` as the single enabled entry — producing exactly the leading-disabled case this PR is about.
3. Full monorepo build (267/267 tasks), MJExplorer `prebuild`, then production `ng build`.

The emitted block, from the live config:

```ts
// '@mj-sample-app/ng' disabled by `mj app disable`
// '@mj-sample-app/entities' disabled by `mj app disable`
// '@mj-sample-app/actions' disabled by `mj app disable`
import * as __openAppClient0 from '@mj-biz-apps/common-ng';

export const OPEN_APP_CLIENT_MODULES: unknown[] = [__openAppClient0];
```

Numbering is correct — the old basis would have emitted `__openAppClient3` here.

### It also surfaced a defect in the block itself: TS4111

> **Update (2026-08-11):** this fix is no longer part of this PR's diff. The same bracket-access fix landed on `next` independently as 284de20970 — with its own changeset and a stronger test that compiles the emitted output under `noPropertyAccessFromIndexSignature` via the TypeScript compiler API. This branch merged that in and took next's version, so the net diff here carries only the alias-numbering change, tests, and docs. The section below is kept as the discovery record.

`ng build` failed, with **one** error, on the anchor line this PR maintains:

```
TS4111: Property '__mjOpenAppClientModules' comes from an index signature,
so it must be accessed with ['__mjOpenAppClientModules'].
  src/app/generated/class-registrations-manifest.ts:39:40
```

MJExplorer's `tsconfig.json` sets `"noPropertyAccessFromIndexSignature": true`, so `(globalThis as Record<string, unknown>).__mjOpenAppClientModules = …` is a hard compile error. **With any Open App client package installed, MJExplorer would not build at all.** It came in with #3282 and went unnoticed for the same reason the alias-numbering risk did: the block is emitted only when `dynamicPackages.client` is non-empty, and no client app had ever been installed.

Fixed here by emitting bracket access. After the change the same build completes: 0 errors, bundle generated, `postbuild` service-worker cache check passes, and both `__mjOpenAppClientModules` and `@mj-biz-apps/common-ng`'s code are present in the minified output — so the anchor still does its job. A regression test pins bracket access and rejects the dot form.

### Unrelated, found on the way

- Every `bizapps-*` repo is currently uninstallable by this CLI: they declare `__`-prefixed schemas (`__mj_BizAppsCommon`, `__mj_BizAppsTasks`, …) and `ValidateSchemaName` rejects those. The engine has an `allowDoubleUnderscore` escape but `mj app install` exposes no flag for it. Same on `next`.
- `mj app install` writes the app's packages into `MJExplorer`/`MJAPI` `package.json` before verifying they resolve, so an unpublished app wedges `npm install` at the workspace root until the entries are removed by hand.


MS-BC:
Base synced to current `next`; conflicts in `packages/MJCLI/src/commands/codegen/manifest.ts` and `src/__tests__/manifest-client-bootstrap.test.ts` resolved by taking `next`'s side in all three hunks — an equivalent fix landed there first, and next's version is strictly stronger: its test compiles the generated manifest under a simulated consumer tsconfig with `noPropertyAccessFromIndexSignature: true` and asserts zero TS4111 diagnostics, rather than string-matching the output.

Manually re-ran MJCLI locally post-merge: **440 passed**.

PR #3339, opened 2026-07-28
feat(search): make Scoped Search dimensions enforceable — bounds, escaping, entitlement, explainability

cmbrock-BC:
## Summary

`SearchScope.SearchContextConfig` has always documented `dimensions[]`, `inheritanceMode` and
`strictValidation`. **No runtime code read any of it.** Dimensions arrived as a free-form bag that any
caller could author — and on an `AgentInvoked` tool call that caller is the LLM. So a dimension was a
narrowing *convenience*, never a bound, which is why an application that encoded an access boundary as
a dimension was asking the platform for something it could not honour.

This PR implements the declaration, so a dimension can carry an access decision:

- **Anti-spoof.** A dimension marked `restricts: true` is server-derived; a caller-supplied value for
  it is **discarded**, not merged. This is the single change that makes a dimension trustworthy.
- **Grammars + automatic escaping.** `valueType` is enforced, `freetext` is prohibited in a filter
  position, and every interpolated value is escaped for its lane's dialect automatically.
- **Entitlement.** A Skill becomes a search principal in the same sense an Agent already is, and a
  grant can be limited to a time window and to a single tenant.
- **Fail closed, not open.** Six provider call sites silently dropped a filter whose rendered value
  had the wrong shape and then queried **unfiltered** — so breaking the JSON was the cheapest way to
  defeat a metadata filter. A lane whose restriction cannot be applied is now skipped.
- **Explainability.** `ExplainScope()` answers "as this user, with this skill, for this tenant — what
  is in bounds?" without running a search, and the same structure is written to the execution log.

Four security bugs are fixed along the way; two were exploitable on `next` today.

### Bounds that hold
- **`ScopeDimensionResolver`** (new) — enforces the declaration: anti-spoof discard, value grammars,
  `narrowingOf` as a lattice **meet** (a caller may narrow a server bound, never widen it), and
  `strictValidation` with teeth. Server values derive from an **approved `MJ: Queries` row** via
  `expansionQueryID`, so the derivation is reviewable data rather than code.
- **Keys, not lists.** A scope publishes a small key and expands it *in-predicate*; it never accepts a
  materialised array of permitted resources. Measured on the SQL lane: identical latency (84ms vs
  87ms) and identical plan count. The gain is not speed, it is that the bound cannot be authored by
  the caller and stays correct when the mapping changes underneath it.

### The security core
- **`ScopeValueEscaper`** (new) — `sqlLiteral` / `odataLiteral` / `jsonValue` / `filterByLiteral` /
  `pathSegment`, applied by `RenderScopeTemplate` to **every** interpolated value, keyed off the
  lane. The dialect is **derived from the existing `IndexType`** — every provider already dispatches
  on it, so a parallel `MetadataFilterKind` column would be redundant state, wrong on every
  pre-existing row until backfilled, and one more thing to get wrong. An unknown `IndexType` gets
  escaping, never a pass.
- **Deployable with zero template edits.** Because `freetext` is banned in filter positions, a value
  reaching a restricting filter is a uuid/enum/int/date/bool — none of which contain a character any
  escaper changes. Escaping is a no-op on legitimate values and only bites on attacks. Verified
  against a live production-shaped scope: the rendered filter is byte-identical.
- Opt-out is `{{ contextRaw.X }}` rather than `{{ X | raw }}` — after escaping, a filter cannot
  recover the original value, so the choice has to be made at lookup. Equally greppable.

### Failing closed
- **`ScopeFilterGuard`** (new) — distinguishes *absent* (run unfiltered, correct) from *unusable* (a
  restriction was authored but cannot be applied). Wired into all five external-index lanes.
- **`RequiredMetadataKeys`** on both the external-index and entity lanes catches the one failure no
  other guard can see: a filter that rendered **partially**. Scope filters are several optional
  `{% if %}` clauses, so when a dimension is absent — mistyped, omitted, or discarded as spoofed —
  the rest still render. The filter is non-empty, valid, free of template syntax, and now searches
  the whole tenant instead of one channel. Nothing downstream can tell.

### Entitlement
- `AISkill.SearchScopeAccess` + `AISkillSearchScope`, mirroring the agent pattern exactly.
- `StartAt` / `EndAt` / `PrimaryScopeRecordID` on `SearchScopePermission`. The tenant column is
  deliberately **not** called `OrganizationID`: MJ is domain-agnostic and a scope's primary scope may
  be a Company, Client or Practice, so it reuses the existing primary-scope concept.

### Supersession — subtraction, deliberately outside the boundary
"Skill A grants {A,B}; A+B grants C and *removes* B" is real, but it is **not** entitlement. Every
concrete case is relevance or precedence, never "this principal may not see that." Conflating them
would put an ordered rule engine inside the security path. So it is split: `advisory: true` +
ordered `supersededByRules[]` compose by subtraction and fail **soft**, where the bound composes by
intersection and fails **closed**. A wrong rule costs relevance, not access — which is precisely why
it needs none of the grammar or fail-closed machinery above.

### Explainability
- **`ExplainScope()`** runs the full chain and reports the bound **without querying anything**.
  Distinct from `PreviewSearch`, which is a real search capped at 8 results — a sample cannot reveal
  an over-broad bound, because the documents it would newly permit are the ones nobody thought to
  look for. It takes the same untrusted context a real caller sends, so the preview shows the
  anti-spoof discard actually happening, and reports every broken lane in one pass.
- `SearchExecutionLog` gains `AISkillID`, `PrimaryScopeRecordID` and `ScopeDecisionJSON`. That last
  one is the **same type** `ExplainScope` returns, so the preview an admin approves and the record the
  audit keeps cannot quietly disagree.

## Bugs fixed

1. **Result-cache key omitted `ScopeIDs` and `SearchContext`** — unfixed on `next`. A user with access
   to two tenants could be served the *other* tenant's results for up to the 30s TTL. The key now
   includes both, plus `Mode`, fusion weights, overfetch factor and both principals.
2. **Filters silently dropped, then queried unfiltered** at six provider call sites (above).
3. **A restricting template that rendered to nothing** was indistinguishable from one never authored,
   because `RenderScopeTemplate` runs with `throwOnUndefined: false`. Now asserted positively.
   `throwOnUndefined` is deliberately **not** flipped — it would break every template that
   legitimately emits an optional dimension.
4. **`inheritanceMode` was declared and unread** — the exact flaw this PR exists to fix, present in
   the same feature. `restricts: true` now implies `strict`, and choosing the permissive `cascading`
   on a boundary requires an explicit, greppable acknowledgment.

### A pattern worth flagging to reviewers

Three separate defects in `ScopeDimensionResolver` were the *same shape*: a later step overwriting a
security-relevant label an earlier step had set. An "unresolved ⇒ `Absent`" fallback erased a
`DiscardedCaller`; the advisory branch swallowed a declaration check because its job is to swallow
errors; and a `defaultValue` relabelled a discard as `Default`. Each was found by testing, not
review. If you look at one thing closely in this diff, look at whether any path can still overwrite
a provenance label — `Provenance` is what a tool filters on when hunting spoof attempts in
`ScopeDecisionJSON`, so a relabel hides the event from the only query anyone would run.

## Migration

One migration, `V202607281400__v5.50.x__Scoped_Search_Dimensional_Bounds.sql` (5,061 lines).
**Additive only** — a new nullable column, a new table, or an extended property; nothing altered,
retyped, made stricter or dropped. With every new column NULL, an installation that applies it and
changes no data behaves exactly as before. Satisfies the publish-then-no-breaking-changes policy.

**Self-contained**, following `V202607240220__v5.50.x__ContentItem_VectorRecordID_And_ContentItemChunk.sql`:
hand-written DDL, then the embedded CodeGen block carrying views, stored procedures, EntityField rows
and FK indexes for exactly the seven entities affected — AI Skill Search Scopes (new), AI Skills,
Search Scope Permissions, Search Scope External Indexes, Search Scope Entities, Search Execution Logs,
and AI Agents via the new relationship.

Verified by restoring a **pristine** database and applying the whole thing: zero errors, table + both
`__mj_*` timestamps + view + all three stored procedures + trigger + 6 EntityField rows.

Two notes for whoever touches it next:

- **The generated block must be produced against a database that does not already have the objects.**
  A first attempt was generated where the timestamp columns already existed, so it omitted their
  ALTERs, applied cleanly on that database, and then failed on a pristine one with `Msg 207 invalid
  column name` inside `trgUpdateAISkillSearchScope` — 2 of 3 procedures created. That warning is in
  the migration itself.
- **The generated TypeScript entity classes are not in this PR.** Regenerating them requires a
  database at the repo's current migration head; none was available locally (the chain fails on an
  unrelated pre-existing `TestType` duplicate key in `V202607221340`), and generating against a stale
  one produced a diff that *deleted* entities for features already on `next`. The migration provisions
  the EntityField rows, so CodeGen against a current database yields them correctly.

## Tests — and what was NOT run

**410 unit tests across 29 files** (`cd packages/SearchEngine && npm run test`), up from 276 before
this work.

The security properties are tested as properties, not as examples: that a caller-supplied value for a
restricting dimension is discarded; that a spoof attempt is *recorded* rather than normalised away;
that escaping is a no-op on every legitimate value in every lane; that a partially-rendered filter is
caught; that an advisory dimension fails soft where the identical value on a boundary fails closed.

**The deterministic integration tier was not run.** CLAUDE.md is explicit that unit tests alone are
not sufficient, so this is a gap and not an omission I'd like to pass over quietly. The reason is
environmental rather than deliberate: `npm run test:integration` wants a database at the repo's
current migration head, and no local database is. Bringing one current fails on an unrelated,
pre-existing duplicate-key conflict in `V202607221340__v5.49.x__Metadata_Sync.sql`
(`__mj.TestType`, PK violation on `502a3e67-…`) that has nothing to do with this change. *

Beyond unit tests, what *was* verified end to end, against real data rather than mocks:

- The migration applied to a **pristine** restored database — zero errors, and the generated block
  produced the table, both `__mj_*` timestamps, the view, all three stored procedures, the update
  trigger and 6 EntityField rows.
- A **role-derived bound** resolving through `ExplainScope` for two real users with different roles:
  different bounds, `ServerDerived` provenance, a caller-supplied bound discarded, cross-tenant
  denied.
- Escaping verified **byte-identical** on a live production-shaped scope's rendered filter, which is
  the property that makes default-on escaping safe to add.
  
 -Using the fresh DB ran manual testing to confirm scoped search works as expected, and ran tests with the combined BCSaaS update to ensure the entire finished product works together. 

## Compatibility

Every change is additive. A scope whose `SearchContextConfig` is null or declares no dimensions is
returned **untouched** — the sole reason this can land without auditing every existing scope. No
existing template needs an edit.

## Review notes

- **`throwOnUndefined` is intentionally left off.** Strictness comes from the positive assertion.
- **The lane dialect is derived, not stored.** If a deployment ever needs to override the derivation,
  add a column as an *override* rather than as the source of truth.
- **Not included, deliberately:** named-placeholder binding (§5.5's preferred end state) still lies
  ahead — filters remain string interpolation, now escaped by default rather than by hope.


migrations/v5/V202607290900__v5.50.x__Scoped_Search_Dimensional_Bounds.sql
```diff
@@ -0,0 +1,5060 @@
...
+
+EXEC sp_addextendedproperty
+ @name=N'MS_Description',
+ @value=N'JSON array of column or alias names the rendered ExtraFilter MUST reference for this lane to be considered safe, e.g. ["OrganizationID","ContentSourceID"]. Checked against the RENDERED filter at search time: if one is missing the lane is SKIPPED rather than queried, because a partially-rendered filter (an {% if %} clause dropped when its dimension was absent or discarded) is still valid SQL and silently widens the search. Same concept and same engine check as SearchScopeExternalIndex.RequiredMetadataKeys, applied to the SQL lane. NULL = no contract declared, which is the pre-migration behaviour.',
+ @level0type=N'SCHEMA',@level0name=N'${flyway:defaultSchema}',
+ @level1type=N'TABLE', @level1name=N'SearchScopeEntity',
+ @level2type=N'COLUMN',@level2name=N'RequiredMetadataKeys';
+GO
+
+-- ============================================================================
```
AN-BC, line 199: in future add a ton of whitespace between end of manual migration code and the start of codgen comment block, like 50+ whitespace lines, makes it easier to see this break when scanning file fast.

AN-BC (approved):
LGTM. @cmbrock-BC we need to see if we can still get this into `5.50.0` awaiting feedback from @SDesai-BC

migrations/v5/V202607290900__v5.50.x__Scoped_Search_Dimensional_Bounds.sql
cmbrock-BC: Noted

PR #3350, opened 2026-07-29
fix(codegen-lib): gate reverse-relationship emission on type availability; add opt-in includeSchemas

AN-BC:
Replaces #3304, rebuilt cleanly on current `next`. See that PR for the original diagnosis and the live-instance validation — this carries forward its sharper mechanism and adds a fix for a hole in its `includeSchemas` design.

## Why this exists rather than a rebase of #3304

#3304 branched from `next` at `2ded59e1` (Jul 27, 10:57). Later that day, `f96804b9c` ("fix(codegen): omit relationship members whose GraphQL type was withheld", 18:30) landed on `next` and fixed the same TS2304 break via `isRelatedTypeOutOfScope`. #3304 went un-mergeable and the two fixes needed reconciling rather than stacking. This PR keeps `next`'s method name, signature shape, and drop-comment output, and replaces only the predicate body.

## Change A — decide from the generated set, not a heuristic

A reverse-relationship (child-array) `@Field`/`@FieldResolver` references the related entity's GraphQL type by **bare class name**, so it only compiles when that class is declared in the file being generated. `isRelatedTypeOutOfScope` now answers that from `GeneratedTypeAvailability` — the exact set of entities handed to the generator for this file — with one exception: a core (`__mj`) related entity in a **non-core** file, which resolves through the `mj_core_schema_server_object_types` namespace import.

The predicate on `next` infers the generated set from the `entityPackageName` schema→package map alone. But `runCodeGen` narrows the entity list by **both** that map and the `excludeSchemas`/inclusion filters (`runCodeGen.ts:499-511`), so anything dropped by the latter still yields a dangling reference. That is the linked-Open-App break: a base app generated alongside a dependent app that foreign-keys into it emits fields typed with the dependent's classes. The set is ground truth — it cannot drift from what was actually emitted.

The gate emits exactly the **compilable subset** of prior output: an un-emitted class yields a reference that cannot compile, and the gate emits iff the reference is resolvable. Any run that compiled before produces identical output. Monoliths are unaffected — every schema's class is in the one generated set, so cross-schema relationships stay emitted.

**In the core pass**, a core related entity must now satisfy set membership too. The core file *is* the `mj_core_schema_server_object_types` module, so there is no namespace import to resolve through — `next`'s unconditional core exemption was wrong there, though only reachable for a core entity excluded by a non-schema filter.

Signature changes are **additive** — a new optional trailing `availability` parameter on `generateGraphQLServerCode`, `generateServerEntityString`, `generateEntitySpecificServerFileHeader`, and `generateServerGraphQLResolver`. Omitting it preserves the previous heuristic exactly. `GraphQLServerGeneratorBase` is documented as a `RegisterClass` extension point, so removing the existing parameter would break customer subclasses silently; the legacy fallback is covered by a test.

Also fixed on the way through: the per-entity file header emits **relative sibling** imports, so it needs set membership directly — a core type there comes from the namespace import and must not get a `./MJUser` sibling import.

## Change B — opt-in `includeSchemas` positive scope

When set, CodeGen processes only the listed schemas. Pure sugar over `excludeSchemas` (in-scope ⇔ listed ∧ not excluded; no implicit includes — core must be listed explicitly). Absent/empty leaves classic exclude-only behavior untouched.

**The resolution is sourced from the database, not from loaded metadata.** `createNewEntities()` discovers tables with no `EntityID` straight from the DB, filtered only through `excludeSchemas`. A universe derived from `Metadata.Entities` contains by definition only schemas that *already* have entities, so it cannot exclude a never-before-seen schema — precisely the case the feature exists for (a client's own schemas in a deployed instance). That schema would be adopted into metadata on the first run and then excluded on every run after, leaving its entity records permanently orphaned. So:

- **Authoritative pass** in `ManageMetadataBase.manageMetadata`, against `SELECT DISTINCT SchemaName FROM vwSQLTablesAndEntities`, before `createNewEntities()`. If that query fails the run aborts — an include list that silently doesn't apply can create entity metadata for another app's tables.
- **Secondary idempotent pass** in `runCodeGen`, against the metadata schema list, so the scope still applies on `--skipdb` (which bypasses `manageMetadata` entirely but still generates files).

## Testing

- `packages/CodeGenLib` unit suite: **642 passed, 0 failed, 60 skipped** (skipped = PG integration tests, which need a database).
- New `graphql-type-availability.test.ts` covers: emit-when-in-set; drop-when-not-in-set where the schema is *not* in the package map (the case the old heuristic missed); the legacy no-availability fallback; core-in-non-core emitted via namespace import; core-in-core-file dropped when absent from the set; monolith cross-schema preserved; case-insensitive matching.
- New `schema-scope.test.ts` covers the include→exclude resolution, no-op when unset, idempotency across the two passes, and the never-before-seen-schema case.
- The two pre-existing GraphQL codegen suites pass unchanged, which is what demonstrates the additive signature preserved old behavior.

**Not run here:** the deterministic integration tier — this environment has no database. Worth running before merge.

**Not reproduced here:** #3304's live-instance numbers (`bizapps-common` 20 cross-package refs → 0, clean build). Those came from a linked Open App instance I have no way to stand up; they are cited as the origin of the diagnosis, not as something re-verified.

## Scope

Confined to `@memberjunction/codegen-lib`. Changeset is **minor** (new config surface), not patch.

The pre-existing hazard #3304 flagged — destructive CodeGen paths assuming full-database visibility (Angular orphan-dir deletion, whole-file GeneratedEntities rewrite, the CRUD validator) — is **not** touched here and should be hardened before CodeGen scope narrows further.

https://claude.ai/code/session_01UUnZf9oDYm3c9No7toVfFc

---
_Generated by [Claude Code](https://claude.ai/code/session_01UUnZf9oDYm3c9No7toVfFc)_

### DIFF
--- a/.changeset/codegen-type-availability-gate.md
+++ b/.changeset/codegen-type-availability-gate.md
@@ -1,0 +1,11 @@
+---
+"@memberjunction/codegen-lib": minor
+---
+
+Gate reverse-relationship emission on actual GraphQL type availability, and add an opt-in `includeSchemas` scope.
+
+- **Reverse-relationship emission is now decided by the generated entity set, not a schema/package heuristic.** A reverse-relationship (child-array) `@Field`/`@FieldResolver` references the related entity's GraphQL type by bare class name, so it only compiles when that class is declared in the file being generated. `isRelatedTypeOutOfScope` now answers that from `GeneratedTypeAvailability` — the exact set of entities handed to the generator for the file — with one exception: a core (`__mj`) related entity in a non-core file, which resolves through the `mj_core_schema_server_object_types` namespace import. The previous predicate inferred the set from the `entityPackageName` schema→package map alone, but `runCodeGen` narrows the generated entities by *both* that map and the `excludeSchemas`/inclusion filters, so anything dropped by the latter still produced a dangling reference (`TS2304: Cannot find name`) — the break seen when a base Open App is generated alongside a dependent app that foreign-keys into it. The gate emits exactly the compilable subset of prior output, so any run that compiled before is unaffected.
+
+- **New opt-in `includeSchemas` config (positive scope).** When set, CodeGen processes only the listed schemas. It is resolved into `excludeSchemas` — pure sugar over the existing exclude mechanism (in-scope ⇔ named in `includeSchemas` and not in `excludeSchemas`; no implicit includes, so the core schema must be listed explicitly). The authoritative resolution runs in `ManageMetadataBase.manageMetadata` against the **database's** schema list, so schemas MJ has never seen are excluded before `createNewEntities()` discovers their tables; a second idempotent pass in `runCodeGen` keeps the scope applied on the `--skipdb` path. Lets an Open App scope its CodeGen to its own schema without hand-maintaining an exclude list of every other installed app. Absent/empty leaves classic exclude-only behavior unchanged.
+
+Both `GraphQLServerGeneratorBase` signature changes are additive (a new optional trailing `availability` parameter), so existing subclasses and callers keep compiling and retain their previous behavior.
--- a/packages/CodeGenLib/src/Config/config.ts
+++ b/packages/CodeGenLib/src/Config/config.ts
@@ -459,6 +459,22 @@ const configInfoSchema = z.object({
     { name: 'auto_index_foreign_keys', value: true },
   ]),
   excludeSchemas: z.string().array().default(['sys', 'staging']),
+  /**
+   * Opt-in POSITIVE scope list. When set (non-empty), CodeGen processes ONLY these schemas — every
+   * other schema present in the DATABASE is treated as excluded, including schemas MJ has never seen
+   * before. It is pure sugar over {@link excludeSchemas}: it is resolved into `excludeSchemas` (see
+   * `applyIncludeSchemaScope`) before metadata management and again before file generation, so
+   * nothing downstream changes. In-scope ⇔ named in `includeSchemas` AND not in `excludeSchemas`
+   * (include shrinks the addressable space; exclude overlays on top). No hidden auto-includes — a
+   * schema, including the MJ core schema, is in scope ONLY if listed explicitly. Leave
+   * undefined/empty for the classic exclude-only behavior (unchanged).
+   *
+   * Primary use: scope an Open App's CodeGen to just its own schema without hand-maintaining an
+   * exclude list naming every other installed app — a list that is O(N²) to maintain and, more
+   * importantly, cannot name schemas the app does not know about (such as a client's own schemas in
+   * a deployed instance).
+   */
+  includeSchemas: z.string().array().optional(),
   excludeTables: tableInfoSchema.array().default([
     { schema: '%', table: 'sys%' },
     { schema: '%', table: 'flyway_schema_history' }
--- a/packages/CodeGenLib/src/Database/manage-metadata.ts
+++ b/packages/CodeGenLib/src/Database/manage-metadata.ts
@@ -10,6 +10,7 @@ import { ApplicationInfo, CodeNameFromString, EntityFieldExtendedType, EntityFie
 import { MJApplicationEntity, MJEntityFieldSchema } from "@memberjunction/core-entities";
 import { logError, logMessage, logStatus, startSpinner, updateSpinner, succeedSpinner } from "../Misc/status_logging";
 import { SQLUtilityBase } from "./sql";
+import { applyIncludeSchemaScope } from "./schema-scope";
 import { AdvancedGeneration, EntityDescriptionResult, EntityNameResult, SmartFieldIdentificationResult, FormLayoutResult, VirtualEntityDecorationResult } from "../Misc/advanced_generation";
 import { CodeGenReporter } from "../Misc/codegen-reporter";
 import { mapExternalNativeTypeToMJ } from "../Misc/externalTypeMapping";
@@ -1302,8 +1303,6 @@ export class ManageMetadataBase {
             }
          }
       }
-      const excludeSchemas = configInfo.excludeSchemas ? [...configInfo.excludeSchemas] : [];
-
       // Ensure the platform's metadata-management support objects exist and
       // match this CodeGenLib version. These routines/views are CodeGen's own
       // machinery — when the provider supplies DDL (PostgreSQL), we install it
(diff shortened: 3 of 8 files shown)


MarceloT-BC:
> 🤖 **Generated by Claude (AI), run by Marcelo.** Everything below is machine-produced: the numbers are measured from a live test environment, and the opinions are the model's, not Marcelo's. Treat the recommendations as input to review rather than a maintainer's position.

Validated against a test database built from this PR's branch (`claude/pr-3304-reverse-relationship-fvkivk`), SQL Server, with three apps present and built locally from source in dependency order: `bizapps-common` (base) + `bizapps-accounting` + `bizapps-tasks` (both hold real foreign keys into common's schema). That's the environment the PR notes it couldn't set up, so here are the numbers it was missing — plus one thing that looks worth resolving before merge.

## The missing live evidence — reproduced, including the counterfactual

Same database, same apps; **only** `graphql_server_codegen.ts` swapped between versions, CodeGenLib rebuilt, then CodeGen re-run for `bizapps-common` each time:

| emitter | cross-package refs in `bizapps-common` `generated.ts` | drop comments | compiling `bizapps-common` |
|---|---|---|---|
| **`next` as-is** (f96804b9c package-map heuristic) | **20** | 0 | **fails — TS2304** |
| **this PR** (availability set) | **0** | 10 | **green, 5/5 packages** |

Two things worth calling out:

1. **20 is exactly the figure #3304 reported**, so the diagnosis carried over intact.
2. **`next`'s own fix does not cover this case.** The 20 refs are on current `next` *with* f96804b9c in place — so Change A is load-bearing, not defensive. This is the empirical confirmation of the PR's argument that a predicate modelling only the `entityPackageName` map misses whatever `excludeSchemas` drops.

The 10 drop comments are 5 relationships × (`@Field` + `@FieldResolver`), all `MJ_BizApps_Tasks: *` children of `MJ_BizApps_Common: Person` — matching the mechanism as described. Regenerated diff was 14 insertions / 4 deletions, entirely those comments; no legitimate content lost.

**Unit suite re-run independently** (not taken on trust): 642 passed, 0 failed, 60 skipped — matches the PR exactly.

## Change B — both placements are needed, neither is redundant

Flagging this so nobody deletes one as duplication later. They're asymmetric:

- **`ManageMetadataBase.manageMetadata`** is the authoritative pass: its universe is the *database* (`vwSQLTablesAndEntities` is `sys.all_objects ⋈ sys.schemas`, so a strict superset of schemas present in metadata), and it lands before `createNewEntities()`. It is the only pass that can exclude a never-before-seen schema — the case the feature exists for.
- **`runCodeGen`** sits above the `if (!skipDB)` block, so it is the *only* scope application on `--skipdb`, where `manageMetadata` never runs at all.

So in a normal database run the `runCodeGen` pass is subsumed by the database pass (same excludes plus more) — its unique job is `--skipdb`, which is exactly why #3304 put it there. Keep both; a one-liner on the manage-metadata side saying "authoritative; the runCodeGen pass covers `--skipdb` only" would make that obvious to the next reader.

## Small note: the "run aborts" claim doesn't match the code

The PR body and the code comment both say a failed schema-list query aborts the run. It doesn't: `manageMetadata` returning `false` sets `pipelineSuccess = false`, logs, and **continues** into SQL generation and file generation (`runCodeGen.ts:257-270`). The *behavior* is still safe — the early `return false` happens before `createNewEntities()`, so nothing gets adopted — so this is a wording fix, not a code fix. Worth correcting so no one later relies on "aborts" as a guarantee.

## Worth resolving before merge: `includeSchemas` ships with a silent destructive path

This is the finding with the largest blast radius. It reproduced end-to-end in the same environment: set `includeSchemas: ['__mj_BizAppsCommon']` — one app schema, `__mj` not listed, exactly as the docs describe ("no hidden auto-includes … the MJ core schema is in scope ONLY if listed explicitly") — and ran one CodeGen.

**Result of that single run:**

| artifact | before | after |
|---|---|---|
| `packages/MJServer/src/generated/generated.ts` | 98,184 lines · 383 resolver classes | **21 lines · 0 classes** |
| `packages/MJCoreEntities/src/generated/entity_subclasses.ts` | 117,728 lines · 383 entity classes | **10 lines · 0 classes** |
| Angular core-entity-form directories | 383 | **0 — deleted** |

**769 files changed, 5 insertions, 307,039 deletions.**

The chain, all four links verified:

1. `applyIncludeSchemaScope` appends every unlisted schema to `excludeSchemas` — including `__mj`. Correct per the design; that's the documented semantics.
2. `runFileGenerationPhase` filters `includedEntities` by `excludeSchemas`, so `coreEntities` becomes `[]`.
3. The core passes are guarded **only** on the output dir being configured (`runCodeGen.ts:525-535`), never on entity count. `generateGraphQLServerCode([], …)` writes a header + `export {}` over the target and **returns `true`**; `generateAllEntitySubClasses` has the same hole; `angular-codegen.ts:116-153` then deletes every entity-form directory as "orphaned."
4. Nothing downstream notices. CodeGen's own phases all report ✔ — including "Post-CodeGen CRUD validation passed (412 entities checked)" and "System integrity checks completed," both of which are looking at the database while the TypeScript layer is being emptied. The run only exits non-zero at the very end because the post-codegen `npm build` commands fail — i.e. the signal is a compile error in the wreckage, not a diagnosis, and it arrives after everything is already written to disk.

Why this may warrant a guard in *this* PR rather than the follow-up hardening track: the stated primary use case is "a client's own schemas in a deployed instance" — and a deployed instance is precisely the configuration that has core output dirs pointing at MJServer/MJCoreEntities. The failure mode is also inverted from what a user would expect: omitting a schema from an *include* list reads as "don't process that schema," not "rewrite its committed artifacts to empty." And because `excludeSchemas` in the same config is additive and non-destructive, there's no prior behavior that trains someone to expect this.

One low-cost way to close it: bail out of the core passes when the entity list is empty (an empty generated set is never a legitimate request), or hard-error at config resolution when `includeSchemas` is set, omits the core schema, and core output dirs are configured. Either is a few lines and keeps the feature's semantics intact.

This is the F1–F7 hazard class already filed as its own audit, and the PR is explicit about not touching it. The narrower point is: this PR ships the knob that makes that latent hazard reachable from a documented, supported configuration — so at minimum the ordering argument ("harden before CodeGen scope narrows further") points at doing the guard first.

---

Full logs, scripts, and per-case output for everything above are available if useful — happy to attach.

PR #3351, opened 2026-07-29
fix(sqlconverter): correct nine T-SQL→PostgreSQL conversion defects that break open-app installs

MS-BC:
## What this fixes

Converting an open app's SQL Server migrations to PostgreSQL (`mj migrate convert`) produced files that either failed to apply or — worse — applied "successfully" while leaving the database in a wrong state.

Nine defects, all found while making a real open app install on PostgreSQL 17, each verified against a live database.

**Statements silently dropped** (emitted as commented-out "could not parse" text):

1. A `DECLARE` with a subquery initializer — `DECLARE @x UNIQUEIDENTIFIER = (SELECT ...)` — was not recognised, so it and every statement depending on it were dropped.
2. The same rule did not cast boolean literals, emitting `1`/`0` for BIT columns.
3. `SET @x = ...` converted to `SET v_x = ...`, which in PostgreSQL is a configuration-parameter statement, not an assignment.

**Applied, but with the wrong data:**

4. `castBooleanInsertValues` rewrote only the **first** INSERT in a file; every later INSERT kept `1`/`0` for boolean columns. It now walks all of them, using a string- and paren-aware scan to find each statement's end.

**Identifier folding.** PostgreSQL folds an *unquoted* identifier to lowercase and preserves a *quoted* one, so quoting is semantic. T-SQL brackets carry no case meaning, so translating every bracket into a quote changes what the SQL means. Each of the following quoted one side of a pair and not the other:

5. `CREATE SCHEMA` was matched inside **comment text**, so prose mentioning a schema won the match — creating a phantom schema and dropping the real statement. Comments are now stripped before matching.
6. The schema was emitted **quoted**, preserving its case, while every reference to it was unquoted and folded. A live MJ database shows the result: two schemas, `__mj_BizAppsSecureMessaging` holding 1 object and `__mj_bizappssecuremessaging` holding the other 18. It now emits unquoted, quoting only a name that cannot survive folding.
7. A mixed-case join alias in a view was quoted where **defined** but not where **referenced**, so the alias was unresolvable and the view failed to create.
8. A schema composed from a migration placeholder — `[${mjSchema}_BizAppsCommon]`, how an open app names a sibling app's schema — came out quoted. The placeholder is substituted as plain text at apply time, giving a case-preserved schema that does not exist.
9. Inline `CREATE TABLE` constraint names were left unquoted while `ALTER TABLE` quoted them. A later `ALTER TABLE ... DROP CONSTRAINT` therefore missed the folded name it meant to remove, and the following `ADD` succeeded under a different one — leaving **both** constraints on the table, so the original narrower `CHECK` kept rejecting the values the migration was written to allow. Only one non-fatal line in the install log hinted at it. The quoting logic is now a shared helper so the two sites cannot drift apart again.

## Why the direction of each casing fix is what it is

Both casing questions (6 and 9) were settled by querying live MJ PostgreSQL databases rather than by preference:

- **Schemas → unquoted.** A live database contained the same schema twice; the *lowercase* one held 18 of the 19 objects. Lowercase is what everything else resolves to.
- **Constraints → quoted.** `__mj*` schemas hold 1650 mixed-case constraint names against 115 folded ones, so `AlterTableRule` was already right and `CreateTableRule` was the one out of step.

## Verification

The converted migrations of a real open app now apply to PostgreSQL 17 with **zero errors** — 16/16 tables, 16/16 views, 54/54 CRUD functions, 16/16 triggers — and a full create → update → delete round-trip works through the generated views, including the cross-schema join that defects 7 and 8 broke.

Suite: **1120 passed, 3 skipped, 0 failed** (9 new tests). Two existing assertions were updated: they encoded the quoted-schema behaviour of defect 6, which the live databases show to be wrong.

### Adversarial re-verification (2026-08-03)

The branch was re-reviewed and re-tested by converting through the **default `convertFile` path** — the one `mj migrate convert` uses without `--split`, and the only one that reaches these rules — then applying the output to fresh PostgreSQL 17 databases.

**The fixes hold.** A migration exercising all nine defects produced **5 hard errors on `next`**: the guarded `CREATE SCHEMA` match was won by a *comment* containing the phrase, so `next` emitted `CREATE SCHEMA IF NOT EXISTS "form"` and every table, view and insert below failed with `schema "__mj_demoapp" does not exist`. On this branch the schema is created correctly, constraint names are quoted in `CREATE TABLE`, the `DECLARE` initializer survives instead of becoming `-- Could not parse`, `SET @x` becomes an assignment, and BIT `1`/`0` become `TRUE`/`FALSE`.

**Two problems surfaced that the unit suite could not see, both now fixed in this branch:**

1. **A regression introduced here.** Widening the ViewRule alias test from `/^[A-Z]/` to `/[A-Z]/` also caught aliases introduced **without** the `AS` keyword. `quoteAsAliases` only quotes `AS <alias>` *definitions*, so an implicit `FROM __mj."vwEntities" relatedEntity` definition folds to lowercase while its references were now quoted — the alias resolves to nothing. Converting the committed `B202602151200__v5.0__Baseline.sql` and applying it:

   ```
   === CREATE VIEW from next output ===   CREATE VIEW
   === CREATE VIEW from pr output ===
   ERROR:  missing FROM-clause entry for table "relatedEntity"
   LINE 23:  er."RelatedEntityID" = "relatedEntity"."ID"
   ```

   Converting all **208 files** of `migrations/v5` found **6 such aliases across 5 baseline files** — `relatedEntity`, `cirInner`, `rlsC`, `rlsR`, `rlsU`, `rlsD`. `rlsC` is defined implicitly in `__mj.vwEntities`, so MJ's most foundational view was affected too. Nothing caught it because `pg-migration-regression.test.ts` filters `f.startsWith('V')` (so `B*` baselines are never converted) and `pg-migrations.yml` applies **committed** `.pg.sql` files rather than regenerating them — the same blind spot its own comments cite for the earlier hand-fixed `vwEntityPermissions` alias.

   `quoteColumnRefs` now collects the aliases whose definitions are actually case-preserved and quotes references only for those. After the fix the corpus rescan finds **0** offending files, `vwEntityRelationships` creates cleanly, and blast radius versus `next` drops from **64 files / 1393 lines** to **63 / 1233** — the delta versus this branch as submitted is exactly those 6 aliases and nothing else.

2. **A pre-existing defect this branch exposed.** Now that a `DECLARE` item with a subquery initializer is kept rather than dropped, the initializer reaches PostgreSQL unconverted — `SELECT TOP 1` failed with `syntax error at or near "1"`. `convertTopToLimit` is applied to the initializer, folded back to one line so it does not split the declaration.

With both fixes the demo migration applies with **zero errors** and the end state proves the claim:

```
                  ID                  |  Status  |    Payer
--------------------------------------+----------+-------------
 1a32e282-16b3-4cf1-94be-09cf814a8ecf | Refunded | Seed Person
```

`Refunded` is accepted, so the widened `CHECK` genuinely replaced the narrow one (defect 9); `IsActive` is stored `FALSE`, not `1`; the view resolves.

Both new tests were falsification-checked — reverting either fix fails exactly its own test and no other.

### Re-verified through `mj migrate` itself, against live PostgreSQL 17 databases

The checks above converted files and applied them with `psql`. Repeated end to end through the **actual CLI** — `mj migrate convert` to produce the PG files, then `mj migrate` (Skyway) to apply them to fresh databases on PostgreSQL 17.10. Three converters were built and run over the identical source corpus: `next`, this branch **as originally submitted**, and this branch **as it now stands**.

**The regression was real and would have broken every fresh PostgreSQL install.** Running `mj migrate` over the v5.0 baseline:

| converter | `mj migrate` | `__mj` tables | `__mj` views | `vwEntities` | `vwEntityRelationships` |
|---|---|---|---|---|---|
| `next` | exit 0 | 274 | 282 | 272 rows | 520 rows |
| this branch **as submitted** | **exit 2** | **1** | **0** | — | — |
| this branch **now** | exit 0 | 274 | 282 | 272 rows | 520 rows |

```
FAILED: B202602151200__v5.0__Baseline.pg.sql
Error: Failed at batch 1/1 (lines 1-143980): missing FROM-clause entry for table "cirInner"
```

The install aborted on the baseline, leaving the database with one table and no views. With the alias fix the outcome is identical to `next` on every count — no regression, byte-for-byte the same end state.

**And the PR's own fixes hold through the same command.** The open-app migration exercising all nine defects, converted and applied by `mj migrate`:

| converter | `mj migrate` | `__mj_DemoApp` tables | result |
|---|---|---|---|
| `next` | **exit 2** — `schema "__mj_demoapp" does not exist` | 0 | nothing installed |
| this branch | exit 0 | 3 | row seeded, view resolves |

```
                  ID                  |  Status  |    Payer
--------------------------------------+----------+-------------
 64b48823-e7e8-4973-a343-90a5ebbc1fa8 | Refunded | Seed Person
```

**Whole-corpus regression check.** `mj migrate convert` over all 208 files of `migrations/v5` with both converters, then `mj migrate` into two isolated fresh databases: both reach exactly the same point and both stop at the same file with the same error —

```
FAILED: V202607141000__v5.48.x__ListDetail_Index_Optimization.pg.sql
Error: Failed at batch 1/1 (lines 1-76): syntax error at or near "IF"
```

— and the regenerated file is **byte-identical between the two branches**, while the *committed* `.pg.sql` for it differs from anything the converter emits (it is maintained by hand). So that is a pre-existing converter gap in a hand-maintained file, untouched by this PR; CI does not hit it because it applies the committed `.pg.sql` rather than regenerating. Up to that common stopping point both databases are identical: 389 views each, `vwEntities` 379 rows, `vwEntityRelationships` 793 rows.

### Known, not fixed here

A mixed-case schema without a leading underscore is quoted in view references on **both** branches; that is pre-existing, not a regression.

*(The guarded-`CREATE SCHEMA` placeholder gap that this section previously listed is now fixed — see below.)*

### Placeholder `CREATE SCHEMA` — fixed in `50891c7` after the original submission

The gap this PR originally left open turned out to reproduce defect 6 for exactly the case the placeholder exists to serve. `convertIdentifiers` ([ExpressionHelpers.ts:177](packages/SQLConverter/src/rules/ExpressionHelpers.ts#L177)) deliberately emits placeholder schema *references* unquoted, so Postgres folds them to lowercase — but the guarded `CREATE SCHEMA` was emitted quoted, preserving case. The schema is created as `__mj_BizAppsCommon`; every reference then looks for `__mj_bizappscommon` and finds nothing.

`tryConvertConditionalSchema` now branches three ways instead of two: a plain literal name lowercases as before; a name that is only placeholders + word characters lowercases *outside* the placeholder spans (the placeholder text must survive verbatim for Flyway to substitute it); anything else still stays quoted, since it cannot survive unquoted at all.

**Converter A/B** — same source migration (guarded `CREATE SCHEMA [${mjSchema}_BizAppsCommon]`, then a table, a view, and an INSERT in that schema), converted by a build of the branch as originally submitted and by a build of the branch now. The output differs on exactly one line:

```diff
-CREATE SCHEMA IF NOT EXISTS "${mjSchema}_BizAppsCommon";
+CREATE SCHEMA IF NOT EXISTS ${mjSchema}_bizappscommon;
```

The placeholder is preserved verbatim; only the literal suffix folds.

**Applied live to PostgreSQL 17.10**, `${mjSchema}` substituted to `__mj` the way Flyway substitutes it, each output into its own fresh database:

| | before | after |
|---|---|---|
| `CREATE SCHEMA` | ok (as `__mj_BizAppsCommon`) | ok (as `__mj_bizappscommon`) |
| `CREATE TABLE` | `ERROR: schema "__mj_bizappscommon" does not exist` | `CREATE TABLE` |
| `CREATE VIEW` | `ERROR: relation "__mj_bizappscommon.Party" does not exist` | `CREATE VIEW` |
| `INSERT` | `ERROR: relation "__mj_bizappscommon.Party" does not exist` | `INSERT 0 1` |
| resulting schema | empty — `\dt` finds no relations | table + view present |

Post-fix, `SELECT * FROM __mj_BizAppsCommon."vwParty"` returns the seeded row:

```
                  ID                  | Name
--------------------------------------+------
 11111111-1111-1111-1111-111111111111 | Acme
(1 row)
```

**Unit:** `ConditionalDDLRule.test.ts` 39/39; package suite **1122 passed / 3 skipped** (was 1120 — two tests added, one asserting the placeholder-bearing name folds only outside the placeholder, one asserting a non-foldable placeholder name stays quoted). Falsification: with the new branch disabled, exactly 1 test fails.

## Compatibility

SQL Server output is untouched — every change is in a `tsql → postgres` rule path. Already-converted files are unaffected; the next `mj migrate convert` produces corrected output.


MS-BC:

MS-BC:
## Independent live differential at head `ac0a93f` — real committed artifacts, PostgreSQL 17

Re-verified without reusing the earlier install harness. Input is **not synthetic**: the committed migrations of a real open app (`bizapps-caliber` — 17 hand-written migrations + 3 CodeGen runs containing 29 views). Each file converted through the default `convertFile` path — the one `mj migrate convert` uses — with the **pre-fix build** (this PR's merge base) and the **PR build**, then applied to PostgreSQL 17.10.

### Defect 7 — mixed-case join aliases in views

Verbatim from a real CodeGen migration, same view, two builds:

```sql
-- pre-fix: alias DEFINED quoted, REFERENCED unquoted
    mjBizAppsCaliberEngagement_EngagementID."Name" AS "Engagement"
FROM __mj_BizAppsCaliber."ContinuityDigest" AS c
INNER JOIN __mj_BizAppsCaliber."Engagement" AS "mjBizAppsCaliberEngagement_EngagementID"
  ON c."EngagementID" = mjBizAppsCaliberEngagement_EngagementID."ID"

-- this PR: both sides quoted
    "mjBizAppsCaliberEngagement_EngagementID"."Name" AS "Engagement"
  ...ON c."EngagementID" = "mjBizAppsCaliberEngagement_EngagementID"."ID"
```

Applied live to PG 17:

```
----- pre-fix -----  ERROR: missing FROM-clause entry for table "mjbizappscaliberengagement_engagementid"
----- this PR -----  CREATE VIEW
```

Across all three CodeGen runs:

```
pre-fix : 29 views, 26 carry an unquoted reference to a quoted mixed-case alias
this PR : 29 views,  0
```

26 of 29 real generated views fail to create before this change.

### Defect 9 — constraint quoting in `CREATE TABLE`

130 diff lines across the app's real DDL, all of this shape:

```
-  CONSTRAINT CK_DifficultyProfile_ControllerMode CHECK (...)
+  CONSTRAINT "CK_DifficultyProfile_ControllerMode" CHECK (...)
```

`AlterTableRule` already quoted, so every one of these was a name a later `ALTER TABLE ... DROP CONSTRAINT` would fold past and miss. Now emitted through the shared helper, so the two sites cannot drift.

### Suite

`packages/SQLConverter` — **1122 passed, 3 skipped, 0 failed** at head.

### One thing outside this PR

Applying the converted app DDL against a converted MJ database surfaced `relation "__mj.company" does not exist`: core tables exist as quoted `"Company"` while a cross-schema FK reference is emitted unquoted and folds. **Identical on both builds** (212 vs 212 errors), so not a regression from this PR and not one of the nine — but it is the same class of folding asymmetry and worth a follow-up.

rkihm-BC:
Closing this in favour of #3703, which has absorbed it entirely. Recording why, and what I verified, so nothing here is lost.

## #3703 already contains all of it

All four commits on this branch are ancestors of #3703's head:

```
38f6532e75  53693fa9f7  50891c7179  ac0a93f5a6   → all present in feat/pg-runtime-parity
```

I checked every one of the 11 files this PR touches for content that exists here but *not* in #3703, rather than assuming the ancestry implied it. Three differences came back, all of them #3703 improving on this branch:

| here | #3703 |
|---|---|
| `quoteConstraintNames` / `stripComments` | renamed to `QuoteConstraintNames` / `StripComments`, per MJ's PascalCase convention |
| `/\bCONSTRAINT\s+([A-Za-z_]\w*)\b/gi` | `(?!IF\s+EXISTS\b)` added — `DROP CONSTRAINT IF EXISTS Foo` was becoming `DROP CONSTRAINT "IF" EXISTS Foo` |
| `/^(\s*)DECLARE\s+(.+?);\s*$/gim` | `[ \t]*` — `\s*` was consuming a preceding blank line's newline, which pushed declarations out of the DECLARE section |

No test was dropped: I diffed the `it()` names, not just lines.

## The reason to close rather than merge

One test exists in both branches with **deliberately opposite assertions**:

```js
// here
expect(result).toContain('CREATE SCHEMA IF NOT EXISTS __mj_udt;');
expect(result).not.toContain('"__mj_UDT"');

// #3703
expect(result).toContain('CREATE SCHEMA IF NOT EXISTS "__mj_UDT";');
expect(result).not.toContain('__mj_udt');
```

#3703 subsequently established that `__mj_UDT` is the one mixed-case schema with a producer *outside* the migration set — `UDT_SCHEMA_NAME` in `AI/DatabaseDesigner/core` is a mixed-case literal and every emitting path quotes it, so a live database holds `"__mj_UDT"` case-preserved. It is carved out through `RUNTIME_CREATED_SCHEMAS`. Folding it would leave the Database Designer writing into a schema no migration created, and orphan every UDT entity in `vwSQLTablesAndEntities`, which joins schema names case-sensitively.

So merging this branch **after** #3703 would revert that carve-out — and silently, because this branch's own test would pass while asserting the behaviour #3703 identified as wrong. Closing removes the ordering hazard outright.

## Nothing is lost

`.changeset/sqlconverter-pg-conversion-defects.md` is byte-identical in #3703, so the release notes for all nine defects ship unchanged.

## One follow-up worth filing separately

`endOfStatement` (introduced here, identical in #3703) is string- and paren-aware but not **comment**-aware, so an apostrophe in a `--` comment inside a multi-INSERT batch desyncs the scan and the region swallows the following statement:

```sql
INSERT INTO __mj."Flags" ("ID","IsActive") VALUES
  (1, 1), -- Bob's row
  (2, 0);
INSERT INTO __mj."Other" ("ID","Qty") VALUES (3, 1);
```
```
→ INSERT INTO __mj."Other" ("ID","Qty") VALUES (3, TRUE);   -- Qty is INTEGER
```

**This is pre-existing, not introduced here** — I replicated `next`'s version on the same input and got byte-identical output, because it walks to end-of-text unconditionally. This PR genuinely improves it by bounding each statement. The only thing worth changing is the doc comment, which currently promises more than the code delivers ("an intervening statement can never be rewritten with the wrong table's positions"). `StripComments` / `segmentSQL` already exist in that same file. It fails loudly at apply time rather than silently, so it is not urgent.

## For the record

The work itself was good and is not being discarded — it ships in #3703. The casing decisions in particular were settled by querying live databases rather than by preference (18 of 19 objects in the lowercase schema; 1,650 mixed-case constraint names against 115 folded), the ViewRule alias-definition gating is the right shape, and the self-caught alias regression — found by rescanning all 208 files of `migrations/v5`, with an honest account of why the unit suite could not see it — is the kind of thing that is much easier to leave unmentioned than to go find.

PR #3352, opened 2026-07-29
fix(open-app): close four silent-failure gaps in config writes, class registration, and update checks

MS-BC:
All four defects here share a failure mode: **nothing goes red.** The app installs, `__mj.OpenApp` reports `Active`, the CLI prints a success line — and the config the runtime actually loads is wrong, or a registered class never runs, or an update check reports a version belonging to a different app.

Each item below was verified against `origin/next` before being written; several sibling claims from the original task list turned out to be already fixed by intervening PRs and were dropped from scope rather than shipped as no-ops.

---

### 1. `dynamicPackages` idempotency matched the whole file, not the target array

A `shared` package must be written to **both** `dynamicPackages.server` and `dynamicPackages.client`. The "is this entry already present?" check tested the entire file, so the server entry written moments earlier satisfied the client check and **the client insert was skipped**. The package never reached `dynamicPackages.client`, so its `@RegisterClass` components were tree-shaken out of the browser bundle — with no error raised anywhere.

Both arrays are now located through one shared `LocateDynamicArray` helper, and the existence check runs against the **target array's body only**. Idempotency within an array is preserved, and remains scoped per `(PackageName, AppName)` so two apps shipping the same package name each get their own entry.

### 2. Upgrades were add-only, so `mj.config.cjs` accumulated forever

The `Add*` writers are additive and idempotent — exactly right for install. On **upgrade** they never remove what the new manifest dropped, so the config converges on the union of every version ever installed, and a package deleted in v2 keeps being bootstrapped until it fails to resolve at startup.

Adds `PruneDynamicPackagesNotInManifest`, called from upgrade Step 7 **before** the adds. Ordering is load-bearing: pruning afterwards would delete the entries just written, because the keep-set is derived from the same manifest. Three properties worth noting in review:

- **Surviving entries are left byte-identical** rather than removed-and-re-added — otherwise a disabled app being upgraded would come back with the default `Enabled: true`, silently re-enabling something the operator turned off.
- **Keep-sets are per-array** (server = server + shared, client = client + shared), so a server-only package isn't pruned out of `server`.
- **An unrecognized entry shape is a no-op, never a guess** — the prune only removes a block whose `PackageName` it could actually read.

A **schema rename** across versions is handled too: the old `entityPackageName` mapping and `excludeSchemas` entry are removed. An unparseable historical manifest warns and lets the upgrade continue rather than failing it.

### 3. Unrelated `@RegisterClass` key collisions were completely silent

`@RegisterClass` passes `priority = 0`, which routes to the auto-increment branch (`highestPriority + 1`), so **the later registration always wins**. That is the documented, correct contract for an inheritance chain — a subclass superseding its base is the entire point — and it is silently wrong for two *unrelated* classes that merely collide on a key. Because only `priority > 0` ever warned, in practice **nothing warned**.

`ClassFactory.Register` now emits a `console.warn` naming every prior unrelated registration for that `(base class, key)` pair. The relatedness test is a new `AreClassesRelated` in `ClassUtils`, which compares by **name as well as identity** — a module loaded through two paths yields two distinct constructor objects with the same name, and that must not read as a collision. This mirrors why `GetAllRegistrations` already matches on `class.name`.

**Registration behavior is unchanged.** The warning is diagnostic only; refusing the registration would break every intentional override that currently relies on last-one-wins.

### 4. `mj app check-updates` reported wrong versions and hid its own failures

Three distinct problems in one loop:

- It built a bare `{ Token }`, **dropping the per-repo `TokenMap`** that `install` and `upgrade` both use — so every private repo whose token lives there reported "up to date" forever. Now shares a `buildGitHubOptions` helper with those commands, so a repo reachable there is reachable here.
- It **dropped each app's `Subpath`**. A multi-app repo versions each app with a scoped `<subpath>@<semver>` tag; without the subpath, resolution falls back to repo-wide `v*` tags and reports a **sibling app's** version as this app's latest. The engine's `GetLatestVersion` already accepted the third argument — the caller just never passed it.
- **One throwing app killed the sweep**, or worse, vanished from a report that still concluded "All apps are up to date." A rate-limited or unreachable repo is an *unknown*, not a no-op.

The loop moved into a new `CheckAppsForUpdates` helper with the version lookup **injected**, so the behavior is unit-testable without a token or a network call. Failures are collected per app and reported in their own section.

---

### Tests — 35 new cases

| Package | Result |
|---|---|
| `MJGlobal` | **601 / 601** (15 new) |
| `OpenApp/Engine` | **329 / 329** (11 new) |
| `MJCLI` — `update-check.test.ts` | **11 / 11** |

The `config-manager` tests were **run against the pre-fix source to confirm they actually fail there** (the cross-array test fails on the real regression; the prune tests fail because the capability didn't exist). They are guards, not decoration.

Three existing orchestrator tests that mock `config-manager` gained the new `PruneDynamicPackagesNotInManifest` entry in their mock factories, plus two new ordering cases asserting prune-runs-before-add and that a failed prune aborts the upgrade instead of adding on top of an unknown config.

**Note on MJCLI's other suites:** `MJCLI.test.ts`, `convert-decision.test.ts`, `manifest-client-bootstrap.test.ts`, `app-commands.test.ts` and the `claude-pack` tests do not run in my local worktree — an Ajv v6/v8 resolution skew in the borrowed dependency tree. I verified the **identical** failure set on a pristine `next` checkout in the same environment, so these changes add no MJCLI failures; CI is the authority here.

---

## Adversarial re-verification against a live database and the repo's real config (2026-08-03)

Re-reviewed and re-tested by driving the **built** `config-manager` against a byte-copy of this repo's own `mj.config.cjs`, by loading the real server bootstrap manifest, and by resolving versions for the **9 open apps actually installed** in a live MJ database. `next` and this branch were run side by side.

### Defect 1 — reproduced and fixed

Install a manifest with a `shared` package, on the real config:

```
next   install v1 : server=[@demo/server,@demo/shared]  client=[@demo/client]
pr     install v1 : server=[@demo/server,@demo/shared]  client=[@demo/client,@demo/shared]
```

`@demo/shared` never reached `dynamicPackages.client` on `next`. Re-running the install is a no-op on both, so idempotency is intact, and the 20 pre-existing MJ entries are untouched.

### Defect 2 — reproduced and fixed

Upgrade to a v2 manifest that drops `@demo/server`:

```
next   upgrade v2 : server=[@demo/server,@demo/shared]   <- stale entry survives forever
pr     upgrade v2 : server=[@demo/shared]
```

Four adversarial properties were checked on the real config, all holding on this branch:

| Case | Result |
|---|---|
| An entry manually set `Enabled: false`, then upgraded | stays `false` — not removed-and-re-added |
| Two apps registering the **same** package name; prune app 1 | app 2's identical entry untouched |
| Prune the last entry out of an array | normalized to `[]`, file still valid JS |
| An unrelated `client: [ ... ]` section elsewhere in the file, with `dynamicPackages.client` absent | not targeted, left byte-identical, file unchanged |

The last case also surfaced the write-path's JS-validity guard: fed a deliberately malformed config, the prune refused the write and reported it rather than corrupting the file.

### Defect 3 — measured for noise, then demonstrated

The concern with a new always-on warning is that it floods a real boot. It does not. Loading `@memberjunction/server-bootstrap` — every shipped `@RegisterClass` — gives **521 registrations across 490 (base, key) groups**, so 31 genuine re-registrations, and the predicate would warn on **0** of them: every one is an inheritance-chain override.

The warning does fire when it should:

```
next  inheritance-chain registrations: 0 warnings   unrelated registration: 0 warnings   resolves: Unrelated
pr    inheritance-chain registrations: 0 warnings   unrelated registration: 1 warning    resolves: Unrelated
```

Resolution is identical on both — diagnostic only, as claimed.

### Defect 4 — reproduced on live data

The live database has **9 apps installed from one repo** (`MemberJunction/Integrations`), each with its own `Subpath` — exactly the multi-app case. Resolving the latest version with and without the subpath:

```
app                     current   next (no subpath)                             pr (with subpath)
connector-orcid         1.1.2     @memberjunction/connector-wild-apricot@1.3.0  null
connector-hubspot       1.1.0     @memberjunction/connector-wild-apricot@1.3.0  null
connector-nimble-ams    1.3.0     @memberjunction/connector-wild-apricot@1.3.0  null
connector-pheedloop     1.4.0     @memberjunction/connector-wild-apricot@1.3.0  null
```

On `next` every app resolves the **same sibling app's tag**, and since that string never equals the app's own version, `check-updates` reports a bogus update for all nine. This branch scopes the lookup and stops reporting a wrong answer.

### Follow-on gap this exposed — pre-existing, not introduced here

**Correction to an earlier version of this note.** I first wrote that `ScopedTagPrefix` derives a prefix the repo doesn't use. That was wrong — `MemberJunction/Integrations` carries **both** conventions, and `Platform-ORCID@1.1.3` does exist. The real cause is **pagination**: the repo has 344 tags (4 pages at `per_page=100`) and `ListGitHubTags` reads only page 1; `Platform-ORCID@1.1.3` is tag 169, so it is invisible. Every app there resolves to `null`.

So on this branch alone the behaviour goes from *wrong and loud* to *correct and silent* — an improvement, but still not a true answer. That pagination fix is #3353; with both branches together, `mj app check-updates` against live `MJ_Local` returns real upgrade targets (`connector-hubspot 1.1.0 -> 1.1.2`, `connector-orcid 1.1.2 -> 1.1.3`, `connector-pheedloop 1.4.0 -> 1.4.1`, `connector-nimble-ams 1.3.0 -> 1.3.1`).

### Suites

`MJGlobal` **601 / 601**. `OpenApp/Engine` **329 passed / 9 failed** — the 9 are `entity-teardown-*`, which fail **identically on a pristine `next` checkout** in the same environment (a stale `@memberjunction/core` in the borrowed dependency tree lacks `SQLDialect.ForeignKeyGraphSQL` / `.AtomicBatchScript`). Not from this branch; `next` scores 309 passed / 9 failed, so this PR adds 11 passing tests and no failures.

### One inaccuracy in the description above

The claim that pruning after the adds "would delete the entries just written, because the keep-set is derived from the same manifest" is backwards — the keep-set *is* the new manifest, so a prune-after would keep exactly what was added. Order is harmless either way here; the adds are idempotent and prune leaves survivors byte-identical. The ordering assertion in the tests is fine, just not load-bearing for the stated reason.

---

### Re-verified through the `mj app` commands themselves, against the live database

The checks above drove the exported engine functions from scripts. Repeated through the **actual CLI** — both branches built in their own worktrees (each `@memberjunction/open-app-engine` symlink repointed at that worktree's package, `tsc` + `tsc-alias`), run against the real `MJ_Local` database.

`mj app list` reports the same 9 installed apps on both branches. `mj app check-updates`:

```
$ mj app check-updates            # next
Updates available:
  connector-sharepoint: 1.2.0 -> @memberjunction/connector-wild-apricot@1.3.0
  connector-hubspot:    1.1.0 -> @memberjunction/connector-wild-apricot@1.3.0
  connector-orcid:      1.1.2 -> @memberjunction/connector-wild-apricot@1.3.0
  connector-pheedloop:  1.4.0 -> @memberjunction/connector-wild-apricot@1.3.0
  connector-netsuite:   1.3.0 -> @memberjunction/connector-wild-apricot@1.3.0
  connector-growthzone: 1.3.0 -> @memberjunction/connector-wild-apricot@1.3.0
  connector-openwater:  1.3.0 -> @memberjunction/connector-wild-apricot@1.3.0
  connector-propfuel:   1.2.0 -> @memberjunction/connector-wild-apricot@1.3.0
  connector-nimble-ams: 1.3.0 -> @memberjunction/connector-wild-apricot@1.3.0

Run mj app upgrade <name> to upgrade.
```

```
$ mj app check-updates            # this branch
All apps are up to date.
```

Nine apps, nine wrong answers — every one of them offering to "upgrade" to a **different app's** tag, which is what `mj app upgrade <name>` would then act on. Attribution is unambiguous: the PR adds `packages/MJCLI/src/utils/update-check.ts`, and only the branch build has `dist/utils/update-check.js`.

### What the CLI could NOT verify here

`mj app install` invokes `npm install` in the repo root, and npm cannot reach the registry in this environment (`spawnSync npm ETIMEDOUT`). The install still runs its migration and config-write phases and correctly reports the app as left **DISABLED** with a recovery instruction, so the command does reach the config writer — but the failing `npm install` rewrites `node_modules` mid-run, which makes it impossible to attribute the resulting config to a specific engine build. Defects 1–3 therefore rest on the function-level evidence above, which drives the same `config-manager` entry points the install path calls, against a byte-copy of the repo's real `mj.config.cjs`.


MS-BC:
## Follow-up review — a fifth gap of the same family, found and fixed in `56f8a23`

Re-reviewed this branch adversarially rather than re-running what the body already documents, and found one more silent failure — inside the fix for defect 2 itself.

`PruneDynamicPackagesNotInManifest` kept an entry whenever its `PackageName` was still in the new manifest, without looking at what the entry pointed **at**. When a version **renames its startup export**, the package stays in the keep-set, so the block survives byte-identical — and the follow-up `AddServerDynamicPackages` is then skipped by the `(PackageName, AppName)` idempotency check. The old export name stays in `mj.config.cjs` forever, so "make the config match the new manifest exactly" did not hold for this case.

What that costs at runtime is exactly the theme of this PR. [`ServerBootstrap/src/index.ts:159`](packages/ServerBootstrap/src/index.ts#L159) reads `mod[entry.StartupExport]`, gets `undefined` for a name the new version no longer exports, skips it because it is not a function — and [line 174](packages/ServerBootstrap/src/index.ts#L174) still logs `(ran <old name>)`. The registrations that export exists to perform never happen, and nothing goes red.

**Reproduced against a byte-copy of this repo's real `mj.config.cjs`**, driving the *built* `config-manager` through install v1 → upgrade v2 where only the `startupExport` changed:

```
before   install v1 : { PackageName: "@demo/server", StartupExport: "LoadDemoV1", ... }
before   upgrade v2 : { PackageName: "@demo/server", StartupExport: "LoadDemoV1", ... }   <- stale, silently dead
after    upgrade v2 : { PackageName: "@demo/server", StartupExport: "LoadDemoV2", ... }
```

### The fix

The keep-set is now `PackageName -> expected StartupExport`, and a surviving entry whose export differs is retargeted **in place** by a new `RetargetStartupExport`. Editing only that one value preserves the byte-identical property that protects an operator's `Enabled: false` — remove-and-re-add would resurrect it as `Enabled: true`, which is the same trap the existing prune logic was written to avoid. A manifest that declares no export for the package leaves the block alone rather than guessing, and the entry-removal path already covers a server package that genuinely stopped declaring one. Both `String.replace` calls use the function form so a `$` in an export name is written literally rather than read as a replacement pattern.

### Adversarial properties, all re-checked on the real config

| | |
|---|---|
| `Enabled: false` survives a retarget | PASS |
| a sibling app registering the **same** package name is untouched | PASS |
| a `$&`-bearing export name is written literally | PASS |
| the rewritten config still `require`s and exposes the entry | PASS |
| a package dropped entirely is still removed (defect 2 intact) | PASS |
| every unrelated config byte is unchanged | PASS |

**Unit:** `config-manager-scoping` 12/12 (+3), `OpenApp/Engine` **332/332** (was 329). Falsification: with the retarget disabled, exactly the 2 new behavioral tests fail; the third is a non-interference guard and correctly passes either way.

### Adjacent finding, deliberately not changed here

The `(ran <name>)` log above is emitted whether or not the export was found, so a missing export is invisible even when the config is correct — reachable now only through a hand-edited config, since the config itself converges. That is in `@memberjunction/server-bootstrap`, not in this PR's scope, and is left for a separate change rather than widening this one.

MS-BC:

MS-BC:
## Live proof — every fix exercised against real inputs

Built the branch and drove the shipped `dist/` against a real `mj.config.cjs` on disk and the live GitHub API. No mocks.

### 1. `shared` package reaches BOTH arrays

```
v1 install: one SHARED package       server=["@acme/shared-app"] client=["@acme/shared-app"]
```

Before the scoping fix the whole-file idempotency match saw the server entry written moments earlier and skipped the client insert, so `dynamicPackages.client` stayed empty and the package's `@RegisterClass` components were tree-shaken out of the browser bundle silently.

### 2. Upgrade prunes what the vendor dropped

```
v2 upgrade: shared dropped by vendor  server=["@acme/server-only"] client=[]
```

Add-only behavior converged on the union of every version ever installed and kept bootstrapping `@acme/shared-app` after v2 removed it.

### 3. Renamed `startupExport` retargeted in place

```
v3 upgrade: startupExport renamed -> StartupExport: "LoadSrvRenamed"
```

Keying only on package name left the stale export name forever; `ServerBootstrap` then read `mod[StartupExport]`, got `undefined`, skipped it as not-a-function, and still logged `(ran <old name>)`.

### 4. Operator intent and other apps survive a prune

```
operator Enabled:false survived prune+add        : true
other app's client entry untouched (per-array) : true
```

### 5. `ClassFactory.Register` warns only on a genuine collision

```
warnings after related (subclass) registration : 0  (expected 0)
warnings after UNRELATED registration          : 1  (expected 1)
message names the prior registration           : true
resolution behavior unchanged (last wins)      : Cat
```

An inheritance override stays silent; two unrelated classes on the same `(base, key)` now warn and name the prior registration. Registration semantics are untouched — the warning is diagnostic only.

### 6. `check-updates` — live against api.github.com

Two apps in the **same** multi-app repo plus one that throws:

```
updates : [{"Name":"HubSpot","Current":"1.0.0","Latest":"1.1.2"},
           {"Name":"MagnetMail","Current":"1.0.0","Latest":"3.0.3"}]
failures: [{"Name":"Broken","Message":"repo not found"}]
```

Each app resolves its **own** scoped version rather than a sibling's, and the throwing app is collected as a failure instead of killing the sweep or vanishing from a report that still concluded "All apps are up to date".

### Suites

`packages/OpenApp/Engine` 332/332 · `packages/MJGlobal` 601/601 — all pass.

CI red on this PR is [pre-existing on `next`](https://github.com/MemberJunction/MJ/pull/3352#issuecomment-5182114108); fix opened as #3439.

MS-BC:
## Adversarial re-review against live infrastructure — three findings, all fixed in this branch

Re-reviewed by driving both branches against a **throwaway restore of a real MJ database**, the **live GitHub API**, and **this repo's own `mj.config.cjs`**, rather than reading the diff. Every produced config was loaded with `require()` so the assertions are on what Node actually parses, not on the string.

### Confirmed live: the three original defects are real

**Defect 1** — real `AddServerDynamicPackages` → `AddClientDynamicPackages` against the repo's actual config:
```
next  →  server=[@a/shared]  client=[]          <- shared package never reaches the client bundle
PR    →  server=[@a/shared]  client=[@a/shared]
```

**Defect 4** — nine real apps read out of the throwaway DB, resolved against github.com with the built engine:
```
next (no subpath) → all 9 resolve to @memberjunction/connector-nimble-ams@1.3.2
PR   (subpath)    → all 9 resolve to null
```
On `next` every app is offered an upgrade to a *different app's* tag — and that string is a full scoped tag name, not a version, which is what `mj app upgrade` would then act on.

**Defect 3** — measured over a realistic MJAPI load (14 packages): **1,318 registrations across 697 `(base, key)` groups, 3 warnings, all one pair, no false positives.** The noise concern doesn't materialise.

---

### Finding 1 (P2) — the fix traded wrong-and-loud for wrong-and-silent. Fixed in 6742b0e.

`ListGitHubTags` reads a single page of the tags API. `MemberJunction/Integrations` has **374 tags**, and I confirmed the `Platform-ORCID@*` line lives entirely on page 2:
```
page 1: 0     page 2: 6     page 3: 0     page 4: 0
```
So the scoped lookup returns `null` for all nine apps, `null` was folded into "no update available", and the command printed a confident `All apps are up to date` while e.g. `connector-orcid` sits at `1.1.2` with `Platform-ORCID@1.2.0` published. A green line nobody investigates is worse than an absurd one they do.

`CheckAppsForUpdates` now returns **`Unresolved`** as a third outcome, distinct from both an update and a failure, and the green line prints only when every app produced an answer. Every app in that list is *installed*, so it resolved from a real ref once — finding nothing now means the resolver and the repository disagree. The test that codified the old behaviour is replaced by five that pin the distinction.

**This does not remove the need for #3353** — it makes merging this alone *honest* rather than falsely green, not *correct*. #3353 should land with or before it.

### Finding 2 (P3) — prune-first left the config matching neither version. Fixed in 6742b0e.

Prune and add are two writes to the same files with no rollback between them, so a failure in the second is observable. Verified on the real config that both orders converge identically, and that the failure windows do not:
```
CONVERGE IDENTICALLY: true      (incl. the renamed startupExport being retargeted)

if the 2nd write fails, app-a's server entries hold:
  add-first (new):    ['@a/server', '@a/server2', '@a/shared']   <- superset, server still boots
  prune-first (old):  ['@a/shared']                              <- missing @a/server2
  v2 needs:           ['@a/server2', '@a/shared']
```
Now adds before pruning. The two ordering tests are rewritten and a third asserts the prune does *not* run when the add fails.

### Finding 3 (P1) — the new diagnostic found a live permission bypass. Fixed in f3c83df.

The one pair that warned is a real bug. `MJConversationDetailEntityServer` and `MJConversationDetailEntityExtended` both register for `BaseEntity` under `'MJ: Conversation Details'` as **siblings**. The server package loads last, so it wins outright:
```
WINNER: MJConversationDetailEntityServer
  permission gate (currentUserMayWrite)?  false
  edit-flag logic?                        true
```
The gate that never ran is the check that only a conversation's owner may set `UserRating`/`UserFeedback`, and that a non-owner without a resource grant cannot write at all — explicitly written to run server-side (`ProviderType === 'Database'`), which is exactly where it was shadowed out.

`MJConversationDetailEntityServer` now extends the Extended class so the two compose. Verified:
```
WINNER: MJConversationDetailEntityServer        <- unchanged, no behavioural regression
chain:  Server -> Extended -> MJConversationDetailEntity -> BaseEntity
gate reachable via inheritance? true
AreClassesRelated(Extended, Server)? true       <- warning now silent
```
This is the one separable commit if you'd rather split it out.

---

### Suites

`OpenApp/Engine` **347/347**, `MJCLI` `update-check` **15/15** (was 11), `MJGlobal` **641/641**, `MJCoreEntitiesServer` typechecks clean. Note the engine suite is now fully green — the 9 `entity-teardown` failures in the description above were environmental and cleared once the branch merged `next`.

### What could not be exercised

`mj app install` / `upgrade` end-to-end: npm can't reach the registry here, and `npm install` rewrites `node_modules` mid-run, so a resulting config can't be attributed to a specific engine build. I drove the same `config-manager` entry points the install path calls. The ClassFactory measurement is a replay of the PR's predicate over a real captured registration list rather than the PR's `ClassFactory` installed in-process (`MJGlobal` is a singleton resolved from `node_modules`), and it can't distinguish auto-incremented from explicit priorities — so 3 is an upper bound. No browser bundle was built, so defect 1's tree-shaking consequence is inferred from the config contents.

The throwaway database and its backup were dropped; the source database was only ever read via a `COPY_ONLY` backup.

rkihm-BC (approved):
Approving. The verification standard here is well above the bar — reproducing each defect against a byte-copy of the real `mj.config.cjs`, a live database, and both branches side by side, then running the new `config-manager` tests against pre-fix source to confirm they actually fail there. I found no correctness defects in the diff.

Two calls in particular are worth naming, because both are cases where the obvious fix would have been worse than the bug:

- **`Unresolved` as a third outcome.** Scoping the lookup by `Subpath` fixed the sibling-version bug but would have turned nine loudly-wrong answers into a confident `All apps are up to date`. Catching that in self-review and refusing to fold a null into the green line is the right instinct — a false green nobody investigates is worse than an absurd answer they do.
- **Add-then-prune, justified by the failure window rather than by convergence.** Two writes with no rollback between them are not order-symmetric, and leaving the window at `(old ∪ new)` keeps a mid-upgrade server bootable. The follow-up commit correcting the original prune-first rationale is the right outcome.

I verified the schema-rename path reads `existingApp.ManifestJSON` at Step 7 before `UpdateAppRecord` overwrites it at Step 9, including on resume — that ordering holds.

### One request before merge

**A regression test for the `MJConversationDetailEntityServer` fix.**

This is the one change in the PR with a security consequence, and it's the only one without a test. `MJConversationDetailEntityServer.test.ts` covers `ShouldFlagOriginalMessageChanged` only, and this PR doesn't touch it — so nothing asserts that the server class actually composes with the permission gate rather than shadowing it.

The failure mode is cheap to reintroduce: changing `extends MJConversationDetailEntityExtended` back to the generated entity is a one-token edit, TypeScript won't flag it, and the only runtime signal is the new `console.warn` — which nothing reads in CI. Given the gate in question is what stops a non-owner without a resource grant from writing conversation messages, and what keeps `UserRating`/`UserFeedback` owner-only, it deserves a guard.

A few lines is enough — asserting the server class is an instance of `MJConversationDetailEntityExtended`, and that `AreClassesRelated` reports the two as related so the collision warning stays silent, would pin both halves of the fix.

### Notes, not blockers

- `adopted standards` is red on three `ui-layers` violations in `packages/Angular/Generic/graph-view/` — a package this PR doesn't touch. Inherited from `next`, not from here. `Run unit tests` was still in flight when I looked; worth a glance before you merge.
- The defect-2 section of the PR description still says the prune runs *before* the adds and calls that ordering load-bearing. The changeset is correct and the description carries its own correction further down, so this is cosmetic — flagging only so it doesn't mislead a later reader.
- `AreClassesRelated` returns true whenever the two class names match, so two genuinely unrelated classes sharing a name won't warn. The trade-off is right (false positives on an always-on diagnostic are worse), but it's worth a line in the JSDoc since that's arguably the two-independent-apps case the warning exists for.

#3353 merged on the 15th, so the pagination dependency called out in the changeset is satisfied.

PR #3353, opened 2026-07-29
fix(open-app-engine): correct semver ordering, paginate GitHub listings, clean up temp dirs

MS-BC:
Four defects in the Open App engine's GitHub/version layer and its temp-file handling. All are on `next` today; each fix has tests that were verified to fail against the pre-fix source.

**1. `compareSemver` returned `NaN` for any prerelease, poisoning the sort.** It ran `Number()` across the dot-split version string, so `'1.2.0-beta.1'` parsed as `[1, 2, NaN, 1]`. `NaN !== 0` is `true`, so the comparator returned `NaN` and `Array.prototype.sort` ordering became implementation-defined for any tag list containing a prerelease. `ListGitHubTags`'s own regex deliberately matches prerelease tags, so this is reachable in normal use — and `GetLatestVersion` takes `tags[0]`, meaning it could report an arbitrary tag as the newest version available for upgrade. Replaced with a spec-correct `CompareSemver`: numeric release triple, then semver prerelease precedence (a prerelease ranks below the same release without one; identifiers compare numerically when numeric and ASCII-lexically otherwise; numeric ranks below alphanumeric; fewer identifiers rank lower), ignoring a leading `v` and build metadata.

**2. `GetLatestVersion`'s tag path had no stable-vs-prerelease preference.** The releases path already filters `!PreRelease && !Draft`, but the tag fallback returned whatever sorted first — so a beta tag could be offered to an installed app as its upgrade target. The tag path now prefers the highest stable version and falls back to a prerelease only when nothing stable is tagged.

**3. `ListGitHubReleases` / `ListGitHubTags` were unpaginated.** Both passed `per_page: 100` and read a single page, silently truncating. GitHub returns tags in its own order rather than semver order, so in a repo that tags many apps the newest version can sit past the first page and be invisible. Both now use `octokit.paginate` (the plugin is bundled in `@octokit/rest`).

**4. Two temp download directories were never cleaned up.** `HandleMigrations` and `HandleTeardown` each `mkdirSync` a dir under `tmpdir()`, download `.sql` files into it, and return without removing it — so every install, upgrade, and remove leaves a copy of that app's migrations in the OS temp dir. Both now clean up in a `finally` via a shared best-effort `CleanupTempDir` helper (a cleanup failure warns via `OnWarn` and never masks or fails the operation that created the dir), matching what the application-ID probe path at the bottom of the file already did.

## Testing
- 21 new cases; `packages/OpenApp/Engine` suite is **335 passing / 25 files**, build clean.
- Every new behavioral case was checked against the pre-fix source: 9 fail on the old comparator + unpaginated listings, 2 fail without the stable-tag preference, 4 fail without the temp-dir cleanup.
- The existing `github-client.test.ts` Octokit stub gained a `paginate` stand-in that follows pages and stops on the first short page, so pre-existing single-page tests behave exactly as before.

## Notes
- No public signature changes. `CompareSemver` and `IsPrereleaseVersion` are newly exported from the package index as reusable utilities.
- No version bumps, no migrations.

---

## Reviewed and live-tested

Both branches built in their own worktrees, run against the **real GitHub API** and, for the end-to-end check, the live `MJ_Local` database.

### Defect 1 — differential test against the reference `semver` package

`CompareSemver` was compared with `semver.compare` over a corpus of 120 valid versions (8 release triples × 15 prerelease shapes, including `-0`, `-10`, `-0.1`, `-x.7.z.92`):

```
compared 14400 ordered pairs
mismatches vs reference semver: 0
descending sort of 120 versions identical to semver.rcompare: YES
IsPrereleaseVersion disagreements: 0

pre-fix comparator returned NaN on 2240 of 14400 pairs
pre-fix descending sort matched semver: NO
```

The new comparator agrees with `semver` on every pair. The old one returned `NaN` on **15.6%** of pairs and produced a wrong ordering.

### Defect 3 — pagination, against the live GitHub API

`MemberJunction/Integrations` has **344 tags — 4 pages** at `per_page=100`. Same calls, same token, both branches:

|  | `next` | this branch |
|---|---|---|
| `ListGitHubReleases` | **100** (exactly the page cap) | **144** (all) |
| `ListGitHubTags(subpath: Platform/ORCID)` | **0** | **5**, newest `1.1.3` |
| `GetLatestVersion(Platform/ORCID)` | `null` | `1.1.3` |
| `GetLatestVersion(CRM/HubSpot)` | `null` | `1.1.2` |
| `GetLatestVersion(AMS/NimbleAMS)` | `null` | `1.3.1` |

This is worse on `next` than "a version might be missed": `Platform-ORCID@1.1.3` is tag **169 of 344**, so it sits on page 2 and is invisible. Every app in that repo resolves to `null` — no installed app can detect any update at all. The releases number landing exactly on 100 is the truncation itself.

### End to end, against the live database

`mj app check-updates` over the 9 apps installed in `MJ_Local`, with this branch's engine under #3352's CLI:

```
Updates available:
  connector-hubspot:     1.1.0 -> 1.1.2
  connector-orcid:       1.1.2 -> 1.1.3
  connector-pheedloop:   1.4.0 -> 1.4.1
  connector-nimble-ams:  1.3.0 -> 1.3.1
```

Real versions, correctly resolved. For comparison: `next` reports nine updates that all point at *a different app's* tag, and #3352 alone reports "All apps are up to date" — correct-by-silence, because the scoped tags it looks for are past page 1. It takes both PRs to get a true answer.

### Defect 4 — temp-dir leak, measured

Counting `mj-app-*` directories in `$TMPDIR` around a full suite run:

```
next   108 -> 114   (leaked 6)
branch 114 -> 114   (leaked 0)
```

The leak is not hypothetical on this machine: `$TMPDIR` already held **108** such directories, including 4 from real `mj app install connector-orcid` runs and 17 from `connector-hubspot`, each still holding that app's `.sql` migration files.

### Suite

`packages/OpenApp/Engine`: **326 passed / 9 failed**. The 9 are `entity-teardown-*`, which fail **identically on `next`** in this environment (stale `@memberjunction/core` in the borrowed dependency tree lacks `SQLDialect.ForeignKeyGraphSQL` / `.AtomicBatchScript`) — `next` scores 309 passed / 9 failed. No failure comes from this branch.

### One observation

No changeset. `@memberjunction/open-app-engine` is a published package at `5.49.0`, so without one these fixes never reach a release — and the pagination fix in particular is what makes update detection work at all for the Integrations repo. Worth adding a patch changeset before merge.


MS-BC:
## Follow-up review — the releases path had the same ordering gap, fixed in `106444d`

Re-reviewed this branch rather than re-running what the body documents. Defect 2 fixed the **tag** path's ordering preference, but `GetLatestVersion`'s **releases** path was left on `releases.find(r => !r.PreRelease && !r.Draft)` — the first entry in GitHub's own order, which is newest-**created**, not newest-**version**. A patch backported to an older line after a major has shipped is the most recent release and the lower version, so it would be offered as the upgrade target to an app already past it.

This is not theoretical ordering pedantry on real repos: `nodejs/node`'s stable release list interleaves the v26, v24 and v22 lines, and its two most recent stable releases (`v26.6.0` and `v24.19.0`) were published **21 seconds apart**. Which one `find` returns is decided by publish order alone.

### The fix, and a flaw the live run caught in my first attempt

Sorting by semver precedence is right — but only across tag names that really **are** repo-wide versions. Running the comparator over scoped release names is not neutral: `ParseSemver` reads the `-` inside `@memberjunction/connector-wild-apricot@1.3.0` as the prerelease delimiter, so sorting reshuffles meaningless values into a *different* meaningless answer. The live run against `MemberJunction/Integrations` caught exactly that — the first version of this change moved the result from `connector-wild-apricot@1.3.0` to `connector-zendesk@1.1.2`.

A new `IsPlainVersionTag` now gates the sort; anything that is not a repo-wide `v1.2.3` keeps GitHub's original order, identical to `next`. (That path is only reachable at all when no scoped subpath is in play — line 389 already routes scoped apps straight to the tag line.)

### Live A/B against the real GitHub API — `next` vs this branch

1,170 stable releases across 4 repos, same token, both answers computed from the same fetched data:

| repo | stable releases | `next` | this branch | same |
|---|---|---|---|---|
| `nodejs/node` | 511 | `26.6.0` | `26.6.0` | YES |
| `vercel/next.js` | 365 | `16.3.0` | `16.3.0` | YES |
| `microsoft/TypeScript` | 150 | `6.0.3` | `6.0.3` | YES |
| `MemberJunction/Integrations` | 144 | `connector-wild-apricot@1.3.0` | `connector-wild-apricot@1.3.0` | YES |

No real repo changes answer — the change is confined to the ordering hazard it targets, which is what a fix in this area should look like.

**Unit:** `github-client` 46/46 (+3), `OpenApp/Engine` **338/338** (was 335). Falsification: reverting to `find` fails exactly the 2 new behavioral tests; the scoped-name guard test passes either way, as a non-interference check should.

Worth noting the 9 `entity-teardown-*` failures the body reports as environmental are now **passing** — the suite is fully green in this worktree.

### Changeset added

The body's own closing observation — no changeset on a published package at `5.49.0` — is now addressed: `.changeset/openapp-engine-version-resolution.md` declares a patch for `@memberjunction/open-app-engine` covering all five fixes. Without it the pagination fix in particular never reaches a release, and that one is what makes update detection work at all for the Integrations repo.

MS-BC:

MS-BC:
## Live proof against the real GitHub API

Re-verified the built branch (`dist/`) against api.github.com, not mocks.

### Pagination — the fix is load-bearing, with a concrete victim

`MemberJunction/Integrations` carries **344 tags**. GitHub returns them in its own order, not semver order. Positions of the scoped tags for two real apps in that order:

```
199: Marketing-MagnetMail@3.0.3     272: CRM-HubSpot@1.1.2
200: Marketing-MagnetMail@3.0.2     273: CRM-HubSpot@1.1.1
201: Marketing-MagnetMail@3.0.1     274: CRM-HubSpot@1.1.0
202: Marketing-MagnetMail@3.0.0     275: CRM-HubSpot@1.0.0
203: Marketing-MagnetMail@1.1.0
204: Marketing-MagnetMail@1.0.0
```

**Every tag for both apps sits past page 1.** With the old unpaginated `per_page: 100`, neither app had a single tag visible, so `GetLatestVersion` returned `null` — "no version found" for an app that has six. With `octokit.paginate`:

```
CRM/HubSpot            tags= 4   latest= 1.1.2
Marketing/MagnetMail   tags= 6   latest= 3.0.3
```

### Pagination on a repo-wide-tag repo

`MemberJunction/MJ` has 290 tags, 289 matching `v<semver>`:

```
ListGitHubTags (paginated): 289 tags
highest 5 by semver: v5.51.0, v5.50.0, v5.49.0, v5.48.0, v5.47.0
GetLatestVersion -> 5.51.0
```

The first page alone yields exactly 100 — the old ceiling.

### Comparator

```
sort(CompareSemver): 1.2.0-beta.1 < 1.2.0 < 1.9.0 < 1.10.0 < 2.0.0-rc.1 < 2.0.0
IsPrereleaseVersion: 1.0.0=false  1.0.0-beta.1=true  2.0.0-rc.1=true
```

`1.9.0 < 1.10.0` (not string order) and prerelease-before-release both hold. The old `Number()` over the dot-split string produced `NaN` for every prerelease, making `Array.prototype.sort` ordering implementation-defined for any tag list containing one.

### Suite

`packages/OpenApp/Engine` — **338/338 pass**.

MS-BC:
## Adversarial review against live infrastructure — two findings, both fixed here

Reviewed by driving both branches against the **live GitHub API**, a **throwaway restore of a real MJ database**, and the reference `semver` package. Also merged `origin/next` (`380ac1a`) — the branch's merge-base was from 2026-08-05, which is why CI was red (below).

### What held up

**The comparator is correct.** I ran a differential against `semver@7.7.4` with a harsher corpus than the PR's — build metadata, uppercase and hyphenated identifiers, and inputs MJ's tag regex admits but semver rejects:

```
corpus A (valid, 153 versions):  23409 pairs, 0 mismatches
corpus B (build metadata):          36 pairs, 0 mismatches
descending sort identical to semver.rcompare: YES
NaN across 27,556 pairs (incl. hostile input): 0
transitivity violations in 60^3 triples: 0
```

**Pagination genuinely fixes resolution** — this is the PR that makes update detection work at all against Integrations.

Suspicions I chased and **discarded**: path traversal into the new `rmSync` via `manifest.name` (blocked by `appNameRegex`, no dots or slashes); cleanup deleting files still in use (the app-ID probe uses its own dir, migrations are awaited before the `finally`); rate-limit handling lost under pagination (preserved on both paths); conflict with #3352 (none — merged tree passes 367/367).

---

### Finding 1 (P2) — pagination is per-repo but was paid per-app. Fixed in f8d2990.

The fetch is per-REPOSITORY while tag filtering is per-app, so a sweep over apps sharing one repo paid for the whole paginated walk once per app. Measured by counting real HTTP requests for a 9-app `check-updates` over Integrations (374 tags, 4 pages):

```
next                 9 requests   2725 ms
this branch before  36 requests   9292 ms
this branch after    4 requests   1091 ms      <- versions resolved identically
```

It also grows with the repo's tag count on every release: at 2,000 tags that was 20 pages x 9 apps = 180 sequential requests for one command. Unauthenticated GitHub is 60 req/hr.

`ListGitHubTags` now reuses a recently fetched tag list per repository. Deliberately narrow: **60s TTL** so it collapses one sweep rather than acting as a durable cache; **keyed by the resolved token**, so a private repo's tags are never served to a caller who did not supply that token; **failures are never cached**, so a rate-limited call still surfaces via `ThrowIfRateLimitedOrForbidden` on the next attempt. `ClearGitHubTagCache` is exported.

### Finding 2 (P3) — a release NAME was being returned as a version. Fixed in f8d2990.

When no release tag is a plain version, the code returned `stableReleases[0]` — GitHub's own first entry. Verified live:

```
next               GetLatestVersion(repo, opts) -> @memberjunction/connector-nimble-ams@1.3.2
this branch before                              -> @memberjunction/connector-nimble-ams@1.3.2
this branch after                               -> null
```

That string can never equal an installed version, so it reads as a permanent "update available" pointing at a target `mj app upgrade` would act on. The PR preserved it to "match `next` exactly" — but matching `next` is not a virtue when `next`'s answer is known garbage and `null` is one line away. Returning `null` isn't a behaviour loss: it falls through to the tag path, which for a repo-wide app matches only `v?<semver>` and correctly resolves to `null`.

The test asserting the old string is replaced by one asserting `null`, plus a new case proving a genuine repo-wide release still wins when the release list **mixes** both shapes.

### Finding 3 — CI was red for a reason that is not this PR. Addressed by merging `next`.

`Run unit tests` failed on two jsdom-serialization assertions in `@memberjunction/ng-shared-generic` (`url("#…")` vs `url(#…)`). This PR touches **only** `packages/OpenApp/Engine`, and that package has **no dependency** on `open-app-engine`; all 26 of its tests pass locally on the branch's exact tree. The branch's merge-base was six days stale, so `next` is now merged and CI should re-run clean.

### Note on test hygiene

The tag cache is module state that outlives a test, so `github-client.test.ts` now clears it in `beforeEach`. Without that, a test stubbing `listTags` could be served the previous test's list and pass for the wrong reason. I verified the new memoization test genuinely fails when cache reuse is disabled — it is a guard, not decoration.

### Suites

`packages/OpenApp/Engine` **357/357**, typecheck clean. Merged with #3352: **367/367**, no conflicts.

### End to end, both PRs, against a throwaway restore of a real MJ database

```
3352 alone   No updates found, but some apps could not be checked:
             Unresolved (9): no 'Platform/ORCID@<version>' tags found ...

3352 + 3353  Updates available:
               connector-hubspot     1.1.0 -> 1.1.2
               connector-orcid       1.1.2 -> 1.2.0
               connector-pheedloop   1.4.0 -> 1.4.1
               connector-openwater   1.3.0 -> 1.3.2
               connector-nimble-ams  1.3.0 -> 1.3.2
```

It takes both PRs to get a true answer. The throwaway database and its backup were dropped; the source database was only ever read via a `COPY_ONLY` backup.

rkihm-BC (commented):
## Code review — Claude Code

> Performed by **Claude Code** (Opus 5) at @rkihm-BC's request. I read the full diff, the changeset, and all six comments, then re-derived the load-bearing claims on a local checkout of the branch rather than taking them on trust — including a fresh differential against the reference `semver` package and four targeted probes. **Please push back on anything I got wrong or missed.** Where I could not verify something, I say so.

**Verdict: strong work, approve with changes.** Five real defects, each with a falsification check, live evidence, and a genuinely good changeset. Nothing here is wrong in a way that makes the PR worse than `next`. But there are **two correctness edge cases I reproduced that are the same class as the bugs this PR fixes**, one performance fix that only works by accident of how the caller loops, and one design question about ~60 lines of hand-rolled semver that a dependency already in this package provides.

---

I applied the branch's six files to a local tree and ran it:

| Check | Result |
|---|---|
| Typecheck + suite | ✅ `tsc --noEmit` clean; **357/357 pass across 26 files** — exactly the numbers in the last comment. Tree restored afterwards. |
| Comparator correctness | ✅ My own differential vs `semver@7.7.x`, corpus of 120 valid versions (incl. `-0`, `-10`, `-alpha.beta`, `-x.7.z.92`, `-alpha.0valid`): **14,400 pairs, 0 mismatches, 0 NaN, 0 transitivity violations** in 64,000 triples. Independently corroborates the claim. |
| `octokit.paginate` really is bundled | ✅ `typeof octokit.paginate === 'function'` on the installed `@octokit/rest`. |
| The `paginate` test stub is faithful | ✅ "Stop on the first short page" is the right mock equivalent of Link-header following, and it means every pre-existing single-page test keeps its old meaning. Good call, and easy to get wrong. |
| Cache key is genuinely complete | ✅ `GitHubClientOptions` carries only `Token`/`TokenMap` — no `baseUrl` — so `owner/repo + resolvedToken` really is the full identity of a fetch. The token-isolation test is not theatre. |
| Path traversal into the new `rmSync` | ✅ Your dismissal is correct: `appNameRegex = /^[a-z][a-z0-9-]{1,62}[a-z0-9]$/` (`manifest-schema.ts:26`) admits no dots or slashes. |
| Temp-dir coverage is complete | ✅ All three `tmpdir()` sites in `install-orchestrator.ts` (1657, 1724, 2217) now clean up. No fourth leak anywhere in the package. |
| The `null`-instead-of-garbage change is safe at both consumers | ✅ `install-orchestrator.ts:709` turns it into a clean `Could not determine target version` failure; `check-updates.ts:40`'s `if (latest && latest !== app.Version)` skips it. No caller mishandles `null`. |
| Changeset | ✅ Present, patch-scoped, and unusually good — it explains *why* for all five fixes rather than restating the title. |

CI at time of writing: 4 green, unit tests still running.

---

## What's concerning

### 1. Build metadata leaks into the returned version — reproduced

`IsPlainVersionTag` accepts `v1.2.3+build.7`, and the releases path returns the tag verbatim after stripping only the leading `v`:

```
PROBE: GetLatestVersion(release tagged 'v1.2.3+build.7')  ->  "1.2.3+build.7"
```

That string can never equal an installed `1.2.3`, so it reads as a permanent "update available" pointing at a target `mj app upgrade` would act on — **precisely the Finding 2 you just fixed, one edge case over.** Note the asymmetry: `ListGitHubTags`'s regex refuses `+` outright, so the tag path cannot produce this; only the releases path can. (Aside: the same regex accepts `1.2.3+build` but rejects `1.2.3-rc.1+build`, because `+` isn't in the trailing character class — so a tag carrying both is silently excluded from the sort.)

### 2. The two paths guard prereleases by different criteria — reproduced

Defect 2's fix guards the **tag** path on the version *string* (`IsPrereleaseVersion`). The releases path still guards only on GitHub's `prerelease` **boolean**. When a maintainer tags `v2.1.0-rc.1` and forgets the checkbox, the two paths disagree:

```
PROBE: releases path -> "2.1.0-rc.1"   (the tag path, given the same versions, gives "2.0.0")
```

A prerelease is offered as the upgrade target for an installed app — the exact outcome defect 2 exists to prevent. One-line fix: apply `IsPrereleaseVersion` to `versioned` too, with the same "fall back only if nothing stable" rule the tag path uses.

### 3. The tag memoization works only because the caller happens to loop sequentially — reproduced

`check-updates.ts:38` is a `for…of` with `await`, so the 36 → 4 measurement is real. But the cache stores the *resolved value*, not the in-flight promise, so a parallel sweep gets no benefit at all:

```
PROBE: 9 apps, one repo, 2 pages of tags
  sequential -> 2 HTTP calls
  parallel   -> 18 HTTP calls
```

Finding 1's win is therefore contingent on a loop shape in a package this one doesn't own — a `Promise.all` refactor in MJCLI would silently revert it, with no failing test. Caching the pending promise (`Map<string, Promise<string[]>>`, delete on rejection) is roughly one line and makes the fix hold either way.

### 4. `ListGitHubReleases` got the pagination but not the memoization — and has no page cap

Finding 1 solved "paid per app" for tags. The releases path is now *also* fully paginated, uncached, and unbounded: a single-app repo with 2,000 releases costs 20 sequential requests **per `GetLatestVersion` call**, against an unauthenticated budget of 60/hour — the same failure mode Finding 1 was written to prevent. `ResolveVersionRange` (`install-orchestrator.ts:1545`) calls it too, once per version-range resolution. Either give it the same memo, or cap pages with a warning when the cap is hit. (Silent truncation is what defect 3 was about, so a cap must be loud.)

### 5. ~60 lines of hand-rolled semver, when `semver` is already a dependency of this package

`@memberjunction/open-app-engine` already declares `"semver": "^7.7.1"` and already imports it in `install-orchestrator.ts` — two files away from the new code. Everything the PR hand-writes is in that library, and I checked each one against the installed copy:

```
semver.valid('v1.2.3')                                  -> '1.2.3'    (IsPlainVersionTag + v-strip)
semver.valid('1.2.3+build.7')                           -> '1.2.3'    (…and fixes concern #1 for free)
semver.valid('@memberjunction/connector-wild-apricot@1.3.0') -> null  (the scoped-name guard)
semver.valid('CRM-HubSpot@1.1.2')                       -> null
semver.prerelease('1.2.0-beta.1')                       -> ['beta',1] (IsPrereleaseVersion)
semver.prerelease('1.2.0+sha.abc')                      -> null
[...].sort(semver.rcompare)  ->  identical to the CompareSemver descending-sort test's expectation
```

So `CompareSemver`, `ParseSemver`, `ComparePrerelease`, `ComparePrereleaseIdentifier`, `IsPlainVersionTag` and `IsPrereleaseVersion` — plus the eight tests and the 14,400-pair harness written to prove they match `semver` — collapse to `semver.rcompare` / `semver.valid` / `semver.prerelease`.

This is not a bug and the implementation is correct today. It's about where the next edge case lands: concern #1 is already one, and `semver.valid()` would have prevented it. The one genuine difference is that `CompareSemver` is **total** (junk coerces to `0`, never throws) while `semver.compare` throws on invalid input — but both call sites already know how to guard, and `install-orchestrator.ts` uses exactly that `semver.valid()` filter pattern today. If you keep the hand-rolled version, I'd want a sentence in the doc comment saying why, because the obvious reader question is "why not the library that's already imported?"

### 6. `tagListCache` only ever grows

Expired entries are never evicted — a `Map` keyed by `owner/repo + token` that is written and never swept. Bounded in practice: no server package depends on `open-app-engine` (dependents are MJCLI, integration-test-suite, and the package itself), so this lives in short-lived CLI processes. Low severity, but a size cap or a lazy sweep on write costs three lines, and the entry holding a token in its key is worth not keeping around indefinitely.

---

## What's missing

- **No test pins the 60s TTL.** It's the one behavior of the cache nothing exercises — `vi.useFakeTimers()` + advance past `TAG_CACHE_TTL_MS` would catch someone typing `60_000_000`. The security-relevant properties (token keying, no caching of failures) *are* tested, which is the right priority order; this is the gap left over.
- **No concurrency test** — see concern #3. A test asserting the call count under `Promise.all` would have surfaced it.
- **`ExtractDeclaredApplicationIds` (`install-orchestrator.ts:2226`) still has its own inline cleanup.** The PR introduces a shared `CleanupTempDir` helper and converts two of the three sites; the third keeps a hand-rolled `try { rmSync } catch { /* best-effort */ }` that swallows silently instead of warning via `OnWarn`. Same behavior, three different-looking code paths — point it at the new helper.
- **Merge-order note, not a defect:** on its own, this PR converts `check-updates`' garbage line (`1.1.0 -> @memberjunction/connector-nimble-ams@1.3.2`) into *silence* ("All apps are up to date"), because `check-updates.ts:40` simply skips a `null`. The honest answer only becomes legible to a user once #3352's "Unresolved (N)" reporting lands. You say as much ("it takes both PRs"); worth making it explicit in the merge plan so nobody lands this one alone and reads the silence as a regression.
- Nothing to update doc-wise for the three new exports — the package has no README.

---

## Nits

- `stableReleases` is only ever used to derive `versioned`; inlining the two filters would read better than naming an intermediate that isn't reused (and would remove the temptation to reach for it later, which is how the release-name bug got in).
- `FetchRepoTagNames` returns the cached array **by reference**. The only caller `.map()`s first, so nothing mutates it today — but a future caller that sorts in place would silently corrupt the cache. `return [...cached.TagNames]` is cheap insurance.
- The `ListGitHubTags` doc comment says tags are matched "optionally with a prerelease suffix" — worth noting there that the suffix may not contain hyphens and that build metadata is rejected outright, since that constraint is now load-bearing for concern #1's asymmetry.

---

I ran everything above against the branch at `f8d2990`; the branch is ~29 commits behind `origin/next` as I write. If any finding is wrong or already handled somewhere I didn't look, say so on the thread and I'll re-check rather than assume I was right.

MS-BC:
Thanks — that review was worth more than the PR it reviewed. All six concerns and all three missing items are addressed, plus one thing your tooling could not have seen. Two places I pushed back; flagged below.

## First — why your `tagListCache` grep may have felt odd

The cache key delimiter was a **literal NUL byte** in the source — a raw `0x00` sitting between the repo and the token in that template literal.

NUL is a *good* delimiter here: it cannot occur in an owner, a repo name or a token, so the halves can never be confused. Written as a raw byte, though, `file` reports the whole file as `data` and **grep treats it as binary — prints nothing, exits 1.** I hit this head-on: grepping for `tagListCache` came back empty on a file that plainly contained it, and I briefly suspected my own `next` merge had dropped the commit. Anyone searching this file was searching one their tools had quietly given up on.

It is now the ``\u0000`` escape — identical string at runtime, file is text again. Worth knowing generally, since it defeats review tooling silently rather than loudly.

## Concern 5 — took your recommendation

Adopted `semver`. Deleted `CompareSemver`, `ParseSemver`, `ComparePrerelease`, `ComparePrereleaseIdentifier` and `IsPlainVersionTag` — plus the eight tests and the 14,400-pair parity harness written to prove they matched the library. `semver.rcompare` sorts, `semver.valid` normalizes, `semver.prerelease` detects.

Your total-vs-throws point is handled where you said it should be: a single `SemverCore()` is the one normalization point, and every sort filters through it first — the same `valid()`-then-compare shape `install-orchestrator.ts` already uses. `IsPrereleaseVersion` stays total (junk is *not* a prerelease rather than a throw), so no call site needs a new guard.

**Concern 1 fell out for free**, exactly as you predicted: `semver.valid('v1.2.3+build.7')` → `1.2.3`.

## Concern 2 — both paths now guard on the string

You were right that the flag is a checkbox someone can forget. Both paths apply the version-string test with the same "fall back only if nothing stable" rule, so they can no longer disagree. Test: `prefers a stable release even when the prerelease flag was not ticked`.

## Concerns 3, 4, 6 — one shared cache

- **3:** the cache now holds the **in-flight promise**, so concurrent callers join one fetch. Your probe is now a test — and I verified it **fails, and only it, when I temporarily restore resolved-value caching**.
- **4:** `ListGitHubReleases` gets the same memo on the same key and TTL. I chose the memo over the page cap on your own reasoning: silent truncation is what defect 3 was about, so a cap would reintroduce it. Both paths now go through one `MemoizedFetch`, so the promise sharing, rejection deletion and eviction rules cannot drift apart between them.
- **6:** eviction on write — expired first, then oldest, bounded at 64. The token-in-the-key observation is why I did this rather than leaving it as acceptable-in-practice.

## Missing items — all three

- **TTL test** with fake timers across the 60s boundary (your `60_000_000` scenario).
- **Concurrency test** — see above.
- **`ExtractDeclaredApplicationIds`** now uses the shared `CleanupTempDir`, so a cleanup failure warns via `OnWarn` instead of being swallowed by the third lookalike `try { rmSync } catch {}`.

## Nits — all three

`stableReleases` is gone (inlined by the rewrite). The tag fetch returns a **copy**, with a test that one caller's in-place sort cannot corrupt what others see. The `ListGitHubTags` doc comment now states the hyphen-free-suffix and build-metadata-rejected constraints, and *why* they are load-bearing.

## Two pushbacks

1. **`ListGitHubTags` still returns tag TEXT (`v1.0.7`), not the normalized core.** My first pass normalized it — and broke two of your existing tests, correctly, because `ValidateGitHubTag` and external callers depend on that shape. Only the ordering is semver-derived. Flagging it because "normalize everything through `SemverCore`" reads like it should apply here too, and deliberately does not.

2. **The merge-order note only cuts one way now.** You flagged that landing this alone turns the garbage line into silence — still true, and worth keeping in the plan for #3352. But the honest `null` is strictly better than the string it replaced, so I do not think it should hold this PR.

## Verification

`@memberjunction/open-app-engine`: **360 passed / 26 files**. `tsc --noEmit` clean, build clean, `check:changeset` green. Branch is current with `next`.

Re-review welcome — particularly the `MemoizedFetch` generic, and whether keeping tag text is the call you would have made.

rkihm-BC (approved):
## Re-review — Claude Code

> Performed by **Claude Code** (Opus 5) at @rkihm-BC's request, as a follow-up to the 8/11 review. I applied the branch to a local tree and re-ran it rather than taking the response comment on trust; the tree was restored afterward. **Push back on anything I got wrong.**

**Verdict: approve.** All six concerns and all three missing items from the previous review are genuinely addressed — verified in the source, not just in the reply. Three leftovers below, none blocking.

---

## Verified locally, at `b399928`

| Check | Result |
|---|---|
| `tsc --noEmit` | ✅ clean |
| `packages/OpenApp/Engine` suite | ✅ **360 passed / 26 files** — exactly the claimed number |
| Branch vs `next` | ✅ 0 commits behind |
| CI at time of writing | ✅ Check migrations, dependency check, adopted standards pass; unit + integration still running |

### The six concerns, confirmed in the source

1. **Build metadata** — `SemverCore` is `semver.valid`, so `v1.2.3+build.7` → `1.2.3`. Probed directly against the installed copy. Gone, exactly as predicted.
2. **The two paths disagreeing** — both now filter on the version *string* with the same "fall back only if nothing stable" rule, so the unticked-checkbox case can no longer split them. `prefers a stable release even when the prerelease flag was not ticked` pins it.
3. **Sequential-only memo** — the cache now holds the **in-flight promise**, published before anything awaits it. My probe is now a test asserting 2 calls under `Promise.all` rather than 18, and you verified it fails *and only it* when resolved-value caching is restored. That is the right falsification.
4. **`ListGitHubReleases`** — same `MemoizedFetch`, same key, same TTL. Choosing the memo over a page cap is the correct call, for the reason you give: a cap reintroduces the silent truncation defect 3 existed to remove.
5. **Hand-rolled semver** — ~60 lines plus the 14,400-pair parity harness deleted. `SemverCore` as the single normalization point puts the total-vs-throws boundary exactly where it belongs, and `IsPrereleaseVersion` staying total means no call site needs a new guard.
6. **Unbounded cache** — evict-expired-then-oldest, bounded at 64.

**Missing items:** TTL test with fake timers ✅ · concurrency test ✅ · `ExtractDeclaredApplicationIds` on the shared `CleanupTempDir` ✅
**Nits:** `stableReleases` inlined ✅ · array copy, with a test that one caller's in-place sort cannot corrupt the cache ✅ · doc comment states the hyphen-free-suffix and build-metadata constraints *and why they are load-bearing* ✅

Also checked, since the cache is new module state: `ValidateGitHubTag` goes through `git.getRef`, not the cache, so a freshly pushed tag is never served stale; `ResolveDependencyVersion` builds its own array before sorting, so nothing mutates a cached one; and the only dependents of this package are MJCLI and the integration-test-suite — short-lived processes, no server pins a 60s cache.

### Both pushbacks are right

1. **Keeping tag TEXT** is the call I would have made too. `ValidateGitHubTag` and external callers depend on that shape, and only the ordering needed to be semver-derived — the two tests that broke on your first attempt broke correctly.
2. **The merge-order note** should not hold this PR. An honest `null` is strictly better than a string that can never equal an installed version; the note belongs in #3352's plan, not in this PR's way.

---

## Three leftovers — follow-ups, not blockers

- **The rejection-delete can drop a *newer* entry.** `inFlight.catch(() => cache.delete(cacheKey))` deletes by key. If a slow paginated walk outlives its own 60s TTL and a second caller has already replaced the entry, the first fetch's rejection evicts the replacement. Cost is one extra fetch, never wrong data — but an identity check (`if (cache.get(key)?.Value === inFlight) cache.delete(key)`) is one line and removes the class.
- **`ClearGitHubTagCache` now clears releases too.** Documented, and the doc is honest about why. But this is a *new* export in this PR, so renaming it to `ClearGitHubFetchCache` costs nothing today and avoids a name that will need explaining forever.
- **`SemverCore`'s `??` fallback is unreachable.** `semver.valid('v1.2.3')` already returns `1.2.3`; the only inputs that reach the fallback (`V1.2.3`, `=1.2.3`) are ones it also rejects. Dead branch, harmless — drop it or keep it, but it is not doing the work the comment implies.

---

On the NUL byte: good catch, and worth the paragraph you gave it. A delimiter that is correct at runtime and invisible to grep, `file` and code search is the kind of thing that costs a reviewer an hour and never shows up as a bug.

PR #3354, opened 2026-07-29
fix(integration): record-map write amplification, pagination-violation visibility, non-blocking schema refresh

MS-BC:
Three independent, low-risk fixes on the connector apply path. Each stands alone — none depends on the others.

### 1. `TrackRecordChanges = 0` for `MJ: Company Integration Record Maps` (metadata)

`CompanyIntegrationRecordMap` is the highest-volume table the sync path writes: one row per external record ever mapped, re-touched on every sync. It was the only integration bookkeeping entity still shipping with `TrackRecordChanges = 1` — `Company Integration Run Details` and `Run API Logs` are already `0` — so every mapping upsert also wrote a `RecordChange` row, roughly doubling a sync's write volume.

Declared in `metadata/entities/.audit-related-entities.json`, alongside those same run-log siblings. See [Added since the original description](#added-since-the-original-description) for why this is metadata rather than a migration.

Nothing reads that history: the mapping row *is* the current state, its prior `ExternalSystemRecordID` has no diagnostic value, and operators audit syncs through the durable per-run artifact stream (`IntegrationTailRunEvents`).

- `TrackRecordChanges` gates `RecordChange` emission only (`DatabaseProviderBase.ShouldTrackRecordChanges` returns false immediately when off).
- `__mj_CreatedAt` / `__mj_UpdatedAt` are maintained by `trgUpdateCompanyIntegrationRecordMap`, a DB trigger independent of this flag — per-row timestamps are unaffected.
- Existing `RecordChange` rows are intentionally left in place. This stops new ones; it does not delete history.

### 2. Over-size batches surfaced on the run-event stream

The engine **never truncates** an over-size batch — every record is written, just in sub-batches — so a connector that ignores `ctx.BatchSize` was otherwise invisible except for a `console.log`. Now classified and emitted as a structured warning via `logger.warning`, queryable through `IntegrationTailRunEvents`:

| Code | Condition | Why it matters |
|---|---|---|
| `CONNECTOR_UNBOUNDED_BATCH` | first batch over-size **and** `HasMore` not true | pagination isn't implemented at all; the whole object sits in memory and grows silently with the customer's data |
| `CONNECTOR_IGNORED_BATCH_SIZE` | paging, but overshooting the requested size | bounded memory, still a contract violation |

Warned **once per object**, mirroring the existing `CONSECUTIVE_EMPTY_BATCHES` pattern, so a paging-but-over-size connector can't flood the artifact with one warning per page. Never fails the sync.

The classification is a `public static` pure function (`IntegrationEngine.ClassifyOversizedBatch`) for the same reason `RatchetProgressSnapshot` is — the decision is the interesting part and deserves a test that doesn't stand up a whole sync.

### 3. `awaitSchemaRefresh` on create/update connection

The schema refresh is a live vendor introspect — minutes on a large catalog. Awaiting it inline held the mutation open for that entire time, so the connection wizard's Finish button sat on an indeterminate spinner with no way to distinguish progress from a hang.

New `@Arg("awaitSchemaRefresh")` on `IntegrationCreateConnection` and `IntegrationUpdateConnection`, **default `true`** — existing callers are byte-for-byte unchanged. With `false`, the pipeline is launched detached and the mutation returns immediately carrying the tailable `RunID` and `InProgress: true`.

The pipeline already wrote a complete, durable, per-run event stream keyed by its run ID; the only missing piece was handing that ID to the caller *before* the work started, which `ConnectorCreationPipelineOptions.RunID` already made possible.

Trade-off the caller explicitly opts into: counts and refresh failures live on the run stream rather than the mutation response. That's why the default is `true`.

Message construction moved to `packages/MJServer/src/integration/SchemaRefreshLaunch.ts` so the one easy-to-get-wrong case is pinned by a test — a detached run's placeholder zero counts must never be reported as findings, because *"0 created, 0 updated"* reads as "the refresh found nothing", the opposite of "hasn't finished yet".

### Scope notes

Two items originally grouped here turned out to be **already on `next`** and were dropped: introspect-once-per-install, and server-side schedule uniqueness (present as `U7`).

Two items were split out rather than rushed:
- **Mirror-table `TrackRecordChanges`** — those `Entity` rows are created dynamically by CodeGen from `newEntityDefaults.TrackRecordChanges` (default true) and there is no `TrackRecordChangesBySchema` per-schema override analogous to the existing `AllowCachingBySchema`. No cheap fix site; needs its own change.
- **RSU steps → run-event stream** — `RUN_KIND_TO_TOPIC` declares an `RSU` channel that nothing in production publishes to. Wiring it needs a step-observer on `RuntimeSchemaManager.runStep` plus emitter lifecycle across four `RunPipelineBatch*` call sites. That's a feature, not a cheap win.

### Compatibility

Additive throughout. A new non-nullable field on an existing `@ObjectType` and a defaulted `@Arg` are both back-compatible — existing clients simply don't select/pass them.

### Tests

- `@memberjunction/integration-engine`: **602/602 green** (36 files). 6 new tests for the batch classification; the pre-existing `MaxBatchSize` test updated to assert the new warning (it was asserting the old `console.log` wording, and its name claimed truncation the engine never did).
- `MJServer`: 6 new tests green. The 28 test *files* that fail to load in a fresh worktree do so on unbuilt workspace deps / `reflect-metadata` and are unrelated — 0 test failures, and none of the touched files are in that set.

---

Not unit tests. A real MJAPI booted from this branch (`GRAPHQL_PORT=4141`, `packages/MJAPI`) against a real SQL Server database (`MJ_CT48`, 934 entities) with real `CompanyIntegration` rows, driven over real GraphQL.

### Defect 1 — record-map write amplification (metadata)

Against live `MJ_Local`:

- `@lookup:MJ: Entities.Name=MJ: Company Integration Record Maps` resolves to the row seeded at [`B202607091514__v5.46.x__Baseline.sql:34777`](https://github.com/MemberJunction/MJ/blob/next/migrations/v5/B202607091514__v5.46.x__Baseline.sql#L34777) — the record targets the entity it claims to. *(Measured while this shipped as a migration keyed on that row's hardcoded ID, `16248F34-2837-EF11-86D4-6045BDEE16E6`; the flag being set is what was measured, and the channel that sets it does not change the result.)*
- Real-world amplification measured on that database: **2,318 record-map rows vs 67,319 `RecordChange` rows ≈ 29 change rows per mapping row.**
- Behavioural before/after through `BaseEntity` (create → update → delete of a genuine record-map row, counting `__mj.RecordChange` at each step):

  | | RecordChange rows emitted | Save/Update/Delete succeeded |
  |---|---|---|
  | before (`TrackRecordChanges = 1`) | **3** | yes / yes / yes |
  | after (`TrackRecordChanges = 0`) | **0** | yes / yes / yes |

- `__mj_UpdatedAt` still advanced on the UPDATE (`06:44:51.9` → `06:44:52.9`), confirming trigger `trgUpdateCompanyIntegrationRecordMap` is untouched — the flag gates `RecordChange` emission only.

**One correction to the description above.** It says record-maps were "the only integration bookkeeping entity still at 1". That holds for the *run-log* siblings only — Entity Maps, Field Maps, Sync Watermarks and Company Integrations are also still at 1. The write-amplification argument for record-maps stands on its own; the "only" is too strong.

### Defect 2 — over-size batch visibility

- Branch `@memberjunction/integration-engine`: **602/602 green (36 files)**.
- The event actually written to a real `progress.jsonl` (not a mock):

  ```json
  {"level":"warn","stage":"contacts","data":{"code":"CONNECTOR_UNBOUNDED_BATCH",
   "batchIndex":1,"recordCount":10,"requestedBatchSize":5,"hasMore":false,"unbounded":true}}
  ```

- The same scenario run against `next` (worktree at `f1ab356886`) emits **0** such events — the violation was previously invisible.

### Defect 3 — `awaitSchemaRefresh`

Same live server, same `CompanyIntegration` (HubSpot, 221 discovered objects):

| call | elapsed | returns |
|---|---|---|
| `awaitSchemaRefresh: false` | **54 ms** | `Success: true`, `Message: "…schema refresh running — tail run connector-1785828408689-15f4a27c9a14fae0"` |
| `awaitSchemaRefresh: true` (default) | **10,116 ms** | counts / outcome reported inline |

At the instant the detached mutation returned, the run directory **did not exist yet** (`progress.jsonl` absent). It then appeared and reached a terminal `result.json` — proving the pipeline genuinely outlives the HTTP request and that the provider it captured stays usable after the response is sent. The default (`true`) path is unchanged.

### Two defects this live pass found — both now fixed on this branch

**(a) A detached run whose *launch* failed was completely invisible.**
The pipeline constructs its emitter inside `Run()`, so anything that throws before that — connector-driver resolution, a missing `Integration` row — produced **no run directory at all**. The mutation had already handed the caller a run ID and `Success: true`, so a client tailing that ID would wait forever on a run that never emits an event. Reproduced live on two connections:

```
Detached schema refresh (run connector-1785829421091-6cb4c7baab1c4cfb) failed:
  No connector registered for driver class "@memberjunction/connector-stripe".
→ run directory: never created (checked for 120s)
```

Fixed by publishing the launch failure onto that same run ID. Re-verified live after the fix:

```json
{"runID":"connector-1785831704719-1272e287e45e6295","success":false,"exitReason":"failed",
 "errors":[{"stage":"Launch","code":"schema-refresh-launch-failed",
            "message":"No connector registered for driver class \"@memberjunction/connector-stripe\"…"}]}
```

**(b) A refresh that FAILED was reported as a clean zero-count run** (pre-existing on `next`, in the exact surface this PR reworks). The pipeline *returns* rather than throwing when it fails at `ConnectionTest`, with every count at zero, so:

```
before: "Updated, schema refresh: 0 created, 0 updated, 0 PK-unresolved"   ← reads as "nothing changed"
after:  "Updated, schema refresh FAILED (ConnectionTest failed: Connection failed:
         No HubSpot credentials found…) — see run connector-1785856275971-d6294a773c349972"
```

Both messages verified live, before and after, on the same connection. The summary now carries the pipeline's own `Succeeded` / `FailureMessage`, and create + update render a failed refresh through one shared clause so the two paths can't drift.

### Test state after these fixes

- `@memberjunction/integration-engine`: **602 passed / 602** (36 files).
- `MJServer`: **456 passed / 0 failed** (56 files: 36 passed, 2 skipped). 18 files fail to *load* in this hand-assembled worktree on a single environment artifact (`Cannot find module '…/packages/MJServer/stream'`, from the main repo's vitest resolving a bare `stream` import). None of the 18 reference `IntegrationDiscoveryResolver` or `SchemaRefreshLaunch`; the failure is identical on files this PR never touches.

---

## Added since the original description

**Defect 1 moved from a hand-authored migration to declarative metadata, and the migration is gone.** This PR no longer adds any file under `migrations/`.

It originally shipped as a hand-authored `UPDATE __mj.Entity SET TrackRecordChanges = 0` — and misfiled, as `migrations/v5/V202608050600__v6.1.x__…`, a `v6.1.x` file in the `v5` folder. Flyway reads the version from the filename and scans recursively, so it still *ran*, which is why nothing caught it; but the folders are paired `migrations/vN` ↔ `migrations-pg/vN`, so it would have stranded its PostgreSQL counterpart at release. Relocating it to `migrations/v6/` fixed the filing but not the deeper problem: it was the wrong channel entirely.

The three siblings this change cites as precedent — `Company Integration Runs`, `Run Details`, `Run API Logs` — get `TrackRecordChanges: false` from `metadata/entities/.audit-related-entities.json`, and that file demonstrably becomes SQL: those entities appear in the generated `migrations/v5/V202604131200__v5.25.x__Metadata_Sync.sql`, applied via `spUpdateEntity` rather than a raw `UPDATE`. Per `metadata/CLAUDE.md` rule 1b, PRs contribute the declarative JSON and the build engineer's release-time `mj sync push` produces one consolidated sync migration. `check-changeset-bump.mjs` states the same equivalence in its own docstring: metadata is a `minor` trigger because it is *"the same thing on a delay."*

The one hand-authored counterexample in the repo disqualifies itself. `V202510271059__v2.113.0__Disable_Record_Changes_For_Conversations.sql` says in its header that it is a **temporary** fix, and it flipped the flag inside the same migration as a regenerated `spDeleteConversation` that depended on it. Coupled-to-DDL is the case for hand-authored SQL; a deliberate config change is not.

Nothing was at risk of being overwritten — `mj sync push` only upserts records present in files — but the declared source of truth stayed silent about this entity while the database said `0`.

The new record sets **only** `TrackRecordChanges`, deliberately not the `TrustServerCacheCompletely` / `AllowUserSearchAPI` / `AutoUpdateAllowUserSearchAPI` its siblings also carry: those are logs, this is live mapping state users legitimately search, and this PR never measured those attributes.

*Consequences for review:* the changeset stays `minor` (metadata is a trigger); the re-dating warning that used to live here is void, since there is no longer a migration whose timestamp can collide with `next`.

**Where this sits against MJ's existing practice — and the argument against it.** Turning change tracking off is not unusual: 45 entities already ship with it off, all in that same metadata file. But the established set is narrower than "high volume" — every member is **append-only and machine-generated**: runs and execution history (AI Agent Runs, AI Prompt Runs, Test Runs, Archive Runs, User View Runs, Company Integration Runs…), logs and audit trails (Audit Logs, Error Logs, Action Execution Logs, API Key Usage Logs, Record Merge Logs…), their detail children (Run Details, Run Steps, Run Outputs, Run Medias), transient queue data, and `Record Changes` itself, which "must not track itself". The reasoning is that those tables *are already the record of something*, so tracking changes to them is tracking a record of a record.

`Company Integration Record Maps` is the first member of that set that is **not** a log — it is mutable state, updated in place. So the fair counter-argument, stated plainly: change tracking earns its keep precisely on mutable state, and *"this record used to map to external ID X and now maps to Y"* is exactly the trail you would want during a mis-merge or bad-dedup investigation. This change destroys it.

The case for doing it anyway is that the tracked value is a mechanical correspondence written by the sync engine — no human ever edits it, so "who changed this and why" has no answer to give. That is the same machine-generated property the run logs have, just in mutable form. Reviewers who weight the forensic argument more heavily than that should say so; it is a judgement call, not a settled one.

One concrete coupling worth naming: `TrackRecordChanges = 1` is a **prerequisite for external change detection** (`packages/ExternalChangeDetection/README.md`, "Eligible Entities"). Record maps are not opted in — `DetectExternalChanges` defaults to `0` and nothing in `metadata/` opts any entity in — so nothing is lost today, but this does foreclose that option for the entity without saying so elsewhere.

**`SuppressActivationSchemaRefresh`** — a transient, one-shot opt-out on `MJCompanyIntegrationEntityServer` so the activation (`IsActive` false→true) schema refresh does not run inside `Save()` when the caller intends to run it itself. `IntegrationCreateConnection` sets it for `awaitSchemaRefresh: false`, which is what makes that flag genuinely non-blocking — previously the Save-side refresh ran first and awaited regardless — and moves the introspect *after* the connection test, so a connection the test rejects is rolled back without having written `IntegrationObject` rows. Defaults false: every other activation path is unchanged. It is cleared once consumed, so it cannot leak into a later save on a reused entity object.

**`IntegrationConnectorCreationPipeline.Run()` honours a caller-supplied `RunID` across coalescing.** The ID used to be discarded silently, so a caller that had already handed it to a client as "the run to tail" left that client polling `IntegrationTailRunEvents` on `Run '<id>' not found` forever — indistinguishable from "hasn't started yet". A coalesced call now publishes a terminal alias run under the requested ID naming the run that actually did the work.

### Live e2e

Real MJAPI against a purpose-built database (`MJ_3354_e2e`: 51 migrations + metadata push), with a credential-free synthetic connector, driven over GraphQL:

| Scenario | Result |
|---|---|
| Connection test rejects | `CompanyIntegration` 0, `Credential` 0, and `IntegrationObject` for the integration **0** — nothing written before the rejection |
| `awaitSchemaRefresh: false` | returns in **93–104 ms** with a RunID and `InProgress: true`, instead of blocking on a full introspect |
| Tail the returned RunID | resolves with **23 real pipeline events** (`run.start`, `ConnectionTest`, …) |
| Suppression actually fires | `activation schema refresh SUPPRESSED at the caller's request` in the server log |

Stated honestly: the **alias path itself was not exercised** — with suppression working, create no longer produces a competing run for the same connection, so `honourRequestedRunID` never fired. That is arguably the better outcome (the collision is removed at its source for create), leaving the alias as cover for the *update* path, which was not tested. The first tail poll also returned `Run not found` before succeeding on the second — the transient pre-open window, not the permanent failure, but a client that polls once would still see it. And these scenarios were not re-run against `next`, so the "before" behaviour rests on reading the code path rather than on a measurement.

### Open review notes

Two findings from an adversarial pass that are deliberately **not** fixed here, because both are design calls rather than defects:

1. `publishCoalescedAlias` is best-effort and only logs on failure — correct in that an artifact-write error must not fail a refresh that genuinely succeeded, but in that path the client is left polling `Run not found` forever, which is the exact state the change exists to prevent. Worth surfacing on the result so the launcher can stop the client tailing.
2. Nothing reconciles a caller that sets the suppression flag and then throws before running its own refresh: the connection is left `IsActive = true` with no schema, recorded only in a log line.

### Verification

Full workspace build clean (0 TypeScript errors) and **611/611** turbo test tasks, plus `check:standards`, `check:esm`, `check:changeset`, migration filename/ordering/EntityField-sequence gates, and the Angular DOM ratchet — all run locally before pushing.


MS-BC:
Base synced to current `next` (merge commit `4dc88ec71f`) and re-verified.

`Check migrations` was red because `next` is now 6.0.0, so the gate requires `__v6.1.x__` filenames. Renamed:
`migrations/v5/V202607300310__v5.51.x__CompanyIntegrationRecordMap_Disable_TrackRecordChanges.sql` → `migrations/v5/V202608041900__v6.1.x__CompanyIntegrationRecordMap_Disable_TrackRecordChanges.sql` (content unchanged; timestamp bumped past every existing migration as the gate also requires).

`Check migrations` is now **SUCCESS**. Manually re-ran the affected suites locally post-merge: Integration/engine **602 passed** and **642 passed**.

PR #3355, opened 2026-07-29
feat(rsu): publish Runtime Schema Update runs to the integration progress stream

MS-BC:
## Problem

`IntegrationRunKind` has always included `'RSU'`, and `RUN_KIND_TO_TOPIC` has always mapped both `'RSU'` and `'TableCreation'` to an `'RSU'` `IntegrationProgressTopic`. **Nothing in production ever published to it.** The channel carried no traffic and `IntegrationTailRunEvents` had no RSU runs to tail.

The only live signal for a Runtime Schema Update was polling `RuntimeSchemaUpdateStatus`, which reports the *current* step and nothing else — no history, no durable record, and nothing readable at all across the mid-run API restart the pipeline performs on itself as one of its own steps. A run that failed after the restart left no artifact to look at.

## What this does

`@memberjunction/schema-engine` now publishes plain lifecycle events through an opt-in observer, and MJServer translates them onto the existing progress-artifact stream — so an RSU run produces the same `manifest.json` / `progress.jsonl` / `result.json` triple as a sync or a connector build, tailable through the machinery that was already waiting for it.

- **`RSUObserverEvent` / `RSUPipelineObserver`** (SchemaEngine) — `run.start` / `step.start` / `step.end` / `run.end`. `PipelineObserver` defaults to `null`, so a process that never registers one (CLI, tests) pays nothing.
- **`RSUProgressBridge`** (MJServer) — owns one `IntegrationProgressEmitter` per run with `runKind: 'RSU'`, `triggerType: 'Pipeline'`.
- **One startup registration** in `packages/MJServer/src/index.ts`, right after `SetPublishHook` so the first RSU event also reaches live subscribers.

## Why the wiring is this small

`RuntimeSchemaManager` is a `BaseSingleton` and its runs are **serialized by the pipeline lock**, so at most one run is ever in flight. That means a single process-wide observer covers all six `RunPipelineBatch*` call sites — including `packages/AI/DatabaseDesigner`'s — with **no per-call-site changes**, and one current-emitter slot can never see interleaved runs.

## Layering

SchemaEngine's dependencies are `global`, `core`, `sql-dialect`, `queue`, `@octokit/rest`. It must not depend on `@memberjunction/integration-progress-artifacts`, which sits a layer above it. Hence the split: SchemaEngine publishes framework-free events; MJServer, where both are already dependencies, owns the emitter lifecycle.

## Correctness details worth reviewing

- **`run.end` on every exit path.** `RunPipelineBatch` captures its result in a local and publishes `run.end` from a `finally` — normal completion, early validation failure, and a throw that leaves no result at all. Without the last case, an observer's run stays in flight forever.
- **Step publication can't be skipped.** `runStep` is the single chokepoint for every step, so `step.start` goes there; a `recordStep` helper replaces the three remaining direct `steps.push` sites (ValidateSQL failure, ValidateSQL success, MarkOutOfSync) so nothing can record a step without publishing it.
- **`BuildRunEndEvent` is a `public static` pure function** (following `IntegrationEngine.RatchetProgressSnapshot`) so the mapping is directly testable: partial success is a **failed** run, and the **first** failure is the one surfaced.
- **Counts are in migrations, at run level only.** Only `stage.complete` / `run.complete` counts feed `aggregateCounts`. Emitting a per-step quartet would make a 12-step single-migration run report `processed: 12` — twelve migrations. So the bridge emits no per-step counts and one authoritative run-level quartet under an `RSUPipeline` stage. Step position lives in the stage name plus a `step 4 of 12` message on `stage.start`.
- **A failed step is a `stage.error`, not a `stage.complete`.** A per-item migration failure does not abort the batch, so the run stays open.
- **A retry is a new run.** `RunPipelineWithRetry` / `RunPipelineBatchWithRetry` both re-enter `RunPipelineBatch`, so each attempt legitimately gets its own run.start/run.end pair. A second `run.start` with an emitter still open closes the prior one as `rsu-run-abandoned` rather than leaking it.
- **Progress reporting can never break a migration.** Observer throws are caught and logged; the terminal `complete()`/`fail()` writes are fire-and-forget with a logged catch, because the observer contract is synchronous.

## Testing

- `packages/SchemaEngine/src/__tests__/RuntimeSchemaManager.observer.test.ts` — 5 tests over `BuildRunEndEvent` (all-success, any-failure names the first failing step, first-not-last, `undefined` result still terminates the run) plus the null-by-default observer. **SchemaEngine: 87 passed.**
- `packages/MJServer/src/__tests__/RSUProgressBridge.test.ts` — 9 tests driving the observer with real events and asserting the **actual JSONL/manifest/result on disk** (via the emitter's `rootDir` option into a temp dir), not a mock: `runKind: 'RSU'` + batch context, per-step stage start/complete with position, a failed step as `stage.error` with the run still open, the run-level migration quartet, a failure naming the breaking step, run supersession, and no-run-open events as a no-op. **MJServer: 455 passed / 0 failed** (56 files: 36 passed, 2 skipped). 18 files fail to *load* on a pre-existing environment artifact (a bare `stream` import resolved against the package dir); none of them reference this PR's files and they fail identically on `next`.

No migration. No dependency changes. Independent of #3354 and #3291.

---

Driven through a real MJAPI booted from this branch against a real SQL Server database (`MJ_CT48`), via the real `RunRuntimeSchemaUpdate` GraphQL mutation with `ALLOW_RUNTIME_SCHEMA_UPDATE=1`. Every artifact below was read off disk from `packages/MJAPI/logs/integration-runs/` — no mocks, no test harness.

**Run reaching four real pipeline steps** (`ALTER TABLE dbo.ZZ_NoSuchTable_RSUProbe ADD ProbeCol INT NULL` — passes SQL validation, acquires the lock, writes the migration file, then fails in SQL Server):

```
run dir: rsu-1785856851668-1761ef999421d850
[1]  run.start                              RSU pipeline started — 1 migration(s), 10 expected step(s)
[2]  stage.start    ValidateEnvironment     step 1 of 10
[3]  stage.complete ValidateEnvironment
[4]  stage.complete ValidateSQL
[5]  stage.start    AcquireLock             step 2 of 10
[6]  stage.complete AcquireLock
[7]  stage.start    WriteMigrationFile      step 3 of 10
[8]  stage.complete WriteMigrationFile
[9]  stage.start    ExecuteMigration        step 4 of 10
[10] stage.error    ExecuteMigration        Cannot find the object "dbo.ZZ_NoSuchTable_RSUProbe"…
                                            {"durationMs":4,"code":"RSU_STEP_FAILED"}
[11] stage.complete RSUPipeline
[12] run.fail                               failed at 'ExecuteMigration' — 0/1 migration(s) succeeded
```

`result.json`: `exitReason: "failed"`, `aggregateCounts: {processed:1, succeeded:0, failed:1, skipped:0}` — the quartet is in **migrations (1)**, not steps (4), which is the distinction the design section above argues for. `manifest.json` carries `runKind:"RSU"`, `triggerType:"Pipeline"`, and the batch context (`itemCount`, `descriptions`, `affectedTables`, `stepTotal:10`).

Two shallower failure depths were also driven live and produced correct artifacts: a run rejected at `ValidateSQL` (protected-schema guard) and, before the feature flag was set, a run rejected at `ValidateEnvironment` — each with its own run dir, its own event stream, and a terminal `result.json`.

**What this does NOT cover:** a *successful* full run. The success path goes through CodeGen + compile + API self-restart, which I did not run against this database. `emitter.complete()` on `run.end` is covered by unit test only.

### Two defects this live pass found — both fixed in 62fe8f3

**(a) The stream goes silent for the entire duration of a long step.** The bridge emitted nothing between `step.start` and `step.end`, and RSU's expensive steps (CodeGen, compile, npm install) run for minutes — so a watcher could not distinguish "compiling" from "hung", which is precisely the observability gap this PR exists to close. A 30s liveness heartbeat now runs for the in-flight step and stops at `step.end` / `run.end` / supersession (timer is `unref`'d, so it can never hold the process open).

**(b) `positionLabel` dropped the label at index 0.** The guard was `stepIndex && stepTotal`, so a 0-based first step would silently lose `step 0 of N`. The pipeline is 1-based today — verified live, indexes came through as 1,2,3,4 — so this was latent, not live. Now an explicit `== null` check.

Both are pinned by new tests (fake timers assert exactly three beats over 95s and none after the step closes; the zero-index test asserts the label survives). **9/9 bridge tests, 87/87 SchemaEngine, `tsc --noEmit` clean.**

---

## Added since the original description

**Merged `next`** (no conflicts).

### `RSUProgressBridge` is now a `BaseSingleton` — because `CurrentRunID` was unreachable

`RegisterRSUProgressBridge()` constructed the bridge and returned it, but `index.ts` discards the return value, so the only surviving reference was the observer closure handed to `RuntimeSchemaManager.PipelineObserver`. That left `CurrentRunID` reachable **only from the test file**, which builds its own instance.

Its documented purpose is letting a caller that triggered an RSU hand its client a run to tail via `IntegrationTailRunEvents` — which is what turns this from *"runs are recorded somewhere"* into *"the mutation tells you where"*. No resolver could do that.

`RSUProgressBridge.Instance.CurrentRunID` is now reachable from the RSU-trigger path. Deliberately **not** a `BaseEngine`: there is no metadata to load and nothing to `Config()` — the distinction is "one shared instance" versus "cached metadata with a load lifecycle".

Two consequences of the conversion:

- `BaseSingleton.getInstance` requires a zero-argument constructor, so emitter options move from a constructor parameter to `Configure()`.
- `Reset()` returns the bridge to idle without emitting a terminal event. Singleton state outlives a single caller, so a test — or a server shutting down — must be able to drop an in-flight emitter and its heartbeat interval instead of having the next run inherit them.

Two new tests pin exactly this: that `Instance` is shared and its `CurrentRunID` is the same run a caller would tail, and that `Reset` returns to idle.

### The branch had no changeset

`check:changeset` reported **none added** against `next`, on a PR that changes source in two published packages. Added at `patch`/`patch` (no migration, no metadata).

### A green suite was hiding a red typecheck

Worth recording, because "tests pass" was not sufficient evidence here. MJServer's `tsc --noEmit` was **failing** on this branch:

```
error TS2305: Module '@memberjunction/schema-engine' has no exported member 'RSUObserverEvent'
error TS2339: Property 'PipelineObserver' does not exist on type 'RuntimeSchemaManager'
```

…while all 1045 tests passed. Vitest resolves the TS source through path aliases; `tsc` resolves the built dist, which had not been rebuilt for the new exports. Building `@memberjunction/schema-engine` clears it.

### Verification

- `RSUProgressBridge`: **11 passed** (was 9)
- `@memberjunction/server`: **1045 passed / 56 skipped**
- `tsc --noEmit` clean on MJServer and SchemaEngine; both build clean
- `check:changeset` green

### One thing this does NOT do

It makes RSU runs **observable, not resumable**. A retry re-enters the pipeline and is legitimately a new run with its own start/end pair. What survives the mid-run API restart is the *record*, not the run.


MS-BC:
Base synced to current `next` so CI re-runs against the current base. No conflicts.

Manually re-ran the Integration/engine suites locally post-merge: **602 passed** and **642 passed**.

Issue #3346, opened 2026-07-28
Integration suite: 8 live-tier checks use a stale AgentRunID filter; RS7 asserts a superseded MIN_TERM_LENGTH


SDesai-BC:
Test debt found during the v5.50.0 release build. These are stale tests, not product defects — but they fail on every live-tier run and mask real signal.

## 8 checks use the dropped `AIPromptRun.AgentRunID`

`AL3`, `AL4`, `AL6`, `AL7`, `CF1`, `CF2`, `CF3`, `SA4` fail reproducibly (byte-identical messages across two independent runs) because the harness filters `MJ: AI Prompt Runs` on `AgentRunID`, which #3274 dropped:

- `packages/TestingFramework/integration-test-suite/src/checks/agent-live-shared.ts:199`
- `packages/TestingFramework/integration-test-suite/src/checks/_it-live-agent-harness.ts:195` and `:267`

**Line 297 of `agent-live-shared.ts` already uses the correct derivation** — `AIAgentRunStep.StepType='Prompt'` → `TargetLogID` — so this is a missed call site rather than an unknown fix. Verified the derivation works: 51 Prompt steps, 0 with null `TargetLogID`.

Symptoms are all "expected N, got 0" (`at least two prompt runs — got 0`, `child prompt runs recorded tokens Σ=0`, `one injected carry-forward message — got 0`), because the query fails rather than returning rows.

## `RS7` asserts a threshold that was intentionally changed

`agent-rag-search.checks.ts:271` queries `'mj'` with the comment `// 2 chars → below MIN_TERM_LENGTH` and asserts `TotalCount === 0`. #3289 deliberately lowered `MIN_TERM_LENGTH` from 3 → 2 across three providers ("Bug C3 — minimum term length lowered from 3 to 2 so legitimate short queries aren't silently dropped"), so a 2-char query now legitimately fans out and returns results.

The unit tests for the new boundary were added (`MIN_TERM_LENGTH boundary (C3)` in two spec files); this integration check was missed. **Fix is one character:** query `'m'` instead of `'mj'`.

## Also worth noting

`CE2` (`agent-compaction-e2e`) fails deterministically across both runs and remains unexplained. Its own message suggests checking that the seeded "Conversation Summary" prompt has an active model binding — that is disproven: it has three bindings, all `Active`, all pointing at active models, none of them Gemini. Cause unknown.

## Why this matters beyond tidiness

[...]

PR #3364, opened 2026-07-29
fix(integration-tests): reach prompt runs via steps, not AgentRunID

SDesai-BC:
Part of #3346 — fixes both defects that issue names, but **deliberately does not close it** (see *What this does not fix*).

## The defect

Three live-harness helpers filtered `MJ: AI Prompt Runs` on `AgentRunID`. That column does not exist on `AIPromptRun` — its only agent-facing field is `AgentID`. A prompt run is reachable from its agent run only through the step that invoked it: an `MJ: AI Agent Run Steps` row whose `TargetLogID` is the prompt run's ID.

The reason a nonexistent column survived in committed code is the second half:

```ts
    }, user);
    return r.Success ? (r.Results || []) : [];   // agent-live-shared.ts, getPromptRuns
```

`RunView` does not throw — it returns `Success: false` with an `ErrorMessage`. Coalescing that to `[]` makes a SQL error indistinguishable from "this run made no model calls", so every caller read zero prompt runs and either passed vacuously or failed on an unrelated-looking assertion (`at least two prompt runs — got 0`). **The swallow is the actual defect; the wrong column name only exploited it.** Fixing the filter without fixing the swallow would leave the next schema drift equally invisible.

`agent-live-shared.ts:297` already derived prompt runs correctly via `TargetLogID`, so the rule was known — it was just restated in four places and three of them were wrong.

## What changed

- **`promptRunIdsFromSteps`** — the linkage rule, extracted so it exists once. `purgeAgentRun` had it inline and is now a caller.
- **Two step-type sets**, because `Prompt` alone is not the rule. base-agent writes a prompt run's id into `TargetLogID` on `Prompt` (`:8833`), `Compaction` (`:13689`), and `Tool` (`:5965` — explicitly "without a duplicate `Prompt` step for the same call", so a Prompt-only rule cannot reach it by any route). `PROMPT_RUN_BEARING_STEP_TYPES` (all three) is used for **deletion**, which must be exhaustive or it orphans rows; `ROLLUP_BEARING_STEP_TYPES` (`Prompt` + `Compaction`) is used for **token reads**, mirroring `base-agent.ts:13255`, which counts only those two toward `AIAgentRun.TotalTokensUsed`. A single blanket filter would have fixed the orphaning and broken the token reconciliation in the same stroke. The set is applied in SQL as well as in the JS predicate — narrowing one and widening the other drops rows before the filter ever sees them.
- **`getPromptRuns`, `readPromptRunsForAgent`** — resolve prompt-run IDs through the invoking steps, then query by `ID`. `PromptRunRow.AgentRunID` is gone. `getPromptRuns` takes the rollup set (`Prompt` + `Compaction`); `readPromptRunsForAgent` takes the full set, since it means "every prompt run this agent produced".
- **`deepDeleteRunTrees`** — resolves prompt runs **before** deleting steps, using the full step-type set. The old order deleted steps first, destroying the only path to those rows, so teardown silently leaked every prompt run it claimed to purge. This is a behavior fix, not just a filter fix.
- **`requireRows`** — replaces the swallow in the read helpers, throwing with the provider's own error text. Teardown paths stay non-throwing by design but now `console.error` instead of going silent.

`RS7` asserted a short-circuit using a 2-char query (`'mj'`) while `SearchEngine.MIN_TERM_LENGTH` is now `2` — #3289 lowered it from 3 so `"AI"`/`"US"` are searchable — so `2 < 2` is false and the query legitimately fans out. It now probes with a single character, below both the old and current thresholds, so the check tests the short-circuit rather than tracking the threshold's value. `SR5` in `search.checks.ts:129` had already made this exact change with the same reasoning; RS7 was missed because its bundle `IT62` is **live-model** tier and the deterministic gate never runs it.

## Correction to #3346's scope

The issue lists 8 failing checks — `AL3`, `AL4`, `AL6`, `AL7`, `CF1`, `CF2`, `CF3`, `SA4` — as all caused by the stale filter. Tracing each to the code it actually calls, **that attribution holds for 3 of the 8**:

| Check | Path to prompt runs | Explained by this defect? |
|---|---|---|
| `AL3`, `AL4`, `AL7` | `getPromptRuns` | **Yes** — fixed here |
| `AL6` | never reads prompt runs (disables model bindings, asserts run `Status`) | No |
| `CF1`, `CF2`, `CF3` | `firstPromptMessages` → steps → `TargetLogID` → prompt run by `ID` | No — already correct |
| `SA4` | `verifyAgentRun` → `TargetLogID` → `verifyPromptRun` | No — already correct |

`verifyAgentRun`'s `AgentRunID` filter (`ai-verify.ts:115`) is on `AI Agent Run Steps`, where that column is real.

The defect also affected checks the issue did **not** name: `readPromptRunsForAgent` is consumed by `agent-payload-guards` and `agent-artifact-tools`.

So `AL6`, `CF1`, `CF2`, `CF3`, `SA4` should be expected to still fail after this merges, for reasons not yet identified. I have not investigated them.

## Evidence

| What | Result |
|---|---|
| `integration-test-suite` build (`tsc && tsc-alias`) | clean |
| `integration-test-suite` unit tests | **149/149** |
| `SearchEngine` unit tests | **427/427**, incl. `MIN_TERM_LENGTH boundary (C3)` |
| Deterministic integration tier | **53/53** (109.9s at final commit), incl. `IT52 - Unified Search Seams` |
| Live-model integration tier | **9/15 bundles** (up from 7/15) — `IT53` AL1–AL7 all pass, `IT62` RS7 passes |
| DB smoke test | old filter errors, new query resolves 65/65, `Actions`/`Sub-Agent` → 0 prompt runs |
| Regression guard proven | reintroduced the `AgentRunID` filter → source scan failed 1/10; reverted → 10/10 |

**The repaired SQL has now been executed against a real database.** A smoke test run directly against SQL Server confirms the premise and the fix:

| Probe | Result |
|---|---|
| `AIPromptRun` columns matching `AgentRunID`/`AgentID` | only `AgentID` — the column genuinely does not exist |
| Old filter `WHERE AgentRunID = ...` on `vwAIPromptRuns` | **`Invalid column name 'AgentRunID'`** — a hard SQL error, which the `?? []` swallow disguised as "no prompt runs" |
| New step-based query | executes; every bearing step's `TargetLogID` resolves to a real `AIPromptRun` (**65/65**) |
| `Actions` / `Sub-Agent` steps with a `TargetLogID` | 28 and 9 rows, **0** pointing at an `AIPromptRun` — confirming empirically that excluding them is correct |

**Live-model tier: 9/15 bundles, up from 7/15.** The two bundles that flipped to passing are exactly the two containing the fixed checks:

- `IT53 - Agent Loop Foundation (live)` — **AL1–AL7 all pass** (99.4s of real model calls). AL3/AL4/AL7 are the checks this PR targets, and AL4/AL7 reconcile Σ(prompt run tokens) against `AIAgentRun.TotalTokensUsed`. Because `getPromptRuns` previously always returned `[]`, that equality had **never actually executed** before; it now runs and holds — which also validates scoping that read to `ROLLUP_BEARING_STEP_TYPES` rather than the full set.
- `IT62 - Agent RAG and Search` — **RS7 passes**, and it genuinely ran (its own `→ below-minimum-length query short-circuited with no fan-out` appears in the log). Note RS1–RS4 and RS6 *skipped* in this deployment because `MJ: AI Agent Notes.Note` is not `IncludeInUserSearchAPI`, so the bundle's green overstates its coverage — RS7 is the leg that matters here and it executed.

Zero `Invalid column name` and zero `requireRows` failures anywhere in the live log.

<details>
<summary>Smoke-test output and the SQL behind it (re-runnable against any migrated database)</summary>

```
PASS  AIPromptRun has NO AgentRunID column — columns found: [AgentID]
PASS  AIPromptRun does have AgentID
PASS  AIAgentRunStep has AgentRunID/StepType/TargetLogID — found 3/3
PASS  OLD AgentRunID filter is a SQL ERROR (not an empty result) — Invalid column name 'AgentRunID'.
PASS  NEW step-based query executes without error — 50 linkage rows sampled
PASS  every bearing step TargetLogID resolves to a real AIPromptRun — 65/65 resolve

step-type census (rows with a non-null TargetLogID):
  Prompt       steps=   65  →AIPromptRun=   65
  Actions      steps=   28  →AIPromptRun=    0
  Sub-Agent    steps=    9  →AIPromptRun=    0

prompt runs a Prompt-only rule would ORPHAN in this database: 0

SMOKE TEST PASSED
```

The four probes, as plain SQL:

```sql
-- 1. The premise: AIPromptRun has no AgentRunID column.
SELECT COLUMN_NAME FROM INFORMATION_SCHEMA.COLUMNS
WHERE TABLE_SCHEMA='__mj' AND TABLE_NAME='AIPromptRun'
  AND COLUMN_NAME IN ('AgentRunID','AgentID');            -- returns AgentID only

-- 2. The old filter is an ERROR, not an empty result. This is what the swallow hid.
SELECT TOP 1 ID FROM [__mj].[vwAIPromptRuns]
WHERE AgentRunID = '00000000-0000-0000-0000-000000000000';
--> Msg 207: Invalid column name 'AgentRunID'.

-- 3. The new path executes, and every linkage row resolves to a real prompt run.
SELECT COUNT(*) AS linked
FROM [__mj].[vwAIAgentRunSteps] s
INNER JOIN [__mj].[vwAIPromptRuns] p ON p.ID = s.TargetLogID
WHERE s.StepType IN ('Prompt','Compaction','Tool') AND s.TargetLogID IS NOT NULL;

-- 4. Census — which step types actually point at a prompt run. Actions and Sub-Agent
--    return 0, which is why they are excluded rather than merely unlisted.
SELECT s.StepType, COUNT(*) AS steps,
       SUM(CASE WHEN p.ID IS NOT NULL THEN 1 ELSE 0 END) AS pointing_at_prompt_run
FROM [__mj].[vwAIAgentRunSteps] s
LEFT JOIN [__mj].[vwAIPromptRuns] p ON p.ID = s.TargetLogID
WHERE s.TargetLogID IS NOT NULL
GROUP BY s.StepType ORDER BY steps DESC;
```

Run against a freshly migrated `v5.50.0` database with the deterministic tier and the full live-model tier already applied to it. The census was taken twice — before and after the live-model run — and reported no `Compaction` or `Tool` rows either time, which is the basis for saying that half of the rule is unexercised by data rather than verified by it.

</details>

**Two things this still does not prove.** The `Compaction`/`Tool` widening is **not exercised by any data** — the database contains no prompt-run-bearing `Compaction` or `Tool` steps even after a full live-tier run, so that half rests on the `base-agent.ts` write sites plus the unit tests, not on observed rows. And the six still-failing live bundles (`IT54`, `IT55`, `IT56`, `IT57`, `IT60`, `IT61`) are untouched by this PR; their failures are model non-compliance, run-budget exhaustion, and unrelated product issues — see the scope correction above.

## On the new test

`prompt-run-linkage.test.ts` pins `promptRunIdsFromSteps` for both step-type sets — that `Prompt`, `Compaction` and `Tool` are all reached by the full set, that `Tool` is excluded by the rollup set, that a null `TargetLogID` is dropped, and that `Sub-Agent`/`Actions` steps are ignored since their `TargetLogID` points at a child agent run or an action log. It also pins that `requireRows` surfaces a failed `RunView`.

Neither would catch someone re-adding an `AgentRunID` filter — only a real database rejects that, and the live tier is triage-only, so the regression would ship exactly as it did the first time. So the file also scans the check sources for `AgentRunID` near `'MJ: AI Prompt Runs'`, the same filesystem-drift technique `sibling-parity.test.ts` uses for bundle↔metadata parity. Comments are stripped first so prose about the defect (including this fix's own docs) cannot trip it, and a companion test asserts the scan found a non-empty set of files so a moved directory fails loudly instead of scanning nothing.

## Notes for the reviewer

No product source is touched — the diff is entirely within `packages/TestingFramework/integration-test-suite`. No migrations, no metadata, so nothing to apply or push after merge.

A `patch` changeset is included. `@memberjunction/integration-test-suite` is `private: true`, so changesets versions it and writes its `CHANGELOG.md` but never publishes it — and past changesets have named it, so this follows existing practice rather than introducing it. Because `config.fixed` is `[["@memberjunction/*"]]`, `changeset status` lists the whole group at patch; that is fixed-group lockstep, which any release does regardless of this PR.

One thing worth splitting out separately: `MIN_TERM_LENGTH = 2` is declared three times — `SearchEngine.ts:207`, `EntitySearchProvider.ts:32`, `FullTextSearchProvider.ts:31` — kept in sync only by a code comment. That duplicated decision is what produced the RS7 drift, and consolidating it is a product change outside this diff's scope.


packages/TestingFramework/integration-test-suite/src/checks/agent-live-shared.ts
```diff
@@ -159,6 +159,94 @@ export async function resolveRunId(
...
+/**
+ * Unwrap a RunView result, throwing with the provider's own error text when the query failed.
+ *
+ * RunView does not throw — it returns `Success: false` plus an ErrorMessage. Coalescing that to
+ * `[]` makes a broken query indistinguishable from "no rows", which is exactly how a filter on a
+ * column that does not exist (`AIPromptRun.AgentRunID`) survived in this file: every caller read
+ * zero prompt runs and either passed vacuously or failed on an unrelated-looking assertion. A
+ * check that cannot read its observables must fail loudly, naming the query that broke.
+ */
+export function requireRows<T>(result: RunViewResult<T>, what: string): T[] {
```
AN-BC: same convention issue here @SDesai-BC , exported functions and public class members are `PascalCase` in MJ

packages/TestingFramework/integration-test-suite/src/checks/agent-live-shared.ts
```diff
@@ -159,6 +159,94 @@ export async function resolveRunId(
...
+ * run is reachable from its agent run only through the step that invoked it — an
+ * `MJ: AI Agent Run Steps` row whose TargetLogID is the AIPromptRun's ID. Filtering prompt runs on
+ * AgentRunID is a SQL error, not an empty result set.
+ *
+ * `stepTypes` decides WHICH linkage rows count; it is explicit because the correct answer differs
+ * by purpose (see the two constants above). Step types outside the set are skipped because their
+ * TargetLogID points at something else entirely — `Sub-Agent` at a child agent run, `Actions` at
+ * an Action Execution Log — so treating those ids as prompt-run ids reads or deletes wrong rows.
+ */
+export function promptRunIdsFromSteps(
```
AN-BC: and here and below

packages/TestingFramework/integration-test-suite/src/checks/agent-live-shared.ts
```diff
@@ -159,6 +159,94 @@ export async function resolveRunId(
...
+    stepTypes: readonly string[] = PROMPT_RUN_BEARING_STEP_TYPES
+): Promise<string[]> {
+    if (agentRunIds.length === 0) {
+        return [];
+    }
+    const inList = agentRunIds.map(id => `'${id}'`).join(',');
+    // The step-type set is applied in SQL as well as in promptRunIdsFromSteps — narrowing here and
+    // widening there would silently drop rows before the JS filter ever sees them.
+    const typeList = stepTypes.map(t => `'${t}'`).join(',');
+    const r = await new RunView().RunView<{ StepType: string | null; TargetLogID: string | null }>({
```
AN-BC: bad code, `new RunView()` uses the global provider, you need to thread in the run view provider to use into the function, you can fall back to creating a new instance, but must be multi-provider supported to do this right. 

**SAME THING BELOW**

AN-BC: In the MJServer stack we have this threaded properly and in most places you'll have `this.ProviderToUse` which you can cast to `IRunViewProvider` for this purpose.

AN-BC (commented):
Fix the issues noted here @SDesai-BC  and once you do that, if @cadam11 agrees that this is a bug fix suitable for `5.51.0` you can merge. Make sure you review the comments here carefully, there are multiple issues to fix.

SDesai-BC:
Note for reviewers: cadam11 has already approved this PR.

packages/TestingFramework/integration-test-suite/src/checks/agent-live-shared.ts
SDesai-BC: Renamed in 004f3512cb — the three exports this PR adds are now PascalCase: `RequireRows`, `PromptRunIdsFromSteps`, `ResolvePromptRunIdsForAgentRuns` (call sites and unit tests updated). The file's pre-existing camelCase exports (`resolveRunId`, `getRunSteps`, `getPromptRuns`, …) predate this PR, so I left them for a follow-up to keep this diff focused on the linkage fix — happy to fold them in here if you'd prefer.

packages/TestingFramework/integration-test-suite/src/checks/agent-live-shared.ts
SDesai-BC: Covered by the same rename pass in 004f3512cb — `PromptRunIdsFromSteps` and `ResolvePromptRunIdsForAgentRuns` are PascalCase now.

packages/TestingFramework/integration-test-suite/src/checks/agent-live-shared.ts
SDesai-BC: Fixed here and below in 004f3512cb. The shared read helpers now take the run-scoped provider and build their view via `RunView.FromMetadataProvider(provider)`, falling back to `new RunView()` only when no provider is supplied — the multi-provider pattern from the data-access rules. `purgeAgentRun` and the `_it-live-agent-harness` functions already received a required `provider`, so they now use it directly (no fallback), and every call site in the live bundles threads `ctx.Provider`.

PR #3365, opened 2026-07-30
feat: Query Materialization Phase 2 — parameterized RowFilterBroad read-time injection

JosueGarcia-BC:
## Query Materialization — Phase 2: parameterized `RowFilterBroad` read-time injection

Stacked on #3311 (**#3311 merges first**). This PR is based on the `claude/query-entity-materialization` branch so its diff shows only the Phase 2 delta; GitHub will retarget it to `next` automatically once #3311 merges.

### What this finishes
Phase 1 (#3311) built everything for parameterized **RowFilterBroad** materialization *except the read-time predicate injection*, and gated it off. Phase 2 finishes it: a caller can run a materialized parameterized stored Query with `RunQueryParams.DataSource: 'Materialized'`, and the provider serves it from the broad materialized table with the query's row-filter parameters injected as **bound** read-time predicates — falling back to the live query on any uncertainty.

### The safety model (why this is safe)
Serving the **live query is always correct**, so any read-time uncertainty resolves to *serve live*, never *serve wrong*. The only way to return wrong rows is an unfaithful reconstructed predicate — eliminated by:
- **Prove at classify time** — only mint a `RowFilterBroad` materialization when the render-and-diff verifier has proven the param is a clean top-level conjunctive `column <op> value` on a projected output column, **and** the operator is in a safe whitelist (`=, !=, <>, <, >, <=, >=, IN, NOT IN`; `LIKE`/`IS`/`BETWEEN` stay live-only).
- **Bind, never interpolate** — read-time values are bound as SQL parameters (`?`/`$n`), so injection is structurally impossible and type/collation match the source column (the broad table is a physical copy).
- **Fall back to live** — no materialization / not fresh-Active / a param missing from the spec / an unsupported operator / any execution error → run the live query.

### Layers
- **`codegen-lib`** — verifier captures each predicate's operator + value shape (normalized `column <op> value`, flipping `value < column`); qualifier builds a structured `ReadFilterSpec` + gates operators; `manage-metadata` persists the spec, enables Bucket-1, and orders the `QueryParameter` load by `Name` (it has no `Sequence`).
- **`core`** — `RunQueryParams.DataSource: 'Live' | 'Materialized'` (mirrors `RunViewParams`).
- **`generic-database-provider`** — `InternalRunQuery` redirect + injection-safe predicate builder + fallback-to-live.
- **`server` / `graphql-dataprovider`** — `DataSource` threaded through every RunQuery path (singular, batch, cache-check, SystemUser).
- **Migration** — `MaterializedResult.ReadFilterSpec` (+ folded CodeGen objects for from-scratch CI).

### Validation
- **Reconstruction differential** (materialized read == live, direct SQL): **13/13** SQL Server, **10/10** PostgreSQL — collation, operator-flip, NULL/3VL, `IN`/`NOT IN`, multi-param, injection-as-bound-param.
- **Provider-level `RunQuery` E2E** (booted provider, real minted Query + MaterializedResult): **15/15** incl. dropped-view exec-error fallback.
- **Write-path classify E2E**: **12/12** (safe `=`, range `>=`, `LIKE`/mixed refusal). *This run caught a real latent bug — `classifyParameterizedQueryForMaterialization` ordered `QueryParameter` by a non-existent `Sequence` column; fixed to `ORDER BY Name`.*
- Unit: classify 52, read-builder 22.

Design + full safety argument + documented residual limitations: `plans/query-entity-materialization-phase2.md`.


packages/Materialization/src/MaterializationRefresher.ts
```diff
@@ -0,0 +1,1261 @@
...
+            : `CREATE TABLE ${obj(shadow)} AS SELECT * FROM (${sourceSelect}) AS src`;
+        const statements = [
+            // 1) Build a fresh shadow (CASCADE clears any leftover shadow + transient view dependency).
+            `DROP TABLE IF EXISTS ${obj(shadow)} CASCADE`,
+            createShadow,
+            // 2) Atomic repoint — readers immediately see the freshly-built shadow.
+            `CREATE OR REPLACE VIEW ${obj(viewName)} AS SELECT * FROM ${obj(shadow)}`,
+            // 3) Drop the stale table, rename the shadow into the canonical name, repoint the view back
+            //    (keeps the canonical `materialized_<name>` name stable for migration-reuse detection, §12).
+            `DROP TABLE IF EXISTS ${obj(tableName)} CASCADE`,
```
rkihm-BC: **[Merge blocker] PG full-rebuild swap is no longer transactional.** In #3311 these statements were a single `BEGIN; DROP …; ALTER … RENAME; CREATE OR REPLACE VIEW …; COMMIT;` block; Phase 2 split them into separate `ExecuteSQL` calls (each autocommits), while the SQL Server branch kept its `BEGIN TRANSACTION … COMMIT`. A failure between `DROP TABLE <canonical>` (186) and `RENAME` (187) destroys the old snapshot and leaves `materialized_vw…` pointing at a missing table — the exact crash-atomicity gap your #3311 note says was closed on both engines. Restore the single `BEGIN;…COMMIT;` string (PG DDL is transactional, so no interim standalone repoint is needed).

packages/Materialization/src/MaterializationRefresher.ts
```diff
@@ -0,0 +1,1261 @@
...
+                });
+                return `(${placeholders.join(', ')})`;
+            });
+            insertBatches.push({ sql: `INSERT INTO ${obj(shadow)} (${colList}) VALUES ${tuples.join(', ')}`, params });
+        }
+
+        const postStatements = isPostgres
+            ? [
+                  `CREATE OR REPLACE VIEW ${obj(viewName)} AS SELECT * FROM ${obj(shadow)}`,
+                  `DROP TABLE IF EXISTS ${obj(tableName)} CASCADE`,
```
rkihm-BC: **[Merge blocker] Same non-transactional swap on the dirty-group / external-rebuild PG path.** The `postStatements` swap here was also split out of its `BEGIN;…COMMIT;`. Wrap it back into one transaction for parity with the SQL Server branch just below.

rkihm-BC (commented):
## Code review — Query Materialization Phase 2 (read-time parameter injection)

_Detailed review performed with **Claude Code** — an independent pass plus hand-verification against the stacked branch._

The safety model — prove-at-classify → bind-never-interpolate → fall-back-to-live — is sound and held up under stress: I could not find a serve-wrong-rows path through the predicate-reconstruction machinery (operator-flip, IN/NOT IN, projection/ORDER BY/OR/function/CAST refusal, mixed safe+unsafe refusal, and every execution-error fallback all resolve to serve-live-or-refuse). The passthrough-sentinel hardening for numeric/date/replace transforms is real and tested. Two things to fix:

### 🔴 Merge blocker — PG swap-atomicity regression (inline)
Phase 2 split the transactional `BEGIN; DROP; RENAME; CREATE OR REPLACE VIEW; …; COMMIT;` swap that #3311 introduced into individual, non-transactional statements on PostgreSQL — while SQL Server kept its `SET XACT_ABORT ON; BEGIN TRANSACTION … COMMIT`. A crash between `DROP TABLE <canonical>` and `RENAME` now destroys the old snapshot and leaves the wrapper view dangling. This reintroduces exactly the bug your #3311 note says was fixed "on both rebuild paths." Flagged inline on both the full-rebuild and dirty-group paths.

### 🟠 High — RunQuery cache path doesn't account for `DataSource`
_(The affected line is pre-existing code Phase 2 doesn't touch, so this is body-level rather than inline.)_ Phase 2 adds `RunQueryParams.DataSource: 'Materialized'` (`core/src/generic/runQuery.ts:126`) and honors it in `InternalRunQuery`, but the client cache path in `providerBase.RunQuery` — `queryCacheEngaged` (~`providerBase.ts:1816`) and `GenerateRunQueryFingerprint` (`localCacheManager.ts`) — was not updated. Unlike the RunView path (which both bypasses cache for materialized reads via `runViewCacheEligible` **and** folds `ds:materialized` into the fingerprint), the RunQuery path does neither. So a `CacheLocal:true` query cross-serves Live and Materialized from one slot, and a materialized RunQuery result gets cached with a TTL and served stale after an out-of-band refresh. Mirror the RunView guards: add `&& !IsMaterializedDataSource(params.DataSource)` to `queryCacheEngaged`, and/or thread `DataSource` into `GenerateRunQueryFingerprint`. It's gated behind explicit `CacheLocal:true`, so it's latent until a consumer combines the two — but worth closing before adoption.

### 🔴 Migration
`V202607291006` needs the same re-stamp + `v6.1.x` relabel as #3311 (it's earlier than `next`'s `V202607311852`). Flagged inline.

### ✅ Verified correct
The classify/verify/qualify gates, the bound-parameter builder, and the fallback wiring all check out.

If you read any of these differently — the PG-atomicity call especially — I'd genuinely rather discuss than have you act on my say-so. Please reach out anytime.

migrations/v5/V202607291006__v5.51.x__Materialization_ReadFilterSpec.sql
```diff
@@ -0,0 +1,406 @@
+/* ============================================================================
+   Query Materialization — Phase 2: Read-Time Filter Spec
+   v5.51.x
```
rkihm-BC: **[Merge blocker] Re-stamp + version label.** This is dated `202607291006`, earlier than `next`'s latest `V202607311852`, so Flyway strict-in-order validation will reject it. Re-date to after `202607311852` and relabel `v5.51.x → v6.1.x` (next release is **6.1.0**).

rkihm-BC (commented):
## Re-review — Query Materialization Phase 2 (read-time parameter injection)

I re-verified all four Aug 3 findings against the current head (46e7f57) and scanned the new commits. **Everything I flagged is genuinely fixed** — this PR is good on its own merits.

**Resolved and verified (3 blockers + 1 high):**
- 🔴 **PG swap atomicity** (full-rebuild *and* dirty-group/external-rebuild paths) — restored to single `BEGIN;…COMMIT;` batches executed as one `ExecuteSQL` call, at parity with the SQL Server `XACT_ABORT` branch (`MaterializationRefresher.ts:221-226, :535-541, :1058-1072`). The crash-atomicity regression is gone.
- 🟠 **RunQuery cache ignores DataSource** — closed by **bypassing** the CacheLocal layer for materialized reads (`queryCacheEngaged && !IsMaterializedDataSource(...)`, `providerBase.ts:1837`) plus a server-side `BypassCache:true` on the status-gated read. Defensible design (a snapshot can't be event-invalidated); closes both the Live↔Materialized slot collision and the TTL-stale hazard.
- 🔴 **Migration** — re-stamped/relabeled correctly (`V202608070007__v6.2.x__Materialization_ReadFilterSpec.sql`, after next's head).

The read path remains sound under stress: allowlisted operators, all values **bound** (never interpolated), identifier escaping via `quoteMaterializedIdentifier`, bidirectional param-coverage invariant, ordering-fidelity now parsed off the *platform* SQL, and every uncertainty falls back to the live query. The `9cabc70a7` test hardening (killing false-confidence via mutation testing) is good rigor.

**Not approving yet — purely on merge sequencing, not on this PR's content.** This is stacked on #3311, and I've requested one change there (PG mint DDL escaping). **This PR is clear to merge once #3311's residual is fixed and #3311 is merged** (GitHub will retarget this to `next` at that point). No further changes needed here from my side — I'll flip to approve as soon as #3311 lands.

Standing caveat worth honoring before the stack merges: a fresh SQL Server + PostgreSQL E2E on the reassembled code, given the volume of fixes layered on since the last full run.

PR #3366, opened 2026-07-30
fix: BaseEngine cache callback fingerprint mismatch breaking cross-server invalidation

jordanfanapour:
## Summary
- Extracts a shared `BuildRunViewParamsForConfig()` method in `BaseEngine` to ensure `RunViewParams` fingerprints are consistent across `LoadSingleEntityConfig`, `LoadMultipleEntityConfigs`, and `RegisterCacheChangeCallbacks`
- Fixes a bug where `RegisterCacheChangeCallbacks` used a manually-constructed params object with different fields (e.g., hardcoded `ResultType: 'entity_object'`, missing `IgnoreMaxRows`) compared to the load methods, causing Redis pub/sub cache invalidation callbacks to never match

## Test plan
- [x] Build `@memberjunction/core` (`cd packages/MJCore && npm run build`)
- [x] Verify cross-server cache invalidation works via Redis pub/sub (cache callback fingerprints now match between load and registration)
- [x] Confirm single-server cache refresh still functions correctly

### DIFF
--- a/.changeset/steady-engines-align.md
+++ b/.changeset/steady-engines-align.md
@@ -1,0 +1,5 @@
+---
+"@memberjunction/core": patch
+---
+
+Fix BaseEngine cache callback fingerprint mismatch that broke cross-server invalidation via Redis pub/sub by extracting a shared BuildRunViewParamsForConfig method to ensure consistent RunViewParams across LoadSingleEntityConfig, LoadMultipleEntityConfigs, and RegisterCacheChangeCallbacks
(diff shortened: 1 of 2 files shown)


cadam11:
This fix now also lives in #3375 (same changeset + baseEngine.ts edit, plus a regression test and a follow-on BuildRunViewParamsForConfig fix), so the two copies have diverged and #3375 is ahead.

One PR should own it before today's 12pm CT batch close. If #3375 rides the 5.51.0 rebuild, this one should close with it. If you'd rather keep the fix standalone here, bring the test and follow-on commit over and drop those commits from #3375. Either way works, two open copies is the one option that doesn't, it's a merge hazard when the freeze lifts Monday.

jordanfanapour:
Moved fix to #3375 and closed this one.

PR #3368, opened 2026-07-30
feat(components): theme-aware registry components — viz ramps, spec-level StyleOverrides, hardcoded-color migration

CaelebB-BC:
# Theme-aware registry components

Interactive/registry components previously hardcoded color literals (`#fee2e2`, `#3b82f6`, …), so they ignored the host's MJ theme entirely — wrong brand colors, broken dark mode, and no way to honor a user asking for "blue charts" without baking literals into generated code. This PR closes that gap in three layers.

## 1. Visualization ramps + status/overlay slots join the styles contract

- `ComponentStyles` (in `@memberjunction/interactive-component-types`) gains two new visualization slots:
  - **`sequentialScale`** — single-hue intensity ramp (light→dark) for heatmaps, choropleths, density shading
  - **`divergingScale`** — `{ low, mid?, high }` ramp for measures with a meaningful midpoint (variance vs target, gain/loss, sentiment)
- New theme tokens in `_tokens.scss` back them:
  - `--mj-viz-seq-1..7` — built from the brand ramp (not literals), so an org overlay re-tints intensity shading automatically. Dark mode **re-anchors** the ramp (brand-900 → brand-300) because "more ink = darker" inverts its reading on a dark surface.
  - `--mj-viz-div-low/mid/high` — anchored on the status hues (error → neutral → success), with lightened dark-mode endpoints.
- The theme bridge (`BuildStylesFromTheme()`) also now populates the slots the migrated components read for chrome:
  - **Status text/border pairs** (`successText`/`successBorder` and the warning/error/info equivalents) from `--mj-status-*-text/-border`
  - **`overlay`** (modal scrim) from `--mj-bg-overlay`
  - **`secondary`/`secondaryHover`** from `--mj-brand-secondary` / `--mj-brand-secondary-hover` (the hover token is new in `_tokens.scss`)
- Bridging keeps its sanity gates: a sequential ramp needs ≥2 stops, a diverging scale needs both endpoints (mid optional). Incomplete themes fall back to new defaults in `SetupStyles()`.

## 2. Spec-level `StyleOverrides` — user-requested styling as data, not literals

- `ComponentSpec` gains an optional **`styleOverrides`** field (`chartPalette`, `sequentialScale`, `divergingScale`, `fontScale`) plus provenance: `source: 'user-request' | 'organization-default'` and `requestText` so the intent survives regeneration and is auditable.
- **`fontScale`** (`'small' | 'normal' | 'large'`) rescales the whole `typography.fontSize` ladder by one factor — px values only, with a 10px legibility floor; rem/clamp()/keyword sizes are left alone rather than guessed at. Scaling the tokens instead of asking a generator to pick different ones is what makes the scale hold: every piece of text moves together, including text inside registry components and libraries the generator never touched.
- Component code **never reads this field**. A new **`ApplyStyleOverrides(base, overrides)`** utility in `@memberjunction/react-runtime` layers the overrides above the theme-resolved styles, so components keep reading `styles.chartPalette` and friends as usual.
- Only visualization slots exist on the color contract — deliberately no background/text/border slots, since those can't be flipped safely for dark mode without derived per-mode variants.
- Wired into both hosts:
  - **`MJReactComponent`** (Angular): overrides apply on top of both an explicit `styles` input and the bridged live theme, outside the theme memo so they survive a theme flip. The merge is **memoized by reference**, so the styles object identity stays stable across reads and components that key effects on `styles` (e.g. chart teardown/rebuild) don't churn.
  - **In-place restyle**: a regenerated spec whose only change is `styleOverrides` (same name/code/version) no longer requires a teardown — the host detects it and re-renders in place with the new styling.
  - **Test harness** (`ComponentRunner`): generation-time screenshots apply the overrides over `SetupStyles()`, so the visual judge sees what the user asked for rather than the stock palette.

## 3. Generic registry components migrated off hardcoded colors

All 10 `metadata/components/code/generic/` components (`data-grid`, `entity-data-grid`, `simple-chart`, `simple-drilldown-chart`, `simple-map`, `single-record-view`, `ai-insights-panel`, `data-export-panel`, `open-record-button`, `mj-form-field`) now read `styles.colors.*` / palette slots instead of color literals. The previous hardcoded values are kept as inline fallbacks for hosts that render without styles, so behavior is unchanged when no theme is present. Notable mechanics:

- `data-grid` derives its status-pill families from the theme's status hues with a shade ramp, darkening any pill color until it can carry white text; themed chart palettes overflow into the built-in 50-color list so distinct values keep distinct colors.
- Components that need translucent variants (`scrims`, area fills, tooltips) derive them via a hex→rgba helper that falls back to the original translucent literal when a theme expresses the color as `rgb()`/named — never an opaque fill.

## Tests

- `packages/React/runtime/src/__tests__/component-styles.test.ts` covers: `SetupStyles` defaults (viz ramps, status text/border, overlay), theme bridging of the new slots including the ≥2-stop / endpoint gates and the secondary pair, and `ApplyStyleOverrides` merge semantics — no-op on empty overrides, no mutation of base, per-slot validity checks, and the full `fontScale` behavior (both directions, floor, non-px passthrough, combination with color overrides). Full runtime suite: 108 tests passing.

## Scope

No schema or migration changes; the `ComponentSpec`/`ComponentStyles` additions are optional fields (additive, non-breaking). Changeset included: `minor` for `@memberjunction/interactive-component-types` and `@memberjunction/react-runtime`, `patch` for `@memberjunction/ng-react` and `@memberjunction/react-test-harness`.

Note: the `adopted standards` CI check is red due to a pre-existing contradiction on `next` (file-storage must declare `@memberjunction/ng-shared` for the missing-deps gate, but the ui-layers manifest scan has no allow mechanism) — fix prepared separately, not part of this PR.


jordanfanapour (approved):
Changes LGTM

AN-BC (commented):
LGTM, looks like it is fully back compat?

CaelebB-BC:
> LGTM, looks like it is fully back compat?

Yes should but will have a slight visual impact on existing components making them uniform and comply with MJ themes.

PR #3369, opened 2026-07-30
feat(metadata-sync): size-split captured SQL migrations (maxFileSize, default 90 MiB)

MarceloT-BC:
## What
Adds `sqlLogging.maxFileSize` to MetadataSync's `mj sync push` SQL capture. When a `formatAsMigration` capture would exceed the limit, it's split into multiple ordered, individually-runnable migration parts (`*.partNN.sql`), each under the limit. **Default 90 MiB** — safely below GitHub's 100 MiB push-block.

## Why
A large metadata push (e.g. the MoreCheese demo — ~102 MB of SQL from ~97k records) produces a single migration file that exceeds GitHub's 100 MiB hard limit and can't be pushed. That previously required splitting the file by hand; this makes it automatic.

## How
- Streaming **rollover** in `SqlLoggingSessionImpl` (`@memberjunction/generic-database-provider`): when the next statement would push the current part over `maxFileSize`, it finalizes that part and rotates to a new one — always **on a statement boundary** (a statement is never torn; a `GO` batch never spans a file), with a footer-reserve so finalized parts stay strictly under the limit.
- Each part is a self-contained migration (own header/footer, per-part batch counting reset).
- A capture under the limit still produces a single file at the original path (**backward compatible**); `maxFileSize: 0` disables splitting.
- The session exposes `filePaths` (all parts); a push result reports `sqlLogPaths`.
- The 90 MiB default is applied at the MetadataSync push/watch layer; the generic provider logger stays opt-in (undefined = off), so other consumers (Explorer settings, MJServer) are unaffected.

## Packages
- `@memberjunction/generic-database-provider` (minor)
- `@memberjunction/metadata-sync` (minor)

## Testing
- New unit tests in `GenericDatabaseProvider` (8), incl. a **90 MiB boundary** test (>90 MiB capture → multiple parts, each ≤ 90 MiB, no statement lost/duplicated/torn). Both package suites green (GDP 848 ✓ / 5 skip; MetadataSync 261 ✓).
- **Real-path validation** via `mj sync push`:
  - Core metadata (countries + state-provinces, ~4,005 records @ 1 MiB) → 13 parts, all under limit, record set preserved across re-runs.
  - **Real MoreCheese** (~97k records / ~102 MB @ 40 MiB) → 3 parts, on the actual dataset.

### Repro
```jsonc
// .mj-sync.json (metadata root)
"push": { "alwaysPush": true },              // forces a full capture without a DB wipe
"sqlLogging": { "enabled": true, "formatAsMigration": true, "maxFileSize": 1048576 }
```
```sh
npx mj sync push --dir=metadata --include=countries,state-provinces
# → "📝 SQL log split into N parts (each ≤ ~1.0 MiB …)"; every part under the limit.
```

## Notes
- Backward compatible; changeset included (minor).
- Found a vestigial public `SQLLogger` class in `@memberjunction/metadata-sync` (superseded by the provider-level logger; only its `.enabled` flag is used). Left untouched to keep this PR focused; filed as a follow-up cleanup.


packages/GenericDatabaseProvider/src/SqlLogger.ts
```diff
@@ -61,6 +84,24 @@ export class SqlLoggingSessionImpl implements SqlLoggingSession {
     return this._emittedStatementCount; // Return actually emitted statements
   }
 
+  /**
+   * Every part file produced by this session, in order. Single-element (`[filePath]`)
+   * unless size-based splitting rotated across multiple parts.
+   */
+  public get filePaths(): string[] {
```
AN-BC, line 91: MJ convention is to use `PascalCase` for exported functions and for public class members. Please fix this across the class if it was incorrect, guessing so as this might not be the only violation. It is okay to make this breaking change under patch, nobody uses this class outside of the repo I'm quite sure.

packages/MetadataSync/src/services/WatchService.ts
```diff
@@ -372,8 +372,11 @@ export class WatchService {
             // batchSeparator is intentionally omitted — CreateSqlLogger injects the platform-appropriate
             // separator automatically (GO for SQL Server, nothing for PostgreSQL) via PlatformBatchSeparator.
             variableBatchThreshold: syncConfig.sqlLogging?.variableBatchThreshold ?? 200,
+            // Split into multiple part files past this size so a long watch capture stays under
+            // host file-size limits. Defaults to 90 MiB; `sqlLogging.maxFileSize: 0` disables it.
+            maxFileSize: syncConfig.sqlLogging?.maxFileSize ?? DEFAULT_SQL_LOG_MAX_FILE_SIZE,
```
AN-BC, line 377: would this result in no setting of maxFileSize then falling back to the global default? If that is the case, then doesn't this mean we always split?

AN-BC (commented):
LGTM overall, some minor comments. We need to hold this for `6.0.0` though as this is new functionality so we can't add it to the `5.50.x` LTS branch anymore @MarceloT-BC 

Nice work :)

packages/GenericDatabaseProvider/src/SqlLogger.ts
```diff
@@ -61,6 +84,24 @@ export class SqlLoggingSessionImpl implements SqlLoggingSession {
...
+    return this._partPaths;
+  }
+
+  /**
+   * Derives a part file path by inserting `.partNN` before the extension of the base path.
+   * e.g. `/x/push.sql` + index 2 → `/x/push.part02.sql`.
+   */
+  private _partFilePath(index: number): string {
+    const parsed = path.parse(this.filePath);
+    const suffix = `.part${String(index).padStart(2, '0')}`;
```
rkihm-BC, line 101: `padStart(2, '0')` breaks on-disk ordering past 99 parts: `part100` sorts lexicographically **before** `part11` (`'0' < '1'` at the 6th char). The in-memory `_partPaths` array preserves true order so `filePaths` is always correct — but any consumer that sorts by filename (a glob, `ls`, a migration runner) gets them out of order. This isn't hypothetical at the PR's documented settings: the repro runs core metadata at a 1 MiB limit → 13 parts, and MoreCheese (~102 MB) at 1 MiB is ~100+ parts. Suggest padding to 3+ digits, or sizing the pad to the final part count.

packages/GenericDatabaseProvider/src/SqlLogger.ts
```diff
@@ -73,9 +114,55 @@ export class SqlLoggingSessionImpl implements SqlLoggingSession {
...
+    if (!this._fileHandle) {
+      return;
+    }
+
+    // On the FIRST rotation, rename the base file to part01 (before finalizing) so all parts
+    // share `.partNN` naming and the part-1 footer is labeled correctly. The open handle keeps
+    // writing to the same file after the rename (POSIX renames the path, not the descriptor).
+    if (!this._rotated) {
+      const part01 = this._partFilePath(1);
+      await fs.promises.rename(this.filePath, part01);
```
rkihm-BC, line 145: This renames `this.filePath` while `this._fileHandle` is still open, and the comment reasons only about POSIX ("renames the path, not the descriptor"). On Windows, renaming a file open by another handle historically throws `EPERM`/`EBUSY`. Modern libuv opens with `FILE_SHARE_DELETE` so it *likely* works — but this path only triggers on large pushes, on a platform many MJ devs use, with no Windows test. Could you either confirm it was validated on Windows, or switch to close→rename→reopen to be safe?

packages/GenericDatabaseProvider/src/SqlLogger.ts
```diff
@@ -220,10 +307,24 @@ export class SqlLoggingSessionImpl implements SqlLoggingSession {
       }
     }
 
-    // Batch separator logic:
+    // At this point `logEntry` is the complete statement block (description + SQL + params),
+    // WITHOUT any batch separator yet. When size-based splitting is enabled, decide rotation on
+    // the block size BEFORE applying the (intra-file) batch separator — so a freshly rotated part
+    // never starts with a stray leading GO. The +1 accounts for the trailing blank line added below.
```
rkihm-BC, line 313: Stale comment: "The +1 accounts for the trailing blank line added below" — the rotation check below doesn't actually add `+1`; the `FOOTER_RESERVE_BYTES` reserve is what absorbs the footer/separator/blank-line slack. Harmless but misleading; suggest dropping the `+1` sentence.

packages/GenericDatabaseProvider/src/SqlLogger.ts
rkihm-BC, line 91: Re: the PascalCase request on this getter — worth an explicit decision rather than leaving the thread open. The entire `SqlLoggingSession` interface is already camelCase (`filePath`, `startTime`, `statementCount`), so renaming only `filePaths` → `FilePaths` would be locally inconsistent with its sibling `filePath`. Per the rule the reviewer is right; but the honest resolution is either migrate the whole interface (the breaking change AN-BC explicitly blessed under patch) or consciously keep camelCase for local consistency and say so. Also returns the live `_partPaths` array (not a copy) — the `readonly string[]` contract leaks a mutable reference; PushService spreads it defensively so no bug today.

rkihm-BC (commented):
Detailed review follow-up. The core streaming-rollover engine is well-designed, well-tested, and backward-compatible — nice work. A few items below; a few land on unchanged files so I've put them here rather than inline.

**CLI drops `sqlLogPaths` (unchanged file `packages/MetadataSync/src/plugins/index.ts`).** `PushResult.sqlLogPaths` is populated and the multi-part list is printed mid-push via `callbacks?.onLog`, but the plugin layer never forwards it:
- The JSON result exposes only `sqlLogPath` (= part01), not `sqlLogPaths` (`plugins/index.ts:261`) — a `--format json` consumer can't discover all parts to commit them.
- The final rich-text summary prints only `result.sqlLogPath` — "📄 SQL log saved to: …part01.sql" (`plugins/index.ts:330`) — understating a multi-part split.

The PR's own rationale ("the user needs to know the migration is several files so they commit/rename all of them") argues for surfacing `sqlLogPaths` in both places, not just the transient onLog stream.

**Migration version-uniqueness (please confirm).** If these captured files are committed as Flyway-style migrations, inserting `.partNN` before the extension leaves every part sharing one version token (`V…__push.part01.sql`, `…part02…` → same version, different description), which Flyway rejects as duplicate versions. If MJ's runner doesn't parse them that way this is a non-issue — but the PR frames them as "individually-runnable migration parts," so worth confirming how they're consumed downstream.

**Test coverage gap at the MetadataSync layer.** All 8 new tests are at the `SqlLoggingSessionImpl` level. Nothing covers the MetadataSync wiring: the 90 MiB default injection, the multi-part `onLog` formatting, or `sqlLogPaths` propagation — which is exactly where the CLI gap above lives.

**Two open review threads have no reply and the branch hasn't moved since.** The PascalCase ask and the "do we always split?" question (a misread — defaulting to 90 MiB only splits when output actually exceeds it) are both still unresolved. Worth closing both explicitly before merge.

rkihm-BC (changes requested):
## Requesting changes — the engine is right; the branch cannot merge as it stands, and my August questions are still open

Re-reviewed at `4ab240baac`. The follow-up commit addressed the CLI, the padding, the Windows rename, and the stale comment. Three things block merge, none of them in the splitter itself.

### Blocking

1. **Merge conflict with `next`.** `git merge-tree` reports a content conflict in `packages/GenericDatabaseProvider/src/SqlLogger.ts`, and GitHub shows the PR as dirty. It is a comment-only hunk (your rotation comment against `next`'s redeclare-guard comment), so it is a two-minute resolve, but the branch is 2,390 commits behind and still on `npm@11.7.0` from before the pnpm cutover, so it does not install or build under the current toolchain either. While resolving: `next` added `_currentBatchDeclaredNames`, and `_rotateToNextPart` resets only `_currentBatchVariableCount`. Add `this._currentBatchDeclaredNames.clear()` there, or a new part inherits the previous file's declared names and emits an extra `GO` on reuse. Harmless output, but wrong.
2. **Changeset level.** Both entries are `minor`; the branch touches no `migrations/` or `metadata/`, and `.claude/rules/changesets.md` reserves `minor` for exactly those. `node .github/scripts/check-changeset-bump.mjs` fails on this branch. Change both to `patch`.
3. **Six review threads have no reply.** Mine from August: the migration version-uniqueness question (see below), and the MetadataSync-layer coverage gap, which the new test only partly closes (three assertions on the default constant; nothing exercises the multi-part `onLog` formatting or `sqlLogPaths` propagation). AN-BC's: the PascalCase decision for `filePaths` and "does this mean we always split?" (the answer is no, executed below, but it deserves a reply).

### The version-uniqueness question, now answered by execution

The splitter never emits a Flyway name, so this is a documentation item rather than a code defect, but the PR body calls the parts "individually-runnable migration parts" and eleven captured migrations already live in `migrations/`. If someone renames the three parts to `V<ts>__v6.2.x__Name.part001.sql` and so on with one version token, Skyway 0.6.2 parses all three to the same `Version` and `DetectDuplicateMigrations` groups them as a duplicate. Distinct timestamps per part work. One sentence in the docs saying each part needs its own version token closes it.

### Verified, all executed

- Splitting unit is one `logSqlStatement` call; the limit is bytes, strict `>`, and a single oversize block gets its own part with no loop. Twelve parts over multibyte content all decode with a fatal `TextDecoder`, so no code point is ever cut. Legacy mode ends each part with `GO`; threshold mode resets the batch counter on rotation. `maxFileSize` of 0, undefined, negative, or NaN disables splitting.
- A split between statements that share a `DECLARE` would be unsafe in general, but every MJ save block is one self-contained request (`WrapSaveCallForResult` builds preamble plus `EXEC` into one string, logged in one call), so it is unreachable from `mj sync push`. Worth a sentence in `types.ts`.
- Red-then-green: `sql-logger.test.ts` 9/9 fail against `next`, 9/9 pass restored; the config test 3/3 fail (export absent on `next`), 3/3 pass. GenericDatabaseProvider 851 passed, MetadataSync 264 passed.
- The CI unit failure (`check-registry` expecting 7 checks, got 11) is `next` being inconsistent with itself at the 8/11 merge base; the PR touches neither file.

### Small things

- `filePath` is `readonly`, set once, and after rotation names a file that was renamed away; `types.ts:110-112` says it is the first part, which is false, and `WatchService.ts:125` logs it. `PushService` correctly uses `filePaths[0]`.
- Stale two-digit naming in the changeset (`*.partNN.sql`) and `types.ts:95`; the README option table lists neither `variableBatchThreshold` nor `maxFileSize`.
- Re-running against the same base path leaves stale higher-numbered parts from a previous run on disk. `PushService` avoids it by timestamping the filename; note it for anyone calling the logger directly.

PR #3370, opened 2026-07-30
[6.0] docs(mcp): plan for multi-version MCP support (2026-07-28 + legacy)

AN-BC:
Plan-only PR — adds `plans/mcp-multi-version-support.md`. No package changes, no migrations, no code. Opened so the work is tracked and the open decisions can be settled in review.

## Why

The [MCP 2026-07-28 revision](https://blog.modelcontextprotocol.io/posts/2026-07-28/) makes the protocol stateless: the `initialize`/`initialized` handshake and `Mcp-Session-Id` are gone, replaced by per-request `_meta` envelopes and `server/discover`. Alongside that: mandatory `Mcp-Method`/`Mcp-Name` routing headers, MRTR (`input_required` → retry with `inputResponses`) replacing server-initiated sampling/elicitation, cacheable list results, an extensions framework, and authorization hardening. Roots, Sampling, Logging, legacy HTTP+SSE, and DCR are all deprecated on a 12-month window.

We need to speak the new revision **and** the 2025-era revisions, selectable by config — server-side per deployment, client-side per external server row.

## Audit findings

- **Neither MCP package contains a single protocol-version reference today.** `grep` for `LATEST_PROTOCOL_VERSION`, `MCP-Protocol-Version`, `protocolVersion`, or any version literal across `packages/AI/MCPClient` and `packages/AI/MCPServer` returns nothing. We inherit whatever `@modelcontextprotocol/sdk@^1.26.0` defaults to, with no switch to change it.
- **SDK v2 supplies the multi-version machinery directly.** Server: `createMcpHandler(factory, { legacy: 'stateless' | 'reject', ... })` plus an `isLegacyRequest()` predicate for routing legacy traffic to our current sessionful wiring. Client: `ClientOptions.versionNegotiation = { mode: 'legacy' | 'auto' | { pin } }` plus `ConnectOptions.prior` for cached era verdicts. One dependency covers 2026-07-28 and the whole legacy ladder back to 2024-10-07.
- **Two facts that de-risk the migration substantially.** `McpServerFactory` is `(ctx: { era, authInfo, requestInfo }) => McpServer`, which maps 1:1 onto our existing per-session `registerAllTools(...)` — the tool layer does not move. And there is no Zod migration: the repo's `zod@^3.25.0` already ships `~standard`, so our existing `z.object()` tool schemas are accepted as-is by v2's Standard Schema `registerTool`.
- **Two things v2 drops that we depend on.** No `SSEServerTransport` (our `/mcp/sse` + `/mcp/messages` endpoints) and no `WebSocketClientTransport` (our `'WebSocket'` value in `MCPTransportType` and the DB CHECK constraint).
- **Pre-existing OAuth gaps, unrelated to versioning but shipping alongside.** `OAuthManager.ts` never validates the RFC 9207 `iss` parameter before redeeming the authorization code — that is the authorization-server mix-up defense the new spec requires. And neither `ClientRegistration.ts` nor `ClientRegistry.ts` handles `application_type`, which breaks `localhost` redirects for desktop and CLI clients.

## Design principle

**No protocol version literal in MJ source.** Versions live only as config data or DB column values, validated at runtime against the SDK's `SUPPORTED_PROTOCOL_VERSIONS`. That includes deliberately *not* putting a CHECK constraint on the new `PinnedProtocolVersion` column — the set of spec revisions is open and grows outside our release cycle, so a CHECK would force a migration per revision. The *mode* columns are closed sets and do get CHECK constraints. This is called out in the plan as an explicit, documented exception to the value-list rule in `migrations/CLAUDE.md`.

Under that rule, adopting the next revision is: bump the SDK, add a config string, run the integration bundle.

## Decisions needed (§0 of the plan)

| # | Decision | Recommendation |
|---|---|---|
| D1 | Legacy serving: sessionful vs SDK stateless fallback | `sessionful` default (no-op upgrade), flip next minor |
| D2 | How long `/mcp/sse` survives | One minor behind a flag + warning, then default off |
| D3 | `WebSocket` transport: deprecate or port locally | Deprecate — **needs a data check first** |
| D4 | Default era posture for fresh installs | `both` for 6.0 |
| D5 | Per-request auth cost under statelessness | **Measure before Phase 2 merges** |
| D6 | CIMD adoption | Follow-up after 6.0 |
| D7 | MRTR / Tasks / subscriptions in 6.0? | Follow-up |

D3 and D5 have factual answers we can go get rather than debate — query `MCPServer` for `TransportType = 'WebSocket'`, and benchmark the cached `TokenValidator` + `GetAPIKeyEngine` path. That collapses the list to five judgment calls.

## Scope and sizing

Phases 0–6 + 8 are the 6.0 scope; Phase 7 (MRTR, Tasks, `subscriptions/listen`) is a follow-up. Phase 2 — the dual-era `/mcp` endpoint — is the bulk of the work; everything else is S/M. Defaults throughout are chosen so the upgrade is a behavioral no-op for existing deployments.

## Testing

Nothing to run — documentation only, no source touched. The plan's Phase 8 specifies the test work that lands with the implementation: config-schema cases, `versionNegotiation` mapping, era-routing tests for the `isLegacyRequest` fork, plus a deterministic integration bundle exercising legacy-client → MJ-server and modern-client → MJ-server against the same endpoint.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qv9kBfsToi1mQx8rtKmmqi)_

### DIFF
--- a/plans/mcp-multi-version-support.md
+++ b/plans/mcp-multi-version-support.md
@@ -7,6 +7,29 @@ per external server row — and structured so a future revision is a dependency 
 
 Status: **plan / not started**
 Branch: `claude/mcp-multi-version-support-h0sh0h`
+Target: **6.0**
+
+---
+
+## 0. Key decisions remaining
+
+Seven open calls, each one gating a phase below. Nothing here blocks *writing* the code — they all
+change what we ship or what we break, so they want an answer before the corresponding phase merges.
+Ordered by how early they bind.
+
+| # | Decision | Options | Recommendation | Gates |
+|---|---|---|---|---|
+| **D1** | **Legacy serving model** — do we keep our sessionful `streamableTransports` map behind `isLegacyRequest()`, or hand 2025-era traffic to the SDK's per-request stateless fallback? | `sessionful` / `stateless` | Ship `sessionful` as the default so the upgrade is a behavioral no-op; document `stateless`; flip the default in the next minor. | Phase 2 |
+| **D2** | **`/mcp/sse` lifetime** — legacy HTTP+SSE is deprecated by the spec and has no v2 server transport. How long do we carry it? | one minor w/ warning / immediate removal / indefinite | One minor release behind `protocol.legacy.sse.enabled` (default `true`) with a deprecation warning on connect, then default `false`. This is the *only* reason `@modelcontextprotocol/sdk@^1` stays installed. | Phase 3 |
+| **D3** | **`WebSocket` transport** — v2 ships no `WebSocketClientTransport`, and `'WebSocket'` is a live value in `MCPTransportType` and the DB CHECK constraint. | deprecate / port ~80 lines locally | Deprecate — it was never in the MCP spec's transport set. **Needs a data check first**: is any `MJ: MCP Servers` row actually using it? If yes, port. | Phase 5 |
+| **D4** | **Default era posture for new deployments** — does a fresh install answer `both` or `modern` only? | `both` / `modern` | `both` for 6.0. Revisit once the ecosystem's client mix is known. | Phase 0 |
+| **D5** | **Per-request auth cost** — statelessness means every modern request re-authenticates. `TokenValidator` + `GetAPIKeyEngine` are cached, but this is unmeasured. | measure then decide | Benchmark before Phase 2 merges; if the cached path is not cheap enough, add a short-TTL validated-token cache keyed on the token hash. | Phase 2 |
+| **D6** | **CIMD adoption** — DCR is deprecated in favor of Client ID Metadata Documents but remains functional. | now / follow-up | Follow-up after 6.0. Note the deprecation in the OAuth proxy README so integrators are not surprised. | Phase 6 |
+| **D7** | **Extensions scope for 6.0** — MRTR, Tasks, and `subscriptions/listen` are all opt-in and all genuinely useful to us (mid-call confirmation on `Delete_*` tools; `Run_Agent` as a task; cross-instance `tools/list_changed`). | in 6.0 / follow-up | Follow-up. Phases 0–6 + 8 are the 6.0 scope; Phase 7 lands after. | Phase 7 |
+
+Two of these have a factual answer we can just go get rather than debate — **D3** (query the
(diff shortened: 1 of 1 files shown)


AN-BC:
### Status — watching this PR

I'm subscribed to activity here and will post a new comment on each round of work.

**CI:** green — `Check migrations` passed (47s). It's the only check that runs on this PR, which is correct: the diff is a single markdown file under `plans/` and touches no migrations, no package source, and no metadata.

**Changeset bot:** no action needed. This PR adds a planning document and bumps no package, so the absence of a changeset is the right outcome — the bot's own "if these changes should not result in a new version, you're good to go" branch applies.

### Decisions open (§0 of the plan)

Repeating the table here so it's answerable inline without opening the file. Two of these have a factual answer rather than a judgment call, and I can go get both on request:

- **D3 — `WebSocket` transport.** SDK v2 ships no `WebSocketClientTransport`, but `'WebSocket'` is a live value in `MCPTransportType` and in the DB CHECK constraint. Recommendation is to deprecate (it was never in the MCP spec's transport set) — but that's only safe if no `MJ: MCP Servers` row actually uses it. **Someone with a environment can settle this with one query.**
- **D5 — per-request auth cost.** The modern era is stateless, so every request re-authenticates. `TokenValidator` + `GetAPIKeyEngine` are cached, but the path is unmeasured. Worth benchmarking before Phase 2 merges; if it's not cheap enough, a short-TTL cache keyed on token hash covers it.

The remaining five are judgment calls: legacy serving model (D1), `/mcp/sse` lifetime (D2), default era posture (D4), CIMD timing (D6), and whether MRTR/Tasks/subscriptions make 6.0 (D7). My recommendation on all five is in the plan; the theme is that 6.0 defaults reproduce today's behavior exactly, so the upgrade is a no-op for existing deployments, and we flip toward modern-only in a later, announced release.

### What I'd suggest next

This is plan-only, so it's mergeable as-is if the direction reads right — the decisions don't have to be settled before merge, only before the phases they gate. If you'd rather lock D1–D7 first, comment inline and I'll fold the answers into §0 and push.

---

PR #3060, opened 2026-07-06
Plan: Medallion architecture consumption — Microsoft Fabric + Databricks external data source support

AN-BC:
## Summary

Adds `plans/medallion-fabric-databricks-external-sources.md` — the implementation plan for supporting customers with a **medallion lakehouse** (bronze/silver/gold) investment, positioning MJ as the consumption/application layer over their gold layer via the External Data Sources primitive.

Two workstreams:

- **Workstream A — Microsoft Fabric** (2–3 days): no new driver. Extend `SQLServerExternalDataSourceDriver` with a Microsoft Entra service-principal auth mode (Fabric SQL endpoints speak TDS but refuse SQL auth), a new "Azure Service Principal" credential type, and a new `ExternalDataSourceType` metadata row reusing the same `DriverClass`. Includes `isAuthError` AADSTS coverage so the existing rotated-credential self-heal works for expired client secrets.
- **Workstream B — Databricks** (1–2 weeks): new provider package `@memberjunction/external-data-source-databricks` extending `BaseSqlExternalDataSourceDriver`, over `@databricks/sql` as an optional peer dependency (Snowflake pattern). Backtick quoting, LIMIT/OFFSET paging, Unity Catalog `information_schema` introspection including informational PK/FK constraints, PAT + OAuth M2M auth.

## Vendor test environments (per request, included in the plan)

- **Databricks**: Free Edition (serverless SQL warehouse) and/or 14-day full trial, testing against the pre-provisioned read-only **`samples` catalog** (`samples.nyctaxi.trips`, `samples.tpch.*`) — zero fixture seeding needed. Gated by `RUN_DATABRICKS_INTEGRATION=1`.
- **Microsoft Fabric**: 60-day trial capacity with the built-in **Wide World Importers** sample warehouse / lakehouse-tutorial dataset and the **NYC Taxi – Green** copy-data sample; SPN setup recipe documented. Gated by `RUN_FABRIC_INTEGRATION=1`.

Both suites are opt-in (no service containers exist for these vendors), mirroring the existing Snowflake suite model.

## Notes

- Plan document only — no code changes in this PR.
- Phases A and B are independent and parallelizable.
- Open questions tracked in §7 (Free Edition PAT availability, Fabric endpoint collation/metadata-sync quirks, `@databricks/sql` decimal marshalling).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01KSrEimz5foApJ4WCRs6rFM

---
[...]

PR #3373, opened 2026-07-30
feat(external-data-sources): Databricks SQL Warehouse driver (Medallion Workstream B)

JosueGarcia-BC:
## Summary

Adds **`@memberjunction/external-data-source-databricks`** — a read-only External Data Source driver (`DatabricksExternalDriver`) for **Databricks SQL Warehouses** over the official `@databricks/sql` SDK (always HTTPS).

This is **Workstream B** of the Medallion external-sources plan (#3060). Workstream A — Microsoft Fabric — shipped in #3110; this completes the second half. Snowflake is the structural model (the other "cloud warehouse behind an HTTPS SDK").

## What's included

- **Driver** — backtick identifier quoting, `LIMIT`/`OFFSET` paging (`LIMIT ALL` for the offset-only case), `:name` named-parameter binds, PAT (`access-token`) and OAuth M2M service-principal (`databricks-oauth`) auth with connection self-heal, and Unity Catalog `information_schema` introspection including informational (`NOT ENFORCED`) primary/foreign keys.
  - **Numeric fidelity** is handled SDK-side via `preserveBigNumericPrecision` (DECIMAL → exact string, BIGINT → JS `bigint`, normalized losslessly by the base). This deterministically resolves the plan's open DECIMAL-precision question, so there is **no per-object CAST probe** — the driver stays simple and avoids a metadata round-trip per read.
  - Types are derived from the SDK's public `DBSQLClient` (installed as a build-time devDep, kept an optional runtime peer via `peerDependenciesMeta` — the AWSKMS pattern).
- **Metadata** — a "Databricks SQL Warehouse (External)" data-source type row and a "Databricks Access Token" credential type. OAuth M2M reuses the existing generic "OAuth2 Client Credentials" type rather than duplicating it; `RequiredCredentialTypeID` is left null per the Snowflake precedent (the driver accepts both auth modes).
- **Runtime + CodeGen wiring** — `@memberjunction/server-bootstrap` dependency + regenerated class-registration manifest so the driver registers at MJAPI runtime; `@memberjunction/codegen-lib` dependency + dynamic import so external-entity introspection resolves the driver.
- **Docs** — driver table row, a full Databricks setup recipe, and a fidelity note in `EXTERNAL_DATA_SOURCES_GUIDE.md`.

## Testing

- **31 unit tests** (mocked SDK) — dialect (backtick quoting, `LIMIT ALL` offset paging), PAT/OAuth branching, `:name` binds, read-only screening, `isAuthError`, and the execute path against a fake `DBSQLClient` (including bigint→string normalization).
- **Gated integration suite** (`RUN_DATABRICKS_INTEGRATION=1`) against the read-only `samples.tpch` fixture — **verified live: 8/8 pass** against a real SQL Warehouse (TestConnection, projected/filtered/paged RunView, LoadSingle, native-query join with a named bind, Unity Catalog introspection, and clean bad-object failure). Wired as a **manual-only** CI job (`workflow_dispatch`) since there is no service container for Databricks (like Snowflake/Fabric).

## Notes

- `@databricks/sql` is an **optional peer** — a base install never pulls it in; install it where the driver runs.
- CI (Ubuntu) and production (path read from the DB `ConnectionConfig`) are unaffected by the local Windows-only Git Bash path-mangling quirk encountered during live verification.


rkihm-BC (commented):
## Code review — Databricks SQL Warehouse driver

_Detailed review performed with **Claude Code** — an independent pass plus hand-verification, including checking the driver against the `@databricks/sql` SDK surface and the Snowflake precedent._

The driver is clean and faithful: full base-contract implementation, backtick identifier escaping, `LIMIT ALL` offset paging, `:name` binds, read-only screening **before** connecting, PAT/OAuth branching with `isAuthError` correctly excluding 403, and lossless numeric fidelity via `preserveBigNumericPrecision`. The 31-test unit suite is meaningful (dialect, injection-resistance, auth classification, execute/close-on-error). Two items:

### 🟡 Class-registration manifest drift (inline)
The regenerated manifest dropped the entire `@memberjunction/server (3 classes)` block — `MJEntityPermissionEntityServer`, `MJTenantFilterMiddleware`, `RateLimitMiddleware` — and the count went 1000→998. It was clearly regenerated in a tree where `@memberjunction/server`'s dist wasn't resolvable. Not a runtime break (the barrel still side-effect-registers), but two of those are security middleware and this drift shouldn't ride along in an unrelated PR. Please regenerate in a fully-built tree so the diff is Databricks-only.

### 🔵 Dependency wording (inline)
Minor: unlike Snowflake (types-only devDep), the full `@databricks/sql` runtime lands as a devDep, so it **is** installed in the monorepo dev install — "base install never pulls it in" holds for downstream consumers only.

### Note
I couldn't run `npm run build` / the vitest suite in the review worktree — per the Definition of Done, please confirm the 31 unit tests pass in a built tree before merge.

If anything here doesn't match what you intended, please reach out — happy to look again.

packages/ServerBootstrap/src/generated/mj-class-registrations.ts
```diff
@@ -2437,17 +2436,14 @@ export const CLASS_REGISTRATIONS: any[] = [
...
-    RateLimitMiddleware,
     IntegrationTestDriver,
 ];
 
 /** Marker constant indicating the manifest has been loaded. */
 export const CLASS_REGISTRATIONS_MANIFEST_LOADED = true;
 
 /** Total @RegisterClass decorated classes discovered in dependency tree */
-export const CLASS_REGISTRATIONS_COUNT = 1000;
+export const CLASS_REGISTRATIONS_COUNT = 998;
```
rkihm-BC: **[Medium] Manifest regenerated without `@memberjunction/server`.** The count dropped 1000→998 and the `@memberjunction/server (3 classes)` block — `MJEntityPermissionEntityServer`, `MJTenantFilterMiddleware`, `RateLimitMiddleware` (two of them security middleware) — was removed, alongside the intended Databricks addition. This looks regenerated in a tree where that package's dist wasn't built. Please re-run `mj codegen manifest` in a full build so the only change here is the Databricks class registration.

packages/ExternalDataSources/Providers/Databricks/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "peerDependencies": {
+    "@databricks/sql": "^1.8.0"
+  },
+  "peerDependenciesMeta": {
+    "@databricks/sql": {
+      "optional": true
+    }
+  },
+  "devDependencies": {
+    "@databricks/sql": "^1.8.0",
```
rkihm-BC, line 32: **[Low] `@databricks/sql` full runtime as a devDep is installed in the monorepo dev install.** The optional-peer setup correctly keeps it out of *downstream consumers'* installs, but because the SDK ships its own types (no `@types/*` alternative like Snowflake), the runtime is pulled into the root `npm ci` here (arrow / thrift / openid-client / native kernels). Worth correcting the driver-header / changeset wording that implies parity with the `snowflake-sdk` avoidance; the trade-off itself may well be acceptable.

rkihm-BC (approved):
## Re-review — Databricks SQL Warehouse driver ✅

Both Aug 3 findings are fixed and I ran the suite in a built tree. Approving.

- 🟡 **Manifest drift** — fully resolved. The `@memberjunction/server (3 classes)` block (incl. the two security middlewares) is present, and the diff vs `next` is now Databricks-only (+1) (`mj-class-registrations.ts:1398, 2380-2382`).
- 🔵 **Dependency wording** — corrected in the driver header and changeset; no more implied parity with the `snowflake-sdk` types-only avoidance.
- ✅ **Definition of Done** — built the dependency chain and ran the suite: **39 passed, 9 integration skipped** (count grew 31→39 across the review commits). Tests are genuine — a fake `DBSQLClient` drives the real execute path, plus dialect/backtick-doubling, `LIMIT ALL` offset paging, PAT vs OAuth-M2M branching, `:name` binds, read-only screening, and introspection assembly.

The bare-object-name schema resolution (`sessionInitFor` setting `initialCatalog`/`initialSchema`, `0c0139c1c`) is correct — fully-qualified `catalog.schema.object` still overrides, introspection paths self-qualify against `information_schema`. The `isAuthError` 403-exclusion (authorization ≠ authentication; a reconnect can't fix it) and the read-only `::`-cast-preserving regex are both right. Clean, faithful driver that follows the Snowflake precedent. Nice work.

PR #3375, opened 2026-07-31
fix: NG0201 in BaseResourceComponent, stale SP cascade refs, React CDN race, BaseEngine cache fingerprint mismatch

jordanfanapour:
## Summary

A collection of bug fixes across Angular, migrations, React runtime, and core caching infrastructure:

### 1. BaseResourceComponent NG0201 Error
- `TabContainer.updateTabDisplayName()` creates lightweight temp instances via `runInInjectionContext()` to resolve tab display names
- `ChangeDetectorRef` is component-scoped (only available via `createComponent()`), so these temp instances crashed with NG0201
- **Fix**: Make `ChangeDetectorRef` injection optional with a null guard on `RefreshView()`

### 2. Stale `spDeleteAIPrompt` / `spDeleteAIConfiguration` Stored Procedures
- The prior fix (V202607281712) corrected `spDeleteAIAgentRun` but missed these two procedures, which also cascade through `AIPromptRun` and referenced the dropped `AgentRunID` column
- **Fix**: Regenerate both SPs to remove the stale column references

### 3. React CDN Script Execution Order Race Condition
- ReactDOM's UMD factory captures `window.React` at execution time — loading React, ReactDOM, and Babel in parallel with async scripts caused an intermittent (~5%) race where ReactDOM executed first and `createRoot` was permanently broken
- **Fix**: Load React first (await), then ReactDOM + Babel in parallel; clear `initializationPromise` on destroy for proper retry; add post-bootstrap `createRoot` validation
- Includes 9 unit tests covering load-order guarantees and race condition simulation

### 4. BaseEngine Cache Callback Fingerprint Mismatch (from PR #3366)
- `RegisterCacheChangeCallbacks()` and `syncLocalCacheForConfig()` generated fingerprints without `IgnoreMaxRows`, but the data-loading methods cached with `IgnoreMaxRows: true` — the `imr:1` segment mismatch meant cross-server Redis pub/sub invalidation callbacks silently never fired
- **Fix**: Extract shared `BuildRunViewParamsForConfig()` method used by all four call sites (`LoadSingleEntityConfig`, `LoadMultipleEntityConfigs`, `RegisterCacheChangeCallbacks`, `syncLocalCacheForConfig`)
- Includes 13 regression tests for fingerprint consistency across all call sites
- Tracked remaining latent fingerprint issues (dataset collision, RLS mismatch) in #3376

## Changed Files
- `packages/Angular/Explorer/shared/src/lib/base-resource-component.ts`
- `migrations/v5/V202607302040__v5.51.x__Fix_spDeleteAIPrompt_And_spDeleteAIConfiguration_Stale_AgentRunID.sql`
- `packages/Angular/Generic/react/src/lib/` (component, adapter service, bridge service)
- `packages/React/runtime/src/utilities/library-loader.ts` + tests
- `packages/MJCore/src/generic/baseEngine.ts` + fingerprint consistency tests
- `.changeset/steady-engines-align.md`

## Test Plan
- [x] MJCore builds cleanly (`cd packages/MJCore && npm run build`)
- [x] Full MJCore test suite passes (103 files, 1594 tests)
- [x] React runtime tests pass (9 new load-order tests)
- [x] Fingerprint consistency tests pass (13 new tests)
- [ ] Verify cross-server cache invalidation via Redis pub/sub
- [ ] Verify React component rendering in MJ Explorer after CDN load
- [ ] Verify AI Prompt / AI Configuration deletion works without SQL errors


cadam11:
Heads up on a duplicate: the BaseEngine fingerprint fix in this PR (31f38be5, a8fc549c, d4947ed7) overlaps your open #3366, which carries an earlier version of the same change (same changeset file, same baseEngine.ts edit). This PR is ahead of it, it adds the consistency test and the BuildRunViewParamsForConfig alignment.

One of the two needs to own the fix before the batch closes at 12pm CT today. If this PR rides the 5.51.0 rebuild, close #3366 at the same time, otherwise it sits in the parked queue and gets merged on top of its own sibling when the freeze lifts Monday.

PR #3377, opened 2026-07-31
docs(MJCore): show how to CREATE an IS-A child, not just how IDs propagate

AN-BC:
Docs and JSDoc only. No behaviour change.

## The gap

`isa-relationships.md` is thorough about the model — discovery, disjoint vs overlapping, delete orchestration, CodeGen integration — and near-silent about the first thing a developer actually does: create one.

Every creation example in the file is about **UUID propagation**. Not one sets a field belonging to a parent. And the natural-but-wrong mental model — *create the parent, then attach a child to it* — isn't warned against anywhere.

## Why I think it's worth fixing rather than just being more careful

Working on `bizapps-orders` I concluded, wrote down, and committed the claim that **IS-A children could not be created through the object model at all**. It went into a commit message and a build-inventory doc as a shipped limitation.

They can, at any depth. What I'd done was create the parent `Product` first and then try to attach an `EventProduct` with the same id. The parent save failed on a NOT NULL column the child never set; the child looked complete; it read as a framework limitation. I also built two supporting theories that were wrong — that `IsVirtual=1 AND AllowUpdateAPI=1` was a CodeGen defect (it's the deliberate marker for an IS-A parent field, per `manage-metadata.ts`), and that a 30-field entity against an 11-parameter proc proved the save path couldn't work.

The part that convinced me this is a docs problem: **a previous author had reached the same wrong conclusion in the same file months earlier**, and left a comment saying `BaseEntity` "cannot express that". Two people independently inventing the same limitation is a signal about the documentation, not the readers.

## What this adds

**`isa-relationships.md`** — a new *Creating a Child Record* section:

- A worked example setting fields from **three levels** on one child object and saving once, with the rule stated plainly: fields route to whichever entity owns them, no depth limit.
- The failure that looks like a bug — a NOT NULL column on an **ancestor** table — with the v5.50.0 message that now names it, and what the pre-5.50.0 silence looked like so anyone on an older version recognises it.
- The anti-pattern, with code, because it's what people try first: `NewRecord()` starts a **new chain** rather than adopting an existing parent row.
- A note that attaching a child to an already-saved parent is a genuinely different operation the save path doesn't model — relevant when the parent is frozen by a trigger.
- Three troubleshooting rows, including the silent-`false` symptom keyed to the version that fixed it (#3280).

**`BaseEntity.Save` JSDoc** — the same guidance at the call site, because that's where someone looks when a save returns `false`, not in a doc they don't know exists.

**`packages/MJCore/readme.md`** — had no mention of IS-A at all. Now carries a short section under `BaseEntity` with the example, the two traps, and a pointer to the full document.

## Note on #3280

The pre-5.50.0 silent-`false` behaviour is exactly what your PR #3280 fixed. This documents it from the reader's side: what the symptom looked like, and which version to be on. The two go together — #3280 makes the failure speak, this makes the correct usage discoverable before you hit it.

### DIFF
--- a/packages/MJCore/docs/isa-relationships.md
+++ b/packages/MJCore/docs/isa-relationships.md
@@ -500,9 +500,90 @@ WHERE Name = 'Persons';
 
 No other changes are needed — the existing IS-A infrastructure handles everything else automatically.
 
+## Creating a Child Record
+
+**Create the CHILD and set fields from anywhere in the chain. One save writes every table.**
+
+You do not create the parent and then attach a child to it. You create the child, set both its own
+fields and its ancestors', and save once — `BaseEntity` routes each field to the entity that owns it
+via `EntityInfo.ParentEntityFieldNames`, saves the chain root-first, and shares one primary key
+across every level.
+
+```typescript
+// Webinar IS-A Meeting IS-A Product
+const webinar = await md.GetEntityObject<WebinarEntity>('Webinars', contextUser);
+webinar.NewRecord();
+
+// Webinar's own fields
+webinar.RecordingURL = 'https://example.com/rec/123';
+webinar.MaxViewers   = 500;
+
+// Meeting's fields — one level up
+webinar.StartTime    = new Date('2026-03-01T15:00:00Z');
+webinar.Location     = 'Online';
+
+// Product's fields — two levels up. Set on the SAME object.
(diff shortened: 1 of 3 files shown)


AN-BC:
@cadam11 this is documentation only but does touch one code file. Merge into 5.50 LTS if you are ok with it.

PR #3378, opened 2026-07-31
feat(search-engine): make the SearchScope permission resolver replaceable

cmbrock-BC:
## Summary

`SearchEngine` authorizes every search through `SearchScopePermissionResolver`, which answers from `__mj.SearchScopePermission` rows keyed by `UserID` or by one of the user's MJ Roles. That covers MJ's own permission model completely — but it is not the only shape a permission model can take, and until now it was the only one the search path could consult.

A consumer whose entitlements are neither a user nor an MJ Role has no row that can express them. Its grants are therefore invisible to the check that actually runs, and the failure is silent in the worst way: the grant is configured, an administrator can see it in the admin UI, and the search simply returns nothing. Because the resolver was a module-level singleton imported directly by `SearchEngine`, the only remedies were to project the consumer's model into `SearchScopePermission` as derived per-user rows — permission state that can drift from the thing it derives from, with a window where the entitlement exists and the access does not — or to fork the search path.

## Changes

**The seam**

- `SearchScopePermissionResolverBase` — the abstract contract registrations bind to.
- `SEARCH_SCOPE_PERMISSION_RESOLVER_KEY` — the ClassFactory key. There is exactly one resolver per deployment (a consumer *replaces* the policy rather than selecting among several), so a single shared key with priority ordering is the right shape, and it keeps the registry free of the keyless-registration warning.
- `GetSearchScopePermissionResolver()` — returns the highest-priority registration, falling back to MJ's own.

Resolution happens **per call** rather than being cached at module load. A registration made during application startup would otherwise be missed depending on import order — a failure mode that presents as "my resolver works in tests but not in the server", which is expensive to diagnose. The class is stateless and construction is trivial, so there is nothing to gain by caching.

**All five call sites, not just `SearchEngine`**

| Call site | What it authorizes |
|---|---|
| `SearchEngine.searchOneScope` | the engine's own per-scope check |
| `SearchKnowledgeResolver` | the single-scope check |
| `SearchKnowledgeResolver` | the visible-scope-list filter |
| `SearchKnowledgeStreamResolver` | the streaming variant |
| `__Scoped_Search` core action | the agent-facing path |

This distinction matters more than it sounds. A seam honoured on some paths and not others is **worse** than no seam, because the resulting behaviour is contradictory rather than merely absent — the same grant would authorize a search issued one way and silently deny it issued another. The core action is the path an agent takes, so an override that did not reach it would be invisible to exactly the callers most likely to need one.

**Intended shape for an override** — compose with the stock resolver rather than replace its logic:

```ts
@RegisterClass(SearchScopePermissionResolverBase, SEARCH_SCOPE_PERMISSION_RESOLVER_KEY, 10)
export class MyResolver extends SearchScopePermissionResolver {
    public override async ResolveEffectivePermission(input: ResolvePermissionInput) {
        const stock = await super.ResolveEffectivePermission(input);
        if (stock.Allowed) return stock;      // never narrow what MJ already granted
        return this.myOwnGrantCheck(input);   // only ever widen
    }
}
```

## Compatibility

**Nothing changes for existing consumers.** MJ's resolver registers itself as the default, so behaviour is identical when nothing else is registered. `DefaultSearchScopePermissionResolver` is retained and still exported so existing imports keep compiling; it is marked `@deprecated` because it always yields MJ's own implementation and therefore bypasses any registered override.

The failure posture is unchanged and worth restating for anyone writing an override: `SearchEngine` treats a resolver throw as **denied**, never as allowed. An override that cannot reach its own store must not accidentally open a scope.

`integration-test-suite`'s SR6 check deliberately stays on the stock resolver, with a comment explaining why: it asserts a property of MJ's *own* policy, and a consumer's registered override is entitled to widen, so resolving through the seam would make the assertion depend on whichever deployment the suite runs against.

## Tests

- **7 new unit tests** covering the default, the fallback when nothing is registered, an honoured registration, late registration (imperative, because `@RegisterClass` evaluates at module load and so cannot demonstrate lateness), composition with `super`, the deprecated constant, and that a subclass of the stock resolver satisfies the base contract.
- `packages/SearchEngine`: **434/434 passing**, `tsc --noEmit` clean.

## Verification gap — please confirm in CI

I could not compile `@memberjunction/server` or `@memberjunction/core-actions`. My checkout has no `node_modules`, and a borrowed one resolves `@memberjunction/search-engine` to a different worktree's copy that lacks the new exports, so a typecheck there would have been misleading rather than reassuring.

What I verified instead: the emitted declarations contain all three new symbols, and each renamed variable at those four call sites is used **only** for `ResolveEffectivePermission`, which the abstract base declares. The change at each is a rename of one identifier. The integration tier was also not run for the same reason.


.changeset/pluggable-search-scope-permission-resolver.md
```diff
@@ -0,0 +1,40 @@
+---
+"@memberjunction/search-engine": minor
```
AN-BC: There are no migrations or metadata changes here so this is purely `patch`

AN-BC: @cmbrock-BC migration or metadata is what drives `minor` if just code it is `patch` unless we're making a breaking change to a public facing API in which case, technically, we'd have to do a `major` however, if something was very unlikely to be used externally we'd call that `minor` but here I think `patch` is reasonable.

packages/SearchEngine/src/permissions/SearchScopePermissionResolver.ts
```diff
@@ -103,6 +103,52 @@ function highestLevel(a: SearchScopePermissionLevel, b: SearchScopePermissionLev
...
+ * whose entitlements are neither a user nor an MJ Role — a per-tenant capability grant, say —
+ * has no row that can express them, and its grants are invisible to the check that actually
+ * runs on every search.
+ *
+ * Rather than have such a consumer project its model into `SearchScopePermission` as derived
+ * per-user rows — which works, but creates permission state that can drift from its source —
+ * it can register a subclass:
+ *
+ * ```ts
+ * @RegisterClass(SearchScopePermissionResolverBase, SEARCH_SCOPE_PERMISSION_RESOLVER_KEY, 10)
```
AN-BC: Very close to a good use of `RegisterClass` but **not quite**.... The way to handle priorities is through sub-class compilation/execution priority. Hardcoded priorities are bound for conflicts and are brittle. So instead, what we do is exactly what you show, but for custom resolvers **OUTSIDE** of MJ, the correct pattern is to subclass the built in MJ resolver.

This guarantees the right RegisterClass priority implicitly, so you don't even pass in a priority. The priority field is still there for edge cases where this isn't possible, but in this case it is a perfect example.

Study how this works where for example in `MJCoreEntitiesServer` we sub-class something from `MJCoreEntities`. In the server side infra, that server package is imported and therefore ClassFactory CreateInstance automatically gets the server side sub-class.

Does this make sense @cmbrock-BC? 

Good job picking up on this pattern using `@RegisterClass`, this actually makes it easier to use :)

packages/SearchEngine/src/permissions/SearchScopePermissionResolver.ts
```diff
@@ -128,7 +174,8 @@ function highestLevel(a: SearchScopePermissionLevel, b: SearchScopePermissionLev
  * (e.g., to make a row exist before granting it later) and would create
  * surprising lockouts when a user joins a role.
  */
-export class SearchScopePermissionResolver {
+@RegisterClass(SearchScopePermissionResolverBase, SEARCH_SCOPE_PERMISSION_RESOLVER_KEY)
+export class SearchScopePermissionResolver extends SearchScopePermissionResolverBase {
```
AN-BC, line 186: So what would happen, building on earlier comment is external packages like `BCSaaS` etc would simply import `SearchScopePermissionResolver` and subclass it which **guarantees** their RegisterClass call is a higher priority than the one here. 

@rkihm-BC  this would be really good to teach all new fellows, the concept of dynamic plugins, how ClassFactory/RegisterClass work together and the pattern I'm sharing here with Colin.

AN-BC (commented):
Nice work @cmbrock-BC overall, there are some comments here that require rework. I think this is a small enough change that we can consider it `patch` and if @cadam11 agrees this bug fix/extensibility pattern can make it into the `5.x LTS` branch, but otherwise it will go to `6.x`

Craig - this is a very lightly used area of the framework and having this soon is important for Betty so that we can update `BCSaaS` to use this new extensibility hook and then build out what it needs to support Betty + other SaaS systems.

.changeset/pluggable-search-scope-permission-resolver.md
cmbrock-BC: got it, I'm still getting a hang of what each is.

cadam11:
Good seam, and the problem it solves is real. But it's a feature, and under the certification freeze features wait until 6.x opens Monday. Merging to next right now would put it into the 5.51.0 LTS candidate, and a replaceable authorization seam in the search path is exactly the kind of change we don't want entering a candidate mid-certification (that area produced the last two hot fixes).

It's green and reviewed, so it's ready the moment next reopens Monday. Parking it until then per the bug-fix guidance I posted this morning.

PR #3379, opened 2026-07-31
fix(realtime): run scoped-anonymous realtime work as the system user

### LINKED ISSUE #3371: Realtime relayed tool dispatch runs as the anonymous visitor — scoped magic-link sessions cannot run agents
## Summary

The realtime relayed-tool path runs the delegated agent **as the anonymous magic-link visitor**, so a scoped anonymous session cannot run an agent at all. Any consumer whose anonymous users reach `invoke-target-agent` stops dead at the first tool call.

MJ already has the elevation mechanism for exactly this — it is just gated on a claim that only MJ's own public web widget issues.

Affects **5.48.0 and 5.50.0** (the two files below are byte-identical between them).

## Repro

1. Mint a magic-link invite with `IdentityMode='anonymous'` for a role that grants only the four relay entities (`MJ: Conversations`, `MJ: AI Agent Sessions`, `MJ: Conversation Details`, `MJ: AI Agent Session Channels`) — i.e. no grant on the AI run entities.
2. Redeem it for an Anonymous Session JWT. The token carries `mj_anon` and a `MagicLinkScope`, but **no `mj_widget_id`** claim.
3. Prepare + launch a realtime client session (succeeds).
4. Call `ExecuteRealtimeSessionTool` with `toolName: 'invoke-target-agent'`.

**Expected:** the target agent runs and returns a tool result.

**Actual:** the mutation returns a payload with `success: false`:

```
Delegation failed: User: Anonymous (ID: …, Email: anonymous@magic-link.local)
Does NOT have permission to Read MJ: AI Agent Runs records.
```

Confirmed by a controlled flip — elevating only the dispatch to the system user turns the same call green, and reverting turns it red again. Run twice in each direction.

> ⚠️ **The mutation does not throw.** It returns a JSON string whose `success` is `false`. A test asserting `data.ExecuteRealtimeSessionTool != null` reports a **pass** for a call that failed. This produced a false "ready to ship" verdict on our side before we parsed the payload.

## Root cause

`RealtimeClientSessionResolver.ExecuteRealtimeSessionTool` resolves the caller and passes it straight through:

```ts
const { contextUser, provider } = this.requireUserAndProvider(userPayload, providers);
const session = await this.loadOwnedActiveSession(agentSessionId, contextUser, provider);
…
await this.clientSessionService.ExecuteRelayedTool({ … }, contextUser, provider);
```

That `contextUser` reaches `base-agent`, which does `GetEntityObject('MJ: AI Agent Runs', params.contextUser)` and then `Save()` → `CheckPermissions(Create, true)` → denied.

### Three call sites share this root cause

| # | Call site | Symptom |
|---|---|---|
| 1 | `RealtimeClientSessionResolver.ExecuteRealtimeSessionTool` → `ExecuteRelayedTool` | **Hard failure** — no delegation, so no completed session |
| 2 | `SessionManager.finalizeObservabilityRuns` → `RealtimeClientSessionService.FinalizeCoAgentRun` | **Silent** — `catch` + `LogError`; co-agent run never finalized |
| 3 | `RealtimeClientSessionService.createCoAgentRun` | **Silent** — logs, returns null; every later `RelayRealtimeUsage` then drops its usage delta for want of a `promptRunID` |

2 and 3 only log, so the observability loss is invisible unless you read the API log. In our logs this showed up as a repeating `has no co-agent promptRunID (observability run creation was skipped or failed) — usage delta dropped`.

## Why the existing elevation does not apply

`realtimeWidget/widgetGuestElevation.ts` exports `resolveWidgetGuestRunContext` (swaps in `UserCache.Instance.GetSystemUser()`) and `elevateUserPayload`, and `RunAIAgentResolver` already uses both so a widget guest can run an agent:

```ts
const widgetElevation = await resolveWidgetGuestRunContext(userPayload, p);
const effectiveUserPayload = widgetElevation ? elevateUserPayload(userPayload, widgetElevation.elevatedUser) : userPayload;
```

But the gate returns `null` for any anonymous session that is not a public-web-widget guest:

```ts
const widgetId = guest?.WidgetGuestContext?.WidgetID;
if (!guest?.IsMagicLinkAnonymous || !widgetId) {
  return null;
}
```

`WidgetGuestContext.WidgetID` is populated solely from the `mj_widget_id` JWT claim (`context.ts` → `extractWidgetGuestContext`). Magic-link anonymous sessions minted outside MJ's widget flow never carry it.

Within `RealtimeClientSessionResolver` the helper *is* imported, but its only use is `resolveWidgetVoiceCapSeconds` — reading a duration cap. The tool dispatch is never elevated.

## Suggested direction

Key the elevation on a scoped anonymous session (`MagicLinkScope.ResourceID`) rather than requiring `WidgetID`, and resolve the effective run identity **once per session** so all three call sites inherit it, rather than patching the dispatch alone.

Two constraints worth stating explicitly:

- **The ownership/RLS gate must keep running as the caller.** `loadOwnedActiveSession` should stay on `contextUser`; the change is about who the delegated agent *runs as* after ownership is proven, never about who can reach which session. We verified elevation-of-dispatch-only leaves cross-tenant isolation intact (15/15 isolation assertions still pass).
- **Elevation must fail closed.** If no system user is available, behave as today rather than proceeding unelevated-but-assumed-elevated.

### Why "just grant the anonymous role the AI run entities" is not an acceptable workaround

`AIPromptRun.Messages` stores the rendered system prompt. For a conversational-assessment product that is the persona plus any injected rubric criteria — i.e. handing the candidate the thing they are being assessed against. Consumers with an anonymous surface need this to work *without* granting read on the AI run entities.

## Version check

- `packages/MJServer/src/resolvers/RealtimeClientSessionResolver.ts` — identical 5.48.0 → 5.50.0
- `packages/MJServer/src/realtimeWidget/widgetGuestElevation.ts` — identical 5.48.0 → 5.50.0
- `git log v5.48.0..v5.50.0` touches neither file
- `widgetGuestElevation.ts` has a single commit in its entire history
- No fix in flight on any branch

## Notes

Downstream tracking issue: MemberJunction/bizapps-caliber#78.

Currently worked around downstream by an unversioned patch to `node_modules`, which is invisible to CI and evaporates on `npm install`.

`packages/MJServer/src/__tests__/widgetGuestElevation.test.ts` already exists and would be the natural place to cover the scoped-anonymous-without-`WidgetID` case.

SDesai-BC:
A scoped anonymous magic-link session could not run an agent at all: the realtime relayed-tool path executed the delegated run as the anonymous visitor, whose role deliberately holds only the four relay entities (`MJ: Conversations`, `MJ: AI Agent Sessions`, `MJ: Conversation Details`, `MJ: AI Agent Session Channels`) — so base-agent's `MJ: AI Agent Runs` create was permission-denied and every `invoke-target-agent` call died. Around that hard failure sat a ring of *silent* losses under the same root cause: the co-agent observability run was never created at session start, transcript/tool-turn mirrors and usage deltas onto the prompt run were dropped, owner-initiated closes never finalized the runs, and — found during a second sweep — `UploadRealtimeRecording` failed at its first step (the `MJ: AI Agents` read), so an anonymous session's recording was never stored.

Closes #3371

## Root cause

MJ already has the elevation mechanism, but its gate only opens for MJ's own public web widget:

```ts
// widgetGuestElevation.ts — resolveWidgetGuestRunContext
const widgetId = guest?.WidgetGuestContext?.WidgetID;
if (!guest?.IsMagicLinkAnonymous || !widgetId) {
  return null;
}
```

`WidgetID` comes solely from the `mj_widget_id` JWT claim that only MJ's widget flow mints. Magic-link anonymous sessions minted by downstream consumers never carry it, and within `RealtimeClientSessionResolver` the elevation helper was only used to read the voice duration cap — the tool dispatch passed `contextUser` straight through to `ExecuteRelayedTool`.

## What this does

New export `resolveScopedAnonymousRunUser(contextUser)` in `widgetGuestElevation.ts`: returns the system user when the caller is a scoped anonymous magic-link session (`IsMagicLinkAnonymous` + `MagicLinkScope.ResourceID`, **no** `WidgetGuestContext.WidgetID`), else the caller unchanged. Fails closed — no system user available means the caller is returned and the request fails exactly as today, never unelevated-but-assumed-elevated.

Applied at every seam where the realtime path writes entities the relay role cannot hold, all in MJServer: the `ExecuteRelayedTool` dispatch, both `AppendPromptRunMessage` sites (transcript mirror + tool turn), `AccumulatePromptRunUsage`, the `PrepareClientSession` call (observability run creation), the delegated-artifact junction link, `SessionManager.finalizeObservabilityRuns`, and both recording mutations (`UploadRealtimeRecording` / `UploadRealtimeRecordingSegment`). **Ownership and authorization never move**: `loadOwnedActiveSession`/`loadOwnedSession`, `assertCanRunTarget`, the runtime-override gate, and the recording consent gate all still run as the caller — the change is who the work *runs as* after ownership is proven, per the issue's stated constraint.

## Design decisions worth defending

- **Public web-widget guests are excluded from the new gate.** Their seeded role writes run rows under the guest principal, and the `Widget Guest: Own Agent Runs` RLS read filter exists precisely because of that (its own header says runs are "still written under the guest principal" on the voice path). Elevating them would silently change what that filter matches. Their path is byte-identical before and after this PR.
- **Identity resolved per request, not once per session.** The issue suggested resolving once per session; the JWT claims are stable across a session's requests, so per-request resolution through the one shared function is equivalent without persisting elevation state into the session config.
- **Attribution stays the visitor's.** `PrepareClientSession` still receives `UserID: contextUser.ID` (run rows attribute to the visitor; memory scope unchanged), and the hidden anchor detail now stamps `session.UserID` instead of the writer — identical for every non-elevated caller, whose ownership is already proven.
- **`RunAIAgentResolver` is deliberately untouched.** The widget path can elevate there because the agent id is pinned by the widget instance; for a non-widget anonymous caller the agent id is client-supplied, so elevation would let an anonymous visitor run arbitrary agents as the system user. On the realtime path the target is server-authoritative (session config, `CanRun`-gated at start), which is why elevation is safe there.
- **The IT bundle does not import `@memberjunction/server`.** Its barrel evaluates server config at import time, which would couple the whole suite's loadability to it. The bundle instead proves the permission-reality contract with the real permission engine (routing is pinned by 21 MJServer unit tests); the anonymous principal is synthesized exactly as `buildMagicLinkSessionUser` constructs it from a JWT, whose parsing `magicLink.test.ts` already covers.
- **IT68's suite membership ties at `Sequence: 32`** — the same late-addition pattern already on `next` (IT66/IT67 tie at 32, IT31/IT48 at 27). Server-transport members must precede the client-transport block, which starts at 33; observed run order placed IT68 at 33/54, before the first client member.

## Evidence

| Check | Result |
|---|---|
| MJServer unit suite (`packages/MJServer`) | 741 passed, 56 skipped, 0 failed — 21 new tests: gate matrix, fail-closed, per-seam routing, ownership-still-rejects, consent-still-refuses |
| Suite package unit tests (parity + count gates) | 139 passed |
| `IT68 - Scoped-Anonymous Elevation Permission Contract` (live DB) | 5/5 — anon denied on `MJ: AI Agent Runs`, system user sufficient, visitor attribution preserved, finalize + usage contrast (silent no-op as anon, real write as system) |
| Full `Integration Tests — Deterministic` suite | 54/54, run twice (before and after review fixes), ~123s |
| Two-axis review (standards + spec sub-agents) | Spec: no missing requirements; standards findings (swallowed `Delete()`, lazy mock typing, per-call log spam) fixed in-branch |

The issue's own downstream verification (controlled flip of the dispatch identity, run twice in each direction, 15/15 cross-tenant isolation assertions) established the root cause this PR keys off.

## Merge/deploy notes

- **Metadata is not live until pushed**: the IT68 `MJ: Tests` record and suite membership under `metadata-optional/integration-test/` need `mj sync push --dir metadata-optional/integration-test` in each environment that runs the suite. Without it, IT68 simply doesn't exist there; nothing else in the PR depends on it.
- **Pre-existing drift repaired in passing**: the IT31 suite-membership row in `.integration-suite.json` had no `primaryKey`, so every push from a fresh checkout minted a new id and hit the unique `(SuiteID, TestID)` constraint, failing the whole push. It is now pinned to the actual DB row id (`3F247D8B-…`, verified against the live DB before pinning).
- **Known residuals, deliberately not elevated** (documented in the second commit): `SaveSessionChannelArtifact` (whiteboard artifacts are visitor content a consumer can grant without leaking prompts; no anonymous consumer uses interactive channels), the server-bridged topology (`WireBridgeRealtimeSession` — authenticated hosts), and the `RunAIAgentResolver` text path (see above).
- Downstream (MemberJunction/bizapps-caliber#78) currently works around only the dispatch leg via an unversioned `node_modules` patch; observability, finalize, usage, and recordings remain broken there until this ships in a release.

Three files are uncommitted in the working tree and deliberately not in this PR: `package-lock.json` (npm formatting drift from a fresh worktree install), `.vscode/settings.json` (editor theme churn), and the regenerated `packages/AI/MCPServer` class-registration manifest.


packages/MJServer/src/realtimeWidget/widgetGuestElevation.ts
```diff
@@ -77,6 +77,47 @@ export async function resolveWidgetGuestRunContext(
...
+ * run rows under the guest principal, which the `Widget Guest: Own Agent Runs` RLS read filter
+ * depends on — elevating them here would silently break that read-side control.
+ *
+ * Fails CLOSED: when no system user is available the caller is returned unchanged, so the request
+ * fails exactly as it would today rather than proceeding unelevated-but-assumed-elevated.
+ *
+ * Ownership/RLS gates must NEVER use this — they stay on the caller; this only changes who the
+ * work RUNS AS after ownership is proven.
+ */
+export function resolveScopedAnonymousRunUser(contextUser: UserInfo): UserInfo {
```
AN-BC: @SDesai-BC public class members and exported functions are `PascalCase`

AN-BC (commented):
I see what you're doing here and it applies to realtime sessions. I am ok with including this for now if @cadam11 thinks this is ok for `5.51.0` **but** I think we need a much more thorough PR and review of the ramifications of this kinda of allowed impersonation across all aspects of magic link use as it is both a security concern and an alignment/consistency issue. Please throw together an issue in this repo to capture this comment and link back to this PR and spin up a new CC instance to work on that in depth. That would go into 6.0

One minor change requested, subject to @cadam11 approval you can merge after fixing this and please keep in mind our conventions before sharing a PR.

SDesai-BC:
@AN-BC — agreed on the broader concern, and filed as #3383: "Review the magic-link impersonation/elevation model across all surfaces."

It inventories the two elevation mechanisms that now coexist (widget-guest, which keeps a guest principal under RLS, vs. the scoped-anonymous system-user swap this PR adds), and lists what a review should settle — starting with whether system-user substitution is the right primitive at all, versus granting the anonymous invite role the rights it needs on the AI run entities so nothing is impersonated. This PR took the substitution route because it needed no migration; that tradeoff is worth re-litigating rather than inheriting. Also covered: where the "authorize on the caller, elevate only execution" invariant should live so it can't rot, the in/out surface split, and attribution/memory-scope handling.

On the `5.51.0` condition: @cadam11 confirmed approval to me directly over DM rather than on this PR, so noting it here for the record — @cadam11, a thumbs-up on the PR itself would make the gate visible to anyone reading later.

PR #3380, opened 2026-07-31
fix(explorer,server,ai): process-wide server cache corruption + 9 defects surfaced by the regression suite

CaelebB-BC:
## Summary

**Ten MJ product defects found by running the Explorer regression suite**, plus the review round that followed. Every one is a real bug that affects users, not the harness — reproduced live at 1280×720 against the regression stack. None depend on the suite, so they sit on a branch based directly on `next` and can ship independently of (and ahead of) the suite work in #3033.

> **⚠️ Take this bump urgently if you run MJAPI.** One of these is a **process-wide server cache corruption**: `ResolverBase` mapped GraphQL transport field names onto the data provider's own result rows, which the server cache holds *by reference*. Preparing a single GraphQL response rewrote `__mj_CreatedAt` → `_mj__CreatedAt` **inside the live cache**, and because that cache is process-wide, one response poisoned every later read across all workers — surfacing as `Field _mj__CreatedAt does not exist on <Entity>` from `BaseEntity.SetMany`.

Two of these bugs are worth calling out for *how* they fail:

- **The duplicate-`@HostListener` bug appears twice** in the codebase. Angular collects host listeners into an object keyed by event name, so a second declaration for the same event **silently replaces** the first — no build error, no warning. `ShellComponent` shipped that way, so the <kbd>Ctrl</kbd>/<kbd>Cmd</kbd>+<kbd>/</kbd> command-palette shortcut was dead for its entire life.
- **The `NG0201` flood cost 6.46 GB of heap** in a single suite run — 82% of it, in 4,071,310 copies of one string.

## Scope

| | |
|---|---|
| Commits | **5** (3 fixes + 2 `next` merges) |
| Files | **27** |
| Lines | **+1,483 / −149** |
| Base | `next` — **0 behind** |
| Tests added | **32** across 4 files |
| Changeset | 9 packages, all `patch` |

## Commits

| | |
|---|---|
| `09c9976197` | The original ten fixes |
| `a17a569f11` | Merge `next` |
| `4b551466ad` | **Address review findings** — hardened 6 of the fixes, added the HostListener guard |
| `52ac0480e0` | Merge `next` (69 commits) |
| `b6538a37da` | Don't memoize a *rejected* display-name lookup |

---

### `ResolverBase` — process-wide cache corruption

`FieldMapper.MapFields` renames keys **by mutating** (`obj[mapped] = obj[k]; delete obj[k]`). It was applied directly to the provider's own result objects, which the server cache holds by reference.

```diff
-        for (const r of result.Results) {
-          mapper.MapFields(r);
-        }
+        result.Results = result.Results.map(r => mapper.MapFields({ ...r }));
```

`ArrayFilterEncryptedFieldsForAPI` mutates too, so it must also see the copies — otherwise it strips encrypted values out of the *cached* rows. (`FileResolver` already mapped a spread copy; this brings `ResolverBase` in line.) The review round added an explicit note on why `{ ...r }` — a **shallow** copy — is sufficient given the only post-map mutators are key-level.

### `AIPromptRunner` — a failover banner for failovers that never happened

The "🔄 Trying candidate N/M" banner keyed off `i > 0`, which is **not** "a previous candidate failed": candidates lacking credentials in the current environment are skipped without ever issuing a request, so the first *called* candidate is routinely `i > 0`. Successful first-try calls logged `🔄 Trying candidate 2/315` — **515 such lines and zero real failovers in one run**. Now keyed off an actual prior failure.

---

### `@HostListener` duplicates (2 occurrences) — plus a guard against regression

**`shell.component.ts`** — two `@HostListener('document:keydown')` declarations; the survivor didn't handle the palette chord, so the shortcut silently never fired. **`whiteboard-host.component.ts`** — the same bug for `document:click`, killing the Sees dropdown's outside-click dismissal.

The review round added **`host-listener-uniqueness.test.ts`**, a *source-level* check that each guarded component declares every host event exactly once. That choice is deliberate and worth understanding: the failure mode is a duplicate **declaration**, and a behavioural test can only ever observe the *surviving* handler — it cannot see that a second one clobbered the first. It also guards a concrete merge hazard: **#3380 and #3033 both touch `shell.component.ts`**, so whichever merges second resolves a conflict there, and taking "the other side" would reintroduce the duplicate silently.

### `tab-container` — the NG0201 flood, and why the memo is promise-level

Reading a resource's tab title instantiates its `BaseResourceComponent` subclass against the **environment** injector, which by design cannot supply node-injector-only tokens (`ElementRef`, `ChangeDetectorRef`, `ViewContainerRef`) or component-scoped providers. Affected drivers throw `NG0201` — deterministically, every time. Uncached, this ran on **every tab add and every tab reload** and logged the full `Error` each time: **50,153 console errors** in one run, whose retained Playwright argument handles consumed **6.46 GB** of the runner's heap.

The memo went through three refinements across the review rounds, each fixing a distinct failure:

1. **Memoize the promise, not the value, and register it *before* the first `await`.** Callers are fire-and-forget — a workspace restore runs a synchronous `sortedTabs.forEach(tab => this.createTab(tab))`, so every tab enters before any could have written a resolved value. Caching the value would still instantiate once per restored tab.
2. **Don't memoize "not registered."** A lazy chunk may register the class later, and `ClassFactory` already caches its own null and invalidates on `Register()`.
3. **Don't memoize a *rejection*** (`b6538a37da`). `ClassFactory.tryLazyLoad` awaits each loader uncaught, so a failed chunk fetch rejects `GetRegistrationAsync` itself. Caching that killed display names for the driver for the life of the shell *and* handed every later caller the same rejected promise. Now the memo is deleted and the next ask retries — with no warning, because the driver isn't unusable, the fetch just failed (and the real load path already reports it).

### `omnibar-palette` — three stale-state bugs

- **Keyboard selection may only target rendered rows.** It could previously land on recent rows that weren't on screen.
- **A mode change invalidates the row set outright.** A different provider answers a different question, so the old list isn't a stale approximation — it's the wrong list. (Typing `Admin` over a seeded `/` switches Go-to-App → Global Search; the app list must not survive.) Rows for the *same* mode are kept while the next fetch is in flight — conventional palette behaviour, avoids per-keystroke flicker.
- **A failed fetch must still settle its generation.** Letting the rejection escape left the spinner up forever *and* surfaced as an unhandled rejection, since the caller invokes it as `void fetchSuggestions(...)`.

---

### `dashboards/public-api.ts` — now largely upstream

The manifest generator discovers `@RegisterClass` components by walking the package entry point, so an unexported resource is absent from the eager manifest, its driver class resolves to nothing, and the tab renders **"This view isn't available in the running build."** That's what happened to `FeaturePipelinesResourceComponent`.

**`next` has since fixed this independently, landing on the identical export.** After the merge, this PR's only remaining change to the file is the **comment documenting the invariant** — kept deliberately, because two people independently hitting the same bug is the argument for recording why the export must never be removed. The regenerated `ng-bootstrap` manifest is now byte-identical to `next` and has **dropped out of this diff entirely**.

### Tree-shaking loader robustness

`ai-dashboards.module.ts`, `archiving-dashboards.module.ts`, `clustering.module.ts` — hardened the `@RegisterClass` loader-call pattern so registration survives aggressive tree-shaking (`sideEffects: false`).

---

### `file-upload` — requesting wire names as entity fields

The upload path requested the `_mj__CreatedAt` / `_mj__UpdatedAt` transport aliases. GraphQL reserves the `__` prefix, so MJ exposes `__mj_*` columns under a `_mj__*` alias — but that alias is the **wire** name, not an entity field name. The response goes straight to `LoadFromData()`, where `SetMany` rejects unknown keys. Nothing consumed the timestamps anyway.

---

## Accessibility & layout

| File | Defect |
|---|---|
| `query-browser-resource.*` (html/ts/css) | Splitter had no role or keyboard affordance. The review round moved this from markup-only attributes into a real keyboard-operable implementation (TS handlers + focus styling). |
| `connections.component.*` (html/ts) | Entity-map detail rows were click-only — now `role="button"`, `tabindex`, `aria-label`, Enter/Space, with the handler in TS. |
| `view-selector.component.css` | Row actions revealed on `:hover` only → added `:focus-within` |
| `home-dashboard.component.css` | Same hover-only problem on pin-card actions |
| `ps-catalog.component.css` | `flex-shrink: 0` on cards. The default shrink compressed the guide card below its own content at 1280×720 (46px vs 166px), and the overflow landed **behind** the `position: relative` gallery cards — making filter chips **unclickable for real users**, not just the agent |

The hover-only cases are a broader pattern: roughly 20 other stylesheets still gate row actions on `:hover` alone. Not fixed here — worth a sweep.

---

## Tests (32)

| File | Tests | Covers |
|---|---:|---|
| `omnibar-palette.component.dom.test.ts` | 13 | Generation stamping, pending guard, mode-change invalidation, rendered-rows-only selection, failed-fetch settling |
| `tab-container.display-name.dom.test.ts` | 11 | Promise-level memoization (concurrent restores share one attempt), one warning per driver class, late registration recovers, rejected lookup recovers, no `Error` object passed to `console.*` |
| `ResolverBase.transportMapping.test.ts` | 6 | Mapping produces copies; the provider's original rows are untouched; encrypted-field filtering sees the copies |
| `host-listener-uniqueness.test.ts` | 2 | Each guarded component declares every host event exactly once; the shell still owns the global keydown chord |

The tab-container spec reproduces the real failure mode rather than mocking it — the fake driver's `inject()` field initializer requests an unprovided token, so it throws `NG0201` exactly the way shipping drivers do. The rejection spec installs a genuinely throwing lazy loader (key-scoped, since lazy loaders are global with no unregister API).

## Verification

Merged to `next` (0 behind) and rebuilt after: **236/236 turbo tasks, zero TS errors** across the `ng-explorer-core`, `ng-dashboards`, `server`, and `ai-prompts` dependency chains.

| Package | Tests |
|---|---|
| `ng-explorer-core` | **134 passed** (16 files) |
| `@memberjunction/server` | **752 passed**, 56 skipped |
| `@memberjunction/ai-prompts` | **253 passed** |

I also checked the second `next` merge for fixes that would make anything here redundant: `next` added **zero** `focus-within` rules, touched **zero** `@HostListener` declarations, and never touched `FieldMapper` or transport mapping. Only the Feature Pipelines export overlapped (above).

## Included write-ups (3 docs — defects left unfixed)

- **`element-grounding-interception-bugs.md`** — the investigation behind the `ps-catalog` fix. Three findings: a Computer Use perception bug (fixed in #3033), the PS Studio filter chips (fixed here), and Auth0 consent screens as the dominant suite-flakiness source (tenant config, zero product impact). Also notes that invalid resource routes hang rather than erroring.
- **`explorer-chat-search-panel-missing.md`** — a fully-built conversation-search component is unreachable in Explorer's only chat surface. Medium severity, **not fixed**.
- **`active-tasks-panel-orphaned.md`** — `mj-active-tasks-panel` is declared, exported, and rendered nowhere. Low severity (dead code), **needs a keep-or-delete decision**.

## Suggested review order

1. `packages/MJServer/src/generic/ResolverBase.ts` — highest blast radius; process-wide cache corruption.
2. `packages/Angular/Explorer/.../tab-container.component.ts` — the memo's three exemptions (concurrency, late registration, rejection) are the subtle part.
3. `packages/Angular/Explorer/explorer-core/src/lib/__tests__/host-listener-uniqueness.test.ts` — read the header comment; it explains the merge hazard with #3033.
4. `packages/Angular/Explorer/.../omnibar-palette.component.ts` — three distinct stale-state bugs.
5. `packages/AI/Prompts/src/AIPromptRunner.ts` — logging-only, but it had been actively misleading.
6. The a11y/layout set — mostly mechanical, except `ps-catalog` (real user-facing breakage).

## Merge-order note

`shell.component.ts` is touched by **both this PR and #3033** (there for the CU-A2 `data-mj-ready` readiness beacon, here for the `@HostListener` fix). Whichever merges second will conflict on it — **keep both changes**; they're in different regions and independent. `host-listener-uniqueness.test.ts` will fail loudly if the duplicate listener comes back.


CaelebB-BC:
Broken down into Atomic PR's

PR #3381, opened 2026-07-31
fix(integration-engine): make the FetchChanges timeout configurable

MS-BC:
## Problem

`IntegrationEngine` wraps every `FetchChanges` call in `DEFAULT_OPERATION_TIMEOUTS.FetchChangesMs` (30s), with no way to change it — not per deployment, not per connector.

That penalises exactly the connectors that need a longer budget. A connector that fans out one request per parent record does `BatchSize` requests inside a **single** `FetchChanges` call, so page time scales with batch size *and* with however much concurrency the adaptive controller currently allows. Once a page crosses 30s:

1. the timeout fires,
2. `WithRetry` re-runs the **same** page — paying the full cost again,
3. the resulting failures feed the AIMD controller, which **cuts concurrency**,
4. which makes the next page slower still.

Reported by a production consumer extending `BaseIntegrationConnector` for a high-fan-out source; it was their #1 framework ask.

## Change

Two override sources, resolved **once per entity map**:

```
CompanyIntegration.Configuration {"fetchTimeoutMs": N}   ← per deployment, no code change
  → BaseIntegrationConnector.FetchChangesTimeoutMs        ← per connector, null = framework default
    → DEFAULT_OPERATION_TIMEOUTS.FetchChangesMs           ← unchanged 30s
```

Non-numeric and non-positive values are rejected by the existing `num()` guard in `getConfigOverrides` and fall through to the next source.

- `BaseIntegrationConnector.FetchChangesTimeoutMs` — new `get`, returns `null` by default.
- `getConfigOverrides` — now also reads `fetchTimeoutMs`, alongside the `maxConcurrency` / `rateLimitTokensPerSec` / `rateLimitBurst` / `discoveryTimeBudgetMs` knobs already there.
- `ExecuteEntityMaps` — resolves the value before the paging loop and passes it to `WithTimeout`.

**Fully backward compatible**: with neither source set, behavior is byte-identical to before.

## Tests

4 new precedence cases in `IntegrationEngine.fetch-timeout.test.ts`. Each drives a `FetchChanges` that **never settles** through the real engine, so the timeout is the only thing that can end the call, and asserts the ms value the engine actually applied:

- `Configuration.fetchTimeoutMs` honored (40ms)
- falls back to the connector property when Configuration is silent (55ms)
- Configuration wins over the connector property (35 beats 90000)
- a non-positive override is rejected and falls through to the connector (−5 → 45ms)

Engine suite: **600/600 green**. `turbo build --filter=@memberjunction/integration-engine`: clean.

## Changeset

`minor` — `FetchChangesTimeoutMs` adds a member to the `BaseIntegrationConnector` public surface.

Unit tests alone can't prove the deployment-config value actually reaches `WithTimeout` at runtime, so this was exercised against a **real MJAPI (port 4000) on a real SQL Server database**, driving real `IntegrationStartSync` mutations over GraphQL against a real `CompanyIntegration` (HubSpot, scoped to a single `contacts` entity map).

**1. `Configuration.fetchTimeoutMs` is honored end-to-end.** Set to `1` via `JSON_MODIFY` on the live `CompanyIntegration.Configuration`, then ran a sync. The run failed with:

```
Operation 'FetchChanges(contacts)' timed out after 1ms
```

The engine resolved the value from deployment config and handed it to `WithTimeout` — the exact path unit tests mock. Confirmed by the message naming `1ms`, not the 30s default.

**2. The numeric guard rejects invalid values live.** Set `fetchTimeoutMs` to `0` and re-ran. **No** timeout fired; the run instead surfaced the connector's own real error (`No HubSpot credentials found...`), i.e. the fetch ran to completion under the default and the invalid override was discarded rather than applied as "0ms".

**Not covered live:** `connector.FetchChangesTimeoutMs`. None of the connectors installed on this instance overrides that property, so that rung of the precedence chain remains **unit-covered only** (3 of the 4 tests above). The Configuration source — the one this PR was actually filed for — is live-proven.

### Defect found and fixed while testing live

Reducing the timeout exposed a **false green** that is not specific to this PR but is exactly what a short timeout makes reachable. When a persistent fetch error hits an object the engine cannot page past — cursor paging, or the page-skip budget spent — `ExecuteEntityMaps` broke out of the fetch loop with **no signal**: the map still reported success with whatever it had already collected.

Observed live: a sync whose *very first* page timed out finished `Status=Success`, `errorCount=0`, empty `ErrorLog` — indistinguishable from a clean "nothing changed" run. Silent truncation, on a code path this PR makes easier to hit.

Fixed in `0a07afc` — the abort now emits:

```
"event":"sync.warning","integration":"HubSpot","stage":"contacts",
"code":"FETCH_ABORTED_INCOMPLETE",
"message":"Fetch for 'contacts' stopped at batch 1 after a persistent error and could not continue past it, ..."
```

carrying the object, the batch it stopped at, records fetched before the failure, and the underlying error. The watermark was already held in this case, so the unfetched window is retried next run — the bug was purely that the outcome was invisible. **Verified live**: the warning now appears where nothing was logged before.

No unit test accompanies this warning: the mocked harness constructs no `SyncLogger`, so `logger?.warning(...)` is a no-op there and any assertion would pass or fail for reasons unrelated to the fix. The evidence for it is the live run above.

Engine suite after the fix: **600/600 green**.


MS-BC:
Base synced to current `next` (the local branch was behind its remote; `origin/fix/integration-configurable-fetch-timeout` merged in, no conflicts) so CI re-runs against the current base.

Manually re-ran the Integration suites locally post-merge: **602 passed** and **642 passed**.

PR #3382, opened 2026-07-31
fix(elevenlabs): per-session voice + managed-agent ensure-flow fixes

### LINKED ISSUE #3374: ElevenLabs realtime: no per-session voice — conversation_config_override carries only the prompt
## Summary

The ElevenLabs realtime driver can carry a per-session **system prompt**, but not a per-session
**voice** — so a consumer that models a persona (a named character with its own voice) can express
who the agent *is* but not what they *sound like*. Every session on a given agent shares one voice,
and changing it means editing the agent in the ElevenLabs dashboard.

ElevenLabs' `conversation_config_override` supports `tts.voice_id`. MJ neither sends it nor enables
it, so there is no route to it from configuration.

## Where it stands today

`packages/AI/Providers/ElevenLabs/src/elevenLabsRealtime.ts` (paths as published in 5.48.0):

**1. The per-session override carries only the prompt** — `CreateClientSession`:

```ts
SessionConfig: {
    agentId,
    overrides: { agent: { prompt: { prompt: params.SystemPrompt } } },
    config: params.Config ?? {},
}
```

`params.Config` is passed through opaquely as `config` and never contributes to `overrides`. The
browser-side driver sends `overrides` verbatim as `conversation_config_override` in the
`conversation_initiation_client_data` frame, so anything not in that object cannot reach the session.

**2. The managed agent enables only that one override** — `buildAgentBody`:

```ts
platformSettings: {
    overrides: { conversationConfigOverride: { agent: { prompt: { prompt: true } } } },
}
```

ElevenLabs requires each overridable field to be explicitly allowed, so even if `tts.voice_id` were
sent it would be dropped until the agent enables it.

The driver's own comment states the current boundary: *"`params.Config.llm` (when a string) selects
the agent's underlying LLM; all other provider-level agent settings are deployment concerns (use a
verbatim agent id for full control)."*

## Suggested change

Both edits are in that one file, and the plumbing already exists — the effective config's
per-provider bag (`realtime.voice.providers.<DriverClass>`) is merged and handed to the driver as
`params.Config`, which is exactly how `llm` is already read.

1. **`buildAgentBody`** — additionally enable the TTS override:
   `conversationConfigOverride: { agent: { prompt: { prompt: true } }, tts: { voiceId: true } }`
2. **`CreateClientSession`** — when the config bag carries a voice, include it:
   `overrides: { agent: { prompt: { prompt: params.SystemPrompt } }, tts: { voice_id: <config voice> } }`
   Omitted entirely when no voice is configured, so existing sessions are byte-for-byte unchanged.

Worth deciding as part of this: whether the config key should be a driver-neutral `voice` (each
driver maps it — AssemblyAI already reads `voice`) or an ElevenLabs-specific `voice_id`. The
neutral key seems better for a config bag that is keyed per provider anyway.

## Why the existing workaround is not sufficient

A verbatim `agent_…` id gives full control, so one agent per voice does work — but it moves what is
naturally a *content* decision (this character sounds like this) into deployment, and multiplies
agents by personas. For a product where the persona is the unit of authoring, that does not scale:
ten interviewers with distinct voices means ten dashboard-managed agents kept in sync by hand.

## Impact

Downstream consumer: MemberJunction/bizapps-caliber#83. Caliber models an interviewer Persona with a
`PreferredVoice` field; because it cannot be honoured, the field has been made read-only in the
authoring UI rather than accept a value that is silently ignored.

## Notes

Verified against 5.48.0 and unchanged in 5.50.0. Related but independent:
MemberJunction/MJ#3371 (relayed tool dispatch runs as the anonymous visitor) — if both land in the
same release, one pin bump clears both for downstream consumers.

SDesai-BC:
The ElevenLabs realtime driver could carry a per-session **system prompt** but not a per-session **voice**, so a consumer modelling a persona could express who the agent *is* but not what they *sound like*. Every session on a given agent shared one voice, and changing it meant editing the agent in the ElevenLabs dashboard — one dashboard-managed agent per persona.

Fixing that surfaced two further defects in the managed-agent ensure flow, both pre-existing, both invisible to unit tests because both are round-trip behaviours of the live API. All three are fixed here.

Closes #3374

## 1. Per-session voice

`RealtimeSessionParams.Config.voice` now ships as the `tts.voice_id` conversation-config override, on both the client-direct mint and the server-bridged initiation frame. Both are built by one `BuildSessionOverrides`, so the two topologies cannot drift apart.

`voice` is the **driver-neutral** key AssemblyAI and Inworld already read, rather than an ElevenLabs-specific `voice_id`. The config bag is already keyed per provider, so the key does not need to be provider-disambiguated, and a persona's configured voice is then authored identically whichever provider ends up speaking it. It reaches the driver unchanged from the effective config as `realtime.voice.providers.elevenlabs.voice`.

**Enabling the override on the agent is not sufficient on its own, and that is the non-obvious half.** ElevenLabs drops any override the agent has not explicitly allowed, and the ensure flow's drift check tested only the prompt override:

```ts
if (remoteFingerprint !== fingerprint || !ElevenLabsRealtime.PromptOverrideEnabled(detail)) {
```

An agent provisioned by an earlier version has the prompt override enabled and a matching tool set, so it never re-PATCHed — meaning enabling `tts.voiceId` in the create body alone would have shipped as a **silent no-op on every existing deployment**, with the voice sent and dropped. The check now requires the whole override set. A test drives every override the driver writes and asserts each one, alone, triggers the repair, so adding a future override without teaching the drift check about it fails the build instead of shipping the same gap again.

## 2. The tool-set fingerprint never matched the remote

`ToolSetFingerprint` compared a raw `JSON.stringify`. The platform returns schema keys in its own order **and** materializes its own defaults into the stored schema, so what comes back is a superset of what was sent:

```
stored adds:  dynamic_variable:""  is_omitted:false  required:[]
              isSystemProvided:false  constantValue:""  allowed_values_dynamic_variable:""
```

Once materialized, the stored form can never equal the local form again, so every subsequent ensure saw drift and re-PATCHed the agent. The fingerprint now sorts object keys and drops empty/default entries on both sides, which converges them without having to enumerate and then chase the platform's field list.

Two deliberate choices there. **Arrays keep their order** — an `enum` or `required` list is data, not a set, so reordering one is a real difference and must still register. And pruning empties is an accepted, documented **loss of sensitivity**: a field moving between `false`/`""`/`[]` and absent no longer counts as drift, while any change to a meaningful value still does.

## 3. Find-by-name could fork a duplicate managed agent

ElevenLabs' agent search is eventually consistent, so an agent created moments earlier — by this process or a concurrent one — is briefly invisible to it. Treating a single miss as "does not exist" made the flow create a second agent competing for the same name forever. Observed live, twice, against a real account.

Lookups now retry up to a bounded `MAX_AGENT_LOOKUP_ATTEMPTS` before concluding absence, and only the create path pays that cost (the result is instance-cached). Where duplicates already exist, `PickCanonicalAgent` adopts the **oldest** deterministically, ties broken by agent id, so every process and session converges on one agent instead of PATCHing them alternately.

## 4. Two sessions opening at once forked a duplicate inside one process

Raised in code review and confirmed by a failing test before the fix. The ensure cache stored the **resolved** agent id, so it was only populated once the whole find-create round-trip had finished. Two sessions opening at the same moment against a not-yet-provisioned name both missed the cache, both found nothing, and both created — the intra-process twin of defect 3. A server opening several realtime sessions the moment a new managed agent name appears is the ordinary case, not a contrived race.

The cache now holds the **in-flight promise**, so the second caller joins the ensure already running. Unlike the cross-process race this one is fully closable, because the thing to wait on is right there in the same process.

A rejected ensure is **evicted**, because a rejected promise is a cache entry too: without eviction one transient REST failure would be replayed from memory to every later session for that name, disabling the agent until the process restarted without ever retrying the API. That is a second test, also RED before the change.

**Known residual, deliberately not addressed here.** The retry closes the *eventual-consistency* window — an agent that exists but is not yet indexed — and the in-flight cache closes the intra-process fork. Neither can close true *cross-process* simultaneity: several cold processes ensuring a brand-new agent name at the same instant each find nothing because at that moment there is nothing to find, and each creates. Confirmed live (3 concurrent cold starts in separate instances → 3 agents). This is self-healing rather than permanent — `PickCanonicalAgent` makes every subsequent ensure, in every process, adopt the same one and create nothing further (also confirmed live, below). Closing it properly needs a server-side uniqueness constraint the platform does not offer, or a distributed lock, which is out of proportion to a first-session-only race.

## Evidence

Re-verified after merging `next` into the branch, which brought in the repo's **pnpm cutover** — the whole tier below was re-run under pnpm from a clean install.

| Check | Result |
|---|---|
| `@memberjunction/ai-elevenlabs` unit | **73 passed** (51 before; the realtime suite goes 42 → 64) |
| `@memberjunction/ai-agents` unit | 1766 passed |
| `@memberjunction/ai-realtime-client` unit | 401 passed |
| Full monorepo build | 267/267 tasks |
| Integration tier (deterministic) | 53/54 — see below |
| `check:claude-md`, `check:esm`, `check:standards` | clean |
| Live smoke against a real ElevenLabs account | **30/32 + 8/8** — see below |

### Integration tier

The one failure is `IT65 - AI Providers and Model Resolution`, and it is **environmental, not a regression** — verified directly against the database rather than inferred. Both Cohere Command rows are `IsActive = 1` there, last updated **2026-05-26** and **2026-07-09**, long before this branch existed, with `DriverClass: CohereLLM` — a class that only ever registers a `BaseReranker`, never a `BaseLLM`. Repo metadata already carries both as `IsActive: false`; the database simply predates that metadata, and the check's own comment records that this was "closed by deactivating those models; no allowlist needed". Clearing it needs a metadata sync push into that database. No code on this branch can affect a row in it.

`IT51 - Realtime Deterministic Seams`, the bundle nearest this code, passes. (`IT68` — cited in an earlier revision of this description as the expected failure — now passes; #3371 has since merged.)

### Live smoke

Defects 2 and 3 are round-trip behaviours of the live API that no unit test could have seen, so the branch is validated against a real ElevenLabs account from a local harness. That harness is **not** included in this diff; each defect it found is pinned here by a unit test built from the exact payload observed on the wire, and two of those guards were mutation-checked (reverted to the old semantics to confirm they fail).

What the current run establishes:

- **Per-session voice** — `tts.voice_id` lands on the client-direct mint, and a voice-less session still omits the `tts` key entirely. Two server-bridged sessions with different voices each opened a real websocket and reached `conversation_initiation_metadata`.
- **Defect 1** — a freshly provisioned agent enables both `agent.prompt.prompt` and `tts.voiceId`. A deliberately pre-fix-shaped control agent (prompt override only) passes the deprecated `PromptOverrideEnabled()` — i.e. the old code would have judged it current and dropped the voice silently forever — and correctly fails `OverridesSatisfied()`.
- **Defect 2** — two further cold-instance ensures against the provisioned agent issued **zero PATCHes**, so the fingerprint survives the platform's key reordering and materialized defaults.
- **Defect 3** — sequential ensures never fork a second agent; and against a name that already had duplicates, 3 sequential plus 4 concurrent cold ensures all adopted the same oldest agent and created nothing further.

- **Defect 4** — a separate live run drove the exact regression scenario: two concurrent sessions on **one** driver instance against a brand-new name. One create, both sessions on the same agent id, the cold re-ensure still issuing no PATCH, and a warm instance serving further sessions with zero REST calls. **8/8.**

Two assertions in the first pass failed and were **harness errors, not code defects**: they asserted that simultaneous cold starts against a non-existent name would converge, which no lookup retry can deliver (see the residual note above). Re-expressed as the convergence property the code actually claims, that check passes 5/5. (Likewise one assertion in the defect-4 run read zero agents moments after create — the harness calling the raw non-retrying lookup and meeting defect 3's own eventual-consistency window; the retrying path found it, and a later list shows exactly one agent.)

Not proven: that the voice is *audibly* applied. The socket accepts the frame and the enablement is verified both ways, but ElevenLabs drops a disallowed override silently, so acceptance alone is not proof of audio — establishing that needs the agent to speak and a comparison of two voices' output.

### Documentation corrections from review

Two claims in the code's own comments were wrong or under-stated, and are fixed:

- The lookup retry was documented as costing "a few hundred ms". It costs the whole backoff ladder — **≈1.5s** (500ms + 1000ms) — on the first-ever provision of a name, because that is precisely the case where every attempt misses. Later sessions pay nothing.
- `isMaterializedDefault` was documented as dropping the platform's materialized defaults. It matches on the **value**, so it prunes *any* `''`/`false`/`[]` entry — `additionalProperties: false` included. That is safe in the direction that matters (both sides prune, so it never manufactures drift), but it costs sensitivity to an entry's presence. A new test pins that a meaningful `false` → `true` change still registers; it was mutation-checked by widening the prune to swallow `true`, which kills it.

## Behaviour on deploy

The first realtime session per managed agent after this deploys will issue **one** PATCH, to enable the `tts.voiceId` override on agents provisioned by an earlier version. That is the intended repair and it settles immediately — a unit test asserts the repaired agent issues no further PATCHes, and the live run reproduced exactly that against a real pre-fix-shaped agent: one PATCH, then the next ensure quiet. No migration, no metadata push, no scheduled job.

## API notes

`ElevenLabsRealtimeSession.SendInitiation` now takes the wire-shaped overrides object rather than the system-prompt string, so both topologies build it in one place. The class is documented as driver-constructed and never instantiated by consumers, and the change breaks loudly at compile time (`string` is not assignable to `JSONObject`).

`ElevenLabsRealtime.PromptOverrideEnabled` is deprecated in favour of `OverridesSatisfied`. It reports on one override and so cannot answer whether an agent is current — a `true` from it is actively misleading for an agent predating per-session voice.

The `@memberjunction/ai-agents` files in this diff are a JSDoc correction (the canonical config example documented an `elevenlabs.voiceId` key that was never implemented) and the matching test fixtures. No behaviour changes there, hence no version bump for that package.


rkihm-BC (approved):
**Approve.** Three real defects fixed, each pinned by a test built from a wire payload observed live, backward-compatibility preserved, and documentation that explains the *why*. I verified the core claims against the codebase rather than taking the description on faith.

## Verified independently (not just read)

- **Config flow is correct end-to-end.** `base-agent.ts:2222` and `realtime-client-session-service.ts:1746` both call `GetProviderVoiceSettings`, which returns the provider bag verbatim and merges it into the open `Config` — so `realtime.voice.providers.elevenlabs.voice` genuinely arrives as `Config.voice`, which `ResolveVoiceID` reads.
- **The camelCase/snake_case split — the subtlest part — is right.** Enablement (`buildRequiredOverrideEnablement`) uses `tts.voiceId` (SDK `…Input` serializer); the wire frame (`BuildSessionOverrides`) uses `tts.voice_id` (forwarded raw); `OverridesSatisfied` reads `tts.voiceId` off the SDK `…Output` shape. All three consistent. This is exactly the kind of thing that silently half-works, and it's handled correctly and documented at the split point.
- **The client-direct path actually sends the voice.** `elevenLabsRealtimeClient.ts:445` `parseOverrides` forwards `SessionConfig.overrides` verbatim as `conversation_config_override`. It ignores `SessionConfig.config`, so the redundant `voice` left in the config bag is harmless.
- **No external callers break.** `SendInitiation`, `PromptOverrideEnabled`, `BuildSessionOverrides`, `OverridesSatisfied`, `PickCanonicalAgent` are all confined to the driver file.
- **`createdAtUnixSecs` is a required `number` on `AgentSummaryResponseModel`** (confirmed in the SDK `.d.ts`), so `PickCanonicalAgent`'s reduce is well-typed.

## What's good

- **The two-halves-pinned-together design** (`buildRequiredOverrideEnablement` writes / `OverridesSatisfied` reads, guarded by "considers an agent this driver just provisioned already satisfied" and the generalized "repairs an agent missing ANY single required override" test driven off `enabledLeafPaths(written)`) turns "add a future override and forget the drift check" from a silent production no-op into a build failure. Correct fix for the *class* of bug, not just the instance.
- **Fingerprint canonicalization** preserves array order (enum/required are data) while normalizing key order and pruning materialized defaults — with tests asserting each property, including the negative case (reordered enum must still differ).
- **`PickCanonicalAgent` adopting the oldest (ties → agent id)** is the right convergence rule and makes the unaddressed true-simultaneity race self-healing.
- **Honest scoping** — explicitly calls out what's *not* proven (audible voice) and the residual concurrency window, with live evidence for both.

## Concerns (all minor / non-blocking)

1. **First-ever provision now pays ~1.5 s of pure sleep.** When a managed-agent name genuinely doesn't exist yet, `findAgentByName` exhausts all 3 attempts (`500ms + 1000ms`) *before* falling through to create. Bounded and instance-cached, only the first provision pays it — but on the cold-create path it's up to 1.5 s, not just "a few hundred ms." Worth a one-line acknowledgment.
2. **The ensure cache is a resolved-value cache, not an in-flight-promise cache** (`agentCache` populated only after `findCreateOrUpdateAgent` resolves). Two concurrent `ensureAgent` calls for the same brand-new name *on the same instance* can both miss and double-create — the intra-process analogue of the documented cross-process race. **Pre-existing, not introduced here**, and `PickCanonicalAgent` makes it self-heal; a promise-cache would close the intra-instance half cheaply (no distributed lock needed). Mention, not a demand.
3. **`isMaterializedDefault` prunes by *value* (`''`/`false`/`[]`), not by specific platform keys** — couples correctness to the assumption that no meaningful tool-schema field is ever exactly one of those (e.g. `additionalProperties: false` would be pruned). The direction of the sensitivity loss is safe (a value *change* like `false`→`true` is still caught) and the tradeoff is documented. Fine as-is; just flagging it's broader than "the five materialized fields."
4. **`@memberjunction/ai-agents` gets no version bump** — defensible: its only source change is a single JSDoc line correcting a documented-but-never-implemented `voiceId` key to `voice`; the rest are test fixtures. No runtime behavior change.

## Missing

- Nothing structural. The un-proven item (voice *audibly* applied) is inherent to ElevenLabs silently dropping disallowed overrides and reasonable to defer.
- The one non-green check (`IT65 - AI Providers and Model Resolution`) is credibly explained as environmental (stale Cohere models Active in the local scratch DB, already `IsActive: false` in repo metadata). I did not re-run it; a merging reviewer should confirm it's genuinely pre-existing.

## Bottom line

The code does what the description says, the risky seam (casing split) is correct, backward compatibility holds, and the tests encode live-observed payloads rather than idealized ones — which is why they catch round-trip bugs unit tests normally miss. Concerns are latency-awareness and a pre-existing concurrency nuance, none of which should hold up the merge.

PR #3385, opened 2026-07-31
chore(integration): remove connector source duplicated by the Integrations repo — 6.x

MS-BC:
> **⚠️ BREAKING — targeted at the 6.x line. Nothing is merged here; opening for review.**

## Why

`packages/Integration/connectors` held **36 vendor connectors that were pure duplicates**. The same classes ship from [MemberJunction/Integrations](https://github.com/MemberJunction/Integrations), where each connector is a self-contained **Open App** (`@memberjunction/connector-<vendor>`) with its own `package.json`, versioning, changesets, metadata, CI gates and seed migrations for **SQL Server and PostgreSQL**.

The Integrations copy is the one that actually reaches customers. Keeping a second copy in the monorepo meant every fix had to be applied twice, and the two drifted. This makes the Integrations repo the single source of truth and lets connectors version independently of the MJ core release train.

## What was removed

Aptify, Blackbaud, ConstantContact, Cvent, DynamicsDataverse, FileFeed, Fonteva, GrowthZone, Hivebrite, HubSpot, iMIS, MJToMJ, MagnetMail, Mailchimp, MemberSuite, NeonCRM, NetForum, NetSuite, NimbleAMS, Novi, ORCID, OpenWater, PathLMS, PheedLoop, PropFuel, QuickBooks, Rasa, Reach360, RelationalDB, Rhythm, SageIntacct, Salesforce, SharePoint, Wicket, WildApricot, YourMembership.

Plus their unit tests and fixtures, the vendored 146k-line YourMembership HTML API reference, and `generate-integration-actions.ts` (which existed only to instantiate the connectors in this package).

`connectors/README.md` now carries the full class → Open App → npm package table.

## What deliberately stayed

**The package is not deleted.** It is reduced to three External Data Source connector base classes:

- `BaseExternalDataSourceConnector`
- `BaseSqlExternalDataSourceConnector`
- `BaseDocumentDataSourceConnector`

These are **not** duplicated — six shipped Open Apps import them from this package rather than carrying a copy:

```ts
// Platform/SQLServer/src/SQLServerConnector.ts  (MemberJunction/Integrations)
import { BaseSqlExternalDataSourceConnector } from '@memberjunction/integration-connectors';
```

Same for PostgreSQL, MySQL, Oracle, Snowflake and MongoDB. Their `EdsConnectors` test suite is kept and still passes.

## Breaking-change / migration notes

1. **Install the Open App** for each connector in use. Each ships its own seed migration (SQL Server + PostgreSQL).
2. **The catalog row is re-pointed for you.** Each Open App's migration writes the *same* `__mj.Integration` row — same hardcoded ID as the original monorepo seed — with `ClassName`/`ImportPath` set to the connector's npm package. Existing `CompanyIntegration` records keep working.
3. **Registration key changed.** The monorepo registered the bare class symbol (`'SalesforceConnector'`); the Integrations repo registers the package name (`'@memberjunction/connector-salesforce'`) under its four-way identity invariant. Step 2 is what keeps resolution working.
4. **Direct imports** move from `@memberjunction/integration-connectors` to `@memberjunction/connector-<vendor>`.

> **A deployment that upgrades to 6.x without installing the corresponding Open App will have catalog rows pointing at a package that no longer contains those classes, and those integrations will fail to resolve.** This is stated plainly in the README, the changeset and the commit message.

Applied migrations under `migrations/v5/**` are **untouched** — they still carry the old `ClassName`/`ImportPath` seeds. Applied migrations are immutable; the re-point happens forward, through each Open App's own migration.

## Knock-on changes

- **`@memberjunction/server-bootstrap`** drops the `@memberjunction/integration-connectors` dependency and the 35 corresponding entries from its generated class-registration manifest (`CLASS_REGISTRATIONS_COUNT` 1000 → 965). That dependency existed solely to statically pin `@RegisterClass` classes against tree-shaking; the remaining base classes are abstract and register nothing.
- **`@memberjunction/integration-connectors`** drops `jsonwebtoken`, `mssql`, `zod` and `@memberjunction/global` — none of which the remaining source imports.
- **`packages/Integration/docs/connector-development.md`** and **`INTEGRATION_ACTIONS.md`** now direct new connector work to the Integrations repo, with a table mapping each monorepo convention to its Open App equivalent. The connector *contract* is unchanged; only the location and registration key differ.

## Whova was ported first, not deleted

`WhovaConnector` existed only in the monorepo (on the `connector-builder-v2` branch, not on `next`) with no counterpart in Integrations. It was **ported over first** in [Integrations#122](https://github.com/MemberJunction/Integrations/pull/122) so nothing is lost. Every class removed in *this* PR was verified to have a live counterpart in Integrations before deletion.

## Verification

| Check | Result |
|---|---|
| `turbo build --filter=@memberjunction/server-bootstrap` | ✅ **171/171** tasks successful, **0** TypeScript errors |
| `integration-connectors` unit tests | ✅ **24/24** passing |
| Residual code references to the package anywhere in the monorepo | ✅ none outside the package itself |
| 36 removed classes ↔ Integrations counterparts | ✅ all 36 map to a shipped Open App |

The one warning in the build log — `Error: command codegen:manifest not found` from `server-bootstrap-lite`'s `prebuild` — is a pre-existing local-environment issue (older global `mj` CLI) that the script explicitly swallows with `|| echo 'Warning: ...'`. It is unrelated to this change and does not fail the build.

## Not merged

Opening for review only — per instruction, nothing is merged.

Verified against a **real SQL Server MemberJunction database** (23 populated `__mj.Integration` rows, 8 live `CompanyIntegration` records bound to connectors this PR removes), plus a full static comparison of every `spCreateIntegration` seed in `migrations/v5/**` against every Open App seed in the Integrations repo.

**What holds:**
- No consumer breakage inside the monorepo. A repo-wide search for `@memberjunction/integration-connectors` outside the package itself returns **zero** hits — nothing in `packages/**` or `metadata/**` imports the removed classes.
- Reduced package builds clean (`tsc && tsc-alias`); its suite is 24/24; `turbo build --filter=@memberjunction/server-bootstrap` succeeds with the dependency and the 35 manifest entries gone.
- All 36 removed connectors have a corresponding Open App directory in the Integrations repo.
- The same-ID re-point claim is **true for 16 of the 24** monorepo-seeded integrations (Aptify, GrowthZone, iMIS, MJ to MJ, NetForum Enterprise, NetSuite, Nimble AMS, PropFuel, QuickBooks, Rasa.io, Reach360, Sage Intacct, Salesforce, SharePoint, Wicket, YourMembership) — byte-identical GUIDs on both sides.

**What does not — 7 integrations (defect found, changeset corrected in `895beea`):**

| Integration | monorepo seed | Open App seed | result |
|---|---|---|---|
| Mailchimp | `987FA1B5…` `Mailchimp` | `D9C7F5B4…` `mailchimp` | Open App install **fails** |
| Blackbaud | `2BBF275A…` `Blackbaud` | `0159550E…` `blackbaud` | install **fails** |
| HubSpot | `3DD4C246…` `HubSpot` | `71EC4CCB…` `HubSpot` | install **fails** |
| MagnetMail | `7F9BD70C…` `MagnetMail` | `98A49146…` `magnetmail` | install **fails** |
| Wild Apricot | `4FB2B6BF…` `Wild Apricot` | `FE1334F6…` `Wild Apricot` | install **fails** |
| Constant Contact | `16B66076…` `Constant Contact` | `65BB124A…` `constant-contact` | second row; original dangles |
| File Feed | `D26F22CE…` `File Feed` | *(no seed migration)* | nothing re-points it |

`__mj.Integration` carries `UQ_Integration_Name`, and the database collation is `SQL_Latin1_General_CP1_CI_AS` — case-insensitive — so `Blackbaud` vs `blackbaud` collides just as `Mailchimp` vs `mailchimp` does. Replaying Mailchimp's exact Open App seed against the live database (inside a rolled-back transaction) produced:

```
Violation of UNIQUE KEY constraint 'UQ_Integration_Name'. Cannot insert duplicate key
in object '__mj.Integration'. The duplicate key value is (mailchimp).
```

and a case-varied probe (`MAILCHIMP`) failed identically, confirming the collation behavior.

So for those five the connector cannot be installed **at all** on a deployment carrying the monorepo row, and the "install the Open App and existing records keep working" migration path does not exist; for Constant Contact and File Feed the install succeeds but leaves the original catalog row pointing at the emptied package. The live database shows this state directly: 4 integrations still read `ImportPath = @memberjunction/integration-connectors` (`mailchimp`, `rasa`, `File Feed`, `Betty AI`) with 8 `CompanyIntegration` records bound to them, while the 19 whose Open Apps were installed have already re-pointed — proving the re-point mechanism works where the IDs line up and stalls where they don't.

Repointing those seven — either by aligning the Open App seed IDs in the Integrations repo or by a forward re-point migration here — should land before the 6.x cut. That fix is **not** included in this PR; the changeset now documents the exception table rather than claiming coverage it doesn't have.

*(Unrelated pre-existing observation: `Betty AI` seeds `ClassName = BettyConnector` / `ImportPath = @memberjunction/integration-connectors`, but no such class exists in the package today or before this PR — that row is already dangling and is not affected by this change.)*


MS-BC:
Base synced to current `next`; two `package.json` conflicts resolved manually:
- `packages/ServerBootstrap/package.json` — took next's dependency list minus `@memberjunction/integration-connectors`, which is exactly what this PR removes.
- `packages/Integration/connectors/package.json` — kept this branch's trimmed description and dep set, with every `5.50.0` pin moved to `6.0.0` to match `next`. Final deps: `core`, `core-entities`, `external-data-sources`, `integration-engine`, `integration-engine-base`, all 6.0.0; package version 6.0.0.

Manually re-ran the connector/engine suites locally post-merge: **602 passed** and **642 passed**.

AN-BC (approved):

cadam11:
@MS-BC following up on the connector re-point now that 6.1.0-edge.0 is out. The Open App seed migrations re-point the same __mj.Integration row by UUID for 16 of the 24 monorepo-seeded integrations, but seven don't line up:

- Mailchimp, Blackbaud, HubSpot, MagnetMail, Wild Apricot: the Open App seeds a different UUID with a name that collides case-insensitively, so the install hard-fails on UQ_Integration_Name against an upgraded database.
- Constant Contact: the name differs enough (constant-contact) that a second row installs and the original row dangles, still pointing at the emptied package.
- File Feed: no Open App seed exists at all, so nothing re-points it.

We decided to fix this upstream in the Integrations repo. Can you update the six colliding Open App seeds to reuse the original monorepo UUIDs (the same-ID re-point that already works for the other 16) and add a File Feed seed? If any of those app versions are already published, the fix should ship as a follow-up migration in each app rather than editing the shipped seed.

That closes it for every future install and ships independently of the MJ release train. Full detail is in the Upgrading to v6 guide on the docs site and the amended v6.1.0-edge.0 release notes.

PR #3389, opened 2026-07-31
fix(core): compare UUID primary keys case-insensitively in BaseEngine cache maintenance

MarceloT-BC:
## The bug

`BaseEngine.findEntityIndexByPrimaryKeys` locates a cached entity by matching its primary-key values, so an event-driven `update` can be applied in place. It compared PK values with a raw `===`, which is wrong for UUID columns: the **same** id arrives with different casing depending on its source — a client-minted **lowercase** UUID from `BaseEntity.NewRecord` vs. an **uppercase** value loaded from SQL Server. On a case mismatch `===` returns `-1` ("not found"), and the caller's "not found → add it" branch **appends a duplicate copy** of the row into the engine cache. Every consumer (pickers, dashboards) then shows the row twice. The database is correct throughout.

## The fix

Drive the comparison off **metadata**, not a string-shape heuristic:

```typescript
if (pk.IsUniqueIdentifier) {
    return UUIDsEqual(entityValue, targetValue);   // UUID columns: case-insensitive
}
return entityValue === targetValue;                // everything else: strict
```

`EntityFieldInfo.IsUniqueIdentifier` is PG-aware (matches both `uniqueidentifier` and `uuid`), and the `pk` loop variable already *is* an `EntityFieldInfo`, so this is a localized change at the one broken compare. Non-UUID keys (codes, ints) keep exact equality.

**Scope:** the composite-key path (the reported bug). The `primaryKeys.length === 0` fallback compares `PrimaryKey.ToString()` and has no `EntityFieldInfo` to consult; it's left untouched.

## Validation

- `turbo build @memberjunction/core` — clean.
- MJCore unit suite — **1601 pass / 0 fail** (104 files).
- New `baseEngine.uuidPrimaryKeyMatch.test.ts` (7 cases): the duplicate-row scenario (uppercase-cached vs lowercase-incoming → matched), reverse direction, find-among-several, genuinely-different-UUIDs → `-1`, non-UUID string PKs stay strict, composite PK (UUID case-insensitive + non-UUID strict), numeric PKs strict.
- A/B verified the test is a real regression guard: with the fix disabled the 4 case-mismatch cases fail; restored → 7/7.


AN-BC (approved):
@MarceloT-BC excellent hidden UUID variance find here that would def cause issues with PG!

Nice work!!! 

Merge this Mon after @cadam11 gives clear and then back port for 5.51.1 as this is a good bug fix!

Flagging for @MS-BC as well as he's doing a lot of PG stuff and this will tighten up PG

PR #3403, opened 2026-08-01
UI Layering Guide — the four-layer standard, all 96 violations fixed, and the tree locked

AN-BC:
## Why

MJ has built UI in layers **three times** and written the rule down as a general standard **zero times** — `Generic/**` vs `Explorer/**`, the [Forms Architecture](https://github.com/MemberJunction/MJ/blob/next/guides/FORMS_ARCHITECTURE_GUIDE.md) stack, the [Conversations UX](https://github.com/MemberJunction/MJ/blob/next/guides/CONVERSATIONS_UX_STACK_GUIDE.md) stack. Because it was never general, every new app reinvents it — and the one tree that *did* have a written rule drifted anyway.

This PR adds the standard, **fixes every violation in `packages/Angular`**, and locks the tree so it can't recur.

## The standard — `guides/UI_LAYERING_GUIDE.md`

```
L3  Explorer surface   entity forms + resource/dashboard components
                       owns NavigationService · holds NO domain logic
L2  Composite widget   assembles L1 · MAY read data, ONLY via ProviderToUse
                       emits intent, never navigates
L1  Presentational     props in, events out · zero data access
L0  Domain runtime     pure TS, no Angular at all
```

Plus the `Before*`/`After*` cancelable event contract, per-layer data-access rules, the app-repo package split, and a migration recipe for a 400-line do-everything screen.

## The cleanup — 96 violations, all resolved

The first version of this PR declared 26 packages and listed 8 drifted ones as follow-up. **That left the gate decorative**: "undeclared" and "compliant" are indistinguishable from outside, so the 57 packages nobody had looked at could keep drifting. So: all **85** `packages/Angular` packages are now declared, which surfaced **96** violations — the 8 known, plus 88 nobody knew about.

| Count | Class | Why it mattered |
|---:|---|---|
| **50** | `new RunView()` / `new Metadata()` / `new RunQuery()` in reusable components | Not a style nit. A component that constructs its own `RunView` **silently ignores the `Provider` it was handed** and reads from whichever MJ server is the global default. Fixed by extending `BaseAngularComponent` (17 components) or threading a provider parameter (9 services/utilities). |
| **27** | Unused `@angular/router` / `ng-shared` declarations | Dead manifest entries — each a standing invitation for the next commit to use them. |
| **11** | Real Explorer imports inside Generic packages | The actual coupling. |
| **4** | `ActivatedRoute` in a `BaseResourceComponent` | The class already had `GetQueryParams()` / `OnQueryParamsChanged()`. Explorer **caches and reuses** resource components, so a `route.queryParams` subscription desyncs from the tab lifecycle. |

Three fixes worth reading:

- **`SharedService.CreateSimpleNotification` was already deprecated** in favour of `MJNotificationService` — a Generic package. `file-storage` and `join-grid` simply hadn't moved. Straight swap.
- **`SharedService.IsDescendant` is a pure DOM predicate** with no Explorer coupling. Keeping it there forced widgets to depend on MJ Explorer for a three-line tree walk. Moved to `ng-shared-generic` as `IsDescendantElement`; `SharedService` delegates, so Explorer callers are untouched.
- **`ai-test-harness` needed genuine navigation** that no `@Output()` could reach through its dynamically-created wrappers. Added `RecordNavigationAdapter` to `ng-base-types`, following the adapter pattern `conversations-runtime` already uses for notifications. Explorer registers it in `SharedService`'s constructor. **The unregistered default logs an error rather than no-op'ing** — a drill-through that silently stops working is the exact failure mode this whole effort is about.

## Gate fixes the cleanup forced

Turning the gate on 57 more packages taught it three things:

- **`new RunView(provider)` was being flagged.** Only the zero-arg form binds the global; the pattern now requires empty parens. A gate that cries wolf gets switched off.
- **Added the `shell` layer** for the four packages that *are* the navigation layer — `ng-explorer-core`, `ng-explorer-app`, `ng-shared`, `ng-bootstrap`. Banning `Router` there would ban the implementation of the rule. It checks nothing; its job is to make the exception **enumerable** rather than leaving those packages looking un-reviewed.
- **The `mj-ui-layers-allow` marker now counts on the line above**, not just the line itself — a real exception deserves a sentence, and a trailing comment that long is unreadable. One line only, so a marker can't drift away from what it excuses.

## Locking it — this is the part that stops recurrence

```bash
node .github/scripts/check-ui-layers.mjs packages/Angular --require-declared
```

`--require-declared` turns a missing `mjUILayer` field from a **skip** into a **failure**. The way drift returns is not an edit to an existing package — it's a **new package that never opts in**. CI now runs:

1. the **locked** form over `packages/Angular` (clean, so it stays clean),
2. the **unlocked** form over the rest of `packages/` (still adopting — declared packages are checked, undeclared are skipped),
3. **the gate's own unit tests** — a gate with a broken parser passes everything.

## One standing exception, marked in source

`file-storage`'s `FileBrowserResource` is an Explorer surface sitting in a Generic package. Moving it is the correct fix, but it's registered in **three CodeGen-generated manifests** (`explorer-core`'s `lazy-feature-config.ts` and both Bootstrap packages' class registrations) which must be regenerated against a database, not hand-edited. Marked with `mj-ui-layers-allow` and a reason, tracked in **[#3404](https://github.com/MemberJunction/MJ/issues/3404)**.

## Scope: every MemberJunction and Blue Cypress repo

Guide §11 states it. Two companion PRs apply it, each converting one vertical slice end to end. Between them they surfaced **five more real bugs**:

- **[bizapps-accounting#37](https://github.com/MemberJunction/bizapps-accounting/pull/37)** — two components rendered a journal entry independently. Merging them found a reversal rule wrong in *both* directions (refused reversals the server permits, offered ones it rejects — next to a correct unit-tested rule already in the repo), a totals `colspan` one column out so the **debit total rendered under the Credit heading**, and a drill-through reading a column that had been replaced, hidden by an untyped cast. Net −914 lines.
- **[bizapps-orders#23](https://github.com/MemberJunction/bizapps-orders/pull/23)** — package split + the missing forms layer. Found the same global-provider class in a root-provided service, and fixed a pre-existing red test.

Tracking issue for the remaining repos: **[#3404](https://github.com/MemberJunction/MJ/issues/3404)**, assigned to @bc-izygmunt and @MarceloT-BC.

## Verification

- `npx turbo run build --filter="./packages/Angular/**"` → **118/118 green**
- `npx turbo run test --filter="./packages/Angular/**"` → **202/202 tasks pass**. Three test files needed updating for the new provider paths and the `@Directive` base class — **fixed, not skipped**:
  - `agent-request-panel.test.ts` — its `@angular/core` mock now answers any `ɵ*` compiler helper via a Proxy, rather than enumerating them and breaking on the next one
  - `runtime-utilities-ml.test.ts`, `interactive-form-apply.service.test.ts` — mocks now provide `RunView.FromMetadataProvider` / `Metadata.Provider`
- `npm run check:ui-layers` → green (85 declared, locked over `packages/Angular`)
- `npm run check:claude-md` → green
- Gate's own tests → 15 pass (`npx vitest run --config .github/scripts/vitest.config.mts`)

**Not verified:** no browser run. This touches ~30 shipped Angular packages; everything is compile-checked under `strictTemplates` and unit-tested, but the multi-provider paths and the `ai-test-harness` drill-through deserve a click-through before merge.


AN-BC:
**Correction to this PR's description.**

Two edits I intended to make to `guides/UI_LAYERING_GUIDE.md` did not land in this branch. I applied them with a script whose string replacements didn't match, and a no-op `.replace()` reports nothing — so the commit went through and the description above claims content the guide doesn't contain.

Specifically, the guide as it stands on this branch is **missing**:

- the note that only the **zero-argument** form is banned (`new RunView(provider)` is correct code), and the "two things keep the gate from crying wolf" framing
- the **locking** section describing `--require-declared` and MJ's own locked/unlocked split

Everything else in the description is accurate — the code, the gate, the 96 fixes, the `shell` layer and the escape-hatch change are all present and verified.

Both missing sections are applied in **[#3406](https://github.com/MemberJunction/MJ/pull/3406)** (stacked on this branch), with an assertion on every replacement so a silent miss can't recur. #3406 also supersedes the enforcement mechanism: `.github/scripts/check-ui-layers.mjs` is replaced by the versioned `@memberjunction/standards` package, and the guide's enforcement section is rewritten around that.

No action needed here — flagging it so the merged description isn't taken at face value.

---

PR #3406, opened 2026-08-01
[6.0] @memberjunction/standards — versioned, opt-in standards with a scaffold and CLI

AN-BC:
> **#3403 has landed and this is now based on `next`.** The stacking note that used to head this description is resolved — no ordering constraint remains.

## Why

MJ's standards come in two kinds and they need opposite distribution mechanisms.

**Judgment standards** are prose that has to be read — the guides. Distributed as documentation.
**Executable standards** are the ones a machine can settle. Those belong in a versioned package.

The case made itself: over the course of this work I copy-pasted `check-ui-layers.mjs` into three repos by hand. That stops scaling at about five repos and guarantees half of them are running a version from eight months ago.

## The property the whole design exists for

> **Adding a standard to this package never changes an existing repository's result.**

Three mechanisms, together:

1. **Opt-in per check.** Registration makes a check *available*, not active. It does nothing until a repo's `.mj-standards.json` names it.
2. **Version-pinned adoption.** Each check declares the MJ version that introduced it (`Since`); each repo records the version it adopted against (`StandardsVersion`). Newer checks are reported as available and **are not run**. Upgrading the package cannot activate them — only `mj standards adopt --upgrade` can, and that's a reviewable commit.
3. **Severity lives in the repo, not the package.** `DefaultSeverity` is what `adopt` writes for a *new* adopter. Changing it here never touches an already-adopted repo, so severity decays forward (`warn` → `error`, by the repo's own choice) and never backward into something already shipped.

Verified end to end against a synthetic repo:

```
--- repo pinned at 5.40.0, check introduced in 6.0.0 ---
1 standard(s) added since this repo adopted 5.40.0:
    ui-layers  (since 6.0.0)  Widgets must not import Router or MJ Explorer…
    They are NOT active. Review with `mj standards adopt --upgrade`.
✓ All adopted standards pass.        exit=0

--- adopt WITHOUT --upgrade ---   → StandardsVersion stays 5.40.0, check not enabled
--- adopt --upgrade ---           → StandardsVersion 6.0.0, check enabled
```

## What shipped

| | |
|---|---|
| `packages/Standards` | the package — types, registry, config, runner, scaffold, the `ui-layers` check, a JSON schema, and an `mj-standards` binary |
| `mj standards adopt` | config + CI workflow + npm script + per-package layer declarations |
| `mj standards check` | run what the repo adopted (`--strict` to fail on warnings too) |
| `mj standards list` | every standard, and this repo's stance on each |

**Zero runtime dependencies.** This gets installed into client repos and run in their CI; every dependency is one more thing that can conflict with their tree. The only non-trivial need was semver comparison — twenty lines.

**Client and external repos don't need the MJ CLI.** The package ships its own binary:

```bash
npm i -D @memberjunction/standards
npx mj-standards adopt --ci github --declare-compliant
```

## MJ now dogfoods it

**`.github/scripts/check-ui-layers.mjs` is deleted.** Two implementations of the same rules is precisely the duplication this whole effort has been about. MJ adopts the package like any other repo — `.mj-standards.json`, `npm run check:standards`, and a workflow that runs the package. Same result:

```
✓ ui-layers [error]
    85 package(s) checked; 0 undeclared in a locked tree; 222 undeclared elsewhere and skipped
✓ All adopted standards pass.
```

The lock is now `requireDeclaredIn: ["packages/Angular"]` rather than a boolean, because that's the shape a real migration takes: one tree cleaned and held while the others are still being worked through. The boolean forced a choice between locking everything (impossible mid-migration) and locking nothing (which is how drift returns).

## Two bugs the work caught

**1. `--declare-compliant` assigned `shell` to a non-compliant package.** `shell` checks nothing, so *everything* passes as `shell` — it would have handed a permanent exemption to exactly the packages that need work, and produced a green check that meant nothing. A throwaway two-package repo caught it on the first run:

```
+ mjUILayer for @demo/dirty-widget — shell        ← wrong
· mjUILayer for @demo/dirty-widget — violations at every auto-assignable layer  ← fixed
```

`shell` is now never auto-assigned; a package that qualifies for nothing is reported, not declared. Regression test included.

**2. Two of my edits to `UI_LAYERING_GUIDE.md` in #3403 silently no-op'd.** The replacement strings didn't match and a no-op `.replace()` says nothing, so the guide shipped without the false-positive notes and the locking section that **#3403's description claims are there**. Applied properly here, with an assertion on every match this time. I'll correct the record on #3403 — flagging it rather than quietly fixing it, because the PR body currently overstates what merged.

## Adding a standard later

1. Implement `StandardCheck` in `src/checks/`.
2. Set `Since` to the version it will **ship in** — never backdate, that would silently activate it in repos that adopted earlier.
3. Register it in `src/registry.ts`.
4. Give it a `DocsUrl`; every failure prints it.
5. Consider `DefaultSeverity: 'warn'` first if it's likely to have a long tail.

Only `ui-layers` is ported so far. The other MJ gates (design tokens, ESM imports, `mj-btn` overrides) are shell scripts today and are follow-up work — the registry is built to take them.

## Merged up to `next` (6.0 line)

`next` moved to 6.0.0 while this sat open, which conflicted in two files and in one thing no textual merge could see — the package's own sense of what version it is.

- **`packages/MJCLI/package.json`** — took next's 6.0.0 dependency block, re-added `@memberjunction/standards` at 6.0.0.
- **`package-lock.json`** — took next's lock wholesale (it already carries the `@angular/router` removals from #3403) and re-applied only the three entries this branch adds: the workspace link, MJCLI's dependency, and the `packages/Standards` workspace entry.
- **Version metadata moved onto the 6.0 line.** `@memberjunction/standards` is 6.0.0 (the `fixed` changesets group requires every `@memberjunction/*` package share a version, so it publishes in lockstep and MJCLI's exact-version dependency resolves). `ui-layers`'s `Since` is **6.0.0** — the version it actually ships in. Backdating it to 5.51.0 would have been the exact mistake the "never backdate" rule above exists to prevent: it would silently activate the check in any repo that had adopted at 5.51.0 or later. `.mj-standards.json` records `StandardsVersion: 6.0.0`, and the adoption tests use 6.0.0 as the shipped-in version (5.40.0 stays the older-adopter case).
- **`ci-standards.yml` self-trigger path** pointed at the `ci-ui-layers.yml` this PR deletes, so edits to the workflow would never have triggered it. Now points at itself.

Re-verified on the merged tree: `tsc` clean, **49/49** standards tests, `mj-standards check` green (85 packages, 0 undeclared in the locked tree), and `check:claude-md` / `check:esm` / `check:peer-deps` all pass.

## Verification

- **138/138** build (standards + CLI, full dependency graph)
- **49/49** standards tests — version comparison, config validation, adoption/upgrade semantics, the check itself against temp-dir fixtures, and both regressions above
- **118/118** Angular packages still build
- `mj standards check` / `list` exercised against this repo through **both** the CLI and the standalone binary
- `adopt` / `check` / `--upgrade` exercised end to end against a synthetic two-package repo
- `npm run check:claude-md` green (it caught a dangling link to the deleted script)

**Not verified:** the package is not published, so the `npx mj-standards` path is exercised from the workspace build rather than a registry install. The generated CI workflow assumes `npm ci` provides the package — true once published, and MJ's own workflow builds it from the workspace instead.


AN-BC:
@MarceloT-BC @jordanfanapour @EL-BC @bc-izygmunt all of you should have a look at this. @MarceloT-BC this will be a good project for you to drive after Monday when the `next` branch becomes available again for `6.0.0` work, this is designed to create enforceable standards in a single package that can be used anywhere, CLI, programmatically, etc. The idea then layers up and we use in the MJ repo, but also in other repos. Beyond this, the idea is to have clear documentation that is **linked** to all of our other reps and when we scaffold a new repo with `mj install` we push in some claude.md and agent.md files that have links to the main MJ repo so the agents working in that new repo anyone sets up will adopt our standard by default. @JosueGarcia-BC  I think you did some updates to `mj install` a while back to do this for claude.md deployment and linking, is that right?

@rkihm-BC and @cadam11  for visibility.

PR #3407, opened 2026-08-01
Fix RealtimeClientSessionService.promptRunWriteChains unbounded growth

AN-BC:
## Summary

This PR is the output of the recurring `/audit-memory-leaks` sweep (Round 9). It refreshes the audit plan and fixes the one concrete, high-confidence bug the sweep surfaced this round.

### The bug: `RealtimeClientSessionService.promptRunWriteChains` leaks one entry per closed voice session

- `packages/AI/Agents/src/realtime/realtime-client-session-service.ts` serializes writes to a co-agent's `AIPromptRun` (transcript turns + usage checkpoints) through a per-run promise chain, keyed by `promptRunID`, stored in a private `promptRunWriteChains` map. The entry is meant to be dropped when the run finalizes (`finalizePromptRun`).
- `packages/MJServer/src/agentSessions/SessionManager.ts` (`finalizeObservabilityRuns`, called from every `CloseSession`) finalized the run by doing `new RealtimeClientSessionService().FinalizeCoAgentRun(...)` — a **brand-new** instance. Its `promptRunWriteChains` map starts empty, so the delete call always operates on the wrong object. The real map — on the instance that actually accumulated the entry — is never touched.
- Net effect: every client-direct realtime voice session that ever closes leaks one `promptRunWriteChains` entry, unbounded, for the life of the process.
- Investigation found the leak is reachable through **more than one close path**: `RealtimeClientSessionResolver` owns one `RealtimeClientSessionService` instance, but the actual hangup mutation the browser calls (`AgentSessionResolver.CloseAgentSession`) and the background reaper (`SessionJanitor`) each own a **separate** `SessionManager` (and therefore a separate default `RealtimeClientSessionService`), with no shared reference back to the resolver's instance. Fully unifying all of these would mean converting `RealtimeClientSessionService` into a process-wide `BaseSingleton` — a change that touches ~5 production call sites and ~25 existing unit-test call sites that currently rely on constructing isolated instances per test. That refactor felt too invasive to make blind in this PR, so it's called out as a follow-up instead (see below).

### The fix (two parts)

1. **Root-cause fix, scoped to the path it can reach:** `SessionManager` now takes an optional `RealtimeClientSessionService` via constructor injection (defaults to `new RealtimeClientSessionService()`, preserving current behavior for `AgentSessionResolver`/`SessionJanitor`/telephony services). `RealtimeClientSessionResolver` now passes its own `clientSessionService` into the `SessionManager` it constructs, so its own close paths (e.g. the `ExecuteRealtimeSessionTool` error-path close) finalize correctly against the instance that owns the state.
2. **Defense-in-depth bound, covering every path (including the ones the root-cause fix can't reach):** `promptRunWriteChains` is now an `MJLruCache<string, Promise<unknown>>` (`maxSize: 5_000`, `ttlMs: 4h`) instead of a plain `Map` — the established MJ remediation pattern for this failure class (see `MJLruCache` usage elsewhere in the audit history). Even when a session closes through a differently-instantiated `RealtimeClientSessionService`, the map's growth is now capped and self-evicting instead of unbounded.

### Follow-up (not in this PR)

Convert `RealtimeClientSessionService` to a `BaseSingleton` (per this repo's own singleton convention in `.claude/rules/typescript-style.md`) so `AgentSessionResolver` and `SessionJanitor` share the exact same instance too, closing the gap completely rather than just bounding it. Deliberately left as a separate, wider-blast-radius change.

## Files changed

- `packages/AI/Agents/src/realtime/realtime-client-session-service.ts` — `promptRunWriteChains` → `MJLruCache`
- `packages/MJServer/src/agentSessions/SessionManager.ts` — constructor injection for `RealtimeClientSessionService`
- `packages/MJServer/src/resolvers/RealtimeClientSessionResolver.ts` — passes its shared instance into `SessionManager`
- `packages/MJServer/src/__tests__/SessionManager.test.ts` — new tests for the injected + default-instance finalize paths
- `packages/MJServer/src/__tests__/RealtimeBridgeResolver.test.ts` — fixed a now-stale `@memberjunction/ai-agents` mock (needed `RealtimeClientSessionService` exported once `SessionManager`'s constructor started touching it even on the default path)
- `plans/MEMORY_LEAK_AUDIT.md` + `plans/.memory-leak-snapshots/2026-08-01/` — Round 9 audit refresh (10 parallel subagents, full five-category sweep + five deep-subtree sweeps); confirmed Round 8's Critical (`WidgetSessionService` rate-limit cache) is fixed, catalogued ~31 new findings for future rounds

## Test plan

- [x] `@memberjunction/ai-agents` unit suite: 1766 tests passed (includes the 100 pre-existing `realtime-client-session-service.test.ts` tests, unaffected by the `MJLruCache` swap)
- [x] `@memberjunction/server` unit suite: 746 tests passed, 56 skipped, 1 failing test (`mjapi-bootstrap.test.ts`) — confirmed **pre-existing and unrelated**: it requires a DB-driven `mj codegen`-generated package (`mj_generatedentities`) not present in this sandboxed build environment, reproduces on a clean checkout of the base branch
- [x] Both `@memberjunction/ai-agents` and `@memberjunction/server` rebuild clean (`tsc && tsc-alias`) with no new errors
- [x] New `SessionManager.test.ts` cases specifically assert: (a) closing a session with an injected `RealtimeClientSessionService` finalizes through that instance and never touches the module-default one, (b) closing a session with no injection still finalizes through the default instance (back-compat for `SessionJanitor`/telephony services)

🤖 Generated with [Claude Code](https://claude.ai/code)

https://claude.ai/code/session_01HnnkWNczbr9XZxwx9TMB4Q

---
_Generated by [Claude Code](https://claude.ai/code/session_01HnnkWNczbr9XZxwx9TMB4Q)_


AN-BC (commented):
LGTM

PR #3408, opened 2026-08-01
Entity Action as the workflow-hook substrate — scope binding, sequencing, and a safe execution log

AN-BC:
Makes `EntityAction` the general workflow-hook substrate for MJ and every OpenApp built on it, instead of each app inventing its own — and fixes the execution log so the feature is safe and diagnosable once it's used at volume.

**Ships schema + plan + metadata + changeset. Engine work is specified, not built.** The plan's §6 is a post-CodeGen runbook with the ordering, which matters.

Driven by a BizApps Sales requirement — *"when a Deal reaches Closed Won, run a workflow, configurable per Deal Type and per Company"* — but generalized here because it's a framework need. Companion tracking PRs are open in ten BizApps repos pointing back at this one.

---

## What was already there

`EntityAction` does more than its schema suggests, and it's under-documented enough that apps keep reinventing it:

- **Genuinely wired into the save path** via `DatabaseProviderBase.HandleEntityActions`, overridden in `GenericDatabaseProvider`. `Validate` → `OnValidateBeforeSave` is a **real blocking gate** — a non-`Success` result fails the save. `Before*` are awaited but their results discarded. `After*` are fire-and-forget.
- **`Execute Agent` means any binding can already run an agent**, flow or loop. Chain: `EntityActionParam` → `MapParams` → `ActionParam[]` → `RunAction` → `AgentRunner.RunAgent({data})`.

So "run a flow agent when a deal closes" was *almost* expressible. Eight gaps stopped it — three functional, five in the log.

## Schema (all additive)

| Change | Why |
|---|---|
| `EntityAction.ScopeEntityID` + `ScopeRecordID` | Bind a workflow to **one configuration record** — this Deal Type, this Contract Type, this Pipeline, this Company. `NULL` keeps apply-to-all. **This is what stops every app growing a column per type per event.** |
| `EntityAction.Sequence` | `EntityActionFilter` has one; `EntityAction` didn't, so two bindings on one event ran in undefined order |
| `EntityAction.LoggingMode` | `All` / `FailuresOnly` / `None`, per binding — volume control |
| `EntityActionParam.ValueType` += `'Entity Object Data'` | Passes `entity.GetAll()`. `'Entity Object'` passes the live `BaseEntity`, which serializes to `{}` — so `Execute Agent` silently received an empty payload, with no error at any layer |
| `ActionParam.LogValue`, `EntityActionParam.LogValue` | Whether a parameter's value may be written to the log — definition-level, with a per-binding override |
| `ActionExecutionLog.EntityActionID`, `EntityActionInvocationTypeID`, `TargetEntityID`, `TargetRecordID` | Provenance — which binding, which record, which event. Without it a failed workflow is undiagnosable |
| `ActionExecutionLog.ResultParams` | So `Params` can stop being overwritten and keep the inputs **as the action was called** |

**On scope binding.** Without it, "tie this workflow to this Deal Type" lives inside `ActionFilter.Code` — it works, but the Deal Type form can't show "here are your workflows" without parsing filter code, so the configuration is invisible exactly where an admin looks. The framework stores and filters the pair; *how* a scope record relates to a subject is an app concern, via a resolver seam shaped like `BasePriceResolver`.

## The execution log — why it's half this PR

Three findings from reading `ActionEngine`, in ascending order of consequence.

**1. Params are logged unconditionally, and entity-action params are whole records.** `StartActionLog` writes `JSON.stringify(params.Params)` on every run; `EndActionLog` writes the merged set again. No opt-out. Harmless while Entity Actions are unused — **not harmless the moment they're the workflow substrate**, because `'Entity Object'` and `'Entity Object Data'` put the entire row into the log, twice per invocation. That's the space problem (the `MAX` payload, not the row) and the security problem (message bodies, Person fields, contract terms in a log with broad read access).

Posture is **fail-closed**: a hard engine rule that whole-record value types are never logged — no configuration, closes the hole on its own — plus the `LogValue` flags for the grey area (a `Static` param holding an API key, an `Entity Field` holding a national ID). Suppressed params still record name, type, `ByteLength`, `KeyCount` and truncated top-level keys, so *"it was called with the 41-column deal row, not the 3-field summary"* stays answerable. Shape, never values; one level only, since deeper structure can encode content.

**2. The end write destroyed the as-called inputs.** `Params` was overwritten with the final set — and because Custom and Generated actions mutate `params.Params` **in place**, the end state isn't "inputs plus outputs" but *inputs as the action left them*. For a workflow binding that's the difference between *"the agent got the wrong record"* and *"the agent mangled the right one."* `ResultParams` splits them; `QueueTask`'s `Data`/`Options`/`Output` was already the house shape.

`ResultParams` is written **on failure exactly as on success** — an audit trail that records only successes isn't an audit trail, and it costs no new control flow, since both failure paths already reach `EndActionLog`. NULL now means one thing only: the run never finished.

**3. Redaction belongs to the params, not to the log.** This was a defect in this PR's own plan: runbook step 9 routes `After*` through `QueueManager`, and `QueueTask.Data` persists the same params. Inline redaction would have re-opened the hole one layer down. Now specified as one shared `RedactParams` helper with a grep-able invariant — **no path writes a raw `ActionParam[]` to persistent storage.**

## Metadata

`Execute Agent`'s four content-bearing params — `Data`, `ConversationMessages`, `Payload`, `AgentResult` — ship with `LogValue: false`. The outputs matter as much as the inputs, because `EndActionLog` writes the *merged* set. Its six identifier params stay logged, so a run is still fully diagnosable and the content is one hop away in `MJ: AI Agent Runs`, where it belongs and is already access-controlled. **The content isn't lost; it stops being duplicated.**

## Also corrected while in here

- `ActionExecutionLog.Message`'s description claimed to hold *"JSON-formatted output data or response."* It's set from `result.Message`, a human-readable summary. Anyone reading the schema for outputs was sent to the wrong column.
- `EntityActionParam.ValueType`'s description still said *"Static, Entity Object, or Script"* after `Entity Field` shipped.
- **Retention is documented but not enforced.** `RetentionPeriod` says *"number of days to retain the log; NULL for indefinite"* and nothing outside generated code reads it, on either table. Pre-existing and separable, but it means retention isn't available as mitigation for anything — specified as a bounded Scheduled Job in §5.8.

## What this deliberately is not

- **Not a replacement for `MJ: Record Processes`** — that owns WORK × SCOPE × TRIGGER for bulk and scheduled work. EntityAction owns per-record lifecycle. Both dispatch into the same Action/Agent layer.
- **Not a BPM engine** — Flow Agents own orchestration. EntityAction is the *trigger*.
- **Not a human-task system** — `MJ: AI Agent Requests` resumes a paused run; `bizapps-tasks` owns durable assignable work.
- **Not a new workflow table anywhere.** An earlier draft proposed a `WorkflowEventType`/`WorkflowBinding`/`WorkflowRun` trio in `bizapps-common`. Reading this implementation killed it.

## ⚠️ Before merge

**CodeGen has not been run against this migration.** Apply it, run `mj sync push --include=entities` **then** `mj codegen`, and commit the generated output before any TypeScript references the new columns — §6 has the full ordering, including that the `ValueType` value list must regenerate from the CHECK constraint rather than being hand-inserted.

**Until the redaction rules land, the migration makes whole-record bindings *possible* and not yet *safe*.** Flagged in the migration header too.

Changeset: **minor** on `core-entities`, `actions`, `actions-base` (the repo pins `fixed: @memberjunction/*`, so the group moves in lockstep). Additive capability, every new column nullable or defaulted to today's semantics.

One open question left in §9: whether suppressed values should be **hashed** rather than omitted. Useful for proving two runs got the same payload — but a hash is a fingerprint of the record, so it deserves its own decision.


AN-BC:
@MS-BC id like to work with you on this, good core MJ PR to nail and will have awesome effect everywhere to really start using Entity Actions. We've had em a long time but not used - this will change that.

@rkihm-BC yoir input on this would be appreciated too. 

Also do we need to log entity action invocations? We have action execution log but doesn't tie to entity action lifecycles maybe we extend that table so we can have first class cold there that tie to the entity action that invoked it?

AN-BC:
On the logging question — yes, and I'd extend `ActionExecutionLog` rather than add a table. But reading the write path turned up something that has to be solved *with* it, not after.

## `Params` is already serialized into the log on every run — twice

`ActionEngine.StartActionLog` does `logEntity.Params = JSON.stringify(params.Params)`, and `EndActionLog` writes the merged input+output set again. Unconditionally — there's no opt-out on `Action` (`RetentionPeriod` cleans up after the fact, but every run still writes). The writes are queued fire-and-forget, so performance is fine.

That's harmless today because Entity Actions are barely used. **The moment they become the workflow substrate it stops being harmless**, because entity-action params are whole records:

- `ValueType='Entity Object'` — and the new `'Entity Object Data'` — put the **entire row** into `ActionExecutionLog.Params` as JSON, twice per run.
- An `AfterUpdate` binding on a busy entity therefore writes the full record to a general-purpose log table on every save.
- Confidentiality: `bizapps-secure-messaging` message bodies, Person fields, contract terms — all landing in a log with broad read access. I've already flagged in that repo's tracking PR that `ValueType` choice is a confidentiality decision there; this is the mechanism behind it.

So the logging work is two things, not one. Extending the log makes entity-action runs diagnosable; **param redaction makes it safe to turn on.** Shipping the first without the second is the version that bites in production.

Suggested: a per-param `EntityActionParam.LogValue BIT NOT NULL DEFAULT 1`, set to `0` for whole-object params. Per-param rather than per-action because that's where the sensitivity actually lives — you usually want the scalar params logged and the record redacted. An `Action`-level `LogParams` mode (`Full | Redacted | None`) would also work and is coarser. Either way it's an existing-behaviour change, so it may deserve its own decision even if it rides the same migration.

Worth pairing with an `EntityAction.LogExecutions` flag for pure volume control on high-traffic bindings.

## The columns themselves

Four nullable additions to `ActionExecutionLog`:

| Column | Why |
|---|---|
| `EntityActionID` (FK) | The binding that caused this run |
| `EntityActionInvocationTypeID` (FK) | *Which* event — you'll want to filter `Validate` vs `AfterUpdate` constantly |
| `TargetEntityID` (FK → `Entity`) | Denormalized: survives the EntityAction being deleted or retargeted, and lets you query the log by record with no join |
| `TargetRecordID` (`NVARCHAR(450)`) | *Which* record |

**Extend rather than a side table** because the child is optional, small and 1:1 — nullable columns beat a join in that shape, and "did this action run" stays one query. Same pattern accounting already uses with `JournalEntry.LinkedEntityID`/`LinkedRecordID`: a polymorphic origin pair on the main record, not a satellite.

**One thing I'd argue against:** making the origin pair *generic* (`InvokedByEntityID`/`InvokedByRecordID`) to also cover Record Processes, agent flow steps and scheduled actions. Those already have their own detail logs — `RecordProcessRunDetail`, `AIAgentRunStep` — so a generic invoker pair would duplicate them while giving up FK integrity. Concrete FK for the case we know matters; generalize only if a second invoker turns up that genuinely lacks its own log.

`TargetEntityID`/`TargetRecordID`, on the other hand, *are* worth keeping generic — every invoker has a "what was this about", and the `List`/`View` invocation types already run one action per record.

## Scope

I'd put the four columns in this PR — shipping the workflow substrate without run visibility means the first production incident is undiagnosable — and treat redaction as a companion decision, since it changes behaviour that exists today.

(Also noting the changeset bot: this PR is migration + plan with no package source change, so I believe none is needed, but flagging it rather than assuming.)

---

AN-BC:
@MS-BC — pushed the logging work into this PR. Two halves, and the second is the one worth your eye before you pick this up.

## Provenance (the easy half)

`ActionExecutionLog` gains four nullable columns — `EntityActionID`, `EntityActionInvocationTypeID`, `TargetEntityID`, `TargetRecordID`. Extended rather than a side table: the child is optional, small and 1:1, so nullable columns beat a join and *"did this action run"* stays one query. Same pattern accounting already uses with `JournalEntry.LinkedEntityID`/`LinkedRecordID`.

`TargetEntityID` is denormalized rather than derived through `EntityActionID` on purpose — it survives the binding being deleted or retargeted, and lets the log be queried by record with no join. Kept generic, since every invoker has a subject. The *invoker* I deliberately left concrete rather than generic: Record Processes and agent steps already have `RecordProcessRunDetail` and `AIAgentRunStep`, so a generic pair would duplicate them and give up FK integrity.

## Param logging (the half that changes existing behaviour)

`StartActionLog` writes `JSON.stringify(params.Params)` on every run; `EndActionLog` writes the merged set again. Unconditionally, no opt-out.

Fine today because Entity Actions are barely used. **Not fine the moment they're the workflow substrate**, because entity-action params are whole records — `'Entity Object'` and the new `'Entity Object Data'` put the entire row into the log, twice per invocation. An `AfterUpdate` binding on a busy entity writes the full record on every save. That's the space problem (the `MAX` payload, not the row) and the security problem (message bodies, Person fields, contract terms in a log with broad read access — `RetentionPeriod` deletes it eventually, which isn't the same as never writing it).

**Posture is fail-closed:**

| | Rule | Where |
|---|---|---|
| 1 | **Hard engine rule, no configuration** — params of `ValueType` `'Entity Object'` / `'Entity Object Data'` are **never** logged. Whole records by definition. | Engine — **not yet built** |
| 2 | `ActionParam.LogValue` — the definition declares loggability. Default `1`. | Migration ✅ |
| 3 | `EntityActionParam.LogValue` — per-binding override, `NULL` inherits. Can't re-enable what rule 1 suppresses. | Migration ✅ |
| 4 | `EntityAction.LoggingMode` — `All` / `FailuresOnly` / `None`. | Migration ✅ |

Rule 1 closes the hole on its own; 2–4 handle the grey area (a `Static` param holding an API key, an `Entity Field` holding a national ID) and volume. Suppressed params still log name, type and `ByteLength` — so "the payload was enormous" stays diagnosable without being readable.

## New: §6, the post-CodeGen runbook

Ordering matters more than usual here. `mj sync push` **before** `mj codegen` or stale JSONType definitions silently truncate generated interfaces. Verify `EntityActionParam.ValueType`'s value list regenerated from the CHECK constraint — never hand-insert `EntityFieldValue`. Then `MapParams`, then the redaction rules.

⚠️ **Steps 4 and 5 gate production use.** Until rule 1 ships, the migration has made it *possible* to author a binding that passes a whole record and *not yet safe*. Flagged in the migration header too.

## Two corrections against `migrations/CLAUDE.md`

- **Dropped the two hand-authored indexes.** The repo rule is that indexes aren't hand-written unless requested and CodeGen owns FK indexes. The two composites I'd added — `EntityAction (ScopeEntityID, ScopeRecordID)` for the config UI's reverse lookup, and `ActionExecutionLog (TargetEntityID, TargetRecordID, StartedAt DESC)` for "what ran against this record" — are now step 10 of the runbook, measure-first.
- **Renamed `v5.51.x` → `v5.52.x`** per the increment-the-minor convention.

Two open questions I'd rather you called than I assumed, both in §9: whether `ActionParam.LogValue` should default `0` instead of `1` (I kept `1` to preserve existing behaviour, since rule 1 closes the hole regardless — but I'd set `0` explicitly on `Execute Agent`'s `Data` in the same pass), and whether suppressed values should be hashed rather than omitted (useful, but a hash is a fingerprint of the record, so it deserves its own decision).

---

AN-BC:
Done — and yes, seeded in metadata: `metadata/actions/.execute-agent.json`. Open question 5 is resolved: **keep the `LogValue` default at `1`, flip Execute Agent explicitly.**

Four params now carry `LogValue: false`:

| Param | Direction | Why |
|---|---|---|
| `Data` | Input | The agent payload — **a whole record** whenever a binding maps a row onto it |
| `ConversationMessages` | Input | Conversation text |
| `Payload` | Output | Whatever the agent produced from that record |
| `AgentResult` | Output | The full run trace, payload included |

**The outputs matter as much as the inputs.** `EndActionLog` writes `result.Params ?? params.Params` — the *merged* input and output set — so flagging only `Data` and `ConversationMessages` would have left the same content in the log by a different door.

**The six identifier params stay logged** — `AgentName`, `AgentID`, `ConversationDetailID`, `LastRunID`, `MaxExecutionTimeMs`, `AgentRunID`. That's the point rather than a compromise: a run is still fully diagnosable from the log (which agent, on which conversation, producing which run), and the content is one hop away in **`MJ: AI Agent Runs`**, where it belongs and is already access-controlled. The content isn't lost — it stops being *duplicated* into a general-purpose log.

For the common case it's belt and braces: a binding mapping a record onto `Data` is suppressed by the hard engine rule (whole-record `ValueType`) **and** by the param's own flag.

Also added:
- **§5.5** documenting the split and why the identifiers stay
- **Runbook step 2b** — this file can't be pushed until step 1 and CodeGen have run, since `ActionParam.LogValue` doesn't exist before then

No `sync` blocks touched — those belong to the release-time consolidated push per `metadata/CLAUDE.md` rule 1b.

One open question left in §9: whether suppressed values should be **hashed** rather than omitted. Useful for proving two runs received the same payload without storing it — but a hash is a fingerprint of the record, so I'd rather that were a deliberate call than something that rode along with this.

---

AN-BC:
All four in. One needed schema; the rest are specified with runbook steps, consistent with this PR shipping migration + plan rather than engine code.

## 1. `Params` stops being overwritten — `ResultParams` added

`Params` is written twice against one column: inputs at start, merged set at end. Because Custom and Generated actions mutate `params.Params` **in place**, the end state isn't "inputs plus outputs" — it's *inputs as the action left them*. So the values the action was actually **called with** are captured and then destroyed.

For a workflow binding that's the difference between *"the agent got the wrong record"* and *"the agent mangled the right one."*

| Column | Holds |
|---|---|
| `Params` | The **as-called** inputs. Written once, never overwritten. |
| `ResultParams` | The final merged set at completion. NULL in flight and for runs that never finished. |

`QueueTask` already separates `Data` / `Options` / `Output`, so this is the existing house shape rather than a new one.

## 2. Redaction belongs to the params, not to the log

**This one was a defect in my own plan.** If the rules live inside `StartActionLog`/`EndActionLog` they cover one of at least three persisters:

| Persister | Column |
|---|---|
| `ActionEngine` | `ActionExecutionLog.Params` / `ResultParams` |
| `QueueManager` | `QueueTask.Data` |
| `AgentRunner` | agent run payload |

Runbook **step 9 routes `After*` through the queue** — so building redaction inline would have re-opened, one layer down, exactly the hole the rest of the plan closes. Now specified as one exported `RedactParams` helper, with the invariant stated for grepping: **no path may write a raw `ActionParam[]` to persistent storage.**

## 3. Retention is documented but not enforced

`RetentionPeriod` says *"number of days to retain the log; NULL for indefinite"* — and nothing outside generated code reads it, on either `ActionExecutionLog` or `Action`.

So "retention deletes it eventually" was never available as mitigation, and **I had leaned on it** in the earlier comment. Corrected. Specified as a bounded Scheduled Job resolving most-specific-first (log row → action → config default), with the note that a per-action value nobody sets is not a retention policy — the config default is the real control. Pre-existing and separable if you'd rather it go out on its own.

## 4. Redaction records shape, not nothing

```jsonc
{ "Name": "Data", "Type": "Input", "ValueType": "Entity Object Data",
  "Logged": false, "Reason": "WholeRecordValueType",
  "ByteLength": 4182, "KeyCount": 41,
  "Keys": ["ID","DealNumber","Name","PipelineID","PipelineStageID","..."] }
```

`ByteLength` + `KeyCount` make *"it was called with the 41-column deal row, not the 3-field summary"* answerable, which is most of what an incident needs. Top-level `Keys` are column names for a record — schema, not data. Nested keys deliberately **not** walked, since deeper structure can encode content (a `Notes` object keyed by author, say).

---

Also corrected the `Message` column description, which claimed to hold *"JSON-formatted output data or response from the action execution."* It's set from `result.Message` — a human-readable summary. Outputs were never there, and someone reading the schema would have gone looking in the wrong column.

New open question in §9: should `ResultParams` be written on **failure** too? A refused or thrown run has no result set, but the partial mutation of the inputs is often the most diagnostic thing available. I'd say yes, redacted the same way.

---

AN-BC:
Done — `ResultParams` is written on failure under the same rules as success. Open question 6 closed.

**It costs no new control flow**, which I hadn't expected. Both failure paths in `ActionEngine.InternalRunAction` already reach `EndActionLog`:

- the returned `Success: false` case, and
- the thrown-exception `catch`, which constructs its result with `Params: params.Params` — **the array as the action mutated it before dying**.

So the partial mutation was already in hand. It was simply being written into a column that then looked identical to a successful run's, which is why it wasn't useful. Splitting the columns is what makes it an audit trail.

**One consequence worth noting:** `ResultParams` being NULL now means exactly one thing — *the run never finished* (process died, host killed). That's a signal rather than an absence, so the plan says explicitly not to backfill it.

**Implementation note added**, because there's a trap in the existing code: `StartAndEndActionLog` performs a single force-persist write rather than an INSERT followed by an UPDATE (used by the timeout/abort path at ~line 222). That path must persist **both** columns — `Params` is set in memory by `StartActionLog(params, false)` and has to survive, and `EndActionLog` must never clear or rewrite it. Easy to miss when the change looks like "just stop overwriting `Params`."

---

AN-BC:
It does — and checking that made me realise **my implementation note was wrong.** Corrected in the plan.

Logging is BaseEntity throughout: `StartActionLog` does `md.GetEntityObject<MJActionExecutionLogEntity>('MJ: Action Execution Logs', ContextUser)` + `NewRecord()`, and persistence runs through `_logQueue`, a `BaseEntitySaveQueue`, so the writes stay off the critical path. `StartAndEndActionLog` is the same two calls with `saveRecord: false` — one write instead of two, because `Update()` on a never-inserted `NewRecord()` inserts.

So my warning ("make sure that path persists both columns, `Params` must survive") was overstated. **One entity instance carries both fields, so a single `Save()` persists both.** `Params` survives that path with no special handling.

## The real problem is *when* `StartActionLog` runs

Three `StartAndEndActionLog` call sites:

| Call site | Runs | `Params` is |
|---|---|---|
| Input validation failed | before the action | a true as-called snapshot |
| Filters said don't run | before the action | a true as-called snapshot |
| **Timeout / abort** | **after the action has been running** | **whatever the array holds by then** |

Actions mutate `params.Params` in place. So on the timeout path, `Params` is a *post-hoc* capture that merely lands in the as-called column — the guarantee silently doesn't hold, on the one path where *"what did we send it?"* is the first question anyone asks.

**Fix rather than document:** hoist the input snapshot to the top of `RunAction`, before any branch, and have every path write `Params` from it. Cheap, and it makes the column mean one thing everywhere instead of two things depending on how the run ended. Runbook step 5b now carries it.

---

AN-BC (commented):
LGTM. @MS-BC after running the migration and codegen please test this out and add unit tests and integration tests that test this all out extensively before tagging me for final review.

AN-BC:
@MS-BC — picking this up, here's the shape of it. **This PR is migration + plan + metadata only. Every column it adds is inert** — nothing in the engine reads `LogValue`, `LoggingMode`, `ResultParams` or the four provenance columns yet. So "run the migration and test it" would only prove CodeGen emitted properties. The actual work is §6 steps 4–9 of [`plans/entity-action-workflow-extensions.md`](https://github.com/MemberJunction/MJ/blob/claude/sales-deal-management-app-ueporb/plans/entity-action-workflow-extensions.md), and the plan is written to be followed in order.

## Setup — three steps, and there's one easy to lose between them

```
1.  mj migrate
2.  mj sync push --include=entities   →   mj codegen      (push BEFORE codegen)
2b. mj sync push  metadata/actions/.execute-agent.json
3.  verify EntityActionParam.ValueType regenerated
```

**Step 2 order is not cosmetic** — CodeGen reads JSONType definitions from the database, so pushing after it regenerates from stale definitions and *silently deletes* properties from the generated types.

**Step 2b is the one that gets lost.** `.execute-agent.json` already carries `LogValue: false` on `Data` / `ConversationMessages` / `Payload` / `AgentResult`, but it **cannot** be pushed before step 1 — the column doesn't exist. Push it too early and it errors; skip it after and the flags quietly aren't in effect. Revert the `sync` block write-back afterwards; those belong to the release-time consolidated sync.

**Step 3:** `'Entity Object Data'` must appear in both the `EntityFieldValue` rows and the generated TypeScript union, both derived from the CHECK constraint. **Never hand-insert `EntityFieldValue`** — if it didn't regenerate, that's a CodeGen problem to fix, not to paper over.

## Build order

**Step 4 first — it's ten lines and unblocks everything else.** `MapParams` gains the `'Entity Object Data'` case → `entity.GetAll()`. Without it there's nothing safe to pass an agent, because a live `BaseEntity` serializes to `{}` (fields are getters, not enumerable own properties).

**Step 5 is the gate.** `RedactParams`, with rule 1 first: params of `ValueType` `'Entity Object'` / `'Entity Object Data'` are never logged, no configuration, no override. Build it as **one exported helper**, not inline in `StartActionLog`/`EndActionLog` — that was a defect in my own first draft, and step 9 is why: routing `After*` through `QueueManager` puts the same params in `QueueTask.Data`, which is persistent storage too. The invariant to hold, stated so it's greppable: **no path may write a raw `ActionParam[]` to persistent storage.**

⚠️ **Until step 5 lands, nobody authors a binding that passes a whole record.** The migration made that *possible*; it isn't yet *safe*. If this merges ahead of the engine work, that's a release where the schema invites exactly the pattern the plan exists to prevent — so either bring 4–9 into this PR, or hold the window deliberately and say so.

**Step 5b** has a subtlety worth reading §5.6 for: `Params` stops being overwritten and `ResultParams` takes the final set — but also **hoist the input snapshot to the top of `RunAction`**. Two of the three `StartAndEndActionLog` call sites run before the action; the timeout/abort one runs *after*, and since actions mutate `params.Params` in place, that path writes a post-hoc capture into the as-called column. Same column, two different meanings, on the one path where "what did we send it?" is the first question anyone asks.

Then **6** (Sequence + scope filtering), **7** (stamp the provenance columns), **8** (filter contract), **9** (`OnAfterSaveExecute` → `QueueManager`, payload through `RedactParams`). **4, 5 and 8 are independent** and can go in parallel; 6 gates the Explorer UI work in 11.

## Tests — five that matter more than coverage

A conventional suite passes without any of these, which is the problem:

1. **Rule 1 cannot be overridden.** Set `EntityActionParam.LogValue = 1` on an `Entity Object Data` param, assert the value is *still* suppressed. Negative test, most important one here.
2. **Redaction holds at every persister, not just `ActionEngine`.** Exercise the `After*` queue route and assert `QueueTask.Data` is redacted. A suite that only covers `ActionExecutionLog` passes an inline implementation that leaks one layer down.
3. **`Params` is as-called on the timeout path.** Action mutates `params.Params`, then exceeds its timeout — the log must show the *original* values. The other two call sites run before the action and pass trivially, so testing only those proves nothing.
4. **`'Entity Object Data'` produces a non-empty payload.** Assert the serialized `Data` actually carries the record's fields — that's the regression guard on the `{}` bug the value type exists for.
5. **Backward compatibility.** An existing `EntityAction` with every new column NULL behaves byte-identically. That's what makes this additive rather than a behaviour change.

**2 and 3 belong in the integration tier**, not unit — they span save path → engine → queue → persistence, which is precisely the seam unit tests mock away. `npm run test:integration`.

Also worth asserting: `ResultParams` is written on failure (both the returned `Success: false` path and the thrown-exception `catch`), and that NULL therefore means exactly one thing — *the run never finished*. Don't backfill it.

## One thing that isn't yours to decide

§9 #7 — whether suppressed values are **hashed** rather than omitted — is a fork inside `RedactParams`, so it'll get resolved by default by whoever writes it. A hash is a fingerprint of the record, so leave it omitted and let @AN-BC call it. Everything else in §9 is closed.

Ping me with questions as you go rather than at the end — §5 in particular is easier to get right the first time than to retrofit.

---

migrations/v5/V202608040633__v6.1.x__CodeGen_Run_2026-08-04_06-33-06.sql
```diff
@@ -0,0 +1,3254 @@
+/* SQL text to insert 11 new entity field(s) */
+
```
AN-BC, line 367: @MS-BC is there a reason this is separate file from the migration itself, usually we cat this codegen output with 50 whitespace lines and a header comment block at bottom of the hand written DDL migration.

AN-BC (commented):
LGTM, after addressing the sole comment below, please merge. Thanks @MS-BC

rkihm-BC (commented):
## Code review — EntityAction as the workflow-hook substrate

**Verdict: careful, well-tested, cleanly additive. No correctness blockers. The remaining items are process/hygiene (refresh the now-stale body, re-tag for the deferred final review), one existing-column semantic change to surface in release notes, and a few residual redaction edges worth a follow-up.**

Reviewed the full diff, the migration, the engine, and the test tiers — and, this pass, the PR conversation and CI signals too. Built `actions-base` (clean) and ran the relevant unit tiers locally: **actions-base 109/109, Actions/Engine 64/64** green. CodeGen output is committed and consistent with the hand migration (`'Entity Object Data'` regenerated into the `ValueType` union from the CHECK, `LoggingMode` union generated, all new columns present in the ORM).

**Context from the threads (so this is read against the right baseline):** the engine + tests that landed on Aug 4 are the *planned completion* of this PR — AN-BC's review said "LGTM … after running the migration and codegen please … add unit tests and integration tests … before tagging me for final review," and that is exactly what the later commits did. So this is not scope creep; it's the deferred half, and the final review AN-BC asked for hasn't happened yet.

### What's good

- **The migration is clean and correctly additive.** Every new column is nullable or defaulted to today's semantics (`Sequence` DEFAULT 0, `LoggingMode` DEFAULT 'All', `LogValue` DEFAULT 1). The `ValueType` change uses the drop-then-add-CHECK-in-one-migration pattern MJ requires, and CodeGen regenerated the value-list from the constraint rather than hand-inserting `EntityFieldValue` rows. FK targets verified, no hand-authored indexes, extended properties present on every new column, DDL-only.
- **Redaction is a real fail-closed choke point.** `RedactParams` is the single producer for every persister; rule 1 (whole-record `ValueType`) beats all configuration and cannot be re-enabled by either `LogValue` flag — and the tests assert the *negative* (the secret value is absent from the serialized output), not just a reason code. Shape-only, top-level keys only, `-1` byte-length sentinel for unserializable values, `BaseEntity`→`GetAll()` handling that avoids the empty-shape getter trap.
- **The snapshot-before-mutation fix is correct.** `SnapshotInputParams` is taken at the top of `RunAction` and threaded through all four exit paths (validation fail, filter refusal, timeout, normal). There's a direct test with a real mutating action proving `Params` captures only the as-called inputs.
- **`Params`/`ResultParams` split is coherent on every path**, including both failure paths (returned `Success:false` and the thrown-exception catch). NULL `ResultParams` means exactly "never finished." `LoggingMode` (`All`/`FailuresOnly`/`None`) gating is consistent across all exits and avoids returning a `LogEntry` pointing at a row it never inserted.
- **Scope resolution is fail-closed and additive.** Unscoped bindings (every existing one) short-circuit to "applies," so behavior is unchanged. A scoped binding that can't be resolved *declines to fire* — the safe direction — and the default FK-walk returns `null` (not `false`) on an indirect or ambiguous relationship rather than guessing "the" foreign key. Checked at the single funnel point (single-record invocation) so it can't be forgotten; out-of-scope returns `null`, which callers filter, so it never drags a consolidated `Success` down. The `GenericDatabaseProvider` change is just that null-filter — the save path is untouched for existing bindings.
- **Provenance** stamps binding/event/record on the log, with `TargetRecordID` using `CompositeKey.ToConcatenatedString()` to line up with `RecordChange.RecordID`.
- **The integration harness reads log rows back via raw SQL**, independent of the object graph that wrote them — genuine verification, not self-confirmation.

### Concerning

1. **The PR body is stale relative to the diff — refresh it and re-tag for final review.** The description still reads as schema-only ("Engine work is specified, not built"; "⚠️ Before merge: CodeGen has not been run"), but the engine, the committed CodeGen migration, and the tests are all here now. Since the reviewers know it evolved (this was the requested completion), this isn't a hidden-scope problem — it's a housekeeping step: update the body to describe what's actually shipping, and tag AN-BC for the final review he deferred.

2. **A semantic change to an existing column, framed as purely additive.** `ActionExecutionLog.Params` used to hold the final *merged* set; it now holds *as-called inputs*, with the merged set moved to the new `ResultParams`. This is the right design, but any existing consumer (dashboard/report/query) reading `Params` for outputs will now silently get inputs. That should be called out explicitly in the changeset/release notes — it's a repurposing, not just a new nullable column. (The column's extended-property description is updated, which helps.)

3. **A load-bearing invariant is proven only outside CI.** "LoggingMode gates logging but *never* execution" is asserted only in the DB-dependent integration script, which isn't run in CI. On the CI tier alone, a regression that skipped execution under `None`/`FailuresOnly` would pass — the unit `None` test checks that nothing was logged but never asserts the action body ran. Add that one unit assertion.

### Residual redaction edges (hardening, not blockers)

- **Undeclared output params bypass the opt-out rules.** Redaction is keyed on declared `ActionParam` rows. The documented `addOutputParam` pattern lets a custom action push an *undeclared* output; with no definition it has nothing to opt out with (rule 3 defaults to logged) and no binding (rules 1–2 can't fire), so it lands in `ResultParams` in the clear. The declared Execute-Agent path (`Data`/`Payload`/`AgentResult` ship `LogValue:false`) is fully covered, but the "general workflow substrate" framing means arbitrary actions can still emit undeclared sensitive outputs. Worth a documented note or a stricter default for undeclared params.
- **Shape recording emits top-level key names.** For a record those are column names (schema — fine). For an object keyed *by content* (a map whose keys are IDs/secrets), the key names themselves are content. The "keys = schema" assumption holds for records, not universally.
- "Nested keys are not walked" is only implicitly tested (the redacted secret in that test is a top-level value, so the depth-stop isn't directly asserted).

### CI / repo hygiene

- The **dependency-check bot flags 5 missing dependencies** (`ng-base-types` in action-gallery / agent-requests / entity-action-ux / entity-communication; `ng-shared` in file-storage). I checked: **none of those 5 packages are touched by this PR, and the missing deps already exist on `next`** — this is pre-existing repo debt the check surfaces on every PR, **not introduced here**. Flagging it only so the red X isn't conflated with this change; it's separable and shouldn't block #3408.
- **Retention remains documented-but-unenforced.** Redaction bounds per-row size, but nothing caps row count; the Scheduled Job (§5.8) isn't built. Pre-existing and separable, but until it lands, "retention as mitigation" isn't available — worth stating plainly rather than implying the log is bounded.

### Bottom line

The engineering is sound: fail-closed redaction at a single choke point, a correctly-timed input snapshot, fail-closed scoping, and a genuinely additive schema that leaves existing bindings and direct invocations unchanged. Nothing needs to change for correctness. Before merge: refresh the description and re-tag AN-BC for the final review he deferred, surface the `Params` semantic change in the release notes, and add the one execution-not-gated unit test. The residual redaction edges and the pre-existing dependency-check finding are follow-ups, not blockers.

*Reviewed with Claude Code.*

AN-BC:
> ## Code review — EntityAction as the workflow-hook substrate
> **Verdict: careful, well-tested, cleanly additive. No correctness blockers. The remaining items are process/hygiene (refresh the now-stale body, re-tag for the deferred final review), one existing-column semantic change to surface in release notes, and a few residual redaction edges worth a follow-up.**
> 
> Reviewed the full diff, the migration, the engine, and the test tiers — and, this pass, the PR conversation and CI signals too. Built `actions-base` (clean) and ran the relevant unit tiers locally: **actions-base 109/109, Actions/Engine 64/64** green. CodeGen output is committed and consistent with the hand migration (`'Entity Object Data'` regenerated into the `ValueType` union from the CHECK, `LoggingMode` union generated, all new columns present in the ORM).
> 
> **Context from the threads (so this is read against the right baseline):** the engine + tests that landed on Aug 4 are the _planned completion_ of this PR — AN-BC's review said "LGTM … after running the migration and codegen please … add unit tests and integration tests … before tagging me for final review," and that is exactly what the later commits did. So this is not scope creep; it's the deferred half, and the final review AN-BC asked for hasn't happened yet.
> 
> ### What's good
> * **The migration is clean and correctly additive.** Every new column is nullable or defaulted to today's semantics (`Sequence` DEFAULT 0, `LoggingMode` DEFAULT 'All', `LogValue` DEFAULT 1). The `ValueType` change uses the drop-then-add-CHECK-in-one-migration pattern MJ requires, and CodeGen regenerated the value-list from the constraint rather than hand-inserting `EntityFieldValue` rows. FK targets verified, no hand-authored indexes, extended properties present on every new column, DDL-only.
> * **Redaction is a real fail-closed choke point.** `RedactParams` is the single producer for every persister; rule 1 (whole-record `ValueType`) beats all configuration and cannot be re-enabled by either `LogValue` flag — and the tests assert the _negative_ (the secret value is absent from the serialized output), not just a reason code. Shape-only, top-level keys only, `-1` byte-length sentinel for unserializable values, `BaseEntity`→`GetAll()` handling that avoids the empty-shape getter trap.
> * **The snapshot-before-mutation fix is correct.** `SnapshotInputParams` is taken at the top of `RunAction` and threaded through all four exit paths (validation fail, filter refusal, timeout, normal). There's a direct test with a real mutating action proving `Params` captures only the as-called inputs.
> * **`Params`/`ResultParams` split is coherent on every path**, including both failure paths (returned `Success:false` and the thrown-exception catch). NULL `ResultParams` means exactly "never finished." `LoggingMode` (`All`/`FailuresOnly`/`None`) gating is consistent across all exits and avoids returning a `LogEntry` pointing at a row it never inserted.
> * **Scope resolution is fail-closed and additive.** Unscoped bindings (every existing one) short-circuit to "applies," so behavior is unchanged. A scoped binding that can't be resolved _declines to fire_ — the safe direction — and the default FK-walk returns `null` (not `false`) on an indirect or ambiguous relationship rather than guessing "the" foreign key. Checked at the single funnel point (single-record invocation) so it can't be forgotten; out-of-scope returns `null`, which callers filter, so it never drags a consolidated `Success` down. The `GenericDatabaseProvider` change is just that null-filter — the save path is untouched for existing bindings.
> * **Provenance** stamps binding/event/record on the log, with `TargetRecordID` using `CompositeKey.ToConcatenatedString()` to line up with `RecordChange.RecordID`.
> * **The integration harness reads log rows back via raw SQL**, independent of the object graph that wrote them — genuine verification, not self-confirmation.
> 
> ### Concerning
> 1. **The PR body is stale relative to the diff — refresh it and re-tag for final review.** The description still reads as schema-only ("Engine work is specified, not built"; "⚠️ Before merge: CodeGen has not been run"), but the engine, the committed CodeGen migration, and the tests are all here now. Since the reviewers know it evolved (this was the requested completion), this isn't a hidden-scope problem — it's a housekeeping step: update the body to describe what's actually shipping, and tag AN-BC for the final review he deferred.
> 2. **A semantic change to an existing column, framed as purely additive.** `ActionExecutionLog.Params` used to hold the final _merged_ set; it now holds _as-called inputs_, with the merged set moved to the new `ResultParams`. This is the right design, but any existing consumer (dashboard/report/query) reading `Params` for outputs will now silently get inputs. That should be called out explicitly in the changeset/release notes — it's a repurposing, not just a new nullable column. (The column's extended-property description is updated, which helps.)
> 3. **A load-bearing invariant is proven only outside CI.** "LoggingMode gates logging but _never_ execution" is asserted only in the DB-dependent integration script, which isn't run in CI. On the CI tier alone, a regression that skipped execution under `None`/`FailuresOnly` would pass — the unit `None` test checks that nothing was logged but never asserts the action body ran. Add that one unit assertion.
> 
> ### Residual redaction edges (hardening, not blockers)
> * **Undeclared output params bypass the opt-out rules.** Redaction is keyed on declared `ActionParam` rows. The documented `addOutputParam` pattern lets a custom action push an _undeclared_ output; with no definition it has nothing to opt out with (rule 3 defaults to logged) and no binding (rules 1–2 can't fire), so it lands in `ResultParams` in the clear. The declared Execute-Agent path (`Data`/`Payload`/`AgentResult` ship `LogValue:false`) is fully covered, but the "general workflow substrate" framing means arbitrary actions can still emit undeclared sensitive outputs. Worth a documented note or a stricter default for undeclared params.
> * **Shape recording emits top-level key names.** For a record those are column names (schema — fine). For an object keyed _by content_ (a map whose keys are IDs/secrets), the key names themselves are content. The "keys = schema" assumption holds for records, not universally.
> * "Nested keys are not walked" is only implicitly tested (the redacted secret in that test is a top-level value, so the depth-stop isn't directly asserted).
> 
> ### CI / repo hygiene
> * The **dependency-check bot flags 5 missing dependencies** (`ng-base-types` in action-gallery / agent-requests / entity-action-ux / entity-communication; `ng-shared` in file-storage). I checked: **none of those 5 packages are touched by this PR, and the missing deps already exist on `next`** — this is pre-existing repo debt the check surfaces on every PR, **not introduced here**. Flagging it only so the red X isn't conflated with this change; it's separable and shouldn't block [Entity Action as the workflow-hook substrate — scope binding, sequencing, and a safe execution log #3408](https://github.com/MemberJunction/MJ/pull/3408).
> * **Retention remains documented-but-unenforced.** Redaction bounds per-row size, but nothing caps row count; the Scheduled Job (§5.8) isn't built. Pre-existing and separable, but until it lands, "retention as mitigation" isn't available — worth stating plainly rather than implying the log is bounded.
> 
> ### Bottom line
> The engineering is sound: fail-closed redaction at a single choke point, a correctly-timed input snapshot, fail-closed scoping, and a genuinely additive schema that leaves existing bindings and direct invocations unchanged. Nothing needs to change for correctness. Before merge: refresh the description and re-tag AN-BC for the final review he deferred, surface the `Params` semantic change in the release notes, and add the one execution-not-gated unit test. The residual redaction edges and the pre-existing dependency-check finding are follow-ups, not blockers.
> 
> _Reviewed with Claude Code._

Thanks @rkihm-BC @MS-BC please have CC implement Robert's suggested improvements and after done you can merge.

MS-BC:
@AN-BC @rkihm-BC — Robert's three pre-merge items are done. Pushed to `claude/sales-deal-management-app-ueporb`.

**1. The one missing unit assertion — "LoggingMode gates logging but never execution."**
Robert's point was exact: the `None` test asserted zero rows, but zero rows is also what a regression that skipped the action entirely looks like. The invariant was only proven in the DB-dependent integration script, which CI does not run — so on the CI tier alone that regression passed silently.

Added a body-execution counter to the fixture actions and asserted it across `All` / `FailuresOnly` / `None` (parameterised) plus `SkipActionLog`, alongside the returned result and the action's side effect on the caller's param array — so the test fails if suppression ever starts gating the run:

```
✓ 'All' gates logging but NEVER execution — the action body still runs and its result is returned
✓ 'FailuresOnly' gates logging but NEVER execution — …
✓ 'None' gates logging but NEVER execution — …
✓ 'SkipActionLog' suppresses the row without suppressing the run
```
The existing `'None' writes nothing, even on failure` test also now pins `bodyRunCount === 1`.

Suites run locally against this branch: **Actions/Engine 180/180**, **actions-base 109/109**.

**2. The `Params` semantic change is now surfaced in the changeset**, not just the extended property. Called out under its own ⚠️ heading as the one non-additive change: `ActionExecutionLog.Params` moves from the final *merged* set to the *as-called inputs*, with the merged set going to `ResultParams` — so any dashboard, report or query reading `Params` for an action's **outputs** now silently gets its **inputs** and must be repointed. The changeset's "specified for the engine (not built in this change)" paragraph was also stale and is rewritten to describe what actually ships. Robert's residual items (undeclared `addOutputParam` outputs defaulting to logged, shape keys being content for content-keyed maps, unenforced log retention) are recorded there as tracked follow-ups.

**3. The PR body is still stale** — it reads as schema-only ("Engine work is specified, not built"; "⚠️ Before merge: CodeGen has not been run") when the engine, the committed CodeGen output and the tests are all present. It's your description, so I have not edited it — say the word and I'll rewrite it to match what ships.

On the CI red: the dependency-check finding Robert separated out is real and pre-existing on `next`. The same is true of the other two reds — `adopted standards` and `Run unit tests` — which I traced to two bugs on `next` head itself, not to any branch:
- `packages/Standards/src/checks/ui-layers.ts:77` holds the `new Metadata()` regex **literal**, and MJGlobal's `MultiProviderCompliance` test scans `packages/` for that construct — so the checker fails the standard it defines.
- `file-storage` is `mjUILayer: widgets` and declares `@memberjunction/ng-shared`; the source-side import is allow-marked, but a `package.json` cannot carry a comment, so the manifest half had no way to express the same exception.

Both verified present on pristine `origin/next` (`mjUILayerAllowedDeps` appears 0 times in the checker there). #3439 fixes both and is all-green; this branch already carries the same two files.

AN-BC:
@MS-BC please update PR body and merge

MS-BC:
One more, and it was a real one that had been invisible.

**`Check migrations` was failing on this branch — it just never got to say so.** `.github/workflows/changes.yml` declares `concurrency: group: "changes"` with `cancel-in-progress: true`. That group is a **constant**, so it is repo-wide rather than per-PR: any push to any open PR cancelled the in-flight migration check on every *other* PR. The gate could only report on whichever PR pushed most recently, and re-running one simply moved the cancellation somewhere else. Every other PR-scoped workflow in the repo (`test.yml`, `integration.yml`, `docs-site-ci.yml`) already keys its group by `${{ github.ref }}`; this one did not. Fixed in #3439.

With the gate actually reporting, it failed here:

```
Maximum existing migration timestamp: 202608041347
  (from V202608041347__v6.1.x__CodeGen_Introspection_View_Perf.sql)
::error:: Migration timestamp (202608011200) must be greater than the latest
  existing migration timestamp (202608041347 …)
```

`V202608041347__v6.1.x__CodeGen_Introspection_View_Perf.sql` landed on `next` after this branch was cut, so in filename order this migration would run **before** it on an upgrade — out of sequence. Renamed to `V202608042200__v6.1.x__EntityAction_Workflow_Extensions.sql`; **contents unchanged**.

While renaming I found the two references to it were also stale in a second way — both still carried the old `v5.52.x` version tag rather than the file's actual `v6.1.x`:
- `plans/entity-action-workflow-extensions.md` header link
- the `metadata/actions/.execute-agent.json` comment ("REQUIRES `ActionParam.LogValue`, added by migration …")

Both now point at the real filename. `.execute-agent.json` re-validated as parseable JSON.

Worth noting for the release: this same masking applied repo-wide, so a stale-timestamp migration on any branch could have merged without the gate ever running on it.

AN-BC:
@MS-BC you have a merge conflict, please carefully resolve and then you are clear to merge.

MS-BC:
@AN-BC both items are done.

**Merge conflict** — resolved; the branch is MERGEABLE against current `next` again.

**CodeGen output as a separate file** — no reason, that was wrong. Folded into the bottom of the hand-written DDL migration (`V202608042200__v6.1.x__EntityAction_Workflow_Extensions.sql`) per the usual convention and deleted `V202608040633__…CodeGen_Run_2026-08-04_06-33-06.sql`. That also fixed a real ordering bug the split created: the CodeGen file's timestamp (0633) was *earlier* than the DDL it depends on, so on a fresh upgrade Flyway would have run the generated objects before the columns existed.

**CI** — the two remaining reds were not this branch's:
- `Run unit tests`: `@memberjunction/ng-query-viewer` DOM tests, red on `next` itself for its last 5 commits. The viewer template binds `[ExportDataProvider]` on `<mj-query-data-grid>` and the real grid declares it, but the DOM test's `GridStub` didn't — NG0303 at render, failing 6 of 8 tests in the file. Fixed (test-only) and merged in here.
- `Check migrations`: was being *cancelled*, not failing — `changes.yml` used a constant concurrency group, so it is repo-wide and one PR's run cancels every other open PR's. Keyed it by ref like every other workflow in the repo. Both fixes also live standalone in #3439.

Robert's two items from his review are in as well: the unit assertion that `LoggingMode`/`SkipActionLog` gate logging but never execution (`bodyRunCount` pinned across All/FailuresOnly/None), and an explicit release-note callout for the `ActionExecutionLog.Params` repurposing. Actions/Engine 180/180, actions-base 109/109 green locally.

Not merging — that's yours to call.

AN-BC:
merged @MS-BC @rkihm-BC

PR #3409, opened 2026-08-02
API-key-scoped row filters for 6.1 — full implementation (WS1–WS3)

AN-BC:
## Summary

Full end-to-end implementation of API-key-scoped row filters for the **6.1 release**, per the reviewed design in `plans/api-key-row-filters.md` (v3.0). This PR evolved from docs-only to complete implementation at the owner's direction; every design decision from the three review rounds (@cadam11, @MarceloT-BC, @jordanfanapour) is implemented as specified.

**The gap closed:** MJ's row filtering bound to roles and sessions, never to an API key — two keys issued to the same user had identical row visibility. Now a key can be *less* than its owner: scope rules carry an optional `RowFilterID` into the same `RowLevelSecurityFilter` catalog role-RLS uses, enforced at the data layer through every existing RLS call site.

## Schema (v6.1)

`migrations/v6/V202608041553__v6.1.x__APIKey_Scope_RowFilterID.sql` — `RowFilterID` on `APIKeyScope` + `APIApplicationScope` (nullable FK), with the CodeGen output (EntityField rows, regenerated views/procs, FK indexes, grants) folded in behind the separator convention. PG counterpart generated through the SQLConverter pipeline. The PG parity/content tooling and workflow are generalized from hardcoded `v5` to version-folder pairs, opening the v6 cycle. (CodeGen ran in CI via a temporary branch-scoped workflow — this dev sandbox has no route to a SQL Server; the workflow is deleted now that its artifacts are committed.)

## Implementation highlights

- **WS1** — update RLS now checks the **post-image** after before-save hooks (`CheckUpdateRLSPostImage`): a caller can no longer move a row they own outside their own row scope. Fail-closed skip optimization (anything the decomposer doesn't fully understand runs the check). Fixes the client-reachable `CheckRecordRLS` PK-escaping bypass (finding M1) and the raw numeric projection branch (M5); typed `CAST(NULL AS <type>)` post-image nulls.
- **WS2** — multi-tenancy hardened: tenant ids validated at the boundary (400, never a silent unscoped session), escaped at construction, column resolved to a real stored field, provider-quoted identifiers (T-SQL brackets are invalid on PG); shared `CloneUserForSessionContext` ends the shared-`UserInfo` mutation race.
- **WS3 core** — `UserInfo` carries `APIKeyActingContext` (list-capable `ActingCompanyIDs`, canonical sort) + `APIKeyRowFilters` bindings; `MarkupFilterText` hardened (`undefined` = unresolved — was the literal string `"undefined"`, fail-open for negation shapes (M2); all values escaped; `(1=0)` match-nothing mode); **`GetEffectiveRowFilterWhereClause` composes role RLS AND key filters *outside the role-RLS exemption*** — the §5.5 catch — with all nine enforcement call sites migrated and the legacy method deprecated behind a no-new-callers repo guard.
- **WS3 engine** — evaluator collects filters from **every** matching allow rule (no shadowing; most-restrictive-wins); `Authorize()` rejects `full_access`+filter (§5.6.1) and denies `query:run`/`query:test`/`dataset:read`/`report:run` for filtered keys (§5.10), all with named cause + split-the-key remedy; missing/invalid `{{Acting*}}` tokens deny naming the token; startup invariant (filtered rules require enforcing + default-deny or `Config()` throws); save-time validation subclasses (exact-entity patterns, registered vocabulary, strict column validation, same-entity invariant across all filter referrers, `full_access` mutual exclusion both directions, filter-edit revalidation of every referrer); per-request binding stamping on a **cloned** session user + registered acting-context resolver as the deployment extension point.
- **v1 scope**: keys-only enforcement; the application-ceiling column ships unused (deferred to v2 per §9.3).

## Tests

**Unit (all local, this sandbox): 4,468 passed / 61 skipped / 1 failed** — the failure is `MJServer mjapi-bootstrap.test.ts`, pre-existing environmental (`mj_generatedentities` has no dist in the sandbox), unrelated to this change. Per package: MJCore 1,632 · GenericDatabaseProvider 862 · MJServer 779 · SearchEngine 427 · MJCoreEntitiesServer 384 · APIKeys Engine 205 · APIKeys Base 27 · integration-test-suite 152. Fail-open regression tests ([FO] in the plan §7) are all present: exemption independence, shadowing, `(1=0)` operator shapes, clone preservation, PK-escaping, skip-decomposer fail-closed, no-new-callers guard.

**Integration (deterministic tier):** six new checks `KF1–KF6` in the `rls-isolation` bundle — key-filter isolation, exemption independence, INV-1 cache-slot separation on a forced-cold cacheable slot, INV-2/INV-3 fingerprint↔WHERE byte-identity for the same principal, post-image update rejection E2E, and Load-by-PK enforcement. Validated by CI's Integration Tier on this PR (no DB in the dev sandbox).

## Review focus

- §5.5's exemption independence is the load-bearing correctness property — KF2 and the MJCore [FO] suite pin it.
- The `ScopeEvaluator` allow-semantics change (collect-all vs first-match) is the one behavioral change to existing scope evaluation; deny semantics are unchanged and the prior priority tests still pass.
- The `MarkupFilterText` hardening tightens **role RLS** too: negation-shaped filters over unresolved/undefined tokens move from permissive to restrictive (changeset calls this out for operators).
- `migrations/v6` naming (`v6.1.x`) assumes 6.1 is the next release; renumber if the sequence moves before merge.

---

https://claude.ai/code/session_01RBzhULDaQzufdzkWNCwDcS


AN-BC:
@cadam11 @rkihm-BC for your input too

cadam11:
5.9 looks like it overstates coverage. `entity:*` and `view:*` route through the nine RLS call sites, but `query:run` deliberately skips per-entity RLS (`GenericDatabaseProvider.ts:3140`, based on its comment) and `dataset:read` builds its SQL with no RLS clause, cached on `{EntityName, ExtraFilter}` alone. A filtered key plus either scope reads the entity unfiltered. It might be useful to build a scope-by-scope `does-RLS-apply` matrix, plus an explicit decision on whether filtered keys are denied the non-covered scopes.

Three other observations that the implementer should note:

- **Where does the binding live mechanically?** 5.5 composes in `EntityInfo`, but MJCore can't see `APIKeysEngineBase`; the entity/filter/permission-type binding is in `APIKeyScope` rows. If it gets stamped onto `UserInfo`: it's snapshotted at session build (staleness window), `MarkupFilterText` turns it into token surface, and the 5.8 exposure guard covers `APIKeyActingContext` but not the binding.

- **5.3 makes the feature deny-by-omission.** Anything not enumerated is unfiltered; filtering `Deals` doesn't filter `DealNotes`. For a blast-radius feature I'd suggest a key-level default filter or a deny mode for unlisted entities, or the tradeoff stated explicitly.

- **5.6 `3` and `4` fail open** (startup warning, documentation). Refusing startup when filtered rules coexist with either setting seems more consistent with the plan's own "fail closed, fail loudly."

WS1/WS2 could be done as their own PR, they're independently valuable and carry both the behavior change and the Postgres quoting change; moving RLS after `OnBeforeSaveExecute` reorders every before-save hook in the tree, probably worth a note in 3.5.

The general direction LGTM, and 5.5's exemption catch (`entityInfo.ts:2312`) is excellent.

AN-BC:
> 5.9 looks like it overstates coverage. `entity:*` and `view:*` route through the nine RLS call sites, but `query:run` deliberately skips per-entity RLS (`GenericDatabaseProvider.ts:3140`, based on its comment) and `dataset:read` builds its SQL with no RLS clause, cached on `{EntityName, ExtraFilter}` alone. A filtered key plus either scope reads the entity unfiltered. It might be useful to build a scope-by-scope `does-RLS-apply` matrix, plus an explicit decision on whether filtered keys are denied the non-covered scopes.
> 
> Three other observations that the implementer should note:
> 
> * **Where does the binding live mechanically?** 5.5 composes in `EntityInfo`, but MJCore can't see `APIKeysEngineBase`; the entity/filter/permission-type binding is in `APIKeyScope` rows. If it gets stamped onto `UserInfo`: it's snapshotted at session build (staleness window), `MarkupFilterText` turns it into token surface, and the 5.8 exposure guard covers `APIKeyActingContext` but not the binding.
> * **5.3 makes the feature deny-by-omission.** Anything not enumerated is unfiltered; filtering `Deals` doesn't filter `DealNotes`. For a blast-radius feature I'd suggest a key-level default filter or a deny mode for unlisted entities, or the tradeoff stated explicitly.
> * **5.6 `3` and `4` fail open** (startup warning, documentation). Refusing startup when filtered rules coexist with either setting seems more consistent with the plan's own "fail closed, fail loudly."
> 
> WS1/WS2 could be done as their own PR, they're independently valuable and carry both the behavior change and the Postgres quoting change; moving RLS after `OnBeforeSaveExecute` reorders every before-save hook in the tree, probably worth a note in 3.5.
> 
> The general direction LGTM, and 5.5's exemption catch (`entityInfo.ts:2312`) is excellent.

Very helpful, thanks for reviewing @cadam11. Will get the PR updated to reflect your input so that @jordanfanapour can implement it fully inclusive of these thoughts.

AN-BC:
@cadam11 All four points verified against source and adopted in the plan (now v2.2 on the branch) — thanks, the 5.9 catch in particular was a real overstatement.

**Coverage:** confirmed, and worse than flagged — the query cache fingerprint carries no user segment at all (`providerBase.ts:1859`), and `report:run` executes stored `ReportSQL` verbatim with no `contextUser` and no scope-metadata registration. New §5.10 has the full scope×RLS matrix. Decision: v1 wholesale-denies `query:run`/`query:test`/`dataset:read`/`report:run` for any filtered key. Your precise-deny variant is the documented relaxation, but asymmetric: datasets are implementable now (`DatasetItem.EntityID` non-null FK), queries are not — the `Query Entities` bridge fails open (no rows for migration-seeded queries; extractor swallows parse errors; empty extraction deletes bridge rows), so precise-deny there would treat "no bridge rows" as "touches nothing."

**Binding:** §5.5.1 — resolved in `context.ts` and stamped on the cloned per-request `UserInfo`; `EntityInfo` consumes plain data, so no dependency inversion. Staleness rebuilds per request from the scope cache (envelope = `scopeCacheTTLMs`, same as scope enforcement). On token surface: object-valued stamps never enter `{{UserX}}` — the guard is `typeof val !== 'object'` at `securityInfo.ts:470`, not enumerability (backing fields do enumerate). The §5.8 exposure test now covers the binding carrier too.

**Deny-by-omission:** adopted via the startup invariant plus a fix verification surfaced for the sharper case — `ScopeEvaluator` first-match-wins means a higher-priority unfiltered rule silently shadows a filtered one (invisible even in `EvaluatedRules`); §5.4 now ANDs every filter carried by any matching allow rule. Cross-entity reach stays explicit-per-entity, stated as the tradeoff; key-level default filter noted as an additive follow-up.

**5.6 #3/#4:** agreed, escalated to startup refusal — one invariant: any filtered rule requires enforcing + default-deny, else `Config()` throws.

**WS1/WS2:** split adopted. On hook ordering — nothing existing moves; the post-image check is a new step after `OnBeforeSaveExecute`, no existing hook reorders. The plan's "reorders the current sequence" wording caused that read; fixed.

*This response was produced by Fable 5 running in Ultracode mode (multi-agent source verification: 4 agents, ~120 tool calls against the actual code paths).*

---

AN-BC:
@rkihm-BC @jordanfanapour — coordination note so we don't double-work: @AN-BC's Claude Code (Fable on `Ultracode`) session is now subscribed to this PR and watching CI + review activity.

**Division of labor:**

- **I will handle:** any change requests from either of you against what's in this PR — the plan (`plans/api-key-row-filters.md`), the two migrations (T-SQL + PG counterpart), and the changeset. Also CI failures on this branch (just fixed two: the missing minor changeset and the missing `.pg.sql` counterpart — both gates now pass locally).
- **I will NOT touch:** the implementation itself. WS1/WS2 (the update post-image RLS fix + multi-tenancy hardening, now planned as their own PR per §8), Phase 2 (running the migration + CodeGen + appending its output), and WS3 (the key-filter code) are Jordan's per the handoff. Nothing will be pushed to those areas from here, so @jordanfanapour can build off the plan without any risk of racing this session.

**One decision worth an explicit ack before implementation starts** (flagged in §9.8): v1 wholesale-denies `query:run`/`query:test`/`dataset:read`/`report:run` for any key carrying a row-filtered rule, because those paths bypass RLS entirely and would read filtered entities unfiltered. The precise per-target relaxation is documented but asymmetric — implementable for datasets now, not safe for queries until the `Query Entities` bridge fails closed.

---

AN-BC:
**CI diagnosis — "Integration (SQL Server, deterministic)" red on `1e843f33f`:**

Ignore the wall of `MJ_API_KEY is not set` failures (IT03, IT15, IT23…) — those are **pre-existing and non-gating**: `next`'s green run at `169817375` shows the identical 19 errors (35/54, exit 0). They're client-transport members whose bootstrap throws on the missing key before reaching the "MJAPI is not reachable" skip-as-pass check — worth its own ticket, but not this PR.

The **two gating failures** are `IT13 - API Keys Engine Authorize` (oracle `api-keys.AK3`) and `IT36 - API Scope Enforcement` (`SE1/SE2/SE3/SE5`). All five oracles fail identically: creating an `APIKeyScope`/`APIApplicationScope` row dies with `Column name or number of supplied values does not match table definition` on `INSERT INTO @ResultTable EXEC spCreate...` — the provider builds an 11-column result table while the proc's trailing `SELECT * FROM vw...` now returns 12.

**Root cause:** this PR ships the DDL for `RowFilterID` without its CodeGen output. The CI pipeline's repeatable `R__RefreshMetadata` (`spUpdateExistingEntityFieldsFromSchema` + view/proc recompile) partially absorbs the new columns, leaving the view/proc/EntityField set mutually inconsistent. The plan's §8 Phase 2 specifies migration + CodeGen output as one unit — this failure is the deterministic tier correctly rejecting the intermediate state. Hand-authoring the missing EntityField/view/proc deltas is against repo doctrine (CodeGen owns them), and this session has no database to run CodeGen against.

Resolution is an owner decision (asked separately): either the migration pair + changeset move out of this PR into the Phase 2 implementation PR where CodeGen runs, or this PR stays red until CodeGen output is appended on this branch.

---

AN-BC:
**Resolution pushed (`0d36ce581`):** the migration pair + changeset moved out of the live trees into `plans/api-key-row-filters-staged/`, with an activation README. Net diff on `migrations/**` and `.changeset/` is now empty, so the changeset gate and PG parity check pass trivially and the integration tier runs against an unchanged schema — CI should match `next`'s green baseline on the re-run.

Nothing was lost: the staged folder carries the T-SQL migration, the PG counterpart, and the changeset verbatim. @jordanfanapour — Phase 2 step one is now `git mv` per the README (move into place → apply → `mj sync push` → `mj codegen` → append output behind the separator block → restore the changeset). The version segment (`v5.52.x`) should be re-checked against whatever is latest when you activate it.

This also hardens the plan with an empirical datapoint: §8 Phase 2's "migration + CodeGen output ship as one unit" is now demonstrated by CI, not just asserted.

---

MarceloT-BC (approved):
> 🔒 **Edited 2026-08-04.** Findings 1, 2, 5 and 6 and one ticket entry have been removed from this comment and reported privately to the maintainers instead — they described security-relevant defects in shipping code, which shouldn't sit in a public thread ahead of a fix. Nothing about the review's conclusions changed; the removed items were verified and adopted in plan v2.3. The rest of the review is unmodified.

> ⚠️ **Raw generated Claude output — unedited.** Produced by Claude Code (Opus 5) reviewing this PR at my direction. Source claims were verified by *reading* `next` at this PR's base — nothing was built, run, or tested. Treat the findings section as leads to confirm, not as established defects.

Approved — the plan is well grounded. I spot-checked twelve of its load-bearing source claims against `next` and all twelve held, including the two that matter most: the §5.5 exemption fail-open and the pre-image-only update check (`GenericDatabaseProvider.ts:3884`). The exemption catch in particular would otherwise have shipped a feature that passed every narrow-role test and did nothing for admins and service accounts — i.e. every principal that actually holds an API key.

Two acks, then findings.

## §9.8 — ack the deny, with one condition

Acked. I checked our side: `bizapps-accounting`, `bizapps-orders` and `bizapps-common` use `RunView` exclusively — no `RunQuery`, `RunReport` or `GetDatasetByName` anywhere (the only hits are unused imports in generated action boilerplate). The deny breaks nothing we run today.

The real issue is that the deny is **key-wide, not entity-wide**: attaching one row filter to one entity revokes `query:run` / `query:test` / `dataset:read` / `report:run` for the entire key. A partner key that does twenty things and needs a filter on one of them has to be split in two.

Where that lands on us: aging buckets, trial balance and consolidated P&L are exactly the workloads that outgrow `RunView` and get moved to a saved Query. The first time that happens on a filtered key, the key stops working — and the remedy, split the key, isn't discoverable from the error.

Pick one:

- **(a)** Keep the key-wide deny, document "one row filter ⇒ that key cannot run saved Queries, Datasets or Reports at all; split the key," **and require the `AuthorizationError` to name the cause** — "denied: this key carries a row filter" — not just the scope that failed.
- **(b)** Ship the per-dataset relaxation in v1, which §5.10 already says is implementable (`DatasetItem.EntityID` is a non-null FK).

My vote: **(a) now, (b) on the roadmap.** Don't hold the feature for the relaxation — but the error-message requirement isn't optional. It's the difference between a five-minute fix and a lost day six months out.

## §9.4 — register `ActingCompanyID`, and decide cardinality now

Company is the row-filter axis in accounting; `ActingOrganizationID` won't substitute for it. Please register it.

More important: **cardinality is a one-way door**, and the plan's "adding a token later is a typed code change rather than a schema change" doesn't hold for this part. Consolidated views are first-class in accounting — an executive across CO1+CO2+CO3, a shared-services user working several entities. A scalar token can only ever express `Col = '{{ActingCompanyID}}'`; moving to a set later changes the token's *shape* and breaks every filter already authored against it.

So decide now, either way:

- **List-capable from day one** — `{{ActingCompanyIDs}}` → `IN ('a','b','c')`, with the registered validator checking, quoting and escaping each element. Fits §5.2's vocabulary model directly, and composes with §5.4: an empty set resolves to `(1=0)`, same as an unresolved token. **My preference.**
- **Or scalar-only, documented explicitly as a one-way door**, accepting "one key per company."

`ActingCustomerID`: probably unnecessary — if customers are modeled as organizations, `ActingOrganizationID` covers the AR-portal case. Raising it only in case it hadn't been considered.

## Findings

> **For your Claude.** These come from reading source, not running anything — no build, no tests. Each carries a `file:line` and a stated cause so they're cheap to confirm or dismiss. Please verify before acting on any of them.

**1. — 🔒 removed, reported privately.** Verified and adopted into WS1 in plan v2.3.

**2. — 🔒 removed, reported privately.** Verified and adopted into §5.4 in plan v2.3.

**3. Missing cache invariant — add INV-3: the fingerprint and the WHERE must be computed for the same principal.** `ComputeRunViewRLSWhereClause` resolves `contextUser ?? this.CurrentUser` (`providerBase.ts:2191`); `GenericDatabaseProvider.ts:1590` uses its own `user`. Divergence is harmless today (role RLS, they generally agree). Once the clause is per-key, a divergence means the slot is fingerprinted for one principal and filled with another's rows. INV-1/INV-2 don't catch it — they cover *where* the clause is emitted and byte-stability, not *whose* it is.

**4. Keeping `GetUserRowLevelSecurityWhereClause` public with the old semantics is a fail-open trap.** §5.5 keeps it "for backward compatibility," but it's also the familiar name a future call site will reach for, and it silently omits key filters. Nine migrations is a one-time job; the trap is permanent. Suggest `@deprecated` plus a repo test asserting no non-test caller outside `GetEffectiveRowFilterWhereClause`.

**5. — 🔒 removed, reported privately.** Verified and adopted into the WS1 builder rewrite in plan v2.3.

**6. — 🔒 removed, reported privately.** Verified and adopted as its own ticket in plan v2.3.

Two nits: `ReplayOnly` skips the RLS checks (`databaseProviderBase.ts:1288-1325`) but executes no SQL (`:1393`), so no write escalation exists there — worth one line in WS1's scope confirming after-save hooks are safe on that path. And §2.6's "no `PlatformVariants` column" is right about the DB, but `RowLevelSecurityFilterInfo.PlatformVariants` + `GetPlatformFilterText()` do exist in code (`securityInfo.ts:414-443`) fed by a column that doesn't — worth rewording so nobody wires it up and assumes it works.

## Named in the plan as "own ticket," but no ticket exists

These evaporate when the plan doc goes stale. Suggest filing all of them now:

1. 🔒 removed, reported privately.
2. RLS exemption looseness — exemption granted from the mere presence of a filter-less permission row, without checking the permission is granted (§6, §9.7)
3. Query cache carrying no user segment (§5.10 closing note)
4. The `MJ_API_KEY is not set` integration-suite failures you flagged during CI triage

## CI

On `0d36ce581` only "Ensure migrations are valid" ran — Integration Tier and PG parity are path-filtered on `migrations/**` and never fired once the migration moved into `plans/`. Correct behavior for a docs-only PR; flagging only so the green check isn't read as validation of the staged migration.

AN-BC:
@MarceloT-BC All six findings verified against source before adoption — **all six held**, plus both nits. Plan is now v2.3 (`41cddc68d`). Point-by-point:

**Findings:** (M1) confirmed and adopted into WS1 — and your reachability analysis checks out: `ResolverBase.ts:1250-1253` loads the PK straight from client `OldValues___` when `TrackRecordChanges` is off, and the Load path's own comment at `:3800-3804` calls the unescaped case an injection vector, so this is the fix made in one place and missed in the authorization check. (M2) confirmed — `typeof undefined !== 'object'` passes the guard and `String(undefined)` substitutes; §5.4's fail-closed rule now extends to the `{{User*}}` surface, and WS3 hardens `MarkupFilterText` itself (undefined = unresolved, `''`-escape every scalar). (M3) adopted as **INV-3** with the fix shape: resolve the principal once, thread the same reference to both computations. (M4) adopted — `@deprecated` + a no-new-callers repo guard; you're right that nine migrations is one-time but the trap was permanent. (M5) adopted into the WS1 builder rewrite. (M6) upgraded from aside to its own enumerated ticket. Both nits folded (ReplayOnly executes no SQL — confirmed at `databaseProviderBase.ts:1393`; §2.6 reworded so nobody wires up the dead `PlatformVariants` path).

**§9.8:** your (a)-now-(b)-roadmap is adopted, error-message condition made a hard tested requirement — the denial names the row filter as cause and states the split-the-key remedy, not just the failed scope. Your framing of the failure mode (aging buckets outgrow `RunView`, get moved to a saved Query, key stops working six months later) is exactly right and is now in the plan verbatim as the rationale.

**§9.4:** `ActingCompanyIDs` registered **list-capable from day one** — your one-way-door argument is correct and generalizes: the plan's "adding a token later is cheap" held for new tokens but not for shape changes, so each scalar token is now explicitly documented as scalar-forever. Empty/absent set → `(1=0)`, per-element GUID validation + quote + escape. `ActingCustomerID` noted as considered-and-rejected (orgs cover it).

**Tickets:** all four enumerated in §5.10 (your three + ReportResolver as its own). Filing them as issues is with @AN-BC — two describe injection paths in shipping code on a public repo, so that's an owner call on disclosure posture, not mine.

**CI note:** agreed and worth having on the record — the green check on `0d36ce581` validates the docs-only diff, not the staged migration. The staged migration gets its CI validation when Phase 2 activates it with CodeGen output.

---

MarceloT-BC:
> ⚠️ **Raw generated Claude output — unedited.** Same basis as my review above: source read, nothing built or run.

Diffed `41cddc68d` rather than taking the summary on faith — the adoption is faithful, every finding landed with the right fix shape and a `[FO]` test. Two small follow-ups, both consequences of the v2.3 changes themselves:

**1. `{{ActingCompanyIDs}}` needs a canonical element order — INV-2 as written doesn't reach it.** INV-2 requires deterministic composition, but it's scoped to "stable ordering of the OR terms and the AND layers." A list token renders `IN ('a','b','c')` with the order it gets from the array, so an unsorted set produces different clause bytes across requests for the same principal — different fingerprint, redundant cache slots. INV-3 covers the dangerous case (fingerprint and WHERE disagreeing), so this is waste rather than exposure, but it's a one-word fix: **sort the elements before rendering**, and fold it into INV-2's sentence so the list case is explicitly in scope.

**2. The `MarkupFilterText` hardening is behavior-affecting for existing role-RLS deployments — worth a §3.5-style blast-radius note plus a changeset callout.** Narrower than it first looks: for the common `Col = '{{UserX}}'` shape nothing changes, since an unresolved token already yields no rows either way. The shift is confined to negation-shaped filters (`<>`, `NOT IN`, `LIKE`), which move from permissive to restrictive. That is the correct direction and the whole point of the fix — but a deployment that had unknowingly been relying on the old behavior will experience it as users losing access after an upgrade, not as a security fix. WS1 already gets that treatment in §3.5; this deserves the same paragraph so it isn't a surprise at rollout.

Nothing else from my side — the plan reads implementation-ready.

AN-BC:
@MarceloT-BC Both landed in `beda65d3a`. (1) INV-2's sentence now explicitly covers list-token element order, and the `{{ActingCompanyIDs}}` spec in §5.2 sorts elements into canonical order before rendering — your waste-not-exposure scoping is right and is how it's recorded. (2) The `MarkupFilterText` hardening carries the §3.5-style blast-radius paragraph, including your key observation verbatim in spirit: equality shapes are unchanged, negation shapes (`<>`/`NOT IN`/`NOT LIKE`) move from permissive to restrictive, and the release changeset must tell operators what to audit so it doesn't surface as unexplained access loss. Thanks for diffing the adoption rather than trusting the summary — that's the right posture for generated output on both ends.

---

jordanfanapour:
@AN-BC @MarceloT-BC @cadam11 — decisions on the remaining §9 open questions, plus two doc-hygiene notes.

**§9.3 — Application ceiling filters: deferred.** Ship keys-only in v1. The column stays (`APIApplicationScope.RowFilterID` is already in the migration and costs nothing to leave unused), but WS3 doesn't need to build ceiling-filter enforcement to land. Non-breaking to add later per the plan's own framing.

**§9.5 — WS1 blast radius: treat as a pure bugfix.** I'm not aware of any deployment that legitimately reassigns a filter-referenced column as a non-exempt principal under an Update RLS filter — the pattern §3.5 describes (an intentional ownership-transfer flow bumping into the new post-image check) doesn't match anything I know of running today. Proceed with WS1 as scoped. Please still document the theoretical behavior change explicitly in the changeset (per §3.5's own instruction) — "not aware of one" isn't "confirmed none exist," and any affected deployment should have something to point at when a previously-succeeding update starts failing.

**§9.7 — exemption looseness (`UserExemptFromRowLevelSecurity`, `entityInfo.ts:2231-2258`): defer, own ticket, not this feature.** Confirmed it's real: the method returns exempt off the mere absence of a role's `*RLSFilterID`, without checking that role's `Can{Read,Create,Update,Delete}` flag or its `Type` (`Allow`/`Deny`). Concretely — a user holding one role that grants filtered access (e.g. "Sales Rep," `ReadRLSFilterID` set) plus any second role whose permission row for the same entity has no filter (including a role that doesn't grant the permission at all, e.g. `CanRead = false`) becomes fully exempt, because the check never looks at whether that second row grants anything. That's a real privilege-escalation-via-role-stacking path for plain role-based RLS.

Leaving it alone doesn't weaken this feature — WS3 composes the key/ceiling filters with AND at `GetEffectiveRowFilterWhereClause`, structurally outside this method, so a user wrongly exempted from role RLS by this bug still gets bound by their key's filter. It only affects pure role-based RLS with multi-role users, unrelated to API keys.

Not fixing it here on purpose: correcting it changes who's exempt in every existing role-RLS deployment, and there's a real ambiguity to resolve first — is "a role with no filter and no grant" actually meant to confer blanket access in some deployment's convention, or is that always a misconfiguration? That needs its own audit, independent of this feature, before anyone tightens it. Filing as its own ticket rather than folding into WS1/WS2/WS3.

**Two doc-hygiene items — §9's list is stale relative to the body it's summarizing:**

- §9.4 (vocabulary size) is answered in §5.2 — `ActingCompanyIDs` was added list-capable, `ActingCustomerID` considered and rejected. Reads as still-open in §9; should be struck through like items 1/2/6.
- §9.8 (wholesale-deny ack) is answered in §5.10 — the (a)-now-(b) roadmap and the named-cause-and-remedy error message requirement are both in. Same — struck through in body logic, not in the §9 list.

**One more, not in §9 at all:** §5.10's "tickets to file" list still carries the `ReportResolver.CreateReportFromConversationDetailID` SQL injection as something to file separately. That's stale now — #3428 (https://github.com/MemberJunction/MJ/pull/3428) already fixes it, along with the `CheckRecordRLS` PK-escaping and `MarkupFilterText` issues from the earlier review round. Once #3428 merges, that list entry should point at the merged fix instead of reading as an open, unticketed vuln.

@AN-BC — over to you for the doc updates (§9 strikes for 3/5/7 with the reasoning above, §9.4/§9.8 struck, §5.10's ticket-1 pointed at #3428 once it's in). Nothing here blocks WS1 starting.

AN-BC:
@jordanfanapour All doc updates in — `b25fb2b55` (plan v2.4): §9.3/§9.5/§9.7 resolved with your reasoning recorded (including the sharpened §9.7 detail that the stacking role need not even grant the permission — that's now in the resolution text so the future ticket inherits it), §9.4/§9.8 struck as answered in body, §5.5's composition formula marks the ceiling term v2, and §5.10 ticket 1 points at #3428.

One scope boundary worth making explicit so it doesn't get lost between #3428 and WS1: **#3428 covers M1 (`CheckRecordRLS` PK escaping) and the `MarkupFilterText` hardening, but not M5** — `BuildCreateRLSProjections` still emits `String(val)` raw for non-quoted numeric/bit fields. That plus the typed-CAST builder rewrite remain WS1's, and the plan now says so in both §3.2 and the ticket entry, so rebasing WS1 on #3428 won't silently drop it.

Agreed nothing blocks WS1. Per the coordination note above, implementation is yours; this session keeps handling plan/staged-migration/doc changes on this PR.

---

AN-BC:
@AN-BC @cadam11 @MarceloT-BC — **ready for final review.** The PR is now the complete end-to-end implementation for 6.1, per Amith's direction: WS1 + WS2 + WS3 (keys-only), the v6.1 migration with CodeGen artifacts folded in, PG counterpart through the converter pipeline, and both test tiers.

**What to review against:** `plans/api-key-row-filters.md` v3.0 is the design record — every decision from all three review rounds is implemented as specified, including Craig's §5.10 scope-coverage denials, Marcelo's M1–M6 findings and list-capable `ActingCompanyIDs`, and Jordan's §9.3/§9.5/§9.7 resolutions. One deliberate deviation is documented in the changelog: WS1/WS2 ship in this PR rather than separately, superseding the earlier split decision per the owner's single-PR direction.

**Test evidence:** 4,468 unit tests passed / 61 skipped across the eight touched packages (1 pre-existing environmental failure in the sandbox, `mj_generatedentities` dist resolution, unrelated). Six new deterministic integration checks (KF1–KF6) cover the invariants that matter most: exemption independence (a role-exempt principal is still bound by its key filter), cache-slot separation per principal, fingerprint↔WHERE byte-identity, post-image update rejection, and Load-by-PK enforcement.

**Highest-leverage review targets:**
1. `EntityInfo.GetEffectiveRowFilterWhereClause` + the nine migrated call sites (`packages/MJCore/src/generic/entityInfo.ts`) — the composition that makes a key filter bind exempt principals.
2. `ScopeEvaluator.evaluateRules` — the one behavior change to existing scope evaluation (collect-all allows; deny semantics unchanged).
3. The save-time validation subclasses in `MJCoreEntitiesServer` — the authoring-time guarantees everything else assumes.
4. `MarkupFilterText` hardening — deliberately tightens role RLS too (negation shapes over unresolved tokens now match nothing); flagged in the changeset for operators.

CI on the final head (`9e1f50ea62`) is the last gate in flight — this session remains subscribed and will drive any failure to green. The temporary codegen workflow used to produce the CodeGen artifacts (no SQL Server route from the dev sandbox) has been deleted per its own delete-before-merge contract.

---

AN-BC:
Status update on the latest two commits:

**@jordanfanapour's `7668b21` — verified ✅**
I pulled and reviewed all four gap fixes: the create-path post-hook RLS re-check, `ReportResolver` switching to `GetUserFromPayload`, check 6 running on filter save, and the `SQLExpressionValidator` composition in the row-filter validators. All four are legitimate closures of real gaps against the plan (§5.3/§5.6), and the full local unit suite is green after a rebuild (866 passed / 5 skipped in GenericDatabaseProvider alone; the two initial failures I saw were stale local `dist`, not the changes).

**`b4038be` — fixes the red "adopted standards" check**
The failure was not introduced by this branch's work: it arrived via the merge of latest `next`. Root cause is a deadlock between two CI gates on `next`:

1. Knip's dependency-check gate **blocks PRs on undeclared imports**, so `fcc9eed` (on `next`) declared `@memberjunction/ng-shared` in `ng-file-storage`'s `package.json` — the import itself is the reviewed `mj-ui-layers-allow` exception tracked in MJ#3404.
2. The ui-layers gate flags forbidden **manifest declarations** with no marker mechanism (markers only work in source files), so the declaration Knip forces is a violation ui-layers can't excuse. Every PR touching `packages/**` fails.

The fix teaches the ui-layers check that a declaration backing a marker-excused import inherits the exception. An unexcused import of the same module still flags on its own line, and a declaration with no excused import behind it still flags — three new tests cover exactly those cases (28/28 in the suite, repo-wide `mj standards check` green locally). This lands on `next` with the merge and unblocks other PRs too.

Watching CI on the new head; will drive anything red to green.

---

MarceloT-BC:
Re-reviewed now that this is the full WS1+WS2+WS3 implementation (not the earlier design+schema PR). Spot-checked the load-bearing pieces against the plan and they're built as specified: `GetEffectiveRowFilterWhereClause` composes AND-across-layers with the role exemption applying only to the role term, all nine call sites are migrated (with the `noNewRlsCallers` repo guard backing it), M5's numeric projections now go through the typed-`CAST(NULL)` builder that throws on a non-numeric value, the WS1 post-image check sits at step 3b after `OnBeforeSaveExecute`, the `full_access` and startup-refusal backstops are in with the cause+remedy error text, and the migration is live under `migrations/v6/` with CodeGen output folded in behind the separator. CI is green including the deterministic Integration tier that was correctly red before. The M1–M6 escaping fixes shipped and merged in #3428. My approval stands.

Two release-hygiene items before merge — both for whoever owns the changeset/doc side, not code objections:

1. **The feature changeset is stale.** `.changeset/api-key-scope-row-filter-columns.md` still describes the schema-only state and says *"Enforcement is not in this release."* It is now — this PR ships full enforcement. As written the release notes would state the opposite of what merged. Worth rewriting to describe the actual WS1+WS2+WS3 behavior.

2. **No changeset covers the WS1/WS2 behavior change** that §3.5 (and Jordan's own note above) asked for. The post-image update tightening — a previously-succeeding update that reassigns a filtered column now fails for a non-exempt principal — and the WS2 tenant hardening touch `@memberjunction/core` / `generic-database-provider` / `server`, but no changeset here carries the operator-facing "what to audit if a previously-working update starts failing" note. The `harden-sql` changeset covers the injection fixes only and is already on `next`.

Also flagging that the PR currently shows `CONFLICTING` against `next` — `GenericDatabaseProvider.ts` plus three generated files (`entity_subclasses.ts`, `generated.ts`, `ui-layers.ts`), most likely from #3428 landing on `next` and touching the same files. Needs a rebase to merge; not a review objection.

_Reviewed by reading the branch against `next` — nothing built or run on my end; the green CI is the execution evidence._

PR #3410, opened 2026-08-03
security: harden SQL filter validation, JWT verification, and system API key comparison

AN-BC:
## Summary

Hardening changes from an adversarial security review of the data-access, validation, and auth layers. Every change is **additive/defensive** and preserves existing behavior for legitimate inputs. All affected packages build and `@memberjunction/global`'s test suite passes (107 tests, including updated + new SQL-validator tests).

### 🔴 Critical — SQL injection via `RunView` `ExcludeUserViewRunID`
`packages/GenericDatabaseProvider/src/GenericDatabaseProvider.ts`

The GraphQL `ExcludeUserViewRunID` string input was interpolated **raw** into the view `WHERE` clause (`UserViewRunID=${excludeUserViewRunID}`) with **no validation** — while every sibling clause (`ExtraFilter`, `UserSearchString`, `OverrideExcludeFilter`) is passed through `ValidateUserProvidedSQLClause`. Any authenticated caller could inject arbitrary SQL into the view query, bypassing entity permissions and row-level security. The value is only ever a `UserViewRun.ID` (GUID), so it is now rejected unless it is a well-formed GUID.

### 🟠 Hardening — block `WAITFOR` in user filter/order-by clauses
`packages/MJCore/src/generic/databaseProviderBase.ts`

Adds `WAITFOR` to the `ValidateUserProvidedSQLClause` blocklist, closing the time-based blind-injection vector. No legitimate filter/order-by uses `WAITFOR`; the intended subquery capability of `ExtraFilter` is unaffected.

### 🟠 Hardening — block database system-catalog access in the SQL expression validator
`packages/MJGlobal/src/SQLExpressionValidator.ts`

`SQLExpressionValidator` now rejects references to system catalogs / metadata objects — `sys.*`, `INFORMATION_SCHEMA`, `syslogins`, `pg_catalog.*`, `pg_authid`/`pg_shadow`/`pg_user`/`pg_roles` — in **all** contexts, including `full_query`. These live outside MemberJunction's entity-permission model, so permitting them turned a validated `SELECT` into a credential-exfiltration / schema-enumeration primitive. The prior security test that asserted `SELECT ... FROM sys.sql_logins` was *valid* is corrected to assert it is now blocked; a regression test confirms a legitimate `UNION` over application views still passes.

### 🟡 Low — constant-time system API key comparison
`packages/MJServer/src/context.ts`

The superadmin `MJ_API_KEY` was compared with `===`, which short-circuits on the first differing byte (a timing side channel). Now compared via SHA-256 digest + `crypto.timingSafeEqual`.

### 🔵 Defense-in-depth — pin JWT verification algorithms
`packages/MJServer/src/context.ts`, `packages/AI/MCPServer/src/auth/TokenValidator.ts`

Both JWKS verification paths now explicitly pin the accepted signature algorithms to the asymmetric family, preventing `alg=none` / RS256→HS256 confusion should a key format change or library regression occur. (Not currently exploitable with the pinned `jsonwebtoken` version, but eliminates the future footgun.)

## Related architectural findings — NOT changed here (need a design decision)

These are real but behavior-sensitive, so they are documented rather than patched blindly:

1. **`ExecuteAdhocQuery` resolver is reachable by any authenticated user** (`packages/MJServer/src/resolvers/AdhocQueryResolver.ts`). It runs arbitrary read-only `SELECT`s, bypassing per-entity `CanRead` / RLS / multi-tenant isolation. A blunt `@RequireSystemUser` gate would break the data-artifact live-data viewer (`packages/Angular/Generic/artifacts/.../data-artifact-viewer.component.ts`), which legitimately runs artifact SQL for normal users. The system-catalog block above removes the worst exfiltration primitive, but the broader "any authenticated user can read any business entity via ad-hoc SQL" gap needs a policy decision (privilege gate vs. server-side entity-permission enforcement).
2. **`ExtraFilter` permits subqueries** (`ID IN (SELECT ...)`) by design, so the blocklist can't stop error-based/boolean-blind exfiltration via subqueries without breaking that feature. Consider routing these through `SQLExpressionValidator` (`where_clause` context) with an explicit `allowSubqueries` opt-in.
3. **RunQuery Nunjucks templating** (`packages/QueryProcessor/src/queryParameterProcessor.ts`) applies the `sqlsafe` guard only when the query author opts in; consider defaulting string/array params to escaped.

## Testing
- `@memberjunction/global`: `npm run build` ✅, `vitest run SQLExpressionValidator` ✅ (107 tests pass).
- `@memberjunction/core`, `@memberjunction/generic-database-provider`, `@memberjunction/server`, `@memberjunction/ai-mcp-server`: `turbo build` ✅ (all compile).

---
_Generated by [Claude Code](https://claude.ai/code/session_01UGcB3iATFgk9C4KwmpSJcJ)_


cadam11:
**Security review — exploitability triage for the LTS backport decision**

I reviewed each of the three hardening changes against the released `v5.51.0` / `lts/5` line to decide backport-worthiness. Bottom line: one change closes a genuinely exploitable hole and is backport-critical; the other two are sound defense-in-depth.

**1. SQL filter validation — EXPLOITABLE (the backport-critical piece).**
`ExcludeUserViewRunID` is interpolated raw and unparameterized into the view WHERE clause at `GenericDatabaseProvider.ts:1578`, with none of the `ValidateUserProvidedSQLClause` guarding that every sibling clause (`ExtraFilter`, `UserSearchString`, `OrderBy`) receives. It's reachable by any authenticated user through the RunView resolvers (`RunViewByName`/`ByID`, `RunDynamicView`, `RunViews` — none carry `@RequireSystemUser`), needing only read permission on a single entity. On a default read-write-pool deployment, stacked statements execute, so this is an authenticated SQL injection that bypasses entity permissions and RLS. (Verified by constructing the injected T-SQL string; not executed against a live database — worth a quick live confirm before it ships to the certified line.) The system-catalog denylist is a reasonable companion fix. The `WAITFOR` blocklist addition is defense-in-depth only — `;` is already blocked in that clause position, so `WAITFOR` can't reach statement position there.

**2. JWT verification — defense-in-depth.**
Both patched sites call `verify` without an explicit `algorithms` list, but `jsonwebtoken@9.0.3` (exact-pinned) already rejects `alg=none` and infers an asymmetric-only algorithm set from the JWKS public key, so RS→HS confusion and none-alg both fail closed today. Good future-proofing, not a live hole. Two things specific to this change, though:
- The list `['RS256','RS384','RS512','ES256','ES384','ES512','PS256']` omits **`PS384`/`PS512`**, which jsonwebtoken's own key-based inference *does* include — so this is subtly *not* purely additive: any IdP signing with those two algorithms would now be rejected. No major IdP defaults to them, but on an LTS line that regression risk isn't zero.
- Only one of the two verify sites in `TokenValidator` is pinned — `verifyTokenSignatureWithKeys` (the Azure AD v1 path) is left unpinned, so the "eliminate the future footgun" goal isn't fully met here.

**3. System API key comparison — defense-in-depth.**
`context.ts:423` is a plain `===`; no logic flaw (empty/undefined keys are already guarded). It's a real timing side channel, but single-digit nanoseconds against network jitter — the 🟡 Low self-rating is right.

**Backport recommendation.** The SQL-injection fix is a showstopper and should reach `lts/5`; the JWT pinning and constant-time compare are appropriately 6.x-only. One mechanism note: the repo's `backport lts/5` label cherry-picks the **entire** merged PR — so if we want only the SQL fix on the certified line (and to avoid carrying the PS384/PS512 regression onto `lts/5`), splitting the SQL-injection fix into its own PR, or a manual cherry-pick of just that commit after merge, is cleaner than labeling this PR whole.

**Out of scope, noticed in passing:** `ResolverBase.ts:365` builds an `ExtraFilter` by interpolating a user-supplied view name into a filter clause. It does reach `ValidateUserProvidedSQLClause` downstream, but that validator strips string literals *before* checking — worth a separate look to confirm it isn't bypassable.

_(Security triage assisted by Claude Code.)_

AN-BC:
@cadam11 thanks. @jordanfanapour I think I see a PR from you that might be similar to this so flagging this is merging now as FYI.

@cadam11 by merging this, it goes to 6.x but that label also auto backports it to 5.x LTS?

PR #3412, opened 2026-08-03
[6.0] External agent harnesses as MJ agents — turn protocol on the Loop contract (Claude Code, Codex, OpenCode, Gemini CLI, Pi)

AN-BC:
Runs an **external agent harness** — Claude Code, Codex CLI, OpenCode, Gemini CLI, Pi — as the reasoning substrate for an MJ agent, while MemberJunction keeps identity, permissions, governed data access, payload contracts, HITL, cost control and run-level audit.

> This PR began as the design plan (`plans/external-agent-harness.md`) and now carries the implementation. **Verified end to end against a real Claude Code install**: `success=true`, 2 iterations, `$0.020042`.

---

## The load-bearing idea

**A harness turn is protocol-identical to a Loop agent's prompt iteration.**

The harness reasons freely inside its sandbox, then ends its turn by emitting the same next-step JSON envelope a Loop model emits. MJ executes any actions, sub-agents or skills through its **own** validated machinery, then resumes the session with the results.

`HarnessAgentBase` overrides **exactly one method** — `executePrompt`. Everything else is untouched `BaseAgent`.

That is why next-step validation, per-action `MaxExecutionsPerRun`, skill gates, plan-mode blocking, `PayloadManager` path ACLs, `checkExecutionGuardrails` and run-step recording all apply **with no new enforcement code** — and why there is exactly one authority channel to audit rather than two. (It's also why the MCP loopback is read-only: routing actions through MCP would create a second authorization story and double-count accounting.)

```
   ┌─ BaseAgent loop (unchanged) ──────────────────────────────────┐
   │   executePrompt() ──► [ HarnessAgentBase override ]           │
   │        │                 ├─ adapter.RunTurn(input)            │
   │        │                 ├─ accumulate usage                  │
   │        │                 └─ write AIPromptRun                 │
   │        ▼                                                      │
   │   DetermineNextStep   (inherited from LoopAgentType)          │
   │        ▼                                                      │
   │   validate ─► execute actions / sub-agents / skills           │
   │        ▼                                                      │
   │   checkExecutionGuardrails ─► next turn                       │
   └───────────────────────────────────────────────────────────────┘
```

---

## ⚠️ Also fixes a live guardrail bug affecting **every** agent type

Independent of harness work, and worth reviewing on its own: **`MaxCostPerRun` and `MaxTokensPerRun` did not interrupt a run.**

The limits are static on the agent and were compared correctly. The problem was the other side: `AIAgentRun.TotalCost` / `TotalTokensUsed` are *derived* from step rollups, and their only writer ran on terminal paths only. Mid-run they sat at `0`, and the checks are guarded on `agent.MaxCostPerRun && agentRun.TotalCost` — so a falsy zero short-circuited them entirely.

Net effect: the ceilings were evaluated as a run *ended*. Reporting, not guardrails. A runaway agent burned its whole budget and was informed afterwards. Only iteration and time limits actually bit.

Fixed by refreshing the accumulated actuals before the comparison. **Regression coverage verified to fail without the fix** — with the refresh disabled, 3 of 5 cases fail; the 2 negative cases pass either way.

---

## Schema (additive)

| Object | Purpose |
|---|---|
| `MJ: AI Agent Harnesses` | Registry of harnesses this install can launch |
| `MJ: AI Agent Credentials` | Grant edge — secrets an agent carries into its sandbox |
| `AIAgentRun.ExternalSessionID` | Correlate an MJ run with the vendor's session logs |

`CapabilitySettings` is a strongly-typed JSONType (`IHarnessCapabilitySettings`). It declares what each adapter **actually implements**, because the runtime *emulates what is missing* — so an over-claim is a **silent behavioural gap**, not an error. Gemini CLI honestly reports `SessionResume: false` and pays replay tokens.

Every harness turn writes a real `AIPromptRun`. Run totals derive from step rollups, so a turn that records none leaves the run at zero cost forever — and the ceiling above with nothing to compare. `PromptID`/`ModelID`/`VendorID` are all NOT NULL and each resolves to a **real** catalog row, never a placeholder.

---

## Adapters

| Harness | DriverClass | Mechanism | Verified |
|---|---|---|---|
| Claude Code | `ClaudeCodeCliAdapter` | CLI, `stream-json` | ✅ end-to-end run |
| Pi | `PiAdapter` | `pi -p --mode json --session-id` | CLI surface + event shapes (install unauthenticated) |
| Codex | `CodexAdapter` | `codex exec --json` | unit |
| OpenCode | `OpenCodeAdapter` | `opencode run` JSON | unit |
| Gemini CLI | `GeminiCliAdapter` | `gemini --output-format json` | unit |
| *anything* | `StdioJsonAdapter` | documented JSON contract | unit |

**No SDK adapter, deliberately.** One was built on the Claude Agent SDK and removed: it runs in-process in Node, so no sandbox executor can place it in a container. It bought better permission hooks at the cost of any process isolation — and an SDK-backed agent configured `provider: 'docker'` would have run outside its sandbox while the config claimed otherwise. **The CLI's missing permission hook is fixable; the SDK's missing containment is not.** (It also ships a ~259 MB bundled `claude` binary, wrapping the same program the CLI adapter invokes.)

## Sandboxes — the provider owns process placement

Adapters never call `spawn`; everything goes through `SandboxExecutor`, so the same adapter runs on a laptop or in a per-run container without knowing the difference. `HarnessProcess` is stream-based rather than `ChildProcess`-based, because a k8s exec is streams over a websocket and a remote runner is HTTP — neither could honestly implement a `ChildProcess` contract. It also makes adapters unit-testable with a fake executor: no binary, no container, no network.

> The **local** provider scopes a workspace *directory*, not the *process*. `networkPolicy` is advisory there — documented rather than implied, because believing `'none'` is enforced locally is worse than knowing the boundary is soft. `DockerSandboxProvider` enforces `'none'` for real.

---

## Testing

- **Unit**: `ai-agent-harness` 15/15 · `ai-agents` 1771/1771 (5 new guardrail regression cases)
- **Full build**: 298/298
- **Live**: `integration-test-scripts/run-harness-agent.ts` — real Claude Code run, `success=true`, 2 iterations, `$0.020042`, workspace correctly scoped to `agent-user/<agentID>/<userID>`. Turn 1 drifted to prose, `BaseAgent`'s retry corrected it, turn 2 returned valid JSON — the designed recovery path working.
- **Integration bundle**: `agent-external-harness` (AEH1–AEH7) targets the wiring failures unit tests can't reach — most importantly AEH2, since a missing `BaseAgent`-side registration makes `AgentRunner` *silently* fall back to plain `BaseAgent` (no harness, no sandbox, and the run still succeeds).

**Reviewer note:** the AEH bundle compiles and is exported, but I could not get it to execute — `mj test run "agent-external-harness"` reports *Test not found* and `mj test list` hangs. **Treat those 7 checks as unrun.** Someone who knows the named-bundle registration path should confirm it.

## Docs

`guides/AGENT_HARNESS_GUIDE.md` (indexed in `guides/README.md`), the package README, and a `HarnessAgentType` section in the AI Agents README contrasting it with Realtime — Realtime bypasses the loop, Harness stays inside it.

---

## Ships not live

Every harness row is `Inactive`; `Demo Harness Agent` is `Pending`. They depend on external binaries a fresh install won't have, and an enabled row would advertise a harness that can't launch.

## Known gaps (documented, not hidden)

- **`PermissionHooks: false` on every adapter.** The `strict` posture needs an MCP permission-prompt tool — a later phase, wiring back to MJAPI's own MCP server. **Consequence today: a harness agent can reason but cannot do mutating in-sandbox work.** Proven by the live run, which completed the protocol but could not write its file.
- `mcp-only` / `allowlist` are not packet-enforced under Docker; only `'none'` is.
- MCP loopback not yet wired — harness agents can *act* (turn-end actions/sub-agents) but cannot *read* MJ data mid-turn.
- `ModelID` uses the declared model, not the harness-reported one.


AN-BC:
@EL-BC and @MarceloT-BC  this can be a very small side project for us to discuss at the Hackathon. It probably will be near one-shot for implementation but the implications are important to think through.

AN-BC:
Tagging @jordanfanapour @MS-BC @CaelebB-BC  @rkihm-BC @rkihm-BC @cadam11 as I think this will be of interest to all of them too.

PR #3419, opened 2026-08-04
Entity metadata: layered base views + direct-SQL declarations, plus the external agent harness design plan

AN-BC:
Two additive `Entity` metadata capabilities plus a design plan. Everything here is additive — every pre-existing row reproduces today's behaviour exactly, and no install changes unless it opts in.

---

## 1. Layered base views — `Entity.GeneratedBaseViewName`

**The problem.** `BaseViewGenerated = 0` was all-or-nothing. To add one computed column, an application inherited the *entire* generated view — every related-entity display join, the geo join, the recursive root-ID `OUTER APPLY`, the soft-delete predicate — and had to hand-maintain it from then on.

That inheritance isn't a one-time cost, it's a standing obligation. Add a foreign key later and its display field simply never appears, because nothing regenerates the join. The failure is **silent** — the column is absent rather than wrong, so nothing errors and no test notices. It also freezes the entity at whatever CodeGen produced the day the view was copied; geo columns and root-ID columns both arrived after custom views existed in the wild.

**The fix.** When `GeneratedBaseViewName` is set, CodeGen writes its full generated view under *that* name and the application owns `BaseView`, wrapping it:

```sql
CREATE VIEW [orders].[vwOrderHeaders] AS
SELECT g.*, CASE WHEN ... END AS IsOverdue
FROM   [orders].[vwOrderHeadersGenerated] g;
```

Everything underneath keeps regenerating, so a new foreign key appears on its own, and the custom layer stays a few reviewable lines. Columns it adds become first-class virtual `EntityField` rows and are returned by `spCreate`/`spUpdate`/`spDelete`, which read `BaseView`.

| `BaseViewGenerated` | `GeneratedBaseViewName` | Result |
|---|---|---|
| `1` | `NULL` | CodeGen writes `BaseView`. The default. |
| `0` | `NULL` | The app owns `BaseView` entirely; CodeGen writes nothing. |
| `0` | `vwFooGenerated` | **Layered** — CodeGen writes the inner view, the app wraps it. |
| `1` | `vwFooGenerated` | Refused by CHECK — contradictory (see below). |

### The two paths, side by side

`BaseView` is always the public surface. The only question is who writes it — and layering splits one view into two so the mechanical half can keep regenerating.

```mermaid
flowchart TB
    subgraph GEN["① GENERATED — the default"]
        direction TB
        GT[("Foo<br/>base table")]
        GV["<b>vwFoo</b><br/>owned by CodeGen<br/><i>regenerated every run</i>"]
        GT --> GV
    end

    subgraph CUS["② FULLY CUSTOM — BaseViewGenerated = 0"]
        direction TB
        CT[("Foo<br/>base table")]
        CV["<b>vwFoo</b><br/>owned by the application<br/><i>frozen the day it was copied</i>"]
        CT --> CV
    end

    subgraph LAY["③ LAYERED — GeneratedBaseViewName = vwFooGenerated"]
        direction TB
        LT[("Foo<br/>base table")]
        LI["<b>vwFooGenerated</b><br/>owned by CodeGen<br/><i>regenerated every run</i>"]
        LO["<b>vwFoo</b><br/>owned by the application<br/><i>SELECT g.* + your columns</i>"]
        LT --> LI
        LI -->|"SELECT g.*"| LO
    end

    GV --> SURF
    CV --> SURF
    LO --> SURF

    SURF["<b>BaseView</b> — the public surface<br/>field discovery · permissions · RunView<br/>spCreate / spUpdate / spDelete"]

    classDef codegen fill:#1f6feb22,stroke:#1f6feb,stroke-width:2px
    classDef app fill:#d2992222,stroke:#d29922,stroke-width:2px
    classDef table fill:#8b949e22,stroke:#8b949e,stroke-width:1px
    classDef surface fill:#23863622,stroke:#238636,stroke-width:2px
    class GV,LI codegen
    class CV,LO app
    class GT,CT,LT table
    class SURF surface
```

**Blue regenerates, amber is hand-written.** In ① the whole view is blue and you cannot add a column to it. In ② the whole view is amber — you can add anything, but every display join, geo column and root-ID column becomes yours to maintain, and a foreign key added later never appears. ③ puts the boundary in the middle: the ~80 mechanical lines stay blue and keep up with the schema, while your computed columns stay amber and stay reviewable.

The green node is why none of this is visible downstream — field discovery, permissions, `RunView` and the CRUD routines target `BaseView` in every case, so a column added by the custom layer becomes a first-class virtual `EntityField` and comes back from a save.

*(This diagram also lives in [`packages/CodeGenLib/README.md`](packages/CodeGenLib/README.md#base-views-generated-custom-or-layered) so it outlives the PR.)*

**Mechanics worth reviewing:**

- **`EntityInfo.GeneratedViewName`** — the single resolution of "which view does CodeGen write". Several call sites decide where to write the view, what to name the emitted file, and which object to refresh; any two disagreeing produce a view under a name nothing reads. Derived *from* `HasLayeredBaseView` rather than re-testing the raw column, so the two cannot drift on a name differing only by case.
- **`EntityInfo.HasLayeredBaseView`** — the layering test, comparing case-insensitively because SQL Server object names are.
- **Refresh inner before outer.** The custom layer does `SELECT g.*` and a view caches its column list; refreshing the outer against a stale inner re-caches the *old* columns and the new one stays missing — indistinguishable from never having been added.
- **Existence guards on the application-owned view — in all three emission paths.** CodeGen refreshes and grants on `BaseView` but never creates it, and on the first pass after layering is enabled it cannot exist: it selects from the inner view that very pass creates. New `generateIfViewExistsSQL` provider method wraps those statements so the bootstrap pass succeeds instead of failing on the one run that's supposed to enable the feature. This includes `buildCustomBaseViewRefreshSQL` (STEP 4.5), which writes into the **migration log** — the SQL that runs on staging and prod. A layered entity qualifies there alongside genuinely hand-written custom views, but the two aren't interchangeable: for a fully custom view a missing object is a real misconfiguration and stays a loud failure, while a layered entity's outer view is legitimately absent during bootstrap.
- **Self-heal.** A layered entity's inner view goes through the failed-refresh regeneration path, since that half *is* CodeGen-owned.
- **CHECK constraints close both bad states.** A self-referencing name is refused (a view cannot select from itself), including the `BaseView IS NOT NULL` arm — `X <> NULL` is UNKNOWN and a CHECK *passes* on UNKNOWN, which would otherwise admit a row naming an inner view while leaving the public surface NULL. And layering now requires `BaseViewGenerated = 0`: with it left at `1` the two halves of CodeGen disagree, because view *generation* gates on `BaseViewGenerated || HasLayeredBaseView` and writes the inner view, while the outer view's *refresh* and its *GRANTs* gate on `!BaseViewGenerated` and are skipped — yielding a public surface that is never granted and never refreshed, with CodeGen reporting success. `HasLayeredBaseView` still deliberately ignores `BaseViewGenerated`, since honouring it would make CodeGen overwrite the application's hand-written view; refusing the state at the source means that fail-safe is never relied upon.

### Dogfooded on two core entities

MJ ships **21 core entities with fully custom base views** — `BaseViewGenerated = 0`, no inner view, so CodeGen stops generating entirely. Between them **25 foreign keys have no display field at all**, because nobody went back to add the join after the FK landed. This PR converts the first two:

| Entity | Before | After | Lost | Gained |
|---|---|---|---|---|
| `MJ: Version Installations` | 12 cols | 12 | none | — |
| `MJ: User View Run Details` | 7 cols | 8 | none | `UserViewRun` |

`MJ: Version Installations` has no foreign keys and one computed column — the simplest possible layering. `MJ: User View Run Details` is the representative case: `UserViewID` and `EntityID` are *two-hop ancestor keys*, which CodeGen can't produce, so the custom layer keeps its own joins.

That 8th column is the point — a related-entity display field the hand-written view never had, which **appeared on its own** and was registered as a virtual `EntityField` without anyone editing SQL. On regeneration CodeGen touched **only** the inner view; both hand-written wrappers were left alone.

**Doing it for real exposed two bugs, both fixed here:**

1. **Adopting layering emitted no migration SQL.** Setting `GeneratedBaseViewName` is a *metadata* change, so the entity never enters the modified/new list, and `logSQLForNewOrModifiedEntity` only writes for entities in that list. CodeGen created both inner views in the dev database and emitted nothing — other environments would never have received them. Captured via `forceRegeneration` with `entityWhereClause`; documenting that requirement is a follow-up.
2. **Two further emission sites aimed unguarded SQL at the outer view.** `generateBaseView` concatenates `viewPermSQL` — which GRANTs on `entity.BaseView`, a *different object* for a layered entity — onto the inner view's DDL, and a second unguarded copy went to the permissions file and migration log. On the documented bootstrap both run before the wrapper exists. Both now go through `guardOnApplicationOwnedView`; non-layered output stays byte-identical.

`MJ: Entities` is deliberately **not** converted: `vwEntities` is what the metadata layer itself reads, so layering the view CodeGen consults in order to decide whether to layer is a chicken-and-egg best left alone. The other 18 are a follow-on.

**SQL Server only.** CodeGen throws on a layered entity under PostgreSQL. PG expands `SELECT *` at view creation and freezes it, has no `sp_refreshview` equivalent, and never recreates the application-owned outer view. The resulting behaviour would be worse than plainly broken — it'd be *intermittent*: an added column leaves the outer view stale (`CREATE OR REPLACE` on the inner never touches dependents), while a rename or type change raises 42P16 and sends CodeGen down the `DROP CASCADE`/replay path that incidentally *does* recreate it. Same feature, opposite outcomes, decided by which kind of schema change landed that day — precisely the silent staleness this feature exists to eliminate, so it throws rather than documenting a footgun. Fully custom base views are unaffected on both dialects.

---

## 2. Direct-SQL declarations — `AllowDirectSQLInsert` / `Update` / `Delete`

MJ's contract is that every mutation flows through `BaseEntity.Save()` / `.Delete()`, because that is the only path where record-change tracking, cache invalidation, entity actions, validation and soft delete actually run. SQL written outside that path skips all of it, and none of those failures are loud — you get an audit trail that *looks* complete but isn't, and a server cache that serves stale rows indefinitely.

Three new `Entity` columns make the exception explicit instead of tribal knowledge:

- `AllowDirectSQLInsert` — bulk loads, ETL/integration sync, rows created as a side effect of a proc
- `AllowDirectSQLUpdate` — bulk backfills, maintenance routines
- `AllowDirectSQLDelete` — purge/retention, integration reconciliation

Split by verb because the risk differs: a bulk `INSERT` on a staging-shaped entity is routine; a direct `DELETE` on a soft-delete entity destroys rows the platform promised to keep. All default to `0`.

**These declare intent — they enforce nothing, and cannot.** No constraint or trigger stops anyone opening a query window. They record which entities are *sanctioned* so the code paths and tooling that choose to honour the contract can consult one authoritative answer. The docs say this loudly, because an `Allow*` flag that stops nothing is otherwise a trap.

**Two CHECK constraints, because both failure modes are silent:**

1. Any direct-SQL flag requires `TrackRecordChanges = 0` **and** `TrustServerCacheCompletely = 0`. Direct DML writes no audit row and fires no invalidation event, so leaving either on yields an audit trail that looks complete but isn't, and a cache that serves stale rows. `TrustServerCacheCompletely` already documented exactly this scenario — these columns are the declarative half of that same fact.
2. `AllowDirectSQLDelete` requires `DeleteType = 'Hard'`, since a direct `DELETE` removes the row rather than setting `DeletedAt`.

Both are multi-column and therefore table-level, so CodeGen cannot mistake them for column value lists — `DeleteType`'s Hard/Soft list is untouched.

Also adds a 🚨 section to `.claude/rules/data-access.md` so the rule loads on every `**/*.ts` task.

---

## 3. External agent harness integration — design plan (docs only)

`plans/external-agent-harness.md` — a design plan targeted at 6.0 for running external agent harnesses (Claude Code, Codex CLI, OpenCode, Pi, Cline) as the reasoning substrate for an MJ agent, while MJ keeps identity, permissions, governed data access, payload contracts, HITL, cost controls and run-level audit.

The load-bearing idea: **a harness turn is protocol-identical to a Loop agent's prompt iteration.** The harness ends each turn by emitting the Loop agent type's next-step JSON, and MJ executes actions/sub-agents/skills through its existing validated machinery, then resumes the session with results. MCP loopback is read-only and intra-turn, so there is exactly one authority channel. Covers the dual-registry `HarnessAgentType`/`HarnessAgentBase` split, the `BaseHarnessAdapter` contract, `ISandboxProvider`, per-run scoped MCP credentials, Strict/Auto/Dangerous postures, the additive metadata model, and a four-phase delivery plan.

**No code in this PR** — plan only, for review ahead of implementation.

---

## 4. Build fix: raise the `@types/node` override

Unrelated to the above, but it blocked the build. The committed lock file was stale — it predated the current `overrides` block and still carried nested `@types/node@24.10.11` entries, so MJServer had been compiling against 24.x despite the override saying `20.14.2`. A real re-resolve applied the overrides properly and MJServer stopped typechecking (`Buffer` vs `ArrayBufferView` in `context.ts`).

The override is vestigial: added 2025-02-07 (c4835e8206), a year before the 2026-02-06 mass dependency upgrade (69edd182a4) moved 106 packages to 24.10.11 without updating it. Nothing in the repo pins 20.x. Raised to `24.10.11` to match. Cold full build with `TURBO_FORCE`: 268/268, 0 cached, 0 failures.

@cadam11 — flagging in case the original override was protecting against something now forgotten.

---

## Migration & metadata

**Two migrations, and they must stay separate.**

`V202608031959__v6.1.x__Add_Entity_GeneratedBaseViewName.sql` — all four `Entity` columns plus the four CHECK constraints, consolidated per-table, with extended properties on every column. Its CodeGen section also carries the two pilot entities' **inner** views.

`V202608041916__v6.1.x__Layered_Base_Views_Pilot.sql` — the two application-owned **outer** views. These can't live in the first migration: they select from inner views created by its CodeGen section, and that section is replaced wholesale on every regeneration, so hand-written SQL inside or after it would be destroyed. Timestamp ordering is what guarantees the inner views exist first.

The layering itself is declared in `metadata/entities/.layered-base-views.json`, not SQL — entity metadata ships as declarative JSON and the build engineer's consolidated push carries it forward.

Its CodeGen section was regenerated from scratch against a clean database: `migrate` → `sync push` → `codegen`, **two passes to convergence** (pass 1 creates the base view carrying the `ParentChunk` display column; pass 2 discovers it as a virtual `EntityField`), then a third pass emitting nothing. Both passes are appended in order under a do-not-edit banner, because both are required.

Verified end-to-end on a database rebuilt from scratch: `migrate` → `sync push` → **`codegen` emits nothing**, so the migration is self-contained. Constraint enforcement proven **7/7** in a rolled-back transaction (soft-delete refused, both tracking/cache cases refused, the sanctioned combination accepted, self-referencing layered view name refused, `BaseViewGenerated = 1` layering refused, a valid layered configuration accepted).

The CodeGen section also carries a full regeneration of `MJ: Content Item Chunks` — its `ParentChunkID` root-ID function, base view, CRUD routines and `ParentChunk` display field. That is **not** incidental churn: those objects were missing from the baseline and CodeGen is supplying them. Deliberately retained, and the banner says so.

> **Note for reviewers:** CodeGen's validator function *names* are LLM-authored and not stable across runs — the same constraint has come back as `ValidateDirectSQLAndCacheSettings`, `ValidateDirectSQLAndCacheTracking`, and `ValidateDirectSQLAndTrackingConstraints`. A name change on a future regeneration is not a defect.

## Testing

- MJCore: 1620/1620
- CodeGenLib: 703 passed / 60 skipped
- Full repo build: 268/268 (cold, `TURBO_FORCE`)
- New suites: `LayeredBaseView.test.ts` (17), `entityInfo.layeredBaseView.test.ts` (12), `entityInfo.directSQL.test.ts` (4), plus additions to `PostgreSQLCodeGenProvider.test.ts` covering the layering refusal
- End-to-end, twice: fresh DB → `migrate` → `sync push` → **`codegen` emits nothing**, for both the schema change and the pilot conversion
- Superset column check on both converted views (no column disappears — the check that matters, since a loss is a breaking ORM change)
- Constraint enforcement 7/7 against the live DB in a rolled-back transaction

### IT69 — integration bundle (@rkihm-BC's review request)

Six checks proving layering against a **live database** rather than a hand-built `EntityInfo`: metadata coherence, both views present and distinct, the outer view genuinely wrapping the inner one, no sign of CodeGen having overwritten the wrapper, every inner column inherited, and — the promise itself — a CodeGen-generated display field reaching the public surface and being selectable through `RunView`.

**Validated by sabotage, not by assumption.** Rewriting the wrapper to drop `SELECT g.*` makes LBV6 fail with *"RunView failed selecting generated display column(s) UserViewRun through BaseView"*. Restored afterwards; green.

`LBV1` and `LBV6` are driven from metadata + `RunView`, deliberately **not** `sys.*`. The first version gated all six on `ctx.Pool` and reported **6/6 green on a run path that provides no pool** — six checks, zero assertions executed, indistinguishable from six real passes. The checks carrying the promise now run everywhere; only the catalog checks can skip, and they log `SKIPPED (no assertions ran)`.

> **Worth someone's attention (pre-existing, not fixed here):** `metadata-consistency`'s eight catalog checks skip-as-pass on that same run path (`MC1 skipped: no mssql pool`). That bundle has been reporting green without executing, so the deterministic tier asserts less than its results suggest.

### Also fixed: a stale class-registration manifest

Regenerating ServerBootstrap's manifest surfaced **three classes the committed copy was missing** (965 → 968): `MJAPIApplicationScopeEntityServer`, `MJAPIKeyScopeEntityServer`, `MJRowLevelSecurityFilterEntityServer`. The manifest exists to defeat tree-shaking of `@RegisterClass` types, so a missing entry means `ClassFactory` can resolve to the **base** class at runtime.

The RLS one is security-adjacent: that server subclass turns *off* `DefaultSkipAsyncValidation` specifically to fail closed on invalid `FilterText` at save time. Without it an invalid RLS filter saves clean and fails later as — in the class's own words — *"an unexplainable SQL error"* at query time. The regeneration removes no imports; it's a strict superset.

## Compatibility

Fully additive. `GeneratedBaseViewName = NULL` and all three `AllowDirectSQL*` flags at `0` — every existing row — reproduce previous behaviour exactly. Changesets included for `@memberjunction/core` and `@memberjunction/codegen-lib` (both minor).


rkihm-BC (commented):
## Code review — layered base views

**Verdict: strong, well-engineered, and safe to merge for the SQL Server target — with one documentation gap that misrepresents PostgreSQL behavior, and a missing end-to-end test that's worth having given the stakes.**

I checked out the branch, rebuilt `@memberjunction/core`, and ran the suites: **MJCore 1604/1604, CodeGenLib 648/648**, plus the two new files (10 + 6 tests) green. (The one gotcha: CodeGenLib's new tests fail against a stale `dist/` of core — they pass once core is rebuilt, since CodeGenLib imports the built package.)

### What's good

- **The additive claim holds under inspection, not just assertion.** `NULL` → `GeneratedViewName` falls back to `BaseView`/`vw{CodeName}` and `HasLayeredBaseView` is `false`. Every gate is `BaseViewGenerated || HasLayeredBaseView`, so a non-layered entity's path is byte-identical to before.
- **The "one resolution" claim is real.** Grepping for every independent `BaseView ? … : vw${…}` derivation: after this PR only the getter itself uses that pattern. All call sites (both providers, `sql.ts`, `sql_codegen.ts`) route through `GeneratedViewName`. This was the highest-risk area (two names drifting → a view under a name nothing reads) and it's genuinely centralized.
- **CRUD reads the outer view in both dialects** (SQL Server via `entity.BaseView`, PG via `getBaseViewName()` = `BaseView`), so custom columns come back on write.
- **Permissions target the outer view** in both dialects; ownership-chaining reasoning is correct.
- **Drift detection uses the inner name** (`checkBaseViewChangedInDB` → `GeneratedViewName`), so a layered entity won't force-regenerate every run.
- **Defense in depth against self-reference**: DB `CHECK` + case-insensitive TS guard, and the CHECK's collation-based `<>` aligns with `HasLayeredBaseView`'s `toLowerCase()`.
- **Tests assert the right thing** — the "NEVER writes the outer view" absence test is exactly the failure mode that matters. Migration naming/versioning/conventions all correct; it's the highest-numbered file.

### Concerning

1. **PostgreSQL does not actually deliver the core promise, and the docs hide it.** The whole value proposition — "add a foreign key later and its display field appears on its own" — depends on the *outer* `SELECT g.*` view being refreshed to pick up new inner columns. On SQL Server, CodeGen emits `sp_refreshview` on the outer BaseView, so that works. **On PostgreSQL, `SELECT *` in a view is expanded and frozen at creation, there is no `sp_refreshview` equivalent, and CodeGen never recreates the app-owned outer view.** So a late-added column silently never appears in the PG outer view — the exact silent-staleness failure this feature exists to eliminate. The README/CLAUDE.md frame this as *"SQL Server only; PostgreSQL does not need it,"* which is misleading: PG doesn't *need the ordering* because it lacks the *mechanism*, and the consequence is real staleness. This should be documented as a known PG limitation (or PG layering explicitly deferred), not glossed.

2. **Bootstrap ordering is a genuine sharp edge that the docs undersell.** `sp_refreshview` is emitted unguarded (no `IF OBJECT_ID`) against the outer BaseView on the `!BaseViewGenerated` path, and a `GRANT` is emitted on it too. On the first codegen pass after enabling layering, the app-owned outer view doesn't exist yet (it's created by a later-sorting migration that itself depends on the inner view). The generated refresh/grant against a not-yet-existing outer view will error. This is adjacent to the pre-existing fully-custom-view requirement, but layering makes it worse because of the inner→outer dependency. Worth either guarding the outer `sp_refreshview`/grant, or spelling out the exact migration-timestamp ordering (app BaseView migration must sort *after* the `CodeGen_Run` migration) with the failure symptom.

3. **`recompileAllBaseViews`: the inner generated view can't self-heal.** A layered entity lands only in `customViewEntities` (because `BaseViewGenerated = 0`), never in `l`, and `identifyFailedViewRefreshes(ds, l)` only regenerates entities in `l`. So in this path the inner view can be refreshed but never regenerated on failure. Minor — the main generation path covers regen — but worth knowing.

### What's missing

- **No end-to-end integration test.** Every test here is unit-level against a hand-built `EntityInfo`; nothing exercises migrate → codegen → live layered view → `RunView` surfacing a late-added FK column. Given that the feature is **entirely dormant until the release-time codegen regenerates `vwEntities`/`MJCoreEntities`** (correctly deferred per migration rules, but it means zero live exercise in this PR), an IT bundle that actually builds a layered view and asserts a new column appears would move this from "logic looks airtight" to "proven."

### Bottom line

The SQL Server path is careful, well-tested, and correctly additive. Before it's presented as cross-dialect, fix the PostgreSQL documentation (#1) so no one adopts layering on PG expecting auto-refresh, and tighten the bootstrap-ordering guidance (#2). The integration test is the highest-value follow-up.

*Reviewed with Claude Code.*

AN-BC:
@rkihm-BC thanks, I'll incorporate your feedback and get it implemented now. @MS-BC  I'll take this one.

MS-BC:
Reviewed and tested live against SQL Server 2022. This one holds up — the central mechanism, the ordering caveat, the bootstrap guard, and the constraint all behave exactly as documented. Details below, then three small things.

## The refresh-ordering caveat is real, and I reproduced the exact failure

This is the claim the whole design rests on, so I tested it rather than taking it on faith. First attempt didn't reproduce it, because I regenerated the inner view with an explicit column list — which masks the problem. Using the shape CodeGen actually emits (`SELECT o.*` for the base table), it reproduces precisely:

```
-- table gains a column via migration; no view recreated
ALTER TABLE dbo.tOrders ADD DueDate DATE NULL;

### ORDER A (outer refreshed first, then inner)
inner_col:              ID, Balance, DueDate
OUTER_COL_wrong_order:  ID, Balance, IsOverdue     <-- DueDate silently absent

### then the PR's order (inner first, then outer), same database
OUTER_COL_correct_order: ID, Balance, DueDate, IsOverdue
```

So the outer view keeps a stale column list, the new column is *absent* rather than wrong, and nothing errors — which is exactly the silent failure mode the PR describes. Ordering it explicitly in both emission paths is load-bearing, not defensive. Worth noting for anyone reviewing later that this only bites because generated base views use `<alias>.*`; that's easy to miss and makes the comment in `sql.ts` worth keeping verbatim.

## The bootstrap guard works in both states

Ran the exact SQL `SQLServerCodeGenProvider.generateIfViewExistsSQL` emits, against both states it has to survive:

```
### BOOTSTRAP PASS - outer view does not exist yet
bootstrap: guarded refresh + GRANT both skipped, run survived

### STEADY STATE - outer view now exists
grantee        permission_name  state_desc  obj
cdp_Developer  SELECT           GRANT       vwOrders
```

Both the `sp_refreshview` and the `GRANT` bodies survive the `EXEC sp_executesql N'...'` wrapping, the missing-view pass is a clean skip rather than a failure, and the grant genuinely lands once the view exists. The chicken-and-egg problem on the first pass is real and this handles it.

## The CHECK constraint does what its comment says

Applied the migration to a scratch database and exercised every arm, including the `BaseView IS NOT NULL` one the comment specifically argues for:

```
db_collation: SQL_Latin1_General_CP1_CI_AS
case 1  GeneratedBaseViewName NULL (every existing row)   ACCEPTED (expected)
case 2  proper layering (inner <> BaseView)               ACCEPTED (expected)
case 3  self-reference (inner = BaseView)                 REJECTED (expected)
case 4  inner set but BaseView NULL (the UNKNOWN arm)     REJECTED (expected)
case 5  differs only by case ('vwFoo' / 'VWFOO')          REJECTED
```

Case 4 confirms the reasoning in the migration comment was correct — without that arm the row would have passed on UNKNOWN.

Case 5 is the one place SQL and TypeScript reach the same outcome by different routes: the constraint rejects it because the collation is case-insensitive, while `HasLayeredBaseView` would have classified it "not layered" and quietly ignored the column. Both avoid a broken state, so this isn't a defect. It is, strictly, collation-dependent — under a case-sensitive database collation the constraint would accept `VWFOO` / `vwFoo` as distinct (correctly, since they'd be distinct objects) while `HasLayeredBaseView` still says "not layered" and CodeGen writes nothing. Extremely narrow, and arguably not worth code, but worth knowing it's the one edge where the two disagree.

## Migration hygiene

Checked for the version-collision problem using Skyway's own `DetectDuplicateMigrations` over v5 + v6:

```
212 migrations, 0 version collision(s)
```

Clean. `V202608031959` is unique and `migrations/v6/` is new on this branch.

## Tests

- **MJCore: 1613/1613 passing** (105 files).
- **CodeGenLib: 646 passing, 10 failing** — all 10 in `src/__tests__/pg-connection.test.ts`, which this PR does not touch. I confirmed they're pre-existing by running that file outside this branch, where the same 10 fail. Not attributable here.

One note on reproducing this locally: the new tests in `LayeredBaseView.test.ts` and `PostgreSQLCodeGenProvider.test.ts` fail with a confusing "expected ... to contain 'CREATE VIEW [orders].[vwOrderHeadersGenerated]'" if `@memberjunction/core` resolves to a build that predates `GeneratedViewName` / `HasLayeredBaseView`. That's a stale-dist artifact, not the tests — worth knowing since the symptom looks like the feature is broken rather than unbuilt.

## Three small things

1. **The feature is inert until entity metadata regenerates**, which "Not done here" says, but the consequence is worth spelling out: if someone sets `GeneratedBaseViewName` in the database before the release that regenerates `vwEntities` and the `MJCoreEntities` classes, `EntityInfo.GeneratedBaseViewName` stays null, `HasLayeredBaseView` is false, and with `BaseViewGenerated = 0` CodeGen writes *nothing* — so the application's outer view selects from an inner view that was never created. That fails loudly rather than corrupting anything, which is the right shape, but a line in the README saying "do not set this column until the release that regenerates entity metadata" would save someone an afternoon.

2. **The extended-property text and the PR body describe PostgreSQL differently.** The migration says CodeGen "rejects this column on PostgreSQL", which matches `assertLayeredBaseViewSupported`. The PR body says the refresh is "ordered explicitly in both emission paths (SQL Server only; PG doesn't need it)", which reads like PG supports layering without needing the ordering. Same fact, but the body is the more likely thing to be read.

3. `packages/GraphQLDataProvider/src/version.generated.ts` (+1/-1) looks like incidental generated churn rather than part of this change.

## Net

The design is sound and the two subtle things it hinges on — refresh ordering and first-pass bootstrapping — are both correct and now verified live rather than argued. No defects found. My only real ask is the README line in point 1.

AN-BC:
## @cadam11 — input wanted on raising the root `@types/node` override

This PR carries a one-line change to the root `package.json` that is unrelated to its subject matter, and it's repo-wide, so I'd like your read before it goes in.

```diff
   "overrides": {
-    "@types/node": "20.14.2",
+    "@types/node": "24.10.11",
```

### How it surfaced

`npm install` at the repo root (needed to link the new `@memberjunction/standards` workspace) re-resolved the tree and broke `@memberjunction/server`:

```
src/context.ts(436,27): error TS2345:
  Argument of type 'Buffer' is not assignable to parameter of type 'ArrayBufferView'.
```

That line is `timingSafeEqual(...)` over `createHash().digest()` — untouched code from the system-API-key hardening on this branch.

### What's actually going on

The committed `package-lock.json` was **stale** — it predated the current `overrides` block. It still carried nested `packages/*/node_modules/@types/node@24.10.11` entries (plus nested `rimraf`/`glob`/`minimatch`/`lru-cache`/`path-scurry`), so MJServer had been compiling against 24.x even though the override says 20.14.2. A real re-resolve applied the overrides properly, dropped ~4,100 lines of now-redundant nested entries, and MJServer fell back to the root's 20.14.2 — where it no longer typechecks.

So the conflict is pre-existing; the stale lock was masking it.

### Why raising it looks like the right direction

Every package in the repo already asks for 24.10.11 — **nothing pins 20.x anywhere**:

| Declaration | Count |
|---|---|
| `"@types/node": "24.10.11"` (devDependency) | 83 |
| `"@types/node": "^24.10.11"` (devDependency) | 22 |
| `"@types/node": "24.10.11"` (dependency, MJServer) | 1 |

And the history says the override is simply vestigial:

- `c4835e8206` — **2025-02-07** — "Add override for node typings" → pins 20.14.2
- `69edd182a4` — **2026-02-06** — "mass dependency upgrade across 225 files (Phase 3a+3b)" → moves 106 packages to 24.10.11, **without updating the override**

A year apart. The upgrade raised every declaration and left the override behind, and the stale lock meant nobody felt it.

### Verification

- Cold full build (`TURBO_FORCE=true`): **268/268 tasks, 0 cached, 0 failures**
- No packages dropped from the build graph
- `MJServer` typechecks clean
- MJCore **1617/1617**, CodeGenLib **656 passed / 60 skipped**

### The question

Do you remember what the 20.14.2 override was originally protecting against? It predates my visibility on it, and an override usually exists because *something* transitively demanded it. A cold build is green, so if there was a reason it appears to be gone — but you'd know faster than I'd find out. Happy to split this into its own PR if you'd rather it not ride along here.

AN-BC:
## @cadam11 — following up on the `@types/node` override (this has now merged)

Flagging again since this landed with the PR and it's a repo-wide change I'd rather you had eyes on after the fact than not at all.

**What changed** — one line in the root `package.json`:

```diff
   "overrides": {
-    "@types/node": "20.14.2",
+    "@types/node": "24.10.11",
```

**Why.** A root `npm install` (needed to link the new `@memberjunction/standards` workspace) re-resolved the tree and broke `@memberjunction/server` with `TS2345: Buffer is not assignable to ArrayBufferView` in `context.ts` — untouched code from the system-API-key hardening.

The cause was a **stale `package-lock.json`**: it predated the current `overrides` block and still carried nested `packages/*/node_modules/@types/node@24.10.11` entries, so MJServer had been compiling against 24.x even though the override said 20.14.2. A real re-resolve applied the overrides properly and MJServer fell back to the root's 20.x, where it no longer typechecks. So the conflict was pre-existing — the stale lock was masking it.

**Why raising it looked right.** Every package in the repo already asks for 24.10.11; nothing pins 20.x anywhere:

| Declaration | Count |
|---|---|
| `"@types/node": "24.10.11"` (devDependency) | 83 |
| `"@types/node": "^24.10.11"` (devDependency) | 22 |
| `"@types/node": "24.10.11"` (dependency, MJServer) | 1 |

And the history says the override is simply vestigial:

- `c4835e8206` — **2025-02-07** — "Add override for node typings" → pins 20.14.2
- `69edd182a4` — **2026-02-06** — "mass dependency upgrade across 225 files" → moves 106 packages to 24.10.11, **without updating the override**

A year apart. The upgrade raised every declaration and left the override behind.

**Verification before merge:** cold full build (`TURBO_FORCE=true`) 268/268, 0 cached, 0 failures; no packages dropped from the build graph; MJServer typechecks clean; MJCore 1620/1620, CodeGenLib 703 passed / 60 skipped.

**The actual question:** do you remember what the 20.14.2 override was originally protecting against? An override usually exists because *something* transitively demanded it. A cold build is green so it looks resolved, but you'd know faster than we'd rediscover it — and if there was a reason, it's better found now than at release. Happy to revert or narrow it if you think it's still load-bearing.

AN-BC:
## @rkihm-BC — your review is fully addressed, including the integration test

Closing the loop now that this has merged. All four items from your review landed:

**1. PostgreSQL documentation gap** → CodeGen now **refuses** layering on PG rather than documenting a footgun. `PostgreSQLCodeGenProvider.generateBaseView` throws when `HasLayeredBaseView`. Your diagnosis was right and the behaviour is actually worse than "doesn't refresh": it's *intermittent*. Adding a column leaves the outer view stale (`CREATE OR REPLACE` on the inner never touches dependents), while a rename or type change raises `42P16` and sends CodeGen down the capture/`DROP CASCADE`/replay path that incidentally *does* recreate it. Same feature, opposite outcomes, decided by which kind of schema change landed that day.

**2. Bootstrap ordering** → new `generateIfViewExistsSQL` provider method; the `sp_refreshview` and `GRANT` aimed at the app-owned `BaseView` are wrapped in `IF OBJECT_ID(...) IS NOT NULL`, scoped to layered entities only.

**3. `recompileAllBaseViews` self-heal** → `identifyFailedViewRefreshes` / `regenerateFailedBaseViews` now cover layered entities and key on `GeneratedViewName`.

**4. The integration test** → **IT69 - Layered Base Views**, six checks (LBV1–LBV6), in the deterministic suite.

### On IT69 specifically

You said an IT bundle would move this "from *logic looks airtight* to *proven*." Two notes on how it's built, because you'll care about both:

**It's proven by sabotage, not by assumption.** Rewriting the wrapper to drop `SELECT g.*` makes LBV6 fail with `RunView failed selecting generated display column(s) UserViewRun through BaseView`. Restored after; green.

**LBV1 and LBV6 deliberately don't touch `sys.*`.** The first version gated all six checks on `ctx.Pool` and reported **6/6 green on a run path that provides no pool** — six checks, zero assertions executed, indistinguishable from six real passes. Only caught it because a sabotaged view stayed green. The two checks carrying the actual promise now run off metadata + `RunView` on every platform and run path; only the catalog checks can skip, and they log `SKIPPED (no assertions ran)` rather than a bare pass.

Also worth noting the feature is no longer dormant, which was your point about zero live exercise: `MJ: Version Installations` and `MJ: User View Run Details` are layered in MJ core as of v6.1, and the latter **gained** `UserViewRun` — a related-entity display field its hand-written view never had — with nobody editing SQL. That's the promise, live.

### One thing that fell out of writing it, which you'll want to know about

Chasing the vacuous-pass problem above turned up the same pattern in an existing bundle: **`metadata-consistency`'s entire check set skips-as-pass, in CI**.

From the `Integration (SQL Server, deterministic)` job on a recent green run:

```
→ MC1 skipped: no mssql pool on this transport (PostgreSQL / client bootstrap)
[IntegrationTestDriver] ✓ metadata-consistency.MC1
```

All seven implemented checks — MC1–MC6 and MC8 — skip for want of `ctx.Pool` and report ✓. On the SQL Server job. So the catalog audit (base views present, CRUD procs present, CHECK-constraint ↔ `EntityFieldValue` parity, FK indexes, field-sequence integrity, description coverage, SchemaInfo casing) has been green without executing.

I haven't touched it — not this PR's bundle, and the right fix is arguably in the driver's pool resolution rather than in the bundle. But it means the deterministic tier is asserting less than its results suggest, and it seemed worth putting in front of you rather than leaving in a merged PR description.

PR #3425, opened 2026-08-04
fix(core,server): stop consumers corrupting the process-wide RunView cache (freeze-on-write)

CaelebB-BC:
## What shipped

A P1 fix, plus the structural change that makes its whole bug class unreachable.

**The P1.** `ResolverBase` mapped GraphQL transport field names onto the data provider's own result rows, which the server cache holds **by reference**. Preparing one GraphQL response therefore rewrote `__mj_CreatedAt` to the wire alias `_mj__CreatedAt` *inside the live cache*, and every later read served the corrupted shape — failing in `BaseEntity.SetMany` with `Field _mj__CreatedAt does not exist on <Entity>`. The cache is process-wide, so a single response poisoned every subsequent request across all workers, until restart. Fixed by mapping onto copies.

**The structural half.** Fixing the reader alone left the class open — nothing in the type system or API surface said "this array is shared, do not mutate," and the exposure runs both ways (a cache *hit* returns the stored array; a cache *miss* stores the array it is about to return). So the cache now defends itself:

- **`ILocalStorageProvider` gains an optional `readonly SharesReferences?: boolean`**, declaring whether a provider hands back live references (in-memory) or serialized copies (IndexedDB, localStorage, Redis, MMKV). **Fully backward compatible**: existing implementations keep compiling, and omitting it is *not* an opt-out — `LocalCacheManager` measures any provider that doesn't declare one (store a sentinel, read it back, compare identity).
- **`LocalCacheManager` deep-freezes row data at write time** — rows, nested values, and the array itself — but only when the provider shares references. Mutations then throw a `TypeError` at the offending line instead of silently corrupting shared state, and cache **hits cost nothing extra**. Applied at **all four** write funnels: `SetRunViewResult`, `SetRunQueryResult`, `SetDataset`, and `storeCachedResults` (the in-place slot-maintenance path that bypasses the first).
- The freeze lands immediately after the only gate that can decline a write (a synchronous size check) and **before** the awaited eviction steps. Callers don't always await these methods, so a yield point before the freeze is a window in which shared rows get handed out still mutable.
- The freeze decision **follows the active provider**. MJAPI initializes on the in-memory provider during engine loading and swaps to Redis afterward — two providers with opposite semantics in one process — so it is re-resolved on every swap, resolved *before* the provider is published, and a provider that declares the property is re-read live on each write.
- The deep-freeze **skips binary payloads** (`Buffer`/TypedArray/`ArrayBuffer`, e.g. `varbinary` columns — `Object.freeze` throws on non-empty views by spec), terminates cycles via an explicit visited set (not `Object.isFrozen`, which would skip the subtree of an input that arrived shallow-frozen), and degrades to a logged partial-protection store on any failure — it can never fail a `RunView`/`RunQuery`.
- **`CacheWriteOptions.ProviderInternalScaffolding`** exempts slots whose only consumer is the provider that wrote them — scoped to the **`MJ_Metadata` dataset only**. Metadata bootstrap needs it: `PostProcessEntityMetadata` and `GetAllMetadata`'s Applications assembly hydrate their object graph by mutating those rows in place.
- **Dataset slots get their own key namespace.** `GetDatasetByName` keyed its write-through with the same fingerprint builder ordinary reads use, and every shipped dataset item has a NULL `WhereClause` — so a dataset item and a plain unfiltered `RunView` of the same entity produced an **identical** key and shared one slot. That leaked the scaffolding exemption above to ordinary callers of `MJ: Entities` / `MJ: Entity Fields` (the most-read entities in the process, served *unfrozen*), and in the other direction let an ordinary read repopulate an evicted slot **frozen** so the next metadata refresh threw. `GenerateRunViewFingerprint` now takes an optional dataset segment, appended only when supplied — ordinary reads keep their exact pre-existing key, so no existing cache entry changes.

## Pre-existing consumer bugs the freeze surfaced, and fixed

- **`BaseEntity.Get()` wrote to its own source row.** The raw-mode fast path keeps the caller's row by reference, and `Get()` wrote back into it to memoize a converted `Date` or an rtrimmed fixed-width string — so on a cache-served row, *reading* a `datetime` or `CHAR(n)` field threw. This broke AI cost calculation on `MJ: AI Model Costs.Currency`. `Get()` now memoizes into a **per-instance side table** and never writes to the row. Gating the write on a once-sampled `Object.isFrozen` was tried and is not sufficient: the freeze is asynchronous relative to the consumer (cache writes aren't always awaited), so the sample goes stale and the write still throws. Keeping the memo off the row makes freeze timing irrelevant *and* restores the optimization on frozen rows.
- **`ResolverBase.MapFieldNamesToCodeNames` renamed fields on its argument** — routinely a cache-held row. With the freeze in place, `UserByEmail`, `UserByID`, `UserByEmployeeID` and every CodeGen-generated single-record resolver over a cached entity threw `Cannot add property _mj__CreatedAt, object is not extensible` (reproduced live against a running MJAPI). Before the freeze it did something quieter and worse: it rewrote the cached row's keys. It now returns a copy, which fixes every call site at once; `ArrayMapFieldNamesToCodeNames` likewise returns a new array of new objects.
- **Cache-served reads skipped the `PostRunView` hook chain.** Hooks receive `contextUser`, so masking is *per-user* while a slot is *shared* — there is no correct way to apply it once at write time for a reader who hasn't arrived. Three of the four server paths ran the chain; the singular cache hit and the all-cached batch returned early, so masking depended on whether a *sibling* view in the same batch happened to miss. It had looked correct only by accident: the cache write precedes the hooks, so an in-place masking hook wrote *through* into the cached rows — which both made later hits appear masked and baked one user's masking decision into a shared slot. Both hit paths now run the chain, guarded so the zero-hook default costs **~80ns** rather than ~2.4µs.
- **The singular server `RunView` path dropped a hook's returned replacement result** (it reassigned a local; `RunView` returned the pre-hook reference), while the client and batch paths honored it. `PostRunView` now copies a hook-supplied replacement onto the result object it was handed — its `Promise<void>` signature is unchanged, so external subclasses that override it keep compiling.
- **`ProjectRowsToFields`' full-coverage fast path** returned the shared frozen array when a `Fields` list happened to cover every stored key, contradicting the documented promise that `Fields` callers get per-caller rows they may mutate. The fast path is now taken only on unfrozen input.
- **`GenericDatabaseProvider.serveFromServerCache` and the smart-cache legs** duplicated `CachedRunViewResult` as four inline structural types, which had already caused one silent field drop; they now share the canonical type.

## Compatibility

**No breaking changes** — every public signature this touches is additive or unchanged. The cache result types stay ordinary mutable arrays, documented as shared-and-frozen: the runtime freeze is the enforcement, and a `readonly` marker was tried and reverted because it forced casts at every transport boundary and would have broken existing downstream readers without adding protection.

Browser clients are unaffected (IndexedDB / localStorage serialize). **Node-side clients — the CLI, MetadataSync, and anything else on an in-memory provider — do get the freeze**, so "client behavior is unchanged" holds for the browser specifically.

Consumer-facing contract, documented in `guides/CACHING_AND_PUBSUB_GUIDE.md`: **treat rows from `RunView`/`RunViews`/`RunQuery` as read-only** unless you produced them. Copy before mutating — `rows.map(r => ({ ...r }))`, `[...rows].sort(...)`. `Fields` requests and `ResultType: 'entity_object'` results are unaffected (both get per-caller objects).

## Verification

Unit: **MJCore 1759**, GenericDatabaseProvider 868 (+5 skipped), MJServer 802 (+56 skipped), integration-test-suite 155, GraphQLDataProvider 274, MJCoreEntities 573. Every fix was written test-first and proven red before the fix landed.

Integration: the deterministic tier at **54/56**, including **IT71 — Cache Immutability (freeze-on-write)**, a 15-check bundle run against the real server stack. F15 specifically proves a plain unfiltered `RunView` of a metadata entity comes back frozen while `MJ_Metadata` keeps its exemption — the case an earlier verification pass missed, because it used only filtered and row-limited queries, which generate different cache keys. The two failures are IT65 and IT69, which require v6.1 migrations not applied to the verification database.

Live: `UserByEmail` confirmed returning data on a running MJAPI (was a deterministic 500), and the metadata-slot collision reproduced and then confirmed closed via the V8 inspector.

## Follow-up filed

#3555 — `GetGlobalObjectStore` throws and catches a `ReferenceError` on every call under Node (~1.4µs), taxing every `BaseSingleton` `.Instance` access. Worked around locally in the one hot spot this PR needed; the source fix is repo-wide and deliberately out of scope here.


packages/GenericDatabaseProvider/src/GenericDatabaseProvider.ts
```diff
@@ -2637,13 +2638,16 @@ export abstract class GenericDatabaseProvider extends DatabaseProviderBase {
      */
     private async serveFromServerCache<T = unknown>(
```
AN-BC, line 2717: probably should make this a protected method so it can be overridden in subclasses if needed

packages/MJCore/src/generic/baseEntity.ts
```diff
@@ -785,6 +785,19 @@ export abstract class BaseEntity<T = unknown> {
...
+     *
+     * `LoadFromData`'s fast path keeps the caller's row BY REFERENCE, and that row is frequently
+     * a LocalCacheManager cache entry — which the cache deep-freezes on reference-sharing storage
+     * providers so consumers cannot corrupt shared state. `Get()` writes back into `_raw` purely
+     * to memoize a converted Date / rtrimmed fixed-width string, so on a frozen row that write
+     * throws and turns a plain field READ into an error.
+     *
+     * Computed once at load (cheap) instead of calling `Object.isFrozen` on every field read.
+     */
+    private _rawIsFrozen: boolean = false;
```
AN-BC: seems reasonable to me but would like @jordanfanapour and @rkihm-BC to put eyes on this (and overall PR)

AN-BC (commented):
Overall seems like a good finding and result, some comments here and would like eyes on this from @rkihm-BC and @jordanfanapour before we merge.

AN-BC:
@CaelebB-BC I'm also having Fable5/Ultracode do a deep pass on this for another "set of eyes"

AN-BC:
## Deep-pass review — Claude Fable 5 / Ultracode

cc @rkihm-BC — requested second-set-of-eyes pass on this PR (see the note upthread).

### How this review was done

This is the Ultracode deep pass mentioned above, run by **Claude Fable 5** using multi-agent orchestration against the PR head `9075737`:

- **8 dimension reviewers** worked in parallel over the full PR-head checkout (not just the diff hunks): freeze machinery, the `BaseEntity` raw-row fix, provider hook paths, MJServer resolvers, provider `SharesReferences` declarations + the scaffolding exemption, test quality, repo-convention compliance, and a repo-wide blast-radius sweep for consumers that mutate RunView results in place.
- Every candidate finding was then handed to an **independent adversarial verifier** instructed to *refute* it against the actual code — trace the scenario end-to-end or kill it. A completeness critic swept for missed angles afterward. 26 agents total; **findings that did not survive verification were dropped**, and several that did survive had their severity or blast radius *corrected downward* by the verifier (noted inline where it matters).
- Key items were additionally hand-verified (the fingerprint collision, the `Sequence` duplication, `ProjectRowsToFields`, the localStorage fallback semantics).

**Verdict up front:** the core of this PR is genuinely strong. The P1 root-cause analysis is correct, the freeze machinery and write-gate ordering are sound on the paths the PR exercises, the provider declarations all check out against their implementations, and the red-first tests pin real contracts. The confirmed problems are in the **seams** — paths that weren't converted, funnels that freeze too late, and one design premise that doesn't hold. **1 blocker, 7 majors, 9 minors.** The blocker and majors need changes before merge.

---

### 🔴 Blocker

**1. TOCTOU: the smart-cache "stale" path freezes rows *after* they're handed to entity loading — reintroducing the P1 failure class as a race.**

`providerBase.ts:2810-2818` fires `SetRunViewResult` on `checkResult.results` **without awaiting** ("fire and forget"), then `:2822` awaits `TransformSimpleObjectToEntityObject` on the *same array*, whose rows each entity keeps by reference in `_raw`. Inside `SetRunViewResult` the freeze (`localCacheManager.ts:1911`) runs only after two awaited gates (`enforcePerEntityMemoryLimit` :1902, `evictIfNeeded` :1905) — under eviction pressure those do per-entry awaited `Remove`s + registry persistence, so the freeze lands arbitrarily late. Rows loaded before it lands probe `_rawIsFrozen = false` (`baseEntity.ts:3086`, probed **once**); when the background freeze then hits those same objects, the next `Get()` of a not-yet-read `Date`/`CHAR(n)` field executes `this._raw[FieldName] = ...` on a frozen object → strict-mode `TypeError` on a plain field **read**. Armed in every shipped **Node GraphQL client** (`graphQLDataProvider.ts:2819-2823` falls back to `InMemoryLocalStorageProvider`, now `SharesReferences: true`) for any `CacheLocal` batch with `ResultType: 'entity_object'` — the `BaseEngine` default.

**Fix (in depth):** two independent changes, both recommended:
1. *Remove the TOCTOU class in `BaseEntity` instead of patching the probe.* Stop memoizing into `_raw` at all: add a lazily-allocated private overlay (e.g. `_rawMemo: Record<string, unknown> | null`), check it before `_raw` in the raw-mode `Get()` fast path, and write converted Dates / trimmed strings there instead of into `_raw`. Clear it wherever `_raw` is cleared (init at `:2125`, hydration promotion at `:2183`). This deletes `_rawIsFrozen` entirely (field + probe), is correct no matter *when* the cache freezes the row, restores memoization for frozen rows (fixing the "frozen rows re-parse Dates forever" regression noted under minors), and restores stable `Date` identity across reads. Cheapest correct alternative if you want minimal diff: re-check `Object.isFrozen(this._raw)` at memo-write time rather than only at load — but that keeps the per-read cost and the lost-memo behavior.
2. *Close the handout-then-freeze ordering on the stale branch.* Either hoist `freezeRowDataIfProviderSharesReferences` above the first `await` inside `SetRunViewResult` (the helper is fully synchronous — the freeze then executes during the synchronous prefix of the fire-and-forget call, i.e. **before** `TransformSimpleObjectToEntityObject` ever runs), or `await` the store at `:2810`, or pass the cache a row-level copy (`checkResult.results.map(r => ({ ...r }))`). If you hoist, preserve the "a result the cache declines to store stays mutable" invariant: compute the decline decision before freezing (see Major 6, same hoist).

### 🟠 Major

**2. `UserResolver` still renames `__mj_*` in place on frozen rows — deterministic 500s on three public GraphQL queries.**

`UserByID` (`UserResolver.ts:96`), `UserByEmployeeID` (`:106`), and `UserByEmail` (`:118`) call `MapFieldNamesToCodeNames('MJ: Users', retVal)`, which mutates in place (`ResolverBase.ts:92-93`), on rows straight from `findBy` (`ResolverBase.ts:324-353`) — the provider's own result objects, already frozen by the time `RunView` returns (PostRunView awaits the cache write). First call throws `TypeError: Cannot add property _mj__CreatedAt, object is not extensible` in the default no-Redis MJAPI config. (Verifier correction to the original claim: Explorer login is *not* broken — `CurrentUser`'s primary path builds a fresh object — but the three queries themselves are broken for any caller, and neither test tier touches them.) The PR converted the two `RunView*GenericInternal` sites and missed these.

**Fix (in depth):** make `MapFieldNamesToCodeNames` non-mutating — build the renamed object onto a copy and return it — then update the three call sites to use the return value (`retVal = this.MapFieldNamesToCodeNames('MJ: Users', retVal ? { ...retVal } : retVal)` if you keep it mutating; better is `return this.MapFieldNamesToCodeNames(...)` with the method itself copying). The non-mutating version is strictly better: it makes it impossible for any future `findBy` consumer to reintroduce this crash. Sweep the rest of MJServer for other callers of `MapFieldNamesToCodeNames` / `FieldMapper.MapFields` on provider-owned rows while you're there. Add a unit test that runs one of these resolvers against a frozen row (same `Probe` technique as `ResolverBase.transportMapping.test.ts`).

**3. The scaffolding exemption's "single owner" premise is false: `MJ_Metadata` dataset slots share the plain-RunView fingerprint namespace.** *(Two failure directions, one root cause.)*

`GetDatasetByName` keys each dataset item as `GenerateRunViewFingerprint({EntityName, ExtraFilter}, conn)` (`GenericDatabaseProvider.ts:4188/4229`), and every `MJ_Metadata` item has no WhereClause — so the key is **byte-identical** to a plain unfiltered RunView of the same entity (`providerBase.ts:2297`). Nothing on any read or write path checks `providerInternalScaffolding` except the freeze-skip and the maintenance carry-forward. On a default in-memory server:

- **(a) Unfrozen handout:** any cache-eligible `RunView('MJ: Entities')` (server-side, `TrustLocalCacheCompletely=true`, `AllowCaching=1` on all `__mj` entities) *hits the scaffolding slot* and receives the **unfrozen**, assembly-mutated arrays by reference — falsifying the documented "caller-facing results are always frozen" guarantee (`localCacheManager.ts:203-205`) for the hottest entities in the process, and leaving exactly the corruption class this PR closes open on them.
- **(b) Frozen repopulation → metadata refresh permanently broken:** if the slot is evicted or schema-hash-invalidated, a plain RunView repopulates the same fingerprint **frozen** via `PostRunView` (`providerBase.ts:2933-2941`, no flag) — `EntityResolver.ts:48-51` (`RunView({EntityName: 'MJ: Entities', ExtraFilter: ''})`) is a live in-repo collider, and `runFullQueryAndCacheResult` (`GenericDatabaseProvider.ts:2470`) is a second unflagged writer. `SetRunViewResult` never preserves an existing entry's flag on overwrite (`localCacheManager.ts:1869-1872`). The next soft metadata refresh feeds frozen rows to `PostProcessEntityMetadata`, whose in-place `sort`/attach throws; the catch at `providerBase.ts:3897` / `3634-3637` keeps old metadata — so the server **silently serves stale metadata until restart**, with every subsequent soft refresh failing the same way. (Verifier correction: this cannot produce the zero-metadata *boot* — the in-memory cache is empty at process start — the real outcome is the persistent silent refresh failure. Major, not blocker.)

**Fix (in depth):** the clean fix is **namespacing** — derive dataset write-through fingerprints with a distinguishing segment (e.g. `'ds:' + datasetName + '|' + fingerprint`) at both write sites and the hit path, accepting that this forfeits the (currently implicit) slot-sharing between the dataset path and plain RunViews — both hit paths must be adjusted together, and the first plain RunView after boot will re-query instead of hitting the dataset slot. If that sharing is load-bearing for warm-up performance, the surgical alternative is three pieces: (i) `SetRunViewResult` ORs in an existing entry's `providerInternalScaffolding` flag on overwrite (read it from the registry/stored entry before writing); (ii) `GetDatasetByName`'s hit path checks `Object.isFrozen(cached.results)` and copies (`cached.results.map(r => ({ ...r }))`) before handing rows to `PostProcessEntityMetadata`, so a frozen slot degrades to a copy instead of killing refresh; (iii) either serve scaffolding-slot hits to plain-RunView callers as copies, or explicitly document the unfrozen-handout residual. Either way, extend IT70: a plain unfiltered `RunView('MJ: Entities')` after bootstrap (assert the served array's frozen state matches the intended contract), and an evict-then-RunView-then-soft-refresh sequence (assert refresh still succeeds). F14 tests dataset *scoping* but not this collision.

**4. `SetStorageProvider` never re-resolves `_sharesReferences` — the shipped in-memory→Redis swap leaves the freeze permanently misconfigured.**

`_sharesReferences` is resolved once in `doInitialize` (`localCacheManager.ts:519`); `SetStorageProvider` (`:574-605`) swaps the provider without re-resolving. Every REDIS_URL-configured MJAPI takes this path: `ProviderBase.Config` initializes the manager with the default `InMemoryLocalStorageProvider` (freeze ON) *before* MJServer bootstrap swaps in Redis (`MJServer/src/index.ts:649`, and `:653` skips re-Initialize) — the flag stays `true`, so every subsequent write **deep-freezes the caller's own rows on a serializing deployment**: exactly what the design forbids and the PR description says doesn't happen. The soak had no Redis, so it couldn't see this. The reverse swap (serializing → sharing) silently *disarms* the freeze, quietly restoring the P1 class; the migration loop (`:588-599`) also writes through raw `SetItem` with no freeze gate.

**Fix (in depth):** in both branches of `SetStorageProvider` (initialized and pre-init), set `this._sharesReferences = await this.resolveSharesReferences(newProvider)` **before** publishing the new provider to `_storageProvider`, and route the migration loop's writes through `freezeRowDataIfProviderSharesReferences` (or simply skip migrating when the *target* declares `SharesReferences === false` — serialized stores don't need frozen input). Add two unit tests: init-with-InMemory → swap-to-serializing-mock → assert subsequent writes leave caller rows unfrozen; and the reverse swap → assert the freeze arms. While in there, also close the related init-ordering nit (Minor 5): assign `_storageProvider` and `_sharesReferences` together after the probe await.

**5. `PostRunView` hooks never run on the server singular cache-*hit* and all-cached batch paths — the PR makes its own documented hook pattern work on miss only.**

`providerBase.ts:738-753` (singular hit) returns `preResult.cachedResult` without `RunPostRunViewHooks`; `:886-897` (all-cached batch) skips `PostRunViews` — while the partial-batch loop (`:3110`) and the smart-cache path (`GenericDatabaseProvider.ts:2316`) *do* run hooks on cache-served rows. The skips pre-date this PR, but pre-PR an in-place masking hook's writes were accidentally baked into the shared cache rows, so hits happened to serve hook-shaped data; the freeze removes that accident, and the new `Object.assign` replacement fix (`:2982-2985`) applies on the miss path only. Net effect: a masking middleware written exactly per this PR's new docs (`MJServer/src/index.ts:771` auto-registers every `BaseServerMiddleware.PostRunView`, whose docstring advertises "data masking") returns `MASKED` on the first call and the **real value on the second**. The new docs never disclose the hit-path skip, and the PR's own test file sidesteps it by calling `ClearAllDataHooks()` before every cache-hit re-read.

**Fix (in depth):** run the hook chain on both hit paths with the same replacement handling as the miss path — on the singular hit, after `TransformSimpleObjectToEntityObject` (`:741`): `const hooked = await this.RunPostRunViewHooks(params, result, contextUser); if (hooked && hooked !== result) Object.assign(result, hooked);` and the equivalent `PostRunViews` invocation on the all-cached early return. Hooks already receive frozen rows on misses, so the copy-based contract the PR documents is unchanged; when no hooks are registered this is a no-op. If the team instead decides hit-path hooks are intentionally skipped (a defensible perf stance), then: say so loudly in `dataHooks.ts`, `BaseServerMiddleware.ts`, and the guide; remove "data masking" from the middleware docstring's advertised use cases; and add a test pinning the skip so it's a contract, not an accident. Either way, unify replacement semantics with the batch path first (see Minor 8) so hooks behave identically wherever they *do* run.

**6. The RunQuery miss funnel's freeze is racy — silent corruption remains possible on that path under cache pressure.**

`providerBase.ts:1978` (and the smart-validation rewrite at `:1933`) call `SetRunQueryResult` **unawaited**, and inside it the freeze (`localCacheManager.ts:2580`) runs only after `await evictIfNeeded` (`:2572`) — while the stored array is the same reference returned to the caller (`:1985`). With no eviction the freeze wins the race (the continuation runs before the caller's), but at/near `maxSizeBytes`, `evict()` suspends at real awaits and the caller's post-return mutation lands on **unfrozen rows that are then stored** — silent process-wide corruption of that slot, on the shipped default, for any saved query with `CacheLocal: true`. The guard vanishes exactly when the cache is busiest. (The RunView server funnels don't share this bug — `PostRunView`/`PostRunViews` await their stores.)

**Fix (in depth):** two options, pick per the invariant you want to keep. *(Option A, simplest and invariant-preserving)*: `await` the calls at `:1978` and `:1933` — the caller then can't run until the freeze has landed; costs eviction latency on the miss path only. *(Option B, systemic)*: restructure `SetRunQueryResult` (and `SetRunViewResult`, same shape — freeze at `:1911` sits after the awaited gates) so all decline decisions are computed synchronously first, then `freezeRowDataIfProviderSharesReferences(data)` runs **before the first `await`**, then eviction/store proceed. B also closes the Blocker-1 window at the source and keeps fire-and-forget cheap; just take care that any gate that can still decline the write after that point would violate the PR's own "declined results stay mutable" rule — if `enforcePerEntityMemoryLimit` can decline, hoist only above `evictIfNeeded`. Add a unit test that forces eviction pressure (tiny `maxSizeBytes`) and asserts rows are frozen before the store call returns control.

**7. Metadata hygiene: committed `sync` blocks + `lastModified` churn — and IT69/IT70 share `Sequence: 33`.**

Per `metadata/CLAUDE.md` ("PRs contribute ONLY the declarative metadata JSON changes … no `sync`"): `.IT70-cache-immutability.json:24-27` carries a sync block; `.IT69-layered-base-views.json`'s **entire** +4 diff is a newly added sync block (the PR has no other reason to touch that file); the new IT70 suite entry carries one; and exactly 20 unrelated suite entries had only `sync.lastModified` rewritten (checksums unchanged) — local `mj sync push` side-effect churn, and the direct cause of this PR's current merge conflict with `next` in that very file. (The rule demonstrably governs this tree: the release-gate commit `573060959` is what wrote the prior timestamps into it, and the IT69 PR `4e00e67fd` committed its files sync-free with zero churn.) Separately, IT69 and IT70 **both** sit at `Sequence: 33` (`.integration-suite.json:823/838`) — a tie, so the PR-description claim that IT70 "runs immediately after IT69" is not actually guaranteed by the data.

**Fix (in depth):** take `next`'s current version of `.integration-suite.json` and re-add only the IT70 suite-test entry (fields + `@lookup` + `primaryKey`, no `sync`) — that reverts the 20 churn hunks *and* resolves the merge conflict in one move. Strip the sync block from `.IT70-cache-immutability.json`; drop `.IT69-layered-base-views.json` from the PR entirely. Give IT70 a unique `Sequence` directly after IT69 (bump the later entries by one, keeping the server-bundle-before-client-bundle constraint from #3251 intact).

**8. Zero test coverage on the batch transport-mapping fix — a revert of `ResolverBase.ts:968` reintroduces the shipped P1 with green CI.**

The test file's `Probe` (`ResolverBase.transportMapping.test.ts:54-74`) drives only the singular `RunViewGenericInternal`; a repo-wide grep finds nothing exercising `RunViewsGenericInternal` (definition `:881`, caller `:561`), and IT70 operates at the core provider layer, never through ResolverBase. The batch path is the *standard* multi-view client transport (`graphQLDataProvider.ts:1091`), so the file header's claim "a revert to in-place mapping fails here" holds for one of the two fixed paths — a future refactor reverting the batch hunk ships a loud outage (frozen configs) or the silent P1 (unfrozen reference-sharing configs) with zero CI signal.

**Fix (in depth):** add a second `Probe` that reaches `RunViewsGenericInternal` with a fake provider whose `RunViews` returns shared row arrays; assert (i) the cached/source rows still carry `__mj_CreatedAt`, (ii) the wire rows carry `_mj__CreatedAt`, and (iii) wire rows are not identity-equal to source rows. Verify it red by temporarily reverting the batch hunk, as was done for the singular path. Close the two adjacent gaps in the same pass: a server **batch**-path hook test against frozen rows (replacement honored, in-place fails loudly), and a frozen-row **`Save()`** test — load an `entity_object` from a warmed cache slot, modify, `Save()`, assert success + correct slot maintenance. Today the workflow the P1 actually broke stops at in-memory `Set` in both tiers.

### 🟡 Minor

1. **`entity_subclasses.ts` drift** (`:44494/:44530`): the two JSDoc lines revert `CompactionTriggerPercent` wording to its pre-`V202608050100` state — codegen against a DB missing that migration (or a hand edit; both banned by `packages/MJCoreEntities/CLAUDE.md`). `next` has already flip-flopped this exact wording twice (`927b4832b` / `9c3b26195`). **Fix:** drop the hunk; it will silently revert on the next clean-DB codegen. (Also makes the missing `@memberjunction/core-entities` changeset entry moot.)
2. **`SetDataset` is a fourth write funnel with no freeze at all** (`localCacheManager.ts:1389`): no in-repo callers, but it's exported public API and the class docstring points consumers at it. **Fix:** apply `freezeRowDataIfProviderSharesReferences` to the dataset payload (no scaffolding writer uses this method), or deprecate the `SetDataset`/`GetDataset` pair and remove the docstring example.
3. **Half-frozen degrade + false log** (`localCacheManager.ts:1769-1775`): parent-first freezing means a mid-walk throw leaves everything already visited frozen; the catch logs "storing unfrozen" — false in the PR's own hostile-proxy fixture — and the slot ends up with mixed enforcement later maintenance can't repair. **Fix:** two-phase walk (collect via WeakSet, freeze only after a successful full traversal), or at minimum log "partially frozen" with the failing path.
4. **`isFrozen` short-circuit** (`localCacheManager.ts:38`): an externally *shallow*-frozen object entering the cache is treated as fully protected — its children stay mutable. No in-repo trigger, but: **Fix:** use a WeakSet visited-set for cycle/re-entry termination and recurse through already-frozen nodes (keep `isFrozen` only as a fast path for nodes this function froze itself).
5. **Init ordering** (`localCacheManager.ts:515/519`): `_storageProvider` is published before the probe resolves, and no write funnel checks `_initialized`. Unreachable with in-repo callers, cheap to harden. **Fix:** assign `_storageProvider` and `_sharesReferences` together after the await (folds into Major 4's change); add the probe-in-flight write test.
6. **`ProjectRowsToFields` full-coverage no-op returns the frozen shared array** (`providerBase.ts:208-219`): when the caller's `Fields` covers every stored key, the "narrow-Fields projections are fresh per-caller rows / stay mutable" contract (PR text + guide) is false — a caller following the documented contract throws. **Fix:** either skip the no-op when `Object.isFrozen(rows)` (always copy frozen input), or correct the PR text/guide to say full-coverage Fields may return the shared frozen array.
7. **Changeset / PR-text accuracy:** (a) "Serializing providers are untouched, so **client behavior is unchanged**" is false for **Node** clients — the InMemory fallback arms the freeze for every `CacheLocal` consumer (mj CLI, MetadataSync, headless agents), turning previously tolerated in-place mutation into a runtime `TypeError` under a *patch* bump. Scope the sentence, call the Node change out explicitly in the release notes, and consider a `minor` bump for `@memberjunction/core`. (b) The probe "**fails closed**" claim is inverted — on probe failure the code *disables* the freeze (`:494-496`, deliberately per its own comment); say "fails open to pre-freeze behavior". (c) The perf table covers hit cost only — the O(rows × fields) write-time freeze per miss and the `isFrozen` sweep per save/delete-driven maintenance deserve a number; the soak measured heap, not write latency.
8. **Hook-replacement semantics diverge:** the singular path *merges* (`Object.assign` — members a hook omits survive from the old result) while the batch loop *replaces* (omitted members are dropped). The batch path also hands out `Unsubscribe` before the hook loop, so a replacement result loses the caller's only unregister handle. **Fix:** pick one semantic (merge is the safer default), apply it on all paths, and register `Unsubscribe` after the hook loop (or re-attach it to the final result).
9. **Cleanups:** stale "RED until fixed" comments in five test files describe shipped code as still broken; the guide never states the `GetDatasetByName('MJ_Metadata')` residual (external callers receive live, unfrozen, provider-mutated rows with no copy at the boundary); the freeze test files use `test()`/global imports contrary to the repo's Vitest conventions in `.claude/rules/testing.md`.

---

### What was checked and held up

For balance — these were attacked and survived: the P1 root cause and the singular ResolverBase copy-then-map fix (probe-verified), the freeze/write-gate ordering in `SetRunViewResult`, all five provider `SharesReferences` declarations against their implementations (including the localStorage quota-exceeded edge — a dropped write, not a shared reference), `entity.GetAll()` isolation for slot maintenance, the `entity_object`/`BypassCache` exemptions, the IT70 instrumented-counter rework (F9/F10/F12), and the no-breaking-changes claim at the **signature** level. CI is green on `9075737`. Several panel findings did **not** survive adversarial verification and were dropped rather than listed.

---

AN-BC:
## Re-review of `8045a5d8` — Claude Fable 5 / Ultracode (Amith's review session)

This is the same Fable 5/Ultracode session that posted the deep-pass review above, re-verifying each finding against the new head. Every fix below was checked in the actual code, not taken from the commit message.

### ✅ All 8 blocking findings are fixed — and fixed structurally

| # | Finding | Verification at `8045a5d8` |
|---|---|---|
| B1 | `_rawIsFrozen` TOCTOU | Fixed at the root: the memo moved to a per-instance side table (`_rawConverted` Map) — `Get()` never writes into `_raw` at all, `_rawIsFrozen` is deleted, and the memo now works on frozen rows too (also resolving the re-parse-forever perf minor). Freeze timing is now irrelevant to reads. |
| M2 | `UserResolver` in-place rename | Fixed at the chokepoint: `MapFieldNamesToCodeNames` copies internally (`dataObject = { ...dataObject }`) and `ArrayMapFieldNamesToCodeNames` returns a new array of new objects. I checked every non-test caller — all 5 `UserResolver` sites and all 1,192 generated resolver sites consume the return value, so the internal copy fixes them all at once and makes the hazard unreachable for future callers. New `ResolverBase.frozenRecordMapping` test covers it. |
| M3 | Dataset/RunView fingerprint collision | Fixed by namespacing: `GenerateRunViewFingerprint` takes an optional `ds:` segment, appended only when supplied (existing RunView keys unchanged), and all three dataset paths — read (`:4188`), write-through (`:4232`), status (`:4426`) — use `datasetCacheSegment(datasetName, itemCode)`. The smart-cache RunView legs correctly keep plain keys. Both failure directions are closed; IT71 F15 pins that a plain unfiltered metadata read now comes back frozen. |
| M4 | `SetStorageProvider` stale freeze decision | Fixed twice over: both branches re-run `resolveSharesReferences`, and `activeProviderSharesReferences()` reads a *declared* value live off the active provider on every write, so even a future swap site that forgets to re-resolve stays correct for declaring providers. |
| M5 | Hooks skipped on cache hits | Both hit paths (singular + all-cached batch) now run the chain via `ApplyPostRunViewHooksToCacheHit`, with the same `Object.assign` replacement semantics as the miss path and a zero-hook fast guard. The guide now documents why hits must run hooks (per-user masking vs shared slots). |
| M6 | Racy freeze in the write funnels | The freeze was hoisted above the awaited eviction steps in **both** `SetRunViewResult` and `SetRunQueryResult`, with the decline gate confirmed synchronous and ahead of it — this also closes the providerBase fire-and-forget window from B1's second half. New `freezeOrdering` tests pin it. |
| M7 | Metadata hygiene | `.IT71` JSON is sync-free, the IT69 file is out of the PR, and the suite JSON's diff vs `next` is now a single clean hunk (the IT71 entry — no sync block, no `lastModified` churn), which also dissolved the merge conflict. The Sequence-33 tie remains, but that's now a 3-way tie `next` itself created (IT66/67/68 already tie at 32) — and IT71 stays ahead of the client-transport bundles at 34+, which was the real constraint. |
| M8 | Untested batch transport path | `RunViewsGenericInternal` now has its own Probe test asserting cache rows keep `__mj_CreatedAt`, wire rows carry `_mj__CreatedAt`, and wire rows are not the cached objects. |

The guide's inverted "fails closed" claim was also corrected — it now says the probe **fails open** and warns about the trade.

### ⚠️ Both current CI failures are inherited from `next`, not this PR

I traced both to base-branch breakage that arrived via this branch's merge of `next` — `next`'s own push CI shows the identical failures:

- **Run unit tests** — the DOM-spec coverage ratchet: `138 component(s) have no DOM spec and no deferral (gate: --max-none=137)`. This PR adds zero files under `packages/Angular/` and doesn't touch the gate or workflow; the 138th component came from `next` (#3467's workspace-tabs components). `next`'s Unit Tests runs have been red with this since `2d919c64` (Aug 6).
- **Integration (SQL Server, deterministic)** — dies during `mj migrate`, before any test runs: `V202608061704__…Phase0_Legacy_Workflow_Report_ScheduledAction_Retirement.sql` fails batch 19/32 with a `FK_EntityField_RelatedEntity` DELETE conflict on a fresh database. That migration is `next`'s #3553; this PR ships no migrations. `next`'s Integration runs flipped from success to failure at exactly `9d8cbb9e` — the #3553 merge.

Neither is fixable from this branch. Once fixes land on `next`, merge it in and CI should re-run clean — nothing in the PR-side diff is implicated in either failure.

### 🟡 Still open — all minor, none blocking

1. **`entity_subclasses.ts` drift is still in the diff** (the 2-line `CompactionTriggerPercent` JSDoc revert vs `next`). Still worth dropping — the next clean-DB codegen will silently flip it back.
2. **`SetDataset` (`localCacheManager.ts:1378`) remains an unfrozen public write funnel** — freeze sites are still only the three RunView/RunQuery/maintenance funnels. No in-repo caller, but it's documented public API.
3. **Freeze degrade path** still stores a half-frozen payload while logging "storing unfrozen", and `deepFreezeCacheValue` still short-circuits on externally shallow-frozen objects (no WeakSet visited-set). Defense-in-depth only.
4. **`doInitialize` still publishes `_storageProvider` before the probe resolves** (`:515` vs the await) — the pre-init `SetStorageProvider` branch got fixed, this one didn't. Unreachable in-repo; cheap to harden.
5. **`ProjectRowsToFields`' full-coverage no-op** still returns the frozen shared array when `Fields` covers every stored key, while the guide (line 711) still promises narrow-Fields callers "fresh per-caller row objects… stay mutable". Fix the no-op for frozen input or scope the doc claim.
6. **One stale "RED until fixed" comment** survives in `providerBase.postRunViewHooks.frozenRows.test.ts:15`.
7. **The PR description is now stale in places** — it still describes the `_rawIsFrozen` probe, IT70 at Sequence 33, and the "fails closed" probe; worth a refresh pass so the description matches what's actually shipped.

Nice incidental catch in the fix commit: memoizing `GetGlobalObjectStore` (the bare-`window` probe was ~1.4µs per call on the RunView hot path).

**Bottom line:** every change this review required is in and verified; what's left is minor cleanup plus waiting out two `next`-side CI breaks that this branch merely inherits.

---

AN-BC:
## Final verdict on `5b9df35e` — Claude Fable 5 / Ultracode (Amith's review session)

Re-verified the defense-in-depth commit against the code. **Every finding from both review rounds is now closed** — this PR has no open items from this review.

Confirmed at `5b9df35e`:
- **`SetDataset` freezes** (`localCacheManager.ts:1421`), placed before the awaited eviction like the other three funnels.
- **`deepFreezeCacheValue` walks via an explicit `WeakSet`** — "already frozen" no longer conflated with "already walked", so shallow-frozen input gets its children frozen too.
- **The degrade log tells the truth now**: "failed partway; storing the entry with whatever was frozen before the failure (partial protection, not none)".
- **Resolve-before-publish in *both* init paths** — `doInitialize` and both `SetStorageProvider` branches assign `_sharesReferences` before `_storageProvider`, closing the mid-probe window.
- **`ProjectRowsToFields`' fast path is gated on `!Object.isFrozen(rows)`** (`providerBase.ts:225`), so the documented "Fields callers get mutable rows" promise now holds on cache hits — with the allocation-free path kept for DB-miss input.
- **The `entity_subclasses.ts` drift is gone** (zero diff vs `next`), and the last stale "RED until fixed" markers are cleared.

Each of these ships with red-first coverage (`freezeHardening`, `projectRowsFrozenNoOp`, updated hook tests) — same discipline as the rest of the branch.

Two cosmetic leftovers, take or leave: the batch **miss**-path hook loop still *replaces* the result where the singular/hit paths *merge* via `Object.assign` (and `Unsubscribe` is attached before that loop), and the PR description still describes the superseded `_rawIsFrozen` design, IT70, and a "fails closed" probe — worth a refresh so it matches what's shipping.

On CI: the two red checks remain the inherited `next` breaks identified earlier (DOM-spec ratchet from #3467; #3553's Phase-0 migration failing `FK_EntityField_RelatedEntity` on fresh databases). Nothing on this branch is implicated — once `next` is fixed, merge it in and this should go green.

From this review's perspective the branch is **approve-ready**: the original P1 fix, the freeze architecture, and every seam the two review rounds opened are now closed and pinned by tests.

---

AN-BC:
**CI status update on `d8d7626b`** — Claude Fable 5 / Ultracode (Amith's review session)

Quick triage so nobody debugs this branch for it: the `next` merge picked up #3564, and the Phase-0 migration now **passes** — but Integration still fails, one stage later, during `mj sync push`:

```
❌ Processing failed for MJ: Entities at MJ: Entities[29]
   Lookup failed: No record found in 'MJ: Entities' where Name='MJ: Report Snapshots'
```

That's `metadata/entities` (owned by `next`) still referencing an entity the now-working Phase-0 retirement migration just deleted — `next`'s own Integration Tier run on `879bc533` (the exact state merged here) fails identically. The unit-test check is likewise still the pre-existing DOM-spec ratchet (`138 > 137`, verbatim). Both remain `next`-side; this PR touches neither `metadata/entities` nor anything under `packages/Angular/`. The approve-ready verdict above is unchanged.

---

AN-BC:
**`next` is green again — a re-merge should clear this PR's CI** — Claude Fable 5 / Ultracode (Amith's review session)

Both inherited breaks are fixed on `next`: the Integration Tier has been green on every push since 16:54 UTC (the Phase-0 `MJ: Report Snapshots` metadata inconsistency is resolved), and Unit Tests went green at `978b2586` (18:19 UTC) after the workspace-tabs DOM specs landed. This branch's head (`d8d7626b`) predates those fixes, so its red checks are stale — merging current `next` into `CB-local-cache-freeze-on-write` should bring CI fully green with no code changes needed on this side. Review status is unchanged: approve-ready.

---

AN-BC:
**Heads-up: `next` has claimed IT71 — the bundle needs one more renumber** — Claude Fable 5 / Ultracode (Amith's review session)

The PR is conflicted with `next` again in `.integration-suite.json`, and this time it isn't just a mechanical merge: `next` now ships its own **`IT71 - Task Graph Orchestration`** (plus `IT72 - Entity Graph`, also at Sequence 33), so this branch's `IT71 - Cache Immutability (freeze-on-write)` collides on the number — the same race that already forced IT69 → IT70 → IT71 here. Next free number is **IT73**: same drill as before (rename the test/JSON/checks references, keep the existing UUIDs so synced records update in place), then re-merge `next`. Review status is otherwise unchanged — CI was fully green on `a308b2ed` before `next` moved.

---

rkihm-BC (commented):
## Review — blocking items inline, the rest as follow-ups

Read the full diff at `a329997` against merge-base `71817db`, plus the three Ultracode rounds above. **The engineering here is excellent** — the P1 diagnosis is right, freeze-on-write (cost per miss) over copy-on-read (cost per hit) is the right trade for a read-dominated cache, the freeze-before-first-await ordering is load-bearing and correct, and moving the `BaseEntity` conversion memo into `_rawConverted` removes the TOCTOU class rather than narrowing it. I spot-verified the earlier rounds' main claims — the fingerprint collision, `ProjectRowsToFields`' frozen gate, the `MapFieldNamesToCodeNames` call-site sweep, freeze ordering, and index alignment in the all-cached hook loop — and they all hold up.

I also swept the repo myself for consumers this freeze would newly break (`.Results.sort/push/splice/reverse/shift/unshift/pop`, `.Results.forEach` with writes, `for…of` row assignment). Three hits, all safe: two are `ResultType: 'entity_object'` (fresh array), one is browser-only. **In-repo is clean.**

### Blocking (3 inline comments)

1. **`IT72` collides with `next`'s own IT72** — see inline on `.IT72-cache-immutability.json`.
2. **The suite-sequence renumber is stale and now creates ties plus a gap** — see inline on `.integration-suite.json:501`.
3. **Re-merge `next`** after 1 and 2. `next` moved underneath this branch again (IT72–IT76 landed, and IT23/IT25 took sequences 36/37), which is what invalidated the renumber.

Nothing in the product code is blocking. The corruption this fixes is worse than anything left open below.

---

### Follow-ups (non-blocking — happy for these to land after merge)

**4. The last hardening round traded O(1) re-entry for O(rows × fields) on the save path.** `deepFreezeCacheValue` (`localCacheManager.ts:37`) now terminates on a per-call `WeakSet` and no longer short-circuits on `Object.isFrozen`. The reason is sound — an externally *shallow*-frozen input would have had its subtree skipped — but the cost lands on `storeCachedResults`, which runs on **every save/delete event, for every cached fingerprint of that entity**, and re-walks the whole carried-forward result set: `Object.freeze` + `WeakSet.add` + an `Object.values()` allocation per row, plus every field value. That was zero pre-PR and O(new rows) in this PR's first iteration. It also sits right next to `estimateResultsSize`, which samples rows *specifically* to avoid O(n) work on this exact path.

The fix keeps both properties: a **module-level `WeakSet` of objects this function froze itself**, checked as a fast skip, with the full walk retained for anything it hasn't seen. That's what round 1 actually suggested ("keep `isFrozen` only as a fast path for nodes this function froze itself") — the implemented change dropped `isFrozen` entirely instead. Worth one benchmark on a realistically-sized slot either way: every perf number in the description and guide is hit-side.

**5. `SetStorageProvider`'s migration loop bypasses the freeze** (`localCacheManager.ts:615`). It reads each entry from the old provider and `SetItem`s it into the new one raw, so swapping *to* a reference-sharing provider republishes every migrated slot **unfrozen** — the P1 class, reopened for those slots until each is rewritten. Not reachable in-repo (only in-memory→Redis ships), but this is presented as a structural guarantee, and serializing→sharing is exactly what an embedded host or test harness does. Either route those writes through `freezeRowDataIfProviderSharesReferences`, or skip migration when the target declares `SharesReferences === false`. No test on this path in either direction.

Related, worth a line in the guide: on a Redis deployment, rows cached during the pre-swap in-memory window stay frozen forever while post-swap rows are not — two mutability contracts in one process depending on when you loaded.

**6. Doc/comment drift.** This PR's comments are unusually load-bearing, which makes the stale ones cost more than usual. All four are artifacts of the last two fix rounds:

| Where | Says | Actually |
|---|---|---|
| `localCacheManager.ts:26` | "the `isFrozen` short-circuit both skips already-frozen subtrees (cheap re-entry for in-place slot maintenance)" | Ten lines below, the code explains at length why it deliberately does **not** key off `isFrozen` |
| `localCacheManager.ts:30` | Dates are "skipped, not frozen" | Dates *are* frozen (harmlessly — internal slots are unaffected) |
| `localCacheManager.ts:483` | probe "fails closed to `false`" | Fails **open** — `false` disarms the freeze. The guide got corrected; this comment didn't |
| `localCacheManager.ts:1992` | "`BaseEntity` samples `Object.isFrozen` once at load" | That mechanism was deleted in this same PR |

Plus: the changeset and guide both say the freeze is applied at "both write funnels" and then list three — there are **four** (`SetDataset` was added in the last round); both still say a freeze failure "degrades to a logged, **unfrozen** store" while the code now correctly logs *partial* protection; **the PR description is materially stale** (describes IT71, the superseded `_rawIsFrozen` probe, and "fails closed"); and `.IT72-cache-immutability.json`'s `Description`/`ExpectedOutcomes` stop at F14 though the bundle ships F15 — arguably the most important check, since it's the case an earlier verification pass missed. That text lands in the DB record.

**7. Changeset scope.** `MapFieldNamesToCodeNames` / `ArrayMapFieldNamesToCodeNames` are `protected` members of `ResolverBase` — our documented extension point for custom resolvers — and they change from mutate-in-place to return-a-copy. I verified every in-repo caller consumes the return value (all five `UserResolver` sites, the CodeGen template, every generated resolver), so we're fine; but a downstream caller that invoked either **for the side effect** silently stops mapping, with no error, just untransformed field names on the wire. That deserves an explicit "if you subclass `ResolverBase`" line in the changeset, not only the JSDoc.

Same theme on the bump itself: the changeset does say Node-side clients get the freeze, but it's mid-bullet. This is a `patch` on `@memberjunction/core` that converts previously-tolerated in-place mutation into a runtime `TypeError` for every server-side consumer of `ResultType: 'simple'` results. I'd argue for **`minor` on `@memberjunction/core` and `@memberjunction/server`**, and putting *"if you see `TypeError: Cannot assign to read only property` on a RunView row…"* at the **top** of the release note rather than in the middle of the guide.

**8. Two small ones.** (a) `serveFromServerCache` in `GenericDatabaseProvider.ts` is still `private` — @AN-BC asked for `protected` in an inline comment above and it got lost under the bot rounds. (b) Hook-replacement semantics still diverge: the singular and both hit paths **merge** via `Object.assign`, the batch miss loop at `providerBase.ts:3173` **replaces** (`results[i] = await …`), and `Unsubscribe` is attached at `:3132` *before* that loop so a replacing hook drops the caller's only unregister handle. Small — but this PR just made "hook returns a replacement" the required pattern, so it's the wrong seam to leave inconsistent.

**Also worth a guide note (no code change):** dataset namespacing is the right fix for the collision, but the consequence is that `MJ: Entities` / `MJ: Entity Fields` are now cached **twice** — the dataset slot and the plain-RunView slot, with genuinely distinct row objects — and both are maintained on every save event, compounding #4. On a large instance that's a real number against `maxSizeBytes`.

---

Deterministic integration tier is green at 11m23s including IT72; unit tests were still running when I looked. Once the three blockers are cleared this is a merge from my side.

---
_Reviewed with [Claude Code](https://claude.com/claude-code) assistance._

metadata-optional/integration-test/tests/integration/.IT72-cache-immutability.json
```diff
@@ -0,0 +1,24 @@
+{
+  "fields": {
+    "TypeID": "@lookup:MJ: Test Types.Name=Integration Test",
+    "Name": "IT72 - Cache Immutability (freeze-on-write)",
```
rkihm-BC: 🔴 **Blocking: `IT72` is already taken on `next`.**

`next` ships `.IT72-entity-graph.json` = **"IT72 - Entity Graph (related-record collections)"**, and IT73–IT76 are gone too. This would land a second IT72. The `@lookup` is by `Name` so `mj sync push` won't fail — it'll just quietly give us two tests numbered IT72.

**Next free number is IT77.** Same drill as the IT69 → 70 → 71 → 72 renames this branch has already been through — **keep the existing UUIDs** so the synced records update in place rather than creating new ones:

- this file → `.IT77-cache-immutability.json`, `Name` → `IT77 - Cache Immutability (freeze-on-write)` (keep `E857259C-DB91-470B-84DD-EC62E08EA9AD`)
- the suite entry's `@lookup:MJ: Tests.Name=…` on line 922 of `.integration-suite.json` (keep `CB7AAFCA-…`)
- the `IT72-*` strings inside `cache-immutability.checks.ts` (the F13 fingerprint/query labels and the `IT72 F13 Probe Entity` name)

This is the fourth renumber `next` has forced on this branch — worth doing it and merging promptly rather than racing again.

metadata-optional/integration-test/test-suites/.integration-suite.json
```diff
@@ -498,7 +498,7 @@
           "fields": {
             "SuiteID": "@parent:ID",
             "TestID": "@lookup:MJ: Tests.Name=IT03 - Client GraphQL Cache Integrity",
-            "Sequence": 34,
+            "Sequence": 36,
```
rkihm-BC: 🔴 **Blocking: this renumber was correct when written, but `next` has since invalidated it.**

These two bumps (`IT03` 34→36 here, `IT15` 35→37 on line 516) made room at 35 back when 36/37 were free. `next` has since placed **IT23 at 36 and IT25 at 37**, so post-merge the deterministic suite reads:

```
34  (empty gap)
35  IT72 - Cache Immutability      ← new
36  IT03 …  +  IT23 …              ← tie
37  IT15 …  +  IT25 …              ← tie
```

**Drop both bump hunks** and give the new test the next free sequence instead (see the comment on line 922). The only ordering constraint that actually matters — the server bundle running ahead of the client-transport bundles — is satisfied by anything ≤ 33, or by simply leaving IT03/IT15/IT23/IT25 exactly where `next` put them.

Reverting these two hunks also shrinks this file's diff vs `next` to a single clean insertion, which is what round 2 got it down to before `next` moved again.

metadata-optional/integration-test/test-suites/.integration-suite.json
```diff
@@ -915,6 +915,17 @@
           "primaryKey": {
             "ID": "B7A9F916-2992-4EC0-99F1-38CD8EB80BC5"
           }
+        },
+        {
+          "fields": {
+            "SuiteID": "@parent:ID",
+            "TestID": "@lookup:MJ: Tests.Name=IT72 - Cache Immutability (freeze-on-write)",
```
rkihm-BC: 🔴 **Blocking (pairs with the two comments above): this entry needs the rename and a non-colliding sequence.**

Two changes here once the test is renumbered to IT77:

1. `TestID` → `@lookup:MJ: Tests.Name=IT77 - Cache Immutability (freeze-on-write)` — keep `CB7AAFCA-4913-4DCF-9D36-B6CD50E5BE96` as the `primaryKey` so the synced record updates in place.
2. `Sequence: 35` → pick a value that doesn't reintroduce the tie/gap described on line 501. Simplest is to slot it in with the other server bundles (`next` already has a 3-way tie at 32 and a 7-way at 33, so a tie there is the established norm and costs nothing) and leave IT03/IT15/IT23/IT25 untouched at 34–37.

Worth noting the PR description's claim that this "runs immediately after IT69" hasn't been true for a while — `next` now has seven tests at Sequence 33. Not a problem in itself, just don't let the description keep asserting an ordering the data doesn't guarantee.

PR #3430, opened 2026-08-04
fix(metadata-sync): load dynamicPackages.server so non-MJ entity subclasses register

### LINKED ISSUE #3415: mj sync pull: BizApps/BCSaaS entity subclasses not registered — causes silent data loss and duplicate records
## Summary

`mj sync pull` silently drops all but one record when pulling entities from non-`MJ:` namespaces (e.g., `MJ_BizApps_Common: Organizations`, `BC: AI Configurations`). Each pull run saves only one record instead of all records, and every subsequent pull appends a duplicate of that lone record rather than updating it.

## Root Cause

`@memberjunction/metadata-sync` bootstraps via:

```js
import '@memberjunction/server-bootstrap-lite/mj-class-registrations';
```

This manifest only covers `@memberjunction/*` packages and registers only `MJ: *`-prefixed entity subclasses. It does **not** load the `dynamicPackages.server` entries declared in `mj.config.cjs` — in our case:

```js
server: [
  { PackageName: '@mj-biz-apps/common-server',          StartupExport: 'LoadBizAppsCommonServer' },
  { PackageName: '@bluecypress/bcsaas-server-bootstrap', StartupExport: 'LoadBCSaaSServer' },
]
```

Those two packages register the typed entity subclasses for `MJ_BizApps_Common: *` and `BC: *` entities, each with field getters like:

```js
get ID() { return this.Get('ID'); }
```

Without them, `ClassFactory.CreateInstance(BaseEntity, 'MJ_BizApps_Common: Organizations')` falls back to bare `BaseEntity`, which has no such getters.

## Bug Cascade

**1. RunView with `ResultType: 'entity_object'` falls back to bare BaseEntity**

ClassFactory logs a warning and falls back. The returned entity object has no typed field getters.

**2. Primary key extraction returns `undefined`**

In `PullService.processIndividualRecords`:

```js
const primaryKey = {};
for (const pk of entityInfo.PrimaryKeys) {
    primaryKey[pk.Name] = record[pk.Name]; // record['ID'] === undefined on bare BaseEntity
}
```

Result: `primaryKey = { ID: undefined }` for **every** record.

**3. All records share the same lookup key**

```js
createPrimaryKeyLookup({ ID: undefined }) // → "ID:undefined"
```

Every DB record produces the identical key.

**4. FileWriteBatch overwrites all-but-one record**

In `FileWriteBatch.applyChanges()`, the `update-array` path:

```js
const index = contentArray.findIndex(r =>
    createPrimaryKeyLookup(r.primaryKey || {}) === change.primaryKeyLookup // "ID:undefined"
);
if (index >= 0) contentArray[index] = change.data; // ← overwrites previous winner
else contentArray.push(change.data);
```

Records are processed in parallel via `Promise.all`. The first record is appended; every subsequent one finds the previous entry (same key) and **overwrites** it. Only the last async operation to resolve survives — non-deterministically.

**5. Written file has `primaryKey: {}`**

`JSON.stringify({ ID: undefined })` → `{}`. The surviving record is written with an empty `primaryKey`, so on the next pull `loadExistingRecords` can't match it and appends a duplicate.

## Symptoms

- `mj sync pull` on a non-`MJ:`-namespaced entity appears to succeed but only writes **one record** per run regardless of how many exist in the DB.
- Which record "wins" is a race condition — specific records may never appear across multiple pulls.
- Each subsequent pull **duplicates** the previously written record (since `primaryKey: {}` never matches anything in the existing file).
- ClassFactory logs visible warnings: `no registration found for base class 'BaseEntity' with key 'MJ_BizApps_Common: Organizations'`.

## Reproducer

Configure a `.mj-sync.json` targeting `MJ_BizApps_Common: Organizations` (or any non-`MJ:` entity) and run `mj sync pull`. With 10+ records in the DB, only 1 appears in the output file. Run again — a duplicate appears.

## Proposed Fix

`@memberjunction/metadata-sync`'s provider initialization (in `packages/MetadataSync/src/lib/provider-utils.ts`) should load `dynamicPackages.server` from `mj.config.cjs` before processing records, the same way `ServerBootstrap` does (`packages/ServerBootstrap/src/index.ts`).

Concretely, after setting up the DB provider, iterate `config.dynamicPackages?.server`, dynamically import each enabled package, and invoke its `StartupExport` function — analogous to `loadDynamicAppPackages()` in ServerBootstrap.

This would register BizApps and BCSaaS entity subclasses so that `record['ID']` returns the correct value and primary key extraction works for all entity namespaces.

## Environment

- `@memberjunction/metadata-sync` v5.51.0
- The `server-bootstrap-lite` manifest was introduced as a CLI startup optimization (commit `3e53843793`) — before that, a broader bootstrap was used that appears to have loaded these packages.

## Workaround

Manually add missing records to the JSON file (bypassing the pull) until a fix is released.

cadam11:
Fixes #3415 — silent data loss and duplicate records in `mj sync pull` for non-`MJ:` namespaces (BizApps/BCSaaS).

## Root cause

`metadata-sync` bootstraps only the `@memberjunction/*` class manifest and never loaded `dynamicPackages.server` from `mj.config.cjs`. Entities like `MJ_BizApps_Common: Organizations` therefore fell back to bare `BaseEntity`, whose typed PK getters don't exist → primary-key extraction returned `undefined` → every record keyed as `"ID:undefined"` → the write batch collapsed them onto one non-deterministic survivor saved with `primaryKey: {}` → each later pull appended a duplicate. Full verification with file:line receipts is on the issue.

Latent since at least v5.48 — **not** a startup-mode regression: this is `@RegisterClass` entity registration, not engine pre-warm, so the loader runs in both `full` and `task` modes.

## The fix

Mirror `ServerBootstrap.loadDynamicAppPackages` (minus GraphQL resolver-path collection, which has no meaning in a CLI process) in metadata-sync's provider initialization, on **both** DB-platform paths: after provider init, dynamically import each enabled `dynamicPackages.server` entry and invoke its `StartupExport`. Robustness contract preserved: no-op when absent, per-package try/catch, `ERR_MODULE_NOT_FOUND` tolerated with a hint, anything else warns and continues, never crashes the CLI. App registrations load after the manifest, so they win ClassFactory priority — same ordering as the server.

Also adds a typed `dynamicPackages` field to `MJConfig` (previously reachable only through the index signature) with the shared entry shape.

## Tests

7 new tests (`dynamic-server-packages.test.ts`, fixture-based — the fixture's file URL plays the PackageName role): absent/empty no-op · import + StartupExport invocation · import without kicker · `Enabled: false` skip · not-installed tolerance · throwing StartupExport warns-not-throws · continues past a failing entry. Package suite: **268/268 green**. Dependency-chain build: 121/121.

## Reviewer notes

- `mj sync push` was verified data-safe (metadata-driven `.Get()`/`.Set()`) — no push changes needed. It will still faithfully propagate files a *broken* pull already wrote; recovering those files is manual.
- The same gap exists in the other lite-manifest CLI consumers (MJCLI, CodeGenLib, MJCodeGenAPI, MCPServer, A2AServer) — follow-up issue coming rather than one six-package PR.
- Line triage (does this ship as a 5.51.x patch?) goes through the LTS process separately; the changeset here is a normal Edge patch.

https://claude.ai/code/session_019EfGkq8cW9drueo1vEN4sM

### DIFF
--- a/.changeset/metadata-sync-dynamic-packages.md
+++ b/.changeset/metadata-sync-dynamic-packages.md
@@ -1,0 +1,5 @@
+---
+"@memberjunction/metadata-sync": patch
+---
+
+Load dynamicPackages.server (Open App server packages) during provider initialization so non-MJ-namespace entity subclasses are registered before sync operations run. Fixes silent data loss in `mj sync pull` for BizApps/BCSaaS entities (#3415): bare BaseEntity fallback made primary-key reads return undefined, collapsing every record onto one key and duplicating it on later pulls.
--- a/packages/MetadataSync/src/__tests__/dynamic-server-packages.test.ts
+++ b/packages/MetadataSync/src/__tests__/dynamic-server-packages.test.ts
@@ -1,0 +1,70 @@
+/**
+ * Tests for loadDynamicServerPackages (#3415): dynamicPackages.server entries from
+ * mj.config.cjs must be imported (registering non-MJ-namespace entity subclasses)
+ * before sync operations instantiate entity objects. The fixture stands in for a
+ * real Open App server package; its file URL plays the PackageName role since
+ * import() accepts both bare specifiers and URLs.
+ */
+import { describe, expect, it, beforeEach, vi } from 'vitest';
+import { loadDynamicServerPackages } from '../lib/provider-utils';
+
+const FIXTURE = new URL('./fixtures/fake-server-pkg.mjs', import.meta.url).href;
+
+type GlobalWithCounters = typeof globalThis & {
+  __fakeServerPkgImports?: number;
+  __fakeServerPkgKickerRuns?: number;
+};
+const g = globalThis as GlobalWithCounters;
+
+describe('loadDynamicServerPackages', () => {
+  beforeEach(() => {
+    g.__fakeServerPkgKickerRuns = 0;
+  });
+
+  it('is a no-op when the section is absent or empty', async () => {
+    await expect(loadDynamicServerPackages(undefined)).resolves.toBeUndefined();
+    await expect(loadDynamicServerPackages([])).resolves.toBeUndefined();
+  });
+
+  it('imports the package and invokes its StartupExport', async () => {
+    await loadDynamicServerPackages([{ PackageName: FIXTURE, StartupExport: 'LoadFakeServer' }]);
+    expect(g.__fakeServerPkgImports).toBeGreaterThanOrEqual(1);
+    expect(g.__fakeServerPkgKickerRuns).toBe(1);
+  });
+
+  it('imports without invoking anything when no StartupExport is declared', async () => {
+    await loadDynamicServerPackages([{ PackageName: FIXTURE }]);
+    expect(g.__fakeServerPkgKickerRuns).toBe(0);
+  });
+
+  it('skips entries with Enabled: false', async () => {
+    await loadDynamicServerPackages([{ PackageName: FIXTURE, StartupExport: 'LoadFakeServer', Enabled: false }]);
+    expect(g.__fakeServerPkgKickerRuns).toBe(0);
+  });
+
+  it('tolerates a package that is not installed (ERR_MODULE_NOT_FOUND)', async () => {
+    await expect(
+      loadDynamicServerPackages([{ PackageName: '@mj-test/definitely-not-installed' }])
+    ).resolves.toBeUndefined();
+  });
+
+  it('warns but does not throw when a StartupExport throws', async () => {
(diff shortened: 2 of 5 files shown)


cadam11:
Superseded by the #4199 work merged to next on 2026-09-03 (commit 41b0d28215, `@memberjunction/dynamic-packages`). The MJ CLI prerun hook now loads `dynamicPackages.server` for every heavy command, including `mj sync pull` and `mj sync push`, so the fix here would add a second redundant loader on the MetadataSync path. Closing without merge.

PR #3434, opened 2026-08-04
fix(ng-file-storage): pin ng-notifications to the workspace version

cadam11:
**One sentence:** `ng-file-storage` pins `@memberjunction/ng-notifications` at `5.51.0` while the workspace package is `6.0.0`, so npm resolves the *published* copy from the registry instead of linking the local sibling — and drags in a nested 36-package 5.51.0 dependency tree with it.

> **Rebased and rescoped.** This PR originally restored the `ng-shared` dependency to unbreak the `next` build. That landed separately (`fcc9eed56d`), and the layering red it caused was fixed on `next` by reclassifying the package to `surface` (`c9b7423c3f`). Both are gone from this PR. What remains is the one fix that never landed.

## The bug

```diff
-    "@memberjunction/ng-notifications": "5.51.0",
+    "@memberjunction/ng-notifications": "6.0.0",
```

That's the whole source change. It looks cosmetic and isn't.

The **top-level** `node_modules/@memberjunction/ng-notifications` entry is a workspace link, which is what makes this easy to miss. But because the spec doesn't match the workspace version, npm gave file-storage its own nested tree:

```
packages/Angular/Generic/file-storage/node_modules/
  @memberjunction/actions-base   5.51.0   (registry)
  @memberjunction/core           5.51.0   (registry)
  @memberjunction/global         5.51.0   (registry)
  ... 36 packages total
```

**This one package has been compiling against a published 5.51.0 dependency tree while every other package in the repo used the 6.0.0 workspace siblings.** Correcting the pin drops all 36 nested entries — that's the 460-line lockfile shrink.

## For the reviewer

- **The lockfile diff is −460/+1 and that is the point**, not collateral damage. Every removed entry is under `packages/Angular/Generic/file-storage/node_modules/`. Worth spot-checking that claim — nothing outside that prefix should be touched.
- **The previous revision moved the dependency line as well as changing it**, which is what made this conflict with `next`. The line now changes in place and stays alphabetical.
- **Origin of the pin:** a pre-6.0 branch state carried forward by the UI-layering sweep, not a deliberate version choice.

## Verification

- `turbo build --filter=@memberjunction/ng-file-storage`: **27/27 successful** against the 6.0.0 siblings
- Merges cleanly into `next` (verified with `git merge-tree`)
- Lockfile regenerated by `npm install` at the repo root, not hand-edited

https://claude.ai/code/session_01Mn3NY1z42C27NgyPEf491p


AN-BC:
@cadam11 was attempting to fix the issue now in that earlier resolution. Would be ideal.

PR #3440, opened 2026-08-04
fix(explorer): decouple session landing app from user-sortable Sequence order

MattC-BC:
## Summary

`UserApplication.Sequence` is a user-owned display preference (switcher/card order), but the shell's bare-root landing was blindly activating `apps[0]` from the Sequence-ordered list. Consequences: dragging any app above Home — or landing in a Sequence-0 tie, which is reachable without ever touching the ordering UI — silently changed where every fresh session (including magic links) opened, and if that app failed to produce a tab there was no fallback: the session was stranded on the loading screen. This PR separates the two roles and adds a failure chain.

- **Landing rule** — new `ApplicationManager.GetDefaultLandingApp()`: the user's active, access-granted app with the lowest `Application.DefaultSequence` (Home ships at `-1`; not hardcoded by name, so orgs that replace Home still get a deterministic landing). Degrades to the first app in the user's order when nothing is flagged below the schema default. The shell's bare-root path uses this instead of `apps[0]`.
- **Deterministic ordering** — `UserInfoEngine.UserApplications` / `GetUserApplicationsForUser` now share one comparator: `Sequence` → `Application.DefaultSequence` → name. Sequence-0 ties no longer resolve purely alphabetically (the "Accounting beats Home" case).
- **Failure fallback chain** — the bare-root path validates each candidate's `CreateDefaultTab()` BEFORE activating it and falls through (declared-default → next app in order → terminal dialog) instead of handing the session to a dead app. The dialog-dismiss fallback paths (`redirectToFallbackApp`, formerly `redirectToFirstApp`) use the same candidate order. Complements #3336, which unsticks the loading screen but leaves the session on the broken app.
- **Nav-item asymmetry** — `CreateDefaultTab()` now honors the `isDefault` nav item (`find(isDefault) || [0]`), matching what explicit app switches and deep links already do, so landing on an app opens the same tab as clicking it.

Unchanged by design: saved-workspace restore still wins at bare root, scoped magic links still short-circuit to their app, deep links still win, and Sequence remains freely reorderable — now with zero landing side effects. Explicitly NOT the fix: pinning Home to Sequence 0.

## Test plan
- [x] Unit: new `UserInfoEngine.applicationOrdering.test.ts` (+4, incl. the Sequence-0 tie regression) — MJCoreEntities 573 passed
- [x] Unit: `CreateDefaultTab` isDefault resolution + `GetDefaultLandingApp` (+5) — base-application 24 passed; explorer-core 113 passed
- [x] Integration: deterministic tier 50/53 — the 3 failures are pre-existing local DB drift (IT50 codegen-artifact consistency, IT52 search seams), unrelated to these packages
- [x] Live: fresh session at `/` lands on Home with a non-Home app dragged to top of app config
- [ ] Live: unscoped magic link lands on Home; scoped magic link still opens its scoped app
- [x] Live: break the landing app's `DefaultNavItems` → session falls through to a working app instead of hanging
- [ ] Live: reordering apps in config changes switcher order only, never the landing app


MattC-BC:
@rkihm-BC - this is ready to be merged into next. The long and short of it is that this fixes the app nav reordering functionality so that if a user puts another app before home, root doesn't point to any other page other than the home screen while still allowing the other app to be ordered first within the app UI.

rkihm-BC:
Reviewed the full diff, tests, and comments — solid, well-targeted fix. Correct root-cause diagnosis (not a symptom patch), the display-vs-landing separation is the right call, the fail-safe `CreateDefaultTab()`-before-activate fall-through is the highlight, and I verified the `isDefault` asymmetry claim against the explicit-switch paths (`shell.component.ts:1045`/`:2000`). No regressions found.

One gap before merge is **verification, not correctness**: the automated coverage is unit-only and heavily mocked, so the seam this PR exists to protect (`UserInfoEngine` ordering → `ApplicationManager` → shell activation) isn't exercised end-to-end, and all four live boxes in the test plan are still unchecked. Could you confirm the two highest-value scenarios manually before we merge?

- [ ] **Fresh session at `/` lands on Home** with a non-Home app dragged to the top of the app config (the core guarantee)
- [ ] **Break the landing app's `DefaultNavItems` → session falls through** to a working app instead of hanging the loading screen (the fail-safe path)

The other two live boxes (magic-link scoping, reorder-changes-switcher-only) and an integration-tier bundle can reasonably be fast-follow.

### Optional polish (non-blocking)
1. **Defensive-default asymmetry.** `UserInfoEngine.applicationDefaultSequence` guards a missing app with `?? 100`, but `GetDefaultLandingApp`'s `reduce` relies on `BaseApplication.DefaultSequence` being set — if `appInfo.DefaultSequence` were ever `undefined`, `Object.assign` overwrites the class default and the `<` comparison goes `NaN`. Can't happen today (schema non-null), but a `?? 100` / `Number.isFinite` guard in the reduce would make the two sides symmetric.
2. **Double `CreateDefaultTab()` on the landing path.** `openLandingApp()` validates via `CreateDefaultTab()`, then `SetActiveApp()` fires the `ActiveApp` subscription which calls it again to open the tab. Idempotent today, so just a one-line comment noting the validate-then-rebuild would help future readers (and flags the trap if `CreateDefaultTab` ever gains side effects).
3. **Test-realism note.** The "keeps the earlier app in Sequence order when DefaultSequence ties" test sets `applications$` directly to `[Bravo, Alpha]`, bypassing the comparator — in production `applications$` is comparator-sorted, so that scenario surfaces as `[Alpha, Bravo]` and lands on Alpha. The test asserts `reduce` stability correctly but could mislead a reader into thinking prod returns Bravo; a clarifying comment would close the gap.

rkihm-BC:
Re-reviewed `4152655` — all three optional polish items landed cleanly and accurately:

1. **NaN-symmetry guard** — `GetDefaultLandingApp` now wraps `DefaultSequence` in `Number.isFinite(...) ? ... : 100`, mirroring `UserInfoEngine.applicationDefaultSequence`'s `?? 100`. The reduce can no longer go order-dependent on an `undefined` field, and the two sides are symmetric. Good call making it real code rather than just a comment. ✅
2. **Double-`CreateDefaultTab` note** — the validate-then-rebuild comment on `openLandingApp` is exactly right, including the explicit trap-if-side-effects warning. ✅
3. **Tie-test clarity** — the comment now states `[Bravo, Alpha]` bypasses the comparator and production would land Alpha; the assertion is about reduce tie-stability. Removes the misread I flagged. ✅

Verified by inspection (no test assertions changed — both existing cases still hold since -1/100 are finite; didn't run the suite locally to avoid disturbing an unrelated worktree, happy to if you'd like a green tick on the record). Nothing new introduced. From my side the code is in good shape — the only open item remains the manual live-verification of the two highest-value scenarios (Home-lands-at-root with a non-Home app dragged up; broken-landing-app fall-through) before merge.

rkihm-BC (approved):
Approving. ✅

This is a clean, well-targeted fix: correct root-cause diagnosis (decoupling the user-owned `Sequence` display preference from where a session lands, rather than the tempting shortcut of pinning Home to Sequence 0), the fail-safe `CreateDefaultTab()`-before-activate fall-through removes a real dead-end, and the `isDefault` asymmetry fix makes landing consistent with explicit switches. Strong typing throughout, one shared comparator, complete rename, and excellent explanatory comments at every non-obvious site.

All three review follow-ups (`4152655`) landed accurately — the `Number.isFinite` NaN-symmetry guard, the double-`CreateDefaultTab` note, and the tie-test clarification. Nothing outstanding on the code.

One non-blocking recommendation carried over from review: before merge, please tick the two highest-value manual scenarios, since no automated test covers them —
- fresh session at `/` lands on Home with a non-Home app dragged to the top of app config
- breaking the landing app's `DefaultNavItems` falls through to a working app instead of hanging

Nice work.

MarceloT-BC:
@MattC-BC The PR says landing "degrades to the first app in the user's order when nothing is flagged below the schema default." In practice that branch never runs — shipped metadata has Chat 1, Data Explorer 2, Lists 50, all below 100. So any user without Home lands on Chat, not on their own first app. That case is silently unhandled, not degraded.

Underneath it: Application.DefaultSequence is documented as a seed for UserApplication.Sequence (a display-order default), and this PR makes it the landing authority without changing that. So an admin retuning DefaultSequence for switcher order now moves everyone's landing app — the same coupling the PR removes, one level up.

Everything else I found is either already hedged in the PR body or minor.

PR #3441, opened 2026-08-04
WS1: Update RLS post-image check + WS2: tenant injection/session-isolation hardening

jordanfanapour:
## Summary

Phase 1 of the API-key-scoped row filters plan (WS1 + WS2 — see `plans/api-key-row-filters.md` on open PR #3409). Two independently-valuable bugfixes, shipped ahead of the feature itself. WS3 (API-key row filters) is not part of this PR.

- **WS1 — Update RLS now validates the post-image, not just the pre-image.** `DatabaseProviderBase.Save()` previously only checked Update RLS against the row as it exists in the DB before hooks ran, so an update could move a row out of the caller's filter (e.g. reassign `OrganizationID` to an org the caller doesn't belong to) and still succeed. A new `CheckUpdateRLSPostImage` gate now runs after `OnBeforeSaveExecute`, reusing `CheckCreateRLS`'s synthetic-row technique against the entity's pending values. `CheckCreateRLS` now also runs a second time, post-hook, closing the analogous gap for new records — additive, the original pre-hook/pre-image checks and their messages are unchanged.
- **WS1 — generalized, safer RLS projection builder.** `BuildRLSProjections` replaces `BuildCreateRLSProjections`: emits typed `CAST(NULL AS <type>)` for every nulled field in update mode (fails the save on an unresolvable type rather than emitting an untyped NULL), and validates numeric field values before interpolating them (fixes the previous raw `String(val)` branch — finding M5).
- **WS1 — cost-control skip, fail-safe by construction.** A new SQL-AST decomposition parser (`DecomposeSimpleAndClause`, via `@memberjunction/sql-parser`) skips the post-image check only when the resolved filter fully decomposes into simple `Column <op> <literal>` AND-terms with nothing dirty; anything it can't fully prove decomposed (OR, functions, subqueries, no dialect wired) always runs the check.
- **WS2 — multi-tenancy injection fix (opt-in feature, default off).** `attachTenantContext` now validates every tenant id against a bounded identifier pattern regardless of source and throws rather than degrading to an unscoped session; `createTenantMiddleware` returns 400 on a malformed or repeated header. The tenant predicate is escaped, resolves the column against real entity metadata, and quotes via the provider's `QuoteIdentifier` instead of a hardcoded SQL Server bracket (fixes a latent PostgreSQL incompatibility as a side effect).
- **WS2 — session isolation.** `createTenantMiddleware` previously mutated `userPayload.userRecord` in place, which may be the shared `UserCache` instance for both JWT and API-key sessions. A new shared `cloneUserInfoForSession` helper (also folded into the existing magic-link session-scoping path) clones to a fresh `UserInfo` before stamping any per-request state.

**Behavior-change note (WS1):** a save that reassigns a filter-referenced column away from a value the caller's Update RLS filter permits will now be rejected, where it previously succeeded. This is the intended fix, not a regression — see the changeset for the full rationale and the recommended remediation (narrow the filter, don't grant RLS exemption).

No schema changes, no CodeGen in this phase. Does not touch PR #3409's branch or staged migration.

## Test plan

- [x] Unit tests green: `@memberjunction/core` (1610/1610), `@memberjunction/generic-database-provider` (860/865, 5 pre-existing skips), `@memberjunction/server` (775/831, 56 pre-existing skips), `@memberjunction/sqlserver-dataprovider` (82/82), `@memberjunction/postgresql-dataprovider` (149/149)
- [x] Real end-to-end regression test: a before-save hook that mutates a filter-referenced field is observed by the post-image check through the actual (non-mocked) `Save()` orchestration — not just asserted via mocked call ordering
- [x] Transaction-group regression test: the post-image check runs and can block a save *before* the item reaches `TransactionGroup.AddTransaction`
- [x] Composite-PK entities flow through the new projection builder correctly
- [x] Deterministic integration tier: `npm run test:integration` — 54/54 passed, including IT06 (RLS Multi-User Cache Isolation), IT27 (Core Entity Write-Side), IT28 (Permission Engine), IT64 (View Security — two-identity RLS)
- [x] Rewrote `multiTenancy.security.test.ts` / `multiTenancy.test.ts`, which previously *documented* the injection/mutation defects as accepted behavior, to assert the fixed behavior instead
- [x] `tsc --noEmit` clean across all touched packages


AN-BC (approved):
LGTM, if testing went well merge away

jordanfanapour:
Closing — superseded by #3409, which now ships as the single end-to-end WS1+WS2+WS3 PR per Amith's direction. This branch's WS1/WS2 work served as the reference implementation for reviewing #3409: it caught a real gap in #3409's WS1 (the create-path post-hook RLS check was missing entirely) and three smaller WS2 defense-in-depth items (repeated-header handling, fail-closed identifier quoting, a concurrency test), all of which have now been ported into #3409. Branch (`ws1-ws2-rls-tenant-security`) left intact for reference.

PR #3444, opened 2026-08-04
feat(explorer): records-style record opens, origin crumbs, mobile UX

MattC-BC:
## Summary

Records become their own surface in MJ Explorer (the 6.0 navigation headline):

- **Records-style record opens** (`Shell.RecordOpen.Style`, default `records`): opening a record keeps you in the app you're standing in; records live in a dedicated golden-layout region with native tabs (close/drag-to-split/pin) and never pollute the main nav tab bar. A persistent count-badged **Records pill** makes any open record at most 2 clicks away from anywhere. `classic` remains as a config escape hatch.
- **Origin breadcrumbs**: every record pane opens with a two-segment crumb ("← App › Page") capturing *where you came from* at navigation time — page segment restores the exact source page (query params included), app segment goes to the app landing. Automatic for any app using standard MJ navigation; record-from-record opens crumb to the parent record; deep links/history re-opens show no crumb rather than a wrong one.
- **Move to Workspace**: records can be deliberately docked into the main layout (and moved back) for mixed compositions — power-user preservation of the old behavior, per feedback.
- **Golden-layout tab redesign** (all layouts): entity type icon in app color with hover→⋯ menu affordance, flex tab interior, token-based GL theme replacing stock PNGs, inter-tab pipes, always-visible close, WCAG 2.5.8 24px targets + full menu keyboarding.
- **Scrolling tab strip**: tabs keep natural width and the strip scrolls (wheel/trackpad + edge nudge arrows with fades + activation scroll-into-view) — replaces GL's overflow dropdown, which was unusable with many tabs.
- **Mobile records UX** (≤768px): the strip gives way to a record bar + bottom-sheet switcher listing every open record with origin subtitles; splits flatten at render *without* touching the persisted layout (desktop splits survive phone sessions); breakpoint crossings rebuild safely.
- **New shared primitives**: `mj-bottom-sheet` (ui-components), `ExplorerBreakpointService` (canonical 768px shell breakpoint), `FlattenLayoutToSingleStack`, `RecordSwitcherService`, pane-level `RecordOriginCrumbComponent`.

## Deployment-facing behavior changes (release notes)

- **Records style is ON by default** — deployments opt out via instance config `Shell.RecordOpen.Style = 'classic'`. Config-load failure fails *open* to records.
- The GL tab redesign + token theme apply in **classic mode too** (unconditional visual change).
- One-time saved-main-layout reset on first boot when an existing workspace contains record tabs (they re-home to the records region).
- `mj-slide-panel`'s state class renamed `visible` → `sp-open` (collision with a global utility); bespoke AI detail panels migrated onto `mj-slide-panel`.
- Font Awesome is now a hard dependency of the GL theme (glyphs replace stock PNGs).
- Workspaces round-trip across app versions; a mixed-version fleet will re-run the one-time reset per client until upgraded.

## Verification

- Unit suites (all green): ng-shared 75, ng-base-application 33, ng-explorer-core 148, ng-ui-components 364.
- Integration tier: 50/53 — the 3 failures are pre-existing environment drift (dev DB behind migrations; a search-config issue), none in paths this branch touches.
- Two independent multi-agent audits (4-lens on the desktop work, 3-lens on mobile) with all findings remediated; key invariants unit-tested (flatten clone/count, close-guard ordering, persist suppression).
- Live-verified via Playwright + hands-on: crumb round-trips (incl. promote/demote and cross-rename edge cases), mobile bar/sheet at 390px, desktop split survival across breakpoint crossings, drag-reorder from a scrolled strip.
- `check:ui` design-token/button gates clean; changeset included (patch ×4 packages).

## Test plan

- [ ] Open records from several apps — each stays in its source app, appears in the records strip, and its crumb returns to the exact source page
- [ ] Records pill resumes last-viewed record from any nav page; strip reaches the rest
- [ ] Move to Workspace / Move to Records round-trips a record without losing form state
- [ ] With 10+ records: strip scrolls (wheel + nudge arrows), drag-reorder and drag-to-split still land correctly
- [ ] Mobile width (≤768px): record bar + bottom sheet switch/close records; drawer Records pill opens the sheet; desktop splits intact after a phone session
- [ ] `Shell.RecordOpen.Style = 'classic'` restores legacy record-open behavior (with the new tab visuals)
- [ ] Dark mode: tabs, crumb, sheet, nudge arrows all legible

cc @BarnattW-BC for the DOM-level component tests.


MattC-BC:
@rkihm-BC  - this is the PR based on the demo we just had.

rkihm-BC:
Reviewed the full branch (core logic directly + four parallel deep-dives across the layout/state core, the tab-container, the navigation/crumb/config layer, and the bottom-sheet/theme layer). High-craft work overall — `layout-transforms.ts` and its tests are exemplary, the `slide-panel` `visible → sp-open` rename is complete and clean (all four AI detail panels genuinely migrated, dead CSS removed, stale test deleted), and the GL theme is design-token clean. Full notes are extensive; here are the three items I'd want fixed before this lands as a 6.0 headline.

### 🔴 1. The `classic` escape hatch fails open in the wrong direction on any casing/whitespace variance
`shell.component.ts:213`:
```ts
this.resolvedRecordOpenStyle = raw === 'classic' ? 'classic' : 'records';
```
This strict, case-sensitive compare is the single opt-out for a default-behavior flip. `InstanceConfigEngine.GetBoolean` parses `'true'` case-insensitively, so an admin will reasonably type `Classic`, `CLASSIC`, or `classic ` (trailing space) — and every one silently fails open to `records`. The operator asks for the legacy escape hatch and gets the new behavior anyway. Fix: `raw?.trim().toLowerCase() === 'classic'`. (The fail-open on a failed config *load* is correctly implemented and well-ordered — this is only the value comparison.)

### 🔴 2. `mj-bottom-sheet` (new shared Generic-layer primitive) is a `role="dialog"` with no `aria-modal`, no focus trap, and no body scroll-lock — and the PR claims scroll-lock it doesn't have
`bottom-sheet.component` has `role="dialog"` + `aria-label` + focus-on-open + focus-restore + ESC + scrim-click (all good), but: no `aria-modal="true"`; Tab walks out into the background behind the scrim; and there is no `body`/`documentElement` overflow manipulation anywhere in the component, so background content scrolls behind the open sheet. Because this is a **new shared primitive** other teams will adopt, the a11y baseline matters more than for a one-off. Either add `aria-modal` + a focus trap + scroll-lock, or at minimum add `aria-modal` + scroll-lock and strike the scroll-lock claim from the PR description. (Lifecycle/timer cleanup in this component is excellent — not a concern.)

### 🔴 3. The "flatten deep-clones so persisted splits survive" invariant only protects the mobile path; desktop restore still shares references with persisted config
`layout-transforms.ts` correctly `structuredClone`s, but the desktop records restore goes through `sanitizeLayoutNode`, which does a shallow `{...node}` and copies `componentState` **by reference** into state Golden Layout then mutates at runtime (`UpdateTabStyle` `Object.assign`s into container state). GL is handed live references into `config.recordsLayout`. It may be tolerated in practice (GL re-serializes on `stateChanged`), but it's undocumented and contradicts the invariant the PR leans on. Either clone in `sanitizeLayoutNode` or document precisely why desktop is safe. There's also no automated test for the headline "desktop splits survive a mobile session" suppression invariant — it currently lives only in untested `tab-container.component.ts`.

Happy to share the fuller set of 🟡 items (query-param merge-vs-replace asymmetry in the crumb fallback path, an untracked `configSub` at `tab-container.component.ts:777`, `SafeDetectChanges` swallowing all errors, and two functions ~200 lines with duplication) if useful — but the three above are the ones I'd gate on.

MattC-BC:
@rkihm-BC 

All three addressed in 3e3a92c214:

1. Classic escape hatch - now raw?.trim().toLowerCase() === 'classic', so any casing/whitespace variant opts out correctly.

2. Bottom sheet - went with the full version since it's a shared primitive: aria-modal="true", real focus trap (Tab/Shift+Tab cycle inside the sheet), and body scroll-lock that restores the exact prior overflow value, including destroy-while-open. 5 new tests.

3. Desktop restore refs - you were right, so instead of patching the shallow copy I moved the whole sanitize pass into layout-transforms.ts as a pure SanitizeLayoutNodeForLoad with componentState deep-cloned. GL now never holds references into persisted config on any path, and that invariant is unit-tested directly (simulated runtime Object.assign never reaches the persisted tree). Bonus: the old method's as any casts are gone.

Suites: 75 / 37 / 148 / 369, token gates clean. And yes, please share the yellow list - the untracked configSub and the SafeDetectChanges error-swallowing sound worth fixing before merge.

rkihm-BC:
Re-reviewed against `3e3a92c2` / current HEAD — **all three 🔴 items verified fixed**, and cleanly:

1. **Classic escape hatch** — `shell.component.ts:216` now `raw?.trim().toLowerCase() === 'classic'`. Any casing/whitespace variant opts out correctly. ✅
2. **Bottom sheet** — `aria-modal="true"` (html:13), real focus trap cycling Tab/Shift+Tab within the sheet, and body scroll-lock that snapshots + restores the exact prior `overflow` (incl. destroy-while-open). Went the full route since it's a shared primitive — right call. ✅
3. **Desktop restore refs** — nicely done: rather than patching the shallow copy you promoted the whole pass into `layout-transforms.ts` as a pure `SanitizeLayoutNodeForLoad` with `componentState` deep-cloned via `structuredClone`, GL manager now imports it (`golden-layout-manager.ts:882`), the invariant is unit-tested directly, and the old `as any` casts are gone. GL no longer holds references into persisted config on any path. ✅

---

### 🟡 The yellow list (you flagged the first two as merge-blockers — agree)

**A. Untracked `configSub` — real use-after-destroy gap** — `tab-container.component.ts:777`
```ts
const configSub = this.workspaceManager.Configuration.subscribe(loadedConfig => {
  if (loadedConfig) { configSub.unsubscribe(); /* … initializeGoldenLayout … */ }
```
It self-unsubscribes on the first non-null config, but it's **not** pushed to `this.subscriptions`. If the component is destroyed while config is still null (config never loads / slow load + fast nav-away), the subscription outlives the component and its callback runs `initializeGoldenLayout` on a dead component. Fix: `this.subscriptions.push(configSub)` (keeping the self-unsubscribe), or gate the callback body on a `private destroyed` flag set in `ngOnDestroy`. While there: the deferred `setTimeout(…, 0)` bodies (breakpoint rebuild, move-to-records, deferred inits) have no destroyed-guard either — the same `destroyed` flag checked in those bodies closes the whole class.

**B. `SafeDetectChanges` swallows *all* errors** — `safe-detect-changes.ts`
```ts
try { cdr.detectChanges(); } catch { /* re-entrant CD */ }
```
`markForCheck()` already covers the re-entrancy case, so the bare `catch` is a broad net: a genuine template-binding error thrown during `detectChanges` now vanishes silently — exactly the bug class that's hardest to trace later. Narrow it, e.g. only swallow when `err?.message` matches the known "already in progress"/re-entrancy signature and rethrow (or `console.error`) otherwise; at minimum `console.debug(err)` in dev so it's not invisible.

**C. `ReturnToRecordSource` — merge-vs-replace asymmetry between its two return paths** — `navigation.service.ts:707`
The exact-tab-still-open path nulls out drifted keys before restoring (REPLACE semantics, ~line 758-764), but the fallback path (source tab closed → re-resolve via `SwitchToApp(..., sourceQueryParams)`) routes through `applyQueryParamsToTab`, which **merges** (`:1400` — only nulls removed, existing keys survive). So "back to where you came from" can land on a hybrid state when the source tab was closed, which the method's own doc-comment ("the page as it was") promises it won't. Not a crash — a fidelity gap between the two paths.

**D. `resolveSourceContext` drops three advertised context fields** — `navigation.service.ts:590-606`
The `options.recordSource` branch copies `sourceAppId/AppName/NavLabel/TabId/Label/QueryParams` but **not** `sourceNavItemName`, `sourceRecordEntity`, or `sourceRecordId` — which `RecordSourceContext` advertises as caller-settable. An overlay/dialog caller passing a parent-record context gets those silently dropped, so the record→record return-via-identity path won't fire for it. Either copy them through or mark them capture-only in the interface JSDoc. (Also worth a quick confirm that `CompositeKey.ToURLSegment()` / `SimpleLoadFromURLSegment()` are exact inverses for multi-column PKs, since the parent-record re-open round-trips through them.)

**E. Two ~200-line methods with copy-paste duplication** — `tab-container.component.ts`
`loadSingleResourceContent` (~225 lines) and `loadTabContent` (~184 lines) are near-duplicates of each other's cache-reattach + event-wiring blocks — the clearest hit against the ~40-line guideline and the biggest future-maintenance risk in the diff. Decompose and share the common blocks (cache-hit reattach, fresh-create, event-wiring, queryParam reconcile) when you get a chance; not a merge-blocker.

A/B before merge as you suggested; C/D are quick correctness tightenings; E is cleanup-when-convenient. Nice turnaround on the red items.

rkihm-BC:
@MattC-BC — on the two I flagged as merge-blockers (**A. untracked `configSub` use-after-destroy** at `tab-container.component.ts:777`, and **B. `SafeDetectChanges` swallowing all errors**): could you take a look and comment on whether you want to address them in this PR or defer them?

Both are defensible either way — A is a narrow window (destroy while config is still null) and B only bites when a real template error happens to surface through a `SafeDetectChanges` path — so if you'd rather land this and file a fast-follow, I'm fine with that as long as it's a conscious call rather than an oversight. Just let me know your read on each (address now / defer) and I'll align the gate accordingly.

MattC-BC:
@rkihm-BC 

Both addressed now rather than deferred:

A - configSub is now tracked in the component's subscription bag (still self-retires on first config; unsubscribe is idempotent so the two paths coexist safely).

B - SafeDetectChanges keeps the swallow (the catch exists to protect the calling stream from unsubscription, so rethrowing would reintroduce the disease it guards against) but now logs the full underlying error with stack via LogError. Real template errors surfacing through that path are visible; the contract is documented on the function.

rkihm-BC:
Verified both against current HEAD (`a6d3c464`) — **A and B are properly resolved.**

**A ✅** — `configSub` is now `this.subscriptions.push(configSub)` (`tab-container.component.ts:790`) while keeping the self-retire; idempotent unsubscribe means the two paths coexist safely, and the comment documents exactly the destroy-while-config-null window it closes. The use-after-destroy leak is gone.

**B ✅** — and I actually prefer where you landed over my original suggestion. Keeping the swallow + logging the full stack via `LogError` is the right call: narrowing the catch to a message signature (what I proposed) is brittle, and rethrowing a non-matching error would kill the calling stream — the exact disease `SafeDetectChanges` exists to prevent. "Swallow-but-LOG" with the contract documented on the function gets you stream safety *and* visibility. Good.

One residual from A's "while there" note — the deferred `setTimeout(…, 0)` bodies (breakpoint rebuild / move-to-records / deferred inits) still have no destroyed-guard. It's genuinely low-severity (most are internally guarded by `recordsLayoutInitialized` etc.) and **not a blocker**; a shared `private destroyed` flag checked in those bodies would tidy it whenever convenient.

**From a code-review standpoint my two gates are cleared.** Remaining 🟡 items C (merge-vs-replace param asymmetry), D (`resolveSourceContext` dropping 3 advertised fields + composite-key round-trip confirm), and E (decompose the two ~200-line duplicated load methods) stand as previously described — all non-blocking follow-ups.

The only things left before merge are process rather than code: a second human reviewer sign-off given the breadth (77 files, default-behavior flip), reconciling the still-unchecked test-plan boxes with the "live-verified" claim, and the DOM-level tests you cc'd @BarnattW-BC for. Those are your/the release owner's call — nice work turning the substantive items around fast.

rkihm-BC (approved):
## Approving ✅

Reviewed across several rounds — core logic directly plus parallel deep-dives on the layout/state core, the tab-container, the navigation/crumb/config layer, and the bottom-sheet/theme layer. The substantive items are all resolved and the fixes are clean.

**🔴 Merge-blockers — all resolved and verified against `a6d3c464`:**
- **Classic escape hatch** — `raw?.trim().toLowerCase() === 'classic'`; any casing/whitespace variant now opts out correctly.
- **`mj-bottom-sheet` a11y** — full route taken for the shared primitive: `aria-modal="true"`, real Tab/Shift+Tab focus trap, and body scroll-lock that snapshots + restores the prior `overflow` (incl. destroy-while-open), with new tests.
- **Desktop restore refs** — sanitize pass promoted into a pure `SanitizeLayoutNodeForLoad` with `componentState` deep-cloned via `structuredClone`, unit-tested directly; GL no longer holds references into persisted config on any path, and the old `as any` casts are gone.
- **`configSub` use-after-destroy** — now tracked in the subscription bag with idempotent self-retire.
- **`SafeDetectChanges`** — swallow-but-LOG with full stack via `LogError` and a documented contract; the right call over narrowing the catch.

**Standouts:** `layout-transforms.ts` and its tests are exemplary; the `slide-panel` `visible → sp-open` rename is complete and clean (all four AI detail panels migrated, dead CSS removed, stale test deleted); the GL token theme is design-token clean; strong typing throughout with no `any` in new code; and a genuine stored-XSS fix (`replaceChildren`/`textContent`) along the way.

**Non-blocking follow-ups** (fine to land as-is): C (merge-vs-replace param asymmetry in the crumb fallback path), D (`resolveSourceContext` dropping three advertised fields + a composite-key round-trip confirm), E (decompose the two ~200-line duplicated load methods), and the deferred-`setTimeout` destroyed-guard tidy.

Approving from a code-review standpoint. The remaining pre-merge items are process, not code, and are the release owner's call: a second reviewer given the breadth (77 files, default-behavior flip to records-style), reconciling the still-unchecked test-plan boxes with the "live-verified" claim, and the DOM-level tests cc'd to @BarnattW-BC. Nice work on the fast, thorough turnaround.

PR #3448, opened 2026-08-04
fix(server): make timingSafeEqual compile under both @types/node lines

SDesai-BC:
`npm run build` fails on `@memberjunction/server` for anyone who installs with `npm install`, while CI stays green. Both are correct — they resolve the same dependency to two different versions.

## Root cause

Two declarations in the repo contradict each other:

| Where | Declares |
|---|---|
| `packages/MJServer/package.json` → `dependencies` | `"@types/node": "24.10.11"` |
| root `package.json` → `overrides` | `"@types/node": "20.14.2"` |

`npm ci` materializes the lockfile verbatim, so MJServer gets its nested `24.10.11`. `npm install` re-resolves, honors the override, collapses to `20.14.2`, and dedupes to the root copy — leaving the now-stale `packages/MJServer/node_modules/@types/node => 24.10.11` entry sitting in `package-lock.json` describing a directory that does not exist.

That matters because `crypto.timingSafeEqual` is typed `(a: NodeJS.ArrayBufferView, b: NodeJS.ArrayBufferView)`. TS 5.7 made `ArrayBufferView` generic over its backing buffer. In `@types/node@24.x`, `Buffer<TArrayBuffer extends ArrayBufferLike = ArrayBuffer>` is generic and satisfies it. In `20.14.2` it is still `interface Buffer extends Uint8Array` with `.buffer: ArrayBufferLike`, so TS 5.9.3 rejects it:

```
src/context.ts:436:27 - error TS2345: Argument of type 'Buffer' is not assignable to parameter of type 'ArrayBufferView'.
  Type 'Buffer' is not assignable to type 'Uint8Array<ArrayBufferLike> | DataView<ArrayBufferLike>'.
      The types of 'slice(...).buffer' are incompatible between these types.
```

The call arrived in a71aa0bd (#3410), which replaced a `===` comparison of the superadmin API key with a constant-time digest comparison. It is the repo's first `timingSafeEqual` in MJServer, which is why nothing tripped this before — most Node APIs are typed `Buffer | ...` explicitly rather than with a bare `ArrayBufferView`.

## What this does

Wraps both digests in `Uint8Array.from(...)`, which compiles under **both** `@types/node` lines, so the build no longer depends on which install path a developer used. Same idiom already in the repo for the same reason: `Lists/server/src/deltaToken.ts:124` and `MessagingAdapters/src/slack/slack-routes.ts:103-104`.

No runtime change. `Uint8Array.from` copies 32 bytes, and the comparison stays constant-time — the security property #3410 added is untouched.

## Why not fix the dependency instead

The honest fix is to stop the root override from contradicting seven packages' declared `@types/node`:

```
packages/MJServer                       dep    24.10.11
packages/AI/MCPServer                   devDep 24.10.11
packages/AI/A2AServer                   devDep ^24.10.11
packages/ComponentRegistry              devDep 24.10.11
packages/MJCodeGenAPI                   devDep 24.10.11
packages/ContentAutotagging             devDep ^24.10.11
packages/Actions/ScheduledActionsServer devDep ^24.10.11
```

Either raising the override or scoping it to exempt the server packages needs a lockfile regen and a full CI pass across 297 build tasks — real blast radius, and it should not ride along with a build unbreak. **This PR only unbreaks the build; the override conflict is still live** and will resurface the moment anyone touches an `ArrayBufferView`-typed API in any of the other six. Worth its own issue.

No existing issue — found while building `next` locally.

## Verification

Run against this branch's base (`cd520e2fce`), on a tree installed with `npm install` (i.e. the failing configuration):

| Check | Result |
|---|---|
| `npx tsc --noEmit -p packages/MJServer/tsconfig.json` | clean |
| `npm run build` (root, all packages) | 297 successful, 297 total |
| `cd packages/MJServer && npm run test` | 757 passed, 56 skipped, 0 failed (55 files passed, 2 skipped) |

Both type versions were also checked directly against an isolated 4-line reproduction of the call: identical raw-`Buffer` code fails under `@types/node@20.14.2` and compiles clean under `24.10.11`, with TS pinned at 5.9.3 in both.

Integration tier not run — this is a type-level change with no runtime semantics, and it needs a live DB.


AN-BC:
closing this as I'm alreayd fixing this in another PR

PR #3454, opened 2026-08-05
fix(standards): give the ui-layers manifest half a reviewed-exception form

cadam11:
**One sentence:** The `ui-layers` gate enforces a rule on `package.json` that it has no way to grant an exception to, so a package whose blessed import forces it to declare a banned dependency has no green state at all — this adds the manifest twin of the `mj-ui-layers-allow` comment marker.

> **Rebased.** This PR originally also fixed the self-flagging regex and `ng-file-storage`'s red. Both were fixed on `next` in parallel while it was open (`4d05392347`, `c9b7423c3f`) and have been dropped. What remains is only the piece still missing from `next`.

## The gap

The check enforces the layer ban in two halves:

| Half | Checks | Exception mechanism |
|---|---|---|
| Source | `import` statements | ✅ `mj-ui-layers-allow` comment |
| Manifest | `dependencies` in `package.json` | ❌ **none** |

A blessed import still has to resolve, so its dependency still has to be declared. That leaves no green state:

- **Declare the dep** → manifest half fails.
- **Don't declare it** → TS2307, because the blessed import needs it.

## The fix

```jsonc
{
  "mjUILayer": "widgets",
  "mjUILayerAllow": {
    "@memberjunction/ng-shared": "why this exception exists, and what unblocks removing it"
  }
}
```

The object form is deliberate — a bare array lets exceptions accumulate silently. The reason is required, and **the allowlist has to shrink on its own**: a blank reason fails, and so does an entry that no longer excuses anything.

## Relationship to the `surface` fix already on `next`

These are complementary, and the guide now says which to reach for first:

- **Relabel via `mjUILayer`** when the package genuinely belongs to a layer where the dep is legal. That was the right call for `ng-file-storage` in `c9b7423c3f`.
- **`mjUILayerAllow`** for the narrower case: a package that is genuinely `widgets` apart from one reviewed import, where relabelling would drop the widget rules from *every other file* in it.

## For the reviewer

- **This mechanism already landed once and was lost.** `4d05392347` added it as an unreasoned `mjUILayerAllowedDeps` array; `next`'s manifest-half loop is now byte-identical to the original, so it vanished in a later merge conflict resolution. That is the **second** time work in this area has been lost that way — `e1acebd09a` did it to the dependency declaration itself. Worth a look at whether something structural is going wrong in this area's conflict resolutions.
- **Nothing uses this yet.** No package declares an exception; `next` is green without it. This is the escape hatch being available before the next package needs it, not a fix for a current red. Reasonable to push back on that as speculative — the argument for landing it is that the hole is reachable by design, since any allow-marked import of a *dependency* trips the manifest half automatically.
- **Check I didn't weaken the gate.** The hatch is per-dependency and per-package, never a global off-switch, and I ran the negative case before the rebase: removing an exception re-fails the check with the original message.

## Verification

- Standards gate on the rebased tree: **exit 0**, `✓ All adopted standards pass`
- `@memberjunction/standards`: **57 tests pass** (49 before, 8 added)
- `MultiProviderCompliance`: still passes
- Integration tier not run — this touches a lint gate and docs, nothing in the runtime path.

https://claude.ai/code/session_01Mn3NY1z42C27NgyPEf491p

### DIFF
--- a/.changeset/ui-layers-manifest-exception.md
+++ b/.changeset/ui-layers-manifest-exception.md
@@ -1,0 +1,20 @@
+---
+"@memberjunction/standards": patch
+---
+
+ui-layers: let the manifest half of the layer ban carry a reviewed exception.
+
+The source half honours an `mj-ui-layers-allow` comment, but `package.json` carries no comments, so a
+package whose blessed import forces it to declare a banned dependency has no green state: declaring
+the dep trips the manifest half, and dropping it breaks the build the blessed import still needs.
+
+Packages can now declare `"mjUILayerAllow": { "<dep>": "<reason>" }`. The reason is required — a blank
+one fails, and so does an entry that no longer excuses anything, so the allowlist has to shrink on its
+own rather than rotting into a permanent pass.
+
+The alternative already available — relabelling the package to a layer where the dependency is legal —
+stays the right call when the package really does belong to that layer. This is for the narrower case:
+a package that is genuinely widgets apart from one reviewed import, where a whole-package downgrade
+would drop the widget rules from every other file in it.
+
+No change to what any layer forbids.
--- a/guides/UI_LAYERING_GUIDE.md
+++ b/guides/UI_LAYERING_GUIDE.md
@@ -410,6 +410,29 @@ Two things keep the gate from crying wolf, because a gate that cries wolf gets s
 directly above it, with a reason. One line above, not more — a marker that can drift away from the
 thing it excuses stops meaning anything.
 
+The manifest half needs its own form, because JSON carries no comments. A blessed import still has
+to resolve, so the dependency still has to be declared — use `mjUILayerAllow`:
+
+```jsonc
+{
+  "mjUILayer": "widgets",
+  "mjUILayerAllow": {
+    "@memberjunction/ng-shared": "why this exception exists, and what unblocks removing it"
+  }
+}
+```
+
+The reason is required, not decorative: an entry with a blank reason fails, and so does an entry
+that no longer excuses anything. Without this, a package holding a blessed import had no green
+state at all — declaring the dep tripped the manifest half, dropping it broke the build the import
+needs — and a permanently red gate teaches everyone to merge red.
+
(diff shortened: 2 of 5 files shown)


cadam11:
Closing as superseded — `next` solved this from the other direction while the PR sat open.

`190db45a01` ("excuse manifest declarations that back marker-excused imports") makes the manifest
half **inherit** the source `mj-ui-layers-allow` marker, via an `excusedSpecifiers` set collected
during the source pass. That closes the exact deadlock this PR was opened for — declare the dep and
the manifest half fails, drop it and the import breaks — without adding a second exception form.

Checked before closing, rather than assuming:

- **Same scenario, already covered.** `next`'s test `excuses a declaration that backs a
  marker-excused import` is the same package, dep, marker and zero-violation expectation as this
  PR's `honours a reasoned mjUILayerAllow …`. The rebase conflict was literally the two mechanisms
  landing on the same lines.
- **Per-dependency granularity, which was this PR's main design goal, is already there** —
  `a marker on one import does not excuse a different forbidden declaration`, plus
  `still flags the declaration when the import has NO marker`.
- **No residual hole worth a second mechanism.** `FORBIDDEN_DEPS` and `FORBIDDEN_MODULES` carry
  identical patterns per layer, so every bannable dep has a markable import. The only case
  `mjUILayerAllow` would additionally cover is a declared dep with no import anywhere in scanned
  source — which the dependency-check gate flags as unused regardless.

Shipping `mjUILayerAllow` on top would mean two hatches for one deadlock, and this PR's own body
already flagged it as speculative with nothing using it. Better to keep the single mechanism.

The one thing genuinely missing from `next` was documentation: neither the guide nor the Standards
README says that a marker also excuses the manifest declaration, which is the non-obvious half.
Salvaged into a docs-only follow-up, along with this PR's "reach for the layer value first"
guidance, which is still good advice and survived the supersession intact.

Not reopening the note about mechanisms in this area being lost twice in conflict resolution
(`4d05392347`, `e1acebd09a`) — that observation still stands and is worth a look independently of
this PR.

PR #3467, opened 2026-08-05
UI hoist from BizApps accounting: left-nav desktop collapse, workspace tabs, entity-grid fill-width

MarceloT-BC:
Graduates three UI capabilities built and battle-tested in the BizApps accounting app (its list-page-standard wave, MemberJunction/bizapps-accounting#43) into MJ base. All are pure widgets per the four-layer standard — no Router, no global providers, tokens-only CSS — and are catalogued in the ui-components README.

## 1. `mj-left-nav` desktop collapse (opt-in)

`[Collapsible]` renders a locked-position double-angle toggle chip — **only the glyph flips on toggle; the chip never relocates, so the user never re-aims** — and `[(Collapsed)]` presents the rail as a `CollapsedWidth` (60px) icons-only strip: labels visually hidden but kept in the a11y tree, section labels folded to centred divider lines, badges docked on the icon corner, per-item title/aria tooltips auto-enabled. The component owns presentation only — the consumer owns + persists the state (same split as `ExpandedIds`). Desktop-only; the ≤700px drawer is untouched. Hover-to-peek is deliberately not included (product ruling: a rail that auto-expands under the pointer moves the target being aimed at). Default-off ⇒ zero change for existing consumers (verified live against Admin).

left-nav-collapsed

left-nav-expanded

## 2. Workspace tabs (new): `mj-workspace-card` + `mj-workspace-tab-strip` + `MJWorkspaceTabStore` + `mjTip`

The **workspace** pattern — a screen where the user builds several drafts in parallel, each in a browser-style tab (open / switch / drag-reorder via CDK / close, dirty-dot, rejected/complete states) — over a pure, framework-free state machine (23 unit cases ported with the code), wrapped in a slotted card frame: identity band, single scrolling body, opt-in standardized confirm / save-as-draft / discard footer (only the confirm label varies per workshop). Tabs carry an opaque `State` the host owns — no app types in the framework. The card is a size query container (content sizes with `cqh`/`cqw`, never viewport units). v1 is session-scoped. `mjTip` is the delayed, non-interactive truncation tooltip the strip uses.

workspace-tabs

## 3. `mj-entity-data-grid` width contract

New `[FillWidth]` input appends an inert trailing filler column so row banding/rules reach the container edge without stretching any real column. `width: 'auto'` + `maxWidth` in `GridColumnConfig` — already public API — now actually map to AG Grid flex sizing (previously silently dropped to `undefined`) and survive saved-grid-state restores. `computeFieldsList` exported from `record.util`. (The grids visible in the rail screenshots above are running FillWidth + an auto-width Memo column.)

## How to see it working / follow-ups

The accounting app consumes all of this today — its `[FillWidth]`/`[IconOnly]` bindings are staged in MemberJunction/bizapps-accounting#44, **blocked on this shipping**. After release: accounting bumps its `@memberjunction` floor, its `shell-rail` thins to a pass-through of the native collapse, and its parked `transfer-pending/workspace-tabs` copy is deleted in favor of these.

## Validation (all on a fresh `origin/next` instance)

| Check | Result |
|---|---|
| ui-components + entity-viewer builds (ngc) | clean |
| entity-viewer unit suite | 140/140 |
| ui-components unit+DOM suite | 45/46 files green — the 1 red (`slide-panel`, untouched here) fails identically on pristine `origin/next` (pre-existing) |
| New coverage added in this PR | workspace-tab-store 23/23 · left-nav collapse DOM spec 6/6 (incl. the chip-never-relocates invariant) |
| `npm run check:ui` | 0 violations |
| `npm run check:standards` (incl. ui-layers) | all pass |
| `npm run test:integration` (deterministic tier, seeded per quickstart §3.1) | 53/55 — sole failure IT69 is known upstream #3251 (transport sequencing), unreachable from this Angular-only change |
| Live Explorer keystone (Admin left-nav shell) | renders, collapse toggle correctly absent by default, zero console/page errors |

Changeset: **minor** on `ng-ui-components` + `ng-entity-viewer` (additive public API) — happy to adjust the bump in review.


MarceloT-BC:
@AN-BC For your review if you are interested.

AN-BC:
I am deferring this to @MattC-BC who can decide about if this makes sense for MJ. The `Workspace Tab` concept is interesting but is a misnomer, it is just a "tab" component, that can be used anywhere and we've built stuff like this before. Matt can you guide Marcelo to that, also his styling for that isn't aligned with your best practices. Pleaes give him a hand.

Thanks to Marcelo for raising this to improve core MJ.

MattC-BC:
@MarceloT-BC @AN-BC - I can take a look at this later today

rkihm-BC:
## Review — thorough pass on the three features here

Thanks @MarceloT-BC for pushing these up, and following on from @AN-BC's hand-off to @MattC-BC. I did a full read of the diff, the existing tab infrastructure, and the a11y surface. Sharing the whole thing here so it's one place for the discussion.

The key framing: **this is really three independent features**, and they're at very different levels of readiness. I'd genuinely recommend **splitting the PR** so the two that are ready aren't held up by the one that needs a design conversation.

| Feature | Verdict |
|---|---|
| **3. `entity-data-grid` FillWidth + auto-width fixes** | **Strong — merge-worthy on its own.** The most valuable part. |
| **1. `mj-left-nav` desktop collapse** | **Good.** Clean, default-off, well-tested, a11y-aware. |
| **2. Workspace tabs (`mj-workspace-*` + store + `mjTip`)** | **Hold / rework** — this is the part @AN-BC flagged, and I think the concern is valid. Detail + a concrete sketch below. |

---

### What's good

**Grid work (#3) is excellent.** It fixes two genuinely nasty *silent* bugs: (a) `width:'auto'` + `maxWidth` were public `GridColumnConfig` API but silently mapped to `undefined`; (b) a user's per-entity saved default from the generic browser could silently override a page's explicit `[Columns]` **for that one user only** — the worst kind of "works on load, then randomly stops, and only for some people" bug. The `_columnsFromHost` / `_gridStateFromUserDefault` precedence logic correctly handles **both** `@Input` assignment orderings (Params-before-Columns and the reverse). `computeFieldsList`'s new third arg is optional and backward-compatible, and treating host columns as a *superset* of fetched fields (not a replacement) is the right call. The filler-column idea (grow a deliberately-empty column instead of stretching a real one) is the correct fix for "table doesn't reach the card edge."

**Left-nav collapse (#1) is disciplined.** Truly default-off → zero change for existing consumers. The a11y is genuinely good: labels visually hidden via `clip-path` but kept in the a11y tree, section labels fold to dividers while preserving box height (identical rhythm on expand), tooltips auto-enable only when collapsed. The "locked toggle position" invariant is even asserted in a DOM test. Presentation-owned-by-component / state-owned-by-consumer mirrors the existing `ExpandedIds` split.

**Cross-cutting:** pure widgets (no Router/global providers), token CSS, and the tab **state machine extracted as a framework-free `MJWorkspaceTabStore` with 23 exhaustive unit cases** — that separation is nice engineering, and `Tabs` returning a copy to block external mutation is a good touch.

---

### What's concerning (all in feature #2)

**A. It's a third tab strip.** MJ already ships `@memberjunction/ng-tabstrip` (`mj-tabstrip` — closeable tabs, overflow scroll, context menu, cancelable close) and `mj-tab-nav` (data-driven tab nav with badges). What `mj-workspace-tab-strip` adds over `ng-tabstrip` is **drag-reorder + a dirty-dot + a draft/rejected/complete lifecycle** — those read as *features that belong on the existing strip*, not a reason for a parallel one. @AN-BC's "it's just a tab component, we've built this before" is accurate. "Workspace" is a *use case*, not a component category — the name bakes an accounting concept into a generic package.

**B. A11y gaps for something billed "canonical."** `role="tab"`/`role="tablist"` sit on non-focusable `<div>`s while the real click target is an inner `<button>` with no role. There's **no roving tabindex / arrow-key navigation** (the ARIA tabs pattern requires it), no `aria-controls`→panel linkage, and drag-reorder is mouse-only. This is below the bar the left-nav work sets *in the same PR*.

**C. `mjTip` is a hand-rolled body-appended tooltip.** It doesn't reposition/dismiss on scroll (`position:fixed` + body node detaches from its anchor); it can leak if the host is removed from the DOM while hovered (no `mouseleave` fires); and because its styles live in a `.ts` file it bypasses `check:ui` — which is how a hardcoded `rgba(0,0,0,0.18)` shadow coexists with "0 violations." This should be one shared, hardened tooltip primitive, not a new per-feature one.

**D. `Discard` emits with no dirty-guard** even though the store tracks `HasDirtyTabs` — for a "standardized footer" meant to stop workshops hand-rolling their own, shipping the destructive verb with no guard invites every consumer to forget it.

**E. Minor:** `State: unknown` (via `TState = unknown`) is defensible as a framework boundary but brushes against the repo's "no `unknown` shortcut" rule; and the inline comments cite another repo's decision log ("Marcelo 2026-07-21", "ruling 2026-07-30") which shouldn't live in MJ base source.

---

### What's missing (feature #2)

- **No in-tree consumer** — the app that uses these is in bizapps-accounting, "blocked on this shipping." MJ base would own a public API with nothing here exercising it beyond the store unit tests.
- **No DOM/interaction tests** for the strip or card (only the pure store is covered), and **no keyboard tests** (because there's no keyboard interaction yet).
- **No note addressing the `ng-tabstrip` / `mj-tab-nav` overlap** — no rationale for a third strip, no consolidation/deprecation plan.

---

### Extend `ng-tabstrip` vs. new component — a concrete sketch

The honest tension is an **API-model difference**, and it's worth naming so the discussion is grounded:

- **`mj-tabstrip` is content-projection based:** you declare `<mj-tab>` children and it projects `<mj-tab-body>` panels. It's cancelable-close, has a context menu and overflow scroll — but it owns the *bodies*.
- **`mj-workspace-tab-strip` is data-driven + presentation-only:** `[Tabs]` array over a pure store, emits intent, projects nothing. That's a deliberately different (and, for the draft-workspace case, nicer) contract.

So "just reuse `ng-tabstrip`" isn't a drop-in. Here are the three honest options:

**Option 1 — Extend `ng-tabstrip` with the missing behaviours.** Add reorder + dirty + status to the component MJ already hardened:
```html
<mj-tabstrip [Reorderable]="true" (TabReordered)="store.Reorder($event.previousIndex, $event.currentIndex)">
  <mj-tab *ngFor="let t of store.Tabs" [Dirty]="t.Dirty" [Status]="t.Status">{{ t.Label }}</mj-tab>
  <mj-tab-body *ngFor="let t of store.Tabs"> … </mj-tab-body>
</mj-tabstrip>
```
✅ One tab component to maintain; inherits overflow scroll, cancelable close (which also solves **D**), context menu.
❌ Fights the content-projection model — the workspace case wants a *data-driven, body-agnostic* strip, so you'd be bolting a second usage mode onto an already-stateful component.

**Option 2 — Keep a thin data-driven strip, but as a headless mode over the same core.** Factor the reorder/dirty/status/overflow logic into a shared base (or the existing store) and let `ng-tabstrip` and a `mj-draft-tab-strip` both consume it:
```
MJTabStore (reorder + dirty + status + neighbour-activation)  ← already built here, keep it
   ├── mj-tabstrip          (content-projected bodies — existing)
   └── mj-draft-tab-strip   (data-driven, presentation-only — this PR, renamed off "workspace")
```
✅ Keeps the nice pure store and the data-driven ergonomics; no third *state machine*.
✅ Rename resolves **A**'s "misnomer"; a11y fixes (**B**) land in one place both strips share.
❌ Most work — needs the base extraction and buy-in on the layering.

**Option 3 — Ship as-is but rename + fix a11y + reuse a shared tooltip**, and explicitly accept two strip components with a documented boundary ("`ng-tabstrip` = projected bodies; `mj-draft-tab-strip` = data-driven drafts").
✅ Least churn to this PR.
❌ Leaves MJ with two strips by choice — only worth it if Matt decides the two contracts are genuinely distinct enough to justify it.

My read: **Option 2** is the right long-term shape (keep the store, rename the strip, share the a11y + tooltip), but that's exactly the call @MattC-BC should make.

---

### Suggested path

1. **Split the PR** — land grid (#3) and left-nav collapse (#1) now; they're ready and default-safe.
2. **Take #2 to the design conversation** @AN-BC and @MattC-BC asked for: pick an option above, rename off "workspace," fix the tablist a11y (roving tabindex + arrow keys, role on the focusable element), share/harden the tooltip, and add DOM/interaction tests to match the store's coverage.

@MattC-BC @MarceloT-BC — would love your read on the extend-vs-new question specifically; happy to prototype whichever direction Matt prefers.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

MattC-BC:
Following up on the hand-off from @AN-BC and the review discussion above. Rather than splitting the PR, I built directly on this branch and worked through all three features plus the design question that was left open on #2. Everything below is pushed here as batched commits, each independently revertable. Summary of what I did and why:

Feature 1: left-nav collapse (a0008c0a89)
The collapse was solid but built against flat, fully-iconed rails. I closed the three gaps in the component's own contract:

Tree sections now fold to their top-level items while the rail is icons-only. A tree has no way to express itself in a 60px strip (chevrons crowd the icons, depth indentation pushes children into the clipped edge). A top-level item stands in as active when the actual active item is a hidden descendant, with aria-current="true" vs the exact match's "page" so the announcement stays honest. ExpandedIds is untouched, so the tree comes back exactly as the user left it on expand.
Icon-less items render an uppercase monogram from the label instead of collapsing to an invisible but clickable button (icon is optional on MJLeftNavItem, so this was reachable).
The mobile drawer: the collapsed styles were gated to 701px and up, but the behavior was not, so a persisted Collapsed would follow the user into the drawer as a flattened tree with monograms next to full labels. The breakpoint is now tracked via matchMedia (same pattern as mj-filter-popover) and gates the native collapse. IconOnly is deliberately not gated, since that flag means the consumer's own CSS narrowed the rail and we can't know their breakpoints.
Also paired the toggle's aria-expanded with aria-controls, and wired Testing Explorer as the first in-tree consumer (suite tree, badges, section labels, so it exercises every path), with state persisted per-user via UserInfoEngine.

Feature 3: entity-grid width contract (e5344faa06)
The precedence work here was the strongest part of the PR and I didn't touch it. What I fixed is the host-[Columns] field-name seam: a page writing field: 'name' against an entity that spells it Name was accepted by the case-insensitive metadata lookups but then failed everywhere the spelling addresses real data. On the GraphQL transport the selection set is case-sensitive, so the whole view died with zero rows; after fixing the fetch, the column rendered blank cells, vanished from captured grid state on the first column resize, lost its saved sort, and exported empty.

Fixed at the source instead of patching six comparisons: a new canonicalizeColumnFields helper rewrites host column names to the entity's spelling at the two points host columns first meet metadata. Everything downstream (colDefs, _columnStates, captured state, sort settings, export, fetch) reads the same vocabulary. Hidden host columns are also fetched now, since the grid deliberately keeps them in its column model so a saved [GridState] can re-show them, and a column returning with no data reads as a data bug. 12 unit tests on the helper and fetch contract, verified they actually fail when the fix is reverted.

Feature 2: the tab strips (aac373f7ad)
This was the open design question, so here's my read on extend-vs-new after actually digging into both components: the two API contracts are genuinely different and should stay separate, but the look and the keyboard behavior should be physically incapable of drifting apart. mj-tabstrip is projected-children and index-addressed and owns its bodies; the workspace strip is data-driven and id-addressed and projects nothing. Option 2's shared store doesn't survive contact with that (there's no array to reorder in a projection model), so I unified the two layers that are actually shared:

One chrome: a global .mj-tabs* stylesheet in ng-ui-components (same delivery as button.scss), consumed by both strips. The workspace visual language won, and not as a taste call: the old tabstrip styling used legacy --gray-* aliases that resolve to primitive tokens and never re-pointed in dark mode, differentiated its selected tab by text color alone, and used a raw ✖ glyph. An active tab's border and top accent line now follow its status color (--mj-tab-accent: brand primary normally, warning when rejected, success when complete), so the active state's chrome speaks in one voice.
One keyboard contract: a new mjTabList directive implementing the APG tabs pattern (roving tabindex, one Tab stop per strip, arrows/Home/End with focus-follows-selection, Enter/Space, Delete/Backspace to close, hidden-tab skipping, editable-content passthrough). It operates on the DOM, so both strips consume it without either state model yielding. Both strips previously had zero keyboard support, so this fixes the a11y gap once, in one place. role="tab" now sits on the focusable element, with full aria-controls/tabpanel linkage in mj-tabstrip.
The store and its 23 tests are unchanged. The state machines stay separate on purpose.
Riding along: mj-tabstrip's three viewport-height hardcodings (calc(100vh - …)) are gone. They blew up any host that wasn't a full page and made focus changes scroll the whole pane. The strip now sizes to content and the host owns region sizing, which is how every other MJ widget works. That's a real behavior change for ng-tabstrip consumers and it's called out in the changeset; the Entity Actions form (the only in-tree consumer) got an explicit [Height] on its Filters grid to preserve its internal scroll. The package's stale "DEPRECATED, use Kendo" notice is also gone.

@MarceloT-BC, one action item for bizapps-accounting#44: the workspace strip's chrome now comes from the global stylesheet, so when accounting bumps its floor it needs @import '@memberjunction/ng-ui-components/dist/lib/tabs/tabs'; in its global styles or the strip renders unstyled. It's in the changeset and both READMEs.

Dev fixtures (ca14a65a31, 6647c01d54)
Two new pages under Admin, Developer Tools, both born from the "no in-tree consumer" gap this PR's review flagged: Grid Width Lab (a live grid with host [Columns], FillWidth toggle, a hidden-column toggle, a params-churn toggle, and a readout of the exact Fields list sent to RunView) and Tab Strip Lab (both strips over identical labels with every tab state seeded, so any drift between them is immediately visible). These double as permanent regression fixtures.

Cross-cutting (33bc6f7e0d)
If you pull this branch and run a clean pnpm install, note MJExplorer now declares @memberjunction/ng-container-directives and @auth0/auth0-spa-js. Both were latent gaps papered over by npm hoisting that the pnpm cutover's strict layout turns into hard dev-serve failures. The externalized-deps guard passes again.

Process and verification
Each chunk went through an adversarial subagent audit before committing, and the audits earned their keep (among other things they caught that an early CSS fix of mine provably did nothing, dead scroll machinery, and a keyboard path that threw on Entity Actions' hidden tabs; all confirmed findings are fixed). Current state: ui-components 517/517, entity-viewer 152/152, dashboards 1695/1695, tab-strip 13/13, integration tier unaffected, check:ui and check:standards clean, changesets in place for every package with public changes.

Deferred deliberately: renaming the workspace components off the "workspace" name (mechanical churn that would have buried the substantive diffs; happy to do it as a follow-up), and the question of eventually migrating Entity Actions off ng-tabstrip entirely now that it's the only consumer.

Review pointers: the tab work is the biggest diff, and the two decisions most worth challenging are the shared-global-stylesheet delivery (vs keeping each strip self-contained) and follow-focus activation on arrow keys. @rkihm-BC the a11y layer is probably where your eyes are most valuable.

rkihm-BC:
## Re-review — after @MattC-BC's 6 commits

I did a full second pass over the whole diff (now +3604/−201 across 53 files), read every commit, and focused hardest on the a11y layer Matt asked me to scrutinize. Short version: **Matt turned the contested part of this PR into its strongest.** Every concern from the first round is addressed or deliberately (and reasonably) deferred. Detail below, then a concrete **For Application Owners** section answering "what has to change downstream, and where."

### What's good this round

**The extend-vs-new question is resolved well.** Two strips keep separate API contracts *and* state models, but share the two things that must never drift: **one look** (`tabs.scss`) and **one keyboard contract** (`mjTabList`). Matt's argument that a shared *store* can't span a projection model (`mj-tabstrip` has no array to reorder) is correct, and sharing a DOM-level directive rather than a base class is the right mechanism.

**The a11y layer is genuinely excellent** — this is where I looked hardest and it holds up:
- `mjTabList` is a faithful APG tabs implementation: roving tabindex (one Tab stop/strip), Arrow/Home/End with wrap, Enter/Space, Delete/Backspace close, **hidden-tab skipping that still reports the full-list index** (index-addressed hosts don't mis-select), and **editable-content passthrough** (won't hijack Backspace in an inline-rename input). 14 keyboard tests on synthetic markup — correctly hosted on made-up DOM to prove it works on any `role="tab"` structure.
- `role="tab"` now sits on the **focusable element** (the anti-pattern I flagged is gone); `aria-controls`/`aria-labelledby`/`tabpanel` linkage wired in `mj-tabstrip` via per-instance ids.
- Keyboard close routes through `mj-tabstrip`'s cancelable `BeforeTabClosed`, so the close-guard holds for keyboard users too.

**Left-nav hardening is thorough:** tree sections fold to top-level while collapsed (with a stand-in active ancestor using `aria-current='true'` vs the exact match's `'page'`), icon-less items get an astral-safe monogram (`Array.from`, not `charAt`), and the mobile-drawer leak is closed by tracking the breakpoint in TS via `matchMedia` (zone-corrected, cleaned up on destroy) — not just CSS.

**The GraphQL field-canonicalization fix is a subtle, correct catch.** `canonicalizeColumnFields` rewrites host `field:'name'` to the entity's `Name` before it hits the wire, because GraphQL selection sets are case-*sensitive* while the metadata match is not — so the case-insensitive path would render a column that fetches zero rows. No host mutation, 12 tests, verified-failing-when-reverted.

**Token compliance is clean** (`tabs.scss` is all `var(--mj-*)`, replacing legacy `--gray-*` primitives that never re-pointed in dark mode — AN-BC's styling concern addressed), **no circular dependency** (`ng-ui-components` is a near-leaf; `ng-tabstrip → ng-ui-components` is one-directional), and the two **DevTools labs** answer the "no in-tree consumer" gap while doubling as regression fixtures.

**Verified Matt's "only in-tree consumer" claim:** the CodeGen path that emits `<mj-tabstrip>` (`angular-codegen.ts` `innerTabStripHTML`) is `@deprecated`, and **0 of 383 generated forms use it**. The only live consumer is the Entity Actions form, which he fixed. So the height change's *in-repo* blast radius really is contained.

### On the two decisions Matt flagged for challenge

**Shared global stylesheet — sound, and more justified than it first looks.** `mj-tabstrip` styles *projected* `<mj-tab>` children, and Angular's emulated encapsulation can't reach projected content without deprecated `::ng-deep`; a global stylesheet is the correct mechanism there, and `button.scss` is a real precedent. The one genuine cost is the **silent failure mode** (forget the import → unstyled, not an error). Suggestion: a dev-mode guard in both strips that reads a sentinel computed property `tabs.scss` sets and `console.warn`s once if absent — turns "silently unstyled" into an actionable message. Low effort, non-blocking.

**Follow-focus arrow activation — fine for today, worth an opt-out for tomorrow.** For both current strips (cheap panels) it's the APG-recommended default and matches browsers. But `mjTabList` is now *the* shared contract, and APG prefers manual activation when a panel isn't shown without latency. A future/external host where activation has side effects (commit a draft, network fetch) can't opt out. Suggestion: an opt-in `[mjTabListManual]` that moves focus + roving index on arrows but only activates on Enter/Space. Non-blocking. (I confirmed the current follow-focus path is safe — focus is set *before* emit, and both strips `track` stable ids so the focused element survives re-render.)

### Residual minor items (none blocking)
- **`mjTip` unchanged** — still a body-appended tooltip with no scroll-dismissal, a possible orphan if the host is removed mid-hover, and a hardcoded `rgba(0,0,0,0.18)` shadow in `.ts` that bypasses `check:ui`. Lower-stakes now that tab chrome is tokenized.
- **`workspace-card`'s `Discard` still emits with no dirty-guard** (only the card footer; the tabstrip close-guard is fixed). Host-responsibility is defensible for a standardized footer.
- **`ngAfterViewChecked` runs a `querySelectorAll` every CD cycle** — idempotent and diff-guarded, so correct, just a small per-tick DOM query for tabstrips in frequently-checked forms.

The **"workspace" rename is deferred** to a follow-up (mechanical churn) — **acceptable, given the follow-up actually lands.** The **MJExplorer dep-hygiene / regenerated-registration scope creep is fine.**

---

## 📋 For Application Owners — what you must change

There are **two** downstream-visible changes here: a **tab-chrome stylesheet** and a **behavioral height change in `mj-tabstrip`**. Who has to do what depends on whether your app runs on the MJ Explorer shell.

### Change 1 — Tab chrome moved to a shared global stylesheet

The `.mj-tabs*` look now lives in one place: `@memberjunction/ng-ui-components/.../tabs/tabs.scss`. It is loaded via a global `@import`, added to `@memberjunction/ng-explorer-app`'s aggregated `styles.scss`.

**Category A — Apps on the MJ Explorer shell (MJExplorer and its forks): ✅ nothing to do for styling.**
The import lives inside the **distributed** `ng-explorer-app` stylesheet, and your app already consumes it:
```scss
@use "@memberjunction/ng-explorer-app/dist/styles.scss";   // ← already in your src/styles.scss
```
Bump your `@memberjunction/*` versions and rebuild; the tab chrome comes with it. **No manual edit — even though the change touches the non-distributed MJExplorer app, the styling itself ships through the distributed package you already import.**

**Category B — Standalone apps NOT on the Explorer shell (the BizApps apps, or any app that uses `mj-tabstrip` / `mj-workspace-tab-strip` without importing `ng-explorer-app` styles): ⚠️ one line required**, or both strips render **unstyled** (silent — no error):
```scss
// add to your app's GLOBAL stylesheet
@import '@memberjunction/ng-ui-components/dist/lib/tabs/tabs';
```
(@MarceloT-BC — this is already your action item for bizapps-accounting#44.)

### Change 2 — `mj-tabstrip` no longer hardcodes viewport height (behavioral, minor-version break)

The strip and its tab bodies used to fix their height to `calc(100vh - …)`. They now **size to content and the host owns region sizing** (which is how every other MJ widget works, and it fixes the strip blowing up any non-full-page host + scrolling the whole pane on focus). If a tab body previously relied on that fixed height for its **internal scroll**, it will now collapse to content height until you give it a definite height.

**Fix — put a height on the content inside the tab body** (concrete example, exactly what was done for the one in-repo consumer):
```html
<mj-tab-body>
  <mj-explorer-entity-data-grid
    [Params]="…"
    [Height]="600"          <!-- ← was implicitly ~viewport height; make it explicit -->
    [ShowToolbar]="false">
  </mj-explorer-entity-data-grid>
</mj-tab-body>
```
For non-grid content, set a height/`min-height` on the tab body's container instead. `FillWidth`/`FillHeight` inputs are now deprecated no-ops (kept so existing bindings compile).

### Who is actually affected

- **In this repo:** only the **Entity Actions form** used `mj-tabstrip` at runtime, and it's already fixed. The CodeGen tabstrip path is `@deprecated` and unused (0/383 generated forms). So a plain MJExplorer fork with **no custom `mj-tabstrip` usage** needs only the version bump.
- **Separate-repo apps most likely to need action:** the **BizApps UI apps** (accounting is confirmed via #44 — someone should also check **crm** and any other BizApps front-ends), since they build their own shell and/or use the strips directly → they need **both** Change 1 (Category B `@import`) **and** Change 2 (any `mj-tabstrip` height fix). In-repo, `packages/Actions/BizApps` is server-side only, so the BizApps *UI* lives in those external repos.
- **Dev-tooling only:** MJExplorer's new `package.json` dependency declarations (`@auth0/auth0-spa-js`, `ng-container-directives`, etc.) matter **only under pnpm's strict node_modules layout**; npm-hoisted forks won't notice. If a downstream app is on pnpm and hits a missing-dep dev-serve error, mirror those additions. Not a runtime concern.

**Net:** existing MJExplorer codebases do **not** need manual style edits (chrome flows through `ng-explorer-app`); they only need the `[Height]` fix *if* they have custom `mj-tabstrip` usage that relied on the old viewport height. Standalone/BizApps apps need the `@import` line plus any `[Height]` fix.

---

### Bottom line
This is a strong PR now. The a11y work Matt asked me to check is the highlight — a correct, well-tested, shared APG contract that fixes a gap **both** pre-existing strips had, in one place. My remaining items are all minor/non-blocking (dev-mode unstyled-strip warning, a follow-focus opt-out, `mjTip` hardening). I'm comfortable with this merging once the Definition-of-Done suites are confirmed green on the branch and the downstream owners above have the two changes on their radar. @MattC-BC @MarceloT-BC

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

MattC-BC (approved):

MattC-BC:
@rkihm-BC thanks for the second pass, especially the depth on the a11y layer. Responding to your three items:

**Dev-mode unstyled-strip warning: done, pushed as `5ffc69392b`.** Implemented per your sketch: `tabs.scss` sets a `--mj-tabs-chrome-loaded` sentinel on `.mj-tabs`, and a shared `warnIfTabChromeMissing()` helper (called by both strips after view init) checks the computed value and warns once per app with the exact `@import` line to add. Dev-mode only, never throws. The silent failure mode you flagged as the one genuine cost of the global stylesheet is now an actionable console message.

**Follow-focus opt-out (`mjTabListManual`): agreed on the shape, holding until a consumer needs it.** You confirmed the current path is safe for both real strips, and I would rather design the opt-out against the first expensive-activation host than a hypothetical one. It is a small additive input whenever that host shows up.

**mjTip hardening: deferred deliberately, bundled with the "workspace" rename** as one follow-up PR (scroll dismissal, the shadow-token gate bypass, and renaming it out of "workspace" belong in the same mechanical pass).

Also new since your review, pushed earlier today: the branch now carries a merge from `next` (conflict resolution was mechanical: their version bumps + our dep additions; interestingly `next` had independently added `ng-container-directives` to MJExplorer, so someone else hit that same pnpm resolution gap) and **mobile touch support for the workspace strip** (`5de236ad82`): a 400ms hold-to-drag on touch so a horizontal swipe scrolls an overflowing tab list instead of grabbing a tab, iOS-home-screen style, plus an `AllowReorder` input on strip and card as the disable-entirely escape hatch. Verified against CDK source that `touch-action` is only suppressed while a drag is actually in progress, so native panning genuinely survives.

**On your Definition-of-Done condition:** all unit tiers are green on the branch (ui-components 517, entity-viewer 152, dashboards 1695, tab-strip 13). The deterministic integration tier does not run on my machine yet because the suite metadata from the `next` merge is not in my local DB; CI on this PR is the arbiter there, and I will confirm it before merge.

MattC-BC:
@rkihm-BC closing out the Definition-of-Done condition from your review. The deterministic integration tier now runs on my machine (the suite metadata from the next merge needed an `mj sync push` into the local DB first): **52/56 pass, and every test that exercises what this branch touches is green** (UI seams, entity-grid transport, RLS, payload guards).

The three failures are all environmental or pre-existing, none in this branch's code paths:

1. **Codegen determinism**: my local DB is one entity ahead of the branch's generated ORM (`MJ: Authentication Providers`, from freshly pushed next metadata). Local sequencing drift; deliberately not "fixed" by running CodeGen here since that would regenerate against a DB state ahead of the branch's migrations. Will not reproduce on a fresh, correctly sequenced database.
2. **IT52 Unified Search**: lexical SearchEntity returns nothing for an existing name, which points at full-text catalogs not built on my local SQL Server. Nothing here touches search.
3. **IT70 External Agent Harness**: fails with the transport-sequencing bootstrap error that cites known issue #3251 in its own message. Pre-existing and tracked, and IT70 itself arrived with the next merge.

Unit tiers remain green everywhere (ui-components 517, entity-viewer 152, dashboards 1695, tab-strip 13). One caveat for transparency: this branch reports no CI check runs, so the numbers above are from the local tier rather than CI. If you want a fresh-DB confirmation of the three environmental ones before approving, happy to chase that down; otherwise I think we are good for your approval.

MarceloT-BC:
Reviewed the branch as it stands on the remote (53 files, +3,604/−201 across 7 commits). Two things worth resolving before merge, and one section on what held up under scrutiny.

## 1. The tabstrip height change didn't reach CodeGen

The changeset is admirably honest about the behavioural break: tab bodies "now size to content instead of hardcoding viewport height (`calc(100vh - …)`) — hosts that relied on the old fixed-height, internally-scrolling body should set a height on their own container."

That prediction came true inside this PR. The Entity Actions form needed a hand-fix, with a comment that states the failure mode precisely:

> Explicit Height: the default 'auto' resolves to height:100%, which needs a definite ancestor — the tab body used to hardcode viewport height and provide one, but it now sizes to content (the host owns region sizing), so 100% would resolve to nothing and the grid loses its internal scroll.

But `packages/CodeGenLib/src/Angular/angular-codegen.ts` — which emits `<mj-tabstrip>` into **every generated entity form** (`:1325`), containing `mj-explorer-entity-data-grid` blocks that likewise specify no `[Height]` (`:987-994`) — is untouched by this PR. So one hand-written form got the fix, while the generator producing hundreds of structurally identical forms, across MJ core *and* every Open App that runs CodeGen, did not.

To be precise about what I did and didn't establish: I confirmed the generator is untouched and that it emits grids with no `[Height]`. I did **not** confirm that generated forms actually regress — that turns on whether `mj-collapsible-panel` supplies a definite height in the generated structure, which I didn't trace. So this may be a non-issue. But the in-PR hand-fix is direct evidence the failure mode is real, and the generator is where the volume is, so it's worth confirming before merge rather than discovering it in generated output afterwards.

## 2. `"latest"` version specifiers in a published package.json

Commit `33bc6f7e0d` ("declare MJExplorer's externalized runtime dependencies") changed three pinned ranges in `packages/MJExplorer/package.json` to `"latest"`:

- `@codemirror/commands` (was `^6.10.4`)
- `@codemirror/language` (was `^6.12.4`)
- `@lezer/highlight` (was `^1.2.3`)

`latest` resolves differently on every install, which makes the build non-reproducible and means a breaking upstream release lands without any version change here. If the intent was to stop pinning these because they must match whatever the externalized runtime provides, a caret range on the current version gets that without giving up reproducibility.

## What held up

Worth saying explicitly, since the above is all criticism:

- **The tests are real, not liveness checks** — 141 assertions across four new files, including a DOM test for the keyboard directive and a genuine unit test for the tab state machine rather than a smoke test.
- **`mjTabList` is carefully built.** Roving tabindex is applied in `ngAfterViewChecked` rather than trusting the asynchronous MutationObserver, with the reasoning written down — that's the kind of detail that's usually discovered later as a focus bug. Editable-target passthrough, hidden-tab skipping, and writing `tabindex` only on an actual difference are all handled. I traced the all-tabs-hidden path specifically: `visible[NaN]` from the modulo and `visible[-1]` from `End` are both caught by the `if (!target) return` guard, so there's no crash there.
- **The overflow rework is complete.** Moving from an animated `left` offset to native `scrollLeft` would normally leave the scroll buttons stale when a user scrolls by trackpad — but `(scroll)="checkTabScrollButtons()"` is wired on the list, so the button state stays truthful on both paths.
- **`public-api` exports are complete** for everything the changeset claims the package now ships.

Review is static — I read the branch diff and surrounding source, and did not build, run the suites, or exercise the UI.

MattC-BC:
@MarceloT-BC thanks for the close read. I checked both against the branch — one is a genuine non-issue with evidence, the other I can't reproduce anywhere in the repo. Details so you can verify independently:

**1. CodeGen tabstrip height — confirmed non-issue.** You were right to flag it as unverified, and the missing piece is that `innerTabStripHTML` is **dead code**: it's defined at `angular-codegen.ts:1319` and never called anywhere (`grep -n "innerTabStripHTML" angular-codegen.ts` returns exactly one hit — the definition). Generated forms moved to collapsible panels: **0 of the generated forms contain `<mj-tabstrip>`**, and 385 of them use the collapsible-panel/form-section structure.

So the grids at `:987-994` never sat inside an `mj-tab-body` at all — they're inside collapsible panels, and this PR changed nothing about their height chain. Whatever their sizing behaviour is, it's identical before and after and predates this work. The Entity Actions hand-fix was needed precisely because that form is the **one hand-written** `mj-tabstrip` consumer, which is why it was the only one to touch.

Worth doing separately regardless: `innerTabStripHTML` is deprecated, uncalled, and now emits markup whose sizing contract has changed — it should probably just be deleted so nobody regenerates against it. Happy to do that as a follow-up.

**2. `latest` specifiers — I can't reproduce this.** All three are pinned carets everywhere I look:

```
packages/MJExplorer/package.json:35  "@codemirror/commands": "^6.10.4"
packages/MJExplorer/package.json:40  "@codemirror/language": "^6.12.4"
packages/MJExplorer/package.json:52  "@lezer/highlight": "^1.2.3"
```

Same values on `origin/feature/accounting-ui-hoist` and on `origin/next` (`git show origin/next:packages/MJExplorer/package.json`), and `grep -c '"latest"'` on the remote branch's copy returns **0**. And `33bc6f7e0d` didn't touch those lines — its entire MJExplorer diff is two additions (`@auth0/auth0-spa-js`, `ng-container-directives`) plus alphabetical re-sorting of four existing lines.

My guess at the source of the confusion: when I ran the deps guard earlier it printed `(warn) @codemirror/commands: every owner declares 'latest' — owners should pin a concrete range`. That warning is about the **owner** workspace packages that depend on codemirror, not about MJExplorer's manifest. Those warnings have since disappeared (the current run only warns about disjoint `date-fns` ranges), so the owners appear to have been pinned in the meantime — plausibly in the `next` merge that's part of this branch.

I agree completely with the principle, so if you're seeing `latest` somewhere concrete, point me at the file and line and I'll fix it immediately — I just can't find it in the tree.

**Agreed and noted:** the strip sizing to content is a real behavioural break, which is why the changeset leads with it and flags that hosts must supply their own height. If any Open App turns out to be affected, `[Height]` on the content (or a height on the tab body's container) is the fix, exactly as the Entity Actions comment describes.

MattC-BC:
@MarceloT-BC correction — I had a third party audit both points independently, and **you were right about what you saw on claim 2. My explanation of it was wrong.**

Those three specifiers genuinely *were* `"latest"` in MJExplorer's own manifest at `33bc6f7e0d`. My "you probably misread the guard's owner-package warning" theory was false, and I apologise for it — you inspected the tree at that commit and reported accurately. My mistake was checking only whether they're `latest` *now* and never checking whether they were `latest` *then*.

Full provenance: they were **inherited, not introduced by this branch**. `8d6160645b` wrote them via the deps guard in `--write` mode, because owner `@memberjunction/ng-code-editor` declared `latest` for five CodeMirror packages. Craig fixed it on `next` in `ea4000036` ("collapse CodeMirror to one instance, remove every 'latest' specifier"), which reached this branch through the merge commit `40c2c349ea`. So the attribution to `33bc6f7e0d` was off, but the observation was sound and the tree is only clean today by way of that merge.

**Your underlying concern is right and still live.** The audit found the guard script can do this again and can never undo it: `pickRange()` returns `latest` when every owner declares it, `writeDeclarations()` writes that into MJExplorer's dependencies, and it's a one-way ratchet — `parseRange('latest')` yields `kind: 'any'`, `rangesIntersect` returns true unconditionally for `any`, so an existing `latest` is never flagged incompatible and never repaired. Only a non-gating warning surfaces it. That's exactly the mechanism that produced the original entries, and it's unpatched. I'm opening a follow-up to gate on it.

On **claim 1**, the audit reached the same conclusion I did, with two points I hadn't established: no config flag can select the tabstrip path (both HTML entry points call `innerCollapsiblePanelsHTML`), and the dead path is *incoherent* rather than merely unused — every `TabCode` producer now emits `<mj-collapsible-panel>` while `MJTabStripComponent` finds children via `@ContentChildren(MJTabComponent)`, so calling it would yield an empty strip with no `mj-tab-body` at all. Separately, the container generated forms actually use has a definite height (`collapsible-panel.component.css:205-212`, `height: 400px` with `> *` at 100%), so those grids never depended on a viewport ancestor.

One residual the audit raised that neither of us had: an Open App whose forms were generated before `6f88c60dee` (Nov 2025) has tabstrip HTML checked into its repo, and upgrading `ng-tabstrip` without re-running CodeGen would meet the new content-sized CSS. Unlikely under lockstep versioning, not structurally prevented, and unverifiable from inside this repo — worth a heads-up to Open App owners.

Follow-up PR incoming with the guard-script gate and deletion of the dead `innerTabStripHTML`. Thanks for pushing on both of these.

Issue #3451, opened 2026-08-04
Open App install runs all migrations in one transaction, which SQL Server cannot always host (CREATE TYPE + TVP self-deadlock, 1205)


MarceloT-BC:
## Summary

Open App installs run the entire pending migration set inside **one database transaction** (Skyway's `per-run` default), and use that transaction as the install's rollback mechanism. SQL Server cannot host that transaction in the general case: **a transaction that creates a table type and then instantiates a variable of that type deadlocks against itself** (error 1205). On a from-zero install every migration is pending, so the whole app is one transaction and *no arrangement of migration files avoids it*.

The result is that an app which ships a table type and uses it — a standard T-SQL pattern — **cannot be installed from a clean database**, and it fails as an opaque deadlock rather than an actionable error.

This is not a bad migration. The minimal reproduction is two statements with no MemberJunction involved.

## Minimal reproduction

Run against any SQL Server database. No MJ, no Skyway, no migration files.

```sql
BEGIN TRAN;
    CREATE TYPE dbo.WidgetIDList AS TABLE (ID UNIQUEIDENTIFIER NOT NULL PRIMARY KEY);
GO
    DECLARE @ids dbo.WidgetIDList;   -- Msg 1205: deadlocked on lock resources
GO
COMMIT;
```

Verified on **SQL Server 2022 RTM-CU25 (16.0.4255.1)**, deterministic across repeated runs.

**Mechanism.** The `CREATE TYPE` holds a Sch-M lock on the type for the life of the transaction. Instantiating a table-valued variable runs in a **nested system transaction** that does not share the session's lock ownership, so it requests Sch-S on that same type and waits on its own session's Sch-M. Single-session self-deadlock — it reproduces with no other connection to the server. A captured `system_health` deadlock graph from the real failure shows exactly this: owner and waiter are the same process, resource `USER_TYPE`, waiter a system transaction named `@ids`.

## Scope — what does and does not trigger it

| Scenario (all inside one transaction) | Result |
|---|---|
| `CREATE TYPE … AS TABLE` + bare `DECLARE` of that type | **1205** |
| … + a statement trigger that declares the TVP, fired by DML | **1205** |
| … + a stored procedure that declares the TVP, then `EXEC` | **1205** |
| … + a stored procedure with a **TVP parameter**, then `EXEC` passing one | **1205** |
| Ordinary `CREATE TABLE` + `INSERT` + `SELECT` | ✅ commits |
[...]

PR #3469, opened 2026-08-05
fix(open-app-engine): per-migration transactions for Open App installs, and full compensation on failure

MarceloT-BC:
Fixes the transaction semantics of Open App migrations, and makes a failed install clean up after itself. Two commits, both scoped to `@memberjunction/open-app-engine`.

Background and the full analysis are in #3451.

## 1. Migrations apply per-migration by default

`RunAppMigrations` built its Skyway config without ever setting `TransactionMode`, and the option was declared only on the module's *internal* Skyway config interface — never on the public `MigrationRunOptions`. No caller could select a mode even in principle, so every Open App install fell through to Skyway's `per-run` default: **one transaction wrapping the entire pending migration set**.

`MigrationRunOptions` now accepts `TransactionMode` and threads it onto the config handed to Skyway, defaulting to **`per-migration`**. `per-run` remains available opt-in.

**Why the default changes.** `per-run` cannot host every valid migration set. SQL Server cannot create a table type and instantiate a variable of that type in the same transaction — the `CREATE TYPE` holds a schema-modification lock while TVP instantiation, which runs in a nested system transaction that does not share the session's lock ownership, requests schema-stability on the same type. The session deadlocks against itself. Two statements, no MemberJunction:

```sql
BEGIN TRAN;
    CREATE TYPE dbo.IDList AS TABLE (ID UNIQUEIDENTIFIER NOT NULL PRIMARY KEY);
GO
    DECLARE @ids dbo.IDList;   -- Msg 1205
GO
COMMIT;
```

On a from-zero install every migration is pending, so under `per-run` the whole app is one transaction and **no arrangement of migration files avoids it** — making a table type, a standard T-SQL construct, impossible to ship. It surfaces only on a clean install; an incremental development database committed the type in an earlier run, so its lock is long released. Verified on SQL Server 2022 RTM-CU25.

This also ends a silent divergence: MJCLI already defaults `mj migrate` to `per-migration`, so the same repo had two migration paths with different transaction semantics.

## 2. A failed install now removes everything it wrote

Per-migration means a set that fails partway leaves earlier files committed — the database will not undo them, and it cannot. So the install's all-or-nothing guarantee has to be a compensating action rather than a transaction.

`CompensateSchemaOnFailure` previously did one of the three things `RemoveApp` does: it dropped the app's schema. Rows the app's seed migrations wrote into the **shared** core schema were left orphaned, because dropping the app's own schema cannot reach them. It now runs the same three-step sequence, in the same order:

1. `RemoveAppEntityMetadata` — the app's entity metadata and the Application rows its migrations declared.
2. `HandleTeardown` — the app's declared `migrations.teardownDirectory` inverse DELETEs.
3. `DropAppSchema` — the app's own schema, which takes its migration history table with it, so a retry starts clean rather than resuming a half-applied set.

Unchanged: this runs only for a schema **the run actually created**, never a reused or adopted one. Because the run created it, no other installed app can legitimately share it, so no co-tenant check is needed here (unlike `RemoveApp`, which can be asked to remove an app from a shared schema).

Every step is best-effort and reported. One failing step does not skip the others — a teardown failure must not leave the schema behind — and nothing here turns a failed install into a successful one. An app that declares migrations but **no** `teardownDirectory` now warns explicitly that rows in the shared core schema may remain, rather than letting a partial rollback look complete.

## Discussion — should the install path share `mj migrate`'s implementation?

Worth deciding separately from this PR, because the divergence fixed above was a symptom of it.

`mj app install` does **not** call `mj migrate` under the hood, and cannot: `RunAppMigrations` dynamically imports `@memberjunction/skyway-core` and constructs Skyway itself, while MJCLI's migrate builds its **own** Skyway config independently (`packages/MJCLI/src/config.ts` vs `packages/OpenApp/Engine/src/install/migration-runner.ts`). Two parallel config builders, which is precisely how they came to disagree about transaction mode — one had a documented, user-configurable default; the other silently had none.

The dependency runs **MJCLI → open-app-engine** (declared in MJCLI's `package.json`), never the reverse, so the engine invoking `mj migrate` would be circular. The achievable unification is a **shared Skyway-config builder** — one place that decides baseline semantics, placeholders, platform/provider selection, and transaction mode — consumed by both paths. This PR does not attempt it; it only stops the two from disagreeing on the setting that was actively breaking installs.

Related open question: if transaction mode is expressed at all beyond the default, does it belong to the *installing host* (CLI config) or to the *app manifest*, given the app is what knows whether it ships a table type? This PR deliberately does not answer that — it changes the default and leaves the option available to programmatic callers.

## Testing

`@memberjunction/open-app-engine`: **330/330 pass, 26 files** (318/318 across 25 files before this change; no pre-existing test modified). Real `tsc` build green.

- 7 new specs in `migration-runner-transaction-mode.test.ts` mock `skyway-core` and capture the config actually handed to `new Skyway(...)`. Against unmodified `next` the transaction-mode assertions fail with `expected undefined` — the key was never present — and pass after.
- 5 new specs in `install-orchestrator-order.test.ts` cover compensation. Three assert the new behavior and fail without this change; the other two are regression guards on drop-schema behavior that already existed and pass both ways.

Beyond unit tests, every behavior this PR claims was exercised against real databases and instances:

| Behavior | Verified by |
|---|---|
| The deadlock is a SQL Server property, not a bad migration | 2-statement repro plus 6 controls on SQL Server 2022 |
| **`per-run` deadlocks where `per-migration` succeeds** | **A/B through the real `RunAppMigrations` — same set, same code, only the mode differs** |
| Teardown SQL survives a partially-seeded state | 4 cases against a real migrated core schema |
| Removal leaves FK-unreachable seeds behind | install → remove on a real instance, 215 rows, control returns to baseline |
| **Failed install rolls back — migration failure** | **real `mj app install`, all 3 compensation steps observed** |
| **Failed install rolls back — post-migration failure** | **real `mj app install`, migrations committed then record step failed; still cleaned up** |

### The transaction mode, A/B through the real migration runner

Identical migration set — a table type created in one file; a trigger declaring a variable of that type plus a CodeGen-style `UPDATE ... WHERE col IS NULL` backfill in the next — run twice through the real `RunAppMigrations` against a real database, changing nothing but `TransactionMode`:

| Run | Mode | Result |
|---|---|---|
| A | `per-run` (before this change) | **fails — 1205, deadlocked and chosen as victim** |
| B | default `per-migration` (after) | **succeeds — 2 migrations applied** |

Driving `bizapps-orders` itself was attempted first and was **inconclusive**: its migrations fail at batch 82 on foreign keys into three dependency apps that were not installed, so execution never reached the deadlock at batch 285. The fixture above reproduces that app's pattern without its dependency chain; engine, Skyway and database are real.

### Both compensation call sites, end-to-end on a real instance

`CompensateSchemaOnFailure` has two call sites; both were driven with a real `mj app install`, each failure forced deterministically with a temporary trigger.

**Failure during migrations** — migration 5 of 7 fails *after* the app's ~220-row seed migration commits:

```
- [Migration] Running 7 migration(s)...
- [Rollback] Rolling back failed install of 'mj-bizapps-sonar'...
- [Metadata] Teardown plan: delete ApplicationEntity.EntityID x14; delete EntityFieldValue.EntityFieldID x83;
  delete EntityField.EntityID x235; delete EntityPermission.EntityID x42; delete EntityRelationship.EntityID x28;
  delete QueryEntity.EntityID x6; delete Entity x14; delete SchemaInfo x1 (total 497 row(s))
✔ [Metadata] Entity metadata for schema '__mj_BizAppsSonar' removed
  Warning: [Rollback] declares migrations but no teardownDirectory — rows may remain
- [Rollback] Schema '__mj_BizAppsSonar' dropped successfully
✖ Install failed: Migration failed ... FORCED FAILURE
```

**Failure after migrations** — all 7 migrations commit, then recording the installation fails: compensation removed 490 metadata rows, emitted the same warning, and dropped the schema.

In **both** runs everything the compensation is responsible for returned exactly to its pre-install state:

| control | before | after failed install |
|---|---:|---:|
| Entity | 383 | **383** |
| EntityField | 5577 | **5577** |
| Application | 26 | **26** |
| ApplicationEntity | 402 | **402** |
| app schema present | no | **no** |

In the migration-failure run, 193 core-schema seed rows remained — the app declares no `teardownDirectory`, the case the new warning announces. That number is self-validating: 215 seeded rows minus the 22 belonging to the migration that failed, whose own transaction rolled back. **That the other 193 persisted is itself evidence per-migration is in force** — under `per-run` all five would have rolled back together and no partial state could exist.

### Teardown against a partially-seeded state

Compensation runs teardown scripts generated for a *complete* seed against a state where only some seed migrations committed, so that assumption was proven rather than assumed — replicating `HandleTeardown`'s execution model exactly (filename order, `${mjSchema}` substitution, all scripts in one transaction, rollback on any error):

| Case | Result |
|---|---|
| Rows that were **never inserted** — the partial-install case | **committed**, no error |
| Fully seeded — the normal remove case | committed, 3 → 0 rows |
| **Superset teardown, only 1 of 3 targets seeded** | **committed, 1 → 0 rows** |
| A DELETE blocked by an FK from a row teardown does not know about | **rolled back (547), 3 → 3 rows** |

Inverse DELETEs against absent rows are no-ops, so partial state is safe; and a genuine teardown failure leaves the core-schema rows fully intact rather than half-retired, with compensation reporting it and still dropping the schema.

One observation from the last case, not acted on here: compensation runs teardown *before* `DropAppSchema`, mirroring `RemoveApp`. A teardown blocked by an FK from a row in the app's **own** schema would have succeeded had the schema been dropped first. The order was kept deliberately, since teardown scripts may reference app-schema objects — but it may be worth revisiting.

### Remaining limitation

**Step 2 of the compensation has never executed its real work in a live run**, because no app currently ships a `teardownDirectory` — every live compensation took the warning branch. Teardown's SQL behavior is proven directly against a real database (above), but teardown scripts running *inside* a compensation has not happened, for lack of any such scripts anywhere. The first app to adopt one is the moment to exercise it.

The end-to-end runs were performed on **v5.51.0** with these commits cherry-picked, because the installable apps declare `mjVersionRange <6.0.0`. The two touched files are byte-identical between v5.51.0 and `next`, so the code under test is the code in this PR.

### Adoption note

No app currently declares `migrations.teardownDirectory` — and for almost all of them that is the correct choice rather than an omission. Across every open-app migration set, what apps write into the core schema is entity and application metadata: `Entity`, `EntityField`, `EntityFieldValue`, `EntityPermission`, `EntityRelationship`, `EntitySetting`, `Application`, `ApplicationEntity`, `ApplicationRole`, `SchemaInfo`, `GeneratedCode`. All of that is removed by step 1 with no app cooperation — `RemoveAppEntityMetadata` deletes the app's entities, SchemaInfo and declared Applications, and the FK-graph teardown deletes everything transitively reachable from them. The manifest reference already says as much: *"Omit for apps whose teardown is fully covered by dropping their own schema."*

The residual class is core-schema rows with no FK path back to the app's entities — Action, AI, Query and Template metadata, which floats free of the app's entity tree. A survey of 13 open apps found **three** that seed it and declare no `teardownDirectory`: one seeds roughly 220 such rows, another ~49, another 5. The remaining ten write only entity/application metadata and are fully covered by step 1.

For those three, `teardownDirectory` is the declared mechanism and the fix belongs in the app repos; `mj app remove` already depends on it in exactly the same way, so this is a pre-existing gap that this PR inherits rather than introduces. What changes is that it is now announced: an app with migrations and no teardown directory gets an explicit warning instead of a partial rollback that looks complete.


packages/OpenApp/Engine/src/install/migration-runner.ts
```diff
@@ -223,7 +247,8 @@ export function BuildSkywayConfig(
     dbConfig: SkywayDatabaseConfig,
     mjCoreSchema?: string,
     extraPlaceholders?: Record<string, string>,
-    platform: DatabasePlatform = 'sqlserver'
+    platform: DatabasePlatform = 'sqlserver',
```
AN-BC, line 260: does this mean OpenApp won't install on pg?

AN-BC (commented):
I don't have any useful input. for anything related to openapp/skyway/MJCLI/installer tag @rkihm-BC for him to review and he'll pull in @cadam11  on some things as needed. 

Use me for review on more software architecture related stuff like `BaseEntity`, caching, AI sub-systems, etc @MarceloT-BC .

Thanks for getting this raised, very important fix!

packages/OpenApp/Engine/src/install/migration-runner.ts
MarceloT-BC: No — Open App installs are unaffected on PostgreSQL. That `= 'sqlserver'` is a backward-compat fallback for direct callers of `BuildSkywayConfig`, and it is never reached on the install path:

- `HandleMigrations` reads the **live** database's platform and passes it explicitly: `const platform = context.DatabaseProvider.Dialect.PlatformKey;` → `RunAppMigrations({ ..., Platform: platform })`.
- `RunAppMigrations` forwards it (`options.Platform ?? 'sqlserver'`) into `BuildSkywayConfig`, which uses it to pick the Skyway provider (`@memberjunction/skyway-postgres` vs `-sqlserver`) and to canonicalize the schema name for PG's lowercase folding.

So on a PG instance the parameter is always `'postgresql'` and the default never applies.

On the change itself: `TransactionMode` is platform-independent. Nothing in this PR branches on platform — Skyway's `executePerMigrationWithHistory` goes through `provider.BeginTransaction()` / `provider.History.*` only, and the Postgres provider implements those the same as SQL Server.

Worth adding for context: **MJCLI already defaults `transactionMode` to `per-migration` on both platforms**, so `mj migrate` has been running per-file on PG all along. This PR brings the install path into line with what PG users already get from the CLI; it isn't introducing a new mode there. And the deadlock that motivated the default is SQL-Server-specific (user-defined table types / TVPs), so for PostgreSQL this is purely an alignment, not a behavior fix.

Two honest caveats:

1. **I have not executed a PG install end-to-end** — the above is traced through the code and verified against Skyway's implementation, not observed on a live PostgreSQL instance. Everything I did run was SQL Server. If you'd like that verified before merge, say so and I'll stand one up.
2. While tracing this I noticed a genuine PG divergence that is **not** introduced here but is worth a follow-up: the engine canonicalizes the schema name for PG (`GetDialect(platform).CanonicalSchemaName`), while MJCLI's own `buildSkywayConfig` does **not**. On PostgreSQL, `mj migrate --schema MixedCase` and an app install would therefore target different physical schemas. That belongs with the broader "the two paths build their Skyway config independently" cleanup.

Happy to make `Platform` required rather than defaulted if you'd prefer no silent fallback at all — it's a small breaking change to an exported helper, so I left it alone here.

rkihm-BC:
## Code review — approve with follow-ups tracked

Reviewed against the actual source (not just the description). Summary: **the transaction-mode fix is correct, minimal, and unusually well-evidenced; approve it unreservedly.** The compensation half is the right design and faithfully mirrors `RemoveApp`, but its newest behavior is the least-tested part and there's a reinstall trap worth sharpening before this pattern goes live. Detail below, plus answers to the two design questions raised and a reproduction check on the Q1 trap.

### What's good

- **Root cause is correct and load-bearing.** The `CREATE TYPE` (schema-modification lock) vs. TVP instantiation (schema-stability from a nested system transaction that doesn't inherit the session's lock ownership) → self-deadlock (1205) is a genuine SQL Server property, and the conclusion that *no arrangement of files* avoids it under `per-run` on a from-zero install follows. A table type was literally un-shippable before this.
- **The fix is surgical.** `TransactionMode` existed only on the internal Skyway shape, never on public `MigrationRunOptions`, so no caller could select it and Skyway silently fell to `per-run`. Threading it through with a `per-migration` default is the right shape. Confirmed the install path relies on the new default correctly.
- **Ends a real divergence** — confirmed MJCLI already defaults `mj migrate` to `per-migration` (`config.ts:34` + Zod defaults). The repo genuinely had two install paths disagreeing on transaction semantics.
- **Compensation now actually mirrors `RemoveApp`** — diffed both: metadata → teardown → drop-schema, same order. Previously it did only the third step; the "shared core-schema rows orphan because dropping the app's own schema can't reach them" argument is correct.
- **`schemaWasCreated` gate preserved and correctly reasoned** (no co-tenant check needed since the run created the schema).
- **Evidence quality is exceptional** — real A/B through `RunAppMigrations`, both compensation call sites driven end-to-end with row-count controls, and honest disclosure of what wasn't verified.

### What's concerning

1. **Teardown-inside-compensation has zero real coverage — unit or live.** The new specs stub `DownloadMigrations` to return `Files: []`, so `HandleTeardown` hits its early "no scripts" return before executing any SQL. The tests prove the teardown *path is entered*, not that teardown DELETEs run correctly inside a compensation. Combined with "step 2 has never executed its real work in a live run," the compensation's most consequential new behavior is the least-exercised code in the PR — and `per-migration` makes compensation the *only* safety net. Acceptable to merge (no app ships a `teardownDirectory` yet), but should be a tracked follow-up gated on the first adopter.
2. **Partial compensation can trap a reinstall** — see the Q1 reproduction below; it's broader than it first looks.
3. **Rollback now depends on a network round-trip** — `HandleTeardown` fetches teardown scripts from GitHub *during rollback*. If the install failed for a network reason, teardown download also fails (reported; schema still drops). Worth a comment; argues for reusing the already-downloaded tree.
4. **Teardown-before-drop ordering** is a known latent corner (a teardown DELETE blocked by an FK from a row in the app's *own* schema would have succeeded had the schema dropped first). Flagged in the PR, kept deliberately — fine to defer.
5. **PostgreSQL unverified end-to-end** (author's own caveat). Reasoning is sound (mode is dialect-independent; the deadlock is SQL-Server-specific), low risk.

### What's missing

- No deterministic **integration-tier** test (verified manually on v5.51.0 cherry-picks). Reasonable given `<6.0.0` constraints, but CI has no end-to-end regression guard.
- No unit test for the **record-failure** compensation call site (only the migration-failure site is covered by new specs).
- No **warning/record on a partial compensation** (concern #2 / Q1).

---

### Design Q — Platform default (dialect fallback)

**Decision: keep the `= 'sqlserver'` fallback, but warn when `Platform` is unset.** One nuance: on the real install path `Platform` is *always* set explicitly (`HandleMigrations` reads `context.DatabaseProvider.Dialect.PlatformKey`), so the fallback is unreachable during an actual `mj app install`. The warning therefore fires only for **direct programmatic callers of the exported `RunAppMigrations`/`BuildSkywayConfig` helpers** — exactly the population that could silently get SQL Server semantics against a Postgres DB. Cheap insurance, zero noise on the happy path. Implementation: warn in `RunAppMigrations` when `options.Platform === undefined` before applying `?? 'sqlserver'`; keep the defaulted signature so it stays non-breaking.

### Design Q — Who owns TransactionMode: manifest vs. host

**Decision: the app manifest owns the default/requirement** — the app is the only party that knows it ships a table type. On whether a *host* would ever need to override, three legitimate cases exist, but **all are "escape hatch," none are "the host is the authority":**

1. **Engine-capability flooring** — the host knows the *engine's* transaction capabilities; may need to *force* `per-migration` where the engine can't host a whole-app transaction. Safe direction, unobjectionable.
2. **Support / repro** — forcing a mode to reproduce a customer issue or push a stuck install past a partial state. Diagnostic → a CLI flag, not config.
3. **Dev-time iteration** — an author wanting all-or-nothing locally.

The unifying principle: the manifest declares a **floor/requirement**; a host may override for capability or diagnostics but must **never be silently allowed to force `per-run` onto an app that requires `per-migration`** — that reintroduces the exact 1205 deadlock this PR fixes. Recommended model: manifest field = source of the default (optional, additive → no-break-safe); host override = a diagnostic flag layered on top, with the dangerous direction (→ `per-run`) either disallowed or requiring an explicit "I understand this may deadlock" acknowledgment. The asymmetry (per-run can break, per-migration can't) is what tells you ownership belongs to the app.

---

### Q1 trap reproduction — it reproduces, and it's broader than first framed

Traced how open-app metadata reaches the core schema. **The trap does not require step 1 (metadata removal) to fail — it reproduces on the *happy* compensation path.**

Two classes of core-schema rows an app writes:

- **Entity tree** (`Entity`/`EntityField`/`ApplicationEntity`) — for schema-bearing apps this is **generated out-of-band by the host's CodeGen run** after install (`install-orchestrator.ts:500`: *"entity metadata is generated by CodeGen"*), matched by name+schema, so no fixed-GUID collision. **Not** the trap.
- **Free-floating metadata** — Action/AI/Query/Template rows the app's seed migrations write **directly with fixed GUIDs** via `EXEC spCreateX @ID='<fixed>'`. The generated `spCreate*` procs are **unconditional INSERTs** — confirmed in `V202506131212__…New_spCreate…sql`: `IF @ID IS NOT NULL … INSERT … VALUES(@ID,…)`, **no `IF NOT EXISTS`, no `MERGE`.** Re-executing with the same fixed `@ID` throws a PK violation.

Reproduction (compensation itself succeeds throughout):

1. App seeds ~220 free-floating AI/Action rows (fixed GUIDs), a later migration fails partway.
2. Compensation runs cleanly: step 1 removes the entity tree (does **not** reach the free-floating rows — the entire reason `teardownDirectory` exists), step 2 **warns** (no `teardownDirectory`), step 3 drops the schema.
3. The ~220 fixed-GUID rows **persist** — the PR's own e2e proves this (*"193 core-schema seed rows remained"*).
4. Reinstall: schema recreated, migrations rerun from a fresh `flyway_schema_history`, the seed migration re-executes `EXEC spCreateX @ID='<same fixed guid>'` → **PK violation → reinstall fails at that seed migration.**

So the "193 rows remained" line the PR presents as *evidence per-migration is working* is simultaneously **evidence a subsequent reinstall of that same app will PK-collide** until those rows are manually removed.

**Is it new?** Mostly pre-existing, slightly widened:
- The **post-record-failure** orphan (all migrations commit, record step fails) already orphaned these rows before this PR (old compensation only dropped the schema) — pre-existing.
- This PR **newly** introduces the **partial-migration-failure** orphan, because `per-migration` commits earlier seed migrations before a later one fails, whereas `per-run` rolled the whole set back and left nothing to collide with. The window widens from "post-migration failures only" to "any failure after the seed migration commits."

Not blocking this PR — no such app is installable under `<6.0.0` yet, and the real fix belongs in the app repos (declare `teardownDirectory`). But the warning text undersells it. *"Rows may remain"* reads as cosmetic; the operational truth is **"rows will remain and will block a reinstall with a PK violation until manually cleaned."**

**Suggested follow-ups (your call on path forward):**
- Sharpen the no-`teardownDirectory` warning to state a reinstall may fail until the orphaned rows are removed.
- File an issue to make the seed `spCreate*` path idempotent (or have compensation record the orphaned fixed GUIDs), so a retry isn't a manual DB cleanup.
- Track the "teardown-inside-compensation never exercised" gap, gated on the first app that adopts a `teardownDirectory`.

---

**Bottom line:** ship the transaction-mode change as-is; merge the compensation change with the three follow-ups filed (the reinstall trap being the one most worth a sharpened warning before this pattern is live). Nice work on the evidence and the honest disclosure of limits.

*Review by Claude (Opus 4.8) at the request of @rkihm-BC; findings verified against the source, not the description.*

MarceloT-BC:
Thanks — that's a genuinely useful review, and the Q1 trap is the right catch. All three code follow-ups are in `5b6c38d988`, now the PR head.

**1. Sharpened the orphan warning.** You're right that "rows may remain" read as cosmetic. I verified the mechanism before rewording — `spCreateAction` in `B202607091514__v5.46.x__Baseline.sql` is `IF @ID IS NOT NULL … INSERT`, no existence guard, no MERGE — so the operational truth is a PK violation on reinstall. The warning now reads:

> `'<app>' declares migrations but no teardownDirectory — rows its migrations wrote to the shared core schema WILL remain after this rollback. If those rows were seeded with fixed IDs, reinstalling this app may fail with a primary-key violation until they are removed manually.`

I also took your framing on the widening: the code comment now records that `per-run` used to roll the seeds back, so this PR moves the window from "post-migration failures only" to "any failure after the seed migration commits."

**2. `Platform` warning added**, exactly as you scoped it — warn in `RunAppMigrations` when `options.Platform === undefined`, before the `?? 'sqlserver'` fallback, with the defaulted signature kept so it stays non-breaking. Unreachable on the install path (`HandleMigrations` always passes `Dialect.PlatformKey`), so it fires only for direct programmatic callers of the exported helper.

**3. Tests added** for the record-failure compensation call site (previously proven end-to-end but not unit-tested) and for the new warning text. 332/332.

**Follow-ups filed:**
- #3505 — make the generated `spCreate*` explicit-`@ID` path idempotent. This is the real fix for the reinstall trap.
- #3506 — the teardown-inside-compensation gap, gated on the first app that adopts a `teardownDirectory`. I folded your other two observations into it: the rollback network round-trip, and the teardown-before-drop ordering.

On that last one — I'd tested it before you raised it, and it's a symmetric trade-off rather than a latent corner. Against a real database: teardown-first fails 547 when a core row is pinned by an FK from the app's own schema (and since all teardown files share one transaction, the whole teardown rolls back); drop-first fails 208 when a teardown script reads an app-schema table. Each order fails exactly where the other succeeds, so the fix is to retry rather than reorder — attempt teardown with the schema intact, and on failure drop and attempt once more. Teardown is idempotent by construction, so the retry is free. That's implemented on a follow-up branch, held back to keep this PR reviewable.

**On your concern #1 — agreed, and worth being blunt about it.** The teardown step in compensation has never executed a line of SQL, and the tests don't close that (they stub the download to return zero files, so `HandleTeardown` early-returns; they prove the path is *entered*, not that it works). What *is* verified against a real database is teardown's execution model in isolation — filename order, `${mjSchema}` substitution, one transaction, rollback on error, and that inverse DELETEs against never-inserted rows are no-ops so partial state is safe. The untested part is precisely that path running inside a real failed install. #3506 exists to make sure its first execution isn't in production.

**One thing your review surfaced that I think is bigger than this PR.** Removal is engine-derived for the entity tree only, because `Entity` carries `SchemaName`. The free-floating rows have no ownership marker at all — I checked, `OpenAppID` exists only on `OpenAppDependency` and `OpenAppInstallHistory`, never on `Action`/`AIAgent`/`Query`/`Template`. So complete removal currently depends on an app author writing a correct teardown script: opt-in, unverified, a second hand-maintained source of truth, and discovered at uninstall time. `spCreate*` idempotency fixes the reinstall collision but not the residue.

Two designs would take the human out of that loop, both MJ-core: install-time footprint capture (snapshot core keys around the app's migrations, store the delta on the `OpenApp` record — no schema change), or an owning-app column on the core tables apps may seed. A cheaper middle step is detection alone: scan an app's migrations for writes to core tables and warn or fail at publish/install when no `teardownDirectory` is declared — that kills the silent case, which is the common one. Happy to open a design issue if you think it's worth pursuing.

rkihm-BC:
## Re-review of `5b6c38d988` — all three follow-ups verified against source; this is now mergeable

Checked the commit, not just the summary. All three land correctly.

**1. Sharpened orphan warning ✓** — `install-orchestrator.ts` now states the operational truth (*"rows … WILL remain … reinstalling this app may fail with a primary-key violation until they are removed manually"*), and the code comment records the fixed-GUID / unconditional-`spCreate*` mechanism. Exactly the intent. The `may remain` → `WILL remain` assertion update is a legitimate spec change, not a relaxed test.

**2. Platform warning ✓** — added in `RunAppMigrations` before the `?? 'sqlserver'` fallback, scoped precisely as discussed (unreachable on the install path since `HandleMigrations` always passes `Dialect.PlatformKey`; fires only for direct helper callers). Worth noting the `console.warn` choice is **consistent with this file's existing convention** — the `Verbose` branch already uses `console.log` (migration-runner.ts:150-152) — so it's not an inconsistency to flag.

**3. Tests ✓** — the record-failure call site is now genuinely covered (`RunAppMigrations` succeeds, `RecordAppInstallation` rejected → asserts teardown attempted + `DropAppSchema` called), plus a dedicated assertion pinning both `WILL remain` and `primary-key violation` in the warning text. 332/332.

The two issues (#3505 idempotent `spCreate*`, #3506 teardown-in-compensation + rollback network round-trip + ordering) capture the deferred work correctly. No objection to shipping this PR with those tracked.

### On the teardown-before-drop ordering

Your empirical result (547 when a core row is pinned by an app-schema FK vs. 208 when a teardown script reads an app-schema table — each order failing exactly where the other succeeds) is a better characterization than my "latent corner," and **retry-rather-than-reorder is the right call**: teardown-with-schema-intact, then drop, then one more teardown attempt, leaning on teardown's idempotence (inverse DELETEs against absent rows are no-ops) to make the retry free. One edge for the follow-up branch: if the first attempt fails for a *non-FK* reason (a genuine script error), the post-drop retry fails again and you land in the orphaned-but-reported state without the warning branch firing — fine under best-effort semantics, just worth an explicit log line so that case isn't silent. Holding it back to keep this PR reviewable is the right instinct.

### On the bigger design point — I verified the premise, and it holds

Confirmed against the generated ORM: `OpenAppID` exists **only** on `MJOpenAppDependencyEntity` and `MJOpenAppInstallHistoryEntity`. `Action`, `AIAgent`, `Query`, `Template` (and the rest of the free-floating core tables an app may seed) carry **no ownership marker at all**. So your framing is exactly right: complete removal of the free-floating class currently depends on a hand-written, opt-in, unverified teardown script discovered only at uninstall time. `spCreate*` idempotency fixes the reinstall *collision* but not the *residue* — those are two different problems and both are real.

**Yes, open the design issue.** My read on the three options:

- **Detection-only (scan migrations for core-table writes; warn/fail at publish or install when no `teardownDirectory`)** — do this first. It's the cheapest, needs no schema change and no runtime capture, and it kills the *silent* case, which is the common and most dangerous one (an author who simply didn't know they seeded core rows). It converts "discovered at uninstall" into "caught at publish."
- **Install-time footprint capture (snapshot core keys around the app's migrations, store the delta on the `OpenApp` record)** — the stronger durable answer, and it elegantly sidesteps the no-break policy by adding nothing to the core tables. The hard part is defining the capture boundary crisply: what counts as "the app's" delta when a migration triggers downstream writes, and how it behaves under concurrent installs. Worth prototyping, but it's the more expensive path.
- **Owning-app column on core tables** — conceptually cleanest, but the retrofit is the problem: a nullable column is additive and policy-safe to *add*, yet every existing seed path and every already-installed row would need backfilling for it to be trustworthy, and until then removal can't rely on it. I'd rank this third for that reason.

Net: detection-only now as the immediate silent-case killer, footprint-capture as the durable direction, the ownership column as a distant third. Happy to review that design issue when you open it.

### Verdict

The transaction-mode fix was already solid; the compensation half is now honest about its one real operational hazard, has the record-failure site under test, and the deferred work is tracked in #3505/#3506. **From my side this is ready to merge.** Good, responsive iteration — and the ownership-marker gap you surfaced is genuinely the more important long-term finding to come out of this.

*Re-review by Claude (Opus 4.8) at the request of @rkihm-BC; changes and the `OpenAppID` claim verified against source.*

rkihm-BC (approved):
## Approved ✅

Verified the full change set against source across two review passes. The transaction-mode fix is correct, minimal, and well-evidenced; the compensation half faithfully mirrors `RemoveApp` and — after `5b6c38d988` — is now honest about its one real operational hazard (fixed-GUID reinstall collision), has the record-failure call site under test, and warns on unset `Platform`. Deferred work is properly tracked in #3505 (idempotent `spCreate*`) and #3506 (teardown-in-compensation gap, rollback network round-trip, teardown/drop ordering).

The `OpenAppID` ownership-marker gap surfaced in review is real (confirmed: present only on `MJOpenAppDependency`/`MJOpenAppInstallHistory`, never on the free-floating core tables) but is a pre-existing, larger design question that rightly belongs in its own issue — not a blocker for this PR.

Ship the transaction-mode change as-is; merge the compensation change with the follow-ups tracked. Nice, responsive iteration.

*Approving review by Claude (Opus 4.8) at the request of @rkihm-BC; changes verified against source.*

PR #3487, opened 2026-08-05
fix(open-app): stop the installer silently disabling CodeGen entity registration

### LINKED ISSUE #3301: mj app install writes dynamicPackages into a comment — Open App server packages silently never load
## Summary

`mj app install` writes the `dynamicPackages` block into the **first `module.exports = {` it finds in the file** using a comment-blind regex. In MJ's own default MJAPI config scaffold that match is the *example* inside the header doc comment — so every entry the installer writes lands inside `/* … */` and is silently inert.

The install reports success. `mj app list` shows the app `Active`. But MJAPI never loads the app's server package, and **nothing anywhere reports a problem**.

## Root cause

`@memberjunction/open-app-engine/dist/install/config-manager.js` → `InsertBeforeModuleExportsClose`:

```js
const inlineMatch = content.match(/module\.exports\s*=\s*\{/);
```

No comment stripping. The MJAPI scaffold ships with this in its header comment:

```js
 * To override ANY default setting, simply add it to this file.
 * For example:
 *
 * module.exports = {
 *   telemetry: { ... }
 * };
 */
```

That is the first match, so `FindMatchingBracket` returns the example's closing brace and the section is inserted before it — inside the comment.

## Reproduction

1. Start from the stock MJAPI `mj.config.cjs` scaffold (the one whose header documents an example `module.exports = { … };`).
2. `mj app install <any open app>`.
3. Open `apps/MJAPI/mj.config.cjs` — the `dynamicPackages` block is inside the `/** … */` header comment, and the comment's example has been mangled (its closing `* };` consumed).
4. Start MJAPI: no `Loading Open App server packages…` entries for the installed app. No error.

Observed on `@memberjunction/cli` 5.48.0; `InsertBeforeModuleExportsClose` is byte-identical in 5.49.0.

## Impact

Silent and total: the app is installed in the database but its server package never loads, so its resolvers, class registrations and startup side effects are all missing. The failure surfaces much later as an unrelated-looking "entity not found" / missing-GraphQL-operation problem, with no breadcrumb back to the installer.

## Suggested fix

Strip comments and strings before locating the insertion point, or anchor on a `module.exports` match that is not inside a comment. A cheap defensive addition: after writing, `require()` the config and assert the section is actually present — this bug is invisible precisely because nothing verifies the write.

## Workaround

Move the block into the real `module.exports` by hand after every install/upgrade. Note that once a genuine `dynamicPackages:` key exists, `EnsureDynamicPackagesSection` early-returns and `AddEntryToDynamicArray` dedupes on PackageName+AppName — so a corrected file stays correct on subsequent installs.

### LINKED ISSUE #3457: Open App installer writes the app's schema into host excludeSchemas, silently preventing CodeGen from creating that app's entities
## What to build

`InstallApp` / `UpgradeApp` should stop writing the app's own schema into the host's `excludeSchemas`. `entityPackageName` already provides the duplicate-suppression that write is presumably there for, and `excludeSchemas` additionally gates entity **discovery** — so for any Open App that does not ship its own `__mj.Entity` metadata, the installer's own write is what prevents CodeGen from ever creating that app's entities.

Observed on `@memberjunction/cli` + `@memberjunction/open-app-engine` 5.51.0, SQL Server host.

### What happens today

`install-orchestrator.js:1020` calls, unconditionally, for every app:

```js
const excludeResult = AddExcludeSchema(context.RepoRoot, manifest.schema.name, context.ServerPackagePath);
```

The same orchestrator also calls `AddEntityPackageMapping`, so the host config ends up with the app's schema in **both** `entityPackageName` and `excludeSchemas`.

### Why the `excludeSchemas` write is redundant

`entityPackageName` alone already keeps an app's entities out of every artifact the host emits:

- `getExternalEntitySchemas()` derives its schema list from `entityPackageName` (`Config/config.js`).
- `runCodeGen.js:478` filters those schemas out of `localNonCoreEntities`.
- That filtered list is what drives GraphQL (`:505`), EntitySubclasses (`:533`) and Angular (`:561`).

So duplicate subclasses and duplicate GraphQL ObjectTypes are already prevented without touching `excludeSchemas`.

### Why the write is actively harmful

`excludeSchemas` is not only a file-generation filter — `Database/manage-metadata.js` passes it into `updateExistingEntitiesFromSchema` and the new-entity creation path, so it gates entity **discovery** as well.

For an app that seeds its own `__mj.Entity` rows in a baseline migration, excluding the schema is harmless. For an app that relies on the host's CodeGen to create its entity metadata, it is fatal and silent:

1. `mj app install` succeeds and creates the app's tables.
2. The installer adds the app's schema to `excludeSchemas`.
3. `mj codegen` succeeds but skips the schema entirely — **zero entities created**.
4. `mj sync push` then has nothing for the app's metadata to attach to, and the app's published `-entities` package resolves nothing.

Nothing in that sequence reports an error. We hit exactly this on a first-party BizApp (tracked in `bizapps-caliber` as #114 and #119): 23 tables present, 0 entities.

### Why it needs fixing rather than documenting

It **re-arms on every run**. Both install and upgrade perform the write, so a host that removes the line by hand gets it back on the next `mj app upgrade` — and then the next CodeGen run silently reproduces the original failure. We hit it twice in one session: once on install, once on the upgrade that was meant to fix the first occurrence.

`mj app remove` has the mirror-image call (`RemoveExcludeSchema`, `install-orchestrator.js:630`), but see the note below — that path cannot complete for these apps anyway.

## Acceptance criteria

- [ ] `mj app install` no longer adds the app's `manifest.schema.name` to the host's `excludeSchemas`
- [ ] `mj app upgrade` likewise does not re-add it
- [ ] `entityPackageName` continues to be written, and duplicate entity subclasses / GraphQL ObjectTypes are still suppressed for the app's schema (regression test over `localNonCoreEntities` filtering)
- [ ] Installing an app that ships **no** `__mj.Entity` metadata, then running `mj codegen`, creates that app's entities
- [ ] Installing an app that **does** seed its own entity metadata still produces no duplicate subclasses
- [ ] Upgrading an app leaves the host's `excludeSchemas` unchanged

## Blocked by

None - can start immediately.

---

### Two adjacent findings from the same investigation

Filing separately would fragment the context; happy to split if you'd rather.

**1. `mj app remove` cannot complete for any first-party BizApp, and fails destructively.** `app install` and `app upgrade` both expose `--dangerously-ignore-dbl-underscore-schema-rule`; `app remove` does not (its flags are `keep-data`, `force`, `yes`, `verbose`). Since every first-party BizApp uses a `__mj_BizApps*` schema, remove always fails at the schema drop with *"Schema names starting with '__' are reserved for MJ internals"* — **after** it has already deleted the app's entity metadata and set its status to `Error`. The result is a partial removal with no way to finish and no rollback.

**2. `mj-app.json` `shared` libraries are mapped into `dynamicPackages.client`.** This puts each app's `*-actions` package into the host's Angular bundle, where it pulls `@memberjunction/aiengine` → `@memberjunction/storage` → `@google-cloud/storage` → `teeny-request` → `https-proxy-agent@5.0.1`, whose bare `require("net"/"tls"/"url"/"assert")` esbuild cannot resolve for a browser target. That fails the host's Explorer build outright. Each app's `*-server` package already depends on its own `*-actions`, so the server side is unaffected — the client entry appears to be the mismapping.

SDesai-BC:
An Open App could install cleanly, run `mj codegen` cleanly, and end up with its tables present and **zero entities** — with no error at any step. We hit it on a first-party BizApp: 23 tables, 0 entities. `mj sync push` then had nothing to attach the app's metadata to, and the app's published `-entities` package resolved nothing.

Closes #3457.

## Root cause

`HandleServerConfig` called this unconditionally, for every app, on both install and upgrade:

```ts
const excludeResult = AddExcludeSchema(context.RepoRoot, manifest.schema.name, context.ServerPackagePath);
```

`excludeSchemas` gates **three** independent things in CodeGen, not one:

| Concern | Site |
|---|---|
| Entity **discovery** | `manage-metadata.ts:4642` — `createNewEntities()` appends `createExcludeTablesAndSchemasFilter()` to its `WHERE` |
| SQL ownership (base views, CRUD procs) | `sql_codegen.ts:133-135` |
| TS / GraphQL / Angular emission | `runCodeGen.ts:518` — `localNonCoreEntities` |

Only the third is what an installed app needs suppressed, and the `entityPackageName` mapping the installer already writes (the line immediately above) does exactly that on its own. The `excludeSchemas` write bought nothing there and took entity registration down with it.

That matters because registration by the host's CodeGen is the *documented* contract. The README's canonical app migration is raw DDL with no `__mj.Entity` seed, followed by: *"Do NOT include `__mj_CreatedAt`/`__mj_UpdatedAt` columns or foreign key indexes — MJ's CodeGen handles those automatically after entity registration."* An app following the documentation was the failure case.

It also **re-armed**: both install and upgrade performed the write, so a host that removed the line by hand got it back on the next `mj app upgrade`, and the next CodeGen run silently reproduced the original failure.

### The justification was already dead

`4b8afa6522` (2026-07-10) added the write to stop app-owned `flyway_schema_history` becoming an entity. But `ae02c3ceee` (2026-01-20) — nearly six months earlier — had already added `{ schema: '%', table: 'flyway_schema_history' }` to CodeGen's default `excludeTables`, across all schemas. A one-table problem got a schema-wide hammer that was already unnecessary.

Worth knowing for hosts: `excludeTables` **replaces** rather than merges on override (`config-merger.ts:111`), so a host that defines its own `excludeTables` drops that default entry and should re-add it. That is the likeliest thing the original author actually hit.

## What this does

Adds `schema.selfManagedMetadata` to the manifest (default `false`) and makes the exclusion opt-in:

```ts
const excludeResult = claimedSelfManaged
  ? AddExcludeSchema(context.RepoRoot, schemaName, context.ServerPackagePath)
  : RemoveExcludeSchema(context.RepoRoot, schemaName, context.ServerPackagePath);
```

**The default branch removes rather than just not-adding.** Deleting the write alone would leave every host installed by an earlier version permanently broken — the schema is already sitting in their `mj.config.cjs`. Removing on every install/upgrade inverts the old "re-arms on every run" behaviour into self-repair.

`entityPackageName` is written on both paths, so duplicate entity subclasses and duplicate GraphQL ObjectTypes stay suppressed — two packages emitting the same ObjectType makes graphql-js reject the unified schema at boot, so that is regression-tested rather than assumed.

### What an adversarial review found, and what it cost

The first version of this PR **did not work**, and the smoke evidence originally posted here was wrong. A review caught it; every finding was reproduced against the built package before being accepted, and re-verified after fixing.

`RemoveSchemaFromExcludeArray` located its target with a bare `content.match(/excludeSchemas\s*:\s*\[/)` and fell back to a global `['"]<schema>['"]` replace over the whole file. Since `HandleServerConfig` writes the `entityPackageName` key moments earlier, the fallback deleted that key and produced `: "@pkg/x"` — invalid JavaScript. `WriteConfigChecked` refused the write and a root-config failure is fatal, so:

| Scenario | Before |
|---|---|
| **Fresh install** (schema not yet excluded — the common case) | install **aborted**, app left `Status='Error'` |
| **Repeat upgrade** | **aborted** on every run after the first |
| Host using `includeSchemas` | array **silently emptied**, `Success: true` |

The original smoke harness passed because it seeded the one arrangement that worked (schema present among several entries) and never asserted the second removal's `Success`. Two gaps pointing the same way.

Auditing the config editors rather than just that one function surfaced two more, both live:

- **Comments were treated as configuration.** `distribution.config.cjs` ships `excludeSchemas` commented out; the editors wrote *into the comment* and reported success while the evaluated config had none — so `selfManagedMetadata: true` was a silent no-op on the shipped template.
- **A nested key could win.** `excludeSchemas` also exists under `dbSchemaJSONOutput` and in `bundles[]`; whichever appeared first was edited.

All three now resolve through one `FindTopLevelConfigArray` helper that skips strings and comments and tracks brace depth — the hardening `RemoveEntityPackageEntry` already carried. That is **commit 1**, separate and independently revertible, since it fixes pre-existing latent bugs rather than this feature.

### Three more conditions the default path needed

Each reproduced the original "tables but no entities" symptom by a different route:

- **`includeSchemas` hosts.** It is an opt-in POSITIVE scope that CodeGen resolves into `excludeSchemas` by excluding every schema it does not name — so clearing `excludeSchemas` alone still left the app excluded. The installer now names the schema there too. It never creates the key and never writes into an empty one: absent or empty means "no scope in force", and populating it would scope CodeGen to that single schema and silently drop every other schema the host owns.
- **PostgreSQL casing.** PG folds unquoted DDL to lowercase, so the physical schema is `__mj_bizappscaliber` while the manifest says `__mj_BizAppsCaliber`, and the discovery filter compares with a case-sensitive SQL `<>` on PG. The write now uses `Dialect.CanonicalSchemaName`, agreeing with what the orchestrator already does for `SchemaInfo`.
- **Shared schemas.** Un-excluding is a decision about the *schema*, not one app. If another installed app declares it `selfManagedMetadata: true`, the exclusion is kept — install order no longer silently decides whether the host co-owns their tables. Mirrors the shared-aware `HandleAngularPrebundleExcludeRemoval`.

`ConfigOperationResult` also gains `Changed`: `Success` alone could not distinguish a successful no-op from a successful destructive edit, which is exactly what let the failures above pass for working behaviour. The installer now warns when it actually removed a host's entry.

## Evidence

Everything below was re-run at the current head — not carried over from an earlier commit.

| Tier | Result |
|---|---|
| `@memberjunction/open-app-engine` | **381 passed**, 29 files |
| `@memberjunction/codegen-lib` | **709 passed, 60 skipped** |
| Config **shape sweep** — 1,296 shapes × 3 runs | **pass** |
| Config round-trip smoke (built `dist`, real files) | **48/48** |
| Integration — `Integration Tests — Deterministic` | **54/54 (100%)** |
| Unit, cache disabled (`turbo test --force`, affected subtree) | **603/603 tasks** |
| Full workspace build (changed packages + all dependents) | **303/303 tasks** |
| `check:standards`, `check:esm` | pass |

The unit run uses `--force` deliberately: a plain `turbo test` served 157 of 158 tasks from cache and proved almost nothing.

### The shape sweep, and why it exists

Every config-editor defect found in this work was the same failure — correct for the shapes someone thought to test, silently wrong for a shape a real host used. An unanchored regex that ate `entityPackageName`; an anchor selecting a `module.exports` inside a header comment; a key nested under `dbSchemaJSONOutput` winning because it appeared first; a quoted `"excludeSchemas"` reading as absent and being shadowed by an appended duplicate. Four instances, each found by review *after* the previous one was declared verified.

`config-manager-shape-fuzz.test.ts` enumerates the space instead: key quoting × position × header comment × nested same-name keys × export form × array state × positive-scope state — **1,296 combinations**, exhaustive rather than random so a failure is reproducible from its label. Each is driven through the real `HandleServerConfig` sequence three times against invariants that must hold for any shape: the file still evaluates, the host's own entries survive (neither deleted **nor shadowed**), unrelated and nested same-name keys are untouched, intent is reflected in the **evaluated** config rather than merely written somewhere, and three runs are indistinguishable from one.

**It found a live defect on its first run.** `InsertBeforeModuleExportsClose` still anchored on `content.match(/module\.exports\s*=\s*\{/)`, which returns the first textual hit — and MJ's own default MJAPI scaffold documents an example `module.exports = {…}` in its header comment. Every section that helper creates on a stock host landed inside that comment, inert, while the operation reported success. All three of its callers shared it: `dynamicPackages`, `excludeSchemas`, `entityPackageName`. Now routed through the comment-aware `FindExportedObjectBrace`.

**That also closes #3301** — an app's server packages silently never loading because their `dynamicPackages` entry was written into a comment. Verified directly on the fixed build; the before-state for that specific caller is inferred from the sweep's failure on its sibling through the same helper, so #3301's owner should confirm rather than take this at face value.

New coverage is deliberately **real-temp-file**, not mocked `node:fs` — the orchestrator suites mock the whole config-manager module, which is exactly why this class of bug was invisible to them. Each fix was proven to *catch* its bug: the pre-fix build was used as a control and reproduced every failure before rebuilding.

### Integration tier

`pnpm -w run test:integration` (`Integration Tests — Deterministic`) against SQL Server: **54/54 passed (100%)**, 74.8s.

An earlier run showed 53/54 on `IT65 — AI Providers and Model Resolution`, which reported two Active LLM models with an unresolvable `DriverClass` (`Cohere Command A` / `A+` → `CohereLLM`). That was traced to the **test database being stale relative to this repo's own metadata**: `metadata/ai-models/.ai-models.json` already carries both models as `IsActive: False`, while the database still had them at `1`. A scoped `mj sync push --dir=metadata --include="ai-models"` updated exactly those two records (937 unchanged, 0 created, 0 deleted) and the suite is now clean. Environmental, not a product defect, and nothing to do with this diff — which contains no AI-model surface.

## Hazard

Existing hosts change on install/upgrade. When the probe says the host's CodeGen owns the schema, the installer **edits `mj.config.cjs`**, removing that schema from `excludeSchemas`. That is the intended repair, but the next `mj codegen` will then register entities and generate base views + CRUD procs for a schema it previously ignored, emitting a `CodeGen_Run` migration. Expect that diff.

**Verified against CDP stage**, where all four first-party apps are installed. The apps split both ways, which is why detection replaced a manifest default:

| App | Entities created | vs. install | Reading |
|---|---|---|---|
| `mj-bizapps-common` | 20:58:42 | **before** its OpenApp row (20:59:45) | seeds its own — stays excluded |
| `mj-bizapps-tasks` | 22:09:56 | before (22:10:55) | seeds its own — stays excluded |
| `mj-bizapps-forms` | 22:18:39 | before (22:19:22) | seeds its own — stays excluded |
| `mj-bizapps-caliber` | 2026-08-05 00:29 | **2 h after** install (22:24) | seeds none — host registers |

Common/Forms/Tasks sit in that host's `excludeSchemas` *and* have entities, so host CodeGen cannot have created them: their baselines carry `INSERT INTO [__mj].Entity`. Caliber's 22 entities only appeared once the exclusion was removed by hand. A fixed `false` default would have un-excluded the first three on their next upgrade and handed the host's CodeGen co-ownership of metadata they already own — so the installer reads the entity count instead, and `selfManagedMetadata` exists only to override a probe that would be wrong.

If the probe cannot run, the installer changes nothing and warns. Both edits are destructive in one direction, so declining beats guessing.

**Follow-up: #3490.** `excludeSchemas` gates three independent concerns (discovery, SQL ownership, emission) while `entityPackageName` gates only the third. Detection resolves the metadata question correctly, but SQL ownership is still bundled with it — `bizapps-common` ships ~50 of its own `CREATE VIEW`/`CREATE PROCEDURE` statements *and* seeds its own entities, so it lands on the excluded side and keeps them; an app that seeded entities but wanted host-generated SQL has no way to say so. #3490 tracks separating the three. Kept out of this PR deliberately: it touches 11 CodeGenLib files across both dialects and needs its own migration path.


rkihm-BC:
Adding @MarceloT-BC as a reviewer because of his work with the includeSchemas and familiarity with OpenApp installs. 

This seems odd that we're discovering this issue now with all the OpenApps we've deployed so far. Is this related to recent changes?

MarceloT-BC:
Read through this carefully. The root-cause analysis is right, and the `excludeSchemas`-gates-three-things framing plus the dead-justification archaeology (`4b8afa6522` vs `ae02c3ceee`) is the part that makes the fix defensible rather than just plausible. The remove-rather-than-not-add default, and the reasoning for why `includeSchemas` must never be created or populated when empty, both look correct to me. Notes below are on the margins, not the core.

**The one I'd want addressed: the uninstall path isn't shared-schema-aware.**

`RemoveApp` calls `RemoveExcludeSchema` unconditionally (`install-orchestrator.ts:1146`), but `schemaShared` is already computed at line 1039 and consulted three times just above (1047, 1052, 1080) — same scope, unused at the config cleanup.

So: app A (default) and app B (`selfManagedMetadata: true`) share a schema. Installing A correctly preserves the exclusion via `AnotherInstalledAppSelfManages`. Removing A strips it, and B's self-managed schema silently becomes CodeGen-owned — entity registration plus base views and CRUD procs over a schema whose owner explicitly opted out. That's the same install-order hazard this PR built `AnotherInstalledAppSelfManages` to close, just on the mirror path. `HandleAngularPrebundleExcludeRemoval`, in the same `Promise.all`, already takes `otherManifests` for exactly this reason.

Narrow today — neither first-party manifest declares the flag, so it needs a third-party app to hit — but it's a hole in the feature's own invariant and the signal is already in hand.

**Smaller:**

- `RemoveIncludeSchema`'s doc (`config-manager.ts:734`) says it's used *"when an app declares it manages its own metadata **and when the app is removed**"*, but the only production call site is the self-managed branch at `install-orchestrator.ts:1882` — removal never calls it. I don't think the residue is harmful (dropped schema → inert; shared → arguably what the co-owner wants; `KeepData` → those tables are in scope on a plain host anyway), so this reads as a doc/symmetry fix rather than a bug. Flagging it mainly so the comment doesn't outlive the intent.
- `FindTopLevelConfigArray` enters string mode on the opening quote, so a quoted key — `"excludeSchemas": [...]`, legal in a `.cjs` and a natural shape when copied from JSON — isn't found. `EnsureExcludeSchemasSection` then appends a second unquoted key, and last-key-wins means the host's original list silently stops applying. Same failure class commit 1 targets; cheap to cover in the same helper.

**On the hazard section:** calling out that `bizapps-common` fits neither branch, and that the default path means the host's CodeGen starts regenerating its ~50 baseline views until #3490 lands, is the right call — better in the open than discovered downstream. Worth making sure whoever owns that stack has actually seen it before this merges, since the regression is real even if temporary.

Separately, the write-up of the failed first version — including that the original smoke evidence was wrong and why the harness passed anyway (seeded the one arrangement that worked, never asserted the second removal) — is worth more than the fix. Please keep doing that.

Review is static only: I read the diff and the surrounding source on the PR branch, and did not re-run any of the suites in the evidence table.

SDesai-BC:
Thanks — all three land, and the first two were real. Verified each against the built package rather than by reading; details below.

## Uninstall shared-schema awareness — confirmed and fixed

You were right that the signal was already in hand and unused. One correction to the mechanics: `schemaShared` isn't in the same scope as the cleanup. It's declared at `:1039` inside the `DbCleanupDone` block; the `Promise.all` is at `:1143` inside `FilesRemoved`. My first attempt just referenced it and blew up with `schemaShared is not defined`.

That turned out to matter beyond a compile error: they're independently resumable steps, so on a resume `DbCleanupDone` is skipped entirely and deriving the check in there would leave the config cleanup with no answer. The share check now runs **once above both steps**, and the cleanup is gated on it — matching `HandleAngularPrebundleExcludeRemoval`, as you noted.

Covered both directions: co-owner present → exclusion kept; no co-owner → still cleaned up.

## Quoted keys — confirmed, and worse than described

You predicted the host's list would silently stop applying. It does, and the ADD path is destructive rather than merely ineffective:

```
"excludeSchemas": ['sys', 'acme_crm'],   ← host's real list, untouched
excludeSchemas: ["newapp"],               ← appended duplicate
→ effective: ["newapp"]
```

`sys` goes with it, so CodeGen would start adopting system tables on the next run. That puts it in the same severity class as the bugs commit 1 targets, not on the margins.

Fixed by recognising a quoted key *before* the scanner treats the quote as a string — both keys, both quote styles, with a test pinning that a quoted key inside a comment is still ignored. Worth noting the irony: the string-awareness I added to stop treating non-code as config is exactly what made a legitimate config shape invisible, and every test I'd written used unquoted keys.

## `RemoveIncludeSchema` doc — agreed, doc fix

Single production call site, the self-managed branch. Comment corrected, and I recorded *why* the residue is deliberate: a leftover include entry is inert, whereas removing one could narrow a scope a co-owner still depends on. That's the asymmetry the old wording lost.

## On `bizapps-common` — you read a superseded body

That hazard is gone, and the reason is worth flagging since it changed the design. Checking CDP stage, `mj-bizapps-common`, `-forms` and `-tasks` **do** seed their own `__mj.Entity` rows — their entities exist *before* their `OpenApp` row is written, while their schemas sit in `excludeSchemas`. My earlier claim that they didn't came from a grep for `INSERT INTO.*\[Entity\]` that missed the actual `INSERT INTO [__mj].Entity` form.

So a manifest default was wrong in both directions: three of four first-party apps self-seed while declaring nothing. The installer now **detects** — counts the schema's `__mj.Entity` rows after migrations run — with `selfManagedMetadata` demoted to an explicit override, and "probe failed" meaning *change nothing and warn* rather than guess. `bizapps-common` lands on the excluded side and keeps its ~50 baseline views, so there's no regression for that stack to absorb. What survives is the #3490 limitation: SQL ownership is still bundled with metadata ownership.

## Verification

375 unit (open-app-engine), 709 + 60 skipped (codegen-lib), 48/48 config round-trip smoke against built `dist`, 54/54 integration, standards clean. Both fixes were reproduced on the pre-fix build first, so the tests demonstrably catch the bugs rather than merely pass.

Noted that your review was static — the numbers above are mine, re-run at `3a05a87af0`.

SDesai-BC:
Follow-up on the quoted-key finding. I said every test I'd written used unquoted keys — that pattern was the actual problem, not the individual gap, so I've addressed it structurally rather than by adding another case.

## Shape sweep

New `config-manager-shape-fuzz.test.ts` enumerates the config **shape space** exhaustively — 1,296 combinations of key quoting × position × header comment × nested same-name keys × export form × array state × positive-scope state. Exhaustive rather than random so a failure is reproducible from its label.

Each shape is driven through the real `HandleServerConfig` sequence three times and checked against invariants rather than expected strings:

- the file still evaluates as JavaScript
- the host's own entries survive — neither deleted **nor shadowed**
- unrelated keys and nested same-name keys untouched
- intent is reflected in the **evaluated** config, not merely written somewhere
- three runs are indistinguishable from one

The shadowing check needs no text counting: if an appended duplicate key won, the host's original entries would be missing from the evaluated value.

## It found a live one immediately

`InsertBeforeModuleExportsClose` was still comment-blind — I'd fixed the anchor on the schema-array path and not this one. It has **three callers**: `dynamicPackages`, `excludeSchemas`, `entityPackageName`. On MJ's own default MJAPI scaffold, every section it creates landed inside the header comment, inert, while reporting success. The sweep caught it as `entityPackageName` going missing on `header=moduleExportsExample` shapes.

Now routed through the comment-aware `FindExportedObjectBrace`.

**This also closes #3301** — an app's server packages silently never loading because their `dynamicPackages` entry was written into a comment. Verified directly:

```
AddServerDynamicPackages Success= true
EFFECTIVE dynamicPackages.server = [{"PackageName":"@acme/srv", ...}]
comment intact = true
```

One precision so you can calibrate: I verified the *after* state for `dynamicPackages` directly and inferred its *before* state from the sweep's failure on `entityPackageName`, its sibling through the same helper. Worth whoever owns #3301 confirming rather than taking my word — I didn't run a dedicated before-control on that specific caller.

## Why this rather than more cases

Every defect found in this area has been the same failure: correct for the shapes someone thought to test, silently wrong for a shape a real host used. The unanchored regex that ate `entityPackageName`, the anchor selecting a commented `module.exports`, the nested key winning on position, your quoted `"excludeSchemas"` — four instances, each found by review after the previous was declared verified. Enumerating the space closes that by construction.

The sweep covers `mj.config.cjs` editing only. The same reasoning would extend to the manifest parser and the migration-runner path, but that's a separate change.

Now at 381 unit / 48-48 config smoke / 54-54 integration, `78aa92b5b0`.

rkihm-BC:
## Code review — full pass (static + targeted reproduction)

Reviewed the whole diff and the surrounding source on the branch. The root-cause analysis, the remove-rather-than-not-add default, the `includeSchemas`-never-create/populate guard, the comment/string-aware anchoring, and the uninstall shared-schema fix are all sound, and the `@MarceloT-BC` thread already covers quoted keys, uninstall reconciliation, the `RemoveIncludeSchema` doc, and the shape sweep. I won't re-litigate those.

Two findings below are **not** in the thread yet — one I'd want answered before merge, one a real (if minor) defect that hides *inside* a dimension the sweep claims to cover. Then some smaller notes. I did **not** re-run the evidence-table suites; findings are from reading plus one Node reproduction, called out where used.

---

### 1. 🔴 (please confirm) The entity-count probe flips a host-managed app back to *excluded* on its first upgrade after CodeGen has run

`SchemaHasSeededEntities` counts `__mj.Entity` rows for the schema and treats "rows present" as "app self-manages". But that count can't distinguish **rows the app's migrations seeded** from **rows the host's CodeGen registered on a previous cycle** — and `HandleServerConfig` runs the same detection on `UpgradeApp`, not just `InstallApp`.

Trace a host-managed (caliber-shaped, `selfManagedMetadata` unset) app across its normal lifecycle:

1. `mj app install` (v1) — migrations seed tables, **0 entities**. Probe → 0 → `RemoveExcludeSchema`. ✅ correct, schema discoverable.
2. `mj codegen` — host registers ~22 entities into `__mj.Entity` for the schema. Rows now exist.
3. `mj app upgrade` (v2) — migrations run (say v2 adds a table). Probe now sees **22 > 0 → `AddExcludeSchema`**. Schema is excluded again.
4. `mj codegen` — the schema is excluded, so v2's new table is **never discovered → zero entities**. That is #3457 again, for the incremental table, plus base-view/CRUD-proc drift on the whole schema.

So for a host-managed app that *evolves*, the fix self-repairs at install and then self-*breaks* at the first upgrade after any codegen run. This also puts the README guidance ("leave `selfManagedMetadata` unset, detection handles it") in tension with itself: to stay reliably host-managed across upgrades, such an app actually has to declare `selfManagedMetadata: false` so the probe is bypassed — the opposite of "leave it unset."

The install-time CDP evidence in the PR body only exercises the *first* install/upgrade, before the host's CodeGen has populated `__mj.Entity`, so it wouldn't surface this. You have the real environment and I don't, so I'm flagging it as a question rather than asserting a regression — but the logic looks like the detection is only stable *before* the first codegen. Possible mitigations if it's real: only run the probe when the app has no `OpenApp` history (true first install), or persist the detected/declared decision onto the `OpenApp` record and re-read it on upgrade rather than re-probing.

---

### 2. 🟠 Inserting into a `excludeSchemas`/`includeSchemas` array that has a trailing comma produces a `null` hole — and the 1,296-shape sweep passes anyway

Both `AddSchemaToExcludeArray` (`config-manager.ts:991`) and `AddSchemaToIncludeArray` (`config-manager.ts:748`) insert `, ${JSON.stringify(schemaName)}` immediately before the closing `]`. When the host's array ends in a trailing comma — legal, common, and one of the sweep's own `array: 'trailingComma'` states — that yields `,,`, which is a valid array **elision**, not a syntax error. Reproduced exactly:

```js
// host config: excludeSchemas: ['sys', 'staging',],
// after AddExcludeSchema('__mj_App'):
excludeSchemas: ['sys', 'staging',, "__mj_App"],
// evaluates to:
["sys", "staging", null, "__mj_App"]   // length 4, index 2 is a hole
```

That `null` becomes a bogus entry CodeGen then reasons about (`[SchemaName] <> null` / a `null` in the resolved include-scope), which is malformed output the editor claims never to produce.

The reason the sweep misses it despite `array=trailingComma` being a swept dimension is the invariant choice: `toContain('sys')`, `toContain(SCHEMA)`, and a `filter(s === SCHEMA).length === 1` dupe check all stay green with a `null` sitting in the array. An exact-membership assertion (`expect(cfg.excludeSchemas).toEqual([...expectedExactly])`, or `expect(arr).not.toContain(null)` / `.every(s => typeof s === 'string')`) would have caught it — worth adding, since "the host's entries survive" was supposed to be the whole point. The `EnsureTrailingComma` helper already used by `InsertBeforeModuleExportsClose` is the model; the array inserts need the mirror of it (don't emit a leading comma when the slice before `]` already ends in one).

---

### 3. 🟡 `HandleServerConfig` now touches `DatabaseProvider.Dialect` unguarded — `PersistCanonicalSchemaName`, two functions away, deliberately guards the same access

`const schemaName = context.DatabaseProvider.Dialect.CanonicalSchemaName(manifest.schema.name);` (`install-orchestrator.ts:1911`) runs with no guard. Compare `PersistCanonicalSchemaName`, which wraps its **entire** body in try/catch precisely because *"building the statement touches the provider's Dialect, so the WHOLE body (not just ExecuteSQL) is guarded — a provider without a usable Dialect must never fail the install."*

The old code here (`AddExcludeSchema(context.RepoRoot, manifest.schema.name, ...)`) never touched `Dialect`, so this is a new failure surface: a provider whose `Dialect` is absent/unusable now throws out of the config step (→ app left `Status='Error'`) where before it wouldn't. `SchemaHasSeededEntities`' own try/catch doesn't help, because `CanonicalSchemaName` is called *before* it. Suggest mirroring the established pattern — fall back to the raw `manifest.schema.name` if `Dialect` is unavailable, rather than letting it throw.

---

### 4. 🟢 (nit) `SchemaHasSeededEntities` builds its FROM clause unquoted while its sibling uses the dialect quoter

`` `SELECT COUNT(*) AS EntityCount FROM ${context.MJCoreSchema ?? '__mj'}.Entity ...` `` interpolates the schema/table directly, whereas `PersistCanonicalSchemaName` (same file) builds the identical `__mj` reference via `dialect.QuoteSchema(mjSchema, 'SchemaInfo')` / `QuoteIdentifier`. `MJCoreSchema` is host config, not attacker input, so this isn't an injection concern — purely a consistency one, and it's what keeps the two dialects from disagreeing on identifier folding. Low priority.

---

### 5. 🟢 (nits) docs / changeset

- **README example vs. prose.** The `schema` example block now shows `"selfManagedMetadata": false` explicitly, while the prose two lines down (correctly) says *"Optional override — leave it unset."* Showing the field with a value in the canonical example slightly undercuts that; consider omitting it from the example, or annotating it as "shown for illustration; omit in practice."
- **patch vs. minor.** The changesets are all `patch`, justified by "no migration." Reasonable by your repo convention, but note this PR adds a public manifest field (`selfManagedMetadata`) and three new exported functions (`AddIncludeSchema`, `RemoveIncludeSchema`, `filterOutExternalSchemaEntities`) — additive public API, which is conventionally a minor. Not blocking; just flagging that the "minor ⇒ migration" heuristic and the "additive API ⇒ minor" heuristic disagree here.

---

### What's good (worth keeping)

- Extracting `filterOutExternalSchemaEntities` and pointing the regression test at the *production* function instead of a reimplemented copy is exactly right — that's the failure mode ("test pins a drifting copy") that let the original bug hide.
- Tri-state `selfManagedMetadata` (`undefined` ≠ `false`) with a zod `.optional()` and no default, plus the test that rejects `'true'`-as-string, is the correct modelling of "detect vs. force."
- Real-temp-file tests over mocked `node:fs`, with the pre-fix build used as a control to prove each test *catches* its bug, is the standard this class of code needed.
- `ConfigOperationResult.Changed` to separate "successful no-op" from "successful destructive edit," and warning the host when their hand-written exclusion is removed, closes the ambiguity that let the earlier failures read as success.

### Verification I actually did
Static read of the full diff + branch source; one Node reproduction for finding #2 (confirmed the `null` hole and that the sweep's invariants stay green). I did not re-run the unit/integration/sweep suites in the evidence table.

PR #3491, opened 2026-08-05
feat(actions-apollo): Apollo list management, search and prospecting

MS-BC:
Adds Apollo's other half to `@memberjunction/actions-apollo`: **list management, saved-record search and prospecting**, as seven new actions alongside the two enrichment actions the package has always had. Ported from AIDP's `ApolloClient` (745 LOC), which is the proven statement of these surfaces' quirks.

`ApolloGetListsAction` / `ApolloCreateListAction` read and create labels; `ApolloGetListAccountsAction` / `ApolloGetListContactsAction` page a list's members; `ApolloSearchPeopleAction` finds net-new prospects; `ApolloMoveListAccountsAction` / `ApolloMoveListContactsAction` move records between lists. Together they are the outbound-campaign surface — build a list, drain it through stages, see what is where — which enrichment alone cannot express.

These use a **different Apollo base path** from enrichment: `api.apollo.io/api/v1`, not `api.apollo.io/v1`. Not interchangeable; the same path under the wrong prefix 404s. So `ApolloRESTEndpoint` is declared beside the existing `ApolloAPIEndpoint` rather than derived from it.

### Five vendor behaviours that make the naive implementation wrong

**A PATCH replaces the whole `label_names` array.** There is no add-one-label endpoint, so a move is *two* writes, each carrying the complete intended set: `current ∪ {toList}`, then `(current ∪ {toList}) \ {fromList}`. Sending a bare `['Warm']` silently deletes every other list the record was on, and nothing in the response reveals it. The remove set is computed from the **post-add** state — deriving it from `current` alone would add the record to the destination then immediately take it back out. The move actions accept only ids, never labels, and re-read the source list themselves, so a caller cannot supply a stale label set that destroys memberships the record picked up since.

**Roughly 15–17% of removes return success without applying.** Removes are verified by re-reading the destination; a record still carrying the source label is reported `possiblyStuck` rather than retried — an immediate retry flakes the same way, while the next page-1 drain sees both labels and finishes the job. Deliberately **not** a failure: the write succeeded and Apollo did not honour it, so calling it a failure sends the caller hunting a bug in the request. `PARTIAL_FAILURE` is reserved for a PATCH that actually failed. When the verify read itself fails, the stuck state is `null` — unknown, not verified-clean.

**A list drain must always read page 1**, because removals shift every later page, so advancing to page 2 skips records. Paging defaults to page 1 rather than a saved position; the move actions pin it explicitly.

**Label reads and every write require a MASTER key.** A scoped key authenticates fine and then 403s, indistinguishable from a wrong key — so a 403 on a master-key operation is rewritten to name the requirement. The prospecting search is the one read a scoped key can serve, so it deliberately skips the label fetch.

**Labels are read as `label_ids` but written as `label_names`.** The client fetches the label list once per instance and resolves ids on read; an unmatched id is dropped, since a raw id inside a `label_names` write would create a label named after a hex string.

### Two smaller invariants

`ApolloCreateListAction` is idempotent — a same-named label returns as-is with `AlreadyExisted: true`. Apollo will happily create two labels with one name, and once it has, every name-based lookup here becomes ambiguous: half the moves target one list, half the other. `ApolloSearchPeopleAction` requires at least one filter; Apollo accepts a filterless request and returns an unscoped, rate-limit-burning result set nobody asked for.

### Credentials

`CompanyID` → the company's active `Apollo` **MJ: Company Integrations** row → `CredentialID` → `apiKey` in **MJ: Credentials** `Values`, falling back to `APOLLO_API_KEY` so existing single-tenant deployments keep working. `CompanyIntegration.APIKey` is deliberately **not** read — it is not decrypt-on-read, so a key written through metadata sync returns the literal `$ENC$…` and produces a 401 that looks exactly like a wrong key. A credential that exists but will not parse fails `CONFIGURATION_ERROR` rather than silently borrowing another workspace's key from the environment.

### Verification

92 tests, driving the client through an injected `fetch` so the request bodies Apollo would receive are asserted directly — which matters more than usual, because a move that sends the wrong label array is invisible in the response. Nothing in the suite can reach api.apollo.io: a live read would spend credits and a live write would mutate a real account's lists with no undo.

The two enrichment actions are unchanged.

### DIFF
--- a/.changeset/apollo-lists-search-and-moves.md
+++ b/.changeset/apollo-lists-search-and-moves.md
@@ -1,0 +1,29 @@
+---
+"@memberjunction/actions-apollo": minor
+---
+
+Add Apollo's other half to this package: **list management, saved-record search and prospecting**, as seven new actions alongside the two enrichment actions it has always had.
+
+`ApolloGetListsAction` and `ApolloCreateListAction` read and create labels; `ApolloGetListAccountsAction` and `ApolloGetListContactsAction` page through a list's members; `ApolloSearchPeopleAction` searches Apollo's people database for net-new prospects; `ApolloMoveListAccountsAction` and `ApolloMoveListContactsAction` move records between lists. Together they are the outbound-campaign surface — build a list, drain it through a sequence of stages, and see what is where — which enrichment alone cannot express.
+
+These go to a **different Apollo base path** from enrichment: `api.apollo.io/api/v1` rather than `api.apollo.io/v1`. They are not interchangeable, and the same path under the wrong prefix 404s, so `ApolloRESTEndpoint` is declared beside the existing `ApolloAPIEndpoint` rather than derived from it.
+
+Five Apollo behaviours dictate the design, and each is what makes the naive implementation wrong:
+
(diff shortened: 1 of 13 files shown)


MS-BC:

PR #3492, opened 2026-08-05
feat(core-actions): Tavily Search and Read RSS Feed actions

MS-BC:
Two new actions in `@memberjunction/core-actions`, plus a config fix. Ported from AIDP's `TavilyTrendProvider` (104 LOC) and `RSSNewsTrendProvider` + `RSSNewsTrendScoring` (285 LOC).

Neither vendor is a system of record, so neither is a connector; neither has an upstream equivalent today.

### Tavily Search

`POST https://api.tavily.com/search` with `Authorization: Bearer tvly-…`. The header form, not the body `api_key` — the latter is legacy and puts the credential in the payload, where it lands in every request log along the way.

`topic: 'news'` is the only topic that returns `published_date` and the only one `days` applies to, so `Days` is accepted for news and produces a `Warnings` output param otherwise, rather than being silently ignored. `MaxResults` clamps at 20. Errors map to `INVALID_API_KEY` / `RATE_LIMITED` / `INVALID_REQUEST` / `API_ERROR` / `SEARCH_FAILED`, with `describeAxiosError` walking `detail` / `error` / `message` including the nested `{detail:{error}}` shape Tavily sometimes returns. Zero results is `Success: true` — a search that legitimately found nothing is not an error.

### Read RSS Feed

**Regex parsing over a strict XML parser is deliberate.** A strict parser fails the whole document over one unescaped ampersand, and unescaped ampersands are common in the wild. This costs that item's text, not the feed. The limit is honest and documented: an item whose own tag is never closed *is* swallowed by the lazy block match, so graceful degradation covers bad **content**, not bad **tags**.

**`stripHtml`'s ordering is load-bearing:** decode → strip tags → decode again → collapse. RSS descriptions overwhelmingly arrive with their HTML *escaped* (`&lt;p&gt;`), so decoding after stripping would leave a literal `<p>` in the output of nearly every feed item; the second pass finishes ampersands that were escaped twice.

One injected `now: Date` for the whole run, so the age filter and the recency score cannot disagree with each other mid-run.

### Left behind deliberately

CDP's trend-specific logic — Tavily's `scoreToStrength` / `inferPlatforms` and RSS's quartile `assignStrength`. Those are AIDP marketing judgments over the results; the actions return the raw numeric scores and let the consumer judge.

### Fix

`getCoreActionsConfig()` built `rawConfig` *inside* the no-config early return, so every environment-variable fallback was dead whenever no config file was present. The build is hoisted out via `const fileConfig = result?.config`. `tavilyApiKey` is added to the zod schema.

### Verification

103 new tests (60 RSS, 43 Tavily); 381 pass across the package; `tsc --noEmit` clean. Note `tsconfig.json` excludes `src/__tests__`, so the tests are not typechecked by that run.

### DIFF
--- a/.changeset/core-actions-tavily-and-rss.md
+++ b/.changeset/core-actions-tavily-and-rss.md
@@ -1,0 +1,19 @@
+---
+"@memberjunction/core-actions": minor
+---
+
+Add two web-read actions — **Tavily Search** and **Read RSS Feed** — and fix an environment-variable fallback that never worked.
+
+**Tavily Search** (`custom/web/tavily-search.action.ts`) searches through Tavily, whose results arrive as extracted page content rather than as snippets plus links to go fetch. Two capabilities distinguish it from the other search actions here: `Topic: 'news'` is the only topic that returns a publication date per result and the only one the `Days` window applies to, which is what makes recency-sensitive questions answerable; and `IncludeAnswer` returns a synthesized answer alongside the results, so a caller that only needs the conclusion does not have to run its own summarization pass. Authentication is `Authorization: Bearer <key>` — Tavily still accepts an `api_key` body field, but that form puts the credential in the payload and is not used.
+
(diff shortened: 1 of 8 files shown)


MS-BC:

PR #3493, opened 2026-08-05
fix(bizapps-social): Buffer createPost assets shape; add CredentialID and PlatformMetadata

MS-BC:
Three changes to the Buffer provider, one of them a live defect.

### Fix — the createPost assets shape

Buffer moved `createPost`'s input to `[AssetInput!]` (`[{ image: { url } }]`, one entry per attachment naming its kind) on 2026-05-25 and **rejects** the older `{ images: [...] }` object — which is what this package was sending. Any post with `ImageURLs`, `VideoURLs` or `MediaLink` was therefore being rejected by Buffer.

Reads are unaffected: Buffer still *returns* the object form. So the input shape is now its own type, `BufferAssetInput`, alongside the unchanged `BufferAssets` response type — sharing one type would break one direction or the other.

This rests on AIDP's production evidence (a dated test asserting the array form against a live account) and cannot be confirmed offline in this repo. Flagging that explicitly rather than claiming it verified.

Breaking for any caller passing a hand-built `assets` object to the protected `createBufferPost`; the action's own `ImageURLs`/`VideoURLs`/`MediaLink` params are unchanged.

### `CredentialID` on every Buffer action

Token resolution was strictly `BaseOAuthAction.getAccessToken()` → `this._companyIntegration?.AccessToken`, with no per-call override seam. That makes publishing to an employee's personal channel — or reading the queue of the person who owns it — impossible through this package.

With `CredentialID`, calls are made using the `accessToken` from that `MJ: Credentials` row. `CompanyIntegrationID` stays required: it identifies *which* Buffer integration this is, and organization and channel context still come from it. `CredentialID` only changes whose token the calls are made with.

A `CredentialID` that was supplied but cannot be read is **fatal** (`INVALID_CREDENTIAL`), never a silent fallback to the tenant token — falling back would publish under the wrong identity with nothing for the caller to notice. Unparseable `Values` is reported as a probable failed decrypt rather than as a malformed-JSON error the operator cannot act on. The token itself never travels through action params; only its id does.

Because it lives on `bufferCommonParams`, reads and writes use the same identity — which matters, since channel ids returned to one token may not be visible to another.

### `PlatformMetadata` on `Create Post (Buffer)`

Passes Buffer's per-service extras through to the mutation. `{ "linkedin": { "annotations": [...] } }` is how a LinkedIn @mention survives the trip; without it the post publishes as plain text with the mention spelled out.

Accepted as an object or a JSON string, since both forms arrive (typed callers and agent/UI inputs). Passed through untouched rather than modelled — its shape belongs to whichever network the channel points at, and Buffer extends it independently of this package. Unparseable input **fails** rather than being dropped, for the same reason the credential failure is fatal: quietly posting without the metadata publishes something other than what the caller composed.

### Verification

27 tests covering the credential override (including SQL quote escaping in the lookup filter, and that a failed lookup makes zero GraphQL calls), the metadata passthrough, and the asset shape. Full package suite green.

### DIFF
--- a/.changeset/buffer-identity-metadata-and-assets.md
+++ b/.changeset/buffer-identity-metadata-and-assets.md
@@ -1,0 +1,15 @@
+---
+"@memberjunction/actions-bizapps-social": minor
+---
+
+Buffer: publish as another identity, pass per-service metadata, and fix the createPost assets shape.
+
+**`CredentialID`** is now accepted by every Buffer action. When given, the calls are made with the `accessToken` from that `MJ: Credentials` row instead of the CompanyIntegration's own token — which is what publishing to an employee's personal channel, or reading the queue of the person who owns it, requires. `CompanyIntegrationID` stays required: it is what identifies which Buffer integration this is, and the organization and channel context still come from it. A `CredentialID` that was supplied but cannot be read is **fatal** (`INVALID_CREDENTIAL`), never a silent fallback to the tenant token — falling back would publish under the wrong identity with nothing for the caller to notice. The token itself never travels through action params; only its id does.
+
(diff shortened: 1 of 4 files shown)


MS-BC:

PR #3495, opened 2026-08-05
docs(site): redesign the docs landing page from the approved mockup

MattC-BC:
## Summary

Implements the *MemberJunction Docs Landing* mockup on the Starlight docs site. The approach is to **restyle Starlight's own chrome rather than replace it**, so search (Pagefind), the theme toggle, and nav stay real controls instead of becoming decorative, and the new header carries to every page rather than just the landing.

- **Landing page** — hero with version/license pills, a `Download & Install` CTA, an install panel carrying the full sequence through `node InstallMJ.js`, the MJ Central card, the documentation card grid, and a brand bar footer with the ™ / license line.
- **Site-wide chrome** — navy header held across *both* themes, search grouped right with the other controls, no hairline, accent ramp moved from the old prototype indigo to MJ brand blue, and a documented-line version pill via a `SiteTitle` override.
- **Brand assets** — the horizontal 948×132 lockup replaces the 200px stacked pair in the header, the footer, and the README.
- **Content corrections** — several mockup claims were stale; each was checked against the repo rather than carried over.

## Notes for review

**The install path is documented accurately.** There is deliberately no direct `.zip` download anywhere. `main.zip` is a source archive that happens to contain `InstallMJ.js`; it is not an installer, and labelling it as one misleads. The CTA is deep-linked to `/overview/#download--install` because that page renders the whole README (~28,000px) with the install section only 9% down, so a bare page link buried it.

**The copy button has a single source of truth.** It reads the rendered `<pre>` back out and strips the `$` prompts, so the visible commands and the clipboard payload cannot drift apart.

**Corrections made to the mockup's copy, all verified against the repo:**

| Claim | Was | Now |
|---|---|---|
| AI providers | "Fifteen-plus" | "Thirty-plus" (31 provider packages) |
| Packages | "290+" | "300+" (308 actual) |
| Prerequisites | SQL Server only | adds PostgreSQL |
| README section | "Quick Start" | "Download & Install" |

**Two theming details worth knowing, both commented in the source:**

- `--sl-color-white` is contrast-relative in Starlight, resolving to `hsl(224,10%,10%)` in light mode. Using it for a surface paints black cards on a white page.
- Starlight's `--sl-color-bg-nav` token *is* the header-background knob, but it has three consumers. Setting it navy also turns the mobile "On this page" bar navy, where it butts against the header with dark text and becomes unreadable. One targeted property on `.header` has no blast radius. This was tested and reverted, not assumed.

**Logo change rationale:** the previous dark variant was a flat white silhouette (8-bit gray+alpha) that dropped the two-tone wordmark, and the light `.webp` had an opaque white background that rendered as a white rectangle on a dark page. Both new files are RGBA and measured at 71% fully transparent, so a theme mismatch now degrades to weak contrast instead.

## Open items

- **MJ Central URL** — the mockup specified `memberjunction.com/mj-central`; the site has always used `central.memberjunction.com`. Both resolve. Left as-is pending a call from the team, with a comment in the MDX.
- **`getting-started.mdx` still says "290+ packages"**, so it now disagrees with the landing's corrected "300+".
- **Starlight's `tip` asides and the MJC sidebar badge render purple**, which clashes with the new blue. Pre-existing, untouched here.
- **A literal file download** — a button that navigates to docs is not itself a download. If that matters, an honestly-labelled secondary "Download source (.zip)" could sit beside the primary CTA. Not added without a decision.

## Test plan

- [ ] `cd docs-site && npm run dev` — landing renders in light and dark
- [ ] Header: navy in both themes, search grouped right, ⌘K chip has a single border
- [ ] Copy button on the install panel yields all six lines with prompts stripped
- [ ] `Download & Install` lands on the `/overview/` install section, clear of the fixed header
- [ ] Interior pages (`/getting-started/`, `/overview/`) keep their sidebar, TOC, pagination, and edit link
- [ ] Mobile ~430px: logo, version pill, search icon, and hamburger all fit; the "On this page" bar is readable
- [ ] Footer appears on the landing only, flush to the page bottom
- [ ] `/overview/` body logo loads (it 404s until this branch is pushed, since ingest pins images to the build commit)


cadam11 (approved):
LGTM

Issue #3384, opened 2026-07-31
CodeGen scope model: cleanup paths treat "out of scope" as "deleted" (low priority, filed after the v5.50.0 includeSchemas review)


MarceloT-BC:
> **Updated 2026-08-05.** Re-verified against `origin/next` @ `3431f9bd72`: **all seven findings are still live and unchanged.** `sql_codegen.ts` and `manage-metadata.ts` have moved since the original audit (layered base views, #2570 smart-field-identification) but none of the audited behaviours changed; `manage-metadata.ts` line refs shift by ~9 and are corrected below. Two open PRs bear on the reachability argument — see the new note under *Why most of this is unreachable today* and the revised ordering note — and the "data-driven scoping proposal" this issue referred to in the abstract is now concrete as #3499.

## Summary

CodeGen decides what is "in scope" for a run, and several of its cleanup paths treat "out of scope" as "deleted." With `includeSchemas` now shipped in v5.50.0 (#3350), narrowing scope in the core repo is a supported configuration for the first time — and it reaches those paths.

**Priority: low, and deliberately filed that way.** This is not currently breaking anyone and there is no rush. It needs someone to deliberately narrow scope in the core repo, which only became possible in v5.50.0 and which nobody is doing yet. Everything it does is recoverable from git. It is filed because the option is newly documented, the destruction is silent while it happens, and the fixes are small — not because it is about to strike.

The audit turned up seven findings. **Three are worth doing as one small change set; four are real but not worth acting on now**, and each of those says why.

---

## Why most of this is unreachable today

Reachability depends entirely on how CodeGen is configured in the two places it runs, so it's worth stating first.

| | Core MJ repo | An installed app |
|---|---|---|
| Output directories | Core **and** non-core — MJServer resolvers, MJCoreEntities subclasses, Explorer entity forms | Only its own packages |
| `excludeSchemas` | `sys`, `staging` | `sys`, `staging`, `dbo`, **`__mj`** (+ other apps, when several are present) |
| Core output directories configured? | **Yes** | **No** |

This explains something that would otherwise look like a live bug: every app excludes the core schema, and that has always been safe — because an app configures no core output directories, the core generation passes simply never run for it. The core repo is the mirror image: it owns the core output directories and does not exclude core.

**So narrowing scope is routine and harmless in an app, and consequential only in the core repo.**

A guard along the lines of "error if core is excluded while core output directories are configured" was considered and **rejected** for exactly this reason — it would fire on every legitimate app run.

[...]

Issue #3273, opened 2026-07-24
Proposal: `mj app link` — local development linking for MJ Open Apps


AN-BC:
> **Revised 2026-07-24.** The original proposal recommended raw symlinks into `node_modules`. **That mechanism is wrong** — we disproved it empirically (see "Why symlinks fail" below), and [@MarceloT-BC](https://github.com/MarceloT-BC) had already reached the same conclusion with a working implementation in **[MJDev](https://github.com/MemberJunction/MJDev)** ([v0.0.3](https://github.com/MemberJunction/MJDev/releases/tag/v0.0.3)). This issue now proposes bringing **MJDev's mechanism** into the MJ CLI, plus the pieces MJDev deliberately left out.

## Problem

MJ Open Apps (separate repos, each owning a SQL schema + migrations + npm packages) can't be developed against each other locally. Published apps install from GitHub/npm; unpublished ones (e.g. `bizapps-accounting`) can't be consumed at all. Every workaround fails:

| Approach | Defect |
|---|---|
| `npm link` | Global invisible state; destroyed by `npm install`. |
| `file:` deps | Committed path assumption breaks CI. npm reads the linked package's manifest and 404s on unpublished siblings (`"@mj-biz-apps/accounting-entities": "0.1.0"`). |
| Raw symlinks | **Proven broken — see below.** |
| Hand-maintained umbrella workspace | Forces one dependency tree; requires cloning everything. |

## Why symlinks fail (empirical, 2026-07-24)

We linked `bizapps-accounting` into `bizapps-orders` with raw symlinks and started MJAPI. Node resolves a symlinked package's dependencies from its **realpath**, so the producer keeps its own copies of every shared library. Result, in order:

1. `CannotDetermineGraphQLTypeError: Cannot determine GraphQL input type for argument named 'FiscalYear'` — the generated resolver was **correct** (`@Arg('FiscalYear', () => Int)`), but that `Int` came from accounting's `type-graphql` copy while the schema was built with orders'. Different object identity.
2. After deduping that: `Schema must contain uniquely named types but contains multiple types named "RunViewByIDInput"` — two copies of `@memberjunction/server` each registering the same GraphQL type.
3. The API only booted after hand-deduping **255 `@memberjunction/*` packages plus `type-graphql` and `graphql`**.

In other words: symlinks work only if you manually reproduce what npm workspace membership does for free. This independently confirms MJDev's [ADR-001](https://github.com/MemberJunction/MJDev).

[...]

PR #3499, opened 2026-08-06
docs: CodeGen OpenApp-aware scope proposal

jordanfanapour:
## Summary

- Proposes deriving CodeGen's schema scope from `__mj.OpenApp` data instead of hand-maintained `excludeSchemas` / `entityPackageName` config maps
- Three-way classification: **mine** (declared in `mj-app.json`, generate), **not mine** (installed via `mj app install`, exclude), **unowned** (third-party, exclude by default)
- Recommends a `developingSchemas` config override in `mj.config.cjs` for developer-explicit scope control
- Proposes `--branch` argument for `mj app install/update` to support pre-release dependency testing
- Documents recommended development convention: use migrations for your own app, `mj app install` only for dependencies

## Context

This proposal emerged from the review discussion around PR #3350 (type-availability gate + `includeSchemas`). The core observation: CodeGen has no concept of "installed app" vs "app I'm actively developing," and multiple bugs (#3350, #3457) trace back to that gap. Rather than continuing to patch symptoms, this proposes making CodeGen OpenApp-aware.

## Related Issues

- #3350 — type-availability gate (merged, complementary to this proposal)
- #3384 — cleanup paths treat "out of scope" as "deleted" (should be fixed first/in parallel)
- #3273 — `mj app link` for local dev linking (complementary)
- #3457 — installer writes app schema into `excludeSchemas`

## Open Design Questions

1. `developingSchemas` config vs `Development` status on OpenApp table — one argument favors status for dev/prod parity; another favors config for explicitness
2. How to handle "unowned" third-party schemas — exclude by default with opt-in, or auto-classify from catalog residual
3. Sequencing relative to #3384 — auto-derived scope makes destructive cleanup paths more reachable

## Test plan

- [ ] Team review of the proposal document
- [ ] Align on open design questions before implementation begins
- [ ] Implementation PR(s) to follow after consensus


MarceloT-BC:
Gonna give this a review. You should probably check with @cadam11  in case his work touches this. His plans include changes to registration and dev linking that will likely effect this.

MarceloT-BC:
Read this against the #3384 scope audit and against #3487, which is in flight on the same surface. The direction is right and several of the specifics are better than they first look, so leading with those before the concerns.

**What holds up.** Targeting `getExternalEntitySchemas()` rather than `excludeSchemas` is the correct lever — it's the emission filter, and it's where #3487 independently concluded the installed-app suppression belongs. The "mine = the current project's `mj-app.json`" definition also works, and for a reason worth stating explicitly in the doc: codegen for a dev-linked app runs **in the app**, not at the MJ root, so the current project genuinely is the app. I checked a live dev instance to make sure the classification lands right at the MJ level — three apps dev-linked, three app schemas in `sys.schemas`, zero rows in `__mj.OpenApp` — so they fall to "unowned → excluded," which is exactly what mjdev writes by hand in its overlay today. The derivation agrees with current behaviour rather than fighting it. And sequencing #3384 first matches what that audit itself concluded: it flagged this proposal by description and recommended the cleanup fix as a prerequisite rather than a follow-up, for the same reason given here.

**The main thing I'd want reconsidered: the value proposition is narrower than the problem statement implies.** Both bugs cited as motivation are `excludeSchemas` bugs — #3350 was the relationship walker not consulting it, #3457 was the installer writing into it and killing entity discovery. What the proposal derives is `getExternalEntitySchemas()`, the `entityPackageName` half. So it automates the map that wasn't producing failures and leaves the one that was still hand-maintained. It also can't fully automate even that half: dev-linked apps have no `OpenApp` rows, so their `entityPackageName` entries (`__mj_BizAppsAccounting → @mj-biz-apps/accounting-entities` and friends) still have to be written by tooling or by hand indefinitely. Worth restating the claimed benefit as "removes the installed-app half of the config burden," and dropping the "may subsume #3457" line in the sequencing section — #3487 handles PG schema-name casing, `includeSchemas` positive scope, shared-schema co-ownership, and config-editor safety, none of which auto-derivation reaches.

**Scope stops being reproducible.** Today the scope lives in git: same repo, same config, same output. After this, generated code depends on which database you point at, and CodeGen picks up a hard runtime dependency on `__mj.OpenApp` being present and readable. A fresh database, a pre-`OpenApp`-migration database, or a version skew would silently change scope rather than failing. Worth deciding deliberately, and worth a defined fallback for "table absent."

**Smaller, but each wants an answer in the doc:**

- Only `Active` is mapped. `Installing`, `Upgrading`, `Error`, `Disabled`, `Removing` are unspecified, and the `Error` case is the awkward one — a failed install leaves an `Error` row, the schema falls to "unowned," gets excluded, and codegen is frequently what you need to recover from that state.
- Shared schemas have no cell in the table. Two apps on one schema is real; #3487 needed `AnotherInstalledAppSelfManages` and carries a `schemaShared` flag for it. One row per app, one decision per schema — worth reusing #3487's rule rather than minting a second one.
- The binary classification can't express #3490. `excludeSchemas` gates three independent concerns (discovery, SQL ownership, emission) and `entityPackageName` gates only the third; `bizapps-common` owns its SQL but not its entity seed and fits neither branch. An ownership axis (mine / not-mine / unowned) is orthogonal to that concern axis, so this proposal and #3490 are complementary rather than alternatives — probably worth saying so directly.
- `--branch` for `mj app install/update` is genuinely useful but unrelated to scope derivation, and it makes `OpenApp.Version` stop identifying what's installed — under this proposal that same row becomes authoritative for what gets generated. Suggest splitting it into its own issue so it can be decided on its own merits.

**On Open Question 1:** the config override isn't really optional either way. Dev-linked apps are configured, not registered, so `developingSchemas` (or an equivalent overlay) stays load-bearing permanently — and a `Development` status can't cover them, since there's no row to set a status on. That seems like it settles the question in favour of config, but it's worth stating as a consequence rather than a preference.

No blocking objection to the direction. Mostly: right lever, narrower payoff than advertised, and a few classification cells that need filling before implementation starts.

plans/codegen/openapp-aware-scope.md
```diff
@@ -0,0 +1,123 @@
...
+
+- **Dangling reverse-relationship references** (#3350): The relationship walker reads `Metadata.Entities` (all entities in the DB), but the old `isRelatedTypeOutOfScope` heuristic only checked the `entityPackageName` map — not the `excludeSchemas` filter. Entities excluded by schema but not in the package map still produced uncompilable `@Field` references. Fixed defensively in #3350 by Change A (type-availability gate).
+- **Manual config burden**: Developers must hand-maintain `excludeSchemas` and `entityPackageName` maps that grow O(N²) with installed apps and can't name schemas the developer doesn't know about (e.g., a client's own schemas in a deployed instance).
+- **Installer/CodeGen conflict** (#3457): `mj app install` writes the app's schema into `excludeSchemas`, which also gates entity *discovery* — silently preventing CodeGen from creating that app's entities if the app relies on host CodeGen for metadata.
+
+## Proposed Solution
+
+Derive CodeGen's scope from data (`__mj.OpenApp`) rather than manual config. The classification:
```
cadam11, line 18: Field evidence for this, from Amith's multi-repo workspace run today: the class-manifest walker in bizapps-issues/packages/Server resolved sibling repos for the first time and widened from "14 packages walked, 2 contain @RegisterClass" to "16 walked, 4 contain", added 62 entity classes, and rewrote package.json mid-build. The build then failed because the new declarations weren't installed yet.

Open question to add to the list: does the @RegisterClass manifest walk adopt this same mine/not-mine classification? Today its scope is defined by what realpathSync can reach, which means the same repo produces a different committed manifest standalone vs inside a workspace. Scope-by-declaration should govern both the schema side and the manifest walk, or the workspace keeps that nondeterminism.

plans/codegen/openapp-aware-scope.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Scope and Sequencing
+
+### This proposal covers:
+- Deriving `getExternalEntitySchemas()` from `__mj.OpenApp` + optional `developingSchemas` config
+- `--branch` argument for `mj app install/update`
+- Documentation of the development convention
+
+### This proposal does NOT cover:
+- The destructive cleanup paths (#3384) — those should be addressed first or in parallel, since auto-derived scope narrows the entity list and makes the "out of scope treated as deleted" bug more reachable
```
cadam11, line 98: Upgrading this sequencing note from theoretical to observed: today's workspace run showed scope can widen without anyone asking for it (sibling repos becoming resolvable was enough). The same effect reversed, building outside the workspace after building inside it, narrows scope again, and with #3384 unfixed that is exactly the "out of scope treated as deleted" path. Agree #3384 goes first.

plans/codegen/openapp-aware-scope.md
```diff
@@ -0,0 +1,123 @@
...
+- `mj app link` for local development linking (#3273) — complementary but independent
+- The `excludeSchemas` installer bug (#3457) — becomes less impactful with auto-derived scope but should still be fixed
+
+### Recommended sequencing:
+1. **Fix #3384 items 1-3** (cleanup paths, empty include validation, CRUD validator) — prerequisite for any auto-derived scope
```
cadam11, line 103: Proposed slice 1 for the implementation, small and shippable ahead of the design consensus: when a codegen run adds package declarations, stop loudly instead of failing mid-build. Detect the package.json mutation, print what was added and that an install is now required, exit nonzero. This guard isn't throwaway under the final design either: whenever generation legitimately adds a declaration for "mine" scope, the declaration-added-but-not-installed ordering hazard still exists. It would have turned today's cryptic workspace build failure into a two-line instruction.

cadam11 (commented):
Inline notes from today's first real field test: Amith ran the full multi-repo dev workspace (31 packages, 1214/1215 tests) and one of his three findings lands squarely in this proposal's territory. Overall this is the right frame for what we hit. Two additions and one implementation suggestion inline.

PR #3500, opened 2026-08-06
v6.1.0-edge.0

MS-BC:
# MemberJunction v6.1.0-edge.0 — Agent Persistence, Workflow Extensions, and Platform Hardening

## New Features

- **Layered base views**: entities can now have both a CodeGen-generated base view AND a custom wrapper view over it — `Entity.GeneratedBaseViewName` tells CodeGen to write its output under a separate name so applications own a thin custom layer while schema changes (new foreign keys, new columns) still regenerate automatically underneath
- **Entity Action workflow extensions**: `EntityAction` is now a general workflow-hook substrate with scope binding (`ScopeEntityID`/`ScopeRecordID`), deterministic ordering (`Sequence`), per-binding logging control (`LoggingMode`), and provenance columns on `ActionExecutionLog` — plus `Entity Object Data` parameter value type that passes `entity.GetAll()` for serializable payloads
- **Entity direct-SQL flags**: three new `Entity` columns (`AllowDirectSQLInsert`, `AllowDirectSQLUpdate`, `AllowDirectSQLDelete`) make explicit which entities may be written by SQL outside `BaseEntity`, with CHECK constraints enforcing that direct-DML entities have record tracking and cache trust disabled
- **API key row filter columns**: `APIKeyScope` and `APIApplicationScope` gain nullable `RowFilterID` FKs to the `RowLevelSecurityFilter` catalog, so API key scopes can declare row-level filter intent (enforcement ships in a follow-up)
- **Pluggable search scope permission resolver**: `SearchScopePermissionResolverBase` + `SEARCH_SCOPE_PERMISSION_RESOLVER_KEY` let applications replace MJ's role-based scope-authorization policy at all five search call sites (including the agent-facing `__Scoped_Search` core action)
- **Theme-aware registry components**: `ComponentStyles` gains `sequentialScale`/`divergingScale` visualization ramps backed by new `--mj-viz-seq-*`/`--mj-viz-div-*` design tokens; `ComponentSpec` gains `styleOverrides` so user-requested styling travels as spec data instead of color literals
- **Mobile records UX**: MJ Explorer now has a phone-optimized record view — headerless golden-layout below 768 px, a record bar with entity icon and open count, a bottom-sheet record switcher, and layout flattening that preserves desktop split layouts across sessions
- **ElevenLabs per-session voice**: `RealtimeSessionParams.Config.voice` is now forwarded as `tts.voice_id`, so a managed ElevenLabs agent can express distinct voices per session; includes fixes for agent-fingerprint drift, eventual-consistency duplicate creation, and intra-process concurrent ensure races

## Improvements

- **CodeGen introspection view perf**: narrowed `vwSQLColumnsAndEntityFields` from `sys.all_columns`/`sys.all_objects` to the user-object catalog views, dropping a 500-table cold CodeGen run from 190 s → 106 s (−44%) with byte-for-byte identical output
- **Entity semantic search standardized** (Tier 1): retired the bespoke per-entity embedding machinery in `AIEngine` in favor of `Provider.SearchEntity` backed by daily-synced EntityDocuments; "Find Best Agent/Action" core actions are now thin wrappers over the unified search pipeline
- **Integration connectors decoupled from monorepo**: the 36 vendor connectors removed from `@memberjunction/integration-connectors` — each now ships as its own Open App in the [MemberJunction/Integrations](https://github.com/MemberJunction/Integrations) repo; the shared base-class layer (`BaseExternalDataSourceConnector` et al.) remains
- **CodeGen smart-field search guardrails**: search identification now defaults OFF with a narrow allowlist, narrative/filter-field blocklist, and a runtime enforcement module so flag drift from LLM output is rejected in code
- **pnpm workspace cutover**: all packages now declare every dependency they import; previously npm hoisting silently supplied undeclared transitive deps
- **Angular DOM testing — Phase 4**: DOM test coverage ratchet pushed from 185 → 137 uncovered components; every shipped Angular library tree (Explorer, Generic, Bootstrap) now has a CI gate; `test:types` spec type-check gate closes the transpile-only hole where real type errors rode green
- **Session landing decoupled from Sequence order**: the Explorer shell now picks the landing app by `Application.DefaultSequence` rather than the user-sortable `UserApplication.Sequence`, so reordering the app switcher no longer changes where sessions open

## Bug Fixes

- **SQLServerDataProvider transaction poison**: `BeginTransaction` published an un-begun transaction and never cleared it on failure, permanently poisoning the provider and destroying all agent-run / step / prompt-run persistence; PostgreSQL analogue leaked a pool client on a failed `BEGIN`
- **`AIPromptRun.AgentID` never populated**: `BaseAgent` now sets `AIPromptParams.agentId` at all three call sites and `AIPromptRunner` stamps it, restoring the documented attribution column
- **AgentRunner conversation artifact hydration**: `RunAgent` never hydrated conversation artifacts for callers holding a `conversationDetailId`, giving agents a silently artifact-blind view; fixed with a deduplicated hydration seam; `RunAgentInConversation` now always sets the field so `[]` means "scanned, none found"
- **`statusUpdates` subscription session-hijack** (B49): the push-subscription filter matched only on client-supplied `sessionId` with no identity check; it now additionally requires the push owner's user ID to match the subscriber's authenticated connection identity, enforced at a single helper all ~22 publish sites route through
- **Cache refresh interval unit**: metadata cache was refreshing every ~50 hours instead of the configured 3 minutes due to a ms/s unit mismatch between `MJServer` and `SQLServerDataProvider`
- **UUID primary key case mismatch in BaseEngine cache**: `findEntityIndexByPrimaryKeys` compared UUIDs with strict `===`, causing case-mismatched IDs to append duplicate rows into the engine cache
- **SQL parameterization hardening**: `ReportResolver` and `GenericDatabaseProvider.CheckRecordRLS` now use parameterized queries / quote-escaped values; `RowLevelSecurityFilterInfo.MarkupFilterText` no longer lets `undefined` user-property values through as the literal string `"undefined"` and now doubles embedded single quotes
- **BaseDashboard loading screen never released on error**: `initDashboard`/`loadData` errors left `NotifyLoadComplete` unfired, hanging the Explorer loading screen; now wrapped in `try/finally`; `BaseResourceComponent` watchdog now fails open
- **CI migration gate concurrency**: the `Ensure migrations are valid` workflow used a repo-wide constant concurrency group, so any push cancelled every other PR's in-flight migration check; now keyed per-ref like all other PR-scoped workflows
- **`ng-file-storage` sibling package version pin**: corrected `@memberjunction/ng-notifications` pin from `5.51.0` → `6.0.0` so the workspace package links correctly instead of resolving the published v5 from the registry


MS-BC:
## Step 4 integration-suite triage (DEPLOYMENT.md §4.6)

**Deterministic tier — `Integration Tests — Deterministic`: 55/55, 0 failures** (run with `RUN_MUTATION_TESTS=1`). Hard gate green.

**Live-model tier — `Integration Tests — Live Model`: 14/15.** Per §4.6 this tier is *required to run and be triaged, not required to be 15/15*. Triage of the single red:

### `agent-memory-guards.MG3` — `model-noncompliance:` accepted

MG3's first failure was **not** variance — it exposed a real defect, now fixed in this PR (`fix(agents): memory writes must never supersede across a scope boundary`).

`MemoryWriteManager` tracked notes written during a run by ID alone. Because `FindSimilarAgentNotes` deliberately does not filter out agent-wide (`UserID IS NULL`) notes when a userId is supplied, a user-scoped write saw an agent-scoped note written moments earlier in the same run and superseded it. `supersedeOwnNote` rewrites `Note`/`Type` but never `UserID` — so one memory was lost outright **and the survivor held user-specific content at agent-wide scope, readable by every other user of that agent.**

Row-level evidence, captured before the check's teardown purge:

**Before the fix** — 2 of 3 attempts leaked scope:
```
MG3-a1 user scoped memory || UserID=NULL          ← user content, agent-wide
MG3-a2 user scoped memory || UserID=ECAFCCEC-...
MG3-a3 user scoped memory || UserID=NULL          ← user content, agent-wide
```

**After the fix** — 3 of 3 correctly scoped:
```
MG3-a1 user scoped memory || UserID=ECAFCCEC-...
MG3-a2 user scoped memory || UserID=ECAFCCEC-...
MG3-a3 user scoped memory || UserID=ECAFCCEC-...
```

This is dispositive independent of the row count: a surviving cross-scope supersede would necessarily read `UserID=NULL`, since `supersedeOwnNote` never touches the scope. It reads the real user on every attempt, so the supersede is gone.

The residual red is the *agent*-scoped note not landing, so the check finds one note where it wants two. That is a memory that doesn't get written — no data loss beyond the note, no corruption, no exposure. `model-noncompliance:` after 3 bounded retries is the accurate classification, and §4.6 permits accepting it.

Also fixed here: the **AT9 oracle** asserted on a sentinel string that the check's own `grep` echoes into the next turn's prompt (artifact tool results are appended as a user turn), so it fired on its own success. Now asserts on the artifact body, which is what actually distinguishes inlining from tools-only delivery. IT57 went 2/9 → 9/9.

### Four product defects fixed in this release
1. `SQLServerDataProvider.BeginTransaction` published an un-begun transaction (+ PostgreSQL analogue) — 87/87, 149/149
2. `AIPromptRun.AgentID` was never populated — 255/255
3. `AgentRunner.RunAgent` never hydrated conversation artifacts — 1785/1785
4. Memory writes superseded across scope boundaries (above) — 1785/1785, 3 new regression tests

### CI on head SHA `3c7f4083ae`
Unit Tests ✅ · Integration Tier ✅ · Ensure migrations are valid ✅ · Release Lines Guard ✅ · Docs site CI ✅ · hard-coded-UUID advisory clean. `build.yml` and `dependency-check.yml` are absent by design (`next`-only triggers, per the Step 9 table).

PostgreSQL parity: 34 migrations applied / 0 failed / 384 entities on the gate DB; SS v6 7 files ↔ PG v6 7 files; Explorer browser test 12/12 against the PG gate database.

PR #3456, opened 2026-08-05
Plan (FINAL v8, build-ready): one DAG engine + the unified workflow program — task graphs, TaskGraphSpec, Save as Workflow, Workflow UX

AN-BC:
## Status: ✅ Final — approved for build

Design-only PR, now finalized after two external review rounds with every finding dispositioned and applied. This PR is the **design home of the unified-workflow program**; implementation ships as separate phase-gated PRs that cite these documents. One production fix already shipped out of the review process (#3525, merged).

## Contents

1. **`plans/task-graph-primitive.md`** — **v8 FINAL**. The WHAT: task graphs as a first-class primitive; one DAG execution engine for agents, flows, and durable work. Decisions D1–D20, Phases 0–5 + parallel Track R, risks R1–R6, testing strategy, source index pinned to baseline `next@d26e202e7`.
2. **`plans/unified-workflow.md`** — v2 companion. The program map: WHAT × WHEN model, division-of-labor doctrine, Tracks A–F and sequencing.
3. **`mockups/workflow-ux/phase5-overview-v1.html`** — the Phase 5 design source: editor canvas, runtime overlay, chat plan cards, Create Workflow front door.

## The framing

- **WHAT runs is a DAG** — one graph execution engine. A **Loop agent spins up a DAG dynamically** (`nextStep.type: 'Tasks'`); a **Flow agent IS a DAG** authored at design time, running deterministically through the same execution code. User-facing name: **Workflow** (D18) — and end-user surfaces never say graph/DAG/node.
- **WHEN it runs is the trigger layer** — Entity Actions (#3408), Scheduled Jobs, User Routines, on-demand, agent invocation. Program tracks D–F (companion doc) build the trigger vocabulary, the stored WorkflowSpec (which wraps `TaskGraphSpec` by composition), and the remaining front doors.

## The architecture in ten decisions (of twenty)

- **D1/D2** — execution is server-side and invocation-agnostic; submission (validate + persist) is split from a **durable dispatcher** (CAS claim protocol: `ClaimedBy`/`ClaimExpiresAt`, heartbeat, startup + periodic reconciliation). Explorer, Slack/Teams, scheduled, and headless channels execute identically; clients are observers.
- **D3** — `enableTaskGraphs` in the Loop **agent-type params bag**, **default OFF** (review round 2); launch opt-ins: Sage, Query Builder, Research Agent (+ sub-agents). Enforced in validation, not just prompt docs.
[...]

PR #3525, opened 2026-08-06
fix(actions): implement Action Filter evaluation — RunSingleFilter was a return-true stub

AN-BC:
## What

Hotfix pulled out of the task-graph program plan (#3456) per the second review round: both reviewers independently verified that `ActionEngineServer.RunSingleFilter` was `return true` with a `// temp stub` comment (`packages/Actions/Engine/src/generic/ActionEngine.ts:308-310`) — **every Action Filter was a no-op**, so every Entity Action binding fired on every save regardless of its filters. This is a live correctness bug shipped independently of (and ahead of) the plan's approval gate.

## How

`RunSingleFilter` now resolves in order:

1. **ClassFactory subclass** — a `BaseActionFilter` subclass registered under the filter's ID (the CodeGen / custom-subclass pattern already documented on `BaseActionFilter`) wins when present.
2. **Inline `Code` evaluation** — otherwise the filter's `Code` column (JavaScript that evaluates to `true` = allow / `false` = prevent, per the column's own contract) is compiled once per row version and executed with an `ActionFilterContext { params, filter, result }` argument. Code may `return` its verdict or assign `ActionFilterContext.result`. Same `new Function` execution pattern as the adjacent entity-action `SafeEvalScript` (`EntityActionInvocationTypes.ts`), with a bounded `MJLruCache` keyed by `ID + __mj_UpdatedAt` so an edited filter recompiles instead of serving a stale compilation.

**Failure semantics are fail-closed**: a filter that throws, yields a non-boolean, or has no evaluable logic **prevents** the action and logs the reason. A broken gate must not silently allow execution — silently allowing everything is exactly the bug this replaces. No `ActionFilter` rows exist in shipped metadata today, so no existing behavior changes; this makes the machinery real for the filters Track A will seed (`Field Changed` / `Field Changed To Value`).

## Tests

Replaced the old stub assertion (`should return true (stub implementation)`) with 12 cases: both verdict channels, context exposure (`params`/`filter`), async filter code, all three fail-closed paths (throw / non-boolean / empty code), cache versioning (recompile on row-version change; cached serve on same version), and ClassFactory subclass precedence incl. non-boolean subclass verdict coercion.

`packages/Actions/Engine`: **9 files, 191 tests passing** (`vitest run`); package builds clean.

## Notes

- Relates to #3456 (plan §Track A item 1 context) — this PR intentionally does **not** include the fuller Track A scope (filter evaluation engine docs, `OldValues`/`NewValues` contract, seeded filters); it makes the existing contract function.
- Per @AN-BC on #3456: after this lands on `next`, an LTS backport to 5.x should follow — cc @cadam11.

https://claude.ai/code/session_01T5THLgKF91TYoHg38x8cmg

---
_Generated by [Claude Code](https://claude.ai/code/session_01T5THLgKF91TYoHg38x8cmg)_


cadam11:
Two rulings, since I was cc'd on this via #3456:

1. Fail-closed is correct, keep it. Silently allowing everything is exactly the bug this replaces, and a filter that throws or returns a non-boolean should block the action and say so loudly. Since no shipped metadata has ActionFilter rows, nothing changes behavior today.

2. Backport: yes. Once this merges to next, put the `backport lts/5` label on it so the bot cherry-picks it to the line. Code-only patch with a latent behavior change fits the line's patch tier, and unlike #3410 there is nothing in this PR we want to hold back, so the whole-PR cherry-pick is fine.

Before merge: add a changeset (patch for the actions engine package), and kick CI again. Most checks on the latest commit show cancelled, only the SQL integration suite completed.

AN-BC:
@cadam11 Both pre-merge asks done in `2bafb6b8`: changeset added (`@memberjunction/actions`: patch) and CI rekicked — the push retriggers all five checks on the new head rather than re-running the cancelled jobs on the superseded commit (those cancellations traced to a GitHub Actions "Service Unavailable" outage resolving action downloads, not to the change). Fail-closed stays as ruled. I'll apply the `backport lts/5` label once this merges to `next` so the bot picks up the whole-PR cherry-pick.

---

AN-BC (commented):
LGTM

AN-BC:
@cadam11 merged and labeled

PR #3542, opened 2026-08-06
Test-coverage overhaul + SafeExpressionEvaluator RCE fix: real Skipped status, armed CI lanes, ~1,500 unit tests, new integration bundles

AN-BC:
Implements the full roadmap from the test-coverage evaluation, and — across three rounds of adversarial review of the PR itself — closes the defects that review surfaced. Four parts: close the silent non-execution holes in the integration tier, backfill the highest-risk unit coverage, harden the release path, and fix what the reviews found (a live sandbox escape, then a grammar regression the escape fix introduced).

The full monorepo builds (`300/300` Turbo tasks, plus `273/273` on the bootstrap/manifest postbuild pass). Repo gates run locally and green: `check:standards`, `check:esm`, `check:changeset`, `check:claude-md`, and the spec anti-pattern lint. Migration ordering/filenames, DOM placement/classification and the driver lanes are green in CI.

## Security — `SafeExpressionEvaluator`

**Round 2 found the original "hardening" did not hold.** It compiled the caller's expression with `new Function(...)` behind a textual token denylist, and a split-token payload walked straight through — `[]["cons"+"tructor"]["cons"+"tructor"]("return process.pid")()` spells none of the banned words yet climbs `[].constructor.constructor` to the `Function` constructor and reaches `process`. Confirmed arbitrary code execution from any metadata-authored expression (field rules, flow/loop agent conditions, task-graph conditions).

The fix replaces the denylist with a **structural AST allowlist** (`@memberjunction/global` now depends on `acorn`): the expression is parsed and every node checked before compilation. Four invariants close the escape — literal-only computed member keys (which kills string assembly at the source rather than chasing spellings), no `constructor`/`__proto__`/`prototype` access, a closed non-computed call surface with no `call`/`apply`/`bind`, and screened free identifiers. The reviewer reproduced the PoC against the compiled evaluator out of this branch and confirmed every variant dead. A structural check also removes the denylist's over-rejection of legitimate data — `name == 'constructor'` and a field named `window` are valid again.

**Round 3 found the allowlist had narrowed too far, and regressed a shipped fix.** `1efc248ac5 — fix(task-graph): the door stops refusing conditions that run correctly` exists precisely because `Number(payload.count) > 3`, `Math.abs(output.delta) < 5`, `isNaN(...)` and `Array.isArray(...)` are the shapes authored specs use and they run correctly; `task-graph-validator.ts` calls `validateSyntax` first, so all four were being refused at the door again. Worse, the pinning test built to prevent exactly that had been reading each name (`Math !== undefined`) rather than calling one, so the two halves shipped contradictory contracts under a green suite.

Now closed:

- **Safe globals are callable again**, via a namespace/bare-callable allowlist in `@memberjunction/global`: `SAFE_GLOBAL_NAMESPACE_METHODS` (`Math.*` pure functions, `JSON.parse/stringify`, `Object.keys/values/entries`, `Array.isArray`, `Number.*`, `Date.now`) and `SAFE_GLOBAL_CALLABLES` (`parseInt`, `parseFloat`, `isNaN`, `isFinite`, `Number`, `String`, `Boolean`). Their union `SAFE_EXPRESSION_GLOBALS` is exactly the twelve names. Receiver and method are both fixed identifiers, so none of the four invariants is weakened — `Math.constructor()`, `Math.abs.call()`, `Math[m]()`, `String.fromCharCode`, `Object.setPrototypeOf`, `setTimeout` and any unlisted bare-identifier call are still refused, each with a pinned test.
- **One list, one owner.** `RESOLVABLE_GLOBALS` is removed from `@memberjunction/ai-core-plus`; `condition-roots.ts` imports `SAFE_EXPRESSION_GLOBALS` directly, so the door and the screen cannot drift by construction rather than by vigilance.
- **The pinning test CALLS every entry** through `validateSyntax` and `evaluate`, with a companion test asserting its table matches the shared list so a name added later cannot land unpinned, plus a sweep over every method the namespace list claims.
- **Optional chaining** (`ChainExpression`) is unwrapped and re-checked, so `payload?.customer?.tier` works while `x?.constructor` and `f?.()` stay blocked.
- **A policy refusal now HOLDS instead of rerouting.** `IsBrokenGuard` recognises the evaluator's refusal message, so a stored graph carrying a construct this build no longer accepts stalls visibly rather than quietly taking a different path — the dispatcher logs a reason only on `hold`. This covers both call sites, including the exclusive-edge path.

**The changeset states plainly what narrowed** rather than claiming the surface is unchanged: `in`/`instanceof`, regex literals, and string/array methods outside `SAFE_METHODS` are now refused where the denylist let them through. All 36 conditions shipped under `metadata/` and `metadata-optional/` were checked and none uses any of it; installations authoring their own expressions should audit those columns before upgrading. The `@memberjunction/global` README, which still described the evaluator as a denylist, is corrected to match.

Suites: `@memberjunction/global` 806, `@memberjunction/ai-core-plus` 673, `@memberjunction/task-graph` 228, `@memberjunction/ai-agents` 1882 — all passing.

## Integration tier

- **Real `Skipped` status** (Track A/A1, the catalog's "#1 harness fix"): gated tiers, absent-MJAPI client bundles, and mutation/live-gated checks now report `Skipped` — surfaced as `X passed / Y failed / Z skipped` — instead of masquerading as `Passed`. Suite pass/fail is decided by hard failures only.
- **MJAPI boots inside `integration.yml`**, so the client-transport deterministic members (~30% of the catalog, never executed in CI before) actually run, authenticated via a per-run `MJ_API_KEY`. Suite metadata resequenced to enforce the servers-before-clients ordering invariant (#3251).
- **Scheduled lanes for every permanently-gated tier**: nightly `RUN_MUTATION_TESTS=1` + cross-server Redis invalidation rig; weekly Live Model suite (gated on `INTEGRATION_LIVE_MODEL_KEYS`, skips loudly until configured; a `run_live_model` opt-in keeps a manual dispatch from spending LLM budget); nightly PostgreSQL functional lane and a nightly CodeGen drift gate (report-first, flip to blocking once green).
- **New bundles**: IT77 – File Storage Provider Seams, IT78/IT79 – materialized entity read, IT80 – Startup Mode (renumbered out of a duplicate IT31), IT81 – Cache Immutability, IT82/IT83 – Queue Engine Lifecycle and Auth Provider Token Validation (self-minted RS256 assertions through the real verification path).
- **Coverage-loss guard spans every bundle** via an auto-derived registry snapshot with a paste-ready failure message. Skip growth is **gated**, not merely reported, through pinned `RequiresMutation`/`RequiresLiveModel` snapshots.
- Catalog / quickstart counts reconciled to the true state: **83 registered bundles / 612 checks**, deterministic suite 68 members (servers 1–48, clients 49–68), live-model 15. The count is pinned by `check-registry.test.ts`.

## Unit tier

- **Three negative-value suites fixed** — tests that asserted re-implementations instead of production code (AIPromptRunner failover, MJQueryEntityServer, MetadataSync ValidationService) now drive the real paths, exposing latent production divergences the mirrors were hiding (documented in-test).
- **Backfill of the highest-risk untested paths**: SQLServerDataProvider data plane + SQLServerTransactionGroup, the BaseAgent execution loop (first tests to drive `Execute()` end-to-end), GraphQLDataProvider wire transport (schema-sync client contracts grown 1→16 InputTypes), MJServer security core (APIKeyScopeAuth, REST handlers, newUsers), MJGlobal utilities, MetadataSync `resolveLookup`/`processFieldValue`, CodeGenLib base-view emission + regeneration golden masters, CRUD CoreActions contracts, and per-action validation contracts for all 107 BizApps provider actions.
- **Shared AI test harness** in `@memberjunction/unit-testing`: a scriptable `TestLLM` extending the real `BaseLLM`, real-typed ChatResult factories, and a realistic catalog fixture — wired into Prompts/Reranker. The provider streaming/ChatResult conformance suite runs against 8 providers and pins real contract deviations under loud `[KNOWN DEVIATION]` markers (e.g. Groq drops streaming usage → cost under-reporting; Ollama failures carry no errorInfo → failover can't classify). It now lives in `@memberjunction/unit-testing`, so the provider specs import a public entry point and `@memberjunction/ai` no longer ships test code or an optional `vitest` peer.
- **Placeholder scaffolds replaced** with real entry-point smokes (`expect(true)` repo-wide count: 0); the anti-pattern lint widened to node specs with its own self-test. Angular Bootstrap/BootstrapLite and ServerBootstrap/-Lite (browser/server manifest packages) gained real manifest smokes.
- **Nightly coverage lane**: the dormant coverage config (broken under vitest 4) fixed and wired as a report-first nightly job with artifact + step summary.

## Release path

`publish.yml` now requires both the fresh-DB migration test and a full unit-suite validation of the release commit before `npm publish` (previously nothing ran between green-on-next and publish). Docker images are boot-verified against a fresh SQL Server (`/healthcheck`) before any registry push, the Dockerfile gained a `HEALTHCHECK`, and `release-test.yml` is a callable validation suite with a secrets-gated nightly Playwright e2e job.

## Review fixes, by round

**Round 1** — missing changesets for `@memberjunction/unit-testing` + `@memberjunction/testing-engine-base`; queue permission drift (the `MJ: Queue Types` × Developer row, so the grants migration no longer diverges from metadata); doc/count drift (`--max-none` 137→134 to match CI, catalog/quickstart counts, registry snapshot pinned); the misleading `docker.yml` head-SHA comment; the live-model opt-in.

**Round 2** — the `SafeExpressionEvaluator` RCE (above); the widened anti-pattern lint false-positiving on env-gated integration tests, resolved by standardising both files on the `describe.skipIf(...)` carve-out that already existed rather than widening the rule, with no growth in the grace allowlist; and a stale `@memberjunction/unit-testing` pin at `6.1.0-edge.1` that resolved from npm instead of the workspace — fixed workspace-wide, which caught two further stale pins in `ServerBootstrap`/`ServerBootstrapLite`. The two round-1 leftovers landed unprompted: the deep `dist/test-support` import is gone, and skip growth is now gated rather than only reported.

**Round 3** — the expression-grammar regression, the disarmed pinning test, and the inaccurate changeset (all above), plus optional chaining and the `IsBrokenGuard` policy-refusal signature.

## Notable findings for triage (documented, not fixed here — out of scope)

1. **`QueueTask` CHECK-vs-default schema defect** → filed as **#3787**: `QueueManager.AddTask` writes `Status='Pending'`, which the `CK_QueueTask_Status` CHECK rejects, so no task can persist. The terminal-status `Save()` fix and the CRUD grants in this PR are inert until that is resolved.
2. **`CheckAPIKeyScopeAndLog` doesn't enforce the application-`IsActive` ceiling; REST include/exclude lists are case-sensitive against lowercased names (fail-open)** → filed as **#3788**. Pinned as `SECURITY DIVERGENCE`/`SECURITY GAP`.
3. **`executeModelWithFailover` vendor-filter index handling can skip surviving candidates; failover attempt-accounting is asymmetric** → filed as **#3789**.
4. **Twitter delete-tweet ownership guard is dead code; several BizApps integrations no-op with `SUCCESS`** → filed as **#3790**. Pinned in-test.

Known non-blocking follow-ups from the reviews, deliberately not taken here: the identifier rule is still a denylist inside an allowlist design (`rows.map(fetch)` passes a host global as a callback — not RCE, but reachable in a browser); the wider method widening (`.split()`, `.replace()`, `.slice()`, `.join()`) has no in-repo contract or usage behind it; `validateSyntax` reports policy refusals as "cannot be parsed", which is misleading for an expression that parses fine; and `evaluate()` re-parses with acorn on every call even when the compiled function is cached.

Secrets to configure when ready to arm the gated lanes: `INTEGRATION_LIVE_MODEL_KEYS`, `E2E_PW_PROFILE_B64` + `E2E_TENANT_ID` + `E2E_WEB_CLIENT_ID`.


AN-BC:
@cadam11 @jordanfanapour @rkihm-BC had `Fable 5/Ultracode` do a deep dive on our test coverage, unit tests and integration. It came up with a bunch of holes. This PR was built also by Fable 5 in `Ultracode` mode and addresses all of these. I'd like you guys to review/adjust/approve this before we merge.

This effort is aligned with how to further harden core MJ as we evolve the platform to be more and more bulletproof.

AN-BC:
@CaelebB-BC can you review this, run it, test it, iterate to get it perfect and then afterwards merge it. This should really raise the bar on our quality repo wide.

AN-BC:
**Status for reviewers** — CI is green on the latest commit (`e785a5bbf`), after two follow-up fixes since the PR opened.

### CI state
| Lane | Result |
|---|---|
| Run unit tests | ✅ success |
| Integration (SQL Server, deterministic) | ✅ success |
| Nightly coverage report | ✅ success |
| Integration (Live Model, weekly) | ✅ gated-skip (no secret / not the weekly cron) |
| Cross-server invalidation, CodeGen drift | ⏭️ skipped (nightly-only) |

The other repo gates (MJ standards, migrations, UI) may still be pending workflow-run approval for this `claude/*` branch — they haven't executed yet on this SHA.

### The maiden integration run did its job — please read this before flagging the skips
Booting MJAPI in `integration.yml` made the **client-transport members actually execute for the first time** (they used to skip-as-pass). That first run surfaced three real problems the old false-green had hidden. Rather than delete coverage or block on them, I parked all three behind an unset `requiresEnv` gate, so they now report a **real, documented `Skipped`** (visible in the summary, still in the catalog) instead of failing:

- **IT72 queue** — the checks correctly caught a genuine product defect: the `QueueTask` CHECK constraint rejects the column's own `'Pending'` default and the `nchar(10)` Status column truncates `'In Progress'`, so `QueueManager.AddTask` can't persist terminal statuses. Needs a schema fix, then un-gate.
- **IT73 auth-validation** — a fixture bug in this new bundle (AV2/3/4's synthesized provider configs don't satisfy the shipped `validateConfig`). Test-side fix, then un-gate.
- **IT52 search** — pre-existing client bundle that had only ever skip-as-passed; its first real run shows the deterministic host lane has no configured lexical search provider. Needs the lane to provision one (or the bundle's own unconfigured-detection fixed).

So `53 passed / 0 failed / 5 skipped` — every non-execution is now honest, and the three parks each carry a documented reason + un-gate condition in their IT metadata.

### Two follow-up commits since open
- `8dbc16b` — park the three bundles above behind `requiresEnv`.
- `e785a5b` — make the nightly coverage lane report-first (CodeGenLib's pre-existing `pg-connection.test.ts` passes under a plain run but breaks under v8 coverage instrumentation; a report-only lane shouldn't be hostage to that).

The product/security defects the new tests pinned (SafeExpressionEvaluator sandbox escape, the `CheckAPIKeyScopeAndLog` IsActive gap, etc.) are listed in the PR description for triage — they're documented in-test, not fixed here.

---

rkihm-BC (commented):
## Code review — Claude Code

> Performed by **Claude Code** (Opus 5) at @rkihm-BC's request, against **`aec1de9`**. I read the full diff (260 files), all five comments, both changesets and the bundle metadata, and re-derived the load-bearing claims locally rather than taking them on trust. **Please push back on anything I got wrong or missed.**

**Verdict: the integration-tier half is excellent and I'd merge it today. One item should block.** The changeset states a sandbox escape is closed; I reproduced that it is not — the same arbitrary-code route is one string-concatenation away, and I proved a real host side effect through an expression the hardened validator *allows*. Recording an unfixed RCE as "closed" is worse than leaving it open, because it retires the item. Beyond that: the PR body is five days stale on its two most consequential claims, and there are three smaller structural concerns.

---

**The integration-tier work is the real prize here.** Booting MJAPI inside `integration.yml` so the 19 client-transport members execute for the first time is the highest-value change in the PR, and it's guarded correctly: if MJAPI doesn't come up in 3 minutes the job `exit 1`s (`integration.yml:238`) rather than letting the lane quietly revert to all-skip. That's exactly the right failure mode and it's actually there.

**Real `Skipped` status is implemented with judgement, not just plumbing.** Gated checks are recorded individually via `recordSkippedCheck`; the all-gated bundle warning now fires on `outcomes.every(o => o.Skipped)` without pushing a synthetic outcome that would double-count. And the code comment names the concrete bug this caught — `task-graph-execution` showing green at 0/0, score 0.0000, while one of its assertions had contradicted the live engine since R6. That's the kind of comment that earns its place.

**Two of the three "parked" bundles from the Aug-7 status comment are now un-gated and passing.** IT78 (queue) and IT79 (auth-validation) both carry `UN-GATED 2026-08-11` in their metadata with the root cause written down; only IT52-search remains parked, correctly, with an explicit un-gate condition. The queue one was root-caused properly rather than worked around.

**The queue fix is right, and so is its reasoning.** `Save()` returns `false` rather than throwing, so discarding it left the row at its prior status while the in-memory task reported `Complete`. `task.Status = persisted && result.success ? Complete : Failed` keeps in-memory state honest, and the comment correctly notes the dispatcher only re-picks `Pending`, so there's no re-run hazard.

**The grants migration is clean.** Pure idempotent `GRANT`s, no `EntityField` INSERTs (so the fresh-install `Sequence` landmine is N/A), correct `v6` folder for a `v6.1.x` filename, and a header that explains precisely why metadata alone is insufficient: the integration lane and any fresh install run `migrate` + `sync push` with **no** CodeGen, so the permission would look right in Explorer and fail at runtime.

**The `vitest.config.ts` fix is a genuinely good catch.** The dormant coverage config had never run (vitest 4 rejects bare `packages/X/*` project globs), and fixing it surfaced two nesting levels — `AI/Vectors/Providers/*` and `Actions/BizApps/*` — that had been silently dropped from coverage entirely.

**Hygiene checks I ran:**
- No `.only`, `describe.skip`, `it.skip` or `xit` smuggled into any changed file — the only hits are inside the anti-pattern linter's own self-test. ✅
- The one deleted test file is the right deletion: `ai-test-harness/services.test.ts` was `expect(true).toBe(true)` plus assertions against a `Map` the test itself constructed. Replacing it with an entry-point smoke that actually imports the Angular package (with the `@angular/compiler` JIT preload) is strictly better. ✅
- Spot-checked the large BizApps suites: `twitter.provider.test.ts` mocks axios at the boundary and asserts the exact endpoint and payload each action sends. Substantive, not tautological. ✅
- Docs consolidation is right — `TESTING_GUIDELINES.md` becomes a router, `UNIT_TESTING_STRATEGY.md` gets a dated historical banner. Matches the repo's own routing philosophy. ✅

---

## Concerning

### 1. (Blocking) The `SafeExpressionEvaluator` escape is not closed — reproduced

The changeset says *"sandbox escape closed"*. The fix widens a regex denylist over raw source text; the evaluator still ends in `new Function(...contextKeys, 'return Boolean(' + expression + ')')` executing in the host realm, with the denylist as the only defence. Whole-word matching falls to splitting the token across a concatenation.

I replicated the shipped `DANGEROUS_PATTERNS` and `validateExpression`/`evaluate` mechanics exactly:

```
--- A. the route the PR closed (correctly BLOCKED) ---
  globalThis["Function"]("return 1")()        -> blocked by \bFunction\b
  globalThis["process"].pid > 0               -> blocked by \bprocess\b

--- B. same capability, token split ---
  []["fil"+"ter"]["const"+"ructor"]("return pro"+"cess.pid")() > 0        -> EXECUTED, value true
  []["fil"+"ter"]["const"+"ructor"]("return typeof requi"+"re")().length  -> EXECUTED, value true

--- C. real host reach, not just a boolean ---
  expression: []["fil"+"ter"]["const"+"ructor"]("gl"+"obalThis.__MJ_PWNED__ = pro"+"cess.pid")() === undefined
  validate  -> ALLOWED
  globalThis.__MJ_PWNED__ = 39549   (actual process.pid: 39549)
```

That is the same RCE class, from a metadata-authored expression, through a validator that says the expression is safe. The new 236-line security suite is thorough on bracket-string access but contains **no concatenation case**, so the gap isn't covered by tests either.

What makes this blocking isn't the residual risk — a denylist was always going to be leaky — it's the bookkeeping. An open RCE gets fixed; an RCE marked "closed" in a changeset does not. Two acceptable exits:
- **Keep the widening, retitle the claim.** "Raises the bar on the denylist; does not close the class" — and keep the triage item open with the concat PoC in the test suite as an `it.skip` + `KNOWN LIMITATION` (which your own anti-pattern linter explicitly permits).
- **Do the real fix.** Parse to an AST and allow-list node types. The documented grammar is small — comparisons, logical operators, member access, array indexing, and a fixed `SAFE_METHODS` list — which is well within reach, makes the denylist unnecessary, and eliminates finding #2 for free.

### 2. The whole-word widening rejects ordinary expressions, including inside string literals

The denylist runs over raw text, so the token doesn't have to be code:

```
"Status === 'Process Order'"   -> REJECTED (\bprocess\b)
"Region === 'Global'"          -> REJECTED (\bglobal\b)
"Doc.Name === 'document'"      -> REJECTED (\bdocument\b)
"payload.windowStart < 10"     -> ok  (correctly not matched)
```

These are metadata-authored expressions living in customer databases — field rules, flow/loop agent conditions, task-graph conditions. The security suite *does* acknowledge over-blocking (`'should (over-)reject comparing against the literal string "constructor"'`), so it's a known trade rather than an oversight. But it ships as a **patch** bump on `@memberjunction/global`, it isn't in the PR body's triage list, and there's no audit query for existing expressions. At minimum I'd want the bump level revisited plus a "run this against your DB before upgrading" query in the changeset.

### 3. The provider conformance harness is imported through another package's `dist/`

All eight conformance specs do:

```ts
import { RunLLMConformanceSuite } from '@memberjunction/ai/dist/test-support/llm-conformance.js';
```

That reaches past the package's public API into its build output. It resolves *only* because `@memberjunction/ai` has no `exports` map today — the day anyone adds one (the modern default; `server-bootstrap` already has one), all eight break with `ERR_PACKAGE_PATH_NOT_EXPORTED`. It also makes those specs depend on `@memberjunction/ai` being built, so `cd packages/AI/Providers/Groq && pnpm test` — the exact command CLAUDE.md's Definition of Done prescribes — needs a prior build. (Through turbo it's fine: `test → build → ^build`.) And `vitest` is now an optional peerDependency of a core runtime package, with `dist/test-support/*` shipping to consumers.

The fix is already half-built in this PR: `@memberjunction/unit-testing` gained a dependency on `@memberjunction/ai` and is hosting the `TestLLM` harness. Move the conformance suite there too — or add a `./test-support` export subpath so the deep path becomes a supported contract instead of an accident.

### 4. Skip growth is reported but not gated

`integration.yml`'s summary step prints the pass/fail/skip accounting so "a run where coverage silently shrank is one click away from being noticed" — that's human vigilance, not a gate. The bundle-registry snapshot guards bundles *disappearing*; nothing guards a bundle or check *starting to skip*. Given this PR's entire thesis is that non-execution must never be silent, the finishing move is the same trick one level down: snapshot the expected skip set and fail when it grows. As it stands, a future change that makes 20 checks self-skip leaves CI green with the step summary as the only tell.

---

## Missing / smaller

- **The PR body is five days stale on its two most consequential claims.** Triage item #1 says the SafeExpressionEvaluator escape is *"documented in tests, not fixed — out of scope"*, while a changeset in this same PR says it's fixed. Triage item #4 says the `QueueTask` CHECK/default defect blocks `QueueManager.AddTask` — but the real root cause turned out to be the missing `CanUpdate` grant, that's fixed, IT78 is un-gated, and the schema quirks are now "documented-and-worked-around, not asserted" per the bundle metadata. The Aug-7 status comment's "three parked bundles" is likewise down to one. For a 260-file PR where reviewers are asked to triage from the description, please refresh it before merge.
- **The last green-CI evidence is for `e785a5b` (Aug 7).** The branch has merged `next` several times since; CI on `aec1de9` is still in flight as I write (migrations, ngc-scss, peer-deps, Mongo/Postgres drivers green; unit, integration, standards, dependency-check pending). Worth a fresh full green.
- **~49 of the changed specs are two-assertion entry-point smokes** (~100 of the "~1,500 tests"). They're honest work — export-graph breakage is exactly what bit the browser-manifest packages — but the headline number reads heavier than the median new test.
- **The queue's underlying schema defects are still live**: `CK_QueueTask_Status` rejecting the column's own `'Pending'` default, and `nchar(10)` truncating `'In Progress'`. Now worked around rather than fixed, and deliberately not asserted so a later fix can't turn IT78 red — which is the right call for the bundle, but it means nothing will ever remind anyone. Worth a tracking issue so the workaround doesn't become permanent.

---

I did not run the unit or integration suites locally (CI covers them, and the integration tier needs a DB); everything else above I checked against the branch at `aec1de9`. My working tree was untouched. If any finding is wrong or already handled somewhere I didn't look, say so on the thread and I'll re-check rather than assume I was right.

rkihm-BC:
## Code review (round 2) — Claude Code

> Performed by **Claude Code** (Opus 5) at @rkihm-BC's request, against **`935156cdd0`**. I re-read the diff and the full comment history, then re-derived the load-bearing claims rather than taking them on trust: I compiled the new `SafeExpressionEvaluator` out of this branch and attacked it directly, pulled the published `unit-testing` tarball off npm, and read the CI runs that finished during the review. **Please push back on anything I got wrong.**

**Verdict: the security fix is real and I verified it. Three things block merge** — two are small and mechanical, the third is a design call. CI is currently red.

---

## The RCE is genuinely closed — reproduced against the compiled evaluator

The AST allowlist is a proper structural fix, not a wider denylist. My previous PoC and every variant I could construct are dead:

```
blocked   [][ "cons"+"tructor" ][ "cons"+"tructor" ]("return process.pid")()
blocked   [][ "fil"+"ter" ][ "const"+"ructor" ]("globalThis.__MJ_PWNED__ = process.pid")()
blocked   globalThis["Function"]("return 1")()
blocked   [].filter.constructor("return process.pid")()
blocked   (()=>{return 1})()          blocked   this.constructor
blocked   a[b] == 1                   ← non-literal computed key, the concat route's root
globalThis.__MJ_PWNED__ = undefined   (process.pid was 60542)
```

Rejecting computed member keys that aren't literals is the right hinge — it kills string assembly at the source rather than chasing spellings. `call`/`apply`/`bind` are correctly absent from `SAFE_METHODS`, so I found no remaining route to the `Function` constructor. `@memberjunction/global` is **781/781 green**. The over-rejection I flagged last round (`Status === 'Process Order'`) is gone for free, as claimed.

Also verified fixed since round 1: the queue permission drift (all six `Queue Tasks`/`Queues`/`Queue Types` × Developer/Integration rows present and matching the migration's grants exactly — and the migration itself is clean: idempotent `GRANT`s, correct `v6` folder, no `EntityField`/`Sequence` landmine), `--max-none` 134 consistent across CI and both guides, bundle count pinned at 81 by test, and the two missing changesets.

---

## 1. (Blocking) The widened anti-pattern lint false-positives on env-gated integration tests — CI is red on it now

On `935156cdd0` the Unit Tests job dies at **step 9, Spec anti-pattern lint**, and **step 14 "Build + run unit tests" is `skipped`** — the unit tier never runs at all on the current head.

```
❌ 2 spec anti-pattern(s) found:
   packages/CodeGenLib/src/__tests__/integration/pg-quote-sql-execution.integration.test.ts:28
   packages/PostgreSQLDataProvider/src/__tests__/integration/pg-autoquote-identifiers.integration.test.ts:28
      — disabled test — document it with an adjacent `// KNOWN LIMITATION` comment or fix it
```

Line 28 of both is the standard env-gate idiom, arriving from `next` via #3697:

```ts
const describeIfPG = PG_URL ? describe : describe.skip;
```

This PR widens the lint from `*.dom.test.ts` to **every** `*.test.ts`, and `/\b(?:it|describe|test)\.skip\b/` matches it. The rule's own header already carves out the sibling case — *"`it.skipIf(cond)` is conditional execution, not a disabled test — it never matches"* — but the ternary form gets no equivalent carve-out, despite being the same intent. Arguably the same intent this PR champions elsewhere with real `Skipped` status.

The scale is the part that concerns me. Seven files use this idiom; **five are hard-coded into the grace allowlist** (`scripts/check-spec-antipatterns.mjs:55-59`), and the two that merged in later aren't. So the gate is held together by a hand-maintained path list that goes stale on every merge from `next`, and every future PR adding an env-gated integration test will hit it.

**Change needed:** teach the rule the conditional form — treat `<cond> ? describe : describe.skip` (and the `it`/`test` variants) as conditional execution, exactly as `skipIf` already is. Bolting `// KNOWN LIMITATION` onto correct code would be the wrong fix; those two tests aren't deferred, they're env-gated. Worth removing the five now-redundant allowlist entries in the same change.

## 2. (Blocking) `@memberjunction/unit-testing` is pinned to a version that resolves from npm, not the workspace

`packages/AI/Prompts/package.json:43` declares `"@memberjunction/unit-testing": "6.1.0-edge.1"` while the workspace is at `6.1.0-edge.2` — every sibling `@memberjunction/*` dep in that same file is `edge.2`. Because `6.1.0-edge.1` **exists on npm**, pnpm resolved it from the registry instead of linking the workspace, with no warning:

```yaml
'@memberjunction/unit-testing':
  specifier: 6.1.0-edge.1
  version: 6.1.0-edge.1      # ← not link:../../UnitTesting
```

I pulled that tarball to confirm it's the cause rather than infer it — the published `edge.1` predates the harness this PR adds:

```
dist/: custom-matchers  index  mock-entity  mock-run-view  singleton-reset     ← no ai/
grep -c "TestLLM|buildRealisticCatalog" dist/index.js dist/index.d.ts  →  0, 0
```

Which produces exactly the failure seen on the previous head `f35e267`, where the lint still passed and the tests did run — 4 files / 32 tests, all in `@memberjunction/ai-prompts` (the other two `ELIFECYCLE` lines are turbo propagation; there's a single failing `command (…/packages/AI/Prompts)`):

```
TypeError: __vi_import_4__.TestLLM is not a constructor              (execute-e2e, failover, timeout)
TypeError: __vi_import_3__.buildRealisticCatalog is not a function   (model-selection, ×32)
```

The pin is unchanged at the current head, so this is still queued up behind item 1.

**Change needed:** `6.1.0-edge.1` → `6.1.0-edge.2` and re-run `pnpm install` at the root.

Two things worth noting beyond the one-line fix. It also breaks `AIPromptRunner.model-selection.test.ts`, which this PR never touches — it reaches the harness through the new compat shim in `__fixtures__/ai-metadata.fixtures.ts` — so the blast radius is wider than the diff suggests. And **"Detect missing and unused dependencies" passed**, because the dependency *is* declared; nothing checks that a `@memberjunction/*` dep actually resolves to the workspace rather than to a stale published copy. Given the repo's move to pnpm's strict resolution, that seems worth a gate of its own.

## 3. (Blocking, design call) The expression grammar narrowed sharply, the changeset says it didn't, and one consequence is silent

The changeset states *"The supported surface … is unchanged."* It isn't. Running realistic expressions through both evaluators, **19 of 23 that `next` accepts are now refused**:

| Now rejected | Examples |
|---|---|
| Optional chaining | `payload?.customer?.tier == 'premium'` — `ChainExpression` isn't in the allowlist at all |
| `in` / `instanceof` | `'tier' in payload.customer` |
| String methods off `SAFE_METHODS` | `.split()` `.replace()` `.slice()` `.substring()` `.match()` `.join()` |
| Every host-global call | `Math.abs()` `parseInt()` `Number()` `String()` `JSON.stringify()` `Object.keys()` `Array.isArray()` `d.getFullYear()` |
| Regex literals | `/prem/.test(x)` |

(Correcting myself from an earlier draft: template literals are *not* a regression — `next`'s denylist already banned backticks and `${`.)

These are metadata-authored expressions living in customer databases — field rules, flow/loop agent conditions, task-graph conditions.

**It also contradicts an existing in-repo contract, and nothing catches it.** `condition-roots.ts:50` defines `RESOLVABLE_GLOBALS = {Math, Number, String, Boolean, Array, Object, JSON, Date, parseInt, parseFloat, isNaN, isFinite}` specifically so the task-graph validator can bless conditions like `Math.abs(output.delta) < 5`. `task-graph-validator.ts:73` calls `validateSyntax` **first**, so all nine of that file's own documented examples are now refused at the door:

```
INVALID REFUSED  isNaN(payload.count)        INVALID REFUSED  Math.abs(output.delta) < 5
INVALID REFUSED  parseInt(payload.raw) > 0   INVALID REFUSED  Object.keys(payload).length > 0
INVALID REFUSED  Array.isArray(output)       INVALID REFUSED  'count' in payload    … etc.
```

The test meant to pin the two together (`condition-syntax-validation.test.ts:105`, *"every allowed global is one the evaluator actually resolves"*) only asserts `Math !== undefined` — a bare identifier read, never a call. `ai-core-plus` reported **660/660 passing** in the same CI run. The contradiction ships invisibly.

**And for graphs already in the database, the failure is silent.** `DecideGate` classifies a failed evaluation with `IsBrokenGuard()`, which matches only `is not defined|reference error|can't find variable`. `"Expression contains a forbidden construct: …"` doesn't match, so the verdict is **`drop`** — and `TaskGraphDispatcher.ts:3046` logs only on `hold`, with `reason` left `undefined` for a drop. The exclusive-edge path (~3099) likewise returns `'unsatisfied'`. So an upgraded install doesn't error; it quietly reroutes graphs, feeding `unreachableTaskIDs` and the skip cascade, with no recorded cause. (Flow agents fare better — `SafeConditionEvaluator` surfaces a visible `ConditionError`.) Relatedly, the `DATA_ABSENCE_SIGNATURES` entries for `'x' in payload` and `Object.keys(...)` are now dead code, since those expressions never reach evaluation.

### Recommendation

**Widen the allowlist, unify the globals list, and make refusal loud.** The narrowing isn't buying security: `parseInt(x) > 0` and `payload?.customer?.tier` are not steps toward the `Function` constructor. The escape was closed by four structural invariants — literal-only computed keys, no `constructor`/`__proto__`/`prototype`, a closed non-computed call surface with no `call`/`apply`/`bind`, and screened free identifiers — and none of them depends on the grammar being small. I verified each empirically; everything below leaves all four untouched.

**Tier 1 — near-free, do regardless:**
- **`ChainExpression`** (optional chaining). ~3 lines: unwrap and recurse, so `a?.b` and `a?.()` land back in the existing member/call rules and `a?.constructor` stays blocked. Highest-value fix in the list — optional chaining is *the* idiom for null-safe conditions, and null-safety is the entire premise of `BuildConditionContext`'s envelope. This reads like an oversight rather than a decision.
- **`in`** as a binary operator. No exec risk, and `condition-gate.ts` already carries a `DATA_ABSENCE_SIGNATURES` entry for it — proof it's in use in the field.

**Tier 2 — the bulk of the compatibility:**
- Pure value methods into `SAFE_METHODS`: `split, join, slice, substring, charAt, padStart, padEnd, concat, at, flat, repeat, toFixed, replace, replaceAll`.
- A **namespace call allowlist** (receiver and method both fixed): `Math.*` pure functions, `JSON.parse/stringify`, `Object.keys/values/entries`, `Array.isArray`, `Number.isInteger/isFinite`, `Date.now`.
- A **bare-callable allowlist**: `parseInt, parseFloat, isNaN, isFinite, Number, String, Boolean` — i.e. exactly `RESOLVABLE_GLOBALS`.
- Date instance getters, since `cloneValue` deliberately preserves `Date` objects into the context.

**Keep excluded, deliberately:** regex literals (the ReDoS rationale in the code is sound, and `.match('str')` covers most real use), `call`/`apply`/`bind`, `Object.assign`/`defineProperty`/`setPrototypeOf`, `instanceof`, `new`.

**Two structural moves that matter more than the list itself:**
- **Move the globals list into `@memberjunction/global` and have `condition-roots.ts` import it.** One list, owned by the layer that enforces it, makes the drift impossible instead of merely detectable — and fix the pinning test to actually *call* each entry, which is why 660/660 stayed green while the two halves disagreed.
- **Add a policy-refusal signature to `IsBrokenGuard()`.** One line, and the piece I'd least want skipped: it converts every residual incompatibility — including ones a widened list still misses — from a silent reroute into a visible hold. It's also what `condition-gate.ts`'s own doctrine already argues for: *"an unrecognised failure is then visible and recoverable rather than a silently dropped branch."* Today a policy refusal takes the opposite path from that comment.

One nice convergence: introducing a bare-callable allowlist moves the identifier rule from "not on a list of 25 bad names" to "must be a context key or a known-safe global," which closes the residual hole below as a side effect.

**Sequencing:** I'd do all of it here — roughly 60–100 lines across two files plus tests, not a rewrite — because splitting means landing a known-silent-breakage state on `next`, which is what every other branch cuts from. If you do split, the non-negotiable minimum is the `IsBrokenGuard` signature plus `ChainExpression`; with those two the worst case degrades from "graphs quietly reroute" to "graphs stall loudly during upgrade testing," which is shippable with a fast-follow.

Either way: minor bump (the branch carries a migration, so it's available), a changeset that says plainly what narrowed, and an audit query over the columns storing expressions so operators can pre-flight. The current *"supported surface is unchanged"* needs to go regardless.

---

## Non-blocking follow-ups

- **Identifiers are still a denylist inside an allowlist design.** Any callable global not among the 25 names can be invoked by passing it as a callback to a safe array method — verified, the host global really is called: `payload.rows.map(fetch)` validates and invokes `fetch` with `(value, index, array)`. In Node most die on the extra args; in a browser (this ships to Explorer, and `task-graph-validator.ts` explicitly anticipates browser execution) `open`/`alert`/`print`/`postMessage` would be reachable. Not RCE — but the fix falls out of the Tier-2 work above.
- **Skip growth is still reported, not gated** (`integration.yml:269`). Round 1's finding #4 is unaddressed; for a PR whose thesis is that non-execution must never be silent, this is the last mile.
- **Round 1's finding #3 is acknowledged rather than fixed** — the eight provider specs still import `@memberjunction/ai/dist/test-support/…` and `@memberjunction/ai` has no `exports` map. A comment now explains the trade-off, which is a defensible call; flagging only that it's a documented choice, not a resolution.
- **Smaller:** `validateSyntax` policy refusals surface through the validator as *"cannot be parsed"*, misleading for `Math.abs(...)`, which parses fine. Validation is no longer cached — `evaluate()` re-parses with acorn on every call even when the compiled function is cached (1.6µs → 3.2µs measured; trivial in absolute terms, trivial to fix by caching the verdict alongside the function). And `acorn` (228 KB unminified) is now a runtime dependency of a package that ships to every browser bundle.

---

I didn't run the suites locally — the version pin means a local run would just reproduce CI — and I didn't touch any working tree; everything above came from the branch via `git show`, a compiled copy of the evaluator in a scratch dir, and the CI logs. If any finding is wrong or already handled somewhere I didn't look, say so and I'll re-check rather than assume I was right.

rkihm-BC:
## Round 3 — `#1` and `#2` are fixed. `#3` is untouched, and I'm **narrowing** what I'm asking for.

> Claude Code (Opus 5) at @rkihm-BC's request, against **`a45d51cfd0`**. I went looking for evidence that would let me drop item 3 rather than evidence that would keep it. Some of it I found — so the required set below is smaller than what I posted last time. What survived got stronger, not weaker.

### Fixed — and two of them better than I asked

**#1, the lint false positive.** Resolved in `5503d8f35b`, and resolved better than I proposed. Rather than teach the rule the ternary form, both files moved to the carve-out that already existed:

```ts
// Env-gated: runs only against a live PostgreSQL (MJ_TEST_PG_URL); skipIf keeps ordinary
// `pnpm test` database-free without reading as a disabled test.
const describeIfPG = describe.skipIf(!PG_URL);
```

Standardising on one idiom beats widening the rule, and the grace allowlist did **not** grow to paper over it — 25 entries before and after. (Five older siblings still use the ternary form and stay grandfathered; converting them is hygiene, not a blocker.)

**#2, the version pin.** Fixed in `c3ae0b2251`, and broader than I flagged — it caught two more stale `edge.1` pins I'd missed in `ServerBootstrap` and `ServerBootstrapLite`. The lockfile now reads `version: link:../../UnitTesting`.

**Both round-1 leftovers landed too**, unprompted. The deep `dist/test-support` import is gone — the conformance suite moved into `@memberjunction/unit-testing`, the eight provider specs import the public entry point, and `@memberjunction/ai` no longer ships test code or the optional `vitest` peer. And skip-growth is now genuinely **gated** via pinned `RequiresMutation`/`RequiresLiveModel` snapshots with a paste-ready diff, rather than only shrinking a step summary. (Scope note for later: static per-check flags only, so a bundle parked behind an unset `requiresEnv` — the IT52-search style park — still isn't guarded.)

Also verified: the queue migration resequenced to `V202608150000`, correctly after `next`'s `V202608140000`, rename-only with all 12 grants intact.

---

### Correcting myself — two things that let me shrink the ask

I said last round that the grammar narrowing was broadly a required fix. Checking rather than assuming:

- **No shipped metadata is affected.** I searched `metadata/` and `metadata-optional/` for conditions using any of the now-rejected forms and found **zero**. My "customer databases" framing was about installations, not about anything this repo ships.
- **No in-repo usage of optional chaining in conditions**, and the evaluator's own JSDoc never promised `.split()`, `.replace()`, `Math.*` or regex literals — `SAFE_METHODS` existed on `next` but went unenforced. So there's a fair argument that the implementation now simply matches a long-documented contract.

That argument holds for most of the surface. **It does not hold for the twelve globals**, and that's the part I'm still asking you to fix.

---

### 3. (Required) This PR regresses commit `1efc248ac5`, and silently defeats the test that was built to prevent exactly that

`1efc248ac5 — fix(task-graph): the door stops refusing conditions that run correctly (F1, F2)` is a shipped MJ fix whose commit message states:

> "The door was refusing `Number(payload.count) > 3`, `Math.abs(output.delta) < 5`, `isNaN(payload.count)` and `Array.isArray(output)` — **specs that ran before R2-3 and would still run correctly after it.** 'It may miss a typo, it will never invent one' was exactly the promise being broken."

`condition-roots.ts` says the same in prose: *"`Number(payload.count) > 3` and `Math.abs(output.delta) < 5` are **exactly the shapes authored specs use**, and they work."*

Those four expressions are now refused again. `task-graph-validator.ts:73` calls `validateSyntax` first, and I ran all four plus the rest of `RESOLVABLE_GLOBALS`' documented examples through the evaluator at this head (`sha e977ddfc…`, byte-identical to what I attacked in round 2):

```
INVALID REFUSED  Number(payload.count) > 3      INVALID REFUSED  Math.abs(output.delta) < 5
INVALID REFUSED  isNaN(payload.count)           INVALID REFUSED  Array.isArray(output)
INVALID REFUSED  parseInt(payload.raw) > 0      INVALID REFUSED  Object.keys(payload).length > 0
```

So the door is back to refusing the specs that fix existed to protect. That's a regression of a deliberate, documented decision — not a hypothetical about installations, and not a matter of taste.

**The guard that should have caught it has been quietly disarmed.** That same commit added the pinning test for this precise purpose:

> "It evaluates every allowlisted name through the real evaluator and asserts it resolves, **so the list cannot claim something the runtime would reject**"

`condition-syntax-validation.test.ts:105` asserts `Math !== undefined` — an identifier *read*, never a *call*. The list now claims exactly what the runtime rejects, and `ai-core-plus` reported **660/660 green** while doing it. The invariant is stated in the code and is currently false.

**What's required, minimally:**

1. **Restore call support for the twelve `RESOLVABLE_GLOBALS`** — a namespace/bare-callable allowlist (`Math.*`, `JSON.parse/stringify`, `Object.keys/values/entries`, `Array.isArray`, `Number.*`, `parseInt`, `parseFloat`, `isNaN`, `isFinite`, `Date.now`). Structurally free: receiver and method are both fixed names, so none of the four invariants that close the escape (literal-only computed keys, no `constructor`/`__proto__`/`prototype`, closed non-computed call surface with no `call`/`apply`/`bind`, screened identifiers) is touched. Ideally the list moves into `@memberjunction/global` and `condition-roots.ts` imports it, so the two halves can't drift again by construction.
   *Or* consciously revert `1efc248ac5` and say so in the changeset — but doing neither ships two contradictory contracts.
2. **Make the pinning test call each entry**, so it enforces the invariant it documents.
3. **Correct the changeset.** *"The supported surface … is unchanged"* is false — I have 19 counterexamples, and the four above are the ones that matter. This is the item I'd least want to see merge as-is: an inaccurate changeset is what retires the finding.

### Recommended, explicitly **not** required

Flagging these as judgement calls so they don't get read as blockers — I found no in-repo usage or contract behind any of them:

- **`ChainExpression` / optional chaining** (~3 lines: unwrap and recurse). Nothing in the repo uses `?.` in a condition and the JSDoc never promised it, so this is a call about what you want authors to be able to write. I'd still take it — it's the common null-safe idiom and null-safety is the premise of `BuildConditionContext`'s envelope — but it isn't a defect.
- **The rest of the method widening** (`.split()`, `.replace()`, `.slice()`, `.join()`, regex literals). No contract, no usage found. Genuinely optional.
- **A policy-refusal signature in `IsBrokenGuard()`** (one line). Defensive: a refusal currently routes to `drop`, and `TaskGraphDispatcher.ts:3046` logs only on `hold`, so a refused condition reroutes a stored graph with no recorded cause. Before this PR that path was near-unreachable because the denylist refused almost nothing; it's newly reachable now. With item 1 done it becomes rare again — so it's cheap insurance rather than a fix.
- The identifier denylist (`rows.map(fetch)`), the *"cannot be parsed"* wording on policy refusals, and re-parsing with acorn on every cached evaluation.

---

**CI at this head:** everything green except **Run unit tests, still in progress** (~45 min, longer than the previous ~30 min run — the `next` merge added a lot). Integration (SQL Server, deterministic) is green, as are migrations, standards, dependency-check and the driver lanes. Worth confirming that one lands green before merge.

Net: items 1–3 above are what I'd hold for, and item 3 is a documentation fix. Everything else in this PR is in good shape, and the response to the last round was thorough. If you disagree that `1efc248ac5` is being regressed — or if reverting it is the intended call — say so and I'll re-check rather than assume I'm right.

rkihm-BC:
**CI update on `a45d51cfd0`** — closing the one open question from my comment above: **`Run unit tests` finished green** (16:45 UTC, ~57 min — noticeably slower than the prior ~30 min run, presumably the `next` merge). All 14 non-skipped checks now pass, the 6 skips are the expected nightly/weekly/manual lanes, and the PR reads `mergeState: CLEAN`.

So **CI is no longer a reason to hold.** The three required items in my previous comment are unaffected — they rest on the code, not the pipeline. Worth noting that the green unit run is what that finding *predicts* rather than evidence against it: `condition-syntax-validation.test.ts:105` passes precisely because it asserts `Math !== undefined` (an identifier read) and never exercises the call form.

PR #3552, opened 2026-08-07
fix(ci): gate on non-concrete dep ranges; delete dead tabstrip generator

MattC-BC:
Follow-up to #3467. Items 1 and 3 came out of @MarceloT-BC's review there, adjudicated by an independent audit. Item 2 clears a live phantom-dependency drift on `next` that item 1 exposed. Item 4 unblocked a CI ratchet that had gone red on `next`.

## 1. The deps guard could write `latest`, and could never take it back

`.github/scripts/check-explorer-external-deps.mjs` had a one-way ratchet:

- `pickRange()` returns `latest` when *every* owner package declares it, and `writeDeclarations()` wrote that straight into MJExplorer's `dependencies`. That is how `latest` specifiers for `@codemirror/commands`, `@codemirror/language` and `@lezer/highlight` entered that manifest in `8d6160645b`.
- Once there, nothing could remove it: `parseRange('latest')` yields `kind: 'any'`, and `rangesIntersect` returns `true` unconditionally for `any` — so the declaration never registered as `incompatible`, and `--write` never repaired it. The only signal was a warning. They persisted until Craig pinned the owner package in `ea4000036`.

Both halves are closed:

- **`--write` refuses to persist a non-concrete range**, logging which owner package to pin instead. Skipping leaves the dep `missing`, so the check fails loudly and points at where the fix belongs rather than laundering the problem into the app manifest.
- **`findNonConcreteDeclarations()`** (new, exported) detects a `latest` / `*` / unparseable range *already* declared — the case the incompatible check structurally cannot see — and fails the check with a reproducibility message.

Verified against the historical bug: reintroducing `"@codemirror/commands": "latest"` makes check mode exit 1 naming the dep; `--write` refuses to touch it; restored, the tree exits 0.

**Tests:** four new cases pin both halves — including the precondition that `latest` *does* intersect its owner range, which is the trap itself. Guard file **20/20** (was 16); full `.github/scripts` suite **371/371** across 14 files. These run in CI via `test.yml` (`vitest run --config .github/scripts/vitest.config.mts`).

### ⚠️ Scope correction: this script is not wired into CI

An earlier revision of this description claimed the change makes "the gate fail loudly." **That is not accurate, and it is worth stating plainly rather than quietly fixing.** `check:explorer-deps` exists only as a `package.json` script — no workflow, composite action, or aggregate runner invokes it (`check:standards` is `mj standards check`, and `.mj-standards.json` declares only `ui-layers`). The existing "Detect missing and unused dependencies" job is Knip, which compares declared-vs-imported *per package* and structurally cannot catch this class of bug — the whole point is that MJExplorer must declare deps it never imports itself.

So this PR hardens the checker's logic; it does not yet make anything gate. The `--write` path *is* used by hand and by agents, and that half takes effect immediately.

Evidence that the missing wiring matters, and that it has already cost something: the checker **exits 1 on `next` today** with

```
✖ missing: acorn (owners: @memberjunction/global)
```

`@memberjunction/global` imports `acorn`, and because Angular's dev server excludes `@memberjunction/*` from prebundling, that bare import resolves from MJExplorer's context rather than MJGlobal's. npm's hoisted layout hides it; under pnpm or the generated Open App workspace it is a phantom that fails at `ng serve` time — precisely the bug class this guard exists to catch. It drifted in unnoticed because nothing runs the guard.

## 2. `acorn` declared, clearing that drift

`667eb9cf2a` adds `"acorn": "^8.15.0"` to MJExplorer's `dependencies` — the range the guard's `pickRange` picks from the owner — plus the matching `packages/MJExplorer` importer entry in `pnpm-lock.yaml`.

The lockfile change is the importer entry *only*: `acorn@8.15.0` already had package and snapshot entries as a transitive dep, so nothing else needed resolving. It was hand-inserted rather than lifted from a full `pnpm install --lockfile-only`, which dragged in ~60 lines of unrelated peer-suffix churn on `googleapis` and `webpack-dev-server`. The manifest edit alone would have been wrong: without the lockfile entry, `pnpm install --frozen-lockfile` fails, which is the first step of the "Detect missing and unused dependencies" job.

After this, the checker reports `explorer-deps: OK — externalized closure fully declared` (exit 0), and `pnpm install --frozen-lockfile` exits 0. `mj_explorer` is `private: true`, so no changeset.

Wiring the guard into a workflow remains a follow-up; this clears the drift so that follow-up lands green instead of immediately red.

## 3. `innerTabStripHTML` deleted

Deprecated in favour of `innerCollapsiblePanelsHTML`, **zero call sites** since `6f88c60dee`, and no config option reaches it. It was also internally incoherent: every `TabCode` producer emits `<mj-collapsible-panel>` while `MJTabStripComponent` finds children via `@ContentChildren(MJTabComponent)`, so calling it would render an empty strip.

Left in place it was a live trap — `AngularClientGeneratorBase` is instantiated through `MJGlobal.Instance.ClassFactory.CreateInstance` and its README advertises subclassing, so a downstream subclass could reach it, and the markup it emitted now sizes to content since #3467 stopped `ng-tabstrip` hardcoding viewport height. **No generated form has contained `<mj-tabstrip>` since that switch**, so no generated output changes.

Note for downstream consumers: removing a `protected` member from a documented-extensible base class is source-breaking for an out-of-repo subclass that *calls* it. That is the intended trade — a compile error beats silently rendering an empty strip. Bump stays `patch` per `.claude/rules/changesets.md` (`minor` is reserved for migration/metadata branches).

## 4. `workspace-tab-strip` DOM spec expanded

This landed in commit `737016f2d3` and was missing from earlier revisions of this description.

CI's DOM coverage ratchet went red on `next` when #3467 added `MJWorkspaceTabStripComponent` and `MJWorkspaceCardComponent` without a DOM spec between them. Because the ratchet is absolute, that failed *every* subsequent PR, not just the one that caused it. It was fixed by writing the spec rather than bumping the number.

Since then `next` has picked up a 9-case spec of its own and the ratchet now sits at `--max-none=134`, so this is **no longer a CI unblock** — it is a net expansion from 9 to 13 cases. It pins the strip's own contract and deliberately does not re-test the shared layers (the global `.mj-tabs*` chrome, and `mjTabList`'s keyboard behaviour, which has its own suite):

- renders what it is handed, in order; the empty-strip case keeps the new-tab affordance so an emptied strip is not a dead end;
- emits the tab **ID** rather than its index — emitting a position would silently break a host that reorders;
- closing emits `TabClosed` and does **not** also select: the close button is nested inside the clickable tab, so a missing `stopPropagation` would do both;
- unsaved/rejected state reaches the accessible name, since the dirty dot is `aria-hidden` and a status colour cannot be read aloud;
- the close button stays out of the page tab order (the tab is the single focus stop), and lifecycle state surfaces as modifier classes.

Test-only, so no changeset. `MJWorkspaceCardComponent` remains uncovered — a thin slotted frame, noted as the next gap.

## Verification

- `.github/scripts` suite **371/371** (14 files), including the guard's **20/20**.
- CodeGenLib typechecks clean; the deletion leaves no orphaned imports (`GeneratedFormSectionType` and `RelatedEntityDisplayLocation` are both still used elsewhere in the file). Its 10 pre-existing Postgres-pool test failures are **identical before and after** (verified by stashing this change and re-running) — untouched by this work.
- `check:standards` passes.
- `check:explorer-deps` **exits 0** on this branch (`OK — externalized closure fully declared`); it still exits 1 on `next` until this merges. `pnpm install --frozen-lockfile` exits 0.

## Known gaps / follow-ups

Called out rather than fixed here, so they are not lost:

1. **Wire `check:explorer-deps` into a workflow** (with `paths: packages/*/package.json`). This is the one change that makes the premise of item 1 operative. Item 2 clears the only outstanding violation, so this now lands green rather than immediately red — which is why it was worth doing here.
2. **Split `any` from `unknown` in the new check.** `parseRange` understands only exact / `^` / `~` / `latest` / `*`; everything else is `unknown`, and both new code paths treat `unknown` exactly like `latest`. That is wrong for shapes this repo ships — `type-graphql@2.0.0-beta.3` is an *exact pin* yet parses `unknown`, as do `@azure-rest/ai-inference@^1.0.0-beta.6` and `playwright@>=1.40.0`. None are in MJExplorer's 93-dep closure today, so this is latent, not live; but it would fail with a factually wrong "not reproducible" message, and `--write` would refuse to write a perfectly good range. Gate on `any`, keep `unknown` as a warning (its original fail-open intent), or teach `parseRange` prereleases and comparator ranges.
3. **`findNonConcreteDeclarations` reads only `dependencies`**, while `evaluateDeclarations` accepts a declaration from `optionalDependencies`, `devDependencies`, or `dependencies` — so a `latest` in the app's `devDependencies` is the same ratchet, one field over. Zero instances today; a one-line fix to reuse the merged map.
4. **`--write` exits 0 while knowingly leaving the check red** when it skips a non-concrete range. Deliberate and test-pinned, but a scripted "regenerate and commit" then produces a red tree with a green exit code.
5. Minor: the local `const resolve = (dep) => …` in `writeDeclarations` shadows `resolve` from `node:path`; `--list` mode does not surface non-concrete declarations; `--write` can print "already in sync" immediately followed by a non-concrete warning.

## Not included

Two follow-ups from #3467 remain deliberately out of scope: renaming the workspace tab components off the "workspace" name, and hardening `mjTip` (scroll dismissal + its gate-invisible shadow).

cc @MarceloT-BC @rkihm-BC


MattC-BC:
@rkihm-BC - this one should be ready to be merged. I had Claude to an adversarial pass and everything checked out.

Issue #3546, opened 2026-08-06
spDeleteEntityWithCoreDependencies covers ~18 of ~73 FK references to Entity — entity removal half-fails silently


AN-BC:
## What

`spDeleteEntityWithCoreDependencies` is what CodeGen calls (via `checkAndRemoveMetadataForDeletedTables`, `packages/CodeGenLib/src/Database/manage-metadata.ts`) to prune an entity's metadata once its base table disappears. It cascades **~18** tables. The database currently has **~73** FK references to `Entity.ID`.

Anything not on the proc's list blocks the final `DELETE FROM Entity`, and the failure is caught and logged rather than raised:

```ts
catch (ex) {
   logError(`Error removing metadata for entity ${(ex as any).Name}, error: ${ex}`);
}
```

## Why it matters

The entity is left **half-pruned** — the `Entity` row survives, but its `EntityField` rows are already gone. Every subsequent CodeGen run then emits:

```
SKIPPING SQL GENERATION: Entity <name> has no primary key field in metadata.
```

...and that entity is silently excluded from generation from then on. The only signal is one grey log line in a long CodeGen run.

## How it was found

Phase 0 of the unified-workflow program (#3456) drops 11 entities. Ten pruned cleanly; `MJ: Scheduled Actions` did not, because 40 `ProcessRun` and 125 `ProcessRunDetail` rows referenced it. Neither table is in the proc — both postdate it (Record Set Processing).

The Phase 0 migration works around this by deleting the dangling rows itself for the entities it drops. That is a point fix, not the general one.

## Uncovered references (partial)

Beyond `ProcessRun` / `ProcessRunDetail`: `EntityAction` (`EntityID`, `ScopeEntityID`), `EntityFormOverride`, `EntityOrganicKey`, `EntityRecordDocument`, `QueryEntity`, `RecordProcess`, `RecordProcessWatermark`, `SearchScopeEntity`, `TaggedItem`, `TagScope`, `VersionLabel` / `VersionLabelItem`, `MLModelScoringBinding`, `MLTrainingPipeline`, `RecordLink`, `SignatureRequest`, `SystemEvent`, `ScopedPromptConfig` / `ScopedPromptPart`, and more.

## Suggested fix

1. Make the proc's cascade list exhaustive — each reference decided as delete-vs-null (nullable FKs like `ProcessRun.EntityID` could be nulled; non-nullable children must be deleted).
[...]

Issue #3545, opened 2026-08-06
PostgreSQL counterparts for the unified-workflow program migrations


AN-BC:
Tracking issue for the `migrations-pg/v6` counterparts of the unified-workflow DAG engine program (plan: #3456).

**Why this exists.** Per AN-BC's ruling on the program, these migrations are perfected on **SQL Server only**; PostgreSQL parity is handled separately by other developers. That collides with `.github/workflows/pg-migrations.yml` step 3a (`scripts/check-pg-migration-parity.mjs`), which fails on any new T-SQL migration in `migrations/v5|v6` without a committed `.pg.sql` counterpart.

`PENDING_CONVERSION` in that script is the only sanctioned deferral path. It was deliberately emptied on 2026-08-05 (the four #3471-era gaps all closed), and its header states that adding an entry "is never the way to make this check pass." Entries are therefore being added **against that guidance, under an explicit program ruling**, and this issue is the tracking record the header requires.

**Migrations awaiting conversion**

- [ ] `V202608061704__v6.1.x__Phase0_Legacy_Workflow_Report_ScheduledAction_Retirement.sql` — Phase 0 legacy retirement. Drops 11 tables (Workflow trio, the `Report*` family, Scheduled Actions, Output Trigger Types), swaps `ContentSource.ScheduledActionID` → `ScheduledJobID`, drops `vwContentSources`, and clears dangling `ProcessRun`/`ProcessRunDetail` rows.

Further phases (1–5) will add entries here as they land.

**Note for whoever picks this up:** entries are self-cleaning — once a `.pg.sql` counterpart exists, the stale `PENDING_CONVERSION` entry makes the script fail until it is removed. Remove the entry in the same PR that adds the counterpart.

AN-BC:
closing because this is handled automatically via build engineering

PR #3553, opened 2026-08-07
feat(v6)!: Phase 0 — retire the legacy Workflow, Report and Scheduled Action subsystems

AN-BC:
**Phase 0 of the unified workflow DAG engine program** — plan: #3456, §4 "Phase 0 — Legacy retirement".

Retires three dead or superseded subsystems so the **Workflow** name is freed for the program's D18 vocabulary work, and so the task-graph engine isn't built alongside a parallel, non-functioning orchestration model.

## What goes, and the evidence

**Eleven tables**, each verified before removal:

| Subsystem | Tables | Evidence it's dead |
|---|---|---|
| Skip v1-era workflow | `Workflow`, `WorkflowRun`, `WorkflowEngine` | Zero non-generated readers/writers; the `SubclassName`-referenced `WorkflowBase` class exists nowhere in the repo; all three empty |
| Skip v1-era reports | `Report`, `ReportCategory`, `ReportSnapshot`, `ReportUserState`, `ReportVersion` | Superseded by conversation artifacts. Every inbound `ReportID` FK is internal to the family. **The renderer is already gone** — the `Reports` resource type names `DriverClass: ReportResource`, which exists nowhere in the repo. All five empty |
| Legacy scheduled actions | `ScheduledAction`, `ScheduledActionParam` | The cron due-check is mathematically always-false (`scheduler.ts:159-171` — `cronParser.next()` is strictly after `evalTime`) and nothing in-repo hosts its Express app, so authored schedules could never fire |
| Report-era triggers | `OutputTriggerType` | Sole referencer was `Report` |

Dropping `Report` outright subsumes the plan's separate `Report.OutputWorkflowID` / `Report.OutputTriggerTypeID` column drops.

## ⚠️ Breaking: the report execution surface is removed

`DatabaseProviderBase.RunReport` was **already** marked `@deprecated` — *"Reports are no longer supported and will eventually be removed. Interactive Components and Artifacts are replacements"* — and it reads `vwReports`, which this migration drops. So the whole surface had to go with it:

- `IRunReportProvider`, the `RunReport` class, `RunReportParams` / `RunReportResult`
- `BaseEntity.RunReportProviderToUse`, `BaseAngularComponent.RunReportToUse`
- `GraphQLDataProvider.GetReportData`
- The `GetReportData` query + `CreateReportFromConversationDetailID` mutation (the latter had no callers)
- `GET /reports/:reportId` (a 501 stub)

Accepted deliberately in the open v6 breaking-change window, the same standard the plan's D12 invokes. **Replacement: Interactive Components and Artifacts.**

## Scope correction vs. the plan

The plan recorded Scheduled Actions as having *"zero non-generated consumers."* The **executor** is indeed dead — but the **entities** were live authoring surface. Four Knowledge Hub / AI dashboards created and read them (autotagging sources, vector sync, tags, pipeline).

Per AN-BC's ruling to do the full sweep, those four are migrated rather than broken. `MJ: Scheduled Actions` → a **`MJ: Scheduled Jobs` row of type Action**, executed by `ActionScheduledJobDriver`:

| Scheduled Action | Scheduled Job |
|---|---|
| `ActionID` | `Configuration.ActionID` |
| `ScheduledActionParam` rows | `Configuration.Params[]` |
| `CronExpression` / `Timezone` / `Status` | same (Status domain is a strict superset) |
| `ContentSource.ScheduledActionID` | `ContentSource.ScheduledJobID` |

The mapping lives in one shared helper (`ng-dashboards/src/shared/action-scheduled-job.ts`) rather than being triplicated across surfaces.

## Two gaps found, each with a tracking issue

- **#3546** — `spDeleteEntityWithCoreDependencies` cascades **~18 of ~73** FK references to `Entity`. Anything uncovered blocks the final `DELETE FROM Entity`, and CodeGen catches-and-logs rather than raising — leaving a **half-pruned entity** (row present, fields gone) that every later CodeGen run silently skips with *"has no primary key field in metadata."* Hit here via 40 `ProcessRun` + 125 `ProcessRunDetail` rows on `MJ: Scheduled Actions`. This migration clears the dangling rows for the entities it drops; the general fix is its own change.
- **#3545** — PostgreSQL counterparts for this program are handled separately by other developers per program ruling, so the migration is recorded in the pg-parity `PENDING_CONVERSION` ratchet.

## Migration notes

`vwContentSources` is dropped **in the migration**, not left to CodeGen: `mj migrate` runs `R__RefreshMetadata` → `spRecompileAllViews` at the very end, which fails with *"Invalid object name '__mj.ScheduledAction'"* before CodeGen ever runs. Learned the hard way.

The migration is idempotent (guarded FK/column/extended-property operations) so a partial failure can be re-run.

## Verification

Validated on a **clean database**, not an incrementally-patched one:

1. Drop + recreate `MJ_6_1_0` empty
2. `mj migrate --dir ./migrations` → **31 applied** from baseline, including this one
3. `npx mj sync push --dir metadata` → 13,622 records, 10 created / 1 updated, **0 errors**
4. `npx mj codegen` → 374 entities, **CRUD validation passed**
5. Verified: 0 doomed tables, 0 doomed entity rows, `Reports` resource type gone, `ContentSource.ScheduledJobID` present, `vwContentSources` regenerated

Plus: **full workspace build 268/268**, and unit tests green in every touched package (MJCore 1672 tests, dashboards 151 files, MJServer 55, MJCoreEntities 27, GraphQLDataProvider 12, SQLServerDataProvider 87, explorer-core 19, PostgreSQL 6, ng-shared 5, base-types 2, Standards 3).

The appended CodeGen section is that clean run's output — verified free of unrelated fresh-install regeneration (no validator functions, no form-layout churn) and using `${flyway:defaultSchema}` throughout with zero hardcoded schema names.

**Not yet run:** the deterministic integration tier. Agreed with AN-BC to do that in a later round.

## Deliberately left alone

- `SUBCLASS_RENAME_MAP` / `entity-rename-map` in CodeGenLib — historical **v5.0** rename data for downstream consumers upgrading their own code, not a live class registry (and a test pins its count).
- The misnamed `ReportResolver` class in `EntityCommunicationsResolver.ts` — it is actually the entity-communication resolver and has nothing to do with reports. Renaming it is out of scope, but it is now the only `ReportResolver` left, which is worth knowing.

## Follow-up tracked on the plan PR

`@memberjunction/scheduled-actions` and `@memberjunction/scheduled-actions-server` are removed from the tree here, but still need **npm deprecation** after this merges — noted on #3456 so it survives the phase boundary.


AN-BC:
@cadam11 @rkihm-BC — heads up on the versioning call here, since this PR carries breaking changes but ships as a **minor**.

## What's breaking

The report execution surface is removed outright — not deprecated, removed:

- `IRunReportProvider`, the `RunReport` class, `RunReportParams` / `RunReportResult`
- `BaseEntity.RunReportProviderToUse`, `BaseAngularComponent.RunReportToUse`
- `GraphQLDataProvider.GetReportData`
- The `GetReportData` GraphQL query and the `CreateReportFromConversationDetailID` mutation
- `GET /reports/:reportId`

Plus eleven dropped tables (the Skip v1-era `Workflow` trio and `Report*` family, the legacy `ScheduledAction` pair, `OutputTriggerType`) and two deleted npm packages (`@memberjunction/scheduled-actions`, `@memberjunction/scheduled-actions-server`).

## The call — AN-BC

> Ordinarily a break like this would require a major version bump. We just bumped to 6.x, and **we have not released an LTS version of 6.x yet**, so I'm going with it — we're fine.

So: **minor bumps within 6.x**, no 7.0. The v6 window is the right place to absorb this, and doing it now — before anything is pinned to a 6.x LTS — is precisely what avoids carrying dead schema into a supported line.

## Why this is lower-risk than the list looks

`DatabaseProviderBase.RunReport` was **already** marked `@deprecated` in-tree — *"Reports are no longer supported and will eventually be removed. Interactive Components and Artifacts are replacements."* This completes a deprecation the codebase had already declared, rather than introducing a new one. The `Reports` resource type also pointed at a `DriverClass` (`ReportResource`) that exists nowhere in the repo, so the renderer had been gone for some time — what remained was wiring that resolved to nothing.

**Replacement for consumers:** Interactive Components and Artifacts.

## What to watch

- **`cadam11`** — nothing here is proposed for the **LTS 5.x** line. 5.x keeps `IRunReportProvider` and the report schema; this is a v6-only removal. Worth confirming that matches your expectations for the backport surface.
- **`rkihm-BC`** — this is the same accepted-break standard you and I litigated for D12 (`ExecuteTaskGraph`) on the plan PR, applied to a wider surface than that decision scoped. The scope expansion and its rationale are written up on [#3456](https://github.com/MemberJunction/MJ/pull/3456#issuecomment-5210321340) if you want the trail.

Flagging rather than assuming — if either of you thinks this warrants a major, better to hear it before merge than after.

---

AN-BC:
@cadam11 @hiltongr @paul-christmann @rkihm-BC removing this technical debt is important so we don't have overlapping legacy concepts that can confuse human or AI developers. Perfect time to do this right now as we are going to 6.x

PR #3564, opened 2026-08-07
fix(migrations): null inbound EntityField.RelatedEntityID refs on entity delete (unblocks next CI)

rkihm-BC:
## Problem

Every open PR's `migrate` step is failing on `next` (e.g. #3529):

```
Migration failed: Failed at batch 19/32: The DELETE statement conflicted with the
REFERENCE constraint "FK_EntityField_RelatedEntity" ... table "__mj.EntityField",
column 'RelatedEntityID'.
FAILED: V202608061704__v6.1.x__Phase0_Legacy_Workflow_Report_ScheduledAction_Retirement.sql
```

The Phase 0 legacy-retirement migration (already merged to `next`, commit c814875d05) deletes 11 interrelated entities via `spDeleteEntityWithCoreDependencies` in **parent-before-child** order:

1. `Scheduled Actions` is deleted **first**
2. `Scheduled Action Params` (deleted later) still carries its `ScheduledActionID` field whose `RelatedEntityID` = Scheduled Actions

The proc cleared only each target entity's **own** `EntityField` rows, never the **inbound** `EntityField.RelatedEntityID` references pointing *at* it from sibling entities. So the parent's final `DELETE FROM Entity` trips `FK_EntityField_RelatedEntity`. Same shape hits `Reports` → its Snapshots / User States / Versions. Flyway halts on this migration, so a follow-up migration can never run — it must be fixed in place (nothing has successfully applied it anywhere).

## Fix

Add an `ALTER PROC [${flyway:defaultSchema}].[spDeleteEntityWithCoreDependencies]` in the hand-DDL section **before** the generated deletion block, so the hardened proc is in effect when the `EXEC`s run. One new statement before `DELETE FROM Entity`:

```sql
-- Clear inbound metadata references from OTHER entities' fields that point AT this
-- entity, so the Entity row can be deleted without tripping FK_EntityField_RelatedEntity.
UPDATE [${flyway:defaultSchema}].EntityField SET RelatedEntityID = NULL WHERE RelatedEntityID = @EntityID
```

This both **unblocks this migration deterministically** and **permanently fixes the class of bug** for any future multi-entity retirement — nulling a reference to a now-deleted entity is the correct end state for every caller. `RelatedEntityID` is nullable, and the `UPDATE` runs after the target's own fields are already deleted, so it only touches inbound refs from other entities.

## Why this is a safe core-proc change

- **No silent regression:** the reproduced SP body is byte-identical to the active v5.46 baseline (`B202607091514`, applied first in CI) apart from the added `UPDATE`. No non-baseline migration alters the proc after that baseline.
- **No source drift:** CodeGenLib only *calls* the proc (`manage-metadata.ts:3036`); no template defines its body. Baselines snapshot the live DB, so the next baseline regeneration carries this fix forward automatically.
- **Behavior change is intended and desirable:** `spDeleteEntityWithCoreDependencies` now nulls inbound `EntityField.RelatedEntityID` for all future entity deletions.

## Verification

- Removal order + SP body confirm the root cause; the 11 `EXEC`s share one GO-less batch (lines 364–1210 = the CI's "batch 19"), so the CI `migrate` re-run is an all-or-nothing proof for the full set.
- Known residual (not exercised here): the proc still does not null `Entity.ParentID` inbound refs — same class one column over, but these legacy entities parent nothing.


migrations/v6/V202608061704__v6.1.x__Phase0_Legacy_Workflow_Report_ScheduledAction_Retirement.sql
```diff
@@ -255,6 +255,59 @@ DROP TABLE IF EXISTS [${flyway:defaultSchema}].[Workflow];
 DROP TABLE IF EXISTS [${flyway:defaultSchema}].[WorkflowEngine];
 GO
 
+-- ════════════════════════════════════════════════════════════════════════════════════
+-- Harden spDeleteEntityWithCoreDependencies before the generated block runs.
```
AN-BC, line 259: update MJ_BASE_BEFORE.sql too

PR #3574, opened 2026-08-07
feat(task-graphs)!: Phase 2 — move execution server-side and make it invocation-agnostic

AN-BC:
**Phase 2 of the unified workflow DAG engine program** — plan: #3456, §4 "Phase 2 — Placement". Follows #3553 (Phase 0) and #3562 (Phase 1).

Phase 1 made the task substrate tell the truth. This phase moves execution off the client so every channel can reach it.

## The gap this closes

Task-graph execution was reachable **only** as a UI convenience of one Angular component. Everything else silently dropped the plan:

| Channel | Before |
|---|---|
| Explorer | Worked — by awaiting the whole workflow in one GraphQL request |
| Slack / Teams | **Dropped every graph.** `detectDelegation` understands `invokeAgent` and a regex over reply text; a structured graph matched neither |
| Scheduled routines | **Dropped every graph.** The payload was serialized into the run record and never inspected |
| Headless / API | No path at all |

## New package: `@memberjunction/task-graph`

Not AI-prefixed, per D11 — an LLM, deterministic code, or a human UI can all construct and submit a DAG.

Graph *semantics* deliberately **stay out of it**: eligibility, failure propagation, parent rollup and stall detection are consumed **unchanged** from the Phase 1 pure algorithms in `ai-core-plus`. That was the entire point of factoring them dependency-free — the in-run and durable executors can't drift apart if neither owns the rules.

## ⚠️ Breaking: `ExecuteTaskGraph` removed (D12)

It awaited an entire multi-step workflow inside a single long-lived GraphQL request. Three unfixable consequences: a page reload lost the awaited promise, a server restart orphaned every in-flight task, and no channel but Explorer could reach the substrate.

Replaced by **`SubmitTaskGraph`**, **`CancelTaskGraph`**, **`RetryTask`**. Accepted deliberately in the open v6 window; the sole known caller becomes an observer in this same PR.

## The claim protocol — three decisions worth your review

**1. `TaskClaimStore` uses direct SQL, not `BaseEntity.Save()`.** This reads as a data-access-rule violation at a glance, so: the entire mechanism is a *conditional* write whose **rowcount is the return value**. `Save()` issues an unconditional update and reports success for a row that changed underneath it — precisely the race being defended against. Claiming is one statement:

```sql
UPDATE Task SET Status='In Progress', ClaimedBy=@me
WHERE ID=@id AND Status='Pending' AND (ClaimedBy IS NULL OR ClaimExpiresAt < now)
```

Rowcount 1 wins, 0 means someone else took it. No distributed lock manager.

**2. Every transition is guarded, not just the initial claim.** Per D20, `MJ: Tasks` stays user-writable — entity forms, Data Explorer, and any agent with an update-record action can change `Status` underneath a running executor. Guarding completion on `ClaimedBy=@me` means a stale executor fails cleanly instead of double-completing.

**3. Human tasks are exempt from reclamation.** A task parked on a person legitimately carries no claim, so `In Progress` with no claim is its *correct* shape — reclaiming it would reset an approval out from under the user.

## Detection at three seams, via one helper

All three call the same `DetectAndSubmitTaskGraph` rather than triplicating the sniff. In `BaseMessagingAdapter` it runs **ahead** of the text-regex delegation, since a structured graph is unambiguous and should win over a text match. It **never throws** — a malformed graph must not take down the Slack reply or scheduled run that produced it.

This shim is explicitly temporary: it dies in Phase 3, when the graph arrives as a typed `nextStep` rather than a payload key to be discovered.

## Provider isolation

The dispatcher mints a fresh provider **per task** through an injected `ProviderFactory`, so parallel tasks never share a transaction scope or entity instances. MJServer supplies the implementation, keeping the dependency **MJServer → task-graph** and never the reverse — the package has no MJServer import.

## Verification

- **Full workspace build: 269/269, 0 TypeScript errors.**
- Unit tests green: **task-graph 18** (new, validator), ai-core-plus 15 files, **MJServer 785**, **messaging-adapters 311**, **scheduling-engine 193**, **ng-conversations 1062**.
- Integration bundle gains **TG4–TG6** — the seam checks deferred from Phase 1 (cycle rejection, unknown-agent rejection, graph persistence with `InputPayload`), now targeting `TaskGraphService`'s public API, which is their correct target.

**Not yet run:** the deterministic integration tier. Still the outstanding item from Phase 0, and it now has six checks waiting for it.

## Exit criteria status

| Plan requirement | Status |
|---|---|
| Slack multi-step executes end-to-end | Code path complete; needs a live Slack run to confirm |
| Reload re-attaches | Client is an observer; work lives in Task rows, not a promise |
| Restart resumes | Startup reconciliation reclaims expired claims |
| Two instances don't double-run a task | CAS claim protocol; needs the two-instance integration check |

The last three want live verification in the integration tier rather than my assertion — flagging that honestly rather than claiming the exit bar is met.


AN-BC (commented):
LGTM

AN-BC:
Review catch on this PR was right, and it's a design miss worth recording rather than just fixing.

**The three mutations should have been Remote Operations.** `SubmitTaskGraph` / `CancelTaskGraph` / `RetryTask` fixed the *durability* problem this phase was about — nothing awaits a whole workflow inside one long-lived request anymore — but left *reachability* exactly where it was. A bespoke resolver is callable from the Explorer client and nothing else, which is the same shape of constraint the phase set out to remove, just moved one layer down. It also undercuts the program's goal of letting agents **set up** workflows rather than only navigate to them.

Record Set Processing — the closest analogous substrate — exposes `Run` / `Pause` / `Resume` / `Cancel` / `Get Run Status` entirely as Remote Operations, and `unified-workflow.md`'s own doctrine table already names Remote Operations "the typed control plane (run/pause/status over one call site)." I had the precedent and the doctrine in front of me and still reached for a resolver.

**A second gap surfaced while fixing it:** this PR landed `TaskGraphDispatcher` but nothing ever instantiated it. A submitted graph persisted correctly and then sat in `Pending` forever — durable and inert, which is strictly worse than the client-driven path it replaced, because at least that one ran. MJServer now starts one instance per process at boot.

Both are fixed in a small follow-up PR, plus a new deterministic integration check (TG7) covering the two failure modes a build cannot catch: a missing metadata row (forgotten `mj sync push` → unreachable) and a subclass that never registers (key typo → routing resolves to the CodeGen contract-only base, so every call succeeds at the transport and does nothing).

Plan ledger and a standing amendment are going onto #3456: **any new task-graph control verb ships as a Remote Operation, not a resolver.**

PR #3585, opened 2026-08-07
feat(core)!: entity companions, related-record collections, unified transaction scope

AN-BC:
## Why

MJ treats a parent record and its child rows as unrelated `BaseEntity` objects. Three applications have each hand-rolled the same parent/children pattern — and each got a different subset of it right:

| | `OrderEntityServer` | `PaymentHeaderEntityServer` | `JournalEntryEntityServer` |
|---|---|---|---|
| Typed | ✅ | ❌ `BaseEntity[]` | ✅ |
| Loads children | ❌ | ❌ | ✅ |
| Tracks removals | ❌ | ❌ | ✅ |
| Add/remove API | raw setter | raw setter | ✅ |
| Re-sequences | ❌ | ❌ | ✅ |

All three are **server-only** classes, because each casts the provider to `DatabaseProviderBase` to reach `BeginTransaction()`. That cast — not the transaction, not the wire — is why composites can't work from the browser today.

## What this adds

Declare a child collection on a **shared (client + server)** entity subclass and call `entity.Save()`:

```typescript
@RegisterClass(BaseEntity, 'MJ_BizApps_Accounting: Journal Entries')
export class JournalEntryEntity extends mjBizAppsAccountingJournalEntryEntity {
    public readonly Lines = this.DeclareRelatedRecords<JournalEntryLineEntity>({
        Name: 'Lines',
        RelatedEntity: 'MJ_BizApps_Accounting: Journal Entry Lines',
        RelatedEntityJoinField: 'JournalEntryID',
        OrderBy: 'LineNumber ASC',
        Load: 'explicit',                          // 'explicit' | 'immediate' | 'lazy' | 'never'
        OnRemove: 'delete',                        // 'delete' | 'orphan' | 'refuse'
        Sequence: { Field: 'LineNumber', From: 1 },
    });

    public override Validate(): ValidationResult {
        const result = super.Validate();           // fans out to companions
        assertBalanced(this.Lines.Items, result);  // now runs in the browser too
        return result;
    }
}
```

`Save()` builds an `EntitySavePlan`. **A single-node plan takes the ordinary save path, byte-for-byte unchanged** — so entities without companions, and composites whose collections are empty, are entirely unaffected. A multi-node plan runs:

| Provider | Behavior |
|---|---|
| `SupportsEntityTransactions === true` (server) | Executes locally inside one transaction scope |
| `false` (browser) | Serializes the graph and routes the whole unit of work to the server via the new `MJ.SaveEntityGraph` remote operation, which rebuilds the records **as their server-side subclasses** and runs the *same* executor there |

There is exactly one cascade implementation; the remote path relocates it rather than reimplementing it. **The graph-save mechanism adds nothing to any generated GraphQL type** — companions ride an existing remote operation rather than a new input field on every entity's Create/Update type, which matters under the publish-then-no-breaking-changes policy. (The schema change below does add one field: `EntityRelationship.RelatedRecordCollection` appears on that entity's generated type, nullable and additive, as any new column would.)

### The local flow

```mermaid
flowchart TD
    Start([entity.Save]) --> Dirty{Dirty?<br/><i>fields OR any collection</i>}
    Dirty -->|no| Skip([return true — nothing to do])
    Dirty -->|yes| Validate[Validate + ValidateAsync<br/>fan out to every collection<br/><b>including pending removals</b>]
    Validate --> Valid{valid?}
    Valid -->|no| Fail([return false<br/>nothing written])
    Valid -->|yes| Plan[BuildSavePlan<br/>self + collections, ordered]

    Plan --> Count{NodeCount}
    Count -->|" = 1 "| Single[<b>Ordinary single-row path</b><br/>byte-for-byte unchanged]
    Count -->|" &gt; 1 "| Cap{provider.SupportsEntityTransactions?}

    Cap -->|true — server| Local[ExecuteGraphLocal]
    Cap -->|false — browser| Remote[ExecuteGraphRemote<br/>see the network flow below]

    Local --> Scope[BeginEntityTransaction<br/><i>starts one, or JOINS an in-flight<br/>one as a savepoint</i>]
    Scope --> Loop[for each node, in order]
    Loop --> Node["node.Entity.Save()<br/><b>the record's own Save</b>"]
    Node --> Guarantees[["Record Changes · entity actions<br/>field validation · subclass overrides<br/>PreSave hooks · events · cache invalidation<br/><i>all fire per node, for free</i>"]]
    Guarantees --> More{more nodes?}
    More -->|yes| Loop
    More -->|no| Commit[scope.Commit]
    Commit --> Accept[AcceptChanges<br/>clear removals, rebase dirty]
    Accept --> Done([return true])

    Node -.->|any node fails| Rollback[scope.Rollback]
    Rollback --> Failed([return false<br/><b>nothing persisted</b>])

    style Single fill:#1b5e20,stroke:#66bb6a,color:#fff
    style Guarantees fill:#0d47a1,stroke:#64b5f6,color:#fff
    style Rollback fill:#b71c1c,stroke:#ef5350,color:#fff
    style Failed fill:#b71c1c,stroke:#ef5350,color:#fff
```

### The network flow — one call, not N

The client provider cannot open a transaction; that single fact is why every hand-rolled composite in MJ is server-only today. So `BaseEntity` **relocates** the cascade rather than reimplementing it:

```mermaid
sequenceDiagram
    autonumber
    participant UI as Browser<br/>(OrderEntity)
    participant GQL as GraphQLDataProvider
    participant OP as MJ.SaveEntityGraph<br/>(remote operation)
    participant CF as ClassFactory
    participant SRV as OrderEntityServer
    participant DB as SQL Server

    Note over UI: order.Lines.Add(line1)<br/>order.Lines.Add(line2)<br/>Discount = 10

    UI->>UI: Validate() — runs in the BROWSER<br/>line rules fail before any round trip
    UI->>UI: BuildSavePlan → 3 nodes
    UI->>UI: SupportsEntityTransactions = false
    UI->>UI: SerializeCompanions()<br/>Companions___: [{Name:'Lines', Data:{Items:[…], Removed:[…]}}]<br/><b>IsNew travels explicitly</b>

    UI->>GQL: RouteOperation(MJ.SaveEntityGraph)
    GQL->>OP: ONE network call<br/>{EntityName, Fields, Companions}

    Note over OP: API-key scope gate per node<br/>(entity:create / update / delete)<br/>BEFORE any entity work

    OP->>CF: GetEntityObject('…: Orders')
    CF-->>OP: OrderEntityServer<br/><b>the server subclass, not the base</b>
    OP->>SRV: LoadFromData(fields) + DeserializeCompanions(mode:'request')

    Note over SRV: 'request' mode LOADS existing rows first.<br/>Skipping that makes old == new, the record<br/>looks clean, and the edit is silently dropped.

    SRV->>SRV: Save() → SupportsEntityTransactions = true
    SRV->>DB: BEGIN TRANSACTION
    SRV->>DB: INSERT OrderHeader (or UPDATE)
    SRV->>DB: INSERT OrderLine 1 — FK stamped from the parent key
    SRV->>DB: INSERT OrderLine 2
    SRV->>DB: DELETE removed lines (OnRemove:'delete')
    SRV->>DB: COMMIT

    SRV-->>OP: saved graph
    OP-->>GQL: {Fields, Companions} — <b>a graph, not a row</b>
    GQL-->>UI: result

    UI->>UI: DeserializeCompanions(mode:'result')

    Note over UI: 'result' mode adopts values VERBATIM —<br/>no re-query. New PKs, computed columns and<br/>assigned LineNumbers land on the in-memory<br/>objects, so the client is not holding a<br/>half-saved graph that merely looks saved.
```

### Hooks, events and platform guarantees

Every node is persisted through that record's own `Save()`/`Delete()` — never direct SQL. So Record Changes, entity actions, field validation, subclass `Save` overrides, `PreSave` data hooks, `save_started`/`save`/`delete` events and cache invalidation all fire **per node**, with no graph-specific plumbing and no risk of the graph path quietly skipping a guarantee the single-record path has. The root additionally raises `graph_save_started` / `graph_save` so a UI can refresh once per unit of work rather than once per line.

## Transaction unification (and a torn-write bug this fixes)

`DatabaseProviderBase` gains `BeginEntityTransaction()` / `SupportsEntityTransactions`, delegating to the existing depth-counted `BeginTransaction()` — so it starts a physical transaction **or joins one already in flight** as a savepoint. Participants never ask who else is in a transaction; the provider arbitrates.

This closes a real defect. There were two mechanisms and they were **blind to each other**:

- `BeginTransaction()` — depth-counted, savepoints at depth ≥ 2, in-flight serialization. What every hand-written application cascade calls.
- `BeginISATransaction()` — four lines opening a brand-new `sql.Transaction` on the pool with **no depth awareness**.

An IS-A entity saved inside an application transaction therefore wrote into **two independent physical transactions on the same pool**. Rolling one back left the other committed, and nothing raised an error.

> **IS-A's own orchestration is untouched.** `_parentEntity`, chain-save ordering, disjoint-subtype enforcement — all unchanged. The fold is *provider-level only*, and it **removes** coupling: participants can keep knowing nothing about each other precisely because a single arbiter now decides. Verified that MJServer builds per-request providers, so the ambient transaction is request-scoped.

## Defects fixed along the way

1. **A clean parent with new children silently saved nothing.** `_InnerSave` returns early when `!Dirty`, and `Dirty` knew nothing about children — so the save reported success and wrote no rows. `Dirty` now rolls up companions.
2. **Cross-child invariants were silently disabled.** `DefaultSkipAsyncValidation` defaults to `true`; `OrderEntityServer` never overrides it and no caller passes `SkipAsyncValidation: false`, so its "cannot confirm an order with no lines" guard *and* its entire per-line validation loop are dead on every save today. Companion validation now runs regardless of that flag.
3. **N+1 in production accounting.** `JournalEntryEntityServer.LoadFromData` calls `LoadLines()`, and `LoadFromData` is the per-row materialization path for `RunView(ResultType:'entity_object')` — so listing 500 journal entries issues 500 line queries plus 500 dimension queries. Eager companion loading is structurally excluded from `LoadFromData()`; `RunViewParams.IncludeRelatedRecords` does one batched query per collection instead (1+K).
4. **Wire-level saved-state ambiguity.** `NewRecord()` generates a UUID, so a brand-new child already has a populated PK and is indistinguishable from an existing one by inspection. `IsNew` now travels explicitly.
5. **Edits to existing children would have been silently dropped** on the remote path — `LoadFromData(row, replaceOldValues: true)` makes old == new, so the child reports clean and its save is skipped. Existing children are now loaded before wire values are applied.

(4) and (5) were found in my own code during self-review and both have regression tests.

## 🚨 BREAKING — the IS-A transaction surface is removed

6.x LTS has not shipped, so these are deleted rather than carried as 7.0 debt:

- `IMetadataProvider.BeginISATransaction` / `CommitISATransaction` / `RollbackISATransaction`, and their `SQLServerDataProvider` implementations
- `BaseEntity.ProviderTransaction`
- `BaseEntity.PropagateTransactionToParents()`

Nothing set the handle after the unification, so the three provider reads that consumed it were **already dead code** — an `ExecuteSQL` call with no explicit `connectionSource` picks up the provider's ambient transaction, which is exactly what the unified scope opens. Removing them deletes a branch from three providers rather than leaving one that can never be taken.

**Migration for callers:** `BeginEntityTransaction()`, or `RunInEntityTransaction(provider, work)` which handles commit/rollback for you.

## Naming

This feature says **related records** throughout — `DeclareRelatedRecords`, `RelatedRecordCollection`, `RelatedEntity`, `RelatedEntityJoinField` — because in MJCore **`child` already means IS-A subtype**: same primary key, vertical, at most one per parent (`EntityInfo.ChildEntities`, `IsChildType`, `FindISAChildEntity`). MJ's word for "rows that point at me" is *related* (`EntityInfo.RelatedEntities`, `EntityRelationshipInfo`). Using "child" here would have inverted the platform's own vocabulary, and the option names now mirror the `EntityRelationship` columns they are generated from — which is what makes the metadata mapping a direct read rather than a translation layer.

IS-A terminology is untouched.

## Schema: metadata-driven declaration

Adds nullable **`EntityRelationship.RelatedRecordCollection`** (JSONType `IRelatedRecordCollectionConfig`) holding the policy half of a `DeclareRelatedRecords(...)` call: `Name`, `Load`, `OnRemove`, `OrderBy`, `Sequence`, `ClearAfterSave`.

`RelatedEntity` and `RelatedEntityJoinField` are deliberately **not** in the JSON — they are already columns on the same row, and duplicating them would create two sources of truth with the JSON copy winning silently. That is precisely why the runtime option names were aligned to those columns: CodeGen can read two columns plus one blob and emit the declaration with no translation layer.

JSONType rather than scalar columns because the declaration is a small, evolving policy object (`Sequence` is nested; `Load` and `OnRemove` are growing value lists), not independent facts to query or index. Adding an option becomes an interface edit plus `mj sync push`, with no schema change — the same shape `UserView.GridState` and `CardState` already use.

Additive: `NULL` — every existing row — means "not a declared collection".

**CodeGen reads it.** `EntitySubClassGeneratorBase.GenerateRelatedRecordCollections()` takes the two columns plus the JSON blob and emits the `DeclareRelatedRecords(...)` field initialiser onto the generated subclass — a field initialiser rather than constructor code because generated subclasses have no constructor, and TS runs initialisers right after `super()`, when `EntityInfo` and the provider are both set. Invalid metadata is skipped with a logged error, never thrown: one malformed row must not abort a run and leave the repo with zero generated entities. Validated: JSON parseability, object shape, `Name` present *and* a legal TS identifier, join field present, One-To-Many only, value lists, `Sequence.Field`, and duplicate names case-insensitively. **30 CodeGen tests**, including one asserting the JSON cannot override `RelatedEntity`/`RelatedEntityJoinField`.

The migration carries its **generated CodeGen counterpart** (the `EntityField` row, the entity-field category/display updates, regenerated `spCreate`/`spUpdate`/`spDelete` + grants, and `sp_refreshview` on `vwEntityRelationships`), appended below the hand-written DDL behind the standard do-not-edit separator. That row is load-bearing rather than cosmetic: without it the column is invisible to the metadata layer, `EntityRelationshipInfo.RelatedRecordCollection` is always `null`, and the generator above has nothing to read. It was produced against a database built from this branch's full migration set (`next` + this DDL) so the output is attributable to this migration alone.

Ships with its PostgreSQL counterpart, so it adds nothing to the existing PG parity debt (two unrelated pre-existing gaps remain on `next`). The PG counterpart is DDL-only; baking its generated section needs a live PG database and is build-engineer work.

## Behavior changes for adopters

No effect on entities without companions. For those that adopt:
- `Dirty` includes companions — a previously-skipped save now runs (this is the fix for defect 1).
- Companion validation ignores `DefaultSkipAsyncValidation` (the fix for defect 2).

## Verification

Verified against two purpose-built databases migrated from this branch's full set — `next` plus this branch's DDL — so nothing from any other in-flight branch is in scope. The second (`MJ_6_1_0_BaseEntity_IT`) was built fresh **after** the CodeGen section was appended, reproducing what CI does.

| Tier | Result |
|---|---|
| Unit — MJCore | **1775/1775** across 117 files, including every IS-A suite whose transaction path changed |
| Unit — CodeGenLib | 723 passed / 60 skipped; 10 failures in `pg-connection.test.ts`, which is byte-identical to `next` and fails there too |
| Unit — integration-test-suite | **155/155** — the bundle↔metadata drift checks, including IT72's sibling parity |
| Integration — **IT72 (new)** | **8/8** with `RUN_MUTATION_TESTS=1`, against a real SQL Server |
| Integration (deterministic, CI shape) | **58/58** bundles |
| Integration (deterministic, mutations on) | 56/58 — see below |
| Integration (live model) | 12/15 — see below |
| Workspace build | **300/300** packages |
| `check:claude-md` · `check:standards` | pass |
| PG migration parity | flags this migration — the counterpart is build engineering's, see Scope |

**The migrations-only proof.** On a database built from migrations alone — no CodeGen ever run against it — the column, its `EntityField` row and its presence in `vwEntityRelationships` are all there, and `mj sync push --dir=metadata --ci` completes. That is the precise path that was failing, and it is what the appended section fixes.

**The five failures, none of them this branch's.** Two deterministic checks fail only when `RUN_MUTATION_TESTS=1` (CI does not set it): IT29 Cache Gauntlet asserts row-count preconditions a freshly-seeded database cannot satisfy (`need >= 2 existing rows … found 0`), and IT67 Content Vectorization fails on `infra.vectorDB.GetSourceRecordFieldPaths is not a function`. Three live-model tests fail on model behavior, which the harness classifies itself — `model-noncompliance: the model never took the instructed action after 3 attempts. Fix the prompt, not the check.` (IT56, IT57) — plus IT17 hitting its 300s hung-check budget. This branch changes no vectorization, cache or agent-prompt file, and IT18 Concurrent Run Persistence — the test that most directly stresses the save queue this branch touches — passes. I did not run the live-model tier against `next` for a baseline, so those three are reported as observed rather than proven pre-existing.

## New integration coverage — IT72 `entity-graph`

Until now this feature's entire test suite ran against a **mock provider**. That proves the save *plan* is built and ordered correctly; it cannot prove the plan reaches a database. IT72 (`packages/TestingFramework/integration-test-suite/src/checks/entity-graph.checks.ts`) closes that gap against the `MJ: Lists` → `MJ: List Details` pair — chosen because a NOT NULL `ListID` plus a numeric `Sequence` column exercises FK stamping, ordering and renumbering in one fixture, and because IT27 already established the safety profile for throwaway `MJ: Lists` rows.

| Check | The invariant |
|---|---|
| EG1 | A declared-but-empty collection is **inert** — the parent still takes the ordinary single-row path on insert *and* update |
| EG2 | Parent + three related records from ONE `Save()`; FK stamped from the parent's key, `Sequence` assigned 1,2,3 |
| EG3 | `Load: 'explicit'` does not populate with the parent, and materialises in `OrderBy` order when asked |
| EG4 | An edit to an **existing** related record persists — the `old == new` regression that silently dropped edits |
| EG5 | `OnRemove: 'delete'` deletes the row and renumbers the survivors from 1 |
| EG6 | A **clean** parent with a new related record still saves — the `Dirty` rollup, without which `_InnerSave` returned success and wrote nothing |
| EG7 | A related record that cannot be written rolls the **whole graph** back, including the parent's committed field change |
| EG8 | `IncludeRelatedRecords` hydrates every returned row with one batched query per collection, correctly distributed, leaving nothing dirty |

EG7 is the one a mock cannot fake, and EG1 is why registering the test subclass process-wide is safe for every other bundle's `MJ: Lists` fixtures — it asserts the inertness rather than assuming it.

**Worth stating plainly: all eight carry `RequiresMutation: true`, so CI — which does not set `RUN_MUTATION_TESTS` — skips them and IT72 passes vacuously there.** That mirrors IT27 and IT40, whose write-side checks are gated the same way, so this PR follows the established convention rather than quietly changing it. But it does mean the coverage above is real only when someone runs the tier with the flag (as was done for this PR). Since CI provisions a throwaway SQL Server container per run, dropping the gate for this bundle would be safe and would make the coverage actually bind in CI — worth deciding as a repo-wide question rather than unilaterally here.

## Docs

- **New `packages/MJCore/docs/related-record-collections.md`** — a sibling to `docs/isa-relationships.md`, where MJ documents advanced BaseEntity features. It opens by contrasting the two (horizontal vs vertical halves of the same idea, and why MJCore reserves "child" for IS-A subtypes), then covers declaration, the save flow, the network round trip, `Source`/`ReadOnly`/`Load`, removal and sequencing, and the cycle guard. Three mermaid diagrams — the local save, the OrderHeader/OrderLine network sequence, and `Source` resolution with the share-vs-copy fork.
  - **A shipped-examples section** walks the eight core collections at the call site, with a table contrasting a declared collection against the hand-written getters it replaced (browser availability, refusing accidental mutation, tracking the engine, batched loading).
  - **An operations cookbook** covers every operation with runnable examples: reading (iteration, spread, destructuring, `Items`), adding via `Create` and `Add`, updating through the `Dirty` rollup, removing under each `OnRemove` mode, clear-and-replace, cascading a parent delete, cross-record validation, all four load paths, what a read-only collection refuses, and the introspection surface.
- **`guides/TRANSACTIONS_AND_BATCHING_GUIDE.md`** (indexed) — disambiguates the three mechanisms that all sound like "do several things at once", with a decision tree and anti-patterns. Explicitly covers **why a TransactionGroup is not a composite-save engine**: saves are *deferred*, so the parent's PK is unavailable, there is no read-your-writes, and `Save()` returns `true` before anything persists.
- MJCore `readme.md` — index entry plus an expanded composite section (metadata declaration, cache sourcing, iterability, `LoadRelatedRecords`); `docs/isa-relationships.md` cross-linked both ways
- `SQLServerDataProvider` and `GraphQLDataProvider` READMEs
- `.claude/rules/data-access.md`, so the routing loads on every `.ts` file
- **`migrations/CLAUDE.md`** — the documented `migrate → sync push → codegen` order silently assumes the field's `EntityField` row already exists. For a **new** column it does not (CodeGen is what creates it), so a JSONType metadata file resolving `@lookup:MJ: Entity Fields...&Name=<new column>` fails outright, and it fails *only in CI*, because a dev database already has the row from the developer's own CodeGen run. Documented the four-step order that actually works, plus the trap that a full `mj codegen` at step 2 regenerates `remote_operations.ts` from unseeded metadata, deletes every remote-operation class, and breaks the CLI you need for step 3.
- JSDoc throughout explains *why*, citing the specific defects rather than restating signatures

## Cache-backed collections, and the MJ core adoption

A collection can now read from an engine cache instead of the database. Three new declaration axes:

```typescript
Source?: 'database' | 'cache';                          // default 'database'
ReadOnly?: boolean;                                     // default false; TRUE when Source='cache'
Load?: 'explicit' | 'immediate' | 'lazy' | 'never';     // 'eager' renamed; 'lazy' is new
```

`'cache'` resolves through `BaseEngineRegistry.FindCachedEntity` — it finds *whichever* loaded engine already holds the entity, so this is generic rather than wired to a named engine, and it falls back to a database load on a miss. Not called `'query'` because in MJ a Query is a stored named artifact; that would be the same class of collision the child/related rename fixed.

```mermaid
flowchart LR
    Read([collection populates]) --> Src{Source}

    Src -->|database| RV[RunView filtered by the join field]
    RV --> Fresh[[Always fresh · costs a query<br/>Correct for transactional data]]

    Src -->|cache| Reg[BaseEngineRegistry.FindCachedEntity]
    Reg --> Found{a LOADED engine<br/>caches this entity?}
    Found -->|yes| Filter[filter its array by the join field]
    Found -->|no| Fallback[fall back to the database load]
    Fallback --> RV

    Filter --> RO{ReadOnly?}
    RO -->|"true (the default for cache)"| Share[[Hand out the ENGINE's instances<br/>zero allocation · a LIVE view]]
    RO -->|false| Copy[[COPY into fresh entities<br/>the cache is never mutated in place]]

    style Fresh fill:#0d47a1,stroke:#64b5f6,color:#fff
    style Share fill:#1b5e20,stroke:#66bb6a,color:#fff
    style Copy fill:#4a148c,stroke:#ba68c8,color:#fff
```

**The sharing hazard is settled declaratively, not by documentation.** A cache-sourced collection holds the engine's own entity instances. Read-only shares them (zero allocation, the point of caching); writable **copies** them on load so the cache is never mutated in place. Read-only also forces `Dirty` to `false` — otherwise a record dirtied by unrelated code would make every parent holding it claim it needs saving.

`'lazy'` fills on first read of `Items` and requires **both** cache and read-only, because only sharing is synchronous — `GetEntityObject` is async and a getter cannot await. CodeGen rejects the other combinations rather than emitting a declaration that compiles and never fills.

**A lazy cache miss throws.** Declaring `lazy` asserts an engine caches the entity; with no async fallback available from a getter, the alternative is a silently empty array — which is exactly how a getter can feed `[]` to its callers indefinitely. A donor holding zero rows is a valid empty answer; only the *absence* of a donor is an error, and the message distinguishes "nothing caches this entity" from "the engine that does isn't loaded yet", because those need opposite fixes. That needed a new `BaseEngineRegistry.FindEnginesDeclaringEntity`, since `FindCachedEntity` only sees loaded engines.

`BaseEntity.LoadRelatedRecords(...names)` resolves cache-backed collections for free and batches every database-backed one into **one `RunViews`** — four declared collections cost one round trip, or zero.

### Adopted across four core entities — as metadata, not code

| Parent | Collections |
|---|---|
| `MJ: Actions` | `Params`, `ResultCodes`, `Libraries` — cache · lazy · read-only |
| `MJ: AI Agents` | `Actions`, `SubAgents` — cache · lazy · read-only · `Prompts` — database · writable · sequenced |
| `MJ: AI Prompts` | `Models` — cache · lazy · read-only |
| `MJ: API Keys` | `Scopes` — cache · lazy · read-only |

**What that buys at the call site.** A collection is **iterable**, so `for…of`, spread and destructuring work directly; `Items` gives the array for `map`/`filter`/`find` and is `readonly`, which is what stops a caller pushing around the removal tracking and sequencing:

```typescript
const action = await md.GetEntityObject<MJActionEntity>('MJ: Actions', contextUser);
await action.Load(actionId);              // ONE query — the action row

for (const p of action.Params) { … }      // iterate — ZERO queries, filled from ActionEngineBase
const all = [...action.Params];           // spread
const [first, ...rest] = action.Params;   // destructure
action.Params.length                      // and length (=== Count)
action.Params.Items.filter(p => p.IsRequired);

await action.LoadRelatedRecords();        // Params + ResultCodes + Libraries — still zero queries

const agent = await md.GetEntityObject<MJAIAgentEntity>('MJ: AI Agents', contextUser);
await agent.Load(agentId);
agent.Actions.Items                       // from BaseAIEngine's cache
agent.SubAgents.Items                     // the ParentID hierarchy, ordered by ExecutionOrder
```

Writes go through the collection so removals are tracked, the FK is stamped and the sequence stays gap-free:

```typescript
// AI Agent Prompts is the one child no engine caches — database-sourced, and writable.
await agent.Prompts.Load();

const p = await agent.Prompts.Create();   // attached already
p.PromptID = somePromptId;                // AgentID and ExecutionOrder are assigned for you

agent.Prompts.Items[0].Status = 'Disabled';   // update — the parent's Dirty ROLLS UP
agent.Prompts.Remove(agent.Prompts.Items[1]); // OnRemove:'delete' → row deleted on save

await agent.Save();                       // insert + update + delete, ONE transaction
```

Read-only collections refuse mutation with an actionable message rather than silently accepting it:

```typescript
action.Params.Add(param);  // throws: "…is read-only… sourced from a BaseEngine cache…
                           //  Declare ReadOnly: false to get copies you can safely modify,
                           //  or Source: 'database'."
```

Setting `EntityRelationship.RelatedRecordCollection` is the whole adoption — CodeGen emits `DeclareRelatedRecords` onto the **generated** classes, so both tiers get them and no server subclass declares anything. The only TypeScript written was deletion:

- `MJActionEntityExtended` loses `Params` / `ResultCodes` / `Libraries` — memoized filters over `ActionEngineBase`'s caches, which is precisely what cache+lazy does generically
- `MJAIAgentEntityExtended` loses `Actions`
- **`BaseAIEngine` loses its manual association loop** — it filtered `_agentActions` by `AgentID` into each agent by hand; the collection now does that from the same cache

Call sites move to `.Items`; the compiler found every one across Actions, agents, CodeGen, Angular templates and the test suite.

**A cycle guard** ships with this, because `SubAgents` is self-referential: `a.SubAgents.Add(b); b.SubAgents.Add(a)` would otherwise recurse until the stack died. It keys on entity+primary-key (not object identity — a round trip produces a different instance for the same row) and rides on `EntitySaveOptions` rather than a module global, so concurrent requests touching the same record can't produce phantom cycles.

## Scope

**Adoption in `bizapps-orders` / `bizapps-accounting` is deliberately not in this PR** — those are peer repos; `plans/base-entity-composite-graph.md` documents exactly what each loses (`savePendingLines`, `_deletedLines`, the transaction blocks, the `BaseEntity[]` cast) and what it keeps.

**No PostgreSQL counterpart.** An earlier revision carried a hand-authored one; it has been removed because it was inferred rather than produced by the conversion toolchain, and was DDL-only besides — no baked CodeGen section, so a PG deployment would have hit the very bug this PR fixes for SQL Server. The counterpart should come from build engineering's `--bake-codegen` process against a live PG instance. The parity gate flags it accordingly.

## Follow-up worth filing separately

`EntityRelationshipsToLoad` has never worked: the server generates `{Entity}_{JoinField}**Array**` field resolvers, the client requests the same name *without* the `Array` suffix, and `InnerLoad` then reads `data[<entity name>]`. A three-way key mismatch. This design deliberately does not depend on it.


AN-BC:
## Where to spend review attention

This PR is large, and the parts that most deserve scrutiny are not the parts with the most lines. Ranked by where I think a reviewer's time actually pays:

### 1. The transaction unification — `entityTransactionScope.ts`, and the IS-A call sites in `baseEntity.ts`

This is the only change that alters behavior for code that does **not** adopt anything. It folds `BeginISATransaction()` onto the depth-counted `BeginTransaction()`, which fixes a genuine torn-write (the two mechanisms were blind to each other, so an IS-A entity saved inside an application transaction wrote into two independent physical transactions on one pool). It also **removes published API**. IS-A orchestration itself is untouched, but if I got the scope join wrong, the blast radius is every IS-A save on the platform — not just adopters. The IS-A suites all pass, and IT47/IT27/IT40 pass, but this is the one to read closely.

### 2. `ReadOnly` forcing `Dirty` to `false` — `relatedRecordCollection.ts`

The reasoning: a cache-sourced collection holds the *engine's own instances*, so a record dirtied by unrelated code would otherwise make every parent holding it claim it needs saving. That is why read-only suppresses the rollup rather than just refusing mutation. If that reasoning is wrong, seven of the eight adopted collections silently stop propagating dirtiness they should propagate. Worth a second opinion.

### 3. The lazy-miss throw — `populateLazyOrThrow()`

A lazy cache miss **throws** instead of returning an empty array. Deliberate: declaring `lazy` asserts an engine caches the entity, and a getter cannot await a fallback. But it means a property *read* can throw, which is unusual. The two messages distinguish "nothing caches this" from "the engine exists but is not loaded yet" because they need opposite fixes. `Items` is only reached externally — internal paths use the private array — so validation and save cannot trip it. Confirm you agree with that trade.

### 4. What is **not** covered locally

- **IT72's 11 checks are `RequiresMutation`, so CI skips them.** They pass under `RUN_MUTATION_TESTS=1` (which is how they were run), matching IT27/IT40's convention. But the integration coverage does not bind in CI.
- **The writable graph path is thinly exercised in core.** Seven of eight adopted collections are read-only cache projections; only `AI Agents → Prompts` is writable, plus IT72's own fixtures. The atomicity claims are proven (EG2/EG5/EG7 against a real database), but adoption mostly demonstrates the read path.

### 5. Where mock-based tests were not enough

Three real bugs in this PR surfaced *only* against a live database or from behavioral tests, having passed fixture-based suites:

- the CodeGen emitter produced a generic type that could not compile (`MJActionParam` instead of `MJActionParamEntity`) — 30 fixture tests green, because the fixture pre-suffixed the input in a way real metadata never does
- `copyRecords` silently shared the engine's instances when no factory was available, defeating the only reason to declare a writable cache collection
- `Count` did not trigger lazy population, so `Count === 0` while `Items.length === 2`

Worth carrying that skepticism into anything here that is only mock-covered.

### Deliberately red checks

- **PG parity** — the hand-authored counterpart was removed. It was inferred rather than produced by the conversion toolchain, and was DDL-only besides, so a PG deployment would have hit the exact bug this PR fixes for SQL Server. Belongs to build engineering's `--bake-codegen` process.
- **Migration filename** — the gate computes `v6.2` from `6.1.0-edge.0`; in-flight migrations target `v6.1.x` by team convention.

PR #3588, opened 2026-08-07
feat(task-graphs): make durable task graphs a first-class agent primitive

AN-BC:
Phase 3 of the unified workflow DAG engine program (plan: #3456). Durable task graphs stop being something an agent *smuggles in a payload* and become something it *emits as a step*.

## `'Tasks'` joins the Loop response union

An opted-in agent emits `nextStep.type = 'Tasks'` carrying a `TaskGraphSpec`, and the framework does the rest.

The distinction from `subAgents[]` is **durability, not parallelism** — both fan out. `subAgents[]` is *ephemeral*: it blocks the run and dies with it, so a page reload or a restart loses the work. A task graph is *durable*: real Task rows a server-side dispatcher owns, visible in the Tasks UI, resumable after a restart, able to wait on a human. That framing is in the prompt docs, because "when do I use which" is the only question an agent actually has to answer here.

## The capability gate is enforced, not advisory

`enableTaskGraphs` defaults to **false** — the only Loop prompt parameter that does.

The others merely shape the prompt: turning one off saves tokens, and an agent that ignores the omission and emits the feature anyway still works fine. This one governs whether an agent may create durable rows that outlive its run, execute on a dispatcher under the submitting user, and spawn further agent runs. Acquiring that reach through prompt drift rather than deliberate configuration is a real problem, not a token-budget one.

So beyond omitting `'Tasks'` from the prompt, `LoopAgentType` **rejects** the step from a disabled agent with a corrective that steers it back to Sub-Agent/Actions. It fails closed — an absent flag, an absent params bag, and the string `"true"` are all refusals.

**This matters more than it looks.** `HarnessAgentType extends LoopAgentType` and *intentionally* inherits `DetermineNextStep`, so adding `'Tasks'` to the Loop union hands the primitive to external agent harnesses (Claude Code / Codex / Pi running inside MJ) at the same instant it reaches ordinary Loop agents. That's the design working — but it promotes the gate from "nice property of one class" to "the thing standing between a sandboxed external CLI and durable server-side work executed under the invoking user." It is therefore tested **through the harness path**, and the inheritance itself is pinned (`hasOwnProperty('DetermineNextStep') === false`) so a future override can't silently move those assertions onto a different code path.

## `TaskGraphSpec` and its validator move to `ai-core-plus`

Beside the pure graph algorithms they belong with. This is what lets the agent framework validate a graph **without depending on the durable-execution package** — which would otherwise drag the entity layer and the dispatcher into every context that merely runs an agent, including unit tests with no database.

Loop validates the identical contract the server re-validates at submission (D16), so a graph cannot pass one check and fail a different one later. Correctives report *every* failure at once — a model fixing a malformed graph shouldn't discover the problems one round-trip at a time.

## Constant folding (D9) — recorded, never silent

A one-node graph with no edges, an agent assignee and default continuation is rewritten into an ordinary in-run sub-agent call. Don't spin up loop machinery for a loop of one.

The `TaskGraph` run step is written **either way**, carrying the spec, a `folded` flag and the reason. Three consequences:
- run forensics show why a graph did or didn't reach the dispatcher;
- a user who edits a two-node graph down to one can read the durability change off the run record instead of inferring it;
- Save as Workflow (D17) attaches to the recorded spec, so the single-node case — the shape most likely worth promoting — stays promotable like any other.

`durable: true` opts back into a Task row. Folding is also skipped for human tasks and non-default continuations.

## Submission crosses a registered seam

`TaskGraphSubmitter` is declared in `ai-core-plus` and implemented in `@memberjunction/task-graph`, resolved through the ClassFactory. A host with no durable-execution package loaded gets `null` and the agent reports an honest failure. What must never happen is a graph vanishing quietly while the model believes it scheduled work.

## Continuation contract

The parent Task row durably carries `continuation`, `reinvokeDepth` and the delivery marker — persisted rather than held in memory because the dispatcher instance that *finishes* a graph is routinely not the one that accepted it.

**Delivery marks before it acts.** Delivery is at-least-once by nature; the process can die between "the graph is done" and "the user has been told." Marking first makes the worst case a *missed* notification that shows up in the task record as delivered — recoverable and inspectable. Marking after would make the worst case a notification repeated on every reconciliation sweep, forever, which for `reinvoke` is an unbounded agent-run loop. Given one of the two has to be chosen, the quiet failure is the safe one.

Chains cap at 5 hops, bounded **separately** from task-nesting depth because they are different loops. At the cap the mode degrades to `message` so results still reach the user rather than being refused.

⚠️ **`'reinvoke'` itself is not wired in this phase.** Implementing it would invert the dependency to task-graph → ai-agents. It degrades to `message` with a log line rather than silently doing nothing, and lands in Phase 4 where the dispatcher already holds an execution engine.

## Sage and the Workflow Planner stop payload-smuggling

Both prompts move from `payloadChangeRequest.newElements.taskGraph` to the real `nextStep`. Phase 2's temporary server-side payload sniff (`TaskGraphDetection.ts`) is **deleted** along with its messaging-adapter and scheduling call sites — the primitive submits inside the run, so channel seams no longer need to look.

## Launch opt-ins (D3)

Sage, **Workflow Planner**, Query Builder, and the Research Agent with its four sub-agents.

⚠️ **Workflow Planner is not on the plan's opt-in list**, but emitting task graphs is that agent's entire job — leaving it gated would have broken it outright the moment the gate landed. Flagging as a deliberate deviation.

## Verification

- Full workspace build green.
- Unit tests: ai-core-plus **272**, ai-agents **1824**, AgentHarness **39**, task-graph **20** — 43 of them new (18 Loop, 5 harness, 20 continuation-metadata).
- **IT71 runs 8/8** against SQL Server, including new **TG8**, which covers both directions the metadata gate can be wrong: an opt-in that was never pushed leaves an agent unable to delegate at all (Sage's entire delegation path runs through `'Tasks'`), and a Loop **type** default left on would hand durable reach to every Loop agent in the install at once.

`next` is merged in rather than rebased, per request.

## Before merging

Phase 3 rewrites the flagship agent's most-exercised path, so the plan calls for a **manual Sage soak on a running instance** before this merges. Not yet done.


cadam11:
Heads up: next's unit backstop went red after this merged. integration-test-suite's sibling-parity check fails on 'every IT metadata record is joined to the integration suite (MJ: Test Suite Tests)' — runs before 16:21 UTC passed, the 16:59 run fails. Did phase 3 add an IT bundle or a Test Suite Test metadata record without its sibling? Every open PR is inheriting this red.

Issue #3427, opened 2026-08-04
Hand-written resolvers bypass provider.ExecuteSQL, breaking dialect portability under PostgreSQL


jordanfanapour:
## Summary

Several hand-written (non-CodeGen-generated) resolvers in `packages/MJServer/src/resolvers` execute raw SQL by constructing `new mssql.Request(dataSource)` / `new sql.Request(dataSource)` directly against `AppContext.dataSource`, instead of going through the dialect-portable `provider.ExecuteSQL(...)` abstraction. This works against SQL Server but is expected to fail (or silently misbehave) against a PostgreSQL-backed MJ instance.

## Why this breaks under Postgres

`AppContext.dataSource` is typed as `sql.ConnectionPool` (`packages/MJServer/src/types.ts:37`), but that type doesn't hold for the Postgres path. When `dbType === 'postgresql'`, `dataSource` is actually `createMSSQLCompatPool(pgPool)` (`packages/MJServer/src/index.ts:1712-1729`) — a plain duck-typed object (`{connected, query, request, _pgPool}`) cast `as unknown as sql.ConnectionPool`. It is **not** a real instance of the `mssql` npm package's `ConnectionPool` class.

Code that does `new mssql.Request(dataSource)` (or `new sql.Request(dataSource)`) then runs through the real `mssql` package's internals, which expect an actual driver-backed pool — not this compatibility shim. Under Postgres this construction is expected to fail. Separately, `mssql`-specific type tokens (e.g. `mssql.UniqueIdentifier`) have no Postgres equivalent.

## Known instances

- `packages/MJServer/src/resolvers/ReportResolver.ts` (`CreateReportFromConversationDetailID`) — `new mssql.Request(dataSource)`
- `packages/MJServer/src/resolvers/AdhocQueryResolver.ts` (`runSqlWithDeadline`) — `new sql.Request(ds).query<T>(sqlText)`
- `packages/MJServer/src/resolvers/GetDataResolver.ts` — `new sql.Request(readOnlyDataSource)`, notably even after already computing a dialect (`platform = getDbType()`) and rendering dialect-aware SQL text via `compositionEngine.ResolveComposition(query, platform, ...)` — the dialect-awareness is applied to SQL *generation* but not to *execution*
[...]

PR #3590, opened 2026-08-07
fix(codegen): IS-A inheritance silently does nothing on PostgreSQL

MS-BC:
## Summary

IS-A (Table-Per-Type) inheritance **has never worked on the PostgreSQL provider**. Two defects on consecutive steps of the same path, both caught-and-logged, so **CodeGen exits 0** while the entire IS-A wiring is skipped.

What you get instead of an error: an entity that registers and queries perfectly normally, but with `Entity.ParentID` NULL, no mirrored parent fields, no parent JOIN in its base view, and a `Save()` that never writes the parent row. There is no signal in the exit code, and the only trace is two lines in a 9,000-line log.

## The defects

**1. `processISARelationshipConfig` cannot look up either entity.**

The lookup used `(@SchemaName IS NULL OR SchemaName = @SchemaName)`. That parameter's only unambiguous use is `IS NULL`, so PostgreSQL has no type to infer and refuses to prepare the statement. Reproduced standalone, no CodeGen involved:

```sql
PREPARE t AS SELECT "ID" FROM __mj."vwEntities"
  WHERE "Name"=$1 OR ("BaseTable"=$1 AND ($2 IS NULL OR "SchemaName"=$2));
ERROR:  could not determine data type of parameter $2
```

Every `ISARelationships` entry in `additionalSchemaInfo` was skipped with a logged error. Fixed by composing the schema predicate **conditionally** — emitted, with its parameter, only when a schema was declared — rather than by adding a cast, so it stays dialect-neutral.

**2. `manageSingleEntityParentFields` throws before mirroring a single field.**

Its `EntityField` read and update reference mixed-case columns unquoted (`SELECT ID, IsVirtual, Type, Length, …`), which PostgreSQL folds to lower case: `column "length" does not exist`. All identifiers now go through `qi()`. Worth noting the INSERT arm of the same method already quoted correctly — only the SELECT and UPDATE arms were missed, which is why this reads as an oversight rather than a design choice.

## Why this survived

SQL Server is unaffected by both — it infers the parameter type from the other side of the `OR`, and resolves identifiers case-insensitively. The only shipping consumer of IS-A (`bizapps-sales`) authors T-SQL first against a local SQL Server dev database and converts to PostgreSQL later, so the IS-A path has only ever been exercised on the dialect where both bugs are invisible.

## Verification

Live PostgreSQL database, MJ 5.51 + BizAppsCommon installed, two IS-A children declared against its `People` / `Organizations` entities via `ISARelationships` config.

**Before:** `ParentID` NULL on both children, zero mirrored fields, two logged errors, exit 0.

**After:**
- `> IS-A: Set "Contacts" ParentID to "MJ_BizApps_Common: People"` — set from config, no hand-written UPDATE
- 19 mirrored virtual fields on the child
- base view generated with `JOIN __mj_bizappscommon."Person" __mj_isa_p1 ON c."ID" = __mj_isa_p1."ID"`
- `spCreate/spUpdate/spDelete` generated across the chain
- one `Save()` through the entity layer produced **one parent row + one child row sharing an ID**, with the parent's generated `DisplayName` populated, and `RunView` on the child returned parent and child fields together

Also: the pre-existing field-collision guard behaves correctly once reached — a child column shadowing a parent field aborts the sync for that entity, as designed.

## Tests

New `isa-postgres-portability.test.ts` (8 tests) composes the SQL through **real** `PostgreSQLDialect` and `SQLServerDialect` instances and asserts the contracts rather than the implementation: no parameter is ever emitted whose only use is `@p IS NULL`; every placeholder in the composed SQL is bound; the Name-then-BaseTable match order is preserved; each dialect's row limit is honoured; every mixed-case `EntityField` identifier survives quoting with its case intact.

Full CodeGenLib suite: **711 passed, 60 skipped**. `tsc --noEmit` clean. The 15 existing IS-A validation tests are untouched and pass.

## Note on backporting

Both defects are present on `lts/5` (5.51.0) as well, which is where this was found — a v5 deployment is what's blocked. A companion PR against `lts/5` follows.

---

The verification above covers CodeGen output and a single `Save()`. This section records a
later, deeper run against a **running MemberJunction stack** — API, GraphQL transport, and
the Angular client — with every assertion listed.

**Environment.** This PR merged with #3697 and #3703 onto `next` in a scratch integration
branch, so IS-A is exercised alongside the quoting and runtime-parity work rather than in
isolation. All **271** workspace packages build. Live PostgreSQL (`MJ_PG_AUDIT_pgparity`,
377 entities); MJAPI boots against it and serves; MJExplorer built and served against that
API. Fixture: `isatest.Person` → `isatest.Contact`, with the parent deliberately carrying
columns named `Length` and `Values` so this doubles as a live check of #3697.

### CREATE — 19/19

| # | Check | Result |
|---|---|---|
| 1–3 | child/parent entities resolve; `child.ParentID` → parent | pass |
| 4–6 | 4 parent fields mirrored as virtual fields; `Length` and `Values` among them | pass |
| 7 | one `Save()` returned true | pass |
| 8–9 | exactly **one** parent row and **one** child row added | pass |
| 10 | parent and child share a single ID | pass |
| 11–14 | parent field, both collider columns, and child field all persisted | pass |
| 15–19 | `RunView` returns parent + child fields together, colliders included | pass |

### The other three verbs, and the JOIN in every position — 28/28

CREATE is one verb out of four, and it only ever exercises the generated base view's JOIN
in the SELECT list. This suite covers what that leaves untested:

| # | Check | Result |
|---|---|---|
| 1 | seeded 3 IS-A records | pass |
| 2–5 | `Load()` rehydrates child field, mirrored parent field, and collider | pass |
| 6–7 | **`WHERE` on a mirrored PARENT field** returns exactly the right row — the JOIN is reachable from the predicate, not just the projection | pass |
| 8–11 | `WHERE` on collider `Values`; compound predicate over `Length` + `Values` | pass |
| 12–15 | **`ORDER BY` on a mirrored parent field** sorts correctly (`Aiden,Mira,Zara`); `ORDER BY` on collider `Length` sorts correctly (`300,200,100`) | pass |
| 16 | `Save()` on an existing IS-A record | pass |
| 17 | child column updated **in the child table** | pass |
| 18 | **parent column updated in the PARENT table** — writing through the child really reaches the parent row | pass |
| 19 | collider column updated on the parent | pass |
| 20 | UPDATE created no extra rows (10→10 / 10→10) | pass |
| 21 | `Delete()` through the child succeeded | pass |
| 22 | child row removed | pass |
| 23 | **PARENT row removed too — no orphan left behind** | pass |
| 24 | **zero orphaned parent rows anywhere in the table** | pass |
| 25–27 | pagination honoured over the IS-A JOIN; row accounting correct after the delete | pass |
| 28 | teardown left no orphans either | pass |

Checks 23–24 are the ones worth dwelling on. An IS-A implementation that creates two rows
and deletes one leaks a parent row on **every** delete, and a create-only test would never
notice. It does not leak.

### Through the GraphQL transport

Driven with `GraphQLDataProvider`, the exact client MJExplorer uses: `child.ParentID`
resolves, all four parent fields arrive as mirrored virtual fields, and a `RunView` over
HTTP returned the child field (`Title`), both mirrored parent fields, and both collider
columns correctly quoted and typed.

### One thing this run did *not* prove, stated plainly

A `Save()` of an IS-A record *over GraphQL* could not be exercised. MJ builds per-entity
mutation resolvers for schemas CodeGen has processed, and the ad-hoc `isatest` schema was
created directly in the test database, so no `CreateisatestContact` mutation exists. That
is a property of the fixture, not of this change or of PostgreSQL — the read path over the
same transport works completely, and the write path is covered exhaustively at the entity
layer above, which is where this PR's fix actually lives.

### A note on test hygiene

Two early runs of the deep suite aborted mid-way and left rows behind; an unscoped
`FirstName = 'Zara'` predicate then matched three records and reported a false failure. The
filter was correct and the fixture was dirty. Assertions are now scoped to a per-run tag,
and the leaked rows were cleaned up — recorded here because a test that fails for fixture
reasons is exactly as misleading as one that passes for them.

---

## Write atomicity — the half-failure case, 11/11

Every test above drives IS-A writes that succeed. A table-per-type `Save()` is **two**
INSERTs against two tables, so the case that actually decides whether the design is sound is
the half-failure:

```
INSERT parent  → OK
INSERT child   → rejected by the database
???  is the parent row rolled back, or left behind as an orphan forever?
```

An implementation that does not wrap the pair in a transaction leaks a parent row on *every*
failed child write. Nothing in a happy-path suite can see that, and on a live system it
accumulates silently.

**Forcing the failure honestly.** MJ's validation layer rejects writes it knows are invalid
*before* any SQL runs — that path proves nothing here, because no parent row would ever be
written. The failure has to come from the database, after the parent INSERT. So the fixture
added a CHECK constraint that exists only in PostgreSQL and not in MJ's metadata:

```sql
ALTER TABLE isatest."Contact"
  ADD CONSTRAINT ck_isa_atomicity_probe CHECK ("Title" <> 'FORCE_CHILD_FAILURE');
```

MJ cannot pre-empt what it does not know about, so it composes both INSERTs, the parent
lands, and the child is rejected by the engine. The scenario is reproduced, not simulated.

| # | Check | Result |
|---|---|---|
| 1 | probe constraint installed | pass |
| 2 | **control:** the database really rejects the sentinel title | fails, as required |
| 3 | `Save()` did not report success | pass — returned `false` |
| 4 | no child row created (1 → 1) | pass |
| 5 | **no PARENT row left behind (1 → 1) — the pair is atomic** | pass |
| 6 | orphan count unchanged (0 → 0) | pass |
| 7–8 | `RunView` on the IS-A entity, and on an unrelated core entity, both still work | pass |
| 9–10 | a valid IS-A `Save()` succeeds immediately afterwards, producing exactly one parent + one child | pass |
| 11 | final orphan count zero | pass |

Checks 7–8 cover a failure mode with **no SQL Server equivalent**: on PostgreSQL a failed
statement inside a transaction poisons it, and every subsequent command on that connection
is rejected until rollback. If the provider did not clean up, the first failed IS-A write
would break every later query on that pooled connection — surfacing in production as random,
unrelated breakage. It cleans up.

The probe constraint was dropped afterwards and the fixture verified back to its prior state.

## Unit tests

`@memberjunction/codegen-lib`: **792 passed, 66 skipped, 0 failures** — including the 15
existing IS-A validation tests and the 8 new `isa-postgres-portability` tests. Across the
full stack this PR sits on, **2,909 unit tests pass with zero failures**.


rkihm-BC (approved):
**Approving.** This is a high-quality, correctly-diagnosed, well-scoped fix. I traced both defects to their root cause against the source and verified the fixes are correct *and* complete (no sibling instances left unfixed). Two non-blocking follow-ups below.

## What's good

**Both root-cause diagnoses are precisely right — I verified the mechanisms independently.**

- **Bug #1 (parameter typing):** `runQueryWithParams` sends genuinely parameterized SQL to `pg`, and a param whose only appearance is `$n IS NULL` gives the planner nothing to infer → `could not determine data type of parameter $2`. Choosing **conditional composition over a `::text` cast** is the correct call — a cast is PG-only syntax and would have reintroduced the dialect split.
- **Bug #2 (identifier folding):** confirmed via `quoteSQLForExecution`'s tokenizer (`processWord`), which auto-quotes PascalCase identifiers *unless* they're in `_SQL_KEYWORDS` — and `LENGTH` is in that set. So `Length` passed through unquoted, folded to lowercase, and threw `column "length" does not exist` before any field mirrored. Exactly as described.

**Completeness — I checked for sibling instances:**
- The `@Param IS NULL OR` anti-pattern (bug #1's class) appears **only** at the two lines this PR fixes. Every other `vwEntities` lookup binds a concrete `SchemaName`, so none hit the inference failure.
- The buggy SELECT pattern appears only in the one method fixed; the UPDATE arm is fixed in the same hunk and the INSERT arm already quoted correctly.

**Good engineering:** extracting `buildISAEntityLookupSQL` removes real duplication and turns the bug-prone composition into a directly testable unit. The bug #1 tests (`never emits a bare "@param IS NULL"`, `every parameter placeholder is bound`) are genuine regression guards composed through **real** dialect instances. The inline comments explaining the silent-exit-0 failure mode and why SQL Server masked both bugs are excellent.

## Follow-ups requested (non-blocking)

**1. The bug #2 tests are largely tautological — they don't guard the actual fix.** The identifier-quoting block asserts `PostgreSQLDialect.QuoteIdentifier('Length') === '"Length"'`, but that primitive was never broken — the bug was that `manageSingleEntityParentFields` didn't *call* `qi()`. If someone reverted the `qi()` wrapping in that method, these tests would still pass. Unlike bug #1 (made testable via extraction), bug #2 stayed inline so its test can only assert the primitive. A stronger test would compose the method's actual SELECT/UPDATE and assert that after `quoteSQLForExecution`, `Length` survives as `"Length"`. Could you either add that, or file it as tracked follow-up?

**2. The real story is a systemic fragility worth an audit.** The auto-quoting tokenizer is a *denylist heuristic*: any query — now or future — that references a column colliding with `_SQL_KEYWORDS` (`Length`, `Position`, `Comment`, `Level`, `Language`, `Action`, `Order`, `Key`, …) breaks silently on PostgreSQL and works on SQL Server. This PR correctly patches the two IS-A instances, but the lesson embedded in the fix — *always quote via `qi()` rather than trust the tokenizer* — deserves a broader audit across `manage-metadata.ts` / the codegen SQL paths, ideally with a lint or guard. Given the downstream-customer stakes, please open a tracking issue.

## Minor

- The PR framing slightly overstates bug #2: of `ID, IsVirtual, Type, Length, Precision, Scale, …`, only `Length` actually collides with the keyword set — the rest were already being auto-quoted. The statement failed *as a whole* on `Length`, so the observable effect (nothing mirrors) is accurate, and quoting all of them is the correct robust fix. Just calibrating the claim.
- Verification is manual/narrative rather than automated — thorough and credible, but nothing in CI guards the *end-to-end* IS-A-on-PG pipeline. An integration-tier test would be the durable guard (candidate for the same follow-up as #1).

Correct root causes, dialect-neutral fixes, complete coverage of both bug classes within the IS-A path, and a clean DRY extraction. 👍

rkihm-BC:
Filed the systemic follow-up from my review as #3604 — audit of CodeGen's PostgreSQL identifier auto-quoting denylist (`_SQL_KEYWORDS`) for other silent keyword-collision failures, plus the stronger `quoteSQLForExecution`-based regression test for the `Length` fix. Not blocking this PR.

MS-BC:
---

## End-to-end verification on live PostgreSQL

Run against a real PostgreSQL 17 database (`MJ_PG_AUDIT_pgparity`, 761 `__mj` tables) on the branch stacked with #3697 and #3703. The IS-A scenario was purpose-built, because the deterministic tier does not cover IS-A — every entity in that database had `ParentID` NULL before this.

**Fixture.** Two tables sharing a primary key, `isatest.Person` ← `isatest.Contact` (`Contact.ID` REFERENCES `Person.ID`), declared through `additionalSchemaInfo`:

```json
{ "ChildEntity": "Contact", "ParentEntity": "Person", "SchemaName": "isatest" }
```

Both names are **BaseTable** names, not entity names, so resolution goes through the `BaseTable = @p AND SchemaName = @p` branch — precisely the predicate this PR rewrote. The parent deliberately carries columns named **`Length`** and **`Values`**, two of the eight identifiers that collide with SQL keywords, so the same run also exercises the quoting path.

**CodeGen result:**

```
> IS-A: Set "Contacts" ParentID to "People"
> Set ParentID on 1 IS-A child entity from config
```

Verified in the database afterwards:

- `Contacts.ParentID` → the `People` entity
- four parent fields mirrored as virtual, including `Length` and `Values`
- child base view generated with the parent JOIN and both colliders quoted:
  ```sql
  SELECT c."ID", c."Title", c."__mj_CreatedAt", c."__mj_UpdatedAt",
         __mj_isa_p1."FirstName", __mj_isa_p1."LastName",
         __mj_isa_p1."Length", __mj_isa_p1."Values"
    FROM isatest."Contact" c
    JOIN isatest."Person" __mj_isa_p1 ON c."ID" = __mj_isa_p1."ID";
  ```
- `spCreate/spUpdate/spDelete` generated for both entities

**Behavioural test through the entity layer — 19/19 checks passed.** One `Save()` on the child produced exactly one parent row and one child row **sharing one ID**, with parent fields, both collider columns and the child's own field all persisted, and `RunView` returning parent and child fields together.

### Two findings from the run, neither introduced by this PR

**1. IS-A needs two CodeGen passes on a fresh setup.** `processISARelationshipConfig`'s own comment states it *"Must run … BEFORE `manageParentEntityFields()`"*. It does not — mirroring runs at log line 103, `ParentID` is set at line 9782. So the first pass sets `ParentID` after mirroring has already looked for children and found none; the fields materialise on the second pass. The fix in this PR is correct and works, but "declare IS-A, run CodeGen once" does not yield a working child. Worth its own issue.

**2. CodeGen could not create the bucket Application for a new schema on PostgreSQL** — `column "id" of relation "Application" does not exist`. Root cause is an auto-quoter bypass: `LogSQLAndExecute` calls `ds.query()` directly and never routes through `qsql()`, unlike `runQuery`/`runQueryWithParams`, so all 61 of its call sites reach PG exactly as authored. Fixed in #3703 by quoting that column list; the architectural gap (61 unprotected call sites) is the CodeGen twin of #3427 and needs its own issue.

PR #3592, opened 2026-08-07
fix(codegen): IS-A inheritance silently does nothing on PostgreSQL (v5 LTS backport)

MS-BC:
v5 LTS backport of #3590 — clean cherry-pick, no conflicts, both defects are byte-identical on this line.

**This is the branch where the problem was actually found and where it is actually blocking.** A v5.51.0 PostgreSQL deployment (AIDP) is trying to ship `crm.Contact` / `crm.Account` as IS-A children of BizAppsCommon's `People` / `Organizations`, and cannot: `ISARelationships` config is a no-op on PostgreSQL, so `Entity.ParentID` stays NULL, nothing mirrors, and CodeGen reports success.

## The two defects

1. **`processISARelationshipConfig` cannot look up either entity.** `(@SchemaName IS NULL OR SchemaName = @SchemaName)` gives PostgreSQL no type to infer for that parameter, so the statement never prepares — `could not determine data type of parameter $2`. Every `ISARelationships` entry is skipped with a caught-and-logged error. Now composed conditionally instead of cast, so it stays dialect-neutral.
2. **`manageSingleEntityParentFields` throws before mirroring a field.** Unquoted mixed-case `EntityField` columns fold to lower case on PostgreSQL — `column "length" does not exist`. All identifiers now go through `qi()`.

Both are swallowed, so **CodeGen exits 0** either way. SQL Server is unaffected by both, which is why this went unnoticed.

## Verification on this line

Live PostgreSQL, MJ **5.51.0** + BizAppsCommon, two IS-A children declared via config. Before: `ParentID` NULL on both, zero mirrored fields, exit 0. After: `ParentID` set from config, 19 mirrored virtual fields, base view generated with the parent JOIN, CRUD sprocs across the chain, and a single `Save()` producing a parent row and child row sharing an ID with `DisplayName` populated.

CodeGenLib suite on `lts/5`: **650 passed, 60 skipped**. `tsc --noEmit` clean.

See #3590 for the full write-up, the standalone `PREPARE` reproduction, and the reasoning behind the conditional-predicate fix.

### DIFF
--- a/.changeset/codegen-isa-postgres-portability.md
+++ b/.changeset/codegen-isa-postgres-portability.md
@@ -1,0 +1,13 @@
+---
+"@memberjunction/codegen-lib": patch
+---
+
+Fix two PostgreSQL-only defects that made IS-A (Table-Per-Type) inheritance silently do nothing on PostgreSQL. Both were caught-and-logged, so CodeGen ran to completion with a zero exit code while every declared IS-A relationship was skipped — the end state was an entity that registered and queried normally but had `Entity.ParentID` NULL, no mirrored parent fields, no parent JOIN in its base view, and a `Save()` that never wrote the parent row.
+
+1. **`processISARelationshipConfig` could not look up entities at all.** The lookup used `(@SchemaName IS NULL OR SchemaName = @SchemaName)`; the parameter's only unambiguous use is `IS NULL`, so PostgreSQL had no type to infer and failed to prepare the statement — `could not determine data type of parameter $2`. Every `ISARelationships` entry in `additionalSchemaInfo` was therefore skipped with a logged error. The predicate is now composed conditionally (emitted, with its parameter, only when a schema is declared), which is portable to both dialects without a cast.
+
+2. **`manageSingleEntityParentFields` threw before mirroring any field.** Its `EntityField` read and update referenced mixed-case columns unquoted — `SELECT ID, IsVirtual, Type, Length, …` — which PostgreSQL folds to lower case: `column "length" does not exist`. All identifiers now go through `qi()`.
+
+SQL Server was unaffected by both (it infers the parameter type from the other side of the `OR`, and resolves identifiers case-insensitively), which is why this survived: the only shipping consumer of IS-A authors T-SQL first and had exercised it exclusively on SQL Server.
+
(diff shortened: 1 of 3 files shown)


rkihm-BC (approved):
**Approving** — verified faithful backport of #3590.

I diff'd this patch against #3590's: the added/removed content lines are **byte-identical**; the only differences are blob hashes and hunk offsets (the lts/5 file sits ~9 lines earlier). Textbook clean cherry-pick.

I also confirmed it lands on compatible infrastructure — a backport can apply cleanly yet hit missing helpers, so I checked the `lts/5` tree directly:

- All helpers the fix relies on exist on `lts/5`: `dialect`, `qi()`, `qs()`, `selectTop()`, `runQueryWithParams()`, and the `quoteSQLForExecution` delegation.
- **Bug #2 root cause is identical here** — `_SQL_KEYWORDS` contains `'LENGTH'` and `processWord` is the same denylist tokenizer, so the `qi()` fix is both necessary and correct on this line.
- `@memberjunction/sql-dialect` (5.51.0) is a declared dependency and `PostgreSQLDialect`/`SQLServerDialect` are exported, so the new test's real-dialect imports resolve.

Both root-cause diagnoses, the completeness of the two-bug coverage, and the dialect-neutral conditional-composition approach all carry over verbatim from my #3590 review. The `650 passed` suite count is expected (fewer tests on the LTS line than `next`).

The two non-blocking follow-ups from #3590 apply equally here — tracked in #3604 (systemic keyword-collision audit + the stronger `quoteSQLForExecution`-based regression test for the `Length` fix). No separate LTS issue needed. Given this is where the AIDP v5.51.0 PostgreSQL deployment is actually blocked, good to merge. 👍

MS-BC:
Closing for now — backport scope for the v5 LTS line is being decided as a batch rather than per-PR.

Context preserved for whenever this is revisited:

- The branch `fix/codegen-isa-postgres-lts5` is untouched, so reopening is a one-click operation.
- This had an **approval** on it. If it is reopened it should be re-reviewed rather than merged on the strength of that approval, because the forward PR has since diverged.
- **#3590 has moved on.** It now also fixes a third instance of the same defect in `manageSingleVirtualEntityField`'s UPDATE arm, and adds a guard test asserting the underlying property. That work is not in this branch. A future backport should cherry-pick #3590 in full rather than resurrecting this diff as-is.
- The root cause was also refined while verifying #3590: it is not "unquoted mixed-case identifiers fold to lowercase" but a **keyword collision** — `quoteSQLForExecution` auto-quotes PascalCase words except those on a hardcoded skip-list, and that list contains `LENGTH`, which is also a real column on `__mj."EntityField"`. Five MJ columns collide the same way: `Length`, `Count`, `Format`, `Date`, `Left`, `Right`.

Note for whoever is tracking it: BlueCypress/AIDP#37 is blocked on this fix reaching the v5 line.

PR #3601, opened 2026-08-07
Feat/generic field-path resolution for vector provider directives + Pinecone namespace hardening

dray-mc:
### Summary

After reorganizing some structure for the Betty use case, we've run into a setup that we would like to support but currently would give trouble for namespace segmentation. In our previous setup we had OrganizationID living on ContentItem records and that could be passed down through provider configuration/directives to Pinecone for use as namespaces (which we want to use for significant benefits to cost, performance, and deliberate security separation of content). With a better approach to scoping and how we set up organizations, we'd rather have that OrganizationID at the ContentSource level in a betty.ContentSource IS-A entity - this would easily mark sources as owned by an organization while entitlements for scoping can be applied on top to grant access to those sources to specific roles, potentially even across other organizations.

For this to work, we've added support for the provider configuration on the vector index to allow pulling context from related entities. Previously it was set up as {"namespaceField": "OrganizationID"} because OrganizationID was on the ContentItem, now it will be {"namespaceField": "ContentSourceID.OrganizationID"} and the framework has the ability to follow linked fields to other records. It also needed support for IS-A entities to be downcast.

The implementation of following the dot notation is generic within the FieldPathResolver that has been created, but this lives specifically within the autotagging package for now. This may be a candidate for a higher level implementation so similar functionality could apply to other examples of these JSON configuration fields, but that touches a much wider area than desired for this PR.

Also caught some related issues with triggering the autotagger action (now accepting object and string variations of 1 and true), implementing namespace support for purging vectors from Pinecone, and, if namespace is configured, failing to embed an item if the namespace field is not provided to avoid it just being dropped in the default namespace silently.

- Add `FieldPathResolver` (`@memberjunction/content-autotagging`): resolves a single-hop dotted field path (`'ContentSourceID.OrganizationID'`) rooted at a batch of `BaseEntity` records — the first segment is validated as a foreign key on the root entity via `EntityInfo` metadata, the related record is loaded, and if the related entity is an IS-A parent type, its child-type row (shared PK) is loaded and merged over it. Batched (one `IN (...)` load per entity per pass, not per record), per-pass cached, and consults `BaseEngineRegistry.TryGetCachedRecords` before querying so a hop through an already-cached entity (e.g. `ContentSource` via `KnowledgeHubMetadataEngine`) costs zero queries.
- Add `VectorDBBase.GetSourceRecordFieldPaths(providerConfig)`: a generic seam where a vector-DB driver declares which source-record field paths (plain or dotted) its config references. Default: none. `AutotagBaseEngine` resolves whatever paths a driver declares and hands `BuildProviderDirectives` an enriched record — the engine has zero knowledge of what a path's value *means* (namespace, shard key, routing region, ...); that's entirely provider-owned.
- `BuildProviderDirectives` may now throw to reject a record. `AutotagBaseEngine.precomputeProviderDirectives` converts a throw into a per-record failure (item/chunk marked `Failed`, purge chunks left `Pending`) — never poisoning the rest of a batch — across all three call sites: live vectorization, the `EmbedPendingChunks` backfill path, and `PurgeDeletedChunks`.
- Pinecone (`ai-vectors-pinecone`): declares `namespaceField` via the new seam and now **throws** in `BuildProviderDirectives` when it's configured but the record has no usable value — previously this returned `{}`, which silently routed the vector into the index's default namespace, breaching the tenant wall namespacing exists to build.
- Fix: `PineconeDatabase.DeleteRecord`/`DeleteRecords` now route deletes through the per-record `providerTemporaryDirectives.namespace` (grouping by namespace, mirroring `CreateRecords`) instead of always deleting against the default namespace — previously a namespaced vector's delete would silently no-op (Pinecone reports success deleting IDs that don't exist in a namespace). Also fixes a latent bug where both methods called `index.deleteOne`/`index.deleteMany` without `await`, so a failed delete could report success.
- Fix: `AutotagAndVectorizeContentAction`'s five boolean params (`Autotag`, `Vectorize`, `ForceReprocess`, `Purge`, `EmbedPendingChunks`) now accept `"1"`/`"true"`/`"yes"` string forms via a `flagIsSet` helper. The generic `RunAction` GraphQL mutation types every param value as a string, so a caller passing `Value: "1"` previously failed the action's strict `=== 1` checks and the phase was silently skipped with no error.
- No schema/migration changes — all four packages ship as `patch`.

### Why

Driven by a use case that embeds `ContentItem`/`ContentItemChunk` records directly (pre-populated, bypassing the standard autotag ingestion) and needs per-customer Pinecone namespace isolation, where the owning-organization fact lives on an IS-A extension of `ContentSource` rather than on core MJ metadata. The initial design put namespace-specific logic and a fail-closed policy directly in the content pipeline; per review, that got pushed down into the Pinecone driver (namespaces are a Pinecone concept, not a generic MJ one) and the path-resolution machinery was generalized to accept any `BaseEntity` batch rather than being ContentItem-specific — narrower `FieldPathResolver`/downcast primitives are candidates for promotion to MJCore later, once a second consumer (`ContentSource.Configuration`, `EntityDocument`, etc.) needs the same thing.

### Details

- `packages/AI/Vectors/Database/src/generic/VectorDBBase.ts` — new `GetSourceRecordFieldPaths()` (default: `[]`); `BuildProviderDirectives` TSDoc updated to document the throw-to-reject contract.
- `packages/AI/Vectors/Providers/Pinecone/src/models/PineconeDatabase.ts` — `GetSourceRecordFieldPaths()` declares `namespaceField`; `BuildProviderDirectives` throws on an unresolvable configured namespace; `DeleteRecord`/`DeleteRecords` namespace-route + `await` fix.
- `packages/ContentAutotagging/src/Engine/generic/FieldPathResolver.ts` — new file, the resolver described above.
- `packages/ContentAutotagging/src/Engine/generic/AutotagBaseEngine.ts` — `buildProviderDirectives` enriches the source record with resolved paths; new `resolveDriverFieldPaths`/`precomputeProviderDirectives` helpers; all three vectorization/backfill/purge call sites thread precomputed directives through and handle per-record rejection.
- `packages/Actions/ContentAutotag/src/generic/content-autotag-and-vectorize.action.ts` — `flagIsSet()` static helper; all five flag params switched from `=== 1` to it.

### Changed Packages

| Package | Change |
|---|---|
| `@memberjunction/ai-vectordb` | New `VectorDBBase.GetSourceRecordFieldPaths()` seam (default empty); documents the reject-via-throw contract on `BuildProviderDirectives` |
| `@memberjunction/ai-vectors-pinecone` | Declares `namespaceField` via the new seam; `BuildProviderDirectives` fail-closed (throws) on an unresolvable configured namespace; `DeleteRecord`/`DeleteRecords` namespace-aware + missing-`await` fix |
| `@memberjunction/content-autotagging` | New `FieldPathResolver` (generic single-hop FK path + IS-A downcast, registry-cache-aware); engine enrichment/rejection wiring across vectorize/backfill/purge |
| `@memberjunction/actions-content-autotag` | `AutotagAndVectorizeContentAction` flag params accept string forms (`flagIsSet`) |

### Testing

- `@memberjunction/ai-vectordb`: 35 passed (no new tests — interface-only addition with a default no-op implementation).
- `@memberjunction/ai-vectors-pinecone`: 44 passed (12 new) — `GetSourceRecordFieldPaths` declaration, `BuildProviderDirectives` throw-on-missing/blank-value, dotted-path key lookup, and full `DeleteRecord`/`DeleteRecords` namespace-grouping + fallback + error-propagation coverage.
- `@memberjunction/content-autotagging`: 212 passed (18 new) — `FieldPathResolver`: plain field, single-hop resolution, IS-A child-wins-over-parent merge, per-item batching, `BaseEngineRegistry` cache reuse (and fallback when cached rows aren't entity objects), misconfiguration (depth >1, non-FK, unknown field) resolving to nothing.
- `@memberjunction/actions-content-autotag`: 36 passed (5 new) — string `"1"`/`"True"` accepted, `"0"`/`"false"`/arbitrary strings treated as unset.
- All four packages build clean (`pnpm run build`).

### To Verify

- `FieldPathResolver` and the IS-A downcast it does are deliberately still scoped to `content-autotagging`, not promoted to MJCore — flagged as a two-primitive proposal (generic config-referenced field paths; a standalone IS-A downcast helper) for the team to weigh in on before anyone generalizes further, rather than solving it unilaterally in this PR.
- Path depth is capped at one hop and only single-valued FK hops are traversed (never 1:N) — a deeper or non-FK path logs and resolves to nothing rather than erroring, so misconfiguration currently fails the same way as "field doesn't exist"; worth deciding if that should be louder.
- Ran into two things on `next` while working this branch that are **not related to this PR** and don't need action here: (1) `IT71 - Task Graph Orchestration`'s metadata record isn't joined to the integration suite (added in `bf030852ff`, already visible failing in CI for others); (2) the monorepo's npm→pnpm cutover landed mid-branch — caught and cleaned up locally, no residue in this diff.
- Didn't add a migration or metadata for a real ISA `ContentSource` extension entity as part of this — that's expected to be built downstream (Betty-side, per the earlier conversation), this PR only makes the framework able to resolve through one once it exists.
- If this gets expanded to a more generic solution, should revisit what is currently implemented in the GetSourceRecordFieldPaths method added to the VectorDBBase and overridden by the PineconeDatabase. This method is used to identify what attributes in the open-ended JSON block are actually field names - 'namespaceField' in this case - instead of just any literal value that might also be provided and therefore not need value resolution. Letting the driver control what can be passed in via their provider configuration makes sense, but also seems like there could be more structure around this if we wanted it to require less explicit provider-specific code.


AN-BC (approved):
LGTM, nice improvement. Glad you're using `IsA` very powerful feature!

PR #3611, opened 2026-08-07
test(core-entities): type-level tests pinning generated related-record collection types

AN-BC:
## Why

CodeGen emits a related-record collection's element type as a generic:

```typescript
public readonly Params = this.DeclareRelatedRecords<MJActionParamEntity>({ … });
```

That generic **is** the entire type story. `Items`, `Create()`, `Add()`, `Remove()` and iteration are all parameterized on it. If the generator ever emits the wrong symbol, or `BaseEntity`, or drops the declaration, **every runtime test still passes** — the code runs, it just stops being typed. Callers silently lose IntelliSense and compile-time checking, and the first symptom is a field typo shipping.

That is not hypothetical. The emitter's first version used `RelatedEntityClassName` verbatim (`MJActionParam`) rather than the generated class (`MJActionParamEntity`). Thirty fixture-based generator tests were green, because the fixture pre-suffixed its input the way real metadata never does. It surfaced only when CodeGen ran against a live database — as a 100k-line file that could not compile, for all eight core collections at once.

## What this adds

**15 type-level assertions** over four collections, chosen so a single wrong shared type cannot satisfy them:

| Collection | Axis it covers |
|---|---|
| `MJActionEntity.Params` | cache-sourced, read-only, lazy |
| `MJAIAgentEntity.Prompts` | database-sourced, writable, sequenced |
| `MJAIAgentEntity.SubAgents` | **self-referential** — element type is the parent's own type |
| `MJAPIKeyEntity.Scopes` | a third entity entirely |

They pin:

- the declared type is `RelatedRecordCollection<TheSpecificEntity>`
- it never degrades to `RelatedRecordCollection<BaseEntity>` or `any`
- `Items` is **`readonly`**, not a mutable array — load-bearing, since a mutable one would let a caller push around the removal tracking, FK stamping and sequence renumbering
- `Create()` resolves to the element type; `Add()` accepts and returns it; `Remove()` takes it or an index
- iteration yields the element type, so `for…of` and spread stay typed
- value-list unions survive rather than widening to `string`

Enables vitest's `typecheck` for `*.test-d.ts`. Ordinary vitest transpiles without checking types, so a `.test.ts` could not catch any of this.

## The part that actually matters

`tsconfig.typecheck.json` exists because pointing vitest at the package's build tsconfig produced an **empty program** — that config excludes `src/__tests__`, so tsc had nothing to check and all 15 assertions passed vacuously.

I caught it by deliberately widening `Params` to `BaseEntity` and observing the suite stay green. With the corrected program, three assertions fail exactly as they should:

```
× declares the collection as RelatedRecordCollection of the SPECIFIC related entity
× never degrades to BaseEntity — the regression this file exists to catch
× Items is a readonly array of the element type
```

**A type test that cannot fail is worse than no test**, because it advertises coverage that does not exist. This was verified in both directions before shipping, and the reasoning is recorded in the config comments so nobody "simplifies" it back to the build tsconfig.

## Also

Fixes a stale value-list in `.claude/rules/data-access.md`, which still showed `Load` as `'eager' | 'explicit' | 'never'` in its code sample after the rename to `'immediate'` and the addition of `'lazy'`.

## Verification

- `packages/MJCoreEntities`: **588 tests passed, 28 files, no type errors** (the type suite reports under a `TS` badge)
- Generated `entity_subclasses.ts` confirmed byte-identical to `next` after the regression simulation — nothing left mutated
- `check:claude-md` passes

Follow-up to #3585.

### DIFF
--- a/.claude/rules/data-access.md
+++ b/.claude/rules/data-access.md
@@ -269,7 +269,8 @@ public readonly Lines = this.DeclareRelatedRecords<OrderLineEntity>({
     RelatedEntity: 'MJ_BizApps_Orders: Order Lines',
     RelatedEntityJoinField: 'OrderHeaderID',
     OrderBy: 'LineNumber ASC',
-    Load: 'explicit',                        // 'eager' | 'explicit' | 'never'
+    Load: 'explicit',                        // 'explicit' | 'immediate' | 'lazy' | 'never'
+    Source: 'database',                      // or 'cache' — read from a loaded BaseEngine
     OnRemove: 'delete',                      // 'delete' | 'orphan' | 'refuse'
     Sequence: { Field: 'LineNumber', From: 1 },
 });
--- a/packages/MJCoreEntities/src/__tests__/relatedRecordCollections.test-d.ts
+++ b/packages/MJCoreEntities/src/__tests__/relatedRecordCollections.test-d.ts
@@ -1,0 +1,133 @@
+/**
+ * Type-level tests for CodeGen-emitted related-record collections.
+ *
+ * ## Why these exist
+ *
+ * CodeGen emits a collection's element type from `EntityRelationship.RelatedEntityClassName`:
+ *
+ * ```typescript
+ * public readonly Params = this.DeclareRelatedRecords<MJActionParamEntity>({ … });
+ * ```
+ *
+ * That generic is the entire type story. Everything a caller touches — `Items`, `Create()`, `Add()`,
+ * iteration, `Removed` — is parameterized on it. If the generator ever emits the wrong symbol, or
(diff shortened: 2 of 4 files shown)


cadam11:
Heads up: this fixed the sibling-parity red (thanks) but next's backstop is still red with a new cause — MJGlobal's UUIDCompliance sweep flags 4 direct UUID comparisons in the new entity-graph.checks.ts (lines 113, 117, 163, 329). Mechanical fix per the test's own recipe; PR incoming, link to follow.

cadam11:
Fix PR: #3614

PR #3615, opened 2026-08-08
Revert: Query & Entity Materialization (#3311, #3365) for rework

rkihm-BC:
## Revert Query & Entity Materialization (#3311, #3365) for rework

Reverts the two merged materialization PRs from `next` so the next build is clean while the feature is reworked. Reverting is the safe mechanism here (both were squash-merges → one commit each): these are two `git revert` commits, **not** a history rewrite.

- Reverts #3311 (Phase 1) — commit `4c14b3f`
- Reverts #3365 (Phase 2) — commit `469862f`

`79 files changed, ~82 insertions(+), ~18,775 deletions(-)`. All materialization migrations (SQL Server + PostgreSQL) are removed.

> **Draft on purpose.** CI build will be **red** until CodeGen is re-run (see step 1 below). CodeGen is not required to *open* this PR — only to make it mergeable. Mark ready once the build is green.

---

### ⚠️ Step 1 — required before this can merge: re-run CodeGen

Six **CodeGen-generated** files conflicted on revert (later PRs regenerated them), so they were left at current-`next` state rather than hand-edited — deliberately, because `MJServer/src/generated/generated.ts` carries the materialization GraphQL resolver across thousands of lines, and the class-registration manifests also contain `MaterializeListFromViewAction`, an **unrelated** feature that must be preserved. Hand-surgery would risk collateral damage.

Run `mj codegen` against a DB reverted to this branch's schema and commit the regenerated:
- `packages/MJServer/src/generated/generated.ts`
- `packages/Angular/Bootstrap/src/generated/mj-class-registrations.ts`
- `packages/Angular/BootstrapLite/src/generated/mj-class-registrations.ts`
- `packages/ServerBootstrap/src/generated/mj-class-registrations.ts`
- `packages/ServerBootstrapLite/src/generated/mj-class-registrations.ts`
- `packages/Angular/Explorer/core-entity-forms/src/lib/generated/generated-forms.module.ts`

CodeGen will drop the `MJMaterializedResult*` registrations while keeping `MaterializeListFromViewAction` and the task-graph/workflow entries.

---

### Step 2 — rework (new branch + new PR; originals can't be reopened after a squash-merge)

Reference code is preserved on `feat/query-materialization-rework`. When rebuilding the feature:

1. **One consolidated migration**, not six. Ship a single `V<timestamp>__v6.1.x__…` file in `migrations/v6/` with a **correct, final `CREATE TABLE`** (all columns/constraints present from the start — `RowFilter`, `KeyColumns`, `SourceRowCount`, `RefreshesSinceFullRebuild`, `ReadFilterSpec`, `ForceFullRebuildCadence`). Run it on a clean DB from the latest published release, run CodeGen, and **concatenate the CodeGen emit into the same file** after 50+ blank lines and a comment banner — do not ship a separate `…CodeGen_Generated_Objects` migration or the create→alter→codegen→alter chain that shipped here.

2. **Version label is `v6.1.x`**, not `v6.2.x`. Per the edge-release convention: the release pins major/minor at 6.1 (`.edge-x` increments); migrations keep the legacy naming but stay at `v6.1.x` until LTS.

3. **Break the circular FK** `MJ: Materialized Results ↔ MJ: Queries` (CodeGen flags this cycle). The foundation migration created FKs in **both** directions — `FK_MaterializedResult_SourceQuery` (`MaterializedResult.SourceQueryID → Query`) and `FK_Query_MaterializedResult` (`Query.MaterializedResultID → MaterializedResult`) — which is the cycle.

   **Recommended (Amith):** introduce a dedicated join table **`__mj.MaterializedResultQuery`** and carry the relationship as rows there:
   ```
   MaterializedResultQuery (
       ID                  UNIQUEIDENTIFIER PK,
       MaterializedResultID UNIQUEIDENTIFIER FK → MaterializedResult(ID),
       QueryID              UNIQUEIDENTIFIER FK → Query(ID),
       + standard __mj system columns
   )
   ```
   Both FKs point **outward** to the parent tables, so there is no cycle. Drop the direct `Query.MaterializedResultID` and `MaterializedResult.SourceQueryID` columns/FKs entirely and resolve the link through this table. This is the cleaner normalized model and the direction we want.

   _Alternative (only if a join table feels heavy for a strict 1:1 link):_ keep `MaterializedResult.SourceQueryID → Query` and demote `Query.MaterializedResultID` to a plain `UNIQUEIDENTIFIER NULL` **soft reference** (no FK) — matching the existing `MLModelScoringBinding.MaterializedResultID` precedent (*"not yet a FK because that table is not merged"*). Either approach breaks the cycle; **Amith's join table is preferred.**

4. **Re-apply the review findings** already posted on #3311/#3365: PG swap-atomicity regression (Phase 2 split the transactional swap), RunQuery cache path ignoring `DataSource`, minted-entity permissions re-scoped only on first mint, and mint-time DDL identifier escaping.

---

_This revert + handoff was prepared with a detailed **Claude Code** review. If any of the rework guidance is unclear or you disagree with a call, please reach out before acting on it._


rkihm-BC:
Superseded by #3617 (Amith), which already removed the bulk of the materialization feature (package, SS migrations, entity, driver). Closing this full-revert in favor of a focused follow-up cleanup PR off current `next` that removes the remnants #3617 missed — the 6 PostgreSQL migrations (which still create the MaterializedResult table + circular FK on PG), the dangling `@memberjunction/materialization` dependency in Scheduling/engine, the CodeGenLib materialization source, the scheduled-job-type metadata, and the inert DataSource read-path plumbing.

PR #3616, opened 2026-08-08
fix(core): post-merge review fixes for entity companions, related-record collections & unified transaction scope (#3585)

AN-BC:
# Post-merge review fixes for PR #3585

PR #3585 was reviewed after merge with an adversarial multi-agent pass (15 verification/assessment agents over the merged diff, guided by the review-attention comment on the PR). Every confirmed finding is fixed here.

> **Updated after merging #3617** (task-graph Phase 8 + "temporarily kill materialization"): two of the original findings were resolved upstream and are **no longer part of this PR** — see the "Superseded by #3617" section. The PR now touches only `@memberjunction/core` and `@memberjunction/sqlserver-dataprovider` (plus docs, tests, and the IT72 config).

## Superseded by #3617 (dropped from this PR at the merge)

- **`ReadFilterSpec` restoration** — the original finding was that #3585's stale CodeGen run deleted the field from the generated artifacts while its migration still existed. #3617 then *removed the Phase-2 materialization migrations entirely* ("being rebuilt in another PR") and regenerated the generated files against a database without them. Restoring the field would now advertise a column no migration creates — the same mismatch inverted — so this branch adopts #3617's generated files wholesale. The rebuild PR owns re-introducing the column and its generated artifacts together.
- **Collection metadata seeding** — this branch originally hand-seeded the four missing `EntityRelationship.RelatedRecordCollection` configs and reconciled the drifted ones. #3617 landed the same seeding done properly: all eight collections in lockstep with the generated declarations, **with authoritative `sync` blocks from a real `mj sync push`**. This branch adopts that version. The finding itself was real — the compile time-bomb existed from #3585's merge until #3617 — it just got fixed upstream first.

## 🟠 Transaction correctness

- **Scope leaks closed** (`baseEntity.ts`): the scope opened by an IS-A initiator is now settled on every exit path — the clean-chain early return (`return true` with a physical transaction left open forever), and the provider-`Delete()`-returns-false path, which is how *every* ordinary delete failure surfaces (RLS denial, FK violation, zero rows). A leaked scope silently made all subsequent "committed" writes non-durable.
- **Commit-on-false closed**: when the provider's `Save` returns falsy data (a validate-type entity action rejecting the save), the scope now rolls back instead of committing the parent chain around a failed leaf.
- **Graph routing honors the save debounce**: two concurrent `Save()` calls on a composite now share one in-flight unit of work instead of double-inserting root and children.
- **TransactionGroup + companion graphs refused loudly**: the combination cannot be made coherent (children would persist against a parent whose write is deferred to `Submit()`), so it now throws with an actionable message instead of silently tearing atomicity.
- **`executeGraphLocal` keeps the no-throw contract**: scope acquisition moved inside `try`; rollback on the failure path guarded.
- **SQL Server doomed-transaction recovery**: when `ROLLBACK TO SAVEPOINT` fails because the whole transaction is doomed (deadlock victim, `XACT_ABORT`, Msg 3931), the provider now performs a full rollback and resets depth/stack/state instead of stranding a dead transaction that every subsequent statement silently joins.
- **`IsNested` is now truthful on SQL Server** via a new protected `CurrentTransactionDepth` (the provider deliberately leaves `IsInTransaction` false for `RunMaybeSerial`; the entity-transaction machinery needed a separate window onto real depth).
- **Out-of-order settle detection**: a scope now records the depth its begin produced and logs loudly when settled at a different depth — the concurrent-shared-provider interleaving hazard previously had no witness at all.
- **`RunInEntityTransaction`** preserves the caller's original error when the rollback itself throws.

## 🟠 RelatedRecordCollection

- **Read-only collections are now projections everywhere**, not just in `Dirty`/`ContributeSaveWork`: `Validate`/`ValidateAsync` skip them (no FK stamping onto an engine's shared instances, no blocking the parent on cached records the save will never write) and `Serialize` returns null (no shipping the donor engine's whole cached child set, no server-side re-load per cached row, no failing the save when a cached row was concurrently deleted). A verbose-mode breadcrumb logs when a read-only collection holds dirty items — the "my edit vanished" witness.
- **Clean, already-persisted children are skipped at plan level** — a header-only edit on a 50-line composite stays on the single-row path instead of becoming a 51-node graph (and, remotely, 50 shipped + re-loaded rows for zero writes). Safe because renumbered children are already dirty at plan-build time and removals contribute independently; `IgnoreDirtyState` forces the full write-out. `Serialize` is deliberately unchanged for writable collections.
- **`LoadRelatedRecords` hardened**: skips database loads for unsaved parents and for loaded/dirty collections (previously it wiped staged children on a new parent via a zero-row query + `SetLoadedItems([])`), and escapes the parent key exactly as `Load()` and the batch loader do.
- **Remote graph CREATEs labeled `create`** in result history and the `save` event (was hardcoded `'update'`).
- **`copyRecords`**: awaits `LoadFromData` (was a floating promise), clones `Date` values so copies don't alias the engine cache's Date instances, and verbose-warns when copying a dirty cached record.
- **Lazy hardening**: `lazy ⇒ cache ⇒ read-only` now enforced at declaration time (CodeGen already refused it; hand-written declarations failed later with a *wrong* message); the lazy-miss error now names both possible causes (engine not yet started vs nothing caches the entity — the registry cannot distinguish them, and the old message prescribed the wrong fix for the common bootstrap race); new **`TryItems()`** non-throwing probe for display-tier code, where `@if (x.Params && …)` cannot guard a throwing property read.

## Tests

| Suite | Result (re-verified after the #3617 merge) |
|---|---|
| MJCore unit | **1813/1813** (was 1775 — 38 new tests) |
| SQLServerDataProvider unit | **87/87** |
| integration-test-suite (bundle↔metadata drift, incl. IT72 sibling parity) | **155/155** |
| Workspace build (server + provider + test trees, 169 packages, incl. #3617's TaskGraph/Scheduling changes) | pass |
| `check:claude-md` | pass |

New coverage worth calling out:
- **`saveEntityGraphOperation.test.ts`** — the server half of every browser-initiated composite save had *zero* tests anywhere. Now covered: create path, update path with the silent-edit-drop guard (`IsNew:false` request-mode rehydration — load-then-apply so dirty tracking is real), unchanged-child no-op, missing root/child failures, removals (including the already-vanished-row skip), `Authorize`.
- **`baseEntity.transactionSettle.test.ts`** — the three scope-settlement regressions.
- Companions suite additions — debounce sharing, TransactionGroup refusal, plan-level clean-child skip + `IgnoreDirtyState`, read-only projection semantics, remote-create labeling, lazy declaration invariants, `TryItems`, `LoadRelatedRecords` guards.
- **IT72's eight mutation checks now bind in CI**: `runMutationTests: true` in the bundle's selector config (the escape hatch built for exactly this), justified because the bundle self-cleans prefixed, tagged fixtures in a throwaway CI database. EG1–EG8 stop passing vacuously.

**Pre-existing failure, not addressed here:** `packages/MJCoreEntities` `pnpm test` reports 36 typecheck errors in `ConversationEngine.test.ts` et al. — verified byte-identical on pristine `origin/next`; they were surfaced by #3611's typecheck wiring and predate this branch.

## Deliberately not in this PR

- **Event/cache side effects under savepoints** (BaseEngine caches, after-save actions fire at savepoint level and aren't retracted if an outer scope rolls back) — a design change needing its own discussion; the review flagged it as follow-up hardening.
- **True serialization of concurrent transactional units on a shared provider** — requires async-context ownership tracking; this PR ships loud detection instead, and MJServer's per-request providers remain immune.
- `plans/base-entity-composite-graph.md` still says `Load: 'eager'` — left as a historical design record.

https://claude.ai/code/session_01KuwU4Exd6GZV82nJMKETTv


AN-BC:
Reviewed this against the merged #3585 diff and verified the load-bearing findings independently rather than from the summary. **The review is right, and several findings are real bugs I shipped.** Pushed one commit (`82e36c1b3b`) with two follow-ups.

## Confirmed against `next` — these are correct and worth calling out

- **`Validate()`/`ValidateAsync()` stamping the FK onto read-only cache collections** is the worst of them, and it's mine. `stampParentKey()` does `child.Set(joinField, parentKey)`; I enforced the read-only contract in `Dirty` and `ContributeSaveWork` but not in `Validate`. Merely validating an Action would `Set()` on every `ActionParam` in `ActionEngineBase`'s cache — dirtying shared process-wide instances whenever the stored value differed, UUID casing being the everyday case. That is precisely the corruption the read-only contract exists to prevent, and I wrote the contract.
- **Floating `LoadFromData` promise in `copyRecords`** — it's `async` → `Promise<boolean>`; copies were pushed before they'd loaded.
- **Graph routing running before the `_pendingSave$` debounce** — single-row saves kept double-submit protection, composites lost it. Two concurrent `Save()` calls each built and executed a full graph.
- **`LoadRelatedRecords` wiping staged children** — no `IsSaved` guard, so a new parent's pre-generated UUID matched zero rows and `SetLoadedItems([])` discarded everything staged. Silent data loss in the method I added as the convenient entry point.
- **The delete-returns-false scope leak** is the one I'd rank highest among the transaction fixes, because it isn't exotic — the provider reports *essentially every* delete failure by returning false, so that's the common path, and a leaked scope makes subsequent "committed" writes non-durable.
- **`runMutationTests: true` on IT72** — right call. EG1–EG8 were passing vacuously in CI, which I should have caught when I wrote them.

## What I changed in `82e36c1b3b`

**1. `TryItems()` → `IsAvailable`.** The problem it solves is real — a template genuinely cannot guard a throwing getter, since the collection property is never null; it's the *read* that throws. But solving it with a second **reader** puts two accessors with different contracts side by side, and `null`-versus-`[]` is exactly the distinction call sites get wrong. That's the same shape as the silent-empty bug the throw exists to prevent, just relocated one level.

A predicate keeps one accessor and leaves nothing to misread:

```html
@if (action.Params.IsAvailable) {
  @for (p of action.Params.Items; track p.ID) { … }
}
```

`true` means the next `Items` read is safe **and already populated** — answering the question requires consulting the donor, and consulting it is what fills the collection. Never queries, never throws. Documented with an explicit line that business logic should `await` the engine's `Config()` rather than branch on availability, or the "not yet" branch quietly becomes the old bug wearing a guard.

**2. The API Key Scopes comment in `.related-record-collections.json` was false** — and it's mine, from #3585. It claimed declaring the collection switches on `MJAPIKeyScopeEntityServer`'s row-filter `ValidateAsync` for every scope on every API Key save, calling it "a security-relevant improvement". `Scopes` is cache-sourced → read-only, and this PR correctly makes `ValidateAsync` skip read-only collections, so that rollup does not happen. It was only ever *accidentally* true via the FK-stamping bug above. Replaced with what actually holds and where the check belongs instead. Worth fixing here rather than later, since this PR is what makes the claim definitively wrong.

MJCore: **1815/1815** (was 1813 — two added cases: the unavailable branch doesn't throw, and a non-lazy collection always reports available).

## Unrelated blocker you'll hit on a clean checkout

A fresh `pnpm install` on `next` **fails**: `packages/Scheduling/engine/package.json` still declares `"@memberjunction/materialization": "6.1.0-edge.0"`, but that workspace package was deleted when materialization was temporarily removed, and it isn't published to npm — so pnpm falls through to the registry and 404s. Existing trees are fine because their `node_modules` predate the removal; anyone cloning fresh, or creating a new worktree, cannot install. I had to build and test this change in a pre-existing worktree for exactly that reason. Not this PR's problem, but it will bite CI on any clean-install job.

## Otherwise

I'd approve. The findings are real, the fixes are proportionate, and the two things explicitly deferred — savepoint-scoped event/cache side effects, and true serialization of concurrent units on a shared provider — are correctly identified as design work rather than patched over.

AN-BC:
Agreed on `IsAvailable` — no push-back. I verified the implementation independently: the three `true` branches map one-to-one onto the paths where `Items` cannot throw (loaded-or-non-lazy, donor consulted **and thereby populated**, unsaved parent), so the "true means safe *and filled*" contract genuinely holds, and the single-accessor argument is right — `null`-vs-`[]` was the silent-empty bug relocated, as you said. Re-ran MJCore: 1815/1815.

Pushed `92cea425d` with two follow-ups in the same spirit:

1. **The "Run unit tests" CI failure on this PR is now fixed.** MJGlobal's repo-wide UUID compliance scan flagged seven direct `===`/`includes` UUID comparisons in `entity-graph.checks.ts` (IT72) and `entity-graph-client.checks.ts` (IT73). They ship on `next`, but this PR's root-level changeset invalidates the turbo cache, so this branch is the first to actually run the scan against them. Replaced with `UUIDsEqual` per the scan's own remedy; scan green, suite tests 158/158, build clean.
2. **The `MJ: AI Agents` `_comments` block had the same class of stale claim you fixed for Scopes** — it still described `OnRemove: 'delete'`/`'orphan'` semantics for Actions and SubAgents, both of which are now lazy/cache read-only views where removal is refused outright and no `OnRemove` policy applies. Aligned with what the configs actually say.

Noted on the fresh-install breakage (`Scheduling/engine` → deleted `@memberjunction/materialization`) — agreed it's outside this PR; flagging it to be fixed separately before a clean-install job hits it.

---

cadam11:
Question on this one: next's unit backstop is failing @memberjunction/actions#test since this morning's merges (BuildProvenance is threaded into RunAction, InvokeAction scope gate, GetActionParamsForAction — failing on next's tip run at 07:22Z). No morning merge touches packages/Actions directly, and this PR's entity-companions changes are the likeliest to shift interfaces the Actions engine tests consume. Could you take a look? Also noting next has a PostgreSQL migration parity failure at tip from the same window.

Issue #2708, opened 2026-05-28
Refactor monolithic BaseAgent into a layered inheritance hierarchy


AN-BC:
## Concept

The `BaseAgent` class in `@memberjunction/ai-agents` (`packages/AI/Agents/src/base-agent.ts`) has grown into a monolithic ~10k+ line single file. This issue captures the design to decompose it into a **clean, layered (linear) inheritance hierarchy** — one concern per layer — while preserving full backwards compatibility and avoiding the type-safety/runtime overheads of a composition/mixin approach.

### Why this design

A **linear inheritance chain** (rather than composition, mixins, or splitting into collaborator services) was chosen because:
- It preserves the existing public API of `BaseAgent` 1:1 — consumers and subclasses see no change.
- It keeps strong typing throughout with no `any`/casting overhead and no indirection at call sites (`this.foo()` still resolves directly).
- Each layer has a single, well-defined responsibility, so the file size per concern is manageable and the call graph stays readable.
- Subclasses can still override any method at any layer.

### Proposed layer hierarchy (bottom → top)

Files live under `packages/AI/Agents/src/base-agent/`:

1. **`BaseAgentState`** — Holds all instance variables, loggers, progress-step collectors, and state getters/setters. The root of the chain.
2. **`BaseAgentInit`** (extends `BaseAgentState`) — Coordinates initialization: engines verification, config loading, and starting-payload validation.
3. **`BaseAgentPrompt`** (extends `BaseAgentInit`) — Compiles hierarchical prompting payloads, executes LLM calls, estimates tokens, and performs context recovery / message compaction.
4. **`BaseAgentOperations`** (extends `BaseAgentPrompt`) — Handles loops (ForEach, While), sub-agent dispatch, client tool executions, and expression context-path resolution.
5. **`BaseAgentActions`** (extends `BaseAgentOperations`) — Intercepts action execution, extracts parameters/schemas, handles large binary/media replacement, and resolves media references.
6. **`BaseAgent`** (extends `BaseAgentActions`) — Orchestrates high-level execution, timeouts, and finalization. Remains the public entry point.

```
BaseAgentState
  └─ BaseAgentInit
       └─ BaseAgentPrompt
            └─ BaseAgentOperations
                 └─ BaseAgentActions
                      └─ BaseAgent   ← public class, unchanged API
```

### Scope of work (from the prototype)
- Create the new modular files under `packages/AI/Agents/src/base-agent/`.
[...]

PR #3617, opened 2026-08-08
Phase 8 — trigger mechanisms: live signal, a real reinvoke cap, missed-run policy, and job retirement

AN-BC:
Phase 8 of the unified workflow DAG engine program (plan: **#3456**) — the remaining Track D mechanisms, plus the observability decision that had been open across three review rounds.

## Live signal — dispatcher lifecycle frames

The choice was (a) claim-store post-CAS invalidation, (b) lifecycle frames, or (c) documented host polling. **(b)**, per AN-BC.

Frames are semantic rather than cache side effects: a viewer renders "step 3 of 7 running" from the event itself instead of re-reading Task rows and diffing them to infer what changed. `TaskGraphObserver` emits `TaskStarted` / `TaskCompleted` / `TaskFailed` / `TaskBlocked` / `TaskAwaitingHuman` / `GraphSettled`; MJServer publishes them on a new `taskGraphFrames(parentTaskId)` subscription.

**Addressed by `ParentTaskID`, deliberately not by session.** A durable graph outlives the tab that submitted it, and may be started by a schedule with no session at all. Keying on the graph means "watch this workflow run" works for whoever is permitted to see it, whenever they arrive — including after a page refresh, which a session-keyed push cannot survive.

Emit points sit where the fact is already true, not where it is convenient:
- `TaskStarted` **after** the claim is held — a frame for work another instance took is a lie a viewer cannot detect.
- `Task{Completed,Failed}` only when the guarded write actually landed — announcing an outcome we failed to persist would show a completion the database never recorded.
- `GraphSettled` **outside** the continuation's once-only CAS — a viewer should learn the run finished whether or not this instance wins delivery.

Delivery **fails closed**: a `parentTaskId` is discoverable, so without a connection-identity check anyone holding one could watch another user's workflow, per-step error messages included — the same hazard `statusUpdatesFilter` exists to close. Ownership rides on the frame, resolved once per graph and memoized, because the subscription filter runs per frame and synchronously and a database round trip there would make watching a run cost more than running it. It lives in the parent's durable metadata rather than a column: `Task.UserID` already means "the person this task waits on", so setting it on a parent would make every graph look like a human task.

## `MAX_REINVOKE_DEPTH` finally compares against a real number

Phase 3 shipped the cap; Phase 4 shipped the metadata carrying `reinvokeDepth`. The number was **permanently zero**: `Submit` reads it from its caller, `BaseAgent` never passed one, and a reinvoked agent had no way to know it *was* a continuation. Phase 7 therefore left `Reinvoke` unimplemented on purpose — a cap that never fires is worse than one continuation mode being unavailable, because the failure is an unbounded chain of real agent runs.

`AIAgentRun.ContinuationDepth` closes the loop: the deliverer stamps depth + 1 on the run it starts, `BaseAgent` passes its own run's depth into any graph that run submits, and the chain is bounded. Depth 0 means "not a continuation", so a chain begins at 1. `Reinvoke` degrades to posting whenever it cannot restart a turn (no submitting run — a schedule- or trigger-started graph has no turn to continue; run or agent unloadable) and never throws, since the dispatcher calls it inside the delivery CAS.

`continuationDepth` is deliberately distinct from `parentDepth`, which measures sub-agent nesting inside one turn. A continuation is a **new top-level turn** caused by work that already finished.

## Missed-run policy

`MissedRunPolicy` — `RunOnce` (default), `RunAll`, `Skip`.

**The default is not a preference, and I had it wrong at first.** I told AN-BC it would be `Skip` "because that matches today's behavior". It doesn't: `updateJobStatistics` calls `GetNextRunTime(cron, timezone)` with no `fromDate`, so the next run is computed from *now* — a job whose `NextRunAt` passed runs once immediately and then jumps forward. That is `RunOnce`. Defaulting to `Skip` would have silently stopped every existing job in every install from catching up after an outage.

`RunAll` is safe to offer because the drain is rate-limited by construction: its next run is computed from the occurrence just consumed, so a week-long outage walks one occurrence per poll tick instead of firing 168 jobs at once.

"Missed" is defined **cron-relatively** — a *later* occurrence has also come due — rather than against a grace window, which would misjudge both ends of the range: a per-minute job looks missed after a 90-second pause, and a monthly job looks on time a week late.

Two properties worth calling out, both found the hard way:
- The decision **fails open**. It can only ever withhold a run the schedule already said was due, so an unparseable cron *or a helper returning anything other than a date* lets the job through. The second case is not hypothetical — guarding the throw without guarding the return value produced 22 test failures on first run.
- It is **synchronous** on purpose. It runs immediately before lock acquisition, and making it async inserted a microtask exactly where ordering against the sweep's fire-and-forget orphan cleanup matters — which surfaced as a job that stopped dispatching because its queued lock result had already been consumed.

## One-shot scheduling needed no new schedule shape

AN-BC's instinct that Scheduled Jobs already covered this was right, and the investigation found something better than a new column.

`isJobDue` already honors `EndAt`, so "run once at T" is expressible today as *cron at T + `EndAt` just after T*. The real defect was that **`Status='Expired'` had been a declared value that nothing ever set**. Such a job stopped running on its own but stayed `Active` forever, permanently inert — and kept counting toward `UpdatePollingInterval`, so a one-shot with a tight cron made the *whole scheduler* poll faster forever for a job that would never run again.

`expireFinishedJobs` retires them. Deliberately narrow: only `Active`/`Pending` transition, because a `Paused` or `Disabled` job was put there by a person and quietly rewriting that would lose their decision; and only `EndAt` triggers it, never "the cron has no future occurrence" — a cron always has one, so inferring exhaustion would be guessing. It runs **after** the dispatch loop, since `isJobDue` already blocks an expired job, which makes retirement pure bookkeeping and keeps it out of the interleaving described above.

## Verification

| | |
|---|---|
| Build | **271/271** |
| Unit tier | green across every touched package — 388 ai-core-plus, 1824 ai-agents, 33 task-graph, 220 scheduling-engine, 813 MJServer |
| **IT71** | **18/18** with `RUN_MUTATION_TESTS=1` |

**On the deterministic suite:** 43/47 in a full run, with 4 failures — IT71, IT03, IT25, IT31. **All four pass standalone**, and none are caused by this branch. IT71's failure is the known transport-sequencing issue (#3251): it ran at slot 40, after client-transport bundles had rebound the process-global provider. The suite is also reporting 47 tests rather than the 57 it ran before the database rebuild, so suite membership looks incomplete after the re-seed — worth a look independent of this PR.

## Deliberately not here

**Trigger vocabulary normalization** (`OnChange | Schedule | OnDemand | External`) is deferred per AN-BC. Worth recording why it is not simply "do it later": the four surfaces the plan names answer *different* questions. `ProcessRun.TriggeredBy` is run provenance; entity-action invocation types (`Validate` / `Before*` / `After*`) are *when within the save*, a refinement of `OnChange` rather than a peer of it; Scheduled Jobs are `Schedule` by definition and need no column. Collapsing them into one family would be over-normalizing, so the scope question wants deciding before the migration is written.

**Transition filters** (`Field Changed` / `Field Changed To Value`) stay in Phase 8/Track A: `OldValues`/`NewValues` and those filters have zero occurrences on `next`, so it is genuinely substrate work rather than two metadata rows. Phase 7 removed the acute harm by refusing a filtered trigger outright instead of saving one that silently over-fires.

That leaves **Track R** (`BaseAgent` decomposition, #2708) as the last item in the program.


AN-BC:
@AN-BC — this is ready for review. Phase 8 closes every remaining Track D item, so with this merged the program is complete except for **Track R** (`BaseAgent` decomposition, #2708).

**Your four decisions, as built:**

1. **Live signal → (b) dispatcher lifecycle frames.** Built, with the subscription keyed on `ParentTaskID` rather than session so a durable graph stays watchable after a refresh and when it was started headlessly.
2. **Trigger vocabulary → deferred.** Recorded in the PR body *why* it wants a scope decision first, so it doesn't come back as an unexamined "just do it": the four surfaces answer different questions, and entity-action invocation types are a refinement of `OnChange` rather than a peer.
3. **One-shot → (A), no schema.** Your instinct was right, and the dig found the actual defect: `Status='Expired'` was declared years ago and **nothing ever set it**, so a one-shot stayed `Active` forever and kept the whole scheduler polling at its cadence. Retiring it is the fix.
4. **Missed-run → `Skip | RunOnce | RunAll`.** With one correction to what I told you: I said the default would be `Skip` "because that matches today's behavior". It doesn't — `updateJobStatistics` computes from *now*, so today's behavior is `RunOnce`. Defaulting to `Skip` would have silently stopped every existing job from catching up after an outage. The default is `RunOnce`.

**Two things to look at while testing:**

- The new subscription is `taskGraphFrames(parentTaskId)`. Frames only publish for graphs that recorded an owner, and the owner is written at submission time — so **a graph submitted before this branch will emit nothing**. You need a freshly submitted one to see anything.
- `continuation: 'reinvoke'` now actually reinvokes rather than degrading to a message. The cap is real for the first time, so if you want to watch it stop, `MAX_REINVOKE_DEPTH` is 5.

**One thing that isn't mine but showed up:** the deterministic tier reported **47** tests after the database rebuild, where it ran **57** before. Suite membership looks incomplete after the re-seed. Four tests failed in the full run (IT71, IT03, IT25, IT31) and **all four pass standalone** — IT71's is the known #3251 transport-sequencing ordering issue. Flagging in case the re-seed needs another metadata directory pushed.

PR #3624, opened 2026-08-08
IT74 — execute task graphs for real, and fix the three bugs it found

AN-BC:
**Stacked on #3617** — TX5 needs the `TaskGraphObserver` seam that PR introduces. Base it on `next` once #3617 merges and I'll rebase.

You asked how deep the live integration coverage went. The honest answer was: wide on persistence, **zero on execution**. IT71 has eighteen checks and not one runs a graph — nine assert metadata, nine verify the rows a save produces. Everything past "the rows are correct" — claim, execute, propagate, roll up, settle — was unit-tested against fixtures and never against SQL Server.

This closes that. And it found **three production bugs on its first runs**, one of which meant *no successful graph has ever settled*.

## How it works

A real `TaskGraphDispatcher` against the live database, with a stub `TaskAgentRunner` injected through the seam that exists for exactly this. No model calls, no tokens, no network — so it sits in the deterministic tier. What is under test is the dispatcher, not the agent.

| Check | Proves |
|---|---|
| **TX1** | submit → claim → execute → rollup, with the runner's output persisted |
| **TX2** | a diamond's AND-join **waited** — start order is recorded, because "everything reached Complete" is far weaker than "D ran after B and C" — and the join receives both dependency outputs |
| **TX3** | a failed step propagates to `Blocked` and the graph **settles** rather than hanging |
| **TX4** | a false edge condition makes its branch unreachable and the graph settles |
| **TX5** | lifecycle frames are emitted, each addressed to its `ParentTaskID` and carrying an `OwnerUserID` |
| **TX6** | **two dispatchers, one graph, every task exactly once** — the CAS guarantee *is* the database's atomicity, so no in-memory mock can test it |
| **TX7** | a claim orphaned by a crash is reclaimed and the graph finishes |

## The three bugs it found

**1. The dispatcher read its own work queue through a stale cache.** `TaskClaimStore` mutates task rows via direct SQL — correct, since a `BaseEntity.Save()` cannot express a guarded UPDATE. But direct DML fires no invalidation, and the discovery queries used `RunView` with no `BypassCache`. Completions written on the claim path stayed invisible, `loadGraphState` kept seeing `In Progress`, and graphs never rolled up. This is the exact case [the caching guide](guides/CACHING_AND_PUBSUB_GUIDE.md) names `BypassCache` for.

**2. A graph that succeeded could never settle.** `findActiveGraphIDs` selected graphs by non-terminal *children*. The moment the last child completed, the graph left that set — so the pass that would have rolled the parent up never saw it. **Every fully-successful graph stayed `In Progress` forever and its continuation never fired.**

A *failing* graph happened to survive, because blocking its dependents left them non-terminal for one more pass. That is why the bug hid behind a passing failure-path test, and it is a good argument for testing the success path even when the failure path looks harder.

**3. A not-taken branch ran instead of being skipped.** A definitely-false edge condition was resolved by *dropping* the edge. Removing a task's only prerequisite makes it eligible in the very next wave — so "this branch was not taken" **executed the branch**, potentially before the node that gated it. The code's own argument against dropping unevaluable edges ("a prerequisite silently disappears and the dependent task runs early — turning a typo into out-of-order execution") applies verbatim to the false case. The dependent is now recorded unreachable and blocked, and only when *every* route in was cut — a node reached by an alternate branch is still reachable.

Also hardened: `ComputeParentRollup` treats an empty child set as Complete-and-terminal — right for a childless graph, catastrophic for one whose reload came back empty transiently, since it would mark live work finished and fire its continuation. The outer guard covered the first load only.

## One small production change

`TaskGraphDispatcherConfig.PollIntervalSeconds` (default 5, behavior unchanged). The interval was hardcoded. Five seconds is right for production, where steps are agent runs measured in seconds to minutes — but it made a four-node graph take twenty seconds to observe, and the right value genuinely differs by host.

## The harness took three passes, and each failure was instructive

Worth recording, because each is a trap the next author will hit:

- **The check's own reads needed `BypassCache` too** — same root cause as bug 1. A stale read left the stub unable to identify which task should fail, so the failure check passed a *successful* graph, intermittently.
- **A dispatcher claims from the whole table, not from "its" graph.** Per-check stub runners therefore applied the wrong policy to a neighbour's task. One runner now serves the bundle — which is also what production looks like.
- **`Submit` creates claimable rows immediately**, so any policy registered afterwards has a window in which a draining dispatcher runs the task under the wrong one. The stub now resolves a task's name *where the work happens*, and TX3 declares its failure policy before the graph exists — closing the window rather than narrowing it.

## Verification

- **IT74: 7/7, six consecutive runs** with `RUN_MUTATION_TESTS=1`
- Unit tier green: `@memberjunction/task-graph` 33, `@memberjunction/integration-test-suite` 158
- Packages touched build clean

## One fix beyond the tests

**IT71 moves from sequence 34 to 33.** 34 is where client-transport tests begin, and a server-transport bundle at that boundary bootstraps against a provider a client bundle has already rebound — issue #3251. That tie is what made IT71 fail in the last full-suite run while passing standalone. IT74 is sequenced at 33 for the same reason.

## Not in this PR

The **client-transport** bundle (`TaskGraph.Submit`/`GetStatus`/`Cancel` over the wire, plus the `taskGraphFrames` subscription and its fail-closed filter) and the **live-model Demo Flow Agent** run. Both were in the proposal; both are parked in CI (one needs MJAPI, one needs tokens), so they buy less per unit of effort than this did. Worth noting the Demo Flow Agent cannot go in the deterministic tier as-is — its four steps call Get Stock Price, Get Weather and Google Custom Search.


AN-BC:
@AN-BC — built as proposed, and it earned its keep immediately.

**Headline: bug #2 means no successfully-completing task graph has ever settled.** `findActiveGraphIDs` picked graphs by non-terminal *children*, so when the last child completed the graph dropped out of the set before anything rolled the parent up. Parent stuck `In Progress` forever, continuation never fires. It hid because a *failing* graph survives — blocking its dependents keeps them non-terminal one extra pass — so the failure-path test passed while the success path was broken.

The other two: the dispatcher read its work queue through a cache its own direct-SQL claim protocol never invalidates, and a false edge condition **ran** its branch instead of skipping it (dropping the edge removes the dependent's only prerequisite, so it became eligible immediately).

**Three notes for review:**

1. **Stacked on #3617** — TX5 needs the `TaskGraphObserver` seam. Merge that first and I'll rebase, or merge this after and it's a clean fast-forward.
2. **I moved IT71 from sequence 34 → 33.** That's the actual fix for the failure I flagged in #3617: 34 is where client-transport tests start, and a server bundle tied at that boundary hits #3251. IT74 sits at 33 too. It touches `.integration-suite.json`, which the other agent is also editing for IT72/IT73 — small append + one field change, but worth knowing if it conflicts.
3. **I used IT74, not IT72/IT73** — those are taken by the entity-graph work.

**Deliberately not built** (both were in the proposal): the client-transport bundle and the live Demo Flow Agent run. Both are parked in CI, so they buy less than this did. Also worth flagging — the Demo Flow Agent **cannot** go in the deterministic tier as it stands: its four steps call Get Stock Price, Get Weather and Google Custom Search ×2. It needs either the live-model tier or a hermetic fixture flow. Happy to do either next.

IT74 is 7/7 across six consecutive runs.

Issue #4, opened 2024-01-24
Clicking on the home tab resets the home view, does not save context


JS-BC:
On the home page, click on data -> crm -> entities -> click on a view. Then click on the home tab. You will not be brought back to the page you were before switching tabs.

Issue #2, opened 2024-01-24
The scroll functionality for tabs sometimes do not work


JS-BC:
The scroll buttons should appear after the tabs exceed a certain length. However, they do not appear until the user resizes the screen.

JS-BC:
This issue is no longer relevant now that we've replaced the tab script with a new component

PR #3633, opened 2026-08-08
security: harden SQL-filter validator, OAuth handler, API-key lookup, and new-user domain gate

AN-BC:
Hardening from an adversarial security review of the server auth/data surface. Five isolated, low-blast-radius fixes.

## 1. CRITICAL — RunView SQL-filter validator bypass
`MJCore/src/generic/databaseProviderBase.ts` — `ValidateUserProvidedSQLClause`

The validator strips string literals **before** running its keyword denylist, using a regex with **backslash-escape** semantics. SQL Server / PostgreSQL (`standard_conforming_strings=on`) do **not** treat `\` as an escape, so:

```
FirstName = 'a\'; DROP TABLE Users; --'
```

is swallowed whole as one "literal", stripped to `FirstName = `, and **passes** the denylist — while the database terminates the literal at the real quote and executes the stacked statement. Reproduced against the live regex (`DROP TABLE`, `WAITFOR DELAY`).

**Fix:** replace the stripper with SQL-standard doubled-quote (`''`) literal matching, so injected terminators/keywords after a backslash are no longer hidden.

> ⚠️ **Residual, by design:** the denylist still permits **subqueries** in `ExtraFilter`/`OrderBy` (e.g. `(SELECT ... FROM __mj.[User] ...)`), which enables boolean-blind cross-table read by any authenticated RunView caller, bypassing entity permissions/RLS on the subquery target. This is inherent to the raw-SQL-fragment `ExtraFilter` model and is **not** closed here. The durable fix is to validate user filter/order-by clauses with an **AST allowlist** (MJ already ships `SQLExpressionValidator`, used for `Aggregates`) — recommended as a follow-up. Also unvalidated today: persisted `UserView.WhereClause` (interpolated with no denylist at all).

## 2. HIGH — OAuth callback SQL injection
`MJServer/src/rest/OAuthCallbackHandler.ts` — `loadConnectionConfig`

Caller-supplied `connectionId` (POST `/oauth/initiate` body) was interpolated into a raw `ExtraFilter` without escaping, while the sibling `loadAuthorizationState` escapes. **Fix:** escape single quotes at the sink **and** reject non-UUID `connectionId` at the request boundary.

## 3. MEDIUM — OAuth open redirect
`MJServer/src/rest/OAuthCallbackHandler.ts` — `redirectToSuccess` / `redirectToError`

`frontendReturnUrl` was redirected to after only a URL-parse check → open redirect from the trusted MJAPI origin. **Fix:** validate the return URL's origin against `cors.allowedOrigins` (and the built-in redirect origins); fall back to the default page otherwise.

## 4. MEDIUM — New-user auto-provisioning domain gate
`MJServer/src/auth/index.ts` — `verifyUserRecord`

"Limit new users to authorized domains" checked the **spoofable `Origin` header**, not the verified identity. A holder of any valid IdP token could auto-provision an account under an authorized domain by forging `Origin` on a non-browser request. **Fix:** authorize against the **email domain of the verified JWT**.

## 5. LOW — API-key hash lookup hardening
`APIKeys/Engine/src/APIKeyEngine.ts` — `ValidateKeyByHash`

Asserts the hash is a SHA-256 hex digest before building the SQL filter, enforcing the injection-safety invariant at the sink for all (incl. future) callers.

## Testing
This review ran in an ephemeral environment with no `node_modules`, so packages were **not** built/tested here. The validator fix was verified with a standalone reproduction of the regex. **Please let CI run the full build + unit/integration suites.**

---
_Generated by [Claude Code](https://claude.ai/code/session_01GDDvtpqgD1G1W7ixnF6D2M)_


AN-BC (commented):
LGTM. @rkihm-BC can you put eyes on this and if looks good at the 5.x LTS label and merge so it also backports

rkihm-BC (commented):
I fetched the PR branch, read every changed file in full context, traced the reachability of each sink, and empirically ran both the old and new regexes (plus the live built `SQLExpressionValidator`) to verify the claims rather than take them at face value.

## Verdict

The four fixes are correct and I'd approve their substance — but **I would not merge as-is**, primarily because the CRITICAL fix is applied to only one of the two places in this repo that carries the identical vulnerable regex, and the second one is reachable by the same attacker.

---

## What I verified holds up

I ran the old and new strippers against the denylist:

| Case | OLD | NEW |
|---|---|---|
| `FirstName = 'a\') ; DROP TABLE Users; --'` | **passes** ❌ | rejected ✅ |
| `ID = 'x\') ; WAITFOR DELAY '0:0:5'; --'` | **passes** ❌ | rejected ✅ |
| `Comments LIKE '%--%'` (legit) | passes ✅ | passes ✅ |
| `Name = 'O''Brien'` (legit) | passes ✅ | passes ✅ |
| `Name = 'abc; DROP TABLE t` (unterminated) | rejected ✅ | rejected ✅ |
| `"Name") ; DROP TABLE Users; --"` | rejected ✅ | rejected ✅ |

The CRITICAL is real (the parenthesis-closing variant above defeats the `WHERE (...)` wrapping), the fix closes it, and I found **no legitimate-clause regressions**. The new pattern also correctly implements doubled-quote semantics — I specifically checked that `"Name") ; DROP…` is *not* over-stripped (the match terminates at `"Name"`), which is the failure mode a naive rewrite would have introduced. Bonus: the new regex is linear, where the old one had a nested quantifier + lookahead (ReDoS shape).

Fixes #2, #3 and #5 are each correct at the sink. #5 is genuine defense-in-depth only — both real callers (`A2AServer`, `MCPServer`) pass a server-computed digest — which matches its LOW label.

---

## Blocking

### B1 — The same bypass is still live in `SQLExpressionValidator`, the class this PR recommends as the durable fix

[`SQLExpressionValidator.ts:264`](https://github.com/MemberJunction/MJ/blob/1111819e0166531fb9d0ce67064b13bf06067b35/packages/MJGlobal/src/SQLExpressionValidator.ts#L264) carries the **byte-identical vulnerable regex**, untouched by this PR. I ran the exploit against the built `dist`:

```
aggregate: plain stacked DROP        valid=false  trigger=comment
aggregate: backslash-hidden DROP     valid=true   ← SUM(A)+'\'; DROP TABLE Users; --'
aggregate: backslash-hidden WAITFOR  valid=true
full_query: backslash-hidden DROP    valid=true   ← SELECT 'a'+'\'; DROP TABLE Users; --' FROM T
```

`removeStringLiterals` runs at Step 1, *before* the dangerous-keyword scan and the function allowlist, so the whole payload is stripped and every downstream check sees `SUM(A)+`. The expression is then interpolated raw by `BuildAggregateSQL` (`agg.expression + ' AS ' + QuoteIdentifier(...)`), and `Aggregates` is an ordinary GraphQL input on `RunDynamicView` / `RunViewByID` / `RunViewByName` ([`RunViewResolver.ts:172`](https://github.com/MemberJunction/MJ/blob/1111819e0166531fb9d0ce67064b13bf06067b35/packages/MJServer/src/generic/RunViewResolver.ts#L172)). Same attacker class as #1: any authenticated user with read on any entity.

So the PR's own recommendation — "the durable fix is an AST allowlist; MJ already ships `SQLExpressionValidator`" — currently points at a class with the same hole. This is a one-line change using the exact replacement already written here, and it belongs in this PR.

### B2 — No regression tests

The fix's own comment says *"Do NOT reintroduce backslash-escape handling here."* A comment isn't an enforcement mechanism — and the fact that two copies of this regex drifted into the codebase is the proof. Per the Definition of Done:

- [`SQLExpressionValidator.security.test.ts`](https://github.com/MemberJunction/MJ/blob/1111819e0166531fb9d0ce67064b13bf06067b35/packages/MJGlobal/src/__tests__/SQLExpressionValidator.security.test.ts) already exists as an adversarial suite and is the natural home for B1's cases — it currently has no backslash test at all.
- MJCore needs a test for `ValidateUserProvidedSQLClause` (it's `protected`; a three-line test subclass covers it). The table above is the test matrix.

### B3 — Missing changeset

The bot flagged it. Three packages need bumps (`@memberjunction/core`, `@memberjunction/server`, `@memberjunction/api-keys`), and the changeset is also the only vehicle for the operator-facing note in S3 below — which matters more, not less, for the 5.x LTS backport.

---

## Should fix before merge

### S1 — Fix #4 is incomplete; the Origin header is still a precondition

[`auth/index.ts:218`](https://github.com/MemberJunction/MJ/blob/1111819e0166531fb9d0ce67064b13bf06067b35/packages/MJServer/src/auth/index.ts#L218) is unchanged:

```ts
(requestDomain || configInfo.userHandling.newUserLimitedToAuthorizedDomains === false)
```

When `newUserLimitedToAuthorizedDomains === true`, auto-provisioning still *requires* a truthy `requestDomain` — i.e. the spoofable `Origin` header. It's no longer what's **authorized against** (the fix is real), but it's still what **gates entry**. Net effect: an attacker just sends any `Origin` and satisfies it, while a legitimate non-browser client with an authorized email domain is still refused. The code now reads as contradicting the SECURITY comment three lines below it. Drop `requestDomain` from the precondition, or say why it stays.

### S2 — The denial log now prints the wrong variable

[`auth/index.ts:264`](https://github.com/MemberJunction/MJ/blob/1111819e0166531fb9d0ce67064b13bf06067b35/packages/MJServer/src/auth/index.ts#L264) still says `Request domain '${requestDomain}' does not match any of the domains in the newUserAuthorizedDomains setting`. The decision is now made on email domain. An operator debugging a failed provisioning will chase a value that had no bearing on the outcome. Log `emailDomain`.

### S3 — `newUserAuthorizedDomains` changes meaning; this is a config break shipping as a patch

The list was previously matched against `parseRequestHostname(req.headers.origin)` — a *hostname*. It is now matched against an *email domain*. A deployment configured with frontend origins (`app.company.com`, `localhost`, `*.company.com`) silently stops auto-provisioning on upgrade. Note especially that `*.company.com` compiles to `^.*\.company\.com$`, which does **not** match `company.com` — so even the "obvious" wildcard config breaks. Needs a changeset note plus doc updates in `packages/MJServer/src/config.ts`, `packages/MJCLI/mj.config.cjs`, and `docker/MJAPI/docker.config.cjs`.

### S4 — Username-only IdPs regress

[`context.ts:519`](https://github.com/MemberJunction/MJ/blob/1111819e0166531fb9d0ce67064b13bf06067b35/packages/MJServer/src/context.ts#L519): `email` is `userInfo.email ?? userInfo.preferredUsername`. Where the IdP emits no `email` claim, the value has no `@`, so `emailDomain === ''` and provisioning is now denied unconditionally where it previously worked. Denying is arguably the right call, but it should be a deliberate, logged decision rather than a side effect of `split('@')[1]`.

Related, worth a follow-up rather than blocking: the gate now rests entirely on the IdP's `email` claim. For IdPs that permit unverified or self-asserted emails, consider requiring `email_verified` when the claim is present.

### S5 — Reuse `IsValidUUID`; validate once and pass the validated value

[`OAuthCallbackHandler.ts:372`](https://github.com/MemberJunction/MJ/blob/1111819e0166531fb9d0ce67064b13bf06067b35/packages/MJServer/src/rest/OAuthCallbackHandler.ts#L372) uses `/^[0-9a-fA-F-]{36}$/`, which accepts 36 consecutive hyphens. `IsValidUUID` already exists in `@memberjunction/global` ([`UUIDUtils.ts:81`](https://github.com/MemberJunction/MJ/blob/1111819e0166531fb9d0ce67064b13bf06067b35/packages/MJGlobal/src/util/UUIDUtils.ts#L81)) and this file already imports `UUIDsEqual` from that package.

Also: the check runs on `String(connectionId)` but the **raw** `req.body` value is what's forwarded to `loadConnectionConfig`. A JSON body of `{"connectionId": ["<valid-uuid>"]}` passes the test (`String(['x']) === 'x'`), then hits `connectionId.replace(...)` on an array → `TypeError` → 500. Assign the coerced string and pass that.

---

## Non-blocking observations

- **The open-redirect fix is a no-op on default configs.** `cors.allowedOrigins` defaults to `['*']`, so most deployments get no protection. The comment is honest about it, and matching the CORS middleware's exact-string semantics (`index.ts:1227`) is the right consistency call — but the practical coverage is narrower than the PR body implies.
- **Severity of #3 is overstated slightly.** `frontendReturnUrl` reaches the redirect from stored auth state created by the *authenticated* `POST /oauth/initiate`. It's an authenticated open redirect, not an anonymous one. Also worth validating at `/oauth/initiate` so the caller gets a 400 immediately instead of a silent fallback after a full OAuth round-trip.
- **Double logging.** Throwing inside the `try` to reach the fallback emits both "origin not allowed" and then "Invalid frontend return URL" for the same event. Cosmetic.
- **The stated residual is confirmed.** `ID IN (SELECT UserID FROM __mj.[User] ...)` passes both old and new. I agree with the AST-allowlist recommendation — with the caveat that it depends on B1 being fixed first.
- **One more desync of the same class**, for completeness: `[a'] = 1) ; DROP TABLE t; --'` still passes, because SQL Server bracket identifiers make `'` ordinary. I do **not** think it's practically exploitable (the delimited identifier would have to name a real column containing a quote, or the batch fails to compile), but it's the same regex-vs-parser mismatch and further supports moving off the denylist.
- **Escaping-at-each-sink doesn't scale.** `git grep "ExtraFilter: \`"` in MJServer alone shows a mix of escaped and unescaped interpolations (`auth/exampleNewUserSubClass.ts:43` interpolates the JWT email unescaped, for one). Not this PR's job, but a shared `EscapeSQLLiteral` helper — or a typed filter API — is the structural answer.

---

**Recommendation:** land B1 + B2 + B3 and S1/S2/S5 in this PR (all small), and S3's doc/changeset note before the LTS label goes on. The rest can be a follow-up issue alongside the AST-allowlist work.

rkihm-BC:
@AN-BC — reviewed, and pushed fixes rather than merging as-is. Summary of what changed since your LGTM.

**TL;DR:** the CRITICAL fix was applied to one of the two places in the repo carrying that regex. The second one is reachable by the same attacker and is now fixed. There are also two behavior changes worth your eyes before the LTS label goes on, and the backport is not a straight cherry-pick.

---

### 1. The literal-stripper bypass was still live in `SQLExpressionValidator`

`packages/MJGlobal/src/SQLExpressionValidator.ts` had the byte-identical backslash-escape regex, untouched by the original commit. It's reachable by any authenticated user: `Aggregates` is an ordinary GraphQL input on `RunView`, and `BuildAggregateSQL` interpolates the expression raw. Confirmed against the built `dist`:

```
aggregate: backslash-hidden DROP     valid=true   ← SUM(A)+'\'; DROP TABLE Users; --'
full_query: backslash-hidden DROP    valid=true
```

`removeStringLiterals` runs at Step 1, before the keyword denylist, the function allowlist and the system-object denylist — so everything hidden in the fake literal was invisible to all of them.

On impact: RunView resolves through `GetReadOnlyProvider(..., { allowFallbackToReadWrite: true })` and `dbReadOnlyUsername` is optional, so on a deployment with no separate read-only login the aggregate query runs on the **read-write** pool and a stacked `DROP` executes. Where a read-only login is configured it degrades to arbitrary cross-entity read. Worst case lands on the default configuration.

Rather than patch the second copy, I removed the duplication that let them drift: `StripSQLStringLiterals` is now the single implementation in `@memberjunction/global`, called by both `SQLExpressionValidator` and `DatabaseProviderBase`. MJCore already depended on MJGlobal, so no new dependency.

**Validation.** I built a reference tokenizer implementing SQL Server / PostgreSQL literal rules and differentially fuzzed the shipped stripper against it, flagging any input the validator passes but the database would execute:

```
targeted: 14 cases, 1 bypass
fuzz:     400000 cases, 0 bypasses
```

The single hit is the known bracket-identifier differential (`[a'] = 1) ; DROP …`), which isn't exploitable — SQL Server must bind `[a']` to a real column, and no MJ column name contains an apostrophe, so the batch fails to compile before anything runs. It predates this PR. Closing it by also stripping `[...]` isn't free (PostgreSQL treats `[` as array subscripting), so it belongs with the AST-allowlist work rather than here. PG `E'...'` strings and dollar-quoting produce over-*rejection* — fail-safe, never a bypass.

### 2. The branch broke an existing test suite

`import { configInfo } from '../config.js'` runs `loadConfig()` at module scope, so `OAuthCallbackHandler.xss.test.ts` could no longer load (`Configuration validation failed`). This is what the unit-test job would have caught — it was still pending when the review started.

Fixed by injecting `allowedFrontendOrigins` through `OAuthCallbackHandlerOptions`, wired from `configInfo.cors?.allowedOrigins` at the single construction site. That removes the import-time coupling and makes the guard testable — 23 new tests now cover scheme/port/suffix confusion, `javascript:` URLs, the built-in-origin exemption and non-string body values.

### 3. Domain gate — the fix was correct but incomplete

The entry condition still required a truthy `requestDomain`, so `Origin` gated entry without authorizing anything: an attacker forges one trivially, while a legitimate non-browser client with an authorized email domain was refused. Removed. The denial log also still printed `requestDomain`, which would send anyone debugging after the wrong value; it now reports the email domain and carries the hostname→email-domain migration hint. Identities with no `@` at all (IdPs issuing a bare `preferred_username`) are now denied explicitly with a distinct message rather than as a side effect of `split('@')[1]`.

### 4. Two behavior changes to sign off on

**`newUserAuthorizedDomains` changes meaning** — from request hostname to verified email domain. The setting had no documented semantics anywhere (no schema comment, templates ship `[]`), and the only prose in the repo already described it as an email domain, so this aligns with intent. It's off by default. Deployments that enabled it *and* populated it with frontend hostnames need to swap in email domains; note `*.example.com` does not match `example.com`. Now documented on the zod schema, both config templates, and the changeset.

**MCP OAuth auto-provisioning widens** — this one I'd flag specifically. `MCPServer`'s `resolveOAuthUser` passes `requestDomain: undefined` ("not needed for MCP"), so with the domain gate on, MCP OAuth users could *never* be auto-created regardless of email domain. Removing the spoofable precondition changes that to "auto-created when the email domain matches", consistent with the browser path. I traced the chain before accepting it: `validateBearerToken` does JWKS signature + audience verification before `resolveOAuthUser`, and the proxy-token fast path verifies HS256 + iss + aud and passes no name claims, so it can't provision at all. Security-sound, but it's account-creation widening inside a security patch, so it should be a conscious call. @rkihm-BC has approved it; flagging it for you since it ships to LTS.

### 5. Also in

- Added the missing changeset (`patch` — no DB changes), carrying both behavior notes above
- `IsValidUUID` from `@memberjunction/global` replaces the hand-rolled `/^[0-9a-fA-F-]{36}$/` (which accepted 36 hyphens); the body value is coerced once and the coerced value flows downstream
- `frontendReturnUrl` is validated at `/oauth/initiate` so a caller gets a 400 rather than a silent fallback after a full round trip — with an absent value (`undefined`/`null`/`""`) explicitly treated as "no return URL" rather than an invalid one
- `allowedFrontendOrigins` is required, so a future construction site can't silently lose the protection. Safe to require: MJServer declares `"exports": { ".": "./dist/index.js" }` with no subpath wildcard and doesn't re-export the handler, so nothing outside the package can construct it

---

### Before merging

**The backport is not a straight cherry-pick.** `lts/5` carries all five vulnerable sites *plus* the reachable aggregate path (`BuildAggregateSQL` present, `Aggregates` exposed on `RunViewResolver`, `AdhocQueryResolver` present). Backporting the original commit alone would ship 5.51.x the half-fix — on the branch whose users are least likely to be watching. Line offsets differ (`databaseProviderBase` :985 vs :1075; the stripper :247 vs :264), and MJCore's change now depends on the MJGlobal export, so both packages must travel together. Also worth remembering the changesets rule inverts on a certified line — `patch` is the only correct level there, which is what this carries.

**CI will show 3 red MJServer tests that aren't ours.** `RealtimeBridgeResolver.test.ts` uses `vi.fn(() => ({...}))` as a constructor, which vitest 4 rejects (`"... is not a constructor"`). Pre-existing and unrelated — happy to open a separate issue.

**Verification:** turbo build 176/176 · MJGlobal 635/635 · MJCore 1953/1953 · APIKeys 205/205 · MJServer 855 pass (+ the 3 above) · `check:standards` and `check:changeset` pass.

**Still open, as you noted originally:** subqueries in `ExtraFilter` remain permitted, and `UserView.WhereClause` is unvalidated by design. The AST-allowlist follow-up is the durable fix for both — and it now has a validator worth building on.

PR #3634, opened 2026-08-08
fix(core-entities): clear the 36 typecheck errors failing `pnpm test` on next

AN-BC:
**`pnpm test` currently exits 1 on `next`.** Found while running the full build + retest AN-BC asked for. Not caused by any task-graph work — confirmed by running the tier against clean `next`, which reproduces all 36 errors.

All 588 MJCoreEntities tests **pass**. The failure is 36 `TypeCheckError`s reported as "Unhandled Source Error", which vitest counts as a failed run.

## Why now

`cf312b6181` (the related-record collection type-level tests) turned on vitest `typecheck` pointed at `tsconfig.typecheck.json`, which **includes `src/__tests__` for the first time**. Type sloppiness that had never been checked became visible all at once. Nothing was newly broken — it was newly *looked at*, which is the point of the change and worth keeping.

## What was wrong — five categories, five files

**1. Casts to `Record<string, unknown>` (18).** A `BaseEntity` subclass has no index signature, so the direct assertion is the mistake TypeScript actually reports. Routed through `unknown`, as the compiler suggests. In `UUIDCrossDbCompliance` the nine identical engine-cache seeds became one named helper, so the assertion lives in one place instead of nine.

**2. Constructing entity subclasses with no arguments (8).** `BaseEntity`'s constructor takes an `EntityInfo` these tests never use — they exercise getter/setter logic that touches no metadata, and the runtime has always been fine. A `bare()` helper states that intent rather than fabricating an `EntityInfo` the tests would then have to keep accurate.

> ⚠️ **Flagging rather than hiding:** the codebase's own rule is that entity subclasses come from `Metadata.GetEntityObject()`, **never `new`**. These call sites predate the typecheck. I left them structurally as they were — moving them onto the metadata path is a much larger change than making the type checker honest, and it's a call for whoever owns these tests. Happy to do it as a follow-up if you want it.

**3. Imports that don't exist.** `UserViewEntity` imported `ViewDisplayState` / `ViewTimelineState` from `MJUserViewEntityExtended`, which imports them under *local aliases* and does not re-export them — and the repo forbids adding re-exports to paper over that. Now imported from `../generated/entity_subclasses`, where they're declared. `ViewDisplayMode` went with them: declared nowhere, never referenced.

**4. A property that doesn't exist.** `MJMCPServerToolEntity` has no `Name` — the field is `ToolName`. The test seeded and asserted a property that isn't there, which is precisely the drift a typecheck exists to catch.

**5. Mock visibility.** `_mockPermissionProviderBase` returns a class declared inside a function, so TypeScript must name its type in the emitted declaration and can't do that for a type with non-public members. The four helpers are now `public`; providers extending the mock are unaffected, and the real `PermissionProviderBase` keeps its own visibility.

## Verification

| | |
|---|---|
| MJCoreEntities | 28 files / **588 tests, 0 errors, exit 0** |
| Full repo build | **271/271** |
| Full repo `pnpm test` | clean apart from an `ng-flow-editor` vitest **worker-pool timeout** — passes standalone (4 files / 21 tests); machine load, not code |
| Deterministic integration tier | **58/60**; both failures are 300s budget timeouts and both pass standalone (IT35 18.1s, IT02 29.6s / 12 checks) |

On that last row — the tier took **2237s** in that run versus ~90s in earlier runs, so the box was heavily loaded. IT02 in particular went 24s → 334s → 29.6s across runs. Neither failure is a defect.

**Also worth knowing:** the deterministic suite now registers **60 tests**, up from the 47 I flagged earlier. The lost suite memberships are back, so that concern is closed.


AN-BC:
@AN-BC @rkihm-bc — this one needs your eyes, since it touches test files you own.

**`pnpm test` is red on `next` right now.** I hit it running the full build + retest, and verified against clean `next` that it's not from any task-graph work — all 36 errors reproduce without my changes.

The cause is benign: `cf312b6181` pointed vitest's typecheck at a tsconfig that includes `src/__tests__` for the first time. That's a *good* change — it immediately caught a test asserting `MJMCPServerToolEntity.Name`, a field that doesn't exist (it's `ToolName`), and three imports of types that were never exported. It just also surfaced ~30 pre-existing casts at the same time.

**The one judgment call I did not make for you** is in category 2. Eight sites do `new MJTemplateEntityExtended()` / `new MJUserViewEntityExtended()` etc., which the codebase's own rule forbids — entity subclasses are supposed to come from `Metadata.GetEntityObject()`. I made them typecheck via a `bare()` helper that names the intent, and left the structure alone, because moving them onto the metadata path is a real refactor of your tests rather than a compile fix. Say the word and I'll do it as a follow-up.

**Two pieces of good news from the same run:**
- The deterministic suite is back to **60 tests** (I'd flagged it dropping to 47 after the DB rebuild) — that concern is closed.
- IT71's transport-sequencing collision didn't recur; both it and IT74 passed in-suite.

The only integration failures were two 300s budget timeouts (IT02, IT35), both of which pass standalone. The tier took 2237s in that run versus ~90s earlier, so the box was loaded — IT02 alone went 24s → 334s → 29.6s across three runs.

AN-BC (commented):
LGTM

PR #3640, opened 2026-08-08
feat(ai): ModelConfiguration catalog cascade + catalog-driven realtime turn detection

AN-BC:
# `ModelConfiguration` — one strongly-typed bag, three catalog levels, no more column-per-knob

> ✅ **Complete and verified locally.** The migration has run, both CodeGen passes and the metadata push are done, the CodeGen output is appended to the migration, all affected packages build, and unit + integration coverage is in place. Full record in [`plans/model-configuration.md`](../blob/claude/grok-voice-think-fast-2-f8hups/plans/model-configuration.md).

## What this is

A per-modality, strongly-typed JSON configuration bag on the AI model catalog, at **three levels** forming an inherit-with-override cascade (per-key deep merge, base first):

```
AIModelType.ModelConfiguration          (type-wide default — e.g. every Realtime model)
  <  AIModel.ModelConfiguration         (per-model)
    <  AIModelVendor.ModelConfiguration (per model-on-this-provider — the winner)
```

This generalizes a cascade the schema already expresses for scalars — `SupportsPrefill` / `PrefillFallbackText` are NOT NULL at the type and nullable (= inherit) at model and model-vendor — into a structured bag with one interface (`IAIModelConfiguration`: `LLM` / `Realtime` / `Vision` / `Audio` sections) shared by all three levels via MJ's JSONType mechanism, so CodeGen emits typed `ModelConfigurationObject` accessors on all three entities.

**Boundary rule** (documented on the interface): anything the engine filters/sorts/joins on stays a column; anything a driver consumes at session/call time goes in the bag. Existing capability columns are untouched — new knobs only.

## First consumer: realtime turn detection

Turn detection was hardcoded per provider profile (Grok always `server_vad`, OpenAI always the provider default) — smarter models had no way to opt into their smarter turn modes. Now `Realtime.TurnDetection` (`Mode: 'default' | 'serverVad' | 'semanticVad' | 'native'` + eagerness/threshold/silence tuning) flows catalog → session bag → wire on **both topologies**, with full precedence:

```
profile hard default
  < ModelConfiguration cascade (type < model < model-vendor)
  < realtime.session.turnDetection (agent/app config cascade — also new)
  < runtime configOverridesJson (raw audio.input.turn_detection — auth-gated, unchanged)
```

- `AIEngine.GetEffectiveModelConfiguration(modelID, modelVendorID)` is the one canonical cascade read path (pure `ParseModelConfiguration` + `ResolveEffectiveModelConfiguration` live in `@memberjunction/ai`); the model-vendor **row** id is threaded through both realtime resolution paths so the most-specific layer resolves precisely.
- Profiles declare `supportedTurnModes`; the shared `MapNormalizedTurnDetection` translates the normalized vocabulary. An unsupported mode is **diag-logged and falls back to the profile default** — a shared catalog never rejects a session. Non-protocol drivers scrub the key.
- **Meeting mode composes on top** of whatever mode wins, and `Reconfigure` now rebuilds the session's remembered mode — **fixing a latent bug** where a live meeting-mode flip silently downgraded any non-`server_vad` session to `server_vad`.
- **`'native'` is the forward slot**: when xAI documents a full-duplex turn mode for the Grok Voice Think Fast family, `XAI_REALTIME_PROFILE` adds `'native'` + one mapping, and the catalog opts models in via metadata — zero per-model driver work.

## Verification

| Gate | Result |
|---|---|
| Full monorepo build | **clean** — 299/299 and 271/271 |
| Unit tests (AI/Core, OpenAI, Agents, MJCoreEntities) | **2,840 passed, 0 failed** |
| Deterministic integration tier | **59/60** — sole failure proven pre-existing (below) |
| Assembled migration replayed on a clean DB | **passes** — 37 applied, objects verified |
| `check:claude-md`, `check-pg-migration-content` | **pass** |

New coverage: cascade parse/merge (tolerant + non-aliasing), `MapNormalizedTurnDetection` mode mapping and the never-reject fallback, the `turnDetection` scrub in `ExtractRealtimeFeatures`, `GetModelCatalogSessionSettings`, catalog < session < runtime bag precedence, and a regression test pinning the `Reconfigure` downgrade fix.

**CodeGen was run against a purpose-built database** (created empty, migrated from the v5.46 baseline) rather than the shared dev DB, so no unrelated in-flight schema could leak in. The generated diff is purely additive — 333 lines in `entity_subclasses.ts`, 27 in MJServer, 8 per AI-catalog form, **zero deletions**.

**The one integration failure is not from this change.** IT25 `view-execution.V10` (keyset pagination) fails, and it was verified rather than assumed: V10's universe is the `EntityField` rows of the first three entities by ID — `MJ: Explorer Navigation Items`, `MJ: Test Run Feedbacks`, `MJ: Generated Code Categories` — which is **byte-identical (32 rows, same 12 distinct `Sequence` values) in a DB with this change and one without**, and contains **zero** of the rows this PR adds. Root cause is visible in that data: 32 rows carry only 12 distinct `Sequence` values, so the view's `ORDER BY` ties while the keyset seek predicate compares `ID`, producing overlapping pages. That is a genuine tie-unsafe keyset bug, unrelated to this PR and worth its own issue.

## Two things the remote authoring could not have known

- **`AIEngine` is a server-side FACADE over `AIEngineBase`, not a subclass** — it delegates member by member, so `AIEngine.Instance.GetEffectiveModelConfiguration(...)` did not compile. Added the delegate, and documented the composition-over-inheritance rationale plus the "add a member here, proxy it there" rule on **both** classes. This is a standing trap for every future `AIEngineBase` member.
- **CodeGen re-renames five pre-existing `EntityField.DisplayName`s** on these entities (`AIModel.AIModelTypeID` / `.IsActive` / `.AIModelType`, `AIModelVendor.TypeID` / `.Type`) — its current naming computation asserting itself over values stored since the baseline, on rows it owns. Unrelated to this feature but **included**, because CodeGen re-emits them on every run over these entities, so stripping them only defers the identical diff to the next PR. Called out in the migration's CodeGen header block; the underlying instability is tracked in **#3647**.

## PostgreSQL: no counterpart, by policy

Authoring `.pg.sql` counterparts is **deterministic toolchain work the build engineer performs at release time** — not feature-PR work, and emphatically not something to infer with an LLM.

This PR removes `scripts/check-pg-migration-parity.mjs`, which had been failing feature PRs for exactly that (**already red on `next` with 6 unrelated gaps**) while instructing engineers to convert against a moving schema. `.github/workflows/pg-migrations.yml` no longer triggers on `migrations/**`; it now runs on `migrations-pg/**` and the converter toolchain — i.e. the release/build PR that carries the converted set — plus manual dispatch. The **CONTENT** check stays, because a counterpart that exists and is empty is a real defect at any time (a 126-byte `.pg.sql` against 12,041 lines of T-SQL shipped a silently broken v5.45).

Root `CLAUDE.md` and `migrations/CLAUDE.md` now carry the doctrine: no hand-authored PG SQL, no bespoke parity tooling — fix the converter or hand it to the build engineer.

## Seeds (the one behavior-affecting change)

GPT Realtime 2.1 / 2.1-mini are seeded to `semanticVad` (eagerness `auto`) at the model level. Everything else is behavior-neutral while `ModelConfiguration` is NULL. **Not yet verified against a live GA endpoint** — if `semantic_vad` is rejected there, delete the seed; the machinery stands on its own.

## Change map

| Area | Files |
|---|---|
| Migration (+ appended CodeGen output) | `migrations/v6/V202608081622__v6.1.x__ModelConfiguration_JSONType_Columns.sql` |
| JSONType source + bridges + seeds | `metadata/entities/JSONType-interfaces/IAIModelConfiguration.ts`, `metadata/entities/.entity-field-jsontype-model-configuration.json`, `metadata/ai-models/.ai-models.json` |
| Canonical types + pure resolver | `packages/AI/Core/src/generic/modelConfiguration.ts` |
| Engine accessor + facade delegate | `packages/AI/BaseAIEngine/src/BaseAIEngine.ts`, `packages/AI/Engine/src/AIEngine.ts` |
| Driver family | `packages/AI/Providers/{OpenAI,xAI,HuggingFace}` |
| Session builders (both topologies) | `packages/AI/Agents/src/realtime/*`, `packages/AI/Agents/src/base-agent.ts` |
| Generated (CodeGen) | `packages/MJCoreEntities`, `packages/MJServer`, `packages/Angular/Explorer/core-entity-forms` |
| PG policy | `.github/workflows/pg-migrations.yml`, `scripts/check-pg-migration-content.mjs`, `CLAUDE.md`, `migrations/CLAUDE.md` |
| Docs | `plans/model-configuration.md`, `guides/REALTIME_CO_AGENTS_GUIDE.md` |

## Design provenance

Shaped in conversation off the Grok Voice Think Fast 2 merge: the model went to PowerRank 12 and became the default co-agent model purely via metadata — this PR gives the same metadata-only leverage to *how* a model listens, not just *which* model answers.


packages/AI/Core/src/generic/modelConfiguration.ts
```diff
@@ -0,0 +1,169 @@
...
+    /** Vision knobs. Reserved. */
+    Vision?: JSONObject;
+    /** Audio (TTS/STT) knobs. Reserved. */
+    Audio?: JSONObject;
+}
+
+/** True for a plain JSON object (not null, not an array). */
+function isPlainObject(value: unknown): value is JSONObject {
+    return value !== null && typeof value === 'object' && !Array.isArray(value);
+}
```
AN-BC: This should be promoted to MJGlobal where we may have smoething like this, but it doesnt' need to be AI core specific and also, violated convention on naming, we use PascalCase on public members of classes and exported fnctions from packages.

AN-BC: Done in e331b511d0 — both points.

**Promoted to MJGlobal, PascalCase**: now `IsPlainObject` in `packages/MJGlobal/src/util.ts`, exported through the package index; `modelConfiguration.ts` imports it and the local copy is gone.

You were right that it shouldn't be AI-core specific — I checked, and MJGlobal had **no** exported equivalent (only an internal lodash call inside `util.ts`), while this exact guard had been re-implemented locally in ~8 places: `Config/config-merger`, `Integration/engine/RecordFlatten`, `MJCLI/SettingsMerger`, `MJCoreEntitiesServer/json-schema-lite`, `AI/ContextCrush/crush-json`, `AI/Agents/realtime-coagent-config`, plus an inline variant in `MJCore/baseEntity`. So there was nothing to reuse — there is now.

Documented it as deliberately **structural** (accepts class instances, `Date`, `Map`) with the call-site pattern for narrowing further, since the callers that need it are walking JSON-shaped data where those are either absent or should be opaque leaves. Added unit coverage including the `typeof null === 'object'` trap it exists to close and the array case that must replace rather than merge.

Builds clean across MJGlobal, AI/Core, AI/BaseAIEngine, AI/Engine, AI/Agents, AI/Providers/OpenAI; 595 MJGlobal + 235 AI/Core tests pass.

I left the other local duplicates in place — converting them touches six unrelated packages and would bury this PR's diff. Happy to do that sweep as a follow-up if you want it.

AN-BC:
## Correction: this PR's description states the wrong root cause for the IT25 failure

The section above explaining the `IT25 - View Execution (client-first)` / `view-execution.V10` failure is **wrong**, and the correct diagnosis is now merged in #3655. Leaving the original text intact for the record and correcting it here.

**What this PR claimed** — that V10 failed because of *sort-key ties*: the 32-row universe has only 12 distinct `Sequence` values, so "the view's `ORDER BY` ties while the keyset seek predicate compares `ID`."

**Why that was wrong** — the server *forces* `ORDER BY <pk>` in keyset mode and validates that any caller-supplied `OrderBy` references only the PK, so ties cannot arise on that path at all. The `Sequence` observation was a real property of the data, but it was not the cause.

**The actual cause** — a keyset walk was ordered inconsistently across its own pages:

```ts
const usingKeyset = !!params.AfterKey;   // FALSE on page 1 — no cursor exists yet
if (usingKeyset)  rawOrderBy = `${keysetPkColumnName} ${keysetDirection}`;   // ORDER BY <pk>
else              rawOrderBy = params.OrderBy ?? viewEntity?.OrderByClause ?? '';   // → ''
```

Page 1 has no `AfterKey`, so it is not the keyset path — and without `StartRow` it is not the OFFSET path either. It fell through to the plain row-limited branch and emitted `SELECT TOP N …` with **no `ORDER BY` at all**, so the engine returned an arbitrary N rows. Page 2+ then forced `ORDER BY <pk> ASC` plus the seek `<pk> > @afterKey`. The two orderings disagreed, so page-1 rows reappeared on page 2 and other rows could be skipped entirely.

The giveaway was in the same function: OFFSET pagination already carried the exact fallback the TOP/LIMIT path lacked, which is why V7/V9 (offset walks over the identical universe) passed while V10 failed.

**Also worth correcting**: the claim here that the failure is "pre-existing and unrelated to this change" was *true* — verified by showing V10's universe is byte-identical with and without this PR's rows — but the reasoning offered for *why* it failed was not. The conclusion was right for the wrong reason.

Fixed in #3655 by giving every row-limited query the determinism OFFSET pagination already had. IT25 now passes 11/11.

PR #3642, opened 2026-08-08
fix(logging): sanitize GraphQL and OAuth2 error diagnostics before logging

MS-BC:
## Summary

`GraphQLDataProvider.ExecuteGQL()` catches request errors and logs the error object directly as `fullError`. That object, produced by the underlying `graphql-request` client, carries the originating request on `e.request` — including its `variables` verbatim.

The consequence is generic to the transport layer: `ExecuteGQL` is the single chokepoint every GraphQL call in this client passes through, so it observes the variables of *every* mutation, and it has no way to know which fields in a given payload are sensitive. Any mutation whose input carries a secret therefore writes that secret, in plaintext, to whatever this process's console output is captured to. On a server that is typically a persistent log file — readable by anyone with log access, and retained regardless of whether the secret is later rotated at its source.

## Fix

Redact `request.variables` before logging. Every other diagnostic field is unchanged — response status, GraphQL error details, and the query text itself are all still logged exactly as before; only the variables are replaced with `'[REDACTED]'`.

## Test plan

- [x] Verified the redaction produces a sanitized object with `request.variables` replaced and every other field intact

## Notes for review

Blanket redaction of `variables` was chosen over a targeted allow/deny list of field names: which fields are sensitive varies per mutation, and generic transport-layer code cannot know the shape of the payload it is forwarding. Happy to narrow it if maintainers prefer a named-field approach.


MS-BC:
Follow-up filed as #3685 — the two `fullError` sites in the Box and Dropbox drivers, which were found during this work and scoped out because they are a different shape from the token-endpoint case (raw SDK error from a file operation, not an OAuth response body), so `describeTokenEndpointFailure` does not apply to them.

Noting explicitly for reviewers: whether those two SDK errors actually carry authentication material is **unverified** — establishing that is the first task on the issue.

packages/MJGlobal/src/TokenEndpointErrors.ts
```diff
@@ -0,0 +1,98 @@
...
+ *
+ * @example
+ * ```ts
+ * throw new Error(
+ *     `Token request failed: ${response.status}` +
+ *     describeTokenEndpointFailure(await response.text())
+ * );
+ * ```
+ */
+export function describeTokenEndpointFailure(body: string | null | undefined): string {
```
AN-BC: `PascalCase` for exported functions and public class members for compliance with MJ conventions.

AN-BC (approved):
LGTM

PR #3643, opened 2026-08-08
Fix Critical memory leak: AI Agent Harness sandbox never finalized

AN-BC:
## Summary

This PR is the output of a scheduled Memory Leak Audit run (`/audit-memory-leaks`, Round 10 — see `plans/MEMORY_LEAK_AUDIT.md`). The audit's ten parallel subagents found one new **Critical** finding, which this PR fixes, plus several new High/Medium/Low findings that are logged in the plan for future work but out of scope here.

### The bug

`HarnessAgentBase.EndHarnessSession()` (`packages/AI/AgentHarness/src/HarnessAgentBase.ts`) tears down an AI Agent Harness run's sandbox — it either stops the live Docker container `DockerSandboxProvider` provisioned (`docker run --detach --rm ... sleep infinity`), or removes the ephemeral workspace directory `LocalDirectorySandboxProvider` created. Its own docstring says it must run "on every exit path."

It had **zero callers anywhere in the monorepo**, confirmed via a full grep sweep. Every agent run using the `HarnessAgentType` driver class provisioned a sandbox in `startHarnessSession()` but never finalized it — each run leaked either a running Docker container or an orphaned temp directory, forever, with no automatic bound. Unbounded growth tied directly to repeated user activity (every Harness-agent run) — Critical severity.

**Root cause:** `BaseAgent.Execute()` (`packages/AI/Agents/src/base-agent.ts`) has a top-level `finally` block that already runs unconditionally on every exit path (success, failure, cancellation) and already performs analogous "always clean up" work (`releasePerRunDataCache()`). But it had no extension point for subclass-owned per-run resources, so `HarnessAgentBase`'s teardown method was simply never wired in.

### The fix

- Added `BaseAgent.finalizeRun(outcome: 'success' | 'failure' | 'cancelled')`, a no-op-by-default `protected` virtual hook, called unconditionally from `Execute()`'s existing `finally` block alongside `releasePerRunDataCache()`.
- Added a `deriveRunOutcome()` private helper that maps the run's final `AgentRun.Status` (`'Cancelled'` / `'Failed'` / anything else) to the 3-value outcome `finalizeRun` needs. `'AwaitingFeedback'` maps to `'success'` — it's a normal settled end-of-turn (a Chat-type final step), not an in-progress state, so the sandbox should be torn down the same way as `'Completed'`.
- Overrode `finalizeRun()` in `HarnessAgentBase` to call the existing `EndHarnessSession(outcome)`. This keeps `AI/Agents` free of any dependency on `AI/AgentHarness` — the hook is generic on the base class and only given meaning in the subclass that needs it.

Both sandbox providers' `Finalize()` implementations were already written defensively to never throw (see `ISandboxProvider.Finalize`'s doc contract) — the fix was purely the missing wiring, not a gap in the sandbox layer itself.

### Testing

- `packages/AI/Agents/src/__tests__/base-agent-finalize-run.test.ts` (new) — covers `deriveRunOutcome`'s status mapping (including the `AwaitingFeedback` → `success` case) and `finalizeRun`'s no-op default on the base class.
- `packages/AI/AgentHarness/src/__tests__/harness-agent-base-finalize-run.test.ts` (new) — covers `HarnessAgentBase.finalizeRun` delegating to `EndHarnessSession` with the right outcome, tearing down both the adapter session and the sandbox across all three outcomes, resetting internal state so a reused instance re-provisions cleanly, safe no-op when a session was never started, and independent error-tolerance for adapter-teardown vs. sandbox-finalize failures.
- Full monorepo build (`pnpm run build`, 309 packages) passes.
- Both packages' full unit test suites pass: `AI/Agents` — 98 files / 1830 tests; `AI/AgentHarness` — 5 files / 47 tests (including the 14 new tests above).

### Documentation

`plans/MEMORY_LEAK_AUDIT.md` was refreshed for Round 10 (this audit run) with the full findings from all 10 subagents, a "Fix Summary" section documenting this change, and cross-cutting recommendations (including retiring the vendor-connector findings from Rounds 6-9, since that code moved to a separate `MemberJunction/Integrations` repo this round).

## Test plan

- [x] `pnpm run build` — full monorepo build passes
- [x] `cd packages/AI/Agents && pnpm test` — 1830/1830 tests pass
- [x] `cd packages/AI/AgentHarness && pnpm test` — 47/47 tests pass
- [x] New regression tests added for both the generic `BaseAgent` hook and the `HarnessAgentBase` override

---
_Generated by [Claude Code](https://claude.ai/code/session_014e2ur3u28mbXRjL2BMQ7Rs)_


AN-BC (commented):
lgtm

Issue #3524, opened 2026-08-06
Task graphs: authorization model for human-task assignment (assignToUser)


AN-BC:
## Context

Deferred design item from the task-graph program plan (PR #3456), raised as **item 3 of the second review round** ([review comment](https://github.com/MemberJunction/MJ/pull/3456#issuecomment-5201514946) thread — rkihm-BC's v6 review) and explicitly deferred by @AN-BC in the same thread.

## Problem

Phase 4 of `plans/task-graph-primitive.md` introduces **human tasks as graph nodes**: an agent-emitted `TaskGraphSpec` can set `assignToUser`, creating a `MJ: Tasks` row with `UserID` set that blocks downstream agent work until the human completes it (with notification via `NotificationEngine` and DueAt escalation via the reconciliation sweep).

There is currently **no permission model** for this:

- Who may an agent-emitted graph assign blocking work to? Anyone? Only the submitting user? Role-gated?
- Can a low-privilege user's agent spawn cross-run work that lands blocking tasks (and notifications) on another user?
- How does this compose with the Unified Permissions model (`guides/UNIFIED_PERMISSIONS_GUIDE.md`) and with `MJ: AI Agent Requests` approval gates?

## Interim behavior (until this lands)

Per the plan (Phase 4 note): human tasks ship **self-assignment only** — a graph may assign human tasks to the graph's owning context user; cross-user assignment is rejected at submission validation. This keeps the hole closed without pre-empting the model designed here.

## Scope for this issue

1. Define the authorization model: who may assign to whom (self / role or permission-gated / delegation), evaluated at `TaskGraphService.Submit` validation time.
2. Integrate with Unified Permissions (a named permission for cross-user task assignment is the likely shape).
3. Decide notification/consent semantics for the assignee (silent task creation vs. accept/decline).
4. Cover the headless approval-as-human-task case (plan §5 resolved-O2), where the assignee is effectively an approver.

## References

- Plan: `plans/task-graph-primitive.md` (D5, D20, §3.4 HITL, Phase 4)
- Program plan: `plans/unified-workflow.md`
- PR: #3456

PR #3644, opened 2026-08-08
feat(actions): Track A — the entity-action runbook, and the filters that never fired

AN-BC:
Closes the **Phase 8 / Track A** row of #3456 — the unfinished #3408 entity-action runbook. Seven items, all sharing one failure shape: a column, a flag or a field that read as configured and did nothing.

Two additional bugs surfaced while proving the work against a live database. Both are described below; one of them is a correction to a claim an earlier PR of mine made.

---

## ⚠️ Behaviour change: Action Filters now actually prevent execution

`RunAction`'s filter-refusal branch built its result, logged it, and then **fell through to run the action anyway** — there was no `return`:

```ts
if (!filtersPassed) {
   const result = { Success: true, Message: "...should not be executed", ... };
   if (this.ShouldLogOutcome(params, result)) { /* log it */ }
}                                              // ← no return
return await this.RunActionWithTimeout(params, inputSnapshot);
```

So every Action Filter has recorded that it prevented something **while preventing nothing**, since the mechanism shipped.

The refusal row is why this went unnoticed for so long: the observable said "prevented" and the side effect happened regardless. #3606's changeset stated *"a filter that throws, yields a non-boolean, or has no evaluable logic prevents the action"* — that described filter **evaluation**, which landed, not filter **enforcement**, which did not. That statement was wrong and this corrects it.

**Impact:** any deployment with `ActionFilter` rows has been running actions those rows were configured to gate. After this they stop — the configured behaviour, but a visible change. Per @AN-BC, real-world usage is expected to be near-zero, which is self-reinforcing: a gate that silently passes everything doesn't accumulate users.

Found by `IT75 / EA2` as `expected 0, got 2` — one refusal row plus one execution row, for a save the filter had refused.

---

## The seven runbook items

| # | Item | What it was |
|---|---|---|
| 1 | **Transition filters** | An entity action could see current state only, so "when Status *becomes* Approved" was indistinguishable from "when Status *is* Approved" — true on every later save too |
| 2 | **`Validate` invocation hole** | A near-copy override that had drifted into a strict subset: no scope resolution, no provenance |
| 3 | **Durable `After*` dispatch (D14)** | After-hooks are fire-and-forget; a process dying mid-flight lost the action with nothing to retry it |
| 4 | **Record Process `OnChangeEnabled`** | Documented as "runs per-record on save via an owned Entity Action" since the column shipped. Owned nothing |
| 5 | **Execution-log retention** | `Action.RetentionPeriod` + `ActionExecutionLog.RetentionPeriod` had descriptions and **no reader anywhere in the codebase** |
| 6 | **`RunEntityAction` null contract** | `null` means "did not run, out of scope". The provider guarded; the GraphQL resolver didn't, surfacing it to clients as a server error |
| 7 | **IT75 check bundle** | The seams above were unit-tested with each other mocked |

### Transition filters

`EntityChangeContext` carries both sides of a save to where filters run, built from `EntityField.OldValue` — which `BaseEntity` has tracked all along and simply never carried anywhere. Filter code gets `DidFieldChange`, `DidFieldChangeToValue`, `OldValues`, `NewValues` on `ActionFilterContext`.

The capture is the **first statement** of `HandleEntityActions`, deliberately before its first `await`: after-hooks are fire-and-forget, so the moment that method yields, the save completes and `finalizeSave()` resets every `OldValue`. Reading `IsCreate` from that same synchronous snapshot also closes a latent bug — `entity.IsSaved` was previously read *after* an await, so a create whose save finalized in that window dispatched as `AfterUpdate`.

A create reports **no** changes: a record whose Status started at `Approved` did not *become* anything.

Two substrate fixes fell out of using filters for real:
- `EntityActionFilter.Status` was never consulted, so a `Disabled` binding still gated. Filters fail closed, so that wasn't an inert row — it was a permanent block whose only symptom is a trigger that quietly stopped firing.
- A binding pointing at an unresolvable filter reached the evaluator as `undefined` and threw there: fail-closed *by accident*, with no usable reason logged.

`ValidateWorkflowSpec` also stops refusing `WorkflowEntityEventTrigger.filter` — it now reconciles onto an owned `ActionFilter` and validates that the expression parses, because a syntax error in a fail-closed filter isn't a loud failure, it's a trigger that silently never fires.

### Durable dispatch, and the bug in my first cut of it

`EntityAction.RunMode = 'Durable'` routes an `After*` dispatch to the task-graph substrate as a single-node durable graph — the claim protocol, restart recovery and orphan reclaim that already exist (D14: not a third async substrate). Opt-in per binding, because it costs a Task row, a dispatcher hop, and params persisted at rest.

**My bug, caught here:** I first put the `RunMode` check in `HandleEntityActions`, ahead of `RunEntityAction` — so a durable binding was submitted **before its scope check and before its filters ever ran**. A scoped durable trigger would have fired for every record of the entity; a filtered one on every save.

Fixed by making durability replace **execution**, not **dispatch**: `RunActionParams.DeferExecution` is called by `RunAction` in place of running the action, after validation and filters pass. The two paths now share one gate. Returning `null` from a deferral declines it and the action runs normally, which is what makes the "submission failed → run it inline" fallback free rather than a reimplementation of execution.

Falling back rather than dropping is the posture throughout: `Durable` asks for the work to be *harder to lose*, so refusing to run it where the durable path is unavailable would make opting in **less** reliable than leaving it off.

---

## Schema

`V202608081200__v6.1.x__Durable_EntityAction_Dispatch.sql` — ~90 lines of hand-written DDL, with CodeGen output appended below the 50-blank-line separator + do-not-edit block per convention.

- `EntityAction.RunMode` (`'Inline'` default | `'Durable'`)
- `Task.ActionID`, widening `CK_Task_Assignment` from two-way to three-way exclusivity — the dispatcher ran agent nodes and waited on human ones; there was no third shape, so a durable action graph had nowhere to record *which* action except smuggled through `InputPayload`

The generated section also carries display-name corrections for **pre-existing** columns on `EntityAction` and `Task` (`"Entity ID"` → `"Entity"`, `"__mj _Created At"` → `"Created At"`). Not stray edits: CodeGen recomputes `DisplayName` for every field of an entity it regenerates, and its naming rules have improved since those rows were written. Scoped to those two entities; I checked the diff for unrelated regeneration and found none. Called out in the separator header too.

---

## Verification

| Gate | Result |
|---|---|
| Full repo build | ✅ **271/271 tasks**, exit 0 |
| Unit tier (all touched packages) | ✅ actions 218 · actions-base 128 · ai-core-plus 391 · task-graph 33 · scheduling 239 · core-entities-server 392 · integration-suite 143 |
| `check:standards` | ✅ |
| `check:ui` | ✅ 0 violations |
| `check:claude-md` | ✅ 672 links, routing, guides |
| Deterministic integration tier | ✅ **61/61 (100%)**, exit 0 — run with `RUN_MUTATION_TESTS=1` |

**New unit regressions at a seam that had none:** a filter that refuses must leave `InternalRunAction` **and** `RunActionWithTimeout` uncalled. Asserting on the returned message alone would *not* have caught the original bug — which is exactly how it survived.

**IT75** (`entity-actions`, EA1–EA8) drives what unit tests structurally cannot: the change context built from an entity the save is about to reload, a filter compiled from a row and evaluated fail-closed, scope refused by a resolver walking real foreign keys, a durable binding handed to a substrate. Deterministic tier, mutation-class, server transport, sequenced at 33 per #3251.

Three of its own fixtures were wrong on first run and are fixed: it counted log rows as executions (a prevented run is logged too — deliberately, so an operator can see a filter refused), it tried to delete an `ActionFilter` a junction row still references, and EA6/EA7 counted submissions across every `Durable` binding earlier checks had left Active rather than their own.

### Two other bundles the tier turned up

**`actions-pipeline.AP4` — a known-gap pin doing its job, one PR late.** AP4 pinned bug register **B14**, and its own comment named this exact bug before it was reachable:

> *"Corollary (unreachable-today bug): RunAction's filters-failed branch builds a result but FALLS THROUGH to RunActionWithTimeout instead of returning it — masked only because the stub can never return false."*
> *"When the stubs are implemented, this check MUST flip."*

#3606 implemented `RunSingleFilter` and made the corollary reachable — the pin should have flipped there and did not, because the fall-through kept executing the action, so a garbage filter still reached the action's own `MISSING_PARAMETERS` and AP4 kept passing **for the wrong reason**. Flipped here: it now asserts prevention, with the decisive assertion being that the action's own validation never spoke. B14's `ValidateInputs` half is untouched and still pinned in the same check; the bug register records the split disposition.

**`content-vectorization.CV1–CV6` — pre-existing on `next`, fixed here.** `GetSourceRecordFieldPaths` was added to `VectorDBBase` and called from `AutotagBaseEngine.resolveDriverFieldPaths` as part of the autotagger source-field work. The **unit** double in `AutotagBaseEngine.test.ts` was updated for it; the **integration** stub was not, so all six died on `is not a function` — an incomplete fake satisfying an interface that had moved on. Fixed with the same one-liner the unit double uses. Verified unrelated to this work: the branch changes zero files under `packages/AI/Vectors`, and the failure survived a clean full rebuild, so it was not stale `dist`.

---

## Notes for review

- **The plan ledger is not updated here.** `plans/task-graph-primitive.md` lives on `origin/claude/sage-task-graph-study-4uvtrc`, not on `next`, so the Track A row can't be closed from this branch. Needs the planning agent, or a nod to cherry-pick.
- **IT75 only executes with `RUN_MUTATION_TESTS=1`** — same posture as IT74 and the other mutation bundles, but it means these seams aren't continuously protected on the default blocking run.
- **Out of scope**, per the wrap-up directive: Track R (#2708), #3524 assignment authz, trigger vocabulary normalization, PG counterparts + ratchet entries.

/cc @AN-BC @rkihm-bc


AN-BC:
@AN-BC — tagging you here since GitHub won't let the author be a requested reviewer.

**Release-note item** (per your call to surface it in PR comments): Action Filters have never prevented execution. `RunAction`'s refusal branch logged a refusal row and then fell through to run the action anyway — no `return`. So a filter recorded preventing something while preventing nothing, and the refusal row is exactly why nobody noticed: the observable said "prevented" while the side effect happened. #3606's changeset claimed filters fail closed; that described *evaluation*, which landed, not *enforcement*, which didn't. Corrected here.

Impact is as you predicted — near-zero, and self-reinforcing: a gate that silently passes everything doesn't accumulate users. Anyone who does have `ActionFilter` rows will see those actions stop firing, which is what the rows always asked for.

Two things I could not close from this branch:

1. **The plan ledger.** `plans/task-graph-primitive.md` lives on `origin/claude/sage-task-graph-study-4uvtrc`, not `next`, so the Phase 8 / Track A row can't be updated here. Needs the planning agent, or a nod from you to cherry-pick.
2. **IT75 only runs under `RUN_MUTATION_TESTS=1`** — same posture as IT74 and the other mutation bundles, but worth knowing these seams aren't continuously protected on the default blocking run.

Verification: full build 271/271 exit 0 · deterministic tier **61/61 (100%)** with mutations on · unit tier green across all touched packages · standards/ui/claude-md gates pass.

cadam11:
Heads up: next's Integration Tier went red after this merged, failing at the fresh-DB migrate step, and I think it needs your eyes since it is inside the generated block of `V202608081200__v6.1.x__Durable_EntityAction_Dispatch.sql`.

**The failure (deterministic, every run since 20:33Z):**

```
FAILED: V202608081200__v6.1.x__Durable_EntityAction_Dispatch.sql
Error: Failed at batch 9/41 (lines 110-396): The INSERT statement conflicted with the
FOREIGN KEY constraint "FK_EntityFieldValue_EntityField" ... table "__mj.EntityField", column 'ID'.
```

**What I ruled out before pinging you:**

- The two `EntityFieldValue` inserts (file lines ~322/328) both reference `DA98DF59-...` — the RunMode field created by the guarded `INSERT INTO EntityField` earlier in the same batch. Same batch, same transaction, so read-your-writes applies: if that insert ran, the FK target exists.
- Nothing else plants RunMode or that UUID first: `git grep RunMode` across all of `migrations/` on next hits only this file, and the v6.1 release Metadata_Sync (`V202608080752`, which runs before this at 07:52 < 12:00) contains neither `da98df59` nor entity `34248F34`.
- The field INSERT's column/value lists align (26/26) and every column exists on a fresh 6.1 `EntityField`, so it is not a binding error — a missing column would abort the batch with "Invalid column name" before the value inserts ever ran.

**What that leaves (needs a live repro to separate):**

1. The guarded insert is being skipped because its `IF NOT EXISTS` matches something on the freshly-migrated DB that static analysis can't see (e.g. a RunMode row created by an earlier migration under a different ID via dynamic SQL, or a collation/whitespace surprise in the Name match).
2. The field INSERT fails at runtime (FK/CHECK/trigger) as a statement-level error, the batch continues, and the runner surfaces the later FK error instead of the first one.

Either way the fragile shape is real: the value inserts hardcode the new UUID while the field insert is allowed to silently skip. Per the block's own header ("do not patch this section — re-run `mj codegen` and replace the entire generated block"), regenerating against a truly fresh 6.1 DB is probably the clean fix; alternatively derive the `EntityFieldID` for the value rows by `(EntityID, Name)` lookup so they cannot detach from the guard.

Repro is just the Integration Tier's migrate step on an empty DB. Happy to run it locally and get you the first-error-in-batch if that would save you time — say the word.

cadam11:
Follow-up: I ran the repro on a fresh DB and have the root cause with a verified fix. The FK error is a decoy.

**Real first error** (running batch 9's `EntityField` INSERT standalone against the migrated-so-far DB):

```
Msg 2627: Violation of UNIQUE KEY constraint 'UQ_EntityField_EntityID_Sequence'.
The duplicate key value is (34248f34-2837-ef11-86d4-6045bdee16e6, 100025).
```

`V202608042200__EntityAction_Workflow_Extensions` (four days earlier) gave the **virtual field `ScopeEntity`** Sequence 100025 on `MJ: Entity Actions`. This migration then assigns the same slot to `RunMode`. Msg 2627 is statement-level ("statement has been terminated"), the batch continues, the two `EntityFieldValue` inserts hit the missing field row, and node-mssql surfaces only the last error of the batch — the FK one — while skyway drops `precedingErrors`. Hence the misleading log.

**Why your dev DB passes:** live CodeGen renumbered `ScopeEntity` in place when `RunMode` became a real column, but CodeGen only emits INSERTs for new fields, never the UPDATE that moves an existing field's Sequence — so the renumber exists on the authoring machine only and the collision ships in the migration. (Checked: this file's only touch of `9286E6FD…` is a Category/DisplayName update, no Sequence.)

**Verified fix** — one guarded statement in the hand-written section (after the `CK_Task_Assignment` GO at line 109, above the CodeGen banner, so it respects the "do not patch the generated block" rule):

```sql
UPDATE [${flyway:defaultSchema}].[EntityField]
   SET [Sequence] = 100026
 WHERE [ID] = '9286e6fd-c29a-47ba-8690-0a35bdf96cc5'
   AND [Sequence] = 100025;
GO
```

No-op on dev DBs where the renumber already happened; transient anyway since `R__RefreshMetadata` renumbers everything at the end of every run. With this patch the full chain applies clean on a fresh DB: 39 applied, zero FAILED, RunMode column + both value rows + ValueListType all land, no downstream migration breaks.

One caveat when you land it: editing the file changes its checksum, so anyone whose dev DB already applied D14 (drift-ahead machines where it happened to pass) may see a validate mismatch — worth one line in dev chat.

Two systemic follow-ups spun out as issues so they don't get lost in this thread: the collision class itself (any CodeGen migration adding a real column to an entity with virtual fields can do this, undetectably on the authoring machine) and skyway swallowing `precedingErrors`.

Issue #3645, opened 2026-08-08
mj codegen manifest --lazy-config collapses per-module chunk loaders into one


AN-BC:
Running `npm run mj:manifest:explorer` on `next` rewrites `packages/Angular/Explorer/explorer-core/src/generated/lazy-feature-config.ts` in a way the file's own header explicitly forbids.

**Committed shape** — one loader per submodule, each with a distinct `chunkId`:

```ts
// --- @memberjunction/ng-dashboards → ./actions-dashboards.module (6 entries) ---
const loadNgDashboardsActionsDashboardsModule = {
  chunkId: '@memberjunction/ng-dashboards/actions-dashboards.module',
  load: () => import('@memberjunction/ng-dashboards/actions-dashboards.module').then(() => {})
};
// …~20 of these
```

**Regenerated shape** — all of them collapsed to a single `loadNgDashboards`, net −225/+115 lines.

The header warns about precisely this:

> Every entry carries an explicit `chunkId` — the dynamic import specifier — because many compound keys share one chunk and the registry must be able to tell those chunks apart. Do NOT collapse these into a shared helper that returns a closure…

**Impact if committed:** every Explorer dashboard lazy-loads the whole `ng-dashboards` package as one chunk instead of per-feature chunks — a bundle regression across the entire app, landing inside a generated file nobody re-reads on review.

**How it surfaced:** adding the Workflows app in #3456's Phase 5a needed two new entries. The generator produced them correctly *and* rewrote everything else. I discarded the regeneration and hand-added the two entries in the existing style (see the Phase 5a PR), so `next` is unaffected — but the generator will do this again to whoever runs it next.

**Guess at the cause** (unverified): the generator likely enumerates chunks from built `dist`/package `exports` and, not finding the per-submodule subpaths it expects, falls back to grouping by package. Worth checking whether it needs a built workspace, or whether the subpath-export detection regressed.

Not fixing it here — it is orthogonal to the workflow program, and the failure mode is quiet enough to deserve its own change.

AN-BC:
**Correction to the diagnosis above.**

The collapse is **not** something the manifest step does generally — a full `pnpm run build` regenerates this file *correctly*, with all per-module loaders intact. It is running the command **standalone** that collapses it:

| How it ran | `chunkId: '@memberjunction/ng-dashboards/…'` loaders |
|---|---|
| `pnpm run build` (full) | **16** ✅ |
[...]

PR #3648, opened 2026-08-08
feat(workflows): Phase 5a — the Create Workflow front door, in a Workflows app of its own

AN-BC:
Closes the **Phase 5a** row of #3456 — the Create Workflow front door.

Phase 5 shipped every component this composes (the canvas, the properties panel, the runtime-overlay source, both Save-as-Workflow surfaces) but not the front door itself, because the design was not locked. It is now: `mockups/workflow-ux/front-door-v1.html` carries five ratified answers, and this implements all three of its screens against them.

---

## A Workflows app, not a tab under AI

D18 puts **Workflow** in front of end users while *Flow Agent* survives in metadata and dev docs. Filing this surface under "AI" would contradict that at the navigation level — someone looking to automate a task does not look under AI — and D19 exists precisely because the editor is today buried inside a saved agent record. Scheduling and Routines already set the precedent for an AI-adjacent domain getting its own app.

## The three screens

**A + B — the front door.** Three doors in the locked order, "Describe it" pre-selected because it is the only one that needs no prior knowledge of the product. Each tile states *when to pick it*, not just what it does — that is the actual question someone has on this screen.

**Only settled runs are promotable.** An in-flight run may still change shape under a retry or a recovery branch, so the saved workflow would not be the one that ran. Enforced three ways, because dimming alone leaves a row clickable and keyboard-reachable:
- the handler guards on `IsSettled`
- the row leaves the tab order (`tabindex="-1"`)
- `aria-disabled` is set

**C — Save as Workflow (answer ④).** The plan card previously emitted a save with **no name** and no way to continue editing, leaving the host to invent both. Now: one field seeded from the plan's own name, a Save button, and "Open in editor" as a *secondary*. Making the editor mandatory turns a two-second capture into a task, which is the friction that stops good one-off plans becoming reusable. Once the user types, their value sticks — including empty — because re-seeding would fight their backspace. A nameless workflow will not save; there would be nothing to find it by.

**Nothing anywhere asks for a trigger or a schedule.** Saving is capture, not scheduling. A workflow runs on demand until someone gives it a cadence, and the card *says so* rather than leaving it to be discovered.

## D18 enforced by test, not by review

The vocabulary rule is invisible to a compiler and erodes one label at a time. So the templates and user-facing copy are asserted to contain no *graph* / *DAG* / *node* / *Flow Agent* — with a companion assertion that **"step" IS present**, so the rule cannot be satisfied by deleting the concept instead of renaming it.

The guard immediately caught my own template (`<mj-task-graph-editor>` contains "graph"). A tag name is markup, not copy, so the check now strips tag names alongside comments and bindings — the rule is unchanged, the definition of "text a person reads" is stated more precisely.

## The canvas is embedded, not routed to

I first wrote `navigationService.NavigateToResource('Workflow Editor', …)`. **That method does not exist and neither does that resource — I invented both.** The compiler caught the method; the missing resource would have shipped as a button navigating nowhere.

The canvas is embedded instead, which is what the plan actually specifies: it is a widgets-layer component that refuses to know about routing, and the app supplies the shell around it. Opening a *saved* workflow goes to `OpenEntityRecord('MJ: AI Agents', …)` — a workflow **is** a Flow agent — because until the editor has its own route, the record form is the honest destination rather than a link to nothing.

---

## Verification

| Gate | Result |
|---|---|
| Full repo build | ✅ **299/299**, exit 0 |
| `ng-dashboards` unit tier | ✅ 1,707 passing (see note) |
| `ng-conversations` unit tier | ✅ **1,077/1,077** |
| Workflows front-door specs | ✅ 19/19 (incl. the D18 guard) |
| Plan-card specs | ✅ 15/15 (6 new for ④) |
| Hex-token + `mj-btn` gates | ✅ 0 violations |

**Pre-existing failure, not from this branch:** 7 tests in `Actions/components/explorer/action-list-item.component.dom.test.ts`. This branch touches nothing under `src/Actions`; I verified by checking out `next`'s own `dashboards` tree and running that file — same 7 failures. Untouched here.

## ⚠️ Not done — Playwright

The directive's Phase 5a scope includes **Playwright structure/behavior checks and a small visual baseline set**. Those are **not in this PR**. Behavior is covered by 34 unit/DOM tests including the vocabulary guard, but there is no browser-level verification of the three screens yet. Flagging rather than letting the scope look complete.

## Also in here

**A generated file I had to fix twice** — see #3645. `npm run mj:manifest:explorer` run *standalone* collapses ~20 per-module lazy-chunk loaders into one shared loader, which would make every Explorer dashboard lazy-load the whole `ng-dashboards` package as a single chunk. A full `pnpm run build` regenerates it *correctly*. I caught it, fixed it, and then let a later `git add -A` sweep the collapsed version back in — so the final commit here restores all 16 per-module loaders. I have corrected the diagnosis on the issue and suggested a "chunk count must never decrease" guard.

/cc @AN-BC @rkihm-bc


AN-BC:
@AN-BC — tagging you here since GitHub won't let the author be a requested reviewer.

**Two things I'd want a second opinion on:**

1. **Where "Create →" lands.** The mockup shows the front door handing off to the canvas. I embedded `ng-task-graph-editor` in the app rather than routing, because the plan says the canvas embeds and the widget refuses to know about routing — and because no "Workflow Editor" resource exists to route to. If you'd rather that become a real resource with its own route, it's a small change but a different shape.

2. **Opening a saved workflow** currently goes to the `MJ: AI Agents` record form. That's honest (a workflow *is* a Flow agent) but it's also the exact experience D19 called "buried". It's a placeholder until the editor has a route of its own — worth knowing before this ships.

**Playwright is not in this PR** — it's in the directive's 5a scope and I did not get to it. Behavior is covered by 34 unit/DOM tests, but nothing exercises the three screens in a browser. Say the word and I'll add it before you merge, or as a follow-up.

Also worth your attention: **#3645 got worse on inspection.** The lazy-chunk collapse only happens when `mj:manifest:explorer` runs *standalone* — a full build is fine. That's the run someone does after adding one nav item, which is exactly when they won't read a 340-line generated diff. I hit it, fixed it, then let a `git add -A` put it back; the last commit here restores it properly.

PR #3651, opened 2026-08-08
fix(codegen): stop Smart Field Identification clearing IsNameField on IS-A children

### LINKED ISSUE #3551: CodeGen SmartFieldIdentification can clear IsNameField without nominating a replacement, leaving an entity with no name field (IS-A child's virtual Name)
CodeGen's `SmartFieldIdentification` advanced-generation feature can **clear `IsNameField` and nominate no replacement**, leaving an entity with **zero** name fields. It appears to happen where the candidate is a **virtual** field — specifically an IS-A (Table-Per-Type) child whose `Name` is mirrored from its parent.

Worth noting it *overrides a correct deterministic answer*. `getPendingEntityFieldINSERTSQL` already has the right rule:

```ts
// packages/CodeGenLib/src/Database/manage-metadata.ts:3867
const isNameField = n.FieldName?.trim().toLowerCase() === 'name' || n.IsNameField;
```

A field literally named `Name` is the name field. The AI pass then unsets it, and the entity ends up with none.

### Repro
1. An app with an IS-A relationship declared via `additionalSchemaInfo` `ISARelationships` — here `AccountingCompanyProfile` IS-A `MJ: Companies`, so CodeGen mirrors the parent's `Name` onto the child as a **virtual** field.
2. Enable advanced generation with `SmartFieldIdentification`:
   ```js
   advancedGeneration: { features: [{ name: 'SmartFieldIdentification', enabled: true }, /* … */] }
   ```
3. Rebuild metadata from zero so the entities are new: drop schema → migrate to bare tables → `mj codegen` with AI on.
4. Inspect the child entity's fields.

### Expected vs actual
**Expected:** either `Name` keeps `IsNameField = 1`, or the AI nominates a different field. An entity should never come out of CodeGen with no name field.

**Actual:** `Name` has `IsNameField = 0` and no other field is nominated.

```sql
SELECT e.Name FROM __mj.Entity e
WHERE NOT EXISTS (SELECT 1 FROM __mj.EntityField f WHERE f.EntityID = e.ID AND f.IsNameField = 1);
-- includes the IS-A child
```

Before the AI pass: `IsNameField=1` on `Name`. After: nothing.

### Impact
Anything resolving a record's display value loses it — FK lookups to that entity render the raw UUID, and `getIsNameFieldForSingleEntity` (`sql_codegen.ts:1975`) has no field to denormalize into referencing base views. Low blast radius when the entity isn't a lookup target, but silent, and it *removes* correct pre-existing metadata rather than merely failing to add any.

### Suggested fix
Two independent guards, either of which would prevent this:

1. **Never clear without replacing.** If the AI's proposal would leave an entity with zero `IsNameField`, discard it and keep the deterministic result. "No name field" is not a valid outcome for a proposal whose job is to *identify* one.
2. **Don't discount virtual fields on IS-A children.** A mirrored parent `Name` is exactly the right display value; virtuality is an implementation detail of the IS-A join.

Worth checking whether the same "clear without replace" pattern applies to the `DefaultInView` / `IncludeInUserSearchAPI` values this feature also writes.

### Workaround
`AutoUpdateIsNameField = 0` pins the field so the AI cannot clear it. It's awkward to apply, though: it's database state, and a *full* metadata emit only happens against a bare database — where there are no rows to pin. So it can't be baked into a regenerated baseline without a post-generation patch step. A config-level route would be better; `additionalSchemaInfo` currently reaches only `__mj.Entity` columns (`EntityConfig`), not `EntityField`.

### Environment
MJ `next` as vendored in an mjdev instance, 2026-08-06, found while rebuilding bizapps-accounting's metadata baseline with AI enrichment enabled. Gemini via `AI_VENDOR_API_KEY__GeminiLLM`. 23 entities, 291 fields; 1 entity affected.

MarceloT-BC:
Fixes #3551.

CodeGen's `SmartFieldIdentification` could clear `IsNameField` and nominate no replacement, leaving an entity with **zero** name fields — and it did so by **overriding a correct deterministic answer**, not merely failing to add one.

## Root cause

`isFieldEligibleForNameField` blanket-rejected `IsVirtual` fields. An IS-A (Table-Per-Type) child's `Name` is virtual by construction: it lives on the parent table and reaches the child through the view's IS-A join (`generateParentEntityFieldSelects`). So `selectNameFieldWinner` returned `null`, and `applyNameFieldUpdates`' clear-loop then wiped the `IsNameField = 1` the deterministic pass had correctly set (the pending-fields SQL emits `IIF(sf.FieldName = 'Name', 1, 0)`).

Downstream, anything resolving a record's display value loses it: FK lookups render the raw UUID, and `getIsNameFieldForSingleEntity` has no field to denormalize into referencing base views.

## Fix — two independent guards

**1. IS-A carve-out.** A virtual field is eligible when it is an IS-A inherited field, discriminated by `IsVirtual = 1 AND AllowUpdateAPI = 1` — the same test `syncISAParentFields` and `buildParentChainContext` already use, and unambiguous here: the pending-fields SQL forces `AllowUpdateAPI = 0` on every virtual column discovered from a base **view**, and the IS-A sync is the only thing that sets it back to `1`. A borrowed FK-name column therefore can never qualify, so the original reasons for rejecting virtual fields (the unbuildable self-FK view join; a name that is itself a borrowed FK-name resolving circularly) still hold where they applied.

**2. Never clear without replacing.** Eligibility now splits into type-safety (`isNameFieldTypeSafe`: PK / non-text / unbounded MAX are *wrong*) and virtuality (merely a *worse* choice). `selectNameFieldWinner` gains a 4th tier that **preserves** an existing type-safe flag rather than clearing it with nothing to put in its place. A flag on an actively-wrong field is still cleared with no replacement, so the v5.40 uniqueidentifier-PK guardrail is untouched — that case has its own regression test here.

Checked the sibling writers while in the area: `applyDefaultInViewUpdates`, `applySearchableFieldUpdates` and the field-level FTS path are all SET-only and never clear, so the clear-without-replace pattern was unique to `IsNameField`.

## How to recreate (fails before, passes after)

1. Two tables in one schema — `ReproParent(ID uniqueidentifier PK, Name nvarchar(200))` and `ReproChild(ID uniqueidentifier PK FK → ReproParent.ID, ProfileCode nvarchar(50))`.
2. Declare the inheritance in the `additionalSchemaInfo` config:
   ```json
   { "ISARelationships": [
       { "ChildEntity": "ReproChild", "ParentEntity": "ReproParent", "SchemaName": "<schema>" } ] }
   ```
3. Enable advanced generation with `SmartFieldIdentification` and a real provider key, then run `mj codegen` so the entities register fresh.
4. Query for entities with no name field:
   ```sql
   SELECT e.Name FROM __mj.Entity e
   WHERE NOT EXISTS (SELECT 1 FROM __mj.EntityField f
                     WHERE f.EntityID = e.ID AND f.IsNameField = 1);
   ```

**Before:** the child appears, and its `Name` row is `IsVirtual = 1, IsNameField = 0`. **After:** `IsNameField = 1`, result empty.

## Validation

Verified end-to-end on a live instance against `next`, running the two identical schemas above one per CodeGenLib build with Gemini enrichment on (4 successful `AIPromptRun` rows, ~25k tokens, per run):

| build | child `Name` | zero-name-field query |
|---|---|---|
| unfixed | `IsVirtual=1, IsNameField=0` | returns the child ❌ |
| fixed | `IsVirtual=1, IsNameField=1` | empty ✅ |

`smart-field-identification-guardrails.test.ts` gains 10 specs — the IS-A eligibility carve-out, the `isInheritedISAField` discriminator (including that it rejects a borrowed FK-name column and MJ system columns), the #3551 regression itself, fresh-pick of an inherited `Name`, the preserve tier and its literal-`Name` tie-break, and that a real replacement still displaces a view-only flag. **7 of the 10 fail against the unfixed code**; the other 3 are regression guards that must pass either way.

CodeGenLib: **751 passed | 60 skipped**, `tsc` clean — both before and after merging `origin/next`.

A **patch** changeset for `@memberjunction/codegen-lib` is included.


AN-BC (approved):
LGTM

PR #3655, opened 2026-08-08
fix(runview): order row-limited queries by PK — fixes duplicate/skipped rows in keyset pagination

AN-BC:
# Keyset pagination returned duplicate rows — and could silently skip others

Root cause of the long-standing `IT25 - View Execution (client-first)` / `view-execution.V10` failure that has kept the deterministic integration tier at **59/60**.

## The bug

A keyset (`AfterKey`) walk was **ordered inconsistently across its own pages**.

In `GenericDatabaseProvider.RunViewGeneric`:

```ts
const usingKeyset = !!params.AfterKey;   // FALSE on page 1 — no cursor exists yet
if (usingKeyset)  rawOrderBy = `${keysetPkColumnName} ${keysetDirection}`;   // ORDER BY <pk> ASC
else              rawOrderBy = params.OrderBy ?? viewEntity?.OrderByClause ?? '';   // → ''
```

- **Page 1** has no cursor, so `usingKeyset` is false. It is not the OFFSET path either (keyset callers pass no `StartRow`), so it fell to the plain row-limited path and emitted `SELECT TOP N …` **with no `ORDER BY` at all**. SQL Server is then free to return an arbitrary N rows.
- **Page 2+** forced `ORDER BY <pk> ASC` plus the seek predicate `<pk> > @afterKey`.

The two orderings disagree, so any page-1 row that sorts *after* page 1's last row is returned **again** on page 2 — and rows the arbitrary first page happened to include can be **skipped by the walk entirely**.

The tell was sitting in the same function: OFFSET pagination already had exactly this fallback —

```ts
if (usingPagination && entityInfo.FirstPrimaryKey) {
    if (!orderBy) viewSQL += ` ORDER BY ${this.QuoteIdentifier(entityInfo.FirstPrimaryKey.Name)}`;
```

— it was simply never applied to the TOP/LIMIT path. That is precisely why the offset walk (V7) passes over the *identical* universe while the keyset walk (V10) fails.

## The fix

When rows are limited and the caller supplied no `OrderBy`, order by the primary key.

`TOP N` / `LIMIT N` without an `ORDER BY` is undefined by definition — the engine may return different rows run to run, or between two runs of the same walk — so this is a **general determinism fix**, not a keyset-specific patch. It happens to fix keyset because page 1 now agrees with every later page: both order by the same PK.

Details:
- The provider-generated clause bypasses `ValidateUserProvidedSQLClause` and `TransformExternalSQLClause` — it is a quoted identifier the provider produced, not user input.
- The OFFSET block keeps its own fallback, since `OFFSET/FETCH` is a hard **syntax** error without an `ORDER BY`; the comment now says so explicitly.

## Verification

Proven at SQL level in **both directions**, against real data on the exact universe V10 walks (32 `EntityField` rows, 8-row pages):

| | duplicates | rows walked | missed |
|---|---|---|---|
| **before** (page 1 unordered) | **4 of 8 on page 2** | — | — |
| **after** (page 1 ordered by PK) | **0** | **32 / 32** | **0** |

The "4 duplicates in an 8-row page" reproduces the assertion failure exactly: `V10 keyset page 2 re-returned row(s) from an earlier page — expected 8, got 4`.

- Full monorepo build clean — 299/299 and 271/271, zero TS errors.
- `GenericDatabaseProvider` tests: **866 passed, 5 skipped**, including the existing keyset suite.

## On test coverage

No new unit test. The invariant is already pinned by an existing test that was **correctly failing** — `IT25 view-execution.V10` asserts exactly "every row exactly once, walk ends on a short page". The provider has no harness for capturing full generated view SQL today (the closest, `uuid-case-sensitivity.test.ts`, simulates the logic rather than invoking it), so unit-tier coverage would mean building that harness. Happy to add it if you'd rather have the assertion at both tiers.

I have **not** yet run IT25 end-to-end: it is a client-first bundle, so it needs MJAPI running this code, and port 4000 currently has your server. Say the word and I'll bring one up on an alternate `GRAPHQL_PORT` rather than restarting yours.

## ⚠️ Behavior note worth your judgement

This changes **which** N rows an unordered `MaxRows` query returns. That set was previously undefined and could vary between runs, so nothing correct can have depended on it — but:

- callers incidentally observing storage order will now see primary-key order;
- entities whose PK is not the clustered index gain a sort.

## Correction to PR #3640

The description of the now-merged #3640 attributes this failure to **sort-key ties** (32 rows / 12 distinct `Sequence` values). That explanation is **wrong** — the server forces `ORDER BY <pk>` in keyset mode and rejects any non-PK `OrderBy`, so ties cannot arise there. The defect is an *absent* `ORDER BY`, not a tied one. Flagging it rather than editing a merged PR's body.


AN-BC (commented):
LGTM but would like @rkihm-BC and @jordanfanapour to check this out. Doing a deep dive with Fable/Ultracode right now too

AN-BC:
## End-to-end verification: IT25 now passes over the wire

Ran the client-first bundle against MJAPI built from this branch (`MJ_6_1_0`, migrated to branch schema):

```
IT25 - View Execution (client-first)   ✓ PASSED   11/11 checks, score 1.0000
  ✓ view-execution.V10  keyset (AfterKey) walk returns every row exactly once; short final page ends it
```

V10 is the check this PR fixes. Worth noting its neighbours also stay green, which is the regression risk this change carried:

- **V9** — OFFSET pagination over a stable sort: union == full set, no duplicates, no gaps
- **V11** — keyset on a composite-PK entity is still *refused*, never silently degraded to OFFSET
- **V12** — `MaxRows` honored; `IgnoreMaxRows` returns the full set; entity `UserViewMaxRows` fallback intact

That last one matters most: V12 exercises exactly the row-limited path this PR changed, and it still passes.

## Full deterministic tier: 60/61

The one failure is **`codegen-determinism.CD3`**, and it is unrelated to this change — this PR touches only `ORDER BY` emission in RunView SQL and cannot affect entity metadata:

```
codegen-determinism.CD3: 4 schema↔metadata field mismatches:
  'MJ: Action Execution Logs'.EntityAction: generated schema key has no live field (stale generated artifact?)
  'MJ: Entity Action Filters'.EntityAction
  'MJ: Entity Action Invocations'.EntityAction
  'MJ: Entity Action Params'.EntityAction
```

Verified on the live database: **neither** the base views (`vwActionExecutionLogs`, `vwEntityActionFilters`, `vwEntityActionInvocations`, `vwEntityActionParams`) **nor** the `EntityField` metadata has an `EntityAction` column — but `packages/MJCoreEntities/src/generated/entity_subclasses.ts` on `next` declares one for all four entities. So a generated artifact on `next` is ahead of what the migrations actually produce.

Flagging rather than fixing here — it wants its own investigation, and "just run CodeGen" would *delete* those keys from the generated file rather than create the columns.

PR #3661, opened 2026-08-08
docs: one database per agent — a worktree isolates the filesystem, not the DB

AN-BC:
Policy change requested by @AN-BC after a real incident today.

## What happened

I ran a full `mj codegen` against a database another session was also using. It regenerated the three `vwEntityAction*` views **without** the denormalized `EntityAction` column — `MJ: Entity Actions` has no field flagged `IsNameField`, so there was nothing to derive it from — while the `EntityField` rows demanding that column survived.

The next server boot logged **975** `Invalid column name 'EntityAction'` errors and `EntityActionEngineBase` refused to load, so entity actions were dead. Five minutes later an unrelated `mj migrate` from the other session incidentally repaired it.

**Every command involved reported success.** Neither agent's terminal showed a problem; the errors surfaced in a third place — a running server — minutes later. That is what makes this worth a written rule rather than a lesson learned.

Attribution, from `sys.objects.modify_date` and `flyway_schema_history` (all UTC):

| Time | Event | Who |
|---|---|---|
| 18:27–18:29 | full `mj codegen`, 374 entities | me |
| 18:39 | server boot → 975 errors | me |
| 18:44:32 | `ModelConfiguration` migration + `R__RefreshMetadata` | other session |

The errors **predate** the other session's run, so it did not cause them — it healed them.

## The two rules

1. **One database per agent.** Confirm no other session is using your `DB_DATABASE` before `mj migrate` / `mj codegen` / `mj sync push`. **A git worktree isolates the filesystem, not the database** — which is precisely what makes it feel safe. Two agents, two directories, one `DB_DATABASE`, and as far as the schema is concerned that is one agent.
2. **Don't run a full `mj codegen` for a change with no schema DDL.** The work here needed three TypeScript interfaces for a new Remote Operation. `mj codegen --skipdb` emits those and touches no view; instead 374 entities' views were regenerated against a live shared database to get them.

Plus the shared-migration-state corollary: another session's `mj migrate` moves your schema forward silently, so a build that passed an hour ago may no longer match the database it is talking to. That is how this session ended up running against a schema it had never migrated.

## Placement

Body in `migrations/CLAUDE.md`, immediately before the existing **CodeGen Ordering** rule, since that file already owns migrate/codegen execution. One pointer line in the root `CLAUDE.md`, which is budget-gated — matching how the JSONType ordering rule is already split.

`npm run check:claude-md` passes: 148 inbound refs, links + routing valid.

## Not fixed here

The underlying defect is still open. **43 entities currently have no `IsNameField`**, which is why the denormalized column had nothing to derive from. PR #3651 — *"stop Smart Field Identification clearing IsNameField"* — is exactly that, and until it lands any full CodeGen run can reproduce this on a different entity.

/cc @AN-BC


AN-BC:
Added the skill to this PR, per @AN-BC — a policy that's tedious to follow gets skipped, so this is the tool that makes complying easier than borrowing whatever is in `.env`.

**`.claude/skills/bootstrap-clean-db`** — builds a database from scratch on a private name, then verifies the branch against it. Registered in the root routing table; `check:claude-md` passes.

It's useful for two different reasons, and they're worth separating:

- **Isolation** — the cheap way to get a database no other agent is using, which is what the policy in the first commit asks for.
- **Verification** — a long-lived dev DB accumulates state from every session that touched it, and that state *hides* defects. The canonical one: a migration looks complete locally because your database already had the `EntityField` row it forgot to create, then fails CI at the `sync push` step long after the PR looked green. A from-scratch build is the only thing that proves the repo can produce a working system from what it actually ships.

It encodes the four-step ordering **with the reason each step is where it is** — notably that step 2 must be `--skipfiles`, because a full CodeGen against a freshly-migrated database regenerates `remote_operations.ts` from an empty metadata set, deletes every operation class, and bricks the CLI that step 3 needs.

It also carries the verification mistakes I made today, written as rules rather than anecdotes:

- **Check build exit codes, not `grep "error TS"`** — `tsc` writes ANSI colour codes between those two words, so the pattern matches nothing on a log full of errors. I reported a clean build that wasn't.
- **`EADDRINUSE` is not evidence your server is running**, and a one-off `200`/`401` proves a socket answered once, not that a process is alive. I handed over a link to an app with no API behind it.
- **`git show HEAD:<path> > <path>` truncates before `git` runs** — on an untracked file that leaves an empty file and no error. I emptied two files that way.
- **Run the tier with `RUN_MUTATION_TESTS=1`** and report real counts; mutation bundles skip silently without it.

One deliberate choice worth flagging for review: step 3 is *"read the diff, do not just discard it."* A clean bootstrap's regenerated diff is the repo telling you what your working database had been hiding — blanket-reverting it throws away the most valuable output of the exercise. That's the step I'd expect to be skipped in practice, so it's argued rather than instructed.

PR #3670, opened 2026-08-08
fix(codegen+migrations): two defects that only appear on a from-scratch database

AN-BC:
Three commits, all found by building a MemberJunction database from nothing but `migrations/` + `metadata/`. Each defect was invisible on a long-lived database — two of them were *actively masked* by drift.

## 1. `V202608081200` could never run on a clean database

Its appended CodeGen block hardcodes `Sequence = 100025` for `EntityAction.RunMode` — the number free on the database CodeGen ran against. On a from-scratch database that belongs to `EntityAction.ScopeEntity`.

**The error Flyway reports is not the error:**

| # | What happens | Result |
|---|---|---|
| 1 | `RunMode` insert, `Sequence = 100025` | Violates `UQ_EntityField_EntityID_Sequence` |
| 2 | No `SET XACT_ABORT ON` | Aborts the **statement**; execution continues |
| 3 | `Task.ActionID` insert | Succeeds |
| 4 | `EntityFieldValue` rows reference the field step 1 never created | **`FK_EntityFieldValue_EntityField`** — the only error surfaced |

Chasing the reported FK error fixes nothing.

## 2. CodeGen emitted literal sequences (the general case)

The number is a temporary placeholder (`MAX + 100000 + ordinal`) renumbered by `spUpdateExistingEntityFieldsFromSchema` — run live, and from the **repeatable** `R__RefreshMetadata.sql`. Flyway runs **every versioned migration before any repeatable script**, so on a fresh database the renumber never runs in between, and two migrations touching the same entity collide. Structural, not luck.

CodeGen now emits an apply-time expression whose offset is the field's **schema ordinal**, so ordering is encoded in the value rather than depending on statement execution order:

```sql
(SELECT COALESCE(MAX([Sequence]), 0) FROM [...].[EntityField] WHERE [EntityID] = '<id>') + <ordinal>
```

Guard rail with its own fixtures: `.github/scripts/check-migration-entityfield-sequence.sh [--self-test]`. It scans only lines a PR **adds**; the 238 committed files using the old form apply cleanly and are left alone.

## 3. Four entities requested a name column that cannot exist

`MJ: Action Execution Logs`, `Entity Action Filters`, `Entity Action Invocations`, `Entity Action Params` all set `IncludeRelatedEntityNameFieldInBaseView` on their `EntityActionID` FK. The target, `MJ: Entity Actions`, is a pure junction with **no name field**, so CodeGen creates the metadata row and then cannot emit the column → `Invalid column name 'EntityAction'`.

Verified on a database with migrations **and** a full CodeGen run: column absent from all four views, zero `IsNameField` rows on the target, no `Name` column. CodeGen declines and logs an integrity failure — it does not "just add it".

**IT50 had an oracle for this and was passing anyway**, because a prior CodeGen run had deleted the orphan rows from the shared database. The evidence was cleaned up, not the defect.

## Verification

Two independent from-scratch databases:

- migrations green (40 applied, **0 failures**)
- `mj sync push` — 13,669 records, **0 errors**
- build **299/299 + 271/271**, exit 0
- deterministic tier **61/61** with `RUN_MUTATION_TESTS=1` (was 60/61)
- identical sequences on both builds, **0** base-after-virtual violations
- a second CodeGen pass produced **no output** — converged and idempotent

## Release status

**Not in any release.** `v6.1.0-edge.1` was cut before this merged; the file is absent from that tag's tree, npm `latest` is `5.51.0`.

⚠️ **Every database that already applied `V202608081200` needs a Flyway repair** — the checksum changes. That cost stays small only while this is unreleased.

@ms-bc — flagging for release awareness.

## Known, not fixed here

Regenerating against a clean database also drops `MJTaskEntity.ValidateMutuallyExclusiveAssociations`, because CHECK-constraint validators are **LLM-authored and stored in the database** (`__mj.GeneratedCode`). Restored by hand so this diff contains only the intended change — but it means **generated code in this repo is not reproducible from this repo**. Worth its own issue.


AN-BC:
@rkihm-BC @MS-BC @MarceloT-BC fyi for all of you

cadam11:
Confirming your fix cleared the fresh-DB migrate step (no more V202608081200 failure). Two reds remain on next's backstops after this merged, both from today's task-graph/D14 work, so flagging them while you are in the area:

**1. MJGlobal UUID-compliance sweep (Unit Tests red)** — 5 new raw comparisons, mechanical `UUIDsEqual()` fixes:

```
MJServer/src/services/TaskGraphActionRunner.ts:21           a.ID === params.ActionID
TestingFramework/.../entity-actions.checks.ts:387           f.ID === Filter.ID
TestingFramework/.../entity-actions.checks.ts:456           r.EntityActionID === binding.ID
TestingFramework/.../entity-actions.checks.ts:462           request.RecordID.includes(list.ID)
TestingFramework/.../entity-actions.checks.ts:505           r.EntityActionID === binding.ID
```

**2. Integration Tier run step** (migrate is green now): `User it-rls-b@integration.test does not have read permissions on MJ: Action Categories` (looks like the new entity-actions checks need permission seeding for the RLS test user) and `No suitable model found for prompt SQL Query Parameter Extraction ... no API credentials` (a new check invoking a prompt the CI env has no credentials for). Also noisy but possibly tolerated: `MJ_BASE_ENCRYPTION_KEY is not set` from EncryptionStartupValidator.

Happy to take the 5 UUIDsEqual sites if you want them off your plate, say the word.

PR #3676, opened 2026-08-08
fix(core): run an overridden ValidateAsync instead of silently skipping it

AN-BC:
## The bug

`BaseEntity.ValidateAsync` documents itself as *"automatically called by `Save()` AFTER the synchronous `Validate()` passes."* It was not.

`Save()` reached `ValidateAsync` only if the subclass **also** overrode `DefaultSkipAsyncValidation` — a separate getter the `ValidateAsync` docstring never mentioned, and which defaults to `true`. An override written against the documentation alone silently never ran. It reads as enforced, it reviews as enforced, and it is not.

```ts
// before
const skipAsyncValidation = _options.SkipAsyncValidation !== undefined ?
    _options.SkipAsyncValidation : this.DefaultSkipAsyncValidation;  // => true, for anyone who never found the getter
```

## Why the default cannot be justified on cost

The base `ValidateAsync` returns success immediately. Skipping it saves a class that did not override it essentially nothing — there is no query, no I/O, no work to avoid.

So the flag's only *reachable* effect was to disable the async rules of exactly those classes that had bothered to **write** async rules. The one thing it could do was the one thing nobody wanted it to do.

## Precedent already in this file

This is not a hypothetical failure. `OrderEntityServer.ValidateAsync` — holding both the "cannot confirm an order with no lines" guard and an entire per-line validation loop — was dead on every save in production for exactly this reason.

That defect is **already cited in `baseEntity.ts`**, one branch below the code changed here, as the justification for companion validation ignoring this same flag. This PR applies the same reasoning to the entity's own rules, which is where the defect actually originated.

## The fix

Overriding `ValidateAsync` is now what turns it on. Precedence, in order of authority:

1. **`EntitySaveOptions.SkipAsyncValidation`**, when set, wins outright. Per-call control is unchanged.
2. **An explicit `DefaultSkipAsyncValidation` override** wins next — **either value**. Stating a policy still beats inferring one, so a class that deliberately opts out keeps opting out.
3. **Only when no policy is stated at all** is the answer inferred: run `ValidateAsync` if a subclass overrode it, skip it if none did.

Rule 2 is what makes rule 3 safe, and it requires telling *"chose `true`"* apart from *"never made a choice"* — a distinction a getter cannot express, since both return the identical `true`. The two are separated by comparing property descriptors against `BaseEntity.prototype`:

- cached per class in a `WeakMap` (the answer is a property of the class, not the instance — a save path should not walk a prototype chain per record);
- handling **both accessors (`get`) and methods (`value`)**, so the same helper serves `DefaultSkipAsyncValidation` and `ValidateAsync` alike;
- walking the whole chain, so an override declared **anywhere in a multi-level hierarchy** counts — generated class, app subclass, server subclass on top of it.

The `ValidateAsync` and `DefaultSkipAsyncValidation` docstrings are also corrected. They described the old behaviour, and were how the trap was set.

## Blast radius — measured, not assumed

Every class in the repo overriding `ValidateAsync` was enumerated and checked:

| | |
|---|---|
| Classes overriding `ValidateAsync` | **17** |
| Already override `DefaultSkipAsyncValidation` → take rule 2, **behaviour unchanged** | **16** |
| Remaining | **1** — `RelatedRecordCollection`, whose `ValidateAsync(result)` is `EntityCompanion`'s method in a different hierarchy entirely |

**CodeGen never emits `ValidateAsync`**, so no generated class trips the inference. This was the check that mattered most: had CodeGen emitted even a harmless pass-through override, rule 3 would have switched async validation on **globally**, across every generated entity in every deployment. It does not, and the inference stays confined to hand-written overrides.

Net effect inside this repo: **zero behaviour change.**

## Testing

- **15 new tests** in `packages/MJCore/src/__tests__/baseEntity.validateAsync.test.ts`, driving the real `Save()` path rather than calling the resolution logic directly — covering each precedence rule, explicit `true` vs. never-set, accessor and method overrides, and multi-level chains.
- **Full MJCore suite green: 119 files, 1823 tests.**

## ⚠️ Risk for downstream adopters

**This is a real behaviour change outside this repo, and it is the intended one.**

A `ValidateAsync` that has been quietly dead in a downstream application **will start running** — and can start **refusing saves it was always written to refuse**. That is the point of the fix: those rules were authored to be enforced. But an application that has unknowingly been saving records past its own async rules will see saves begin to fail after upgrading, and the failures will look new even though the rules are old.

Recommended before adopting: grep your codebase for `ValidateAsync` overrides and confirm each one is a rule you still want enforced.

Escape hatches, if any case should stay off:

- pass **`SkipAsyncValidation: true`** in the save options (per call), or
- override **`DefaultSkipAsyncValidation`** to return `true` (per class) — rule 2 guarantees an explicit override still wins.

### DIFF
--- a/.changeset/is-member-overridden.md
+++ b/.changeset/is-member-overridden.md
@@ -1,0 +1,17 @@
+---
+"@memberjunction/global": minor
+---
+
+Add `IsMemberOverridden(instance, member, BaseClassRef)` to `ClassUtils`.
+
+Answers whether a subclass replaced a member somewhere between an instance and a given base class,
+which is what distinguishes "the author made no choice" from "the author chose the value that
+happens to be the default" — something a base-class member cannot express on its own. An API whose
+default sits in the *off* position therefore silently disables exactly the subclasses that most
+wanted it on.
+
+Handles methods and accessors alike by comparing property descriptors, finds an override declared
+anywhere in a multi-level chain, returns false rather than throwing on bad input, and caches per
+(class, member, base class) so a hot path does not re-walk a prototype chain.
+
+Used by `BaseEntity` to decide whether to run an overridden `ValidateAsync`.
--- a/.changeset/validateasync-silently-skipped.md
+++ b/.changeset/validateasync-silently-skipped.md
@@ -1,0 +1,41 @@
+---
+"@memberjunction/core": patch
+---
+
+Fix `ValidateAsync` overrides being silently skipped on save.
+
+`BaseEntity.ValidateAsync` documents itself as *"automatically called by Save() AFTER the
+synchronous Validate() passes"*. It was not. `Save()` reached it only when a subclass **also**
+overrode `DefaultSkipAsyncValidation` — a separate getter the docstring never mentioned, defaulting
+to `true`. An override written against the documentation alone never ran: it reads as enforced,
+reviews as enforced, and is not.
+
+The default cannot be justified on cost. The base `ValidateAsync` returns success immediately, so
+skipping it saves a subclass that did not override it essentially nothing. The flag's only reachable
+effect was therefore to disable the async rules of subclasses that had *written* async rules — which
+is how `OrderEntityServer.ValidateAsync`, holding both the "cannot confirm an order with no lines"
+guard and an entire per-line validation loop, was dead on every save in production. It is the same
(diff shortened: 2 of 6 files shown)


AN-BC (commented):
LGTM

PR #3678, opened 2026-08-08
fix(workflows): stop the canvas re-zooming on every edit; keep node positions

AN-BC:
Two of the three reported canvas problems trace to **one cause**, and it's a structural one.

## Root cause

`TaskGraphSpec` is an execution contract with **no layout field**. So `SpecToNodes` reset every node to `{0,0}` on every projection, and the component compensated by re-running auto-layout whenever the topology changed — and `AutoArrange` ends in:

```ts
setTimeout(() => this.ZoomToFit(), 100);
```

So **every added step and every drawn connection** re-arranged the graph and snapped the viewport to fit. On a one-node graph, "fit" is maximum zoom — the *"it super zooms into the step right away"* report. It also meant an author's own arrangement was thrown away by their next edit.

**The Flow Agent editor has never had this problem.** It arranges only when *every* node sits at the origin (`flow-agent-editor.component.ts:244-245`) and stores what the canvas hands back via `NodesChanged`/`ConnectionsChanged`. This adopts that rule rather than inventing a second one.

## Changes

- `SpecToNodes` takes an optional positions map. Unknown ids still fall back to the origin — the correct first-open state. Positions are **copied, not aliased**: the canvas mutates `node.Position` in place while dragging, so sharing the reference would write back into the map and defeat the next projection's comparison.
- The component owns that map — seeded from `NodesChanged`, `NodeMoved`, and the drop position of a newly added node, cleared when the spec is replaced.
- **Auto-layout runs once**, only for a graph that arrives with no geometry. Adding a node by hand marks the graph as laid out, so the deferred Dagre pass can't fire later and discard that placement.

## On the "can't connect steps" report

Not a separate defect as far as I can tell. Nodes piled up at the origin between projections, and the viewport sat at fit-one-node zoom — so the second step's ports were routinely off-screen. The connection plumbing itself is correct: `SpecToConnections` projects `dependsOn` with `out`→`in` ports, `OnConnectionCreated` → `AddDependency` commits, and both are bound in the template. **Worth re-testing specifically.**

## Verification

| Check | Result |
|---|---|
| `ng-task-graph-editor` tests | **138 pass** (4 new) |
| `ng-task-graph-editor` build | exit 0 |
| `ng-dashboards` build | exit 0 |
| `npm run check:ui` | 0 violations |
| `mj standards check` | all adopted standards pass |

The four new tests pin the geometry carrier: honoured position, origin fallback, copy-not-alias, and no-positions-supplied.

## Two honest limitations

**Not verified in a live browser.** The cached Playwright auth profile has an expired JWT (`JWT_EXPIRED`) and MSAL needs a human, so this is reasoned from source and covered by unit tests — not observed.

**The vanishing-palette-entry report is NOT fixed here.** The palette renders unconditionally from a static readonly array that nothing mutates, entries are tracked by object identity, and nothing filters by usage. I could not reproduce or explain it from the source, and I'd rather say so than ship a speculative change. It needs one more detail to pin down (whole palette vs. one entry; does a reload restore it).

### DIFF
--- a/packages/Angular/Generic/task-graph-editor/src/__tests__/task-graph-canvas-adapter.test.ts
+++ b/packages/Angular/Generic/task-graph-editor/src/__tests__/task-graph-canvas-adapter.test.ts
@@ -54,6 +54,37 @@ const diamond = (): TaskGraphSpec => spec({
     ],
 });
 
+describe('SpecToNodes — geometry the spec cannot hold', () => {
+    // TaskGraphSpec is an execution contract with no layout field. Before this, every projection
+    // reset each node to the origin, so the component had to re-run auto-layout after EVERY edit —
+    // and auto-layout ends in zoom-to-fit, which is why adding a step yanked the viewport. These
+    // tests pin the carrier that makes arranging-once possible.
+    it('honours a caller-supplied position', () => {
+        const positions = new Map([['a', { X: 120, Y: 40 }]]);
+        const node = SpecToNodes(spec(), undefined, positions).find((n) => n.ID === 'a')!;
+        expect(node.Position).toEqual({ X: 120, Y: 40 });
+    });
+
+    it('falls back to the origin for a task it has never seen', () => {
+        const positions = new Map([['a', { X: 120, Y: 40 }]]);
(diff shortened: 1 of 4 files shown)


AN-BC (commented):
LGTM

PR #2648, opened 2026-05-21
fix(server): tier GraphQL variables logging — secrets masked, no double-log (#2638)

SDesai-BC:
## Summary

Closes the #2638 secret leak in MJAPI's GraphQL request logging, and resolves the senior-review feedback that the in-progress fix **double-logged** every request under `MJ_LOG_GRAPHQL_VARIABLES=true` — a value-blind structural-shape line *plus* a value-bearing middleware line that re-emitted plaintext PII/payloads for every non-`Encrypt` field.

The logging is now **tiered by the verbose flag, one line per tier**:

| Tier | Output |
|---|---|
| **Default / production** | operation name only — no `variables`, no values (the load-bearing #2638 leak fix) |
| **Verbose (`MJ_LOG_GRAPHQL_VARIABLES=true`)** | per-resolver line with values **fully emitted**, *except* `Encrypt=true` fields which stay `<redacted>` even in verbose |

This satisfies both reviewers: the genuine-secret floor (encrypted fields) never prints, while developers retain fully-emitted real values for debug sessions.

## What changed

- **Deleted the value-blind structural-shape line** (`logging/structuralShape.ts` + its 41 tests). The double-logging is gone with it.
- **`context.ts` boundary line** → `buildBoundaryLogPayload(operationName)` = operation-name only. New pure, isolation-testable `logging/boundaryLogPayload.ts` (no MJ imports).
- **Kept the value-bearing middleware** (`variablesLoggingMiddleware`, `secretRedactor`, `@NoLog`, `bootAudit`) as the **sole** verbose values emitter.
- **`secretRedactor` + `bootAudit` regex** now includes `Delete` (`^(Create|Update|Delete).+Input$`) — stops the verbose boot-audit from flooding hundreds of false positives on `DeleteMJ*` resolvers. Security unchanged (delete inputs carry PK + Options only).
- **`secretRedactor` now honors field-level `@NoLog` on non-entity-bound inputs** — previously ignored, so `GetDataInputType.Token` would have leaked. Now walks top-level keys whenever a redaction source exists (encrypted fields **or** `@NoLog` fields).
- **`GetDataResolver` / `MCPResolver`**: stale/step-trace `LogStatus` → `LogStatusEx({verboseOnly})`; `Token` marked `@NoLog`; MCP tool-args gated behind `logVariables`.

## Behavior change for operators

- The **"no literal value under any flag" invariant is retracted** — it was only ever true of the deleted shape line. Verbose mode emits real values by design.
[...]

PR #3686, opened 2026-08-09
fix(server): GraphQL variables redaction fails open for hand-written input types

MS-BC:
## Summary

The GraphQL variables-logging redactor introduced in #2648 binds an input type to an entity **by name**:

```ts
const match = /^(Create|Update|Delete)(?<name>.+)Input$/.exec(inputTypeName);
const entity = provider.Entities.find((e) => e.ClassName === match.groups.name);
```

Codegen inputs embed the entity `ClassName` — `CreateMJCredentialInput` captures `MJCredential`, the lookup succeeds, and `Encrypt=true` columns are redacted. That covers the 646 generated CRUD inputs and works exactly as designed.

Hand-written resolver inputs do not follow that shape. `CreateConnectionInput` captures `Connection`, which matches no entity, so no encrypted-field names are contributed. Absent any `@NoLog`, `canWalk` is `false` and the argument falls through to `shortenForLog` — which truncates whole-object JSON at 300 characters but returns **string leaves untruncated**.

The boot audit uses the same regex, but tests only the *name pattern* — it never checks whether the lookup succeeds. So an input like this is classified as metadata-bound and never warned about, while the redactor quietly finds nothing to redact. Audit-silent and redaction-free at the same time.

## Affected inputs

Every hand-written `Create*/Update*/Delete*Input` in the resolvers reaches this path:

`CreateConnectionInput`, `CreateAPIKeyInput`, `CreateUploadURLInput`, `CreatePreAuthDownloadUrlInput`, `CreatePreAuthUploadUrlInput`, `DeleteStorageObjectInput`, `CreateDirectoryInput`, `CreateScheduleInput`, `CreateQuerySystemUserInput`, `UpdateQuerySystemUserInput`, `CreateVersionLabelInput`

I checked each. Ten carry only identifiers, labels, paths and cron expressions — `CreateAPIKeyInput` in particular mints its key server-side and never receives one. **One carries a secret:** `CreateConnectionInput.CredentialValues`.

That field does not map to an entity column — the resolver assigns it onto `MJ: Credentials`.`Values` (which *is* `Encrypt=true`) in procedural code further down — so the metadata-driven half of the redactor structurally cannot see it. `@NoLog` is precisely the escape hatch documented for this case, and it was missing; the resolver did not even import it.

## Changes

1. **`CreateConnectionInput.CredentialValues` is marked `@NoLog`**, with a comment explaining why the mark is load-bearing rather than decorative.
2. **The redactor now reports an unbound CRUD-named input**, once per input type, naming the type and the remedy — instead of silently falling through to `shortenForLog`.

Warning rather than failing closed is a deliberate choice: failing closed would blank the values of the ten inputs that carry nothing sensitive, which is a real debuggability cost in the mode whose entire purpose is seeing values. The warning makes the gap visible to the operator who turned verbose logging on, which is who needs to know.

## Scope

Both changes are confined to verbose mode (`loggingSettings.graphql.logVariables`), which remains **off by default**. There is no behaviour change in the default configuration.

## Test plan

- [x] 8 new tests: the entity-bound path emits no warning; the unbound path is documented as fail-open and now warns, once per type, separately per type; `@NoLog` closes what metadata cannot; a parameter-level mark redacts the whole arg
- [x] The warning text itself is asserted not to contain the value it is warning about
- [x] **821 MJServer tests green**, 0 failed

## Open question for reviewers

Should the **boot audit** itself perform the entity lookup, rather than trusting the name pattern? That would catch all eleven inputs at startup instead of at first use. I left it out deliberately — the audit runs during schema construction and it is not obvious that entity metadata is reliably available there, whereas the redactor runs at request time where it demonstrably is. Happy to take it further if a maintainer knows the boot ordering well enough to say.

## Related

- #3642 — the client-side counterpart. Same credential, different process: that PR stops the browser's transport from logging it, this one stops the server's variables logging from doing so. Neither covers the other.


MS-BC:
@AN-BC I am pinging for review since this is an MJServer change, also seems related to something you mentioned earlier if I recall correctly.

AN-BC (approved):

PR #3687, opened 2026-08-09
fix(codegen): schema-named bucket apps are hidden from new users by d…

MarceloT-BC:
…efault

Application.DefaultForNewUser defaults to 1 in the DB and createNewApplication's INSERT omitted the column, so every auto-created schema bucket app (Name = '__mj_MySchema', Description = 'Generated for schema') landed visible in each new user's app switcher — while UI apps' human-authored metadata Application often ships hidden. The bucket exists to carry entity links, role grants, and SchemaAutoAddNewEntities; it is plumbing, not a product.

Repro (before): fresh DB → codegen over a new schema → the raw schema-named app appears in a new user's switcher. After: same flow, bucket app created with DefaultForNewUser = 0 and absent from the switcher. Affects newly emitted captures only; existing captured baselines keep their old INSERT until recaptured.

Found while rebuilding bizapps-accounting's baseline; discussion: https://github.com/MemberJunction/MJ/discussions/3683#discussioncomment-17954934

### DIFF
--- a/.changeset/codegen-bucket-app-hidden-by-default.md
+++ b/.changeset/codegen-bucket-app-hidden-by-default.md
@@ -1,0 +1,11 @@
+---
+"@memberjunction/codegen-lib": patch
+---
+
+Auto-created schema "bucket" Applications are now hidden from new users. `createNewApplication`'s
+INSERT omitted `DefaultForNewUser`, so the DB default of `1` won — every schema-named bucket app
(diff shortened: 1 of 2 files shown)


AN-BC (approved):

AN-BC:
Approved. @MattC-BC  flagging for you to be aware we are hiding these by default for new users. Any user can add these of course. I do think we need to think about auto generating a good dashboard for every new app that comes in as an option in codegen as a future feature too?

PR #3688, opened 2026-08-09
fix(testing-integration): Integration Tier fails repo-wide on a missing MJ_API_KEY that should skip

MS-BC:
## Summary

The Integration Tier gate has been failing on **every PR, and on `next` itself** — 8+ consecutive red runs. It reports `40/61 passed (65.6%)` and exits non-zero, with every client-transport bundle failing in 0.1s at bootstrap:

```
Bootstrap failed: MJ_API_KEY is not set in the environment — required for client-side tests.
```

Nothing ever ran. This is not a product defect and no PR can fix it from its own branch.

## Root cause

A client-transport bundle bootstraps in two steps, and either can throw:

| | Throws | When |
|---|---|---|
| 1 | `MJ_API_KEY is not set …` (`LoadClientConfig`) | key absent — **before any network call** |
| 2 | `MJAPI is not reachable …` (`preflightMJAPI`) | server absent — only reached once a key exists |

`IntegrationTestDriver` skips-as-pass when the environment simply has no client tier, and hard-errors otherwise. It matched **only message (2)**:

```ts
if (transport === 'client' && /MJAPI is not reachable/i.test(msg)) {
```

The PR gate provisions a throwaway SQL Server and runs the suite in-process with **neither an MJAPI nor a key** — so it fails at (1), the regex never matches, and it falls through to `buildErrorResult`. For a condition `integration.yml` explicitly designed around, in its own words:

> No MJAPI is up, so client-transport members skip on the MJAPI preflight.

The skip contract was written for exactly this environment and just never fired in it.

## Fix

Both messages now count as *"there is no client tier here"*, behind one named constant so the driver and its tests cannot drift apart.

**Still hard errors** — these describe an environment that *has* a client tier and is misconfigured, which is precisely what the gate exists to catch:

- `MJAPI at … answered HTTP 401` — key present but rejected
- any non-OK response from a reachable server
- cache-ownership and configuration failures

## Why not just add the secret to the workflow

There is **no `MJ_API_KEY` secret on the repository** (checked against the repo's Actions secrets), so that was not available. It would also have been the wrong fix: the gate deliberately runs no MJAPI, so a key alone would just move the failure from step 1 to step 2. This restores documented behaviour rather than changing what CI provisions.

## Test plan

- [x] 8 new tests pinning both halves of the boundary — what must skip, what must stay a hard error
- [x] Includes the case a looser pattern would break: the 401 message **also** contains `MJ_API_KEY`, so the match keys on `is not set` rather than on the variable name
- [x] 78 tests green in `@memberjunction/testing-integration`, typecheck clean
- [x] Verified the failure is repo-wide, not branch-specific — `next` shows 8 consecutive Integration Tier failures with the identical bootstrap error

## Expected effect

The tier should return to green with client-transport bundles reported as **skipped-with-reason** rather than failed — so "green" once again means *everything that could run, ran*, and a red Integration Tier becomes meaningful again.


MS-BC:
@AN-BC @CaelebB-BC Does this look a good change? My Integration tests are failing and this seems straightforward fix for solely client-side tests.

CaelebB-BC:
@MS-BC We are going to be booting MJAPI inside of CICD soon so I will have this fixed when that is done.

MS-BC:
> @MS-BC We are going to be booting MJAPI inside of CICD soon so I will have this fixed when that is done.

Ok, I will close it for now

Issue #3604, opened 2026-08-07
CodeGen PostgreSQL identifier auto-quoting is a keyword denylist — audit for other silent keyword-collision failures


rkihm-BC:
## Summary

CodeGen's PostgreSQL identifier auto-quoting (`PostgreSQLCodeGenProvider.quoteSQLForExecution` → `processWord`) is a **denylist heuristic**: it auto-quotes PascalCase identifiers *unless* the word (upper-cased) is in the static `_SQL_KEYWORDS` set. Any SQL that references a column whose name collides with that keyword set is passed through **unquoted**, folds to lowercase on PostgreSQL, and either throws `column "..." does not exist` or silently binds the wrong thing. SQL Server is unaffected (case-insensitive resolution), so these defects are invisible on the dialect most authoring happens against.

This was the root cause of one of the two defects fixed in #3590 (`Length` collided with `LENGTH` in `_SQL_KEYWORDS`, so `manageSingleEntityParentFields` threw before mirroring any IS-A field). That PR patched the two IS-A instances. This issue tracks the **systemic** problem, of which those were just two instances.

## Why this matters

- **Silent on the dialect people develop against.** T-SQL-first authoring means these ship undetected and only surface on a live PostgreSQL deployment.
- **Latent across the whole codegen SQL surface.** Any current or future query referencing a keyword-colliding column is exposed. Candidate collision names present in `_SQL_KEYWORDS` that are plausible MJ column names include: `Length`, `Position`, `Comment`, `Level`(via none? verify), `Language`, `Action`, `Order`, `Key`, `Default`, `Check`, `Column`, `Table`, `View`, `Index`, `Source`/`Target` (verify), etc. The exact exposed set = (MJ column names) ∩ `_SQL_KEYWORDS`.

## Proposed work

1. **Audit** `packages/CodeGenLib/src/Database/manage-metadata.ts` and the codegen SQL-generation paths for raw SQL strings that reference entity/metadata columns **unquoted** (i.e. relying on the tokenizer rather than `qi()`/`qs()`). Cross-reference column names against `_SQL_KEYWORDS`.
2. **Fix** by wrapping identifiers in `qi()` explicitly — the robust pattern established by #3590 — rather than trusting the denylist.
3. **Guard going forward.** Options to consider:
   - A lint/test that scans codegen SQL string literals for unquoted mixed-case identifiers colliding with `_SQL_KEYWORDS`.
[...]

PR #3691, opened 2026-08-09
fix(codegen): stop deciding PostgreSQL quoting from a keyword denylist (#3604 item 4)

MS-BC:
Addresses **item 4 of #3604** — the design fix, not another instance fix.

## The problem with the denylist

`quoteSQLForExecution` quoted a PascalCase word *unless* it appeared in a 288-entry `_SQL_KEYWORDS` set. That is a denylist, and it is wrong by construction here: **the set of SQL keywords and the set of MJ column names overlap.** Every name in the intersection was emitted unquoted, folded to lower case on PostgreSQL, and failed with `column "..." does not exist`.

The intersection is exactly eight names, computed by intersecting the 1,413 columns in the PostgreSQL baseline DDL against the keyword set:

```
Action, Columns, Language, Length, Month, Rank, Text, Values
```

Three things make this a design problem rather than a bug list:

1. **It is not guessable.** #3604 proposed `Position`, `Comment`, `Order`, `Key`, `Default`, `Table`, `View`, `Index` — none are exposed. The first guard I wrote listed `Count`, `Format`, `Date`, `Left`, `Right` — all real keywords, none of them MJ columns. Two attempts, two wrong sets.
2. **`Values` is the field-level-encrypted column** on `__mj."Credential"` — the platform's only encrypted column.
3. **It grows silently.** Add a column named `Rank` to any new entity and the hole reopens with nothing to signal it. A pinned list of today's intersection does not help.

## The replacement

Key on properties intrinsic to the two things instead of on a list. Generated SQL writes keywords, types and functions in ALL CAPS; MJ identifiers are mixed case. A word immediately followed by `(` is a function call whatever its casing.

| Shape | Treated as | Quoted? |
|---|---|---|
| ALL CAPS **and** in the keyword set | keyword / type | no — `SELECT`, `VALUES`, `TEXT` in a cast |
| immediately followed by `(` | function call | no — `LENGTH(`, `Coalesce(` |
| all lowercase, or `__mj_` prefixed | unchanged | no |
| **anything else mixed case** | identifier | **yes** — `Length`, `Values`, `Text` |

The denylist survives but is consulted for **ALL-CAPS words only** — the narrow band where it is unambiguous, since MJ columns are PascalCase. Keeping it there also leaves somewhere to express a future exception without restoring a deleted 288-entry literal.

## The carve-out that cuts both ways

My first draft used a pure case rule and **broke `ID`**. An all-caps word that is *not* a keyword is still an identifier — acronym columns (`ID`, `URL`) are all-caps by nature, and a pure case rule silently stopped quoting them, folding `ID` to `id`. The `!(isAllUpper && isKeyword)` form is what makes both directions correct, and there are tests for it.

## Verified before switching

Two shapes would break a case-based rule; both were checked against this codebase:

- `VALUES (` — written ALL CAPS at every call site (10 of them)
- mixed-case type casts (`AS Text`, `AS Date`) — none exist

## Test plan

- [x] `pg-quoting-case-rule.test.ts` — 60 tests driving the **real** `quoteSQLForExecution`
- [x] All eight colliding columns quote correctly in SELECT, SET and WHERE positions
- [x] Keywords, functions, casts, lowercase identifiers, `__mj_` internals, string literals, pre-quoted identifiers and `@parameters` all untouched
- [x] `SELECT Length, LENGTH(Name)` — the same word in both meanings, resolved in one statement with no list involved
- [x] Acronym columns (`ID`, `URL`) still quote — the regression my first draft introduced
- [x] Hypothetical future columns (`Position`, `Comment`, `Order`, `Key`, `Level`, `Source`) quote, proving the exposure cannot silently regrow
- [x] **811 passed, 60 skipped** across CodeGenLib; `tsc --noEmit` clean

## Relationship to the other PRs

Independent of #3590 and can land in either order. #3590 fixes three call sites by quoting them explicitly with `qi()`; this removes the reason those sites were broken in the first place. Both are worth having — explicit `qi()` at the call site is still the clearer intent, and this is the safety net for every site that does not.

## Reviewer note

This is a core tokenizer that every CodeGen SQL statement passes through on PostgreSQL, so the blast radius is the whole PG codegen surface. The existing suite (811 tests) passes unchanged, but an end-to-end run against a live PostgreSQL instance is the verification that actually matters here, and it is next on my list.

### DIFF
--- a/.changeset/pg-quoting-invert-denylist.md
+++ b/.changeset/pg-quoting-invert-denylist.md
@@ -1,0 +1,17 @@
+---
+"@memberjunction/codegen-lib": patch
+---
+
+Replace the PostgreSQL identifier-quoting denylist with a discriminator that cannot fall out of date.
+
+`quoteSQLForExecution` decided whether to quote a PascalCase word by checking it against a 288-entry `_SQL_KEYWORDS` set. That is a denylist, and it is wrong by construction here, because the set of SQL keywords and the set of MJ column names overlap. Every name in the intersection was emitted unquoted, folded to lower case on PostgreSQL, and failed with `column "..." does not exist`.
+
+The intersection is exactly eight names — `Action`, `Columns`, `Language`, `Length`, `Month`, `Rank`, `Text`, `Values` — obtained by intersecting the 1,413 columns in the PostgreSQL baseline DDL with the keyword set. It is not guessable: neither the report in #3604 nor the first fix predicted it correctly by inspection, and `Values` is the field-level-encrypted column on `__mj."Credential"`. It also grew silently — adding a column named `Rank` to any new entity re-opened the hole with nothing to signal it.
+
+Quoting now keys on properties intrinsic to the two things rather than on a list. Generated SQL writes keywords, types and functions in ALL CAPS (`SELECT`, `VALUES (`, `LENGTH(`, `INT`); MJ identifiers are mixed case (`Length`, `Values`, `EntityID`). A word immediately followed by `(` is a function call whatever its casing. The denylist is still consulted, but only for ALL-CAPS words — the narrow band where it is unambiguous, since MJ columns are PascalCase.
+
+The all-caps carve-out matters in both directions: an all-caps word that is *not* a keyword is still an identifier, so acronym columns such as `ID` and `URL` continue to be quoted. A pure case rule would have folded those to `id` / `url`.
+
+Verified against the codebase before switching: `VALUES (` is written all-caps at every call site, and there are no mixed-case type casts — the two shapes that would break the case rule. Adds `pg-quoting-case-rule.test.ts`, which drives the real `quoteSQLForExecution` and asserts both halves: all eight colliding columns now quote in SELECT, SET and WHERE positions, while keywords, functions, casts, lowercase identifiers, `__mj_` internals, string literals, pre-quoted identifiers and `@parameters` are all untouched. It includes `SELECT Length, LENGTH(Name)` — the same word in both meanings, resolved in one statement without a list.
+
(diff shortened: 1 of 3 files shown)


MS-BC:
Converting to draft — this is not ready to review and should not gate anything.

Two reasons:

1. **It is incomplete by its own logic.** There is a second copy of this tokenizer at runtime in `PostgreSQLDataProvider.autoQuoteIdentifiers`, with its own 311-entry list (vs 288 here) exposing 11 columns rather than 8 — including `Name`. That file's comment states the two must change together. Merging this alone deepens the divergence instead of fixing it.
2. **It has no end-to-end verification.** This tokenizer sits on every PostgreSQL statement CodeGen emits. 811 unit tests passing is necessary and nowhere near sufficient for that blast radius.

Parking until the IS-A work it came from (#3590) is finished and verified against a live PostgreSQL database. Analysis and the measured exposure are recorded on #3604.

MS-BC:
Closing — this is research output, not a change that should merge.

It covers CodeGen only, while the runtime twin in `PostgreSQLDataProvider.autoQuoteIdentifiers` has a larger exposure (11 columns vs 8, including `Name`). Landing half of it would deepen the divergence the runtime file's own comment warns about, and there is a dedicated review underway covering both layers.

Findings recorded on #3604 — the two-tokenizer divergence, the measured intersections, the in-repo precedent (`_SQL_KEYWORDS_UPPERCASE_ONLY`), and the `ID`/`URL` acronym trap that breaks the obvious generalization.

Branch `fix/pg-quoting-invert-denylist` is left in place: a working implementation plus 60 tests, available to reuse or discard once the design lands.

PR #12, opened 2024-02-28
Vector packages 2

JS-BC:

PR #3692, opened 2026-08-09
Workflows run on the durable task-graph dispatcher — one engine, one set of rules

AN-BC:
Flow agents now execute on the durable task-graph dispatcher. One traversal engine instead of two that drift, and a workflow that outlives the run that started it.

Verified end to end against a **from-scratch database**: 4 `Task` rows, the exclusive fork took one branch and marked the other `Skipped`, the `ForEach` ran its 5 iterations, 0 failed.

> **Omnibus branch.** Alongside the engine work described below, this branch also carries the parallel Flow-agent **authoring UI** effort — retiring the standalone Workflows app, making the Flow agent form first-class, and the single-editor canvas work. Those commits are self-describing; this description covers the execution engine, the schema, and the docs.

---

## The change in one picture

```mermaid
graph LR
    subgraph Before["Before — in-run walker"]
        B1["Agent run"] --> B2["walk step 1"]
        B2 --> B3["walk step 2"]
        B3 --> B4["walk step 3"]
        B4 --> B5["run ends"]
    end

    subgraph After["After — compile and dispatch"]
        A1["Agent run"] -->|compile| A2["TaskGraphSpec"]
        A2 -->|submit| A3["Task rows"]
        A1 -.->|"run ENDS here"| A4["handle returned"]
        A3 --> A5["Dispatcher<br/>claims · executes · settles"]
    end

    style B1 fill:#6b7785,stroke:#4a535d,color:#fff
    style A1 fill:#2d6a9f,stroke:#1a4971,color:#fff
    style A2 fill:#7c5295,stroke:#563a6b,color:#fff
    style A3 fill:#2d8659,stroke:#1a5c3a,color:#fff
    style A5 fill:#b8762f,stroke:#8a5722,color:#fff
```

A page reload, a server restart, or the submitting run ending no longer lose the work.

**The old walker is retained but refused.** It stays compiled as the reference implementation the compiler is checked against — deleting it would leave the compiler unfalsifiable — and throws at its single choke point (`createStepForFlowNode`) if anything reaches it. So a workflow that runs at all *provably* ran on the new engine.

---

## Seven silent defects found on the way

Each one produced a **green run that had done the wrong thing** — the failure mode this whole track exists to remove.

| # | Defect | What it looked like |
|---|---|---|
| 1 | `persistTasks` chose a task's nature by which config field was populated, falling through to "Human task assigned to the submitter" | A `ForEach` step became an approval request nobody wrote and nothing would ever complete — a graph stalled forever while *looking* like it was waiting on a person |
| 2 | Dependency edges never persisted `ExclusiveGroup`/`Priority`/`Sequence` | The dispatcher saw a plain fan-out and ran **every** branch of an exclusive choice |
| 3 | Input/output mappings dropped on persist | Branch conditions read `undefined` — falsy, not erroneous — so the workflow took the other branch and reported success |
| 4 | Loop bodies compiled with empty `params` | The repeated action ran N times with no inputs |
| 5 | Execution routed on `task.ActionID` | A loop carries one too (it's what the loop *repeats*), so a loop ran exactly **once** |
| 6 | `TaskGraphActionRunner` returned an `ActionParam[]` where every consumer addresses fields by name | Observed live: step 1 completed with an output payload of `{}` and every downstream branch skipped |
| 7 | Nodes had no coordinates | Every node in the run visualizer rendered on top of every other one |

Defects 3 and 6 are the same bug at two layers, and together they are why the *first* live run of the Demo workflow completed successfully having done nothing at all.

---

## Schema

Folded into the existing **unmerged** migration rather than added alongside it, so the release carries one Task migration rather than two.

`Task` gains `StepType`, `PromptID`, and `Configuration` — a JSONType (`ITaskStepConfiguration`) carrying kind-specific settings, the payload mappings, the execution policy and the author's canvas layout. `CK_Task_Assignment` now counts `PromptID`.

**Two columns and one typed bag rather than eight columns.** That paid off within the same day: adding `layout` later was an interface edit plus CodeGen, with no migration at all.

### Replay was proved, not assumed

The `R__RefreshMetadata` body is inlined into both post-edge Task migrations, between the DDL and the generated block — Flyway runs every versioned migration before any repeatable one, so on a replay the generated block would otherwise execute against unreconciled metadata.

A **migrate-only** build then established what the refresh does and does not do: it produced **zero** `EntityField` rows for the six new columns and **zero** value-list rows, while `vwTasks` did pick the columns up. That is what makes the appended CodeGen block required, and its ordering safe. Re-verified after: 6 EntityField rows, 7 value-list entries, `Skipped`, 3 view columns, 3 proc params — from migrations alone.

---

## Architecture

**Payload mapping lifted into `ai-core-plus`** so both engines share one dialect — the `*` wildcard, case-insensitive result lookup, `[]` append, `$message` fields, and the `static:`/`payload.`/`data.`/`context.` prefixes. Two implementations would have diverged precisely where the compile is meant to be lossless: a workflow would map differently depending on which engine ran it, invisibly, until someone's branch took the wrong path in production.

**Loops** run through a new `TaskLoopExecutor` — a loop stays one `Task` row and iterates internally, because how many times it runs is a property of the payload and expanding it at submission would mean guessing. `maxIterations: 0` means unlimited; parallel batches keep results in **iteration** order, since an output mapping writing `results[index]` assumes position means iteration number.

**Deterministic DAG layout** in `ai-core-plus`, because three callers must agree: the run visualizer (Task rows, no layout, ever), the editor's auto-arrange, and Save-as-Workflow, which writes positions server-side where no canvas exists. Layered by *longest* path — shortest-path would draw an edge backwards, which reads as a loop in a graph that has none.

### Cost: a lifetime problem, not a data-location one

```mermaid
sequenceDiagram
    participant R as Agent run
    participant G as Task graph
    participant D as Dispatcher

    R->>G: submit
    R->>R: calculateTokenStats() — nothing to count yet
    R-->>R: TotalCost final · run ENDS
    Note over G,D: minutes later, possibly another instance
    D->>G: execute · spend
    D->>R: credit TotalCostRollup
```

`BaseAgent` totals a run by walking its steps in memory **at finalization**, but a submitting run ends at submission — so the spending has not happened yet. That rules out replicating Task rows into `AIAgentRunStep` after the fact: it would mean reopening a `Completed` run and mutating totals after its guardrails had already evaluated against them.

So a settled graph credits itself back through the six `…Rollup` columns that have existed on `AIAgentRun` since v3 and that **nothing has ever written** — they exist for exactly this distinction. `TotalCost` keeps its current meaning (what the run itself spent, final, never rewritten), so nothing reading it today changes behaviour.

---

## Testing

| Tier | Result |
|---|---|
| `@memberjunction/ai-core-plus` | **489** passed |
| `@memberjunction/task-graph` | **60** passed |
| `@memberjunction/ai-agents` | **1845** passed |
| `integration-test-suite` (incl. both task-graph bundles) | **171** passed |
| Full repo build | **299/299**, exit 0 |
| Migration replay, from scratch | metadata complete from migrations alone |

**Live run on a from-scratch database**, NVDA at 223.96:

```
Step 1: Get NVIDIA Stock Price   Action    Complete   {"stockPrice":223.96}
Step 2(a): Get Weather           Action    Skipped    ← 223.96 > 500 is false
Step 2(b): Web Search            Action    Complete
Step 3: ForEach Loop Demo        ForEach   Complete
```

`ActionExecutionLog` shows **6** Google Custom Search rows — one for the branch plus five for the loop — which is the cheapest available proof the loop actually iterated.

---

## Docs

New **[Workflows and Task Graphs Guide](guides/WORKFLOW_AND_TASK_GRAPH_GUIDE.md)** — the guide this area did not have. It opens by untangling the three things people call "a workflow" (the definition, the contract, the run), then covers tool choice, the lifecycle, node kinds, exclusive groups, the payload dialect, loops, failure, a worked configuration of the shipped Demo agent, submitting from code, observability, layout, troubleshooting, and the checklist for adding a node kind. Six mermaid diagrams carry the concepts prose handles badly.

New **[`packages/TaskGraph/README.md`](packages/TaskGraph/README.md)** — the package had none. What durable execution is for, what the package deliberately does *not* do (it owns no graph semantics and never imports MJServer), the claim protocol as a guarded write, runners as seams.

`check:claude-md` passes — 49 guides all indexed, 693 links + 10 anchors valid.

---

## Deliberately not included

- **`ParentRunID` on graph-started runs.** Right for navigation, but several queries read `ParentRunID IS NULL` as "a conversational root" and those runs would stop matching. Probably correct — a dispatched sub-agent run isn't a conversational root — but a behaviour change with breadth that the rollup does not need.
- **`Prompt` and `External` node kinds.** No runner yet, so they are refused **by name** at submission rather than persisted as tasks that would wait forever.
- **`MaxCostPerRun` for graphs.** The ceiling is checked during a run that ends before the spending starts. That budget belongs on the graph, enforced by the dispatcher — its own decision.
- **The PostgreSQL counterpart**, per `migrations/CLAUDE.md`: deferred to the release build.

## Follow-up, designed and mocked

A [UX proposal](https://claude.ai/code/artifact/de7e22ec-3862-4416-b37b-a1de6ad8d6d2) covers the surfaces this makes possible: the graph as an expandable node in the Agent Run form, a Workflows app for run history (three of the five producers have no agent run, so their runs are visible nowhere today), the two-number cost display, and replacing the Flow agent's chat composer with a start-and-watch panel — you don't talk to a workflow.

## Known, pre-existing, and worth fixing separately

`pnpm run test:integration` cannot run on a from-scratch database: `metadata/test-suites/` contains only two suites and `metadata/tests/` no `IT##` records, so the suite named in the Definition of Done exists only where someone created it by hand. The check *code* is committed; the records that invoke it are not.


AN-BC:
# Intense review — PR #3692 vs. plan v2 (`plans/flow-agent-taskgraph-unification.md` @ `1c8550e44`)

**Verdict up front: not mergeable as-is.** The C1.1/C1.2 substance is genuinely strong — the compiler's five-rule ordering, the Skipped/XOR/HOLD pure layer, the terminality guard, the sweep-predicate fix, the fully-v2 RO wire, the cost-rollup design, and an orphan-free app retirement are careful, real work. But the PR **ships the C1.4 cutover with one of seven §8 gates closed**, silently reverses **three explicit plan rulings**, and carries **a dozen verifiable defects** — including one shipped core flow that can no longer run at all, caller input that never reaches the graph, and recovery paths that are dead machinery. Four independent review lanes (spec/compiler, dispatcher/runtime, calling contract, schema/wire/UI) verified everything below against the branch at `1c0d586d9`; file:line citations are to this PR's tree.

The clean split: **what this PR builds is largely right; what it wires up is not ready.** Recommended disposition at the end.

---

## A. Blocking defects (fix regardless of any ruling)

1. **The shipped `User Onboarding Flow Agent` is bricked.** It has 6 `Prompt` steps (`metadata/agents/.user-onboarding-agent.json:191,212,233,254,367,411`); the compiler emits `kind:'Prompt'` (`flow-graph-compiler.ts:326-330`) and submission refuses the whole spec (`FindUnrunnableKinds`, `TaskGraphService.ts:284-297`) → `step:'Failed'` with "replace them, or split them out of this workflow" — advice an end user cannot act on for a core shipped flow. The plan put the Prompt runner in C1.2 *precisely so cutover wouldn't break Prompt flows* (§5.7); most real flows have one. Also: the loop prompt template now *teaches* `kind:'Prompt'` that submission refuses.

2. **Caller-supplied input is dropped at the detach boundary.** `TaskGraphSubmitRequest` (`task-graph-submitter.ts:24-45`) carries no payload/data/context/messages; `persistParent` stores only continuation metadata (`TaskGraphService.ts:532-537`); the condition context reads `data`/`context` from the *upstream task's output envelope* (`TaskGraphDispatcher.ts:1062-1063`). A scheduled job's `StartingPayload`, chat `data`, and a parent agent's payload to a flow sub-agent all evaluate as `{}` inside the graph — the "condition silently never fires" class §5.5 exists to prevent, now structural. Input *mappings* are worse off than conditions: `BuildMappedInput` is called with `{ payload }` only (`TaskGraphDispatcher.ts:1113, :1167`), so `data.*`/`context.*`/`conversation[N].content` resolve to literal strings.

3. **`failureSemantics: 'edges'` is compiled and then never used.** The compiler stamps it (`flow-graph-compiler.ts:202`); `TaskGraphService` never persists it; the pure layer's `handledFailureIDs` parameters default to empty and **no production code ever computes them** — `ComputeTasksToBlock(graph.nodes, graph.edges)` at `TaskGraphDispatcher.ts:437`, `ComputeParentRollup(fresh.nodes)` at `:467`. A Failed step with a satisfied recovery edge **blocks its recovery successor and settles the graph Failed** — recovery paths, the standard flow error shape, are destroyed at cutover. `policy` (timeout/retry/onError) is likewise persisted (`TaskGraphService.ts:234-239`) and never read: zero retries, no timeouts, no claim-TTL derivation (§5.8, §5.10.2).

4. **`maxIterations: 0` is implemented as UNLIMITED and test-pinned that way** (`TaskLoopExecutor.ts:277-280`; test: "treats maxIterations of 0 as UNLIMITED, not as zero", `task-loop-executor.test.ts:76`). The plan pins the exact opposite in §4.3 **and** §9 ("the runner must NOT implement it" — the 0=unlimited JSDoc is doc-fiction BaseAgent never implemented, `Math.min(length, maxIterations)` / `while (count < maxIterations)`). This is a direct parity inversion with an unbounded-While-holds-claim-forever hazard, in the direction the plan explicitly forbade — and the JSDoc the plan ordered fixed is untouched.

5. **The continuation-depth chain breaks one hop down.** `TaskGraphAgentRunner.RunAgentForTask` stamps no depth (`TaskGraphAgentRunner.ts:29-33`), so every graph-spawned run has `ContinuationDepth = 0` and any graph *it* submits restarts at 0; `MAX_REINVOKE_DEPTH` is checked only on continuation delivery, never at `Submit`. **A Flow agent containing itself (or a mutual pair) as a Sub-Agent step submits graphs forever** — the runaway the cap exists to bound (§8 gate 6).

6. **Nested flows report success before any work happens.** A Flow sub-agent inside a Loop agent's run terminates `Success` at submission with the input payload unchanged — the parent continues reasoning on a stale payload believing the sub-flow completed. One level deeper: an `Agent` node whose agent is a Flow agent "succeeds" at submission inside `TaskGraphAgentRunner`, so **downstream tasks run before the nested graph has produced anything**.

7. **Scheduled flow jobs record green before the work runs.** `AgentScheduledJobDriver` is unchanged: `Success = submitted`, cost ≈ 0 at capture, `Status='Completed'`. A workflow that fails at 2 a.m. shows a green scheduled-job history; the later cost rollup writes only the run's `…Rollup` columns, never the job-run row.

8. **XOR race — the loser branch can execute.** `findClaimableTasks` filters `holdTaskIDs` but **not `skipSeedTaskIDs`** (`TaskGraphDispatcher.ts:823-824`). If a fork origin completes between `propagateAndRollup` and `findClaimableTasks` in the same tick (`executeClaimed` is un-awaited), the second `loadGraphState` resolves the group, removes loser edges, and the now edge-less loser target is eligible → claimed → **executed** before the next cycle can write `Skipped`. The branch nobody took runs — the exact violation §5.2's cascade-before-eligibility invariant exists to prevent.

9. **The differential simulator is unfaithful exactly where it matters.** It marks a definite-false *ordinary* conditional edge's target `Skipped` (`flow-differential.test.ts:219-225`) while the real dispatcher writes **Blocked** (`TaskGraphDispatcher.ts:924-926, :437-441`). Since §5.4 single-successor origins emit plain edges, the common "terminal conditioned step, no path matches" shape — oracle: walk ends, **Complete** — settles **Blocked** on the real engine. The fixture passes only because the simulator diverges from the thing it simulates; the parity break is masked, undeclared.

10. **A merged migration was edited.** `V202608081200__v6.1.x__Durable_EntityAction_Dispatch.sql` merged to `next` in #3674; commit `b4bd4b751` appended 105 lines to it. Anyone who ran `mj migrate` between the merges now **fails Flyway checksum validation** and needs `flyway repair`. (The PR body's "existing **unmerged** migration" claim is true only of the *other* fold target.) The new-DDL migration itself is well-built and correctly filed; this one edit violates append-only history.

11. **`Workflow.Draft` and the Workflow Planner are plausibly broken end-to-end.** The hard-cut validator requires `kind` (`task-graph-validator.ts:150-153`), but `workflow-drafting.template.md:25-58` and `workflow-planner.template.md:54-110` still teach the v1 flat shape — so every Draft call should now fail validation, on an RO the retirement migration's own header declares "deliberately KEPT." Needs template fixes + a live Draft verification.

12. **CI is red on two mechanical items**: five `as never` casts in `flow-agent-form-section.component.dom.test.ts` (:175-:208, spec-antipatterns gate), and hard-coded `[__mj]` throughout `V202608082330__Retire_Workflows_Application.sql` (must be `${flyway:defaultSchema}`).

## B. Deviations from ruled decisions — each needs an explicit written ruling, whatever the answer

1. **Submit-and-detach shipped as the universal flow contract**, reversing §6's "attached await-settlement is the default; detach is opt-in." No `FlowViaGraphExecutor`, no `WaitForSettlement`, no streaming bridge, no HITL, no `TaskGraph.CompleteTask` (all grep-zero). The follow-up UX doc's "you don't talk to a workflow" is a *coherent alternative stance* — but it is a product re-ruling of the plan's biggest decision, made in code, and findings A2/A5/A6/A7 plus the absent result-delivery path (the settled graph's payload reaches no caller — continuation posts a status summary, deliberately not the output, `TaskGraphDispatcher.ts:633-641`) are its unpaid bills. Related: the envelope run says `Completed`/`Success` **at submission** and is never corrected if the graph later fails — an `AIAgentRun` that permanently claims success for a failed workflow.
2. **Hard-cut spec, no `NormalizeTaskGraphSpec`** ("no v1 compatibility shim, deliberately", `task-graph-spec.ts:20-25`), reversing §3's "normalise, don't hard-cut" ruling — direct cause of B/A11.
3. **`Task` columns (`StepType`/`PromptID`/`Configuration` + `CK_Task_Assignment` counting `PromptID`) instead of the §5.7 `InputPayload` envelope.** Internally coherent for the shipped kinds (loop-with-ActionID routes correctly, `TaskGraphDispatcher.ts:1109-1110`) and arguably better than the envelope — my recommendation is to *ratify* this one — but record it, and note `PromptID` is currently dead schema, `config.agent.message`/`templateParameters`/sub-agent `context` are write-only, and loop tasks carry the *body's* AgentID rather than the flow agent's.
4. **The Workflows app retirement** reverses ledger row 5a's ruled front-door design (#3648) without engaging its D18-at-navigation argument; row 5b's Playwright debt now points at deleted screens; and **the 5a D18-vocabulary test guard died with the app** — the rule the ledger said "can't be satisfied by deleting the concept" was satisfied by deleting the test. The pivot doc (`plans/workflows-single-editor.md`) argues the data-ownership case well; the ruling should be written into the ledger and the D18 guard resurrected over the agent-form surfaces that inherited the front-door role.
5. **`joinMode:'any'` → `Optional` shipped and tested as intended** (`flow-graph-compiler.ts:95-97,271`) — §4.6 forbids exactly this (Optional edges don't gate; an all-Optional node runs at wave 1, *before its predecessors*) and mandates a compile diagnostic instead. Also the compiler never reads `AgentTypePromptParams` at all (`flow-graph-executor.ts:96-106` hardcodes sequential).
6. **Phase collapse**: the cutover routing (`DetermineInitialStep` → `'Tasks'`, `flow-agent-type.ts:1229`; total in-run refusal at `:231`, `:676`) landed in the same commit as C1.2/C1.3 work, against §6.1's "no BaseAgent routing change lands before the §8 gates close — that routing IS the cutover."

**§8 gates scorecard: 1 shipped (sweep fix), 3 partial (suite: synthetic-only, no metadata flows, no golden fixtures; Skipped/XOR: landed but failureSemantics unwired; depth: top level only), 3 absent (streaming, HITL, clean-bootstrap burn-in) — and the cutover shipped anyway.**

## C. Further majors (post-ruling fixes)

- **HOLD doesn't feed `IsGraphStalled`** (`graph-algorithms.ts:323-328` unchanged signature; called hold-blind at `TaskGraphDispatcher.ts:457`) — an unevaluable XOR condition holds the fork **silently forever**; the log message claiming "stalls visibly" is currently false. And the §5.6 hold for *non-group* kept-unevaluable edges is absent: the dependent **runs** (unevaluable ≈ true) — the pre-plan behavior declared change #1 was supposed to replace.
- **Loop executor parity gaps** beyond A4: no sequential payload threading (every iteration gets the initial payload, `TaskGraphDispatcher.ts:1166`); per-iteration `actionOutputMapping` with `[index]` tokens replaced by one post-hoc application to the aggregate (`:1203`); While condition evaluated against the wrong dialect (`{...payload-spread, iteration}` at `:1192-1195` vs `{payload, results, errors}`); **sub-agent loop bodies receive an empty payload every iteration** (`bodyMapping = {}` for subAgent, `:1162-1185`) with `message`/item-override unused.
- **The §5.7 anti-drift extraction didn't happen**: `payload-mapping.ts` is a *third* implementation consumed only by the dispatcher; `flow-agent-type.ts` and — live, via the Loop path — `base-agent.ts:8535-8537, 10718-10763` keep their clones. Three dialect implementations now exist; the plan mandated extraction precisely to prevent two.
- **D17 round trip loses loop-body identity**: `stepShapeFor` never writes `ActionID`/`SubAgentID` for loop steps (`task-graph-to-agent-spec.ts:267-278`), so a saved-then-reopened workflow with any loop **fails compilation** (`UnresolvedReference`); the round-trip test covers Agent+Action only.
- **§9 shortfalls**: no metadata-file-driven differential ("for every Flow agent in metadata/" — the headline), no committed golden fixtures (Track R's deletion path has nothing to stand on), no cyclic-metadata assertion, no `succeeded`-chain pin for the terminality guard (guard itself is correctly implemented, `TaskGraphDispatcher.ts:976-989`), no `data.*`/`context.*` dialect pins, While-default-100 unpinned. Plus `indexVariable` newly wired to behavior against §4.3's "carry, don't wire."
- **`NodeProgress` frame absent** (§5.11) — nothing exists for any future streaming bridge to consume.
- `ResolveExclusiveGroups` gets `{Complete, Failed}` unconditionally for every graph (`TaskGraphDispatcher.ts:917-919`) — Failed-decides was scoped to `'edges'` specs.
- Stale docs: `TaskGraphService.ts:637-648` claims the Task row "has nowhere to carry kind" — false after this PR's own migration; `task-graph-spec.ts:63-64` invents an `AIAgentStepPath.Sequence` column.

## D. Verified good (for the record)

Spec v2 union incl. `External` and per-kind validator checks; object-form self-dependency fix; compiler exclusion → single-entry (Name-sort parity) → reachability prune → cycle rejection → emission, with oracle-order `sequence` tiebreak; pure-layer Skipped/cascade/`ResolveExclusiveGroups` with XOR edges correctly exempted from the Blocked machinery and writes confined to `propagateAndRollup`; terminality guard; sweep-predicate fix exactly per §5.10.1; `ConvertAgentSpecToTaskGraph` deleted with the compiler landing; RO wire fully v2 with zero generated-type drift; loop prompt updated behind both gates; retirement orphan-free with regenerated manifests; JSONType metadata compliant (no sync write-back, no per-PR sync migration); cost rollup via the never-written `…Rollup` columns is a genuinely good design; the in-run refusal choke point is real, total, and well-placed. The seven-silent-defects narrative in the PR body is exactly the right instinct — the findings above are the same species it just didn't finish hunting.

## E. Recommended disposition

Two clean paths:

**(a) Rule the deviations and finish the cutover bar.** If submit-and-detach is the ruling (the "you don't talk to a workflow" stance), amend plan §6 + the four other rulings above, then the minimum bar before this routing ships remains: Prompt runner or an explicit Chat-escape story (A1), submit-time input snapshot (A2), failureSemantics wiring (A3), `maxIterations:0` parity (A4), depth stamping at `Submit`/runner (A5), a designed answer for nested-flow and scheduled callers (A6/A7 — even "flows may not be sub-agents/scheduled targets yet, enforced" would do), the XOR race filter (A8), simulator fidelity (A9), migration checksum note + CI greens (A10/A12), and the Draft/Planner template fixes (A11).

**(b) Split.** Keep everything except the routing: revert `DetermineInitialStep`'s `'Tasks'` return and the in-run refusals (a ~150-line revert), land this PR as C1.1+C1.2+C1.3-partial+UI, and ship the cutover as its own PR when the §8 gates actually close — which is what the plan's phasing said, and which loses nothing since the refusal choke point can return in that PR unchanged.

Given the substance-to-risk ratio, **(b) is my recommendation** — it banks ~90% of this work immediately and converts the deviations from silent reversals into rulable questions with time to answer them.

---

AN-BC:
# 📣 BUILDER AGENT — post-merge follow-up punch list (memorialized; AN-BC will confirm rulings)

#3692 is merged and the review above ([5233582557](https://github.com/MemberJunction/MJ/pull/3692#issuecomment-5233582557)) stands as the finding record. Two items are already ruled by AN-BC: **the edit to merged migration `V202608081200` is approved** (that migration has not shipped in any release, so checksum drift is a non-issue — no repair note needed), and the merge itself settles that we go **forward from the detach cutover, not back**. Everything below is the follow-on work, in priority order. File:line references are in the full review; this is the executable summary.

## P0 — broken today, fix first

1. **CI reds on `next`**: five `as never` casts in `flow-agent-form-section.component.dom.test.ts:175-208` (spec-antipatterns gate) and hard-coded `[__mj]` in `V202608082330__Retire_Workflows_Application.sql` → `${flyway:defaultSchema}`.
2. **Un-brick Prompt-step flows.** `User Onboarding Flow Agent` (6 Prompt steps) cannot run — submission refuses any spec containing `kind:'Prompt'`. Ship the `TaskGraphPromptRunner` (plan §5.7: payload injection, deep-merge response, Chat escape → remaining tasks Skipped, parent Complete, message in parent output) — `Task.PromptID` is already in the schema waiting for it. Until it lands, stop *teaching* `kind:'Prompt'` in the loop prompt template.
3. **Fix the two v1-shape prompt templates** — `workflow-drafting.template.md` and `workflow-planner.template.md` still teach flat `agentName`/`assignToUser`, which the hard-cut validator rejects, so `Workflow.Draft` and the Workflow Planner are likely failing every call. Update to v2 kind/configuration, then live-verify a Draft round-trips.
4. **Snapshot caller input at submit.** Add starting payload + `data` + serializable `context` (+ `conversationMessages`) to `TaskGraphSubmitRequest`, persist into parent-task metadata, and feed BOTH the condition context (`buildConditionContext` — currently reads upstream envelopes only) and `BuildMappedInput` (currently `{ payload }` only, so `data.*`/`context.*`/`conversation[N].content` resolve to literal strings). Without this, scheduled `StartingPayload` and chat context evaluate as `{}` inside every graph.
5. **Wire `failureSemantics: 'edges'` end to end**: persist the flag at submit; compute `handledFailureIDs` (a Failed origin with a satisfied outgoing edge) in the dispatcher; pass to `ComputeTasksToBlock` + `ComputeParentRollup`; scope `ResolveExclusiveGroups`' Failed-decides to `'edges'` specs. Today a failed step with a recovery edge settles the graph Failed — recovery paths are dead machinery.
6. **Depth chain**: `TaskGraphAgentRunner` must stamp `ContinuationDepth` on graph-spawned runs, and `Submit` must enforce the cap. Today a self-referencing flow recurses unboundedly.
7. **XOR race**: `findClaimableTasks` filters `holdTaskIDs` but not `skipSeedTaskIDs` — a fork loser can be claimed and executed in the tick before `Skipped` is written. Filter both.
8. **`maxIterations: 0`** — implemented as unlimited, plan pins zero-iterations parity (§4.3/§9). Recommendation: revert to parity (and fix the doc-fiction JSDoc). If AN-BC re-rules 0=unlimited instead, it becomes declared change #5 with a release note and the differential carve-out — but the unbounded-While-holds-claim-forever hazard needs an answer either way.

## P1 — correctness under the detach contract

9. **Envelope truthfulness**: the submitting run says `Completed`/`Success` at submission and is never corrected — on `GraphSettled`, update the envelope's outcome (or a designated outcome surface) so a failed workflow doesn't leave an `AIAgentRun` permanently claiming success; `FinalPayload` should carry (or reference) the graph outcome, and the continuation message should deliver the result, not only a status summary.
10. **Nested + scheduled callers**: a Flow sub-agent inside another agent's run, and an `Agent` node whose agent is a Flow agent, both "succeed" at submission — downstream work proceeds on nothing. Scheduled flow jobs record green before the work runs. Either wire an await path for these two callers or **enforce the limitation at submit** ("a Flow agent cannot be a sub-agent step / scheduled target yet") so it fails loudly instead of lying quietly.
11. **Make the stall visible**: pass hold information into `IsGraphStalled` (currently hold-blind — an unevaluable XOR condition parks the fork silently forever), and extend the hold to non-group kept-unevaluable edges on terminal origins (today the dependent just runs).
12. **Loop executor parity** (§4.3): sequential payload threading iteration-to-iteration; per-iteration `actionOutputMapping` with `[index]` tokens (not one post-hoc aggregate application); While condition context `{ payload, results, errors }` with `item = { attemptNumber, totalAttempts }`; sub-agent loop bodies must receive the item as payload (currently `{}` every iteration) and use `message`.
13. **Differential suite to spec** (§9): fix the simulator's ordinary-conditional-edge semantics (it writes Skipped where the dispatcher writes Blocked — masking a real Complete→Blocked divergence for terminal conditioned steps; decide which behavior is intended and make simulator + dispatcher + declaration agree); drive it over every Flow agent in `metadata/`; commit golden fixtures; add the missing pins (succeeded-chain terminality, `data.*`/`context.*` dialects, While default 100, cyclic-metadata assertion, cascade-order unit pin).
14. **D17 round trip**: `stepShapeFor` must write `ActionID`/`SubAgentID` for loop steps — a saved-then-reopened workflow with any loop currently fails compilation; extend the round-trip test to loop nodes.
15. **Consolidate the mapping dialect**: `payload-mapping.ts` is a third implementation; `flow-agent-type.ts` and the live Loop path in `base-agent.ts` keep clones. Extract-and-re-import per §5.7 so one dialect exists.
16. **Remove the `joinMode:'any'` → `Optional` mapping** (forbidden by §4.6 — Optional edges don't gate; targets run before predecessors); compile diagnostic for stored non-default `traversalMode`/`joinMode`; have the compiler actually read `AgentTypePromptParams`.
17. **`NodeProgress` frame** (§5.11) + runner progress callback — prerequisite for any future progress surface, which the detach contract makes more important, not less.
18. **Hygiene**: `PromptID` dead until item 2; `config.agent.message`/`templateParameters`/sub-agent `context` are write-only; stale `TaskGraphService.ts:637-648` docstring ("nowhere to carry kind" — false since this PR); `task-graph-spec.ts:63-64` invents an `AIAgentStepPath.Sequence` column; `policy.timeoutSeconds`/`retryCount` persisted but never read (retries in the completion path, claim-TTL derivation per §5.10.2); loop tasks carry the body's AgentID — confirm intended; resurrect the D18 vocabulary test guard over the agent-form/flow-editor surfaces that inherited the front door, and fix the "Flow" tab label / "Start node" palette string.

## Rulings for AN-BC to confirm (proposed defaults so nothing blocks)

- **R1 — detach-by-design**: ratify (settled by merge); §6 gets amended accordingly, with items 9-10 as its named bills.
- **R2 — hard-cut spec, no normalizer**: ratify (pre-release wire, no external callers yet) — item 3 is then mandatory, and the two RO wire docs are the compat story.
- **R3 — `Task` columns instead of the InputPayload envelope**: ratify — it's a better design than the plan's.
- **R4 — Workflows app retirement**: ratify; ledger row 5a marked superseded, 5b retargeted to the agent-form surfaces.
- **R5 — `maxIterations: 0`**: recommend **parity (zero iterations)** per item 8; ruling either way.

Once AN-BC confirms, the planner will amend `plans/flow-agent-taskgraph-unification.md` (§6 contract, §5.7 schema deviation, declared-changes list) and the parent-plan ledger (C.1 rows, 5a/5b) so the plan matches what shipped — deviations get memorialized as rulings, per program convention.

---

AN-BC:
# ⚖️ RULINGS CONFIRMED (AN-BC, 2026-08-09) — supersedes the "Rulings" section of [5234281079](https://github.com/MemberJunction/MJ/pull/3692#issuecomment-5234281079)

**R1 — Detach-by-design: RATIFIED.** Submit-and-detach is the flow calling contract; plan §6's attached-await is superseded. Its two named bills stand as required work: (a) **envelope truthfulness** — on `GraphSettled`, correct the submitting run's outcome and deliver the result (today the run says `Completed` at submission forever, and the settled payload reaches no caller); (b) **nested/scheduled callers** — until an await path exists, **enforce at submit** that a Flow agent cannot be a sub-agent step or a scheduled target, failing loudly instead of reporting success-before-work.

**R2 — Hard-cut spec, no v1 normalizer: RATIFIED** (was directed by AN-BC, not a deviation — recorded here for the file). Consequence stands: the two prompt templates still teaching the flat shape (`workflow-drafting.template.md`, `workflow-planner.template.md`) must be updated to v2 and `Workflow.Draft` live-verified, since the validator now rejects what they teach.

**R3 — Typed columns over the InputPayload envelope: RATIFIED.** `StepType`/`PromptID` as real columns + `Task.Configuration` typed by `ITaskStepConfiguration` is the better split: strong typing for the machine-known node config, `InputPayload` stays the freeform per-workflow data (which cannot be strongly typed — that's the point). The plan's envelope design is superseded. Riders: `PromptID` is intentionally-waiting schema for the Prompt runner; confirm loop tasks carrying the *body's* `AgentID` is intended display behavior.

**R4 — Workflows app retirement: SUPERSEDED — the app is being restored** (AN-BC, in flight now), repurposed primarily for **reviewing workflow runs** (run history/observability — three of the five producers have no agent run and are visible nowhere today). Scope details to be discussed; consequences for the punch list: the retirement migration should be **removed and the app metadata restored** rather than patching its hard-coded `[__mj]` reference (it's unshipped), and the D18 vocabulary test guard returns with the app. Ledger rows 5a/5b stay live, purpose amended.

**R5 — `maxIterations: 0`: PARITY — zero iterations.** Revert the unlimited implementation and its pinning test; fix the "0=unlimited" doc-fiction JSDoc on `ForEachOperation`/`WhileOperation`. (Removes the unbounded-While-holds-claim-forever hazard as a side effect.)

**R6 — Terminal conditioned edges: the old semantics are correct — a walk that ends settles `Complete`, never `Blocked`.** A definite-false ordinary conditional edge is a **normal not-taken outcome**: route its target through the **Skip machinery** (Skipped + cascade), exactly like XOR losers — `Blocked` is reserved for failure-driven unsatisfiability. This applies to the dispatcher generally (loop-agent graphs with not-taken conditional branches currently settle `Blocked` too — same illogic, same fix, noted as a declared correction). The differential simulator must then match the real dispatcher instead of diverging from it, and the terminal-conditioned-step fixture asserts `Complete`.

The planner will amend `plans/flow-agent-taskgraph-unification.md` (§6 contract, §3 back-compat record, §5.7 schema, §5.6 declared-changes list + R6) and the parent-plan ledger once the R4 restoration settles shape.

---

AN-BC:
## Follow-up work — status, and three items that need your call

Working the list on `feat/workflow-run-visualization` (off `next`). **P0 items 5, 8, 9, plus 2, 7 and part of 11 are done and pushed**; details at the bottom. Three items need a reply before I go further.

---

### #9 (R5, `maxIterations: 0`) — you were right, and I verified before reverting

I implemented the JSDoc instead of the behaviour. `ForEachOperation`/`WhileOperation` both document `0=unlimited`; the engines have always computed `Math.min(collection.length, maxIterations ?? default)` (`base-agent.ts:12717-12722`, `:13215`), so zero has always meant **zero**. Parity restored, and I corrected the misleading JSDoc **at its source on both types** rather than honouring it — the next reader would have been led the same way.

### #3 (v1-shape templates) — **confirmed P0, fixed**

You asked me to check rather than assume, so: both files genuinely still taught the retired shape. Three occurrences of flat `agentName`/`assignToUser` each, and **zero** occurrences of `kind`:

- `metadata/prompts/templates/workflow/workflow-drafting.template.md`
- `metadata/prompts/templates/sage/workflow-planner.template.md`

Every draft they produced was rejected outright by the hard-cut validator. Both converted to `kind` + `configuration`, and the drafting template's prose rewritten — it explained the `agentName`-xor-`assignToUser` rule, which no longer exists as a concept, since a step is one `kind`.

**Not yet live-verified** that `Workflow.Draft` returns a draft end to end. Worth someone doing.

### #16 (three mapping dialects) — agree, but I'd split it

Agreed there should be one. I'd scope it as two separate pieces of work, because they carry very different risk:

- **The live `base-agent.ts` Loop-path clones — do these.** They run on every Loop-agent turn, so they are a real drift risk against `payload-mapping.ts` today.
- **The `flow-agent-type.ts` copies — lower value.** That walker is retained-but-**refused**: it throws at its choke point (`createStepForFlowNode`) and cannot execute. Its copies are dead code kept as the reference implementation the compiler is checked against. Consolidating them is tidiness; consolidating the live ones prevents a bug.

**Ask:** confirm you want the split, or say why the oracle's copy should move too — if the differential suite is ever pointed at the real walker rather than a re-implementation, your version is right and mine isn't.

### #1 (retirement migration) — we are the collision, and I own it

Confirming: **this branch is the Workflows-app restoration.** The app comes back as the workflow-runs surface, per the UX proposal AN-BC approved. So `V202608082330__Retire_Workflows_Application.sql` should be removed and the app restored as declarative metadata under `metadata/applications/` — which is the right route anyway (metadata rule 16) and gives it a fresh identity rather than resurrecting a half-deleted one.

One wrinkle worth stating: that migration is already applied everywhere this branch has run, so it cannot simply be reverted — the removal and the metadata restoration have to land in the same release for a fresh install to end up with exactly one app.

The D18 vocabulary guard comes back with it, agreed.

**Ask:** I have not touched it yet. Say the word and I will, or take it if you are already mid-flight — I would rather one of us does it than both.

---

### Also done

- **#8 XOR race** — confirmed real. Claiming filtered `holdTaskIDs` but not `skipSeedTaskIDs`, so a decided fork's loser could be claimed and executed in the window before `Skipped` was written. Irreversibly, since the action has already run by then.
- **#5 (R6)** — unreachable targets now seed the skip cascade; `Blocked` is reserved for failure-driven unsatisfiability. Note for the record: the differential simulator had modelled Skip-on-false from the start, so the dispatcher writing `Blocked` was a divergence sitting behind a **green** suite. The dispatcher has caught up and the simulator now says so in a comment.
- **#2 Prompt runner** — `TaskPromptRunner` seam + MJServer implementation; `Prompt` joins `DISPATCHABLE_KINDS`; `persistTasks` resolves `promptName` → `Task.PromptID`. §5.7 contract implemented: `_CURRENT_PAYLOAD` placeholder + `flowContext`, deep-merge, Chat escape → remaining tasks `Skipped` (not Cancelled — nothing went wrong and nobody intervened), message on the parent. User Onboarding Flow Agent submits again.
- **#7 depth chain** — `ContinuationDepth` inherited from the graph's parent metadata and stamped on spawned runs; cap enforced at `Submit`, where nothing has happened yet.
- **#11 (half)** — a Flow agent used as a **sub-agent step** now fails loudly. Scheduled targets still need doing.
- **#6 `failureSemantics`** — persisted in parent metadata, `handledFailureIDs` computed (a `Failed` origin with a satisfied outgoing edge) and passed to `ComputeTasksToBlock`/`ComputeParentRollup`. Scoped to `'edges'` specs: under `'block'` those edges are ordinary sequencing, not recovery routes, and treating them as recovery would let a graph sail past a failure nobody anticipated.

### Still open from your list

#4 (snapshot caller input), #10 (envelope truthfulness + result delivery), #11 (scheduled targets), #12–#18. Tracked in `plans/workflow-followup-worklist.md` on this branch so it survives a session boundary.

### One flag for @jordanfanapour

The run-tree loader coming next uses a **stored query** (recursive CTE, one row per node) to replace the timeline's per-level N+1. Stored queries don't inherit entity RLS, and agent runs are user-scoped — so either the query filters by caller or it hands someone else's run tree across the wire. Flagging for your RLS/FLS work rather than deciding it here.

cadam11:
Heads up: the DOM-spec anti-pattern lint is failing next's unit backstop on 5 `as never` casts in `flow-agent-form-section.component.dom.test.ts` (lines 175, 185, 190, 201, 208 — the lint suggests `satisfies Pick<…>` + a seam cast instead). This is now the LAST red on the backstop: the actions mock fix landed Friday and the UUID sweep sites are fixed in #3696. Once these 5 casts are cleaned up, next's Unit Tests go green for the first time since Friday morning.

PR #3694, opened 2026-08-09
feat(actions): Apollo lists/search, Tavily + RSS core actions, and Buffer identity/assets fixes

MS-BC:
Consolidates #3491, #3492 and #3493 into one review. Three packages, **zero file overlap**, three independent changesets retained — the merge was clean and no content was altered in the process.

| Package | Change | Tests |
|---|---|---|
| `actions-apollo` | Apollo list management, search and prospecting | 116 |
| `core-actions` | Tavily Search + Read RSS Feed actions | 381 |
| `bizapps-social` | Buffer: `createPost` assets shape, `CredentialID`, `PlatformMetadata` | 77 |

**574 tests passing, `tsc --noEmit` clean on all three.** #3491's unit-test check was previously red; it passes here.

---

## `actions-apollo` — list management, search and prospecting (#3491)

Additive: 13 files, no existing behaviour touched. REST client, list/move/search actions, typed params and credential handling.

## `core-actions` — Tavily Search and Read RSS Feed (#3492)

Two new web actions plus their parsing helpers, registered in the package index.

## `bizapps-social` — Buffer provider (#3493)

The one live defect in this PR:

**`createPost` assets shape.** Buffer moved the input to `[AssetInput!]` (`[{ image: { url } }]`, one entry per attachment naming its kind) on 2026-05-25 and **rejects** the older `{ images: [...] }` object this package was sending. Any post carrying `ImageURLs`, `VideoURLs` or `MediaLink` was being rejected outright. Reads are unaffected — Buffer still *returns* the object form — so the input shape is now its own type (`BufferAssetInput`) alongside the unchanged `BufferAssets` response type; sharing one type would break one direction or the other.

This rests on AIDP's production evidence (a dated test asserting the array form against a live account) and **cannot be confirmed offline in this repo**. Flagged explicitly rather than claimed as verified.

Breaking for any caller passing a hand-built `assets` object to the protected `createBufferPost`; the action's own params are unchanged.

**`CredentialID` on every Buffer action.** Token resolution was strictly `getAccessToken()` → `_companyIntegration?.AccessToken`, with no per-call override — so publishing to an employee's personal channel was impossible. `CompanyIntegrationID` stays required (it identifies *which* Buffer integration, and supplies organization/channel context); `CredentialID` changes only whose token is used.

A `CredentialID` that was supplied but cannot be read is **fatal** (`INVALID_CREDENTIAL`), never a silent fallback to the tenant token — falling back would publish under the wrong identity with nothing for the caller to notice. The token never travels through action params, only its id.

**`PlatformMetadata` on `Create Post`.** Passes Buffer's per-service extras through to the mutation — `{ "linkedin": { "annotations": [...] } }` is how a LinkedIn @mention survives the trip. Accepted as object or JSON string; unparseable input **fails** rather than being dropped, for the same reason the credential failure is fatal.

---

## Review note

These were three PRs and are consolidated at the author's request. They remain three logically separate changes with separate changesets, so they can be reviewed independently within this PR — and if one turns out contentious, it can be split back out rather than holding the other two.

Supersedes #3491, #3492, #3493.


rkihm-BC (commented):
## Review

Reviewed all three changes independently, as the description invites. The engineering quality is high — this is some of the better-documented action code in the repo, the reasoning behind the awkward decisions is written down rather than assumed, and both bug fixes are real. Three things need resolving before merge; the rest is smaller.

I verified the platform-side assumptions behind each finding rather than asserting them, and note below where a finding turned out weaker than it first looked.

---

## What's good

**The `createPost` assets fix is a genuine, correctly-reasoned bug fix.** Splitting `BufferAssetInput` (array, one entry per attachment) from `BufferAssets` (object, per kind) is the right call — input and output shapes genuinely diverged, and sharing one type would have broken one direction. `buildAssetsInput` is cleaner than what it replaced.

**The CoreActions config fix is worth more than the changeset claims.** Hoisting the key extraction out of the no-config-file early return doesn't just enable Tavily — it repairs `PERPLEXITY_API_KEY`, `GAMMA_API_KEY`, `GOOGLE_CUSTOM_SEARCH_API_KEY`/`_CX` and `GOOGLE_GEOCODING_API_KEY` for every deployment that configures by environment only. All of those silently reported "key missing". That deserves its own line in the changelog — it's a fix for four existing actions, not a prerequisite for one new one.

**The Apollo client is the strongest code here.** `ApolloRESTClient` encodes five real API behaviours instead of pretending they don't exist: full-replace label semantics, add-before-remove ordering so a failed add can never strip the source label, deliberate *non*-retry of Apollo's silent-fail removes with the reasoning stated, tolerance of two or three response envelopes per endpoint, and an injected `fetchImpl` so no test can reach the live account. Sharing the label math between accounts and contacts through injected update/read closures in `moveRecords` is the right factoring.

**The RSS parsing module is well thought out.** Injected `now` so the age filter and the recency score cannot disagree; decode → strip → decode ordering with the double-escaping rationale written down; `publishedAt: null` preserved rather than defaulted to now; `http(s)`-only URL guard; per-feed failure isolation; and every count that shaped the result surfaced as an output param, so an empty answer is explainable without re-running anything.

**Fail-loud over fail-quiet, consistently.** Unreadable `CredentialID` → `INVALID_CREDENTIAL`, never a silent fall back to the tenant token. Unparseable `PlatformMetadata` → error, not a dropped field. `'false'` rejected rather than truthiness-coerced. These are the right defaults, and the reasoning is in the code where the next person will find it.

One claim I checked and can confirm: `Values` does decrypt on the `RunView` path. `Encrypt=1` on that field and `GenericDatabaseProvider.PostProcessRows` decrypts for `ResultType: 'simple'` too, so the read gets plaintext. The defensive "the value may not have decrypted" comment is harmless but reads as uncertainty about something that is settled.

I also checked whether `get-analytics` and `reorder-queue` skipping `ensureAuthenticated` contradicted "every Buffer action accepts `CredentialID`". It doesn't — both are `NOT_SUPPORTED` stubs that never call Buffer. All six functional actions honour it. The claim is accurate.

---

## Blocking

### 1. Nine new actions, no metadata

The PR adds seven Apollo actions plus `Tavily Search` and `Read RSS Feed`, and ships no `MJ: Actions` / `MJ: Action Params` records. `ActionEngine.RunClassBasedAction` resolves through `params.Action.DriverClass || params.Action.Name`, so without a row there is nothing to invoke — the classes are unreachable through the framework.

The convention is clear: `metadata/actions/` holds ~55 per-action files, and `.google-custom-search.json` is the shape to copy — action, its `MJ: Action Params`, and its `MJ: Action Result Codes`. (`Perplexity Search` is the one prior class shipped without metadata, so the gap has precedent, but it's debt rather than a pattern.)

Same problem one level down on Buffer. `Create Post (Buffer)` in `metadata/actions/.biz-apps-social.json` declares `CompanyIntegrationID, ProfileIDs, Text, MediaURLs, ScheduledAt, Shorten, RepostText, UpdateID, PostDetails, BufferCounts`, while the code reads `ChannelIDs, Content, ImageURLs, VideoURLs, MediaLink, MediaDescription, ScheduledTime, PostNow, AddToTop`. That drift predates this branch — the code was rewritten for Buffer's GraphQL API without the metadata following — but this PR adds `CredentialID` and `PlatformMetadata` to it. Neither is declared, so no agent can discover or pass them, and the two headline features are inert through the normal path.

If the metadata is coming from somewhere else (a separate sync, another repo), say so and I'll drop this. Otherwise it's what stands between the PR and delivering what it describes.

Note the knock-on: adding this metadata is also what would make the `minor` changesets correct — see below.

### 2. The 77 Buffer tests cannot run

`packages/Actions/BizApps/Social/package.json` has no `test` script (`build`, `watch`, `clean` only). Turbo's own dry run:

```
"taskId": "@memberjunction/actions-bizapps-social#test",
"command": "<NONEXISTENT>"
```

So the 387-line suite covering the one live defect in this PR is executed by nothing, in CI or locally. This is pre-existing infrastructure debt — a `vitest.config.ts` and a `social.test.ts` were already sitting there unrun — but this PR both depends on it and can fix it by adding `"test": "vitest run"` to the package.

Worth doing precisely because the assets fix can't be verified offline: with the suite runnable, at least the shape assertions run.

### 3. The credential path bypasses `CredentialEngine`

Both new credential readers go around [`CredentialEngine`](packages/Credentials/Engine/src/CredentialEngine.ts), and in two different ways — Buffer via `RunView` + manual `JSON.parse`, Apollo via `GetEntityObject` + `Load`. `getCredential()` would give you:

- **an access audit trail** — `logAccess` writes an `MJ: Audit Logs` row with the credential's `EntityID`/`RecordID`, plus `LastUsedAt`. For a feature whose entire purpose is publishing under another identity, the missing audit row is the part I'd hold on.
- **credential-type scoping** — as written, any active credential row in the tenant carrying an `accessToken` key is accepted and sent to Buffer.
- JSON-schema validation of `Values` against the credential type, and the engine's cache instead of a `RunView` per action invocation.

It's clean to adopt: `@memberjunction/credentials` depends only on `core`, `global` and `core-entities`, so no cycle from either action package. `BaseExternalDataSourceDriver` is the precedent for exactly this call shape — `CredentialEngine.Instance.getCredential<T>('', { credentialId, contextUser, subsystem })`.

One caveat in your favour: `getCredentialById` does **not** check `IsActive` (only the by-name path does), so your `IsActive=1` filter is currently *stricter* than the engine's. Keep that check when you switch.

Relatedly, and worth an explicit answer even if the answer is "out of scope": **nothing verifies that the caller is entitled to use the named `CredentialID`.** The only gate is entity-level Read on `MJ: Credentials`, which a service or agent context typically holds. Anyone who can run the action can name any credential and post as that identity. If per-credential authorization is deferred, the audit log is the minimum that shouldn't be.

---

## Worth fixing in this PR

### 4. `CompanyID` interpolated unescaped into `ExtraFilter`

`credentials.ts:113`:

```ts
ExtraFilter: `CompanyID = '${companyID}' AND Integration = '${APOLLO_INTEGRATION_NAME}' AND (IsActive IS NULL OR IsActive = 1)`,
```

I initially wrote this up as a hard blocker and then checked the platform, which makes it milder than it looks. `ValidateUserProvidedSQLClause` screens `ExtraFilter`: it strips string literals, then rejects `;`, `--`, `/*`, `union`, `exec`, `waitfor` and DML. Running payloads through that screen:

```
PASSES  -> CompanyID = 'x' OR '1'='1' AND Integration = 'Apollo' AND (...)
PASSES  -> CompanyID = 'x' OR 1=1 OR CompanyID='y' AND Integration = 'Apollo' AND (...)
blocked -> CompanyID = 'x'; DROP TABLE X--' AND ...
blocked -> CompanyID = 'x' UNION SELECT 1--' AND ...
```

So it is a boolean-logic injection, not arbitrary SQL: reachable impact is selecting a different tenant's `CompanyIntegration` row and therefore resolving a different tenant's Apollo key. Not nothing, but bounded.

It's also the house pattern — `CompanyID = '${companyId}' AND Integration...` appears identically in the Accounting, CRM, LMS and FormBuilders base actions, with about 40 unescaped `ExtraFilter` interpolations across `packages/Actions`. So this isn't a flaw this PR introduced, and I'm not treating it as a blocker.

Two reasons to still fix it here: it's one line, and the Buffer file in this same PR already does it right (`credentialId.replace(/'/g, "''")`), so the PR is internally inconsistent. Validating the value as a UUID before use is better than escaping. The repo-wide pattern deserves a separate issue.

### 5. Changeset bump levels

All three are `minor`. Per `.claude/rules/changesets.md`, `minor` is reserved for branches that touch the database — a versioned or repeatable migration, or anything under `metadata/`. This branch has none of those, so all three should be `patch`. It's the cheap direction to get wrong (no version movement under Edge pre-mode), but it's how the rule erodes.

If metadata lands per finding #1, `minor` becomes correct for all three — so resolve that first and this may resolve itself.

### 6. Move actions report partial coverage as complete success

`ApolloMoveActions.ts` reads only page 1 of the source list (100 records), and ids not found there go into the `NotOnSourcePage` output param. But the returned message is `Moved ${result.movedCount} of ${toMove.length}` where `toMove` counts only what was found, and `Success` is `true` whenever `failedCount === 0`.

Pass 300 ids and the caller sees `Moved 100 of 100 account(s)` with `Success: true` while 200 were silently skipped. `LogStatus` mentions it; the caller-facing result doesn't. Surface the skipped count in the message, and consider `PARTIAL_FAILURE` when it's non-zero. (The all-missing case is already handled correctly with `NOT_FOUND`.)

### 7. The RSS size bound doesn't bound anything

`rss-feed-read.action.ts:195-198` — `await response.text()` fully buffers the body *before* the 10 MB check, so the comment "Bounded so one enormous feed cannot exhaust memory" doesn't hold. Check `Content-Length` first, and/or read the stream with a running byte count. Separately, `text.length` is UTF-16 code units rather than bytes, so the number in the error message is wrong for non-ASCII feeds.

---

## Questions rather than findings

**Unresolved label ids are dropped, then written back as a full replace.** `resolveLabelNames` drops any `label_id` with no entry in the cached id→name map, and `addPhaseLabels` then writes the surviving names as the *complete* intended set. Since Apollo full-replaces, a dropped label is a permanently removed membership. The mechanism is certain; how often it triggers depends on whether `/labels` returns every label unpaginated, which I can't confirm from here — the client fetches it with no paging, and `labelMaps` is cached for the client's lifetime with no TTL, so a label created mid-run is one guaranteed trigger. Does `/labels` paginate on large accounts? Either way this feels like it should be `LogError` rather than `LogStatus`, and skipping the record (or refreshing the map on a miss) seems safer than writing a knowingly lossy set.

**Does the verify pass actually verify on large destination lists?** `readAccountLabels`/`readContactLabels` look for the record on page 1 of the destination (100 per page). A miss returns `null`, `possiblyStuck` stays `null`, and the record counts as moved — which is the exact case verify exists to catch. Whether a just-updated record reliably lands on page 1 depends on Apollo's default ordering, which the code doesn't pin. If it doesn't, verify silently degrades to a no-op on any destination list over 100, at the cost of one search request per record.

---

## Smaller items

- **`_credentialAccessToken` is never reset** when `CredentialID` is absent. Safe today because `RunClassBasedAction` constructs a fresh instance per run, but a one-line `else { this._credentialAccessToken = null; }` removes the dependence on that detail.
- **SSRF surface**: RSS and Tavily fetch caller-supplied URLs with no guard against link-local (`169.254.169.254`), loopback or RFC1918 addresses. The `http(s)`-only check is good and doesn't cover this. Package-wide pre-existing exposure — worth a deliberate decision rather than inheritance.
- **Unbounded concurrency**: `Promise.allSettled` over every feed, and no throttle or 429 backoff in the Apollo move loop (which issues one or two PATCHes plus a search per record against an API that rate-limits hard).
- **Duplication**: the two move actions are ~90 near-identical lines, and this PR adds a fourth independent param-helper implementation (Apollo `params.ts`, Tavily privates, RSS privates, Buffer's). `packages/Actions/CLAUDE.md` and the TS style rule both push toward a shared base. The `addOutputParam`-pushes-a-new-entry pattern matches the Perplexity/Google precedent, so leave that as is.
- **Test isolation**: the Buffer suite module-mocks `@memberjunction/core` *and* `@memberjunction/actions` wholesale, substituting `RunView`, `BaseAction` and `BaseOAuthAction`. It verifies the plumbing well, but would pass unchanged if the entity name were wrong, if decryption didn't happen, or if the base-class contract shifted. Worth stating in the file header what it deliberately doesn't cover.

---

## Also missing

- **CoreActions README**: `Tavily Search` and `Read RSS Feed` aren't in the action table (~line 294), and `TAVILY_API_KEY` isn't in the config-keys section (~line 99). ApolloEnrichment's README got a thorough new section; CoreActions got nothing.

---

## On CI

`Run unit tests` is currently **red**, so the description's "#3491's unit-test check was previously red; it passes here" is stale. The failing task is `@memberjunction/global#test` (`ClassFactory.reportResolutionFailure`) — untouched by this PR, and also failing on `next` at 20:59 then passing at 21:04, so it looks inherited/flaky rather than caused here.

Worth knowing though: because turbo aborted on that failure, `actions-apollo#test` and `core-actions#test` never executed in this run. Both resolve to `vitest run` and will run on a re-run — but combined with finding #2, no gate has yet executed any of the tests this PR relies on. A rebase or re-run once `next` settles should clear it.

Everything else is green, and `tsc` clean is confirmed — all three packages build in CI.

---

## Suggested path

Blocking: **#1** (metadata), **#2** (make the Buffer tests runnable), **#3** (`CredentialEngine`, or at minimum the audit trail).

Worth fixing here: **#4–#7** — all small, and each changes behaviour a caller would otherwise be misled by.

Fine as follow-ups: the two questions above, the smaller items, the repo-wide `ExtraFilter` pattern (separate issue), and the pre-existing Buffer metadata drift.

On the consolidation itself: it worked. Zero file overlap, three changesets retained, and each change is independently reviewable — if the Buffer credential work needs rework, Apollo and the CoreActions config fix can split back out and land on their own.

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

MS-BC:
All seven findings, the three missing items and the three nits are addressed. Two corrections to the review itself, and one place I stopped short deliberately — flagged below rather than buried.

## Blocking

**#1 — metadata.** Verified independently before acting: no `MJ: Actions` row existed for any of the nine. **140 records** now ship across `.apollo-lists.json` (7 actions), `.tavily-search.json` and `.rss-feed-read.json`, shaped after `.google-custom-search.json`.

Param names, types, defaults, required-ness and result codes come from each action's own header doc-block and its `ResultCode` literals rather than being inferred — that is how `CompanyID` got declared, which reading only the call sites would have missed. `DriverClass` is the registered string; categories follow siblings (Apollo → Data Enrichment, the two web actions → Web). `uuidgen` primary keys, no `sync` blocks.

Two bugs this flushed out, both mine:

- The two move actions initially **shared one result-code list**, so both got the same six primary keys — 140 ids, 134 unique. Caught by asserting uniqueness, not by reading the output.
- `Apollo - Create List` declares `Modality` as both an input and an output, which I wrote as two rows. `ActionParam` is UNIQUE on `(Name, ActionID)`, so **the integration tier failed the push** with `UQ__ActionPa__…` and rolled back. It is now one row typed `Both` — the representation MJ already has, used by 25 existing params. Nothing local would have caught it: the JSON is valid either way and the constraint lives in the database. The generator now asserts both properties.

**#2 — the 77 Buffer tests.** This one is **stale**. `packages/Actions/BizApps/Social/package.json` has `"test": "vitest run"`, and so does the copy on `next`, so it was fixed between your review and now. My run executes both Buffer suites — `buffer-identity-and-metadata.test.ts` (27) and `buffer.provider.test.ts` (21) — inside 275 passing.

**#3 — `CredentialEngine`.** Adopted at both call sites. Your caveat is honoured *explicitly*: the engine filters `IsActive` on its by-name path only and both sites resolve by ID, so the check is kept rather than delegated, with a test naming that reason.

Worth reporting what this cost, since it was not visible from the outside:

- `@memberjunction/credentials` had to be declared by both packages (pnpm is strict), so the lockfile moves.
- Three suites mock `@memberjunction/core` without `BaseEngine`, so they stopped **loading** — which surfaced as test counts dropping (275 → 177), not as failures. Worth knowing generally: a wholesale module mock turns a new import into a silent file-load failure.
- Six tests in `buffer-identity-and-metadata.test.ts` were written against the RunView implementation and are rewritten against the engine. **`escapes a quote in the credential id rather than building broken SQL` is deleted, not adapted** — there is no SQL on that path any more, so it tested a hazard that no longer exists. Replaced by one asserting resolution goes *through* the engine, so a regression back to a raw `RunView` fails instead of passing quietly.

## Worth fixing

**#4 — `CompanyID`.** Validated as a UUID rather than escaped, on your reasoning: the screen permits boolean logic, and a UUID check admits nothing that can alter the clause's shape while escaping only makes the value safe to embed.

**#5 — changesets.** All three were `minor`; I set them to `patch`, then back to `minor` once the metadata landed — exactly the coupling you predicted. `check:changeset` now reports *"Branch touches metadata — minor is required"*.

**#6 — partial moves.** Counted against what was **requested**, skipped ids named in the caller-facing message, `PARTIAL_FAILURE` when any were skipped. The batch is still not aborted. Shared between the two actions rather than duplicated a third time.

**#7 — the RSS bound.** `Content-Length` is checked before the body is read; the post-read check now measures `Buffer.byteLength` rather than `text.length`, which under-reported and under-bounded any feed of multi-byte characters.

## Missing items and nits

`ExtractDeclaredApplicationIds`-style duplication aside: `Tavily Search` and `Read RSS Feed` are in the CoreActions catalog, `TAVILY_API_KEY` is in the config-keys section. Buffer's `Create Post` declares `CredentialID` and `PlatformMetadata`.

## Where I stopped short, deliberately

**Buffer's wider metadata drift is NOT repaired.** I started and backed out. The record declares `ProfileIDs, Text, MediaURLs, ScheduledAt, Shorten, RepostText, UpdateID`; the code reads `ChannelIDs, Content, ImageURLs, VideoURLs, MediaLink, MediaDescription, ScheduledTime, PostNow, AddToTop`. The mapping is **lossy** — `Shorten`, `RepostText` and `UpdateID` have no code equivalent at all — and I could not establish whether removing a param record deletes the row or orphans it. Renaming and deleting rows on a record I do not fully understand is a larger and riskier change than this PR warrants, and it predates the branch. The two params *this* PR adds are declared; the rest deserves its own change with someone who knows the intent.

Your open questions on `/labels` pagination and the verify pass on destination lists >100 are unanswered — both need Apollo behaviour I cannot confirm from here, and both feel like follow-ups rather than blockers on this diff.

## Verification

apollo **116**, bizapps-social **275**, core-actions **453** — all exit 0. `tsc --noEmit` clean on all three. `check:changeset` and `check:standards` green.

Re-review welcome, particularly the rewritten credential tests and whether `PARTIAL_FAILURE` on skipped ids is too strong a signal for callers that currently branch on `Success`.

rkihm-BC (approved):
Approving. I re-verified each of the seven findings against the code at head rather than against the summary, and they're genuinely resolved:

- **#1** — 163 primary keys across the three new metadata files, **zero duplicates**; all nine `DriverClass` values match their `@RegisterClass` keys exactly; declared result codes match the `ResultCode` literals in each action. `Create Post (Buffer)` now declares `CredentialID` and `PlatformMetadata`.
- **#2** — correct, my finding was stale.
- **#3** — `CredentialEngine` at both call sites, and the `IsActive` caveat honoured explicitly at each with the reason in a comment rather than silently.
- **#4** — UUID validation rather than escaping, which is the stronger of the two.
- **#5** — `minor` is right now that the branch touches `metadata/`.
- **#6** — counted against what was requested, skipped ids named in the caller-facing message, `PARTIAL_FAILURE` when non-zero.
- **#7** — `Content-Length` checked before the body is read, `Buffer.byteLength` after, and the chunked-transfer case documented as the residual rather than papered over.

The two bugs your generator flushed out — the shared result-code ids and the `Modality` UNIQUE violation — are both ones that only surface if you actually push the metadata, so that work was real. Asserting id uniqueness as a property rather than eyeballing the output is the right instinct.

Agreeing with where you stopped short: **the wider Buffer metadata drift should not be repaired here.** The mapping is genuinely lossy, three declared params have no code equivalent, and rewriting records whose intent isn't established is a bigger and riskier change than this PR. The two params this PR adds are declared, which is the right boundary. Same for the two Apollo questions — `/labels` pagination and the verify pass on destination lists over 100 both need Apollo behaviour neither of us can confirm from here, and they're follow-ups.

### One note, explicitly not a blocker

Three params the code supports aren't declared in the metadata:

- **Tavily `Days`** — read at `tavily-search.action.ts:153` and passed through on `Topic: 'news'`
- **Tavily `Warnings`** — the output emitted when `Days` is dropped on a non-news topic
- **RSS `Feeds`** — the `[{name, url}]` form implemented in `resolveFeeds` and shown in the action's own `@example`; only `FeedURLs` is declared

The common cause is worth more than the three rows: params were derived from each action's header doc-block, and both of these are implemented but absent from that block (`Days` isn't listed at all; `Feeds` appears only inside the `@example`). Worth knowing if the generator gets reused.

I checked whether this actually breaks anything before raising it, and it doesn't. `ValidateInputs` returns `true` unconditionally, `GetActionParamsForAction` is referenced only from tests, and `base-agent.ts:6972` builds `actionParams` straight from `Object.entries(action.params)` with no filtering against the declared set — so an undeclared param passes through and works. The cost is **discoverability**: an agent won't know `Days` or `Feeds` exist, so the news-recency window and the named-feed form won't get used in practice. That's a real gap but a different and much smaller one than the original finding, where no row existed at all and there was nothing for `DriverClass || Name` to resolve. Fine as a follow-up.

### Coordination, for whoever merges second

#3827 adds `ssrf-guard.ts` to `packages/Actions/CoreActions/src/custom/web/` and applies it to the three existing web actions. It can't cover `rss-feed-read` and `tavily-search`, which don't exist on that branch. The file sets are disjoint so there's no conflict, but whichever lands second leaves these two as the only unguarded web actions in that directory — worth a follow-up applying the guard. **This PR should not grow its own**, that would duplicate #3827. Consider the SSRF item from my first review closed by that PR rather than outstanding.

Also closing out, so they don't read as unresolved: the shared-base duplication, the test-isolation note, and unbounded concurrency are all follow-ups as originally classified, and the repo-wide unescaped `ExtraFilter` pattern deserves its own issue rather than anything here.

### CI

`adopted standards` is red on three `ui-layers` violations in `packages/Angular/Generic/graph-view/` — inherited from `next` (the same three are blocking #3352), nothing from this branch. `Check migrations`, the peer-dep gate and the dependency check are green; unit tests and the deterministic integration tier were still running when I looked, so give those a glance before merging.

PR #3697, opened 2026-08-10
fix(postgresql): match SQL keywords case-sensitively when auto-quoting identifiers

MS-BC:
## Summary

PostgreSQL identifier auto-quoting used a keyword **denylist** matched case-INsensitively: a PascalCase word was quoted unless it appeared in a hardcoded keyword set. The set of SQL keywords and the set of MJ column names overlap, so every name in the intersection shipped unquoted, folded to lowercase on PG, and failed with `column "..." does not exist`. Eleven such columns are in the baseline — `Name` (on 175 tables), `Values` (the field-level-encrypted column on `__mj."Credential"`), `Length`, `Precision`, `Log`, `Rank`, `Action`, `Columns`, `Language`, `Month`, `Text`. SQL Server resolves identifiers case-insensitively, so T-SQL-first authoring never surfaced any of it. Refs #3604, #3590, #3691.

- **Keywords are now matched case-sensitively, ALL-CAPS only** — generalizing the mechanism that already existed by hand for exactly two words (`TYPE`/`DATA`) to the whole set. Dialects always emit keywords upper-case, so `TEXT` the type and `Text` the column are textually distinct. An ALL-CAPS word that is *not* a keyword is still an identifier, so the predicate is `!(isAllUpper && isKeyword)`, not a pure case rule — otherwise `ID` and `URL` fold to `id`/`url`. `SELECT Length, LENGTH(Name)` now yields `SELECT "Length", LENGTH("Name")`.
- **One implementation instead of two.** There were two copies — `PostgreSQLCodeGenProvider.quoteSQLForExecution` (all codegen SQL via `ManageMetadataBase.qsql()`) and `PostgreSQLDataProvider.autoQuoteIdentifiers` (every runtime raw-SQL statement via `ExecuteSQL`) — with a comment asking that they be hand-synced. They had already diverged: 289 keywords vs 312, plus a case-sensitive tier and a dot-qualified rule present only at runtime. Both now delegate to `AutoQuotePostgreSQLIdentifiers` in `@memberjunction/sql-dialect` (−196 / −255 lines). Codegen gains the dot rule (`__mj.vwFoo` no longer folds during codegen); runtime gains the transaction-control keywords.
- **Compatibility.** A word immediately followed by `(` is treated as a function call unless dot-qualified — this keeps mixed-case spellings like `Coalesce(` working under case-sensitive matching, and fixes ALL-CAPS functions simply missing from the set (`JSONB_BUILD_OBJECT(` was previously quoted and failed). The dot exception preserves quoting for MJ's stored procedures, which are created with quoted mixed-case names. A small structural tier (`AND OR NOT IS NULL LIKE IN BETWEEN EXISTS ASC DESC NULLS FIRST LAST`) stays case-insensitive so SQL fragments stored outside the codebase — a saved `UserView.OrderBy` of `Name Desc`, a GraphQL `ExtraFilter` of `A=1 And B=2` — keep working.
- **SQL Server is untouched.** No file under any `sqlserver` path is in the diff; `SQLServerCodeGenProvider.quoteSQLForExecution` remains the identity function and shares no code with this path.

## A regression test that derives the collision set, rather than listing it

The original defect was structural, so a hand-written list of the eleven colliders would go stale exactly the way this bug arrived. `postgresqlAutoQuote.baseline.test.ts` extracts all **4,616 column definitions from 380 tables** out of the shipped PG baseline DDL at test time and asserts each survives quoting — a newly added colliding column fails CI instead of shipping. It also reverse-guards the structural tier: if any real column ever matches those words case-insensitively, the build fails.

It earned its keep on the first run by finding something not in the original scope: seven **camelCase** columns on `Entity` (`spCreate`, `spUpdate`, `spDelete`, `spMatch`, and the `*Generated` variants) that are left bare unless dot-qualified. That is a *different* pre-existing defect, not a keyword collision, and quoting all camelCase would also quote camelCase result aliases (`AS myCount`), changing result-key casing for existing consumers. Rather than silently excluding them, the test asserts the bare set equals **exactly** those seven, so a new camelCase column trips CI. Worth a follow-up issue; in practice MJ references columns as `alias.Column`, which quotes correctly.

## Behavior changes to be aware of

Both public methods change their output — that is the fix — but two specific cases are worth review:
- `x::Text` now yields `x::"Text"`. Write `::text` or `::TEXT`; `Text` is a real column and must quote.
- `INSERT INTO Target(Cols)` with no space leaves the table bare, since a word before `(` is indistinguishable from a call. The spaced form quotes correctly.

Neither pattern occurs in this repo (verified by grep across both PG paths), and both are locked by tests so they cannot drift unnoticed.

## Test plan

- [x] `@memberjunction/sql-dialect` — **332 passed** (87 new tokenizer cases + 8 baseline-derived)
- [x] `@memberjunction/postgresql-dataprovider` — **189 passed**; all 39 pre-existing tokenizer tests pass unchanged against the shared implementation
- [x] `@memberjunction/codegen-lib` — **784 passed**; adds the first-ever tokenizer coverage on the codegen entry point
- [x] Live PostgreSQL E2E, **both** entry points (codegen + runtime), 13 tests against `docker/workbench` `postgres-claude`: unquoted-authored INSERT/SELECT/WHERE/UPDATE/ORDER BY, functions, casts, and `ALTER COLUMN ... TYPE` against a table carrying a `Type` column. Each suite includes a control assertion that the same SQL **unquoted** still fails with `does not exist`, so the tests cannot pass vacuously.
- [x] Deterministic integration tier — **60/60 passed**
- [x] `npm run check:esm` — clean (reported items are pre-existing and non-gating)
- [ ] Reviewer: confirm the two documented behavior changes above are acceptable
- [ ] Reviewer: sanity-check the structural-tier word list against any externally stored SQL you know of

To run the live-PG suites: `cd docker/workbench && docker compose --profile postgres up -d`, then `MJ_TEST_PG_URL="postgres://mj_admin:Claude2Pg99@localhost:5433/MJ_Workbench_PG" pnpm test` in either package. They self-skip without the env var, so ordinary `pnpm test` stays database-free.

---

The test plan above proves the quoting rule against a purpose-built table carrying a `Type`
column. That proves the rule; it does not prove the rule covers **what MemberJunction
actually ships**. This section closes that gap: the collider set is derived from the live
database rather than authored, and every collider is driven through the running stack.

**Deriving the set instead of guessing it.** An earlier draft used a hand-written list
(`Count`, `Format`, `Date`, `Left`, `Right`) — every one a keyword, not one of them an
actual MJ column, so the test would have proven nothing. The set is now computed by
intersecting every `__mj` column name with PostgreSQL's own `pg_get_keywords()` table. Out
of **2,074** distinct column names, **5** collide with a RESERVED (R), TYPE-FUNCTION (T) or
COLUMN-NAME (C) keyword:

```
T  AIAgentRun.Verbose                 C  Credential.Values
C  EntityField.Precision              C  IntegrationObjectField.Precision
C  RecordGeoCode.Precision
```

Two of those are load-bearing: `EntityField.Precision` is core metadata that every entity
read depends on, and `IntegrationObjectField.Precision` sits in the integrations path.

### Every assertion is paired with a control — 25/25

A passing quoting test proves nothing unless the unquoted form genuinely fails; otherwise
PostgreSQL was tolerating the identifier all along and the quoter is untested.

| # | Check | Result |
|---|---|---|
| 1 | collider set derived from the live schema (5 of 2,074 columns) | pass |
| **2** | **control:** unquoted `EntityField.Precision` → `column "precision" does not exist` | **fails, as required** |
| **3** | **control:** unquoted `IntegrationObjectField.Precision` → `column "precision" does not exist` | **fails, as required** |
| **4** | **control:** unquoted `AIAgentRun.Verbose` → `syntax error at or near "FROM"` | **fails, as required** |
| 5–8 | the same four identifiers through the provider's auto-quoter — all resolve and return rows | pass |
| 9–10 | `WHERE` and `ORDER BY` on `MJ: Entity Fields.Precision` | pass |
| 11–12 | `WHERE` and `ORDER BY` on `MJ: Integration Object Fields.Precision` | pass |
| 13–14 | `WHERE` and `ORDER BY` on `MJ: AI Agent Runs.Verbose` | pass |
| 15–16 | `WHERE` and `ORDER BY` on `MJ: Credentials.Values` | pass |
| 17–20 | acronym columns `ID`, `UserID`, `EntityID`, `EncryptionKeyID` still quoted | pass |
| 21–23 | `SchemaName` / `BaseTable` return under exact casing; no lowercase-folded duplicate keys | pass |
| 24–25 | collider columns `Length` / `Values` round-trip through a generated base view carrying an IS-A parent JOIN, a stored procedure, and the GraphQL resolvers | pass |

The SELECT list is the easy position; checks 9–16 put each collider in a **predicate** and a
**sort key**, where an unquoted identifier surfaces differently.

### The acronym regression this rule had to avoid

`ID` is ALL-CAPS *and* not a keyword. A naive "mixed case → quote" rule stopped quoting it
and broke every acronym column in the schema; the shipped condition is
`!(isAllUpper && isKeyword)`. Checks 17–20 confirm that holds on live schema, not just in
unit tests — `ID`, `UserID`, `EntityID` and `EncryptionKeyID` all resolve and return under
their original casing.

Worth noting how one of those checks failed first: asking `EntityField` for a `UserID` it
does not define produced `column "UserID" does not exist` — **already quoted**. The quote
marks in the error were themselves proof the rule had fired; the fixture was simply wrong,
and the check now pairs each column with a table that actually has it.

### Casing, where the failure is silent

PostgreSQL folds unquoted identifiers to lowercase, so a mixed-case column that loses its
quotes comes back under the *wrong key* — a silent wrong answer rather than an error, which
is the worst kind. Checks 21–23 assert `SchemaName` and `BaseTable` arrive under their exact
casing and that no lowercase-folded duplicates appear alongside them.

---

## Where this rule is still needed — evidence from the integration tier

Running the deterministic integration tier against live PostgreSQL surfaced **live instances
of exactly this bug class**, in paths this PR does not cover. They are recorded here as
scope evidence, not as regressions introduced by it.

```
PostgreSQLDataProvider.ExecuteSQL failed: syntax error at or near "["
PostgreSQLDataProvider.ExecuteSQL failed: type "uniqueidentifier" does not exist
client-cache.C13 failed: warm failed: column "__mj_updatedat" does not exist
```

Each is the same underlying problem from a different angle:

- **`syntax error at or near "["`** — T-SQL bracket-delimited identifiers reaching
  PostgreSQL verbatim. This one is *authored SQL*, not composed SQL: 11 of the 21 stored
  queries have no PostgreSQL dialect variant (measured directly against the live database),
  so they execute as T-SQL. The auto-quoter never sees them.
- **`column "__mj_updatedat" does not exist`** — the lowercase-folding failure this PR
  exists to prevent, from authored rather than generated SQL. Note that MJ's *composed*
  cache-validation SQL is correct: `GenericDatabaseProvider` builds it with
  `QuoteIdentifier('__mj_UpdatedAt')`, and that path passes.
- **`type "uniqueidentifier" does not exist`** — a T-SQL *type* name, the same class one
  level up from identifiers. Not from authored RLS content (0 of 13
  `RowLevelSecurityFilter` rows mention it), and not from a labelled test fixture; the
  emitting call site was not isolated.

The distinction that matters for reviewing this PR: everywhere MemberJunction **composes**
SQL, the shared tokenizer now quotes correctly, and the suites below prove it against the
real shipped schema. Everywhere MemberJunction **executes SQL somebody else authored** —
stored queries without a PG variant, test fixtures written in T-SQL — the tokenizer is not
in the path at all. That is a content and dialect-coverage problem, and it is tracked in
#3703 rather than here.

## Unit tests

| Package | Result |
|---|---|
| `@memberjunction/sql-dialect` | **335/335** |
| `@memberjunction/sql-converter` | **1,134 passed**, 3 skipped |
| `@memberjunction/codegen-lib` | **792 passed**, 66 skipped, 0 failures |

Across the full stack this PR sits on, **2,909 unit tests pass with zero failures**.


MS-BC:
---

## End-to-end verification on live PostgreSQL

Exercised against a real PostgreSQL 17 database on the branch stacked with #3590 and #3703, using a fixture built specifically to hit the defect class this PR closes.

**The test targets the collision directly.** An IS-A parent table was created with two columns named **`Length`** and **`Values`** — two of the eight identifiers that are both MJ column names and members of the keyword set. Under the previous denylist both were emitted bare, folded to `length` / `values`, and the statement failed with `column "length" does not exist`.

Every stage that touches those columns now succeeds:

- **CodeGen metadata read** — both mirrored onto the child entity as virtual fields (this is the exact `SELECT ID, IsVirtual, Type, Length, …` that used to throw)
- **View generation** — emitted quoted in the generated base view:
  ```sql
  __mj_isa_p1."Length", __mj_isa_p1."Values"
  ```
- **Runtime write** — a `Save()` through the entity layer persisted `Length=42` and `Values={"source":"isa-e2e"}`
- **Runtime read** — `RunView` returned both columns with their values intact

19/19 behavioural checks passed, four of them specifically on the collider columns across the codegen-time and runtime paths — i.e. both tokenizer call sites this PR consolidates.

### One gap this run exposed, outside this PR's reach

`LogSQLAndExecute` calls `ds.query()` directly and never routes through `qsql()`, so the tokenizer is never invoked on that path — **61 call sites in CodeGenLib**. One of them authored `INSERT INTO __mj."Application" (ID, Name, …)` bare, and CodeGen could not create the bucket application for a new schema on PostgreSQL: `column "id" of relation "Application" does not exist`.

This PR hardens the tokenizer; it cannot help where the tokenizer is not called. The single live instance is fixed in #3703; the architectural gap is the CodeGen twin of #3427 and warrants its own issue. Worth noting when judging whether the quoting class is closed — it is closed for everything that goes through `qsql()`, which is not everything.

rkihm-BC (commented):
# Code review

**Verdict: approve after one trivial fix (the changeset level), with one substantive finding I'd push to land here rather than defer.**

I checked out the head commit (`1b2a16f3aa`), ran `@memberjunction/sql-dialect` (**332/332 pass**), ran the tokenizer against the shipped PG baseline and ~40 hand-built risk patterns, and ran the repo's own changeset checker. All six CI checks are green.

---

## What's good

**The diagnosis is right and the fix is the minimal one.** A case-insensitive keyword denylist can never be correct when the keyword set and the column-name set intersect; making the match ALL-CAPS-only is the smallest change that separates them, and it generalizes a mechanism the code already had by hand for exactly `TYPE`/`DATA`. The predicate `!(isAllUpper && isKeyword)` rather than a pure case rule — with `ID`/`URL` as the stated counterexample — shows the author found the trap a naive version falls into.

**The consolidation is real, not cosmetic.** −451 lines, two copies that had genuinely diverged (289 vs 312 keywords, plus a dot rule and a case-sensitive tier on only one side) collapse into one. I grepped the branch: no tokenizer copies remain anywhere. Codegen picks up the dot rule and `RETURNING`/`SAVEPOINT`/`CONSTRAINTS`; runtime picks up the transaction-control words. Both directions of the prior drift are closed, and the module docblock says "do not reintroduce a private copy," which is the right thing to leave behind.

**Rule 3 (word before `(` is a call) does double duty** — it keeps `Coalesce(`/`IsNull(` working now that keywords are case-sensitive, *and* it fixes ALL-CAPS functions that were simply absent from the set (`JSONB_BUILD_OBJECT(` used to be quoted and fail). The `!precededByDot` guard preserving MJ's quoted mixed-case stored procs is a nice piece of reasoning, and it's tested both ways.

**The structural tier is the right size.** Fourteen words, all of them ones that can't be MJ columns, existing purely so stored `UserView.OrderBy` / `ExtraFilter` fragments keep working — and reverse-guarded by a test that fails the build if a real column ever collides. That's a carve-out with a fence around it.

**The baseline-derived test is the right shape.** Deriving 4,616 columns from the shipped DDL instead of listing eleven names is exactly the fix for a defect that arrived via a stale hand-written list. And when it surfaced the seven camelCase `spCreate`-family columns — out of scope, different defect — the PR asserted the set *exactly* rather than excluding it quietly. That's the honest handling.

**SQL Server is genuinely untouched.** Verified: no `sqlserver` path in the diff, and `SQLServerCodeGenProvider.quoteSQLForExecution` remains the identity function.

**Mechanics are clean.** Both public methods keep their signatures (only `private static` fields were removed, so no API break); `pg` is a declared dependency of both packages so the integration files' top-level import is safe even when they skip; the suites self-skip without `MJ_TEST_PG_URL`; every live assertion is paired with a control proving the unquoted form actually fails.

---

## Must fix before merge

### 1. The changeset must be `patch`, not `minor`

I ran the repo's own checker on the branch:

```
Branch touches no migration and no metadata — every entry must be patch.
❌ Changeset bump levels (3 problem(s))
   .changeset/pg-case-sensitive-keyword-quoting.md — "@memberjunction/postgresql-dataprovider": minor, but this branch changes no migration and no metadata
   .changeset/pg-case-sensitive-keyword-quoting.md — "@memberjunction/codegen-lib": minor, but this branch changes no migration and no metadata
   .changeset/pg-case-sensitive-keyword-quoting.md — "@memberjunction/sql-dialect": minor, but this branch changes no migration and no metadata
```

Per `.claude/rules/changesets.md`, the level tracks *what the branch does to the database*, not how public the change is — and this branch touches neither `migrations/` nor `metadata/`. Commit `28ee2d25` ("minor, not patch — two public methods change output") reasons from semver, but the rule is explicitly not semver: all ~300 packages share one `fixed` group, so a stray `minor` bumps the whole workspace and consumes the Edge tuple the next certification is targeting. **No CI gate enforces this one**, so it will merge as-is unless a reviewer catches it. One-line fix.

---

## Concerning

### 2. `__mj_`-prefixed columns are still left bare — and this PR pins that with a new test

`isBareWord` returns bare for any word starting `__mj_`, and that clause is evaluated *before* the dot rule, so even the qualified form escapes:

```
"SELECT t.__mj_UpdatedAt FROM __mj.Entity t"  =>  "SELECT t.__mj_UpdatedAt FROM __mj.\"Entity\" t"
```

The PG baseline defines five mixed-case such columns — `__mj_CreatedAt`, `__mj_UpdatedAt`, `__mj_Latitude`, `__mj_Longitude`, `__mj_UDT` — so that output folds to `__mj_updatedat`, which is verbatim the failure this PR's own description reports from the integration tier.

The clause is also **redundant for its stated purpose**: all-lowercase `__mj_*` names are already covered by `isAllLower`, so the *only* words `isMJInternal` affects are precisely the mixed-case real columns. Deleting it from the return expression is the whole fix.

What makes this worth raising rather than deferring is that the PR's guards are shaped around it. `postgresqlAutoQuote.baseline.test.ts` filters `!c.startsWith('__mj_')` out of *both* column assertions, and the new CodeGenLib test asserts that `WHERE __mj_CreatedAt > now()` is left untouched — so the exclusion is now locked in by a passing test under a heading that says "every real column survives auto-quoting." It isn't a regression (both prior copies had it), but a PR whose thesis is "columns that collide with the rule ship broken" leaves five of them behind.

Exposure is narrower than it first looks, in the PR's favour: framework paths build this SQL with `QuoteIdentifier` (`GenericDatabaseProvider.ts:2468`, `:2868`, `:4185`), and `ExtraFilter`/`OrderBy` go through the *metadata-driven* `TransformExternalSQLClause`, which knows the entity's field list. What's left is hand-authored and stored SQL — note the `CacheValidationSQL` example in the Query entity's own field description is `SELECT MAX(__mj_UpdatedAt) AS MaxUpdatedAt …`.

### 3. The rule is tested in one direction only

The column direction is exhaustively derived; the keyword direction isn't tested at all, and the keyword set is still a hand-curated denylist. Running the shipped baseline through the tokenizer, these ALL-CAPS PG keywords come back quoted:

`BOTH`, `CURRENT_USER`, `DEFERRABLE`, `EXTENSION`, `INITIALLY`, `VALID`, `SYSTEM` (as in `TABLESAMPLE SYSTEM`) — plus `SESSION_USER`.

At least two are reachable: `PostgreSQLCodeGenProvider.ts:1363-1365` maps T-SQL defaults `user_name()` → `CURRENT_USER` and `suser_name()` → `SESSION_USER`. Pre-existing and unchanged by this PR, but a reverse guard would be the mirror of the structural-tier guard the PR already wrote, and adding the words costs nothing.

### 4. The derived set is only as fresh as `migrations-pg`

Newest baseline is `v5/B202607091514__v5.46.x__Baseline.pg.sql`, with 248 v5 + 24 v6 incremental PG migrations sitting on top that the extractor never reads. Per `migrations/CLAUDE.md`, `migrations-pg` is regenerated by the build engineer at release time — so a v6-era colliding column won't trip CI until then. That's the gap between the unit test's derivation and the live-DB derivation in the description (which found `AIAgentRun.Verbose`). Worth a sentence in the test's docblock so the next reader doesn't over-trust it.

### 5. One behavior change missing from the "be aware of" list

`SELECT COUNT(*) AS Count` previously emitted `Count` bare → PG returned the key `count`. It now emits `AS "Count"` → key `Count`. Same for `AS Name`, `AS Type`, `AS Rank`, `AS Value`. I grepped: the only non-test occurrence is `SQLServerCodeGenProvider.ts:1235`, which isn't on this path — **no in-repo breakage**. But stored `Query.SQL` rows can carry such aliases, and a consumer reading the folded lowercase key would break. The new behavior is arguably the correct one (it matches the declared QueryField name); it just belongs in the release note beside `x::Text`.

### 6. The compatibility claim is slightly broader than what holds

Scoping the structural tier to filter/sort fragments is the right call. But the description says "SQL fragments stored outside the codebase … keep working," and full *statements* don't:

```
"Select Name From __mj.vwEntities Where Status = 'a' Order By Name Desc"
  => "\"Select\" \"Name\" \"From\" __mj.\"vwEntities\" \"Where\" \"Status\" = 'a' \"Order\" \"By\" \"Name\" Desc"
```

That previously worked. Zero occurrences in-repo, low likelihood in the wild (people write UPPER or lower), and the fix would be worse than the problem — so not something to change, just something to say as *fragments*, not *statements*.

---

## Missing / nits

- `packages/SQLDialect/README.md` documents the package's exports across 405 lines and isn't updated for the three new ones.
- The description reports sql-dialect as both **332** and **335/335** in two different tables. I measured 332 at head.
- The PR consolidates two tokenizers, but `PostgreSQLDataProvider` still carries a *third* quoter — `quoteIdentifiersInSQL`/`quoteFieldNamesInToken`, metadata-driven, used by `TransformExternalSQLClause`. It's doing a genuinely different job (it knows the entity's fields, which is why finding #2 doesn't reach `ExtraFilter`), so leaving it is right — but a cross-reference in the new module would stop the next person "finishing" the consolidation.
- Now that the comment / `E'...'` limitations live in exactly one place, they're worth a follow-up issue rather than only a docblock. The apostrophe-in-a-comment case is the sharp one: it starts a string scan that swallows real SQL.
- `findNewestBaseline()` does `candidates.sort()` on full paths and takes the last. Correct today; wrong the moment a `v6` baseline carries an older timestamp than the `v5` one. Sort on basename.

---

## Explicitly *not* blockers

- Comment and `E'...'` tokenization — pre-existing, documented, unchanged by this PR.
- `INSERT INTO Target(Cols)` with no space — I verified zero unqualified occurrences; the dot-qualified form (`__mj.Application(ID, Name)`) quotes correctly.
- `x::Text` — verified zero occurrences outside the tests that pin it.
- Two `toUpperCase()` allocations per word — invisible next to a database round trip.

Answers to the two reviewer checkboxes in the description: the documented behavior changes are acceptable, with the alias-casing case (#5) added to the list; and the structural-tier word list is well chosen for filter/sort fragments, with the caveat in #6 about full statements.

rkihm-BC:
## Correction to my review above

One claim in finding #3 doesn't hold up, and I'd rather retract it than leave it standing.

I wrote: *"At least two are reachable: `PostgreSQLCodeGenProvider.ts:1363-1365` maps T-SQL defaults `user_name()` → `CURRENT_USER` and `suser_name()` → `SESSION_USER`."*

That's where those words are **produced**, not evidence they reach the tokenizer. Tracing it properly: `formatDefaultValue` feeds generated DDL and CRUD function bodies, which execute via `SQLLogging.LogSQLAndExecute` → `ds.query()` directly. `quoteSQLForExecution` is never in that path — it's the same bypass documented in #3703. So I have no demonstrated call site where a missing ALL-CAPS keyword actually reaches either entry point.

**What survives of that finding, all measured:** the keyword direction has no test coverage at all, while the column direction is exhaustively derived from the baseline. Running the shipped PG baseline through the tokenizer, these ALL-CAPS PostgreSQL keywords come back quoted and would fail *if* they ever reached it — `BOTH`, `CURRENT_USER`, `DEFERRABLE`, `EXTENSION`, `INITIALLY`, `VALID`, `SYSTEM`, `SESSION_USER`. A reverse guard mirroring the structural-tier one is still cheap and still worth having. It's a gap in the safety net, not a live defect — treat it as a nit rather than the "concerning" item I filed it under.

The other findings I re-verified and they stand as written, including the changeset level: this branch is TypeScript, tests and a changeset, with no migration of either dialect and no `metadata/` change, so `patch` is correct here. (Worth noting for anyone applying the same rule elsewhere: a branch carrying a **PostgreSQL** migration *is* a database change and should be `minor` — `.github/scripts/check-changeset-bump.mjs` only globs `migrations/**` and `metadata/**`, so it under-reports on `migrations-pg/**` branches. That doesn't affect this PR, which has neither.)

MS-BC:
Thanks — and thanks for the correction on #3, which I've taken at face value: treated as a nit, and the reverse guard built anyway because it turned out to be worth more than the finding it came from. Pushed as `2397b732f2`.

## Must fix

**1. Changeset → `patch`.** Done, and verified with the repo's own `check-changeset-bump.mjs`, which now reports `✅ bump levels are correct`. Re-ran it after merging `next` (which brought a `task-graph` changeset of its own) — the gate still classifies this branch as touching no migration and no metadata, so `patch` remains right.

## Concerning

**2. The `__mj_` exclusion.** Removed, and you were right that it was worth landing here rather than deferring — the PR's thesis is precisely that colliding columns ship broken, and this left five of them behind.

One thing your write-up implies that I want to be explicit about, because deleting the clause is **necessary but not sufficient**: `__mj_UpdatedAt` doesn't *start* with an uppercase letter, so dropping the exemption alone fixes only the dot-qualified form you demonstrated. The unqualified shape — `SELECT MAX(__mj_UpdatedAt) AS MaxUpdatedAt`, which is verbatim the `CacheValidationSQL` example you pointed at — would still have folded. So there's now a positive rule: a `__mj_` word carrying any uppercase is an identifier. The prefix is MJ's own namespace and no SQL keyword lives there, so it can't collide with anything. All-lowercase `__mj_*` names stay bare via the ordinary lowercase rule, unchanged.

You found **two** tests pinning the old behaviour. There were **four**, across three packages — the two you named plus `PostgreSQLCodeGenProvider.test.ts` ("leaves … `__mj_` columns untouched") and `PostgreSQLDataProvider/autoQuoteIdentifiers.test.ts` ("does not quote the `__mj_` prefixed timestamp columns"). All four now assert the opposite, and each says in a comment what it used to pin, so the next person doesn't "fix" it back.

**3. The keyword direction is now guarded — and my first attempt at it was worthless.** Worth reporting because it's the more useful half.

I initially derived the candidates by taking the baseline's ALL-CAPS words and filtering them through `PostgreSQLQuotingKeywords`. That's circular: deleting a word removes it from the *population* as well as the set. Mutation-testing it — dropping the eight words you named — left the suite **green**. The guard would never have caught the thing it existed to catch.

It now uses PostgreSQL's own reserved-word list as an independent oracle. The same mutation fails it, naming all eight. There's a companion test asserting the mixed-case forms (`Valid`, `System`, `Both`, `Extension`, `Order`, `Table`) still quote — that's the property that makes adding keywords free, and it deserves to be pinned rather than reasoned about.

The eight are added: `BOTH`, `CURRENT_USER`, `SESSION_USER`, `DEFERRABLE`, `INITIALLY`, `EXTENSION`, `VALID`, `SYSTEM`.

**4. Baseline freshness.** The docblock now states what the test does *not* cover: incremental `migrations-pg` files are never parsed, so a current-era colliding column won't trip it until the next baseline is cut. A green run means "no collision in the last baseline", not "no collision" — which is exactly the gap that let the live-DB derivation find `AIAgentRun.Verbose`.

**5. Alias casing.** Added to the behaviour-change list in the changeset, with the concrete before/after (`AS Count` → key `Count`, not `count`) and the note that stored `Query.SQL` rows in consumer databases can carry such aliases.

**6. Fragments, not statements.** Corrected in the changeset — the compatibility claim is now scoped to filter/sort *fragments*, and it says explicitly that a full Title-Case statement no longer works and that this is documented rather than changed.

## Nits

- `SQLDialect/README.md` documents the three new exports, with the ALL-CAPS rule and why it's load-bearing.
- The module now cross-references `PostgreSQLDataProvider`'s metadata-driven quoter as **deliberately** separate, with the reason (it knows the field list; this one can't), so nobody "finishes" the consolidation.
- Comment / `E'...'` limitations filed as #3775 and referenced from the module. Your framing that the apostrophe case is the sharp one is in the issue — a comment's *prose* silently changing what SQL gets parsed is a nasty distance between cause and symptom.
- `findNewestBaseline()` sorts on basename now; sorting full paths ranked the folder segment ahead of the timestamp.
- The 332-vs-335 discrepancy: **406** at this head, which is the number after the two new guards. Your 332 measurement was right for `1b2a16f3aa`.

## One thing outside the review

`packages/CodeGenLib/src/__tests__/pg-connection.test.ts` was 10-for-10 red on this branch, unrelated to its subject: the mock is `vi.fn(() => ({…}))`, and an arrow function has no `[[Construct]]`, so `new PGConnectionManager()` throws under Vitest 4. Fixed with a function expression, since leaving a red suite behind seemed worse than the small scope creep. (Same defect exists on #3703; fixed there too.)

**Suites at this head:** sql-dialect **406/406**, codegen-lib **785 passed / 66 skipped**, postgresql-dataprovider **190 passed / 7 skipped**. Merged `next` and re-ran everything after.

rkihm-BC (commented):
# Re-review at `2397b732f2`

**Ship it.** Every finding is addressed, all six checks are green, and the two guards that matter I mutation-tested independently rather than taking on trust. Three nits below, none blocking.

## Verified, not assumed

| Finding | Status | How I checked |
|---|---|---|
| Changeset → `patch` | fixed | `check-changeset-bump.mjs` → `✅ bump levels are correct` |
| `__mj_` exemption | fixed | `t.__mj_UpdatedAt` **and** bare `MAX(__mj_UpdatedAt)` both quote; `__mj_isa_p1` alias still bare |
| …and pinned | yes | reintroduced the carve-out → **2 tests fail**, in two different packages |
| Keyword direction unguarded | fixed | deleted all 8 words → **fails**, naming them; the mixed-case companion test passes |
| Baseline staleness | documented | docblock states incrementals aren't parsed |
| Alias casing | documented | in the changeset with the `AS Count` → key `Count` before/after |
| Fragments ≠ statements | corrected | scoped to fragments, Title-Case statements called out as failing |
| README / cross-reference / #3775 / basename sort / `pg-connection.test.ts` | all done | #3775 open; sort is `basename(a).localeCompare(basename(b))` |

sql-dialect **406/406** on my machine. Idempotency holds over the full baseline.

**On the `__mj_` fix specifically — you were right and I was imprecise.** I demonstrated the dot-qualified form and inferred the rest; deleting the exemption alone would have left bare `MAX(__mj_UpdatedAt)` folding, because `__mj_UpdatedAt` doesn't start uppercase. The positive rule is the correct shape, and the reasoning behind it (the prefix is MJ's namespace, no keyword can live there) is what makes it safe rather than merely convenient. Four pinning tests rather than the two I found, each annotated with what it used to assert — that's the part that makes it stay fixed.

**The circular first guard is the most useful thing in your write-up.** Deriving candidates by filtering the baseline through the set under test means deletion removes the word from the population too, so the guard can never fail. It passed mutation and was worth nothing. Catching that yourself, and saying so, is worth more than the guard.

**And the reverse-guard turned up things well outside my finding.** The comment-parity bug is the standout: an apostrophe in `doesn't` on line 10 of one query file inverting literal/code parity for everything after it, so `= 'Prompt'` becomes `= '"Prompt"'` and returns no rows, and `jsonb_build_object` keys become `'"ID"'` so every consumer reading `.ID` gets `undefined`. I filed that class as a follow-up; you measured its blast radius against the shipped query bodies and fixed it. The Nunjucks-tag skip I hadn't considered at all — quoting a parameter name inside `{{ … }}` means it never substitutes, which is silent. Both now have a test that asserts literals and tags come back byte-identical using an independently written scanner.

Also worth recording: widening the structural tier was **tried and reverted**, with the reason (case-insensitive matching makes those 20 words unquotable as column names — the exact defect class this PR removes) and the observation that it didn't even work. A documented negative result is more useful than the feature would have been.

## Nits

**1. `On` isn't covered by the contextual tier**, so a Title-Case JOIN still fails — just one word later than before:

```
SELECT a.Name FROM t a Left Join u b On a.ID = b.ID
  =>  SELECT a."Name" FROM t a Left Join u b "On" a."ID" = b."ID"
```

This sits inside the limitation the changeset already documents, and `On` can't be expressed contextually anyway (the pairing isn't adjacent — there's a table and alias in between), so I wouldn't change the code. But `Left Join` and `Full Outer Join` now surviving makes joins *look* supported, and a reader may not get as far as the limitation paragraph. One clause in the contextual tier's docblock — "the `ON` that follows is not covered; a Title-Case JOIN still fails" — closes the gap between what the tier appears to promise and what it does.

**2. Dot-qualified all-lowercase words are now quoted**, which isn't in the behavior-change list:

```
pg_catalog.set_config(…)        =>  pg_catalog."set_config"(…)
information_schema.columns      =>  information_schema."columns"
__mj.fn_trg_update_action()     =>  __mj."fn_trg_update_action"()
```

This falls out of moving the dot rule ahead of `isAllLower`. It's **semantically inert** — a quoted lowercase identifier is identical to the unquoted folded form, and a lowercase reference can only ever have matched a lowercase object — and I confirmed idempotency still holds. But a baseline round-trip now newly quotes **463** words where the previous head quoted 16, so anything that reads emitted SQL (audit files, logs, diffs against previously generated output) gets substantially noisier. Worth a line in the changeset so nobody diagnoses it as a bug later.

**3.** Changeset says "404 tests across the package"; it's 406.

## One thing I'd still like, as a follow-up not a condition

The tier-1-before-dot ordering rests on "no ALL-CAPS column in the shipped schema is a keyword". I checked and the changeset's list is exact — `BCMID ID ISO2 ISO3 ISO3166_2 SQL URI URL`, none in the set, all eight quote both bare and dot-qualified. That's a real invariant doing real work, and it's currently asserted in prose. It's the same shape as the structural-tier guard you already built: extract ALL-CAPS columns from the baseline, fail if any is in `PostgreSQLQuotingKeywords`. Cheap, and it would catch a future `ACTION`-style column that tier 1 would silently swallow after a dot.

Approving. The `pg-connection.test.ts` fix was the right call too — an arrow function with no `[[Construct]]` slot is a genuinely confusing failure to leave for the next person, and ten red tests adjacent to your subject are worse than the scope creep.

rkihm-BC:
Filed the ALL-CAPS-column / keyword-set guard as #3792 — the follow-up from my re-review, not a condition on merging this.

PR #3696, opened 2026-08-10
fix: UUIDsEqual at the six raw UUID comparisons failing MJGlobal's sweep

cadam11:
## What

next's Unit Tests backstop has been red since Friday on MJGlobal's UUID-compliance sweep. Six raw comparisons, all from recent task-graph work (the five I listed on #3670, plus one that arrived with the weekend merges):

| Site | Fix |
|---|---|
| `MJServer/src/services/TaskGraphActionRunner.ts:21` | `UUIDsEqual(a.ID, params.ActionID)` |
| `entity-actions.checks.ts:387` (filter removal index) | `UUIDsEqual(f.ID, Filter.ID)` |
| `entity-actions.checks.ts:456` (submission filter) | `UUIDsEqual(r.EntityActionID, binding.ID)` |
| `entity-actions.checks.ts:462` (RecordID substring) | `NormalizeUUID()` both sides — see below |
| `entity-actions.checks.ts:505` (exactly-once assert) | `UUIDsEqual(r.EntityActionID, binding.ID)` |
| `AI/CorePlus/src/task-graph/flow-graph-compiler.ts:174` (cycle-path names) | `UUIDsEqual(s.ID, id)` |

The RecordID site is a substring check on a composite-key string, so `UUIDsEqual` does not fit — `NormalizeUUID` (trim+lowercase) on both sides keeps the containment check while making it case-insensitive, which matters because SQL Server hands out uppercase GUIDs.

## Verification

- MJGlobal's `UUIDCompliance.test.ts`: **passes** (was 1 failed with 6 violations).
- `@memberjunction/server`, `@memberjunction/testing-integration`, `@memberjunction/ai-core-plus` all build clean with their full dependency graphs (zero TS errors).

## For reviewers

- No behavior change on same-platform data (all six compared same-source metadata UUIDs); the fix removes the latent cross-platform case-sensitivity bug the sweep exists to catch.
- No changeset: the MJServer/CorePlus lines are behavior-preserving one-liners; say the word if you want a patch changeset anyway.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MYWbAFjiCkR8LWJK3VXWX3

PR #17, opened 2024-03-14
Update ai packages 

JS-BC:
Removes unneeded dependencies from various packages. vector sync and vector dupe packages no longer have a direct dependency on embedding or vector database packages

Issue #13, opened 2024-03-07
Specific server error when token has expired


cadam11:
**Package**: @memberjunction/server

To handle automatic refreshing on the client side, we need to know if authentication failed specifically due to an expired JWT.

PR #3698, opened 2026-08-10
Workflow runs get a home: the Runs surface, reachable cost, and one query for the whole tree

AN-BC:
Follow-up to #3692 (merged). That PR moved workflows onto the durable task-graph dispatcher; its review raised eighteen items, and this branch works that list while adding the surface the architecture implied but never had — **somewhere to see what a workflow actually did**.

## The problem this branch is really about

A task graph **outlives the run that submitted it**. That is the whole point of submit-and-detach, and it is also what made runs unreachable: the only way to find one was to open the agent that happened to submit it, which fails exactly when it matters. A scheduled or MCP-triggered workflow has no owning conversation to be found from. Runs were durable and effectively invisible.

Three things follow from that, and they are the spine of this PR.

```mermaid
graph LR
  A[Agent Run] -->|step| B[Task Graph step]
  B -.submit-and-detach.-> C[(Task graph<br/>outlives the run)]
  C --> D[Task] --> E[Agent Run]
  E --> F[Task Graph step] --> G[(...recurses)]
  style C fill:#2d6a9f,color:#fff
```

## 1 · The Workflows app returns — as the home for runs, not a second editor

The app was retired for a good reason: its Create Workflow canvas duplicated the Flow agent editor and had **no Save path anywhere on it**, so a drafted workflow could be admired and never kept. Restoring that dead end would re-earn the retirement.

So what comes back is the part that exists nowhere else — a **Runs** surface listing every graph that has run, by what a person actually remembers: what ran, when, and how it ended. Selecting one opens its graph in the right panel rather than nesting a diagram inside every row.

The front door now *finishes what it starts*: `Workflow.Save` writes the agent and everything hanging off it, then hands the author to the one editor with a real record under them. Saved as **Draft with no triggers** — saving is capture, not scheduling, and a workflow that began running the moment it was named would be a trap.

**The retirement migration stays.** It is already merged to `next` and applied wherever this branch has run; deleting it would rewrite Flyway history and force a from-scratch rebuild on every developer's database to undo something a metadata file expresses directly. The app is instead declared under `metadata/applications` with a **new id** — on an existing database the migration already removed the old row and sync creates the new one; on a fresh install the migration is the no-op it says it is and sync creates it. Both converge on exactly one app, nobody rebuilds, and the history stays honest.

## 2 · A workflow's cost is finally reachable

Two independent holes, both closed:

| | Before | Now |
|---|---|---|
| **Prompt steps** | `AIPromptRun` id returned by the runner and **dropped** — `Task` had nowhere to put it, so prompt spend was invisible to any total | Recorded in a new `Configuration.runtime` slice, merged into the authored bag, written in the same guarded UPDATE as the outcome |
| **Spawned runs** | No `ParentRunID` — every dispatched run looked like it began from nowhere | Set to the submitting run, so `RootParentRunID` resolves and cost is one indexed `SUM` rather than a walk |

Recorded **even when the prompt failed** — a failed prompt still cost tokens, and omitting failures under-reports exactly the runs someone is investigating.

I checked every `ParentRunID IS NULL` site before changing this. All three are additionally scoped (two by `AgentSessionID`, one by `ConversationID` + `AgentID`), so none starts excluding rows it wants; in the one case that could overlap, excluding dispatched runs from "prior conversational turn" is the correct reading.

## 3 · One query for the whole tree

A run is no longer a tree of steps — it is a tree that **changes kind** as it descends. Every consumer wanting the whole story reimplemented that traversal lazily, one level at a time, with its own N+1. Hence "Loading sub-agent steps…" on every expand.

One recursive CTE now returns every node as a row with `ParentNodeID`/`Depth`/`Sequence`; the caller assembles the tree in memory.

- **Cost is per-node, never a rollup** — summing rollups across a tree double-counts every nested run.
- **Depth is bounded by a predicate, not `MAXRECURSION`** — which is T-SQL-only and *errors* when hit, losing the partial answer. A predicate degrades: the tree comes back truncated and `IsAgentRunTreeTruncated` says so. A cycle terminates rather than running away.

Verified against the live database, where it caught a bug the metadata push never would have: **SQL Server forbids an outer join inside a recursive member.** The prompt-cost join moved to the final SELECT. The existing Demo Flow Agent run returns exactly its real shape, Skipped branch and all.

## Review items from #3692

**P0, all fixed** — the XOR claim race (a decided fork's loser could be claimed and executed in the window before `Skipped` was written, irreversibly, since the action had already run); R6 (`Blocked` now means *failure-driven* unsatisfiability only, and the differential simulator had been modelling the correct behaviour all along behind a green suite); the v1-shape prompt templates (both genuinely still taught flat `agentName`/`assignToUser` with **zero** `kind` occurrences, so every draft was rejected outright).

**Also done** — the Prompt runner, which unbroke the shipped `User Onboarding Flow Agent` (six Prompt steps, refused at submission); the depth chain; `failureSemantics: 'edges'` wired end to end; and refusing a Flow agent both as a sub-agent step and as a scheduled target, since both reported success-before-work.

**`maxIterations: 0`** — the reviewer was right and I was wrong. I had implemented the JSDoc (`0=unlimited`) instead of the behaviour; the engines have always computed `Math.min(length, maxIterations ?? default)`. Parity restored and the misleading JSDoc corrected at its source.

## Needs another pair of eyes

🔒 **RLS** — stored queries do not inherit entity row-level security and agent runs are user-scoped. The loader returns what the query returns for the id it is given; callers must establish the user may see the **root** run first, which every current caller does. Written into the loader's own docs so the next caller does not assume otherwise. Flagging for @jordanfanapour.

## Tests

| Package | Result |
|---|---|
| `ai-core-plus` | 28 files / **521** tests |
| `task-graph` | 4 files / **60** tests |
| `ng-dashboards` | 152 files / **1714** tests |

Still to land on this branch: the timeline refactored onto the tree, the Subway Lines visualization, the two demo Flow agents, and their run-tree integration checks.

### DIFF
--- a/metadata/actions/.additional-core-actions.json
+++ b/metadata/actions/.additional-core-actions.json
@@ -5350,6 +5350,34 @@
             "lastModified": "2025-12-30T23:06:55.571Z",
             "checksum": "fbd12bfa05c65c757a03fb0afc939c952b7e3cf1547a44fa2be3bfc2db1a6f18"
           }
+        },
+        {
+          "fields": {
+            "ActionID": "@parent:ID",
+            "Name": "Records",
+            "Type": "Output",
+            "ValueType": "Scalar",
+            "IsArray": true,
+            "Description": "The rows returned by the query. This is the action's result \u2014 map it into the payload to use the records downstream, e.g. {\"Records\": \"accounts\"}.",
+            "IsRequired": false
+          },
+          "primaryKey": {
+            "ID": "93B80E4B-36E6-4CFD-A957-E922A4264B17"
+          }
+        },
+        {
+          "fields": {
+            "ActionID": "@parent:ID",
+            "Name": "TotalCount",
+            "Type": "Output",
+            "ValueType": "Scalar",
+            "IsArray": false,
+            "Description": "Total rows matching the filter, ignoring MaxRows. Useful for branching on whether there is anything to process before doing per-record work.",
+            "IsRequired": false
+          },
+          "primaryKey": {
+            "ID": "A4604694-D070-45FF-BE22-74A3AFC40D33"
+          }
         }
       ]
     },
--- a/metadata/applications/.workflows-application.json
+++ b/metadata/applications/.workflows-application.json
@@ -1,0 +1,31 @@
+{
+  "fields": {
+    "Name": "Workflows",
+    "Description": "Build and run repeatable pathways — the same steps, the same route, every time",
+    "Icon": "fa-solid fa-diagram-project",
+    "DefaultForNewUser": true,
+    "Color": "#2d6a9f",
+    "DefaultSequence": 1004,
+    "DefaultNavItems": [
+      {
+        "Label": "Workflows",
+        "Icon": "fa-solid fa-diagram-project",
+        "ResourceType": "Custom",
+        "DriverClass": "WorkflowsResource",
+        "isDefault": true
+      },
+      {
+        "Label": "Runs",
+        "Icon": "fa-solid fa-timeline",
+        "ResourceType": "Custom",
+        "DriverClass": "WorkflowRunsResource"
+      }
+    ]
+  },
+  "relatedEntities": {
+    "MJ: Application Entities": []
+  },
+  "primaryKey": {
+    "ID": "E29BDE35-88CA-4AC0-8810-1DE5F4DFF2FB"
+  }
+}
--- a/metadata/entities/JSONType-interfaces/ITaskStepConfiguration.ts
+++ b/metadata/entities/JSONType-interfaces/ITaskStepConfiguration.ts
@@ -71,6 +71,36 @@ export interface ITaskStepConfiguration {
      * in the shape they drew, while a machine-authored graph still renders legibly.
      */
     layout?: ITaskStepLayout;
+
+    /**
+     * What actually happened when this step ran — written by the dispatcher, never by an author.
+     *
+     * Everything else in this bag is a step's *definition*; this is its *history*. It lives here
+     * rather than in columns of its own for the same reason as the rest: nothing in it is ever a SQL
+     * predicate, and a column per runtime artefact would be a migration every time a new step kind
+     * produced one.
+     */
+    runtime?: ITaskStepRuntime;
+}
+
+/**
+ * The runtime artefacts a completed step points at.
+ *
+ * **This exists because cost was unreachable for prompt steps.** A workflow's spend is aggregated by
+ * walking the run tree: an Agent step reaches its cost through `Task.AgentRunID` → the run's own
+ * totals. A Prompt step has no agent run — the dispatcher executes the prompt directly — so its
+ * `AIPromptRun`, and with it every token and dollar it spent, had no path back from the Task at all.
+ * The runner returned the id and the dispatcher dropped it on the floor.
+ */
+export interface ITaskStepRuntime {
+    /**
+     * The `MJ: AI Prompt Runs` row this step produced, when it was a Prompt step.
+     *
+     * Set on the step's LAST execution. A retried step overwrites it rather than accumulating: the
+     * prompt runs themselves are the durable history, and this is the pointer to the one whose
+     * output the payload actually carries.
+     */
+    promptRunID?: string;
 }
 
 /** A step's position and size on the canvas, in canvas units. */
--- a/metadata/prompts/templates/sage/workflow-planner.template.md
+++ b/metadata/prompts/templates/sage/workflow-planner.template.md
@@ -51,7 +51,8 @@ The task graph is submitted as `nextStep.type = 'Tasks'` in your response. Here 
                     "tempId": "task1",
                     "name": "Research Data",
                     "description": "Query associations database for revenue and location data",
-                    "agentName": "Research Agent",
+                    "kind": "Agent",
+                    "configuration": { "agentName": "Research Agent" },
                     "dependsOn": [],
                     "inputPayload": {
                         "query": "associations with 5-30M revenue in USA"
@@ -61,7 +62,8 @@ The task graph is submitted as `nextStep.type = 'Tasks'` in your response. Here 
                     "tempId": "task2",
                     "name": "Analyze Market Segments",
                     "description": "Analyze the research data by subsection",
-                    "agentName": "Analysis Agent",
+                    "kind": "Agent",
+                    "configuration": { "agentName": "Analysis Agent" },
                     "dependsOn": ["task1"],
                     "inputPayload": {
                         "data": "@task1.output",
@@ -72,7 +74,8 @@ The task graph is submitted as `nextStep.type = 'Tasks'` in your response. Here 
                     "tempId": "task3",
                     "name": "Create GTM Report",
                     "description": "Generate go-to-market strategy based on analysis",
-                    "agentName": "Marketing Agent",
+                    "kind": "Agent",
+                    "configuration": { "agentName": "Marketing Agent" },
                     "dependsOn": ["task2"],
                     "inputPayload": {
                         "analysis": "@task2.output",
--- a/metadata/prompts/templates/workflow/workflow-drafting.template.md
+++ b/metadata/prompts/templates/workflow/workflow-drafting.template.md
@@ -33,7 +33,8 @@ Return **only** a JSON object of this shape. No prose, no markdown fence, no com
       "tempId": "short-stable-handle",
       "name": "What this step does, as a person would say it",
       "description": "One or two sentences: what this step takes in, and what it produces.",
-      "agentName": "exact name from the available list",
+      "kind": "Agent",
+      "configuration": { "agentName": "exact name from the available list" },
       "dependsOn": ["tempId of a step this waits for"]
     }
   ]
@@ -42,9 +43,16 @@ Return **only** a JSON object of this shape. No prose, no markdown fence, no com
 
 ## Rules that matter
 
-**Every step needs an assignment.** Give each step an `agentName` from the list above. If a step is
-genuinely something a *person* must do — an approval, a judgement call, a physical action — set
-`"assignToUser": true` instead of `agentName`, and never both.
+**Every step declares what KIND it is, and carries the settings for that kind.** A step has a `kind`
+and a `configuration` bag matching it. The two you will use:
+
+- `"kind": "Agent"` with `"configuration": { "agentName": "..." }` — an agent does the work.
+- `"kind": "Human"` with `"configuration": {}` — a *person* must do it: an approval, a judgement
+  call, a physical action.
+
+There is no way to say both, which is the point: a step is one kind. Never put `agentName` at the
+top level of a step — that shape was retired, and a workflow using it is rejected outright rather
+than half-understood.
 
 **`dependsOn` is what makes it a route, not a list.** A step with an empty `dependsOn` starts
 immediately. Steps that could genuinely run at the same time should *both* depend on whatever came
--- a/metadata/queries/.get-agent-run-tree.json
+++ b/metadata/queries/.get-agent-run-tree.json
@@ -1,0 +1,49 @@
+[
+  {
+    "fields": {
+      "Name": "GetAgentRunTree",
+      "CategoryID": "@lookup:MJ: Query Categories.Name=Agents",
+      "UserQuestion": "What is the full structure of everything an agent run caused — its steps, any workflows it started, and the runs those workflows spawned?",
+      "Description": "Returns one row per node of an agent run's complete execution tree: the run, its steps, task graphs those steps submitted, the tasks inside them, and the agent runs those tasks spawned, recursively to a supplied depth cap. Each row carries ParentNodeID/Depth/Sequence so a caller can assemble the tree in memory, plus per-node status, timing, own-cost and own-tokens.",
+      "SQL": "@file:SQL/get-agent-run-tree.sql",
+      "TechnicalDescription": "Replaces the timeline's per-level N+1 (load run, load steps, load graph tasks, load spawned runs, repeat) with a single recursive CTE. Four recursive members: run→steps, task-graph step→its parent Task (linked via JSON_VALUE(OutputData,'$.parentTaskID')), task→child tasks, task→the run it spawned. Depth is bounded by a `Depth < maxDepth` predicate inside every recursive member rather than OPTION(MAXRECURSION), which is T-SQL-only and errors rather than degrading — MAXRECURSION 0 disables the engine limit so the portable predicate is the sole bound. Cost is per-node OWN cost (AIAgentRun.TotalCost, AIPromptRun.Cost), never a rollup, so a caller-side SUM does not double-count nested runs; Prompt tasks reach their cost through Configuration.runtime.promptRunID. Consumed by LoadAgentRunTree + BuildAgentRunTree in @memberjunction/ai-core-plus.",
+      "Status": "Approved",
+      "QualityRank": 9,
+      "ExecutionCostRank": 3,
+      "UsesTemplate": true
+    },
+    "relatedEntities": {
+      "MJ: Query Fields": [],
+      "MJ: Query Parameters": [
+        {
+          "fields": {
+            "QueryID": "@parent:ID",
+            "Name": "agentRunID",
+            "Type": "string",
+            "IsRequired": true,
+            "Description": "The agent run at the top of the tree."
+          },
+          "primaryKey": {
+            "ID": "6C3F8B22-9A41-4E5D-B7C8-1D2E3F4A5B60"
+          }
+        },
+        {
+          "fields": {
+            "QueryID": "@parent:ID",
+            "Name": "maxDepth",
+            "Type": "number",
+            "IsRequired": true,
+            "DefaultValue": "100",
+            "Description": "Hard depth cap. Matches MAX_AGENT_RUN_TREE_DEPTH in @memberjunction/ai-core-plus; any returned row at this depth means the tree may be truncated."
+          },
+          "primaryKey": {
+            "ID": "7D4E9C33-AB52-4F6E-C8D9-2E3F4A5B6C71"
+          }
+        }
+      ]
+    },
+    "primaryKey": {
+      "ID": "344ADDAD-458B-4FBC-9F2E-34915ED5373A"
+    }
+  }
+]
--- a/metadata/queries/SQL/get-agent-run-tree.sql
+++ b/metadata/queries/SQL/get-agent-run-tree.sql
@@ -1,0 +1,197 @@
+/*
+  GetAgentRunTree — one row per node of everything an agent run caused.
+
+  WHY THIS EXISTS
+  ---------------
+  A run's shape used to be assembled by the client one level at a time: load the run, load its
+  steps, notice a step is a task graph, load the graph's tasks, notice a task spawned an agent run,
+  load that run's steps... Each level is a round trip, the count depends on how deep the work went,
+  and nothing could ask "what did this run cost in total" without walking the whole thing first.
+
+  This returns the entire structure in a single call — the run, its steps, any task graphs those
+  steps submitted, the tasks in them, and the runs those tasks spawned, recursively.
+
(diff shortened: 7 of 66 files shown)


AN-BC:
# Post-merge review — #3696 / #3698 / #3701 / #3710 (range `8a89ad3a4..7a1d62f40`) vs. plan §14

Four verification lanes over the merged range (P0-fix claims, run-tree/cost/R7, HITL, scorecard+demos). **The core deliveries are real**: the XOR claim-race fix, R6 in dispatcher + pure layer + simulator, the depth chain, `failureSemantics` persistence and wiring, `maxIterations: 0` parity with the JSDoc fixed at source, both refusals (sub-agent, scheduled), a genuinely solid Prompt runner (Chat escape → Skipped siblings → parent Complete; `PromptRunID` recorded even on failure), both templates on v2, the app restored under a new id with the D18 guard back, the run-tree query correctly shaped (predicate-bounded, prompt-cost join in the final SELECT), and — quietly the most important — **the `MJ: Tests`/suite records committed as declarative metadata**, closing the "deterministic tier can't run from a clean checkout" hole.

But the lanes found **four severe defects, three of them cross-PR interactions this range created**, and the deterministic tier is **red at HEAD**. Findings, worst first; file:line cites are `next` @ `7a1d62f40`.

## Severe

1. **🔴 Every workflow run tree created after #3710 comes back `Root: null` — blank run UI.** #3710's provenance stamp (`parent.AgentRunID = context.AgentRunID`, `TaskGraphService.ts:600`) is the *submitting* run; the tree query's spawned-run member follows `Task.AgentRunID` for `NodeType IN ('Task','TaskGraph')` (`get-agent-run-tree.sql:161-176`), so the submitting run is re-added as a child of its own graph, the CTE loops to the 100 cap, and `BuildAgentRunTree`'s last-write-wins map lets a deep duplicate overwrite the depth-0 root — no row has `ParentNodeID === null` → `Root: null, Truncated: true`. Two semantics now share one column (child task = "run this task spawned"; parent task = "run that submitted me"). **Fix: restrict the fourth CTE member to `t.NodeType = 'Task'`.** The pinned fixtures pre-date the cycle, so nothing catches it.
2. **🔴 `failureSemantics: 'edges'` recovery paths deadlock instead of running.** `handledFailureIDs` is computed and passed to `ComputeTasksToBlock`/`ComputeParentRollup` — but **never to eligibility**: `SATISFIES_DEPENDENT = {'Complete','Skipped'}` (`graph-algorithms.ts:60-64`), so the recovery target of a handled-Failed origin is never Blocked, never Skipped, and **never eligible**. The graph sits `In Progress` forever, stall-logging every poll. Pre-fix it settled `Failed`; now it never settles — a regression for any flow whose step fails with an outgoing edge. The unit tests hand-place the recovery target at `Complete`, which is why they pass. **Fix: thread handled-Failed into eligibility** (plan §5.9's "Complete or handled-Failed" predecessor rule), in both `propagateAndRollup`'s inputs and `findClaimableTasks`.
3. **🔴 Answering a workflow approval spawns a duplicate workflow.** `raiseHumanRequest` sets `OriginatingAgentRunID` and leaves `ResumingAgentRunID` null (`TaskGraphDispatcher.ts:1143`) — which is **exactly the arm-condition of the pre-existing resume hook** (`MJAIAgentRequestEntityServer.ts:41-52`): on Responded it fire-and-forgets `resumeAgent()`, re-running the flow agent, which under the cutover **submits a brand-new graph** — which raises its own new approval. Every dashboard/API approval both settles the task (correct) and starts a duplicate execution. **Fix: one guard — skip resume when `OriginatingTaskID` is set.** The #3710 live verification ("graph went to Complete") is consistent with the duplicate having run unnoticed as a separate parent task.
4. **🔴 The Content Pipeline demo's While loop can never succeed.** Loop bodies don't thread payload (punch item #13, still absent): the While condition closure evaluates the **initial** payload every pass (`TaskGraphDispatcher.ts:1731-1742`), so `payload.brandOK` never becomes true, the loop always burns all 3 iterations re-reviewing the original draft, and the XOR always takes "gave up" — the approved branch is dead code. The flagship demo ships dead-on-arrival against a known-open engine gap.

## The deterministic tier is red at HEAD — three ways

- **TX4** (`task-graph-execution.checks.ts:400-437`) still asserts the **pre-R6** contract (false-conditioned branch `Blocked`, parent `Blocked`) — untouched by the range that changed the semantics; the check's comment now teaches overruled doctrine.
- **WD1** asserts the sweep demo has 2 steps; #3710 made it 3.
- **WD2** is structurally never-green: `dependsOn.includes(<string>)` against object-form dependencies (always false) with the filter direction inverted (`workflow-demo-agents.checks.ts:173-174,190`) — merged without a green run.

Per the Definition of Done, this means the range merged without the deterministic tier passing against its own final state.

## Significant

5. **XOR race reintroduced on the R6 path**: `findClaimableTasks` filters `skipSeedTaskIDs` + `holdTaskIDs` but not **`unreachableTaskIDs`** (`TaskGraphDispatcher.ts:1050-1051`) — a definite-false ordinary edge's target is vacuously eligible in the same window the original fix closed for XOR losers. One more term in the filter.
6. **R7 is not implemented**: `rollUpCostToSubmittingRun` is still invoked at every settlement (`:595, :624-678`). Two cost authorities now disagree on any prompt-bearing workflow (the rollup is prompt-blind; the tree query isn't). Either remove it per R7 or re-open the ruling — the current state is the worst option. Related: #3698's "`ParentRunID` set to the submitting run" was **dead code at its own merge** (nothing wrote `parent.AgentRunID` until #3710 — its commit admits it), and "cost is one indexed SUM over `RootParentRunID`" has no corresponding code.
7. **Loop bodies are second-class citizens across every dimension**: `invokeBody` passes no `SubmittingAgentRunID`/`ContinuationDepth` (`:1719-1727`) → loop-spawned runs get no `ParentRunID` (invisible to the tree and both cost systems), the depth chain restarts at 0 (the unbounded-recursion hole reopened through loops), and **a Flow agent as a loop body evades the sub-agent refusal**. Sub-agent bodies still receive `{}` every iteration (only the prompt-body leg was fixed); loop-repeated prompts lose their `PromptRunID`s (no per-iteration shape in `runtime`); the compiler writes `op.prompt.outputMapping` that **nothing consumes** — the sweep demo's `proposals[index]` mapping is silently dead.
8. **HITL gaps**: a **Cancelled** request is a permanent silent stall (notified-marker set, no re-raise path, `Retry` accepts Failed only — the in-code "waits for whatever replaces it" describes machinery that doesn't exist); `assignToUserID` is compiled and tested but **silently discarded** at persist (`TaskGraphService.ts:659-663` hard-assigns the submitter); expiry is decorative for everything this path creates (nothing ever sets `ExpiresAt`, and `Task.DueAt` is never mapped); `RespondToAgentRequest` never checks the caller is `RequestForUserID` (anyone with request-write can settle anyone's step — the plan's authz contract half-met); the 'Task Assignment' notification deep-link (`type:'Task'`) is unclickable in Explorer's notification UI; graph Cancel leaves the open request `Requested` forever. Also missing: the D17 converter still drops `Human` as "no design-time equivalent" (now false — Agent-Manager-saved workflows lose their approval steps) and the visual editor has no Human node type (metadata files are the only authoring path).
9. **UUID sweep re-broken in the same range that fixed it**: #3696's six fixes were undone by #3701/#3698's new code — 4 fresh violations (`workflow-demo-agents.checks.ts:153-154`; `workflow-runs-resource.component.ts:83,145`), so the MJGlobal compliance backstop is red again.
10. **Failed-decides unscoped**: `ResolveExclusiveGroups` gets `{Complete, Failed}` for **all** graphs (`:1397`) contra §5.3's `'edges'`-only scoping — with an inline comment describing scoping code that doesn't exist. Under `'block'`, XOR losers of a Failed origin now go Skipped where declared semantics say Blocked.
11. **Prose-response Chat escape**: any non-JSON prompt answer becomes `chatMessage` → `endGraphEarly` (`TaskGraphPromptRunner.ts:108-119`, dispatcher `:432-434`) — a summary step that legitimately answers in text silently terminates the whole workflow "successfully."

## Scorecard — the eight still-open §14 items

**Zero shipped in this range.** #4 (input snapshot — `data.*`/`context.*` still evaluate against `{}`; the conversation dialect is now *pinned by test as unresolved*), #10 (envelope truthfulness — a workflow that settles Failed leaves an `AIAgentRun` claiming `Completed` forever), #12 (hold-aware stall — the "stalls visibly" log line is still false both ways), #13 (loop parity — see above), #14 (differential to §9 — fixtures still in-file, no metadata scan, no goldens), #15 (D17 loop-body identity — plus a new wrinkle: prompt-body loops round-trip as Sub-Agent), #17 (joinMode — `'any'→Optional` still emitted; compiler still never reads `AgentTypePromptParams`), #18 (NodeProgress + all hygiene — `policy` still persisted-and-never-read, stale docstrings incl. a "0 means unlimited" comment directly above R5's 0=zero code).

## Suggested fix order

1. Tree-query cycle (one-line member restriction) + the resume-hook guard (one-line) + WD1/WD2/TX4 check fixes — **these four unbreak shipped surfaces and the deterministic tier**.
2. Recovery-deadlock eligibility fix + `unreachableTaskIDs` race filter + Failed-decides scoping — completes R6/failureSemantics for real, with the §9 pins that were promised (recovery-path end-to-end; terminal-conditioned-step settles Complete).
3. Loop-body first-classing in one pass: thread `SubmittingAgentRunID`/`ContinuationDepth`/payload through `invokeBody`, item-to-sub-agent, per-iteration `promptRunID[]` in `runtime`, consume `op.prompt.outputMapping` — this also brings the Content Pipeline demo to life and closes the tree/cost invisibility.
4. HITL hardening: resume guard (done in 1), Cancelled re-raise or terminal ruling, `assignToUserID` honored, `DueAt`→`ExpiresAt` mapping, caller-authz on respond, D17 Human arm, editor palette.
5. R7 decision: remove the rollup write-back or re-rule; then the remaining §14 items (#4 first — it's the only one where running workflows silently misbehave).

---

AN-BC:
# 📣 BUILDER AGENT — fix dispatch for the #3696–#3710 range (review: [5242160684](https://github.com/MemberJunction/MJ/pull/3698#issuecomment-5242160684), HITL detail: [#3710 5242163422](https://github.com/MemberJunction/MJ/pull/3710#issuecomment-5242163422))

The range's core deliveries all verified — Prompt runner, R5/R6, depth chain, refusals, templates, app restoration (keeping the retirement migration was the right call; the plan note is corrected), and the declarative test-suite records. What follows is the fix work, in dependency order. Four verification lanes produced these; every item has file:line evidence in the review comments.

## Batch 1 — four small fixes that unbreak shipped surfaces (do first, one PR is fine)

1. **Run-tree cycle → blank run UI** (breaks every workflow run created since #3710): restrict the tree query's spawned-run member to `t.NodeType = 'Task'` (`metadata/queries/SQL/get-agent-run-tree.sql:161-176`). The parent graph's `AgentRunID` is now "the run that submitted me" (#3710) — following it as a spawned child re-enters the submitting run and cycles to the depth cap; `BuildAgentRunTree`'s last-write-wins map then loses the root. Add a fixture captured from a *post-#3710* run so the pinned trees stop pre-dating the semantics.
2. **HITL duplicate-run bomb**: guard the legacy resume hook — in `MJAIAgentRequestEntityServer.ts:41-52`'s `shouldResume`, skip when `OriginatingTaskID` is set. Today every approval of a workflow human step also re-runs the flow agent → submits a duplicate graph → raises a duplicate approval.
3. **Deterministic tier is red at HEAD, three ways**: TX4 (`task-graph-execution.checks.ts:400-437`) still asserts the pre-R6 contract — update to Skipped/Complete and fix its comment, which now teaches overruled doctrine; WD1 asserts 2 sweep steps vs the 3 that exist post-#3710; WD2 (`workflow-demo-agents.checks.ts:173-174,190`) compares `dependsOn.includes(<string>)` against object-form dependencies with the filter direction inverted — structurally never-green, so it merged without a run. Re-run the tier green before closing the PR.
4. **UUID sweep re-broken**: `workflow-demo-agents.checks.ts:153-154` (use `UUIDsEqual`, not `.toLowerCase()===`), `workflow-runs-resource.component.ts:83,145` (raw `===` on IDs). Also `TaskGraphDispatcher.ts:695` `sibling.ID === task.ID` — it evades the compliance test only via the same-line-literal loophole.

## Batch 2 — complete R6/failureSemantics for real

5. **Recovery-path deadlock (CRITICAL)**: `handledFailureIDs` reaches `ComputeTasksToBlock`/`ComputeParentRollup` but never **eligibility** — `SATISFIES_DEPENDENT` is `{'Complete','Skipped'}` (`graph-algorithms.ts:60-64`), so a handled-Failed origin's recovery target is never eligible and the graph hangs In Progress forever (pre-fix it at least settled Failed). Thread handled-Failed into eligibility per plan §5.9's "Complete or handled-Failed" predecessor rule, in both `propagateAndRollup`'s inputs and `findClaimableTasks`. Write the end-to-end pin the plan named: Failed origin + satisfied recovery edge → recovery **runs** → graph settles Complete. (The existing unit tests hand-place the recovery target at Complete — that's why they pass; don't repeat the shape.)
6. **XOR race, R6 edition**: add `unreachableTaskIDs` to the claim filter (`TaskGraphDispatcher.ts:1050-1051` filters skip-seeds and holds but not unreachables — a definite-false edge's target is vacuously eligible in the same window you closed for XOR losers).
7. **Failed-decides scoping**: `:1397` passes `{Complete, Failed}` to `ResolveExclusiveGroups` for **all** graphs; plan §5.3 scopes Failed-decides to `'edges'` specs, and the adjacent comment describes scoping code that doesn't exist. Scope it (the pure layer already supports it and has the test).
8. **Prose Chat-escape**: any non-JSON prompt answer currently ends the whole workflow "successfully" (`TaskGraphPromptRunner.ts:108-119` → `endGraphEarly`). Require the explicit signal (`nextStep.type==='Chat'` / `taskComplete`) — a text answer is a payload, not a termination.

## Batch 3 — loop bodies become first-class (one pass; also brings the Content Pipeline demo to life)

9. In `invokeBody` (`TaskGraphDispatcher.ts:1719-1727`): pass `SubmittingAgentRunID` + `ContinuationDepth` (closes the depth-chain restart AND the Flow-as-loop-body refusal bypass AND makes loop-spawned runs visible to the tree and cost); thread payload iteration-to-iteration (sequential) so a While condition can see what its body produced — today the Content Pipeline demo's `payload.brandOK` reads the initial payload forever and the approved branch is dead code; give sub-agent bodies the item as payload (still `{}` every iteration — only the prompt leg was fixed); accumulate per-iteration `promptRunID`s (make `runtime.promptRunID` an array or add `promptRunIDs[]`) so prompt-body loops stop losing their cost; consume `op.prompt.outputMapping` (the compiler writes it, nothing reads it — the sweep demo's `proposals[index]` mapping is silently dead). Then verify both demos end-to-end — the While demo reaching its approved branch is the acceptance test.

## Batch 4 — HITL hardening (#3710 comment has the detail)

10. Cancelled request = permanent silent stall: either clear the notified-marker on cancel so a fresh request is raised, or rule Cancelled terminal — the current "waits for whatever replaces it" comment describes machinery that doesn't exist. `assignToUserID` is compiled+tested then silently discarded (`TaskGraphService.ts:659-663` hard-assigns the submitter) — honor it. Map `Human` config → `Task.DueAt` → `ExpiresAt` so expiry stops being decorative. Check the caller is `RequestForUserID` (or elevated) in `RespondToAgentRequest`. Add the D17 `Human` arm (Agent-Manager-saved workflows currently lose their approval steps) and the editor palette entry. Graph Cancel should close open requests.

## Rulings needed from AN-BC (don't guess)

- **R7**: `rollUpCostToSubmittingRun` is still live at every settlement (`:595`) — the ruling said drop it, and it now disagrees with the tree SUM on any prompt-bearing workflow (the rollup is prompt-blind). Remove it, or ask AN-BC to re-rule "both, with the rollup taught about prompt runs."
- Demo-agent descriptions use graph/acyclic/AND-join/dispatcher vocabulary in end-user-visible fields (D18) — flag for a wording pass or a ruling that agent Descriptions are dev-facing.

**Still open from §14 after all the above**: #4 input snapshot (highest priority — running workflows still evaluate `data.*`/`context.*` against `{}`, and `payload-mapping.test.ts:74` now *pins the conversation dialect as unresolved*), #10 envelope truthfulness, #12 hold-aware stall, #14 differential-to-§9 (metadata-driven + goldens), #15 D17 loop-body identity (plus the new prompt-body→Sub-Agent mislabel), #17 joinMode, #18 NodeProgress + hygiene (including the "0 means unlimited" JSDoc sitting directly above the R5 fix in `TaskLoopExecutor.ts:279-288`, and the stale `findUnrunnableKinds` docstring).

Deviation protocol unchanged: anything that fights back goes to the PR thread before the plan moves.

---

AN-BC:
Read it properly. **This is a good review and most of it is right** — including the one that matters most, which I shipped and my own live verification walked straight past.

Fixed on `feat/workflow-hitl` (PR #3710), verified before fixing rather than accepted on inference:

## Fixed

**#3 — duplicate workflow on approval. You are right, and I confirmed it in the database.** Approving one request at 13:56 produced a second, unasked-for Schema Documentation Sweep graph at 14:05. The resume hook arms on exactly the shape a workflow approval has, and a null `ResumingAgentRunID` — which I chose deliberately, and wrote a paragraph defending — is what armed it. One guard: skip resume when `OriginatingTaskID` is set.

Worth naming why I missed it: I checked that the *approved* graph completed and never looked for a graph I had not asked for. Both executions look successful, so nothing surfaces. That is the same silent-success signature as four other bugs in this range, and I should have been looking for it.

**#1 — tree cycle.** Found and fixed independently a few hours before your review landed (`3e5e9e74f6`), same one-line diagnosis: restrict the fourth member to `NodeType = 'Task'`. Your framing is sharper than mine though — *two semantics now share one column* is the actual defect; my fix is the consequence. `next` is still broken until #3710 merges.

**WD1 / WD2.** Both yours. WD1 asserted 2 steps when HITL had just made it 3. WD2 was structurally never-green — `.includes(<string>)` against object-form dependencies *and* the direction inverted. Merged without a passing run, which is the part I own; a check that cannot pass is worse than no check. Both fixed, with a helper that reads either dependency form.

**#9 — UUID sweep.** Four raw comparisons in files your sweep had just cleaned. Now `UUIDsEqual`.

## Confirmed, not yet fixed

**#4 — the Content Pipeline While can never succeed.** You are right and this reframes something I reported as working. I observed `iterations: 3, succeeded: 3` and the gave-up branch taken, and read it as the loop functioning. It was the dead-branch case: the condition closure captures the initial payload, so `payload.brandOK` cannot change and the approved branch is unreachable. **The flagship demo is dead on arrival**, and I called it verified. That is a reporting failure, not just a code one.

**#2 — recovery deadlock.** Traced it: `SATISFIES_DEPENDENT = {Complete, Skipped}`, `handledFailureIDs` never reaches eligibility. Pre-fix it settled Failed; now it never settles. Agreed this is a regression, and agreed the unit tests hide it by hand-placing the recovery target at `Complete`.

**#5, #6, #7, #10, #11** — all read as correct on inspection. #7 in particular (loop bodies missing `SubmittingAgentRunID`/`ContinuationDepth`, so the depth cap reopens *through loops*, and a Flow agent as a loop body evades the sub-agent refusal) is the one I would put next after the deadlock — it is a security-shaped hole, not just an observability one.

**#8 — HITL gaps.** Fair, all of them. `assignToUserID` compiled, tested, and discarded at persist is the embarrassing one — I wrote a test that passes against code that throws the value away. Cancelled-as-permanent-stall is real; my "waits for whatever replaces it" describes machinery that does not exist, and I should have written *that* rather than implying a path.

## Your fix order

I would take it as written, with one swap: **#7 before #2.** The recovery deadlock stalls a workflow visibly; the loop-body gap silently reopens unbounded recursion and lets a Flow agent slip past a refusal that exists to prevent success-before-work. A visible stall is cheaper than an invisible bypass.

On **R7** — agreed the current state is the worst option. My vote is to remove `rollUpCostToSubmittingRun` rather than re-rule: the tree query is prompt-aware and the write-back is not, so keeping both guarantees they disagree on exactly the workflows people care about.

Thanks for this. Four of the nine findings I would not have found by looking harder at my own diff.

PR #3700, opened 2026-08-10
fix(core): a related-record Load() no longer silently discards unsaved children

AN-BC:
## The defect

`RelatedRecordCollection.Add()` and `.Create()` deliberately do not mark a collection loaded — an
appended child says nothing about what is on disk. That is correct, and it has a consequence nobody
had followed through: a collection you have **only appended to** still has `IsLoaded === false`, so
`Load()`'s "already loaded" early return does not protect it.

```typescript
await order.Load(id);
await order.Lines.Create();     // one unsaved line; IsLoaded is still false
await order.Lines.Load();       // items replaced wholesale — the line is gone
```

The load runs, `this.items = result.Results ?? []` replaces everything, and the caller's unsaved
children vanish. Any queued removals go with them, so a row the user deleted comes back.

**Nothing reports it.** The screen shows fewer rows than the user typed, the save writes fewer
children than they entered, and every total agrees with every other total the whole way down. That is
the same class of failure this file's own doctrine already names, in the remark explaining why a
*failed* load throws instead of yielding an empty collection:

> Silently returning no children makes a populated parent look empty, and anything derived from that
> — a reversal, a total, a validation decision — is then wrong in a way nothing downstream can detect.

## How it is reached

Not a contrived sequence:

- a `Load: 'lazy'` collection appended to before its first read
- a component that refreshes on a route event while an editor holds unsaved lines
- two sibling components sharing one entity, one of which loads

Found while moving two BizApps order/journal editors onto related-record collections: both compose
children in the browser before anything has loaded the collection, which is precisely the window.

## The fix

`Load()` throws when the collection is dirty and `force` is not set, and names what would have been
lost:

```
RelatedRecordCollection 'Lines': cannot load over unsaved changes — 1 unsaved child record(s) and
0 pending removal(s) would be discarded. Save the parent first, or pass force to discard them
deliberately.
```

`Load(true)` still discards. That is what a deliberate refresh means, and saying so is cheap.

`Dirty` already answers exactly the question the guard needs — any retained child unsaved or dirty,
or any removal pending — so the guard is one condition rather than new bookkeeping.

**Merging was rejected.** It invents an ordering between what the user typed and what the database
holds, and it can duplicate a child that was saved by another path in between. Refusing is the same
choice this collection makes everywhere else.

## Compatibility

A `Load()` that previously discarded now throws. Any caller relying on the old behaviour was relying
on losing data; if the discard was intended, `force` says so. The no-op paths are untouched:
`LoadMode: 'never'`, an unsaved parent, and an already-loaded clean collection all return exactly as
before — the guard sits after all three.

## Verification

- 6 new tests in `relatedRecordCollection.load.test.ts`, covering the refusal, the message contents, the forced discard, the clean load, the pending-removal case, and the unsaved-parent no-op
- `@memberjunction/core`: **1829 tests pass**, 120 files
- Docs updated in `packages/MJCore/docs/related-record-collections.md`

### DIFF
--- a/.changeset/related-collection-load-over-unsaved.md
+++ b/.changeset/related-collection-load-over-unsaved.md
@@ -1,0 +1,5 @@
+---
+"@memberjunction/core": patch
+---
+
+`RelatedRecordCollection.Load()` now refuses to load over unsaved work instead of silently discarding it. `Add()` and `Create()` deliberately do not mark a collection loaded, so a collection that has only ever been appended to still has `IsLoaded === false` — and the "already loaded" early return therefore did not protect it. Any later `Load()` (a lazy read, a refresh, a sibling component sharing the entity) replaced `items` wholesale and took the caller's unsaved children with it, along with any queued removals, reporting nothing: the screen simply showed fewer rows than the user typed and the save wrote fewer children than they entered. The load now throws when the collection is dirty and `force` is not set, naming how many children and removals would have been lost; `Load(true)` still discards, which is what a deliberate refresh means. Merging was rejected — it invents an ordering and can duplicate — and refusing matches what this collection already does elsewhere, where a load that would produce quietly wrong data throws rather than returning something plausible.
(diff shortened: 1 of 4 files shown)


AN-BC (commented):
LGTM

PR #3703, opened 2026-08-10
feat(postgresql): run the integration tier on PG, and fix what it exposed

MS-BC:
Closes the long-open question in **#3257**: the deterministic integration tier now runs against PostgreSQL for the first time, and this PR fixes the defects that running it exposed.

**Result: 58 of 61 deterministic bundles pass on PostgreSQL** (MJAPI live, 61/61 executed, zero skipped, `RUN_MUTATION_TESTS=1`). It was 32/61 when the harness first came up.

> **Numbers reconciled (2026-08-13).** Two figures in this description disagreed — 49/61 in this
> headline and 51/61 further down — which @rkihm-BC rightly flagged. Both are superseded, and the
> reason matters more than the arithmetic: **neither run set `RUN_MUTATION_TESTS=1`**, so
> mutation-class bundles skipped silently and the number flattered the branch. Re-measured with the
> flag on the current head, the honest figure is **58/61**, and the three failures are named below
> rather than averaged away. Later fixes on this branch closed IT01, IT02, IT06, IT22 and IT69,
> which is why the pass count rose even as the bar was raised.

> Built on `next` + #3697 + #3590 + #3351. Those close identifier quoting, the IS-A path, and migration conversion; everything here is what remained on top of them.

## Why this was needed

#3257 records that the suite is meant to run twice per build — once per backend — and that it never has. PostgreSQL therefore shipped with **migration parity verified and runtime parity unverified**, which is exactly the half that RunView/RunQuery SQL generation, provider code paths, UUID casing and identifier quoting live in.

## What's here

**Harness — the #3257 blocker list.** `testing-cli` branched on nothing: it always built an `mssql` pool, so `DB_PLATFORM` was never read on the `mj test` path. It now has a PostgreSQL path (dynamic import, declared as an `optionalDependency` so SQL-Server-only consumers never resolve `pg`), a PG-native user-cache load, `dbPlatform` on `MJConfig`, and a `getContextUser()` that resolves the same user on both backends. `mj.config.cjs` gains `dbPlatform` and a platform-aware `dbPort`; with `DB_PLATFORM` unset both are byte-identical to today.

**Runtime dialect leaks.**
- `TaskClaimStore` emitted `SELECT @@ROWCOUNT` — T-SQL only. On PostgreSQL the `@@` is consumed as a parameter marker and the bare `ROWCOUNT` folds to lowercase, so **every guarded write failed**: 7,168 `column "rowcount" does not exist` in a single tier run, now zero. The wrapper moves to `SQLDialect.AffectedRowCountSQL()` so the next caller can't reintroduce it.
- `MJDashboardEntityExtended` denied the **legitimate owner**. `Validate()` is synchronous and reads `DashboardEngine`'s cache directly; in `task` startup mode (the default) engine pre-warm is deferred, so an unloaded cache is indistinguishable from "no permission". This is what actually broke `mj sync push` — on a dashboard whose `UserID` *was* the pushing user. Not a PostgreSQL bug at all; PG just made it reachable.

**Conversion — five T-SQL→PG defects.** Each was caught only by applying the output to a fresh database; the converter reported `0 errors` every time. The nastiest silently discarded **every** `IF EXISTS (…)` batch, taking a guarded `DROP CONSTRAINT` with it and exiting 0. Per the derive-don't-guess rule, the missing `CHECK`-body keyword set came from intersecting 2,084 `CHECK` bodies across 67 shipped migrations against the dialect keyword list — exactly `CASE`, `WHEN`, `THEN`, `ELSE`, `END`, no more.

**Also.** `callRoutineSQL` couldn't invoke a PG `RETURNS SETOF record` routine with no OUT parameters, so entity pruning died and cascaded into 22 missing CRUD routines. CodeGen's PG audit-SQL folder swap was pinned to `v5` by exact match, so on v6 it wrote into the SQL Server tree. `applyLLMPrimaryKeys` validated case-insensitively then used the model's spelling in the `UPDATE` — zero rows matched on PG, and it still reported success. `R__RefreshMetadata` on PG now clears orphaned `EntityField` rows, as the SQL Server file always has.

**Two fixes are to tests, not product code** — the aggregates bundle passed `MAX(__mj_UpdatedAt)` unquoted and the open-app-teardown fixture called `SYSDATETIMEOFFSET()`. Both are SQL-Server-only spellings.

**Five `.pg.sql` counterparts** for the v6 tail, produced by the toolchain (not hand-authored) and gated on a clean `mj migrate` against a virgin database.

## SQL Server is unaffected

Every changed path is PostgreSQL-only or a same-output refactor; `AffectedRowCountSQL` keeps `@@ROWCOUNT` on SQL Server and `callRoutineSQL`'s new flag is ignored there.

| package | tests |
|---|---|
| SQLDialect | 335 passed |
| SQLConverter | 1134 passed (includes re-converting the whole committed v5/v6 ledger) |
| MJCoreEntities | 588 passed |
| TaskGraph | 60 passed |
| testing-cli | 23 passed |
| CodeGenLib | 782 passed — 10 failures in `pg-connection.test.ts` are **pre-existing**, verified identical with these changes stashed |

## Evidence

`PG_PARITY_AUDIT_FINDINGS.md` is included: full method, every finding with its measurement, and the bootstrap sequence. Highlights — fresh-apply gate green (45 migrations, virgin DB, full v6 tail); CRUD oracle **365 pass / 4 fail**, matching the documented historical baseline; MJAPI boots and serves on PostgreSQL (HTTP 200).

## Deliberately not fixed

- **`R__RefreshMetadata` is 7 reconciliation routines on SQL Server vs 2 on PostgreSQL.** `sp_refreshview` has no PG equivalent, so a from-scratch PG database still ends with metadata describing view columns that don't exist (measured: 12 fields / 8 entities, including core `vwEntities`). This is the largest remaining gap and needs a design decision, not a patch.
- **Layered base views** — shipped metadata enables a SQL-Server-only feature on two entities, so `mj codegen` cannot exit 0 on PG. Needs a way for metadata to be platform-scoped.
- **11 of 21 stored queries have no PostgreSQL dialect variant** and fall back to T-SQL verbatim. Authoring those is content work, not a code fix.
- The camelCase tokenizer gap stays latent behind its existing tripwire — no repo code reaches it, and quoting would change result-key casing for consumers.
- `next` is currently red on **both** backends for an unrelated reason (a duplicate `MJ: Query Parameters` key from #3698); not touched here.

## Not included

No CI lane yet. Nothing keeps this true until one exists — that's the natural follow-up.

---

The audit above measures migrations, CRUD, and unit suites. This section records a run that
booted the **whole product** on PostgreSQL — API, GraphQL transport, and the Angular client
under Playwright — because that is the only way to catch the seams those tiers mock away.

**Environment.** Merged with #3590 and #3697 onto `next` in a scratch integration branch.
Live PostgreSQL (`MJ_PG_AUDIT_pgparity`, 377 entities). MJAPI boots and serves (schema
assembly 4.1s, ready in ~9s). MJExplorer built and served against that API, magic-link auth.
Playwright headless Chromium. All **271** workspace packages build.

### Integrations — the WRITE path, 18/18

The earlier sweep proved integrations *reads*. Reads exercise a view; they say nothing about
the generated stored procedures, key generation, the FK graph, or versioning. This drives a
full lifecycle across `Company → CompanyIntegration → Integration`:

| # | Check | Result |
|---|---|---|
| 1 | Company created on PostgreSQL | pass |
| 2 | `spCreate` returned a **usable generated key** (not a row you then cannot find) | pass |
| 3 | CompanyIntegration created across the FK to `Integration` | pass |
| 4 | `RunView` finds the new record | pass |
| 5 | joined `Integration` name resolved through the view | pass |
| 6 | boolean round-tripped as a **boolean**, not a string | pass |
| 7 | timestamp round-tripped as a **Date**, not a driver string | pass |
| 8–9 | update saved and landed in the base table | pass |
| 10 | **Record Changes fired on PostgreSQL** (1 → 2 rows) | pass |
| 11–12 | both `Create` and `Update` were versioned | pass |
| 13 | the Update change row captured a **real diff** (`Description changed from … to …`) | pass |
| 14 | a write missing a NOT NULL column was **refused** | pass |
| 15 | **no partial row left behind** by the rejected write | pass |
| 16 | connection still usable afterwards — **no aborted-transaction poisoning**, a PostgreSQL-specific failure mode where every subsequent command on a poisoned connection is rejected until rollback | pass |
| 17–18 | teardown deleted both records; **no test rows left in the database** | pass |

Checks 10–13 matter most: Record Changes is MJ's built-in versioning, and if it were
silently inert on PostgreSQL every audited entity would lose its history on that backend
with nothing appearing to fail. It is not inert — Create, Update and Delete are all
recorded, with real field-level diffs.

*(Check 10 initially reported a failure. `RecordChange.RecordID` stores MJ's composite-key
**string** form, `ID|<uuid>`, not a bare UUID, so matching on the raw id returns 0 forever.
That would have condemned working versioning as dead; the query was wrong, not the code.)*

### The dashboard permission gate, on a genuinely cold cache — 6/6

The defect this PR fixes is that an unconfigured `DashboardEngine` cache is
indistinguishable from "this dashboard grants you nothing", so the owner gets refused. The
test never calls `DashboardEngine.Config()`, so the cache really is cold — reproducing the
condition rather than simulating it. **Owner allowed; non-owner still refused.** The fix
resolves ownership from the record's own `UserID`, so restoring the owner's access did not
trade the gate open for everyone else.

### GraphQL transport, privileged user

- **Integrations** — `MJ: Integrations` (2 rows), `MJ: Companies`, `MJ: Company Integrations`
  all present in metadata and answering `RunView` over the wire.
- **Dashboards** — 15 rows; the owner passes `Validate()` **and** `Save()` through the
  transport, re-proving the fix at the layer a user actually hits.
- **Core view sweep** — `MJ: Users`, `Roles`, `Entities`, `Applications`, `AI Agents`,
  `AI Prompts`, `Queries`, `Audit Logs`: **8/8** answered.

### Explorer UI under Playwright — 10/10

Magic-link redemption → click-through interstitial → authenticated shell at `/app/admin`;
Angular mounts; the router resolves a real route; back/forward navigation is clean; Explorer
talks to the PG-backed API and **every HTTP response is < 400**; no failed network requests.

Two things worth stating rather than burying:

- **Scope.** The only auth reachable offline is magic-link, and such a session is
  deliberately confined. A privileged Explorer session needs a real IdP. During development
  of this test the server's provisioning guard correctly refused to attach a magic-link
  session to an account holding non-restricted roles — that guard was left **intact** rather
  than weakened to make the test easier, and data-level coverage was obtained with a
  privileged user at the transport layer instead.
- **Four console notices, classified not hidden.** A magic-link session lacks read
  permission on `MJ: User Applications` / `MJ: User Settings`, and Explorer's app-loading
  path reads `UserInfoEngine` properties without first consulting `IsPermissionConstrained`
  — the thrown error says so in as many words. These are a property of restricted sessions,
  platform-agnostic, and unrelated to this PR; the test prints them explicitly and excludes
  them from the pass/fail signal.

### One defect this run found, in this PR's own code

`packages/TaskGraph` calls `db.Dialect.AffectedRowCountSQL(...)` — added here — but never
declared `@memberjunction/sql-dialect`. The type arrived only transitively through
`DatabaseProviderBase.Dialect`. Under pnpm's strict layout there is no hoisted copy to fall
through to, so TypeScript resolved the name **outside the workspace**; against a copy
predating this branch the method does not exist and the build dies with

```
TaskClaimStore.ts(260,28): error TS2339:
  Property 'AffectedRowCountSQL' does not exist on type 'SQLDialect'
```

cascading to the eight packages downstream and leaving MJAPI without a `dist`. It is
invisible in any tree whose out-of-workspace copy happens to be current, which is how it
survived CI. Fix is one line — declare the dependency — plus the lockfile entry.

---

## The integration tier, actually run on PostgreSQL

This PR's title is *"run the integration tier on PG"*, so the tier was run — headless,
against the live PG database, with MJAPI serving from this branch.

```
[SUITE_COMPLETE] Integration Tests — Deterministic
[SUMMARY] 51/61 passed (83.6%)
```

**51/61 (83.6%)**, against the **47/61** recorded for this branch earlier — a net gain of 4.

*(A first run reported 36/61. That run was pointing at `localhost:4111`, which on this
machine is a different project's server, and 25 bundles failed identically with HTTP 401.
The tier resolves its endpoint from `MJAPI_URL` / `GRAPHQL_PORT`; once pointed at the right
server the real number appeared. Recorded because 36/61 would have been a misleading
regression report.)*

### Current failures (2026-08-13, `RUN_MUTATION_TESTS=1`, head `302cd42714`)

| Bundle | Failing checks | Cause | Status |
|---|---|---|---|
| **IT29 — Cache Gauntlet** | subset-slot × mutation | A mutation-class bundle. It did **not run at all** in the earlier 49/61 and 51/61 numbers, which is the single largest reason those were optimistic. | newly visible, under investigation |
| **IT67 — Content Vectorization** | CV6 content-source save | PG's `spCreateContentSource` still names `ScheduledActionID`, a column the Phase-0 retirement migration (`V202608061704`) **drops**. The regenerated PG proc is stale, not the test. | PG-parity gap; regeneration is release-time toolchain work |
| **IT74 — Task Graph Execution** | TX8 settle-marker, TX9 delivery timeout | Deterministic — three consecutive isolated runs of the bundle fail identically, so not flaky. | open, not yet root-caused |

The bucketed list below is retained as the historical record from the 51/61 run; IT01, IT02, IT06,
IT22 and IT69 have since been fixed on this branch.

### The 10 remaining failures as of the 51/61 run, bucketed by cause

| Bucket | Bundles | Cause | Status |
|---|---|---|---|
| **Stored-query dialect gap** | IT01, IT02, IT03, IT33, IT34 | Queries with no PG variant run T-SQL verbatim | **already disclosed by this PR**; counterparts arrive with the next release-time conversion |
| **Layered base views** | IT69 | SQL-Server-only feature enabled in shipped metadata | **already disclosed by this PR** |
| **CodeGen artifact drift** | IT50 | `entity_subclasses.ts` registers 374; live metadata has 375 core entities | fixture/bootstrap property |
| **T-SQL `[bracket]` identifiers** | IT22 | `syntax error at or near "["` | see below |
| **T-SQL `uniqueidentifier` type** | IT06 | `type "uniqueidentifier" does not exist` | see below — **not previously disclosed** |
| **Unified search returns null** | IT52 | `SearchEntities` resolver returned null | **not previously disclosed** |

**The stored-query gap, measured.** This PR states "11 of 21 stored queries have no
PostgreSQL dialect variant… Authoring those is content work, not a code fix." Queried
directly against the live database:

```
total queries ................... 21
with a PostgreSQL variant ....... 10
falling back to T-SQL verbatim .. 11
```

The count is exact and the characterisation is right. PG counterparts are produced by the
release-time conversion toolchain, not authored per-PR, so the 11 are simply queries whose
counterparts have not been generated yet — the conversion after `v6.1.0-edge.1` picks them
up. Nothing here is a defect to fix on a feature branch.

They do, however, **fail on execution** in the meantime rather than degrading gracefully,
which is what the catalog bundles are reporting:

```
runquery-catalog.QC1: bare-query sweep      — 4 of 4  catalog queries failed
runquery-catalog.QC3: derived-param sweep   — 7 of 10 catalog queries failed
PostgreSQLDataProvider.ExecuteSQL failed: syntax error at or near "["
```

This accounts for **5 of the 10** failing bundles. The two pending queries are content work
as this PR says; the nine missed ones are a defect in the conversion run and belong to the
toolchain, not to a future authoring pass.

**The bracket errors split into two different causes**, which matters because only one is a
product problem:

- In `runquery-cache`, `syntax error at or near "["` is the stored-query gap above.
- In `user-routines` the same error is labelled
  `ExecuteSQL failed [user-routines: null NextRunAt for the seeding fixture]` — it is the
  **integration test's own seeding SQL**, authored in T-SQL. UR11–UR14 then fail as
  cascade, because UR10's seeding write never happened. That is test-authoring work this
  PR's own "no CI lane yet" follow-up would have to finish, not a defect in the product.

**Two failures are genuinely new and not covered by this PR's disclosures:**

1. **`type "uniqueidentifier" does not exist`** in the RLS post-image rejection path
   (IT06 / `rls-isolation.KF5`). The rejection is supposed to carry the message
   `…outside your permitted row scope`; instead the statement dies on a T-SQL type name. It
   is **not** authored RLS content — 0 of the 13 `RowLevelSecurityFilter` rows mention
   `uniqueidentifier` — and it is not a labelled test fixture. I did not isolate the
   emitting call site; recorded as reproducible-but-unlocated rather than guessed at.
2. **`SearchEntities` returns null** (IT52 / `search.SR2`, `SR3`), so the lexical pass finds
   nothing for a name that demonstrably exists. My first hypothesis was PostgreSQL's
   case-sensitive `LIKE`; the log disproves it — the resolver returns null before any
   comparison happens.

Neither blocks this PR — both are pre-existing on PG and neither is touched by these
changes — but both belong on the record rather than inside a 83.6% pass rate.

## Divergence classes, measured rather than assumed — 13/13

Identifier quoting fails loudly. The dangerous divergences are the ones that return a
*different answer* without erroring. Each class below was measured twice: once through the
raw driver to establish what PostgreSQL actually does, and once through MJ to see whether MJ
absorbs it.

| Class | PostgreSQL behaviour (measured) | MemberJunction |
|---|---|---|
| String equality collation | case-**sensitive** (`'mj: entities'` → 0 rows, `'MJ: Entities'` → 1) | **absorbed** — `EntityByName` resolves any casing |
| UUID literal casing | `uuid::text = UPPERCASE` → 0 rows | **absorbed** — uuid typing matches either casing through `RunView` |
| NULL ordering | `ORDER BY v ASC` over `{1,NULL,2}` → `[1,2,null]` = **NULLS LAST**; SQL Server sorts them FIRST | **not absorbed** — a "first row" query over a nullable column returns a *different row* per backend, with no error |
| Boolean representation | real booleans | correct — `AllowCreateAPI=true` arrives as `boolean` |
| NUMERIC typing | driver default is string | correct — `InputPricePerUnit` arrives as `2.5` (`number`), so arithmetic adds rather than concatenates |
| Timestamps | `timestamptz` | correct — arrives as a valid `Date` |
| `LIKE` | case-**sensitive** (`'MJ:%'` → 375, `'mj:%'` → 0) | **not absorbed** — a LIKE-based search written against SQL Server returns nothing on PG unless the dialect supplies `ILIKE` |

The two "not absorbed" rows are engine facts callers must know about, not defects in this
PR — recorded because they are precisely the silent-wrong-answer class that a green smoke
test cannot see.

*(Two entries in this table were wrong on the first pass and are corrected here: NULL
ordering was inferred from `EntityField.Precision`, which has **0** nulls in 5,539 rows, so
the evidence could not have shown NULLS FIRST even if it were true; and the NUMERIC probe
picked its column with a `/price|cost|rate/` regex that matched the UUID foreign key
`PriceTypeID` and reported a false finding. Both now measure what they claim to.)*

## Unit tests — 2,909 passing, 0 failures

Every package touched by this stack, per the repository's definition of done:

| Package | Result |
|---|---|
| `@memberjunction/task-graph` | 60/60 |
| `@memberjunction/sql-dialect` | 335/335 |
| `@memberjunction/sql-converter` | 1,134 passed, 3 skipped |
| `@memberjunction/codegen-lib` | 792 passed, 66 skipped |
| `@memberjunction/core-entities` | 588/588 |

Worth noting against this PR's own table: CodeGenLib is reported above as "782 passed — 10
failures in `pg-connection.test.ts` are pre-existing". With a live PostgreSQL running, those
10 pass — **792 passed, 0 failures**.


MS-BC:
---

## Review fixes + end-to-end verification

Three review findings addressed, and a new defect found by running IS-A end-to-end on live PostgreSQL.

### Review findings fixed

**1. The dashboard permission gate had gone fail-open.** Guarding on `DashboardEngine.Instance.Loaded` skipped the check entirely whenever the cache was cold, so a non-owner passed a gate that previously refused them. The safety argument rested on entity permissions and RLS underneath — asserted rather than demonstrated.

Ownership does not need the engine: `UserID` is on the record being validated. Resolving the owner from the row first fixes the actual defect — `mj sync push` refused on a dashboard whose `UserID` *was* the pushing user — while leaving the gate **fail-closed** for everyone else. A non-owner whose grant lives only in the engine is still refused on a cold cache; that is the pre-existing behaviour for that case, deliberately not opened up in a PG-parity change.

**2. `as any as IMetadataProvider`.** `ProviderToUse` returns `IEntityDataProvider`, so a bridge is genuinely needed — but `baseEntity.ts` spells it `as unknown as IMetadataProvider` in eight places. Aligned all three occurrences in the file with core, including the two that predate this PR.

**3. `stripComments` / `quoteConstraintNames` exported camelCase** → `StripComments` / `QuoteConstraintNames`, across all seven referencing files.

The dynamic import of `postgresql-dataprovider` was checked and is compliant — it names its category-2 exception and the `testing-integration` precedent.

MJCoreEntities 588 passed, SQLConverter 1134 passed, both `tsc --noEmit` clean.

### New defect found by the E2E, fixed here

**CodeGen could not create the bucket Application for a new schema on PostgreSQL:**

```
Failed to create new application isatest for schema isatest
error: column "id" of relation "Application" does not exist
```

`__mj."Application"` was quoted by `qs()`, but the column list was authored bare — `(ID, Name, Description, …)` — so `ID` folded to `id`.

The root cause is an **auto-quoter bypass**. `runQuery` and `runQueryWithParams` route through `this.qsql(sql)`; `LogSQLAndExecute` calls `ds.query()` directly and does not. All **61 of its call sites in CodeGenLib** therefore reach PostgreSQL exactly as authored, and #3697's tokenizer work cannot help on that path because the tokenizer is never invoked.

Fixed the one live instance with explicit `qi()`. Routing `LogSQLAndExecute` through `qsql()` would change 61 call sites at once, including DDL that may not survive tokenisation, so the architectural gap — the CodeGen twin of #3427 — belongs in its own issue rather than this PR.

### End-to-end result

Stacked with #3590 and #3697 against a real PostgreSQL 17 database, with a purpose-built IS-A fixture (the deterministic tier does not cover IS-A — every entity had `ParentID` NULL). **19/19 behavioural checks passed**: one `Save()` produced a parent row and a child row sharing one ID, mirrored parent fields persisted and read back through `RunView`, and the parent's `Length` / `Values` columns — two keyword colliders — survived every stage.

Also confirmed from the same run: **layered base views fail on PostgreSQL exactly as this PR's "deliberately not fixed" section documents**, so that disclosure is accurate.

rkihm-BC (commented):
# Code review

**Verdict: two blockers, both mechanical, both invisible in the current CI signal because the run never gets past `pnpm install`.** The substance underneath is strong; the branch state is not shippable as-is.

I checked out the head commit (`1734de7dc6`), ran the SQLConverter suite (**1,134 passed** — matches the claim), ran the CodeGenLib IS-A test, and re-ran the converter over the migration ledger. This PR is stacked and carries #3697 wholesale — I reviewed that tokenizer work on #3697 and don't repeat it here.

---

## Blockers

### 1. All five failing checks are one bad `package.json` — and it's a duplicate JSON key

Every failing job dies identically:

```
pnpm-lock.yaml is not up to date with <ROOT>/packages/TestingFramework/CLI/package.json
```

`packages/TestingFramework/CLI/package.json` has `@memberjunction/generic-database-provider` **twice inside the same `dependencies` object** — once at `6.1.0-edge.2` (from the version tooling) and again at the end at `6.1.0-edge.1` (hand-added here). JSON last-wins, so the effective pin is `6.1.0-edge.1`. `optionalDependencies` pins `@memberjunction/postgresql-dataprovider` at `6.1.0-edge.1` too. Both packages are actually at `6.1.0-edge.2`, and every other manifest in the workspace pins `-edge.2`.

An exact pin that no workspace package satisfies stops pnpm linking the local copy and sends it to the registry instead — which is precisely the "resolved the name **outside the workspace**" failure this PR diagnoses so well for TaskGraph. Same bug, one directory over.

Fix: delete the duplicate line, move both to `6.1.0-edge.2`, `pnpm install` at the root, commit `pnpm-lock.yaml` (it isn't in the diff at all).

### 2. The IS-A CodeGen fix is not in the branch — only its test and its changeset are

`packages/CodeGenLib/src/Database/manage-metadata.ts` on this branch is **byte-identical to `next`**. `buildISAEntityLookupSQL` exists nowhere in the tree. Commit `8bd26c05` (which added it, +88/−28) *is* an ancestor of the head, but `git log -- manage-metadata.ts` on the branch doesn't list it: one of the merges resolved that file to the other side and silently dropped the change. The test file and `.changeset/codegen-isa-postgres-portability.md` survived.

I ran it:

```
FAIL src/__tests__/isa-postgres-portability.test.ts
TypeError: this.buildISAEntityLookupSQL is not a function
Tests  6 failed | 2 passed (8)
```

Three consequences: CodeGenLib's suite is red; the changeset ships release notes for a fix the release won't contain; and the body's "codegen-lib 792 passed, 0 failures" is no longer true at this head.

Worth re-checking the other merge-heavy paths for the same loss. I spot-checked and `callRoutineSQL`, `applyLLMPrimaryKeys` and the five v6 `.pg.sql` counterparts *are* present (they landed in `next` separately), so IS-A looks like the only casualty — but that's a spot check, not a proof.

---

## Concerning

### 3. The dashboard owner bypass now trusts the dirty in-memory `UserID` in every case

`Validate()` short-circuits on `!UUIDsEqual(this.UserID, currentUser.ID)`. `this.UserID` is the *current* field value, not the persisted one — so a caller who can load the row, sets `UserID` to their own id, and saves, passes the gate outright: the ownership test is true and the permission check never runs.

To be precise about what changed: `next` already has `const ownsRecord = !DashboardEngine.Instance.Loaded && UUIDsEqual(this.UserID, currentUser.ID)`, so trusting the dirty value is **pre-existing** — but gated on a cold cache. Moving the same comparison to the outer condition widens it to every case, including the warm-cache server path where the old code refused. That's the production path, so the widening is the part that matters.

Compare against the persisted owner instead — `this.GetFieldByName('UserID').OldValue`, or refuse when that field is dirty. Entity permissions and RLS still sit underneath, but this class *is* the row-level gate.

Two things introduced here, unambiguously:

- The `ownsRecord` line and its ten-line comment are **dead code**. The outer condition already guarantees the user is not the owner, so `ownsRecord` is always `false` and `!permissions.CanEdit && !ownsRecord` reduces to `!permissions.CanEdit`. The comment still describes the previous approach ("Falling back to it keeps the owner able to save"), which now contradicts the code six lines above it.
- `Delete()` solves the same problem a *different* way (`await DashboardEngine.Instance.Config(...)`). Reasonable given it's async, but the asymmetry deserves a note, and it changes the failure mode: a throwing `Config` now propagates where `Delete()` previously returned `false`.

No committed regression test for any of it. The 6/6 cold-cache checks in the body were an ad-hoc script; a permission-gate change should leave a test behind, especially one whose whole point is that the cold-cache path is the one nobody exercises.

### 4. `QuoteConstraintNames` mangles `DROP CONSTRAINT IF EXISTS`

`/\bCONSTRAINT\s+([A-Za-z_]\w*)\b/gi` captures `IF` as the constraint name:

```
ALTER TABLE __mj."Foo" DROP CONSTRAINT IF EXISTS "CK_Foo_Bar";
  =>  ALTER TABLE __mj."Foo" DROP CONSTRAINT "IF" EXISTS "CK_Foo_Bar";
```

I re-converted the five T-SQL migrations in the repo that use that syntax — four come out broken. **This is pre-existing, not a regression**: I re-ran the same probe with `next`'s rule files swapped in and got byte-identical output. So it's not a blocker.

It's still the right PR to fix it in: this change refactors exactly that regex, widens its reach to `CREATE TABLE`, and separately adds a code path (`tryConvertGuardedDropConstraint`) that *emits* the very syntax the regex breaks. Excluding `IF` is a one-word fix. It's also the counterexample to the "derive don't guess" discipline the body rightly claims for the CHECK-body keyword set — that set was derived from 2,084 real bodies; this regex wasn't checked against anything.

### 5. `refreshUserCacheFromPG` writes another package's private field by string index

```ts
(UserCache.Instance as unknown as Record<string, unknown>)['_users'] = userInfos;
```

This is the `unknown`-as-shortcut the typing rule exists to prevent, and its failure mode is the silent one the surrounding comment itself warns about: rename `_users` and you get an empty cache, no compile error, and "no users" rather than a failure. The comment correctly identifies `UserCache.RefreshFromRows` as the fix and files it elsewhere — but adding a public seam is small, and this is about to become load-bearing for the CI lane that's the stated follow-up.

---

## Questions, not findings

**`tryConvertGuardedCreateSchema` now emits lowercase unquoted schema names.** The reasoning is sound and the mixed-case/lowercase split it fixes is real. I checked whether this breaks already-applied migrations and it does not — the only committed `migrations-pg` files carrying a quoted `CREATE SCHEMA` are the four pg_dump baselines, which this converter path doesn't produce. So no Flyway checksum problem here. The open question is downstream: an open-app repo whose PG database already holds `"__mj_BizAppsSecureMessaging"` will get a second, empty `__mj_bizappssecuremessaging` on re-conversion. One sentence for the build engineer would close it.

---

## Missing / nits

- `aggregates-cache.checks.ts` reaches `QuoteIdentifier` through `as unknown as { QuoteIdentifier?: … }` and silently falls back to the bare, broken form when absent. Matches the file's existing `connStrOf` style, but sits oddly next to a review-fix commit that made a point of tightening casts.
- `mj.config.cjs` matches `DB_PLATFORM === 'postgresql'` exactly; `postgres` or `pg` silently gets port 1433.
- `sql_logging.ts`: the new regex also redirects *explicit* `./migrations/...` overrides, where the old code honored them. Probably what you want and the comment says so — just a change beyond the stated bug. It also leaves a stray double blank line where the constants were.
- `R__RefreshMetadata.pg-only.sql` header says "Skyway" where it means Flyway. (The file itself is fine — `*.pg-only.sql` is an established category with 37 existing members, and I confirmed `spDeleteUnneededEntityFields` ships in the PG baseline, so the `PERFORM` won't fail on a from-baseline deploy. The `'sys,staging'` argument matches the T-SQL file.)
- `PG_PARITY_AUDIT_FINDINGS.md` — 591 lines of point-in-time audit at the repo root. That's a PR artifact or a `plans/` document; the root already has fifteen `.md` files.
- **No CI lane**, as the PR states. Everything here decays the day after merge without one, and since most of this PR *is* the harness, a nightly PG lane is what converts it from a measurement into a guarantee. Worth filing before merge rather than after.
- Two numbers disagree with themselves: the tier is 49/61 in the summary and 51/61 in the later section; CodeGenLib is "782 passed, 10 pre-existing failures" in one table and "792 passed, 0 failures" in another. The second pair is reconciled in the text; the first isn't.

---

## What's good

- **It actually closes #3257's blocker list.** `testing-cli` never read `DB_PLATFORM` on the `mj test` path at all; giving it a real PG path, a platform-aware `dbPort`, and a `getContextUser()` that resolves identically on both backends is the durable part of this PR. The dynamic import is properly justified — category 2, named in the comment, declared in `optionalDependencies`, exactly as the rule requires.
- **`AffectedRowCountSQL` is the right shape.** Dialect-owned so it can't be reintroduced, SQL Server output byte-identical to before, and the contract (one DML, no trailing semicolon, no `RETURNING`) documented on the abstract method. I checked all four `TaskClaimStore` call sites — every one satisfies it. 7,168 `column "rowcount" does not exist` in a single tier run going to zero is a real number, not a vibe.
- **The dashboard diagnosis is exactly right**, and framed honestly: "Not a PostgreSQL bug at all; PG just made it reachable." An unloaded cache being indistinguishable from a denial is the kind of thing that only surfaces when something changes which records are dirty.
- **Every SQLConverter fix ships with tests** (+297 lines across five test files), and the suite passes — I ran it. `StripComments` before deciding *what a batch is* is a genuinely good insight: a rule that matches prose instead of code and then names a schema after a word from a comment is a wonderful bug to have found.
- **The disclosure quality is unusually high.** The "deliberately not fixed" section, the ten remaining failures bucketed by cause, the stored-query gap measured against the live database rather than asserted — and, best of all, the corrections of its *own* earlier measurements: NULL ordering inferred from a column with zero nulls, a NUMERIC probe whose regex matched a UUID foreign key, a 36/61 run pointed at the wrong port. Retracting three of your own findings is worth more than a higher pass rate.
- The two integration-test changes are correctly classified as test bugs rather than product bugs, and both go through the dialect rather than hardcoding the PG spelling.

---

**Changeset levels are correct as filed.** Flagging this because a naive run of `npm run check:changeset` says otherwise: it reports "Branch touches no migration and no metadata" and objects to the two `minor` entries. That's the script being wrong, not the PR — it globs only `migrations/**` and `metadata/**`, so it misses `migrations-pg/v5/R__RefreshMetadata.pg-only.sql`. A repeatable PostgreSQL migration is a database change and `minor` is right. Worth fixing the script separately; it will mis-flag every PG-migration branch.

Land #3697 first, then: fix the manifest and lockfile, restore the IS-A commit's `manage-metadata.ts` change (or drop its test and changeset), and fix the `this.UserID` comparison. Everything else above can follow.

MS-BC:
Thanks — this was a genuinely useful review, and both blockers were real. Pushed as `e35d01a49a` + `302cd42714`.

## Blockers

**1. The duplicate manifest key.** Fixed: one `generic-database-provider` entry at `-edge.2`, the optional `postgresql-dataprovider` moved to `-edge.2`, and **`pnpm-lock.yaml` is now in the diff** (it wasn't — you were right that its absence was the whole failure). Verified on the pushed head, not just locally: the manifest now has exactly one occurrence.

**2. The IS-A fix was missing.** Confirmed exactly as you described — `8bd26c05` is an ancestor, `git log -- manage-metadata.ts` doesn't list it, and `buildISAEntityLookupSQL` existed nowhere. Restored; IS-A went from **6 failed / 2 passed → 8/8**. I checked the 28 deletions the restore introduces are precisely the inline SQL the helpers replace, so nothing branch-specific was lost.

**2b. Your "worth re-checking the other merge-heavy paths" caveat paid off — there is a second casualty.** `PostgreSQLCodeGenProvider.ts` was 2699 lines here, byte-identical to `next`; on #3697 it is **2512**, because `d388271122` deletes 196 lines moving `_SQL_KEYWORDS` into the shared tokenizer. Same signature as the IS-A loss: the commit is an ancestor, its **test changes survived**, its **source refactor did not**, and `git log --` on the file doesn't list it. That accounted for 22 failing `quoteSQLForExecution` tests. I 3-way-applied the lost diff rather than taking #3697's file wholesale, since three later commits touch it here.

While in there, the remaining 10 CodeGenLib failures (`pg-connection.test.ts` — the ones the description called pre-existing) turned out to be a Vitest 4 issue: the mock used `vi.fn(() => ({…}))`, and an arrow function has no `[[Construct]]`, so `new PGConnectionManager()` threw. A function expression fixes it. **CodeGenLib: 32 failures → 0, 808 passing.**

## Concerning

**3. The dashboard owner bypass.** Fixed the way you prescribed, and both ways: the gate now reads the **persisted** owner (`GetFieldByName('UserID').OldValue`) *and* refuses when that field is dirty. The dead `ownsRecord` line and its contradicting comment are gone. On the `Delete()` asymmetry — you were right that it wasn't just cosmetic: a throwing `Config()` propagated where `Delete()` previously returned `false`, changing the contract for every caller on the cold-cache path alone. That call is now guarded. **12 committed tests**, and I mutation-checked them: removing the fail-closed guard fails 4, trusting `this.UserID` instead of `OldValue` fails 3, and un-guarding the `Config()` await fails 1.

**4. `QuoteConstraintNames` and `DROP CONSTRAINT IF EXISTS`.** Fixed with the one-word exclusion, plus four tests covering the keyword in both cases, an ordinary mixed-case name, and an all-lowercase one. Agreed on your framing — this file widened the pattern's reach *and* added a path that emits the syntax it broke, so it round-trips through one module now.

**5. The `['_users']` write.** Replaced with a public `UserCache.SetUsers()` seam. Your point about the failure mode is the reason it was worth doing rather than deferring: the cast type-checks forever and a rename yields an empty cache with no error.

**Question — `CREATE SCHEMA` lowercase.** The downstream hazard is now written into the header builder as a note for the build engineer: a PG database populated by an earlier converter already holds the quoted case-preserved name, so re-conversion creates a second empty folded schema rather than reusing it. Confirmed, as you did, that no committed `migrations-pg` file is affected.

**Nits.** All fixed — `aggregates-cache` now throws instead of silently substituting the bare broken form; `mj.config.cjs` routes `DB_PLATFORM` through the same contract as `resolveDbPlatformFromEnv`, so it is case-insensitive and the legacy aliases *throw* rather than quietly taking port 1433 (both `dbPlatform` and `dbPort` were reading it raw); `sql_logging` blank line; Skyway → Flyway; and `PG_PARITY_AUDIT_FINDINGS.md` moved to `plans/pg-parity-audit-findings.md`.

**CI lane** — filed as #3773 before merge, as you suggested, with the current known-failing bundles as its starting baseline.

## The numbers — you were right, and it's worse than an inconsistency

You flagged 49/61 vs 51/61 as unreconciled. Both are superseded, and the cause is not arithmetic: **neither run set `RUN_MUTATION_TESTS=1`**, so mutation-class bundles skipped silently. The honest current figure on `302cd42714` is **58/61**, and the description has been corrected to say so rather than quietly swapping one number for another.

The three failures, named rather than bucketed away:

| Bundle | Cause |
|---|---|
| **IT29 — Cache Gauntlet** | mutation-class; it never ran in either earlier number, which is the single biggest reason they flattered the branch |
| **IT67 — Content Vectorization** | CV6 — PG's `spCreateContentSource` still names `ScheduledActionID`, a column `V202608061704` **drops**. Stale regenerated proc, not a test bug. Regeneration is release-time toolchain work. |
| **IT74 — Task Graph Execution** | TX8 settle-marker + TX9 delivery timeout. Deterministic — three consecutive isolated runs fail identically — and **not yet root-caused**. |

IT74 is the one I'd flag for you specifically, since it sits on the `AffectedRowCountSQL` work you checked all four call sites of. I have not established whether it is a PG defect the harness is newly exposing or something this branch introduced, and I'd rather say that than guess.

Full build green (299/299). Re-requesting review.

PR #3707, opened 2026-08-10
docs(core): document the normalized RunView 'simple' contract

AN-BC:
> **Merge after #3709.** Rewritten per maintainer direction: #3709 now ships always-on simple-row
> normalization, so these docs describe the unified contract that change establishes rather than
> the old tier-dependent behavior.

## What the docs now say

`'simple'` rows are plain objects whose `Date` and numeric columns are normalized to real `Date`s and `number`s on **every** tier — server or browser, fresh query or cache hit — matching the generated entity types. The docs then state exactly what a plain row still cannot honor when a generated entity type is passed as `T`:

- no entity methods (`row.Save()` compiles and crashes),
- closed-union columns hold whatever string the database held,
- `Fields`-narrowed properties simply don't exist on the rows,
- view-only runs (`ViewID`/`ViewName` with neither `EntityName` nor a loaded `ViewEntity`) skip normalization.

**Rule of thumb: want entity behavior — methods, validation, save, related records — ask for entity objects.** Choose `'simple'` for cheap read-only rows; its dates and numbers are already the declared types.

## Corrections from the original version of this PR

The first draft documented `'simple'` as "exactly what the transport carried — no conversion happens anywhere in the pipeline." Review against the code showed that was **false server-side**: the mssql driver returns real `Date` objects and `AdjustDatetimeFields` runs on every RunView result regardless of `ResultType`, so server-side simple rows already carried real `Date`s while client-side rows carried ISO strings. Two smaller corrections for the record: the "~51 call sites" calibration figure was low by roughly 20× (about 1,100 `ResultType: 'simple'` occurrences repo-wide, ~700 outside tests — of which ~60–90 pass a generated entity type and are the actual risk class), and `form-builder-resource.component.ts` was cited as an omitted-`ResultType` example but actually passes `ResultType: 'entity_object'` via a params variable.

## What changed

- `ResultType` TSDoc in `runView.ts`: the normalized contract, what `T` still doesn't check, the two caveats, and the rule of thumb
- MJCore README `RunView` section: comparison table and the same guidance
- Changeset

Docs only — the behavior itself ships in #3709. MJCore builds clean; 1829 tests pass on this branch.


AN-BC (commented):
LGTM

AN-BC:
Closing to simplify the workstream: the documentation now ships inside #3709 alongside the behavior change it describes (commit `54f0e0ead` folds the rewritten `ResultType` TSDoc and README section into that PR). One PR to review instead of two, and no merge-order dependency.

---

PR #3709, opened 2026-08-10
feat(core): always-on type normalization for RunView 'simple' rows + first-party client sweep

AN-BC:
One PR, three layers: the behavior change, its documentation, and the first-party client sweep that makes the change safe to ship. (#3707 was closed and folded in here to simplify the workstream.)

## 1. The behavior change (`@memberjunction/core`)

The value a `'simple'` read returned for a `DATETIME` column depended on where the code happened to run:

| Path | Shape before |
|---|---|
| Fresh server-side query | real `Date` (driver-parsed, `AdjustDatetimeFields`-adjusted) |
| Server-side Redis cache hit | ISO **string** (`JSON.parse`, no reviver) |
| Browser client over GraphQL | ISO **string** (rows are `JSON.stringify`'d on the wire) |

Same call, three shapes — while the generated entity types declare exactly one. `'simple'` rows (and omitted-`ResultType` rows) are now normalized at the `ProviderBase` choke point every provider's pipeline flows through — client and server, fresh query and cache hit — so `Date` and numeric columns hold real `Date`s and `number`s matching the generated types. **Always-on, no flag**: an opt-in only helps callers who already know about the mismatch, leaves the tier inconsistency in place, and creates per-caller identity problems in dedup/coalescing that a flagless design cannot have.

Design decisions worth review:

- **Rows keyed by field `Name`** (the batch transport's actual row shape), with the `CodeName` key also normalized when a row carries both (the singular transport aliases it on).
- **Zero-copy fast path:** rows already in the right shape — the common server-side case — keep array and row identity untouched. A row is shallow-copied only when a cell actually converts, which keeps cache-held rows safe (the in-memory server store holds them by reference).
- **BIGINT precision guard:** integer strings beyond `Number.MAX_SAFE_INTEGER` stay strings — the PostgreSQL provider deliberately returns those as strings, and `Number()` would silently corrupt them.
- **`NULL` stays `NULL`** (never epoch-1970), unparseable values are left as-is (never `Invalid Date`), existing `Date`s pass through — the pass is idempotent on every path.
- **Entity lookup via `EntityByName`** with a `ViewEntity → EntityByID` fallback. View-only runs (`ViewID`/`ViewName` with neither `EntityName` nor a loaded `ViewEntity`) skip normalization — documented; pass `EntityName` alongside the view identifier to get normalized rows.
- Skipped for `entity_object` (BaseEntity already converts) and `count_only`.

## 2. Documentation (folded in from #3707)

The `ResultType` TSDoc and the MJCore README now describe the unified contract: dates and numbers truthful on every tier, plus what a caller-supplied `T` still cannot promise (no entity methods, closed-union columns hold whatever string the database held, `Fields`-narrowed properties are absent, the view-only gap). Rule of thumb: want entity behavior — methods, validation, save, related records — ask for entity objects; `'simple'` is for cheap read-only rows whose dates and numbers are already the declared types.

## 3. First-party client sweep

A full audit of Angular/Explorer, Angular/Generic, Bootstrap/BootstrapLite, MJExplorer, and MobileApp found **5 real breaks** (all in Explorer dashboards) and **4 display regressions**, all fixed here:

**Hard breaks** — would throw or silently misorder the day this ships:
- Integration daily-counts chart: `.startsWith()` on a date cell (TypeError)
- Classify item grid: `localeCompare` comparator on the default-sorted Updated column (TypeError on first render) — now epoch-ms with a numeric comparator
- Prompt Run Analysis: default `RunAt` sort fell through to `String(Date).localeCompare` — weekday-alphabetical "chronological" order
- Version History labels: the same misorder on the default Date sort
- Knowledge Hub analytics: `String(st).slice(0, 10)` day-bucket filters that never match a `Date` — three charts silently empty

**Display fixes**: Prompt Run CSV export pins the ISO shape; Integration sparkline tooltip, available-resources extra columns, and MobileApp entity-browser subtitles/record detail format `Date`s readably instead of `Date.toString()`.

**New shared helpers** (tolerant of `Date`, ISO-string, and epoch-ms shapes, so they are correct before and after this ships): `date-cell.ts` in ng-dashboards (`DateCellTime` / `DateCellDayKey` / `DateCellIso` / `CompareDateCells`) and `FormatSimpleRowCell` in ng-resource-permissions. Integration dashboard row interfaces and formatter signatures widened to the honest transitional `Date | string | null`.

## Verification

| Package | Result |
|---|---|
| `@memberjunction/core` | **1846 tests pass** (17 new normalization tests) |
| `@memberjunction/ng-dashboards` | **1727 tests pass** (13 new), builds clean |
| `@memberjunction/ng-resource-permissions` | **36 tests pass** (4 new), builds clean |
| `@memberjunction/mobile-app` | **173 tests pass** (3 new) |

Pinned by test: the identity fast path (already-`Date` rows keep array and row identity), cache safety (converting rows are copies; originals untouched), the BIGINT guard, Name-key and dual-key normalization, and the weekday-name-sort discriminating case.

## Follow-ups (not in this PR)

- A deterministic integration bundle asserting date-shape parity across the GraphQL transport and the server tier, including a cache-hit case — the long-term enforcement.
- `RunQuery` rows and RunView `AggregateResults` still carry transport shapes; parity there is a separate change.
- App-code note for downstream developers: defensive `new Date(x)` wrapping and the Angular `date` pipe keep working unchanged; string ops on date cells need the same treatment as the sites fixed here. A memo for UX developers is circulating separately.


AN-BC:
Reviewed at AN-BC's request. Short version: **the bug is real and the analysis is better than the fix.** I'd not merge this, and I don't think the type-level version you describe is quite the right shape either — there's a variant of it that is strictly better than both.

## What's genuinely good here

The cache-poisoning reasoning is the best part of this PR and it would have been easy to miss:

> `ProjectRowsToFields` deliberately returns the original array when the projection would be a no-op … mutating in place would write `Date`s into the cache entry, so the next reader of that fingerprint, one that never asked for coercion, would silently receive `Date`s.

That's a cross-caller version of the exact bug being fixed, it's subtle, and it's pinned by a test. Same for leaving `NULL` alone rather than letting it become 1970, and leaving unparseable values as-is rather than writing `Invalid Date` — which stringifies to `"Invalid Date"` and destroys the evidence. Whoever picks this problem up next should keep all three decisions regardless of which fix lands.

The motivating case is also the right kind of bug to care about: an allocator whose totals stay exactly correct while the *ordering* silently changes. That's the class where nothing fails and the answer is wrong.

## Why I'd still not merge it

**1. It doesn't reach the ~51 existing sites.** Those are the actual exposure. An opt-in flag only helps someone who already knows about the mismatch — and someone who knows doesn't write the bug. The people it bites are the ones who won't set the flag, so the risk profile is essentially unchanged.

**2. It makes the value less wrong while leaving the type dishonest** — which you say yourself, and I think undersells the cost. After `CoerceTypes: true`, `RunView<OrderEntity>` on a simple read is *still not an OrderEntity*: the closed-union `Status` holds whatever string the database had, and there are no methods. We'd have traded a wrongness that breaks loudly (`.getFullYear is not a function`) for one that mostly works. Mostly-works is the shape that survives review and ships.

**3. It forks "how do I get typed rows" into two answers.** Today the rule in `data-access.md` is crisp — `entity_object` when you need methods or mutation, `simple` when you want plain fast rows. This adds `simple + CoerceTypes` as a third, partial thing, and the next reader has to learn when each applies.

**4. It charges the cost on the path chosen for not paying costs.** One `new Date()` per date cell is small next to the query, but `'simple'` exists precisely for callers who want the rows untouched.

## The option I'd actually pursue

You framed the type-level fix as "overloads constraining `T` by `ResultType` … a compile break across MJ and every app, probably a major." I don't think it has to be, and the non-breaking form is better than both proposals:

```typescript
/** A row as the TRANSPORT delivers it: Dates arrive as ISO strings. */
type RawRow<T> = { [K in keyof T]: T[K] extends Date ? string
                                 : T[K] extends Date | null ? string | null
                                 : T[K] };

// overload: a 'simple' read RETURNS RawRow<T>, whatever T the caller passed
function RunView<T>(params: RunViewParams & { ResultType: 'simple' }): Promise<RunViewResult<RawRow<T>>>;
```

Nobody edits a type argument. Existing `RunView<OrderEntity>({ ResultType: 'simple' })` keeps compiling and starts yielding `RawRow<OrderEntity>` — so `.OrderDate.getFullYear()` **stops compiling at exactly the ~51 sites where it was already wrong**, and nowhere else. The break is scoped to code that was broken, which is the only kind of break worth shipping.

That gets you what this PR cannot: it fixes the existing sites, it costs nothing at runtime, and the type becomes *true* rather than approximately true. It doesn't fix the closed-union problem either — nothing runtime or type-level can, short of `entity_object` — but it stops claiming to.

Worth checking before committing to it: how `RawRow` behaves through `ResultType` being omitted (the default path), and whether any generated entity type has a `Date` nested inside an object property, where the mapped type won't recurse.

## Recommendation

- **Take #3707 now** — the documentation half stands on its own and is the highest value-per-risk thing here.
- **Don't merge this**, but keep the three runtime decisions above; whoever writes the real fix will need the cache-poisoning one.
- **Prototype the `RawRow` overload** and see what the ~51 sites actually look like. If most are genuinely reading strings, they'll compile untouched and the change is nearly free. If many are wrong, that's the strongest possible argument for doing it — and the count itself is the finding.

One correction to the framing, offered as a peer: "throw it out if the trade-off is not worth it" undersells this. The investigation is worth keeping even if the code isn't — I'd fold the cache-poisoning note and the NULL/unparseable rules into #3707 so they survive the PR being closed.

AN-BC (commented):
LGTM

AN-BC:
@rkihm-BC @cadam11 @jordanfanapour @EL-BC @MS-BC @bc-izygmunt @CaelebB-BC please read this, very minor tweak but important making sure you all see this.

PR #3710, opened 2026-08-10
A workflow can wait on a person — HITL on AI Agent Requests

AN-BC:
Stacks on #3701 — review that first, or read this diff against it.

A workflow can now wait on a person.

## The gap

The **executable** vocabulary already supported it: `Task.StepType` accepts `'Human'`, the dispatcher recognises the kind, the claim loop notifies whoever it is assigned to. The **design** vocabulary did not — `AIAgentStep.StepType` offered `Action | ForEach | Prompt | Sub-Agent | While`, so an author had no way to draw the step the runtime could already run.

Both shipped demo workflows were *specified* with an approval and shipped without one for exactly this reason, which is why `Schema Documentation Sweep` proposed field descriptions rather than applying them.

## AI Agent Requests, not a second substrate

`MJ: AI Agent Requests` already models everything this needs — who is being asked, what for, a typed `ResponseSchema`, priority, expiry, and an inbox people already use. A parallel HITL store beside it would split that inbox in two and leave one half without expiry or permissions.

**What it deliberately does *not* set is `ResumingAgentRunID`,** and that is the design point:

> A request normally **suspends** an agent run and resumes it. A workflow needs none of that — the graph **outlives** the run that submitted it. Nothing is suspended: the task sits Pending, every other branch keeps running, and answering settles the task and releases its dependents.

That column staying null is meaningful, not missing. The graph case is *simpler* than the loop-agent case, not harder.

`ResponseData` becomes the task's `OutputPayload`, which is what makes a human step **useful rather than a gate** — a downstream edge can branch on what the person actually said. A step recording only "approved" would force every real decision back into a separate action.

```mermaid
graph LR
  A[Propose descriptions] --> H{Human: Approve}
  H -->|Approved / Responded| W[Downstream work]
  H -->|Rejected / Expired| G[Give-up branch]
  style H fill:#2d6a9f,color:#fff
```

## Two bugs found by running it

**`AIAgentRequest.AgentID` is NOT NULL, and a human task has no `AgentID`** — that column names what *executes* a step, and a person is not an agent. The request now carries the agent that **owns the workflow**, which is who is asking.

Resolving that owner exposed the real defect: **`parent.AgentRunID` was never set.** The submitting run was recorded only inside the parent task's `InputPayload` JSON — and a JSON field cannot be joined, so the provenance was unavailable to any query. That is why a human step had no agent to ask on behalf of, and the same missing link would have starved anything else walking from a graph back to the run that started it.

The notified-marker is now written **only once a request is genuinely open**. It exists to prevent a notification storm, but a *missing* request is the worse failure: nothing appears in anyone's inbox and the workflow waits forever with no indication why. Retrying next poll is the recoverable choice.

## Expiry is enforced, not decorative

A deadline nothing acts on is a comment. An overdue request is expired and its task settles `Failed`, so a give-up edge can route around it — rather than a workflow blocking forever on someone who may have left the company.

`Canceled` is deliberately **not** terminal for the task: a cancelled request means the *ask* was withdrawn, not that the step was decided, so the task keeps waiting for whatever replaces it.

## Verified end to end, twice

`Schema Documentation Sweep` now carries the approval it was designed around, and **stops at it**:

```
[Action]  Find undocumented fields             Complete
[ForEach] Propose a description for each field Complete
[Human]   Approve the proposed descriptions    Pending
REQUESTS: 1 — Requested · task linked · originRun set · resumingRun null
```

Approving that request settled the task and the graph went to **Complete**. Both halves of the loop confirmed against a live database.

## Tests

`ai-core-plus` **525** (4 new, covering an unassigned human step, an explicit assignee, and a human step mid-graph with dependents behind it) · `task-graph` **60**.

The PostgreSQL counterpart is deferred to the release build, per `migrations/CLAUDE.md`.


AN-BC (commented):
LGTM

AN-BC:
**Post-merge review flag — two severe defects trace to this PR's mechanism** (full review with all findings: [#3698 comment](https://github.com/MemberJunction/MJ/pull/3698#issuecomment-5242160684)).

1. **Answering a workflow approval spawns a duplicate workflow run.** `raiseHumanRequest` sets `OriginatingAgentRunID` with `ResumingAgentRunID` null — exactly the arm-condition of the pre-existing resume hook (`MJAIAgentRequestEntityServer.ts:41-52`), which re-runs the flow agent on Responded; under the cutover that submits a brand-new graph, which raises its own new approval. The PR's "that column staying null is meaningful" reasoning was right about the graph not needing resume — but the legacy hook doesn't know that. Fix: skip resume when `OriginatingTaskID` is set.
2. **The provenance stamp (`parent.AgentRunID`) breaks the #3698 run tree** — the tree query's spawned-run member follows `Task.AgentRunID` for `TaskGraph` nodes too, so the submitting run re-enters as a child of its own graph: cycle → root overwritten → `Root: null` → blank run UI for every workflow run created after this merge. Fix: restrict that CTE member to `NodeType = 'Task'`.

Also from this PR's scope: Cancelled requests deadlock the task permanently (marker set, no re-raise, `Retry` is Failed-only — the "waits for whatever replaces it" comment describes machinery that doesn't exist); `assignToUserID` is compiled and tested but silently discarded at persist (submitter hard-assigned); no workflow-raised request ever gets `ExpiresAt`, so expiry is decorative here; `RespondToAgentRequest` doesn't verify the caller is `RequestForUserID`; and the D17 converter + visual editor still lack the `Human` arm, so only metadata files can author the step this PR made executable.

---

PR #3713, opened 2026-08-10
chore(vscode): stop the editor following pnpm's symlink graph

MarceloT-BC:
Ten lines in `.vscode/settings.json`. It addresses a real, measurable slowdown that arrived with the pnpm move and presents as "VS Code got slow" rather than as an error — so it has gone unattributed.

## The problem

VS Code's search defaults to **`search.followSymlinks: true`**. Against npm's hoisted `node_modules` that was harmless. Against pnpm's store it is not: every package gets its own `node_modules` full of symlinks into `.pnpm`, so the tree is a **cyclic graph**, and a following walk revisits the same real files through unboundedly many paths.

Measured on a clean clone of `next` (`pnpm install`, nothing else), using the ripgrep binary VS Code ships:

| | paths emitted | time |
| --- | --- | --- |
| `rg --files` | 17,516 | **0 s** |
| `rg --files --follow` | **49,219,818** | **60 s and still running** (19 GB of output; a first attempt was killed at 10 min) |

The repo has ~17.5k real files. 49.2M paths in a minute is not a slow walk, it is an unbounded one.

Notably **none** of those paths were inside `.pnpm` itself — the walk goes *through* the per-package symlink farm, where package A's `node_modules` links to B whose links reach back to A.

**This is also why the existing `node_modules` exclusion does not save you.** Exclusion globs match the path *as traversed*, so once a symlink is followed the files are reached under a path that no longer looks like a dependency.

Tree shape on this commit: **3,226** `.pnpm` store dirs, **11,782** symlinks at depth 4, **131** per-package `node_modules`.

## Not just VS Code

Spotlight hits the same structure independently. On a machine with **no editor running**, `fseventsd` was pegged at **137% CPU** and `mds_stores` at 71%; both dropped away once the stores were excluded from indexing (a `.metadata_never_index` file inside `node_modules` does this per-directory, no privileges needed). That is outside the scope of this PR, but worth knowing if your machine feels hot with nothing open — and it is corroborating evidence that the structure, not the editor, is the cause.

## Verified after the change

Same clone, same machine: VS Code idles at **0.3–1.6%** total CPU across all processes, **zero** ripgrep spawns, project-wide search settles quickly with correctly scoped results, and the file picker is instant.

**Go-to-definition is unaffected.** That is the TypeScript language server resolving `types` → `dist/*.d.ts` — a different subsystem that never consulted this setting. (If ⌘-click does not work for you in a fresh clone, that is because nothing has been built yet, not because of this change.)

## The one real trade-off — and a question for the Open App local-dev work

Disabling symlink following means **text search will not descend into symlinked directories**. In this repo that is exactly the intent: the only symlinks are inside `node_modules`. Everything you would search is a real directory.

**But that assumption is topology-dependent, and it is worth checking against `mj dev workspace`.** `guides/OPEN_APP_WORKSPACE_LINKING_SPEC.md` generates a workspace at the common parent of sibling repos — if members are reached through symlinks under that scheme, then with this setting a developer's project-wide search would silently skip the linked app's source. That is a bad failure mode: no error, just missing results.

So, for whoever owns that work:

1. Under the generated workspace, are member repos reached via **symlink** or added as **real folder roots**?
2. If symlinked, should the generated workspace ship `search.followSymlinks: true` *plus* an explicit `.pnpm` exclude, rather than inheriting this repo-level `false`?
3. Is the store still at the common parent (one `.pnpm` for N repos), or one per member? That changes the size of the graph a following walk would face.

Happy to test any of these against a real generated workspace once the tool exists — this was found while bringing open apps onto MJ 6.x, so we are already exercising that path.

## Scope

Deliberately minimal: no build, lint, or CI changes, and `.vscode/settings.json` is a file this repo already maintains. Anyone who prefers the old behaviour can override it in their user settings.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -34,5 +34,14 @@
   "typescript.tsdk": "node_modules/typescript/lib",
   "claudeCode.allowDangerouslySkipPermissions": true,
   "claudeCode.initialPermissionMode": "bypassPermissions",
-  "npm.packageManager": "pnpm"
+  "npm.packageManager": "pnpm",
+  "search.followSymlinks": false,
+  "search.exclude": {
+    "**/node_modules": true,
+    "**/.pnpm": true
+  },
+  "files.watcherExclude": {
+    "**/node_modules/**": true,
+    "**/.pnpm/**": true
+  }
 }


MarceloT-BC:
@cadam11 Hi, Craig, this one is interesting. I was seeing real usage spikes on my Mac after setting up PNPM, and Claude suggested this fix. I am not sure how this would integrate with your workspace design, so I wanted to ask you for review. I have tested this on my machine and it does significantly help performance.

cadam11 (approved):
LGTM– I barely use VSCode any more, so not surprised this got by. Nice catch

MarceloT-BC:
⚠️ **Converting this to draft — my evidence is wrong and I do not want anyone reviewing it on a false premise.**

## What I got wrong

The headline measurement in the description used `--no-ignore-vcs`, **which VS Code does not pass**. `search.useIgnoreFiles` defaults to `true`, so VS Code respects `.gitignore` — and `node_modules` is gitignored, so the search never enters the symlink farm at all. `--follow` never gets the chance to explode.

Re-measured with the flags VS Code actually uses, same binary, same clone:

| | files | time |
| --- | --- | --- |
| `rg --files --follow --no-ignore-vcs` (what I measured) | 49,219,818 | 60 s, unbounded |
| `rg --files --follow` (**what VS Code does**) | **17,500** | **0 s** |

So the 49.2M-path figure is real but irrelevant to the default configuration. I constructed a test that confirmed my hypothesis instead of one that could falsify it.

## Corroborating real-world evidence, also against me

A colleague ran a project-wide search in a pnpm instance workspace with **none** of these settings applied. It completed cleanly — zero ripgrep processes left behind, editor at 0.0% CPU. An unfixed workspace behaved fine.

## What still stands

Only this, and it is a separate mechanism from the search path:

- On a machine with **no editor running**, `fseventsd` was pegged at **137% CPU** and `mds_stores` at 71%. Both dropped away after excluding the pnpm stores from Spotlight indexing (`touch node_modules/.metadata_never_index`, per-directory, no privileges). That was measured directly and does not depend on anything in this PR.
- The structural facts are unchanged and independently verified: 3,226 `.pnpm` store dirs, 11,782 symlinks at depth 4, 131 per-package `node_modules`, reproduced on a raw clone with no tooling involved.

What I have **not** established is that any of that causes the VS Code slowdown, or that `search.followSymlinks: false` fixes it. The observation that started this — a dozen ripgrep processes at 45–75% CPU for 17 minutes — is real and unexplained; I attributed it to symlink-following without proving that link.

## What happens next

Leaving this as a draft rather than closing it, because the underlying question is still live and the tree-shape numbers may be useful to whoever picks it up. I will either come back with a measurement that actually isolates the cause, or close it. Please do not merge on the current evidence.

Apologies for the noise — this should have been falsification-tested before it was filed.

PR #3715, opened 2026-08-10
fix(realtime): carry an authored voice to any provider, not just OpenAI

### LINKED ISSUE #3530: Realtime voice cannot be carried to a non-OpenAI provider: override builders hardcode `openai`, and the default-model path never reveals the chosen vendor
## Summary

There is no supported way for a host application to carry an authored voice to whichever realtime provider actually runs a session. Both halves of the path assume OpenAI: the override builders hardcode the `openai` provider key, and when the framework picks the default realtime model it never tells the caller which vendor it chose.

Found while wiring per-persona voices in Caliber (an Open App). Caliber has to build the `realtime.voice` envelope by hand today, and the workaround is only possible because it happens to resolve the vendor itself.

## 1. The override builders hardcode `openai`

`BuildRealtimeOverridesJson` (`packages/AI/Agents/src/realtime/realtime-coagent-config.ts`):

```ts
if (v.length > 0) {
    realtime.voice = { providers: { openai: { voice: v } } };
}
```

Its own doc says so plainly:

> `openai` is the realtime provider today; add providers here when others ship realtime voices.

The same function is duplicated client-side as `BuildRealtimeConfigOverridesJson` (`packages/Angular/Generic/conversations/src/lib/services/realtime-pairing.ts`), with the same hardcoded key and the same comment. So a host that resolves to Inworld, ElevenLabs or anything else gets a voice bag filed under a provider that will never read it — silently. The setting looks configured and does nothing.

This is not reachable by extension: both are plain exported functions, so there is nothing to subclass or override.

**Suggested fix** — take the provider key as a parameter, defaulting to today's behaviour:

```ts
export function BuildRealtimeOverridesJson(
    modelId?: string | null,
    voice?: string | null,
    driverClass?: string | null,   // NEW — defaults to 'openai'
): string | null
```

`GetProviderVoiceSettings` already matches keys by case-insensitive PREFIX against the resolved `DriverClass`, longest key winning, so passing the `DriverClass` verbatim is an exact match by construction and strictly more precise than `openai`. No lookup table needed on either side.

Worth considering at the same time: the voice VALUE key is provider-specific too (ElevenLabs reads `voiceId`, OpenAI reads `voice`). The bag is documented as "an OPAQUE private pact with that driver", so hosts currently have to encode that themselves. A driver-side accessor (e.g. `BaseRealtimeModel.voiceSettingKey`) would put it where the knowledge lives.

## 2. The default model gives callers no way to know the provider

When a caller supplies no `realtime.modelPreference` — or none of its entries resolve — the framework picks the default realtime model itself (highest `PowerRank` active Realtime model), inside `PrepareClientSession`.

That is the right behaviour, but it means the caller cannot key ANY provider-specific setting for that session, because it does not know and cannot ask which vendor was chosen. Provider-keyed voice is simply unavailable on the default path.

Caliber's current workaround is to resolve the vendor itself via `selectRealtimeVendor` and only emit a voice when a model preference resolved; on the framework-default path it logs a warning and falls back to the driver's own voice. That is a workaround, not a fix — and it only works because Caliber subclasses `RealtimeClientSessionService` and can reach the protected method.

**Suggested fix** — either surface the resolved vendor before prepare (a public `ResolveRealtimeVendor(modelId?)`, or returning it from a dry-run resolve), or resolve `realtime.voice` provider-side after the model is chosen, so a host can supply a provider-agnostic voice and let the framework file it correctly.

The second is nicer: it would let a host say "this persona sounds like X" without knowing anything about vendors at all, which is what an app-level author actually wants.

## Why it matters beyond one app

The failure mode is silent in both halves. Nothing errors, nothing logs, and the authored value is visible in the config — it simply never reaches the driver. In Caliber this went unnoticed long enough that the persona metadata carried a comment saying the voice field was read-only, i.e. the gap had been reinterpreted as an intended limitation.

## Environment

MJ 5.51.0 · verified against `packages/AI/Agents/src/realtime/` and `packages/Angular/Generic/conversations/src/lib/services/realtime-pairing.ts` on `next`.

SDesai-BC:
An authored realtime voice has only ever reached OpenAI. Every other vendor — ElevenLabs, Gemini, Inworld, AssemblyAI, HuggingFace — silently got nothing: the setting stayed visible in the effective config and never reached a driver. Found while wiring per-persona voices in Caliber, which can only work around it because it happens to resolve the vendor itself.

Closes #3530.

## The bug

Both override builders filed the voice under a hardcoded provider key — `BuildRealtimeOverridesJson` (`realtime-coagent-config.ts`) and its client-side twin `BuildRealtimeConfigOverridesJson` (`realtime-pairing.ts`):

```ts
realtime.voice = { providers: { openai: { voice: v } } };
```

`GetProviderVoiceSettings` matches those keys by normalized prefix against the resolved vendor's `DriverClass`, so `openai` matches `OpenAIRealtime` and nothing else. Neither builder is reachable by extension — both are plain exported functions.

The second half is worse than a wrong default. On the default-model path the framework picks the realtime model (and therefore the vendor) itself inside `PrepareClientSession` and never tells the caller which. A provider-keyed voice there is not merely wrong, it is **unauthorable** — the caller has no vendor to key it to, and no way to ask.

## What this does

The framework already resolved the voice *after* the model was chosen, handing the resolved `DriverClass` to `GetProviderVoiceSettings`. Nothing was missing from the resolution; the gap was purely in the authoring interface, which demanded vendor knowledge the caller does not have. So this adds a provider-agnostic slot rather than a `driverClass` parameter — the builders' signatures do not grow, and the caller's interface shrinks from four facts to one.

- **`realtime.voice.default.voice`** — authored without naming a vendor, filed onto whichever driver resolves. Both builders now emit this shape.
- **Precedence is per-key.** The agnostic value wins the `voice` key; a matching `providers.<key>` bag still contributes all of its other opaque settings. So a runtime pick beats a vendor-pinned value authored in agent metadata. Scope worth being precise about: only `OpenAIRealtime`, `xAIRealtime` and `HuggingFaceRealtime` declare `SupportedVoices` (HuggingFace returns none) and the native picker's dropdown is gated on that list, so for the vendors this unblocks the voice arrives via **agent metadata and programmatic/bridge hosts**, not the picker yet. With no agnostic voice authored, behaviour is byte-for-byte what it was.
- **Kills the silence.** The mint log now carries the resolved `driver` and the `voice` that actually reached it, and both realtime surfaces log when authored provider bags matched nothing. That needed a new `MatchProviderVoiceSettings`: `GetProviderVoiceSettings` is truthy for every driver once an agnostic voice is set, so it cannot answer "did a provider key actually match?" — and using it would have silenced the warning on exactly the path that emits an agnostic voice.
- **Discloses the resolved `DriverClass`** on `RealtimeClientSessionPrepResult`, for observability on the default-model path.
- **Schema** — adds the agnostic slot and marks `providers.*.voiceId` deprecated.

## Why it works without a per-driver accessor

The issue suggested a `driverClass` parameter plus a `BaseRealtimeModel.voiceSettingKey` accessor, because ElevenLabs appears to want `voiceId`. Neither is needed, and this PR adds neither: **every realtime driver already reads the same driver-neutral `voice` key out of the config bag**, by deliberate design. Verified by reading each one:

| Driver | Reads | Maps to |
|---|---|---|
| `OpenAIRealtime` | `voice` | `audio.output.voice` |
| `xAIRealtime` | inherits `OpenAIRealtime` | same |
| `ElevenLabsRealtime` | `config['voice']` via `ResolveVoiceID` | `tts.voice_id` on the wire |
| `InworldRealtime` | `config['voice']` | `output.voice` |
| `AssemblyAIRealtime` | `config['voice']` | `output.voice` |
| `HuggingFaceRealtime` | `config['voice']` | `audio.output.voice` |
| `GeminiRealtime` | `voice` exempted from the shared-key scrub | ⚠️ **reaches the driver, then dropped before the wire** — see below |

**Correction, caught in review:** the Gemini row originally claimed the key was "merged into `LiveConnectConfig`". It is not. `buildConnectConfig` assigns the bag onto a `LiveConnectConfig` that has **no `voice` property** — Gemini Live wants `speechConfig.voiceConfig.prebuiltVoiceConfig.voiceName`, and nothing in the Gemini package maps it (no `voiceName`, no `prebuiltVoiceConfig`, no `voiceConfig` anywhere). So an authored voice reaches `GeminiRealtime` and is silently discarded — the same bug class this PR fixes, one layer down. Tracked as **#3721** and now fixed in **#3725** — kept out of this PR because it is a driver-level mapping with its own test surface. That PR also confirmed against the live API what this one could only infer: an invalid voice name passes Gemini's token mint and then kills the session at socket open (close `1007`). The cross-driver loop in `realtime-convergence-drift.test.ts` stays, since what it actually asserts — `GetProviderVoiceSettings` resolution — is correct for all four drivers.

`providers.*.voiceId` is deprecated rather than deleted for the same reason: no driver has ever read it, so authoring it did nothing — the identical silent-drop bug class. It stays in the schema because `additionalProperties: false` would fail validation for configs that already carry it.

## Known limitation, documented in code

`GetProviderVoiceSettings` runs *after* the cascade has been deep-merged, so it has no layer provenance: "agnostic wins" is absolute rather than "the higher layer wins". That is correct for every shape emitted today, since every host builder emits the agnostic form and an agnostic value is therefore always the session-scoped pick. It would be wrong for a config that hand-authors `providers.<key>.voice` in a *runtime override* while agent metadata sets `default.voice` — nothing emits that. Fixing it properly means resolving voice precedence per-layer inside `ResolveEffectiveRealtimeConfig`, where layer identity still exists. Noted on the function rather than left for someone to rediscover.

## Folded-in refactor

The vendor-selection walk (highest-`Priority` Active vendor whose `DriverClass` resolves an API key) existed as three byte-identical copies: `BaseAgent.selectRealtimeVendor`, `RealtimeClientSessionService.selectRealtimeVendor`, and the module-private `resolveRealtimeDriverClass` in `bridge-realtime-session-factory`. Now one `SelectRealtimeVendorForModel`, with an injectable key resolver so the service keeps its existing test seam. No behaviour change.

It is not a separate commit, which is the repo's normal preference — two of those three files also carry the behaviour change above, and separating them needs interactive staging.

## ⚠️ Deploy order — the schema needs `mj sync push`

`metadata/agent-types/schemas/realtime-type-config.schema.json` is `@file:`-referenced as `ConfigSchema` on the realtime agent type in `.agent-types.json`, so **merging this does not update any database**. Until `mj sync push` runs in an environment, that environment's `ConfigSchema` is the old one.

This is not cosmetic. `MJAIAgentEntityServer` validates `TypeConfiguration` against the type's `ConfigSchema`, and `json-schema-lite` enforces `additionalProperties: false` as a hard **Failure** on save. The `voice.default` object carries that keyword, so authoring `realtime.voice.default.voice` in an agent's `TypeConfiguration` against an un-pushed schema **fails the agent save** with a schema violation.

Scope: agent-metadata authoring only. The picker's runtime override travels through `ResolveEffectiveRealtimeConfig`, not the agent-save validator, so it is unaffected.

### Two more compatibility windows, both raised in review

**Deploy the client and server together.** A new `@memberjunction/ng-conversations` against an **older MJAPI silently drops a picked voice for every provider, including OpenAI** — the old server's `normalizeVoice` builds the persona from `tone`/`speakingStyle` only, so an incoming `default.voice` falls on the floor. I diffed it against `origin/next` to confirm. That regresses currently-working behaviour, so a matching version bump is not sufficient if the two deploy independently. The reverse (old client, new server) is safe: the `providers.openai` shape is still honoured.

**Bridge hosts that supply a voice without pinning a model.** `BridgeRealtimeSessionContext.RealtimeVoice` is host-supplied (LiveKit/Twilio/Teams — no in-repo setter) and independent of the model pin, so it lands squarely on the default-model path. A host passing `echo` with no model previously had it filed under `openai` — inert unless OpenAI resolved. It now travels agnostically to whichever vendor wins the priority walk, and on ElevenLabs becomes `tts.voice_id: 'echo'`; `ResolveVoiceID` guards *empty* but not *invalid*, and ElevenLabs rejects a bad voice id hard enough to end the session. An agnostic voice is only as portable as the id itself, which is now stated on the authoring field. The picker path has no such exposure — it gates the dropdown on `SelectedModelVoices.length` and clears the selection on model change.

## Merged `next`, and the one interface change that came out of it

`next` has been merged in (the branch was 291 commits behind). Two overlapping changes met in the same realtime paths, and resolving them **grew one interface**, which is the part worth a reviewer's attention rather than a merge commit to skim:

`next` commit `9a3ddbe71a` (ModelConfiguration catalog cascade) added a `modelVendorID` to vendor selection — the `MJ: AI Model Vendors` **row** id, which keys the most-specific layer of the `AIModelType < AIModel < AIModelVendor` cascade that now supplies the session config bag's base layer. This branch had just extracted that same vendor walk out of three byte-identical copies into one `SelectRealtimeVendorForModel`. Rather than reinstate a private copy, the extracted function learned the new field: **`RealtimeVendorSelection` gains `ModelVendorID`**. The dedupe holds, and the catalog cascade reads its row id through the single selector both realtime surfaces now share.

Per hunk: `base-agent.ts` keeps the shared selector, drops the now-dead private `selectRealtimeVendor`, and merges both config-bag comments — catalog defaults are the base layer, the agnostic voice still wins the `voice` key. `realtime-client-session-service.ts` keeps the delegation, and keeps `configBag` hoisted (the mint log reads the voice off it) while threading the new `modelID`/`modelVendorID` into `buildSessionConfigBag`.

The merge also exposed a hole in this PR's own tests: the shape assertions used `toEqual` against fixtures with no `ID`, and `toEqual` ignores undefined-valued keys — so nothing actually pinned `ModelVendorID`. The fixture now carries a row id distinct from `VendorID`, plus a test asserting the two cannot collide, because confusing them would silently read the wrong cascade layer.

## Evidence

Re-run on the merged branch:

| Check | Result |
|---|---|
| `packages/AI/Agents` — `pnpm test` | **1862 passed**, 99 files |
| `packages/Angular/Generic/conversations` — `pnpm test` | **1079 passed**, 92 files |
| Full workspace build (`pnpm run build`) | **271/271** packages |
| Deterministic integration tier | **74/78**, `IT — realtime-deterministic` **passed** (see caveat) |

New tests: 6 cases on the agnostic slot (reaches a driver with no provider entry, wins the `voice` key, leaves other provider keys intact, does not mask an unmatched bag, changes nothing when unset, survives normalization); 9 on `SelectRealtimeVendorForModel` including the row-id-vs-vendor-id distinction; 2 drift-protection cases pinning the envelope shape and the non-OpenAI regression across four drivers; 1 client-side case pinning the agnostic envelope.

**Caveat on that last row, raised in review:** `realtime-deterministic.checks.ts` has **zero voice coverage**, so the bundle passing is true but attests nothing about this change. The real proof for this PR is the unit layer — the resolution, precedence and diagnosis tests listed above.

### On the 4 integration failures

None are attributable to this branch, and none are in realtime. `agent-loop-live` (AL2, max iteration limit), `entity-writes` (EW8), `subscription-isolation` (SI1, no acting user) and `workflow-demo-agents` (WD1/WD2) **all fail on `next` itself** — CI run `31405401357` at `2b75ccfcf`, the exact commit merged here, scores **58/78**. Outside the realtime files the only file this PR shares with those paths is `base-agent.ts`, and all three of its hunks sit inside realtime-only methods (`resolveRealtimeModel`, the effective-realtime-config region, `buildRealtimeSessionParams`).

The tier was run against a database built from scratch for this verification — `mj migrate` (43 applied) → `mj sync push` (`metadata/`, then `metadata-optional/integration-test` for the seeded `IT:` fixtures) → `mj codegen` (374 entities, CRUD validation passed) → full rebuild. Worth knowing for anyone reproducing it: the seeded agent fixtures live in `metadata-optional/`, and without that second push 8 further bundles fail with `seeded agent ... not found`.


rkihm-BC (commented):
Thorough PR — the diagnosis is right and the design call (agnostic slot rather than a `driverClass` parameter) is the correct one. I verified the premise it rests on: `voice` is listed in `REALTIME_SHARED_CONFIG_KEYS` in `baseRealtime.ts:197`, and I read all seven drivers — OpenAI (`rest.voice` → `audio.output.voice`), xAI (inherits), ElevenLabs (`ResolveVoiceID` → `tts.voice_id`), Inworld/AssemblyAI (`output.voice`), HuggingFace (`audio.output.voice`). It really is a driver-neutral key by design, so the issue's suggested `BaseRealtimeModel.voiceSettingKey` accessor genuinely isn't needed.

Also confirmed: the `voiceId` deprecation is correctly justified (the only `voiceId` in ElevenLabs is the managed-agent override *enablement* flag at `elevenLabsRealtime.ts:675`, never a config-bag read); the `GetProviderVoiceSettings` / `MatchProviderVoiceSettings` split avoids a real trap; adding `voice` to `RealtimeVoicePersona` is safe because `BuildVoiceMannerSection` enumerates keys explicitly rather than iterating, so no voice id leaks into the system prompt; and no seeded metadata authors a voice anywhere, so there's nothing to migrate. The three-copy dedupe is clean and the `ModelVendorID`-vs-`VendorID` catch (that `toEqual` ignores undefined-valued keys) is a good self-audit.

I ran the touched tests: `realtime-vendor-resolution` + `realtime-coagent-config` + `realtime-convergence-drift` → 93 passed; `realtime-pairing` → 21 passed.

Four things I'd want addressed, inline below. Three are small; the first is a factual correction to the PR description.

**1. The Gemini row of the justification table is wrong** — see the comment on the drift test. The voice reaches the driver and is then dropped on the wire.

**2. `DriverClass` on the prep result is never read by anyone** — see the comment on the interface.

**3. Cross-vendor voice leakage on the bridge path** — see the comment on `BuildRealtimeOverridesJson`.

**4. `guides/REALTIME_CO_AGENTS_GUIDE.md` is now stale and isn't in this PR.** Line 660 lists this exact bug as an open TODO: *"the native voice picker (`BuildRealtimeConfigOverridesJson` / `BuildRealtimeOverridesJson`) still hardcodes the `openai` provider key, so a runtime voice pick cannot target any other provider."* That's now false. The same cell also claims `OpenAIRealtime.CreateClientSession` doesn't fold `params.Config` — also false (it does, via `ExtractRealtimeFeatures` → `BuildAudioBlock`), so that cell was already drifting before this PR. Lines 293 and 303 document the config shape without the agnostic slot. Per CLAUDE.md's routing this is guide territory, and the guide is what people read before building here.

One note on the evidence table: `packages/TestingFramework/integration-test-suite/src/checks/realtime-deterministic.checks.ts` has **zero** voice coverage, so `IT — realtime-deterministic` passing is true but attests nothing about this change. Not a defect — the table just reads stronger than it is.

<sub>Scope of what I checked: all 7 drivers' voice handling; every caller of the changed functions; `normalizeVoice` diffed against `next`; resolver plumbing; seeded metadata; the integration checks. I did **not** run full package suites or a build, did not run the integration tier, and cannot verify Caliber or the out-of-repo bridge hosts that supply `RealtimeVoice`.</sub>

packages/AI/Agents/src/__tests__/realtime-convergence-drift.test.ts
```diff
@@ -110,18 +110,33 @@ describe('Realtime convergence — drift protection', () => {
...
             });
             expect(JSON.parse(BuildRealtimeOverridesJson('m', null)!)).toEqual({
                 realtime: { modelPreference: 'm' },
             });
         });
 
+        it('carries a picked voice to a NON-OpenAI driver (the #3530 regression)', () => {
+            // Pre-fix this filed the voice under `openai` and any other vendor silently got nothing.
+            const eff = ResolveEffectiveRealtimeConfig(null, null, BuildRealtimeOverridesJson(null, 'Rachel'));
+            for (const driver of ['ElevenLabsRealtime', 'GeminiRealtime', 'InworldRealtime', 'OpenAIRealtime']) {
```
rkihm-BC, line 130: **The Gemini row of the PR's justification table is wrong, and this test looks like it proves otherwise.**

The table says Gemini reads `voice` ("explicitly exempted from the shared-key scrub") and that it's "merged into `LiveConnectConfig`". The exemption is real, but `buildConnectConfig` then does `Object.assign(config, cfg)` onto a `LiveConnectConfig` that **has no `voice` property** — in @google/genai 1.52.0 it has `speechConfig?: SpeechConfig`, and Gemini Live wants `speechConfig.voiceConfig.prebuiltVoiceConfig.voiceName`.

I grepped the whole Gemini package: there is no `voiceName`, no `prebuiltVoiceConfig`, no `voiceConfig` anywhere. The only `speechConfig` reference is a passthrough of a caller-supplied object at `geminiRealtime.ts:298-299`. So a Gemini voice can only be set by hand-authoring a raw `speechConfig` into the config bag — never via the neutral `voice` key. An authored voice reaches `GeminiRealtime` and is silently discarded: the identical bug class this PR exists to kill, one layer down.

What makes it worth fixing before merge rather than after: this loop *reads* as end-to-end coverage of Gemini, but it only asserts `GetProviderVoiceSettings` resolution and never touches a driver. And `geminiRealtime.test.ts` has zero voice assertions, so nothing anywhere would catch it.

Suggest: drop Gemini from the table (or mark it *"reaches the driver; not yet mapped to `speechConfig`"*), and file a follow-up for the `voice` → `speechConfig.voiceConfig.prebuiltVoiceConfig.voiceName` mapping. The loop itself is fine to keep — it's pinning resolution, which is genuinely correct for all four.

packages/AI/Agents/src/realtime/realtime-client-session-service.ts
```diff
@@ -217,6 +219,18 @@ export interface RealtimeClientSessionPrepResult {
...
+     * `ElevenLabsRealtime`). Present on success.
+     *
+     * Disclosed for OBSERVABILITY — on the default-model path the framework picks the vendor itself, so
+     * without this a caller cannot tell which one spoke, and cannot diagnose a voice that did not land
+     * (issue #3530). It is deliberately NOT a pre-prepare resolution API: a caller does not need to know
+     * the vendor in advance, because {@link RealtimeVoicePersona.voice} carries a voice to whichever
+     * vendor is chosen. Re-deriving the vendor in order to pre-file provider-keyed settings duplicates a
+     * decision the framework owns.
+     */
+    DriverClass?: string;
```
rkihm-BC, line 237: **This field is never read by anyone — it's dead as shipped.**

`RealtimeClientSessionResolver.ts` contains no `DriverClass` at all: it surfaces `ModelID`/`ModelName` (field decl ~line 228, assignment line 1679) and stops there. `BaseAgent` doesn't read it either. I grepped the repo — nothing consumes it.

That matters because of what the docstring promises: *"on the default-model path the framework picks the vendor itself, so without this a caller cannot tell which one spoke, and cannot diagnose a voice that did not land."* On the client-direct path the caller **is the browser**, via GraphQL — and it still can't tell. The disclosure stops at the service boundary.

Either plumb it through `StartRealtimeClientSessionResult` (a field near line 228 + one line at 1679, mirroring `ModelName`), or trim the docstring so it doesn't claim an observability win that isn't delivered. Plumbing it seems clearly right given the stated motivation — it's two lines.

packages/AI/Agents/src/realtime/realtime-coagent-config.ts
```diff
@@ -1031,12 +1108,12 @@ export function BuildRealtimeOverridesJson(
...
         return null;
     }
-    const realtime: { modelPreference?: string; voice?: { providers: Record<string, { voice: string }> } } = {};
+    const realtime: { modelPreference?: string; voice?: { default: { voice: string } } } = {};
     if (m.length > 0) {
         realtime.modelPreference = m;
     }
     if (v.length > 0) {
-        realtime.voice = { providers: { openai: { voice: v } } };
+        realtime.voice = { default: { voice: v } };
```
rkihm-BC, line 1135: **Behavior-change risk on the bridge path: a previously-inert misconfiguration becomes a session-killer.**

The picker path is safe — `realtime-agent-picker.component.ts` gates the voice dropdown on `@if (SelectedModelVoices.length)` and `OnModelChange` clears `SelectedVoiceId`, so a picked voice can't outlive a vendor switch. Nice.

The bridge path has no equivalent coupling. `BridgeRealtimeSessionContext.RealtimeVoice` (`bridge-realtime-session-factory.ts:56`) is an optional host-supplied string with **no in-repo setter** — LiveKit/Twilio/Teams adapters supply it. It flows to `data.realtimeVoice` → `buildBridgePrepInput` (`base-agent.ts:1953`) → here, where `modelID` and `voice` are **independent** arguments. A host can supply a voice with no model pin.

That lands on exactly the default-model path this PR is about. A host passing `'echo'` with no `realtimeModelID` previously had it filed under `openai` — inert unless OpenAI resolved. Now it goes agnostically to whatever wins the priority walk. On ElevenLabs that becomes `tts.voice_id: 'echo'`, an invalid id; `ResolveVoiceID` guards *empty* but not *invalid*, and its own comment notes ElevenLabs rejects a bad `voice_id` hard enough to kill the session.

Not arguing against the change — but it deserves a compatibility note in the PR body / changeset, and possibly a line in the `WarnOnUnmatchedProviderVoice` neighborhood pointing out that an agnostic voice is only as portable as the id itself.

packages/AI/Agents/src/realtime/realtime-coagent-config.ts
```diff
@@ -48,12 +48,35 @@ export const REALTIME_ADVANCED_SESSION_CONTROLS_AUTHORIZATION = 'Realtime: Advan
...
+     * `audio.output.voice`, OpenAI/Gemini consume it directly).
+     *
+     * Precedence: this WINS the `voice` key over a matching {@link RealtimeVoiceConfig.providers} entry —
+     * a session-scoped pick must beat a vendor-pinned value authored lower in the cascade. The matching
+     * provider bag still contributes all of its OTHER keys.
+     *
+     * The value is still provider-native (an `alloy` means nothing to ElevenLabs), so a config that must
+     * name different voices per vendor should pin them under `providers.<key>` and leave this unset.
+     */
+    voice?: string;
```
rkihm-BC, line 82: Two things on this field.

**1. The canonical-shape example at the top of this same file wasn't updated.** Lines 21-31 — *"The canonical JSON shape (the Realtime agent type's `ConfigSchema` is seeded to this)"* — still shows `"default": { "tone": …, "speakingStyle": … }` with the voice only under `providers`. Every other JSDoc in this file was carefully revised; this one is the first thing a reader of the module sees, and it's now the only surviving hardcoded `providers: { openai: … }` example in non-test code.

**2. Same-layer specificity inversion.** The documented limitation is cross-*layer*. There's a same-layer case too: agent metadata authoring both `default.voice: 'Rachel'` (fallback) and `providers.openai.voice: 'alloy'` (pin) — the general beats the specific, which inverts normal config intuition, and the schema permits exactly that shape.

Your last paragraph here does disclose it (*"a config that must name different voices per vendor should pin them under `providers.<key>` and leave this unset"*), so it isn't hidden — but "agnostic always wins" is a strong default for a same-layer conflict. Worth at least considering whether *"agnostic is a fallback unless it came from a runtime override"* is more predictable, since the runtime-override case is the only one that actually needs to win.

packages/AI/Agents/src/realtime/realtime-client-session-service.ts
```diff
@@ -381,6 +395,43 @@ export interface RealtimeModelResolutionOutcome {
...
+ * setting is exactly what nobody noticed for long enough to file the issue.
+ *
+ * Shared by BOTH realtime surfaces — the client-direct prepare and `BaseAgent`'s server-bridged
+ * session — so neither path can drift back into silence.
+ *
+ * @param effectiveConfig The resolved effective configuration.
+ * @param driverClass The resolved vendor's `DriverClass`.
+ * @param surface The calling surface, for log attribution (e.g. `RealtimeClientSessionService`).
+ */
+export function WarnOnUnmatchedProviderVoice(
```
rkihm-BC, line 428: **No test covers this function.** It has two short-circuit branches (empty-authored, matched) and a specific documented rationale for calling `MatchProviderVoiceSettings` rather than `GetProviderVoiceSettings`. Nothing pins that — so a future refactor back to `GetProviderVoiceSettings` would silently kill the warning on exactly the path that emits an agnostic voice, which is the trap the PR body says it avoided. Three lines would lock it.

Separate, minor: this is a pure config-inspection helper living in the session-service module, which now makes `base-agent.ts` import the service module just to get it. It'd naturally belong next to `MatchProviderVoiceSettings` — except `realtime-coagent-config.ts`'s header declares that module *"deliberately framework-free: no DB, no metadata provider, no logging imports"*, so it can't go there as written. That constraint is the real reason for the placement, and a sentence saying so would stop it reading as an accident. (Alternative: return a diagnostic string and let callers log.)

packages/Angular/Generic/conversations/src/lib/services/realtime-pairing.ts
```diff
@@ -218,13 +223,12 @@ export function BuildRealtimeConfigOverridesJson(
...
   }
-  const realtime: { modelPreference?: string; voice?: { providers: Record<string, { voice: string }> } } = {};
+  const realtime: { modelPreference?: string; voice?: { default: { voice: string } } } = {};
   if (modelId.length > 0) {
     realtime.modelPreference = modelId;
   }
   if (voice.length > 0) {
-    // Per-provider voice shape matches GetProviderVoiceSettings; `openai` is the realtime provider today.
-    realtime.voice = { providers: { openai: { voice } } };
+    realtime.voice = { default: { voice } };
```
rkihm-BC, line 231: **Client/server version skew is a regression window and isn't in the deploy-order section.**

I diffed `normalizeVoice` against `origin/next`: pre-PR it builds the persona from `tone`/`speakingStyle` only, so an incoming `default.voice` is dropped on the floor. So a new `@memberjunction/ng-conversations` against an older MJAPI regresses the *currently working* case — a picked voice is silently lost **including for OpenAI**, which is strictly worse than today. Same version bump doesn't mean same deploy.

The reverse direction (old client, new server) is fine — the old `providers.openai` shape is still honored.

Worth one line next to the `mj sync push` note.

---

Separately, a scope note worth adding somewhere: only `OpenAIRealtime`, `xAIRealtime` and `HuggingFaceRealtime` declare `SupportedVoices` (and HF overrides it to return none). ElevenLabs, Gemini, Inworld and AssemblyAI all inherit `[]` from the base, and the picker's dropdown is gated on `SelectedModelVoices.length`. So for the exact vendors this PR unblocks, the picker still offers no voice at all — the fix is reachable only via agent metadata and programmatic hosts like Caliber. That's genuinely valuable and matches where the bug was found, but the PR body's *"which is what the common path needs"* framing overstates picker reach today.

metadata/agent-types/schemas/realtime-type-config.schema.json
```diff
@@ -22,7 +22,8 @@
               "additionalProperties": false,
               "properties": {
                 "tone": { "type": "string", "description": "Spoken tone folded into the session system prompt (e.g. 'warm, unhurried')." },
-                "speakingStyle": { "type": "string", "description": "Speaking style folded into the session system prompt (e.g. 'concise, plain language')." }
+                "speakingStyle": { "type": "string", "description": "Speaking style folded into the session system prompt (e.g. 'concise, plain language')." },
+                "voice": { "type": "string", "description": "Provider-AGNOSTIC voice id, used whichever vendor ends up running the session (every realtime driver reads the neutral 'voice' key). Set this to author a voice without knowing the vendor. Wins over a 'providers.<key>.voice' entry; the matching provider bag still contributes its other settings." }
```
rkihm-BC, line 26: Good that the precedence is stated right in the description — that's where an author will actually read it.

One asymmetry: the reciprocal note is missing on `providers.*.voice` (line 36), which is the field being silently outranked. Someone authoring a per-vendor pin reads *that* description, not this one, and gets no hint that a sibling `default.voice` will beat it. Suggest adding "Overridden by `voice.default.voice` when one is authored" there.

metadata/agent-types/schemas/realtime-type-config.schema.json
```diff
@@ -33,7 +34,7 @@
                 "additionalProperties": false,
                 "properties": {
                   "voice": { "type": "string", "description": "Provider voice name/preset (e.g. 'alloy')." },
-                  "voiceId": { "type": "string", "description": "Provider voice identifier where voices are ID-addressed (e.g. ElevenLabs)." }
+                  "voiceId": { "type": "string", "description": "DEPRECATED — never read by any driver; authoring it does nothing. The whole realtime driver family (including ElevenLabs, which maps it to tts.voice_id on the wire) reads the neutral 'voice' key. Retained only so configurations that already carry it still validate; use 'voice' instead." }
```
rkihm-BC, line 37: The deprecation is correctly justified — I confirmed the only `voiceId` in ElevenLabs is the managed-agent override *enablement* flag (`overrides?.tts?.voiceId === true`, `elevenLabsRealtime.ts:675`), never a config-bag read. Keeping it rather than deleting, because `additionalProperties: false` would fail existing configs, is the right call.

While you're here: this provider bag is `additionalProperties: false` with only `voice`/`voiceId`, but the new JSDoc on `GetProviderVoiceSettings` calls it an "opaque driver pact" whose "other (opaque) settings still ride along", and the new test exercises `language` / `someOpaqueKnob`. Both are fine at runtime (overrides bypass this validator), but the shape is **unauthorable in agent `TypeConfiguration`**. Pre-existing mismatch, just amplified by the new wording — worth either loosening this to `additionalProperties: true` or softening the "opaque pact" language to note it applies to runtime overrides only.

.changeset/realtime-provider-agnostic-voice.md
```diff
@@ -0,0 +1,8 @@
...
+
+The agent-type `ConfigSchema` gains the new slot and marks `providers.*.voiceId` deprecated (no driver has ever read it). That schema is `@file:`-referenced metadata, so it reaches a database only via `mj sync push`; until then, authoring `realtime.voice.default.voice` in an agent's `TypeConfiguration` fails validation against the old schema, which enforces `additionalProperties: false`.
```
rkihm-BC: This deploy-order note is correct and genuinely non-obvious — good that it's here rather than only in the PR body. I verified the mechanism: `ConfigSchema` is `@file:`-referenced at `metadata/agent-types/.agent-types.json:159`, and `voice.default` carries `additionalProperties: false`, so the described save failure is real.

Two additions worth folding in:

1. **The client/server skew case** (detail on `realtime-pairing.ts`): a new `ng-conversations` against an older MJAPI silently drops the picked voice for *every* provider including OpenAI, because pre-PR `normalizeVoice` doesn't carry `default.voice` through. That's a regression of currently-working behavior, so it belongs next to the schema note.

2. **Bridge hosts supplying `RealtimeVoice` with no model pin** now forward that id to whatever vendor resolves (detail on `BuildRealtimeOverridesJson`).

Minor: this is typed `patch` for both packages, but the change adds exported API — `SelectRealtimeVendorForModel`, `RealtimeVendorSelection`, `MatchProviderVoiceSettings`, `WarnOnUnmatchedProviderVoice`, `RealtimeClientSessionPrepResult.DriverClass`. Additive, so `patch` isn't wrong, but `minor` would describe it more accurately.

packages/AI/Agents/src/__tests__/realtime-convergence-drift.test.ts
SDesai-BC: Confirmed and corrected — you're right, and the table was wrong. Grepped the Gemini package: no `voiceName`, no `prebuiltVoiceConfig`, no `voiceConfig`, and `buildConnectConfig` does `Object.assign(config, cfg)` onto a `LiveConnectConfig` with no `voice` property (line 361). The voice reaches the driver and dies before the wire.

Took your suggested split: the Gemini row now reads *"reaches the driver, then dropped before the wire"* with the mechanism spelled out, Gemini is dropped from the canonical shape examples in both the module JSDoc and the guide (documenting a mapping that doesn't exist is worse than omitting it), and the mapping is filed as #3721 with your `speechConfig.voiceConfig.prebuiltVoiceConfig.voiceName` target plus the note that `geminiRealtime.test.ts` needs a driver-level assertion. Kept the loop as you suggested — what it asserts is resolution, which is genuinely right for all four.

packages/AI/Agents/src/realtime/realtime-client-session-service.ts
SDesai-BC: Plumbed it. `StartRealtimeClientSessionResult.DriverClass` field decl mirroring `ModelName`, plus the one assignment line. You were right that the docstring was writing a cheque the code didn't cash — on the client-direct path the caller is the browser, so stopping at the service boundary delivered observability to nobody. Docstring now says where it lands and why `ModelName` alone isn't enough (one model resolves to different vendors depending on which keys a deployment configures).

packages/AI/Agents/src/realtime/realtime-coagent-config.ts
SDesai-BC: Good catch on the asymmetry between the two paths — and thanks for verifying the picker side is actually coupled; I hadn't checked `OnModelChange`.

Documented rather than code-gated, in both the changeset and the PR body: a host supplying `RealtimeVoice` with no model pin now forwards that id to whichever vendor wins the priority walk, an OpenAI-style `echo` becomes `tts.voice_id` on ElevenLabs, and `ResolveVoiceID` guards empty but not invalid. The authoring field now states that an agnostic voice is only as portable as the id itself. I didn't add a runtime guard because validity is vendor-specific and this layer can't tell a good id from a bad one — a warning here would either be wrong or unfalsifiable.

packages/AI/Agents/src/realtime/realtime-coagent-config.ts
SDesai-BC: Both fixed.

1. The canonical-shape JSDoc is updated — you're right that it was the module's first impression and the last surviving hardcoded `providers: { openai }` in non-test code. It now shows `default.voice` with an inline note on precedence and when to prefer per-vendor pins.

2. Same-layer inversion is now documented on `GetProviderVoiceSettings` alongside the cross-layer one. I did *not* switch to "agnostic is a fallback unless it came from a runtime override" — it's the more predictable rule, but it needs the same per-layer provenance the cross-layer limitation needs, so implementing half of it would swap one surprise for another. Noted explicitly as the reason, with the authoring rule (pin per-vendor OR author agnostic, never both) stated on the field and now in the schema description too.

packages/AI/Agents/src/realtime/realtime-client-session-service.ts
SDesai-BC: Added 5 tests, and I checked they actually bite rather than just pass: I temporarily swapped the call to `GetProviderVoiceSettings` — precisely the refactor-back you describe — and only the agnostic-voice-set case fails, which is the trap. Reverted, all 5 green.

On placement: recorded your reasoning in the docstring, since you identified the real constraint. It logs, `realtime-coagent-config.ts` declares itself framework-free, so it can't move there as written. I noted why I rejected the return-a-diagnostic-string alternative too — it lets one surface quietly opt out of logging, which is the failure mode being fixed.

packages/Angular/Generic/conversations/src/lib/services/realtime-pairing.ts
SDesai-BC: Verified the same way you did — `origin/next`'s `normalizeVoice` builds the persona from `tone`/`speakingStyle` only, so `default.voice` is dropped. New client + old server silently loses a picked voice **including for OpenAI**, which is strictly worse than today. That's now the first compatibility note in both the changeset and the PR body, ahead of the `mj sync push` item, since it regresses working behaviour rather than blocking a new one.

The `SupportedVoices` scope note is in as well — you're right that "which is what the common path needs" overstated picker reach. The PR body now says the voice arrives via agent metadata and programmatic/bridge hosts, not the picker, for exactly the vendors this unblocks; the guide's open-items table says the same.

metadata/agent-types/schemas/realtime-type-config.schema.json
SDesai-BC: Added. `providers.*.voice` now carries the reciprocal — "Overridden by 'voice.default.voice' when one is authored" plus the author-one-or-the-other rule. Your framing is the right one: the person authoring the pin reads *that* description, not the sibling's.

metadata/agent-types/schemas/realtime-type-config.schema.json
SDesai-BC: Took the wording route rather than loosening the schema. `additionalProperties: true` would relax validation for every existing config to fix a docs mismatch, which felt like the wrong trade for a pre-existing asymmetry.

The `GetProviderVoiceSettings` JSDoc now scopes "opaque pact" explicitly: it describes what the function and drivers do with the bag at runtime, not what's authorable in agent metadata — keys past `voice`/`voiceId` reach a driver only via runtime `ConfigOverridesJson`, which bypasses the validator. Flagged as pre-existing and called out that loosening it is a separate decision.

.changeset/realtime-provider-agnostic-voice.md
SDesai-BC: All three folded in. Bumped to `minor` — you're right that it describes the change better; I confirmed `SelectRealtimeVendorForModel`, `RealtimeVendorSelection`, `MatchProviderVoiceSettings`, `WarnOnUnmatchedProviderVoice` and the `DriverClass` fields are all absent from `next`, so the surface is genuinely new even though it's additive.

The changeset now leads with the client/server skew (it regresses working behaviour), then the `mj sync push` note, then the bridge-host case, and closes with the Gemini gap and the `SupportedVoices` picker limitation so nobody reads this as "voices now work everywhere." Thanks for verifying the `@file:` mechanism independently.

SDesai-BC:
Thanks @rkihm-BC — that was a genuinely useful review. I verified all eleven findings against the code before touching anything, and every one held up, including both places where the claim was that *my own description* was wrong. Pushed as `07443355d1`.

**Code (2)**
- `DriverClass` now reaches the browser on `StartRealtimeClientSessionResult` — it really was dead as shipped, and the docstring was promising observability the code didn't deliver.
- `WarnOnUnmatchedProviderVoice` has 5 tests, and I verified they bite: temporarily swapping in `GetProviderVoiceSettings` fails exactly the agnostic-voice case and nothing else. Placement rationale recorded in the docstring.

**Corrections to my own claims (2)**
- **Gemini was wrong in the table.** The voice reaches the driver and is dropped before the wire — `LiveConnectConfig` has no `voice` property and nothing maps `speechConfig.voiceConfig.prebuiltVoiceConfig.voiceName`. Table corrected, Gemini removed from the canonical examples, mapping filed as **#3721**.
- **`guides/REALTIME_CO_AGENTS_GUIDE.md` updated** — it listed this exact bug as an open TODO, and separately claimed `CreateClientSession` doesn't fold `params.Config` (it does; already stale before this PR). Both shape sites now document the agnostic slot.

**Docs / schema (4)** — canonical-shape JSDoc, reciprocal precedence on `providers.*.voice`, the same-layer specificity inversion, and "opaque pact" scoped to runtime overrides (I chose wording over `additionalProperties: true`, since relaxing validation for every existing config to fix a docs mismatch seemed the wrong trade — happy to revisit).

**Changeset (3)** — `patch` → `minor`, plus the client/server skew case leading (it regresses working behaviour, unlike the others) and the bridge-host case.

Two of your points I deliberately did **not** implement, with reasons in-code rather than silently: the fallback-unless-runtime-override precedence rule (needs the same per-layer provenance the documented cross-layer limitation needs — half of it would swap one surprise for another), and a runtime guard on voice-id validity (vendor-specific; this layer can't distinguish a good id from a bad one). Both are documented as open rather than closed.

Also took your evidence-table note: the `IT — realtime-deterministic` row now carries the caveat that the bundle has zero voice coverage, so it attests nothing about this change, and the `SupportedVoices` picker-reach limitation is stated in the PR body, the changeset and the guide.

PR #3717, opened 2026-08-10
fix: resolve runtime-configured packages from the host under pnpm

### LINKED ISSUE #3677: Open App registration and migrations fail under pnpm: server-bootstrap and open-app-engine import packages they do not declare
Two packages load code by **bare dynamic import** without declaring what they import. Under npm this resolved via hoisting; under pnpm a bare specifier resolves from the **importing package**, which is given only what it declares. Both are blocking on a 6.x (pnpm) monorepo.

Found while dev-linking an Open App into an MJ `6.1.0-edge.0` instance. Both reproduce on `origin/next` @ `6.1.0-edge.1`.

---

## 1. `@memberjunction/server-bootstrap` cannot load Open App server packages

`packages/ServerBootstrap/src/index.ts` reads `dynamicPackages.server[]` from `mj.config.cjs` and does `await import(pkgName)` for each entry. It declares none of them — and cannot, since the names are only known at runtime.

**Symptom** (MJAPI starts, then):

```
Error loading Open App server package @mj-biz-apps/common-server:
  Error: Cannot find package '@mj-biz-apps/common-server' imported from
  .../packages/ServerBootstrap/dist/index.js
```

The package is present and correctly linked into `packages/MJAPI/node_modules/@mj-biz-apps/common-server` — MJAPI declares it. But ServerBootstrap is the module doing the import, and it has no such dependency, so **no Open App can register into MJAPI on pnpm at all.**

## 2. `@memberjunction/open-app-engine` cannot run app migrations

`packages/OpenApp/Engine/src/install/migration-runner.ts` does `await import('@memberjunction/skyway-core')` through an indirection variable. Its own comment states the intended contract:

> The skyway provider packages are published as (optional) dependencies of the host process (e.g. MJCLI).

Only MJCLI declares them. That contract cannot hold under pnpm: a bare specifier resolves relative to the importing module (`OpenApp/Engine`), not to the host entrypoint, so **no host can satisfy it.**

**Symptom:**

```
migrations failed: Migration failed for schema '__mj_BizAppsCommon':
  Cannot find package '@memberjunction/skyway-core' imported from
  .../packages/OpenApp/Engine/dist/install/migration-runner.js
```

---

## Reproduce

1. Create an instance on MJ 6.x (pnpm) and complete deps + build.
2. Dev-link an Open App as a workspace member (e.g. `bizapps-common@next`), so its packages appear in `dynamicPackages.server[]`.
3. Run the app's migrations → failure 2.
4. Start MJAPI → failure 1.

Neither reproduces on 5.x/npm, where hoisting puts both in the root `node_modules`.

## Suggested fix

For **open-app-engine**: declare the skyway packages as `optionalDependencies`, or resolve them with `createRequire` against an explicitly-passed host path rather than a bare specifier — the current comment describes a host-provides contract that the resolver cannot honour.

For **server-bootstrap**: the package names are dynamic, so a declaration is not possible. Resolve each entry with `createRequire` rooted at the **host** (the package whose `mj.config.cjs` supplied the entry, i.e. MJAPI) instead of importing bare. That also makes the resolution origin explicit rather than incidental, which is what made this work on npm by luck.

## Workaround (for anyone blocked today)

pnpm `packageExtensions` in `pnpm-workspace.yaml`, declaring the missing deps on MJ's behalf:

```yaml
packageExtensions:
  '@memberjunction/open-app-engine':
    dependencies:
      '@memberjunction/skyway-core': '^0.6.2'
      '@memberjunction/skyway-sqlserver': '^0.6.2'
      '@memberjunction/skyway-postgres': '^0.6.2'
  '@memberjunction/server-bootstrap':
    dependencies:
      '@mj-biz-apps/common-server': 'workspace:*'   # one line per dev-linked app server package
```

With both applied, the full path works on 6.1.0-edge.0: link → migrate → codegen (384 entities) → sync → build → MJAPI ready → a GraphQL query against an app entity returns `Success: true`.

Reported from MJDev (MJ Dev Manager). Happy to open a PR for either fix if the `createRequire` direction is the one you want.

MarceloT-BC:
## Summary 
Two packages loaded code by bare dynamic `import()` without declaring what they import. npm's hoisted `node_modules` let that resolve by accident; under pnpm a bare specifier resolves from the *importing* package, which only gets what it declares — so Open App registration and app migrations both fail on a 6.x (pnpm) workspace.

## Packages
**`@memberjunction/open-app-engine`** — the skyway packages are now `optionalDependencies` (`^0.6.2`, matching MJCLI's pins), so the existing dynamic import resolves under pnpm's strict layout. The import stays dynamic so the module still compiles and loads when they're absent (per the dynamic-import policy, category 2).

**`@memberjunction/server-bootstrap`** — the `dynamicPackages.server[]` and `codeGeneration.packages` names are only known at runtime, so they can't be declared. `importFromHost()` tries the bare import first (identical behavior on npm layouts), and on a resolution failure retries via `createRequire` anchored at the host application: cwd, then the directory of the `mj.config.cjs` that named the package, then the process entrypoint. One subtlety: MJAPI runs under ts-node's ESM shim, which throws resolution failures as plain `Error`s with **no `code`** — so the failure predicate matches the standard codes (`ERR_MODULE_NOT_FOUND`, `MODULE_NOT_FOUND`, `ERR_PACKAGE_PATH_NOT_EXPORTED`) and, when no code is present, Node's own resolver message.

## How to reproduce the original failures:
Why this affects standard MJ usage: dynamicPackages.server[] is written by mj app install and consumed by server-bootstrap on every MJAPI boot, and RunAppMigrations is the install flow's migration step — both failures live in those paths and depend only on where the two packages sit in the pnpm workspace, not on how the app was added. The repro below uses a workspace-linked app because no Open App has a 6.x-compatible published release yet; the resolution failure is identical for installed packages consumed from a workspace checkout.

on an MJ 6.x (pnpm) instance, dev-link an Open App (e.g. `bizapps-common@next`) with no `packageExtensions` workarounds, then run its migrations and start MJAPI:
- migrations: `Cannot find package '@memberjunction/skyway-core' imported from .../OpenApp/Engine/dist/install/migration-runner.js`
- MJAPI boot: `Error loading Open App server package @mj-biz-apps/common-server: Cannot find package ... imported from .../ServerBootstrap/dist/index.js` (while the package is correctly linked into MJAPI's own `node_modules`.

## Fix
**With this fix**, same setup: migrations apply, MJAPI loads the app server package + its resolver paths, and a GraphQL query against an app entity returns `Success: true`. ServerBootstrap (15) and OpenApp/Engine (332) unit tests pass. `pnpm-lock.yaml` carries the matching 10-line delta for `--frozen-lockfile` CI. `ServerBootstrapLite` was audited and has no bare dynamic imports.

**Scope note:** this PR covers the *resolution* failures only — a module not found at all. A separate question about module **identity** under pnpm (a package found twice rather than not found) is still being researched and will be tracked in its own issue once we have a reproducing topology; it does not affect the fixes here. An earlier version of this section stated a specific mechanism for that second case which we have since disproven — see the correction in the comments.

Fixes #3677


MarceloT-BC:
A note on how this PR relates to the broader failure pattern, since a third instance of it has now turned up in our testing.

All three share one theme: **MJ's server stack assumes module identity, while its dependency graph permits multiple physical copies (or zero reachable copies) of a package.** npm's hoisted `node_modules` made both assumptions hold *by accident* — everything deduped to one hoisted copy that everyone could reach. pnpm's strict layout removes the accident, and the theme surfaces as two distinct failure shapes:

1. **Resolution failures** (this PR): a bare dynamic import resolves from the importing package, which never declared the dependency → module not found at all. Fixed here for `open-app-engine` (declare the skyway packages) and `server-bootstrap` (resolve runtime-configured names from the host).
2. **Identity failures** (companion, to be filed): the module IS found — twice. type-graphql keeps its decorator metadata in per-copy storage, so when the graph forks into two physical copies, registrations land in one copy's storage while `buildSchema` reads another's. Nothing fails to resolve; the schema is just silently wrong. This is the same class as MJ's existing `@memberjunction/global` single-copy invariant, showing up one dependency deeper.

We have npm-vs-pnpm control evidence for the identity case and will file it as a companion issue to #3677. Two candidate fix directions for it: (a) tighten the transitive peer surface so resolution ids can't fork, or (b) make GraphQL type registration idempotent so a second module instance is harmless. We lean toward (b) — it fixes the *class* of bug rather than one occurrence, and it doesn't depend on every future dependency edge staying well-behaved.

Keeping this PR scoped to the resolution failures; the companion issue will link back here.

rkihm-BC (commented):
## Code review — Claude Code

> This review was performed by **Claude Code** (Opus 5) at @rkihm-BC's request. It reads the full diff, the PR description, and the follow-up comment, and it re-verifies the load-bearing claims empirically rather than taking them on trust. **Please push back on anything I got wrong or missed** — I ran the checks below on a local checkout of `pr-3717`, and where I could not verify something I say so explicitly.

**Verdict: approve with changes.** The diagnosis is right, both fixes are the correct shape, and the riskiest part (the error-shape heuristic) holds up under the real loader. There is **one defect I was able to reproduce**, two things worth fixing before merge, and a set of adjacent issues in the touched files.

---

## What I verified independently

| Claim | Result |
|---|---|
| ts-node's ESM shim throws resolution failures with **no `code`** | ✅ Ran `node --disable-warning=DEP0180 --import ./register.js` (MJAPI's exact launcher, `packages/MJAPI/register.js`) → `constructor.name = Error`, `code = undefined`, message `Cannot find package 'x' imported from …`. The regex branch matches. **This branch is load-bearing — without it the whole fix is dead under MJAPI.** |
| Error codes on the other two paths | ✅ Plain-Node ESM → `ERR_MODULE_NOT_FOUND`; `createRequire().resolve` → `MODULE_NOT_FOUND`. All three arms of the predicate are real. |
| The host-anchor fallback actually loads a host-only package | ✅ Built a fixture (an importer package with an empty `node_modules`, a host with `@app/good` linked into its own). Bare import fails → cwd anchor resolves → `RESOLVER_PATHS` returned → `StartupExport` ran. |
| `configFilePath` is populated | ✅ Already present on `next` (`index.ts:241`, `configSearchResult?.filepath`) — not a new field, and correctly named. |
| Typecheck + tests | ✅ Applied the PR file locally: `src/index.ts` compiles clean (the only `tsc` errors are pre-existing ones in a stale generated manifest); **15/15** ServerBootstrap tests pass. Tree restored afterwards. |
| Lockfile delta | ✅ `skyway-*@0.6.2` already has `packages:` and `snapshots:` entries (MJCLI pulls them), so only the importer block is needed — the 10-line delta is structurally complete. I did **not** run `pnpm install --frozen-lockfile` (it would mutate the working tree), so that specific CI step is unverified by me. |
| Scope is complete | ✅ `dynamicPackages.server` has exactly one consumer (ServerBootstrap); the client side is a static CodeGen'd manifest; ServerBootstrapLite has no bare dynamic imports; `RunAppMigrations` is reached only via `install-orchestrator` → MJCLI. |

One point in the PR's favour that isn't stated in the description: **`optionalDependencies` is not merely pragmatic here, it's the only compliant option.** `.github/scripts/check-internal-peer-deps.mjs` forbids any `@memberjunction/*` in `peerDependencies` (a single internal peer silently escalates all ~294 packages to a major bump during an Edge pre-mode window). Since skyway is `@memberjunction/*`-scoped, `peerDependenciesMeta.optional` was off the table. The `^0.6.2` pins match MJCLI, so pnpm dedupes to one physical copy and no identity fork is introduced. All 6 CI checks are green.

---

## Must fix

### 1. The retry loop masks the *next* failure with the error this PR just fixed

`packages/ServerBootstrap/src/index.ts:96-107` — a resolution failure thrown *from inside a module that resolved perfectly well* is indistinguishable from "this anchor couldn't see the package". The loop swallows it, exhausts the anchors, and rethrows the **original** bare-import error.

I reproduced this. Fixture: `@app/broken` is linked into the host and does `import "some-missing-transitive-dep"`.

```
reported message: Cannot find package '@app/broken' imported from …/importer/boot.mjs
>>> TRUE cause  : @app/broken/index.js imports "some-missing-transitive-dep"
```

Under pnpm, a missing transitive dep in an app package is *the* most likely next failure mode — and it will be reported as "`@app/x` cannot be found from ServerBootstrap", pointing whoever debugs it straight back at #3677. That is an expensive wrong signpost for exactly the population this PR serves.

Fix — separate resolution from evaluation, which is also simpler than what's there:

```ts
for (const anchor of anchors) {
  let resolved: string;
  try {
    resolved = createRequire(anchor).resolve(pkgName);
  } catch (resolveError: unknown) {
    if (!isResolutionFailure(resolveError)) throw resolveError;
    continue; // this anchor can't see the package — try the next
  }
  // Resolved. Anything that fails from here is the module's own problem — surface it.
  return (await import(pathToFileURL(resolved).href)) as Record<string, unknown>;
}
```

---

## Should fix

### 2. Both call-site catches still branch on `code === 'ERR_MODULE_NOT_FOUND'`

`index.ts:146` and `index.ts:237`. The PR itself establishes that under MJAPI's loader there is no `code` — so the friendly `Open App server package not found (run 'npm install'?)` line is dead in production, and the benign "app not installed yet" case instead prints `console.warn('Error loading Open App server package …', <full error object>)`. `isResolutionFailure` now exists; use it at both sites.

### 3. Import-only exports maps are silently unreachable (verified)

`require.resolve` resolves under CJS conditions. A package whose exports map declares only `"import"` throws `ERR_PACKAGE_PATH_NOT_EXPORTED` at *every* anchor, and the fallback gives up with the original "cannot find package" message:

```
require.resolve -> ERR_PACKAGE_PATH_NOT_EXPORTED
FAIL: ERR_MODULE_NOT_FOUND | Cannot find package '@app/importonly' imported from …
```

MJ's own packages use a single `"default"` condition, so they resolve today — but Open App server packages are authored **outside** this repo (nothing in CodeGenLib emits their `package.json`), so this is a live gap for third-party apps rather than a theoretical one. At minimum, distinguish `ERR_PACKAGE_PATH_NOT_EXPORTED` in the final throw with an actionable message ("reachable from the host, but its exports map has no CJS-resolvable condition") instead of repeating "cannot find package".

Related and smaller: on a dual package, `require.resolve` selects the **CJS** copy, so `import(fileURL)` can load a second physical instance alongside an ESM copy already in the process — precisely the module-identity fork described in your follow-up comment. Not reachable for MJ-shaped packages, but worth a sentence in the doc comment so nobody widens this later without noticing.

---

## Nits and adjacent issues in the touched files

- **No changeset.** With `fixed: [["@memberjunction/*"]]` in `.changeset/config.json` the packages *will* still publish on the next release, so the fix does ship — but there's no changelog line, and "Open App won't boot under pnpm" is exactly the string someone will search for. Add a patch changeset for both packages.
- **No test for the new logic.** `isResolutionFailure` is a regex against a Node error message — the most churn-prone assumption in the PR — and it's unexported, so it can't be tested directly. Export it (or drive `importFromHost` with fixtures) and cover four cases: coded ESM, coded CJS, uncoded ts-node-shaped, and a genuine load error that must propagate.
- **`migration-runner.ts:9` and `:17`** — the file header still says the dynamic import exists so the module "compiles even when `@skyway/core` is not installed". Wrong package name (`@skyway/core` vs `@memberjunction/skyway-core`), and it now contradicts the declaration this PR adds. The inline comments were updated; the header wasn't.
- **`migration-runner.ts:236-249`** — `catch { throw new Error('… not found. Install @memberjunction/skyway-postgres …') }` swallows *everything*, including a throw from `new PostgresProvider(dbConfig)` on bad connection config, and reports it as "provider not found". That's the same misdiagnosis class this PR exists to eliminate, and the advice is now stale: post-fix, the real cause would be `--no-optional` or a registry that doesn't carry the package. Narrow the catch to a resolution check and attach `{ cause: error }`.
- **Install footprint.** `open-app-engine` now transitively pulls `pg@8.18.0` and `mssql@11.0.1` for every consumer. Fine here (MJCLI already had both; integration-test-suite is server-side), but it isn't free and is worth knowing.
- **A second pin to keep in sync.** `^0.6.2` now lives in both MJCLI and open-app-engine. If one moves to 0.7 and the other doesn't, pnpm will happily install two skyway majors and the migration runner will quietly use the older one. Worth considering (not necessarily in this PR): `skyway-core` is *not* optional for `RunAppMigrations` — it's mandatory. Making it a plain `dependency` with a **static** import would fail loudly at install time, drop the hand-rolled `SkywayConfig`/`SkywayInstance` interfaces, and remove the implicit `any` from `new Skyway(config)`, leaving only the two platform providers optional and dynamic. As it stands, the manifest says "optional" about something that is required.

---

Again — I'm an automated reviewer. If any of the above is wrong, or if I've missed context that makes a finding moot, say so on the thread and I'll re-check rather than assume I was right.

MarceloT-BC:
> **Revised 2026-08-12.** A further round of testing corrected this comment again. Net effect for this PR: **nothing here changes** — the two resolution fixes stand on their own evidence — but the "identity" half I described is narrower than stated and is no longer a pnpm story at all. Edit history has the prior text.

**Correction to my original note, on point 2.** type-graphql `2.0.0-beta.3` (the version MJServer pins) anchors its metadata on `globalThis` — `if (!global.TypeGraphQLMetadataStorage) global.TypeGraphQLMetadataStorage = new MetadataStorage()` — so two physical copies **share** one storage. The identity failure is therefore not "registrations land in one copy's storage while `buildSchema` reads another's."

What actually happens, measured with two genuinely distinct files on disk compiled by tsc under MJ's own decorator flags: **identical** duplicate copies produce a **correct** schema (harmless); **divergent** copies — real version skew — let the **last-imported** copy silently win, with no server-side error; and the loud `Schema must contain uniquely named types` appears only in the narrow case where the two copies expose differently-*named* queries so both types stay simultaneously reachable. I previously had that split backwards.

**Withdrawing the pnpm framing.** pnpm's peer-context duplication is documented ([how-peers-are-resolved](https://pnpm.io/how-peers-are-resolved)) to hard-link the **same version with identical source** once per distinct peer set — which is the harmless case above. Divergence requires version skew, which npm's nested `node_modules` produces just as readily. So there is no pnpm-specific identity defect to file as a companion to #3677, and I've dropped that plan. The remaining, genuinely MJ-anchored observations are recorded on #3741, which is the better home for them.

rkihm-BC (commented):
## Re-review — Claude Code

> Follow-up by **Claude Code** (Opus 5) at @rkihm-BC's request, against `a3c9e95` (*"address review — never mask a resolved module's own failure; test the resolution heuristic"*). I re-ran the branch locally rather than reading the diff alone. **Push back on anything I got wrong.**

**All three findings are resolved, and resolved well** — not patched around. Both nits (changeset, tests) are closed too. Extracting `host-import.ts` was the right call: it turned an untestable inner function into a unit with real coverage and put `index.ts` back to a 12/9 delta.

There is **one new finding I reproduced**: fixes #1 and #2 interact, and for the single most likely failure case they cancel each other out — the true cause #1 recovers is discarded by #2 before the operator sees it. Plus three small items.

---

## Verified locally

| | Result |
|---|---|
| Typecheck | ✅ `tsc --noEmit` clean on both packages (only the pre-existing stale-generated-manifest errors) |
| ServerBootstrap suite | ✅ **26/26** — the original 15 plus 11 new |
| OpenApp/Engine suite | ✅ **332/332** |
| Finding 1 actually fixed | ✅ My own fixture: a host-visible package with `import 'a-missing-transitive-dep'` now throws `Cannot find package 'a-missing-transitive-dep' imported from …/@probe/broken/index.js` — the true cause, not the masked one |
| Finding 3 actually fixed | ✅ An `import`-only exports map now raises the actionable message instead of repeating "cannot find package" |
| Build safety of the new relative import | ✅ See the last nit — I checked this rather than assuming, and it is **not** a problem |

Tests are stronger than what I asked for: real on-disk fixture packages (`good` / `throwing` / `broken-transitive` / `importonly`) rather than mocked errors, so they exercise Node's actual resolver. The four predicate cases are covered, including the two negatives (code-less error with an unrelated message; unrelated code with a matching message).

Two things beyond the review that I want to name as good:

- **`ImportProviderClass`** is a better fix than I suggested for the swallow-everything catch — it separates resolution from construction and attaches `{ cause }`, so a bad connection config now surfaces as itself instead of "provider not found".
- The new provider message calls out **`--no-optional` installs and registries that don't carry the package** by name. That closes the practical concern behind my `optionalDependencies`-vs-`dependencies` note without changing the manifest, which I think is the right trade.

---

## New finding — #1 and #2 cancel out at the call site (reproduced)

`importFromHost` now correctly surfaces a resolved module's own missing transitive dependency. But the call sites' new `isResolutionFailure(error)` branch classifies that error as "the package is missing", logs the generic line, and **throws the message away**:

```
PROBE  importFromHost threw : Cannot find package 'a-missing-transitive-dep' imported from …/@probe/broken/index.js
PROBE  isResolutionFailure   : true
PROBE  operator sees         : "  Open App server package not found (run 'npm install'?): @probe/broken"
PROBE  true cause reaches operator? NO — the message above is discarded
```

For this case the result is **worse than before the rework**: the old `code === 'ERR_MODULE_NOT_FOUND'` check was always false under ts-node's loader, so it fell through to `console.warn(…, error)` and printed the whole thing. Fix #2 made the friendly branch reachable — and the friendly branch is lossy. Under pnpm, a missing transitive dep in an app package is exactly the failure this PR's population will hit next, so it's worth not swallowing.

Cheapest correct version — only claim "the package is missing" when the error is about *that* package:

```ts
const message = error instanceof Error ? error.message : String(error);
if (isResolutionFailure(error) && message.includes(pkgName)) {
  console.log(`  Open App server package not found (run 'npm install'?): ${pkgName}`);
} else {
  console.warn(`  Error loading Open App server package ${pkgName}:`, error);
}
```

Or, if you'd rather not branch on the message, just append it to the friendly line. Same at `index.ts:85` and `index.ts:178`.

---

## Smaller items

- **`IsModuleResolutionFailure` (`migration-runner.ts:270`) is a verbatim duplicate of `isResolutionFailure` (`host-import.ts:26`)** — the same regex-on-a-Node-error-message heuristic, in two packages, and only one copy has tests. The two packages can't depend on each other and cross-package re-exports are out per the repo's TS rules, so a shared home probably isn't worth spinning up today; but a `// Keep in sync with @memberjunction/server-bootstrap's host-import.ts` line on each would at least make the coupling visible to whoever changes one.
- **`ImportProviderClass` casts an unverified export**: `mod[exportName] as new (…) => unknown`. If the package resolves but doesn't export that name, the caller gets a raw `TypeError: PostgresProvider is not a constructor` from the `new` two frames later. A `typeof ctor !== 'function'` guard with a named message is two lines and keeps the honest-error property the rest of this function now has.
- **`RunAppMigrations`'s outer catch flattens to `.message`**, so the new `{ cause }` chain is dropped before any caller sees it. Nothing is lost today — the guidance lives in the message — just noting the `cause` is decorative on that path.
- **Style only, verified not a bug:** `import { importFromHost, isResolutionFailure } from './host-import'` is extensionless, while the sibling `mj-class-registrations.ts` writes `'./generated/mj-class-registrations.js'`. I checked whether this exposes the `check:esm` bug class and it does not — the build is `tsc && tsc-alias -f`, and I confirmed on a fixture that `tsc-alias -f` rewrites the emitted specifier to `./host-import.js` and the output imports cleanly in a fresh native-ESM Node process. Purely a consistency point.

---

CI at time of writing: 5 green, `Run unit tests` still going. Nothing here blocks merge in my view — the new finding is a two-line change to a log branch, and everything else is optional polish.

MarceloT-BC:
Thanks for both rounds — everything actionable is applied, in `a3c9e95` (first round) and `fc8f369` (re-review):

**Round 1:** resolution/evaluation split in the anchor loop (a resolved module's own failure now surfaces as-is); `isResolutionFailure` at both call sites; actionable error for import-only exports maps + dual-package note; patch changeset; 11 fixture-driven tests for the helper; migration-runner header fixed and `ImportProviderClass` narrows the "install the provider" advice to genuine resolution failures with `{ cause }`.

**Re-review:** the new finding was real and is fixed — but with one refinement to your suggested guard, which our regression test caught: a transitive failure's message contains the parent package's name **unquoted** in its `imported from …/node_modules/<pkg>/…` path, so `message.includes(pkgName)` still swallows exactly the case you reproduced. Node's resolution errors always **quote** the missing name, so the shipped guard matches `'${pkgName}'` (quoted) — the friendly line fires only when this package itself is the error's subject. There's a `createMJServer`-level test driving a real on-disk broken-graph fixture to pin it. Also applied: keep-in-sync cross-references on the two duplicated predicates, a constructor-type guard in `ImportProviderClass` (named error instead of a raw `TypeError`), and the `.js` extension on the new relative import.

**Left as-is, deliberately:** `RunAppMigrations` flattening to `.message` (per your note, nothing is lost today), and `optionalDependencies` stays rather than a static required dep — fair point that it overstates optionality for `RunAppMigrations`; happy to take that as a follow-up if maintainers want it, since it changes consumer install footprint.

ServerBootstrap 27/27, Engine 332/332, and the full path re-verified live on a pnpm instance with no workarounds (migrate → boot → app-entity GraphQL `Success: true`).

MarceloT-BC:
Final round (`288d293`), from an adversarial audit of the branch: anchor priority reordered so the mj.config.cjs that named the package is consulted before cwd (removes a wrong-checkout shadowing case), skyway-core's own import now gets the same guided optionalDependencies failure message as the providers (the common `--no-optional` failure mode previously surfaced as a raw resolver error), and a positive-path regression test asserting the friendly 'not found' line fires when a package is genuinely missing. The audit also confirmed clean: no module-identity fork from the file-URL fallback (bare and anchored imports yield the identical namespace), lockfile frozen-install-consistent. ServerBootstrap 28/28, Engine 332/332, full path re-verified live on a workaround-free pnpm instance.

MarceloT-BC:
Filed #3741 to track the rest of this bug class: the adversarial audit found four more dynamic-import sites with runtime-configured or undeclared specifiers (most notably `AIEngine.ts:1535`, which loads DB-configured AI driver modules the AI Engine never declares — the likely next real-world hit under pnpm). The issue requests a full repo-wide audit + a CI guard, and points at the two fix patterns this PR established (declare vs host-anchored resolution). Nothing there blocks this PR.

rkihm-BC (commented):
## Third-round review — Claude Code

> Follow-up by **Claude Code** (Opus 5) at @rkihm-BC's request, against **`288d293`** — note that's one commit past the `fc8f369` your comment describes ("*address adversarial audit — config anchor first; skyway-core gets the guided failure path*"). Re-ran the branch locally rather than reading the diff. **Push back on anything I got wrong.**

**Verdict: ship it.** The re-review finding is fixed, and your refinement to my suggested guard was **correct — I reproduced that my version was buggy and yours is not**. Everything else from both rounds is closed. What's left below is one residual I found while probing the guard (narrow, pre-existing, with a cheap tightening), one behaviour change from `288d293` I want you to confirm is intended, and two nits. Nothing blocking.

---

## Verified locally, at `288d293`

| | Result |
|---|---|
| Typecheck | ✅ clean on both packages (only the pre-existing stale-generated-manifest errors) |
| ServerBootstrap suite | ✅ **28/28** (up from 26 — `288d293` adds one) |
| OpenApp/Engine suite | ✅ **332/332** |
| The original #3677 repro still works under the new anchor order | ✅ host-only package loads from a config-anchored host |

Then I re-ran the shipped call-site guard against five real on-disk failure shapes, comparing it side by side with the version I proposed:

```
A host-only package (the repro)      LOADED OK
B missing TRANSITIVE dep             shipped=WARN(full error)  prior-suggestion=FRIENDLY(lossy)
C genuinely missing package          shipped=FRIENDLY(lossy)   prior-suggestion=FRIENDLY(lossy)
D import-only exports map            shipped=WARN(full error)  prior-suggestion=WARN(full error)
E module throws resolver-shaped msg  shipped=FRIENDLY(lossy)   prior-suggestion=FRIENDLY(lossy)
```

**Row B is the point.** My suggested `message.includes(pkgName)` classifies the transitive failure as FRIENDLY and swallows it — exactly the bug the guard exists to prevent. The shipped `message.includes(\`'${pkgName}'\`)` routes it to WARN carrying `Cannot find package 'probe-missing-transitive' imported from …/@probe/broken/index.js`. Your reasoning was right: the parent's name is present *unquoted* in the `imported from` path, so an unquoted `includes` matches it. Good catch, and the regression test you added is what should have caught mine.

Row D is worth calling out too: the actionable exports-map error reaches the operator in full, because the `Error` you throw has no `code` and its message starts with `Package '…'` rather than `Cannot find …`, so `isResolutionFailure` returns false and it takes the warn path. Correct — see the nit below about that being load-bearing but untested at the call site.

Also confirmed closed since round two: `IsModuleResolutionFailure`/`isResolutionFailure` now carry keep-in-sync cross-references naming each other and which copy holds the tests; `ImportSkywayClass` guards `typeof ctor !== 'function'` with a named message instead of a raw `TypeError`; the relative import is `./host-import.js`. Extending `ImportSkywayClass` to `skyway-core` in `288d293` is a real improvement — the most likely failure (all three skyway packages absent under `--no-optional`) now hits the actionable message at the *first* import rather than a raw resolver error.

Your two "left as-is" calls are both reasonable, and I agree `optionalDependencies` → `dependencies` is a follow-up rather than a merge blocker, since it changes install footprint for every consumer.

---

## One residual (narrow, pre-existing — not introduced here)

Row E: a package that **loads successfully and then throws** an error whose message happens to match Node's resolver shape *and* quotes its own name is reported as *"not found (run 'npm install'?)"* and the real error is discarded.

```js
// @probe/evilmsg/index.js
throw new Error("Cannot find package '@probe/evilmsg' in the widget registry");
```
```
shipped guard → FRIENDLY(lossy)   // the operator is told to run npm install; the real error is gone
```

This is the inherent tail of a message-based heuristic, it predates the quoted guard, and it needs an unusual coincidence. But there's a cheap tightening that eliminates it outright: Node's own code-less resolver messages **always** carry ` imported from `, and the contrived one above doesn't. So in the code-less branch of `isResolutionFailure`:

```ts
return code === undefined && typeof message === 'string'
    && /^Cannot find (package|module) '.*' imported from /.test(message);
```

I checked this against the ts-node shim message that makes the branch load-bearing — `Cannot find package 'x' imported from /path/entry.ts` — and it still matches, so the pnpm-under-ts-node path is unaffected. One thing to confirm before taking it: that the shim never surfaces a *code-less* CJS-form message (`Cannot find module 'x'` with no `imported from`), which the tighter regex would stop matching. Both copies of the predicate would need it.

## One behaviour change from `288d293` — confirm it's intended

Putting `configFilePath` ahead of `cwd` fixes the case you describe (launching instance A's server from instance B's directory). It also flips the opposite case. Demonstrated on a fixture — workspace root holds `mj.config.cjs` **and** a copy of the package, app dir (cwd) holds a different copy:

```
@probe/dual    resolved to: ROOT-COPY   <- config anchor wins under the new order (was APP-COPY)
@probe/apponly resolved to: APP-ONLY    <- config anchor missed it, cwd fallback caught it
```

Fall-through works correctly. The thing to sanity-check is the first line: when a monorepo root config sits above an app whose own `node_modules` carries the authoritative copy, a root-visible copy now wins. Under pnpm the root only carries root-declared deps, so this should be rare — but a dev-linked Open App at the workspace root (close to how this PR's repro is set up) is exactly the shape that could hit it. If MJAPI-style hosts always have their own nearest config, this is moot; I just want it to be a decision rather than a side effect.

## Nits

- **The exports-map error's routing is load-bearing but untested at the call site.** It reaches the operator only because that `Error` lacks a `code` and doesn't start with `Cannot find`. If someone later attaches `code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'` to it — a natural-looking improvement — `isResolutionFailure` flips to true, the message already contains `'<pkgName>'` quoted, and the actionable text gets swallowed by the friendly line. The unit tests cover the throw; nothing pins that it lands on the warn path in `index.ts`. One assertion in the existing `createMJServer` fixture test would lock it.
- The changeset is accurate and scoped; both packages patch-bumped. No notes.

---

CI at time of writing: `Check migrations` and the peer-dep gate green; unit tests, dependency check, integration and standards still running on `288d293`. Everything I ran locally passes. My tree was restored clean afterwards.

PR #3722, opened 2026-08-10
test: fix both failures keeping the unit-test CI gate red — as-never casts + the DOM-spec ratchet

AN-BC:
## Why

`test.yml` has been red on `next` for every recent run. Fixing the first failure exposed a second one behind it — this PR clears both, so the unit-test gate finally produces a signal again. With #3709 (always-on simple-row normalization) now merged, this is the blocker between that change and a trustworthy CI run.

## What changed

**1. The `as never` casts (`core-entity-forms`).** `check-spec-antipatterns.mjs` flagged five `as never` casts in `flow-agent-form-section.component.dom.test.ts`, all feeding `ContributeToSave(… as never)`. `ContributeToSave` takes a `TransactionGroupBase` it only hands through to the editor's `QueueSaveInto` — it never reads a member of it — so the honest double is an identity-marker object behind a single seam cast, the same shape as the file's existing `recordWithId` fake. The call sites now type-check against the real signature instead of erasing it.

**2. The DOM-spec coverage ratchet (`Angular/Generic`).** With the antipattern gate cleared, the next step in the job failed: 136 components with no DOM spec and no deferral, against `--max-none=135`. While the earlier gate was masking the ratchet, two components landed on `next` without specs: `AgentInvocationsComponent` (ng-agents) and `TaskGraphRunViewComponent` (ng-task-graph-editor). This PR adds real DOM specs for both — lazy-load and non-UUID guards, provider-driven projection, live/paused distinction, open-intent emission, and the error-vs-empty honesty cases — and ratchets the gate to the new count of **134** so the improvement locks in.

## Verification

- `node scripts/check-spec-antipatterns.mjs packages` → ✅ clean
- `node scripts/check-dom-spec-placement.mjs packages` → ✅ clean
- `node scripts/dom-test-report.mjs packages/Angular/Generic --max-none=134` → ✅ 134 ≤ 134
- core-entity-forms: **247 tests pass** (32 files) · ng-task-graph-editor: **145 pass** (6 files, 7 new) · ng-agents: **90 pass** (11 files, 6 new)

## Note on the integration job

`integration.yml` fails on `next` for environment reasons that predate this PR — missing AI provider credentials (`No valid API credentials/keys are configured`), `MJ_API_KEY` unset for the client-side bundles, and a `No acting user` error in subscription-isolation. Those are runner-configuration issues, not code issues, and are unaffected by this change; this PR restores the unit-test gate, which is the tier that can be green today.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CPdQW7f7FcLWKsK1JCqeKy)_

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -159,7 +159,7 @@ jobs:
         run: node scripts/classify-explorer-components.mjs --min 85
 
       - name: Generic DOM coverage ratchet
-        run: node scripts/dom-test-report.mjs packages/Angular/Generic --max-none=135
+        run: node scripts/dom-test-report.mjs packages/Angular/Generic --max-none=134
 
       - name: Bootstrap DOM coverage ratchet
         run: node scripts/dom-test-report.mjs packages/Angular/Bootstrap --max-none=0
--- a/metadata-optional/integration-test/test-suites/.integration-suite.json
+++ b/metadata-optional/integration-test/test-suites/.integration-suite.json
@@ -904,6 +904,17 @@
             "lastModified": "2026-08-08T15:50:31.726Z",
             "checksum": "91d8a2f26f1f5ec8ceb4f8af36dc42701163fa21b47280d03bb21bdad61289cc"
           }
+        },
+        {
+          "fields": {
+            "SuiteID": "@parent:ID",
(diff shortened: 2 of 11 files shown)


AN-BC (commented):
LGTM

PR #3723, opened 2026-08-10
Workflow cost is the run tree; a graph runs in the order it was drawn; a workflow step reads like the agent step it is

AN-BC:
Follow-on to #3716. Everything here is one family: **work that produced a plausible number or a plausible-looking run, with nothing that said it was wrong.**

Five commits — cost rollup + HITL · Flow AND-join + timeline · loop passes + silent mappings · execution ordering + payload · reviewer follow-ups.

---

## 1. Workflow cost is the run tree — R7-A, verified live

`rollUpCostToSubmittingRun` did its own arithmetic, wrong twice over and invisibly: it filtered child tasks on `AgentRunID`, so **every Prompt step's spend was absent**, and it read each nested run's `…Rollup ?? …Total`, so the answer depended on whether that nested graph had settled.

R7 originally said delete the write-back. That premise was false — `TotalCostRollup` has four display readers, all with a `?? TotalCost` fallback. Per the re-ruling (Option A): **the tree is the authority; the `…Rollup` columns are its settlement-time cache.**

Writing the sum back makes the column an *output* of the tree, non-circular only because the query reads own cost and never a rollup. Stated at the point of temptation and pinned by **WD5**, which plants an absurd rollup on a real run — the unit tests use fixtures and never execute the SQL. When the tree can't be summed, the columns are **cleared**, not left holding a stale total from an earlier settlement.

## 2. A loop's passes exist — the rollup was 11× low

The tree reaches nested work through six relationships and a loop iteration is none of them. A `While` that spent real money over three passes reported one childless node with no cost.

| | before | after |
|---|---|---|
| `TotalCostRollup` | 0.00049725 | **0.00555375** |
| `TotalTokensUsedRollup` | 939 | **17070** |

`SUM(tree.Cost)` now equals the stored rollup **exactly** on a loop-bearing workflow.

## 3. A graph is dispatched only once its edges exist

Submission wrote children first and dependencies after, leaving a window where every task existed with no prerequisites. A poll landing there claims the whole graph at once. **Observed, not theorised** — graph `C08B36E3`:

```
dependency gating 'Research: focused' written   00:00:59.653
'Research: focused' STARTED                     00:00:59.647   ← 6ms earlier
```

`Close out: approved` ran in the same wave as the research steps, before the draft it was meant to judge existed — and the graph reported Complete. Parent, children and edges now go in **one transaction**, so a half-built graph is never visible.

## 4. Steps see their payload

Two independent bugs, both silent:

- A step with **no input mapping** fell back to the raw input instead of the merged payload. A Prompt step declares no mapping by design, so `{{ _CURRENT_PAYLOAD }}` rendered `{}` and the model wrote from an empty brief. The draft step said *"the research data was empty"* — true of what it was handed, while twenty results sat in the dependency outputs beside it.
- A step with **no output mapping** *replaced* the payload instead of merging. A loop's output is a summary, so it discarded everything the iterations established — including the `brandOK` the reviewer had just set. The edges read `payload.brandOK === true` / `!== true`; against a summary the give-up branch won **every** run regardless of the decision.

**Result — the demo works end to end for the first time:** `broad → focused → draft → review → approved`, strict order, `Close out: approved` **Complete**, `gave up` **Skipped**.

## 5. Silent mappings, silent scrapes, silent suites

- `ApplyOutputMapping` skipped absent outputs with a bare `continue`. It now returns `unmapped` and the dispatcher reports it **naming what the step did return**. The demo's step named `Web Search`'s parameters against `Google Custom Search`; corrected to `Query`/`Items`.
- **Web Search** captured results with `(.*?)</div>` — non-greedy, stopping at the first *nested* div — so the link survived and **every snippet was cut out**. Ten well-formed hits carrying no information. Now sliced between block starts; an all-empty parse is reported loudly. 5 pinning tests.
- A bundle whose every check is gated out reported **PASS with 0/0 executed** — which is how TX4's overruled assertion survived a release. It now records an explicit skip naming the flag that would run it.

## 6. Flow agents cannot AND-join

A Flow agent compiles with `TraversalMode: 'sequential'`, where a step with two outgoing paths becomes **one exclusive group**. So a fan-out means "either", not "both" — my earlier fix had made `Research: focused` Skipped on every run. The demo is a chain now; **WD2 asserts the inverse invariant**: no unconditional fan-out anywhere. A metadata deletion needed a migration, since `mj sync push` is upsert-only.

## 7. HITL

Cancelled requests re-raise instead of stalling forever (widened for NULL `StepType`); cancelling a graph **withdraws** its open requests instead of leaving them in someone's inbox; cross-user `assignToUserID` is refused at submission rather than silently reassigned; and `expiresInHours` finally makes the existing expiry machinery reachable.

## 8. Timeline, and a stray-file bug

Every workflow step wore the same icon — the row `type` was right, then `icon`/`color` were set from a map keyed by node **type** and painted over it. Now by task **kind**, and a Prompt task navigates to its **prompt run**. `AIAgentPermissionHelper` passed its message as `LogError`'s second parameter, which is a **file name**.

## 9. A ForEach kept one item in five

The one in this PR with real data loss. A ForEach over five items reported five successes, reported
Complete, and kept item five.

The body's output mapping was applied **once after the loop**, against the accumulated payload —
where the output parameter it names no longer exists. So every pass deep-merged its raw result into
the shared payload, each overwriting the last, and the mapping matched nothing and wrote nothing.
Nothing was wrong on screen; the step was green.

It is now applied **per pass**, which is also the only moment at which `name[]` can append — the form
a ForEach needs to collect anything. `ApplyOutputMapping` gained an optional `into` so the append has
a list to append to; the default keeps every existing caller byte-identical. Verified live:
`loopResults` went from 1 entry to 5.

The demo agent's own mapping named an output the action does not return (`SearchResults`, which
Google Custom Search has never emitted) and used an `[index]` syntax that has never existed.
Corrected, so the demo demonstrates the thing it is named after.

## 10. What a run says about itself

Seven fixes, each one a case where the screen was confidently wrong:

| Symptom | Cause |
|---|---|
| Every step "created the whole payload from nothing" | `Payload Changes` had nothing to diff — `Task.InputPayload` holds the *authored* input and round-trips back out as spec, so the *resolved* payload-at-start is now recorded separately |
| A loop pass showed `null` on both sides | A pass has no row of its own; it now carries its own input and output. **Not** the running payload before/after — that is quadratic, and the first cut produced a **121KB** `Configuration` for a five-pass demo. A budget bounds it and states what it omitted |
| A settled workflow had a completion with no beginning | `StartedAt` is stamped by the *claim*, and a graph parent is never claimed. It takes the earliest child's start |
| A four-step graph rendered as one enormous box at **265%** | Every node arrived at the origin — the run view computed real geometry and had no way to hand it over — so zoom-to-fit fitted a one-node bounding box. The editor now accepts positions, and the dispatcher settles a computed layout onto steps that have none. Authored arrangements are never overwritten |
| A question mark on every workflow row | `Complete` is not `Completed`. Task rows missed every branch keyed on the step vocabulary |
| A dim, near-invisible icon in dark mode | The marker colour was emitted as `var(--mj-status-info)` where the CSS matches `data-color="info"`. An unmatched attribute selector is not an error, and neither is a variable that resolves to nothing |
| A Completed run containing Pending steps, reading as broken | It is not: a graph **outlives** the run that submitted it. One observed run finished in 225ms while its workflow ran 18 seconds longer. The form now polls while the graph moves and says so |

Skipped steps are hatched and struck through, and the detail panel says which of *skipped* /
*blocked* / *pending* it is — three situations that look identical in JSON.

`AgentRunTreeStatus` claimed to be normalized across the entities. It never was, and `Completed` was
not even a member — so the type disagreed with the data for every agent step in every tree. Both
vocabularies are now declared. **The spec type gate caught this**; vitest could not, because esbuild
strips types without checking them.

## 11. Workflows → Runs

Sortable on every column with real start/end times (unset sorts last in **both** directions —
absent is not a value at one end of the range). A run opens to its canvas; clicking a step shows that
step's record, and a JSON tab shows the whole run.

---

## Verification

- Unit: `ai-core-plus` **542** · `task-graph` **67** · `core-entity-forms` **256** · `task-graph-editor` **149** · `dashboards` **1738** · `core-actions` **283**
- **Spec type gate** (`tsc --noEmit -p tsconfig.spec.json`) clean on every touched package — it is a separate gate from vitest and it caught two real defects this round
- Integration: `workflow-demo-agents` **5/5**, `task-graph-execution` **7/7** (mutation enabled)
- Full build **271/271**; the recursive CTE executed against a real database at every change
- **Live end to end** on `7C8ABB20` (rollup + loop passes), `67401956` (strict ordering + approved branch), and `CA6377BF` (per-pass payloads, resolved payload-at-start, parent `StartedAt`, `loopResults` = 5)

**Deterministic tier:** the 12 failures were **all environmental** — 7 needed `metadata-optional/integration-test` pushed, one needed MJAPI running for a client-transport check, the rest were acting-user / SQL / hung-check issues. None reference task-graph, payload mapping or the rollup.

## Known boundary, stated not hidden

A loop pass is a leaf, so a loop body that dispatches a graph of its own contributes only its run's own spend. Fixing it by traversing `ParentRunID` would double-count runs already reachable via `TargetLogID`; refusing graph-capable loop bodies would ban a legitimate composition. The real fix makes a pass a row rather than a JSON entry — flagged in the SQL where someone would hit it.


AN-BC:
## Live verification — run `69BFFFBF-F319-4E69-86B7-4A14F1A0E175`, graph `AC458370-E365-44FF-AD94-7D4E4E8B48AD`

Submitted with `mj ai agents run -a "Content Pipeline"` and dispatched by a freshly started MJAPI carrying these builds.

### The rollup wrote — all four columns, on one basis

| TotalCost (own) | TotalCostRollup | TokensRollup | PromptTokensRollup | CompletionTokensRollup |
|---|---|---|---|---|
| 0 | 0.00049725 | 939 | 729 | 210 |

The graph-reachability guard passed rather than refusing, so the tree reached the graph from the submitting run and the four columns are populated from one traversal. This is the ruling working end to end on a real submit-and-detach.

### The fan-out fix works

`Research: focused` is **Complete** on this run. On `27A0EF8E` (before the fix) it was `Skipped`, because the fan-out from `Research: broad` compiled to an exclusive group. Both research steps now execute.

### The loop iterates and reads the draft

2 iterations, 2 succeeded, and the reviewer's feedback quotes real draft content — so payload accumulation is holding and this is no longer the `[object Object]` failure.

### Two NEW defects the run surfaced — neither is graph shape, both are real

**1. `Research: focused` produces no output of its own.** Its output payload contains only `broadResearch`; the key `focusedResearch` its `ActionOutputMapping` declares is **absent entirely**, and its payload is byte-identical in length to `broad`'s (2207). The step runs, reports Complete, and its work is discarded.

```
Research: broad    keys: broadResearch    broadResearch: 10 results
Research: focused  keys: broadResearch    broadResearch: 10 results   <-- no focusedResearch
```

**2. Web Search returns results with empty snippets — 10 of 10, on both calls.** Titles and URLs only, no content. This is why the draft said the research was empty: it was *right*. There was nothing to synthesize. The reviewer disagreed because it counted sources rather than content, and the workflow closed on its give-up branch.

So the demo still ends on "gave up", but for two well-localized reasons rather than a graph that could only ever fail. Neither is fixed here — (1) needs the output-mapping path chased for a non-starting step, and (2) is the Web Search action/provider returning contentless results, which may be configuration rather than code.

AN-BC:
# Reviewer on watch — first pass on the two commits (`1a42288e7`, `93cee139e`)

I'm subscribed to this PR now; I'll review each push and post findings here. This pass covers everything currently on the branch.

**Verdict: both commits are right, and the R7-A ruling is encoded exactly as intended.** The rollup as `SumAgentRunTreeCost(LoadAgentRunTree(...))` with assignment-not-accumulation, the non-circularity property stated at the point of temptation (the query header's 🚨), WD5 planting an absurd rollup so a basis regression fails unmistakably instead of plausibly, and the closing equality assertion — `totals === tree` — is literally the ruling as a test. The AND-join analysis is correct (sequential compilation makes every fan-out an exclusive choice — plan §2/§5.4 parity, working as ruled), the demo-as-chain fix loses nothing, and the migration-for-metadata-deletion reasoning (`mj sync push` is upsert-only) is exactly right and worth remembering as a pattern. `FindCrossUserAssignments` and the `LogError` file-name catch are both clean.

Four findings, two worth fixing in this PR:

### 1. MEDIUM — the refusal paths can leave a *stale* rollup, not an absent one

"Each case logs and **leaves the column alone**, so `?? TotalCost` keeps its honest meaning" holds only when the column is null. Sequence: a submitter's graph 1 settles → rollup written; graph 2 settles later but the tree is truncated (or graph 2 isn't reachable) → refusal leaves the **graph-1-era value** in place — a number that excludes graph 2, sitting in the authoritative column. By the function's own standard ("a known-low number presented as a total is worse than no number"), the refusal branches should **null the column when it's non-null** (falling back to `?? TotalCost` honestly) — or at minimum log that a stale value remains. Small change, same function, and worth a line in the doc comment either way.

### 2. MEDIUM — TX4 still asserts the pre-R6 contract on `next` (batch-1 leftover)

`task-graph-execution.checks.ts` at `next` HEAD still demands `AssertEqual(b.Status, 'Blocked', 'the untaken branch must be Blocked…')` and `parent.Status === 'Blocked'` for a **definite-false condition** — while the dispatcher on the same HEAD routes unreachables through the skip seeds (R6, and I verified the `unreachableTaskIDs` claim-filter fix landed too — nice). Whenever the full deterministic tier runs, TX4 contradicts the live engine. Since this PR is already in the R6 cost/semantics neighborhood and its verification only cites the WD bundle (5/5), please fold the TX4 rewrite in here (or its own commit on this branch): untaken branch → `Skipped`, parent → `Complete`, and the check's comment updated — it currently teaches the overruled doctrine as design intent.

### 3. LOW — `reopenCancelledHumanTasks` misses NULL-`StepType` human tasks

The filter is `StepType='Human'`, but human tasks created before the `StepType` column existed (or by any path that doesn't write it) have NULL and would stay dead after a cancel. If you've verified no such rows can exist on any upgraded database, say so in the comment; otherwise widen to `(StepType='Human' OR (StepType IS NULL AND UserID IS NOT NULL))`.

### 4. LOW — carried forward, fine if later batch

Graph **Cancel** still leaves open requests at `Requested` forever (flagged on #3710); the cancel→re-raise cycle also accumulates one Cancelled row per cycle (bounded by human action — acceptable, just noting the inbox debris interacts with the same cleanup). And your two stated gaps — loop iterations unreachable from the tree (so the cache under-reports loop-bearing workflows) and nothing setting `ExpiresAt` — are honestly framed and stay on the §14 list; the loop-tree one is the same root cause as the batch-3 `invokeBody` work (no `AgentRunID`/per-iteration `promptRunID` written back), so they'll likely close together.

### Ruling note (recorded in the plan)

Batch 4 said "honor `assignToUserID`"; this PR **refuses** cross-user assignment until #3524 instead. That's the better call — honoring it without an authz model would let anyone put work in anyone's inbox — and the refusal is loud, named, and at the only free moment. I'm recording it as the interim semantics: *refuse now; "honor" becomes the #3524-unlock work.*

---

AN-BC:
## Reply to the reviewer — all four findings, plus both stated gaps now closed (`a44d73e4b9`)

Thanks — findings 1 and 2 were both real and both are fixed. Detail below, and the two "carried forward" gaps from your finding 4 are closed in this same push rather than deferred.

### 1. MEDIUM — stale rollup on refusal ✅ fixed

You're right, and my own sentence was the giveaway: *"leaves the column alone, so `?? TotalCost` keeps its honest meaning"* is only true when the column is null. Two graphs, first settles and writes, second can't be summed → the first graph's total sits in the authoritative column excluding work that has since happened. `?? TotalCost` protects a reader from null, not from stale.

All three refusal paths now route through `clearStaleRollup`, which nulls all four columns and logs why. A run with no rollup yet is untouched — nothing stale to clear, and writing nulls over nulls would churn Record Changes for nothing. The invariant is now stated plainly: **when the column is present, it equals the tree.**

### 2. MEDIUM — TX4 on the pre-R6 contract ✅ fixed, and worse than it looked

Rewritten to `Skipped` / parent `Complete`, with the comment replaced — it was teaching the overruled doctrine as design intent, which you correctly called out as the bigger harm.

**But the reason it survived is worth its own finding:** the deterministic tier skips every `RequiresMutation` check by default, so `IT — task-graph-execution` was reporting **PASS with 0/0 checks run**. That is why the tier showed it green in my last verification while the assertion contradicted the engine. With `RUN_MUTATION_TESTS=1` the bundle is **7/7** including the corrected TX4. A bundle that passes while executing nothing is the same failure class this PR keeps closing, so it may deserve treatment of its own — a zero-check bundle arguably should not report PASS.

### 3. LOW — NULL `StepType` ✅ widened

Checked rather than assumed: the column **is** nullable and 4 such rows exist in the reference database, though none currently carry a `UserID`. Widened to `(StepType='Human' OR (StepType IS NULL AND UserID IS NOT NULL))`. The notified-marker predicate already scopes this to tasks the dispatcher raised a request for, so the widening can't pull in unrelated work.

### 4. LOW — the two gaps are now closed, not carried

**Loop iterations reachable from the tree.** Your diagnosis of the root cause was right, and one part needs correcting: stamping `ParentRunID` would *not* have been sufficient. The tree never joins on `ParentRunID` — its six members join `s.AgentRunID`, `s.ID`, `OutputData.parentTaskID`, `tk.ParentID`, `s.TargetLogID`, `tk.AgentRunID`. So loop passes needed a record of their own. The dispatcher now writes one entry per pass to `Configuration.runtime.iterations`, and the tree expands them into nodes (outside the recursion — a pass is always a leaf, and T-SQL forbids APPLY in a recursive member).

The magnitude is the headline:

| | before | after |
|---|---|---|
| `TotalCostRollup` | 0.00049725 | **0.00555375** |
| `TotalTokensUsedRollup` | 939 | **17070** |

Same workflow. That 11× difference is the loop — spent, and not counted. And the tree's `SUM(Cost)` now equals the stored rollup **exactly** (`0.00555375` both sides) on a loop-bearing workflow, which is the equality your implementation contract asked to be pinned, holding live rather than in a fixture.

**`ExpiresAt`.** A Human step can now declare `expiresInHours`; the compiler carries it, persist writes it into the row the dispatcher reads, and `raiseHumanRequest` sets the column. Absent still means no deadline — expiring on a timeout nobody chose is worse than waiting.

### Also in this push: the mapping defect behind the demo's "empty research"

`ApplyOutputMapping` skipped absent outputs with a bare `continue`. Skipping is correct; skipping *silently* is what let a step run, cost money, report Complete and contribute nothing. It now returns `unmapped` and the dispatcher reports it **naming what the step did return** — which turns a four-table forensic exercise into one log line. The demo's own mappings named `Web Search`'s parameters against `Google Custom Search`; corrected to `Query`/`Items`, both research steps produce content for the first time (10/10 results with real snippets each, where the draft previously — correctly — reported nothing to synthesize).

### Ruling note

Agreed on recording *refuse now; "honor" becomes the #3524-unlock work* as the interim semantics for `assignToUserID`.

**Verification:** ai-core-plus 537 · task-graph 67 · core-entity-forms 250. Integration `workflow-demo-agents` 5/5, `task-graph-execution` 7/7 (mutation enabled). Full build 271/271. Live end to end on run `7C8ABB20`.

AN-BC:
# Reviewer — `a44d73e4b` verified; one correction accepted, one new LOW, two process items

All four findings confirmed fixed in code, not just claimed: `clearStaleRollup` nulls all four columns on every refusal path with the no-op-when-already-null guard (and the rewritten doc states the invariant exactly — *present ⇒ equals the tree*); TX4 now asserts Skipped/Complete with the old doctrine-teaching comment replaced by an honest account of why it was rewritten; the NULL-`StepType` widening ships with the verified row count rather than an assumption; and the iteration trace is written **inside `absorb`** so a loop that fails partway still records the passes that ran — the right placement, called out for the right reason.

**Your correction of my diagnosis is accepted.** I said stamping `ParentRunID` would make loop-spawned runs "enter the tree" — wrong mechanism: the tree's six members never join `ParentRunID`, so parentage alone was never reachability. `Configuration.runtime.iterations` + the `CROSS APPLY OPENJSON` expansion outside the recursion (a pass is always a leaf; APPLY is illegal in a recursive member) is the correct design, and the live 11× equality — stored rollup = tree SUM to the digit on a loop-bearing workflow — is exactly the pin the ruling asked for. The `JSON_QUERY`-not-`JSON_VALUE` comment will save someone an afternoon someday.

### New LOW — a pass node is a leaf *by construction*, and one shape can hide under it

The iteration expansion emits passes as leaves, outside the recursion — correct for prompt passes and for plain sub-agent passes. But a **loop-body agent that itself dispatches a graph** (a Loop agent with `enableTaskGraphs`, used as a loop body) contributes only its run's *own* spend: the recursion can't descend into pass nodes, so a nested graph under a pass is unreachable — the same under-count class, one level deeper. Rare today (Flow bodies are refused; opted-in Loop agents as loop bodies are exotic), so I'd take a one-line comment in the SQL's expansion block acknowledging the boundary — or, if you'd rather close it structurally, a refusal at submit for graph-capable loop bodies until passes can recurse. Your call which; silence is the only wrong option, per this PR's own doctrine.

### Endorsed, and worth its own item: PASS with 0/0 checks

The TX4 survival mechanism you uncovered — the deterministic tier skipping every `RequiresMutation` check and reporting the bundle **PASS with zero checks executed** — is the strongest finding in this reply, and it's the same species as the suite-sequencing guard from #3649: an invariant that was convention until it broke once. Suggested shape: a bundle whose executed-check count is 0 while it registers ≥1 check reports **SKIPPED** (surfaced in the suite summary), never PASS, and the Definition of Done names `RUN_MUTATION_TESTS=1` as required when engine semantics change. Happy to see it as a small follow-up PR rather than here — but it should exist before the next semantics change, because TX4 will not be the last check this bites.

### Before merge

- **Changeset**: the bot's right that there's none — this touches `task-graph`, `ai-core-plus`, `core-entities`, `ng-core-entity-forms`, all published; prior PRs in this series carried one.
- The interim `assignToUserID` ruling (*refuse now; honor = #3524 unlock*) is recorded in plan §14, as agreed.

Watching for the next push.

---

AN-BC:
**CI note — the Integration (SQL Server, deterministic) failure on `a44d73e4b` is not this PR.**

The run failed 59/78, but every failure is environmental and every PR-relevant bundle is green:

- **PR-relevant, all passing**: `task-graph-execution` ✓ (0/0 — the known mutation-skip hole), `task-graph-orchestration` ✓, `workflow-demo-agents` ✓ 5/5.
- **The 15 failures split into two environmental classes**: four client-transport bundles died on `MJ_API_KEY is not set in the environment`, and the rest (`shipped-agents-live` SA1–SA4, `remote-op-ai-authoring`, `prompt-runner`, `concurrent.CC1`) on `No valid API credentials/keys are configured` for any model vendor — secrets absent from the runner, nothing this diff touches. The two stragglers (`entity-writes.EW8` FK on `Tag`, `subscription-isolation.SI1` acting-user) are outside this PR's blast radius entirely.
- **The base branch reproduces it**: `next` has three consecutive Integration Tier failures in the same window (`70278614b`, `39d1ba9f5`, `c5116180c`), with another `next` run in flight now.

So: CI red on this check, failing on the base branch too — will re-run when the base recovers. No code action on this branch.

One observation for the 0/0-PASS follow-up: `task-graph-execution` reporting ✓ at score 0.0000 in this very log is the defect live in CI — the SKIPPED-not-PASS guard would have made this run's summary honest.

---

AN-BC:
# Reviewer — `12591c5f5` verified; one new LOW, and two items still hold the green light

This commit closes five real defects and I verified each in the diff, not the description:

- **The submission race is the find of the PR.** Timestamped proof (`Research: focused` started 6ms before its gating edge existed), and the fix is the right shape: children + edges in one `RunInEntityTransaction`, with capability-narrowing (`SupportsEntityTransactions`) instead of a type assertion the interface doesn't promise, and honest pass-through degradation for browser providers. This bug would have corrupted execution order on *every* sufficiently-unlucky submit, and it explains "Close out ran before the draft existed" perfectly.
- **`inputPayload` → `payload`** — one word, and every unmapped step (Prompt steps by design) now actually sees what its dependencies produced. The narrative of the draft step truthfully reporting "the research data was empty" while twenty results sat beside it is exactly the *plausible-run-that-lied* family this PR hunts.
- **No-mapping output now merges instead of replacing** — consistent with the mapped path and the doc comment; the unreachable-approved-branch analysis (summary object clobbering `brandOK`, so `!== true` always won) is correct.
- **Cancel now withdraws open requests as `Canceled`** — closing the gap carried from #3710, and `Canceled`-not-`Expired` is the right call for exactly the stated reason: the dispatcher treats expiry as routable failure, which would misstate a deliberate stop.
- **The 0/0-PASS guard landed in-PR** rather than as a follow-up — a fully-gated bundle now emits an explicit `.gated` outcome naming the flag, so the suite can no longer claim coverage it didn't run. Better than my suggested scope.
- Web-search slice-between-starts (with the zero-width guard), the all-snippets-empty loud warning, 5 pinning tests, and the judge-the-given-draft convergence fix all check out.

### New LOW — the parent is persisted *outside* the transaction

`persistParent` runs before `RunInEntityTransaction`, so if the children/edges transaction throws, submit correctly returns `Success: false` — but a parent Task row is already durably in `Pending` with zero children. It never settles (`propagateAndRollup` skips empty reloads, deliberately) and it never dies: the active-graph scan picks it up **every poll, forever** — permanent debris plus a permanent tick of wasted work per failed submit. Two easy shapes: move `persistParent` inside the transaction (edges only reference child IDs, so ordering within the transaction is unchanged), or mark the parent `Failed` in the submit catch. The first is cleaner.

Also verified, no action needed: unmapped **Action** steps now receive the full merged payload too, and `buildParams` forwards every key as an `ActionParam` — I checked the engine side and undeclared params are inert (actions read by name; `buildOutput` only echoes `Output`/`Both`), so the blast radius is execution-log size, not behavior.

### Still holding the green light

1. **Nested-graph-under-a-pass** (from my previous review) — unaddressed and unacknowledged: a pass node is a leaf, so a loop-body agent that dispatches its own graph contributes only its run's own spend. A one-line boundary comment in the SQL's expansion block is the minimum; a submit-time refusal for graph-capable loop bodies is the structural option.
2. **Changeset** — still none, and the scope has grown to 8 packages (the bot's suggested list is now accurate: `core-entities`, `task-graph`, `core-actions`, `ai-engine-base`, `ai-core-plus`, `ng-core-entity-forms`, plus the two testing packages).

Both are small. With those two plus the parent-row LOW addressed, I expect to green-light.

---

AN-BC:
## Reply — all three addressed (`ca4feb4ab6`), plus the tier is now triaged

### NEW LOW: parent outside the transaction ✅ fixed

Correct, and it's the nastier half of the race I'd just fixed: a failed submit left a childless parent durably in `Pending`, which never settles (the rollup skips a graph with no nodes) and never dies, so the active-graph scan picks it up **every poll forever**. I took your first shape — all three writes in one transaction. Ordering inside is unchanged since edges only reference child IDs.

### Nested-graph-under-a-pass ✅ acknowledged and bounded — with a disagreement on the structural option

You're right that I neither addressed nor acknowledged it; that's now a `🚧 KNOWN BOUNDARY` block in the expansion itself. I went with the comment rather than either structural option, and I want to be explicit about why, because I think both alternatives are worse:

- **A `ParentRunID` recursive member would double-count.** A sub-agent run is already reachable through its Sub-Agent step's `TargetLogID`. Adding parentage as a second path returns the same run twice and the SUM counts it twice — which is exactly the compounding the query header exists to warn against, traded for a subtree that is rare today. That would be re-opening the bug this PR closed, in a new place.
- **Refusing graph-capable loop bodies at submission bans a legitimate composition** (a loop over a sub-agent that happens to use a workflow) to protect a number.

The honest fix is to make an iteration's run a first-class node the recursion can descend from — which requires a pass to be a **row** rather than a JSON entry. That's a real design step, not a patch, so the limit is stated where someone would hit it instead of being quietly absorbed. Happy to take the row-per-pass work as its own change if you'd rather it land now.

### Changeset ✅ added

Covers all eight packages, with the behavioural changes described rather than the file list.

---

## The deterministic tier — triaged, and it was an environment gap

I said I couldn't establish these were pre-existing. Now I can: **all 12 failures were environmental, none touch this branch.**

| Reason | Bundles |
|---|---|
| `seeded agent 'IT: …' not found — push metadata-optional/integration-test` | agent-carry-forward, agent-compaction-e2e, agent-loop-live, agent-memory-guards, agent-plan-mode, agent-rag-search, agent-skills-live (7) |
| `no GraphQLDataProvider resolvable in a client-transport run` | transaction-groups — I had stopped MJAPI so the tier couldn't race the dispatcher |
| `No acting user available for 'RecordProcess.RunNow'` | subscription-isolation |
| server-subclass refusal / `Error executing SQL` | entity-server-invariants, entity-writes |
| hung check exceeded the run budget | agent-runner |

Not one references task-graph, payload mapping, or the rollup. `task-graph-execution` and `workflow-demo-agents` were green throughout.

I've now pushed `metadata-optional/integration-test` (265 records) and restarted MJAPI, and the tier is re-running — I'll post the corrected number when it lands. Worth noting for the suite's own sake: **7 of 12 "failures" were a missing optional metadata push**, which is a setup requirement the tier doesn't state anywhere it would be read.

AN-BC:
# Reviewer — `ca4feb4ab` closes everything I was holding; one answer, then approval on your tier number

All three verified in the diff:

- **Parent inside the transaction** — took the cleaner shape, ordering preserved. Agreed it was the nastier half of the same race: your original fix stopped a *mis-ordered* graph; this stops an *immortal zombie* one.
- **The `🚧 KNOWN BOUNDARY` block** is exactly what I asked for, and your double-count argument against the `ParentRunID` member is correct — `TargetLogID` already reaches sub-agent runs, so parentage as a second path would re-open the compounding bug in the very query whose header warns against it. Good refusal, well documented.
- **Changeset** — behavioral descriptions rather than a file list, right minor/patch split.

**On your question — row-per-pass as its own change, not here.** This PR is at three commits and ~1,100 lines closing eleven defects; making an iteration a first-class row is a schema step with its own migration, CodeGen pass, and blast radius, and bolting it on now would bury both. The boundary comment is sufficient for this PR precisely because it's stated where the next person will hit it. File it as the follow-up alongside the 0/0-PASS guard's sibling (the *"7 of 12 failures were a missing optional metadata push the tier never states"* finding wants the same treatment — a setup requirement that lives only in tribal knowledge is the documentation flavor of the silent-failure family this PR hunts).

**Your triage of the deterministic tier is accepted** — none of the 12 touch this branch, and `task-graph-execution` / `workflow-demo-agents` stayed green throughout.

I'm holding formal approval on exactly one thing: the corrected tier number from your re-run with the metadata pushed and MJAPI up. Post it and, assuming the task-graph bundles hold, I'll approve.

---

AN-BC:
## Tier result — corrected, and the residual named

I pushed `metadata-optional/integration-test` (265 records) and re-ran. The full tier then **wedged on a Google/Gemini provider retry loop** (512 retry lines, `attemptNumber: 52`) — the same upstream stall that ate my earlier attempt, so the suite-level number is not obtainable right now. I verified the previously-failing bundles individually instead:

| Bundle | Before | After |
|---|---|---|
| `agent-carry-forward` | ✗ seeded agent not found | **6/6** |
| `agent-plan-mode` | ✗ seeded agent not found | **6/6** |
| `task-graph-execution` | 7/7 | **7/7** |
| `workflow-demo-agents` | 5/5 | **5/5** |
| `transaction-groups` | ✗ TG5 | **4/5** — TG5 still fails |

So the metadata push resolved the agent-bundle failures, exactly as their messages said it would.

**The one residual is `transaction-groups.TG5`**, and it is not this branch's:

> `no GraphQLDataProvider resolvable in a client-transport run — the SEC1 pin cannot execute (wiring regression)`

MJAPI is up (confirmed listening) and it still fails, so it isn't a "server not running" issue — the check needs the run itself to be in client-transport mode. My commits touch none of it; the only `transaction-groups` file in my diff is `IT47-transaction-groups.json`, and that is a `sync` block stamped by my metadata push, not a change to the check. The message calls itself a wiring regression, which reads like a pre-existing item for whoever owns that pin.

Two things worth recording for the suite itself, both of which cost me real time:

1. **7 of 12 "failures" were a missing optional metadata push.** That's a setup requirement the tier never states anywhere a person would read it before running.
2. **A provider stall wedges the whole tier** with no progress output — 512 retries deep and zero bundles reported. Combined with the 0/0-PASS issue this PR just fixed, the suite has two ways to be uninformative while looking like it is working.

AN-BC:
# Reviewer — ✅ APPROVED at `ca4feb4ab` (as comment: both sessions share a GitHub identity, so a formal review-approval is blocked)

The corrected tier result satisfies the last hold: `task-graph-execution` **7/7** (mutation enabled) and `workflow-demo-agents` **5/5** on the final head, the seven agent-bundle failures resolved by the metadata push exactly as their messages predicted, and `transaction-groups.TG5` verified out of scope — its SEC1 client-transport pin names itself a wiring regression, and this branch's only touch on that bundle is a sync stamp.

**What this PR is, in one paragraph:** eleven defects in the same family — work that produced a plausible number or a plausible-looking run with nothing that said it was wrong. A cost rollup that was arithmetic beside the tree instead of a projection of it; loop passes that spent 11× the reported cost invisibly; a submission race that ran a judge before the draft existed; unmapped steps starved of their own dependencies' outputs; a loop summary clobbering the payload that made an approved branch unreachable; mappings, cancellations, deadlines, and assignments that failed silently; and a test bundle that reported PASS while executing nothing. Every fix ships with the invariant stated at the point of temptation, and the four review rounds on this thread each made the PR smaller in claims and larger in verification.

**Recorded for follow-up** (not this PR): row-per-pass as first-class tree nodes (the stated leaf boundary); the two suite findings — the undocumented `metadata-optional/integration-test` setup requirement that manufactured 7 phantom failures, and the provider-stall wedge that eats the tier with zero progress output; and graph-Cancel's remaining cancel→re-raise row growth. The `assignToUserID` "honor" semantics unlock with #3524, per the interim ruling in plan §14.

Green-lit to the maintainer for human review and live testing.

---

AN-BC:
# Reviewer — `a1b87000c` reviewed; approval stands, with one claim that outruns its wiring

The three fixes are real and verified:

- **Task-as-`step` with `ProjectTaskToStepView`** is the right architecture — translate at the boundary, reuse every branch keyed on `type === 'step'`, mark workflow work with `provenance` instead of a parallel type system. The view-model-not-entity warning (never `.Save()` a projection) is exactly the kind of comment that prevents a future bug. The per-parent `StepNumber` counters and the provenance boundary (a Run under a task reverts to ordinary styling) are both correct calls.
- **The sentinel kill** — `new Date(8_640_000_000_000_000)` rendering as a real clock time on every Pending row is a textbook member of this PR's family: fabricated data indistinguishable from data. Nullable `startTime` + em dash, ordering moved to the query where it belongs.
- **The ORDER BY forensics** are right on both counts (T-SQL NULLS FIRST putting unstarted work above executed work; Sequence 0 + no StartedAt falling through to GUID order), and the ⚠️ *persist-order-is-not-author-order* honesty block — naming topological order as the genuine fix and why it can't be had without edges — is the correct way to leave it.

### LOW — the Action-task log link is claimed, not wired

The commit message lists *"an action offered no link to its execution log"* among the fixed defects, and `ProjectTaskToStepView`'s comment says the tree resolves *"an Action task its execution log."* Neither is true yet: the tree's final SELECT only reassigns `SourceEntity`/`SourceID` for **prompt** runs (`runtime.promptRunID`), so an Action task keeps `SourceEntity='MJ: Tasks'` and the projection correctly derives `TargetLogID: null` — the detail panel falls back to the Task row, same as before. Nothing records the log ID: `TaskGraphActionRunner.RunActionForTask` drops it on the floor even though **`ActionResult.LogEntry` already carries the `MJActionExecutionLogEntity`**.

The wiring is small and symmetric with what you built for prompts: return `result.LogEntry?.ID` from the runner, stamp it as `runtime.actionLogID` beside `promptRunID`, resolve it in the tree's final SELECT the same way `pr` is joined. Either wire it (one more small commit) or trim the claim from the comment — a stated link that resolves to null is the documentation flavor of the silent-failure family. Given everything else in this PR, I'd wire it.

Approval from `ca4feb4ab` **stands** — this is presentation-layer, the cost/ordering invariants are untouched, and 250 form tests hold with the four sentinel-pinning tests updated rather than deleted.

---

AN-BC:
# Reviewer — `50cbcea59` verified; the action-log LOW is closed properly, approval stands

All four pieces confirmed in the diff:

- **`runtime.actionLogID` is wired end to end and exactly symmetric with the prompt path**: `ActionResult.LogEntry.ID` carried out of the runner, stamped by `configurationWithRuntime`, resolved in the tree's final SELECT via a second LEFT JOIN with the two-way CASE on `SourceEntity`/`SourceID`. Existing runs falling back to the Task row (nothing recorded a log ID for them) is the honest behavior. This closes my LOW from `a1b87000c` the right way — wired, not claim-trimmed.
- **The payload rename as translation** (`Input/OutputPayload` → `PayloadAtStart/PayloadAtEnd`) completes the view-model, and the note about the rejected alternative — a second field-list panel being "a different panel rather than the same one, which is the opposite of the point" — is the right design reasoning committed where it will be found.
- **The draw/render split** is clean: `TaskGraphRenderType` as a superset, `IsAuthorableNodeType` as a compiler-enforced seam rather than a cast, and the palette deliberately not polluted with un-configurable entries. "AGENT STEP · No agent chosen yet" on a healthy ForEach was indeed a claim that the step was broken — same family as everything else here.
- **The token fix** replaces two **non-existent** custom properties (`--mj-color-info`, `--mj-color-primary` — silent `var()` failures by CSS design) with real semantic tokens, and leaves the warning that prevents recurrence. The hex values on the new render entries match the file's existing palette convention, so no token-rule concern there.

Nothing new to raise. Approval from `ca4feb4ab` stands through `50cbcea59` — the run-view feedback loop this closes was live-test-driven, which is exactly the sequence that should precede a merge.

---

AN-BC:
**Reviewer — `6a653b6f1` verified, approval stands.** Correct catch: the embedded run view bypasses `GetAgentRunTree`, so the ordering fix had to be restated at its own `OrderBy` — same rule (started work in execution order, unstarted last, persist order as tiebreak), verified live on the settled Content Pipeline with the Skipped branch listing last. That's now two independent consumers encoding one ordering rule; if a third appears, it's worth hoisting into a shared helper — noted for the follow-up list, not this PR.

---

AN-BC:
## Final wave — run-view parity, ordering, and three more defects (`a1b87000c2` · `50cbcea595` · `6a653b6f1d`)

### A workflow step now reads like the agent step it is

The run timeline routes on a step's **shape**: `type === 'step'` + `data.StepType` picks the icon and every detail tab, `TargetLogID` is what a row opens, `ParentID` is what makes a loop's passes render as children. Graph tasks carried none of it — I'd given them row types of their own, which got the icon right and missed everything else.

A task is now a `step` carrying a translated view (`ProjectTaskToStepView`), with `provenance` marking it as dispatcher work. Verified on a fresh run:

```
d3 [Action] Research: broad        -> MJ: Action Execution Logs   payloads✓
d3 [Prompt] Draft the piece        -> MJ: AI Prompt Runs          payloads✓
d3 [While]  Review against brand   -> MJ: Tasks                   payloads✓
d4 [Prompt] Pass 1                 -> MJ: AI Prompt Runs
```

`runtime.actionLogID` closes the last gap — actions had no equivalent of `promptRunID`, so a workflow's action could not offer "view the execution log" while an agent step always could. And a task's Input/OutputPayload are mapped to `PayloadAtStart`/`PayloadAtEnd`, which is what gives the shared panel the same before/after diff rather than a raw dump.

> One thing I tried and reverted: rendering the underlying record as a field list. It was a *different* panel, which is the opposite of the goal. Parity comes from carrying the same fields, not from a second presentation.

### Steps were listed in GUID order, and unstarted rows showed an invented time

Two defects in one `ORDER BY`: **NULLs sort first in T-SQL**, so unstarted work sorted *above* work that ran (a settled graph led with its Skipped branch); and with every task at `Sequence 0` and no `StartedAt`, the sort fell through to `NodeID` — a GUID. Separately the projection filled `startTime` with the maximum `Date` as a sort sentinel, and it **rendered**: every Pending task showed the same fabricated clock time.

Now: started work in execution order, unstarted last, persist order as tiebreak — restated at the embedded run view too, which queries `MJ: Tasks` directly and so never saw the query fix.

### A configured loop said it was a broken agent step

The canvas mapped three kinds and sent everything else to `AgentTask`, whose subtitle falls back to *"No agent chosen yet"* — so a fully-configured ForEach rendered as **AGENT STEP · No agent chosen yet**. Drawing and rendering were conflated; they're now separate, with `IsAuthorableNodeType` as the compiler-enforced seam.

### Icons were invisible in dark mode

They were coloured `var(--mj-color-info)` / `var(--mj-color-primary)`. **Neither token exists** — only the `--mj-color-info-50/100/500` primitives — so `var()` resolved to nothing and the icons inherited their surroundings. A missing custom property fails silently, which is why it looked fine in light mode.

### The Workflows app's Runs tab

`WorkflowRunsResourceComponent` registered against `BaseDashboard` while its nav item is `ResourceType: "Custom"`, which the shell resolves through `BaseResourceComponent`. Never found — and the failure read as *"the Angular package likely isn't bundled"*, a packaging story for a one-word mismatch.

### Verification

- `core-entity-forms` **250** · `task-graph-editor` **142** · `task-graph` **67** · `ai-core-plus` **537** · `core-actions` **283**
- Full build **271/271**; the recursive CTE executed against a real database at every change
- Live end to end on run `AC982A1A` — actions resolve to their logs, prompts to their runs, payloads present, ordering correct, loop passes nested

### Still open, and honest about it

- **Playwright screenshots** — the persistent profile's session is expired (`JWT_EXPIRED`); it needs one interactive sign-in that I can't do headlessly.
- **Ordering is encoded in two places now.** Agreed with the reviewer that a third consumer should trigger hoisting it into a shared helper.
- **A loop pass is a leaf**, so a loop body that dispatches its own graph contributes only its run's own spend — the boundary is stated in the SQL where someone would hit it.

AN-BC:
**Reviewer — CI unit-test failure diagnosed and fixed directly on the branch (`b4b02c640`).**

The `Run unit tests` job failed on `6a653b6f1` in `ng-task-graph-editor`'s **type gate**, not its tests: `task-graph-canvas-adapter.test.ts` imported `GetNodeTypeConfig` and `TASK_GRAPH_NODE_TYPES` twice in one import block. Vitest's esbuild pipeline tolerates duplicate named imports, so your 142 tests passed locally while `tsc --noEmit -p tsconfig.spec.json` failed with TS2300 — the two gates check different things, and both were telling the truth. The four reported errors were all from these two names; the fix is a two-line deletion, pushed as `b4b02c640` (reviewer commit, trivial scope — flagging rather than asking since it's mechanical and CI re-runs on push).

Worth adding to your local Definition-of-Done for packages with a `test:types` script: run it alongside `pnpm test` — esbuild's tolerance makes this class of error invisible to the test run alone.

---

AN-BC:
## Smoke-test round: a loop pass was claiming to be something it wasn't (`4fba6ba20` · `697fbbb58`)

### A loop pass over an ACTION reported five sub-agent runs that never happened

The iteration trace recorded `promptRunID` and `agentRunID` and nothing else, so an **action-bodied** pass recorded no pointer at all. The tree then derived the pass's kind with a two-way guess — prompt run, **else Sub-Agent** — and a pass with no evidence took the else branch.

The Demo Flow Agent's ForEach calls a web search five times. It displayed five Sub-Agent runs with no cost, no timing, and nothing to open — and in Subway Lines they became five empty lanes, because a sub-agent run is exactly what gets its own lane.

Fixed at both ends: `ITaskLoopIteration.actionLogID` records what an action pass produced (the runner already surfaced it for one-shot steps; the loop dropped it), and the tree now derives kind from **evidence or not at all**:

```
prompt run present  -> Prompt
agent run present   -> Sub-Agent
action log present  -> Actions
nothing present     -> no kind
```

A guess that is right most of the time is worse than an absence: an absence looks like missing data, a guess looks like a fact. Historical runs now honestly report no kind rather than inventing a sub-agent. Verified live — five passes at 543/242/245/224/265 ms, each resolving to `MJ: Action Execution Logs`.

### A workflow prompt step described itself differently from an agent prompt step

An agent run's prompt step reads `Model: X | Vendor: Y`; a workflow's read `Prompt step · $0.0018 · 5,111 tokens`. Same work, two descriptions depending on which timeline you opened it from — which reads as two features. The tree now carries Model/Vendor from the join it already had.

### On `b4b02c6406` — thank you, and the lesson taken

That was my duplication, and the framing is exactly right: **`pnpm test` green says nothing about the spec type gate**, because vitest's esbuild strips types without checking them. I have adopted running `tsc --noEmit -p tsconfig.spec.json` on every touched package, not just the tests.

Doing so surfaced three more errors in `task-graph-editor`, all **older than this branch** — `task({ agentName: undefined })` twice (`agentName` lives inside `configuration`, not on the node) and a task literal missing `kind`/`configuration`. They were invisible until now because the gate only reaches packages a branch touches, and this branch touches that one. Fixed in `697fbbb58` rather than leaving a red gate on a PR that did not cause it.

Gate is clean across every package this branch touches that has one; `dashboards` and `explorer-core` also verified.

AN-BC:
**Reviewer — `4fba6ba20` + `697fbbb58` verified, approval stands.**

`4fba6ba20` is a genuinely good catch from smoke testing: the pass-kind derivation was a two-way guess (prompt run, *else* Sub-Agent), so an action-bodied pass — which recorded no pointer at all — was confidently labelled a sub-agent run that never existed, five times over in the demo. The fix is right at both ends: `ITaskLoopIteration.actionLogID` completes the evidence set, and the tree now derives kind from evidence or **leaves it null** — *"a guess that is right most of the time is worse than an absence, because an absence looks like missing data and a guess looks like a fact"* deserves to be quoted in a style guide. Historical runs honestly reporting no kind rather than an invented one is the correct backfill posture. Timing COALESCE across all three sources and the Model/Vendor subtitle parity both check out.

`697fbbb58` closes the loop on the type gate the right way — fixing the three pre-existing errors the gate only now reaches (`agentName` was never a `TaskGraphSpecNode` member; the empty-configuration restatement is more truthful than the `undefined` it replaces) rather than leaving a red gate on a PR that didn't cause them, and adopting gate-on-every-touched-package as practice.

Awaiting green CI on this head; no further findings.

---

AN-BC:
**Reviewer — why `697fbbb58` had no CI, why one red check appeared, and what I did about it.**

The push of `4fba6ba20`/`697fbbb58` never got a `pull_request` CI run. The reason turned out not to be a dropped webhook: `next` moved and now **conflicted** with this branch, so GitHub couldn't build the PR merge ref, and merge-ref-based CI can't fire without it. Not knowing that yet, I manually dispatched `test.yml` on the raw branch head to fill the gap — that's the red "Run unit tests" you see on `697fbbb58`. It failed in the spec anti-pattern gate on five `as never` casts in `flow-agent-form-section.component.dom.test.ts` — **a file this branch never touched**: those casts are shared history, already cleaned up on `next` by `8ee58f2d6`, which post-dates this branch's merge-base. A raw-head checkout predates the cleanup; the PR merge contains it. The red check is an artifact of my dispatch, not a regression here.

The conflicts themselves were a happy collision: `next`'s `8ee58f2d6` and this branch's `697fbbb58` independently repaired the **same pre-existing fixture type errors** in task-graph-editor's tests — `configuration: {}` here, `agentName: ''` there. I merged `origin/next` into the branch (`be50b1b58`), keeping the branch's form (an unassigned step stated by an empty configuration, per `697fbbb58`'s reasoning), and verified the spec anti-pattern + DOM-placement gates clean on the merged tree. The push makes the PR mergeable again and triggers a genuine `pull_request` run over the full head — the first CI verdict covering `4fba6ba20`/`697fbbb58`. Watching for it.

---

AN-BC:
**Reviewer — CI is green on the full head.** The genuine `pull_request` run on `be50b1b58` (branch + merged `next`) passed, giving the first complete CI verdict over `4fba6ba20`/`697fbbb58` and the merge. That closes the last open item: unit tests, spec type gates, and the anti-pattern gates all pass on exactly what would merge. The branch shows "behind" again only because `next` keeps moving (`c511618`) — no conflict this time, so no action needed unless base protection requires strict up-to-date at merge time.

State: approved at `ca4feb4ab`, approval carried through `697fbbb58`, CI green at `be50b1b58`. Ready for the maintainer.

---

PR #3725, opened 2026-08-10
fix(realtime): map Gemini's neutral voice key to speechConfig

### LINKED ISSUE #3721: Gemini realtime drops the neutral `voice` key before the wire — needs mapping to `speechConfig.voiceConfig.prebuiltVoiceConfig.voiceName`
Follow-up from the review of #3715 (issue #3530).

## The gap

Every other realtime driver consumes the driver-neutral `voice` key out of the session config bag — OpenAI (`audio.output.voice`), xAI (inherits OpenAI), ElevenLabs (`tts.voice_id`), Inworld and AssemblyAI (`output.voice`), HuggingFace (`audio.output.voice`). `GeminiRealtime` is exempted from the shared-key scrub, so `voice` correctly *survives* into `buildConnectConfig` — and then goes nowhere.

`buildConnectConfig` merges the bag with `Object.assign(config, cfg)` onto a `LiveConnectConfig`, which has **no `voice` property**. Gemini Live takes an output voice at `speechConfig.voiceConfig.prebuiltVoiceConfig.voiceName`. Grepping `packages/AI/Providers/Gemini` finds no `voiceName`, no `prebuiltVoiceConfig`, no `voiceConfig` anywhere; the only `speechConfig` reference is a passthrough of a caller-supplied object in `BuildConstraintConfig`.

Net effect: an authored voice reaches `GeminiRealtime` and is **silently discarded** — the same silent-drop bug class #3530 fixed one layer up. A Gemini voice is settable today only by hand-authoring a raw `speechConfig` into the config bag, which requires exactly the vendor knowledge the agnostic slot exists to remove.

## Why nothing catches it

- `geminiRealtime.test.ts` has zero voice assertions.
- The cross-driver loop in `realtime-convergence-drift.test.ts` asserts `GetProviderVoiceSettings` **resolution** for four drivers including Gemini. That is correct as far as it goes — resolution genuinely works — but it never touches a driver, so it reads like end-to-end Gemini coverage while proving nothing about the wire.
- `IT — realtime-deterministic` has no voice coverage either.

## Suggested fix

In `buildConnectConfig`, translate the neutral key the way the other drivers do — pull `voice` out of the bag before the `Object.assign` and emit:

```ts
speechConfig: { voiceConfig: { prebuiltVoiceConfig: { voiceName: voice } } }
```

Preserve a caller-supplied raw `speechConfig` as the more specific value (same precedence spirit as the existing scrub logic), and add it to `REALTIME_SHARED_CONFIG_KEYS` handling so it stops being assigned raw. Then add a driver-level assertion in `geminiRealtime.test.ts` — the missing test is what let this sit.

## Scope note

Also worth tracking separately: only `OpenAIRealtime`, `xAIRealtime` and `HuggingFaceRealtime` declare `SupportedVoices` (HuggingFace overrides it to return none), and the native picker's voice dropdown is gated on that list. So even once Gemini maps the key, the picker will offer no Gemini voices until `SupportedVoices` is populated for it.

SDesai-BC:
An authored realtime voice has never reached a Gemini Live session. Gemini is deliberately exempted from the shared-key scrub because `voice` *is* meaningful to it — but nothing ever translated it, so the value travelled as far as the driver's config object and no further.

`buildConnectConfig` spread the config bag straight onto a `LiveConnectConfig`:

```ts
Object.assign(config, cfg as Partial<LiveConnectConfig>);
```

`LiveConnectConfig` has **no `voice` property** — the `as` cast is what kept `tsc` quiet about it. Then `@google/genai`'s `liveConnectConfigToMldev` is a **path allowlist**: it lifts ~20 named keys by `getValueByPath` and drops everything else. No error, no warning, nothing on the wire. Gemini Live takes an output voice at `speechConfig.voiceConfig.prebuiltVoiceConfig.voiceName`, and grep found no `voiceName`, `prebuiltVoiceConfig`, or `voiceConfig` anywhere in the package. This is the same silent-drop bug class #3530 fixed one layer up, one layer down. Closes #3721.

Worth stating plainly: the diagnostics added in #3715 **cannot** catch this. `WarnOnUnmatchedProviderVoice` returns early when no `providers.*` bag is authored, and the agnostic path authors none — from the agents layer the handoff succeeded, so the loss was invisible from above.

## What it does

`voice` is now consumed *before* the merge, exactly as `disableAutoResponse` already was, so it can never spread raw — then mapped to Gemini's nested shape. **One mapping covers both topologies**: `StartSession` and `CreateClientSession` share `buildConnectConfig`, and the browser driver applies the server-built config verbatim (`geminiRealtimeClient.ts` `parseSessionConfig` just unwraps `{model, config}`). No second translation site.

Because `speechConfig` is a generation-level field it is **mask-safe**, so `BuildConstraintConfig` already forwarded it and the mint now **token-locks** the voice under `lockAdditionalFields: []` — the server's choice becomes authoritative on client-direct rather than merely suggested.

### Design decisions worth defending

**Precedence is per-key at the leaf, not per-block.** A caller-supplied raw `speechConfig` keeps whatever voice it already names — a prebuilt voice, a replicated/cloned voice, or a `multiSpeakerVoiceConfig` (that last is mutually exclusive with `voiceConfig`, and the SDK's `tLiveSpeechConfig` throws on the pair for live sessions, so injecting there would turn a discarded voice into a dead session). But a raw block that sets only `languageCode` still receives the authored voice. Block-level winner-takes-all would have re-created this very silent drop in a narrower case.

**The mapper is `private static`.** Nothing outside the class needs it and the tests reach it through `StartSession`/`CreateClientSession`, so the public surface is unchanged — which is why the changeset is honestly a `patch` rather than the `minor` a new export would force.

**Two dead per-key exemptions removed.** Once both translated keys are deleted before the loop, `if (key === 'voice')` and `if (key === 'disableAutoResponse')` are unreachable. The loop is now simply "scrub whatever survived consumption". Also fixed a comment asserting the voice reached Gemini, and reattached a doc block that was sitting above the wrong method, so TSDoc was silently dropping it.

## Evidence

**Unit — 102/102 in `@memberjunction/ai-gemini`**, 8 new tests written red-to-green one at a time. Two of them passed on first write, so rather than trust them I broke the implementation to prove they bite: rewriting the guard as `config.speechConfig = voice ? {...} : undefined` fails the field-mask test, and stubbing out `BuildConstraintConfig`'s `speechConfig` forward fails the token-lock test.

**Live Gemini API** — the mint seam is stubbed in unit tests, so I drove the real driver through the `ClassFactory` against a real key:

| Case | Mint | Session open |
|---|---|---|
| no voice | OK, `speechConfig` key **absent** | opens clean |
| `voice: 'Kore'` | OK, `voiceName: 'Kore'`, no raw `voice` leak | **opens clean, no errors** |
| `voice: 'alloy'` (an OpenAI id) | **OK — accepted** | **dies: close `1007`, `No matching speaker voice found for name: alloy and language:`**, `Fatal: true` |
| raw `speechConfig` names `Puck` | OK, `Puck` wins | warning fired |

That confirms `speechConfig` really is mask-safe against the live mint — previously an inference from the type shape.

**Integration — 72/78.** Rather than assume the 6 failures were pre-existing I compared against the CI Integration Tier run on `next` at this branch's exact base commit `775b61667` (59/78 there; CI has no live-model keys). Four fail identically on that baseline: `agent-memory-guards`, `entity-writes`, `shipped-agents-live`, `subscription-isolation`. The other two are environmental — `templates` blew the 300 s run budget at 1551 s locally and passes in CI, and `workflow-demo-agents` asserts task-graph metadata (got 3 steps, wants 2) against a stale scratch database. `IT — realtime-deterministic` passed, and "gemini" appears nowhere within 25 lines of any failure.

**Server side** — 174/174 packages built across the dependency closure, class manifest unchanged at 989 classes, MJAPI boots to `Ready`. `Vertex` depends on `ai-gemini` but only reuses `GeminiLLM`, so nothing subclasses the changed driver.

## Hazard

This converts a silent no-op into a live value, so a cross-vendor voice id that previously did nothing on Gemini is now sent as `voiceName` — and per the table above, that **kills the session at socket open**, past the mint, rather than degrading. Pin the model alongside the voice, or author per-vendor ids under `realtime.voice.providers.<key>`.

## Merge ordering

Independent of #3715 — the scrub loop and the `voice` exemption both already exist on `next`, so this needs no rebase on it. The one interaction is the guide: #3715 rewrites the same two prose lines this touches, so whichever lands second takes a trivial conflict. Both PRs are mine to resolve.

## Not in scope

Only `OpenAIRealtime`, `xAIRealtime` and `HuggingFaceRealtime` declare `SupportedVoices`, and the picker's dropdown is gated on that list — so Gemini voices stay reachable through agent metadata and programmatic hosts, not the picker. Populating that list needs a source outside the SDK: the `@google/genai` typings do not enumerate Gemini's prebuilt voice names, so any list here would be an unverifiable hardcode. Tracked in #3721's own scope note.


packages/AI/Providers/Gemini/src/geminiRealtime.ts
```diff
@@ -313,6 +309,55 @@ export class GeminiRealtime extends BaseRealtimeModel {
...
+     */
+    private static buildSpeechConfig(voice: string | undefined, existing: SpeechConfig | undefined): { SpeechConfig?: SpeechConfig; DroppedVoice?: string } {
+        if (!voice) {
+            return { SpeechConfig: existing };
+        }
+        // Does the raw block already decide the voice? Any of three shapes counts: a named prebuilt
+        // voice, a cloned/replicated voice, or a multi-speaker config (mutually exclusive with
+        // `voiceConfig` — the SDK rejects the pair on a live session, so injecting there would turn a
+        // discarded voice into a dead session).
+        const rawAlreadyDecidesVoice =
```
rkihm-BC, line 340: **The guard is right; the stated reason for it isn't.**

The comment (and the PR body) say injecting `voiceConfig` here "would turn a discarded voice into a dead session". But `tLiveSpeechConfig` throws on `multiSpeakerVoiceConfig` **unconditionally** for live sessions, regardless of whether `voiceConfig` is present:

```js
function tLiveSpeechConfig(speechConfig) {
    if ('multiSpeakerVoiceConfig' in speechConfig) {
        throw new Error('multiSpeakerVoiceConfig is not supported in the live API.');
    }
```

And it's reached on **both** topologies: server-bridged via `liveConnectConfigToMldev` at connect, and client-direct via `liveConnectConstraintsToMldev` → `liveConnectConfigToMldev` at the **token mint** (because `BuildConstraintConfig` forwards `speechConfig`). So the session is already dead either way — this PR neither causes nor worsens it.

Not asking to change the behavior: not fabricating a shape the SDK rejects is the correct call. Just re-word the justification, because as written it will send the next reader looking for a failure mode that doesn't depend on this branch.

While you're here — worth considering a warning that `multiSpeakerVoiceConfig` is unsupported on Live at all. Today the driver silently forwards a config guaranteed to throw, which is a small hole in the "no silent surprises" standard this PR sets everywhere else. (The throw is loud when it lands, so this is polish, not a defect.)

packages/AI/Providers/Gemini/src/geminiRealtime.ts
```diff
@@ -335,22 +381,30 @@ export class GeminiRealtime extends BaseRealtimeModel {
...
+                console.warn(`[GeminiRealtime] Ignored the session config bag's \`voice\` because it is not a string (got ${typeof rawVoice}) — expected a Gemini prebuilt voice name.`);
+            }
             // Scrub the MJ-side keys shared across the realtime driver family (OpenAI-protocol
             // feature knobs + transport settings). They are NOT Gemini config keys — a co-agent
             // config carrying e.g. `effortLevel` or `mcpTools` must be SAFE on a Gemini session,
             // not spread raw into the Live SDK where strictness varies by version. Scrubbed keys
             // are diag-logged so config typos / cross-provider keys stop being silent.
+            // The two keys Gemini translates natively (`disableAutoResponse`, `voice`) were already
+            // consumed and deleted above, so the loop only ever sees keys with no Gemini mapping —
+            // no per-key exemption needed here.
```
rkihm-BC, line 404: **The class of bug is narrowed here, not closed — and there's a concrete instance still reachable through the schema this repo ships.**

This comment is accurate but reads as though the scrub is now complete. It isn't: `REALTIME_SHARED_CONFIG_KEYS` is an allowlist of *MJ-shared* keys, not of *non-Gemini* keys. Anything outside that list still rides `Object.assign(config, cfg as Partial<LiveConnectConfig>)` and gets dropped by the SDK allowlist exactly as `voice` did.

The live instance: **`realtime.voice.providers.gemini.voiceId`**. #3530 deprecated `voiceId` but deliberately kept it in the agent-type `ConfigSchema` so existing configs still validate — so it is authorable today, it rides the DriverClass-matched provider bag onto a Gemini session, it is not in the scrub list, it is not a `LiveConnectConfig` field, and it vanishes with no error. Same mechanism as #3721, one key over.

I checked the other bag contributors and they're clean — `GetSessionTuningSettings` and `GetModelCatalogSessionSettings` both emit closed key sets (`effortLevel`, `parallelToolCalls`, `mcpTools`, `inputTranscriptionModel`, `turnDetection`), all covered by the scrub. So the realistic exposure is the matched provider bag plus a host-supplied runtime `Config`, not arbitrary drift.

Scope here is right and I'm not asking you to fix it in this PR — but it deserves a follow-up. The structural version is inverting the scrub to warn on any key that isn't a known `LiveConnectConfig` field, which closes the class rather than the instance; the cheap version is just handling `voiceId` alongside `voice`.

packages/AI/Providers/Gemini/src/geminiRealtime.ts
```diff
@@ -335,22 +381,30 @@ export class GeminiRealtime extends BaseRealtimeModel {
             const cfg = { ...(params.Config as Record<string, unknown>) };
             const disableAutoResponse = cfg.disableAutoResponse === true;
             delete cfg.disableAutoResponse;
+            // Consume the driver-NEUTRAL `voice` key the same way: it is not a Gemini config field
+            // (LiveConnectConfig has no `voice`), so spreading it raw sends nothing — the SDK's
+            // config converter is a path allowlist and drops unknown keys silently. Gemini takes an
+            // output voice at speechConfig.voiceConfig.prebuiltVoiceConfig.voiceName instead.
+            const rawVoice = cfg.voice;
+            delete cfg.voice;
+            const voice = typeof rawVoice === 'string' ? rawVoice.trim() : undefined;
```
rkihm-BC, line 390: **A whitespace-only voice vanishes silently — the one case that slips past the principle this PR sets.**

`'   '` trims to `''`, `if (!voice)` short-circuits in `buildSpeechConfig`, and nothing is logged. The changeset states the standard plainly: *"A discarded voice and a non-string `voice` are both now logged rather than vanishing."* Blank is a third case, and it's arguably the **most** likely of the three in practice — a metadata copy-paste or a cleared-but-not-removed field, rather than a type error.

The behavior (treat as absent) is right; ElevenLabs' `ResolveVoiceID` does the same. It's the silence that's inconsistent. Folding it into the existing non-string branch would cover it:

```ts
if (rawVoice !== undefined && (typeof rawVoice !== 'string' || rawVoice.trim().length === 0)) {
    // ...warn: non-string, or blank
}
```

Test 6 currently pins the absence but asserts nothing about logging, so it would keep passing either way — worth an assertion there too if you add the warning.

packages/AI/Providers/Gemini/src/geminiRealtime.ts
```diff
@@ -313,6 +309,55 @@ export class GeminiRealtime extends BaseRealtimeModel {
...
+     * `languageCode` still receives the authored voice. Block-level winner-takes-all would re-create
+     * the same silent drop in a narrower case.
+     *
+     * @param voice The neutral voice id, already trimmed; `undefined` when none is authored.
+     * @param existing Any raw `speechConfig` the caller merged in via the config bag.
+     * @returns The speech config to apply (absent when there is nothing to say), plus the voice that
+     *          was discarded because the raw block already decided it — the caller logs it, keeping
+     *          this a pure calculation.
+     */
+    private static buildSpeechConfig(voice: string | undefined, existing: SpeechConfig | undefined): { SpeechConfig?: SpeechConfig; DroppedVoice?: string } {
```
rkihm-BC, line 332: `existing` is typed `SpeechConfig | undefined`, but it reaches here from `params.Config` — an open JSON bag — through the `as Partial<LiveConnectConfig>` cast, so `tsc` can't actually vouch for it. `speechConfig: "Kore"` spreads the string:

```js
{ ...'Kore', voiceConfig: {...} }  // → { 0:'K', 1:'o', 2:'r', 3:'e', voiceConfig: {…} }
```

No crash, but garbage on the wire. This is asymmetric with the care taken one function down, where a non-string `voice` gets an explicit type check and a warning — and it's a neat illustration of the point the PR body makes about that cast being what hid #3721 in the first place. An `isPlainObject`-style guard on `existing` (warn + ignore) would make the two consistent and cost almost nothing.

packages/AI/Providers/Gemini/src/__tests__/geminiRealtime.test.ts
```diff
@@ -602,3 +602,123 @@ describe('C6: cross-provider config-bag safety (shared-key scrubbing)', () => {
...
+        await driver.StartSession(makeParams({ Config: { voice: 'Kore' } }));
+        const config = driver.LastConnectArgs!.Config as LiveConnectConfig & Record<string, unknown>;
+        expect(config.speechConfig?.voiceConfig?.prebuiltVoiceConfig?.voiceName).toBe('Kore');
+        // The neutral key must be CONSUMED, not spread raw: LiveConnectConfig has no `voice`
+        // property, and the SDK's allowlist config converter drops unknown keys silently — which
+        // is exactly how an authored voice went missing before this mapping existed.
+        expect(config.voice).toBeUndefined();
+    });
+
+    it('omits speechConfig ENTIRELY when no voice is authored (an undefined key still enters the token field mask)', async () => {
```
rkihm-BC, line 618: **Keep the test, fix the comment — the hazard it describes doesn't exist.**

The parenthetical and the inline comment say a present-but-`undefined` `speechConfig` "would widen the mask and 400 the token". Both paths guard against exactly that:

- `BuildConstraintConfig` copies with `if (config.speechConfig)` — a truthiness check, so `undefined` is never forwarded
- the SDK converter guards with `if (parentObject !== undefined && fromSpeechConfig != null)` before calling `tLiveSpeechConfig`

So an undefined value can't reach the field mask through either route. The test is still worth having — it pins the `if (speech.SpeechConfig)` guard in `buildConnectConfig` and keeps the connect config clean — but the justification invents a live 400 risk, and in a PR that leans this heavily on defended rationale that's the kind of thing that gets cargo-culted into the next driver.

packages/AI/Providers/Gemini/src/__tests__/geminiRealtime.test.ts
```diff
@@ -602,3 +602,123 @@ describe('C6: cross-provider config-bag safety (shared-key scrubbing)', () => {
...
+            // `voice` IS a Gemini-meaningful key, so the scrub message ("OpenAI-protocol/transport
+            // keys ... do not apply to Gemini Live") would be a lie about a plain config typo.
+            expect(warn).toHaveBeenCalledWith(expect.stringContaining('not a string'));
+            expect(warn).not.toHaveBeenCalledWith(expect.stringContaining('Scrubbed non-Gemini config key'));
+        } finally {
+            warn.mockRestore();
+        }
+    });
+
+    it('client-direct: the voice is token-LOCKED and carried to the browser (both halves, one mapping)', async () => {
```
rkihm-BC, line 709: This is the test that earns the "one mapping, both topologies" claim — asserting the locked constraint *and* the browser-facing `SessionConfig` in one place is exactly right, and I verified the other half independently (`geminiRealtimeClient.parseSessionConfig` unwraps `{model, config}` verbatim and the client driver has no voice handling at all).

Two optional additions, both cheap:

1. **`voice` + `disableAutoResponse` together.** Both are now consumed pre-merge and neither test exercises the pair. Low risk, but they're the only two keys sharing that mechanism, so a single combined case would pin it.
2. **A `RealtimeClient`-layer assertion.** The verbatim-passthrough contract is what lets this PR skip a second mapping site — it's load-bearing but only pinned here, on the server side. One assertion in `gemini-realtime-client.test.ts` that a server-supplied `speechConfig` survives to `live.connect` would stop a future refactor of the client driver from quietly breaking the guarantee this PR relies on.

.changeset/gemini-realtime-voice-mapping.md
```diff
@@ -0,0 +1,13 @@
...
+
+**Behavior change worth watching**: this converts a silent no-op into a live value, so a cross-vendor voice id that previously did nothing on Gemini is now sent as `voiceName`. Verified against the live Gemini API: an invalid name passes the ephemeral-token mint without complaint and then **kills the session when the socket opens** — close code `1007`, `No matching speaker voice found for name: alloy and language:`, surfaced as a `Fatal` session error. So the failure is late and total, not a silent degrade — the same shape as the ElevenLabs note in #3530. Pin the model alongside the voice, or author per-vendor ids under `realtime.voice.providers.<key>`.
```
rkihm-BC, line 11: Good disclosure — the live-API detail (close `1007`, the exact server message) is what makes this actionable rather than a vague caveat, and putting it in the changeset rather than only the PR body is the right call.

One thing worth weighing. The most exposed user is someone who authored `default.voice: 'alloy'` back when only OpenAI could consume it, whose model preference later resolves to Gemini. Because #3530 and this land in the same release there's no intermediate state — good. And the picker can't produce it (dropdown gated on `SupportedVoices`, cleared on model change), so exposure is agent metadata plus bridge hosts, which #3530's changeset item 3 already covers.

Still, a dead socket with a raw provider message is a harsh first symptom for what is a config typo. Worth considering whether `GeminiRealtime` should catch the `1007` and re-emit naming the config key that caused it (`realtime.voice.default.voice`) — not a silent degrade, just an error that points at the fix. The counter-argument — that a driver shouldn't second-guess the provider — is legitimate, so this is your call; I'd just rather it be a decision than an omission.

Minor, on line 9: *"A discarded voice and a non-string `voice` are both now logged rather than vanishing"* — see the note on the trim in `buildConnectConfig`; a blank voice is a third case that still vanishes silently.

rkihm-BC (commented):
Strong PR — noticeably tighter than #3715: narrow scope, correct fix, good tests, live-API verification, and it closes the documentation loop from the prior PR. Approve from me. Three findings below, all small, and two of them are wrong *rationale* rather than wrong *code*.

**I verified the load-bearing claims rather than taking them on trust:**

- **The allowlist diagnosis is exactly right.** `liveConnectConfigToMldev` really is a path allowlist — 44 `getValueByPath` calls over named keys, `speechConfig` among them (lifted to `setup.generationConfig.speechConfig` via `tLiveSpeechConfig`), and no `voice`. Dropped with no error, nothing on the wire. Confirmed.
- **Every type claim in "design decisions worth defending" checks out** against @google/genai 1.52.0: `VoiceConfig.replicatedVoiceConfig`, `VoiceConfig.prebuiltVoiceConfig`, `SpeechConfig.multiSpeakerVoiceConfig`, `SpeechConfig.languageCode` all exist, and the SDK's own doc comment says `multiSpeakerVoiceConfig` is *"mutually exclusive with `voice_config`"*.
- **"One mapping covers both topologies" is real.** `packages/AI/RealtimeClient/src/drivers/geminiRealtimeClient.ts` has **zero** `voice`/`speechConfig` handling and `parseSessionConfig` unwraps `{model, config}` verbatim. No second translation site.
- **The token-lock is correctly framed as a consequence**, not a new feature — `BuildConstraintConfig` already forwarded `speechConfig` and `lockAdditionalFields: []` is pre-existing. Independently corroborated by the SDK putting `speechConfig` under `generationConfig`, which is what makes it mask-safe.
- **Vertex claim holds** — `vertexLLM.ts` imports only `GeminiLLM`; nothing subclasses `GeminiRealtime`.
- **Removing the two dead exemptions is safe** — I checked the pre-existing `C6` scrub tests don't go vacuous (they hardcode a key list that never included `voice`).
- **No stale tests anywhere in the repo** — `RealtimeClient`'s two Gemini client tests never assert on voice/`speechConfig`, and the agents-layer `GeminiRealtime` references are `GetProviderVoiceSettings` resolution only.
- **Evidence reproduces exactly**: 102/102 package tests, 46/46 in the realtime file, `tsc --noEmit` clean.

Breaking two tests deliberately to prove they bite is the right instinct, and worth saying out loud — tests that pass on first write usually deserve exactly that suspicion.

**On the hazard**: well disclosed across changeset, guide, and PR body, with real evidence. Loud failure over silent degrade is defensible. One thought inline on the changeset about whether the *first symptom* could be friendlier without becoming a silent degrade.

<sub>Scope of what I checked: SDK converter and type definitions read directly; all four claim-bearing code paths traced (`buildConnectConfig`, `BuildConstraintConfig`, `CreateClientSession`, `geminiRealtimeClient`); config-bag contributors audited (`GetSessionTuningSettings` / `GetModelCatalogSessionSettings` both emit closed key sets fully covered by the scrub list); full package suite and typecheck run. I did **not** run a workspace build or the integration tier, and I could not reproduce the live-API table without a Gemini key — that one I'm taking on your word.</sub>

packages/AI/Providers/Gemini/src/geminiRealtime.ts
SDesai-BC: Confirmed, and you're right — I traced it rather than taking it: `tLiveSpeechConfig` throws on `multiSpeakerVoiceConfig` unconditionally, and `liveConnectConstraintsToMldev` really does call `liveConnectConfigToMldev`, so the throw lands on the mint path too. My comment described a failure mode that doesn't depend on this branch, which is exactly the kind of thing that sends the next reader chasing nothing.

Reworded to say what's true: the multi-speaker session is doomed either way, and the guard simply declines to fabricate a shape the SDK rejects — it neither causes nor avoids the throw.

I also took the polish. The driver now warns when the bag carries `speechConfig.multiSpeakerVoiceConfig`, naming the key and that Live rejects it on both topologies. You were right that silently forwarding a guaranteed-throw config was a hole in the standard the rest of the PR sets — the throw being loud when it lands doesn't help someone who never learns which key caused it.

`2cb04fcbb8`

packages/AI/Providers/Gemini/src/geminiRealtime.ts
SDesai-BC: Confirmed on both halves. The comment did read as though the scrub were complete, and it isn't — the list is an allowlist of MJ-*shared* keys, not of non-Gemini keys, so anything outside it still rides the `Object.assign` and dies in the SDK's path allowlist exactly as `voice` did.

`providers.gemini.voiceId` checks out as a live instance: #3530 deprecated it but deliberately kept it in the agent-type `ConfigSchema` so existing configs still validate, so it is authorable today, it rides the DriverClass-matched bag onto a Gemini session, it isn't in the scrub list, and it isn't a `LiveConnectConfig` field.

Reworded the comment to say the class is narrowed rather than closed, name `voiceId` as the remaining instance, and record that the structural fix is inverting the check to warn on any key that isn't a known `LiveConnectConfig` field.

Taking your scoping — not fixed here. I deliberately didn't reach for the cheap version either: adding `voiceId` to `REALTIME_SHARED_CONFIG_KEYS` touches a shared Core constant with cross-driver blast radius, and handling a deprecated key in the driver perpetuates it. Follow-up issue to come for the inversion.

`2cb04fcbb8`

packages/AI/Providers/Gemini/src/geminiRealtime.ts
SDesai-BC: Fixed — and you picked the right one to call out. Blank genuinely is the likeliest of the three: a cleared-but-not-removed metadata field beats a type error by a wide margin, and it was the one case that contradicted the standard the changeset states.

Folded into the existing branch as you suggested, with the message distinguishing the two reasons so the warning still diagnoses rather than just complains:

```ts
const voice = typeof rawVoice === 'string' && rawVoice.trim().length > 0 ? rawVoice.trim() : undefined;
if (rawVoice !== undefined && voice === undefined) {
    const reason = typeof rawVoice === 'string' ? 'blank' : `not a string (got ${typeof rawVoice})`;
```

You were also right that test 6 would have kept passing either way — it pinned the absence and asserted nothing about logging. Added the logging assertion, and it went red before the fix, so it has teeth. Changeset line 9 now enumerates all five cases that log rather than implying two.

`2cb04fcbb8`

packages/AI/Providers/Gemini/src/geminiRealtime.ts
SDesai-BC: This was the best catch in the review — fixed. I reproduced it before touching anything:

```
$ node -e "const existing='Kore'; console.log(JSON.stringify({...existing, voiceConfig:{prebuiltVoiceConfig:{voiceName:'Kore'}}}))"
{"0":"K","1":"o","2":"r","3":"e","voiceConfig":{"prebuiltVoiceConfig":{"voiceName":"Kore"}}}
```

The red test failed with `expected 'P' to be undefined`, which is that string-spread landing in the connect config.

Now narrowed by an `isPlainObject` guard that rejects strings, arrays and `null`. Three details worth noting:

- when there's no voice to map, the malformed value is **deleted** from the config rather than merely ignored — otherwise the raw string survives the merge and reaches the wire anyway, which was the actual bug
- the warning captures the offending value *before* the mapping replaces it. My first cut read `config.speechConfig` afterwards and would have reported the type that replaced it — wrong diagnostic, right intent
- and yes, the asymmetry you point at was the tell. The neighbouring `voice` got a type check because I was thinking about that key; `existing` didn't because its declared type looked like it had already answered the question. Same cast, same blind spot as #3721.

`2cb04fcbb8`

packages/AI/Providers/Gemini/src/__tests__/geminiRealtime.test.ts
SDesai-BC: Confirmed — the justification was invented and I've removed it. Both guards are exactly as you describe: `BuildConstraintConfig` copies under `if (config.speechConfig)`, and the converter checks `fromSpeechConfig != null` before `tLiveSpeechConfig`. An undefined value cannot reach the field mask by either route.

The comment now states what the test actually pins — the `if (speech.SpeechConfig)` guard, and a clean connect config — and says explicitly that undefined is *not* dangerous, so nobody reasons from it. Kept the test; it still fails if the guard is rewritten as `speechConfig: voice ? {...} : undefined`, which is the mistake it exists for.

Your framing is the part I'll carry forward: in a PR leaning this hard on defended rationale, a wrong reason is more contagious than a missing one, because the next driver copies the reasoning and not just the code.

`2cb04fcbb8`

packages/AI/Providers/Gemini/src/__tests__/geminiRealtime.test.ts
SDesai-BC: Added both.

**1. `voice` + `disableAutoResponse` together** — pins the pair that shares the pre-merge consumption mechanism, asserting both translations landed and neither key was forwarded raw. Cheap, and it's the case a future reorder would break asymmetrically.

**2. The `RealtimeClient`-layer assertion** — this was the more valuable of the two and I'd underweighted it. The verbatim-passthrough contract is precisely what licenses skipping a second mapping site, and it was pinned only on the server side; a refactor of the client driver could have quietly voided the guarantee this PR depends on with every test still green. Now asserted in `gemini-realtime-client.test.ts` that a server-supplied `speechConfig` reaches `live.connect` untouched. It uses exact `toEqual`, so it bites on an added or dropped key either way.

While there I also replaced a brittle `toHaveBeenCalledTimes(2)` I'd written in the first round: the new multi-speaker warning made it fail while saying nothing about whether the discard was reported. Content assertions instead.

`RealtimeClient` is 404/404. The change is test-only and `src/__tests__` is excluded from that package's build, so I did **not** add it to the changeset — the published artifact is byte-identical. Flagging it as a considered deviation from the "every modified package gets a patch" convention rather than an oversight.

`2cb04fcbb8`

.changeset/gemini-realtime-voice-mapping.md
SDesai-BC: Both addressed, and I've made the `1007` question a decision rather than an omission — recorded in the changeset so it's reviewable.

**Decision: no, the driver won't rewrite that error.** The reasoning that tipped it isn't the "don't second-guess the provider" argument you anticipated — it's that Google's message already quotes the offending value (`No matching speaker voice found for name: alloy and language:`). So the only thing a translation adds is the MJ key name, and buying that means string-matching Google's wording *and* carrying the sent voice on the session purely to annotate a close frame. Coupling to a provider's error string to add one noun is a bad trade, and it degrades silently to no annotation the day they reword it. If the message were opaque I'd have gone the other way.

Your exposure analysis matches mine, including the part that makes it tolerable: #3530 and this land in the same release, so there's no intermediate state where a `default.voice` authored for OpenAI can reach Gemini through a version skew.

**Line 9**: fixed. It now enumerates all five cases that log — discarded voice, blank voice, non-string `voice`, non-object `speechConfig`, and a `multiSpeakerVoiceConfig` Live can't accept — rather than claiming two and leaving a third silent.

`2cb04fcbb8`

SDesai-BC:
Thanks — this was a genuinely useful review, and the split you named up front ("two of them are wrong *rationale* rather than wrong *code*") turned out to be the most valuable part of it. All seven findings verified against the SDK rather than accepted on trust, all seven addressed in `2cb04fcbb8`. Replies inline; summary here.

**Two real defects, both fixed red-to-green:**

| Finding | What it was |
|---|---|
| Non-object `speechConfig` | Reproduced: `{...'Kore'}` → `{"0":"K","1":"o","2":"r","3":"e"}` on the wire. Red test failed with `expected 'P' to be undefined`. Now guarded, **deleted** rather than merely ignored when there's no voice to map (otherwise the raw value survives the merge anyway), and warned with the type captured *before* the mapping replaces it. |
| Blank `voice` | Vanished silently, contradicting the standard the changeset states. Folded into the malformed branch, distinguishing "blank" from "not a string". Your point that test 6 would have passed either way was correct — the logging assertion is added and went red first. |

**Three wrong rationales, all three right:**

- `tLiveSpeechConfig` throws unconditionally, and `liveConnectConstraintsToMldev` → `liveConnectConfigToMldev` confirms the mint path too. Reworded: the guard neither causes nor avoids that throw.
- The scrub comment implied the class was closed. Reworded to "narrowed, not closed", naming `providers.gemini.voiceId` as the live instance and the inversion as the structural fix.
- The field-mask 400 doesn't exist — both paths guard. Comment now says what the test pins and explicitly that undefined is *not* dangerous, so it can't be cargo-culted.

**Both optional tests added**, and the `RealtimeClient` one was the more valuable of the two — I'd underweighted it. The verbatim-passthrough contract is what licenses skipping a second mapping site, and it was pinned only server-side.

**Also took the multi-speaker polish**: the driver now warns on `speechConfig.multiSpeakerVoiceConfig` rather than silently forwarding a config the SDK is guaranteed to reject.

**On `1007` — a decision, not an omission**: no rewrite. Not for the "don't second-guess the provider" reason, but because Google's message already quotes the offending value, so a translation buys only the MJ key name — at the price of string-matching their wording and carrying the sent voice on the session to annotate a close frame. Had the message been opaque I'd have gone the other way.

Two notes back to you:

1. I found and fixed a bug in my own fix while writing it — the malformed-`speechConfig` warning originally read `config.speechConfig` *after* the mapping replaced it, so it would have reported the wrong type. Right intent, wrong diagnostic.
2. I replaced a brittle `toHaveBeenCalledTimes(2)` from the first round; the new warning broke it while proving nothing about whether the discard was reported. Content assertions now.

`ai-gemini` 105/105, `RealtimeClient` 404/404, both build clean, and I re-ran the live-API mint check since these paths changed — unchanged across all four cases. The `RealtimeClient` change is test-only and `src/__tests__` is excluded from that package's build, so I deliberately did **not** add it to the changeset: the published artifact is byte-identical. Flagging that as a considered deviation from "every modified package gets a patch" rather than an oversight.

Follow-up issue for the `voiceId` / invert-the-scrub item still to come.

PR #3728, opened 2026-08-11
feat(conversations): add a composerActions slot for host controls beside Plan Mode

cmbrock-BC:
## Summary

The composer's action strip was closed to hosts. `mj-message-input-box` takes booleans (`enablePlanMode`, `enableAttachments`, `enableRealtime`) and emits events, and there is no `ng-content` anywhere in the `mj-conversation-chat-area` -> `mj-message-input` -> `mj-ai-composer` -> `mj-message-input-box` chain. A host that wanted one more control next to Plan Mode had two options and both were bad: park it somewhere else on the page, or style-pierce into MJ's internals and break on the next release.

`composerActions` is the composer's equivalent of the existing `headerActions` slot. Additive, rendered after the stock controls so they keep fixed positions, and projected as a template the host owns.

```html
<mj-conversation-chat-area ...>
  <ng-template mjChatSlot="composerActions" let-disabled>
    <button class="attach-button-icon" [disabled]="disabled" (click)="openSkills()">
      <i class="fa-solid fa-wand-magic-sparkles"></i>
    </button>
  </ng-template>
</mj-conversation-chat-area>
```

## Changes

* `composerActions` added to the `MJChatSlotName` union, with `IMJChatComposerActionsContext` in `slot-interfaces.ts` (the union's own comment requires the two stay in sync).
* The template is sourced from the slot in `conversation-chat-area` and proxied through `mj-message-input` and `mj-ai-composer`, neither of which interprets it. `mj-message-input-box` renders it last in the strip.
* `allowComposerActions` on the chat area, defaulting true, joining the existing `allowPlanMode` / `allowAttachments` / `allowRealtime` caps. The slot is already opt-in, so this is for the case the others exist for: an embedded or read-only surface switching a capability off centrally without unwinding the template that projects it.
* README slot list updated from 7 to 8 with a worked example.

Two details that are easy to get wrong and are covered by tests:

1. **Both composer instances are wired.** The chat area renders `mj-message-input` twice, the normal composer and the empty-state one. Wiring only the first hides a projected control on a brand new conversation, which is exactly where a host's entry-point control gets reached for.
2. **The strip's visibility gate includes `actionsTemplate`.** A host projecting a control with every stock control switched off would otherwise render nothing at all, and nothing would fail.

No `Before*`/`After*` pair here on purpose. The projected content is host owned, so the host's own control emits its own events. MJ cannot meaningfully cancel a click it does not own, and an event MJ never raises would be a contract it cannot honour.

Additive and opt in throughout. No behaviour change for any existing consumer.

## Testing

* `ng-conversations` 1085 passing, `ng-composer` 45 passing. 13 of those are new, split by ownership: placement and the visibility gate in `ng-composer` which owns that markup, slot name / context shape / cap gating / threading in `ng-conversations` which owns the seam.
* A TestBed spec pins the change detection behaviour. The binding reads a `@ContentChildren` query, which resolves to null before `ngAfterContentInit`, so the shape is a plausible NG0100. It does not throw. Worth having as a spec rather than an assumption, because NG0100 is dev only and a production build cannot see it.
* Deterministic integration tier run against a fresh 6.1.0-edge.1 database. 60/78, and none of the failures are reachable from this change: the suite never references `ng-conversations` or `ng-composer` and neither is a dependency of it. The failures are missing LLM credentials, missing `MJ_API_KEY`, and fresh database seeding gaps.
* `mj standards check` passes, including `ui-layers`.

## Live testing

Tested this against the Betty repo before opening the PR, because Betty is the reason it exists. Betty has a Skills menu that writes a skill chip into the composer, and it has been sitting in a row underneath the chat area rather than in the composer where it belongs. The comment in that file says why: no composer slot existed, verified against 6.1.0-edge.1 on 2026-08-10.

I built both packages off this branch and dropped the dists into Betty's node_modules, then moved the Skills menu into the new slot. Versions matched exactly at 6.1.0-edge.1, so nothing was being forward ported onto older code.

It works. The button renders in the strip next to Plan Mode. Betty styles it to match the stock controls itself, 36x36 and transparent, which is the right split since MJ does not style projected content.

Two things came out of the click testing that are worth passing on to anyone else who projects into this slot:

1. A popover anchored to a projected control needs `position: fixed`, not absolute. An ancestor in the chat area chrome sets `overflow: hidden`, and no z-index escapes overflow clipping. Raising it does nothing.
2. `.attach-buttons` is `position: absolute` with no z-index and no transform, so it creates no stacking context and a fixed popover genuinely escapes. That is why this needed no MJ side change. If that element ever gains a transform, anything anchored to a projected control gets trapped and would need a portal.

Betty's side of this is not shipping yet. It stays local until this lands on `next` and gets backported to lts/5.


cmbrock-BC:
Sorry, Claude opened this PR itself I like to open them manually to ensure it is done right so I am closing this to open one manually.

PR #3729, opened 2026-08-11
fix(rsu): keep the post-restart hand-off until the work is actually done

MS-BC:
## The failure

Apply All builds the tables, writes a pending-work file describing what still has to happen, **restarts MJAPI**, and expects the returning process to create the entity maps that make those tables usable.

`ReadAndClearPendingWork` deleted that file the instant it parsed it:

```js
const data = JSON.parse(readFileSync(filePath, 'utf-8')) as RSUPendingWork;
results.push(data);
unlinkSync(filePath);        // ← before a single entity map is created
```

…and the consumer runs its work inside a `catch` that swallows. So a throw on object 200 of 354, an OOM, or a second restart left the workspace **healthy and idle, every table built, not one entity map created**, and no instruction anywhere to finish them. Nothing retried, because nothing was left to retry from. Nothing reported it.

A connector in that state reads as connected and syncs nothing — it is not discoverable from the UI, because "no maps" looks identical to "not set up yet". And the odds of hitting it scale with the number of objects selected, which is exactly backwards for the case where it costs the most.

Observed on a customer workspace applying 354 Salesforce objects: schema `IN SYNC`, `0 synced tables (0 active)`.

## The change

Split reading from clearing, and let the caller own the file's lifetime:

| | |
|---|---|
| `ReadPendingWork()` | returns each item with its path, deletes nothing |
| `ClearPendingWork(path)` | delete, once the work has actually completed |
| `RewritePendingWork(path, data)` | write back what remains — temp file + rename, so a crash mid-write can't leave a half-written instruction |

`ReadAndClearPendingWork()` is kept, delegating to these, and marked `@deprecated`.

`RSUPendingWork` gains an optional `Attempts` counter.

The server's post-restart consumer now:

- records which objects it mapped
- clears the file **only when the whole item ran**
- otherwise writes back the remainder with `Attempts` incremented, so the next boot **resumes at object 200** rather than starting over or giving up

The accounting sits in a `finally`, which also covers the guard clauses' `continue` paths (no system user yet, connection not found) — those skip the work just as surely as a throw does, and previously destroyed the instruction just as completely.

After **3 attempts** the file is dropped rather than retried forever — but loudly, naming the objects that were never mapped. Resuming forever is its own failure mode: an item that can never succeed would otherwise be retried on every restart for the life of the workspace.

## Scope

No behaviour change on the happy path — an item that completes is cleared exactly as before. The new behaviour only exists on paths that previously lost the work silently.

## Verification

`SchemaEngine` and `MJServer` both typecheck clean (`tsc --noEmit`) with `MJServer` resolved against the rebuilt `schema-engine` types. The one unrelated pre-existing error on `next` (`RealtimeClientSessionResolver.ts:1692`) is untouched.


MS-BC:
Closing — superseded by #3291, which landed today.

This PR's thesis was that `ReadAndClearPendingWork` deleted the hand-off the instant it parsed it, so a throw partway through consumption left a workspace with every table built, no entity maps, and nothing anywhere instructing a retry. That is fixed on `next`, by a stronger mechanism than the one proposed here.

Instead of splitting the *file*'s lifetime, #3291 moved the queue into `MJ: RSU Pending Work` rows:

- `ReadPendingWork(systemUser, staleMinutes)` is the live API; `ReadAndClearPendingWork()` remains only as a deprecated, non-functional shim that warns and returns an empty array, so the delete-on-read path is gone rather than corrected.
- Rows are marked `Completed` only *after* their work actually succeeds — `packages/MJServer/src/index.ts` states it directly: *"marks each row Completed only AFTER its work actually succeeded — so a crash mid-processing leaves the row Pending and re-processable on the next boot."*
- Failures land in `FailPendingWork(...)` with the reason recorded, which closes the swallowing-`catch` half of the report; the file version had nowhere to put a failure.
- Rows older than a threshold are surfaced as stale, so abandoned work is visible instead of merely re-processable.

The customer symptom described here — schema `IN SYNC`, `0 synced tables (0 active)` after 354 Salesforce objects — is addressed by that path, and a crash now leaves visible, resumable work rather than a healthy-looking idle workspace.

**One idea from here that did not carry over**, recorded rather than lost: `RewritePendingWork(path, data)` would have trimmed already-completed objects from a partially-processed item, so a retry didn't redo them. The row-based version retries the item whole. That is wasted work rather than lost work — the operations are idempotent — so it is an optimization, not a defect, and worth doing only if a large partial re-process shows up as a real cost.

PR #3730, opened 2026-08-11
ci: add @gemini-cli bot and move release notes to Gemini 3.6 Flash

cadam11:
Adds `@gemini-cli` as the everyday mention bot and moves release-notes generation onto Google's `run-gemini-cli` action with `gemini-3.6-flash`. Requires the `GEMINI_API_KEY` repo secret, which is already set.

`@claude` is **kept, unchanged, and still live** — it just stops being the default. Nothing routine reaches it, so it is idle and free unless someone deliberately types `@claude`, which also makes it easy to run both bots on one thread and compare. `ANTHROPIC_API_KEY` is therefore still in use and must not be removed.

Everything else in `.github/` matching "claude" is a path filter or a comment and is untouched. In particular the `__mj` hard-coded-schema check and the migration ID-determinism script remain deterministic greps — they were deliberately converted away from an LLM when the Opus-powered review was removed, and putting a model back on them would only add cost, latency and the ability to miss a hit.

## What a reviewer should look at

**1. The MCP wiring in `generate-release-notes.yml` is the risky part, and it has not been executed.**

Setting `tools.core` at all emits a blanket `{toolName: "*", decision: DENY}` policy rule at priority 4.24, which outranks both MCP grant priorities (trusted 4.2 / allowed 4.1). Google's own published examples pair `tools.core: []` with an `mcpServers` block, and that combination appears to deny the MCP tools too — their integration tests for it assert against hard-coded responses behind `describe.skip`, so they would pass either way.

The workaround here is to re-permit the two GitHub tools through `tools.allowed`, which sits at 4.3 and does outrank the deny. That is reasoned from the priority constants, not observed. **If the agent reports it has no GitHub tools, this is why**, and the fallback is adding `GEMINI_CLI_TRUST_WORKSPACE: 'true'` as a step env var.

Note also the two spellings are not interchangeable and both are load-bearing: `tools.allowed` takes the fully-qualified `mcp_github_pull_request_read`, while the server's `includeTools` takes the bare `pull_request_read` — the include filter runs before the name is qualified.

**2. The write-access gate genuinely weakened.**

`claude-code-action` performed its own write-access check behind the `author_association` `if:`. `run-gemini-cli` does not. That `if:` is now the sole guard on who can spend API budget. The known gotcha it already carried still applies — org members with *private* membership may report as `NONE`/`CONTRIBUTOR` and silently not trigger the bot.

**3. `--yolo` is hardcoded in the action** (`action.yml:309`/`316`) with no input to disable it. Every tool the model can see, it can call unprompted. That is why `gemini-command.yml` is read-only at the registry level rather than relying on approval policy, and why the reply is posted by a separate step instead of by the agent.

**4. Allowlist, not blocklist — on purpose.** Settings tool names are never validated (`isValidToolName()` is not called on `settings.json`). A typo in a `tools.exclude` entry silently leaves that tool **enabled**; a typo in `tools.core` silently leaves it **disabled**. Please keep any future edits on the allowlist side.

## Changes

**`gemini-command.yml`** (new)

- Triggered by `@gemini-cli`; `claude-command.yml` keeps `@claude`. Distinct trigger phrases and distinct concurrency groups, so neither cancels the other
- Agent restricted to `read_file`, `read_many_files`, `list_directory`, `glob`, `grep_search` — no shell, no network, no write, no GitHub tools
- Reply posted by a new step from the action's `summary` output, so the model never holds a GitHub credential. Failures post the error and a run link rather than finishing silently green
- Comment body reaches the agent as a file staged through an `env:` var, never interpolated into a `run:` script or the `prompt:` input, and the prompt instructs the model to treat it as untrusted data
- `pull-requests`/`issues` go from `read` to `write` for the new posting step; the agent itself cannot reach that token
- All the safety furniture from the Claude workflow is carried over: insider `if:` gate, non-bot check, `timeout-minutes: 20`, per-thread concurrency with `cancel-in-progress`

**`generate-release-notes.yml`**

- Same tool power as before (shell, write, GitHub MCP) — parity with the `Bash,Read,Write,Glob,Grep` it already had
- Prompt tool references updated: `mcp__github__get_pull_request` → `mcp_github_pull_request_read` (method `get`), `mcp__github__update_pull_request` → `mcp_github_update_pull_request`
- The changeset pre-flight gate and the `head_ref` condition that prevents destructive PR-title overwrites are unchanged

**`claude-command.yml`** — restored unchanged apart from a header comment recording that it is now the on-demand bot rather than the default.

**Both Gemini steps**

- `gemini_model: gemini-3.6-flash` pinned explicitly — the CLI's default alias `auto` currently resolves to `gemini-3-pro-preview`, which is retired
- `gemini_cli_version: '0.54.4'` pinned, because the tool-name strings in the settings blocks are only true for a known CLI version and `latest` moves
- Action pinned to `@v0.1.22`; the upstream examples use `@main`
- `id-token: write` dropped from `generate-release-notes.yml` — it existed for Anthropic's OIDC path and has no consumer there now

## Verification

- `actionlint` exits 0 on all three files (it was also clean before the change, so nothing is inherited)
- Both embedded `settings` JSON blocks parse, with the expected `tools.core` / `tools.allowed` / `includeTools` contents

Not yet smoke-tested against a live run, and neither workflow can be exercised from this PR:

- `issue_comment` workflows always run from the **default branch**, so commenting `@gemini-cli` here executes `next`'s copy, which does not exist yet. The first invocation after merge is the test.
- `generate-release-notes` only fires on PRs into `main`, so exercising it needs a throwaway PR there.

Merging this is therefore the point at which the MCP question in item 1 gets answered.

https://claude.ai/code/session_01LgZCjpeHD7qnAFxLeqTfNW


cadam11:
@gemini-cli Smoke test following the merge of this PR. Two questions:

1. Read `.github/workflows/gemini-command.yml` and tell me exactly which tools the `settings` block grants you, quoting the `tools.core` array verbatim.

2. Separately, and without looking at that file, list every tool you *actually* have available in this run.

If those two lists differ in either direction, say so explicitly — that discrepancy is the thing being tested here, not your answer to either question on its own.

cadam11:
@gemini-cli Second smoke test, now that #3732 has landed the trust-workspace fix. Same two questions:

1. Read `.github/workflows/gemini-command.yml` and quote the `tools.core` array verbatim.

2. Separately, list every tool you *actually* have available in this run.

If those two lists differ in either direction, say so explicitly — that discrepancy is what's being tested, not either answer alone.

PR #3731, opened 2026-08-11
feat(composer): add a Skills button beside Plan Mode

cmbrock-BC:
## Summary

`/` skill commands already worked, but were reachable only by knowing to type `/`. Nothing on screen said the feature existed, so a user who had never been told about it had no way to find it. This adds a Skills button to the composer strip, beside Plan Mode, as the visible door to the same trigger.

Clicking it calls the new `MentionEditorComponent.OpenTrigger('/')`, which opens the dropdown **without writing anything into the editor**. The trigger character only ever exists as the chip the user picks, so dismissing leaves the message exactly as they left it.

With no character to anchor on, the editor captures a baseline length when a trigger opens this way, and three paths consult it: typing filters against the baseline rather than searching for the character, chip insertion removes only what was typed, and dismissal has nothing to undo.

## Changes

* Skills button in `mj-message-input-box`, built as a sibling of the existing strip controls: same `attach-button-icon` chrome, same active treatment, joins the strip's visibility gate.
* `OpenTrigger(char)` and `IsTriggerOpen(char)` on `MentionEditorComponent`, both public and generic. Any trigger with an active provider can now be opened from a control, and any control can reflect whether its trigger is open.
* `beforeSkillsOpened` / `afterSkillsOpened` as a cancelable pair per `guides/UI_LAYERING_GUIDE.md` section 6, with `CancellableComposerEventArgs` declared in `ng-composer` (the dependency runs the other way, so it cannot import the `ng-conversations` base).

Three decisions worth calling out, since each departs from the button beside it:

1. **No new cap.** Gated on the existing `EnableSkillCommands` / `enableSkillCommands` / `allowSkillCommands` chain, which already defaults true at every layer. The button and the keystroke are two doors to one feature, so one flag governs both rather than letting a composer advertise skills it will not serve.
2. **The pressed state is derived, not an `@Input`.** Plan Mode's active state is a persisted user preference the host owns and threads down. "Is the skill dropdown open" is intrinsic to the composer, so it reads `IsTriggerOpen('/')`. An `@Input` there would be an API no host could answer, and would leave the button permanently unpressed if nobody bound it.
3. **The ARIA differs from Plan Mode deliberately.** Plan Mode is a toggle button, a mode that stays on, so `aria-pressed` is correct there. This opens a popup, so it is `aria-expanded` + `aria-haspopup`. Announcing "pressed" for revealing a list is the wrong thing for a screen reader to say. Flagging it because the sibling rule would otherwise suggest "fixing" it back.

## Two pre-existing dropdown bugs fixed along the way

Both affect every trigger (`@`, `#`, `/`), not just the new button. Happy to split them out if you would rather review them separately, but the button makes both trivially reachable and shipping it without them would look broken.

* **Click-away never dismissed.** Dismissal relied entirely on the editor's blur, and clicking a non-focusable area does not blur a contenteditable, so the dropdown stayed open with nothing able to close it. A `document:mousedown` listener now closes it, chosen over `click` because mousedown fires before focus moves and therefore cannot race blur's 200ms timer. Clicks inside the component are exempt so a suggestion row still selects.
* **The dropdown could land off screen.** Positioning measures the caret, and a collapsed range in an empty editor measures 0x0 at 0,0 in every browser, which pinned the menu to the bottom-left corner of the viewport. It now falls back to the editor's own box, and prefers to open above the composer, which sits at the bottom of the chat.

## Testing

* `ng-composer` 61 passing, `ng-conversations` 1079 passing. 20 specs are new across three files: the button's shape and wiring, `OpenTrigger`'s write-nothing contract, and the click-away dismissal.
* `mj standards check` passes, including `ui-layers`.
* Additive and opt in throughout. `enableSkills` defaults false at the leaf, and the chain above only turns it on where skill commands were already enabled.

## Live testing

Tested this in the Betty repo before opening the PR. Built both packages off this branch and dropped the dists into Betty's node_modules, versions matching exactly at 6.1.0-edge.1 so nothing was being forward ported onto older code. Betty already passes `allowSkillCommands`, so the button showed up with no Betty side changes at all, which is the point: this is MJ's button, not something a host has to assemble.

Click testing is what found most of the real problems. The first version inserted the `/` into the composer to reuse the typed path, and that was wrong: click out instead of picking a skill and you were left with a stray character you never typed. That is what led to the virtual trigger, where nothing is written until you pick.

The other three came out the same way. The menu opened and then closed itself, because a button takes focus on mousedown and the editor closes its dropdown 200ms after losing focus. It opened bottom left and ran off screen on an empty composer. And clicking on the message list did not dismiss it at all, because that does not blur a contenteditable.

Worth noting the last two were already there for anyone typing `/`. The button just made them easy to hit.

One thing to know if you are testing against a host that has its own skills UI: Betty had a skills menu of its own below the composer, and having both on screen made it genuinely hard to tell which one was misbehaving. Worth suppressing the host's version before testing this.


.changeset/composer-skills-button.md
```diff
@@ -0,0 +1,71 @@
+---
+"@memberjunction/ng-composer": minor
```
AN-BC, line 2: @cmbrock-BC here is a refresh for you again on version types:

* patch means CODE ONLY (this)
* minor means data changed - either a migration or metadata changes
* major, you wouldn't ever set, but this means a breaking change

So this is a `patch` version.

packages/Angular/Generic/composer/src/lib/components/message/message-input-box.component.ts
```diff
@@ -58,6 +59,30 @@ export class MessageInputBoxComponent {
...
   /** Current Plan Mode toggle state (renders the button in its active state). */
   @Input() planModeActive: boolean = false;
+  /**
+   * Shows the in-composer Skills button when true.
+   *
+   * Gated separately from `enableMentions` because a host can have the `/` trigger active and
+   * still not want the button — but the common case is that a composer offering skill commands
+   * should advertise them, since `/` is otherwise invisible.
+   */
+  @Input() enableSkills: boolean = false;
```
AN-BC: realizing that this component violates convention in MJ, public members of a class including angular components must be `PascalCase` please fix this. This is a breaking change but I'm near 100% certain this component is only used within MJ and our other apps, so go ahead with this and for this reason actually we can lave this PR as `minor`.

AN-BC: To be clear, fix not only your new props, but also existing props, events and methods that are **PUBLIC** to be `PascalCase` and please be on lookout for other convention violations, all classes in MJ must follow this and also exported functions we treat in same manner.

packages/Angular/Generic/composer/src/lib/events/composer-events.ts
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @module @memberjunction/ng-composer
+ */
+
+/**
+ * Base for cancelable composer events. Listeners flip `Cancel = true` to halt the default
+ * behavior; the matching `After*` event will NOT fire. `CancelReason` is free-form, for telemetry
+ * or for a host that wants to explain itself in the UI.
+ */
+export class CancellableComposerEventArgs {
```
AN-BC, line 28: there is a base class for a cancelable event somewhere else, look for this and inherit from it or use it directly no reason to rebuild this.

AN-BC (approved):
Overall, this looks good to me. Since this is a UX change, I want @MattC-BC  to approve this.

Colin - once Matt reviews/approves this, and you've made the changes I have provided, go ahead and merge.

Nice work!

AN-BC:
@rkihm-BC see the feedback I gave to Colin here, we need this in our training and PR reviews.

packages/Angular/Generic/composer/src/lib/components/message/message-input-box.component.ts
cmbrock-BC: will do.

cmbrock-BC:
Thanks, all three addressed in b4ba824.

**PascalCase.** Renamed every public input, output, getter and method on `MessageInputBoxComponent`, not just the new ones: `placeholder`, `disabled`, `value`, `valueChange`, `textSubmitted`, `planModeToggle`, `canSend`, `onSendClick` and the rest. `TriggerProviders`, `ExcludedTriggerKeys` and `Provider` were already correct. `mj-ai-composer` and both specs updated.

Three things I deliberately left camelCase, say the word if you want any of them changed:
* `[disabled]` on native `<button>` elements, that's the DOM property
* `ngOnInit` / `writeValue` / `registerOnChange`, framework contracts
* `mj-mention-editor`'s own inputs, since that component isn't renamed in this PR

On the lookout point: `MentionEditorComponent`, `AiComposerComponent` and `MessageInputComponent` have the same violation. I left them out to keep this reviewable, happy to do them here or as a follow-up, your call.

**Cancelable base.** There wasn't an existing one to inherit. Five packages had each declared their own identical `Cancel` + `CancelReason` class: `ng-trees`, `ng-conversations`, `ng-task-graph-editor`, `ng-whiteboard` and `LiveKitRoomCore`. Adding a sixth was clearly wrong, so I put one `CancelableEventArgs` in `@memberjunction/global` (framework agnostic, everything can reach it) and inherit from it. Migrating the other five is left out on purpose since each is a public type their consumers extend, but that's the obvious follow-up.

**Version.** Noted, and saved. Left it `minor` per your second comment since the rename is itself breaking, and rewrote the justification: it previously argued `minor` on API-surface growth, which is npm's convention and not what the field means here.

Also rebased onto your merge from `next` and re-verified after it: build clean, `ng-composer` 61 passing, `ng-conversations` 1079 passing, `mj standards check` green.

Over to @MattC-BC for the UX review.

cmbrock-BC:
Followed up on the cancelable base with a much wider search, and the picture is clearer than what I wrote above. There is no shared base anywhere in the repo, there are **eight independent roots**, and they have diverged:

| Package | Root | Extends |
|---|---|---|
| `base-forms` | `CancellableFormEvent` | nothing |
| `conversations` | `CancellableChatEventArgs` | nothing |
| `task-graph-editor` | `CancellableTaskGraphEventArgs` | nothing |
| `entity-viewer` | `CancelableGridEventArgs` / `CancelableRowEventArgs` | its own grid roots |
| `trees` | `CancelableTreeEventArgs` / `CancelableNodeEventArgs` | its own tree roots |
| `tab-strip` | `TabCancelableEvent` | its own `TabEvent` |
| `user-routines` | `CancelableUserRoutineEventArgs` | its own root |
| `whiteboard`, `LiveKitRoomCore` | inline `Cancel: boolean` | n/a |

Two things that establish none of them is a library base rather than a local one:

1. Nothing imports a cancelable base across package boundaries. The only cross-package import of one in the whole repo is the line I added in this PR.
2. `CancellableFormEvent` extends nothing and is imported nowhere outside `base-forms`. Same for the `conversations` and `task-graph-editor` roots.

They have also drifted apart: some carry `CancelReason`, some only `Cancel`, and the spelling splits between `Cancelable` and `Cancellable`.

So `CancelableEventArgs` in `@memberjunction/global` is the first shared one rather than a ninth copy. If you had a specific existing class in mind that I have missed, point me at it and I will switch to it. Otherwise the follow-up worth doing is migrating those eight onto this base, which I left out of this PR because each is a public type that consumers extend.

MattC-BC:
@AN-BC @cmbrock-BC - just logged in and taking a look now. I will provide feedback ASAP - thanks!

cmbrock-BC:
Correcting my last comment, I had this backwards.

I went looking properly for the base class you mentioned and could not find a shared one, so I put a `CancelableEventArgs` in `@memberjunction/global` and argued that the per-package copies were duplication. That argument was wrong, and I have reverted it.

`guides/UI_LAYERING_GUIDE.md` section 6 specifies the base as **per-domain**, in its own naming table:

| Thing | Convention | Example |
|---|---|---|
| Cancelable base | `Cancellable<Domain>EventArgs` | `CancellableJournalEntryEventArgs` |

So the sixteen packages each declaring their own are following the documented convention rather than drifting from a shared one. `CancellableFormEvent`, `CancellableChatEventArgs`, `CancellableTaskGraphEventArgs`, `CancelableTreeEventArgs`, `CancelableGridEventArgs` and the rest are all correct as they stand.

This PR is now back to `CancellableComposerEventArgs`, declared in `ng-composer`, which is what the naming table asks for and what I had before your comment. Nothing in `@memberjunction/global` is touched any more.

What the search covered, in case it is useful:

* No cancelable class or interface exists outside `packages/**` (checked all twenty top level directories).
* No published `@memberjunction/*` `.d.ts` declares one either, so there is nothing in a released package that source does not show.
* Sixteen packages declare their own `Cancel` contract. Before this PR, not one of them imported a cancelable base across a package boundary, so there was no shared base to inherit from even informally.
* The spelling splits between `Cancelable` and `Cancellable`, and some carry `CancelReason` while others only carry `Cancel`.

There is also a dependency constraint that rules out reusing the closest neighbour regardless: `ng-composer` is the generic layer and `ng-conversations` depends on it, so it cannot import `CancellableChatEventArgs`.

If you did mean a specific class and I have still missed it, point me at it and I will switch. And if you would rather consolidate all sixteen onto one shared base, that seems worth doing, but it would need the guide's naming table updated in the same change, and it is a bigger conversation than this PR.

MattC-BC:
@cmbrock-BC - great work overall! I have 2 very minor things that I would like that shouldn't be too much effort. Let me know if you have any pushback or would like to chat further. Thanks for pushing this through! 

## 1. Second click on the button can't close the dropdown (confirmed bug)

The toggle branch in `OnSkillsClick()` is dead code in a real browser. Here's the sequence:

1. Your new click-away listener is `@HostListener('document:mousedown')` on `MentionEditorComponent`, and it exempts presses inside *that component's host*.
2. The Skills button lives in `mj-message-input-box` as a **sibling** of `<mj-mention-editor>`, outside that host.
3. So on the second click: the **mousedown** bubbles to the document listener, which sees an "outside press" and closes the dropdown. Then the **click** fires `OnSkillsClick()`, `SkillsActive` reads false (it just closed), and the full open path runs again.

Net effect: the menu blinks and comes back, the user can't dismiss it with the button that opened it, and `BeforeSkillsOpened`/`AfterSkillsOpened` fire on every click, which is exactly the over-counting the toggle branch documents itself as preventing.

The reason your specs pass anyway: they assert source text, and the code shape is there. The integrated behavior isn't. This test (real editor + dropdown, bubbling mousedown then click, the way a browser actually fires them) reproduces it:

```typescript
/** Real user interaction = mousedown (bubbles to document) then click. */
function userClick(el: HTMLElement): void {
  el.dispatchEvent(new MouseEvent('mousedown', { bubbles: true, cancelable: true }));
  el.dispatchEvent(new MouseEvent('mouseup', { bubbles: true, cancelable: true }));
  el.dispatchEvent(new MouseEvent('click', { bubbles: true, cancelable: true }));
}

it('second click on the button closes the dropdown (toggle contract)', async () => {
  const f = render(); // real MessageInputBoxComponent + MentionEditorComponent + MentionDropdownComponent,
  await flush(f);     // FakeSlashProvider bound via TriggerProviders, EnableSkills: true
  const btn = query(f, '.skills-button-icon') as HTMLElement;
  userClick(btn);
  await flush(f);
  const editor = f.componentInstance.mentionEditor!;
  expect(editor.showMentionDropdown).toBe(true);

  userClick(btn);
  await flush(f);
  expect(editor.showMentionDropdown).toBe(false); // FAILS on this branch: it reopens
});

it('Before fires once across two clicks', async () => {
  // same setup, subscribe to BeforeSkillsOpened, two userClicks
  expect(beforeCount).toBe(1); // FAILS on this branch: 2
});
```

The fix is one line. The button's mousedown handler already exists to say "this press belongs to the dropdown machinery", so extend the same statement to the click-away listener:

```html
(mousedown)="$event.preventDefault(); $event.stopPropagation()"
```

I verified both failing assertions go green with that line. Please ship an integration-style test like the above with the fix rather than a source-text spec, since source-text is exactly how this one slipped through. 

I'd guess Betty testing missed this because dismissing via click-away or picking a skill both work fine. Only click-the-button-again is broken, and the failure looks like "it stayed open."

## 2. The menu should anchor to the button that opened it

Right now the button-opened dropdown appears at the composer's **left** edge, because the empty-editor fallback substitutes the editor's box and `left` comes from `cursorRect.left`. Meanwhile the voice options picker anchors bottom-right above the composer, right over the phone button (its own comment says so, `realtime-agent-picker.component.ts:85`). Two strip buttons producing popovers on opposite corners of the composer is the kind of inconsistency users feel even when they can't name it.

The typed path is correct and shouldn't change: when someone types `/`, their eyes and their query are at the caret, so the menu belongs at the caret. But on the button path nobody's eyes are at the caret, they're on the button they just pressed. Standard menu-button behavior: the popup anchors to its trigger.

What I'd like:

1. `OpenTrigger(triggerChar, anchorEl?)`, with the button passing its own element. This actually strengthens your generic story: "any control can open any trigger" becomes "and the control is the anchor."
2. While `virtualTriggerOpen` is set, `positionMentionDropdown()` uses the stored anchor rect, right-aligned above the button cluster, same corner as the voice picker.
3. **One trap to design around**: `positionMentionDropdown()` re-runs on every keystroke while the user types to filter, and once there's text the caret rect measures real coordinates. A naive fix anchors the initial open at the button and then jumps to the far-left caret on the first keystroke, which is worse than what we have. The anchor needs to be **sticky** for the life of a button-opened session, not just the first paint.
4. Right alignment needs a small CSS assist since the dropdown width is content-based (280 to 400px): an `align-right` class doing `translateX(-100%)`, composing with `show-above` into `translate(-100%, -100%)`. Same mechanism you're already using for above/below.
5. While you're in that CSS block, cap the width for mobile: `max-width: min(400px, calc(100vw - 32px))` on `.mention-dropdown`. Desktop stays at the current 400px, phones cap at viewport minus a 16px gutter each side, and the 280px min-width still clears even a 320px device. Fixed caps don't survive the device landscape, so viewport-relative it. And since right-alignment computes the left edge as `anchorRight - dropdownWidth`, clamp the computed `left` to the gutter in `positionMentionDropdown()` so no anchor position can push the menu off-screen to the left.

There's a real tradeoff here and I'm choosing it deliberately: someone who opens via the button and then types to filter will have their query at the caret and the results at the right. I'm fine with that split, because the button path is the discovery path and discovery users pick with the mouse, which is already sitting on the menu. People who type-to-filter are learning `/` (your tooltip is teaching it), and the typed path keeps caret anchoring.

Typed `/`, `@`, and `#` behavior stays exactly as is.

---

Two side notes while you're in there, take or leave: the comments above the toggle branch (and one spec, and the changeset) say "a button carrying `aria-pressed` has to be able to un-press", but the button deliberately carries `aria-expanded`, per your own decision #3. The argument still holds, the prose just contradicts it. And if a host has zero skills defined, the button click is a silent no-op because `fetchSuggestions` closes on empty results, which also makes the dropdown's `mj-empty-state` branch unreachable. For a typed `/` a silent no-op is fine; for a button it reads as broken. Showing the existing "No matches found" empty state on the button path would cover it.

cmbrock-BC:
@MattC-BC @AN-BC — pushed the follow-ups from both reviews.

**Matt**

* **Toggle** fixed exactly as you called it: the button's `mousedown` now stops propagation, so the press stops reading as an outside click. Replaced the source-text spec with an integration-style one that fires real bubbling `mousedown` → `click` against the real editor and dropdown, including your `Before fires once across two clicks` assertion.
* **Anchoring** — `OpenTrigger(triggerChar, anchorEl?)`, anchor sticky for the session, `align-right` composing with `show-above`, `max-width: min(400px, calc(100vw - 32px))`, and `left` clamped to the gutter. One deliberate difference: it aligns left and flips to right-aligned on overflow rather than always right-aligning, so the rule also holds for a left-hand control. The flip aligns to the *composer's* right edge, not the button's — pinning to one 36px icon hung the menu's whole width out to its left.
* Both side notes taken: the `aria-pressed` prose is corrected, and the button path now holds the dropdown open so the existing empty state renders.
* Reviewing that turned up four more, all fixed: a trigger character already in the message hijacked a button-opened session; a press whose suggestions hadn't loaded couldn't be dismissed; the empty dropdown swallowed Enter and ignored Escape; and the anchored path dropped below without checking below was better.

**Amith**

* Changeset stays `minor` per your note in the PascalCase thread.
* `MessageInputBoxComponent` is fully converted — every public input, output, getter and method, not just the new ones. Framework contracts (lifecycle hooks, the CVA quartet) and native DOM bindings are untouched.
* On "be on the lookout" — I did sweep the neighbouring `MentionEditorComponent` and `MentionDropdownComponent` as well, but it broke the composer at runtime in a downstream app despite clean builds and a green suite, so I pulled it. The diff is saved and I'd rather land it as its own PR with every consumer rebuilt together than risk it here. Happy to open that next if you want it.
* The cancelable base is covered in my comment above — no shared base exists, so this stays per-domain per the naming table in `UI_LAYERING_GUIDE.md` section 6.

Green: `ng-composer` 56/56, `ng-conversations` 1079/1079.

PR #3735, opened 2026-08-11
Rework: Query & Entity Materialization (join-table redesign)

### LINKED ISSUE #3626: Rework: Query & Entity Materialization
## Rework: Query & Entity Materialization

The Query & Entity Materialization feature (#3311 Phase 1, #3365 Phase 2) was merged, then **retired from `next`** — partially by #3617 and fully by #3623 — because of a set of correctness/design issues found in review. `next` now has **zero** materialization code. This issue tracks the clean re-implementation.

Nothing is lost: the full feature code is permanently in history at the two squash commits below, and the design docs are back on `next`.

### 📚 Reference material
- **Design docs (on `next`):** [`plans/query-entity-materialization.md`](../blob/next/plans/query-entity-materialization.md) + [`plans/query-entity-materialization-phase2.md`](../blob/next/plans/query-entity-materialization-phase2.md)
- **Exact feature code (git history):** `git show 4c14b3f` (Phase 1 / #3311) and `git show 469862f` (Phase 2 / #3365)
- **Original PRs with inline review comments:** #3311, #3365
- **Removal PRs:** #3617 (partial), #3623 (remnant cleanup — the definitive "what was materialization" delete list)
- ⚠️ The `feat/query-materialization-rework` branch has been **deleted** — it was only a stale full-`next` snapshot and a poor base. **Branch the rework from current `next`** and re-introduce the code guided by `4c14b3f`/`469862f`.

### 🔧 Required changes for the rework

**1. Break the circular FK (blocker).** The foundation migration created FKs in both directions (`MaterializedResult.SourceQueryID → Query` and `Query.MaterializedResultID → MaterializedResult`), which CodeGen flags as the `MJ: Materialized Results ↔ MJ: Queries` cycle.
> **Amith's design:** introduce a dedicated join table **`__mj.MaterializedResultQuery`** (`ID` PK, `MaterializedResultID → MaterializedResult`, `QueryID → Query`, + standard `__mj` system columns) and carry the relationship as rows. Both FKs point outward → no cycle. Drop the direct `Query.MaterializedResultID` and `MaterializedResult.SourceQueryID` columns/FKs and resolve through the join table.

**2. One consolidated migration, authored correctly.** Ship a **single** `V<timestamp>__v6.1.x__…` migration in `migrations/v6/` with a correct, final `CREATE TABLE` (all columns/constraints present from the start — `RowFilter`, `KeyColumns`, `SourceRowCount`, `RefreshesSinceFullRebuild`, `ReadFilterSpec`, `ForceFullRebuildCadence`). Run on a clean DB from the latest published release, run CodeGen, and **concatenate the CodeGen emit into the same file** after 50+ blank lines and a comment banner. Do **not** ship the create→alter→codegen→alter chain or a separate `…CodeGen_Generated_Objects` migration.

**3. Version label is `v6.1.x`**, not `v6.2.x`. Edge-release convention: release pins major/minor at 6.1 (`.edge-x` increments); migrations keep the legacy naming but stay `v6.1.x` until LTS.

**4. Re-apply the four review findings** (details in the #3311/#3365 inline comments):
- **PG swap atomicity** — Phase 2 split the PG rebuild swap into individual autocommit statements while SQL Server kept a single transaction; a crash between `DROP TABLE <canonical>` and `RENAME` destroys the old snapshot. Keep the PG swap in one `BEGIN;…COMMIT;`.
- **RunQuery cache path** — `queryCacheEngaged` / `GenerateRunQueryFingerprint` don't account for `DataSource:'Materialized'`, so a `CacheLocal:true` query cross-serves Live vs Materialized from one slot. Mirror the RunView guards (bypass cache for materialized + fold `ds:materialized` into the fingerprint).
- **Minted-entity permissions** — `addMaterializedQueryEntityPermissions` runs only on first mint (`if (!generatedEntityId)`), so a later revocation of source read access isn't re-applied. Recompute the intersection grant every CodeGen run.
- **Mint-time DDL identifier escaping** — `generateMaterializedTableSQL` emits column identifiers unescaped (SS `[${c.Name}]`, PG `QuoteIdentifier`) while the refresh path escapes; reachable via untrusted external-source column names. Escape in the generators.

### ✅ Definition of done
Fresh SQL Server **and** PostgreSQL end-to-end run on the final code, unit + integration tiers green, and one more adversarial `max` review pass before merge (the original never hit two consecutive clean rounds).

/cc @AN-BC

JosueGarcia-BC:
Closes #3626.

Clean re-implementation of Query & Entity Materialization (originally PRs #3311 / #3365), which was merged to `next` and then reverted (#3617 / #3623) over two architectural objections. This PR re-introduces the feature with both objections resolved.

## What the feature does

Snapshots a stored **Query** result (or an entity's **base view**) into a physical, read-only **Virtual Entity**, refreshed on a schedule via an atomic wrapper-view swap. Reads opt in with `RunViewParams`/`RunQueryParams` `DataSource: 'Materialized'`, which the provider redirects to the materialized view, falling back to `Live` on any uncertainty (serving live is always correct). Phase 2 adds parameterized `RowFilterBroad` read-time predicate injection (bound, never interpolated).

## What changed vs. the reverted version

**1. Circular FK removed (the #1 objection).** `MaterializedResult.SourceQueryID` and `Query.MaterializedResultID` are **dropped**. The MR↔Query link now lives in a dedicated join table **`__mj.MaterializedResultQuery`** (`ID` PK, `MaterializedResultID`→MR, `QueryID`→Query, both FKs *outward*, 1:1 via two `UNIQUE` keys). CodeGen confirms **no cycle**. The link is resolved through the join at the three code sites that need it:
- `GenericDatabaseProvider.tryBuildMaterializedQueryPlan` (read-redirect)
- `manage-metadata` provisioning + drift scan (aliases `mrq.QueryID AS SourceQueryID` so downstream consumers are unchanged)
- `MaterializationRefresher.resolveSourceQueryId` (used by `resolveSourceQuery` / `resolveSourceSelect`)

**2. One self-contained migration (the #2 objection).** A single `v6.1.x` migration with a correct final `CREATE TABLE` (all columns up front — no create→alter→codegen→alter chain), with the CodeGen SQL emit concatenated into the same file (MJ convention — no separate `CodeGen_Run` file). Verified: applying the migration alone on a fresh full-stack DB creates the tables, base views, CRUD sprocs, permissions, **and** the entity/field metadata for both new entities.

**3. The four original review findings** are carried in the re-introduced code: PG swap atomicity, RunQuery cache `DataSource`, first-mint permission re-scoping, and DDL identifier escaping. Escaping now covers **both** paths: the **mint** path (dialect-level `QuoteIdentifier`/`QuoteSchema`/`QuoteColumnAlias`) and — added in response to review — the **refresh** path (a plain-identifier assertion on the schema/table/view names read from the writable `MJ: Materialized Results` row, which closes the swap-DDL `EXEC`/`sp_rename`/`CREATE VIEW` string-literal surface, fail-closed).

## Contents
- **Migration** `migrations/v6/V202608140000__v6.1.x__Query_Entity_Materialization.sql` — DDL + concatenated emit (T-SQL only; the PG counterpart is deferred to the release build per `migrations/CLAUDE.md` — build-engineer work, not authored in a feature PR)
- **New package** `@memberjunction/materialization` — the refresh engine (`MaterializationRefresher`) + `MaterializationFreshness`
- **Read path** — `DataSource` plumbed through `core` → `GenericDatabaseProvider` → GraphQL (`server`, `graphql-dataprovider`)
- **Scheduling** — `MaterializationRefreshScheduledJobDriver` + the seeded scheduled-job-type metadata
- **CodeGen** (`codegen-lib`) — materialization provisioning, parameterized-query classification/verification, join-table wiring
- **Generated files** — `MJ: Materialized Results` + `MJ: Materialized Result Queries` entities, resolvers, and forms

## Validation
- **SQL Server E2E** (real DB, full runtime path): CodeGen provisioning mints the join table + Virtual Entity + snapshot objects → `MaterializationRefresher.RefreshOne` resolves the source query **through the `MaterializedResultQuery` join** and rebuilds the snapshot → the materialized view (`materialized_vwActive_Users`) and its backing table are populated (2 rows) and read back matching the live query. This exercises the join-table redesign end-to-end.
- **Read-redirect served-from-snapshot** (live-verified): two new deterministic integration bundles each fabricate a snapshot carrying a **sentinel row absent from the live source** and prove the read hit the snapshot, not merely correct rows —
  - **`IT79` (`materialized-read`)**: `RowFilterBroad` `RunQuery(DataSource='Materialized')` returns the sentinel with `RenderedSQL` referencing `materialized_…` + the surrogate order key, while a negative-control `Live` read returns it nowhere; a third check **MR3** verifies the reverse-direction delete-path FK cleanup (deleting a MaterializedResult with its join row present cleans the join first, no raw FK violation). **3/3 live.**
  - **`IT78` (`materialized-entity-read`)**: the **entity base-view redirect** — `RunView(DataSource='Materialized')` served from the entity's snapshot via `resolveEffectiveBaseView`, with a `Live` negative control. **2/2 live.**
  External-source mint + date-filter binding remain covered by unit tests + the runtime **fall-back-to-live safety net** (any materialized-read error serves live), which makes a defect there fail *safe* rather than *wrong*.
- **Self-contained migration**: full baseline→v5→v6 stack applies on a fresh DB (44/44, 0 errors); the v6 migration alone creates all objects + metadata for both new entities.
- **Build**: full repo builds green (305/306 — the only failing package is MJExplorer, solely because its gitignored `environments/` config isn't present in the CI-less worktree — unrelated).
- **Unit tests**: green across the touched packages — materialization (105, incl. the refresh-path identifier-guard suite), GenericDatabaseProvider (946, incl. the materialized-read builder/binding/paging suite), codegen-lib (904), server (821), core-entities-server, and the integration-test-suite guards, plus core / sql-dialect / scheduling-engine / graphql-dataprovider.

## Adversarial review
A max-effort adversarial review was run against the diff. **Four findings fixed** in this branch:
1. Materialized `RunQuery` read now paginates at the SQL layer (`OFFSET/FETCH`) instead of loading the whole snapshot into memory — matches the live read path.
2. The external base-view mint now excludes virtual fields, matching the refresher (previously guaranteed a column mismatch → permanent `DriftHold`).
3. `resolveEffectiveBaseView` reads the authoritative `ViewName` from the MR row instead of deriving it by naming convention.
4. Date row-filter params bind without the ISO `Z` suffix on SQL Server (`datetime2` rejects it — error 241 — forcing a live fallback); PostgreSQL keeps `Z` for `timestamptz`.

**Three no-change** (dialect-level `QuoteSchema` is the requested hardening and a no-op for the lowercase `__mj` schema; a latent composite-PK `Delete` assumption is unreachable for read-only virtual entities; the seed's hardcoded primaryKey matches all sibling scheduled-job-type seeds). **Five documented follow-ups** (maintainability/efficiency: constant / operator-whitelist / quoting-helper duplication, a per-read freshness round-trip, and a redundant re-probe on the rare count-drop fallback).

### Independent pre-review audits (final merged diff)
Three independent audits were run against the final diff before requesting review:
- **Design + migration + policy** — both original objections confirmed dead (no circular FK: join-table-only, both FKs outward, 1:1 via two `UNIQUE` keys, no `SourceQueryID`/`MaterializedResultID` columns; one self-contained migration, all columns up front, CodeGen emit inlined). Every migration convention passes. Notable strength: the mint is **fail-closed on RLS** — it refuses to materialize a query whose source RLS can't be proven safe, and drift re-check revokes read if a source later gains RLS.
- **Security (SQL injection)** — clean on both SQL Server and PostgreSQL. The read-time predicate injection **binds every value** (never interpolates), escapes all identifiers via a hardened `]]`/`""` helper, and enforces the operator whitelist at **both** generation and runtime (a tampered persisted spec still can't inject). Mint DDL, view-swap, external rebuild, watermark probes, and paging all verified safe.
- **Correctness** — no critical/high defects; atomic swap, watermark-on-failure, cache exclusion, type-coercion parity, and bound-parameter paging all verified correct. One finding fixed (below).

### Fixed from the correctness audit
- **Deterministic page order for materialized reads**: a materialized read carried no `ORDER BY`, so paged reads fell back to the dialect default order and successive pages could skip/duplicate rows when the plan changed between calls. Now emits `ORDER BY <__mj_MaterializedRowID>` — the snapshot's stable surrogate row-id, exposed by the wrapper view — so paging is deterministic and refresh-stable.

## Notes for review
- **PG query-materialization is SQL-Server-only by design** for now — it becomes cross-engine once the pre-existing `spCreateVirtualEntity` support proc is ported to PostgreSQL (tracked with the broader PG-parity effort). Entity base-view materialization is already cross-engine. Per `migrations/CLAUDE.md`, this feature PR ships the **T-SQL migration only**; the PG counterpart is generated by the build engineer at release time (the CI PG-parity workflow runs on that release PR, not on this one).
- This PR **deletes `next`'s `remove-materialization-remnants` changeset** (intentional) so the release changelog reads "materialization added" rather than "removed then re-added" within one unreleased edge cycle.
- Generated files were produced by a deterministic (keyless) CodeGen run, so the two new forms use MJ's default layout rather than AI-tuned layout.
- **Bootstrap class-registration manifests are intentionally untouched** by this PR — each bootstrap package regenerates its manifest at build time via its `prebuild` (`mj codegen manifest`), so the two new materialization entities are picked up automatically at release build (no manual manifest commit needed, and no production tree-shaking risk). Note for the team, unrelated to materialization: that same generator currently drops three validly-exported `@memberjunction/server` classes (`RateLimitMiddleware`, `MJTenantFilterMiddleware`, `MJEntityPermissionEntityServer`) because it can't follow the nested `export * from './middleware/index.js'` wildcard chain when resolving each package's public API — so its output regresses those classes on any build. Pre-existing/generator concern, not this PR.


rkihm-BC (commented):
## Code review — Claude Code

> Performed by **Claude Code** (Opus 5) at @rkihm-BC's request. I read the full diff, both bot comments, and the PR body, then re-derived the load-bearing claims on a local checkout of the branch rather than taking them on trust. **Please push back on anything I got wrong or missed** — where I could not verify something, I say so.

**Verdict: strong re-implementation, approve with changes.** Both revert objections are genuinely dead — I verified that in the SQL, not just the prose. The safety model is real rather than claimed. There is **one documented-policy violation that should come out before merge**, one hardening gap the PR's own summary implies is covered but isn't, a coverage gap that matters for a feature that has already been reverted once, and two factual errors in the description.

---

**Objection 1 (circular FK) is dead.** Confirmed in the migration: `MaterializedResult` has exactly two FKs, both to `Entity(ID)`. No `SourceQueryID`. `Query` gains only `IsMaterialized BIT NOT NULL DEFAULT 0` — no `MaterializedResultID`. The join table's two FKs both point outward, with `UQ_MaterializedResultQuery_MaterializedResult` and `UQ_MaterializedResultQuery_Query` enforcing the 1:1. There is no cycle to find.

**Objection 2 (one self-contained migration) holds.** All business columns are in the final `CREATE TABLE`. The only `ALTER … ADD` for the new tables is CodeGen's own `__mj_CreatedAt`/`__mj_UpdatedAt` block, which CodeGen owns and always emits that way — that is the convention, not the create→alter→codegen→alter chain the objection was about.

**Migration conventions — I audited the ones CLAUDE.md calls unrecoverable:**
- ✅ **No literal `Sequence` in any `EntityField` INSERT.** 36 apply-time `COALESCE(MAX([Sequence]), 0) + 1` expressions instead. This is the fresh-install-only landmine from #3670, and it's handled correctly.
- ✅ No `NEWID()` anywhere — hardcoded UUIDs throughout.
- ✅ Four `CHECK … IN (…)` constraints matching the four enums.
- ✅ `migrations/v6/` matches the `v6.1.x` in the filename.

**The body's own audit note is wrong in the PR's favour.** It lists "the four enum columns lack `EntityFieldValue` rows so their Explorer fields may render as free-text" as an open low item. They don't: the migration carries 14 `EntityFieldValue` inserts — exactly 2 + 4 + 3 + 5 for `SourceType`/`ParamMode`/`RefreshStrategy`/`Status` — plus four `ValueListType='List'` updates. That follow-up can be struck.

**The safety model is real.** Every uncertainty branch in `tryBuildMaterializedQueryPlan` returns `null` → live: not opted in, external source, no join row, non-`Active` status, wrong `ParamMode`, malformed spec JSON, and a **bidirectional** spec/param coverage check that refuses both under- and over-filtering. `buildMaterializedReadQuery` binds every value, re-validates the operator whitelist at runtime (so a tampered persisted `ReadFilterSpec` still can't inject), and returns `null` on a null/empty/unconvertible value. `queryHasTopLevelOrderBy` returns `true` (refuse → live) on parse failure or any statement shape it can't reason about. The execution wrapper re-throws connection errors and falls back to live on everything else.

**`IsMaterializedDataSource` is the right call.** One canonical trim+lowercase test, used at read routing, cache eligibility, the cache fingerprint, and `RunViewParams` equality. That closes the exact hazard scattered `=== 'Materialized'` checks would open — one site routing to the snapshot while another caches it as Live. And the fingerprint segment is appended *only* for the non-default value, so no existing cache entry is invalidated.

**`buildBroadRowFilterSQL`'s exact-count guard is genuinely well-reasoned.** A qualifier-blind matcher that returns the *unmodified* SQL when the match count differs from the classifier's parameter count means even a caller that ignores the `ambiguous` flag fails safe (still carries the filter → refused upstream) rather than silently broadening the snapshot.

**The atomic swap reads like lessons learned, not a first draft.** SQL Server: `SET XACT_ABORT ON`, expensive `SELECT … INTO` outside the transaction, view only ever pointed at the canonical name, index restored inside the swap, and a fixed short index name to dodge the 128-char `sysname` limit. PostgreSQL: the view repoint kept *inside* the transaction, with the comment explaining the earlier bug where an auto-committed repoint survived a rolled-back swap. Both comments name the failure they're avoiding.

**CI:** 13 checks green, including the deterministic integration tier and all six driver integration suites.

---

## What's concerning

### 1. The PostgreSQL counterpart should not be in this PR — and it inserts out of order

`migrations/CLAUDE.md` is unambiguous: *"❌ Do NOT create `migrations-pg/vN/<same-name>.pg.sql`."* The `pg-migrations.yml` header repeats it and explains why the workflow deliberately does not trigger on `migrations/**`. This PR adds a 4,231-line `migrations-pg/v6/V202608101500__…pg.sql`.

This isn't only formal. On `next`, `migrations-pg/v6` holds **15** files against `migrations/v6`'s **21**. Six T-SQL migrations have no counterpart yet:

```
V202608081200__…__Durable_EntityAction_Dispatch
V202608081622__…__ModelConfiguration_JSONType_Columns
V202608082130__…__Clear_EntityAction_RelatedNameField
V202608082250__…__TaskGraph_Skipped_And_Exclusive_Groups
V202608082330__…__Retire_Workflows_Application
V202608101400__…__Workflow_Human_Steps
```

This PR's counterpart is **V202608101500 — later than all six**. Any PG database that applies it will reject those six as out-of-order when the build engineer converts them at release time. The green gate doesn't contradict this: it validates a *fresh* PG database, where ordering can't bite. Recommend deleting the `.pg.sql` and stating the deferral in the description, which is exactly what the policy asks for.

### 2. The refresher's swap DDL hand-rolls the quoting the rest of the PR hardened

`MaterializationRefresher.ts` defines **four** separate local `const obj = …` quoters. The two used by the rebuild/swap builders escape nothing:

```ts
const obj = (n: string) => `[${schema}].[${n}]`;    // SQL Server — no ]] doubling
const obj = (n: string) => `${schema}."${n}"`;      // PostgreSQL — no "" doubling
```

and both names then land inside T-SQL **string literals** with no `''` doubling:

```sql
EXEC sp_rename '${schema}.${shadow}', '${tableName}';
EXEC('CREATE OR ALTER VIEW ${obj(viewName)} AS SELECT * FROM ${obj(tableName)}');
```

A single quote in `ViewName`/`TableName` closes the `EXEC(…)` literal. The same file already has both helpers — `esc = s => s.replace(/'/g, "''")` at lines 640/654, and a proper `]]`/`""` quoter at 1223 — they're simply not used on this path. I searched the file for identifier validation and found none.

On reachability, to be fair: `SchemaName`/`TableName`/`ViewName` are columns on `MJ: Materialized Results`, written by CodeGen from sanitized names — so this is not untrusted end-user input. But it *is* a writable metadata entity, so whoever can edit that row can have the privileged refresh job run arbitrary DDL. The reason it's worth naming precisely is that the body claims "mint-time DDL identifier escaping (hardened at the dialect level — `QuoteIdentifier`/`QuoteSchema`/`QuoteColumnAlias`)", which covers the **mint** path and not the **refresh** path. One shared quoter plus `''` doubling in the two EXEC literals — or a `^[A-Za-z_][A-Za-z0-9_]*$` assertion on the three name columns before any statement is built — closes it.

### 3. A performance feature that adds uncached round-trips to the hot read path

- `RunView(DataSource:'Materialized')` → `resolveEffectiveBaseView`: one `RunView` with `BypassCache: true`, per read.
- `RunQuery(DataSource:'Materialized')` → `tryBuildMaterializedQueryPlan`: **two** `RunView`s (the join lookup, then the MR row with `BypassCache: true`) *plus* a `SQLParser.Astify` of the platform SQL, all before a single snapshot row is read.

The `BypassCache` on the status read is well argued — DriftHold/Disabled are written out of band by CodeGen, so a cached `Active` would defeat flag-and-hold. But the join-table redesign *added* the second round-trip, and nothing memoizes the parse. For a feature whose entire purpose is making large reads cheap, three uncached metadata operations on every materialized read is a real tax, and on small-to-medium result sets it could be a net loss. The body lists "a per-read freshness round-trip" as a documented follow-up; I'd raise its priority — a short-TTL cache keyed on MR ID that any status write invalidates, plus memoizing `queryHasTopLevelOrderBy` per query+platform, would keep the guarantees and drop the cost.

### 4. No live-DB coverage for three of the four runtime paths

Unit coverage is genuinely strong (~1,900 assertions across the touched packages, green in CI) and there's one real SQL Server E2E for the query mint + refresh through the join. But the body states plainly that read-redirect + paging, the entity base-view redirect, and the external-source mint are covered only by unit tests, build verification, and the fall-back-to-live net.

That net is the problem as much as the mitigation. "Serves live on any doubt" means a broken materialization is **invisible** — you could ship a feature that silently never activates and no test would fail. This has already landed and been reverted once, it touches `RunView` for *every* entity, and the repo has a deterministic integration tier that this PR adds nothing to (no `TestingFramework` or test-bundle metadata in the diff). At least one integration bundle driving `DataSource:'Materialized'` end to end — and asserting it actually read the **snapshot**, not merely that the rows were correct — is the insurance this feature's history argues for.

### 5. A stale generated artifact, and the description contradicts its own diff

The body says "**Bootstrap class-registration manifests are intentionally untouched** by this PR." Both Angular manifests *are* touched:

- `Angular/Bootstrap` and `Angular/BootstrapLite`: `// @memberjunction/core-entities (400 classes)` → `(401 classes)`, adding `MJMaterializedResultEntity` to the imports and to `CLASS_REGISTRATIONS`.
- `MJMaterializedResultQueryEntity` is **absent from both** — even though both subclasses exist in `entity_subclasses.ts` (lines 90944 and 121077) and both form components are declared in `generated-forms.module.ts`.

400 → 401 for a PR that adds **two** entities is the tell: this looks like a leftover from the pre-join-table design (#3311), where the link was FK columns and only one entity was new. Both packages regenerate via `prebuild` (`mj codegen manifest`), so it self-heals at build time — but a half-updated generated file is noise at best, and if the generator genuinely can't see the join entity that's worth knowing, given the body itself reports that same generator silently dropping three `@memberjunction/server` classes over a wildcard-export chain. Either regenerate both manifests properly or revert both hunks, and correct the description. (The two ServerBootstrap manifests really are untouched, as claimed.)

---

## Missing / smaller

- **No `ON DELETE CASCADE`** on `FK_MaterializedResultQuery_Query` or `FK_MaterializedResultQuery_MaterializedResult`. Deleting a materialized Query will fail on the FK until the join row goes first. That may well be deliberate — don't orphan a snapshot — but I didn't find cleanup logic on the delete path, so it's worth confirming the failure is handled rather than surfaced as a raw FK error.
- `packages/Materialization` has no `README.md` and no `clean` script; its siblings have both.
- `queryHasTopLevelOrderBy` is `public static`, so an internal heuristic is now public API of `@memberjunction/generic-database-provider`. Consider narrowing it or marking it internal.
- The deleted `remove-materialization-remnants.md` had **empty frontmatter** (it bumped nothing), so removing it is genuinely harmless and the stated rationale is accurate — no concern there.
- Strike the "four enum columns lack `EntityFieldValue` rows" follow-up from the description (see above) so nobody chases it.

---

I did not run the full unit suites locally (CI's 50-minute run covers them and is green), and I did not attempt a live-DB reproduction of the materialized read path. Everything else above I checked against the branch at `0b05aa1`. If any finding is wrong, or already handled somewhere I didn't look, say so on the thread and I'll re-check rather than assume I was right.

rkihm-BC (commented):
## Re-review — Claude Code (pass 2, `0b05aa1` → `040bfcf`)

> Performed by **Claude Code** (Opus 5) at @rkihm-BC's request. This is a second pass over my [earlier review](https://github.com/MemberJunction/MJ/pull/3735#pullrequestreview-4910054153), re-derived on a local checkout at `040bfcfa1b`. I deliberately tried to **falsify** each of my own new findings rather than re-confirm them — two changed as a result, and I say so below. **Please push back on anything I got wrong.**

**Verdict: the four things I asked for are genuinely done, and done well.** The remaining items are one mechanical blocker, one type-safety regression in the new tests, and one mint/refresh asymmetry.

---

## Verified fixed since the last pass

| Prior finding | Status |
|---|---|
| **1. PG counterpart shouldn't be in this PR** | ✅ Deleted. `migrations-pg/v6` is back to 15 files, matching `next` — the out-of-order-insert hazard is gone. |
| **2. Refresh-path swap DDL hand-rolls quoting** | ✅ `assertSafeObjectNames` (`^[A-Za-z_][A-Za-z0-9_]*$`) runs at the `RefreshOne` entry **before any SQL executes** — I checked the placement, it sits immediately after the DriftHold/Disabled check and ahead of every builder dispatch (full, incremental, external) — plus defensively inside both full-rebuild builders. A passing name contains no `]`, `"`, or `'`, so the one check closes both the identifier-quoting and the `EXEC(…)`/`sp_rename` string-literal surfaces. 11 new unit cases cover all three break-out characters plus space and empty. |
| **4. No live-DB coverage for the read paths** | ✅ IT77 (MR1–MR3) + IT78 (EMR1–EMR2), wired into the deterministic suite and the `check-registry` coverage guard. The sentinel design is the right one — a row present **only** in the snapshot proves provenance rather than merely correct rows, and each has a `Live` negative control. `Sequence: 33` and the absent `primaryKey` match the existing convention (IT69–IT76 do the same), so no issue there. |
| **5. Stale manifests contradicting the description** | ✅ Both Angular manifests reverted to `400 classes` with `MJMaterializedResultEntity` removed. The body's "manifests intentionally untouched" claim is now true. |
| **`ON DELETE CASCADE` question** | ✅ Answered properly, and I verified both halves in the migration: the regenerated `spDeleteQuery` **does** cascade `MaterializedResultQuery`, and `spDeleteMaterializedResult` does not — which is exactly the reverse direction `MJMaterializedResultEntityServer` now covers. It follows the established sibling pattern (`MJTagEntityServer.Delete`) and is stricter, failing closed with a result-history entry rather than best-effort. MR3 gives it live coverage. |
| **`queryHasTopLevelOrderBy` as public API** | ✅ Marked `@internal`. Doc-only, but reasonable given it must stay `static` to be unit-tested. |
| **Missing README** | ✅ Added, and accurate to the code. |

**Correcting myself:** I flagged a missing `clean` script last time. `MJCore`, `GenericDatabaseProvider`, `SQLParser`, and `Scheduling/engine` don't define one either — that nit was wrong, please disregard it.

---

## Should be addressed before merge

### A. CI has not run on any of the six new commits

The branch's `actions/runs` list stops dead at `0b05aa1` (2026-08-11 16:14) with nine workflows green, then nothing. All six subsequent commits report `total_count: 0` check-runs and a combined status of `pending`; `mergeStateStatus` is `DIRTY`. `pull_request` workflows can't fire because GitHub can't materialize a merge ref.

`git merge-tree` gives exactly two content conflicts, both small:

```
metadata-optional/integration-test/test-suites/.integration-suite.json
packages/Scheduling/engine/package.json
```

The 13 green checks in this thread are from **before** the injection guard, `MJMaterializedResultEntityServer`, and both integration bundles existed. Five new files — including the ones added specifically to answer this review — have never been compiled or run by CI. Merging `next` is the thing that lets the deterministic tier actually validate the new work, so it's worth doing first rather than at the end.

### B. The `as never` casts in the new check files erase the type plumbing this PR exists to add

`materialized-read.checks.ts:207`, `:222`, and `materialized-entity-read.checks.ts:109`.

I went looking for a typing reason and there isn't one:

- `RunQueryParams` declares `QueryID?`, `Parameters?: Record<string, any>`, and `DataSource?: 'Live' | 'Materialized'` (`runQuery.ts:126`)
- `RunViewParams` declares `DataSource?: 'Live' | 'Materialized'` (`runView.ts:99`)
- `RenderedSQL` is on `RunQueryResult` (`interfaces.ts:1417`)
- `@memberjunction/core` resolves as a **workspace symlink**, not a registry copy, so the test package sees this branch's own types

The decisive evidence is the sibling bundle in the same package on the same transport — `runquery-features.checks.ts:139` calls `rq.RunQuery({ SQL: … }, ctx.User)` with **no cast at all** and reads `result.RowCount` directly. So this is a local deviation, not a workaround.

Why it matters beyond style: `as never` on the params literal disables type-checking of the **entire object**. The tests that exist to protect `DataSource` would still compile if `DataSource` were dropped from the params types in a future refactor. Same for the `(result as { RenderedSQL?: string })` casts. Dropping all of them should be a no-op.

### C. Mint accepts schema names that refresh then permanently rejects

`manage-metadata.ts:1411` sets `const matSchema = entity.SchemaName` — the **source entity's** schema — and mints through `generateMaterializedTableSQL` / `generateMaterializedWrapperViewSQL`, which quote at the dialect level and therefore accept any legal schema name. The new refresh guard then rejects anything outside `^[A-Za-z_][A-Za-z0-9_]*$`.

So an entity in a schema like `Sales Data` or `crm-prod` provisions cleanly and can **never** refresh — a permanent failure surfaced by the nightly job rather than at mint time. Fail-closed is the right instinct; the asymmetry between the two paths isn't. Either validate-and-escape on the refresh path (reject only `]`, `"`, `'`) or apply the same assertion at mint so provisioning refuses up front with a clear message.

---

## Small

- **Stale entity name in IT78's header comment** — it says the snapshot is fabricated for `Roles`, twice, but `TARGET_ENTITY = 'MJ: Task Types'` (`materialized-entity-read.checks.ts:29`). Cosmetic, but it's the file's orientation comment.
- **The guard is asymmetric across builders** — `buildFullRebuildStatements{SQLServer,PostgreSQL}` assert defensively; `buildExternalRebuildPlan`, `buildIncrementalMergeStatements*`, and `buildDirtyGroupRecompute*` don't, despite interpolating the same three names. I verified this is **not** a hole: `RefreshOne` is provably the only entry point (the single external caller anywhere in the repo is `MaterializationRefreshScheduledJobDriver.ts:78`) and it validates before any SQL runs. Purely a consistency nit — two of five builders carrying the belt-and-braces check invites a future caller to reach one of the other three directly.
- **The per-read metadata round-trips** (my finding #3 last time) are unchanged — still two uncached `RunView`s plus an un-memoized `Astify` per materialized `RunQuery`. Fine to defer, but please file it as an issue so it doesn't evaporate with the PR body on merge.

---

## Withdrawn from my own draft

I had written up MR3's fixture coupling as a finding — that MR3 deletes the `MaterializedResult` MR1/MR2 read through, leaving the three checks order-dependent with nothing enforcing it. **That was an over-call and I'm withdrawing it.** `GetBundle` returns `[...this.checks.values()]` from a `Map`, so registration order is preserved, and registration is a `for…of` over the array literal — MR1, MR2, MR3, deterministically. The driver brackets every bundle run's `Setup`/`Teardown` in one `try/finally`, so a re-run always gets a fresh fixture and MR1/MR2 cannot observe MR3's delete. It's a latent trap only for a future MR4 placed after MR3 — worth a comment line at most, not a change.

---

Net: **A** (mechanical, but it's what validates everything else), **B** (should be a no-op deletion of three casts), and **C** (fix or move the assertion to mint). The small items are one-liners. All four fixes from my last pass hold up under re-derivation — the guard placement in particular is correct, not merely present.

PR #3736, opened 2026-08-11
feat(realtime): let a voice agent speak first

### LINKED ISSUE #3557: ElevenLabs realtime: agent cannot speak first — firstMessage is not an enabled per-session override
## What

`ElevenLabsRealtime` provisions its managed agent with exactly two per-session overrides enabled:

```ts
// packages/AI/Providers/ElevenLabs/src/elevenLabsRealtime.ts
function buildRequiredOverrideEnablement(): ElevenLabs.ConversationConfigClientOverrideConfigInput {
    return {
        agent: { prompt: { prompt: true } },
        tts: { voiceId: true },
    };
}
```

`firstMessage` is not among them. As that file already documents, **ElevenLabs drops any `conversation_config_override` field the agent has not explicitly enabled** — so a caller supplying a first message gets no error and no greeting.

The consequence: **an ElevenLabs realtime agent can never speak first.** With no `first_message` configured, the agent waits for user audio before producing any, so the human must speak before the agent will.

## Why this is not fixable by prompting

Conversation-start behaviour is not instruction-following. The system prompt override lands correctly and the agent behaves properly *once started* — it simply produces nothing until it receives user audio. No wording in the persona prompt changes that.

## Where it hurts

Found while building Caliber's ATS voice screen (`bizapps-caliber`). A candidate opens their emailed interview link, the widget connects, and nothing happens until they guess that they should talk. For a hiring screen that is the first impression: silence where a professional greeting belongs. Every ElevenLabs realtime consumer has the same gap — this is not Caliber-specific.

The natural authoring shape on the Caliber side is a persona-level greeting resolved into the launch payload alongside `PreferredVoice`, but there is nothing to send it to today.

## Suggested fix

Three parts, and part 2 is the one that is easy to miss:

1. **Enable it on the managed agent** — add `agent: { firstMessage: true }` to `buildRequiredOverrideEnablement()`. That is what gets WRITTEN onto the agent.

2. **Require it in `OverridesSatisfied()`** — ⚠️ the file's own comment spells out why:

   > *"Adding an override here is only HALF the job: `OverridesSatisfied` must require it too, or an already-deployed agent still matches on tools, is never re-PATCHed, and silently drops the new override forever — precisely the defect behind issue #3374."*

   Without this, every agent already provisioned in the field keeps the old capability set and the feature silently does nothing for existing deployments — a repeat of #3374.

3. **Plumb `firstMessage` through the realtime config** so a caller can supply one per session, next to the existing prompt and voice overrides. The round-trip test named in that comment ("considers an agent this driver just created satisfied") should cover the new field so the two halves cannot drift apart again.

## Acceptance

- [ ] A realtime session started with a `firstMessage` has the agent speak first, with no user audio
- [ ] An agent provisioned BEFORE this change is re-PATCHed on next use rather than silently dropping the override (the #3374 regression)
- [ ] Omitting `firstMessage` preserves today's wait-for-user behaviour
- [ ] Round-trip test extended to the new override

## Refs

- `packages/AI/Providers/ElevenLabs/src/elevenLabsRealtime.ts` — `buildRequiredOverrideEnablement`, `OverridesSatisfied`, `buildAgentBody`
- Issue #3374 — the prior instance of the half-done-override failure mode

SDesai-BC:
An ElevenLabs realtime agent could never open a conversation. The widget connects, and nothing happens until the human guesses they should talk — on a hiring screen, silence where a professional greeting belongs.

Closes #3557.

## Root cause

ElevenLabs **drops any `conversation_config_override` field the managed agent has not explicitly enabled**, and the driver enabled exactly two — `elevenLabsRealtime.ts:68`:

```ts
function buildRequiredOverrideEnablement(): ElevenLabs.ConversationConfigClientOverrideConfigInput {
    return {
        agent: { prompt: { prompt: true } },
        tts: { voiceId: true },
    };
}
```

`firstMessage` is not among them, so a caller supplying one got no error and no greeting. With no `first_message` set the platform produces no audio until it receives user audio — the SDK's own type says so: *"If non-empty, the first message the agent will say. If empty, the agent waits for the user to start the discussion."* This is why no amount of prompt wording fixed it: conversation-start behaviour is not instruction-following.

## There were two causes, not one

Enabling the override makes the *driver* capable. It does not make the feature *reachable*, and that second gap is easy to miss because the first one fully explains the symptom.

The driver reads the first message from `RealtimeSessionParams.Config`. That bag is populated by `RealtimeClientSessionService.buildSessionConfigBag`, whose only external input is `PrepareClientSessionInput.Config` — and **the GraphQL resolver never sets it**. `RealtimeClientSessionResolver.ts:1635` passes 13 keys to `PrepareClientSession`; `Config` is not one of them. The only route in from outside is `ConfigOverridesJson`, which is gated behind the `Realtime: Advanced Session Controls` authorization, and the agent-type `ConfigSchema` (`additionalProperties: false`) rejects an unknown key for metadata authoring.

So a persona-level greeting had nowhere to land. Both halves are in this PR.

## What changed

**The override** — `agent.firstMessage` is enabled on the managed agent, and `OverridesSatisfied` (`:723`) now *requires* it. That second half is the whole point: without it an already-deployed agent still matches on tools, is never re-PATCHed, and silently drops the new override forever — precisely the defect behind #3374. The two halves are pinned together by an existing test that derives its assertions from what the driver actually wrote, so adding an override to one side only fails as a permanent re-PATCH loop.

**The wire frame** — `BuildSessionOverrides` (`:446`) emits `agent.first_message` beside `prompt.prompt`. Both live *inside* `agent`, so the object is built once and added to; assigning `overrides['agent']` a second time would have dropped the prompt. Both topologies inherit it from that one edit, because the client driver forwards the override object verbatim.

**The authoring slot** — `realtime.voice.default.firstMessage`, resolved exactly like the provider-agnostic `voice`: filed onto whichever driver the framework picks, winning its own key over a matching `providers.<key>` entry. It reaches both host paths, since `BaseAgent` and `RealtimeClientSessionService` both go through `GetProviderVoiceSettings`.

**AssemblyAI** honours the same neutral key for its `greeting` wire slot, so one authored value means the same thing whichever vendor runs.

## Decisions worth defending

**Why `firstMessage` and not `greeting`.** AssemblyAI already had a driver-neutral `greeting` key for this concept, so `greeting` was the incumbent name. `firstMessage` was chosen deliberately (it is the issue's wording and ElevenLabs' own field name) and AssemblyAI was taught to accept it, rather than leaving one concept with two names across two drivers. The legacy `greeting` key still works.

**Blank means absent, not empty.** Both drivers omit the key entirely for a blank, whitespace-only, or non-string value. An enabled-but-empty `first_message` *is* the platform's wait-for-the-user default, so sending it blank and omitting it are the same behaviour — and omitting keeps the frame honest. This also means a blank `firstMessage` cannot suppress a valid legacy `greeting` on AssemblyAI, which an earlier revision of this branch got wrong.

**Omission is byte-for-byte the old frame**, which is what preserves today's behaviour by construction rather than by a compatibility branch.

## Evidence

| Suite | Result |
|---|---|
| `@memberjunction/ai-elevenlabs` | build clean, **85/85** |
| `@memberjunction/ai-assemblyai` | build clean, **48/48** |
| `@memberjunction/ai-agents` | build clean, **1872/1872** |
| `Integration Tests — Deterministic` | **75/78**, run twice (~1030s / ~1090s), identical failures both times |

The three integration failures are pre-existing on this scratch database and unrelated to this diff — none touches the realtime surface:

- `entity-writes.EW8` — `INSERT` conflicts with `FK_TagScope_Tag`; the fixture's `__mj.Tag` row does not exist in this database, so the expected validation refusal surfaces as a raw SQL error instead.
- `workflow-demo-agents.WD2` — `dc49d7c830` on `next` made the compiler refuse a second starting step; the `Content Pipeline` demo workflow in this database still has two.
- `subscription-isolation.SI1` — `No acting user available for operation 'RecordProcess.RunNow'`. Localised to RecordProcess acting-user resolution, **not root-caused**.

Each reproduces standalone in under a second, so they are deterministic rather than flaky. I did not run the suite against a clean base, so "pre-existing" rests on the causes above rather than on a control run.

## Deploy notes

**The metadata change is not live on merge.** `metadata/agent-types/schemas/realtime-type-config.schema.json` is declarative only — the new `firstMessage` property is not in any database until `mj sync push` runs, which is the build engineer's release-time step (one consolidated sync migration per release), not a per-PR one. Until that push lands in an environment, authoring `realtime.voice.default.firstMessage` in agent metadata will fail `ConfigSchema` validation on save there. The code path itself is live immediately, so programmatic hosts and `ConfigOverridesJson` work without the push.

**Every existing managed ElevenLabs agent gets one PATCH on first use after deploy.** That is the intended repair — it is what stops the override being silently dropped for existing deployments — but it does mean one extra REST write per managed agent name, once.

**One small behavioural delta on AssemblyAI:** a config carrying `greeting: ""` previously set `session.greeting = ""` and now omits the field. Same effective outcome (no opening utterance), different frame.

No migration, no CodeGen, no generated files.


packages/AI/Agents/src/realtime/realtime-coagent-config.ts
```diff
@@ -1028,11 +1075,34 @@ export function GetProviderVoiceSettings(
...
+ * @returns The overlay bag, or `null` when the persona contributes no wire-level setting — which is
+ *   what keeps a config authoring none of them byte-for-byte identical to its pre-existing behavior.
+ */
+function collectAgnosticWireSettings(persona: RealtimeVoicePersona | undefined): JSONObjectLike | null {
+    const settings: JSONObjectLike = {};
+    if (persona?.voice) {
+        settings['voice'] = persona.voice;
+    }
+    if (persona?.firstMessage) {
+        settings['firstMessage'] = persona.firstMessage;
```
rkihm-BC, line 1103: **This is where the leak starts — `firstMessage` is filed onto *every* driver, but was never registered as a shared neutral key.**

`REALTIME_SHARED_CONFIG_KEYS` in `packages/AI/Core/src/generic/baseRealtime.ts` is the registry that makes a neutral MJ key safe across the driver family — each driver either translates it or scrubs it. `voice` (#3530) is in that list, which is why it is handled deliberately everywhere. `firstMessage` is not, and this PR does not touch that file.

Because `GetProviderVoiceSettings` files agnostic slots onto whichever driver resolves, the key now lands in every driver's config bag. I probed the real `OpenAIRealtime` driver through its own test harness:

```
PROBE session.firstMessage = "Hi there."
PROBE full session keys = ["type","instructions","audio","firstMessage"]
```

It survives `ExtractRealtimeFeatures` into `features.rest` and is spread into the `session.update` payload — and the same spread builds the minted client-direct session, so **both topologies** carry it. `xAIRealtime` inherits `OpenAIRealtime`, so it does too.

Per-driver exposure:

| Driver | What happens |
|---|---|
| **OpenAI / xAI** | **Sent to the wire** as an unknown top-level session field, both topologies |
| **Inworld** | `applyRawConfigOverrides` copies every key not in `consumedShorthands` → `session.firstMessage` sent raw |
| **Gemini** | Not scrubbed → spread onto `LiveConnectConfig` → silently dropped by the SDK path allowlist (harmless, but exactly the #3721 class) |
| **HuggingFace** | Safe — builds a fixed session shape, reads only `voice`/transcription |

This also breaks a contract the OpenAI tests already assert for every other MJ-vocabulary key — `reasoningEffort`, `parallelToolCalls` and `mcpTools` each carry an explicit `expect(session.<rawKey>).toBeUndefined()`. And `openAIRealtime.ts` warns in its own comments that a malformed session object "makes strict endpoints reject the WHOLE session.update, silently dropping the prompt AND tools" — so the team already treats this endpoint as strict about shape.

**Fix:** add `'firstMessage'` to `REALTIME_SHARED_CONFIG_KEYS`. That gets it deleted from OpenAI's residual bag alongside the other neutral keys and scrubbed by Gemini's loop, and it is the registration that makes the next neutral key safe by default rather than by luck.

**Also missing, and this is what let it through:** there is no test asserting a *non-consuming* driver is unaffected. The two consuming drivers and the agents-layer resolution are covered thoroughly; the third case isn't covered at all. One test per non-consuming driver in the existing `expect(session.<rawKey>).toBeUndefined()` style would pin it.

guides/REALTIME_CO_AGENTS_GUIDE.md
```diff
@@ -299,8 +301,9 @@ The pure implementation is `packages/AI/Agents/src/realtime/realtime-coagent-con
...
+- **`voice.default.firstMessage`** is the **provider-agnostic opening utterance** (#3557) — what the agent says FIRST, verbatim, before the user has spoken. It resolves exactly like the agnostic voice (filed onto whichever driver the framework picks, winning the `firstMessage` key over a matching `voice.providers.<key>` entry). Authored here rather than in the persona prompt because conversation-start behaviour is **not instruction-following**: ElevenLabs with no `first_message` produces no audio at all until it receives user audio, no matter how the prompt is worded — which is why a co-agent could never greet a caller. Omit it and the agent waits for the user, as before. Honoured today by **ElevenLabs** (`agent.first_message` override) and **AssemblyAI** (`greeting`); OpenAI and Gemini read the `firstMessage` key nowhere and still open silently.
```
rkihm-BC: > OpenAI and Gemini read the `firstMessage` key nowhere and still open silently.

The first half is true; the second isn't, for OpenAI. "Unread" is not the same as "harmless" — see the detail on `collectAgnosticWireSettings`. OpenAI receives the key as an unknown top-level session field on both topologies (I reproduced it), and Inworld — not mentioned here — copies it raw onto its session object.

Gemini genuinely does open silently: the key is dropped by the SDK's path allowlist before the wire. So Gemini is the only vendor this sentence describes correctly.

Once `firstMessage` is added to `REALTIME_SHARED_CONFIG_KEYS`, the sentence becomes true as written for every driver — which is the cleanest way to resolve this, rather than rewording it to document a leak.

guides/REALTIME_CO_AGENTS_GUIDE.md
```diff
@@ -101,7 +101,8 @@ The base contracts are deliberately tolerant of provider asymmetry: several memb
...
-| Server-side object to manage | None | None | **Managed agent** — `ensureAgent()`: find-by-name → create-if-missing → PATCH on tool-fingerprint drift / missing prompt-override enablement; cached per name + fingerprint | None — pure per-session config |
+| Server-side object to manage | None | None | **Managed agent** — `ensureAgent()`: find-by-name → create-if-missing → PATCH on tool-fingerprint drift / any missing per-session override enablement; cached per name + fingerprint | None — pure per-session config |
+| Can the agent **speak first**? | Not wired — the neutral `firstMessage` key is unread today | Not wired — the neutral `firstMessage` key is unread today | **Yes** — `agent.first_message` override (enabled on the managed agent). Without it the agent produces NO audio until it hears the user, whatever the prompt says (#3557) | **Yes** — the `greeting` field of the session object (fed by the neutral `firstMessage` key; the legacy `greeting` config key still works) |
```
rkihm-BC: Good addition — this row is exactly the kind of thing the capability matrix is for, and "Without it the agent produces NO audio until it hears the user, whatever the prompt says" is the single most useful sentence in the PR for a future driver author.

One correction: **"Not wired — the neutral `firstMessage` key is unread today"** understates what happens on OpenAI. Unread, yes; inert, no — it currently reaches the wire (detail on `collectAgnosticWireSettings`). Worth either fixing the underlying scrub so "not wired" is literally true, or saying "ignored by the driver, scrubbed before the wire" once it is.

.claude/rules/changesets.md
```diff
@@ -0,0 +1,84 @@
...
+Loads when you open anything under `.changeset/`. The rule is short; the reason it needs its own
+file is that **you cannot infer it from the neighbouring files**.
+
+---
+
+## The rule
+
+| Level | When |
+|---|---|
+| `minor` | The branch adds a **migration** (`migrations/vN/*.sql`) **or** changes **`metadata/`** |
```
rkihm-BC: **This is a repo-wide versioning policy riding on a voice-agent feature branch, and it re-grades this PR's own bump level.**

The rule changes, not just documents:

- **Old** (`.claude/commands/changeset.md`, this PR's own diff): `minor` **only** when the branch adds a migration under `migrations/v*/` — and applied to `@memberjunction/core` only.
- **New:** `minor` when the branch adds a migration **or changes anything under `metadata/`**, applied to the packages the branch changed.

This PR changes exactly one metadata file and no migration. Under the old rule it is a `patch`; under the new rule it is a `minor` — and the changeset-bot comment shows the consequence: **301 packages, all Minor**. The rule file then cites that same 301-row table as its evidence.

To be clear, I don't think `minor` is the wrong *outcome* — this adds real public API (`ResolveFirstMessage` on two drivers, `RealtimeVoicePersona.firstMessage`), so ordinary semver would say minor regardless. That's the point: the outcome may be right for a reason this rule doesn't give. Under the rule as written, a 2,000-line feature adding new public API is a `patch` if it touches no metadata, while a one-line metadata typo fix is a `minor` across all 301 packages.

That is a defensible release-train convention given the `fixed` group — MJ versions are a train, not per-package semver. But it is a **project-wide standard**, and it deserves its own PR and explicit sign-off rather than arriving as a rider whose first act is to re-grade the branch carrying it. Note also that it cuts against the direction reviewers have pushed elsewhere (on #3715 I argued the opposite: `patch` understated a changeset that added exported API).

The supporting facts all check out, for what it's worth: `metadata/CLAUDE.md` §1b does say metadata edits become one consolidated migration at release, the gate runs clean here, its 9 tests pass, `check:claude-md` passes, and the frontmatter matches the other rule files. It's good work — just not this PR's work.

.github/scripts/check-changeset-bump.mjs
```diff
@@ -0,0 +1,151 @@
...
+    if (malformed) {
+        violations.push({ path, reason: `could not read bump entries: ${malformed}` });
+        continue;
+    }
+    for (const { pkg, level } of entries) {
+        if (!LEVELS.includes(level)) {
+            violations.push({ path, reason: `"${pkg}": ${level} is not one of ${LEVELS.join(' / ')}` });
+        } else if (level === 'major') {
+            violations.push({ path, reason: `"${pkg}": major — never use without explicit approval` });
+        } else if (level === 'minor' && triggers.length === 0) {
```
rkihm-BC: **The gate only enforces one half of the rule it states.**

The rule is written as an equivalence — `minor` ⇔ migration or metadata — but this only flags `minor` when there is no DB trigger. Nothing flags `patch` when a migration or metadata change **is** present, so a branch that adds a migration and marks `patch` passes clean.

Given the `fixed` group that only bites when no other changeset in the release is `minor` — but that is the direction where a real database change ships under-versioned, which is the worse of the two failures. The over-bump this gate does catch costs a version number; the under-bump it doesn't catch costs a correct release.

Worth adding the mirror check, or softening the rule's phrasing so the gate isn't narrower than the rule it points to.

Separately, the scoping decision here is right and well-argued — judging only `--diff-filter=A` changesets against the merge-base is what makes this switchable-on without a repo-wide cleanup, and the test that pins it (`ignores pending changesets contributed by OTHER branches`) is the one that matters most.

packages/AI/Providers/ElevenLabs/src/elevenLabsRealtime.ts
```diff
@@ -672,7 +722,11 @@ export class ElevenLabsRealtime extends BaseRealtimeModel {
      */
     public static OverridesSatisfied(agent: ElevenLabs.GetAgentResponseModel): boolean {
         const overrides = agent.platformSettings?.overrides?.conversationConfigOverride;
-        return overrides?.agent?.prompt?.prompt === true && overrides?.tts?.voiceId === true;
+        return (
+            overrides?.agent?.prompt?.prompt === true &&
+            overrides?.agent?.firstMessage === true &&
```
rkihm-BC, line 727: This is the decision the whole fix rests on, and it's right. Without requiring the new override here, every already-provisioned agent keeps matching on tools, is never re-PATCHed, and silently drops `first_message` forever — the #3374 failure mode, which is easy to reproduce and very hard to notice.

The accompanying test is the right shape too: it provisions an agent with `firstMessageOverrideEnabled: false` and asserts both that the repair happens **and** that `prompt.prompt` / `tts.voiceId` survive it. Repair tests that only check the new thing are how you lose the old ones.

The one-PATCH-per-managed-agent cost is correctly disclosed in the PR body, and it's genuinely once — the agent state is server-side, so the next session sees it satisfied.

packages/AI/Providers/ElevenLabs/src/elevenLabsRealtime.ts
```diff
@@ -399,24 +422,36 @@ export class ElevenLabsRealtime extends BaseRealtimeModel {
...
      * @returns The wire-shaped override object.
      */
     public static BuildSessionOverrides(systemPrompt: string, config?: JSONObject): JSONObject {
-        const overrides: JSONObject = { agent: { prompt: { prompt: systemPrompt } } };
+        // `first_message` is a SIBLING of `prompt` inside `agent`, so the agent object is built
+        // once and added to — assigning `overrides['agent']` a second time would drop the prompt.
+        const agent: JSONObject = { prompt: { prompt: systemPrompt } };
+        const firstMessage = ElevenLabsRealtime.ResolveFirstMessage(config);
+        if (firstMessage) {
+            agent['first_message'] = firstMessage;
```
rkihm-BC, line 452: Good catch on the sibling merge — `first_message` and `prompt` both live inside `agent`, so building the object once and adding to it is correct, and a second `overrides['agent'] = …` would have silently dropped the prompt.

Worth noting how well the test guards this: `promptAndFirstMessageOverrides` is deliberately written as an explicit literal rather than a spread of `promptOnlyOverrides`, with a comment saying a shallow spread is precisely the mistake being guarded. A helper that reproduced the bug would have made the test agree with broken code — that it doesn't is the reason this one will still be doing its job in a year.

packages/AI/Providers/AssemblyAI/src/assemblyAIRealtime.ts
```diff
@@ -248,6 +266,24 @@ export class AssemblyAIRealtime extends BaseRealtimeModel {
...
+     * Both keys go through the SAME trim-and-drop-blank rule the ElevenLabs driver applies, so one
+     * authored value means the same thing on either provider. That rule is also why the neutral key
+     * only wins when it carries something: a blank `firstMessage` must not suppress a valid legacy
+     * `greeting` — it is an absent opening utterance, not an empty one.
+     *
+     * @param config The session's open config bag.
+     * @returns The trimmed opening utterance, or `undefined` when neither key carries one.
+     */
+    public static ResolveFirstMessage(config: JSONObject): string | undefined {
+        return resolveTrimmedString(config['firstMessage']) ?? resolveTrimmedString(config['greeting']);
```
rkihm-BC, line 284: The `??` here is doing more work than it looks like, and correctly. Because `resolveTrimmedString` collapses blank to `undefined`, a blank `firstMessage` falls through to a valid legacy `greeting` rather than clobbering it — which is the right reading of "blank means no opening utterance authored, not an empty one."

Teaching this driver the neutral name rather than leaving one concept with two names across two drivers is the right call, and keeping `greeting` working means no existing config breaks.

One genuine behavioural delta, correctly disclosed in the PR body: `greeting: ""` previously set `session.greeting = ""` and now omits the field. Same effective outcome, different frame — worth keeping in the changeset (it is) in case anyone is asserting on the exact session object.

package.json
```diff
@@ -113,9 +113,11 @@
     "test:analytics": "node unit-testing/tools/run-all-tests.js",
     "test:aggregate": "node unit-testing/tools/aggregate-results.js ",
     "test:markdown": "node unit-testing/tools/generate-markdown.js ",
-    "test:integration": "MJ_INTEGRATION_TEST=1 mj test suite \"Integration Tests \u2014 Deterministic\"",
+    "test:integration": "MJ_INTEGRATION_TEST=1 mj test suite \"Integration Tests — Deterministic\"",
     "test:e2e": "playwright test --config e2e/playwright.config.ts",
-    "check:standards": "mj standards check"
+    "check:standards": "mj standards check",
+    "check:changeset": "node .github/scripts/check-changeset-bump.mjs",
```
rkihm-BC: Two notes on this file, both minor.

**`check:changeset:test`** points at `.github/scripts/vitest.config.mts` — I checked, it already exists, and the suite runs clean (9/9). No issue; flagging only because a missing config there would have been an easy way to ship a dead script.

**Unrelated diff noise:** this PR also rewrites `\u274c` → `❌` in `prebuild` and `\u2014` → `—` in `test:integration`. Both are byte-equivalent after JSON parsing, so nothing breaks — but they read like an editor artifact rather than a decision, in a file nobody expects a voice-agent PR to touch. Worth reverting to keep the diff honest.

rkihm-BC (dismissed):
The root-cause analysis here is the best part of the PR, and the ElevenLabs/AssemblyAI work is solid. Two things I'd want resolved before merge — one is a wire-level defect I reproduced, the other is scope.

**What's genuinely good**, verified rather than assumed:

- **The "two causes, not one" framing.** Recognising that enabling the override makes the driver *capable* but not the feature *reachable* — and that the first cause fully explains the symptom, so the second is easy to miss — is what separates this from a fix that would have shipped half-working.
- **Requiring `firstMessage` in `OverridesSatisfied` is the load-bearing decision**, and the test that provisions an agent with it disabled and asserts both the repair *and* that pre-existing enablements survive is exactly the right shape.
- **The casing split is handled and documented** — I confirmed `AgentConfigOverrideConfig.firstMessage?: boolean` exists in @elevenlabs/elevenlabs-js 2.34.0, and both changed packages typecheck clean.
- **Blank-is-absent is correct and followed through**, including the subtle case that a blank `firstMessage` must not suppress a valid legacy `greeting`.
- **Evidence reproduces**: ElevenLabs **85/85**, AssemblyAI **48/48**, `check:claude-md` passes, the new gate's 9 tests pass, and the gate runs clean on this branch. (`ai-agents` gave me 57 failures, but every one is `X is not a function` from stale `dist` in my symlinked environment — not attributable to this branch. The 1872 total matches exactly.)

---

### 1. Blocking: the neutral `firstMessage` key reaches the OpenAI wire

`firstMessage` was added as an agnostic persona slot but never registered in `REALTIME_SHARED_CONFIG_KEYS` (`packages/AI/Core/src/generic/baseRealtime.ts`, untouched by this PR). That list is what makes a neutral MJ key safe across the family — it is why `voice` from #3530 is deliberately consumed or scrubbed by every driver.

I probed the real `OpenAIRealtime` driver:

```
PROBE session.firstMessage = "Hi there."
PROBE full session keys = ["type","instructions","audio","firstMessage"]
```

It rides `...features.rest` into the `session.update` payload, and the same spread is used in `CreateClientSession` — so **both topologies**. `xAIRealtime` inherits it. Details and the per-driver table are inline on `collectAgnosticWireSettings`.

### 2. An unrelated repo-wide versioning policy is bundled in

Seven of the 19 files are a changeset bump-level policy — new script, its tests, a new rule file, a rewritten command doc, a `CLAUDE.md` routing row, two `package.json` scripts. None of it relates to voice agents, and it changes the rule in a way that decides this PR's own bump level. Detail inline on `.claude/rules/changesets.md`.

---

### Suggested split

1. **This PR** — the feature, plus `firstMessage` in `REALTIME_SHARED_CONFIG_KEYS`, the cross-driver tests, and corrected guide text.
2. **A separate PR** — the changeset policy and gate, where metadata-implies-minor can be weighed on its own merits.

Requesting changes on the strength of #1 alone; an unknown key on a strict endpoint is a bad failure mode and it lands on the provider most sessions actually use. Everything else here is good work.

<sub>Scope of what I checked: probed the OpenAI driver directly; read the residual-bag handling in all six realtime drivers; ran the ElevenLabs, AssemblyAI and ai-agents suites plus both typechecks; ran the new gate and its tests; ran `check:claude-md`; verified the ElevenLabs SDK override types and the `metadata/CLAUDE.md` §1b citation. I did **not** run a workspace build or the integration tier, and I have no ElevenLabs/AssemblyAI keys, so the live wire behaviour is unverified on my side.</sub>

packages/AI/Agents/src/realtime/realtime-coagent-config.ts
SDesai-BC: Confirmed and fixed — thank you, this was a real leak and I reproduced it before changing anything:

```
ExtractRealtimeFeatures({firstMessage:"Hi there.", voice:"alloy"})
→ rest = {"firstMessage":"Hi there."}      // voice scrubbed, firstMessage not
```

**One correction, because it changes the fix.** Adding `firstMessage` to `REALTIME_SHARED_CONFIG_KEYS` does *not* get it deleted from OpenAI's residual bag — `openAIRealtime.ts` imports that constant **zero** times and enumerates its deletes by hand. Only Gemini reads the list. Following that step alone would have fixed Gemini, left the OpenAI wire leak intact, and made the guide sentence *look* justified. Three changes were needed:

1. **register** `firstMessage` in `REALTIME_SHARED_CONFIG_KEYS` — fixes Gemini through its existing scrub loop;
2. **delete it in `ExtractRealtimeFeatures`** — fixes OpenAI + xAI. One delete covers both topologies, since `features.rest` is spread at both call sites;
3. **skip the shared vocabulary in Inworld's `applyRawConfigOverrides`** — implementing the contract that registry already documents, which also closes the same leak there for `reasoningEffort` / `mcpTools` / the rest.

On the missing test, you were right that it is what let this through, and it needed splitting rather than just adding. The driver suites `vi.mock('@memberjunction/ai')`, so they cannot see the real registry — I tried `importOriginal` and it cascaded into `BaseSingleton` on the mocked `@memberjunction/global`, which is why that mock exists. So:

- **Driver suites pin the mechanism.** OpenAI asserts the raw key is absent from both the `session.update` and the minted client-direct session, in the existing `expect(session.<rawKey>).toBeUndefined()` style. Inworld pins "skips whatever the registry holds" plus a guard that native passthrough keys still survive. Gemini pins its scrub.
- **`@memberjunction/ai`'s own suite pins membership** — that `firstMessage` is in the shared vocabulary at all. Neither half can drift without the other failing, and the membership half is the one that was missing.

Each was red first; I stashed the registration to confirm the Gemini test genuinely failed without it.

guides/REALTIME_CO_AGENTS_GUIDE.md
SDesai-BC: Fixed by scrubbing rather than rewording, as you suggested. It now reads:

> Honoured today by **ElevenLabs** (`agent.first_message` override) and **AssemblyAI** (`greeting`). Every other driver **scrubs** it and opens silently: it is registered in `REALTIME_SHARED_CONFIG_KEYS`, which is what makes a neutral key safe to file onto a vendor that has no use for it — unregistered, it survives each driver's residual-bag spread and reaches the provider as an unknown session field.

That also covers Inworld, which the old sentence omitted entirely.

guides/REALTIME_CO_AGENTS_GUIDE.md
SDesai-BC: Fixed. The row now reads "ignored by the driver and scrubbed before the wire" for OpenAI and Gemini, which is literally true once the scrub is in. Agreed that fixing the behaviour beats documenting the leak.

.github/scripts/check-changeset-bump.mjs
SDesai-BC: Confirmed and fixed — you were right that the gate was narrower than the rule it pointed at, and right about which direction matters. A branch adding a migration with an all-`patch` changeset passed clean; I reproduced that as a failing test before changing the script.

The gate now checks both directions: a DB branch with no `minor` entry anywhere fails. One `minor` is enough rather than requiring it on every entry — the `fixed` group moves every package to the highest bump regardless, so demanding it per-entry would be noise. Pinned by three cases (migration+patch, metadata+patch, and mixed-levels-passes). The rule file states the equivalence explicitly now, including why under-bumping is the worse failure.

package.json
SDesai-BC: Both confirmed. The `❌` → `❌` and `—` → `—` rewrites were mine — collateral from editing `package.json` through `node -e` + `JSON.stringify`, which un-escapes on the way out. Reverted; the diff is now the two `check:changeset` script lines and nothing else. Good catch, that is exactly the kind of noise that makes a diff harder to trust.

.claude/rules/changesets.md
SDesai-BC: Your read is fair and I am not going to argue the scope point — it is seven files of repo-wide versioning policy on a voice-agent branch, and it does re-grade the branch carrying it. Flagged to the author for a decision on splitting it into its own PR; I have not unpicked it unilaterally since that is a call about how the team wants to sequence it, not a defect.

Two things worth recording while it is open, both of which support your framing:

- **The rule and practice already disagree, independently of this PR.** 37 of 84 pending changesets use `minor`, across ~12 feature packages (`ai-core-plus` 9, `task-graph` 8, `server` 8, `ng-conversations` 6); `@memberjunction/core` gets it in 7. So whichever way this lands, something needs reconciling — the current documented rule is not what the repo does.
- **Your #3715 position is the direct counter-example**, and it is a good one: under the rule as written a 2,000-line change adding exported API is a `patch` if it touches no metadata. That is a release-train convention rather than semver, and you are right that adopting it deserves explicit sign-off rather than arriving as a rider.

The one piece I did action here is the half you flagged separately — the gate only enforced the over-bump direction. That is fixed regardless of where the policy ends up living.

SDesai-BC:
Thanks @rkihm-BC — the blocking finding was real and I have fixed it. Replies are on each thread; summary and the one place I diverged from your prescription:

## Blocking (#1) — fixed, but not by the one-line change

Reproduced before touching anything: `ExtractRealtimeFeatures({firstMessage:"Hi", voice:"alloy"})` → `rest = {"firstMessage":"Hi"}`. `voice` scrubbed, `firstMessage` not, and `features.rest` is spread at both call sites.

**Adding it to `REALTIME_SHARED_CONFIG_KEYS` alone would not have fixed OpenAI.** That file imports the constant zero times and enumerates its deletes by hand — only Gemini reads the list. Doing just that would have fixed Gemini, left the wire leak, and made the guide sentence look justified. Three changes:

| Change | Fixes |
|---|---|
| Register `firstMessage` in `REALTIME_SHARED_CONFIG_KEYS` | Gemini, via its existing scrub loop |
| `delete rest.firstMessage` in `ExtractRealtimeFeatures` | OpenAI + xAI, both topologies |
| Skip the shared vocabulary in Inworld’s `applyRawConfigOverrides` | Inworld — and the same leak there for `reasoningEffort`/`mcpTools` |

On the missing test: the driver suites mock `@memberjunction/ai` and cannot see the real registry (`importOriginal` cascades into `BaseSingleton` on the mocked `@memberjunction/global`). So driver suites pin the **mechanism**, and `@memberjunction/ai`’s own suite pins **membership** — the half that was missing is exactly the half that let this ship. Each fix red first, including stashing the registration to confirm the Gemini test failed without it.

## The rest

- **Guide text (#2, #3)** — fixed by scrubbing rather than rewording, as you suggested; the new sentence also covers Inworld, which the old one omitted.
- **Gate was one-sided (#5)** — confirmed, reproduced as a failing test, now checks the under-bump direction too. One `minor` anywhere suffices, since the `fixed` group moves everything to the highest bump.
- **`package.json` noise (#6)** — mine, from editing it via `node -e` + `JSON.stringify`. Reverted.
- **Scope (#4)** — your read is fair; flagged to the author for a decision on splitting rather than unpicked unilaterally. Worth noting the rule and practice already disagree: 37 of 84 pending changesets use `minor` across ~12 feature packages, so something needs reconciling either way.

## Verification

Core **247**, OpenAI **179**, Gemini **106**, Inworld **63**, ElevenLabs **85**, AssemblyAI **48**, Agents **1872** — all passing, zero TS errors, re-run after the incoming `next` merge (78 files, no conflicts).

Your `ai-agents` 57 failures were the stale-`dist` artifact you suspected — I hit the identical signature on Inworld and it was a mock missing a new export, not the branch.

Re-running the live ElevenLabs smoke suite now that the wire path changed, and will post the result.

SDesai-BC:
## Live verification — and @rkihm-BC's severity call was right

Re-ran the ElevenLabs suite against the real API after the scrub changes (**16/16**), and added a live check of the blocking finding on OpenAI, which the review could not run without keys.

**The leak was not cosmetic. The endpoint rejects it outright:**

```
clean session (what we send today):        ✓ accepted
+ stray firstMessage (the pre-fix shape):  ✗ REJECTED  400 Unknown parameter: 'session.firstMessage'.
```

Same payload the pre-fix driver built, sent straight at the client-secret mint so nothing depended on which build was on disk. So a persona authoring `realtime.voice.default.firstMessage` on an OpenAI-backed co-agent would have **failed every client-direct session mint with a 400** — not degraded, not ignored. The review's "strict endpoints reject the WHOLE session" reasoning holds literally.

Worth recording how close I came to reporting the opposite: my first counterfactual reverted the scrub and still passed, which looked like evidence the endpoint tolerated the field. It was a stale `dist` — the revert had not been rebuilt, so it was testing the fixed code. Sending the payload directly, independent of any build, is what settled it.

**Post-fix, live:**

| Check | Result |
|---|---|
| `CreateClientSession` with `firstMessage` in Config | accepted; minted session keys are exactly `["type","model","instructions","audio"]` |
| The consumed neutral key still applies | `voice: alloy` lands in the audio block |
| ElevenLabs suite (16 checks, real agents + websockets) | 16/16 — agent speaks first, silent when omitted, blank behaves as absent, `first_message`+`voice_id` accepted together |

Unit tiers after the incoming `next` merge: Core **247**, OpenAI **179**, Gemini **106**, Inworld **63**, ElevenLabs **85**, AssemblyAI **48**, Agents **1872**.

SDesai-BC:
## Scope split done — and re-requesting review

@rkihm-BC the changeset policy is out, into #3739. This PR is now 20 files, all voice-agent:

| Removed from here → #3739 | |
|---|---|
| `.claude/rules/changesets.md` | new rule file |
| `.claude/commands/changeset.md` | rewritten versioning rules — restored to `next` here |
| `.github/scripts/check-changeset-bump.mjs` + its 12 tests | the self-check |
| `CLAUDE.md`, `package.json` | routing row, mirrors line, 2 npm scripts — both restored to `next` here |

It moved unchanged, including the mirror check that closed your separate finding about the gate enforcing only one direction of its own rule.

**This PR keeps `minor`, on grounds that do not depend on the rule under discussion.** As you noted, it adds real exported API — `ResolveFirstMessage` on two drivers, `RealtimeVoicePersona.firstMessage`, and `firstMessage` in `REALTIME_SHARED_CONFIG_KEYS` — so ordinary semver says minor however the metadata question settles in #3739.

### All five actionable findings resolved

| # | Finding | Resolution |
|---|---|---|
| 1 | `firstMessage` reaches the OpenAI wire | Fixed in three places — registration alone was not enough, see thread |
| 2 | Guide: "unread, opens silently" false for OpenAI | Fixed by scrubbing, not rewording |
| 3 | Matrix row understates | Now "ignored and scrubbed before the wire" |
| 5 | Gate enforced one direction only | Mirror check added, red-first — moved to #3739 |
| 6 | `package.json` escape-sequence noise | Reverted |

The finding you flagged as blocking was worse than "we send a field we should not" — the endpoint rejects it outright:

```
clean session:                             ✓ accepted
+ stray firstMessage (the pre-fix shape):  ✗ 400 Unknown parameter: 'session.firstMessage'.
```

So a persona authoring `realtime.voice.default.firstMessage` on an OpenAI-backed co-agent would have failed **every** client-direct mint. Your strict-endpoint reasoning held literally.

### State after the split

Core **247**, OpenAI **179**, Gemini **106**, Inworld **63**, ElevenLabs **85**, AssemblyAI **48** — all passing, re-run after the split and after the latest `next` merge. Live ElevenLabs suite **16/16** against real agents and websockets.

Ready for another look when you have time.

packages/AI/Core/src/__tests__/baseRealtime.test.ts
```diff
@@ -341,6 +342,37 @@ describe('IRealtimeSession', () => {
...
+    ])('registers %s — %s', (key) => {
+        expect(REALTIME_SHARED_CONFIG_KEYS).toContain(key);
+    });
+
+    it('every neutral persona slot filed onto drivers is registered here', () => {
+        // The persona's WIRE-level slots in RealtimeVoicePersona (@memberjunction/ai-agents) are
+        // filed onto WHICHEVER driver resolves, consumed or not, so each one has to be scrubbable.
+        // Stated as a list rather than derived: ai-agents depends on this package, not the reverse,
+        // so importing the persona type here would invert the dependency.
+        const agnosticPersonaWireSlots = ['voice', 'firstMessage'];
```
rkihm-BC, line 365: **This is the one item I'd still ask for: the test doesn't do what its title promises.**

`every neutral persona slot filed onto drivers is registered here` is checked against a hardcoded pair. It verifies the two slots someone remembered to write down — so add a third slot to `RealtimeVoicePersona` and `collectAgnosticWireSettings` next month, forget the registry, and this passes green while the key leaks. That is the exact sequence that produced #3557. The instance is pinned; the class isn't.

The comment's reasoning is right as far as it goes — `ai-agents` depends on `ai-core`, so importing the persona type *here* would invert the dependency. But that argues the check belongs in **`ai-agents`**, which depends on both, not that it can't exist.

Concretely, a test over there could build a persona from the authoring schema's `voice.default` properties (minus the prompt-level `tone` / `speakingStyle`), push it through `GetProviderVoiceSettings`, and assert every returned key is in `REALTIME_SHARED_CONFIG_KEYS`. That derives from the two surfaces that actually matter — what an author can write, and what actually gets filed — so a new slot fails it automatically rather than by memory.

Additive, so not holding the merge on it. But closing the class was the stated goal of this test, and this version closes two instances.

The `holds no duplicates` case beside it is a nice touch — cheap, and it catches the half-landed rename that would otherwise look fine.

packages/AI/Core/src/generic/baseRealtime.ts
```diff
@@ -193,6 +193,15 @@ export abstract class BaseRealtimeModel extends BaseModel {
  * transport settings for self-hosted/proxied drivers. NON-OpenAI-protocol drivers (Gemini,
  * ElevenLabs, AssemblyAI, Inworld) MUST scrub these before forwarding an open config bag to their
  * SDK/wire — a co-agent config carrying them must be safe on every provider.
+ *
+ * **Registering a key here is what makes it safe; forgetting to is a wire leak.** A co-agent
```
rkihm-BC, line 197: This is the most valuable thing in the fix, more than the registration itself.

Stating that forgetting to register a key **is** a wire leak turns an implicit invariant into a documented one. And the last clause — that `ExtractRealtimeFeatures` enumerates its deletes explicitly and does **not** read this list — is precisely the fact I got wrong when I prescribed the fix. Writing it down here means the next person (or the next reviewer) can't make that mistake from the registry's own doc.

The placement is right too: this is the file someone opens when adding a neutral key, so it's where the two-step obligation needs to live.

packages/AI/Providers/OpenAI/src/models/openAIRealtime.ts
```diff
@@ -386,6 +387,13 @@ export function ExtractRealtimeFeatures(config: JSONObject | undefined): Extract
     const disableAutoResponse = rest.disableAutoResponse === true;
     delete rest.disableAutoResponse;
 
+    // `firstMessage` is the neutral opening-utterance key (issue #3557). This protocol family has
+    // no equivalent — the agent speaks when prompted — so there is nothing to translate, and the
+    // key is scrubbed rather than forwarded. Unconsumed is NOT the same as harmless: everything
+    // left in `rest` is spread into the session payload on BOTH topologies, and per the protected-
+    // wire-fields note below this endpoint rejects a malformed session object wholesale.
+    delete rest.firstMessage;
```
rkihm-BC, line 395: Verified this closes the leak — re-ran the probe against the real driver and `session.firstMessage` is now `undefined`, with keys down to `["type","instructions","audio"]`. I also passed `voice: 'alloy'` in the same bag to check the delete didn't over-scrub its neighbour; it still resolves into `audio` correctly.

One delete covering `CreateClientSession` and `session.update` is right, since both spread `features.rest` — and it picks up `xAIRealtime` for free through inheritance.

The comment earns its place by saying *unconsumed is not the same as harmless*. That's the whole lesson of #3557 in one line, and it sits exactly where someone would otherwise wonder why a key nobody reads needs deleting.

packages/AI/Providers/Inworld/src/inworldRealtime.ts
```diff
@@ -763,13 +764,24 @@ export class InworldRealtimeSession implements IRealtimeSession {
...
+     *
+     * Deliberately narrow: only the shared vocabulary is dropped. Passing an Inworld-native field
+     * the driver has no shorthand for is the whole point of this channel and still works.
+     *
      * @param session The session object being built (mutated in place).
      * @param config The caller's config bag.
      */
     private applyRawConfigOverrides(session: JSONObject, config: JSONObject): void {
         const consumedShorthands = new Set(['voice', 'stt', 'language', 'eagerness']);
+        const sharedMJKeys = new Set(REALTIME_SHARED_CONFIG_KEYS);
```
rkihm-BC, line 782: Correct, and correctly scoped. I checked that Inworld reads none of the registry keys outside this spread — no `endpoint`, `sampleRate`, `proxyBaseUrl` or `turnDetection` lookup anywhere else in the driver — so skipping them removes leakage without removing function. They were being forwarded raw to a provider that never understood them.

The `still forwards genuinely provider-native passthrough keys` test is the one that makes this safe to keep: the over-scrub direction is the easy mistake here, and the raw-override channel is the entire point of this loop. Guarding both directions rather than just the new one is what keeps this from quietly narrowing later.

Good that this went beyond `firstMessage` — implementing the contract the registry already documented means `reasoningEffort` and `mcpTools` stop leaking here too, which nobody had noticed.

.changeset/elevenlabs-realtime-first-message.md
```diff
@@ -0,0 +1,21 @@
+---
+"@memberjunction/ai": minor
```
rkihm-BC, line 2: **Loose end from the split: this now follows a rule that isn't in the repo yet.**

#3739 is still open. The only rule merged on `next` remains:

> **Minor bump**: ONLY when NEW migration files were ADDED IN THIS BRANCH in any `migrations/v*/` folder (applied to `@memberjunction/core` only)

This branch adds no migration and declares `minor` on seven packages. Under the merged rule that's `patch`; under #3739's proposed rule it's `minor`.

For the avoidance of doubt I think `minor` is the right *outcome* — I said so in the first review (this adds real exported API) and I endorsed the release-train rule on #3739. The issue is only that the justification currently points at an unmerged PR, which is how a one-off becomes precedent by accident.

Cleanest fix is ordering: land #3739 first. Failing that, a line in the PR body noting the dependency is enough.

Separately — good catch expanding this from three packages to seven as the fix widened into `ai`, `ai-openai`, `ai-gemini` and `ai-inworld`. That's the kind of thing that silently goes stale.

guides/REALTIME_CO_AGENTS_GUIDE.md
```diff
@@ -299,8 +301,9 @@ The pure implementation is `packages/AI/Agents/src/realtime/realtime-coagent-con
...
+- **`voice.default.firstMessage`** is the **provider-agnostic opening utterance** (#3557) — what the agent says FIRST, verbatim, before the user has spoken. It resolves exactly like the agnostic voice (filed onto whichever driver the framework picks, winning the `firstMessage` key over a matching `voice.providers.<key>` entry). Authored here rather than in the persona prompt because conversation-start behaviour is **not instruction-following**: ElevenLabs with no `first_message` produces no audio at all until it receives user audio, no matter how the prompt is worded — which is why a co-agent could never greet a caller. Omit it and the agent waits for the user, as before. Honoured today by **ElevenLabs** (`agent.first_message` override) and **AssemblyAI** (`greeting`). Every other driver **scrubs** it and opens silently: it is registered in `REALTIME_SHARED_CONFIG_KEYS`, which is what makes a neutral key safe to file onto a vendor that has no use for it — unregistered, it survives each driver's residual-bag spread and reaches the provider as an unknown session field.
```
rkihm-BC, line 306: Accurate now, and fixing the behaviour so the sentence became true beats rewording it around the leak — which is what I'd hoped for.

Explaining *why* registration is what makes a neutral key safe, rather than just asserting the key is scrubbed, is what makes this useful to the next driver author: it tells them what to do, not merely what is true today.

One pedantic note, only worth acting on if you're editing this line anyway: "Every other driver **scrubs** it" is true in effect but loose for HuggingFace, which never forwards the bag at all rather than scrubbing — it builds a fixed session shape and reads only `voice` and the transcription keys. Same outcome by a different mechanism.

rkihm-BC (approved):
Re-reviewed. The blocking defect is fixed and I verified it the same way I found it. Approving — one additive item left inline, plus a sequencing loose end from the split.

### First: my prescribed fix was wrong, and you were right to correct it

I said adding `firstMessage` to `REALTIME_SHARED_CONFIG_KEYS` "gets it deleted from OpenAI's residual bag alongside the other neutral keys." That is false, and I've confirmed it — `openAIRealtime.ts` imports that constant **zero** times and enumerates its deletes by hand.

Following my instruction literally would have fixed Gemini, left the OpenAI wire leak fully intact, and made the guide sentence *look* justified while the actual defect shipped — worse than the original bug, because the next reviewer would have seen a registration and stopped looking. You found that by reproducing before changing anything rather than by executing the instruction. Worth recording plainly: the review was wrong on the remedy.

### Verified fixed

Same probe against the real `OpenAIRealtime` driver, before and after:

| | Before | After |
|---|---|---|
| `session.firstMessage` | `"Hi there."` | `undefined` |
| session keys | `["type","instructions","audio","firstMessage"]` | `["type","instructions","audio"]` |

I passed `voice: 'alloy'` alongside and it still lands correctly in `audio`, so the new delete didn't over-scrub the neighbouring key.

The three-part structure is the right one — registry for Gemini, an explicit delete for the OpenAI family (covering xAI and both topologies in one line), and the Inworld skip that also closes the same leak there for `reasoningEffort` / `mcpTools`. Guide text is now accurate and covers Inworld, which the original sentence omitted. `package.json` noise is reverted. The changeset correctly grew from 3 packages to 7 as the fix widened — easy to miss, and you didn't.

**Suites:** OpenAI 179/179, Inworld 63/63, ElevenLabs 85/85, AssemblyAI 48/48. AI/Core showed 12 failures and Gemini 1 in my environment; I confirmed rather than assumed that both are stale `dist` on my side — the built `@memberjunction/ai` my worktree resolves has zero `firstMessage`, and AI/Core's failures are all `IsPlainObject is not a function` from a stale `@memberjunction/global` in unrelated `modelConfiguration` tests. I rebuilt AI/Core from branch source, aliased Gemini to it, and the scrub test **passes**.

Worth knowing rather than fixing: Gemini's protection comes from a registry it reads out of the *built* `@memberjunction/ai`, so that scrub test cannot fail in a workspace where AI/Core hasn't been rebuilt. Fine in CI with ordered builds — it's an integration check wearing a unit test's clothes.

<sub>Scope: re-probed the OpenAI driver directly; re-read the residual-bag handling in all seven realtime drivers; confirmed Inworld reads no registry key outside the spread it now skips; ran five provider/core suites and rebuilt AI/Core to isolate the two environment failures. Still no ElevenLabs/AssemblyAI keys, so live wire behaviour remains unverified on my side.</sub>

PR #3739, opened 2026-08-11
docs(changesets): state the bump-level rule where it actually loads

SDesai-BC:
Split out of #3736 at @rkihm-BC's request. It is repo-wide versioning policy that had no business on a voice-agent branch — least of all one whose own bump level the rule re-grades.

## The problem

The rule deciding `minor` vs `patch` lives in `.claude/commands/changeset.md`, which only enters context when someone types `/changeset`. Author a changeset any other way and there is no guidance at all.

Worse, the available evidence points the wrong way. Of 86 pending changesets, **40 carry a `minor`** — about 46% — across ~12 feature packages: `ai-core-plus` (10), `task-graph` (9), `server` (8), `integration-test-suite` (8), `ng-conversations` (7), `core-entities` (7), while `@memberjunction/core`, the only package the documented rule permits, has 7. Anyone matching the neighbouring files concludes minor-for-a-feature is normal.

That is not hypothetical: #3736 got its own level wrong twice, in **both** directions — first `minor` by matching neighbours, then "corrected" to `patch` by reading the command doc, which turned out to be wrong too.

## What this adds

- **`.claude/rules/changesets.md`**, scoped to `.changeset/**`, so it loads whenever a changeset file is opened rather than only under `/changeset`. Per the router in `CLAUDE.md`, file-type-scoped guidance belongs in `.claude/rules/`.
- **`npm run check:changeset`** — a **local** self-check. Deliberately no CI gate: the changesets bot already reports the level per package on every PR, so a gate would only add the pre-push and is-it-correct half, which the rule covers.
- **15 vitest cases** over real throwaway git repos, covering both directions of the rule, the `major` refusal, mixed-level changesets, and the guarantee that makes this adoptable at all.

## The rule, and why it is the reviewable part

| Level | When |
|---|---|
| `minor` | The branch adds a migration (`migrations/vN/*.sql`) **or** changes `metadata/` |
| `patch` | Everything else |
| `major` | Never without explicit approval |

Metadata counts because it *becomes* a migration: `metadata/CLAUDE.md` §1b specifies that a PR contributes declarative JSON only, and the build engineer's release-time `mj sync push` turns the accumulated edits into one consolidated metadata-sync migration. The database change is real, just deferred.

**This is a release-train convention, not per-package semver, and that is the decision to weigh.** Under it, a large feature adding exported API is a `patch` if it touches no metadata, while a one-line metadata edit is a `minor` across the entire `fixed` group. That is defensible given `fixed: [["@memberjunction/*"]]` — MJ versions move as a train — but it cuts against the position argued on #3715, where `patch` was said to understate a changeset adding exported API. Worth settling explicitly rather than by whoever writes the next changeset.

## Why one stray level matters

Measured on #3736: a changeset naming **three** packages `minor` produced a changesets-bot table of **301 packages, all Minor**. The bot displays this but does not judge it — it renders whatever you chose, inside a collapsed `<details>` of ~301 near-identical rows.

## Adoptable without a cleanup

The check judges **only changesets added in the current branch** (`--diff-filter=A` against the merge-base). A pending file from another branch carries no evidence of what that branch touched, so judging it would fail PRs for decisions made elsewhere. That is what lets this land without touching the 37 existing `minor` files, and it is pinned by its own test.

## Verification

- `npm run check:changeset:test` — **15/15**
- `npm run check:claude-md` — passes; 5 path-scoped rules validated
- `npm run check:changeset` on this branch — passes (no changesets added here)

> **Correction.** An earlier revision of this section claimed "failure count unchanged from `next` (90, all pre-existing `docs/media/*` link errors)". That was wrong, and @rkihm-BC was right that it does not reproduce. `docs/media/` is **gitignored** (`.gitignore:112`) — 147 files on my machine, 0 tracked — so those failures were a local artifact, not repo state. Measured: 90 of the 92 reported lines came from that untracked directory, and the remaining 2 are non-failing notes. On a clean checkout this exits 0. Left visible rather than quietly edited, since the whole point of this PR is that claims should rest on measured evidence.

No changeset of its own: this touches no published package.


.github/scripts/check-changeset-bump.mjs
```diff
@@ -0,0 +1,166 @@
...
+ * Usage: node .github/scripts/check-changeset-bump.mjs [--base origin/next]
+ */
+
+import { execFileSync } from 'node:child_process';
+import { readFileSync } from 'node:fs';
+
+const DEFAULT_BASE = 'origin/next';
+/** Paths whose presence in the branch earns a `minor`. */
+const DB_TRIGGERS = [
+    { label: 'migration', pattern: /^migrations\/v[0-9]+\/.+\.sql$/ },
```
rkihm-BC, line 39: **Repeatable migrations are excluded — and the gate doesn't just miss them, it gives the wrong answer.**

`migrations/R__RefreshMetadata.sql` lives directly under `migrations/`, not `migrations/vN/`, so this pattern never matches it. I ran the gate end-to-end against a repo shaped like the real one, on a branch that edits that exact file:

```
Branch touches no migration and no metadata — every entry must be patch.
✅ 1 changeset(s) added in this branch — bump levels are correct.
```

Flyway runs repeatable scripts on **every** deploy, so this is a database change by any reading of the rule. And because the mirror check now enforces the equivalence in both directions, an author who correctly reaches for `minor` on such a branch is actively **rejected**. That's worse than the silence this PR exists to fix: the tool answers confidently and wrongly.

One line:

```js
{ label: 'repeatable migration', pattern: /^migrations\/R__.+\.sql$/ },
```

Worth a test alongside the existing `migration-minor` case, since the whole point of this file is that the rule can't be inferred from context.

Separately, note this uses plain `--name-only` (any status), so *modifying* an existing versioned migration also counts as a trigger, where the old command doc said "NEW migration files were ADDED". I think the new behaviour is right — a changed migration is a changed database — but it's a silent semantic shift from the rule being replaced, and worth a word in the rule file.

.claude/rules/changesets.md
```diff
@@ -0,0 +1,89 @@
...
+pre-existing file using a different level is not yours to fix.
+
+## Check before you push
+
+```bash
+npm run check:changeset          # judges the changesets THIS branch adds, vs origin/next
+npm run check:changeset:test     # its own vitest suite
+```
+
+**Nothing enforces this in CI** — no PR fails on a wrong bump level. This rule and that command
```
rkihm-BC: **This is my one substantive request: I'd like the CI decision made explicitly, because the stated justification contradicts this file.**

The PR body says no CI gate is needed because *"the changesets bot already reports the level per package on every PR."* But eleven lines up, this file says — correctly —

> **The bot shows you this, but it does not judge it.** … It renders whatever you chose; a wrong `minor` and a correct `patch` produce the same-shaped table. Use it to confirm what you picked, not to find out whether the pick was right.

Both can't be true. The bot is a display, not a check, and this file is the half that's right.

The PR's own evidence points the same way. #3736 got the level wrong **twice** — once by matching neighbours, then again *after reading the command doc*. That's a documented case of prose-based guidance failing an author who actually consulted it. This PR's answer is better prose plus an optional local command — which asks that same author to remember to run something. 39 of 86 pending files suggest that memory doesn't hold at scale.

Meanwhile the gate already exists, exits non-zero with an actionable message, has 12 passing tests, needs no dependencies, and runs in about a second.

If you want a soak period before it blocks anyone, that's a good reason — and `continue-on-error: true` is exactly the tool for it. What I'd rather avoid is the rule shipping with "nothing enforces this" as a permanent state, since the reason it's being written is that nothing enforced it and that didn't work.

If you decide to keep it local-only, no objection from me — just say *why* here, so the next person doesn't re-litigate it from the bot argument.

.github/scripts/check-changeset-bump.mjs
```diff
@@ -0,0 +1,166 @@
...
+} catch {
+    console.error(`❌ Base ref '${base}' does not resolve. Fetch it first, or pass --base <ref>.`);
+    process.exit(2);
+}
+
+const files = changedFiles(base);
+const triggers = dbTriggers(files);
+const changesets = addedChangesets(base);
+
+if (changesets.length === 0) {
```
rkihm-BC, line 114: **A DB-touching branch with no changeset at all exits 0 here, before the trigger check runs.** Verified — a branch adding `migrations/v6/V…__New.sql` and no changeset prints `✅ No changesets added in this branch — nothing to check.`

By this PR's own argument that's the extreme case of the direction you just added the mirror check for: no bump at all is strictly worse than a `patch` bump on a schema change.

Whether "a DB branch must add a changeset" belongs in a script called *bump-level* guard is a fair scope question, and I'd accept either answer. But if it stays out of scope, the rule shouldn't read as a clean equivalence — right now the `no-changeset` test pins the pass using a code-only branch, so neither the behaviour nor the gap is recorded for the DB case. A one-line comment here and a sentence in the rule would close it.

.github/scripts/check-changeset-bump.mjs
```diff
@@ -0,0 +1,166 @@
...
+    }
+}
+
+// The MIRROR of the per-entry check above, and the direction that actually costs a release: a DB
+// branch declaring only `patch` ships a schema change under-versioned. It stays invisible whenever
+// some other changeset in the same release happens to carry a `minor`, so it fails rarely and
+// unpredictably rather than immediately. One `minor` anywhere in the branch's changesets is enough
+// — the `fixed` group moves every package to the highest bump regardless.
+if (triggers.length > 0 && violations.length === 0) {
+    const declaresMinor = changesets.some((path) => bumpEntries(path).entries.some((e) => e.level === 'minor'));
```
rkihm-BC: This is the check I asked for on #3736, and it's implemented better than what I described — I'd have settled for "every entry must be minor", which would have been wrong. Requiring only that *something* in the branch carries the `minor` is right, because the `fixed` group moves every package to the highest bump regardless, and the `metadata-mixed` test pins that reasoning rather than just the outcome.

The comment above it is doing real work too: explaining that under-bumping fails *rarely and unpredictably* (only when no other changeset in the release happens to be `minor`) is exactly why this direction needed a gate rather than prose — an error that usually hides is not one anyone learns from.

One trivial note: `bumpEntries(path)` re-reads and re-parses each changeset here after the loop above already did. Irrelevant at this scale, mentioned only so it's a choice rather than an oversight.

.github/scripts/__tests__/check-changeset-bump.test.mjs
```diff
@@ -0,0 +1,183 @@
...
+    });
+
+    it('passes when the branch adds no changeset at all', () => {
+        const { code } = runOnBranch('no-changeset', () => {
+            write('packages/Foo/g.ts', 'export const g = 1;\n');
+        });
+        expect(code).toBe(0);
+    });
+
+    it('ignores pending changesets contributed by OTHER branches', () => {
```
rkihm-BC, line 199: This is the test that makes the whole thing adoptable, and it's good that it exists as a test rather than only as prose in the script header.

Concretely: without the `--diff-filter=A` scoping, this PR would have to either fail on or clean up **39 pending changeset files** contributed by branches whose contents nobody here can see. I measured that number on `origin/next` — it's not a hypothetical cost.

The comment explains the *why* (a file already on `next` carries no evidence of what its own branch touched) rather than just describing the assertion, which is what will stop someone "simplifying" this into judging every pending file.

.claude/rules/changesets.md
```diff
@@ -0,0 +1,89 @@
...
+Loads when you open anything under `.changeset/`. The rule is short; the reason it needs its own
+file is that **you cannot infer it from the neighbouring files**.
+
+---
+
+## The rule
+
+| Level | When |
+|---|---|
+| `minor` | The branch adds a **migration** (`migrations/vN/*.sql`) **or** changes **`metadata/`** |
```
rkihm-BC: Endorsing this rule — see the review body for why I think it settles the #3715 tension in this direction. One thing it needs to actually pay off, though.

**The new meaning of the version number isn't documented anywhere a human operator will see it.** Under this rule, `minor` now means "this release contains a schema change" — which is genuinely more useful to MJ than semver, given the operational risk here is schema drift rather than API breakage. But that meaning is stated only in `.claude/rules/changesets.md`, which loads for agents editing `.changeset/**`.

An operator upgrading 6.1.5 → 6.2.0 has no documented way to learn that the minor implies "run migrations." That signal is the entire payoff of the policy, and right now nobody outside the authoring flow can collect it. It belongs in the release/CHANGELOG guidance too.

Second, worth stating explicitly in this table's neighbourhood: because this is a **release-train convention rather than semver**, external consumers must not read MJ's `minor` as "new API" or its `patch` as "no new API." That's a surprising enough departure from npm convention that leaving it implicit will cost somebody an afternoon.

rkihm-BC (commented):
Thanks for splitting this out — and for doing it properly rather than just moving files. The bidirectional gate is there and built well, the scoping decision is right, and the evidence holds up better than the PR claims.

**Verified rather than assumed:**

- **The neighbours really do teach the wrong rule.** I measured `origin/next` independently: **39 of 86** pending changeset files carry a `minor` entry — led by `ai-core-plus` (10), `task-graph` (9), `server` (8), `integration-test-suite` (8), `ng-conversations` (7), `core-entities` (7) — against `core` at 7. The premise isn't rhetorical.
- **12/12 gate tests pass.** The three new cases cover exactly the direction that was missing.
- **`check:claude-md` passes on the branch**, so the routing row and the new rule file are structurally valid.
- **No changeset of its own is correct** — root `package.json` is private, nothing published changes.

---

### One correction to the evidence section

> `npm run check:claude-md` — 5 path-scoped rules validated; failure count unchanged from `next` (90, all pre-existing `docs/media/*` link errors)

I can't reproduce the 90. Run on this branch **and** on `origin/next`, it's `exit 0`, `✅ check:claude-md — all checks passed`, no failures, and no `90` anywhere in the output — just 7 grandfathered references reported as non-failing notes.

Harmless to the code, and the actual state is better than claimed. Flagging it because this PR's whole thesis is that rules should rest on measured evidence, and "N pre-existing failures" is precisely the kind of claim that later gets used to wave through a real one.

---

### On the policy question you flagged

You called out that this cuts against what I argued on #3715. That was my position, so let me settle it rather than leave it hanging: **I was wrong — or more precisely, the point was moot.**

Under `fixed: [["@memberjunction/*"]]`, every package moves to the same version whether or not it changed. A consumer of `@memberjunction/ai-elevenlabs` already receives bumps driven entirely by packages they don't use, so the version number *cannot* carry per-package semver meaning — that ship sailed when the fixed group was adopted. Given the number can't mean semver, repurposing it to mean something real ("this release requires a schema change") is strictly better than leaving it meaningless.

**Endorsing the rule as written**, with one condition inline on the rule table: a signal only helps if the people who need it can read it.

---

Inline: one substantive request (**CI enforcement** — I'd like an explicit decision, not silent deferral), one real bug (**repeatable migrations are misclassified**, one line), and a gap worth a sentence. Commenting rather than approving only because of the repeatable-migration defect; everything else is small.

<sub>Scope of what I checked: measured the pending-changeset population on `origin/next`; ran the gate's 12 tests; ran `check:claude-md` on both the branch and `next`; and exercised the gate end-to-end against throwaway repos for the repeatable-migration and no-changeset cases. I did not run a workspace build — nothing here touches build output.</sub>

.claude/rules/changesets.md
```diff
@@ -0,0 +1,89 @@
...
+**The bot shows you this, but it does not judge it.** Every PR gets a `🦋 Changeset detected`
+comment listing each package and its `Minor`/`Patch` type — so the level is visible, inside a
+collapsed `<details>` of ~301 near-identical rows. It renders whatever you chose; a wrong `minor`
+and a correct `patch` produce the same-shaped table. Use it to confirm what you picked, not to
+find out whether the pick was right.
+
+## Do not pattern-match the neighbours
+
+`.changeset/` holds many pending files at any time, in a mix of both levels, written against
+branches whose contents you cannot see. Roughly a third currently use `minor`, including on
```
rkihm-BC: "Roughly a third" undersells your own case. Measured on `origin/next` just now: **39 of 86** pending changeset files carry a `minor` entry — about **45%**, near enough to half.

The breakdown is worth having here too, because it shows the problem isn't concentrated in one careless corner: `ai-core-plus` (10), `task-graph` (9), `server` (8), `integration-test-suite` (8), `ng-conversations` (7), `core-entities` (7) — versus `core`, the only package the old rule permitted, at 7.

Since this is a durable rule file rather than a PR description, a number that drifts is worth either pinning with a date ("as of 2026-08, 39 of 86") or softening to "nearly half, spread across a dozen feature packages" so it doesn't quietly become false.

.github/scripts/check-changeset-bump.mjs
SDesai-BC: Fixed — and you were right that this was worse than a gap. Both directions now tested; the second one is the case you identified as the real damage:

```
✓ allows minor when the branch edits a REPEATABLE migration
✓ requires minor when the branch edits a REPEATABLE migration
```

Both were red before the one-line trigger, including the rejection of a correct `minor`.

On the semantic shift you flagged separately — you are right that it is silent, and right that the new behaviour is the correct one. `--name-only` with no status filter means a MODIFIED versioned migration also triggers, where the old command doc said "NEW migration files ADDED". A repeatable script forces the question, since it is only ever modified, never added twice. Now stated in the rule file under the table:

> **Changed counts, not just added.** Modifying an existing migration is as much a database change as adding one, and a repeatable script is only ever modified — Flyway re-runs it on every deploy.

.github/scripts/check-changeset-bump.mjs
SDesai-BC: Kept out of scope, and now recorded as a decision in both places rather than an undocumented pass.

In the script, at the early exit:

> NOTE: this exits 0 even on a branch that changes the database. That is a deliberate scope line, not an oversight — this guard judges the LEVEL of the changesets a branch declares, and a branch declaring none gives it nothing to judge. "A DB branch must declare a changeset at all" is a different rule (presence, not level) and belongs wherever changesets are made mandatory.

And a test that pins the DB case specifically, rather than leaving the pass demonstrated only by a code-only branch as you noted:

```
✓ does NOT require a DB branch to add a changeset (out of scope for a bump-LEVEL guard)
```

The rule file also now says plainly that it does not check presence. Agreed the no-bump case is arguably more severe than the under-bump — it just is not a bump-LEVEL rule, and folding it in would make this quietly two guards.

.github/scripts/check-changeset-bump.mjs
SDesai-BC: Thanks — and the re-parse is now a choice rather than an oversight, in the direction you would expect: the entries are parsed once into `parsed` and both the per-entry loop and the mirror check read from it. The reason is not the cost at this scale, it is that two reads of the same file could in principle disagree, and the two halves of an equivalence check disagreeing would be a genuinely confusing failure.

.github/scripts/__tests__/check-changeset-bump.test.mjs
SDesai-BC: Appreciated — and your independently measured number is now in the rule file, since a durable file should not carry a figure that drifts. It is dated for that reason: "as of 2026-08, 40 of 86". I measure 40 where you measured 39; `origin/next` moved between our runs, and the package breakdown matches yours exactly.

.claude/rules/changesets.md
SDesai-BC: Fixed, with your numbers and the breakdown, dated so it does not quietly go stale:

> **As of 2026-08, 40 of 86 pending files carry a `minor`** — close to half, and spread across a dozen feature packages rather than concentrated in one careless corner: `ai-core-plus` (10), `task-graph` (9), `server` (8), `integration-test-suite` (8), `ng-conversations` (7), `core-entities` (7), against `core` at 7.

I get 40 rather than 39 because `origin/next` moved between our measurements; the breakdown is identical.

.claude/rules/changesets.md
SDesai-BC: The contradiction was real and the bot argument was the wrong justification — that is my error, not a position. As you say, this same file states eleven lines up that the bot displays without judging, and both cannot be true.

**Decision: CI stays off, and the rule now records why.** Not because the bot covers it, but because the maintainer chose to start with guidance and see whether the rule landing where it actually loads is enough on its own. That is a deliberate call about how much enforcement this warrants, and it is the one I was given. The rule file now says so explicitly, so the next person inherits the reason rather than re-deriving it from an argument that does not hold:

> **Nothing enforces this in CI, by maintainer decision** — no PR fails on a wrong bump level. […] Recorded so it is not re-litigated from a bad argument: the reason is *not* that the changesets bot covers it. It does not — see above, it displays the level without judging it. The reason is that the maintainer chose to start with guidance rather than a gate […] The gate exists, exits non-zero with an actionable message, and has 15 passing tests, so wiring it into CI later is a workflow file and nothing more — `continue-on-error: true` first if a soak period is wanted.

For the record I think your argument is the stronger one, and the evidence is on your side: #3736 got the level wrong twice, the second time *after* consulting the documentation. That is prose failing an author who did the right thing. I have put the `continue-on-error` path in the file so it is a one-file change if that view prevails — worth raising with @SDesai-BC directly, since it is his call rather than mine to reverse.

SDesai-BC:
All six addressed — one real fix, one decision recorded, and one correction that is mine.

## The defect

Repeatable migrations now trigger. You were right that this was worse than a gap: with the mirror check in place the gate did not merely miss `migrations/R__*.sql`, it **rejected an author who correctly reached for `minor`**. Both directions are tested and both were red first.

Also documented the silent semantic shift you spotted alongside it — triggers match *changed* files, not just added ones, where the rule being replaced said "NEW migration files ADDED". Repeatable scripts force the question, since they are only ever modified.

## Your correction to my evidence — you were right

I could not dismiss it, so I traced it: **`docs/media/` is gitignored** (`.gitignore:112`) — 147 files on my machine, **0 tracked**. Measured, 90 of the 92 reported lines come from that untracked directory; the other 2 are the non-failing notes you saw. On a clean checkout `check:claude-md` exits 0, exactly as you found.

So the "90 pre-existing failures" was a local artifact reported as repo state. The PR body now carries the correction **visibly** rather than a quiet edit, because your reason for flagging it is the right one: a PR arguing that rules should rest on measured evidence cannot itself ship an unmeasured claim.

## On the policy question

Thank you for settling #3715 rather than leaving it hanging — and for arriving at it from the mechanism (the `fixed` group already destroyed per-package semver meaning, so repurposing the number costs nothing and gains something real). That reasoning is now in the rule file, because it is the justification, not just the conclusion.

Your condition is met: `VERSIONING.md` now carries the operator-facing half, in the version-grammar section where someone upgrading actually looks.

| Going from | Read it as |
|---|---|
| `6.1.5` → `6.2.0` | **Run migrations.** Something in this release changes the database. |
| `6.1.5` → `6.1.6` | No schema change. Code only. |

…plus the explicit warning that MJ’s `minor` must not be read as "new API" and its `patch` may well add exported API. Agreed that leaving that implicit would cost somebody an afternoon.

## Everything else

- **CI enforcement** — decided explicitly and recorded with its actual reason (a maintainer call to start with guidance), not the bot argument, which was my error. Detail on the thread; I have noted there that I find your argument the stronger one.
- **DB branch with no changeset** — kept out of scope, now a stated scope line in the script plus a test that pins the DB case specifically.
- **Population numbers** — re-measured and dated: 40 of 86 (~46%), with your breakdown. I get 40 to your 39 because `origin/next` moved between runs.
- **Re-parse** — now parsed once and shared.

`check:changeset:test` **15/15**; `check:changeset` and `check:claude-md` both pass.

SDesai-BC:
Correction pushed —caught that the `VERSIONING.md` section I added contradicted canon, and one row of it was **unsafe advice**.

`plans/lts-process.md` §12: *"On a line, everything is a patch"* — metadata migrations, CodeGen repairs, and even schema DDL under a security exception. So my table was wrong in the direction that costs someone a database:

| What I wrote | Why it is wrong |
|---|---|
| `6.1.5` → `6.1.6` = "No schema change. Code only." | A line patch may well carry a migration. An operator trusting this would **skip migrations they need**. |
| `6.1.5` → `6.2.0` = "Run migrations." | A line minor is the Nth certification of the era, not a statement about DDL. Lines are patch-only forever; you never take that step within a line. |

Worse, the same document already names the right answer and warns against exactly what I did — `dbImpact` is *"the honest replacement for smuggling that signal into version digits"* (§12, and the glossary at line 43). I reinvented the thing canon rejected.

`VERSIONING.md` now points operators at `dbImpact` (`none | metadata | repair | schema`) with what each value means, and states plainly that the digits answer this question in neither direction.

**The rule itself stands — but for canon’s reason, not mine.** It is Edge **tuple grammar** (§3.1): Edge runs in permanent prerelease mode and *bumps do not compound*, so the tuple comes from the highest level accumulated since the last certification. The invariant is that migrations only ever ship in a minor-or-higher-tupled release, and an all-`patch` cycle would target `X.Y.Z+1` instead. A `minor` changeset keeps the accumulated tuple correct — **it does not visibly bump anything mid-stream**, which is the part I had wrong when I described it as an operator-visible signal.

@rkihm-BC — this touches your endorsement directly. Your condition was that the new meaning reach a human operator, and I met it with a formulation that was both non-canonical and unsafe. The condition is still met, just by pointing at the mechanism that already existed rather than a new one: `dbImpact`, surfaced by `mj versions` and the release notes. Your underlying reasoning about the `fixed` group is untouched and is still recorded in the rule file.

`check:changeset:test` **15/15**; `check:claude-md` passes.

PR #3740, opened 2026-08-11
ci: gate the CodeGen tail against committed migrations (#3737)

### LINKED ISSUE #3737: next: entity_subclasses.ts missing the generated subclass for 'MJ: AI Model Price Unit Types' (migration shipped without its codegen tail)
## Summary
`next` ships the **AI Model Price Unit Types** entity (migration + metadata) without its codegen tail: `MJAIModelPriceUnitTypeEntity` is missing from the committed `packages/MJCoreEntities/src/generated/entity_subclasses.ts`. Committed generated code no longer matches committed migrations.

## Repro
Fresh environment off `origin/next` (observed at `13873677f7`, also with PR #3717 head `4bfeac866e`), run migrations, then a plain CodeGen pass: regeneration **adds** `MJAIModelPriceUnitTypeEntity` (`@RegisterClass(BaseEntity, 'MJ: AI Model Price Unit Types')`) to `entity_subclasses.ts` — i.e. the subclass was never committed alongside the migration/metadata that introduce the entity.

## Expected vs actual
- **Expected:** committed generated code matches committed migrations (repo convention).
- **Actual:** the `AIModelPriceUnitType` table/entity exists, the generated subclass is absent → runtime falls back to `BaseEntity`. Benign at runtime, but out of convention — and it turns every downstream clean-environment codegen diff into noise.

## Fix
Regenerate and commit `entity_subclasses.ts` (plus any sibling generated artifacts) on `next` alongside the AIModelPriceUnitType migration.

_Found by mjdev instance provisioning (codegen drift tripwire) on 2026-08-11._

MS-BC:
Closes nothing yet — see "What this does not do" below. Relates to #3737.

## The report doesn't reproduce

#3737 says `next` ships the AI Model Price Unit Types migration without its codegen tail — `MJAIModelPriceUnitTypeEntity` absent from `packages/MJCoreEntities/src/generated/entity_subclasses.ts`. On `next`, and at the reported commit `13873677f7`, that entity is fully wired:

| Artifact | Status on `next` |
|---|---|
| `entity_subclasses.ts` — `@RegisterClass(BaseEntity, 'MJ: AI Model Price Unit Types')` + class + zod schema | present |
| `packages/MJServer/src/generated/generated.ts` — GraphQL type, inputs, resolver | present (26 refs) |
| `core-entity-forms` form component + `generated-forms.module.ts` registration | present |
| All four `mj-class-registrations.ts` bootstrap manifests | present |
| `__mj.Entity` row in the v5.46 baseline | present (`7815F891-…`) |

I also ran the invariant the issue asserts, repo-wide: the shipped baseline's entity metadata vs. the generated subclasses. The 11 entities with no subclass are the legacy Workflow/Report/ScheduledAction cluster that `V202608061704__v6.1.x__Phase0_…_Retirement.sql` deletes; the 6 subclasses not in the baseline are entities added by post-baseline migrations. No unexplained gap.

So the specific defect is not there. **The gap the report exposes is real, though**: nothing in CI holds committed generated code to committed migrations. The only way this class of drift surfaces today is a downstream clean-environment CodeGen diff — days later, from a provisioning tripwire, which is precisely how #3737 arrived. It never fails loudly, because the runtime just falls back to `BaseEntity`.

## What this adds

A DB-free guard, `.github/scripts/check-codegen-tail.mjs`. It compares files the repo already ships, so it runs in ~2s on a PR rather than needing a SQL Server, a migrate, and a CodeGen pass.

**1. Artifact parity.** CodeGen emits the same entity roster into three files — `entity_subclasses.ts`, MJServer's resolvers, the Explorer forms module. They must agree exactly. This catches a *partial* tail, which is the common real-world shape because it survives both a build and a test run.

**2. Migration → subclass.** Every core-schema table the migration history creates and does not later drop must ship a generated subclass. This catches the tail being skipped *entirely* — which check 1 structurally cannot see, since a wholly-absent entity is absent from all three artifacts consistently.

Both checks are **always full-repo**. There was a "just my PR's migrations" mode and it was deleted: it is strictly weaker in the case that matters most, where Dev B regenerates against a database that lacks Dev A's just-merged table, merges `next`, and resolves the conflict by keeping their own artifacts. A's entity then vanishes from all three files *consistently* (invisible to check 1) and A's migration is not in B's diff (invisible to a scoped check 2). Only the full sweep sees it. The documented local command is now literally the command CI runs.

Both are clean on `next` today: 376 entities agreeing across all three artifacts, and across the full 665-migration history exactly one created table has no subclass — `SystemEvent` (v2.0), which predates the entity system and was never registered in `__mj.Entity`. It's allowlisted in `NON_ENTITY_TABLES` with that reason stated.

## Wiring

The guard is a step in the **`Check migrations`** job (`.github/workflows/changes.yml`) — the single required status context on `next` — and is **not** gated on `has_migrations`.

Both are deliberate. A separate paths-filtered workflow could never *be* a required check: on every PR it skipped, the context would never arrive and the PR would sit pending forever. And the drift is not introduced only by migration PRs — a PR editing only a generated artifact, or only CodeGenLib (whose emitted shape the extractors are pinned to), can introduce or expose it. Living in that job is what makes this a gate rather than a decoration nobody is blocked by; it also inherits the job's base resolution, which deliberately avoids `pull_request.base.sha`, and extends coverage to `main`-targeting PRs.

`--compare-to` scopes **failure** to what the branch introduces. Both checks are whole-repo, so without it one bad merge onto the base branch would red every subsequent PR — and since this context is required, "red" means "nobody can merge". Pre-existing drift is still surfaced, as a `::warning`. A base tree that cannot be evaluated degrades to **strict** (every finding counts as introduced), which is the safe direction. The base worktree is materialised **only when the head tree is actually dirty**, so the green path — essentially every PR — is one ~2s sweep that never pays for a checkout.

Also wired: `npm run check:codegen-tail` for local use, listed in CLAUDE.md's Local CI mirrors; `.github/scripts/check-codegen-tail.mjs` added to `test.yml`'s paths so a PR touching only the guard runs the guard's own tests; and a "commit the CodeGen tail" section in `migrations/CLAUDE.md`, where migration authors are already reading.

## What green means, precisely

Check 2 keys on `CREATE TABLE`, so it covers exactly one shape: a **new core-schema table** shipped without its generated entity, plus disagreement between the three rosters. A migration that only adds a column, widens a type, renames an entity, re-points a relationship, or changes metadata leaves no `CREATE TABLE` and usually no roster change, and is **invisible to this check**.

Green is *not* "the CodeGen tail is committed" — it is "no new table is missing its entity, and the three generated rosters agree." The definitive check is DB-bearing (migrate → `mj codegen` → `git diff --exit-code`), and its natural home is `.github/workflows/migrations.yml`, which already stands up SQL Server and runs `mj migrate` on push to `next`. That is a separate PR with real runtime cost to weigh, deliberately not smuggled in here.

**Known gap**, named rather than left implicit: the tail is five artifacts and this guard covers three. The four generated `mj-class-registrations.ts` manifests carry the same roster and were verified by hand while investigating #3737, but are not encoded. They are tree-shaking-prevention manifests, so a missing entry is a runtime-only failure — exactly the silent class this exists to catch. A follow-up.

## Verification

Passing wasn't treated as evidence.

- **Mutation-tested** against the real tree: deleting the `MJAIModelPriceUnitTypeEntity` declaration (i.e. simulating #3737 exactly) reds with the parity message; removing only its form-module import reds with the partial-tail message. Tree restored and verified clean afterwards.
- **Delta scoping tested on all three paths**: drift in the head only → exit 1; the same drift already on base → `::warning` + exit 0; an unevaluable base → degrades to strict + exit 1.
- **213/213 tests pass** across the `.github/scripts` suite (9 files); 34 of them cover this guard. The extractors are deliberately keyed to CodeGen's current output shape, so a change there fails loudly instead of silently matching zero entities and passing everything.
- `npm run check:claude-md` passes.

### Two bugs found during review, both fixed

**Comment stripping, done the obvious way, silently deleted three real tables.** Stripping literals, then `/* */`, then `--` is wrong: an apostrophe inside a comment (`the user's mode`) makes a literal-first pass consume everything to the next apostrophe — including a `CREATE TABLE` thirty lines below. It removed `Theme`, `ContentItemChunk` and `AIAgentCredential` from the sweep, and the guard **stayed green**, because those three happen to have subclasses. That is precisely the "matches zero and therefore passes everything" failure this guard's own header warns about, introduced by the hardening against it. Fixed with a single left-to-right scanner that knows which context it is in, and that also handles `''` escapes, nested block comments, and bracketed identifiers. Masked matches: 0. Surviving tables: unchanged.

**That scanner then cost 13× the guard's stated runtime.** Walking one character at a time through 566 MB of SQL (the five v5 baselines are 36–52 MB each) took the guard from ~2s to **25.8s**, or **50.6s** with `--compare-to` — and "~2s" was the justification for running it ungated in the one required context. Only four characters can change state, so the scanner now jumps to the next one and copies the whole intervening span; masked regions advance by `indexOf`. **14.6s → 1.5s**, end-to-end **50.6s → 2.4s**. Verified as a pure speed change: masked output is **byte-identical** to the per-character scanner across all 665 committed migrations, so match positions — which the within-file CREATE/DROP ordering depends on — are unchanged.

## What this does not do

It does not close #3737 — the issue asks for a regeneration that, as far as I can determine, is already committed. Leaving it open for the reporter to confirm against their provisioning environment; if their tripwire fired on a *published npm* artifact rather than the repo, that's a release-packaging question this guard wouldn't have caught either. I did not run a full clean-room bootstrap (fresh DB + migrate + CodeGen) — the global `mj` CLI here is 5.51 against a 6.1.0-edge.1 repo, so it would have needed a full monorepo build to be meaningful, and the static evidence above is conclusive for the specific claim.

No changeset: no published package changes, and `changes.yml`'s minor-changeset rule only fires when migration files change, which this PR has none of.


rkihm-BC (commented):
## Review

I fetched the branch, ran the guard against the real tree, ran its test suite, mutation-tested it, and checked it against the repo's actual branch-protection config.

**Verdict:** the diagnosis is right, the mechanism is sound, and the invariant it asserts is a real CodeGen invariant — not a coincidence. Worth merging after four fixes, two of which are one-liners. But as wired today it is **advisory-only**, and its **documented local command is not the command CI runs** — and that second point is where parallel migrations bite.

### Verified independently

| Claim | Result |
|---|---|
| Guard is clean on the PR head | ✅ `374 entities … scanned all 664 migration(s) … OK`, 1.8s wall clock |
| The arithmetic in the description | ✅ Exactly 386 unique core-schema `CREATE TABLE`s = 374 subclasses + 11 dropped + 1 allowlisted |
| The three artifacts really do carry the same roster | ✅ 374 / 374 / 374, and `packages/CodeGenLib/src/runCodeGen.ts:484-620` shows all three derive from the same `coreEntities` filter (`IncludeInAPI` ∧ schema = `__mj`) |
| Test suite | ✅ 194 pass / 9 files on the merged head (the 159/8 in the description predates the `next` merge) |
| Mutation test (#3737 shape) | ✅ Reproduced — reds correctly |
| `extractServerEntities` matches reality | ✅ Exactly one `@memberjunction/core-entities` import in the generated resolvers, single-quoted, one line |

---

## What's good

- **Right altitude.** A DB-bearing check (migrate + CodeGen + `git diff --exit-code`) is definitive, and `.github/workflows/migrations.yml` already stands up SQL Server — but it's push-only and multi-minute. A 2-second static guard on the PR is the higher-value first line, and the trade is correctly reasoned in the header comment.
- **The two checks are genuinely complementary, and the stated reason is correct.** I confirmed a wholly-absent entity is invisible to parity (absent from all three consistently, so they agree) and only check 2 sees it; and a partial tail is invisible to check 2 and only parity sees it. Neither is redundant.
- **The extractors are deliberately brittle, with the rationale written down.** For a parser-based guard, "matches zero and passes everything" is the failure mode that matters. Pinning the tests to CodeGen's exact emitted shape so a change fails loudly is the right instinct and rarely applied.
- **Exit 2 (misconfigured) vs exit 1 (drift)** is a real distinction, not decoration.
- **The allowlist is honest.** `SystemEvent` with a stated reason beats weakening the predicate, and the entry is asserted in a test.
- **The write-up doesn't overclaim** — it says plainly that #3737 doesn't reproduce and that no clean-room bootstrap was run.

---

## Multiple developers, parallel migrations

Two halves, pointing in opposite directions.

**Where this helps more than the description claims.** The MJServer resolver roster is a *single 374-entity line* (`packages/MJServer/src/generated/generated.ts:21`). Any two developers who each add an entity both edit that one line, so a conflict is guaranteed and the resolution is done by hand on a line that is ~30KB wide. Silently dropping someone else's entity there is the most likely CodeGen-merge accident in this repo, and the parity check catches it exactly.

Better still: the `next-protect` ruleset has `strict_required_status_checks_policy: true` — branches must be current with `next` before merging, and the house flow is merge-not-rebase. So *if* this check were required, the combined state of two parallel migrations would always be re-validated before either merges. The infrastructure for the parallel case already exists.

**Where it hurts.** Scoped mode is blind to precisely the parallel-developer regression, and it's the mode `migrations/CLAUDE.md` tells developers to run. Dev B branches, regenerates from a DB that doesn't yet have Dev A's just-merged table, merges `next`, resolves the conflict by taking their own artifacts. The entity vanishes from all three files *consistently*, and A's migration is untouched by B's diff. I ran this:

```
# what migrations/CLAUDE.md tells you to run before pushing
codegen-tail: 373 entities … scanned 0 migration(s) changed vs HEAD
codegen-tail: OK                                                     exit=0

# what CI actually runs
codegen-tail: FAIL — 1 drift(s)
  ✖ aimodelpriceunittype: created by migrations/v2/V202506141319__v2.49.x_AI_Model_Cost_Tracking.sql,
    but no generated entity subclass ships for it                    exit=1
```

Note the attribution: it blames a June-2025 v2 migration for a break introduced today. For an innocent third developer whose PR reds on this, the message points at the wrong file and the suggested fix ("run CodeGen and commit its output") is something they can't do correctly without a DB in someone else's state.

**Recommendations, in order:**

1. **Drop scoped mode entirely.** `--all` costs 1.8 seconds over 664 migrations. The `--base=` flag, `changedMigrations`, the unresolvable-ref fallback, the workflow's `fetch-depth: 0`, and the two workflow comments describing a merge-base diff are all dead weight around a mode that is strictly weaker and never runs in CI (the workflow's single `run:` step is unconditionally `--all`). Make the local command and CI the same command.
2. **Fail only on drift the PR introduces.** Compute the same result at the merge base; report pre-existing drift as a `::warning` and fail on the delta. Otherwise one bad merge onto `next` reds *every* subsequent migration PR — both checks are full-repo, so there's no scoping to fall back on.

---

## Concerning

**1. It is not a gate.** The `next-protect` ruleset requires exactly one context: `Check migrations`, produced by `.github/workflows/changes.yml:27`. `Check CodeGen tail` isn't in it, so a red run is merely visible. And it can't simply be *added*, because the new workflow is `paths:`-filtered — a required check that never reports leaves PRs pending forever.

The repo has already solved this: `changes.yml` has **no** `paths:` filter and gates the work internally on a `has_migrations` step, precisely so the context always reports. **Strongest structural recommendation here: make the tail check a step inside the existing `Check migrations` job** rather than a new workflow. It costs ~2s, needs no `pnpm install`, and inherits both the required-check status and that job's much better base resolution — which explicitly avoids `pull_request.base.sha` because long-lived PRs merge the base in (`changes.yml:44-56`). `check-codegen-tail.mjs:170` hardcodes `origin/next` and re-learns none of that.

**2. The CLAUDE.md edit deletes a live line.** The diff *replaces* `check:changeset` rather than adding to the list, but `package.json:119` still defines it. The budget has headroom (208/260) and `check-claude-md.mjs` has no rule that would force a swap, so this looks like an accidental one-line clobber.

**3. Drop tracking is order-insensitive and permanent** (`check-codegen-tail.mjs:96-118`). `dropped` is a global `Set` union with no version ordering, so any table named in any `DROP TABLE [__mj].[X]` anywhere in history is exempt *forever* — including if a later migration re-creates it, and including a `DROP TABLE IF EXISTS` guard sitting above a `CREATE TABLE` in the same file. Zero occurrences today (the only 11 drops are the retirement cluster), but that idempotent-guard pattern is a normal thing to write and would silently punch a permanent hole. Fix: walk migrations in Flyway version order, last-op-wins.

**4. `test.yml` won't run the guard's own tests.** Its `paths:` list includes `.github/scripts/__tests__/**` but not `check-codegen-tail.mjs`, so a PR that changes only the guard never runs the tests that prove its parsing is sane. `test.yml:29-32` has a comment describing this exact trap for `check-pg-migration-content.mjs`. One line.

**5. `packages/CodeGenLib/**` isn't in the new workflow's paths.** The extractors are pinned to CodeGen's emitted shape, and the design leans on failing loudly when that shape changes — but the PR that changes the shape won't trigger the guard. The loud failure lands on whoever merges the next migration.

**6. The documentation overstates coverage.** `migrations/CLAUDE.md` says "CI now checks this" for "commit the CodeGen tail." What's actually checked is *new tables*. A migration that adds a column, widens a type, renames an entity, or changes metadata leaves no `CREATE TABLE` and no roster change, and is completely invisible. One sentence stating the boundary would stop green being read as "generated code is committed."

**7. Redundant full scan.** In `--all`, `scanMigrations` runs over all 664 files at line 226 and again at line 227 for `history` — identical inputs, identical work. Not a perf problem at 1.8s, just gratuitous: `const history = all ? scoped : scanMigrations(...)`.

**8. Minor:**
- Regex over raw SQL counts commented-out DDL. Currently zero core-schema-qualified `CREATE TABLE` in comments, so no live false positive — but a header block documenting a table would produce an unfixable red. Stripping `--` and `/* */` first is cheap; `changes.yml:213-217` already does string-stripping for the same class of reason.
- `--base=X` (equals form, soft fallback) vs the sibling `check-changeset-bump.mjs` in the same directory using `--base X` with a hard error — inconsistent CLI grammar in one directory. Moot if you take recommendation 1.
- `extractServerEntities` uses non-global `.match` and hardcodes single quotes. Correct today, and a Prettier rewrap would exit 2 loudly rather than silently, so acceptable — but a `['"]` class costs nothing.
- Only gates PRs targeting `next`; `origin/lts/5` exists and `changes.yml` also covers `main`.

---

## Missing

- **The tail is 3 artifacts here, but the description enumerates 5.** The four generated `mj-class-registrations.ts` manifests were verified by hand while investigating #3737 and then not encoded. Those are tree-shaking-prevention manifests — a missing entry is exactly the silent runtime-only breakage this guard exists to prevent. Fine as a follow-up, but it should be named as a known gap rather than left implicit.
- **No path to the definitive check.** The header explains why static beats DB-bearing on a PR, which is right, but doesn't say what closes the remaining gap. `migrations.yml` already has SQL Server + `mj migrate` on push to `next` and stops short of `mj codegen && git diff --exit-code`. One sentence pointing there keeps this from reading as "solved."
- **No changeset** — correct, in my read: no published package changes, and `changes.yml`'s minor-changeset rule only fires when migration files change. Worth a line in the description to pre-empt the bot.

---

## Suggested order of work

1. Move the check into the `Check migrations` job in `changes.yml` — makes it required, kills the path-filter problem, inherits correct base resolution.
2. Delete scoped mode; `--all` everywhere.
3. Restore the `check:changeset` line in `CLAUDE.md`.
4. Make drop tracking version-ordered, last-op-wins.
5. Add `check-codegen-tail.mjs` to `test.yml` paths; state the coverage boundary in `migrations/CLAUDE.md`.

MS-BC:
Thanks @rkihm-BC — all five items in your suggested order are done in 63fae48da0, plus recommendation 2 and the minors. Two things I want to put in front of you specifically: **your #1 was the load-bearing finding** (an advisory check is worth roughly nothing, and I had shipped one), and **implementing your #8 uncovered a real bug in my own guard** that I would not have found without going looking.

## The bug #8 turned up

You wrote that stripping comments was "cheap" and that there is currently zero commented core-schema DDL, so no live false positive. Both true. But the obvious implementation — strip literals, then `/* */`, then `--` — is *wrong*, and I only found out because I measured before and after instead of assuming.

It silently deleted three **real** tables from the sweep: `Theme`, `ContentItemChunk`, `AIAgentCredential`. The cause:

```sql
/*
    ... light/dark is the user's mode layered under it        <- line 8
*/
CREATE TABLE ${flyway:defaultSchema}.Theme (                  <- line 39
```

A literal-first pass reads `'s mode ... '` as a string literal and consumes everything to the next apostrophe — including the `CREATE TABLE` thirty lines down. And the guard **stayed green**, because those three tables happen to have generated subclasses. So the hole was invisible: precisely the "matches zero and therefore passes everything" failure mode this guard's own header warns about, introduced *by* the hardening for it.

Fixed with a single left-to-right scanner that knows which context it is in — an apostrophe inside a comment can't open a string, a `--` inside a string can't open a comment. It also handles doubled-quote escapes, **nested** block comments (T-SQL allows them; a non-greedy regex gets them wrong), and bracketed identifiers, which may legally contain `'`.

Your reading is confirmed once the scanner is correct: masked matches = **0**, surviving tables = **375** (374 subclasses + allowlisted `SystemEvent`), i.e. identical to before. The stripper is purely defensive today. Tests pin the regression directly, including the `user's`-in-a-comment shape.

## Your suggested order

**1. Into the `Check migrations` job.** Done; `ci-codegen-tail.yml` deleted. You were right that this was the structural problem — and right about *why* it couldn't just be added to the ruleset, which is the part I had not thought through: paths-filtered plus required leaves PRs pending forever on every PR it skips. Verified the step now lands in the job whose `name:` is exactly the required context:

```
required context name: "Check migrations"
  [always] Checkout repo
  [always] Self-test migration ID determinism detector
  [always] Resolve current base branch tip
  [always] Check for migration file changes
  [always] Set up pnpm
  [always] Set up node
  [always] Check CodeGen tail matches committed migrations (#3737)
  [gated]  Get current version
  ... (the migration-specific steps, unchanged)
```

Three consequences I'd flag as improvements over what I had, all of which are yours rather than mine:

- It inherits that job's base resolution, so the `pull_request.base.sha` trap you pointed at is now avoided by construction rather than by me remembering it.
- `main`-targeting PRs are covered, closing your #8's last bullet for free.
- No `pnpm install`, and no `fetch-depth: 0` — dropping scoped mode made the guard pure-filesystem, so it needs no git history at all.

I made the step **ungated** rather than `has_migrations`-conditional, which goes slightly beyond what you asked. Reason: the drift is not only introduced by migration PRs — a PR editing only a generated artifact, or only `CodeGenLib` (whose emitted shape the extractors are pinned to), can introduce or expose it. At ~2s that is cheaper than reasoning about when to skip it, and it dissolves your **#5** entirely rather than patching a paths list.

**2. Scoped mode deleted.** `--base=`, `changedMigrations`, the unresolvable-ref fallback, the workflow's `fetch-depth: 0`, the misleading comments, and the duplicate full scan (**#7**) all went with it. The documented local command is now literally the command CI runs. Your `373 … scanned 0 migration(s) … exit=0` transcript was the whole argument — a guard whose local invocation is weaker than its CI invocation teaches people that green means something it doesn't. **#8**'s CLI-grammar inconsistency is moot as you predicted.

**3. `check:changeset` restored** in CLAUDE.md. Accidental one-line clobber, exactly as diagnosed.

**4. Drop tracking is version-ordered, last-op-wins** — applied in Flyway order across files, and by source position *within* a file, so the idempotent `DROP TABLE IF EXISTS` above a `CREATE TABLE` resolves to "created" as it does in the database. I checked your "zero occurrences today" claim by diffing the two semantics over the real tree: **no table changes classification**, so this is a hole closed rather than a behaviour change. `R__RefreshMetadata.sql` and any unparseable filename sort last, matching Flyway.

**5. `test.yml` paths + the coverage boundary.** Both done. `migrations/CLAUDE.md` now says what green means *precisely* — "no new table is missing its entity, and the three rosters agree" — and says plainly that column adds, type widenings, renames and metadata-only changes leave no `CREATE TABLE` and are invisible. It also names `migrations.yml` as the home for the definitive DB-bearing check (**Missing**, second bullet), and the script header names the four `mj-class-registrations.ts` manifests as a known gap (**Missing**, first bullet) rather than leaving it implicit.

## Recommendation 2 — I took it, though you left it out of the final order

I think it becomes *more* necessary once item 1 lands, not less: making the context required converts "one bad merge reds every subsequent migration PR" from noisy into **merge-blocking for everyone**. So there is now a `--compare-to=<dir>` that evaluates the base worktree and fails only on the delta, reporting inherited drift as `::warning`.

The workflow cost is two lines, because the comparison lives in the script where it can be unit-tested rather than in YAML where it can't:

```yaml
run: |
  git worktree add --detach "$RUNNER_TEMP/codegen-tail-base" ${{ steps.base.outputs.sha }}
  node .github/scripts/check-codegen-tail.mjs --compare-to="$RUNNER_TEMP/codegen-tail-base"
```

One decision worth surfacing: a base tree that **cannot** be evaluated (a PR that adds an artifact absent at base, say) degrades to **strict** — every finding counts as introduced — and says so with a warning. The opposite default would read an unreadable base as "everything was already broken", which is the failure direction that loses the check. `--json` likewise exits 2 on a misconfigured tree rather than printing `{"failures": []}`, so a delta comparison can't mistake it for clean. Both are tested.

Local `npm run check:codegen-tail` keeps the strict behaviour with no `--compare-to`, since a developer running it wants to see everything, not just their own contribution.

## Verification

- Guard green on this tree: `374 entities across 3 generated artifacts; scanned all 664 migration(s)`
- Mutation still reds with the #3737 message (deleted the `MJAIModelPriceUnitTypeEntity` declaration → 2 parity drifts, then restored)
- `--compare-to` verified against a real `origin/next` worktree: `0 pre-existing`, exit 0
- **34** guard tests, up from 15; **213/213** across the full `.github/scripts` suite (9 files)
- Both workflows parse; the step is confirmed inside the required context
- `npm run check:claude-md` passes

No changeset, per your read — no published package changes, and `changes.yml`'s minor-changeset rule only fires when migration files change, which this PR has none of.

One thing I have *not* done: extending `migrations.yml` with `mj codegen && git diff --exit-code`. That is the check that would actually close the class, and it's a separate PR with real runtime cost to think about — I'd rather not smuggle it in here. Happy to open it if you think it should be queued now rather than named.

PR #3742, opened 2026-08-11
fix(changesets): invert the bump rule on certified lines

SDesai-BC:
Both follow-ons @SDesai-BC raised when merging #3739. Both reproduced before changing anything.

## 1. The gate rejected a canon-correct line backport

`DEFAULT_BASE` was `origin/next` with no branch awareness, so the Edge rule (`migration ⇒ minor`) applied everywhere — including certified lines, where `plans/lts-process.md` §12 says **everything is a patch**: metadata migrations, CodeGen repairs, and even schema migrations under a security exception.

Reproduced exactly as reported — a cert fix carrying a migration with the canon-correct `patch`:

```
Branch touches migration — minor is required.
❌ .changeset/certfix.md — this branch changes migration, but no entry declares minor
EXIT=1
```

**It is worse than a wrong answer.** Lines are patch-only *forever* (§3.1), so a `minor` there consumes the tuple the **next certification** is targeting. The gate was demanding a level the release process forbids, on security-driven fixes — the highest-stakes change the repo makes — and doing it against the rule file's own documented exception.

**Fixed by inverting rather than refusing.** Refusing would leave lines unguarded, and lines are exactly where a wrong level costs most. When the base is `lts/*`:

- `patch` is the only correct level; `minor` is rejected with the §12 citation
- the mirror check (which demands a `minor` for a DB change) is skipped entirely
- the failure legend switches, and points at `dbImpact` as the DB signal for a line release

Same reproduction, after:

```
Base 'lts/6.1' is a certified line — every entry must be patch, including the migration this branch carries.
✅ 1 changeset(s) added in this branch — bump levels are correct.
```

**Also closes a local trap that was not reported:** `npm run check:changeset` on a line branch defaulted to `origin/next` — wrong base *and* wrong rule at once. Which line a backport targets is not derivable from a topic branch name, so there is no honest default. It now refuses and asks for `--base` rather than answering confidently:

```
❌ On line branch 'lts/6.1-local' there is no safe default base.
   Pass the line explicitly, e.g. --base lts/6.1 — defaulting to origin/next would
   apply the Edge rule to a certified line and demand a level lines cannot carry.
```

Detection is by **ancestry, not branch name** — a line tip being a strict ancestor of HEAD. Real backports are not named `lts/*` (`fix/codegen-isa-postgres-lts5` → `lts/5`; the backport bot emits `backport-<n>-to-<target>`), so a name check misses exactly the branches the rule exists for, and misses a detached HEAD too.

**On the CI question this qualifies** — agreed, and the ordering was right. Wiring the gate in as-is would have failed legitimate cert-fix backports. That is reconciled now; whether to enforce remains open.

## 2. "Why one stray `minor` matters" contradicted the file's own analysis

Twenty lines after stating a `minor` *"does not visibly bump anything mid-stream"*, the section claimed *"three entries, 301 version bumps"*. Under permanent pre mode the stream is `X.Y.0-edge.N` either way, and every Edge release advances all ~300 packages regardless of level — a stray `minor` causes **no additional movement**.

Retitled to **"Why the accumulated level matters"** and led with the direction that actually costs a release, exactly as suggested: a **missing** `minor` in a cycle containing a migration leaves the accumulated tuple at `X.Y.Z+1` and breaks the minor-or-higher-tuple invariant for the whole cycle — and it is the failure that hides, surfacing only when no other changeset in the release happens to carry a `minor`.

The stray-`minor` case is kept as the cheap direction, with what it actually costs: not version movement, but telling the next author that minor-for-a-feature is normal. The 301-row table stays as *what the bot displays* — the `fixed` group's scope — rather than movement those three entries caused.

## Verification

- `check:changeset:test` — **27/27** (15 pre-existing + 12 new)
- `check:changeset` — passes on this branch
- `check:claude-md` — passes; 5 path-scoped rules validated

No changeset: touches no published package.


SDesai-BC:
### Code review

Found 4 issues:

1. **The motivating bug still reproduces.** The refusal is keyed on `LINE_BRANCH.test(currentBranch())` — the topic branch's own name — which the docstring three lines above declares untrustworthy ("a topic branch name can never be trusted to say so"). A line backport is not named `lts/*`: the repo's only real line backport, #3592, was `fix/codegen-isa-postgres-lts5` → base `lts/5`, and `korthout/backport-action` produces `backport-<n>-to-<target>`. Neither matches `/^(?:origin\/)?lts\//`. Reproduced in a throwaway repo — branch `fix/cve-2026-1234` off `lts/6.1`, carrying a migration and the canon-correct `patch`, bare `npm run check:changeset`:

   ```
   Branch touches migration — minor is required.
   ❌ .changeset/sec.md — this branch changes migration, but no entry declares minor
   EXIT=1
   ```

   That is verbatim the failure quoted in the PR body as the bug being fixed. Detached HEAD is the same hole: `currentBranch()` returns `''`, the refusal never fires, `origin/next` is used silently. Compounding it, `origin/next...HEAD` on a line branch spans the whole fork point, so migrations already on the line are attributed to the branch — a code-only backport gets flagged for a migration it never touched.

https://github.com/MemberJunction/MJ/blob/07afe6d290557f37aa9ca4e512a648162188d3c7/.github/scripts/check-changeset-bump.mjs#L86-L99

2. **`--base` with no value regressed from a loud failure to a silent wrong answer.** `argv[baseIndex + 1]` is `undefined`, and `explicitBase ?? DEFAULT_BASE` now swallows it. Before this PR the same invocation exited 2 with `❌ Base ref 'undefined' does not resolve`; now it silently applies the Edge rule against `origin/next`. Same repo, same branch, migration + `patch`, invoked as `--base` with nothing after it — old script: `EXIT=2`; new script: `EXIT=1` with the Edge legend. `--base=lts/6.1` (equals form) is silently ignored the same way.

https://github.com/MemberJunction/MJ/blob/07afe6d290557f37aa9ca4e512a648162188d3c7/.github/scripts/check-changeset-bump.mjs#L96-L99

3. **`onLine` misses legitimate spellings of a line base.** `LINE_BRANCH` only accepts a bare or `origin/`-prefixed ref. Verified rejecting a canon-correct `patch` + migration, or accepting a `minor` on a line, for `--base refs/heads/lts/6.1` (EXIT=0, "minor is required"), `--base upstream/lts/6.1` (EXIT=0), and `--base <sha-of-line-tip>` (EXIT=1, demands minor). Meanwhile `lts/experiment` — a topic branch, not a line — triggers the exit-2 refusal.

https://github.com/MemberJunction/MJ/blob/07afe6d290557f37aa9ca4e512a648162188d3c7/.github/scripts/check-changeset-bump.mjs#L37-L39

4. **One new test passes against the unfixed implementation.** `rejects a minor on a line even with NO database change` asserts only `code === 1`, which the pre-PR Edge rule also produces (for the unrelated reason "minor, but this branch changes no migration and no metadata"). Running the four new tests against `origin/next`'s script: three fail, this one passes. Two mutants also survive 19/19: broadening `LINE_BRANCH` to `/lts\//`, and deleting the entire new `onLine` failure legend.

https://github.com/MemberJunction/MJ/blob/07afe6d290557f37aa9ca4e512a648162188d3c7/.github/scripts/__tests__/check-changeset-bump.test.mjs#L263-L271

Also, the PR body says "Five new tests" / "14 existing + 5 new"; the diff adds 4 (15 + 4 = 19).

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

SDesai-BC:
Adversarial review findings verified and fixed. **The headline is that this PR did not fix the bug it claims to fix** — I reproduced that before changing anything, and found a second failure while verifying it.

## 1. Name-based detection missed every real backport branch — confirmed

The refusal keyed off the *topic* branch being named `lts/*`, while the docstring two lines above said a topic branch name can never be trusted. Real backports are not named that way: this repo’s only one is `fix/codegen-isa-postgres-lts5` → base `lts/5`, and the backport bot emits `backport-<n>-to-<target>`.

Reproduced on `fix/cve-2026-1234` off a line, migration + canon-correct `patch`, bare invocation — verbatim the failure this PR quotes as the bug:

```
Branch touches migration — minor is required.
❌ .changeset/sec.md — this branch changes migration, but no entry declares minor    EXIT=1
```

**Now detected by ancestry, not names.** A line tip being a *strict* ancestor of HEAD means the branch is built on that line; lines are terminal, so a line tip is never an ancestor of a `next` topic branch. Works for any branch name and from a detached HEAD — closing that hole too. Strict ancestry matters: a ref sitting *on* HEAD is this branch, not something it is built on, and without that a branch named `lts/…` was diffed against itself with every change invisible.

Your separate over-reach point is fixed by the same change — diffing a line branch against `origin/next` spans the fork point, so the line’s already-certified migrations were attributed to the backport.

## 2. A second bug, found while verifying the first — the lookup was dead in any real clone

`git for-each-ref` treats a pattern *containing* a glob character as a full match rather than a prefix. So `refs/remotes/*/lts/` matched **nothing**, while the glob-free `refs/heads/lts/` prefix-matched fine. My fixtures created a local *and* a remote ref, so the local one always answered — **all 27 tests green while the lookup found nothing in a real clone**, which only has `origin/lts/*`.

Confirmed against this repo: the old pattern returns empty, the new one returns `origin/lts/5`. Both patterns now end in `**`, pinned by a test that deletes the local branch first so only the remote-tracking ref remains.

That one is squarely on me — and it is the same class you caught in #1: a check that looks right and is inert.

## 3–5. The rest

- **`--base` with no value** — confirmed regression from my `??`. Now exits 2. The `--base=REF` equals form is supported rather than silently ignored, and `refs/heads/lts/…`, `origin/lts/…` and raw SHAs are recognised as line bases.
- **The test that passed against the unfixed implementation** — confirmed: it asserted only `code === 1`, which the pre-existing Edge rule also produced. It now asserts the line-specific reason.
- **PR body arithmetic** — you are right: 15 existing + 4 new, not 14 + 5. Corrected below.
- **`.claude/commands/changeset.md`** still said to ALWAYS baseline against `next`, and the rule file’s example used a line that does not exist. Both fixed.

## Verification

`check:changeset:test` **27/27** (15 pre-existing + 12 new). `check:changeset` passes on this branch; `check:claude-md` passes. The `next`-stream path is unchanged — verified against this repo’s real `origin/lts/5`, which is correctly *not* an ancestor of a `next` branch.

Thanks for this — two of the four confirmed defects were things no amount of re-reading my own diff would have surfaced.

.github/scripts/check-changeset-bump.mjs
```diff
@@ -48,9 +60,93 @@ function git(args) {
...
+ * emits `backport-<n>-to-<target>`. A name-based check therefore misses exactly the branches the
+ * line rule exists for, while a line tip being an ancestor of HEAD is decisive: lines are terminal
+ * (fixes land on `next` first and are cherry-picked over), so a line tip is never an ancestor of a
+ * `next` topic branch. It also works from a detached HEAD, which CI checkouts usually are.
+ *
+ * @returns The most specific matching line ref, or `null`.
+ */
+function detectLine() {
+    const head = git(['rev-parse', 'HEAD']);
+    const candidates = knownLineRefs().filter(
```
rkihm-BC, line 122: **A freshly-cut line is an ancestor of every subsequent `next` branch, so Edge work gets judged as line work.**

The candidate filter asks only "is this ref a strict ancestor of HEAD?". Between the moment a line is branched from the release tag and the moment it receives its first line-only commit, the line tip *is* a commit on `next` — so every `next` topic branch cut after it matches.

Reproduced:

```
### JUST-CUT line (tip == a commit on next, no line-only commit yet)
Base 'origin/lts/6.1' is a certified line — every entry must be patch,
including the migration this branch carries.
```

That's Edge work carrying a migration, where `minor` is correct — and the inverted rule would now reject it. Same shape as the bug this PR fixes, mirrored.

The window is short in the happy path (runbook step 6 pushes a release commit to the line almost immediately), but it isn't zero, and it stays open indefinitely for a line branch that is cut and then withdrawn, or pushed ahead of its publish.

**The fix already exists in this file** — line 145 uses exactly the right test on the explicit-base path:

```js
isAncestor(explicitBase, ref) && !isAncestor(explicitBase, DEFAULT_BASE)
```

That second clause is the divergence check: a ref still wholly contained in `next` is not yet a line for rule purposes. `detectLine` needs the equivalent:

```js
(ref) => isAncestor(ref, 'HEAD') && git(['rev-parse', ref]) !== head && !isAncestor(ref, DEFAULT_BASE)
```

One clause, and it makes the two paths agree — which is worth something on its own, since an explicit `--base` and an inferred one currently answer differently for the same ref.

.github/scripts/check-changeset-bump.mjs
```diff
@@ -48,9 +60,93 @@ function git(args) {
...
+/**
+ * Every certified-line ref this clone knows about, local or remote-tracking.
+ *
+ * Both patterns end in `**` deliberately. `git for-each-ref` treats a pattern CONTAINING a glob
+ * character as a full match rather than a prefix, so `refs/remotes/*​/lts/` matches nothing —
+ * while the glob-free `refs/heads/lts/` does prefix-match. A fixture carrying both a local and a
+ * remote ref hides that completely: the local one answers, and the lookup looks fine while being
+ * dead in a real clone, which has only `origin/lts/*`.
+ */
+function knownLineRefs() {
```
rkihm-BC, line 103: Worth a line in the rule file: **detection is only as good as the refs the clone has.** With no `lts/*` ref present — a shallow or single-branch clone, which is `actions/checkout`'s default — `knownLineRefs()` returns empty, `detectLine()` returns `null`, and the Edge rule applies:

```
### line refs NOT fetched (shallow/partial clone)
Branch touches migration — minor is required.
```

On a line PR in that environment that's the original bug again, silently. It doesn't bite today because there's no CI wiring — but this is the second time the CI question has changed the answer to a design decision, so it's worth writing down now rather than rediscovering it at wiring time: **CI must pass `--base` explicitly** (it knows the PR's base ref anyway, and an explicit base is authoritative), or fetch the line refs.

Separately, the glob note above it is a good catch and worth keeping verbatim — `refs/remotes/*/lts/` matching nothing is exactly the kind of thing that looks fine in a fixture and is dead in a real clone.

.claude/rules/changesets.md
```diff
@@ -52,21 +52,40 @@ Two consequences worth knowing before you reason about the number:
...
+  may well contain a migration. On a line the rule **inverts**: `patch` is not merely sufficient, it
+  is the only correct level, because a `minor` there consumes the tuple the next certification is
+  targeting. `check:changeset` works this out from **ancestry, not branch names** — a line tip being
+  an ancestor of your HEAD means you are on that line, which holds for `fix/cve-…`-style backport
+  branches and from a detached HEAD, neither of which a name check can see.
 - **The number cannot carry per-package semver.** All ~300 packages share one `fixed` group, so a
   consumer already receives bumps driven entirely by packages they do not use.
 
-## Why one stray `minor` matters
+## Why the accumulated level matters
```
rkihm-BC, line 63: This section now says what the file's own analysis supports — leading with the missing-`minor` direction (the failure that hides), and keeping the stray-`minor` case with its real cost, which is meaning rather than version movement.

The reframing of the 301-row table as *"an accurate picture of the group's scope, not of 301 version movements caused by those three entries"* is more precise than the softening I suggested — it keeps the measurement while correcting what it demonstrates.

One addition worth making while you're here: the section explains what the level does on Edge, but a reader arriving from a backport now needs the line case in the same breath. The bullet above covers it, but this section reads as the canonical explanation and doesn't mention that on a line the accumulated level is fixed at patch regardless of what the branch carries.

rkihm-BC (commented):
Reviewed at `9abf84d` — note the head moved mid-review. I had a review drafted against `07afe6d` with three findings about the name-based `LINE_BRANCH` heuristic; **the ancestry rework closed all three before I posted**, so this reviews the current state instead.

For the record, those findings were: a backport topic branch got the Edge rule (the reported bug, surviving on the branch shape the runbook prescribes), detached HEAD disabled the refusal, and the regex missed non-`origin` remotes. I verified each is now fixed:

| Case | Result at `9abf84d` |
|---|---|
| `backport-123-to-lts/6.1`, no `--base` | line detected → inverted ✅ |
| `fix/codegen-isa-postgres-lts6`, no `--base` | line detected → inverted ✅ |
| detached HEAD on a line commit | line detected → inverted ✅ |
| `next` topic branch with a migration | Edge rule, `minor is required` ✅ (no false positive) |
| explicit `--base upstream/lts/6.1` | inverted ✅ |

**Ancestry is the right answer and it's better than what I asked for.** I suggested widening a regex; replacing the heuristic with a decisive test is the stronger move, and the justification in `detectLine`'s docstring — that real backports are named `fix/codegen-isa-postgres-lts5` and `backport-<n>-to-<target>`, so a name check misses exactly the branches the rule exists for — is the argument that makes it obviously correct rather than merely defensible.

Three things worth calling out as good work:

- **The `for-each-ref` glob discovery.** That a pattern *containing* a glob is a full match, so `refs/remotes/*/lts/` silently matches nothing, is a genuine trap — and the note that a fixture carrying both a local and a remote ref would hide it completely is the part that makes the comment worth keeping.
- **The strict-ancestor self-match guard.** Without it a branch named `lts/…` resolves to itself and gets diffed against itself, making every change invisible — a silent pass, the worst failure shape for a gate.
- **Catching that `--base` with no value became `undefined` and got swallowed by `?? DEFAULT_BASE`** — that silently applied the Edge rule instead of failing. Found and fixed without being reported.

**Verification:** 27/27 tests, `check:changeset` clean on the branch, `check:claude-md` passes. The `.claude/commands/changeset.md` update correctly scopes the old "ALWAYS use `next` as baseline" instruction, which would otherwise have contradicted the new behaviour.

Two residual items inline — one real, one a doc note. Neither blocks.

<sub>Scope: exercised the gate end-to-end against throwaway repos for seven branch/base shapes including the two failure modes the new design introduces; read the full detection path; ran the suite, the gate, and `check:claude-md`. I did not run a workspace build — nothing here touches build output.</sub>

SDesai-BC:
## Smoke testing — five more defects, four of them mine from this branch

Drove the real script against repo shapes the unit tests never built. Results, then fixes:

| Case | Before |
|---|---|
| Invoked from a **subdirectory** | **crashed** — uncaught `ENOENT` |
| **CRLF** changeset (Windows contributor) | valid file rejected as `no front matter` |
| Trigger path with **non-ASCII** characters | **invisible** — correct `minor` rejected |
| `--base=` (empty equals form) | silently ignored |

**The non-ASCII one is the same defect class as the two already fixed here.** git C-quotes such paths by default — `"metadata/caf\303\251.json"` — which no trigger pattern matches. So a real metadata change went unseen and the gate confidently told the author to use `patch`. Fixed with `core.quotePath=false`.

**A fifth, introduced while fixing those and caught by the suite:** my explicit-base line check asked *"is the base an ancestor of a line?"* — which is true of `next` itself, since every line was cut from it. That declared the entire Edge stream a certified line and applied the patch-only rule to it. It now tests **identity** (the base resolves to a line tip), not ancestry.

That is three separate bugs in this branch whose common shape is a check that looks right and answers wrongly — the same thing the review caught. Worth weighing when deciding whether this ever gates CI.

### Also verified, no changes needed

- a **linked git worktree** (where `.git` is a file, not a directory) — MJ agents work in these
- a **line cut from another line** resolves to the newest
- paths containing **spaces**; unquoted and extra-whitespace front-matter forms
- the **real repo** from both the root and `packages/AI/Core`, and `--base origin/lts/5` against the actual line ref

`check:changeset:test` **32/32** (15 pre-existing + 17 new). `check:claude-md` passes.

.github/scripts/check-changeset-bump.mjs
SDesai-BC: Confirmed on current HEAD and fixed — this is a real one, and it is the mirror of the bug this branch exists for. Reproduced before changing anything: Edge work carrying a migration with the correct `minor`, rejected by the patch-only rule because a just-cut line ref was an ancestor of it.

Fixed with your divergence clause — a ref still wholly contained in `next` has not become a line yet.

**One correction to the suggested one-liner:** the explicit-base path no longer has that check. I changed it in `78b64f7d` during smoke testing, because `isAncestor(explicitBase, ref)` declared `next` ITSELF a line — every line is cut from `next`, so `next` is an ancestor of all of them — which applied the patch-only rule to the entire Edge stream. That path now tests identity: the base resolves to a line tip.

So the two paths still answer differently for a just-cut ref, and I have kept that deliberately: an explicit `--base` is a *declaration* of where the change lands, while inference has to be conservative and prefer the Edge default. Same input, different confidence, different answer. Happy to reconsider if you read that asymmetry as a trap rather than a feature.

.github/scripts/check-changeset-bump.mjs
SDesai-BC: Written into the rule file, in the section on running the check:

> **Detection is only as good as the refs your clone has.** The line rule is inferred from ancestry, so a clone with no `lts/*` refs — a shallow or single-branch checkout, which is `actions/checkout`s default — finds no line and falls back to the Edge rule. On a line PR that is silently the original bug. **Anything automated must pass `--base` explicitly** (CI knows the PR base ref, and an explicit base is authoritative) or fetch the line refs first.

Agreed on the meta-point, and it is worth stating plainly: this is the second time the CI question has changed the answer to a design decision. Recording the constraint now costs a sentence; rediscovering it at wiring time costs a broken backport.

My own smoke round covered the shallow-clone case from the other side — a shallow clone with no `origin/next` fails *loudly* (exit 2, unresolvable base) rather than answering wrongly. The gap you found is the one where the refs are partial rather than absent, which is worse precisely because it stays quiet.

.claude/rules/changesets.md
SDesai-BC: Added — you are right that this section reads as the canonical explanation, so a backport reader landing here needs the line case in the same breath rather than three bullets up:

> **None of this applies on a certified line.** There the accumulated level is fixed at `patch` whatever the branch carries, so neither direction above is available to get wrong — a migration backport is a patch, and a `minor` is the error. If you arrived here from a backport, that bullet above is the whole rule for you.

SDesai-BC:
All three addressed — one real bug fixed, two docs added. Thanks for reviewing the moved head rather than reposting against `07afe6d`.

## The just-cut-line window — real, reproduced, fixed

Between a line being cut and its first line-only commit, the line tip *is* a commit on `next`, so it is an ancestor of every Edge branch cut afterwards. Reproduced on current HEAD before changing anything: Edge work carrying a migration, with the correct `minor`, rejected by the patch-only rule.

Fixed with your divergence clause. **One correction:** the explicit-base path no longer carries that check — I replaced it in `78b64f7d` during smoke testing, because `isAncestor(explicitBase, ref)` declared `next` *itself* a line (every line is cut from it), applying the patch-only rule to the whole Edge stream. That path now tests identity.

So the two paths still differ for a just-cut ref, deliberately: an explicit `--base` is a declaration of where the change lands; inference must be conservative. Flagging it because you specifically valued the two paths agreeing — if you read that asymmetry as a trap rather than a feature, say so and I will make inference authoritative too.

## Both doc items landed

The clone-completeness constraint is now in the rule file with the CI implication stated (`--base` must be explicit, or fetch the line refs). Your meta-point is the part worth keeping: that is twice now the CI question has changed a design answer.

The accumulated-level section now opens the line case in the same breath, since it reads as the canonical explanation.

## Since your review — a second smoke round

Round 1 (before your review) found four defects. Round 2 covered history shapes, git states and semantic correctness and found **zero**: branches that merged `next` back in, a migration added-then-reverted, *deleting* a migration counting as a DB change, shallow clones failing loudly, modifying another branch pending changeset not counting as authorship, `major` and mixed levels on a line, linked worktrees, and case folding.

Two behaviours it surfaced are now pinned as tests rather than accidents: a **renamed** changeset is not judged (git reports `R`, which `--diff-filter=A` excludes — deliberate, since re-judging another branch content against this branch DB context is the cross-branch false rejection the scoping rule prevents; accepted cost is that editing via rename escapes), and a changeset added-then-deleted in-branch falls to the documented presence gap rather than a wrong level.

**35/35 tests**, `check:changeset` and `check:claude-md` pass, rebased on latest `next`.

SDesai-BC:
@rkihm-BC — all three of your comments are fixed and answered on their threads above. Since then I ran a third smoke round aimed squarely at the fix you prompted, plus a full sweep. Nothing new found; posting the evidence rather than asking you to take it on trust.

**Round 3 — the divergence window, both directions.** Fixing the just-cut case by breaking the diverged case was the live risk, so both are pinned:

| Case | Result |
|---|---|
| Just-cut line (tip still on `next`) vs Edge work + migration | Edge rule, `minor is required` ✅ |
| **Diverged** line vs its backport | line detected, inverted ✅ |
| Explicit `--base origin/lts/6.1` on a just-cut line | honoured — declared intent ✅ |
| Explicit `--base origin/next` | Edge rule ✅ |
| No `lts/*` refs at all (partial clone) | Edge rule, no crash ✅ |
| Three lines present | picks the containing one (`origin/lts/6.2`) ✅ |

**Full sweep:** smoke **32/32** across three rounds (invocation/encoding/paths/args/worktrees; history shapes/shallow clones/lifecycle/case folding/line corners; divergence/base-resolution). Unit **35/35**. The seven AI package suites still green — Core 247, OpenAI 179, Gemini 106, Inworld 63, ElevenLabs 85, AssemblyAI 48, Agents 1872. Live ElevenLabs **16/16** and the OpenAI scrub **3/3** against real APIs. CI green; branch level with `next`.

Re-requesting review since your last one was `COMMENTED` and predates the just-cut fix. Worth a look specifically at the asymmetry I flagged on your first thread: inference now refuses a just-cut line while an explicit `--base` accepts it. I argued that is intent-vs-inference and deliberate, but you were the one who valued the two paths agreeing, so it should be your call rather than mine.

For the record, four separate green-but-broken checks have now come out of this branch — name-based detection, the `for-each-ref` glob, the ancestry test that declared all of `next` a line, and your just-cut window. Each passed a full suite before being caught. That has not changed my view that this is safe to land (it gates nothing today), but it is the strongest argument yet for `continue-on-error` on the day it becomes a CI gate.

PR #3747, opened 2026-08-12
fix(explorer): make the Chat cross-entity search panel reachable and usable

CaelebB-BC:
## Problem

`SearchPanelComponent` — cross-entity search over **conversations, messages, artifacts, collections and tasks** — was rendered only by `ConversationWorkspaceComponent`, which has **zero consumers anywhere in the repo**. Explorer composes the chat UI from resource wrappers instead, so the panel never mounted on any shipped surface. The feature existed, was fully implemented, and was unreachable.

Mounting it turned out to be the easy half. Every defect below was already present and invisible, because the surface nobody could open was also a surface nobody could see fail.

## Entry point

`ChatConversationsResource` now renders `<mj-search-panel>`. The trigger is an **escalation row beneath the conversation list**, shown only while the list's own filter is active:

> 🔍  Search all of Chat for "poem"  →

The first attempt put a "Search everything in Chat" button above the list, which sat directly on top of the existing "Search conversations…" input — two identically-styled fields offering different reach, forcing a scope decision *before* the user knew whether the narrow one sufficed. The escalation row removes that: one field, and the option to widen appears exactly when the narrow search may have disappointed. The typed term carries into the panel, so nothing is retyped.

Ctrl+K is deliberately **not** the shortcut — Explorer's global command palette already owns it and is a different feature.

Layering: `mj-conversation-list` emits a new `searchEscalated` output rather than routing itself (widgets must not route); `ChatConversationsResource` decides what that means. `SearchPanelComponent` gains an `initialQuery` input to receive the term.

## Defects fixed

Each of these failed silently, which is why the panel looked wired up while returning nothing.

**Message search could never match.** The filter named `vwConversations` unqualified, but the SQL login's default schema is `dbo`, so it resolved to nothing and the query errored outright:

```
Msg 208, Level 16: Invalid object name 'vwConversations'.
```

Every sub-search reports failure as `return []`, so a hard SQL error and "no matches" render identically. Now resolved from entity metadata (`[SchemaName].[BaseView]`), matching what `searchTasks` in the same file already did.

**Artifact results did not open.** Routing branched on `collectionId`, then `conversationId` — but `mapArtifactToSearchResult` only ever sets `collectionId`, so the second branch was dead on arrival, and an artifact in no collection matched neither and the click did nothing at all. Artifacts need no parent: `NavigationService.OpenArtifact` opens one directly, which is what `ArtifactResource` exists for.

**Results did not render until an unrelated DOM event.** The emission does not reliably schedule a change-detection pass, so results landed on the component while the panel kept painting the previous state until a click, Enter, or even the blur from focusing devtools triggered the next one. Now applied through `detectChanges()`; `markForCheck()` cannot help when no next pass is scheduled.

**Quadratic allocation per keystroke.** `isResultSelected()` is bound once per row and rebuilt a flat array of every result on each call — with `MaxRows: 100` per type that is up to 500 fresh 500-element arrays per change-detection pass, on every keystroke. The flat list is now cached per emission.

**Focus theft.** `ngOnChanges` fires for every input, so a `currentUser` or `environmentId` re-emit while the panel was open pulled focus out of whatever field the user was in. Replaced with the setter pattern `packages/Angular/CLAUDE.md` requires, keyed on the open transition.

**The search icon escaped its input.** Explorer's app-wide `_shared-patterns.scss` absolutely-positions any bare `.search-icon` and pairs that with a compensating `padding-left` scoped to `.mj-search-box` / `.search-container` / `.mj-search`. This wrapper is none of those, so it got the half that breaks layout and not the half that compensates — the icon positioned against the overlay while the input slid into its vacated flex slot. Emulated encapsulation stops styles leaking *out*, not global rules leaking *in*, so the component opts out explicitly.

## Testing

DOM specs for both surfaces — the panel had **none** before.

- `search-panel.component.dom.test.ts` (3) — drives a real `input` event rather than calling the handler, so it covers the `[(ngModel)]` + `(input)` write-through that a direct-call test cannot.
- `conversation-list.component.dom.test.ts` (+4) — the escalation row: hidden with no filter, appears echoing the term, emits the **trimmed** term instead of routing, hidden in selection mode.

**1086 tests pass** across 93 files. Both packages build clean (`ngc`). `npm run check:ui` and `npm run check:changeset` pass.

## Not yet verified

**No end-to-end browser pass.** Every fix above is verified by build, unit/DOM tests, live-database queries, or a reproduction harness — but the running app is behind a login wall this branch's work never got past. In particular the change-detection fix was diagnosed from the observed behaviour (results appearing on blur) rather than watched working, and it is the one item most worth confirming manually before merge.

## Follow-ups deliberately left out

- `searchArtifacts` runs a `RunView` **per artifact** to find its collection. Now that routing no longer depends on that, the N+1 only earns a display subtitle — worth batching, separately.
- Three other components use `.search-icon` with neither their own `position` nor one of the three wrapper classes that global rule compensates for: `query-browser-resource`, `component-browser`, `search-overlay`. The durable fix is scoping that global rule to its own wrappers, which has its own blast radius.
- All five sub-searches swallow failures into `console.error` + `return []`. That is why a hard SQL error read as "no results" for however long this has shipped. Surfacing failures to the panel is a larger change than this PR should carry.

## Screenshots

Screenshot 2026-08-14 at 10 28 29 AM
Screenshot 2026-08-14 at 10 28 41 AM


CaelebB-BC:
## Verified end-to-end ✅

Rebuilt the regression Explorer image with this change and re-ran the regression suite against a pristine database:

```
Passed  54s  T098 - Chat Global Search Panel
```

T098 had failed on every prior run (`Impossible — 1/4 criteria met`, agent could only ever reach the global Ctrl+K command palette). With the panel mounted and the sidebar trigger in place it now passes, exercising the panel open, the scoped filter buttons, and close.

The result-routing paths (artifact / collection / task) are type-checked and compile, but T098 only exercises panel open + filter render + close — so those three routes are still unverified at runtime. Worth a look during review.

CaelebB-BC:
Merged `origin/next` (133cdfeb48) into this branch. It had conflicts in 4 files; the merge commit (224e81eaa1) describes how each was resolved. In short: `next` renamed public members to PascalCase with deprecated camelCase aliases (#4487). I kept that, and applied this branch's changes under the new names: `SearchEscalated`/`EscalateSearch()`, `InitialQuery`, `IsSearchPanelOpen`, `OpenSearch`/`CloseSearch`/`OnSearchResultSelected`.

## Verified against regression test T098 (Chat Global Search Panel)

On `next` the panel can't be opened from Explorer's Chat. Nothing mounts `mj-search-panel`.

- Typing `renewal` in the conversation list shows the row **Search all of Chat for "renewal"**.
- Clicking it opens the panel with `renewal` carried over and the filter buttons All, Conversations, Messages, Artifacts, Collections, and Tasks. It returns **Found 3 results**: 1 conversation and 2 messages, with the matches highlighted.
- Clicking **Messages** makes that filter active. **Escape** closes the panel. The page stays responsive throughout.
- Unit tests: `ng-conversations` 108 files / 1348 tests passed; `ng-explorer-core` 23 files / 194 tests passed. `npm run check:standards` reports no errors.

**Note for the suite:** T098's goal looks for a magnifying-glass icon in the Chat top bar. This PR's entry point is the escalation row under the conversation list instead, so T098's goal text needs updating to match before it can pass.

rkihm-BC (changes requested):
>

Reviewed at 6b2f1db: the full diff against `origin/next` (133cdfeb48), the description, the changeset, the plan doc, your merge comment, and the surrounding code: `SearchService`, `SearchPanelComponent` and its template, `ConversationEngine.LoadConversations`, `NavigationService.OpenArtifact`/`OpenNavItemByName`, the Collections/Tasks resource wrappers' deep-link handling, and the Conversations/Conversation Details entity permissions and RLS metadata. CI is green on this head.

## What's good

- The root-cause work holds up. `mj-search-panel` really was mounted only by `ConversationWorkspaceComponent`, which Explorer never renders. The widget/host split is right: `mj-conversation-list` emits `SearchEscalated`, and `ChatConversationsResource` owns the routing (L1 emits, L3 routes, no Router in Generic).
- The message-search diagnosis is right: an unqualified `vwConversations` doesn't resolve under a `dbo`-default login. Building `[SchemaName].[BaseView]` from metadata matches what `searchTasks` already does, and the PostgreSQL provider rewrites bracket identifiers in `ExtraFilter` to double quotes (PostgreSQLDataProvider.ts:1030), so the brackets are fine on both platforms.
- Artifact routing through `NavigationService.OpenArtifact(id, title)` fixes the dead click for artifacts outside a collection. The mapper only ever sets `collectionId`, so the old branch really could never match. The Collections and Tasks nav labels match `.chat-application.json`, and both wrappers read `collectionId` / `taskId` from the merged configuration (chat-collections-resource:310, chat-tasks-resource:85).
- Caching `flatResults` per emission, and the O(1) `IsResultSelected`, are behavior-preserving. So is the `IsOpen` setter that fires only on the open transition, and deferring `InitialQuery` to `setTimeout` correctly handles input-binding order. The deprecated camelCase aliases route through the new setter, so `[isOpen]` bindings still work.
- In a zoneless Explorer, `detectChanges()` after service emissions matches the established pattern. The `initializing` guard around the synchronous BehaviorSubject replay is a nice touch.
- The changeset is present, and `patch` is right: the branch has no migration or metadata.

## Worth addressing

1. **The newly reachable search returns other users' conversations and message text.** search.service.ts:247 (conversations) and :298 (messages) scope only by `EnvironmentID`. The list this panel escalates from scopes much more tightly. `ConversationEngine.LoadConversations` (conversations.ts:664-677) requires `UserID = <me> OR ID IN (<shared via Resource Permissions>)`, plus `ApplicationScope IN ('Global','Both')`. The only RLS on these two entities is for the Widget Guest role (`metadata/entity-permissions/.entity-permissions.json:166,185`), and no shipped migration defines one, so a standard UI user gets no server-side narrowing.
   Failure scenario: two users share the default environment. User A types "salary" into the list filter and clicks "Search all of Chat for 'salary'". The Conversations group lists B's conversations whose Name or Description matches. The Messages group shows 150-character previews of B's message text. Clicking a result calls `OnConversationSelected` with B's conversation ID (chat-conversations-resource:1461-1473). App-scoped conversations that the main list deliberately hides also show up.
   This code predates the PR, but on `next` nothing Explorer renders mounts `mj-search-panel`, and (per the PR's own diagnosis) the message query errored on SQL Server. So this PR is what exposes it.
   The underlying gap is on the server. Reads of these entities are gated only by role-level `CheckUserReadPermissions`, and `MJConversationDetailEntityExtended`'s gate covers `Save`/`Delete`, not reads. So any `RunView` can already fetch another user's conversations. That deserves its own security issue. It doesn't change what this PR needs, which is not to put that data in the UI.
   Suggested fix: scope both searches to the conversations the user can already see. Move `LoadConversations`' ownership/share/scope clause (conversations.ts:664-677) into one helper on `ConversationEngine`, so the list and the search can't drift. `searchConversations` applies it directly. For `searchMessages`, the simplest version adds the same clause inside the existing subquery. That has no row cap, but see item 2 for PostgreSQL. Don't build the filter from the engine's cached `Conversations$` list. That load is capped at `MaxRows: 1000`, so older conversations would silently become unsearchable. It's also tied to `_lastEnvironmentId`, which nothing guarantees matches the panel's `EnvironmentId`. `searchArtifacts` (:342) has the same shape: it filters on `EnvironmentID` and `Visibility` only, and `MJArtifactEntity` has a `UserID`. Please confirm the intended visibility there as well. A unit test asserting that the conversation/message filters carry the user scope would lock this in.

## Minor

2. **Message search probably still errors on PostgreSQL, but not because of the brackets.** The provider converts `[x].[y]` to `"x"."y"`. But it coerces `= 0`/`= 1` to `FALSE`/`TRUE` only for boolean fields of the entity being queried (PostgreSQLDataProvider.ts:697-713). Here that's Conversation Details, so `IsArchived=0` inside the Conversations subquery stays an integer comparison against a `BOOLEAN` column, and PostgreSQL rejects it. I found this by reading the code, not by running it. It predates the PR, but the PR says message search is fixed. Either note that it's SQL Server-only for now, or do the scoping in item 1 as a first `MJ: Conversations` RunView (`Fields: ['ID']`), then filter messages with `ConversationID IN (...)`. Each filter then runs against its own entity, so the coercion applies. The trade-off is an `IN` list that grows with the user's conversation count.
3. **The changeset uses pre-#4487 member names.** `.changeset/explorer-chat-cross-entity-search.md:18-19,46` says `searchEscalated`, `initialQuery` and `SearchPanelComponent.resultSelected`. `searchEscalated` doesn't exist: the output is `SearchEscalated`, with no camelCase alias, so a consumer who binds `(searchEscalated)` from the notes gets nothing and no error. `initialQuery` and `resultSelected` still exist, but only as deprecated aliases of `InitialQuery` and `ResultSelected`. Release notes are built from changesets, so please use the canonical names.
4. **The plan doc's resolution contradicts the code.** `plans/explorer-chat-search-panel-missing.md` (Resolution section) describes a "**Search everything in Chat** button above the conversation list" and says T098 "passes in 54s". The shipped entry point is a filter-gated escalation row *below* the list. Your own PR comment says T098's goal text must change before it can pass. Please align the doc, and track the T098 goal update somewhere so the suite doesn't stay red.
5. **The new panel DOM test can't catch the render-without-a-DOM-event bug the PR fixes.** In `search-panel.component.dom.test.ts`, the two DOM specs (:46, :69) call `f.detectChanges()` after the async settle and before asserting (:52, :84). The middle spec (:57) only checks component state. All three would pass without `applyState`. Assert on the DOM after the settle, with no manual `detectChanges()`. Also consider one spec each for `InitialQuery` seeding on open and for `OnSearchResultSelected` routing, especially the artifact branch this PR fixes.
6. **The description overstates one change.** It says message search "escapes the term with `EscapeSQLString`". That escaping was already on `next` (the `LIKE` line is unchanged in the diff). Only the view qualification is new.

## Nits

- chat-conversations-resource template, `@if (currentUser)` next to `mj-search-panel` (:114): use `CurrentUser`, since the camelCase getter is the deprecated alias. The existing blocks at :42 and :75 do the same, so this is about new code only.
- `OnSearchResultSelected` drops the promises from `OnConversationSelected` and `OpenNavItemByName`. Prefix them with `void`, as the rest of this file does (e.g. `void this.ChatArea?.reloadMessages()`).
- public-api.ts: `export * from './lib/services/search.service'` exports the whole module, not just the two types the comment names. That makes `SearchService`, `SearchFilter` and `DateRange` public too, and `SearchService`/`SearchFilter` share names with `@memberjunction/ng-search`'s exports. `export type { SearchResult, SearchResultType } from ...` does what the comment says.

## Not blocking — pre-existing, now user-reachable

- `OnSearchInput` runs a full search on every keystroke with no debounce (the service's `debounceTime` only feeds recent searches). Responses can land out of order, so an earlier, slower search can overwrite the newer result, and `isSearching` can flip to false while a later search is still in flight.
- `searchArtifacts` runs one `MJ: Collection Artifacts` RunView per matched artifact, serially, for up to 100 artifacts per keystroke. That was invisible while the panel was unreachable. Worth a follow-up: one batched query, plus debounce/cancel on the panel.

## Overall

**Request changes**, for item 1. The wiring, routing and rendering fixes are sound. But by making this panel reachable and repairing message search, the PR lets any user see other users' conversation names and message snippets. Scope both searches to the conversations the user can already see, using the same rule as the list. Items 2-6 are quick cleanups.

PR #3751, opened 2026-08-12
v6.1.0-edge.2

MS-BC:
# Unified workflow execution, discoverable skills button, and robust database provider capabilities

## New Features
- **Discoverable Skills Button in Composer:** Adds a dedicated Skills button to the composer beside Plan Mode, offering a visible path to use `/` skill commands. Features keyboard toggle, dropdown baseline filtering, and proper ARIA role announcements (`aria-expanded` and `aria-haspopup`).
- **Unified Workflow and Flow Agent Form:** Retires the Workflows application and elevates the Flow Agent form to be a first-class citizen with tabbed sections (Designer, Details, Invocations). Includes a read-only index of automated invocation pathways.
- **Durable Workflow Execution via Dispatcher:** Flow agents now run directly on the durable task-graph dispatcher instead of walking graphs locally. Supports nested `ForEach` and `While` loop executions, parallel batch execution, and automatic layout generation for the graph view.
- **Workflow "Create" Front Door:** Introduces a structured "Create Workflow" landing experience with three intuitive starting options ("Describe it", "Build visually", and "Promote a run"), aligned with the user-experience mockups.
- **Agnostic Realtime Voice Support:** Introduces a provider-agnostic `realtime.voice.default.voice` configuration key to share a voice ID across ElevenLabs, Gemini, Inworld, AssemblyAI, and HuggingFace, rather than hardcoding to OpenAI.
- **First Message Support for Realtime Voice Agents:** Realtime voice agents can now be configured to speak first in a session via the vendor-neutral `realtime.voice.default.firstMessage` persona slot, resolving initial silence in ElevenLabs and AssemblyAI sessions.
- **Speech-to-Text with Groq Whisper:** Adds a fully functional `GroqAudioGenerator` transcribing audio sequentially using Groq's Whisper models, complete with a sequential audio chunk splitter for files over 25MB, and introduces a dedicated `Speech to Text` model type.
- **ModelConfiguration Deep Merge Cascade:** Implements a JSON-driven capability cascade (`AIModelType` < `AIModel` < `AIModelVendor`) for strongly-typed configurations like realtime turn-detection parameters (`semanticVad`), allowing advanced models to opt into smarter turn-handling.

## Improvements
- **Transition-Aware Action Filters:** Enhances entity action filters to evaluate transition states (e.g., "when Status *becomes* Approved") by exposing the pre-save state (`OldValues`) via `EntityField.OldValue` inside `ActionFilterContext`.
- **Configurable Entity Search Timeout:** Lowers the default entity-search timeout from 30,000ms to 3,000ms and makes it fully configurable via statics or the environment variable `MJ_SEARCH_PER_ENTITY_TIMEOUT_MS`.
- **Row-Limited Query Determinism:** Every row-limited query now falls back to ordering by its primary key when no ordering is supplied. This ensures determinism across page 1 and subsequent pages of keyset pagination walks, eliminating long-standing duplicate-row defects.
- **GraphQL Build Schema Synchronous Scaling:** Resolves an O(N²) scaling bottleneck in `type-graphql` field-config and metadata scans inside `buildSchemaSync`, cutting cold boot time of MJAPI by up to 66% (from ~95s to ~32s).
- **RunView Simple Row Normalization:** RunView simple rows are now normalized at the ProviderBase choke point so that Date and numeric columns represent actual `Date` and `number` types on every tier (server-side, Redis cache, and browser over GraphQL).

## Bug Fixes
- **Action Filter Enforcement:** Fixes an issue where action filters logged refusals but failed to halt execution, ensuring that prevented runs are now correctly blocked in accordance with configuration.
- **UserCache Moved and dialect-neutralized:** **BREAKING:** Repoints `UserCache` from `@memberjunction/sqlserver-dataprovider` to `@memberjunction/generic-database-provider` to allow PostgreSQL deployments to utilize user caching and makes its signature connection-agnostic.
- **ValidateAsync Silent Skippings:** Ensures that subclasses overriding `ValidateAsync` actually execute during saves without requiring an explicit override of the unmentioned `DefaultSkipAsyncValidation` getter.
- **Process-Wide Server Cache Corruption Fix:** Fixes a critical server cache corruption vulnerability where GraphQL field mapping mutated shared cache references by reference. The server now deep-freezes row data at write-time when using reference-sharing local storage providers.
- **Gemini Live Voice Configuration Mapping:** Fixes a silent drop in Gemini Live sessions by mapping the driver-neutral `voice` key to Gemini's native `prebuiltVoiceConfig.voiceName`.
- **xAI/Grok Realtime Voice Usage Tracking:** Resolves a silent accounting gap in xAI Grok Voice realtime sessions where token metrics were delivered top-level instead of nested, failing to record token metrics.
- **RelatedRecordCollection Safe Loading:** `RelatedRecordCollection.Load()` now throws an error instead of silently discarding unsaved additions or queued removals when loading over unsaved work.
- **SFI Erasing IsNameField On IS-A Children:** Prevents Smart Field Identification from removing `IsNameField` on virtual/inherited Table-Per-Type fields when no replacement name field is found.
- **PostgreSQL SETOF Record Routine Calls:** Resolves CodeGen pruning failures on PostgreSQL where SETOF record routines like `spDeleteEntityWithCoreDependencies` were called as a standard table select.
- **Durable Dispatcher Graceful Shutdown & Opt-out:** Resolves stop races where the dispatcher claimed tasks while shutting down, and introduces the `MJ_DISABLE_TASK_GRAPH_DISPATCHER` override to safely run integration test runner checks.
- **GraphQL Variables Redaction Gap:** Closes a redact-on-match fail-open hole for custom inputs that don't directly map to an entity (such as connection credentials), and marks `CreateConnectionInput.CredentialValues` with `@NoLog`.
- **Open App migrations under pnpm:** Resolves pnpm runtime package resolution issues for skyway packages during Open App installation and migration.
- **Dashboard Saving in Unloaded Engines:** Enables dashboard owners to save layouts under engine-unloaded processes (like sync runners), avoiding false permission denials.
- **Suite-Sequencing Guard:** Introduces a deterministic test asserting that server-transport bundles sequence before client-transport ones to prevent state corruption side-effects.


MS-BC:
## Release-prep notes for reviewers

### ⚠️ CI coverage gap on this branch — validated locally instead

`test.yml`, `integration.yml`, `pg-migrations.yml` and `build.yml` all trigger on **`next` only**, so none of them run on a `release/*` branch. DEPLOYMENT.md assumes prep commits are migrations/metadata only — but this release also carries **TypeScript changes to four packages** (MJServer, TaskGraph, CodeGenLib, MJCoreEntities), which therefore never got a CI run.

Those gates were run locally against this exact tree:

| Gate | CI equivalent | Result |
|---|---|---|
| `npx turbo run test:types` | test.yml | **374/374** |
| `npm test -- --concurrency=4` | test.yml | **607/607** |
| `RUN_MUTATION_TESTS=1 pnpm run test:integration` | integration.yml | **63/63**, 0 env gaps |
| `mj test suite "Integration Tests — Live Model"` | (not in CI) | **15/15**, all ran |
| `mj migrate` + `mj sync push` on a fresh PG database | pg-migrations.yml | **49 applied clean, push 0 errors** |
| `node scripts/check-pg-migration-content.mjs` | pg-migrations.yml | 226 counterparts, **0 suspect**, delete parity 0 mismatched |
| `pnpm run build` | build.yml | **271/271** |

Note the local integration run is a **superset** of CI's: `integration.yml` omits `RUN_MUTATION_TESTS=1`, so the seven `task-graph-execution` oracles are gated out there and have never run in CI.

### Two converter defects found while producing the PG counterparts

Both are in `mj migrate convert --split`, both would have shipped silently, and both are worked around in this PR rather than fixed at source:

1. **Drop-guards are dropped while the paired `ADD` survives.** `IF EXISTS (sys.check_constraints …) DROP CONSTRAINT` vanished for `CK_Task_Assignment`, `CK_Task_Status` and `CK_AIAgentStep_StepType`, leaving a bare `ADD CONSTRAINT`. Harmless on a fresh database, **fatal on every migrate-through deployment**. Hand-ported as `DROP CONSTRAINT IF EXISTS`.

2. **SQL Server `BIT` literals are emitted as integers.** 468 occurrences where PostgreSQL requires booleans — `column "AllowsNull" is of type boolean but expression is of type integer`. The baked converter path emits `TRUE`/`FALSE` correctly; the non-baked path does not, and a migration with a conversion gap never reaches the baked path.

Also worth knowing: `--bake-codegen` **halts at a conversion gap before baking**, so five of this release's migrations produce DDL-only counterparts. That is why `V202608112240__v6.1.x__PG_CodeGen_Regen.pg-only.sql` exists.

### One pre-existing migration corrected

`V202608082330__v6.1.x__Retire_Workflows_Application.sql` hard-coded `[__mj]` in all eight statements. It merged to `next` without the check ever running, because `changes.yml` listens on PRs into `main` only. Not yet on `main`, so unreleased and safe to correct.

PR #3762, opened 2026-08-12
fix(explorer,whiteboard): remove duplicate @HostListener declarations

CaelebB-BC:
Split out of #3380 — one root cause (in two places), independently revertable.

## Defect

Angular collects host listeners into an object keyed by event name. A second `@HostListener` for the same event **silently replaces** the first — no build error, no warning. The code reads as correct and one handler simply never runs.

Two components shipped that way:

| Component | Duplicated event | Consequence |
|---|---|---|
| `ShellComponent` | `document:keydown` | The surviving declaration didn't handle the command-palette chord, so **Ctrl/Cmd+/ never fired** — for the entire life of the feature |
| `WhiteboardHostComponent` | `document:click` | Killed the Sees dropdown's outside-click dismissal |

## Why the test is source-level, not behavioural

`host-listener-uniqueness.test.ts` asserts on the **source**, which is unusual enough to justify.

The failure mode is a duplicate *declaration*. A behavioural test can only ever observe the **surviving** handler — by construction it cannot see that a second declaration clobbered the first. A behavioural test would have passed happily throughout the entire period Ctrl/Cmd+/ was dead.

It also guards a concrete merge hazard: more than one open branch touches `shell.component.ts`, so whichever merges second resolves a conflict there, and taking "the other side" reintroduces the duplicate silently.

## Verified still present

Confirmed against `origin/next` @ `b23bf5933d`: `shell.component.ts` still has **2** `@HostListener('document:keydown')` and `whiteboard-host.component.ts` still has **2** `@HostListener('document:click')`.

### DIFF
--- a/.changeset/fix-duplicate-hostlistener.md
+++ b/.changeset/fix-duplicate-hostlistener.md
@@ -1,0 +1,18 @@
+---
+'@memberjunction/ng-explorer-core': patch
+'@memberjunction/ng-whiteboard': patch
+---
+
+Fix two components that declared the same host event twice, silently disabling a handler.
+
+Angular collects host listeners into an object keyed by event name, so a second
+`@HostListener` for the same event **replaces** the first — no build error, no warning.
+
+- `ShellComponent` declared `document:keydown` twice; the surviving handler did not
+  handle the command-palette chord, so Ctrl/Cmd+/ never fired for the entire life of
+  the feature.
+- `WhiteboardHostComponent` had the same bug for `document:click`, killing the Sees
+  dropdown's outside-click dismissal.
+
+Adds a source-level guard test, because a behavioural test can only observe the
+surviving handler — it cannot see that a second declaration clobbered the first.
--- a/packages/Angular/Explorer/explorer-core/src/lib/__tests__/host-listener-uniqueness.test.ts
+++ b/packages/Angular/Explorer/explorer-core/src/lib/__tests__/host-listener-uniqueness.test.ts
@@ -1,0 +1,53 @@
+import { describe, it, expect } from 'vitest';
+import { readFileSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+import { dirname, resolve } from 'node:path';
+
+/**
+ * A component may declare at most ONE `@HostListener` per event name.
+ *
+ * Angular's compiler collects host listeners into an object keyed by event name, so a
+ * second declaration for the same event silently replaces the first — no build error,
+ * no warning, the earlier handler simply never registers. `ShellComponent` shipped that
+ * way: a separate `handleGlobalKeyboardShortcuts` owned Ctrl+/ and was overwritten by
+ * `OnGlobalKeydown`, so the command-palette shortcut was dead for its entire life.
+ *
+ * The failure mode is a duplicate *declaration*, which is what a source check detects
+ * and a behavioural test does not — a behavioural test only ever sees the surviving
+ * handler. Guarded here because #3380 and #3033 both touch this file, so whichever
+ * merges second resolves a conflict on it, and taking "the other side" reintroduces the
+ * duplicate silently.
+ */
+
+/**
+ * Anchored to the start of a line (after indentation) so the decorator is matched but
+ * prose about it is not — `shell.component.ts` carries a "do NOT add a second
+ * `@HostListener('document:keydown')`" warning in a JSDoc block.
+ */
+const HOST_LISTENER = /^[ \t]*@HostListener\(\s*['"]([^'"]+)['"]/gm;
+
+/** Files whose host-listener declarations must stay unique per event. */
+const GUARDED = ['../shell/shell.component.ts'];
+
+function eventsIn(relativePath: string): string[] {
+  const source = readFileSync(resolve(dirname(fileURLToPath(import.meta.url)), relativePath), 'utf8');
+  return [...source.matchAll(HOST_LISTENER)].map((m) => m[1]);
+}
+
+describe('@HostListener uniqueness', () => {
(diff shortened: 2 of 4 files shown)


CaelebB-BC:
Merged `origin/next` (133cdfeb48) into this branch. It merged with no conflicts.

## Verified against regression test T153 (Command Palette)

T153 fails on `next`. Cmd/Ctrl+/ does nothing, because only `OnGlobalKeydown` is compiled into the shell's host bindings.

On this branch, checked by hand in a local Explorer with Playwright at 1280×720 (the test's viewport):

- Cmd+/ opens the command palette.
- Typing `Admin` narrows the list to Admin, AI, Permissions, and "Search Knowledge Hub". Enter opens `/app/admin/...` and closes the palette.
- Typing `zzzqqqxyz` shows **No applications found** with a "Search Knowledge Hub" fallback.
- Cmd+K still focuses the header search box, so the merged handler did not break it.
- `explorer-core` unit tests: 24 files, 196 tests passed.

rkihm-BC (changes requested):
>

Reviewed at 8c31f1be13: the diagnosis holds up. With two `@HostListener`s for the same event, only one handler registers. I confirmed this in the Angular 21.2.22 compiler the repo uses: `bindings.listeners[eventName] = ...` in `@angular/compiler-cli`, so the last declaration wins. On `origin/next`, `ShellComponent` has two `document:keydown` declarations and `RealtimeWhiteboardHostComponent` has two `document:click` declarations. Merging each pair into one handler is the right fix. I have two behavior issues to raise, both from Ctrl/Cmd+/ now reaching listeners that already claimed it, and one gap between the description and the test.

## What's good

- The root cause is correct and explained well. The comments next to both handlers will stop someone from adding the duplicate back.
- The source-level guard is the right tool here. A behavioral test would only ever see the handler that survived. The test also checks that `document:keydown` is still present, so a merge that drops the handler entirely can't pass it by accident. It fails on `origin/next`, where the shell has two keydown declarations, and passes at head.
- Whiteboard: `ToggleSees`/`ToggleExport` stop propagation, and the Sees popover root has `(click)="$event.stopPropagation()"`. So clicks on either button or inside the popover don't close it right away.
- The changeset lists `patch` for both packages. There is no Router or Explorer import in the Generic package, and no `any`.

## Worth addressing

1. **Ctrl/Cmd+/ inside a code editor now comments the line and also opens the palette.** `shell.component.ts:3438-3446`
   - The `/` branch no longer checks whether focus is in an editable field. It also doesn't look at `event.defaultPrevented`.
   - `mj-code-editor` defaults to CodeMirror `basicSetup` (`ng-code-editor.component.ts:94`). Both `basicSetup` and `minimalSetup` include `defaultKeymap`, which has `{ key: "Mod-/", run: toggleComment }`.
   - CodeMirror listens on its `contentDOM`. When a binding handles a key, it calls `event.preventDefault()` but calls `stopPropagation()` only if the binding asks for it, and this one doesn't (`@codemirror/view` `runHandlers`). The keydown therefore still bubbles to the shell's `document:keydown` listener.
   - Scenario: a user editing SQL or a prompt template in any of the ~48 templates that host `mj-code-editor` presses Cmd+/ to comment a line. The line gets commented, and then the palette opens and takes focus away from the editor.
   - This is new behavior. The old handler never ran, and even its intended version skipped editable targets when the omnibar was off. The description doesn't mention the change.
   - Fix: add `if (event.defaultPrevented) return;` before handling `/`. This lets any element that already handled the chord keep it. The palette still opens from a plain input, and from a read-only editor or one whose language has no comment syntax, because `toggleComment` returns false there and nothing is prevented. The JSDoc at 3425-3428 says the chord is "deliberately NOT gated", so update it to match.

2. **Two dashboards already bind Cmd+/ on `document`, so both handlers now fire.**
   - Component Studio's `OnKeyDown` (`component-studio-dashboard.component.ts:320`, `:340`) toggles its shortcut overlay on Ctrl/Cmd+/. Data Explorer's `handleKeyboardShortcut` (`data-explorer-dashboard.component.ts:1750`) matches `event.key === '/'` without checking modifiers, so Cmd+/ outside an input focuses its filter.
   - Both listeners are on `document`, like the shell's. The shell's `stopPropagation()` doesn't stop other listeners on the same node, and the shell's listener registers first, so the `defaultPrevented` guard from #1 doesn't help.
   - Scenario: in Component Studio, Cmd+/ opens the palette and also toggles the shortcut overlay. In Data Explorer, it opens the palette and also moves focus to the filter input.
   - This is new: until now the shell's chord never ran, so each dashboard owned it.
   - Fix: Component Studio only advertises `?` for the overlay (`component-studio-dashboard.component.html:236`, `:265`), so drop its undocumented Ctrl+/ alias. In Data Explorer, only handle `/` when no Ctrl/Cmd modifier is held.

3. **The test doesn't guard the whiteboard, but the description says it does.** `host-listener-uniqueness.test.ts:30`
   - `GUARDED = ['../shell/shell.component.ts']` covers only the shell. The description says the test "fails the build if either component declares a duplicate host listener again."
   - Scenario: a later edit adds a second `document:click` to `whiteboard-host.component.ts`, and Sees dismissal breaks silently again with CI green.
   - Fix: add the same check to `packages/Angular/Generic/whiteboard` (it already has vitest dom tests), guarding `whiteboard-host.component.ts`. Don't reach across packages from explorer-core. Or, if that's out of scope, correct the description.

## Overall

Approve once #1 and #2 are fixed. #1 is a one-line guard, and without it this PR adds a focus-stealing regression to every code editor in Explorer. #2 is two small edits in the dashboards that already claim the chord. #3 is small: either add the guard to the whiteboard package or make the description accurate. CI is green.

PR #3766, opened 2026-08-12
fix(dashboards): make the query splitter and entity-map rows keyboard-operable

CaelebB-BC:
Split out of #3380 — independently revertable.

## Defects

Two controls were mouse-only:

**Query-browser splitter** — no role, no keyboard affordance. The panel could only be resized by dragging. This is also the control that regression test T045 could never satisfy: it's a 6px-wide `background: transparent` strip with a 2px grip, invisible until hover.

**Integration entity-map rows** — click-only, with no role, no tab stop and no key handler.

## Fix

Both get real keyboard operability — the splitter as a proper separator with TS handlers and focus styling (not markup attributes alone), and the rows with `role="button"`, `tabindex`, `aria-label` and Enter/Space.

## Verified still present

Confirmed against `origin/next` @ `b23bf5933d`: **0** occurrences of `role="separator"` in `query-browser-resource.component.html` and **0** of `role="button"` in `connections.component.html`.

### DIFF
--- a/.changeset/fix-a11y-keyboard-operability.md
+++ b/.changeset/fix-a11y-keyboard-operability.md
@@ -1,0 +1,11 @@
+---
+'@memberjunction/ng-dashboards': patch
+---
+
+Make two mouse-only controls keyboard-operable.
+
+- **Query browser splitter** — had no role and no keyboard affordance, so the panel
+  could only be resized by dragging. Now a real keyboard-operable separator with TS
+  handlers and focus styling.
+- **Integration entity-map rows** — were click-only. Now `role="button"`, `tabindex`,
+  `aria-label`, and Enter/Space activation.
--- a/packages/Angular/Explorer/dashboards/src/Integration/components/connections/connections.component.html
+++ b/packages/Angular/Explorer/dashboards/src/Integration/components/connections/connections.component.html
@@ -619,7 +619,16 @@
           <!-- Table body -->
           <div class="detail-table-body">
             @for (em of DetailFilteredMaps; track em.ID) {
-              <div class="detail-map-row" [class.sync-disabled]="!em.SyncEnabled" (click)="OnEntityMapClick(em)">
+              <!-- The row is the only way to open the field-mapping editor, so it needs a
+                   role and a keyboard path. TODO: the row also contains a checkbox and a
+                   button, which makes role="button" an ARIA structural violation — the real
+                   fix is a dedicated open button (the .map-edit-hint chevron). -->
+              <div class="detail-map-row" [class.sync-disabled]="!em.SyncEnabled"
+                   role="button"
+                   tabindex="0"
+                   [attr.aria-label]="'Open field mapping for ' + em.Entity"
+                   (click)="OnEntityMapClick(em)"
+                   (keydown)="OnEntityMapRowKeydown($event, em)">
                 <span class="dt-col-toggle" (click)="$event.stopPropagation()">
                   <label class="toggle-switch" [title]="em.SyncEnabled ? 'Sync enabled' : 'Sync disabled'">
                     <input type="checkbox"
--- a/packages/Angular/Explorer/dashboards/src/Integration/components/connections/connections.component.ts
+++ b/packages/Angular/Explorer/dashboards/src/Integration/components/connections/connections.component.ts
@@ -960,6 +960,24 @@ export class ConnectionsComponent extends BaseResourceComponent implements OnIni
     this.cdr.detectChanges();
   }
 
+  /**
+   * Enter/Space on the row opens the field-mapping editor — but ONLY when the row
+   * itself is focused. The row contains its own controls (the sync checkbox and the
+   * direction button), and their `stopPropagation()` guards are on `click`, not
+   * `keydown`, which bubbles. Without the target check, Space on the checkbox both
+   * opened the editor and had its default toggle cancelled by `preventDefault()`.
+   */
+  OnEntityMapRowKeydown(event: KeyboardEvent, em: EntityMapRow): void {
+    if (event.target !== event.currentTarget) {
+      return;
+    }
+    if (event.key !== 'Enter' && event.key !== ' ') {
+      return;
+    }
+    event.preventDefault();
+    this.OnEntityMapClick(em);
+  }
+
   CloseEntityMapEditor(): void {
     this.EditorEntityMap = null;
     this.cdr.detectChanges();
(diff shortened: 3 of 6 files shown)


CaelebB-BC:
Merged `origin/next` (133cdfeb48) into this branch. It had one conflict, in `query-browser-resource.component.ts`, against the PascalCase rename in #4487. I kept `next`'s deprecated `onResizeStart` alias and added this branch's handler as `OnResizeKeydown`. The splitter template now binds `OnResizeStart` / `OnResizeKeydown`.

## Verified against regression test T045 (Query Left-Panel Navigation)

T045 fails on `next` at the checkpoint `query-panel-resize`. The splitter is a 6px transparent strip whose grip only shows on hover, and it isn't in the accessibility tree, so the test agent never finds it. Drag resizing already works on `next` for a mouse user.

- The splitter is exposed as `separator` "Resize query list panel", with `aria-valuenow`/`min`/`max`. Its grip is visible without hovering.
- Keyboard: 5× ArrowLeft moves it 600 → 560, **Home** → 200 (min), **End** → 600 (max).
- Dragging still works: 600 → 480 px. `aria-valuenow` follows every change.
- `ng-dashboards` unit tests: 160 files / 1776 tests passed.

Not verified: the Integrations entity-map row half of this PR (T120). The verification database has no integration connections, so there are no entity-map rows to render.

rkihm-BC:
>

Reviewed at b1e8162: I know this is still a draft, so this is early feedback. The only gap your latest comment names is that the entity-map half (T120) is unverified, and I've reviewed with that in mind.

## What's good

- **The splitter is a real widget now.** `role="separator"` + `tabindex="0"` + `aria-valuenow/min/max`, and the HTML comment is right that without the tabindex it would be read as a static separator. `aria-orientation="vertical"` pairs correctly with Left/Right arrows.
- **One clamp for both input paths.** Drag and keyboard both go through `setPanelWidth`, and both persist through `persistPanelWidth` (debounced), so a keyboard resize survives a reload the same way a drag does. Home/End land on 200/600, and Shift gives a 40px step. Unhandled keys return before `preventDefault`, so Tab and everything else still work. The keyboard path adds no document listeners, so the existing `ngOnDestroy` cleanup still covers everything.
- **Focus styling uses tokens only** (`--mj-brand-primary`, `color-mix`). The token gate passes.
- **`OnEntityMapRowKeydown` checks `event.target !== event.currentTarget`.** That's the right call, and the doc comment explains why: `stopPropagation` on the child spans only covers `click`. (It's also why swapping in `[mjClickable]` here would be a step backward: its keydown handler has no target check, so it would bring back the Space-on-checkbox bug.)
- The changeset is `patch` on `@memberjunction/ng-dashboards`, which is correct. The deprecated `onResizeStart` alias from #4487 is kept.

## Worth addressing

1. **The entity-map row is a `role="button"` wrapping a checkbox and a button.** See `connections.component.html:626-631`. This is new in this PR, and the TODO at :622-625 already names it. Before this is marked ready, I'd do the fix the TODO describes rather than ship the known violation:
   - An axe scan flags every row as `nested-interactive` (serious, WCAG 4.1.2).
   - ARIA treats the children of `button` as presentational. Screen readers may therefore not expose the sync toggle and the direction button as separate controls.
   - The `aria-label` replaces the row's content as its name. A screen-reader user hears only "Open field mapping for Contacts, button": no external object, no sync state, no direction.
   - Two maps from different external objects into the same MJ entity get identical names.

   Fix: put a real open button in the `.map-edit-hint` chevron slot (:662-664), e.g.
   `<button type="button" mjButton variant="icon" size="sm" [ariaLabel]="'Open field mapping for ' + (em.ExternalObjectLabel ?? em.ExternalObjectName) + ' to ' + em.Entity"><i class="fa-solid fa-chevron-right"></i></button>`.
   It needs no click handler of its own: the click bubbles to the row's `(click)`. Then remove `role`/`tabindex`/`aria-label`/`(keydown)` from the row, and remove `OnEntityMapRowKeydown`. Keep the row's `(click)` so mouse users can still click anywhere on the row. `MJButtonDirective` is already imported in `integration.module.ts`, and the button brings its own `:focus-visible` ring (`box-shadow: var(--mj-focus-ring)`). Keep the button inside the `.map-edit-hint` span, or drop that class from it: the `.map-edit-hint` color/font-size rules (`connections.component.css:2428-2436`) would otherwise override the icon variant's styling, and `check:ui-buttons` won't catch it because it only flags `.mj-btn` selectors. As it stands, the row has no focus style of its own: it gets only the UA outline, and the hover background and chevron highlight (`connections.component.css:2191`, `:2434`) never appear on keyboard focus.

   On the stated remaining work: running T120 against a DB with an integration connection is still needed, but it won't catch this. The row will "work" with Enter either way. Run an axe or accessibility-tree check on the rows as part of that verification.

## Minor

- **Focus is lost when the field editor opens or closes.** The list and the editor are the two branches of one `@if`/`@else` (`connections.component.html:202`), so opening the editor destroys the focused row. `CloseEntityMapEditor` (`connections.component.ts:992`) re-renders the list with focus on `<body>`, so a keyboard user has to Tab back from the top of the page. Now that there is a keyboard path, it's worth moving focus into the editor on open, and returning it to that map's open button after close (remember `em.ID`, and focus the button after `detectChanges`).
- **The separator has no `aria-controls`.** The APG window-splitter pattern points it at the pane it sizes. Add an `id` to `.query-tree-panel` (`query-browser-resource.component.html:4`) and `aria-controls` on the handle.
- **The "visible grip" claim in the description.** The comment also says that on `next` the grip "only shows on hover". It doesn't: on `next` the grip already has `background: var(--mj-border-strong)` at rest (`query-browser-resource.component.css:416`, unchanged here). This PR only adds the `:focus-visible` states. Please correct the description so nobody goes looking for a resting-state change.
- **No unit tests for the two new handlers.** Both are pure and cheap to cover: clamp at 200/600, Shift step, Home/End, no `preventDefault` on other keys, and the row target check. Covering them would pin down the behavior that is currently verified only by hand. The sibling `visual-editor.component.dom.test.ts` shows the pattern for the DOM side.
- **Nit:** the splitter's focus ring is `2px solid var(--mj-brand-primary)`. That's the same color today, but the ring tokens are `var(--mj-ring-width) solid var(--mj-focus-ring-color)` (`_tokens.scss:444-450`). Using them keeps it in step if the ring is re-themed.
- **Nit:** the template comment at `query-browser-resource.component.html:97-102` partly narrates history ("Before this it was a bare `<div>`…"). That belongs in the commit message. The first sentence, about why the tabindex matters, is worth keeping.

## Overall

**Comment.** The splitter half is solid and could ship as-is. I'd mark this ready for review once:
- the entity-map row gets the dedicated open button the TODO describes (#1), and
- that half has been exercised against a database that actually has an integration connection.

PR #3767, opened 2026-08-12
fix(entity-viewer,dashboards): reveal hover-gated actions on focus; stop card collapse

CaelebB-BC:
Split out of #3380 — independently revertable.

## Defects

**Hover-gated row actions (2 files).** `view-selector` and `home-dashboard` revealed their row/pin actions on `:hover` alone. At `opacity: 0` those controls are pointer-only — invisible to a tabbing keyboard user and unreachable without a mouse. For the pin card this is a genuine dead end: the overflow menu is the *only* route to rename/remove/configure.

**`ps-catalog` card collapse.** Cards had no `flex-shrink: 0`. The default shrink compressed the guide card below its own content at 1280×720 — **46px against 166px of content** — and the overflow landed *behind* the `position: relative` gallery cards. That made the filter chips genuinely unclickable for real users, not just for a test agent.

## Verified still present

Confirmed against `origin/next` @ `b23bf5933d`:
- `view-selector.component.css` — **0** `focus-within`
- `ps-catalog.component.css` — **0** `flex-shrink`
- `home-dashboard.component.css` — its one `focus-within` is on an unrelated `.search-input-wrapper`; the `.pin-card` actions are still hover-only

## Scope note

Roughly **20 other stylesheets** still gate row actions on `:hover` alone. Deliberately not swept here — this PR stays revertable, and the sweep deserves its own change.

### DIFF
--- a/.changeset/fix-a11y-focus-and-layout.md
+++ b/.changeset/fix-a11y-focus-and-layout.md
@@ -1,0 +1,18 @@
+---
+'@memberjunction/ng-entity-viewer': patch
+'@memberjunction/ng-dashboards': patch
+---
+
+Reveal hover-gated row actions on keyboard focus, and stop a card collapsing under
+its own content.
+
+- `view-selector` and `home-dashboard` gated row/pin actions on `:hover` alone. At
+  `opacity: 0` they were pointer-only — invisible to a tabbing keyboard user and
+  unreachable without a mouse. Added `:focus-within`.
+- `ps-catalog` cards had no `flex-shrink: 0`. The default shrink compressed the guide
+  card below its own content at 1280x720 (46px vs 166px) and the overflow landed
+  *behind* the `position: relative` gallery cards, making filter chips genuinely
+  unclickable.
+
+Note: roughly 20 other stylesheets still gate row actions on `:hover` alone. Not
+addressed here — worth a dedicated sweep.
--- a/packages/Angular/Explorer/dashboards/src/Home/home-dashboard.component.css
+++ b/packages/Angular/Explorer/dashboards/src/Home/home-dashboard.component.css
@@ -922,7 +922,11 @@
   font-size: 14px;
 }
 
-.pin-card:hover .pin-name .more-btn {
+/* Also reveal on keyboard focus — the pin's overflow menu is the only route to
+   rename/remove/configure, and at opacity 0 it was pointer-only: invisible to a
+   tabbing keyboard user and unreachable without a hover. */
+.pin-card:hover .pin-name .more-btn,
+.pin-card:focus-within .pin-name .more-btn {
   opacity: 1;
 }
 
--- a/packages/Angular/Explorer/dashboards/src/PredictiveStudio/components/ps-catalog.component.css
+++ b/packages/Angular/Explorer/dashboards/src/PredictiveStudio/components/ps-catalog.component.css
@@ -1,4 +1,9 @@
 .ps-catalog { display: flex; flex-direction: column; gap: 16px; }
+/* Cards size to their content. Without this the default flex-shrink:1 compresses the
+   guide card below its own body when the panel is height-constrained (46px vs 166px at
+   1280x720), and the overflowing content — including the filter chips — lands behind the
+   position:relative gallery cards, making the chips unclickable. */
+.ps-catalog > .ps-card { flex-shrink: 0; }
 
 .ps-catalog .guide-card .guide-head { display: flex; gap: 12px; align-items: flex-start; }
 .ps-catalog .guide-head i { font-size: 18px; margin-top: 2px; }
--- a/packages/Angular/Generic/entity-viewer/src/lib/view-selector/view-selector.component.css
+++ b/packages/Angular/Generic/entity-viewer/src/lib/view-selector/view-selector.component.css
@@ -367,7 +367,12 @@
   margin-top: 2px;
 }
 
-.vs-card:hover .vs-card-actions {
+/* :focus-within matters as much as :hover — these buttons (Duplicate, Configure,
+   Open in Tab) were reachable ONLY by pointer. A keyboard user could tab into them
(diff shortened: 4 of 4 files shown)


CaelebB-BC:
Merged `origin/next` (133cdfeb48) into this branch. It merged with no conflicts.

## Verified against regression test T069 (Predictive Studio workbench tour)

T069 fails on `next` at the checkpoint `algorithm-catalog-guide`. At 1280×720 the Algorithm Catalog's "Guide me" card shrinks to a thin strip above the gallery, so its scenario chips can't be clicked. Playwright's click on "Interpretability required" times out.

- The Guide me card shows at full height: heading, all seven scenario chips, and **Clear**.
- Clicking "Interpretability required" shows the banner *"For Interpretability required, your top pick(s): Logistic Regression & Ridge Regression"*. The cards re-rank, and the top picks get Primary badges.
- **Clear** resets the choice, with no active chip left.

At 1440×900 the same flow already worked on `next`. That is why the bug only appears at the suite's viewport.

rkihm-BC (approved):
>

Reviewed at f9c37251f0: three small CSS changes plus a changeset. All three are correct and low-risk. The pin-card part leaves the keyboard path it describes only partly fixed.

## What's good

- `view-selector.component.css:374-375`: the actions are real `<button>`s (`view-selector.component.html:111-128`, `178-190`). `opacity` never took them out of the tab order, so `:focus-within` is the right fix. They now show up while a keyboard user tabs through Duplicate, Configure, and Open in Tab. Both cards (My Views and Shared Views) use the same classes, so both are covered.
- `ps-catalog.component.css:6`: `.ps-catalog > .ps-card { flex-shrink: 0; }` is the right fix. `.ps-card` has `overflow: hidden` (`predictive-studio.shared.css:48`), so its automatic minimum height drops to 0. That let the flex column squeeze it. The gallery (`.gallery`, not a `.ps-card`) is not affected, and `.ps-content` scrolls (`ps-studio-resource.component.ts:180`), so the taller card just scrolls. The chips were already `<button>`s with a working `toggleUseCase` handler, so layout was the only problem.
- No new colors. The changeset bumps `ng-entity-viewer` and `ng-dashboards` as `patch`, which is correct.

## Worth addressing

1. **Pin cards: the first thing you tab to is still invisible.** See `home-dashboard.component.css:862-880`, and `home-dashboard.component.html:103,113,122` (ungrouped) and `210,220,229` (grouped). The `.pin-card` div has no tabindex or key handler, so on the keyboard a pin opens only through the overlay's Open/Run `<button>`. That overlay still shows only on `.pin-card:hover`. Its `pointer-events: none` doesn't stop focus, so the button is in the tab order but painted at `opacity: 0`. The Open button comes before `.more-btn` in the DOM, so Tab lands on it first. With this PR, `:focus-within` then shows the "..." button while focus sits on the invisible Open button. A keyboard user sees "..." appear, presses Enter, and the pin opens instead of the menu. The fix is one more selector on the rule at `:877`, and it covers grouped and ungrouped pins alike:
   ```css
   .pin-card:hover .pin-overlay,
   .pin-overlay:focus-within { opacity: 1; pointer-events: auto; }
   ```
   `.pin-card:focus-within .pin-overlay` also works, and it matches hover more closely: the overlay stays up while focus moves on to "...", the same way hover shows both at once. Either selector reveals the overlay fully.

2. **The pin menu still can't be used from the keyboard, which is the dead end the PR sets out to fix.** The description says the overflow menu "is the only way to rename, remove or configure". The button can now be reached, but the menu it opens is still mouse-only:
   - The menu items are `<div (click)>` with no tabindex or role (`home-dashboard.component.html:281-305`). Tab can't reach them, and focus stays on the "..." button, so nothing in the menu can be chosen. Escape does close it.
   - `ShowPinMenu` places the menu at `event.clientX/Y` (`home-dashboard.component.ts:1457-1458`). A click from Enter or Space on a button carries coordinates of 0, and the menu is `position: fixed` (`.css:1005-1006`), so it opens at the top-left of the viewport, far from the pin.

   Remove is already keyboard-reachable another way: the Edit button (`html:30`) turns on edit mode, where each pin shows a visible "Remove pin" `<button>` (`html:90,198`). Rename and Move to Group are still mouse-only. Rename in edit mode is a click on the `.pin-name` div (`html:137`). So the description overstates both the problem (Remove was never menu-only) and the fix (Rename and Move to Group are still mouse-only).

   This problem was there before the PR, and fixing it is bigger than a CSS tweak. Either make the items `<button>`s, place the menu from `(event.currentTarget as HTMLElement).getBoundingClientRect()`, and move focus to the first item on open, or trim the description's claim and open a follow-up issue. Either works. What matters is that the PR doesn't read as having removed this dead end.

## Minor

- `home-dashboard.component.html:246` (existing): the "..." button on **grouped** pins only calls `$event.stopPropagation()` and never opens the menu. Now keyboard users can see it too. It's worth listing in the same follow-up.
- `ps-catalog.component.css:2-5`: the comment blames the default `flex-shrink: 1` and says the chips "land behind" the gallery cards. The reason the card could shrink below its content is `.ps-card { overflow: hidden }`, and anything below 46px is clipped (not painted), not stacked behind. The column is height-limited in the first place because `.ps-panel { height: 100% }` (`predictive-studio.shared.css:25-28`). The fix is still right. The changeset repeats the "behind" wording, and release notes are built from changesets, so reword both. That would save the next reader from hunting for a z-index problem.
- Same accuracy point for the other two comments (`view-selector.component.css:370-373`, `home-dashboard.component.css:925-927`) and the changeset. They say the buttons were "unreachable" without a pointer. `opacity: 0` leaves a button focusable and exposed to assistive tech, as the first point under What's good notes, so they were invisible, not unreachable. The menu behind "..." is the part that really is unreachable (item 2).

## Overall

Approve. The three CSS changes are correct, stay in scope, and are safe to revert. Item 1 is a one-selector fix I'd fold in here. Item 2 can be a follow-up if the description is toned down to match.

PR #3165, opened 2026-07-15
docs(release): fold v5.48 release lessons into DEPLOYMENT.md + pg-migrate runbooks; fix docs.yml TypeDoc errors

SDesai-BC:
## Release-process improvements from the v5.48 release

Everything the v5.48 release surfaced, folded back into the docs — plus a real CI fix. Two commits, cleanly separated:

### `docs(release)` — DEPLOYMENT.md + pg-migrate runbooks
Captures the field lessons that cost real time this cycle:

- **DEPLOYMENT.md**
  - Recommend a `release/vX.Y-prep` branch + PR into `next` (the pattern used for v5.48).
  - **Step 3**: fresh-DB creation snippet; `MJ_MIGRATION_REQUEST_TIMEOUT=1800000` (the 300s default kills a baseline batch — documented the `Failed to cancel request in 5000ms` → `LoggedIn state` cascade); **don't probe the DB mid-migrate**; **check for CPU-hungry Docker containers starving SQL Server** (the actual root cause of two failed migrates — a clean run is ~1 min); commit the mj-sync `primaryKey`/`sync` writeback with the migration; **codegen is NOT part of Step 3** (optional drift-check only); the changeset step is a *check*, not an automatic add.
  - **Step 4**: point at `.github/scripts/validate-npm-packages.sh` as the authoritative new-package check.
  - **Step 7**: reference `/pg-migrate-experimental` alongside `/pg-migrate-v2` + shared invariants.
- **`pg-migrate-experimental`** — memory pre-flight / `--concurrency=2` on <8GiB VMs; `python3-venv` apt prereq + broken-venv cleanup; **correct workspace names `mj_api`/`mj_explorer`** (`@memberjunction/mjapi` and `ng-explorer` don't exist); generate gitignored MJExplorer `environment*.ts`; `--file` takes a basename; **converter's EntityField-Sequence dedup edits committed `.pg.sql` — detect + revert**; AST blind spots (CTE-targeted DELETE, `sys.indexes` IF-EXISTS guards); `cdp_*` role bootstrap SQL for the gate; **disable ScheduledJobs in the gate DB** (RunImmediatelyIfNeverRun vector-sync OOM-kills MJAPI); `provisioningGuard='warn'` for Owner magic-link (with explicit human sign-off); Phase-4 proportionality note for index/data-only migrations.
- **`pg-migrate-v2`** — same two nonexistent package names fixed.

### `fix(docs-ci)` — the 5 TypeDoc errors failing `docs.yml` on every release
[...]

PR #3769, opened 2026-08-12
fix: stop String.replace expanding $-sequences in dynamic replacements

### LINKED ISSUE #3171: String.replace with dynamic replacement string mangles $-sequences (~13 sites; HIGH: MJInstaller writes corrupted secrets)
## Summary
A repo-wide sweep (prompted by the `@memberjunction/metadata-sync` `{@include}` fix in #3165) found the **same latent bug class in ~13 other places**: `String.prototype.replace(search, replacement)` where the **replacement** is a *dynamic string* (file content, secrets, user input, DB values, SQL/prompt fragments). JS interprets `$$`, `$&`, `` $` ``, `$'` in a **string** replacement as special patterns (and `$1`–`$99` when `search` is a regex with capture groups), so any dynamic replacement containing a literal `$` is silently corrupted. Fix = pass a replacement **function** `() => replacement`, or escape `$`→`$$`.

The `MetadataSync/sync-engine.ts` instance is already fixed in #3165. The reviewer of that PR asked for this sweep — the pattern **does** exist elsewhere.

## HIGH — fix first
`packages/MJInstaller/src/phases/ConfigurePhase.ts` — `replaceOrAppendEnvVar` (line ~841) and the config-patch replacers (~1141, ~1155) write `.env` values (**DB passwords, JWT/AUTH secrets, API keys**) via `content.replace(pattern, \`${name}='${value}'\`)` with **no `$`-escaping**. A secret containing `$$`/`$&`/`` $` ``/`$'` is corrupted → wrong secret written → broken install/auth.
- **The correct pattern already exists in the sibling** `packages/MJCLI/src/lib/legacy-install.ts:401` (`value.replaceAll('$', () => '$$')`) — MJInstaller `ConfigurePhase` regressed it. Copy that escaping (or switch to a function replacement).

## MED — data/SQL/prompt corruption
- `packages/MJCore/src/generic/securityInfo.ts:471` — user property value into an **RLS `MarkupFilterText`** filter (`$&` can eat the token → mis-scoped rows).
- `packages/MJCore/src/generic/entityInfo.ts:2547-2604` — escaped data value into a **SQL WHERE** clause (`'Custom'` normalization); `$` not escaped.
- `packages/GenericDatabaseProvider/src/GenericDatabaseProvider.ts:1911` and `packages/PostgreSQLDataProvider/src/PostgreSQLDataProvider.ts:606` — end-user **search term** into a SQL predicate format (`UserSearchParamFormatAPI`).
- `packages/AI/Prompts/src/AIPromptRunner.ts:2914,4025` — `AssistantPrefill` into a system-prompt template (LaTeX `$$` etc.).
- `packages/AI/ComputerUse/src/engine/ComputerUseEngine.ts:1188-1192,1253-1257` — user goal / step summaries into controller & judge prompts.
- `packages/AI/Vectors/Sync/src/generic/EntityDocumenTemplateParserBase.ts:89` — field value into embedding-document text (closest structural twin of the fixed bug).

## LOW — logs / test-framework / narrow tooling
- `packages/MetadataSync/src/lib/sql-logger.ts:64` (debug SQL log only), `packages/AI/Engine/src/AIEngine.ts:1613` (@deprecated path), Dynamics connector `:617`, TestingFramework oracles, SQLConverter rules, React-linter message builders. (Full table available from the audit.)

## Correctly-handled references (for the fix author)
- `MJCLI/src/lib/legacy-install.ts:401-402` — escapes `$` first (the model to copy).
- `SQLConverter/src/rules/FunctionRule.ts:170` — hardcoded `$$$$` literal with a comment explaining the collapse.

## Recommended
1. Fix the **HIGH** MJInstaller `ConfigurePhase` sites first (own PR + `patch` changeset + a regression test that sets a value containing `$$`/`$&` and asserts the written `.env` is byte-exact).
2. Work down the MED list (SQL/RLS/prompt).
3. **No lint currently catches this class** (the existing `string-replace-all-occurrences` React rule only flags first-occurrence/replaceAll). Consider a small grep/lint gate: flag `.replace(` / `.replaceAll(` whose 2nd arg is a non-literal, non-function expression, so it stops recurring.

## Context
Surfaced during the #3165 code-review hardening loop. `sync-engine.ts` fixed there; this issue tracks the rest.

SDesai-BC:
A DB password containing `$&` was written to MJAPI's `.env` with the **stale password spliced into the middle of it**. Not a typo'd secret — a structurally corrupted file, written silently, surfacing later as "MJAPI can't connect."

`String.prototype.replace(search, replacement)` treats `$$`, `$&`, `` $` ``, `$'` and `$1`-`$99` as metacharacters when `replacement` is a **string**. That is fine for a literal you wrote; it is a defect when the replacement is runtime data, because JS cannot tell data from instructions. This sweep fixes the sites #3171 identified plus those the new gate surfaced, prioritised by blast radius — secrets, SQL generation, security predicates. It is **not** exhaustive: ~90 sites remain repo-wide and a few of those do carry data (`MentionParser.ts` substitutes a mention name into conversation text; `FieldExternalizer.ts` a value into a filename pattern). They are lower-severity and the line-aware gate catches each one the next time anybody touches it, which is the intended stopping point rather than an oversight. Closes #3171.

## The offending line

`packages/MJInstaller/src/phases/ConfigurePhase.ts`, on the path that syncs the root `.env` into MJAPI's when you re-run `mj install`:

```ts
return content.replace(pattern, newLine);   // newLine embeds the user's DB password
```

Driving the real `ConfigurePhase.Run()` over a mock filesystem, before any fix:

| password in root `.env` | what landed in MJAPI `.env` |
|---|---|
| `pa$$w0rd` | `pa$w0rd` — silently wrong secret |
| `pa$&w0rd` | `paDB_PASSWORD='stale-old-password'w0rd` |
| ``pa$`w0rd`` | `paDB_HOST='localhost'` |
| `pa$'w0rd` | `pa` |
| `pa$w0rd` | correct — `$` before an ordinary character is not special |

Two details worth having, because they explain the survival:

- **Only the replace branch is affected.** The append branch of the same function uses string concatenation and is byte-exact. So this fires on re-install over an existing MJAPI `.env`, never a fresh install.
- **The path was already tested.** `ConfigurePhase.test.ts` had a sync test using `DB_PASSWORD='Strong!Pass'` — a special character was deliberately chosen, just not one `replace` treats as a metacharacter.

## What changed

Every site now passes a replacement **function**, whose return value is used literally with no `$` expansion. Where a back-reference was intentional, capture groups are carried through as named callback parameters rather than `$1`.

- **`installer`** — `.env` secret sync (`DB_PASSWORD`, `CODEGEN_DB_PASSWORD`, `PG_PASSWORD`), `newUserSetup` (embeds user name and email), `mjRepoVersion`, Explorer `environment.ts`.
- **`core`** — `RowLevelSecurityFilterInfo.MarkupFilterText`, which substitutes user properties, magic-link scope and `{{Acting*}}` tokens into row-level-security filters. A `$` here **rewrote the predicate** — the exact outcome the neighbouring `'`-escaping was written to prevent. This feeds `GetEffectiveRowFilterWhereClause`, used across RunView reads, Create and Update. Also organic-key `Custom` normalization.
- **db providers** — end-user search terms via `UserSearchParamFormatAPI`, view-template inner SQL, PG identifier quoting, `QueryCompositionEngine.renameSQLIdentifier` (CTE identifiers in composed queries), and SQL Server batch parameter rewriting.
- **`sql-parser`** — `restoreAliases`, which swaps generated aliases back to the caller's original bracketed identifiers.
- **`sql-converter`** — the configured schema name in emitted PostgreSQL, in both `ViewRule` and `InsertRule`.
- **ai** — assistant prefill (routinely contains `$$` for LaTeX or currency), computer-use goals/URLs/step summaries, embedding-document field values, entity field markup.
- **`testing-framework`, `react-linter`, `metadata-sync`** — judge-prompt inputs, diagnostic messages, debug SQL logging.

## Why a function rather than escaping `$` to `$$`

Both work. The function form was chosen because escaping has to be applied correctly at every call site and is invisible when it is missing — `securityInfo.ts` is the proof: it escaped `'` and stopped there, so a `$` walked straight through the defence the comment above it describes. A function cannot be half-applied. It also needs no shared helper, which matters because `MJInstaller` does not depend on `MJCore`.

## New CI gate

`.github/scripts/check-dynamic-replace.mjs` flags `.replace()`/`.replaceAll()` whose second argument is neither a string literal nor a function. **No existing rule covered this**: the React `string-replace-all-occurrences` rule destructures `const [searchArg] = args` and never inspects the replacement.

Three decisions a reviewer would otherwise question:

- **Line-aware, not file-aware.** 90 pre-existing sites remain across the repo, a good share benign — a bare identifier holding a function reference is indistinguishable from one holding a string. Gating whole files would fail PRs for code they merely brushed past. It reports only lines a change touched, and tracks the full call span so a multi-line call whose replacement argument sits three lines below the `.replace(` is still caught.
- **Static string literals are allowed.** An author writing `'$1'` means the back-reference. The bug class is runtime data reaching the slot.
- **`safe-replace:` escape hatch** for genuinely-safe sites, which forces the reason to be written down. It is coarse on purpose — it suppresses every call in its span and matches inside string literals — and the script header now says so rather than implying it is per-site, because a marker that silently fails to apply is worse than one that applies too widely.

## The gate was reviewed as hard as the sweep, and did not survive first contact

Adversarial review found the gate could report **green over code it never parsed** — the one outcome its own header says it must never produce. Each of these was reproduced before it was fixed:

| Defect | How it failed open |
|---|---|
| `git()` had no `maxBuffer` | Node's 1 MiB default; `git diff --unified=0` on a real PR is megabytes (#3735 alone produced 897 KB). Raw `ENOBUFS` stack, exit 1, on an unrelated PR |
| A body line beginning `++ b/` | Under `--unified=0` an added line gains a `+`, so file **content** rendered as `+++ b/x` and matched the file-header regex. Following hunks were attributed to a phantom path and dropped. Reproduced: identical violating change plus one benign line → `✓ 1 file(s) clean`, exit 0 |
| Look-back over raw source | A trailing `//` comment made the next line's `/` read as division; a backtick inside the unmasked regex opened a template frame that never closed, blanking the rest of the **file**. Two real files did this |
| `invokedDirectly` compared raw path strings | Node resolves symlinks when loading an ES module, so any invocation through a symlinked path printed nothing and exited 0 — a silent pass |
| A regex starting with `>` | The JSX `/>` rule was too broad, so `html.replace(/>/g, userValue)` blanked its line. 53 files contain `replace(/>` |
| `--file` with a missing or bad path | Printed `✓ 0 file(s) clean`, exit 0 — a pass over nothing, while `--base` already exited 2 for the same mistake |
| Quadratic scan | `lineAt` re-sliced the prefix per violation: 20k violations took 8.3 s, 100k took 212 s. Now 28 ms and 141 ms |

The workflow step also moved into the existing "no build needed" block. It reads source text and costs about a second, but sat *after* the ~20-minute build+test step — so any unit-test failure stopped it running at all, silencing the gate exactly when a PR was already in trouble.

## Evidence

| What | Result |
|---|---|
| Gate's own tests | **249 passed** |
| Gate vs. pre-fix code | **68** violations across the files this PR converts |
| Gate vs. this branch | **29 files clean**, exit 0 |
| `--all` audit | 90 pre-existing sites, unchanged by this branch |
| Deterministic integration tier | **63/65**, `RUN_MUTATION_TESTS=1`, from-scratch database |
| Runtime `$`-probes | **49/49** |

The tier ran on a database built from scratch — 47 migrations, 376 entities, 13,728 metadata records, then seeded from `metadata-optional/integration-test`. Both failures were traced and neither belongs to this change:

- `IT29 cache-gauntlet` fails on row-count preconditions (`need >= 2 existing rows … found 0`) that a clean database cannot satisfy.
- `IT03/C12` was verified **pre-existing on `next`** by building the merge-base in a second worktree and re-running the check with a merge-base client *and* server against the same database. It fails identically there.

**The coverage gap the earlier revision of this description declared open is now closed.** `rls-isolation.RLS1`/`RLS2` still self-skip, so the integration tier does not exercise `{{UserID}}` token substitution. Targeted runtime probes cover it instead: all seven metacharacter forms are substituted byte-exact through the real `MarkupFilterText`, the resulting predicate is **executed against SQL Server**, accepted, and returns correctly-scoped rows. The same probes cover `UserSearchString`, organic-key clauses and `ExtraFilter`. Unit tests prove the string is right; these prove the SQL is.

One incidental confirmation worth recording: a probe run whose module resolution fell through to an unfixed checkout returned `Name = 'a$b'` for input `a$$b` — the bug reproducing live outside a test harness. The same probe against this branch returns `a$$b`.

## Review round (rkihm-BC)

Three fixes requested, all reproduced before being fixed: a duplicate `"test"` key in the Social manifest (a merge collision with #3542, which appended the same key this branch had already added), `--all` descending into dot-directories so `.angular/cache` vendored bundles drowned the audit (proved locally: 90 → 91 with one file under `.angular/cache`, and that file is git-ignored, which is why CI never saw it), and `escapeRegExp` inserted between `transformCodeOnly`'s JSDoc and the function. Also moved the workflow comment block below the enumerated guard list it had split, and corrected a stale count in it. The dotenv inverse hazard is now named in the changeset rather than only here.

## What reviewers should watch

- **The new gate starts failing PRs** that introduce a dynamic replacement on a touched line. That is the point, but it is a new way for an unrelated PR to go red.
- **`testing-cli` env precedence changes.** `.env` was loaded with `override: true`, so an explicit `DB_DATABASE=… mj test …` was silently discarded and the suite ran — **including mutation tests** — against whatever `.env` pointed at. It is now dotenv's default `false`. Verified at runtime: an inline `DB_DATABASE` is honoured and a wrong one now fails loudly instead of quietly testing elsewhere. `DEPLOYMENT.md` §4.1 told the release engineer the opposite and has been corrected.
- **Two regex-escaping fixes ride along**, because the replacement-side fix was unreachable without them: `PostgreSQLDataProvider.quoteFieldNamesInToken` and the SQL Server batch parameter rewrite both interpolated a caller-supplied name into a `RegExp` unescaped, so a `$` acted as an end-anchor and the pattern matched nothing.
- No migration, no metadata, nothing to sync — every changeset entry across 18 packages is `patch`.

## Known, and deliberately not fixed here

- `sql-logger` and `SQLValidatorOracle` build `@param` patterns without word-boundary anchoring, so `@param1` also matches inside `@param10` and `@id` inside `@idType`. Real, pre-existing, and a different defect class from `$`-expansion.
- `GenericDatabaseProvider` pushes to a cycle-detection stack that is never popped, and `queryCompositionEngine` tokenises comment-stripped SQL while replacing into the raw string. Both pre-existing, both on lines this PR touches for other reasons.
- A from-scratch build does not reproduce `next`'s committed generated code: several `IsNameField`-driven display names differ. This PR contains no migration, metadata or CodeGen change, so it belongs with the materialization work in #3735.

The follow-up comment on #3171 describes a **different** defect in `MetadataSync/src/lib/sync-engine.ts` (the include loop matches against the original content but replaces the first occurrence in the mutated string, so a nested include can replace the wrong one). Its `$`-expansion half was already fixed in #3165; the wrong-occurrence half is untouched here and should be filed separately rather than lost when this closes.


rkihm-BC (commented):
## Code review

Read the full diff, the gate script end-to-end, and ran the gate's own `findViolations` over the repo to check its claims independently.

### What's good

- **The root diagnosis is correct and the fix form is the right one.** Replacement functions are unconditional; `$$`-escaping has to be reapplied at every call site and fails silently when it isn't — `securityInfo.ts` is the proof (it escaped `'` and a `$` walked straight past). Choosing functions over a shared escape helper also avoids an `MJInstaller → MJCore` dependency that doesn't exist.
- **The genuinely dangerous sites are all covered**: `.env` secret sync, `MarkupFilterText` (RLS predicates feeding `GetEffectiveRowFilterWhereClause` on read/create/update), `UserSearchParamFormatAPI` (end-user search terms), organic-key normalization, `renameSQLIdentifier`, `restoreAliases`. I re-audited the ~90 remaining sites; nothing left is in a security or SQL-generation path of comparable severity.
- **The two regex-*search*-side escaping fixes are real bugs, not scope creep.** In `quoteFieldNamesInToken` and the SQL Server batch rewrite the replacement-side fix was literally unreachable — a `$` acted as an end-anchor and the pattern matched nothing. The `a.b` → wrongly quotes `axb` test is a separate live bug caught in passing.
- **Tests drive real code paths.** `ConfigurePhase.test.ts` runs the actual `phase.Run()` over a mock FS and asserts byte-exactness *plus* absence of the stale password — not a re-implementation. Same for `restoreAliases` via the private-method seam and the PG quoter.
- **The gate is unusually well-built for a CI script.** Mode-stack masking (nested `${}`), look-back over the *masked* buffer, `desynced` → fail rather than report clean, `--file`/`--base` misuse exiting 2, `realpathSync` on the entry-point check, binary-searched `lineAt`. Line-aware scoping with call-span tracking is the right call given the pre-existing sites, and moving it into the pre-build guard block is correct — a gate that only runs when tests pass is a gate that goes quiet exactly when it's needed.
- Changesets cover all 18 touched packages, all `patch`, correct per the repo's bump rule.

### Please fix before merge

**1. Duplicate `"test"` key in `packages/Actions/BizApps/Social/package.json`.** `next` already had `"test": "vitest run"`; this PR added a second identical key alongside `test:watch`:

```json
"build": "tsc && tsc-alias -f",
"test": "vitest run",      // added here
"test:watch": "vitest",
"watch": "tsc -w",
"clean": "rimraf dist",
"test": "vitest run"       // pre-existing
```

Behaviourally inert (last wins, same value), but it's invalid-in-spirit JSON in a file people hand-edit. Drop the added `"test"` line, keep `test:watch`.

**2. `--all` audit is ~66% noise on any dev machine.** `SKIP_DIR_SEGMENTS` has no dot-directory rule, so `packages/MJExplorer/.angular/cache/**` gets scanned. Running the gate's own `findViolations` over my checkout: **443 hits, 294 of them from `marked.js` / `xlsx.js` / `mermaid.js` in the Vite cache** — the same vendored third-party code `bundled-libs`/`vendor` are in the skip list to exclude. CI is unaffected (`ls-files --exclude-standard` skips ignored files), but `pnpm check:dynamic-replace:all` — the documented audit entry point — is unusable as shipped. Either skip path segments beginning with `.`, or drive `--all` off `git ls-files`.

**3. `ExpressionHelpers.ts`: the new `escapeRegExp` was inserted between `transformCodeOnly`'s JSDoc and `transformCodeOnly` itself.**

```ts
/**
 * Apply a transformation function only to SQL code segments,
 * preserving string literals and comments unchanged.
 */
/**
 * Escape every regex metacharacter in `literal` ...
 */
export function escapeRegExp(literal: string): string { ... }

export function transformCodeOnly(...)   // ← now undocumented
```

Move the new function above the orphaned comment, or below the function it belongs to.

### Flagging, no change requested

**4. The `testing-cli` `override: true` → `false` flip is a real behaviour change riding along in a `$`-expansion PR.** I think it's the right change — it aligns `mj test` with `migrate`/`codegen`/`sync push`, and the old behaviour made "one database per agent" unenforceable by environment variable. But note the *inverse* hazard it introduces: any environment that exports `DB_*` globally (docker images, CI containers, a stale `export` in a shell profile) now silently wins over `.env`, where `.env` used to be authoritative. Worth a release-note line, not just a PR-body line.

**5. The description overstates the sweep.** "This sweep fixes every site where it was data" — ~90 remain, and some of those *are* data: `ConversationsRuntime/src/mentions/MentionParser.ts:265` substitutes `@${mention.name}` into conversation text, `MetadataSync/src/lib/FieldExternalizer.ts:136`. Stopping there is a defensible line — the line-aware gate catches them on next touch — but better to say that than to claim completeness.

**6. "All 15 checks green" is from an earlier run** — `Run unit tests` is currently pending on the head commit. Everything else is green, including the deterministic integration tier and all three MJInstaller platforms.

### Optional / minor

- The one acknowledged coverage gap: `rls-isolation.RLS1`/`RLS2` self-skip, so no *committed* test executes a `$`-bearing RLS predicate against a live database. The runtime probes that closed it were one-off and aren't in the repo. Unit tests do prove `MarkupFilterText` produces the right string, which is the part that regressed — a committed integration check would be belt-and-braces.
- In `test.yml` the new step's comment block landed between the "1) … 4)" enumerated comment and the four steps it enumerates, so that list now reads as describing the dynamic-replace step.
- `escapeRegExp` now exists in three more copies. The repo already had ~5 pre-existing copies, so I would **not** ask for a shared helper here — noting it only so it isn't mistaken for something this sweep introduced.
- `flow-agent-type.ts:1780` still interpolates `indexVar` into `new RegExp` unescaped — the search-side twin of the two escaping fixes elsewhere. `indexVar` is an authored loop-variable name, so this is theoretical.

### Verdict

Approve after 1–3, all small and mechanical. The core change is correct, the risk ordering is right, the tests exercise real paths, and the gate holds itself to the standard it enforces.

SDesai-BC:
Thanks — all three fixed in b732acc677, each reproduced first.

**1. Duplicate `"test"` key** — fixed. One correction to the account: `next` didn't already have it. The key arrived on this branch with the merge of #3542 (`f5ec13b979`), which appended `"test": "vitest run"` at the end of `scripts` while the branch had already added the same key plus `test:watch` after `build`. A collision neither side could see alone. Trailing one dropped.

**2. `--all` dot-directory noise** — fixed, and confirmed here rather than taken on report: one violating file under `.angular/cache` moved the count 90 → 91, and it's git-ignored, which is exactly why CI never saw it. The skip went in `walk()` rather than `isSkippedPath`, since the latter also filters diff mode and `.github/**` needs to stay scannable there.

**3. Orphaned JSDoc** — fixed; `escapeRegExp` moved below `transformCodeOnly`.

Also took the two presentational notes: the workflow comment block had split the enumerated "1) … 4)" guard list from the steps it describes (moved below them, still pre-build), and its "~95 pre-existing sites" was stale — it's 90.

**4. dotenv inverse hazard** — agreed, and it's now in the changeset rather than only the PR body, so it lands in the release notes.

**5. Description overstatement** — fair, corrected. It now says the sweep is prioritised by blast radius and explicitly *not* exhaustive, naming `MentionParser.ts` and `FieldExternalizer.ts` as remaining data-carrying sites the line-aware gate will catch on next touch.

**6. Stale CI claim** — removed the live-check tally from the evidence table; it goes stale on every push.

On the RLS coverage gap: I ran one-off runtime probes on a from-scratch database that execute a `$`-bearing `MarkupFilterText` predicate against SQL Server and assert correct scoping — all seven metacharacter forms pass. You're right that they aren't committed, so the repo still has no standing check for it. Happy to land one as a committed integration bundle if you'd like it in this PR rather than a follow-up.

`flow-agent-type.ts` `indexVar` is already pinned by a characterisation test recording that it's unescaped, so a future fix shows up as a deliberate change.

Verified after the fixes: gate suite 268, `--all` 90, self-check 28 files clean, SQLConverter 1145, Social 248, `check:changeset` + `check:standards` pass.

rkihm-BC (commented):
## Re-review at `df1762e1`

All three items from my previous review are resolved, and one of them is fixed better than I asked for.

### Required items — done

1. **Duplicate `"test"` key** in `Actions/BizApps/Social/package.json` — removed. ✓
2. **`--all` cache noise** — dot-directory skip added. Placed in `walk()` rather than `isSkippedPath()`, deliberately, so `.github/**` stays scannable in diff mode. That's more correct than the fix I suggested — I'd have broken diff-mode coverage of the gate's own directory. ✓
3. **Orphaned JSDoc** — `transformCodeOnly` moved back under its comment. ✓

### The `BaseTestDriver` commit is a significant find in its own right

`BaseTestDriver.Provider` was returning `new Metadata() as unknown as IMetadataProvider`. I verified independently: `Metadata` has **no `RowLevelSecurityFilters` member at all**, so the getter read `undefined`, `discoverTokenFilter`'s `?? []` turned that into "no filters exist", and `rls-isolation.RLS1`/`RLS2` skipped-as-pass on **every** database while the bundle reported green — with filters actually present the whole time.

That is a fail-open in a security check, which is the same defect shape as the `$`-expansion this PR exists to fix. Worth calling out explicitly in the PR description rather than leaving it as "stop a security check skipping itself" in a commit subject.

The fix returns `Metadata.Provider`, which is properly typed `IMetadataProvider`, so the `as unknown as` cast disappears as a side effect — a strict improvement. `RLS11` now executes `$`-bearing predicates against a live database, closing the coverage gap I'd raised as optional, and both registry coverage guards were updated 15 → 16. `UserInfo.Name` is a plain field rather than an accessor, so the prototype-preserving principal clone in RLS11 is safe. **`Integration (SQL Server, deterministic)` is green on this head, so RLS11 passes in CI.**

### One new problem, introduced by fixing my cosmetic nit

I flagged that the new step's comment block had landed between the `1) … 4)` enumerated comment and the four steps it enumerates. That was fixed by moving the **step** below the four DOM guards, rather than by moving the **comment**. Verified consequence on this exact run (`Run unit tests`, job 94920958153):

```
success   Explorer DOM coverage gate
failure   Generic DOM coverage ratchet
skipped   Bootstrap DOM coverage ratchet
skipped   Dynamic-replace guard      ← never ran
```

This is precisely the failure mode the step's own comment argues against:

> Sitting after the ~20-minute build+test step meant any unit-test failure stopped it from running at all — the gate went silent exactly when a PR was already in trouble and most needed checking.

Four sequential `bash -e` guards now sit upstream of it, any of which takes it down with them. The gate consequently has not run on its own final head at all. Suggested fix: put the step back at the top of the pure-file guard block and move its comment with it, leaving the `1) … 4)` list directly above the four steps it describes.

### Blocking, but not this PR's defect

`Run unit tests` is red: the Generic DOM coverage ratchet reports 136 components with no spec against a cap of 134. This PR touches **zero** files under `Angular/Generic`, and `next` is failing the same gate independently (commit `60406dc17`, 22:16Z) — it arrived through the merge. Needs the ratchet raised or deferrals added on `next` before this can go green.

### Status

- `mergeable: MERGEABLE` (no conflicts), `mergeStateStatus: UNSTABLE`.
- Every other check on this head is green, including the deterministic integration tier, all three MJInstaller platforms, and all six driver suites.
- Owed here: restore the guard step's position — one small edit.
- Owed elsewhere: the `next` ratchet failure.

Once the step ordering is restored and `next` is green, this is good to merge from my side.

PR #3770, opened 2026-08-12
Workflow Run Console — realtime runner + debugger for task graphs

AN-BC:
## What this is

A **realtime runner/debugger at the task-graph level** — the "press Run and watch it live" experience the agent harness gives a Loop agent, but for the durable dispatcher that now executes every workflow. Since the Track C.1 cutover (#3692) a Flow agent *is* a task graph at runtime, so the place to run, watch and debug a workflow is the graph, not the agent run.

Design doc: **`plans/task-graph-realtime-runner.md`** (in this PR). Mockup: **`mockups/workflow-ux/run-console-v1.html`**, extending the committed Phase 5 design contract.

This completes two committed debts and adds the observability half of the hardening program:

- **The 5b UI debt.** `TaskGraphFrameResolver` has published seven semantic frame kinds over `taskGraphFrames(parentTaskId)` since Track C.1 — with **zero client consumers**. Every UI surface polled `MJ: Tasks` on a 3-second timer instead.
- **The R4 charter.** The restored Workflows app was repurposed for run review; this is what run review looks like when the engine can talk.
- **Hardening (#3749) synergy.** Every R2 headline defect shares the signature *"the graph looks fine while a run strands silently."* Each class now surfaces as a named diagnosis: a crashed prompt task's expired claim (R2-1) is a red worker badge, a permanent hold from data absence (R2-3) is an amber path with its reason and an operator override, a deferred settlement (R2-2) shows in the engine strip. The console doesn't replace those fixes — it makes the next ring discoverable in minutes rather than by four-track adversarial review.

## The design decision worth reviewing

**Every debugger control is a gate on *claiming*, never on *running*** — durable declarative state in the parent's metadata bag, which the claim filter reads each pass. Not hooks, not callbacks.

This falls out of the engine's own properties: the dispatcher is multi-instance and survives restarts, so a callback registered in one process gates nothing on the peer that wins the claim and dies with the process. A flag on the row gates every instance uniformly, works when "start paused" is set before any dispatcher has seen the graph, and survives a deploy. The debugger and the dispatcher never talk — they rendezvous on the row, which is the engine's native idiom (the claim protocol is itself nothing but CAS'd shared state).

It also **answers the question the program deliberately deferred** (#3576: *"pausing a claimed task means deciding what happens to its claim"*): pause gates claiming and never touches a live claim, so there is nothing to decide. In-flight steps finish naturally and their completions land. A breakpoint is an authored hold — the exact mechanism `holdTaskIDs` already implements for unevaluable conditions, driven by intent instead of a broken guard.

Timing is honest by construction: verbs take effect at the next poll boundary (≤5s), because you can never interrupt claimed work anyway — only gate what gets claimed next.

## What ships

**Engine (`@memberjunction/task-graph`)**
- New frame kinds emitting state the dispatcher **already computes**: `GateDecision` (why a branch ran / why one is held), `ClaimChanged`, `PassCompleted` (the engine heartbeat), `GraphPaused`/`GraphResumed`, `BreakpointHit`, `NodeProgress`. Change-only with per-edge dedup — the frame-topic analogue of `logUnevaluableConditionOnce`.
- `debug-state.ts` — the pure control model (`ParseTaskGraphDebugState`, `DecideClaimGate`, `OverrideVerdictFor`), parsed with the same tolerant posture as `ParseTaskGraphParentMetadata`: unreadable state reads as "not being debugged", never a guess that could gate real work.
- Guarded `JSON_MODIFY` writes for the debug bag (same two-writer discipline that protects the continuation marker), plus `TrySkipPending` and `TryForceComplete` — the latter **refuses a live claim** rather than racing an executor.
- Edge overrides apply to propagation as well as claiming, so an answered path cascades its skips through the engine's own rules.

**Control plane** — per the #3576 standing rule, every verb is a Remote Operation: `TaskGraph.Pause` / `.Resume` / `.Step` / `.SetBreakpoints` / `.OverrideEdge` / `.SkipTask` / `.ForceCompleteTask` / `.UpdateTaskInput`, owner-or-elevated gated; `RetryTask` gains an optional edited input. Metadata rows included. `ForceCompleteTask` refuses human steps — those keep their first-class `CompleteTask` path with the assignee check.

**Client + UI**
- `GraphQLDataProvider.TaskGraphFrames(parentTaskId)` — first consumer of the subscription, shared and refcounted per graph, mirroring `PushStatusUpdates`.
- Run view (widgets layer, so the **host owns the subscription**): `LiveFrame` input folds frames into the overlay; `ReplayAt` renders the run as it stood at a moment, from row timestamps. Doctrine kept explicit — **frames are advisory, rows are truth**: status frames patch in place, cascade frames trigger a debounced row reconcile, because inferring a cascade client-side would be a second implementation of the engine's rules.
- Workflows → Runs becomes the console: pause/step/step-wave toolbar, engine pass strip, stall banner, step inspector (claim, path verdicts, live runner progress, what-if computed with `ComputeTasksToBlock` — the dispatcher's own algorithm, so the preview cannot disagree with reality), and a replay scrubber.

**No migration** — the debug state lives in the existing metadata bag.

## Verification

Run in this sandbox:
- Full monorepo build green (290 packages).
- Unit tiers green for every touched package: TaskGraph **141**, MJServer **838**, ng-task-graph-editor **161**, ng-dashboards **1746** — including **13 new** `debug-state` tests covering the parse tolerance matrix and the full claim-gate decision table (pause beats breakpoints; a named step releases only itself and only when eligible; a breakpoint pauses the whole graph so siblings don't start either).
- `check:ui`, `check:standards`, `check:changeset` pass.

**Not run here — no database in this environment.** Please cover locally:
1. `mj sync push` then `mj codegen` for the eight new Remote Operation rows (CodeGen will emit typed bases; `TaskGraphDebugOperations.ts` currently declares its own contracts and can be re-parented onto them in a follow-up — the runtime contract is identical either way).
2. The deterministic integration tier.
3. End-to-end on a live dispatcher: set a breakpoint, run a flow agent from chat, watch it pause before the step, inspect, step through, resume to settle. Two-instance behavior for the step-allowance CAS is the piece most worth exercising.

## Open questions carried from the design doc

- **Step default** — one step or one wave? Both ship; the toolbar shows both.
- **D18 boundary** — the engine strip speaks engine vocabulary (passes, claims). Assumed builder-surface-with-collapse; a ruling would settle it.
- **Loop granularity** — a paused graph with an in-flight `ForEach` finishes the whole loop task, since claims gate at task granularity. Fine for v1; per-iteration gating would need the loop executor to consult the flag.
- **Multi-node push** — frames still ride in-process PubSub, so a client attached to another node relies on the reconcile poll. Getting frames onto the Redis channel (the cache-invalidation precedent) is tracked as a separable item.

https://claude.ai/code/session_015nTcYPCohgZTPbCRT2xKo8

---
_Generated by [Claude Code](https://claude.ai/code/session_015nTcYPCohgZTPbCRT2xKo8)_


AN-BC:
## 🔍 Review from the task-graph hardening loop — design is right, four changes needed before merge

Reviewed at `6cd337d` by the agent running the Round 1–3 engine hardening reviews (#3745, #3749, #3771). The core design decision — **every control is durable declarative state gating *claiming*, never a hook gating running** — is correct, and it genuinely answers #3576's deferred claim-pause question. The pure `debug-state.ts` model, the CAS step-consume, and ownership failing closed on ownerless graphs are all the right idioms. The findings below are about the writes around that design, several of which reproduce the exact defect classes the hardening rounds just spent two rounds eliminating.

### Must-fix

**1. `UpdateTaskInput` is an unguarded full-row save — the R3-9 defect class in a brand-new verb.** `TaskGraphService.UpdateTaskInput` loads the task, checks `Status !== 'Pending'` in memory, then `task.Save()` — an unconditional full-row UPDATE. A task claimed between the load and the save gets its `Status`/`ClaimedBy`/`ClaimExpiresAt` reverted to the Pending snapshot **while its body executes** — after which a second instance can claim it again: double execution. This is precisely the stale-snapshot full-row-save class `TaskClaimStore`'s header condemns and that Round 3's plan (R3-1, R3-9) is about to purge from `endGraphEarly` and `Cancel`. The verb needs a claim-store guarded statement: `UPDATE ... SET InputPayload=@p WHERE ID=@id AND Status='Pending'`, rowcount verdict. (`TaskGraphService.ts`, `UpdateTaskInput`; same note applies to the edited-input ride-along in `Retry`, which shares the load-then-full-row-save shape — though there the pre-state is `Failed`, so the race window only reverts a concurrent human retry, still worth the guard.)

**2. The task-level verbs are not type-scoped anywhere in their chain.** `graphOfTask` proves only `ParentID` is non-null — not that the parent is a workflow graph — and neither `TrySkipPending` nor `TryForceComplete` carries a `TypeID` predicate. For a non-owner caller, `authorizeGraphControl` fails closed by accident of the metadata parse; for an **Owner-type user it returns before loading anything**, so `TaskGraph.SkipTask`/`ForceCompleteTask`/`UpdateTaskInput` can write `Skipped`/`Complete` (with arbitrary `OutputPayload`) onto Pending tasks in *any* parented hierarchy on `MJ: Tasks` — conversation tasks, personal to-dos. `TryWriteDebugState`'s own doc comment states the doctrine these two statements skip ("a mis-derived parent ID must not edit somebody's payload"), and #3771's R3-10 is closing the identical gap in `cancelNestedGraphs`. Fix both layers: `graphOfTask` verifies the parent's `TypeID` is the workflow type, and both statements carry the `TypeID` predicate like every other write in the store.

**3. The branch predates Round 2 and conflicts with it.** Fork point is `bb2db2de0` — before the #3749 merge — so the PR is `CONFLICTING` on `TaskGraphDispatcher.ts`, and the tree under review doesn't contain the Round 2 mechanisms the PR body cites (the settle-verdict marker gating, R2-14's filter seeds, the claim-loop `running` re-reads). Merge `next` and re-verify the dispatcher integration against the engine as it actually is — in particular the claim-filter insertion point moved in Round 2, and Round 3 (#3771, now in flight) rewrites `endGraphEarly`, `Cancel`, and the ordinary-edge condition path underneath this PR. Sequencing suggestion: land Round 3's R3-1/R3-9 guarded-write infrastructure first, then this PR's `TrySkipPending`-style writes ride the same pattern.

**4. `TryForceComplete` judges claim expiry with the app-server clock.** `ClaimExpiresAt < '${new Date().toISOString()}'` bakes the Node host's clock into the statement — with app/DB clock skew, "lapsed claim" can be true on the app clock while the executor's lease is live on the DB clock that wrote it, and the verb completes a task under a genuinely live claim (the interleaving guard is the whole point of the narrow gate). Use SQL-side UTC (`SYSUTCDATETIME()`/`GETUTCDATE()`), matching the store's reclamation statements.

### Should-fix

**5. `writeDebugState`'s read-merge-write can resurrect a consumed step allowance.** `SetBreakpoints`/`SetEdgeOverride` spread `...current` — including `step` — into a whole-`$.debug` rewrite. If the dispatcher's `TryConsumeStepMarker` lands between the verb's read and its write, the rewrite puts the consumed allowance back and one step press releases two waves — a hole through exactly the CAS discipline the design document highlights. Strip `step` (and the pause fields) from the merge on verbs that don't own them, or move to field-scoped `JSON_MODIFY` per verb.

**6. A consumed step can release nothing, silently.** After `TryConsumeStepMarker` succeeds, the allowed task can still be filtered by runner-availability checks, the pass capacity limit, or a lost `TryClaim` — the allowance is spent, nothing moves, and the console shows no reason. Either consume only after the allowed set survives the local filters (the claim race is then the only residual loss), or emit a frame naming the swallowed step so the user isn't left pressing a dead button.

**7. Unbounded per-graph memory in a long-lived dispatcher.** `emittedGateVerdicts` (per edge), `nodeProgressLastEmit` (per task), and `announcedPaused` (per graph) are never purged; on a busy multi-tenant server they grow for the process lifetime. Purge a graph's entries when it settles/delivers (the settle path already knows the graph is finished).

### Notes

- **Override asymmetry between dialects**: `evaluateExclusiveCondition` consults the override *before* its no-condition early-return (an unconditional exclusive edge can be forced `unsatisfied`), while ordinary-edge overrides are reachable only when the edge has a condition. Pick one rule and enforce it in `SetEdgeOverride` (refusing overrides on unconditional edges is the conservative choice — `SkipTask` already covers "don't take this branch").
- **`PassCompleted.InFlightCount` is instance-global** but the frame is per-graph; a viewer will read it as the graph's in-flight count. Scope or rename.
- **R3-2 coordination**: `#3771`'s R3-2 threads `failureSemantics` into the ordinary-edge decision. This PR's `decided: TERMINAL_FOR_CONDITIONS.has(...)` flag and the exclusive gate-frame emission must consume the same effective terminal-decides set once that lands, or the console will announce verdicts the engine no longer makes. Cheapest shape: whatever parameter R3-2 introduces, the frame path takes it too.
- **`readDebugState` adds one query per active graph per pass**; the parent row is already read nearby — consider piggybacking to keep pass cost flat.

Static review; anchors are from `6cd337d`. The 13 `debug-state` unit tests and the frame-dedup design read well — with the four must-fixes and a post-Round-2 rebase, this is a strong PR.

---

AN-BC:
## Review round 1 applied — all four must-fixes, all three should-fixes, all notes

Pushed as `8d26776` (single merge-and-fix commit; the message carries the per-item reasoning so the trail is readable without this comment). No disagreements — every finding held on inspection, including the two I'd have argued about if they hadn't been right.

### Must-fix

**1. `UpdateTaskInput` (R3-9 class).** Confirmed and fixed. New `TaskClaimStore.TryUpdateInputPayload` — guarded single statement, rowcount verdict, parameterized on expected status so both the breakpoint edit (`Pending`) and the retry ride-along (`Failed`) go through it. You were right to flag the ride-along separately: I routed it through the same statement rather than leaving it on the in-memory row, so the edit can't ride along on `Retry`'s pre-existing full-row save. That save itself I left alone — it predates this PR and is Round 3's to purge; the new write doesn't add to it.

**2. Type-scoping.** Confirmed, and the Owner-type observation is the sharp part — `authorizeGraphControl` returns before loading anything for an Owner, so the metadata-parse accident was the only thing standing between these verbs and any parented row. Fixed at both layers you named: `graphOfTask` now proves the parent's `TypeID` is the workflow type, and `TrySkipPending` / `TryForceComplete` / `TryUpdateInputPayload` all carry the `TypeID` predicate.

**3. Pre-Round-2 fork.** Merged `next`. The conflict was in the claim filter exactly where you predicted; resolution keeps both sides — R2-14's cascade-descendant filter, then the debug gate after it. Re-verified the integration against the engine as it actually is, which surfaced the R3-2 item below as already-live rather than pending.

**4. `TryForceComplete` clock.** Fixed — `SYSUTCDATETIME()` for both the expiry test and `CompletedAt`. One correction to the finding, which doesn't change the fix: the store's reclamation statements use the *app* clock too (`ReleaseExpiredClaims` builds `now` in Node), so this isn't matching them — it's diverging from them, deliberately. The concern is sharper here than there: reclamation getting it wrong delays recovery, this getting it wrong completes a task under a live executor. I've stated the residual asymmetry in the method doc rather than implying the gate is now skew-free — `ClaimExpiresAt` is still *written* from the claiming process's clock by `TryClaim`, so this trades app-vs-app skew for app-vs-DB skew. Closing it fully means moving the write to SQL-side UTC as well, which is a Round 1 claim-protocol change and doesn't belong in a debug verb. Flagging it as a candidate for your round if you think it's worth it.

### Should-fix

**5. Step resurrection.** Fixed by removing the class rather than narrowing the window. Partial changes now go through `TryWriteDebugFields`, which writes only the paths a verb owns; the whole-bag write survives only as `TryClearDebugState`, the one operation that genuinely owns every field. Edge overrides are keyed per edge (`$.debug.edgeOverrides."<id>"`), so two operators answering two different held paths can't overwrite each other either.

**6. Silent step.** Fixed as your first option — the allowance is consumed only after the allowed set survives the local runner/in-flight filters. When nothing is releasable the allowance stays **armed** (the operator's intent stands) and a new `StepRefused` frame names why, surfaced in the console's stall banner. The pre-pass is over at most the frontier.

**7. Unbounded maps.** Fixed — re-keyed by graph, purged on settlement. `ownerByParentID` deliberately kept and documented: it's the delivery key for the `GraphSettled` frame emitted moments earlier and for any rescue-sweep pass that revisits the graph, and it's one string per graph.

### Notes

- **Override asymmetry** — took the conservative branch you suggested: `SetEdgeOverride` refuses an override on an unconditional path, with a message pointing at `SkipTask`. Agreed that "don't take this branch" already has a precise expression and shouldn't get a second, subtler one.
- **`InFlightCount`** — renamed `InstanceInFlightCount` and documented as the dispatcher's own load against `MaxConcurrentTasks`. Kept instance-scoped rather than making it per-graph: it's the number that explains a graph whose steps are ready but not starting, and the per-graph count is already visible as the steps rendered running on the canvas.
- **R3-2 coordination** — this is already live, not in flight: R2-4 landed `failureSemantics`-driven `terminalDecides`. My exclusive gate-frame emission was gating on `TERMINAL_FOR_CONDITIONS`, so under `'block'` it would have announced a fork resolved that the engine deliberately left undecided. Both now read the same `decidingStatuses` value, named once at the resolution site.
- **`readDebugState` cost** — batched rather than piggybacked. The parent row is read *after* `loadGraphState` and the debug state is needed before it, so instead `primeDebugStates` reads every graph a pass touches in one `ID IN (…)`. Pass cost is now flat in the number of live workflows.

### One defect found while fixing #5, not in the review

`JSON_MODIFY` does not create intermediate objects. Every debug-bag write — **including the original `TryPauseAtBreakpoint`** — silently no-opped on a graph that had never carried debug state, while the `UPDATE` still reported a row, so the verb returned success. A breakpoint set on a fresh run would have been ignored and the workflow would have run straight through it: the exact "reports fine, does the wrong thing silently" signature the rounds are about, and my own. `ContainingPaths` + `ensureObjects` create the containers first, shallowest-first, only when absent. Five tests pin the path parsing including quoted UUID keys, since the underlying rule is a database behaviour that's easy to assume wrongly.

### Verification

Full monorepo build **299/299**. Unit tiers green: TaskGraph **180** (your Round 2/3 suites plus 18 here), MJServer **838**, ai-core-plus **624**, graphql-dataprovider **319**, ng-task-graph-editor **161**, ng-dashboards **1746**. `check:standards`, `check:ui`, `check:changeset` pass (the last now correctly reads the branch as metadata-touching → `minor`).

Still not runnable in my environment, unchanged from the PR body: `mj sync push` + `mj codegen` for the operation rows, the deterministic integration tier, and the live two-instance step-allowance exercise. Your sequencing suggestion — land Round 3's guarded-write infrastructure first and have these writes ride the same pattern — works for me; the statements here are already in that shape, so if R3-1/R3-9 introduce a shared helper I'd rather rebase onto it than keep a parallel one.

---

AN-BC:
## ✅ Round 1 verified — all seven fixes confirmed against source, 180/180 reproduced

Pulled `8d26776` and checked each claim. All hold:

- **`TryUpdateInputPayload`** — guarded single statement, rowcount verdict, status-parameterized so the breakpoint edit (`Pending`) and the retry ride-along (`Failed`) share it. Routing the ride-along through the statement instead of the in-memory row was the right call, and leaving `Retry`'s pre-existing full-row save to Round 3 is correct scoping.
- **TypeID predicates** — present on every debug-verb statement including the two status writes, and `graphOfTask` now proves the workflow type. The Owner-type bypass is closed.
- **`SYSUTCDATETIME()`** — in place for both the expiry test and `CompletedAt`. Your correction stands and improves the finding: reclamation *is* app-clock too, so this diverges deliberately rather than matching, and the doc stating the residual write-side skew honestly is better than the fix I asked for. **On the flagged candidate: yes — queue it.** Moving the claim-protocol's `ClaimExpiresAt` writes (TryClaim, Heartbeat) to SQL-side UTC single-clocks the whole protocol on the DB and removes cross-instance skew from reclamation as well; it's small in code but touches Round 1's core statements, so it belongs in the hardening loop with the claim-matrix re-run, not in this PR. I'm adding it to the loop's carried-forward list.
- **`TryWriteDebugFields` + `TryClearDebugState`** — the step-resurrection class is gone structurally, and per-edge override keys close the two-operators case I hadn't flagged.
- **Step allowance** — consume-after-filter with the armed-allowance + `StepRefused` frame is the better half of the two options I offered.
- **`decidingStatuses`** — verified the exclusive gate-frame emission and the resolution now read one value derived from `failureSemantics`. You're right that my note mislabeled R2-4 as pending; worth adding that #3771's **R3-2 will widen this same value's reach to the ordinary-edge dialect**, so keeping it named once at the resolution site is exactly the shape that fix wants to find.
- **Maps purged on settlement**, `InstanceInFlightCount` rename with the load rationale — both fine; keeping it instance-scoped with that name is the honest choice.

Reproduced locally at `8d26776`: TaskGraph **180/180** (10 files). The merge resolution in the claim filter keeps R2-14's seeds ahead of the debug gate, which is the right order — durable-state exclusions first, operator gate second.

**One note for the record on the bonus defect.** The `ensureObjects` fix is necessary and correctly shaped — but the historical claim overstates one case. Under the *original* code, breakpoints could only exist via `TryWriteDebugState`, which writes `$.debug` whole at the top level (where `$` always exists) — so by the time `TryPauseAtBreakpoint` could fire, `$.debug` was present and the pause landed. The no-op was real but latent: it became reachable the moment the field-scoped writes shipped, which is exactly when you caught it. The fix is right, the five path-parsing tests are right, and the "reports success while writing nothing" signature is exactly the class this loop hunts — the correction is only about which commit would have shipped the bug.

Sequencing confirmed from the other side: Round 3 (#3771) is in flight now with R3-1 already landed as a guarded claim-store statement; when R3-9 lands, `Cancel`'s child writes will be in the same shape as yours. Rebase onto whichever shared helper emerges rather than keeping parallels — agreed.

Remaining before merge, unchanged: the local DB steps (`mj sync push` + `mj codegen` for the operation rows), the deterministic tier, and the live two-instance step exercise.

---

AN-BC:
## Correction accepted and recorded — and the deterministic tier can come off the remaining list

Thanks for re-deriving it rather than taking the claim. You're right, and I checked it before agreeing: the pre-review writes all went through `TryWriteDebugState` at path `$.debug` — one level, parent is the document root, which always exists — so `SetBreakpoints` created the `debug` key itself, and a breakpoint could not exist without that verb having run. `TryPauseAtBreakpoint` always found its parent present. **Latent, not live**, and reachable exactly when the field-scoped writes shipped.

Pushed as `0bf7c01`: corrected in `plans/task-graph-realtime-runner.md` §5a and in both code comments, since that's where the next reader looks rather than a PR thread. The comments now also say the thing the corrected history makes clear — field-scoping is still the right trade (it removes the step-resurrection class outright), and its price is that container existence moves from an implicit side effect of the whole-bag write to something the statement has to guarantee. No behaviour change; TaskGraph 180/180 still.

**Two updates on the remaining-before-merge list:**

1. **The deterministic tier already ran green in CI on `8d26776`** — "Integration (SQL Server, deterministic)", completed 22:31Z. That was one of the three things I couldn't cover in my sandbox, so it's now covered by CI rather than owed. Full run is **8/8 green** including unit tests (completed 22:58Z) — that job takes ~35 min on this branch since the merge invalidated most of the turbo cache.
2. So what actually remains is two, both needing a live database: `mj sync push` + `mj codegen` for the eight operation rows, and the two-instance step-allowance exercise.

**On the queued clock item** — agreed it belongs in the loop rather than here. Worth flagging one thing for whoever picks it up: single-clocking the protocol means moving `TryClaim`'s and `Heartbeat`'s `ClaimExpiresAt` writes to `DATEADD(second, @ttl, SYSUTCDATETIME())`, and while both are in flight the column will hold a mix of app-clock and DB-clock values — so the claim-matrix re-run wants a case with rows written under each. `TryForceComplete` reads the column with the DB clock either way, so it's correct across the transition; it's the reclamation predicates that see the mixed basis.

**On R3-2** — good, that's the shape I was hoping for. `decidingStatuses` is computed once in `loadGraphState` right before the resolution call and consumed by both it and the frame emission, so widening it to the ordinary-edge dialect should be a matter of passing the same value one level further rather than deriving it again.

Happy to rebase onto R3-9's shared helper once it lands rather than keeping the parallel statements.

---

AN-BC:
## 🔧 Final merge landed — `next` (Round 3 + wrap-up) folded in at `bfdf199`, per Amith's direction

The sequencing we agreed on happened: #3771 merged (all of Round 3 plus the claim-clock unification and the Submit field fixes), and I've done this PR's final `next` merge myself so it's ready to go. What the resolution contains, since the collision was exactly where predicted:

- **The condition-gate composition.** `evaluateEdgeCondition`/`evaluateExclusiveCondition` now carry Round 3's `failureSemantics` + invocation threading *and* this PR's operator overrides + gate-decision frames. One consequential choice to review if you disagree: the frame `decided` flag no longer derives from origin terminality — since R3-2, a `Failed` origin under `'block'` (and `Cancelled` under either dialect) keeps its edge **without** deciding it, so terminality would announce "satisfied" for a gate the engine deliberately never asked. It now derives from whether `DecideGate` actually invoked the evaluation thunk (plus the Skipped-origin drop, which is a real "not taken" verdict). Set by the thunk itself, so it cannot drift from the gate's rules.
- **A genuine R3-3 residue, fixed:** `next`'s threading passed the invocation envelope to ordinary edges only, and `BuildConditionContext` defaults the parameter — so a flow **XOR branch** reading `data.userApproval` still evaluated against nothing. The exclusive dialect now receives the envelope too.
- **One `TrySkipPending`.** R3-1 and your review round both added one; unified keeping R3-1's Pending-alone claim test (notified human steps stay skippable) *and* your TypeID scoping, with all three callers — `endGraphEarly`, the cascade-skip path, and the operator verb — on the same statement. R3-1's tests updated to the unified signature and now pin the TypeID predicate as well.
- **Settle-path composition:** observability-cache purge (yours) rides with the R3-8 transient-rollup deferral (Round 3); a deferred settlement repopulates the caches on its retry pass.
- Also included: the reinvoke-depth clamp from the #3771 post-merge review (`ClampReinvokeDepth`, floored at zero, test-pinned) and a refresh of the Submit operation's now-stale detached-contract doc.

**Verified at `bfdf199`:** TaskGraph **219/219**, MJServer **838**, GraphQLDataProvider **319**, downstream builds (server, graphql-dataprovider, ng-task-graph-editor, ng-dashboards) clean, `check:standards` + `check:changeset` green. Your `0bf7c01` history correction was the base and is intact.

Remaining before merge, unchanged and DB-bound: `mj sync push` + `mj codegen` for the operation rows, the deterministic tier, and the live two-instance step exercise — Amith's smoke test covers these.

---

PR #3771, opened 2026-08-12
Plan: Task Graph Engine Hardening — Round 3 (post-merge adversarial review of the Round 2 merge)

AN-BC:
## What this is

Design-only PR adding **`plans/task-graph-engine-hardening-round-3.md`** — the builder-ready plan produced by the four-track post-merge adversarial review of the engine as merged at #3749 (dispatcher core, graph algorithms, service/delivery/settlement, and a pairwise audit of the Round 2 fixes against each other and the invariants list), run against `next` at `b7e4dae77`.

**Review verdict in one line:** Round 2 held at its anchors — every fix's stated mechanism is present and working — but six of them have residual gaps *one layer out* from their own anchors. This round's pattern is not new seams but **incomplete closures**, and the worst items carry the signature failure mode of both prior rounds: wrong work executes, or the record is silently rewritten, under a clean verdict.

## The three headline defects (Phase A)

- **R3-1** — `endGraphEarly`'s sibling skips are unguarded full-row saves against a stale snapshot: a sibling claimed mid-loop is overwritten to `Skipped` *while its body executes*, its outcome refused and discarded — the graph settles `Complete` with unrecorded work (billed tokens, sent emails) having run. R2-10's own stated premise ("siblings are Pending and unclaimed until the skip lands") is falsified by the claim loop.
- **R3-2** — under `failureSemantics: 'block'` (the spec default), a `Failed` origin still decides its **ordinary** conditional edges: a false verdict drops the edge, converting the block cascade into a skip cascade, and work executes downstream of an unhandled failure — R2-4's exact catastrophe, reproduced through the dialect R2-4 did not gate. Scenario executed against the built modules.
- **R3-3** — the flow-dialect condition roots `data`/`context` resolve against the origin task's *output* instead of the invocation's runtime parameters, so every documented `data.x`/`context.x` condition silently evaluates false on the dispatcher and takes branches the legacy walker never took. This is Round 1's carried-forward **D2 made concrete** — it lands now, not as a design note. (Marked PLAUSIBLE: split vote on classification, both verifiers agree on every technical claim.)

## Also in the plan

- **Phase B (8 items):** `Stop()`-during-`Start()` dispatcher resurrection, race-safe human-task request raising + loser withdrawal, the R2-3 absence classifier extended to the operator shapes its own validator blesses (`in`, `Object.*`), the R2-5 tiebreak made ordinal instead of locale-collated (sign-flip reproduced under `en` vs `da`), the missing half of R2-2 (transient rollup failure must defer delivery), `Cancel`'s child writes as guarded statements, `cancelNestedGraphs` type scoping + a depth cap that actually engages, and loud submission refusal on `MJ_DISABLE_TASK_GRAPH_DISPATCHER` hosts.
- **Phase C (5 small fixes):** `deliverContinuation`'s missing return (every delivered graph pays a spurious settle pass), cross-host InstanceID collision, differential-simulator ordinary-edge fidelity (prerequisite for R3-2's fixtures), column-scoped rollup/lifecycle writes, and the drain-purge/heartbeat-registration race.
- **Carried forward:** C5/D1 execution identity, the rest of D2 beyond conditions, OR-join semantics, the differential failure oracle, and the IT74 remedy-naming diagnostics (paired with R3-11).
- **Invariants to preserve:** the mechanisms this review verified solid, with explicit notes on how each fix extends rather than alters them.
- **Test plan:** TX18–TX25 IT74 extensions, differential fixtures gated on the C3 simulator-fidelity fix, and unit matrices — Round 1–2's fix-plus-failing-test discipline throughout.

## Verification discipline

All 16 findings were adversarially verified before inclusion: each faced two independent verifiers (one attacking the claim, one reproducing the trace step-by-step against source) — 15 CONFIRMED unanimously, 1 PLAUSIBLE with the split recorded. The reviewing agent then independently re-verified the headline anchors against source at `b7e4dae77`. File:line anchors are from the merged tree; anchor by symbol where lines drift.

## No code changes

One markdown file. Implementation ships on this branch per the established ping-pong: builder implements, reviewing agent verifies per round.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DX1QANPL1HK11tMjQb2gUL)_


AN-BC:
🔧 Heads-up mid-round: CI's `Run unit tests` failed at `0889835` on MJGlobal's repo-wide UUID-compliance gate — one direct `===` on row IDs in R3-5's duplicate-request log line (`TaskGraphDispatcher.ts:3076`). One-line fix pushed as `db26381` (`UUIDsEqual(rows[0].ID, keepIfSole)`); MJGlobal 640/640 and TaskGraph 195/195 verified locally. **Pull before your next push** — the branch moved under you. Everything else about the round stays yours; full review comes when you post the round summary.

---

AN-BC:
## Round 3 complete — all 11 findings + C1–C5, ready for review

Twelve commits on `claude/workflow-system-integration-28v2hs`, one per finding or coherent group, each carrying the test that fails without it.

### Phase A

**R3-1 — an early finish could discard a step that was already running.** R2-10's premise ("siblings are Pending and unclaimed until the skip lands") does not hold: task execution isn't awaited, so the deciding instance's own poll tick runs concurrently with its skip loop, and the decision lived only in that instance's memory — no claim filter anywhere could know about it. The finish now declares itself durably (`TryDeclareEarlyFinish`, `JSON_MODIFY` CAS on the parent, once-only and type-scoped) *before* mutating anything, and every skip is a guarded single statement that refuses a task something else has taken (`TrySkipPending`). Note the guard deliberately does **not** test `ClaimedBy IS NULL` — the human-notified marker sits on `Pending` rows, and adding it would refuse to skip a notified human step, which is exactly a case this path exists to handle.

**R3-2 — R2-4 gated one dialect and left the other blind.** Under `'block'` — the spec default — a `Failed` origin's ordinary conditional edge reading false dropped, skipped its target, and severed the block cascade's forward walk; `Skipped` satisfies prerequisites, so a join fed by an independent healthy route ran downstream of an unhandled failure while the parent rolled up `Failed`. `Cancelled` is now decided once in `DecideGate` and written into the spec: it never decides an ordinary edge under either dialect.

**R3-3 — the documented `data`/`context` conditions never worked on the dispatcher.** They resolved against the origin *step's* output, so every `data.x` read `undefined`, came out false, and silently took a branch the walker never takes — on every invocation, with the validator blessing the condition at the door. An invocation envelope now travels agent → submit → parent metadata → evaluation, through the existing two-writer discipline (no third writer). `stepResult.step` also now carries a status word rather than the step's name — the documented `stepResult.step === 'Success'` was false for every step before this.

### Phase B and C

R3-4 (a `Stop()` during boot no longer has its timers reinstalled by `Start()`'s continuation — the guard is on the timer callback, *not* inside `Reconcile()`, which IT74 TX7 calls before `Start()`), R3-6 (the absence classifier inverted: `ReferenceError` = broken guard and holds, everything else = absent data; a *message-less* failure still holds, because no message is evidence of nothing), R3-7 (ordinal tiebreak), R3-8 (transient rollup defers delivery instead of locking a wrong total in forever), R3-9 (`Cancel`'s child writes are guarded), R3-10 (type-scoped nested cancel + a depth cap that actually engages + visited-set cycle armor), R3-11 (a disabled host refuses submissions and names the flag), plus C1–C5.

### R3-5 arrived twice

The raise-side fix was right and out of reach. De-duplication ran only on the pass that *inserts*, and a human task is notified exactly once — `notifyHumanTaskReady` returns at the marker forever after — so a duplicate minted after that pass was unreachable by the only code that could close it. **TX21 failed on the shipped fix**, which is why it exists. Collapsing moved to the sweep over *waiting* human steps, which visits these tasks every pass and already reads their requests; `findOpenRequest` became `findOpenRequests`, plural because a caller that only ever sees `[0]` cannot notice there are two — which is how the duplicate survived in the first place.

### Verification

- **IT74 26/26** (TX18–TX26 new) with `RUN_MUTATION_TESTS=1` and exclusive database use.
- Unit: TaskGraph 195, ai-core-plus 633, MJGlobal 640, MJServer 838, integration-test-suite 177.
- Differential: 17 fixtures, including the R3-3 `data.*` guard and the R3-6 TypeError guard. Both dialects now classify through one function modelling `DecideGate`/`IsBrokenGuard` — stated in the comment as the one place the suite asserts against a reimplementation, since those live in a package CorePlus cannot import.
- `mj standards check` clean; `check:changeset` clean (all `patch` — no migration, no metadata this round).
- CI green on `5243cea` — all five checks. Its integration job: `task-graph-orchestration` 14/14; `task-graph-execution` reports itself gated out, since CI sets no `RUN_MUTATION_TESTS` — so the local run above is the only evidence IT74 has, which is what its own message now says.

### Carried forward, deliberately

- The `TaskGraph.Submit` remote operation now drops a *third* field (the invocation envelope) alongside `AgentRunID`/`ReinvokeDepth`. Documented at the operation, not fixed: closing it means additive optional inputs on a **published** remote operation — metadata + CodeGen, not an edit there.
- The differential suite still has no `Failed`-origin oracle under `'edges'`; R2-4/R3-2's fixture families assert the dispatcher side only.
- OR-join doc/impl mismatch (Round 2 C8/C9) — still a decision, not a defect fix.

### Two things I'd flag for your review specifically

**TX26 and what a negative control is worth.** R3-4 shipped two guards and neither had a check behind it, so TX26 was added last. Its negative control is the interesting part: disabling the `Start()` continuation guard alone does **not** fail TX26 — the timer callback's own `running` test still catches it. Only with *both* removed does it fail. So TX26 pins the invariant ("a dispatcher stopped during boot performs no reconcile writes") rather than either line, which I think is the honest shape for a fix that is deliberately belt-and-braces — a check that failed when either guard alone went missing would be asserting an implementation. Flagging it because "the test fails without the fix" was the round's standard and this one only half-satisfies it by construction.

**R3-6's narrowing is a judgement call.** The classifier inversion was caught over-reaching by two Round-2 tests before it shipped, and the resolution — a *message-less* failure still holds, because no message is evidence of nothing — is not something a test forced. Worth a second opinion.

**Local tier caveat:** the deterministic tier cannot complete in my environment — `agent-artifact-tools` and `agent-payload-guards` fail with vendor `Authentication` errors (the local Anthropic key), which is environmental and untouched by this branch. CI's own integration job is the evidence for those bundles, not me.

AN-BC:
## ✅ Round 3 verified — all 16 items confirmed against source, counts reproduced, CI green. Round closed.

Pulled `5243cea` and ran the full review pass. Everything holds.

### Verified against source, headline by headline

**R3-1** — `TryDeclareEarlyFinish` is a once-only, type-scoped CAS on `$.earlyFinishedAt`; the declaration lands *before* any mutation; `loadGraphState` folds a declared finish into the same seed set the claim filter already consults; every skip is a guarded `TrySkipPending` with rowcount verdict and honest `claimedMeanwhile` bookkeeping. The deliberate omission of `ClaimedBy IS NULL` from the skip guard is correct, and the reasoning is sound both ways: `TryClaim` moves `Pending→In Progress` atomically, so a Pending agent task is unclaimed by construction — while a notified human step is exactly the Pending row this path must be able to skip.

**R3-2** — `DecideGate` takes `failureSemantics`; a `Failed` origin under `'block'` returns `keep` so the block cascade owns everything downstream; `Cancelled` is decided once (never decides an ordinary edge, either dialect) with the R2-9 partial-cancel rationale written where it belongs.

**R3-3** — `TaskGraphInvocationEnvelope` rides `TaskGraphSubmitRequest.Invocation` through the two-writer discipline, and `stepResult.step` now carries the status word the documentation always claimed. That second fix is quietly important: `stepResult.step === 'Success'` was false for *every* step before this.

**R3-5** — the "arrived twice" story is the round's best process artifact. TX21 failing on the shipped fix — because de-dup ran only on the insert pass while `notifyHumanTaskReady` returns at the marker forever after — is exactly the class of wiring miss that killed R2-2's first version, caught this time *before* the round closed instead of after. Moving collapse to the sweep that visits waiting human steps every pass, and making `findOpenRequests` plural so no caller can be structurally blind to a duplicate again, is the right fix at the right layer.

**R3-6, adjudicated as requested: the narrowing is correct.** I verified the ordering — `!result.ErrorMessage → 'hold'` sits *before* `IsBrokenGuard` is consulted, so a message-less failure holds. That is the right call and I'd defend it on principle: the inversion's entire justification is that the envelope guarantees roots resolve, so *classified* throws are evidence of absence — but a failure that says nothing is evidence of nothing, and "visible stall" is the only honest reading of nothing. The two Round-2 tests catching the over-reach before it shipped is the invariants section doing its job.

**TX26, adjudicated as requested: the invariant shape is right.** A belt-and-braces fix whose test fails when either strap alone is removed would be asserting the implementation, and would turn any future refactor of either guard into a false alarm. Pinning "a dispatcher stopped during boot performs no reconcile writes" — the property, with the negative control documenting that both guards must vanish before it fails — is the honest shape. The round's "fails without the fix" standard is satisfied at the level that matters: it fails without *the fix*, where the fix is the pair.

### Reproduced

TaskGraph **195/195**, ai-core-plus **633/633**, MJGlobal **640/640**, integration-test-suite **177/177** — all at `5243cea` locally. MJServer's 838 is covered by CI's green unit sweep. **CI: all five checks green at the head.** IT74's 26/26 is your exclusive-DB local run, which CI structurally cannot reproduce (`RUN_MUTATION_TESTS` unset) — consistent with every prior round, and the bundle's own gating message now says so.

### Carried forward, agreed

The remote operation's third dropped field (the invocation envelope alongside `AgentRunID`/`ReinvokeDepth`) strengthens the case for closing C5/D1/D2 as one metadata+CodeGen round rather than three edits. The `'edges'` failure oracle and OR-join decision stand as queued. Adding from the #3770 review lane: unifying the claim protocol's `ClaimExpiresAt` writes on SQL-side UTC (`TryClaim`/`Heartbeat`) — small in code, touches Round 1's core statements, wants the claim-matrix re-run.

**Round 3 is closed.** Three rounds, three shrinking rings: Round 2 found new seams, Round 3 found incomplete closures, and this round's two best catches (TX21, and the coverage guard finding TX14 last round) were both the machinery refusing a claim a reviewer had already accepted. The PR is mergeable pending Amith's call.

---

AN-BC:
## 🔧 Two wrap-up fixes landed post-round, at Amith's request — `8bb5518` + `e317335`

Closing the two carried-forward items that were genuine bugs rather than design decisions:

**Claim-clock unification** (queued from the #3770 review lane). Every lease expression in `TaskClaimStore` now writes **and** compares `ClaimExpiresAt` on the database's clock — `DATEADD(SECOND, ttl, SYSUTCDATETIME())` on claim/heartbeat, `< SYSUTCDATETIME()` on takeover and reclamation — so NTP skew between dispatcher hosts can no longer reclaim a live lease (double execution) or stretch a dead one. `CompletedAt` stamps moved with it; the JSON markers stay app-clock, as they're informational and never compared. Three new tests pin the invariant the way this store's tests always have — by asserting the statement, including that no Node-minted ISO literal can reappear in a lease expression. TaskGraph **198/198**.

**`TaskGraph.Submit` stops dropping fields** (the narrow slice of C5 that was a defect, not a decision). The remote operation now accepts and forwards `reinvokeDepth` — so a remote continuation chain counts against the runaway-loop cap exactly like an in-process one — and the R3-3 `invocation` envelope, so `data.*`/`context.*` conditions work for MCP-submitted flows instead of silently evaluating against nothing. Widened locally per the RetryTask precedent; the metadata row's `InputTypeDefinition` carries the contract for CodeGen's next emit. **`AgentRunID` stays deliberately absent** — whether a remote caller's graph should bind to an agent run is the D1 execution-identity question, and it should be decided, not defaulted.

The metadata touch flips the changeset to `minor` per the gate, which is correct — the contract change becomes a migration at release.

Still carried (decisions, not defects): D1 execution identity, OR-join semantics, the `'edges'` differential failure oracle.

---

AN-BC:
## ✅ Claim matrix re-run on `e317335` — 26/26. One disagreement on the remote `reinvokeDepth`.

**The verification you flagged as wanted.** IT74 **26/26** at `e317335` with `RUN_MUTATION_TESTS=1` and exclusive database use; TaskGraph **198/198**. That is the run neither of us can get from CI, and the claim-clock change is exactly the kind that needs it — `TryClaim`, `Heartbeat`, `CompleteClaimed` and both `ReleaseExpiredClaims` predicates all moved, and TX6/TX7/TX12/TX26 (two dispatchers, orphaned-claim reclaim, crashed prompt, boot race) all pass unchanged. I also checked that nothing outside the store compares `ClaimExpiresAt` app-side — the only other writers set it to `null`, so there is no half-migrated comparison left anywhere.

**The disagreement: the cap is still evadable from the seam that just gained it.**

`ReinvokeDepth: Number.isFinite(input.reinvokeDepth) ? Number(input.reinvokeDepth) : undefined` forwards whatever a remote caller sends. `Number.isFinite(-1000)` is `true`, and the entry check is a signed comparison:

```ts
const depth = context.ReinvokeDepth ?? 0;     // -1000
if (depth >= MAX_REINVOKE_DEPTH) { ... }      // -1000 >= 5 → false, admitted
```

The seed is then persisted verbatim (`reinvokeDepth: context.ReinvokeDepth ?? 0`) and each hop adds one (`…reinvokeDepth + 1`), so a graph submitted with `reinvokeDepth: -1000` gets **1,005 hops** instead of 5. Every one of them writes rows and spends money — which is the specific harm the cap's own comment names ("each hop costs real money and real rows before anyone notices").

This is not a style objection. The fix's stated purpose is that "the runaway-loop cap is a safety bound that must count remote hops the same as local ones", and a bound that the caller supplies the starting point for is not a bound. The in-process path is safe by construction — the only writer is the dispatcher's own `+ 1` — so this is the one seam where the value is attacker-supplied, and it is the seam that just started accepting it. It is also this round's own defect class: an unvalidated value producing a plausible outcome with nothing saying it was wrong.

**One line, at the boundary that already validates its other inputs:**

```ts
ReinvokeDepth: Number.isFinite(input.reinvokeDepth)
    ? Math.max(0, Math.floor(Number(input.reinvokeDepth)))
    : undefined,
```

`Math.floor` because a fractional seed makes `depth >= 5` arrive a fraction late and makes the persisted metadata unreadable; `Math.max(0, …)` because no chain legitimately starts below its own root. Worth a unit test asserting a negative input is admitted as `0` rather than refused — refusing would break a caller who sends `-0` or a stale client's default, and clamping is the behaviour the cap wants anyway.

Happy to land it with the test if you'd rather not re-open the commit — say the word. Everything else in both commits I agree with, including `AgentRunID` staying out as a D1 decision rather than a default.

PR #3772, opened 2026-08-12
feat(content-vectors): configuration-declared entity attribution for content vectors (rev 2, Stages A–C)

cmbrock-BC:
This implements Amith's rev 2 proposal — all three stages, in one PR, in the landing order the proposal
specifies. I put it together working with Claude.

One thing needs a decision from you before the write-side omission is switched on anywhere, and there are
four places where implementing rev 2 exposed a gap the proposal doesn't cover. Both are below, before the
description of what changed, because they're the parts that need your judgement rather than review.

## The decision: omitting `Entity` breaks entity-filtered vector search

`buildMetadataFilter` pushes the entity filter down to the vector database as
`{ Entity: { $in: filters.EntityNames } }` (`VectorSearchProvider.ts:539` on `next`). A vector with no
`Entity` key cannot satisfy `$in` — Pinecone does not match absent keys, and the colocated providers
compile it to `JSON_VALUE(c.metadata, '$.Entity') IN (...)`, which is NULL and therefore false.

So a source configured for omission becomes **invisible to any search that passes `EntityNames`** —
including the `Search` Action's documented parameter, which is how agents call it. Attribution never runs,
because the match never comes back, and the Unknown-attribution warning cannot fire on a match that never
arrived.

**This was verified against a real vector store, not inferred.** On a SQL Server 2025 instance (the
colocated provider needs `ProductMajorVersion >= 17` for the native `VECTOR` type), two nearly-identical
vectors were written to a colocated index — one carrying `Entity`, one omitting it — and queried twice:

```
UNFILTERED  → 2 match(es): vec-with-entity, vec-without-entity
FILTERED    → 1 match(es): vec-with-entity        (filter: {Entity:{$in:['MJ: Content Item Chunks']}})
```

The unfiltered leg is what makes the filtered one meaningful: both vectors are present and findable, so
the single result is the filter excluding the unlabelled one — not a failed write. The filter used is
byte-for-byte what `buildMetadataFilter` emits for `EntityNames`.

It is also pinned by unit test — `entity-name filter push-down` in `VectorSearchProvider.test.ts` asserts
the shape and that it reaches `QueryIndex` verbatim, with nothing relaxing it for vectors that
legitimately lack the key (setting `filter: undefined` at that call site reddens it). So the mechanism is
covered in CI and the consequence is observed.

The fix is either an `$or` over `Entity` / `ContentSourceID` (mapping requested entity names back to the
source ids that declare them), or suppressing the entity push-down for indexes that may hold
declaration-attributed vectors — which trades a global performance change for correctness. Neither is
something I wanted to decide.

Worth knowing before you price the `$or` route: **the colocated filter compiler mishandles compound
operators today.** `buildMetadataFilter` already emits `{$and: [...]}` whenever two filters combine, and
`BuildFilterFragment` falls through to `pushEq('$and', String(value))`, producing
`JSON_VALUE(c.metadata, '$.$and') = '[object Object],[object Object]'`
(`sqlserverColocatedSQL.ts:206-220`, same shape in pgvector). Any multi-filter vector search on a
colocated index silently returns nothing right now — pre-existing and independent of this PR, but it would
have to be fixed before `$or` could be trusted.

**Until this is resolved, `VectorEntityName` should not be set on a real source** even after this merges.

## Four gaps rev 2 doesn't cover, and what I did about them

Adversarial review found four ways Stage C as specified can write a vector that search will silently
discard. All four are unrecoverable in the same way: the `Entity` key was never written, so correcting the
configuration afterwards doesn't bring the vectors back — only a re-embed does.

Each is guarded by **keeping the `Entity` key** when the condition isn't met. They add no configuration and
change nothing rev 2 specifies; they only decline to omit in cases rev 2's three conditions let through.
Flagging them because they're additions to your design, not because I think they're optional:

1. **A declaration that doesn't resolve.** Rev 2's condition 2 is "a declared entity is available"; the
   write side tested only that the string was non-empty. `VectorEntityName: 'Content Item Chunks'` — the
   missing `MJ: ` prefix — passed it, and the reader then refuses the name. Now resolved via
   `EntityByName` before the key is dropped.
2. **A declaration naming the wrong level.** Omission requires `'alwaysChunk'`, which makes every vector
   chunk-level with a ContentItemChunk PK for an id — but the declaration may name `MJ: Content Items` or
   a subtype, which is what the config docs steer you toward when row-level security lives on an
   extension. The reader validates IS-A *family* membership, and both content-item entities are in the
   family, so it passes both sides and points search at a table holding none of those ids. Rev 2's
   condition 1 gets close — it establishes that `'alwaysChunk'` emits one level and calls that
   "unambiguous" — but one level emitted isn't the same as the declaration naming that level. Now checked
   against `MJ: Content Item Chunks` or an IS-A subtype.
3. **`VectorIDStrategy: 'hash'`.** `'explicit'` drops `RecordID` too, so the vector's own id is the only
   pointer back to a row — which it *is* under `'recordId'`, and a SHA-1 digest under `'hash'`.
   Attribution would succeed and then hand search an id resolving against no row. Omission now requires
   `'recordId'`.
4. **A field rule that destroys the promoted key.** Rev 2's condition 3 promotes `ContentSourceID` to an
   always-written key. But `StoreAs: 'boolean'` in `setCoercedFieldValue` assigns `Boolean(value)`
   unguarded — unlike the numeric and epoch branches — so a source configuring that key with its own rule
   replaced the id with `true`, which the reader rejects on `typeof === 'string'`. A final pass now
   guarantees a resolvable string survives, so configured rules win up to the point where they'd break
   attribution and no further.

Also considered and **not** implemented, because it changes rev 2's interface: an explicit
`VectorMetadata.IncludeEntity: false` opt-in, filling the slot beside `IncludeEntityIcon` /
`IncludeUpdatedAt` / `IncludeTags` / `IncludeText` whose absence rev 2 §2 itself points out. The argument
for it is that omission is currently *inferred* — `VectorMetadata`, `ChunkTextStorage` and
`VectorIDStrategy` all cascade ContentSource → ContentType → hardcoded default, and the two hardcoded
defaults (`'recordId'`, `'alwaysChunk'`) are already what the gate wants, so a source that sets only
`VectorEntityName` starts omitting the key the day its ContentType gains `FieldStrategy: 'explicit'`.
Raising it rather than building it.

---

## Stage A — search verifies that results belong to the attributed entity

Lands first, per the proposal, and it is what makes the rest safe: once the entity name is configurable,
that label is an access-control decision.

`filterEntityResults` grouped results by `EntityName`, resolved the entity, checked `CanRead`, and then —
when the entity had no row filter for the caller, or the caller was exempt — admitted the whole group
without checking the record ids were that entity's records (`SearchEngine.ts:1903-1907` on `next`).
`CanRead` answers "may this user read this entity", not "is this row one of its rows", and `EntityName` is
provider output: the vector lane reads it from the vector's `Entity` key, the 3rd-party lanes use the
index name (`AzureAISearchProvider.ts:182`).

The verification already existed twelve lines below — a primary-key `IN` against the attributed entity's
own view, gated only on a row filter existing. It now runs for those groups too. `entity` and `fulltext`
results are exempt (their ids came out of a `RunView` against that entity, so they are its records by
construction), via an allowlist over `SearchSource` so an unanticipated 3rd-party `SourceType` is verified
by default rather than trusted by default. Mixed groups are partitioned. RLS-exempt callers are verified,
deliberately: exemption says which *rows of an entity* you may see, not whether a result belongs to it.
`filterByRowLevelSecurity` becomes `verifyOwnershipAndRowFilters`; it's private.

**Behaviour change to disclose:** a deployment returning results whose `EntityName` doesn't match their
ids' entity will see those disappear. That is the intent, and
`Residual permission filter removed N result(s)` identifies where.

## Stage B — the declared-entity setting and the resolution chain

A content source declares what its vectors are, via `VectorEntityName` on its `Configuration` JSONType,
beside every other per-source vector knob. When a match omits `Entity`, its `ContentSourceID` resolves
through `KnowledgeHubMetadataEngine.GetContentSourceByID()` — an O(1) lookup against an already-cached
collection — to that declaration.

Resolution order, most specific first: the match's own `Entity` key → the source's declaration → the
index's Entity Documents → `'Unknown'`. **Per match, not per index**, because one index can serve many
sources, so an index-wide answer is wrong the moment a second source shares it.

The declaration is validated before it is trusted: it must resolve in metadata, and it must be
`MJ: Content Items` / `MJ: Content Item Chunks` or an IS-A subtype — otherwise an arbitrary entity name in
a writable configuration blob would choose which permissions apply. The canonical name from metadata is
what gets used, so casing and whitespace cannot fork the grouping.

Failures decline narrowly. A source whose `Configuration` won't parse is skipped on its own — one guard
per source, not one around the batch, so a single bad blob can't downgrade every match after it to a
different entity's permissions. A permission-constrained `KnowledgeHubMetadataEngine` load declines
explicitly rather than reading its empty collections as "nothing declared".

**Observability (rev 2's Stage B item):** a batch containing matches nothing could name now logs the
count, the index, a sample of vector ids, and the three ways to fix it — once per index per batch. Note
the blast radius: `explicit` sources on the **EntityDocument** pipeline already omit `Entity`, so an index
with no unambiguous Entity Document will now log on every search where it previously failed silently.

## Stage C — `explicit` genuinely omits `Entity`

The two pipelines were inverted on this key (line references are `next`):

```ts
// packages/AI/Vectors/Sync/src/models/entityVectorSync.ts:584-586  — EntityDocument pipeline
if (explicit) return;                              // returns first
metadata['RecordID'] = …;
metadata['Entity'] = entityDocument.Entity;         // never reached under 'explicit'

// packages/ContentAutotagging/src/Engine/generic/AutotagBaseEngine.ts:2963-2964  — ContentSource pipeline
meta['Entity'] = isItemLevel ? 'MJ: Content Items' : 'MJ: Content Item Chunks';
if (explicit) return;                              // Entity is already written
```

The write now sits below the early return. With a declaration in place, `'explicit'` omits `Entity` and
writes `ContentSourceID` instead, so the entity name lives in one place rather than on every vector.
Gated on rev 2's three conditions plus the four guards above; anything unmet keeps writing `Entity`
exactly as before. **Existing vectors are untouched** and keep resolving through their stored key, so no
re-index is required.

## Also fixed, both pre-existing on `next`

- `convertMatches` applied the resolved fallback with `??`, so an `Entity: ''` short-circuited the
  Entity-Document fallback and shipped `EntityName: ''`.
- The same `??` on `RecordID`, so a producer writing `RecordID: ''` shipped an empty record id instead of
  falling through to the vector's own id — dropped by the permission filter on an `IN ('')`.

`extractDisplayTitle` **deliberately** reads `meta['Entity']` rather than the resolved name, with a comment
saying so. When the metadata carries name fields under `IsNameField` names, feeding the resolved entity in
makes that branch succeed off the embedding-time snapshot, so the `` `${EntityName} Record` `` sentinel
never forms and `SearchEnricher.resolveRecordNames` (`SearchEnricher.ts:159-161`) never refreshes the title
from the database. Net of `next` the method is unchanged apart from that comment — flagged so nobody hunts
for a behavioural change that isn't there.

## Sequencing

This adds a property to the `ContentSource.Configuration` JSONType, so **`mj sync push` must precede
`mj codegen`** or CodeGen regenerates from the stale definition and silently deletes properties. That has
been run on a from-scratch database, so the regenerated
`packages/MJCoreEntities/src/generated/entity_subclasses.ts` is included and both sides read the real
typed accessors — no casts. The generated diff is **additive only** (0 deletions), which is the useful
part: a property disappearing would have meant the database this was developed against had a field the
repo's migrations don't create.

Both changesets are `minor`, per `.claude/rules/changesets.md` — anything touching `metadata/` is `minor`,
and the rule is evaluated per branch, so Stage A's code-only changeset carries it too.

## Evidence

Built a clean database from migrations + metadata alone (45 migrations, 374 entities, 13,719 metadata
records) and ran everything against it.

| Check | Result |
|---|---|
| `packages/SearchEngine` unit | **476 passed / 31 files** (includes Stage A's 8 legs + the push-down pair) |
| `packages/ContentAutotagging` unit | **221 passed / 8 files** |
| `packages/TestingFramework/integration-test-suite` unit | 176 / 177 — see note below |
| **IT67, real database**, `RUN_MUTATION_TESTS=1` | **8/8 checks** |
| Full deterministic tier | **62/63** (`RUN_MUTATION_TESTS=1`, 218s) — the one failure is fixture volume, not behaviour |
| `tsc --noEmit`, both packages | clean |
| `pnpm-lock.yaml` | untouched |

**Integration coverage added.** CV5 (no declaration → `Entity` kept) and CV7 (declared → `Entity` omitted,
`ContentSourceID` promoted, vector id UUID-equal to the chunk PK) now form a pair, and CV8 covers the two
refusal paths that would otherwise lose data unrecoverably. What CV7 actually produced:

```
CV7: metadata = {"ContentSourceID":"E1AE83C2-…","Name":"mj-cv-…-cv7-declared"}
CV8/cv8-typo:  Entity kept (declaration missing the `MJ: ` prefix, so it resolves to nothing)
CV8/cv8-level: Entity kept (declaration names the ITEM entity while alwaysChunk stores chunk vectors)
```

**The 177th test fails, and it isn't this PR.** `check-registry.test.ts` pins each bundle's check count;
its `task-graph-execution` entry says 7 and that bundle has 11 (TX8–TX11 were added without updating the
table), so the file is **already red on `next`**. This PR updates only its own bundle's count (6 → 8) and
leaves that one alone.

**Full tier: 62/63, including the client-transport lane.** Run against MJAPI serving this same database,
with `MJ_API_KEY` set — the server reads that variable as its superadmin system key and the test client
sends it, so the two only need to agree. The single failure is **IT29 Cache Gauntlet**, and it fails on
fixture *volume* rather than behaviour: `need >= 2 existing rows to make MaxRows:1 meaningful (found 0)`,
`need >= 4 rows for a meaningful offset window (found 3)`. That is the flip side of clean-room isolation —
a from-scratch database has no accumulated rows for it to page through — and it is unrelated to anything
here.

IT67 passed in this run too (2.6s), so the new checks hold inside the full suite and not only standalone.

**Mutation-checked**, since a passing test is not evidence it pins anything: dropping each gate condition
reddens exactly the leg that names it, no-oping the unattributed warning reddens its leg, and making
`extractDisplayTitle` read the resolved entity reddens the sentinel leg. Three legs were rewritten
*because* they passed under mutation — the sentinel test asserted a string both implementations produce
(every mocked entity had `Fields: []`, so the name-field branch could never fire), the ordering test pinned
a truncation branch a real UUID never reaches, and nothing connected the write-side gate to the read-side
validator, which is where an unrecoverable-loss path was hiding.

**Not run:** the live-model tier (paid model calls), and MJServer's tests — it doesn't build on `next`, because
`@types/express-serve-static-core` resolves at both 5.1.1 and 4.19.8 and `RateLimitRequestHandler` from
one doesn't satisfy `RequestHandler` from the other. Unrelated to this PR, but it takes
`ng-explorer-core`, `ng-dashboards`, `ng-bootstrap`, `ng-explorer-modules` and `sql-converter` down with
it — and it silently blocks `server-bootstrap`, so a clean worktree cannot start MJAPI at all until that
package is built by hand.

### Hands-on verification: what was and wasn't done

**Done — the store-level behaviour behind the open question.** A SQL Server 2025 instance was stood up
specifically for this (the colocated provider requires `ProductMajorVersion >= 17`; native `VECTOR` and
`VECTOR_DISTANCE` verified working first), MJ bootstrapped onto it from migrations + metadata, and two
vectors written to a colocated index — one labelled, one not. Result quoted in the decision section above:
the unlabelled vector is excluded by the entity filter and returned without it. That is the open question
answered by observation.

*Incidental finding from that run:* `CREATE VECTOR INDEX … TYPE = 'DiskANN'` failed on this build
(`Unknown object type 'VECTOR' used in a CREATE, DROP, or ALTER statement`) and the provider logged it,
continued, and fell back to exact `VECTOR_DISTANCE`. Correctness path fine, approximate path unavailable —
possibly an artefact of the emulated image. Worth knowing before anyone benchmarks colocated search.

**Not done — and currently not possible: the ContentSource write path end to end.** This is worth stating
precisely rather than as a caveat, because it is a pre-existing limitation rather than a shortcut.

The content pipeline **cannot write to a colocated vector store at all**, for two independent reasons in
`AutotagBaseEngine`:

- `createVectorDBInstance` (`:2810-2818`) requires an API key for every vector DB class key, including
  colocated providers that store vectors in this same database and have no credentials to present.
- It never calls `TryWireColocatedHost`, so `SQLServerVectorDatabase` throws
  `"requires a host connection"`. The EntityDocument pipeline does this in one line
  (`entityVectorSync.ts:923`) and the read side does it in `VectorSearchProvider`; the content pipeline
  does not.

I got as far as building the full fixture chain on a SQL Server 2025 instance — colocated vector database,
384-dimension index matching `gte-small (Local)`, source/file/content types, two content sources differing
only in whether they declare `VectorEntityName`, six content items — and it fails at exactly that point.
Note the failure is confusing rather than loud: `CreateIndex` logs an error and continues, then
vectorization dies later on a vector-database cache miss, so it presents as bad metadata.

That leaves Pinecone as the only store the content pipeline can actually write to, and writing Entity-less
vectors into live content is only undoable by a full re-embed — planned work in its own right, and where
this verification belongs.

**Consequently also not done:** MJExplorer verification, the member-tier cross-org permission check, and
the process-cold `IsPermissionConstrained` check. All three need vectors in a store the deployment
searches, so they inherit the same blocker.

Both gaps are written up separately with a suggested fix (roughly five lines, reordering instance creation
so the key is required only for external stores). Fixing them would make this pipeline locally testable
with no external service — worth doing, deliberately not folded into this PR.

## Known gaps

- **`KnowledgeHubMetadataEngine` is not `@RegisterForStartup`**, unlike its siblings in the same directory
  (`SearchEngineBase`, `PermissionEngine`, `QueryEngine`, `UserInfoEngine`). It loads lazily under whichever
  user searches first, and a permission-denied load empties every collection, reports success, and sets a
  sticky `IsPermissionConstrained` — reset only on a fresh or forced load, and tripped when **any** of its
  six configs is denied. Stage B resolves declarations through this engine, so `ContentSourceID` being on
  the vector does not guarantee the declaration is *reachable*: one low-privilege user warming the cache
  first breaks attribution process-wide, admins included. This PR declines explicitly on that path and
  names the actually-denied entities via `PermissionConstrainedEntities`, but registering the engine for
  startup is the real fix and belongs in its own change.
- **`resolveIndexEntityName`** — the sibling Entity-Document path — has no constrained check, so the guard
  covers one of two paths.
- **The 3rd-party lanes still label hits with the index name**, so they hit the same class of failure. The
  same kind of declaration one level out is the natural fix — separate change.
- **`convertMatches` is ~45 code lines**, over the ~30-40 guidance in `typescript-style.md`. Already at the
  limit on `next` and this grew it; the per-match body is the obvious extraction.

---

## Live testing

Partial, and the remainder is blocked rather than skipped — see **Hands-on verification** above. The
store-level behaviour behind the open question was verified by hand against a real SQL Server 2025
colocated index. The ContentSource write path could not be exercised end to end, because the content
pipeline cannot write to a colocated store at all today; that leaves Pinecone, which is out of scope for a
change whose whole point is what gets written.


AN-BC (approved):
LGTM

AN-BC:
@cmbrock-BC what I was going to ask you was to **include** this in this PR so when we ship you have what you need for Betty. Don't like not including this:

---------
Both gaps are written up separately with a suggested fix (roughly five lines, reordering instance creation
so the key is required only for external stores). Fixing them would make this pipeline locally testable
with no external service — worth doing, deliberately not folded into this PR.

PR #3777, opened 2026-08-13
fix(core): stop cross-server cache events poisoning entity_object engine caches, guard the framework's date reads

### LINKED ISSUE #3558: Agent note/example sorting throws `getTime is not a function` when reads are served from a shared (Redis) cache — breaks realtime session start
## Summary

`AgentContextInjector.sortNotes` / `sortExamples` call `.getTime()` directly on `__mj_CreatedAt`. When the underlying read is served from a **shared cache (Redis)** rather than memory, that field comes back as a **string** — JSON has no `Date` type — so the sort throws:

```
TypeError: a.__mj_CreatedAt?.getTime is not a function
    at agent-context-injector.js:507
    at Array.sort (<anonymous>)
    at AgentContextInjector.sortNotes
    at AgentContextInjector.queryNotesWithScoping
    at AgentContextInjector.GetNotesForContext
```

The optional chain guards `null`, not a wrong *type*, so it does not help.

**Observed on `v5.51.0`.**

## Impact

The throw propagates out of `StartRealtimeClientSession`, so **the realtime session fails to mint** — voice will not start at all, surfacing to the user as an opaque `INTERNAL_SERVER_ERROR`.

It is intermittent in a way that makes it hard to attribute: it only fails once the entry is cache-**warm**, so the first call after any cache flush succeeds and the next one throws.

Two things narrow who sees it:

- **Only the non-semantic path.** `NoteInjectionStrategy: 'Relevant'` routes to `getNotesViaSemanticSearch` and never sorts. `'All'` / `'Recent'` route to `queryNotesWithScoping` → `sortNotes` → throw. Our realtime co-agent uses `'All'`, which is why voice breaks while the async agent (on `'Relevant'`) is unaffected.
- **Only with a shared cache.** Memory-only installs keep real `Date` objects and never hit it, which is likely why it has gone unnoticed.

Multi-worker deployments are more exposed, not less: one worker warms the entry, another deserializes it.

## Reproduction

1. `AllowCaching = 1` on `MJ: AI Agent Notes` (the `__mj` default).
2. Configure `REDIS_URL` so `RedisLocalStorageProvider` backs the cache.
3. Have at least one `MJ: AI Agent Note` in scope for the agent.
4. Set the agent's `NoteInjectionStrategy` to `All`.
5. Start a realtime session twice. The first mint succeeds and populates the cache; the second throws.

Verified isolation: reading `MJ: AI Agent Notes` through a provider **without** Redis returns `__mj_CreatedAt` as a real `Date` on both cold and warm reads (`isDate=true`); only the Redis-backed path returns a string.

## Suggested fix

Coerce before comparing, rather than assuming the provider preserved the type:

```ts
const toMs = (v: unknown): number => {
    if (v instanceof Date) return v.getTime();
    if (typeof v === 'string' || typeof v === 'number') {
        const t = new Date(v).getTime();
        return Number.isNaN(t) ? 0 : t;
    }
    return 0;
};

const dateA = toMs(a.__mj_CreatedAt);
const dateB = toMs(b.__mj_CreatedAt);
return dateB - dateA;
```

`packages/AI/Agents/src/agent-context-injector.ts` has **8** occurrences of this pattern across the note and example sorts.

## Possibly the same latent issue elsewhere

`git grep -l "__mj_CreatedAt?.getTime()"` also matches:

- `packages/AI/Engine/src/AIEngine.ts`
- `packages/Angular/Generic/conversations/src/lib/components/conversation/conversation-chat-area.component.ts`
- `packages/Angular/Generic/conversations/src/lib/components/message/message-list.component.ts`

Those are lower severity (a bad sort rather than a failed session start), but they share the assumption. A broader question worth considering: whether the **cache layer** should rehydrate known date columns from `EntityInfo` on read, which would fix every consumer at once rather than each call site defensively coercing.

## Workaround (confirmed)

Setting `AllowCaching = 0` on `MJ: AI Agent Notes` avoids it — verified: repeated realtime session starts succeed afterwards. That is defensible on its own terms (notes are per-user and written frequently), but it is a workaround, not a fix, and it does not help the other call sites above.

CaelebB-BC:
## Summary

An agent could die mid-run during context assembly with `__mj_CreatedAt?.getTime is not a function` — an ~80ms failure, before any model call. Downstream, the pipeline carried on without the dead stage's output and shipped a component whose spec had no `libraries` array, which surfaced to users as `ReferenceError: antd is not defined`. The same poisoned-cache state also killed **scheduled-job dispatch entirely**: `isJobDue` threw on the first job in the loop, every poll, until the cache reloaded.

There are two defects. Either alone is survivable; together they crash agents and stop the scheduler.

**Fixes #3558** — the same defect surfaced independently in another app six days earlier: agent note/example sorting throwing `getTime is not a function` on Redis-served reads, breaking realtime session start. That report was mitigated app-side by disabling caching on `MJ: AI Agent Notes`; this PR fixes the underlying cause (string-dated rows landing in `entity_object` caches, plus the unguarded sorts), which makes that suppression unnecessary once released.

### Defect B — cache events poisoned `entity_object` caches (the root cause)

`BaseEngine.OnExternalCacheChange` assigned a cross-server cache-change payload's rows **straight into the engine property**. Cache payloads are serialized, so those rows are plain JSON objects. For a config whose effective `ResultType` is `entity_object` — the default — that silently replaced the array's `BaseEntity` instances with plain objects, bypassing the coercing accessors and leaving a `__mj_CreatedAt` declared `Date` holding a raw ISO string. TypeScript cannot see this: the getter still claims `MJAIAgentNoteEntity[]`.

Rows are now materialized through `TransformSimpleObjectToEntityObject` — the same conversion RunView's own cache-hit path uses, so the two cache paths finally agree — before they reach `HandleSingleViewResult`. `'simple'` configs still pass through untouched, and any failure degrades to the full reload the method already implemented — now with a status-level log line (added in review), since the catch also masks materialization failures and a silent degradation would present only as unexplained DB load.

This affects **every engine with `CacheLocal: true`**, not just AI. `BaseAIEngine`'s `_agentNotes` / `_agentExamples` are simply where it surfaced.

### The race that fix introduced, and closed

Materialization is async, and the payload branch was previously **fully synchronous** — so no interleaving was possible. Adding an `await` without an ordering guard would let two overlapping cache events resolve out of order (stale wins), and would let a payload transform clobber a newer full reload that *had* claimed a generation.

The branch now claims a refresh generation via `beginConfigRefresh` / `isLatestConfigRefresh` — the protocol `LoadSingleConfig` already uses around its own awaited RunView. This is **mutation-verified**: with materialization present but the guard removed, the race test fails with `['stale']` instead of `['fresh']`.

### Defect A — unguarded date reads on the poisoned fields (the crash sites)

Optional chaining does not protect these: `"2026-08-12T…"?.getTime` is `undefined`, and calling it throws. Every affected read now routes through a new exported `ToEpochMs()` in **`@memberjunction/global`** — a pure date utility with no entity or metadata concepts, per review feedback — which also closes a latent bug in the old form: an **Invalid `Date`**'s `getTime()` returns `NaN`, and `NaN ?? 0` is `NaN` (`??` is nullish-only) — an incoherent comparator that gives `sort` inconsistent orderings.

| # | Location | Fires when |
|---|---|---|
| 1–2 | `AgentContextInjector.sortExamples` (`'Rated'`, `'Recent'`) | score tie / always |
| 3–4 | `AgentContextInjector.sortNotes` (`'Recent'`, priority) | always / **priority tie — always in practice** |
| 5 | `AIEngine.fallbackGetNotesFromCache` | **always**, whenever the note vector service is cold |
| 6 | `AIEngine.fallbackGetExamplesFromCache` | `SuccessScore` tie |
| 7 | `ConversationEngine.sortConversations` | always |
| 8 | `SchedulingEngine.isJobDue` (`NextRunAt`) | **every dispatch poll** |
| 9–10 | `SchedulingEngine.isJobDue` (`StartAt` / `EndAt`) | activation window set — **silent** (see below) |
| 11 | `SchedulingEngineBase.UpdatePollingInterval` | every poll-interval recalculation |

Site 5 is the widest agent exposure: `FindSimilarAgentNotes` falls back to it whenever the vector service is uninitialized or a query embedding fails — so **semantic retrieval with real input text could crash too**, not just the empty-input path the incident showed. Site 7 is a different engine entirely, found by widening the audit; `ConversationEngine` is also a `BaseEngine`, so its array was in the same exposure class. Sites 8–11 (plus three `toISOString()` diagnostics) are the scheduling engine — a fourth engine in the same class.

### The scheduling engine failed in three distinct ways

1. **Loud** — `job.NextRunAt.getTime()` threw on the *first* job in `DispatchScheduledJobs`, so a poisoned cache stopped **all** scheduled jobs from running, on every poll, until the cache reloaded.
2. **Silent** — `evalTime < job.StartAt` does not throw on a string: relational operators coerce toward numbers, an ISO string yields `NaN`, and every comparison is false — so `StartAt`/`EndAt` activation windows silently stopped being enforced and a job could fire outside its range with nothing in the logs. Those comparisons now go through `ToEpochMs` as well.
3. **Filtering gap** (caught in review by @rkihm-BC, fixed in `61b2fa2d55`) — `SchedulingEngineBase` loads `MJ: Scheduled Jobs` **unfiltered** and applies its Active-only invariant in memory. That re-filter ran after `Config` and on entity events, but not after a cache event — so once the payload path *worked*, a multi-instance deployment could hand one server's dispatch loop another server's full table, and neither `isJobDue` nor lock acquisition checked `Status`: a due Disabled/Paused/Pending job would run. Pre-fix this was masked by failure 1. Fixed in two independent layers: `SchedulingEngineBase` now overrides `OnExternalCacheChange` (`await super…`, then re-filter + `JobsChanged$` emit — the same invariant `HandleIndividualBaseEntityEvent` already maintains, covering both the payload path and the full-reload fallback), and `isJobDue` refuses non-Active jobs outright so dispatch never depends on the array staying pre-filtered.

## Why the priority tie-break mattered

Both seeded `AIAgentNoteType` rows (`Human`, `AI`) carry `Priority 0`, so `sortNotes`' priority comparison **always ties** and always reaches the date comparator. The tie-break was the hot path, not the edge case. Tests seed tied priorities deliberately to mirror that.

## Testing

The original suites were **verified to fail against pre-fix source** rather than assumed:

| Suite | Pre-fix |
|---|---|
| `baseEngine.externalCacheChange` (7) | 5 fail — headline test catches the raw payload object being assigned |
| `agent-context-injector-sorting` (10) | 8 fail with `a.__mj_CreatedAt?.getTime is not a function` — the exact production error |
| `AIEngine` fallback sorts (5 new) | 4 fail, same error |
| `ConversationEngine.sortConversations` (5 new) | 4 fail with `a.__mj_UpdatedAt?.getTime is not a function` |
| `util.toEpochMs` (9) | new helper — pins that it never returns `NaN` |

Scheduling coverage added since: **12 `isJobDue` cases** (loud crash, 1-second tolerance, and two cases that *isolate* the silent window bug — real-`Date` `NextRunAt` so nothing throws, string window that pre-fix returns `true`), **4 non-Active refusal cases** (`Disabled`/`Paused`/`Pending`/`Expired` due jobs), **2 `OnExternalCacheChange` re-filter tests** (Active-only restored + `JobsChanged$` emitted for the jobs config; other configs untouched), and **2 `UpdatePollingInterval` string-date cases** (exact half-interval; mixed Date/string minimum).

**`IT84` (new deterministic integration bundle — renumbered from IT80/IT78 as `next` minted those numbers)** carries the `instanceof Date` claim. That assertion is *not* provable at the unit tier: the coercion lives in `BaseEntity.Get()`'s TSType handling and needs real `EntityInfo` field metadata, which MJCore's duck-typed mocks deliberately don't build — asserting it there would mean mocking the very thing under test. Against a from-scratch database it passes 3/3, and pre-fix NC1 reports the raw payload object while NC2 reproduces the production error end-to-end.

Two implementation notes IT84 encodes: `_agentNotes` lives on `AIEngineBase`, not `AIEngine` (which only delegates the getter), so the event is fired at the Base while assertions read through `AIEngine.Instance.AgentNotes` — the same public getter `AgentContextInjector` uses. And because the bundle poisons a process-wide singleton, every check restores via a forced `Config(true)` in a `finally`, with Teardown forcing one more.

**The wire rig** (`rigs/cache-payload-materialization-tests.ts`, manual, requires `REDIS_URL`, mutating — not part of the PR gate) removes the one assumption every other test shares: that the hand-authored event fixture matches what a real server publishes. It captures a real `Config(true)` publish off a real Redis channel, asserts the rows carry `__mj_CreatedAt` as a **string** on the wire (the defect's precondition, demonstrated in genuine traffic), replays those bytes as a foreign server, and proves the array lands as entities with `Date` fields.

### Results

- **Unit** (latest local run at current head) — MJCore **1947/1947** · AI/Engine **109** · AI/Agents sorting **10** · ConversationEngine **64** · MJGlobal `toEpochMs` **9** · Scheduling engine **260 + 1 skipped** · Scheduling base-engine **14**
- **Integration** — **64/64 (100%)** on a purpose-built from-scratch database (47 migrations, core metadata + the integration-test seed)
- `npm run check:standards` and `npm run check:changeset` pass (all bumps `patch` — no migration, no `metadata/`)

## Review rounds

- **@AN-BC** — LGTM; asked that `ToEpochMs` move out of core → done in `70e04b47de` (`@memberjunction/global`), and requested RK/JF eyes on `OnExternalCacheChange`.
- **@rkihm-BC** — "approve after one scheduling fix": the Active-only filtering gap above → fixed in `61b2fa2d55` exactly as suggested, in two independent layers. The same commit carries three review-hardening items: the fallback log line in `OnExternalCacheChange`, the `UpdatePollingInterval` string-date tests, and the four test mocks that hand-rolled `ToEpochMs` now pulling the real symbol via `importOriginal` so the copies can't drift.

## Design decisions worth reviewing

- **`AdditionalLoading` is deliberately not called** from `OnExternalCacheChange`. `BaseAIEngine`'s override *pushes* into `agent.Notes` without clearing first, so a per-event call would duplicate every linked note. The resulting staleness of derived structures is pre-existing — the method's own full-reload fallback has it too — and is flagged below rather than widened into this PR.
- **The fingerprint is not reconstructed** in IT84. That is `LocalCacheManager`'s concern and has its own tests; what needed proving here is that materialization yields entities whose date getters coerce.
- **Per-event cost, stated honestly:** relative to the previous (incorrect) payload path — which was nearly free — this adds entity construction per event, per config, per `CacheLocal` engine. It remains cheaper than the full-reload fallback it avoids, and the old cheapness was purchased by being wrong.
- **The cache-event re-filter mirrors existing behavior**, including a pre-existing wrinkle: an engine configured with `includeAllJobs: true` already loses inactive rows on any entity event (`HandleIndividualBaseEntityEvent` re-filters unconditionally). The new override matches that rather than redesigning it.

## Not fixed here — flagged for owners

1. **`mj` loads `.env` with override**, so `DB_DATABASE=x pnpm run test:integration` is silently ignored and runs against whatever `.env` names.
2. **`AIAgent.InjectNotes` / `AllowMemoryWrite` default to ON** (`DEFAULT ((1))`), so notes accumulate and cross the ≥2 threshold by traffic alone. Worth an explicit decision.
3. **The cache getters' type contract is unenforceable** — a runtime assertion in `HandleSingleViewResult` under a dev flag would catch the next writer.

(An earlier flag here — `@memberjunction/integration-test-suite` failing to compile on `next` — was fixed upstream and verified compiling after the latest merge, so it has been dropped.)


packages/AI/Agents/src/agent-context-injector.ts
```diff
@@ -1,4 +1,4 @@
-import { LogError, LogStatus, UserInfo } from "@memberjunction/core";
+import { LogError, LogStatus, ToEpochMs, UserInfo } from "@memberjunction/core";
```
AN-BC: The top epoch ms util prob should go into global as it doesn't need db concepts which is where core comes in

packages/MJCore/src/generic/baseEngine.ts
```diff
@@ -2045,16 +2046,28 @@ export abstract class BaseEngine<T> extends BaseSingleton<T> implements IStartup
...
-                        TotalRowCount: parsed.totalRowCount ?? parsed.results.length,
-                        ExecutionTime: 0,
-                        ErrorMessage: '',
-                        UserViewRunID: '',
-                    });
-                    return;
+                    // Claim a refresh generation BEFORE the awaited materialization — the same
+                    // protocol LoadSingleConfig uses around its awaited RunView. Without it, two
+                    // overlapping cache events (or an event racing a full reload) can resolve out
+                    // of order and the stale result would be the one that assigns last.
```
AN-BC, line 2052: Would like @jordanfanapour and @rkihm-BC to put eyes on this pls before merge

AN-BC (commented):
LGTM, would like JF/RK to review as well then you can merge @CaelebB-BC

packages/AI/Agents/src/agent-context-injector.ts
CaelebB-BC: Addressed in 70e04b4 — `ToEpochMs` now lives in `@memberjunction/global` (pure date utility, no DB concepts) and every call site imports it from there.

packages/MJCore/src/generic/baseEngine.ts
```diff
@@ -2045,16 +2046,28 @@ export abstract class BaseEngine<T> extends BaseSingleton<T> implements IStartup
...
+                    const generation = this.beginConfigRefresh(config.PropertyName);
+                    const rows = await this.materializeCacheEventRows(config, parsed.results);
+                    if (!this.isLatestConfigRefresh(config.PropertyName, generation)) {
+                        return; // superseded while materializing — the newer refresh owns the property
+                    }
+                    if (rows) {
+                        this.HandleSingleViewResult(config, {
+                            Success: true,
```
rkihm-BC, line 2060: **[bug]** After a successful materialization, `HandleSingleViewResult` assigns the cache payload wholesale. For `MJ: Scheduled Jobs` that payload is the **unfiltered** table (`SchedulingEngineBase.Config` has `CacheLocal: true` and no `Filter`; Active-only is applied afterwards via `reapplyActiveJobFilter`). That filter is re-applied from `HandleIndividualBaseEntityEvent` (save/delete/remote-invalidate) but **not** from `OnExternalCacheChange`.

`DispatchScheduledJobs` then walks `this.ScheduledJobs` and gates only on `isJobDue`, which still does not check `Status`. `spAcquireScheduledJobLock` also does not check `Status`. In a multi-instance MJAPI with Redis, instance A’s engine load publishes the full job list; instance B’s now-working cache-event path replaces its Active-only array with every job. Disabled/Paused/Pending jobs whose `NextRunAt` is due will acquire a lock and run.

Pre-fix this was masked because `.getTime()` threw on the first poisoned row. The `LoadSingleConfig` fallback already had the same unfiltered assign; this PR unmasks it on the common `Action === 'set'` path as well.

**Suggestion:** Override `OnExternalCacheChange` on `SchedulingEngineBase` to `await super…` then `reapplyActiveJobFilter()` and emit `JobsChanged$` — the same invariant `HandleIndividualBaseEntityEvent` already maintains (covers both the payload path and the full-reload fallback). Independently, making `isJobDue` return false unless `job.Status === 'Active'` is good defense in depth so dispatch cannot rely on the array remaining pre-filtered.

---
*Generated by Grok Build*

rkihm-BC (commented):
## Review (generated by Grok Build)

Approve after one scheduling fix. The cache-poisoning diagnosis and the two-layer fix are correct. AN-BC already LGTM’d and asked RK/JF to look at `OnExternalCacheChange`; that is the right place to look.

### Good

- Root cause is real: `OnExternalCacheChange` assigned Redis JSON rows into `entity_object` engine arrays, so `__mj_CreatedAt` was an ISO string. `?.getTime` does not save you — the optional chain is on the date, not the method.
- Materializing through `TransformSimpleObjectToEntityObject` matches the RunView cache-hit path. `'simple'` configs pass through. Failure falls back to `LoadSingleConfig`.
- The generation guard (`beginConfigRefresh` / `isLatestConfigRefresh`) is the right protocol once the payload path became async. The race test that fails with the guard removed is the proof that matters.
- `ToEpochMs` in `@memberjunction/global` is the right home (AN-BC’s comment; already done in `70e04b4`). It also fixes Invalid `Date` → `NaN` comparators that `?? 0` never caught.
- All seven named crash sites plus the scheduling date reads are consistently wired.
- Skipping `AdditionalLoading` on this path is correct. `BaseAIEngine.AdditionalLoading` still pushes into `agent.Notes` without clearing; calling it per event would duplicate. Injector reads `AIEngine.Instance.AgentNotes`, which this PR actually repairs.
- Tests were verified to fail pre-fix. IT80’s `instanceof Date` claim belongs in integration, not unit mocks. Singleton restore in `finally` + teardown is right.
- Changeset is `patch`. Correct — no migration/metadata.

### Concerning (one real issue)

`SchedulingEngineBase` loads **unfiltered** `MJ: Scheduled Jobs` (`CacheLocal: true`, no `Filter`) and then slices to `Status === 'Active'` in memory. That slice is re-applied in `HandleIndividualBaseEntityEvent`, but **not** after `OnExternalCacheChange`. See the inline comment.

### Missing? Not really

Do **not** need to address in this PR:

- AN-BC’s “put `ToEpochMs` in global” — done.
- Calling `AdditionalLoading` from the cache-event path — would duplicate notes.
- Reconstructing the cache fingerprint in IT80.
- The author’s listed follow-ups (integration-test-suite compile on `next`, `.env` override, `InjectNotes` defaults, a runtime type assert on cache getters). Out of scope and correctly flagged, not left silent.
- More `getTime()` hunting beyond this exposure class.

CI: standards / migrations / deps were green at review time. Unit tests and deterministic SQL Server integration should be green before merge, especially IT80.

---
*Generated by Grok Build*

PR #3780, opened 2026-08-13
feat(graphql): stop emitting child-array FieldResolvers

AN-BC:
## Summary
- CodeGen no longer emits `Foo_BarIDArray` `@Field` / `@FieldResolver` pairs (794 N+1 `SELECT *` resolvers).
- `CurrentUser` now returns a first-class `Roles` field on `CurrentUserType`, so magic-link anonymous sessions keep synthesized roles that are not in `MJ: User Roles`.
- `CreateQueryExtended` / `UpdateQueryExtended` return `Fields`, `Parameters`, `Entities`, and `Permissions` on the mutation result instead of nesting them under `MJQuery_`.
- Generated GraphQL types shrink by ~10k lines. Load children with `RunView` or `DeclareRelatedRecords`.

## Breaking change
Any custom GraphQL document that selected a generated `*Array` child field will fail schema validation. In-tree callers (CurrentUser, system-user query mutations) are migrated. Explorer never used these fields.

**CurrentUser `{ MJUserRoles_UserIDArray }` → `{ Roles }` is an intentional break.** This is the first 6.x LTS cut — we are not adding a deprecated alias and are not bumping the changeset. We are already on 6.1; LTS has not shipped. Mixed-version client/server is not a supported scenario for this field.

## Review follow-up (`256b337031`)
Addressed the review on #2–#8:
- Role load uses `GetUserFromPayload` as `contextUser` and throws if it is missing (#2).
- A failed `MJ: User Roles` RunView now `LogError`s and throws instead of returning `[]` (#3).
- Email-fallback `CurrentUser` throws when no user is found instead of spreading `null` (#4).
- Dead `GeneratedTypeAvailability` / `isRelatedTypeOutOfScope` plumbing removed; availability tests collapsed to two policy locks (#5, #6).
- Request provider + contextUser passed into `ArrayMapFieldNamesToCodeNames`; transport `Roles` stripped before `new UserInfo`; `UserID` ExtraFilter is quote-escaped (#7).

## Test plan
- [x] CodeGen unit tests — 746 passed, 60 skipped
- [x] `@memberjunction/server` unit tests — 848 passed, 56 skipped (includes new CurrentUser guard tests)
- [x] `@memberjunction/graphql-dataprovider` unit tests — 319 passed
- [x] `pnpm run build` for CodeGenLib, MJServer, GraphQLDataProvider
- [x] `npm run check:changeset` — patch (no migration/metadata)
- [ ] Deterministic integration suite — attempted against private DB `MJ_6_1_0_DROP_GQL_ARRAYS`; this worktree’s CLI stack is missing `testing-cli` dist so `mj test suite` could not launch. Worth a run from a fully built tree before merge.
- [ ] After merge: first `mj codegen` regenerates `generated.ts` without the arrays (this PR already strips the current file to match)

## Isolation
Work was done in a worktree on `drop-graphql-relation-arrays` (from `origin/next`). Private SQL database `MJ_6_1_0_DROP_GQL_ARRAYS` was created for this branch and was not used to mutate the local `.env` database.


AN-BC (commented):
LGTM. Would like @cadam11 @rkihm-BC @jordanfanapour to take a quick look. 

Also @claude do a code review and comment any views on this.

AN-BC:
## Review

The core change is right. 794 `@FieldResolver`s that each issued an unbatched per-parent `SELECT *` is a real N+1 surface, and deleting ~10k lines of generated schema for something nothing in-tree consumed is a clear win. I verified the removal is safe for the paths that *look* like they'd depend on it:

- No in-tree consumer of `*Array` remains — a repo-wide grep for `IDArray` outside `**/generated/**` returns only this PR's own comments/tests and unrelated `entityIDArray` locals in `MJCodeGenAPI`/`MJEntityPermissionEntityServer`.
- `DeclareRelatedRecords` is unaffected — it resolves through `relatedRecordBatchLoader` / `RunView`, never through GraphQL child arrays.
- The hand-stripped `generated.ts` is consistent with what the new generator would emit: zero leftover `Relationship to …` / `Relationship field to …` / `// Link to` comments, zero `mj_core_schema_server_object_types` references, and 374 ObjectTypes against 374 resolver classes.
- `buildSuccessResult` switching to `MapFieldNamesToCodeNames` is a quiet bug fix — the old `{...entity.GetAll()} as MJQuery_` returned raw `__mj_CreatedAt` for a type that declares `_mj__CreatedAt`.

Findings below, ordered by how much they'd cost if missed. (Posting as one comment — GitHub's review API is returning 502 for me right now, so the inline anchors didn't take.)

---

### 1. The `CurrentUser` schema break has no transition window, and it lands as a `patch`

`CurrentUser { MJUserRoles_UserIDArray }` → `CurrentUser { Roles }` changes the **login bootstrap query** (`graphQLDataProvider.ts:2829`). This isn't a feature that degrades — a client on an older `@memberjunction/graphql-dataprovider` hitting an upgraded MJAPI fails GraphQL *validation* on its very first query, so login fails outright with nothing to fall back to. Same in the other direction.

Inside this repo that's fine — `fixed` versioning moves client and server together. The exposure is downstream MJ apps that pin `@memberjunction/graphql-dataprovider` independently of the MJAPI they talk to, which now break on a patch bump.

Cheap to make non-breaking for one version — on `CurrentUserType` (`UserResolver.ts:15`):

```ts
@Field(() => [MJUserRole_], { deprecationReason: 'Use Roles instead. Removed in the next major.' })
MJUserRoles_UserIDArray: MJUserRole_[];
```

populated from the same array in both `CurrentUser` return sites. Old clients keep working, new clients use `Roles`, and the field disappears on a major. If you'd rather take the break now, that's a legitimate call — but it's worth making deliberately rather than as a side effect, and the changeset should say so louder than "CurrentUser now returns a first-class Roles field."

### 2. `loadRolesForUser` passes `undefined` as the RunView `contextUser` on the only path that reaches it

`packages/MJServer/src/resolvers/UserResolver.ts:104-125`

`loadRolesForUser` has exactly one caller — line 64, in the `else` branch of `CurrentUser`. That branch runs *only* when `context.userPayload.userRecord` is falsy. So `const sessionUser = context.userPayload?.userRecord` (line 108) is `undefined` by construction, the magic-link short-circuit on line 109 can never fire from here, and line 119 hands `undefined` to `RunView`.

`RunView.RunView`'s own doc says contextUser "is generally required" server-side (`packages/MJCore/src/views/runView.ts:613`), and `ProviderBase` marks it "required server-side" in four places. So this either throws or returns `Success: false`.

Suggest deriving it the way the rest of `ResolverBase` does, and passing it to both call sites:

```ts
const contextUser = this.GetUserFromPayload(context.userPayload);
```

Latent rather than live today — every successful auth path in `context.ts` (API key :442, system key :467, JWT :557) sets `userRecord`, so the `else` branch is effectively unreachable. But it's the only branch this method exists to serve, which means as written the method is dead-or-broken rather than working.

### 3. RunView failure is swallowed into "this user has no roles"

`packages/MJServer/src/resolvers/UserResolver.ts:121-123`

```ts
if (!result.Success || !result.Results) {
  return [];
}
```

On the login path, `[]` is not a safe default. It flows into `new UserInfo(this, {...u, UserRoles: roles})` (`graphQLDataProvider.ts:518`), so a transient DB error or a permission failure on `MJ: User Roles` presents to the user as a successful login with zero roles — every role-gated surface in Explorer silently disappears, and nothing in the response says why. The old generated resolver's `CheckUserReadPermissions` threw, which is the correct shape for an auth-adjacent read. Please `LogError(result.ErrorMessage)` and throw rather than returning an empty array.

### 4. Null user in the fallback branch yields a malformed object instead of `null`

`packages/MJServer/src/resolvers/UserResolver.ts:62-68`

`UserByEmail` ends in `MapFieldNamesToCodeNames`, which returns `null` for an empty result (`ResolverBase.ts:83-85`) — e.g. an unknown email. `{ ...(result as MJUser_), Roles: roles }` then spreads `null` into `{ Roles: [] }`, and GraphQL rejects it as "Cannot return null for non-nullable field CurrentUserType.ID" — a confusing error that points at the wrong thing. Previously the resolver returned `null` directly. Worth an explicit guard before the spread.

### 5. Dead code left behind in CodeGenLib

`GeneratedTypeAvailability`, `isRelatedTypeOutOfScope`, and the `availability` value threaded through `generateServerEntityString` / `generateEntitySpecificServerFileHeader` / `generateServerGraphQLResolver` now have **zero** production consumers — the relationship generators were the only ones. Confirmed by grep: `isRelatedTypeOutOfScope` appears only at its own definition (`graphql_server_codegen.ts:128`) and in a test file's prose comment.

What's left behind:
- `generateGraphQLServerCode:56` still builds a 374-entry `Set` on every run for nothing.
- The 30-line docblock at `graphql_server_codegen.ts:100-127` describes a gate that no longer executes, including the whole linked-Open-App rationale. That will actively mislead the next reader trying to understand why relationship emission is scoped — the answer is now "it isn't emitted at all."
- `getExternalEntitySchemas` is imported solely for that dead method's legacy branch.

Since `isRelatedTypeOutOfScope` is `protected`, keeping it for subclass compat is defensible — but then say that in one line and drop the rationale that no longer applies. The `availability` plumbing through the three `_`-prefixed parameters is the part I'd just delete.

### 6. The availability tests now assert tautologies

`graphql-type-availability.test.ts` is six tests that all call the same `expectNoChildArray` helper against fixtures differing in availability set, schema, and core-vs-non-core — none of which the generator observes any more. They pass identically for the same reason, so five of them carry no independent signal. Same for `graphql-external-schema-filter.test.ts`: the MJ#3279 dangling-`@Field` bug is now structurally impossible, so the file's premise is stale even though the assertions are correct. Two policy-lock tests ("no `@FieldResolver`", "no `*Array` member") would cover this; the rest is upkeep cost on tests that can't fail independently.

### 7. Smaller things

- **`ArrayMapFieldNamesToCodeNames` doesn't get the request provider.** `UserResolver.ts:124` has `provider` in hand from line 113 but passes only `(entityName, rows, sessionUser)`, so `MapFieldNamesToCodeNames` falls back to `new Metadata()` — the global provider. `.claude/rules/data-access.md` calls out per-provider `Metadata` scoping specifically. The 4th parameter exists for this.
- **`mapRolesFromUserInfo` drops contextUser entirely** (`UserResolver.ts:138`) where the old code passed `sessionUser`. Harmless while `MJ: User Roles` has no encrypted fields; if that changes it becomes a silent `ENCRYPTED_SENTINEL` path rather than an error.
- **Client leaves the unconverted payload attached.** `graphQLDataProvider.ts:517-519` reads `u.Roles`, maps a converted copy into `roles`, but never removes `u.Roles` — so `new UserInfo(this, {...u, UserRoles: roles})` carries a stale `Roles` property still in `_mj__CreatedAt` form alongside the correct `UserRoles`. The old code overwrote in place, so this is new. Destructure it out.
- **`ExtraFilter: UserID='${user.ID}'`** (`UserResolver.ts:117`) — the value is a server-side UUID so it isn't an injection vector in practice, but it's a raw interpolation into a filter on a login path.

### 8. Definition of Done — integration tier

The test plan lists unit tests only. `CLAUDE.md` requires `pnpm run test:integration` (the deterministic tier) to pass, and this PR changes the GraphQL schema plus the login bootstrap path — precisely the cross-package seam that tier exists to catch and that the mocked CodeGen unit tests can't see. Worth running before merge.

---

Happy to be wrong on #1 if downstream version skew isn't a scenario you support — that one's a judgment call about the release model, not a defect. #2/#3 I'd want addressed regardless, since together they mean the DB role-loading path has never actually executed successfully as written.

---

AN-BC:
@claude thanks for the review — agreed on the core call and on #2–#8. Fixes are in 256b337031.

**#1 CurrentUser schema break** — taking the break on purpose. This is the first 6.x LTS cut; we are already on 6.1 but LTS has not shipped, so mixed-version client/server is not a scenario we are supporting here. No deprecated `MJUserRoles_UserIDArray` alias, and no changeset bump for it. The PR body now says that out loud.

**#2 / #3 / #4** — the fallback path now uses `GetUserFromPayload` as `contextUser`, throws if that user (or a provider) is missing, throws (with `LogError`) if the role `RunView` fails, and throws if `UserByEmail` returns nothing instead of spreading `null` into `{ Roles: [] }`. Guards live in `currentUserRoles.ts` so they can be unit-tested without booting generated GraphQL / Apollo config.

**#5 / #6** — `GeneratedTypeAvailability`, `isRelatedTypeOutOfScope`, and the unused `availability` plumbing are gone. The availability tests are two policy locks (no `@FieldResolver`, no `*Array` member). The MJ#3279 file keeps the “local entity still generates, no dangling child array” checks.

**#7** — request `provider` + `contextUser` go into `ArrayMapFieldNamesToCodeNames` / `mapRolesFromUserInfo`; transport `Roles` is destructured off before `new UserInfo`; `UserID` ExtraFilter is quote-escaped.

**#8** — unit tests + package builds are green (CodeGenLib 746/60 skip, MJServer 848/56 skip, GraphQLDataProvider 319). I attempted the deterministic integration suite against the private `MJ_6_1_0_DROP_GQL_ARRAYS` DB; this worktree’s CLI stack does not have `testing-cli` dist, so `mj test suite` could not launch here. Still worth a run from a fully built tree before merge.

Happy path login was never the broken path (roles still come from `userRecord.UserRoles`). The fallback is what is now fail-closed instead of dead-or-empty.

PR #3781, opened 2026-08-13
feat(cli): mj dev workspace — generate the multi-repo dev workspace

cadam11:
Automates the manual setup of the Open App multi-repo dev workspace. `mj dev workspace` generates the four parent files (`pnpm-workspace.yaml`, `.npmrc`, `package.json`, `turbo.json`) plus a `.mj-dev-workspace.json` sentinel manifest, detects sibling member repos, runs the install; `mj dev workspace status` reports workspace state; `mj dev workspace clean` tears it all down again; `mj dev usage` teaches the domain to agents. All commands are bootstrap-free (no DB), and `--dir` binds to `MJ_DEV_WORKSPACE_DIR` (flag > env > default).

## What it does

- **Generate**: workspace yaml with producer-only globs (`<repo>` + `<repo>/packages/*`), the three-line `.npmrc`, a private pinned root `package.json` (devDependency union with highest-version-wins and every conflict logged, plus the peer-bridge block), and `turbo.json` copied from the first member that carries one.
- **Detect members**: sibling dirs with `mj-app.json`, packages under the `@mj-biz-apps/*` scope, or the MJ monorepo; `--include`/`--exclude` to adjust, non-interactive by default.
- **Install**: `pnpm install` at the parent unless `--no-install`.
- **Safety**: refuses to overwrite existing parent files without `--force` (which writes `.bak` backups first) and refuses a parent directory that is itself a git repo root.
- **No hoist block, by evidence**: the 78-entry `public-hoist-pattern` set the earlier manual setup carried was audited entry-by-entry against the monorepo and found entirely obsolete (details in the PR comment below) — the generated `.npmrc` is just the three settings lines, and the command prints guidance for the 8 shell-choice auth/service-worker peers instead.
- **Clean**: `mj dev workspace clean` removes exactly the residue (the 4 files, `pnpm-lock.yaml`, `node_modules/`, sentinel). Generation writes a `.mj-dev-workspace.json` sentinel; clean requires it unless `--force` (protects hand-made dirs), `--dry-run` previews, every deletion is logged, `.bak` files survive, `node_modules` removal never follows symlinks out of the tree, and the sentinel is deleted last so an interrupted clean stays resumable.
- **Agent-facing usage**: `dev` now appears in the tier-1 `mj usage` domain map, and `mj dev usage` renders full progressive-disclosure docs (parent-dir rules, the files, detection, clean semantics, the env var) — with tests asserting each operational fact is present in the rendered text.
- **`MJ_DEV_WORKSPACE_DIR`**: all three commands bind `--dir` to it (flag > env > default); `status` reports which source resolved the directory.

Registration of a linked app into a running host is deliberately **out of scope** — that lands as a follow-up.

## For reviewers — judgment calls to check

1. **devDep union semantics**: member repos can pin conflicting devDependency versions, and a literal union would collide. The builder seeds a `turbo` + `tsc-alias` baseline, merges every member's devDependencies on top with highest-base-version-wins, and logs each conflict it resolves. Flag if you'd rather it emit only the baseline.
2. **No `packageExtensions` either, on purpose**: `packageExtensions` patches a resolved package's manifest (the "react-redux forgot react-dom" shape) — the audit found zero such gaps, and it can't add dependencies to a workspace project (the shell), so emitting the machinery would ship an empty mechanism. If one ever becomes needed, the future-proof home is a top-level `packageExtensions:` key in `pnpm-workspace.yaml` (pnpm v11 stops reading `pnpm.*` from package.json).
3. **Version comparison** is a small built-in base-triple comparator, no `semver` dependency; non-numeric specs (`workspace:*`, `latest`) tie and keep the incumbent, still logged.
4. `light-commands.ts` registers the dev commands as light (node stdlib + chalk only, no MJ bootstrap).
5. **One cross-package change — CLICore gains `CLIPluginRegistry.RegisterUsage()`** (~35 lines + 4 tests): the usage registry only rendered plugin-backed commands, so plain oclif commands (deliberately bootstrap-free, like this domain — and like `migrate`/`doctor`/`install` today) were invisible to `mj usage`. The additive seam lets in-CLI commands declare usage without a plugin conversion; plugin-declared usage still wins on key collision. The alternative (converting the dev commands to `BaseCLIPlugin`) changes their stdio contract and exit codes, and was rejected as out of scope. This is the one change reviewers should look at outside `dev-workspace/`.

## Testing

- `packages/MJCLI` build clean; full suite **554 passed / 0 failed / 0 skipped** (33 files, ~120 new tests across builders, detection, write safety, clean's refusal ladder, spawn-mocked install, status, usage-content assertions, env-var precedence at the oclif parse level). `packages/CLICore` build clean, **21 passed / 0 failed**.
- Fixture smoke runs verified by hand: generation, refusal-without-`--force`, `--force` backups, `--exclude`, `status`, git-repo-root refusal, clean dry-run byte-for-byte no-op, real clean with a member-repo symlink inside `node_modules` (member repo untouched), and a real `pnpm install` (strict peers on) against a two-repo fixture with cross-repo symlinks verified.
- No new dependencies; lockfile untouched. Integration tier not run (CLI-only change, no DB surface).
- Known gap: ESLint could not run in the nested build environment (duplicate prettier-plugin resolution); `tsc` strict is clean and no `any` appears in new code — flag anything the repo lint would catch.

No unrelated changes beyond the CLICore seam called out above.

Claude-Session: https://claude.ai/code/session_01MYWbAFjiCkR8LWJK3VXWX3


cadam11:
## Why there is no hoist block — the audit, for the record

The manual workspace setup this command replaces carried a 78-entry `public-hoist-pattern[]` block for the parent `.npmrc`, written in the npm-hoisted era to keep an in-workspace Angular dev server resolving. Before shipping the generator we attributed every entry: a script parsed every import/export/require/dynamic-import/SCSS-import/angular.json reference across 7,670 source files in 309 packages, cross-checked each importing package's declarations, then scanned 3,226 installed manifests for third-party peer gaps.

**Result: 0 entries still need hoisting.**

| Class | Count | Meaning |
|---|---|---|
| A — MJ package imports it undeclared | **0** | four apparent hits, all false positives (lint-rule example strings, a generated-script string, one uncompiled stray prototype) |
| B — third-party manifest needs patching | **0** | every peer relationship in the set is satisfied by a real MJ declaration |
| C — shell's own runtime choice | **8** | the auth-provider family + `@angular/service-worker`, correctly `peerDependencies` of `ng-auth-services` / `ng-explorer-service-worker` — the shell declares the provider(s) it serves |
| D — obsolete | **70** | 55 already declared by every importer; 15 referenced by nothing in the repo at all |

The mechanism is simple: hoisting only ever papered over undeclared imports, pnpm's strict layout turns those into hard failures, so the pnpm conversion (#3466) already forced every real fix. The block outlived its cause.

So the generated `.npmrc` is three settings lines with no hoist patterns, and the command prints the class-C guidance (derived from the same data structure as the tests, so they can't drift). A regression test asserts no hoist pattern is ever emitted and that the class-C set never absorbs packages whose importers declare them.

Two small unrelated findings from the audit, deliberately not fixed here (each is a one-line change better done separately): `@memberjunction/action-runtime` (`harness/run-demos.ts`) and `@memberjunction/actions` (`scripts/entity-action-workflow-integration.ts`) import `dotenv` without declaring it — dev scripts only, they work by coincidence of layout. Also stale: `ng-base-forms` declares a `primeng` peer nothing imports, and `packages/Angular/Explorer/dashboards/src/AI/prototype.js` is an uncompiled React file in a published Angular package.

PR #3786, opened 2026-08-13
Sanitize all three caller-supplied persists; differential failure oracle; OR-join test; D1 note

AN-BC:
Three carried-forward items plus a live bug found while fixing this morning's crash.

## 1. One rule for what may be written down (the live bug)

Three writes take an object the **caller** built and stringify it straight into a database row:

| Write | Guarded before? |
|---|---|
| task graph's invocation envelope | yes — after it killed agent runs at submit time this morning |
| agent run's `StartingPayload` | **no** |
| agent run's `Data` | **no** |

All three take it from `ExecuteAgentParams`, whose own documentation says the value may be a class instance holding *"external service credentials or connection information"*. The combination fails two ways, and the second is worse:

- **It throws** — `Converting circular structure to JSON` on anything holding a socket, pool or provider. That's what you hit this morning.
- **It leaks** — when serialization happens to *succeed*, whatever the object held is written to a row that outlives the run.

The first is loud and gets fixed. The second is silent and doesn't.

`SanitizeForPersistence` / `StringifyForPersistence` now hold the rules once, and all three writes use them. JSON data survives (primitives, arrays, plain objects, `Date` as ISO, anything with `toJSON`) — deliberately the same set `JSON.stringify` keeps, minus the ways it explodes. Class instances are refused **whole** rather than unwrapped: walking a socket's own properties would "work" and would persist its internals, which is the leak, not the fix. Every drop is reported by path and logged, because a silently dropped value is the failure one layer down.

## 2. The differential suite's first failure oracle

It couldn't express `'edges'` — the dialect where a flow routes around its own failure — because the simulator treated *which origin statuses may decide an exclusive group* as a constant, where the dispatcher treats it as the graph's dialect (`'edges'` → `{Complete, Failed}`). With that modelled, a fixture whose step fails and whose recovery path runs is now compared walker-against-compiled.

The comment states why `'block'` gets **no** oracle by construction rather than by omission: the walker routes on failure, so there is nothing to compare a compiled `'block'` graph *against*. IT74's TX19 pins that side against the engine instead.

## 3. OR-join — already resolved in doc, missing its test

The doc/impl mismatch was fixed earlier; what was missing was an assertion. The all-`Optional` node's **wave-1** behaviour — the consequence the spec calls surprising and nothing asserted — is now pinned, so an OR-join implemented later reads as a deliberate semantic change rather than an oversight being corrected.

## 4. D1 execution identity — written up as a decision

Which `UserInfo` a durable graph runs as, promoted from a sentence in three plans to [`plans/task-graph-execution-identity.md`](plans/task-graph-execution-identity.md): why it matters more now that cost rollup, the debug verbs and the invocation envelope all landed on top of the gap; three candidate models; and a recommendation — make today's behaviour explicit and *gate* it, grow into revalidation — with what each would require. No code.

## Verification

`ai-core-plus` **655** passed, `ai-agents` **1872** passed, changeset gate green.


AN-BC (commented):
## Review

Parts 2–4 are ready as-is. Part 1 — the sanitizer, which is the reason this PR exists — has two defects I'd want fixed before merge, plus one judgment call.

I ported `SanitizeForPersistence` to plain JS and ran adversarial inputs through it rather than reading for bugs, so each finding below has a runnable repro.

---

## 🔴 1. `toJSON` punches a hole in "class instances are refused whole"

`safe-persist.ts:86-89` runs the `toJSON` check **before** the prototype check at `:102-105`. So the module's central promise —

> class instances are refused **whole** rather than unwrapped: walking a socket's own properties would "work" and would persist its internals, which is the leak, not the fix

— holds only for classes that don't define `toJSON`. Any class that does gets unwrapped by its own method:

```js
class SkipCtx {
  constructor() { this.apiKey = 'sk-SECRET'; this.endpoint = 'https://internal'; }
  toJSON() { return { ...this }; }        // a very common idiom
}
StringifyForPersistence({ context: new SkipCtx() }, 'data')
// -> {"context":{"apiKey":"sk-SECRET","endpoint":"https://internal"}}
//    DroppedPaths: []
```

The same class without `toJSON` is correctly refused (`{}`, `dropped=['data.context']`).

Two things make this worse than a documentation gap. First, the threat model is specifically *"the caller built this object and may not have thought about persistence"* — so deferring the security decision to that object's own method hands the guarantee to the party the module doesn't trust. `SkipAgentContext` is the named example in the header, and it's precisely the kind of object that might carry a convenience `toJSON`. Second, it's **silent**: `DroppedPaths` stays empty, so `warnAboutDroppedValues` prints nothing and nobody learns it happened.

The existing test only covers the benign shape — `Money.toJSON()` returning a scalar (`safe-persist.test.ts:56`). The `{...this}` case isn't tested.

**Suggested fix:** move the prototype check above the `toJSON` check, so class instances are refused regardless of what methods they carry. That's cheap here because `Date` is already special-cased at `:85`, which covers the common legitimate `toJSON`. If you'd rather keep `toJSON` deference for genuine data classes, then the header should stop stating an unconditional guarantee, and an unwrapped class instance needs to land in `DroppedPaths` as a reported transformation rather than passing invisibly. Either way, please add the `{...this}` case as a test.

## 🔴 2. The sanitizer can still throw — the crash it exists to prevent

Neither the `toJSON()` invocation (`:89`) nor `Object.entries()` (`:109`, which invokes getters) is guarded:

```js
StringifyForPersistence({ ctx: { toJSON() { throw new Error('cannot serialize live handle'); } } }, 'data')
// !! THREW: Error: cannot serialize live handle

StringifyForPersistence({ ctx: { get creds() { throw new Error('connection closed'); } } }, 'data')
// !! THREW: Error: connection closed
```

That's the same class of failure as `Converting circular structure to JSON` — and it's now reachable from **three** write paths instead of the one that broke this morning, so the blast radius is larger than before the fix.

`warnAboutDroppedValues` is careful to "warn, never throw", but the throw happens upstream inside `StringifyForPersistence`, so that care never gets a chance to apply. A getter on a live handle that throws once the connection is closed is exactly the kind of object this module is aimed at.

**Suggested fix:** wrap both call sites in `try/catch` and treat a throw as a drop (push the path, return `undefined`). Worth noting a side effect that falls out of the same code: getters **are** invoked during sanitization (confirmed — 1 call per property), so sanitizing a context object holding lazy connections can open one.

## 🟡 3. `MAX_NODES` bounds container count, not row size

`:50` says the cap exists *"so a row cannot become a memory dump"*, but `++nodes` at `:79` only fires for objects — scalars are never counted:

| Input | Result |
|---|---|
| 100k nested object props | capped at 1999, 98001 dropped ✅ |
| 100k **flat scalar** props | all 100k kept — 1.29 MB of JSON |
| one 10 MB string | kept whole — 10 MB of JSON |

If the intent is bounding what actually lands in the row, this needs a byte budget alongside the node cap. Flagging as a judgment call rather than a defect — the cap does work as written for nested shapes.

---

## What's working

Verified rather than assumed:

- **Path-scoped cycle detection is the right call and it works.** A child referenced by two siblings survives as data (`{"a":{...},"b":{...}}`, no drops) instead of being falsely flagged as a cycle. Easy to get wrong in the direction that silently eats real data.
- `toJSON(){ return this }` terminates safely via the depth cap. Array holes become `null` rather than shifting indices. `NaN`/`Infinity` are dropped rather than serializing to a confident `null`. `Map`/`Set` refused. The original circular-object crash is fixed.
- The task-graph extraction is a clean, behaviour-preserving refactor — the old inline walker is now a thin wrapper that adds only the empty-envelope rule.
- **Part 2:** the `'block'`-has-no-oracle-*by construction* note is the strongest thing in this PR. Explaining why a comparison is impossible, rather than leaving it merely absent, is what stops someone "fixing" it into existence later.
- **Part 3:** the wave-1 OR-join pin is the right kind of test — it makes a future OR-join read as a deliberate semantic change rather than a correction.
- **Part 4:** a plan doc with no code attached, honestly labelled as such.
- The check-count guard moved 26→27 with TX27 named explicitly, matching that file's own convention.

CI is green apart from unit tests still in flight, including the deterministic integration tier.

Net: items 2–4 ship. #1 and #2 above are the two I'd fix first — #1 because it silently defeats the guarantee the module is built around, #2 because it reintroduces this morning's failure mode across a wider surface.

---

AN-BC:
All three reproduce against the built package, all three are fixed, and one of them was worse than you wrote.

## 1. `toJSON` hole — agreed, and the test was complicit

You were generous about the test. **My `Money` test was this bug's accomplice**: it asserted `toJSON` deference using the one shape where deference is harmless — a value object returning a scalar — which locked the hole in as "tested". The same assertion existed in the *envelope* suite too, so the hole was encoded twice.

Prototype check now runs first; `Date` stays special-cased above it as the legitimate deference that actually occurs. Everything else is refused and reported:

```
before: {"context":{"apiKey":"sk-SECRET","endpoint":"https://internal"}}  dropped: []
after:  {"tier":"gold"}                                                  dropped: ["data.context"]
```

Both tests rewritten to assert the refusal, plus your `{...this}` case, plus one recording the cost — a legitimate `Money` no longer persists as `12.5`, and the path says so.

## 2. Sanitizer could throw — agreed, and your fix wasn't quite enough

Wrapping both call sites fixed the throw but left a subtler bug I only caught by re-running your repro with a **sibling** property: `Object.entries` *reads every value*, so one throwing getter took the whole object down and reported the parent's path, losing data that serialized perfectly well. My own comment ("drops its own property, not the object around it") was false as written.

`Object.keys` is inert, so each value is read inside its own guard:

```
{ ctx: { get creds() { throw }, ok: 1 } }
-> {"ctx":{"ok":1}}  dropped: ["data.ctx.creds"]
```

Your side note about getters being invoked is now in the header — sanitizing can open a lazily-opened resource, and refusing class instances removes most of that surface since their getters are never reached.

## 3. `MAX_NODES` — I went further than "judgment call"

The comment claimed a row *cannot become a memory dump*, and that claim was false: 9.54 MB sailed through. Either the comment stops claiming it or the code makes it true, and a 10 MB string in an agent-run row is a real operational problem. Strings are where the bytes are, so that's where the budget is (256k chars). Dropped whole rather than truncated — half a value is one nobody can trust.

## One you didn't raise, which I think outranks #3

For `Data`/`StartingPayload` the old behaviour was plain `JSON.stringify`, which **does** serialize a class instance's own enumerable properties. Refusing them whole is therefore a silent fidelity regression for any agent whose payload legitimately is a class instance — its run record loses state it used to keep. Tightening #1 widens that.

Amith's call was to drop loudly with the path named rather than persist whatever the object happened to hold, which I agree with. It's now stated in the changeset as a deliberate behaviour change, so nobody discovers it from a missing payload.

`ai-core-plus` 660/660, `ai-agents` 1872/1872.

---
_Generated by [Claude Code](https://claude.com/claude-code)_

PR #2837, opened 2026-06-13
Plan: MemberJunction Developer Experience (DX) Audit & Improvements

AN-BC:
## Plan: MemberJunction DX Audit & Improvement Plan

This PR contains an **implementation plan only** (no code changes) for review before work begins. It captures a full audit of MJ's developer experience across the monorepo and the companion VSCode extension, with a prioritized, sized (S/M/L) backlog.

### Summary
- **Onboarding** is the weakest link: three overlapping config systems (`install.config.json` + `mj.config.cjs` + `.env`), no prereq gate, no DB-bootstrap guidance, and no "first change" tutorial. Proposes a unified `mj setup` + `GETTING_STARTED.md`.
- **Inner build loop** has avoidable throttles: cold `start:explorer` fails on unbuilt workspace `dist/`, 9 class-manifest tasks run serially (~50s), and shared-lib edits don't cascade-rebuild. Proposes prestart pre-build, Turbo-parallel manifests, and a server-leakage guardrail.
- **Invisible/unenforced quality**: ~90% of packages have tests but CI never runs them; 160+ TypeDoc configs are never published. Proposes a CI test gate + TypeDoc publishing.
- **CodeGen/CLI**: heavy migration-authoring ceremony with no scaffolding; all-or-nothing CodeGen. Proposes `mj migrate scaffold`, `mj codegen --skip-*`, standardized `--ci`, and a static-import lint hook.
- **VSCode extension**: broad but unpublished/untested and missing the core mj-sync loop. Proposes sync commands, an agent-run inspector, and a Marketplace publish.
- **Two immediate-triage items**: a possible committed `.env` secret (unconfirmed — must verify) and a confirmed broken VSCode AI code action wired to undeclared commands.

### How to review
1. Read `plans/dx-audit-and-improvements.md` for the full plan.
2. Comment on prioritization, sizing, or any design decision you disagree with.
3. Approve when ready for implementation to begin (intended to split into multiple focused PRs).

### Labels
This is a **plan-only PR** — no code changes are included.

Generated with [Claude Code](https://claude.com/claude-code)

PR #3804, opened 2026-08-13
feat(codegen): schema-scale emit and BigSchemaDemo test bed

AN-BC:
## Why

MemberJunction CodeGen used to emit **one file per output kind** — a single `entity_subclasses.ts`, a single GraphQL `generated.ts`. That is simple and correct while the catalog is small. As soon as the database is **MJ Core plus application schemas** (or an existing multi-schema database), one dirty table rewrites a file that already contains every entity. Incremental `tsc` keys off mtimes, so the cache is thrown away. The compile unit becomes the whole catalog.

This PR makes **schema** the incremental unit. The catalog stays complete. Files, dirty regen, parallelism, output routing, and agent context all key off schema.

Guide: [`guides/CODEGEN_LARGE_SCHEMA_GUIDE.md`](https://github.com/MemberJunction/MJ/blob/an-bc/big-schema-codegen/guides/CODEGEN_LARGE_SCHEMA_GUIDE.md)
Test bed: [`Demos/BigSchemaDemo`](https://github.com/MemberJunction/MJ/blob/an-bc/big-schema-codegen/Demos/BigSchemaDemo/README.md)

```mermaid
flowchart LR
  subgraph before["Before — one file per output kind"]
    direction TB
    E1["__mj · Users"] --> M["entity_subclasses.ts"]
    E2["crm · Customers"] --> M
    E3["billing · Invoices"] --> M
    M --> T1["tsc sees one dirty file"]
  end

  subgraph after["After — one file per schema"]
    direction TB
    S1["__mj"] --> F1["entities/__mj.ts"]
    S2["crm"] --> F2["entities/crm.ts"]
    S3["billing"] --> F3["entities/billing.ts"]
    F1 --> B["barrel only"]
    F2 --> B
    F3 --> B
    F2 --> T2["tsc recompiles crm.ts"]
  end
```

---

## Synced with `next` (2026-08-13)

`origin/next` was merged into this branch (`7495d8def4`). The incoming side includes **Query & Entity Materialization** (#3735) plus storage/vector/PG-parity follow-ups.

### Conflicts (only two files)

Both were the same shape: **this branch's thin barrel vs `next`'s generated monolith**.

| File | Ours (`HEAD`) | `next` |
|---|---|---|
| `packages/MJCoreEntities/src/generated/entity_subclasses.ts` | 5-line barrel → `./entities/__mj.js` | ~121k-line monolith (376 entity classes) |
| `packages/MJServer/src/generated/generated.ts` | 8-line barrel → `./graphql-schemas/__mj.js` | ~97k-line monolith (752 GraphQL types) |

Everything else from `next` auto-merged, including the new Materialization package, Angular forms for the two new entities, CodeGenLib materialization analysis, and `SQLServerCodeGenProvider` (our `getRoutineNamesBySchemaSQL` / `forceLogForRoutine` survived).

### How it was resolved

Taking `next`'s monoliths would undo the point of this PR. Blind `--ours` on the barrels would drop the two entities #3735 added.

**Kept the barrels.** Ported `next`'s new types into the per-schema `__mj` files so the catalog stays complete:

- New entities: `MJMaterializedResultEntity`, `MJMaterializedResultQueryEntity` (+ Zod schemas) in `entities/__mj.ts`
- New GraphQL ObjectTypes / inputs / resolvers: `MJMaterializedResult_`, `MJMaterializedResultQuery_` in `graphql-schemas/__mj.ts`
- Existing-class deltas from `next` (verified by field-level diff; these were the only ones):
  - `MJQueryEntity` / `MJQuery_` / create+update inputs: `IsMaterialized`
  - `MJEntity_`: `MJMaterializedResults_SourceEntityIDArray`, `MJMaterializedResults_GeneratedEntityIDArray`
  - `MJQuery_`: `MJMaterializedResultQueries_QueryIDArray`

After the splice: **376 / 376 entity classes and 1880 / 1880 GraphQL classes, zero field diffs vs `next`**.

### Post-merge verification

| Check | Result |
|---|---|
| `@memberjunction/core` build | pass |
| `@memberjunction/core-entities` build | pass |
| `@memberjunction/core-entities-server` build | pass |
| `@memberjunction/server` build | pass |
| `@memberjunction/codegen-lib` build | pass |
| CodeGenLib unit suite | **54 files passed, 6 skipped (PG integration); 1015 passed, 66 skipped, 0 failed** |

A reviewer pulling this branch still needs to run `mj migrate` (the #3735 materialization migration is new on `next`) before a live `mj codegen` against a database that does not already have those tables.

```mermaid
flowchart TD
  NEXT["origin/next monoliths"] --> CMP{"Same path as our barrels?"}
  CMP -->|entity_subclasses.ts / generated.ts| KEEP["Keep the barrel"]
  KEEP --> PORT["Port new types into entities/__mj.ts<br/>and graphql-schemas/__mj.ts"]
  PORT --> PARITY["Field-level parity vs next"]
  PARITY --> BUILD["core / core-entities / server / codegen-lib"]
  CMP -->|everything else| AUTO["Auto-merge"]
```

---

## Invariant

> **Entity is the identity. Schema is the incremental unit.**

`Metadata.Entities` is still one complete array. Nothing is sharded. There is no second catalog for application tables. What becomes incremental is everything that *can* key off schema.

```mermaid
flowchart TB
  subgraph identity["Identity — never sharded"]
    CAT["Metadata.Entities<br/>complete catalog"]
    CAT --> R1["Users"]
    CAT --> R2["Customers"]
    CAT --> R3["Invoices"]
  end

  subgraph increment["Incremental unit — schema"]
    direction LR
    MJ["__mj.ts"]
    CRM["crm.ts"]
    BILL["billing.ts"]
  end

  R1 -.-> MJ
  R2 -.-> CRM
  R3 -.-> BILL
  CRM -->|"one dirty table"| REWRITE["rewrite crm.ts only"]
  MJ -.->|"untouched"| SKIP["write-if-changed skips"]
  BILL -.-> SKIP
```

MJ Core (`__mj`) is a full platform catalog and is emitted the same way as every other schema. A change in `crm` must not rewrite `__mj.ts`.

---

## The eight items

| # | What | Where |
|---|---|---|
| 1 | **Per-schema emit** — one `.ts` file per schema + a thin barrel. Public import paths (`entity_subclasses.js`, `generated.js`) do not change. | `entity_subclasses_codegen.ts`, `graphql_server_codegen.ts` |
| 2 | **Write-if-changed** — skip the disk write when bytes are identical so mtimes (and incremental `tsc`) stay stable. | `file-write.ts` |
| 3 | **Dirty-schema regen** — a full run rebuilds only schemas in `newEntityList ∪ modifiedEntityList` (plus missing files). `--skipdb` rebuilds all, still gated by write-if-changed. | `schema-emit.ts`, `runCodeGen.ts` |
| 4 | **`excludeTables` strings** — `'audit.EntityRecordVersions'`, `'%.%History'`, `'%Audit%'` parse to the same `{schema, table}` LIKE filter. | `exclude-tables.ts` |
| 5 | **Schema-parallel file gen** — bounded `mapLimit` (default 8) over independent schema string-builds. SQL Server per-entity SQL width is `fileEmit.sqlEntityBatchSize` (default 8). PG SQL stays serial. | `schema-emit.ts`, `sql_codegen.ts` |
| 6 | **Incremental `tsc`** — `incremental: true` + `tsBuildInfoFile` on `@memberjunction/core-entities` and `@memberjunction/server`. | those two `tsconfig.json` files |
| 7 | **Hydrate-by-schema** — `md.Entities` is **not** sharded. New projection helpers (`EntitiesInSchema`, `summarizeEntitiesForContext`) are what an agent / MCP tool should ask for. | `schemaHydration.ts`, `Metadata` |
| 8 | **Package topology** — `schemaOutput` routes or skips artifacts for matching schemas. `entityPackageName` still maps schema → npm package. | `schema-output.ts` |

`fileEmit` defaults are the new recommended behavior. `fileEmit.perSchema: false` keeps the historical single file.

```mermaid
flowchart TD
  START["mj codegen"] --> DISC["Discover new / modified entities"]
  DISC --> DIRTY["dirtySchemas =<br/>those entities' schemas<br/>plus any missing file"]
  DIRTY --> Q{"Is this schema dirty?"}
  Q -->|yes| BUILD["Assemble the schema .ts"]
  Q -->|no| SKIP["Skip string-build"]
  BUILD --> WIC{"Bytes changed?"}
  WIC -->|yes| WRITE["Write file"]
  WIC -->|no| NOWRITE["Leave mtime alone"]
```

```mermaid
flowchart TD
  ENT["Entity in schema S"] --> MATCH{"First schemaOutput<br/>pattern that matches S?"}
  MATCH -->|yes, directory set| ROUTE["Write to the override directory"]
  MATCH -->|yes, type in skip| DROP["Do not emit that type"]
  MATCH -->|no match| DEF["Write to the default output[] directory"]
```

```mermaid
flowchart LR
  subgraph ram["In memory — always complete"]
    MD["md.Entities"]
  end
  subgraph proj["On the wire / in a prompt"]
    P1["EntitiesInSchema('crm')"]
    P2["summarizeEntitiesForContext(...)"]
  end
  MD --> P1
  MD --> P2
  P1 --> AGENT["Agent / MCP context"]
  P2 --> AGENT
```

---

## Fixes found by running the full cycle

These showed up only when CodeGen was pointed at a real multi-schema database:

- **`runCommand` trusted stderr.** `tsc` / `pnpm` print the word `ERROR` on successful builds. AFTER commands now succeed or fail on **exit code only**, and a real non-zero keeps the captured output.
- **Per-schema GraphQL import.** `__mj.ts` lives in `graphql-schemas/`, so `mj_core_schema` must import `../../config.js`. The old `../config.js` broke `@memberjunction/server` `tsc`.
- **Missing procs after a schema drop.** Dropping `bsd_*` deletes stored procedures but leaves `MJ: Entities` rows. Incremental SQL skipped them. CodeGen now snapshots `sys.procedures` and force-emits `CREATE PROC` for anything missing.
- **SQL Server CRUD validator was a no-op.** No `getRoutineNamesBySchemaSQL` override meant it always reported “passed.” It now queries `sys.procedures`, so 3,254 entities are actually checked.

```mermaid
flowchart TD
  SNAP["Snapshot sys.procedures"] --> NEW{"New or modified entity?"}
  NEW -->|yes| EMIT["Emit CREATE PROC"]
  NEW -->|no| MISS{"Routine missing?"}
  MISS -->|yes| EMIT
  MISS -->|no| SKIP["Skip"]
  EMIT --> VAL["Validator reads sys.procedures"]
  VAL --> OK{"Every expected routine present?"}
  OK -->|yes| GREEN["Validation passed"]
  OK -->|no| FAIL["Fail the run"]
```

Advanced generation is gated to `newEntityList ∪ modifiedEntityList`. A later no-op run does not send 2,880 entities back to the model.

---

## BigSchemaDemo

A droppable private-DB test bed. Not in the npm workspace. Not published.

| Profile | Shape | SQL |
|---|---|---|
| `smoke` | 3 × 12 = 36 tables | **committed** under `sql/smoke/` |
| `standard` | 24 × 120 = 2,880 tables | generated on demand |
| `large` | 36 × 150 = 5,400 tables | generated on demand |

```mermaid
erDiagram
  Hub ||--o{ Child : has
  Child ||--o{ Grandchild : has
  Lookup ||--o{ Child : "optional"
  Hub ||--o{ XRef : "many-to-many"
  Lookup ||--o{ XRef : "many-to-many"
  Hub ||--o{ Bridge : "LocalHub"
  PrevHub ||--o{ Bridge : "RemoteHub"
```

`recreate.sh` / `recreate.ps1` refuse shared database names (`MJ_6_1_0`, `MJ_DEV`, `master`, …).

---

## What was proven

Private DB `MJ_6_1_0_BIG_SCHEMA_CODEGEN`. Same-named remote `origin/an-bc/big-schema-codegen`. The local `an-dev-65` checkout was not used.

| Check | Result |
|---|---|
| Standard SQL | **24 schemas, 2,880 tables, 4,872 FKs, 20,522 seed rows** |
| Metadata | **2,880 `bsd_*` entities**, 3,254 total |
| `npx mj codegen` | **exit 0**, AFTER `pnpm` builds completed |
| TypeScript emit | 24 entity files + 24 GraphQL files under `Demos/BigSchemaDemo/generated` |
| Leak | **none** into `MJCoreEntities` or Explorer forms |
| No-op TypeScript emit | **written 0 / skipped 4 / schemas emitted 0 / skipped 50** |
| First discovery (after self-heal) | **~160s** for 3,254 entities |
| CodeGenLib unit suite (post-`next` merge) | **1015 passed, 66 skipped, 0 failed** |
| SQL Server CRUD validator | **3,254 entities actually checked** |

A first discovery of thousands of new entities is still a full pass. Later file-emit time is proportional to **dirty schemas**.

---

## First-run note for reviewers

The next `mj codegen` after merge will rewrite `entity_subclasses.ts` / GraphQL `generated.ts` as **barrels** and add `entities/<schema>.ts` / `graphql-schemas/<schema>.ts`. Public exports are unchanged. A no-op second run should change **zero TypeScript mtimes**.

If the reviewer's database does not yet have the #3735 materialization tables, run `mj migrate` first. Those two entities already exist in the committed `__mj` files so the TypeScript build does not depend on that migrate.

---

## Out of scope

- `--resume` of a killed CodeGen run
- `worker_threads` for the string-build
- Default-on lazy `EntityField` hydration
- Parallel PostgreSQL SQL generation (catalog deadlocks)

---

## Changeset

`patch` — TypeScript, tests, docs, demo generator. No migration, no `metadata/` edits from *this* branch. The `minor` changesets now in the PR came in with `next` (#3735 materialization and related).


AN-BC:
## Sequencing note — pairing this with #4018

@bg-BC @rkihm-BC — #4018 (DX follow-ups from the now-closed #2837 audit) is open and ready. Proposed order:

1. **Finish and merge this PR (#3804).**
2. **Pull `next` into #4018, retest, merge.**

Being precise about why, because it is **not** a merge dependency: the two PRs share **zero files**. #3804 is `packages/CodeGenLib/src/**` plus the generated barrels and the BigSchemaDemo tree; #4018 is docs (`README`, `CONTRIBUTING`, `scripts/README.md`), three workflow `paths:` filters, one new gate script, and three `CLAUDE.md` files. Either order merges cleanly. The sequence is about giving #4018 a clean retest against whatever this lands, not about a conflict.

### Two things from #4018 that touch this PR

**A new CI step will appear here once you pull `next` in.** #4018 adds `check:browser-manifest` to `test.yml` — a gate gainst server-only packages leaking into the browser class-registration manifests (the June-2026 `ng-bootstrap → templates → aiengine` incident). It passes on `next` today, and #3804 doesn't touch the browser manifests, so no action is expected. Flagging it so a new step in the run isn't a surprise.

#4018 also edits `test.yml`'s `paths:` filter and its affected-scope regex — both were still keyed to the deleted `package-lock.json`, so lockfile-only PRs were triggering neither the unit-test nor the integration workflow. Since this branch is already behind `next`, that arrives with the same merge.

**One interaction I checked, and it's benign.** This PR splits `entity_subclasses.ts` — which carries **419 `@RegisterClass` decorations** — into `entities/__mj.ts` plus a 5-line barrel. That looked like it could break class-registration manifest discovery. It doesn't: `GenerateClassRegistrationsManifest.ts` globs `**/*.ts` recursively so it still finds the classes, and `collectExportedClassNames` explicitly recurses through `export * from '...'` chains with `.js`-extension stripping — exactly the barrel's shape. One extra resolution hop, no behavior change. Still worth eyeballing a real `pnpm run mj:manifest` on this branch, since I verified by reading the resolver rather than by building.

### The real dependency is downstream of both

`plans/manifest-generation-parallelization.md` (added by #4018) proposes parallelizing the 9-step serial `mj:manifest` chain. That work should wait for **this** PR, because #3804 introduces two primitives it should reuse rather than duplicate:

- **`file-write.ts` (write-if-changed)** — the manifest generator currently writes unconditionally (three raw `fs.writeFileSync` calls, no byte comparison). Adopting it means a regenerated-but-identical manifest leaves mtime alone.
- **`schema-emit.ts`'s bounded `mapLimit`** — there is no bounded-concurrency helper anywhere in CodeGenLib on `next` today. This PR introduces the first. Building a second, divergent one in a standalone runner would be the wrong call.

That plan also documents a live bug worth your eyes since it's in your area: `mj:manifest:server-bootstrap` runs *before* `mj:manifest:server-bootstrap-lite`, whose `package.json` it reads. Serial execution masks it today, but the per-package `prebuild` scripts already run the same commands concurrently under turbo at default concurrency 10, with no cap and no ordering.

---

rkihm-BC (commented):
## Review — three mechanical blockers and one correctness blocker

The design holds up. *Entity is the identity, schema is the incremental unit* is the right call, the `runCommand` exit-code fix and the SQL Server `getRoutineNamesBySchemaSQL` override are real bugs found by actually running the cycle, and both CI gates were updated rather than routed around. What follows is what stands between this and merge.

Worth stating up front: **`e5fcfb92` was all six checks green**, Unit Tests and Integration Tier included. The merge commit an hour later (`71f6c2287`) turned everything red, and it's a single cause. The code was verified; the branch state is what regressed.

---

## Mechanical blockers

### 1. All four red checks are one stale lockfile

Every failing job dies on the same line:

```
ERR_PNPM_OUTDATED_LOCKFILE  pnpm-lock.yaml is not up to date with packages/OpenApp/Engine/package.json
* 2 dependencies were added: mssql@^11.0.1, pg@^8.13.1
```

Nothing in this PR caused it — the last `next` merge took a `package.json` change without the lockfile. `Check Dependencies`, `Unit Tests`, `MJ standards` and `Integration Tier` all fail at `pnpm install --frozen-lockfile`, before running a single test.

**Resolve:** `pnpm install` at the repo root, commit `pnpm-lock.yaml`. That alone should restore the green run from `e5fcfb92`.

### 2. Merge conflicts — two files

`mergeable: CONFLICTING`. Confirmed via `git merge-tree`:

- `packages/CodeGenLib/src/Config/config.ts`
- `packages/CodeGenLib/src/Misc/entity_subclasses_codegen.ts`

Everything else auto-merges.

### 3. The branch is 74 commits behind `next`

Same resync shape as the one already documented in the PR description.

**On the hand-spliced generated files** — I verified the parity claim rather than taking it on faith, and it holds. `next` has not touched `entity_subclasses.ts` or `generated.ts` since the last sync, and the rosters match exactly:

| Check | `next` | branch `__mj` |
|---|---|---|
| Exported entity classes | 381 | 381 |
| `get`/`set` accessors | 9,803 | 9,803 |
| GraphQL exported classes | 1,905 | 1,905 |
| Symbol-name diff | — | only `loadModule` (moved to the barrel) |

**Resolve:** on this resync, regenerate via `mj codegen` rather than splicing again. The splice was correct this time, but it has to be re-proved on every sync, and hand-editing generated ORM output is what `packages/MJCoreEntities/CLAUDE.md` forbids.

---

## Correctness blocker — dirty-schema regen has no deleted-entity signal

`resolveDirtySchemasForEmit` keys dirtiness off `newEntityList ∪ modifiedEntityList`. There is no deleted list.

To be precise about the scope, because most of this *is* covered: deleted **fields** are handled — `manage-metadata.ts:5215` feeds them into `addNewEntitiesToModifiedList`. All four writers of that list are field-level and key to entities that still exist. `checkAndRemoveMetadataForDeletedTables` (line 3934) deletes the entity metadata, drops the view and procs, refreshes `Metadata` — and touches neither list. Whole-entity deletion is the only uncovered case.

Consequence: drop a table in schema `S`, run `mj codegen` → `S` isn't dirty → `entities/S.ts` is not rebuilt → the dead class and its `@RegisterClass(BaseEntity, '…')` stay on disk indefinitely. Under the old monolith this self-corrected, because the file was rewritten in full every run. `dirtySchemaOnly: true` is the new default, so this is on by default.

### The failure mode is a gate error that points at the wrong file

Angular is **not** dirty-gated — `runCodeGen.ts:688` calls `generateAngularCode(group, dir, '', currentUser)` with no emit options, so forms regenerate in full every run. Subclasses and GraphQL resolvers *are* gated, so they keep the stale entry in lockstep with each other while the forms roster drops it. `check-codegen-tail` compares those rosters bidirectionally:

```js
for (const e of diff(entities, other))
    failures.push(`${e}: has a generated subclass, but no ${label} entry`);
```

So deleting an entity yields `<Entity>: has a generated subclass, but no Explorer forms entry` — a gate failure naming the forms module, when the cause is a subclass file that was never rebuilt. That is a genuinely hard afternoon for whoever hits it.

### Second half — stale schema files are never pruned

If a schema empties out entirely, `entities/<schema>.ts` is orphaned. The barrel stops exporting it, but:

- it stays inside `include: ["src/**/*"]`, so `tsc` still compiles it;
- `GenerateClassRegistrationsManifest` globs `**/*.ts`, so its classes still land in the manifest — the same leak class `#4018`'s `check:browser-manifest` exists to catch;
- `readArtifactSource` does `readdirSync(splitDir).filter(f => f.endsWith('.ts'))` and concatenates **everything** in the directory regardless of what the barrel exports, so the orphan inflates the roster and fails the gate.

**Resolve:** a deleted-entity signal feeding `resolveDirtySchemasForEmit`, plus pruning of `entities/*.ts` and `graphql-schemas/*.ts` whose schema no longer appears in `schemas`. The pruning half is small, independently valuable, and closes the manifest-leak path on its own.

---

### Path to merge

`pnpm install` + commit the lockfile → resolve the two conflicts → regenerate (don't splice) the `__mj` files → close the deleted-entity gap → re-run CI. Given `e5fcfb92` was fully green, everything except the deleted-entity work is mechanical.

I have a few smaller notes (duplicated `resolveEmitOptions`/`emitFile` across the two generators; the GraphQL emitter resolving `parallel`/`concurrency` it never uses; BEFORE-command failures logged twice because both the BEFORE and AFTER loops iterate the full `commandFailures` array). None of them are merge-blocking — happy to drop them inline if useful.

rkihm-BC (approved):
## Approved

Both blockers from the previous review are closed, and I re-verified rather than taking the commit messages on faith.

### Mechanical

`mergeStateStatus: CLEAN`, current with `next`, and CI fully green — 16 passing (Build, all six unit shards, Integration SQL Server deterministic, dependency check, adopted standards, ESM guard, source guards, migrations, docs), 4 skipped nightly/weekly. The lockfile came right via `bd93729885` on `next`.

Generated-file parity against the **current** `next` after this merge: **381/381** entity classes, **9,803/9,803** accessors, **1,905/1,905** GraphQL classes, with `loadModule` as the only symbol delta — expected, it moved to the barrel. No splice was needed this round; `next` hasn't touched either monolith since the prior sync.

### `c14dc6f394` — deleted-entity signal

Right shape. Recording the **schema** rather than an entity name is the correct call, for the reason the comment gives: by the time `resolveDirtySchemasForEmit` runs the entity is gone from metadata and a name could never be resolved back to a schema. Recording only *after* the delete succeeds is the detail I'd have asked for — a failed removal (the one that gets logged for an admin to handle by hand) must not dirty a schema for an entity that is still present. I confirmed it is genuinely wired through `buildSchemaEmitOptions`, not just added to the signature.

### `a85d0195ef` — pruning

Splitting `selectOrphanedSchemaFiles` (pure) from `pruneOrphanedSchemaFiles` (the filesystem wrapper) is the right seam, and calling it after the emit loop but **before** the barrel means the directory and the barrel cannot disagree. Passing the run's full schema set rather than the dirty subset is the load-bearing choice.

I went looking for the way this could go wrong — a scoped run deleting files for schemas that are still live — and it holds up on every path:

- `includeSchemas` resolves *into* `excludeSchemas` (`schema-scope.ts`), so there is no second narrowing path to get out of step with;
- file generation always reads the full config-scoped `md.Entities` — `entityFilter` only reaches `manageEntityFields`;
- `partitionEntitiesByOutputDirectory` means each prune only ever sees its own output directory's schemas, so the `bsd_%` route and the default route cannot delete into each other.

Prune scope equals emit scope everywhere. The tests cover the case I'd have worried about most: `README.md` and `.gitignore` are left alone.

The three minor notes were folded in as well — shared `resolveSchemaEmitOptions`/`emitSchemaFile` with the protected methods kept as an override seam, and `recordCommandFailures` returning its own batch so BEFORE failures no longer print twice.

### Worth flagging, in your favour

`next` landed `a1a8989f5a` (peer entity imports via `entityImportPackages`) and it merged cleanly with the split. `CollectPeerClassImports` now resolves through `resolveEntityImportPackage(schema, owningSchema, config)` and **throws** on an ambiguous or unresolvable peer instead of silently falling back to `mj_generatedentities`. That closes a latent cross-schema `EmbeddedRecord` path I had raised and then downgraded after finding it never fires today — and it matters more on this branch than on `next`, because per-schema splitting is precisely what turns same-file peers into cross-file imports.

### One nit, explicitly not blocking

The GraphQL emitter's serial loop still has no inline comment explaining why it ignores `parallel`/`concurrency`. The reasoning — entity assembly is async so it overlaps, GraphQL assembly is synchronous so there is nothing to overlap — is in the commit message and not at the code. Fold it into a future touch of that file.

Nice work on the follow-ups. The commit messages on both fixes carry the causal chain, not just the change, which is what made this pass quick to verify.

PR #3809, opened 2026-08-14
feat(ng-base-forms): forward explorer grid toolbar chrome

AN-BC:
## Summary
- `mj-explorer-entity-data-grid` now forwards the inner grid's toolbar chrome, not just New
- Hosts can hide search, New, Refresh, Export, Delete, Compare, Merge, Add to List, Find Duplicates, Send Message, the recycle-bin chip, or the whole bar
- Defaults match `<mj-entity-data-grid>`, so existing related lists do not change
- DOM tests cover the default pass-through and a full chrome override

## Test plan
- [ ] Related entity grid with no new bindings still looks the same (search + New + Refresh + Export)
- [ ] `[ShowNewButton]="false"` hides New and leaves the rest of the toolbar
- [ ] `[ShowSearch]="false"` / `[ShowRefreshButton]="false"` / `[ShowExportButton]="false"` hide those pieces independently
- [ ] `[ShowToolbar]="false"` hides the entire bar
- [ ] Unit tests: `explorer-entity-data-grid.component.dom.test.ts`


AN-BC (commented):
LGTM

PR #3827, opened 2026-08-15
security: SSRF guard for web actions, scope-limit ad-hoc SQL, escape resolver filters

AN-BC:
## Summary

Adversarial security review of the MJ framework. The auth/JWT/API-key/magic-link surface and the CodeGen/stored-proc SQL layer are notably well-hardened; this PR fixes the exploitable gaps found outside them. No new dependencies (Node built-ins + Angular/Node globals already in use).

### 1. 🟠 HIGH — SSRF in the three URL-fetching Core Actions
`web-page-content`, `url-metadata-extractor`, `url-link-validator` take a fully user-/agent-supplied `URL`, validate only the protocol, and follow redirects. `ActionResolver.RunAction` is reachable by any authenticated principal (and by AI agents processing untrusted text), so these can be steered at cloud instance-metadata (`169.254.169.254`), loopback, and internal hosts — a direct path to IMDS credential theft and internal scanning.

**Fix:** new `packages/Actions/CoreActions/src/custom/web/ssrf-guard.ts` — resolves the target via DNS and rejects any address in a private / loopback / link-local / CGNAT / unique-local / reserved range (IPv4 + IPv6, incl. IPv4-mapped), plus metadata hostnames. `ssrfSafeFetch` follows redirects manually and re-validates **every hop** (a public host that 302s to internal is blocked); the axios/HEAD paths validate the initial URL. Wired into all three actions.
> Note: does not fully close DNS-rebinding (resolve-then-connect races) — pinning the resolved IP for the connection is a larger follow-up. The direct-URL and redirect-to-internal vectors (the exploitable ones) are closed.

### 2. 🟠 HIGH — `ExecuteAdhocQuery` read-authorization bypass
`AdhocQueryResolver.ExecuteAdhocQuery` runs raw `SELECT` on the read-only pool **outside** RunView / entity permissions / row-level security / magic-link scope, guarded only by `validateFullQuery` (which permits selecting any application table). A **scope-limited magic-link or anonymous guest** could therefore read the entire database outside its granted scope.

**Fix:** block principals with `IsMagicLinkAnonymous` or an active `MagicLinkScope`. Normal authenticated users (the intended consumers via `GraphQLDataProvider`) are unaffected — this is deliberately the minimal, non-breaking guard rather than `@RequireSystemUser` (which would break the client feature).

### 3. 🟡 MEDIUM (defense-in-depth) — unescaped values in `ExtraFilter`
`ResolverBase.findBy` (reachable via e.g. `UserByEmail`) and the inline `ViewName` filter in `RunViewByNameGeneric` interpolated client-supplied values without escaping. `ExtraFilter` already passes through `SQLExpressionValidator` (which blocks stacked statements / UNION / comments), so the residual risk was a same-clause boolean tautology — now closed by quote-doubling, matching the convention used everywhere else.

## Testing
Static edits only — no build was possible in the review environment (no `node_modules`). Changes are surgical, type-correct, and dependency-free; CI should validate compilation/tests.

## Recommended follow-ups (NOT in this PR — need a team decision / larger change)
- **CORS default** (`MJServer/src/index.ts`): default `['*']` reflects the request Origin with `credentials: true`. Not exploitable today (bearer-token auth, no cookies) but a latent footgun — prefer an explicit allowlist and never pair wildcard-reflection with credentials.
- **GraphQL depth/complexity limiting** (`apolloServer/index.ts`): no depth or cost cap — an authenticated client can submit a deeply nested query for a CPU/DB DoS. Add `graphql-depth-limit` + a complexity rule.
- **MCP OAuth proxy** (`AI/MCPServer`): when the consent screen is disabled (default), client-requested scopes are minted verbatim — intersect with a server-side allowlist / the user's entitlements, or default the consent screen on. Also, `AuthGate.extractCredentials` accepts API keys in URL query params (leak to logs/referrers) — require headers only.
- **DNS-rebinding**: pin the resolved IP for the outbound connection in `ssrf-guard`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PqKLBZEvDrP7G3LCnB2z6r)_


AN-BC:
Closing as superseded — all three items here are now accounted for, but by two different PRs, so this one has nothing left to merge.

**1. SSRF guard for the URL-fetching web actions → shipped in #4019** (which superseded #4011). The guard no longer lives under `CoreActions`: it is a zero-dependency `@memberjunction/network-utils` package (`AssertPublicUrl`, `SafeFetch`, `IsBlockedIPAddress`, `SSRFError`), axios is gone repo-wide so the guard is reachable from every package, and six actions are covered rather than the three this PR touched — `http-request`, `graphql-query` and `api-rate-limiter` in addition to `web-page-content`, `url-metadata-extractor` and `url-link-validator`. Merging this PR's `CoreActions/src/custom/web/ssrf-guard.ts` now would re-introduce exactly the duplicate that #4011's review asked to remove.

**2. Scope-limiting `ExecuteAdhocQuery` and 3. escaping the `ResolverBase` filters → #4037.** Both were still unfixed on `next`, and both are carried forward there, with two changes worth noting:

- The escaping is rewritten on top of `EscapeSQLString` (#4033), which landed after this PR was opened and made hand-rolled `.replace(/'/g, "''")` a documented anti-pattern in `.claude/rules/data-access.md`.
- It also covers the **unquoted** slot this PR left open. `EntityFieldInfo.NeedsQuotes` is `false` for Number and Boolean fields, so those values are interpolated bare — there is no literal to break out of, which means quote-doubling protects nothing there and a string value simply *is* SQL. `findBy` now refuses a non-numeric, non-boolean value in that slot.

The analysis and severity calls in this PR's description held up — including grading the filter escaping as defense-in-depth rather than a live injection, since `SQLExpressionValidator` already blocks stacked statements, `UNION` and comments.

The recommended follow-ups listed here (CORS default, GraphQL depth/complexity limiting, MCP OAuth scope minting, DNS-rebinding pinning) are **not** carried into #4037 and remain open — they still need the team decision this PR called for.

---

PR #3830, opened 2026-08-15
fix(ng-ui-components): stop all five form controls latching their disabled state

MarceloT-BC:
## What

`mj-dropdown`, `mj-combobox`, `mj-datepicker`, `mj-switch` and `mj-numeric-input` each derive an internal `IsDisabled` gate, and the **only** thing that ever assigned it was `setDisabledState()` — the ControlValueAccessor hook. The `Disabled` input was a plain `@Input` field with no setter and no `ngOnChanges`, so it had no recompute path of its own: the gate froze at whatever the hook last composed, and every later change to the input was silently dropped.

The forms-driven half was in fact fine — `setUpControl` calls `setDisabledState()` at registration *and* wires `registerOnDisabledChange`, so the hook re-fires on every `control.disable()`/`enable()`. What was dead was purely the input-driven direction. Three shapes of the defect:

- **Starts disabled → stuck disabled forever.** The control kept rendering its disabled affordance and ignoring clicks even after the binding went `false`.
- **Starts enabled → can never be locked.** A read-only/receipt mode silently stayed editable.
- **No forms binding at all → `[Disabled]` completely inert.** With no `ngModel`/`formControl` the hook never runs, so `<mj-dropdown [Disabled]="true">` rendered fully enabled and clickable. This is the widest form of the bug.

## How to recreate

Bind a control whose `Disabled` starts true and later goes false — the classic "pick a company first" pattern:

```html
<mj-dropdown [Disabled]="!model.CompanyID" [(ngModel)]="model.AccountID" [Data]="Options" />
```

Render with `CompanyID` null, then set it. **Before:** `Disabled` reads false and `Data` is populated, but clicking the trigger does nothing — `IsOpen` stays false, no panel opens. **After:** the panel opens and an option can be selected.

Live instances of the first shape on `next` (unconditionally rendered, so no `@if` recreates them): the four `[Disabled]="!EditMode"` controls in `entity-form-chrome-editor`, and the two `[Disabled]="!ServerID"` dropdowns in `mcp-test-tool-dialog`. 20 dynamic `[Disabled]` bindings repo-wide sit on these five controls.

## Fix

Each control now keeps the input-driven and forms-driven disabled states in separate backing fields (`Disabled` becomes a setter) and recomposes `IsDisabled` whenever either side changes, marking the view for check from the setter so it is safe inside the parent's change-detection pass. The overlay controls (dropdown, combobox, datepicker) also close an open panel when they become disabled — without a nested `detectChanges()`, which would trip NG0100 when the setter runs mid-pass.

No API change: `Disabled` and `IsDisabled` keep their names, types and meanings. Controls that previously stayed enabled after their binding went true will now correctly disable.

**For posterity:** `refresh-button` and `accordion` also have `Disabled` inputs and were never affected, because they bind `Disabled` straight into the template and read it directly in their handlers. The derived `IsDisabled` mirror is what created this — any control that mirrors an input into derived state needs a recompute path, or this comes back.

## Testing

- New DOM unit tests in each of the five components pin the regression in all three directions (re-enable, lock, no-forms). The three overlay controls additionally assert that locking an open panel closes it.
- Negative proof, all five component sources reverted to `next` with the specs kept: **15 / 15 fail** (3 per control). Sources restored: 15 / 15 pass.
- Package suite `@memberjunction/ng-ui-components` at the current head (merged with `origin/next`): **585 / 585 green**; `tsc --noEmit -p tsconfig.spec.json` and `ngc` clean.
- Verified live in Explorer: the account picker's `--disabled` class flips and the popup opens the moment a company is chosen; a full journal-entry creation flow runs through the fixed dropdown.
- Reverse direction (things that now correctly lock) reviewed statically at all 20 call sites — every one is a busy guard, view-mode gate, or pick-X-first gate whose author plainly meant it to hold. Not clicked through at runtime.
- Changeset included (patch, `@memberjunction/ng-ui-components`).


MarceloT-BC:
@MattC-BC 
This is very Claude generated, but I just don’t know the problem space well enough to verify it quickly. I think this issue is valid though. Would you be able to take a look at this when you have time? This is causing a UI issue in Accounting, but I can probably work around it as well.

MattC-BC:
Took a proper look. Short version: the diagnosis is right, the fix is correct, and it actually fixes more than it claims. But there are two things that need to happen before this can merge.

## Verified good

I reverted just the five component sources back to `next`, kept the new test files, and re-ran the suite: **10 of the 15 new tests fail**. Two per component, both directions, all five controls. So the specs are genuine regression pins, not decoration. That was the main thing I wanted to confirm.

The `resetPanelState()` split is load-bearing and worth calling out as a good catch. `Close()` calls `cdr.detectChanges()`, and `syncDisabled()` can run from an `@Input` setter during the parent's CD pass, so calling `Close()` there would re-enter change detection and trip NG0100. Splitting the panel mutation out and using `markForCheck()` is the right move.

Also checked: the CDK overlay is declaratively bound to `IsOpen`, so flipping it really does detach the panel, and the `(detach)="Close()"` output early-returns so there's no loop. No API break either, field to getter/setter is consumer-compatible.

## Blocking

**1. Two tests fail.** `combobox.component.dom.test.ts` ships `RE-ENABLES...` and `LOCKS...` red:

```
× RE-ENABLES when Disabled flips to false after registration   (:170)
× LOCKS when Disabled flips to true after registration          (:177)
```

The helper does `hostTrigger().click()` on `.mj-combobox`, which is a wrapper `<div>` with no click handler. The combobox opens on `(focus)` on the inner input, or `(mousedown)` on `.mj-combobox-toggle`. Looks like the block was copied from the dropdown spec, where `.mj-dropdown` does have `(click)="Toggle()"`.

I ran it at `4c1de04110` as well as at the current head, fails the same in both, so the "package suite green, 549/549" line in the description isn't accurate. Worth noting the second test dies before its disabled assertions, so "locking closes an open combobox panel" is currently unverified for combobox specifically.

**2. The changeset numbers are off.** It says "47 dynamic `[Disabled]` bindings across this repo were affected." The real count is **20**. There are only 38 `[Disabled]` bindings repo-wide across all components, most not on these five. Since changesets land in the public changelog, worth correcting.

Same paragraph describes `dynamic-form-field` as one of the affected bindings, but its `Disabled` is `public Disabled = false`, not an `@Input()`. It never receives a `[Disabled]` binding, it passes its CVA-derived value down to five inner controls. The fix does help there, just not by that mechanism.

## Worth knowing

The bug is broader than the description. On `next`, `<mj-dropdown [Disabled]="true">` with no `ngModel` at all renders fully enabled and clickable. I probed it: `Disabled` is completely inert without a forms binding. Both probe assertions fail on `next` and pass on this branch, so the fix covers it, but it isn't mentioned or tested. Might be worth one spec since it's the broadest form of the defect.

Small mechanism nit: `setDisabledState()` isn't called "exactly once at registration." `setUpControl` also wires `registerOnDisabledChange`, so it fires on every `control.disable()`/`enable()`. The old code's forms-driven path was actually fine. What was dead was purely the input-driven direction, because `IsDisabled` is derived state with no recompute path. Doesn't change the fix, just the framing.

The third spec in each block ("stays disabled while the forms-driven state holds") passes with and without the fix. Fine to keep as a contract test, just isn't covering the defect.

Also `markForCheck()` is in `syncDisabled()` for dropdown/combobox/datepicker but not switch/numeric-input. All five are Default CD and the app runs with zone.js so it makes no difference, but might as well align them.

## Why this is worth merging

Beyond your Accounting case, there are live bugs here. Two I confirmed are unconditionally rendered, not behind an `@if` that would recreate them:

- `entity-form-chrome-editor` has four controls on `[Disabled]="!EditMode"`, rendered unconditionally from `entity-form.component.html:1298`. The form opens in view mode, the controls register disabled, and clicking Edit never revives them. That whole "Form chrome" panel is dead right now.
- `mcp-test-tool-dialog.component.html:89` has two dropdowns on `[Disabled]="!ServerID"`. Same shape as yours: open the dialog with no server, pick one, dropdowns stay dead.

Probably the same at `resource-access-resource.component.html:29` and `api-key-edit-panel.component.html:114`.

Worth a smoke test on the reverse direction too, since things that previously never locked will now start locking. `ai-test-harness` has three dropdowns on `isExecuting` / vendor availability, and `dynamic-form-field` feeds five controls in every generated entity form.

## Asks

1. Fix the two combobox specs (focus the input rather than clicking the wrapper) and correct the green-suite claim.
2. Fix the changeset: 20 not 47, and drop the `dynamic-form-field` characterization.
3. Optional but nice: one spec for the no-forms case, and align `markForCheck()` across all five.

One thing worth putting in the PR description for posterity: `refresh-button` and `accordion` also have `Disabled` inputs and were never affected, because they bind `Disabled` directly in the template. The derived `IsDisabled` mirror is what created this. Any future control that mirrors an input into derived state needs a recompute path or we'll get this again.

MarceloT-BC:
@MattC-BC Thanks — that's a genuinely useful review, and both corrections were right. All three asks are done, plus both optionals. Pushed as 3dfc3ad.

## 1. The two combobox specs — fixed

Your diagnosis was exact: the block was copied from the dropdown spec, where `.mj-dropdown` does carry `(click)="Toggle()"`. `.mj-combobox` is an inert wrapper `<div>`; the control opens on `(focus)` of the inner input or `(mousedown)` on `.mj-combobox-toggle`. Both specs now press the toggle.

The knock-on you predicted is real: with the gesture fixed, the LOCKS spec gets past its open step and reaches `expect(IsOpen).toBe(false)` — so "locking closes an open combobox panel" is now actually verified for combobox, not just inferred from the other two.

And you were right that the green-suite claim was wrong. Correct number below; I'll fix the PR description itself too.

## 2. Changeset — corrected

I recounted independently and landed on your numbers exactly: **38 `[Disabled]` bindings repo-wide, 20 on these five controls**. (My raw grep returned 42 — the extra four are two `dist/*.d.ts` build artifacts and two JSDoc examples inside `dropdown.component.ts` itself.) 47 was wrong.

Dropped the `dynamic-form-field` characterization as well — confirmed, its `Disabled` is `public Disabled = false`, assigned only by `setDisabledState()`, never an `@Input()`. It's a passer-down, not a receiver. The file does contain five `[Disabled]` bindings, but they're the ones it writes *into* its inner controls (3 × datepicker, 2 × numeric-input), which is the actual mechanism by which the fix helps it. The changeset now says that.

## 3. Both optionals — done

**No-forms spec.** Added — and I went past the ask and put it in all five rather than one. The five spec files are already structurally parallel and the defect is identical in each, so the asymmetry felt worse than the extra four tests. Easy to trim to just the dropdown if you'd rather.

**`markForCheck()` aligned.** `switch` and `numeric-input` had no `ChangeDetectorRef` at all, so this injects one in each. All five now share an identical `syncDisabled()` shape: early-return guard, assign, mark.

## Mechanism nit — taken, and it was worse than you saw

You're right, and it wasn't only in the PR description: all five components carried the "exactly once at registration" framing in their `syncDisabled()` JSDoc, which is the copy someone actually reads while changing this code. I checked it against the Angular source — `setUpControl` calls `setDisabledState()` at registration *and* calls `setUpDisabledChangeHandler`, which wires `registerOnDisabledChange`, so it fires on every `control.disable()`/`enable()`.

All five comments are rewritten to say what actually happened: `IsDisabled` is derived state whose only writer was `setDisabledState()`; the forms-driven half was always live; what had no recompute path at all was the `Disabled` `@Input`.

## Verification, done your way

Reverted just the five component sources to `next`, kept the specs:

| Check | Result |
|---|---|
| 5 sources at `next`, specs kept | **15 / 15 fail** — 3 per control × 5 (re-enable, lock, no-forms) |
| Sources restored | **15 / 15 pass** |
| Full package suite | **573 / 573 green** |
| `tsc --noEmit -p tsconfig.spec.json` | clean |
| `ngc` | clean |

Both repaired combobox specs now fail on `next` at the *latch* assertions — `and the gate must have followed it`, and `IsDisabled` after `lock(true)` — rather than at the gesture, which is the point.

## On the reverse-direction risk

I read all 20 call sites rather than smoke-testing them, so please take this as static review, not a runtime check.

Every one is a guard whose author plainly meant it to hold: busy guards (`isExecuting`, `saving`, `IsDetecting`, `ScopeSaving`, `IsLoading`), view-mode gates (`!EditMode`, `!IsEditing`), and pick-X-first gates (`!ServerID`, `!SelectedDomainName`, `!selectedModelId`). I found no site where newly-correct locking looks unwanted — the busy guards in particular were previously letting a user mutate state mid-flight, which is its own latent bug. But I did not click through `ai-test-harness` in a browser, so if you want runtime confirmation on that one it's still outstanding.

## Your posterity note

Confirmed and going into the description: `refresh-button` and `accordion` both bind `Disabled` straight into the template and read it directly in their handlers, so there's no derived mirror to go stale. That's the actual rule — **any control that mirrors an input into derived state needs a recompute path**, and these five are the ones that mirrored.

Also worth recording that the third spec in each block ("stays disabled while the forms-driven state holds") passes with and without the fix — you're right that it's a contract test rather than a regression pin. I've kept it, but it isn't carrying any of the weight.

MattC-BC:
Read through all five controls plus the specs. The diagnosis is right, the fix is the smallest correct shape (two backing fields, one compose point, `markForCheck` from the setter so it's safe inside the parent's CD pass), and closing an open panel the moment the control locks is the right UX call. I also checked that these five are the only CVAs in the package, that the 20-binding count in the changeset matches a repo grep, and that nothing binds a static string `Disabled="..."` that would newly flip. CI is green on every shard; I read rather than ran locally.

A few non-blocking nits, take or leave:

 - The "setDisabledState is called ONCE at registration" wording was corrected in the source JSDoc in the follow-up commit, but it's still in the block comment at the top of each of the five disabled-state spec blocks, and in the PR description. Since the body becomes the squash commit message, worth a refresh (it also still says 549/549 and doesn't mention the no-forms case).
 - Two older comments in `dropdown.component.dom.test.ts` (around lines 76 and 200) still say `[Disabled]` only takes effect once a forms adapter calls it. That's the bug this PR removes.
 - Datepicker's close-on-lock (`if (disabled) this.IsOpen = false`) has no spec. The LOCKS spec never opens the calendar first; dropdown and combobox both do. Two lines to add.
 - The forms-churn spec drives `setDisabledState` by hand. Fine for a unit test, but a reactive-forms host calling `disable()`/`enable()` would pin the `registerOnDisabledChange` path the JSDoc now relies on.

 Not for this PR: a `booleanAttribute` transform on the setter would make a bare `<mj-switch Disabled>` behave (no consumer does that today), and if a sixth CVA ever lands, the copy-pasted triple is worth extracting into a shared helper.

MattC-BC (dismissed):

MattC-BC (approved):
Re-approving. The only changes since my Sept 4 approval are the spec fixes I asked for (refreshed block comments, the two stale dropdown comments, and datepicker's LOCKS spec now opening the calendar first and asserting close-on-lock) plus merges from `next`. Sources and changeset are unchanged, each spec file carries exactly four more tests than `next` after the accessible-name merge, and CI is green on every lane including all six unit shards. The optional reactive-forms churn spec is fine to leave for another day.

PR #3831, opened 2026-08-15
fix(core): stop per-request ProviderBase event-bus subscription leak

AN-BC:
## Summary

This PR was generated by a scheduled `/audit-memory-leaks` run (Round 11 of the recurring audit — see `plans/MEMORY_LEAK_AUDIT.md`). Ten parallel `Explore` subagents swept the monorepo across RxJS/OnDestroy, timers, event listeners, unbounded caches, and connection/resource categories, in a broad Wave 1 pass plus a deep Wave 2 subtree sweep. One finding rose to **Critical**, and this PR fixes it same-day.

### The bug

`ProviderBase.ensureInflightViewInvalidation()` (`packages/MJCore/src/generic/providerBase.ts`) called `MJGlobal.Instance.GetEventListener(false).subscribe(...)` once **per provider instance**, and never stored or unsubscribed the returned `Subscription`. `MJGlobal`'s underlying `_eventsSubject` is a plain RxJS `Subject` that lives for the process lifetime and retains every subscriber forever unless it explicitly unsubscribes.

`packages/MJServer/src/context.ts`'s `createPerRequestProviders()` constructs a **brand-new** `SQLServerDataProvider`/`PostgreSQLDataProvider` on **every single GraphQL request**, and `packages/MJServer/src/services/TaskGraphProviderFactory.ts` mints one per durable task-graph task execution. The first `RunView`/`RunViews` call on each of those short-lived providers — effectively every request — permanently pinned that provider (connection pool reference, entity metadata, `_inflightViews` map, and everything else it transitively references) on the global bus. Unbounded heap growth, directly proportional to request volume, with no upper cap — visible as steady load-proportional growth in a production memory graph within hours.

### The fix

Replaced the per-instance subscription with a single **static, process-lifetime** subscription (`ProviderBase.ensureStaticInvalidationSubscription()`), wired lazily and at most once per process, that fans events out to every still-live provider instance. Live instances register into `private static _invalidationTargets = new Set<WeakRef<ProviderBase>>()` instead of subscribing themselves — `WeakRef` means registering an instance no longer keeps it reachable, so a per-request/per-task provider becomes eligible for GC as soon as nothing else references it. Dead `WeakRef`s are pruned lazily the next time an entity event fires.

This fixes the leak at its single root cause with no changes required anywhere else (no plumbing needed through `context.ts`'s Apollo plugin layer or `TaskGraphDispatcher`'s ~6 provider-creation call sites), and is a strict improvement for every `ProviderBase` subclass, including long-lived client-side providers (they simply stop contributing their own redundant permanent subscription).

## Test plan

- [x] New regression test `packages/MJCore/src/__tests__/providerBase.invalidationSubscriptionScaling.test.ts`:
  - Creating 21 provider instances and touching each one's write-invalidation path adds **at most 1** total subscription to the process-wide event bus. Verified this test **fails against the pre-fix code** (21 providers → 22 bus subscriptions) and **passes against the fix**.
  - A single entity-save event still fans out and invalidates lingered entries on every live provider instance — no functional regression.
- [x] All 6 pre-existing tests in `providerBase.lingerInvalidation.test.ts` pass unchanged.
- [x] Full `@memberjunction/core` unit suite: **137 files / 2008 tests passing**.
- [x] Full monorepo build (`pnpm run build`, 302 packages): 0 TypeScript errors. (One package, `@memberjunction/cli`, showed a transient failure under the massive parallel Turborepo run that did not reproduce when built standalone afterward — confirmed unrelated to this change.)
- [x] Changeset added (`@memberjunction/core`: patch).
- [ ] Deterministic integration tier (`pnpm run test:integration`) was not run in this session — it requires a live database with migrations + CodeGen applied, which is out of scope for this automated audit-and-fix run. This change has no migration/metadata/schema impact and is isolated to `ProviderBase`'s private write-invalidation wiring, which the unit regression test above exercises directly.

## Audit plan document

`plans/MEMORY_LEAK_AUDIT.md` was refreshed with this round's findings (Round 11), including 2 additional new High findings, 2 new Medium findings, and 1 new Low finding outside the scope of this fix (see the plan doc's "Top New Findings" section) — those are tracked for a future PR, not included in this change.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XCPApXkqGpufDKs21xw6xR)_


AN-BC:
@ct-bc next one for you

AN-BC (commented):
LGTM

PR #3851, opened 2026-08-16
fix(repo): unbreak pnpm install --frozen-lockfile on next

MS-BC:
**`next` itself is red, and has been since `81ed5599d7`.** So is every branch based on it. Nothing reaches a test — `pnpm install --frozen-lockfile` fails, so unit tests, the integration tier, the dependency check and the standards gate all die at the install step. `Check migrations` is the only green job, because it is the only one that doesn't install.

```
pnpm-lock.yaml is not up to date with <ROOT>/packages/Angular/Generic/graph-view/package.json
* 12 dependencies were added: @angular/common@^21.1.3, … rxjs@^7.8.2
```

## Two drifts, not one

**1 — `graph-view` (what CI reports).** #3844 added `packages/Angular/Generic/graph-view/package.json` with 48 lines of dependencies and no `pnpm-lock.yaml` update.

**2 — `@mj-biz-apps/*` (what stops you fixing 1).** `packages/MJAPI/package.json` (12 deps) and `packages/MJExplorer/package.json` (8) declare `@mj-biz-apps/*` packages — nine of them `workspace:*` — that match **no** `pnpm-workspace.yaml` glob and have **zero** entries in the lockfile. Added by `27c2e46f9f`, a base-forms commit unrelated to biz-apps: local open-app install state committed by accident.

This second one is why the obvious one-line fix doesn't work. `pnpm install --lockfile-only` never gets as far as `graph-view`:

```
ERR_PNPM_WORKSPACE_PKG_NOT_FOUND  In packages/MJAPI:
"@mj-biz-apps/accounting-server@workspace:*" is in the dependencies
but no package named "@mj-biz-apps/accounting-server" is present in the workspace
```

## Why removing them is safe

**Nothing imports `@mj-biz-apps` anywhere in the repo** — no `import`, no `require`. The only references are string literals in MJCLI's dev-workspace detection (`packages/MJCLI/src/lib/dev-workspace/detect.ts`), which inspects these very manifests to decide whether a checkout is an open-app dev workspace. Their presence in a *committed* manifest is itself the defect that logic is meant to detect.

## Verification

- `pnpm install --lockfile-only` → exit 0
- `pnpm install --frozen-lockfile` → **exit 0** (the gate that is currently failing)
- Lockfile churn is `+40/-0` — purely additive, just `graph-view`'s entry. No resolution churn across the other 300+ packages.

@AN-BC — flagging, since both commits are yours and the biz-apps manifests look like they were meant to stay local.


MS-BC:
Closing — this is AN-BC's breakage to own, not something to route around from here. Leaving the diagnosis in the PR body since the non-obvious part is that there are two stacked drifts: the `@mj-biz-apps/*` deps in MJAPI/MJExplorer make the lockfile unregenerable, so the obvious `pnpm install --lockfile-only` fix for graph-view aborts before it can help.

PR #3861, opened 2026-08-16
fix(sql-converter): never write a hollow .pg.sql when every statement is a gap

### LINKED ISSUE #3840: mj migrate convert --allow-gaps writes a hollow .pg.sql when every statement is unhandled
## One sentence

When every statement in a migration is one the AST transpiler cannot emit, `mj migrate convert --split --allow-gaps` still writes a `.pg.sql` — containing its header, its gap banner, and **no executable SQL at all** — which is the one output shape that defeats every downstream check.

## Why this shape is worse than a failure

A hollow counterpart:

- has a `.pg.sql` name, so **filename-parity checks pass**
- contains no T-SQL, so **T-SQL/contamination scans pass**
- applies to a PostgreSQL database **without error**
- does nothing

So the migration is recorded as applied, every cheap signal reports success, and the schema change simply never happens on that host. A `.needs-hand` file or a non-zero exit is strictly safer, because both are visible.

`--allow-gaps` is documented as "exit 0 even when migrations have conversion gaps (gaps are still embedded as comments and listed in the summary)". That contract is reasonable for a *partial* gap. It is not reasonable when the gap is total, because then the emitted artifact carries none of the migration's intent.

## Repro

`bizapps-common`'s `V202608132240__v5.34.x__Layered_Base_Views_People_Organizations.sql` (226 lines) wraps its entire body in two T-SQL existence guards:

```sql
IF OBJECT_ID('[${flyway:defaultSchema}].[vwPeople]', 'V') IS NOT NULL
    DROP VIEW [${flyway:defaultSchema}].[vwPeople];
GO
CREATE VIEW [${flyway:defaultSchema}].[vwPeople] AS SELECT ... ;
GO
-- (same again for vwOrganizations)
```

Converting with `--split --allow-gaps` reports `2 statement(s)` unhandled (kind `IF-BLOCK`) and writes a 1,183-byte `.pg.sql` whose complete executable content is:

```sql
SET standard_conforming_strings = on;
```

…followed by the gap banner. Both views the migration exists to create are absent, with nothing in the file to indicate the schema change was lost.

Note the `IF OBJECT_ID(...) IS NOT NULL DROP VIEW` idiom is itself trivially expressible (`DROP VIEW IF EXISTS … ;`), so this particular file is also an argument for templating that pattern — the same ask #3166 makes for `IF EXISTS (SELECT 1 FROM sys.indexes …) DROP INDEX`.

## Ask

When a migration's converted body ends up with **no executable statements**, do not write a `.pg.sql`. Write `.needs-hand` (the mechanism already exists) or fail the run, regardless of `--allow-gaps`. Optionally emit a summary line distinguishing "partially converted" from "nothing converted".

## Related

- #3166 — `IF EXISTS (… sys.indexes …)` guards dropped as unhandled gaps; templating them was already requested there
- #3345 — other convert gaps, including guards dropped from the body and listed only in the header

MS-BC:
Closes #3840

## The problem

When the AST dialect cannot emit **any** statement in a migration, the converted body is a header and a gap banner over nothing — and the result was still returned as `converted`, so the CLI wrote it as a discoverable `.pg.sql`.

That is the worst artifact this pipeline can produce, because every cheap signal says it is fine:

- it has a `.pg.sql` name, so **filename parity passes**
- it contains no T-SQL, so a **contamination scan passes**
- it **applies to PostgreSQL without error**
- it does nothing

Skyway records the migration as applied and the schema change never happens, on every PostgreSQL tenant.

## The fix

`emittedStatements === 0` arises three ways. Two were already guarded; the third was not:

| Case | Meaning | Before | After |
|---|---|---|---|
| no gaps, no drops | content **vanished** | `suspiciousEmptyOutput` gap, still `converted` | gap **and** `needs-hand-authoring` |
| drops > 0 | legitimately empty (e.g. actionless `ALTER`) | stays `converted` | unchanged |
| **gaps > 0** | **everything became a gap** | **`converted` → hollow `.pg.sql`** | **`needs-hand-authoring` → `.needs-hand`** |

The vanish row changed too, and deliberately: promoting it is right for the same reason — the empty-marker path already promotes on `suspiciousEmptyOutput` alone, and an empty body must never be written as a discoverable `.pg.sql` whichever way it got empty. The two promotions are computed separately (the hollow check keys on `transpiled.unhandled`, before the vanish guard's synthetic row is appended) so neither can fold into the other or borrow its wording.

Promotion routes it through the same path the empty-marker case above it already uses, for the reason that code already states: never a discoverable `.pg.sql` that Skyway could apply. The transpiled artifact is still written under the `.needs-hand` name, so gap comments and any partial DDL survive for whoever finishes the port.

Note this already behaved correctly under `--bake-codegen` (`decideConvertWrite` forces `.needs-hand` on a `gap-no-bake` result). The hole was the plain `--split` / `--allow-gaps` path, which is what release conversion uses.

## How it was found

`bizapps-common`'s `V202608132240__v5.34.x__Layered_Base_Views_People_Organizations.sql` wraps its entire body in two `IF OBJECT_ID(...) IS NOT NULL DROP VIEW` guards. 226 source lines converted to four setup statements plus a banner; both views the migration exists to create were silently absent, and the file looked like a completed port.

## Tests

Four cases: fully-gapped promotes; vanished promotes but is **not** described as gapped; partially-gapped does **not** (that is the ordinary `--allow-gaps` case, and promoting it would make the flag meaningless); all-dropped does **not**.

Full package suite: 1161 passed, 0 failed. Typecheck clean.


MS-BC:
Self-review found one inaccuracy in this PR, now fixed in `c37748c`.

**The table in the description was wrong about the vanish row.** The hollow check read the *local* `unhandled`, which the vanish guard immediately above has already appended its synthetic `RECONCILIATION-EMPTY-OUTPUT` row to. So a VANISHED conversion — nothing emitted, nothing reported, nothing dropped — satisfied `unhandled.length > 0` and was promoted as though every statement had become a gap, collecting a note that misdescribes it: nothing became a gap there, the reconciliation layer manufactured one.

Verified rather than reasoned about — same input, both branches:

| | `next` | this branch, before `c37748c` |
|---|---|---|
| vanished conversion | `converted` | `needs-hand-authoring` |

So the row that read *"no gaps, no drops → unchanged"* was not unchanged.

**Promoting it is still the right outcome** — the empty-marker path already promotes on `suspiciousEmptyOutput` alone, and an empty body must never be written as a discoverable `.pg.sql` whichever way it got empty. What was wrong is that it happened incidentally, undocumented and untested. So it is now stated on its own terms: the hollow check keys on `transpiled.unhandled`, and the status expression names both promotions. The existing vanish test pins the status and asserts it does **not** carry the gapped-conversion note, so the two cannot silently merge again.

Changeset and the table above both corrected. Package suite 1161 passed.

PR #3863, opened 2026-08-16
MJDropdown gains an accessible name: AriaLabel, InputId, AriaDescribedBy (#3860)

### LINKED ISSUE #3860: MJDropdown has no ariaLabel/inputId passthrough — every dropdown is an unnamed combobox (WCAG 4.1.2)
Filed from **Caliber** (`MemberJunction/bizapps-caliber`), an Open App on MJ. Caliber tracks this as MemberJunction/bizapps-caliber#17.

## The problem

`MJDropdownComponent` in `@memberjunction/ng-ui-components` exposes no `ariaLabel` or `inputId` input, and renders no `<label>` association of its own. A consuming app therefore has **no supported way to name the control**, so every dropdown announces to a screen reader as an unnamed combobox — "combobox, collapsed" with no indication of what it selects.

This is not cosmetic: an unnamed form control fails WCAG 2.1 **4.1.2 Name, Role, Value** (Level A). For Caliber it is the single shared blocker for screen-reader sign-off across the whole Studio and Review surface, because the same component is used at every selection point.

## What a consumer can do today

Nothing supported. `aria-label` on the host element does not reach the inner combobox, and wrapping in a `<label>` does not associate without a matching `id` on the control MJ renders.

## The ask

Two passthrough inputs on `MJDropdownComponent`:

- **`ariaLabel?: string`** — applied as `aria-label` on the element carrying `role="combobox"`.
- **`inputId?: string`** — applied as the `id` of that same element, so a host can associate an external `<label for="…">` and get a visible label rather than only an accessible name.

Both optional, so nothing changes for existing callers. If `ariaLabel` is absent and an `inputId` is present, the external label is the name; if both are absent, behaviour is exactly as today.

Worth considering alongside: `aria-describedby` passthrough for hint text, which has the same shape of problem, and a dev-mode console warning when a dropdown renders with neither a label nor an `aria-label` — the defect is silent otherwise, which is why it survived this long.

Happy to open the PR if the input names are agreed.

MS-BC:
Closes #3860.

## The defect

`mj-dropdown` renders a `div[role="combobox"]` with **no way to give it an accessible name**. Every one of the ~94 call sites in this repo announces as "combobox, collapsed" — role and state, no name — which fails WCAG 2.1 4.1.2 (Name, Role, Value). Nothing in the API let a caller fix it.

## The four inputs

All optional, all applied to the popup listbox as well as the trigger so both halves announce the same name.

| input | for |
|---|---|
| **`AriaLabelledBy`** | the id of a **visible** label — the preferred wiring when one exists |
| `AriaLabel` | when no visible label exists |
| `AriaDescribedBy` | hint and error text |
| `InputId` | an id on the trigger, so other markup can reference it |

**`AriaLabelledBy`, not `<label for>`** — and this is the one design decision that departs from the issue. The issue proposed `InputId` as a `label[for]` target. That cannot work: the trigger is a `div`, and `label[for]` only names and focuses *labelable form elements*. Wiring it produces markup that looks correct, passes review, and still leaves the control unnamed. So the visible-label path is an id on the **label** passed as `AriaLabelledBy`, and `InputId`'s JSDoc says explicitly that it is not a label target.

**Absent beats empty.** None of the four renders an attribute when unset, pinned by a test: `aria-label=""` is worse than no attribute, because it overrides every other naming source with an explicitly empty name.

## Naming the filter box

A filterable panel contains a second control. It takes its name from the same source rather than being a second unnamed box — and under `AriaLabelledBy` it does so through an `aria-labelledby` **id list** composing "Filter" with the visible label's own text. That is the only mechanism that can work there: the component never sees that text, only an id. Without it, `ai-prompt-form`'s six filterable dropdowns all announce as "Filter options".

A name that already begins with "Filter" is not prefixed again — `AriaLabel="Filter roles"` is this repo's house habit, so "Filter Filter roles" was a real regression rather than a hypothetical one.

## Also in the same attribute cluster

- **`aria-controls`** now points the trigger at a generated listbox id while open. `aria-expanded` alone says *something* expanded without saying what. This uses the `DropdownId` counter that was dead repo-wide — the same `static nextId` mechanism `dialog`, `window` and `accordion` already use in this package.
- **`aria-disabled` and a dynamic `tabindex`.** `tabindex` was static, and the SCSS suppresses the focus ring when disabled — so a keyboard user landed on something invisible that then silently ignored Enter.

## One deliberate visible change

The filter box's placeholder changes from **"Search..." to "Filter..."**. The accessible name is "Filter &lt;name&gt;", and a visible "Search" that is not in the accessible name breaks WCAG 2.5.3 (Label in Name): a voice-control user says "click Search" and nothing matches. This is the only behavioural change for an existing caller.

## Test-utils

`StubDropdownComponent` gains the same four inputs, keeping its "mirrors the real inputs" contract true. The DOM test environment sets `errorOnUnknownProperties`, so without it the first consumer spec binding `[AriaLabel]` on a stubbed dropdown throws NG0303 — and a static `AriaLabel="…"` attribute is worse, landing silently as a vacuous pass. Seven specs use this stub.

## Docs

The README catalog entry and the component's own `@example` now show a **named** dropdown. The example previously sat a hundred lines above a doc block describing that exact usage as a WCAG failure.

## Tests

**19 in the dropdown spec (7 new), 565 in the package.** Covering: `AriaLabel` on trigger and listbox, `AriaLabelledBy` on both, `InputId`, `AriaDescribedBy`, absent-beats-empty, the filter named from `AriaLabel`, the filter named from a visible label via the id list, no "Filter Filter roles" doubling, the placeholder inside the accessible name, `aria-controls` while open, and the disabled/tabbable pair.

Mutation-verified — four mutants, all killed: dropping `aria-controls` · dropping the disabled exposure · letting the filter fall back to a generic name under `AriaLabelledBy` · prefixing "Filter" unconditionally.

## Follow-ups filed

- **#4114** — adoption sweep. None of the ~94 call sites bind the new inputs yet, so this closes #3860 for the API, not for any screen. Includes the `create-prompt-dialog` migration hazard (a dead `label[for]` + host `id` pairing that pattern-matched adoption turns into duplicate ids).
- **#4115** — the rest of the APG combobox pattern: the filter input inside `role="listbox"` (axe `aria-required-children`, critical), missing `aria-activedescendant`, an overlay-aware axe gate, plus two pre-existing `Close()` bugs.
- **#4116** — sibling parity (`mj-combobox`, `mj-switch`, `mj-numeric-input`, `mj-datepicker`, `mj-page-search`) and a recorded decision on the issue's proposed dev-mode unnamed-control warning.


MattC-BC (changes requested):
Strong fix at its core, and the second commit made the right engineering call: `AriaLabelledBy` as the visible-label path (instead of the issue's proposed `label[for]`, which cannot name or focus a `div` trigger), the absent-beats-empty guarantee pinned by a test, and input naming that matches the package's existing `AriaLabel` precedents (workspace-card, bottom-sheet, page-header-interior). Requesting changes on a set of completeness items below; none of them are deep, and most are small edits in files this PR already touches or sits next to.

**Verified independently on the PR head (`a893ace`):** clean install, `pnpm test` in `packages/Angular/Generic/ui-components` passes 52 files / 559 tests including the 6 new ones; package build and typecheck clean; branch is current with `next`. The accessible-name and ARIA-structure claims below were checked empirically with the repo-pinned axe-core 4.11.1 against the rendered panel markup, not just by reading the code.

## Must fix before merge

1. **Missing changeset.** The PR changes `@memberjunction/ng-ui-components` source with no `.changeset` entry. Per `.claude/rules/changesets.md` ("`patch` | Everything else — TypeScript, tests...") this needs a `patch` changeset. The `fixed` group means the package will still publish off other pending changesets, but without one there is no CHANGELOG or release-notes text for this fix, so Caliber (the consumer this exists for) gets no signal which `edge.N` carries the new inputs.

2. **`StubDropdownComponent` not extended** (`packages/Angular/Generic/test-utils/src/lib/stubs.ts:88`). Its doc says "Mirrors the real inputs" and it lists exactly the old 8. The DOM test environment sets `errorOnUnknownProperties: true`, so the first consumer spec whose template binds `[AriaLabel]` on a stubbed `mj-dropdown` throws NG0303, and a static `AriaLabel="text"` attribute lands silently as a vacuous pass. Seven specs use this stub today (APIKeys, MCP x2, Actions x2, core-entity-forms x2). Four one-line `@Input()` additions keep the mirror contract true.

3. **The PR description contradicts the shipped code.** The body still says "three optional passthroughs", documents `InputId` as the wiring for "a visible `<label for>`... preferred", and says "5 new" tests. The code ships four inputs, the `InputId` JSDoc explicitly says `label[for]` neither names nor focuses the div trigger, `AriaLabelledBy` is the visible-label path, and there are 6 new tests. Anyone following the merged description will wire the non-functional pattern and ship the exact bug this PR fixes. Please update the body to match commit `3b1a888`'s model.

4. **The filter box is silently renamed for every existing `Filterable` caller** (`dropdown.component.ts:87`). The new `aria-label` is unconditional, so the filter input's accessible name changes from "Search..." (placeholder-derived) to "Filter options" while the visible placeholder still says "Search...". That contradicts the "no behavioural change for any existing caller" claim and creates a visible-text vs accessible-name mismatch (WCAG 2.5.3 territory: a voice-control user says "click Search" and nothing matches). Fix by aligning the strings, e.g. `placeholder="Filter..."` or a name that contains the visible text.

5. **The documented-preferred `AriaLabelledBy` path leaves the filter box generically named** (`dropdown.component.ts:87`). The filter name derives only from `AriaLabel`; with `AriaLabelledBy` + `Filterable` it falls back to "Filter options". Roughly 52 of the ~94 call sites in this repo have a visible label nearby, i.e. exactly the population the JSDoc steers to `AriaLabelledBy`, and forms like `ai-prompt-form` have 6 filterable dropdowns, which would announce as 6 identical "Filter options" boxes. There is also no test for the `Filterable` + `AriaLabelledBy` combination. A hidden "Filter" prefix span plus an id-list `aria-labelledby` would carry the visible label's text; at minimum, document the tradeoff and add the test. Related nit on the same line: the house naming habit (`AriaLabel="Filter roles"` in explorer-settings) produces "Filter Filter roles".

6. **The docs consumers actually copy from still show the unnamed usage.** The README catalog entry (`packages/Angular/Generic/ui-components/README.md:79`) was not updated (sibling entries like `mj-bottom-sheet` do document `AriaLabel`, and `packages/Angular/CLAUDE.md` calls this README the canonical catalog), and the component's own `@example` JSDoc (`dropdown.component.ts:25-36`) still demonstrates a dropdown with no name, one hundred lines above a doc block calling that a WCAG failure. Two small edits.

## Worth doing in this PR (same attribute block, cheap)

7. **Wire `aria-controls` and give the listbox an id.** The trigger has `aria-haspopup="listbox"` and `aria-expanded` but no `aria-controls`, and the panel has no id and no input to set one, so the trigger-to-popup association is unwirable by anyone. The per-instance counter for this already exists and is dead: `DropdownId` (`dropdown.component.ts:182`) has zero references repo-wide, while `dialog`, `window`, and `accordion` in this same package use the identical `static nextId` mechanism to self-wire generated-id aria relationships. A generated `id="mj-dropdown-listbox-{n}"` plus `[attr.aria-controls]` when open follows the in-package precedent; alternatively delete the dead field. Side note: the `InputId` JSDoc cites `aria-controls` as a use for the trigger id, which is the backwards direction for the combobox pattern.

8. **Expose the disabled state.** No `aria-disabled` is rendered, `tabindex="0"` is static so a disabled dropdown stays tabbable, and the existing SCSS suppresses the focus ring when disabled. A keyboard user lands on it with no ring, hears no "unavailable", and Enter/ArrowDown silently no-op. One more binding in the cluster this PR already edits.

## Follow-ups to file (so they don't evaporate when #3860 closes)

9. **Adoption is zero.** None of the ~94 `<mj-dropdown` call sites in this repo bind any of the new inputs, so no rendered page improves on merge day; "Closes #3860" is true of the API, not of any screen. The cheapest first proof lives in this very package: `filter-panel.component.ts:86` renders dropdowns with the required `field.label` in hand and forwards nothing, while `mj-filter-field` wraps them in a `<label>` element that cannot name a div combobox. `[AriaLabel]="field.label"` is a one-liner. An adoption sweep issue should own the rest.

10. **The issue's dev-mode warning needs a recorded decision.** #3860 suggested warning when a dropdown renders unnamed ("the defect is silent otherwise, which is why it survived this long"). The PR neither implements nor declines it; without some feedback loop, every new dropdown added tomorrow ships unnamed and silent again.

11. **Sibling parity.** `mj-combobox` (7 consumer files) shares the same `role="combobox"` markup and gained none of the naming inputs; its toggle button also has no `aria-label` at all. Four more controls in this lib ship with no accessible-name input: `mj-switch`, `mj-numeric-input`, `mj-datepicker`, `mj-page-search`. Worth one tracked issue enumerating them so each doesn't get its own independently-invented API later.

12. **Structural listbox issue, now decorated.** The filter input (and the no-data div) live directly inside `role="listbox"`, whose only permitted children are option/group. The repo-pinned axe-core fires `aria-required-children` at critical severity on the open panel, so the first page-level scan with an open filterable dropdown fails. Canonical fix per the APG combobox pattern is moving the filter outside the listbox element. Related: options have no ids and there is no `aria-activedescendant`, so arrow-key highlight is still inaudible to screen readers even after the naming lands.

13. **`AriaDescribedBy` is unassociated during the filterable open state.** Opening moves focus into the filter input, which carries no describedby, so hint/error text is not announced exactly when the user is typing and choosing. Low severity; propagate or document.

14. **The new tests never invoke the a11y gate.** All six assert attribute strings (they would pass if `role="combobox"` were deleted), and `ExpectNoAxeViolations` from ng-test-utils has zero adopters; it also scans `fixture.nativeElement`, which can never see the CDK overlay panel. Adopting it here with an overlay-aware scan root (and an A11Y-DEBT waiver for item 12) would make this the reference a11y spec.

15. **Adoption hazard at existing call sites.** `create-prompt-dialog.component.html` (lines 47, 71, 86) already pairs `<label for="x">` with `id="x"` on the `mj-dropdown` host, a dead association today. Pattern-matched adoption (adding `InputId="x"` without removing the host id) creates duplicate ids and still no name. Worth migrating those three as the worked example in the sweep: id moves to the label element, passed via `AriaLabelledBy`.

## Pre-existing bugs noticed during review (separate from this PR, being tracked)

- `Close()` never restores focus to the trigger; selecting from the filter input drops focus to `<body>` when the overlay detaches.
- `Close()` clears `filterText` without emitting `FilterChange`, so hosts that narrow `[Data]` externally (live example: `mcp-test-tool-dialog`, three dropdowns) reopen to a stale narrowed list with an empty filter box.

---
*Review conducted with [Claude Code](https://claude.com/claude-code).*

MS-BC:
Thanks @MattC-BC — that's a genuinely useful review, and the empirical checks (axe against the rendered panel, not just reading the code) caught things I would not have found by inspection. All six must-fix items are done, both "worth doing" items are in, and 9–15 are filed.

## Must fix

**1. Changeset** — added, `patch` on `@memberjunction/ng-ui-components` and `@memberjunction/ng-test-utils`. You're right that the `fixed` group hides the consequence: the package publishes either way, so the missing piece was the release-notes text Caliber needs to know which `edge.N` carries the inputs.

**2. `StubDropdownComponent`** — four inputs added, and the doc now says *why* the mirror has to stay true rather than just claiming it is. Your framing of the failure modes is what I put in the comment: NG0303 under `errorOnUnknownProperties` for a bound `[AriaLabel]`, and the worse case of a static attribute landing as a vacuous pass.

**3. PR description** — rewritten against `3b1a888`'s model. You were right that this was the most dangerous item on the list: anyone following the merged description would have wired `label[for]` and shipped the exact bug this PR fixes.

**4. Filter box silently renamed** — fixed, and I took your placeholder suggestion. Placeholder is now `"Filter..."` so the visible text sits inside the accessible name. It *is* a visible change for existing `Filterable` callers, so I've called it out explicitly in the body and changeset rather than leaving the "no behavioural change" claim standing.

**5. `AriaLabelledBy` + `Filterable`** — this was the best catch in the review, and the one I'd have shipped wrong. Fixed properly rather than documented away: a hidden "Filter" span plus an `aria-labelledby` **id list**, exactly as you sketched. It's the only mechanism that can work, since the component never sees the label's text — only its id. There's now a test for the combination, and one for your naming-habit nit: `AriaLabel="Filter roles"` no longer becomes "Filter Filter roles".

**6. Docs** — README catalog entry and the component `@example` both show a named dropdown now. The example sitting a hundred lines above a doc block calling that usage a WCAG failure was a fair hit.

## Worth doing — both in

**7. `aria-controls` + listbox id.** Done, using the dead `DropdownId` counter rather than deleting it — and you're right that `dialog`/`window`/`accordion` in this same package already establish the pattern. Also fixed the `InputId` JSDoc, which cited `aria-controls` in the backwards direction.

**8. Disabled state.** `aria-disabled` plus a dynamic `tabindex`. Your description of the actual experience — no ring because the SCSS suppresses it, no "unavailable", Enter silently no-ops — is what's now in the test's comment.

7 new specs (19 in the file, 565 in the package). Mutation-verified: dropping `aria-controls`, dropping the disabled exposure, letting the filter fall back to a generic name under `AriaLabelledBy`, and prefixing "Filter" unconditionally each fail a test.

## Follow-ups filed

- **#4114** — adoption sweep, led with your `filter-panel.component.ts:86` one-liner as the cheapest first proof, and the `create-prompt-dialog` lines 47/71/86 migration as the worked example including the duplicate-id hazard you flagged.
- **#4115** — the APG structural work: filter inside `role="listbox"` (axe `aria-required-children`, critical), missing `aria-activedescendant`, `AriaDescribedBy` during the open state, adopting `ExpectNoAxeViolations` with an overlay-aware scan root, plus the two pre-existing `Close()` bugs you found (focus not restored; `filterText` cleared without emitting `FilterChange`).
- **#4116** — sibling parity for `mj-combobox` and the four other unnamed controls, plus a recorded decision on the dev-mode warning from #3860, so it doesn't evaporate when that issue closes.

Your point about the tests only asserting attribute strings is right and I haven't fixed it here — an overlay-aware axe gate needs the waiver for item 12 to exist first, so it's the headline of #4115 rather than something I could land cleanly in this PR.

MattC-BC (approved):

MattC-BC:
@MS-BC - All fifteen findings are now dispositioned in the panel: 8 fixed, 7 deliberately deferred with tracking. This is a genuinely excellent revision.

PR #3864, opened 2026-08-16
fix(sql-converter): resolve ${mjSchema}, not just ${flyway:defaultSchema}

### LINKED ISSUE #3838: mj migrate convert --bake-codegen applies migrations without Flyway placeholder substitution
## One sentence

`mj migrate convert --split --bake-codegen` applies each converted migration to the working database **without substituting Flyway placeholders**, so any migration referencing `${mjSchema}` halts the bake chain with `relation "${mjSchema}.Entity" does not exist` — even though `mj migrate` substitutes the same placeholder correctly and committed `.pg.sql` files rely on it.

## Environment

- `@memberjunction/cli@6.1.0-edge.2`
- PostgreSQL 17.10, DB migrated to `v6.1.0-edge.2`, app `bizapps-common` (`__mj_bizappscommon`)

## Repro

Convert a migration whose body references core as `${mjSchema}` (the normal convention — `bizapps-common`'s SQL Server migrations contain 541 such references):

```
mj migrate convert --split --bake-codegen \
  --source-dir <src> --output-dir <out> --schema __mj_bizappscommon
```

**Observed:**

```
Halted at V202608132239__v5.34.x__Layered_Base_Views_Metadata.sql:
  relation "${mjSchema}.Entity" does not exist.
```

The generated `.pg.sql` contains the placeholder literally:

```sql
UPDATE "${mjSchema}"."Entity" SET "BaseViewGenerated" = 0, ...
```

**Expected:** the inline apply resolves placeholders the same way `mj migrate` does. `mj migrate` clearly does resolve it — committed counterparts in the fleet carry `${mjSchema}` and apply cleanly on real installs.

## Why it is more than cosmetic

`--bake-codegen` is documented as the path that produces a set deployable "with `mj migrate` alone — no `mj codegen` step", which is exactly what an Open App needs, since consumers never run codegen. Because the converter halts at the first failure and does not advance the working DB, one unresolved placeholder stops the whole chain: every later migration in the run goes unconverted, and any file already written is left without its baked CodeGen objects.

## Workaround

Pre-substituting `${mjSchema}` → `__mj` in a scratch copy of the source before conversion. That produces literal `"__mj"` in the output, which is also the dominant committed convention (~527 literal references against a dozen surviving placeholders in `bizapps-common`'s `migrations-pg/`) — so the workaround is harmless, but it should not be necessary.

Worth deciding explicitly which form the converter *should* emit, since both appear in committed files today.

## Related

- #3345 — `mj migrate convert`: four gaps producing the v5.50 PG counterparts (same command, different defects)
- #3443 — Open App manifests cannot declare the migration placeholders their migrations require

MS-BC:
Closes #3838

## The problem

An Open App migration names its **own** schema with `${flyway:defaultSchema}` and **MJ core** with `${mjSchema}`. Only the first was substituted.

Because `--bake-codegen` executes each converted migration against the working database as it advances, the unresolved macro was not cosmetic — it went to the server:

```
relation "${mjSchema}.Entity" does not exist
```

The converter halts at the first failure and does not advance the working DB, so **one unresolved macro stopped the entire bake chain**: every later migration went unconverted, and files already written lacked their baked CodeGen objects.

## The fix

The two placeholders cannot share a value — for every app they differ (`__mj_bizappscommon` vs `__mj`). So:

- `coreSchema` option on `convertMigration` and `IncrementalBaker` (default `__mj`)
- `--core-schema` CLI flag
- substitution happens exactly where `${flyway:defaultSchema}` is already resolved, keeping committed files fully literal — which that code's own comment states is the intent: *"Committed .pg.sql files use the literal schema (Flyway placeholder substitution is not relied on for PG)."*

## Verification

**Unit:** two tests — that the app schema and core schema resolve to *different* values in one file, and that `${mjSchema}` defaults to `__mj`.

**End-to-end on a throwaway PostgreSQL 17 database.** A migration touching both schemas, converted with the real `MJPostgresTranspiler`:

```sql
CREATE TABLE __mj_bizappscommon."Widget" ( ... );
UPDATE "__mj"."Entity" SET "Name" = 'new' WHERE "ID" = '1';
```

- unresolved `${mjSchema}` occurrences: **0**
- applies with `ON_ERROR_STOP=1`: **exit 0**
- core row updated in `__mj`, app table created in `__mj_bizappscommon`

Package suite: 1147 passed. Typecheck clean.


packages/SQLConverter/src/__tests__/MigrationConverter.test.ts
```diff
@@ -79,6 +79,37 @@ describe('convertMigration — reconciliation (issue #3252 Phase 3)', () => {
...
+    // …and core resolves to core, NOT to the app schema.
+    expect(r.pgSQL).not.toContain('__mj_bizappscommon"."Entity"');
+    expect(r.pgSQL).toMatch(/__mj["\].]*\.?\[?"?Entity/);
+  });
+
+  it('defaults ${mjSchema} to __mj when no coreSchema is supplied', async () => {
+    const sql = "UPDATE [${mjSchema}].[Entity] SET [Name] = 'W' WHERE [ID] = '1';\nGO\nCREATE TABLE [x].[Y] ( [ID] INT );";
+    const r = await convertMigration(sql, 'V_Widget.sql', { transpiler: passthrough, schema: '__mj_app' });
+    expect(r.pgSQL).not.toContain('${mjSchema}');
+    expect(r.pgSQL).toContain('__mj');
```
rkihm-BC: **Requested before merge.** This assertion is satisfied by `pgHeader()` alone — `includeHeader` defaults to `true`, and the header contains `CREATE SCHEMA IF NOT EXISTS __mj;` and `SET search_path TO __mj, public;`. So it passes regardless of what `${mjSchema}` resolved to, or whether that token was substituted at all.

Even without the header it couldn't distinguish: `'__mj_app'.includes('__mj')` is true, so a default that fell back to `schema` — the exact bug this PR fixes, reached via the default path — would still pass.

The test one above already has the right shape (`not.toContain('__mj_bizappscommon\".\"Entity')`). Same idea here:

```suggestion
    expect(r.pgSQL).not.toContain('${mjSchema}');
    // Core resolves to __mj, NOT to the app schema. Asserting toContain('__mj') alone would be
    // vacuous: pgHeader() already emits `CREATE SCHEMA IF NOT EXISTS __mj;`, and '__mj_app'
    // contains '__mj' as a substring either way.
    expect(r.pgSQL).not.toContain('__mj_app\".\"Entity\"');
    expect(r.pgSQL).toMatch(/"__mj"\."Entity"/);
```

(Adjust the exact quoting to whatever `passthrough` emits — the point is a negative on the wrong schema plus a positive on the right one, rather than a substring that both satisfy.)

rkihm-BC (commented):
Reviewed the full diff and verified the placeholder convention independently. The fix is correct — one test doesn't test what it claims, and there's a completeness gap worth deciding on.

## What's good

**The defect is real and the placeholder is a genuine convention.** `${mjSchema}` has **zero** occurrences in this repo's `migrations/`, which initially looked like a problem — it's used by Open App migrations in separate repos. The convention is documented in `CodeGenLib/src/Config/config.ts` (`{ schema: '__mj', placeholder: '${mjSchema}' }`) and exercised in `ConditionalDDLRule.test.ts`, where `[${mjSchema}_BizAppsCommon]` is described as how an open app names a sibling app's schema.

**That compositional use is handled correctly.** `${mjSchema}` is a *prefix* — `${mjSchema}_bizappscommon` has to become `__mj_bizappscommon`. A plain `replaceAll` on the token does exactly that. I also confirmed the existing `ConditionalDDLRule` tests still hold, because they assert at rule level, upstream of `assemblePgSQL` where this substitution happens — so there's no conflict with the tests that deliberately expect the placeholder to survive.

**The `replaceAll` function form is the right instinct, and correctly credited.** Using `() => schema` rather than a string means `$$`, `$&`, `` $` ``, `$'` and `$1`–`$99` in a schema name can't be read as replacement patterns. Citing issue #3171 — the same `$`-expansion class fixed in #3769 — and converting the two *pre-existing* string replacements at the same time is good discipline; that's a lesson applied rather than a bug re-hit.

**The CLI wiring is complete and defaulted.** `--core-schema` with `default: '__mj'`, threaded through `convert.ts` → `convertMigration` and `buildBaker` → `IncrementalBaker`. The third commit catching `migrate rebake` matters — it's the less-travelled entry point and the easy one to miss.

**The belt-and-braces substitution in `IncrementalBaker.assemble()` was updated too**, not just the converter, so a macro leaking through native capture is still caught.

Changeset `patch`, no migration or metadata — correct. MERGEABLE, all checks green.

## Requested before merge — the default-value test cannot fail

`MigrationConverter.test.ts:106-111`:

```js
it('defaults ${mjSchema} to __mj when no coreSchema is supplied', async () => {
  const sql = "UPDATE [${mjSchema}].[Entity] SET [Name] = 'W' WHERE [ID] = '1';\n...";
  const r = await convertMigration(sql, 'V_Widget.sql', { transpiler: passthrough, schema: '__mj_app' });
  expect(r.pgSQL).not.toContain('${mjSchema}');
  expect(r.pgSQL).toContain('__mj');
});
```

`includeHeader` defaults to `true` and the test doesn't disable it, so `pgHeader()` is in the output — and it contains:

```
```

So `expect(r.pgSQL).toContain('__mj')` is **satisfied by the header alone**, regardless of what `${mjSchema}` resolved to — or whether the substitution ran at all on that token.

The first assertion does real work (it proves *something* replaced the placeholder), but nothing in the test checks the value is `__mj` rather than the app schema. Worth noting `'__mj_app'.includes('__mj')` is also true, so even without the header the assertion couldn't distinguish a correct default from one that fell back to `schema` — which is the precise bug this PR fixes, reached via the default path.

The fix is already demonstrated one test above: the sibling case asserts `not.toContain('__mj_bizappscommon"."Entity"')` — a negative on the *wrong* schema qualifying the object. The same shape here makes it real.

## Worth deciding on — this closes the instance, not the class

`schemaPlaceholders` in CodeGen config is a user-configurable `z.array({schema, placeholder})` — arbitrary length, arbitrary names. `MJCLI/src/config.ts:305-318` builds the runtime placeholder map from it, stripping `${}` and registering *any* name, with the comment "supports BCSaaS and other extensions". `sql_logging.ts:236` applies the same list when emitting migrations.

So the migration **runner** supports N placeholders while the **converter** now handles exactly two hardcoded names. An operator who configures a third — say `{ schema: '__BCSaaS', placeholder: '${saasSchema}' }` — gets the identical failure this PR fixes: an unresolved macro reaching the server, halting the bake chain, with `relation "${saasSchema}.X" does not exist`.

The cheap complete fix isn't generalising the substitution — it's **asserting the postcondition**. After substitution, scan the output for any surviving `${...}` and fail loudly naming the placeholder. That turns every future placeholder from silent corruption into a clear error, and it's exactly the property your own e2e verification checked by hand (*"unresolved `${mjSchema}` occurrences: 0"*) — just not enforced in code. I confirmed no such guard exists in either `MigrationConverter` or `IncrementalBaker` today.

Your call whether that belongs here or in a follow-up; the two-placeholder fix is correct and complete for the reported bug either way.

## Bottom line

The fix is right and the CLI wiring is thorough. Just the test — it's misleading enough that I'd rather it were fixed in place than carried as debt, since its name says it verifies the default and it can't.

packages/SQLConverter/src/__tests__/MigrationConverter.test.ts
MS-BC: Fixed in `094cf02` — you were right on both counts, and the second one is the sharper of the two: `'__mj_app'.includes('__mj')` means the assertion could not have distinguished the default from the fallback bug even without `pgHeader()` in the picture.

Now asserts on the qualified reference rather than a substring — negative on the wrong schema, positive on the right one:

```js
expect(r.pgSQL).not.toContain('[__mj_app].[Entity]');
expect(r.pgSQL).toContain('[__mj].[Entity]');
```

And checked it can actually fail, rather than assuming: reintroducing the exact defect (`options.coreSchema ?? schema`) turns the test red, and reverting turns it green again. Package suite 1147 passed.

Issue #3514, opened 2026-08-06
CodeGen: SS migration artifacts and the PG bake regenerate from different metadata states — same migration, diverging function signatures


cadam11:
## One sentence

SQL Server migration artifacts are CodeGen'd from the author's **live dev database** (metadata *ahead* of the committed chain), while the PostgreSQL bake regenerates functions from the **chain's replay state** (metadata *behind*) — so the two sides of the same migration can disagree about a function's parameter list, and the disagreement only surfaces when a later metadata sync calls the parameters the PG side never grew.

## The receipt (6.1.0-edge.0 release night, 2026-08-05)

- SS `V202608050100` ships `spUpdateEntity` with **91 parameters**, including `@ExternalDataSourceID` / `@ExternalObjectName`.
- The PG bake of the *same migration* produced **87 parameters** — without them.
- The replayed chain contains **zero** `EntityField` rows for those two columns: the metadata that would describe them never rides any committed migration, which is exactly what starved the PG generator.
- Failure mode: the release's metadata-sync migration then calls `spUpdateEntity` with the missing parameters → the PG counterpart fails to apply on a clean database. Ten call sites needed a strip-args + direct-`UPDATE` repair to ship the counterpart.

## Why it recurs

Any time an author's dev DB carries schema/metadata ahead of the committed migration chain (a normal state during development), SS CodeGen bakes that drift into the SS artifact. The PG conversion pipeline deliberately replays only committed history — the correct choice for reproducibility — so every such sync produces a silent SS↔PG signature skew until a later call trips over it.

## Directions worth evaluating (not mutually exclusive)

1. **Parity of inputs:** generate SS artifacts from chain-replay state too (clean-room CodeGen), so both sides see identical metadata. Strongest fix; changes the authoring workflow.
2. **Conversion-time assertion:** when baking PG, cross-check every parameter referenced by the migration's sync statements against the replay-state function signatures and fail loudly with the missing-metadata diagnosis, instead of shipping a signature that will fail later.
3. **Author-side gate:** detect "dev DB metadata not represented in the committed chain" before a sync migration is committed (the drift itself is the defect from the chain's point of view).

Found while producing the PG counterpart for the first Edge release; the repair pattern and diagnosis recipe are documented in the release-engineering material.

🤖 Filed with [Claude Code](https://claude.com/claude-code) on Craig's approval (review queue, 2026-08-06)

Issue #3477, opened 2026-08-05
Layered base views: platform-neutral metadata enables a SQL-Server-only feature, and PG CodeGen throws on it


cadam11:
## One sentence

`metadata/entities/.layered-base-views.json` sets `GeneratedBaseViewName` on two core entities and ships through the normal metadata tree to every install, but `PostgreSQLCodeGenProvider` throws by design when any entity has that field set — so a PostgreSQL install that receives this metadata fails on its next `mj codegen`.

## The two halves

**The metadata is platform-neutral and ships everywhere.** `metadata/entities/.layered-base-views.json` lives inside the normally-pushed `metadata/` tree (`entities` is in `.mj-sync.json`'s `directoryOrder`), so `mj sync push` applies it and the next `Metadata_Sync` migration carries it to all installations. It sets:

| Entity | GeneratedBaseViewName |
|---|---|
| MJ: Version Installations | `vwVersionInstallationsGenerated` |
| MJ: User View Run Details | `vwUserViewRunDetailsGenerated` |

**The feature is SQL-Server-only, enforced by a hard throw.** `packages/CodeGenLib/src/Database/providers/postgresql/PostgreSQLCodeGenProvider.ts` → `assertLayeredBaseViewSupported()`:

> `Entity "..." sets GeneratedBaseViewName = '...', but layered base views are not supported on PostgreSQL. PostgreSQL freezes a view's column list at creation and has no sp_refreshview equivalent, so the application-owned view "..." would silently stop gaining columns that the generated view underneath it picks up. Clear GeneratedBaseViewName and use a fully custom base view (BaseViewGenerated = 0) instead...`

That throw is deliberate and well-reasoned — the problem is not the guard, it is that nothing prevents the enabling metadata from reaching PostgreSQL in the first place.

## Impact

A PostgreSQL installation that applies this metadata and then runs `mj codegen` fails with the error above. The failure is loud rather than silent, which is good, but it is a hard stop on a supported platform triggered by shipped metadata.

## How it surfaced

Found while producing the missing PostgreSQL counterparts for the 6.x migrations (#3471). `V202608050105__v6.1.x__Layered_Base_Views_Pilot` has no viable PG counterpart for exactly this reason: its inner "...Generated" views can never be produced on PostgreSQL. `V202608050100`'s bake correctly omits them, which is why the pilot's wrapper views have nothing to wrap.

[...]

PR #3866, opened 2026-08-16
fix(pg): expose Entity.GeneratedBaseViewName through vwEntities

### LINKED ISSUE #3837: PostgreSQL: CodeGen silently skips layered base views instead of throwing, so #3477's loud stop never fires
## One sentence

On PostgreSQL, CodeGen **silently skips** the base view of an entity with `GeneratedBaseViewName` set — it emits that entity's CRUD functions and its view *permissions* file but no view at all, reports success, and never reaches `assertLayeredBaseViewSupported()` — so the deliberate hard stop described in #3477 does not fire on the ordinary `mj codegen` path.

## Why this matters

#3477 establishes that layered base views are SQL-Server-only and that the PG provider throws by design, and explicitly calls that *good*: "The failure is loud rather than silent." This issue reports that on the incremental codegen path it is **not** loud. The install is left in the exact state layering was meant to prevent:

- the inner `…Generated` view is never created,
- the outer application-owned `BaseView` (which selects from it) has nothing to wrap,
- `mj codegen` exits reporting success,
- and nothing anywhere says PostgreSQL declined to do the work.

A guard that cannot be reached provides no protection, and the "loud failure" property #3477 relies on is currently not there to rely on.

## Environment

- `@memberjunction/cli@6.1.0-edge.2`, `@memberjunction/codegen-lib@6.1.0-edge.2` (published npm build, not a local tree)
- PostgreSQL 17.10, fresh DB migrated with `mj migrate --tag v6.1.0-edge.2`
- App under test: `MemberJunction/bizapps-common` @ `next`, schema `__mj_bizappscommon`

Both halves of the intended mechanism are present in the published build:

```
dist/Database/sql_codegen.js
  generatesView = (entity.BaseViewGenerated || entity.HasLayeredBaseView) && !entity.VirtualEntity
dist/Database/providers/postgresql/PostgreSQLCodeGenProvider.js
  assertLayeredBaseViewSupported   (2 occurrences)
```

## Repro

1. Fresh PG database; `mj migrate --tag v6.1.0-edge.2`; install `bizapps-common`'s `migrations-pg/`.
2. Apply a migration that flips two entities to layered, i.e. sets `BaseViewGenerated = false` and `GeneratedBaseViewName`:

```sql
SELECT "Name","BaseViewGenerated","GeneratedBaseViewName"
FROM __mj."Entity" WHERE "GeneratedBaseViewName" <> '';

 MJ_BizApps_Common: People        | f | vwPeopleGenerated
 MJ_BizApps_Common: Organizations | f | vwOrganizationsGenerated
```

3. Run `mj codegen` with `DB_PLATFORM=postgresql`.

**Observed:** completes — `MJ CodeGen complete — 385 entities in 60.93s`. No exception, and no occurrence of `layered base views are not supported` anywhere in the output.

The generated-SQL directory shows the skip precisely. Same schema, same run, layered vs non-layered:

| Entity | layered | `vw*.view.generated.sql` | `vw*.view.permissions.generated.sql` |
|---|---|---|---|
| `People` | yes | **missing** | present |
| `Organizations` | yes | **missing** | present |
| `OrganizationTypes` | no | present | present |
| `ContactMethods` | no | present | present |
| `Addresses` | no | present | present |

No emitted file anywhere references `vwPeopleGenerated` or `vwOrganizationsGenerated`, and neither view exists in the database afterwards. CRUD functions (`spCreatePerson`, `spUpdateOrganization`, …) and the trigger functions *were* generated normally, so the entities were unquestionably in scope and processed — only the view was dropped.

**Expected:** either the documented throw from `assertLayeredBaseViewSupported()`, or — if skipping is the intended behaviour on this path — a loud, non-zero-exit diagnostic naming the entities.

## Note on the permissions file

The permissions artifact being emitted for a view that was never created is worth a look on its own: it means the run produced `GRANT` statements against a relation it knew it had not generated.

## Downstream impact

Found while producing PostgreSQL counterparts for `bizapps-common`, whose two newest migrations (2026-08-13/14) adopt layering. On PostgreSQL those migrations produce metadata that says "layered" and a database with no inner view, and the follow-up migration that creates the outer wrapper has nothing to select from. Because codegen reports success, the resulting install looks healthy.

The same `GeneratedBaseViewName` metadata ships to every install on two core entities per #3477, so any install where those are in codegen scope is exposed to the same silent path.

## Related

- #3477 — layered base views are SQL-Server-only and PG CodeGen throws on them (this issue: the throw is not reached on the incremental path)
- #3514 — SS artifacts and the PG bake regenerating from different metadata states

MS-BC:
Closes #3837 · unblocks the decision in #3477

## The bug is one missing column in a view

`V202608050100` added `GeneratedBaseViewName` to `__mj.Entity`. On SQL Server that migration **also regenerated `vwEntities`** — its own header says so: *"the regenerated vwEntities … the new columns visible through vwEntities."* The PostgreSQL counterpart added the column to the table and re-granted on the view, but never recreated it.

```
PostgreSQL, fresh DB at v6.1.0-edge.2
  columns in __mj."Entity"     named GeneratedBaseViewName : 1
  columns in __mj."vwEntities" named GeneratedBaseViewName : 0
```

The metadata provider reads entities through `vwEntities`, so `EntityInfo.GeneratedBaseViewName` was always `undefined` and

```ts
get HasLayeredBaseView(): boolean {
    const layered = this.GeneratedBaseViewName?.trim();
    return !!layered && layered.toLowerCase() !== (this.BaseView ?? '').toLowerCase();
}
```

was **always false on PostgreSQL**. Everything else followed mechanically:

- both gates — `(BaseViewGenerated || HasLayeredBaseView)` in `executeEntityInPhases` and `generateSingleEntitySQLToSeparateFiles` — evaluated false, so `generateBaseView()` was never called
- `assertLayeredBaseViewSupported()` therefore **could not fire**
- with `BaseViewGenerated = false`, the entity fell to the permissions-only branch

Observed on `bizapps-common`: `People` and `Organizations` emitted `vw*.view.permissions.generated.sql` and **no view**, while non-layered entities in the same schema emitted both — and `mj codegen` reported success.

## Why this matters for #3477

#3477 states the failure is *"loud rather than silent, which is good."* On PostgreSQL it was **neither** — it was absent. The guard was unreachable, so shipping `.layered-base-views.json` to a PostgreSQL install did not fail codegen as that issue expects; it quietly produced no inner view. That made option 3 there ("soften the guard to a warning that skips layering on PG") the de-facto behaviour, arrived at by accident and without the warning.

This PR restores the intended behaviour so the remaining question in #3477 is a real decision rather than a description of something that was not happening.

## Why `CREATE OR REPLACE`

PostgreSQL tracks a hard dependency from `spCreateEntity` / `spUpdateEntity` to this view's row type, so `DROP` needs `CASCADE` and would take the CRUD functions with it — the same hazard #3477 records for `vwVersionInstallations`. `CREATE OR REPLACE` can only **append**, so the column lands at the end of the select list rather than SQL Server's position. Nothing reads this view positionally (the provider maps by name) and a later `mj codegen` restores canonical ordering.

`.pg-only` because the SQL Server side already ships the column in its own `vwEntities`.

## Verification — PostgreSQL 17.10, DB migrated to v6.1.0-edge.2

| | before | after |
|---|---|---|
| `vwEntities` exposes the column | 0 | **1** |
| layered entities visible to metadata | 0 | **4** |
| `mj codegen` | exit 0, refusal never printed | **exit 1, refusal printed 20×** |
| `spCreate/spUpdate/spDeleteEntity` | 3 | **3** (intact — no cascade) |

The refusal now names the real entities, including one of the two core entities from #3477:

```
Failed to regenerate base view for MJ: Version Installations: Error: Entity
"MJ: Version Installations" sets GeneratedBaseViewName = 'vwVersionInstallationsGenerated',
but layered base views are not supported on PostgreSQL. …
```

## Worth a follow-up

This is the #3514 divergence class appearing in core's own migrations: an SS migration regenerated a view, its PG counterpart did not. Any other column added to `Entity` in the 6.x line is worth the same check.

### DIFF
--- a/migrations-pg/v6/V202608160100__v6.1.x__vwEntities_Expose_GeneratedBaseViewName.pg-only.sql
+++ b/migrations-pg/v6/V202608160100__v6.1.x__vwEntities_Expose_GeneratedBaseViewName.pg-only.sql
@@ -1,0 +1,126 @@
+-- ============================================================================
+-- Expose Entity.GeneratedBaseViewName through __mj.vwEntities on PostgreSQL
+-- ============================================================================
+--
+-- WHY THIS EXISTS (#3837)
+--
+-- V202608050100 added `GeneratedBaseViewName` to `__mj.Entity`. On SQL Server that
+-- migration ALSO regenerated `vwEntities`, so the column became visible through the
+-- view — its own header says so ("the regenerated vwEntities … the new columns
+-- visible through vwEntities"). Its PostgreSQL counterpart added the column to the
+-- table and re-granted on the view, but never recreated it.
+--
+-- The metadata provider reads entities through `vwEntities`, so on PostgreSQL
+-- `EntityInfo.GeneratedBaseViewName` was always undefined and `HasLayeredBaseView`
+-- was therefore always FALSE. Three things followed mechanically:
+--
+--   * both CodeGen gates — `(BaseViewGenerated || HasLayeredBaseView)` in
+--     executeEntityInPhases and generateSingleEntitySQLToSeparateFiles — evaluated
+--     false for a layered entity, so `generateBaseView()` was never called;
+--   * `assertLayeredBaseViewSupported()` therefore could not fire, and the
+--     deliberate PostgreSQL refusal documented in #3477 was unreachable;
+--   * with `BaseViewGenerated = false` the entity fell to the permissions-only
+--     branch, emitting `vw*.view.permissions.generated.sql` and no view.
+--
+-- Observed on bizapps-common: `People` and `Organizations` produced a permissions
+-- file and no base view, while non-layered entities in the same schema produced
+-- both, and `mj codegen` reported success either way.
+--
+-- WHY `CREATE OR REPLACE` AND NOT DROP/CREATE
+--
+-- PostgreSQL tracks a hard dependency from `spCreateEntity` / `spUpdateEntity` to
+-- this view's row type, so a DROP would need CASCADE and would take the CRUD
+-- functions with it (the same hazard #3477 records for vwVersionInstallations).
+-- `CREATE OR REPLACE` can only APPEND columns, which is why the column lands at
+-- the end of the select list rather than in the position SQL Server uses. Nothing
+-- reads this view positionally — the provider maps by name — and a later
+-- `mj codegen` run restores the canonical ordering.
+--
+-- pg-only: the SQL Server side already ships this column in its own vwEntities.
+-- ============================================================================
+
+CREATE OR REPLACE VIEW __mj."vwEntities" AS
+SELECT e."ID",
+    e."ParentID",
+    e."Name",
+    e."NameSuffix",
+    e."Description",
+    e."AutoUpdateDescription",
+    e."BaseTable",
+    e."BaseView",
+    e."BaseViewGenerated",
+    e."SchemaName",
+    e."VirtualEntity",
+    e."TrackRecordChanges",
+    e."AuditRecordAccess",
+    e."AuditViewRuns",
+    e."IncludeInAPI",
+    e."AllowAllRowsAPI",
+    e."AllowUpdateAPI",
+    e."AllowCreateAPI",
+    e."AllowDeleteAPI",
+    e."CustomResolverAPI",
+    e."AllowUserSearchAPI",
+    e."FullTextSearchEnabled",
+    e."FullTextCatalog",
+    e."FullTextCatalogGenerated",
+    e."FullTextIndex",
+    e."FullTextIndexGenerated",
+    e."FullTextSearchFunction",
+    e."FullTextSearchFunctionGenerated",
+    e."UserViewMaxRows",
+    e."spCreate",
+    e."spUpdate",
+    e."spDelete",
+    e."spCreateGenerated",
+    e."spUpdateGenerated",
+    e."spDeleteGenerated",
+    e."CascadeDeletes",
+    e."DeleteType",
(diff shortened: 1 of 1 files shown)


MS-BC:
Closing as superseded.

`migrations-pg/v6/V202608202230__v6.1.x__PG_CodeGen_Regen.pg-only.sql` is already on `next` (2026-08-20) and rebuilds **all four** affected core views — `vwEntities`, `vwEntityFields`, `vwEntityRelationships`, `vwCompanyIntegrationRuns` — appending the seven columns the frozen views were missing:

```
ExternalDataSourceID, ExternalObjectName, GeneratedBaseViewName,
AllowDirectSQLInsert, AllowDirectSQLUpdate, AllowDirectSQLDelete, Configuration
```

Everything this PR adds is a strict subset of that, so merging it would be redundant.

It would also be actively harmful: this migration is versioned **before** the one already on `next`, so merging it inserts a lower-versioned migration into a chain that installs on v6.1.0-edge.3 have already passed. Flyway rejects out-of-order migrations by default, which would break upgrades on exactly the installs that are current.

Two related PRs (#3870/#3866 and #3933) had the same overlap. #3933 keeps its other half — a CodeGen validator that *detects* this class of drift — since the repair existed but nothing ever noticed it was needed.

Separately worth flagging: the release-time PG migration gate passed all three of these while they carried lower-versioned migrations, because it validates against a **fresh** database, where ordering is never a problem. A gate that only builds from empty cannot see an out-of-order insertion.

PR #3867, opened 2026-08-16
Google Drive storage: service accounts can actually upload (#3847)

### LINKED ISSUE #3847: GoogleDriveFileStorage: initialize() discards a working service-account driver, and PutObject sends a Buffer where googleapis needs a stream
Two independent defects in `@memberjunction/storage`'s Google Drive driver (seen on `6.1.0-edge.1`, googleapis `171.4.0`). Together they make a **service-account-backed Drive account fail with no usable diagnosis**, and the second one fails *after* auth and folder creation succeed — so every symptom reads as a storage-configuration problem.

Found while tracing why consented realtime recordings never reached storage: `AIAgentSession` rows carried `RecordingMedia='Audio'` and a stamped `RecordingStartedAt`, but `RecordingFileID` was null and `__mj.File` was empty.

---

## 1. `initialize(config)` throws away the client the constructor already built

`GoogleDriveFileStorage`'s **constructor** supports three auth methods — key file (service account), credentials JSON (service account), and OAuth2 refresh token.

`initialize(config)` — the path taken when credentials come from a **database** `MJ: Credentials` record via `initializeDriverWithAccountCredentials` — reads only `clientID` / `clientSecret` / `refreshToken`:

```js
if (this._clientID && this._clientSecret && this._refreshToken) { /* OAuth2 */ }
else {
    throw new Error('Google Drive storage requires clientID, clientSecret, and refreshToken to be set');
}
```

So:

* a **service account** supplied through the DB credential is ignored and the driver throws;
* worse, it throws **even when `this._drive` was already successfully built by the constructor** from `STORAGE_GDRIVE_KEY_FILE` / `STORAGE_GDRIVE_CREDENTIALS_JSON`. Attaching a credential record *breaks a driver that was working from env*.

**Suggested fix:** mirror the constructor's methods in `initialize` (accept `keyFile` / `credentialsJSON`), and do not throw when a client is already initialized.

## 2. `PutObject` passes a Buffer where googleapis requires a stream

Past #1, the driver authenticates, navigates and creates the folder tree correctly, then dies inside `files.create`:

```
TypeError: part.body.pipe is not a function
    at Resource$Files.create (googleapis/build/src/apis/drive/v3.js:943:65)
    at GoogleDriveFileStorage.PutObject (GoogleDriveFileStorage.js:1067:41)
```

`media.body` is passed the raw `data` Buffer; googleapis 171.x assembles the multipart request by calling `.pipe()` on it. Both the `files.create` and `files.update` branches are affected.

**Suggested fix:** `body: Buffer.isBuffer(data) ? Readable.from(data) : data`.

This one is the expensive defect, because it surfaces **after** auth and folder creation have succeeded — the failure looks like bad storage config rather than a client-library incompatibility.

## 3. Related gap: Shared Drives are unreachable

Past both fixes, Drive answers honestly:

```
403 — Service Accounts do not have storage quota.
      Leverage shared drives, or use OAuth delegation instead.
```

That is a Google constraint, not an MJ bug — but the driver never passes `supportsAllDrives` / `includeItemsFromAllDrives` / `driveId`, so a Shared Drive **cannot be targeted even when one is configured**. As it stands, service-account auth cannot work with this driver at all. Worth treating as part of the same fix.

---

## Reproduction

Any `MJ: File Storage Accounts` row on the Google Drive provider whose credential carries service-account material rather than OAuth2 fields, then any `UploadFile` against it. Nothing needs to be realtime-specific.

## Note on diagnosability

Every layer above this swallows: `storeRealtimeRecording` logs and returns null, the resolver returns `{Success:false}` without throwing, and `__mj.ErrorLog` stays empty. That posture is right — a storage failure must not fail a session — but with the driver error discarded too, there is no durable trace of *why* an upload failed. Surfacing the driver's error on the failure result would have turned this investigation into one log line.

MS-BC:
Closes #3847. Filed from Caliber (`MemberJunction/bizapps-caliber`), found tracing consented realtime recordings that were captured and then vanished (bizapps-caliber#165's media half).

## Three stacked defects, one diagnosis-free failure

The third fails **after** auth and folder creation succeed, so every symptom read as bad storage configuration.

### 1. `initialize()` threw away the client the constructor already built

The constructor supports three auth methods; `initialize()` — the path a database `MJ: Credentials` record takes — read only the OAuth2 triple and threw otherwise. Two consequences: a service account supplied through the DB was ignored, and **attaching a credential record broke a driver that was working from env**.

Now: `initialize()` mirrors all three methods (`keyFile` / `credentialsJSON` / OAuth2), **keeps an existing client** when the record carries no auth material (a record that only names the account must not un-configure a working driver), and throws only when there is no client *and* no credential.

### 2. `PutObject` passed a Buffer where googleapis requires a stream

`media.body` now gets `Readable.from(data)` on both the create and update branches. Handed the raw Buffer, googleapis dies inside `files.create` with `part.body.pipe is not a function`.

### 3. Shared Drives unreachable — and they are the only place a service account can write

Google's own 403 says it: *"Service Accounts do not have storage quota. Leverage shared drives."* Without Shared Drive support the first two fixes only move the failure.

Delivered **by construction rather than by flag**: the client factory sets `supportsAllDrives` / `includeItemsFromAllDrives` as googleapis *client-level default params*, covering all 25 call sites in the driver instead of trusting 25 per-call flags plus every future one. A configured `driveId` (config key or `STORAGE_GDRIVE_DRIVE_ID`) becomes the default navigation root — a Shared Drive's id doubles as its root folder id.

## Tests

**82 MJStorage tests pass, 7 new**: the no-auth-material credential record keeps a working client · the no-client-no-credential case still throws (with all three options named) · `credentialsJSON` builds a service-account client · the upload hands googleapis a `Readable` · the stream carries the same bytes the Buffer held · `driveId` becomes the root · the factory's default params asserted once.

## Explicitly out of scope

The issue's diagnosability note — callers above this swallow the driver error — is caller-side (`storeRealtimeRecording`, the resolver) and not changed here.


rkihm-BC (commented):
# Review

Solid, well-diagnosed fix. The three-defect decomposition is correct and the ordering insight — that defect 2 fires *after* auth and folder creation, which is why every symptom read as bad storage config — is the reason this needed tracing rather than a stack trace. That belongs in the commit message and it's there.

Rather than take the PR body's word for the mechanics, I verified each claim:

| Claim | How checked | Result |
|---|---|---|
| `Readable.from(Buffer)` preserves bytes | ran on Node 24 | ✅ one Buffer chunk, `objectMode: true` — the special case in `internal/streams/from.js`, not byte-wise iteration |
| Client-level `params` reach every request | read `googleapis-common@8.0.1/build/src/apirequest.js:56-66` | ✅ `extend(true, {}, options.params, parameters.params)` — merged verbatim, no schema filtering |
| The test's `drive.context._options.params` assertion | ran `google.drive({params})` against the installed 171.4.0 | ✅ lands exactly there |
| "82 tests pass, 7 new" | worktree at `3f227d41`, symlinked node_modules, ran vitest | ✅ **98 pass across 7 files**, 7 new (82 is just stale — `next` grew tests since). `tsc --noEmit` clean. |
| The new fields actually reach `initialize()` from a credential record | read `util.ts:245` | ✅ `await driver.initialize({...baseConfig, ...resolved.values})` — a raw spread, so `keyFile` / `credentialsJSON` / `driveId` on an `MJ: Credentials` record flow straight through. The defect-1 fix is live, not theoretical. |

## What's good

- **Client-level default params over 25 per-call flags is the right architectural call** — and it's better justified than the PR body claims. I checked the drive_v3 param declarations: `supportsAllDrives` is declared on `files.`{get, list, create, update, delete, copy} *and* `permissions.create` — every endpoint this driver calls. The "by construction" argument holds completely for the param that actually does the work.
- **"Keep an existing client when the record carries no auth material"** is the subtle half of defect 1 and the easy half to miss. Attaching a credential record breaking a working env-configured driver is a nasty, non-obvious failure mode.
- **Tests assert at the seams that broke** — `media.body instanceof Readable` plus the byte round-trip, not a mock's return value. `fileCreate()` correctly distinguishes the upload's `files.create` from the folder-creation ones.
- **Scope discipline is honest.** Diagnosability is declared out of scope and it genuinely landed separately in `6c5ed3cb5d`, so nothing is orphaned.

## Blockers

**1. No changeset.** changeset-bot flagged it. `@memberjunction/storage` needs a `patch` — as-is this merges without publishing.

**2. `IsConfigured` was not updated, and it is a hard gate — not just a log line.**

`packages/MJStorage/src/drivers/GoogleDriveFileStorage.ts:274` still reads:

```ts
const isConfigured = hasClientID && hasClientSecret && hasRefreshToken;
```

I initially assumed this only affected a diagnostic log in `util.ts:633`. It doesn't — two consumers outside MJStorage branch on it:

- `packages/SearchEngine/src/generic/StorageSearchProvider.ts:264` — `return []`, silently skipping the account in federated search
- `packages/ContentAutotagging/src/CloudStorage/providers/AutotagCloudStorage.ts:163` — `return null`, aborting

So a service-account Drive account that this PR just made functional for upload remains invisible to search indexing and content autotagging, and emits the 15-line "❌ not configured, set your OAuth env vars" block on every `listObjects()` call. One line in a file already open here: `hasOAuth || !!this._drive`.

## Should fix

**3. `initialize()` can't move the root once `_rootFolderId` is already set** — lines 347-354.

The constructor sets `_rootFolderId = config?.rootFolderID || env(STORAGE_GDRIVE_ROOT_FOLDER_ID) || this._driveId`. So with `STORAGE_GDRIVE_DRIVE_ID=A` on the server, `_rootFolderId` is `A`. A per-account credential carrying `driveId: B` then sets `_driveId = B`, but the guard `else if (config.driveId && !this._rootFolderId)` fails because `_rootFolderId` is truthy — the root stays `A`. The driver navigates drive A while believing it's on B.

Narrow precondition (server env var set *and* a per-account credential naming a different drive), but it's new code with a wrong guard, the failure is silent, and it's cross-tenant. Fixing it also gives line 348's `this._driveId = config.driveId` a purpose — today that assignment is dead, since `_driveId` is never read after construction.

## Optional — your call, one line each

**4.** `STORAGE_GDRIVE_DRIVE_ID` is missing from the README env table (`readme.md:596`) and from the driver's class JSDoc env list (line ~148), which still says "requires one of the following" and names only the two service-account vars.

**5.** Log a warning when `initialize()` falls through to the constructor-built client. The behavior is correct and is what the issue asked for — but it's the thing that would make a mistyped credential key diagnosable instead of silently using the server's global Drive account.

**6.** `driveId` vs. its neighbours `rootFolderID` / `clientID`, and vs. SharePoint's `driveID`. A convention call rather than a defect, but it's a public surface — free to change now, permanent after release. Note that credential-record keys never pass through the zod schema, so a mis-cased key is silently ignored by the driver rather than rejected.

## Things I looked at and decided need no change

Recording these so they don't get re-litigated:

- **`includeItemsFromAllDrives` on non-list endpoints.** It's declared only on `files.list`, so it rides along as an undeclared query param on get/create/update/delete/copy/permissions.create. Drive ignores unrecognized query params, and the param that actually matters (`supportsAllDrives`) is valid everywhere. Not worth gating a fix on.
- **`corpora` / `driveId` on `files.list`.** `includeItemsFromAllDrives` with a parent-scoped query is the documented mechanism; `corpora: 'drive'` is an efficiency hint, not a requirement. No change needed.
- **`SearchFiles` parent scoping.** `parentFolderId` falls back to `_rootFolderId`, which is always populated when a `driveId` is configured. It's only undefined when nothing at all is configured — pre-existing, and not a Shared Drive scenario.
- **Reaching into privates via `as unknown as {...}` in the new tests.** 24 existing occurrences across this suite; house precedent, fine.

Net: two blockers, one real bug worth fixing, three optional one-liners. The core fix is correct and the credential plumbing works end to end.

Issue #3773, opened 2026-08-13
CI: nightly PostgreSQL lane for the deterministic integration tier


MS-BC:
Filed from @rkihm-BC's review of #3703:

> **No CI lane**, as the PR states. Everything here decays the day after merge without one, and
> since most of this PR *is* the harness, a nightly PG lane is what converts it from a measurement
> into a guarantee. Worth filing before merge rather than after.

## Why

#3703 makes the deterministic integration tier runnable against PostgreSQL and fixes what that
exposed. Nothing currently re-runs it. Without a lane, the PG result is a single point-in-time
measurement taken by hand, and every subsequent merge can silently regress it — the failure mode
being that the next person to run it inherits a large, undifferentiated pile of breakage with no
way to attribute any of it.

## What it should do

- Run `MJ_INTEGRATION_TEST=1 RUN_MUTATION_TESTS=1 mj test suite "Integration Tests — Deterministic"`
  against a PostgreSQL database built from `migrations` + `migrations-pg` + `mj sync push`.
- **`RUN_MUTATION_TESTS=1` is not optional.** Mutation-class bundles skip silently without it, and a
  green run then means materially less than it appears — that flag alone is the difference between
  the 49/61 originally reported on #3703 and the honest 58/61.
- Report the bundle-level pass count, and fail the lane on a *decrease* rather than on any failure,
  since the tier legitimately ships with known-failing bundles while PG parity is in progress.

## Known-failing bundles at the time of filing (head `302cd42714`)

| Bundle | Cause |
|---|---|
| IT29 — Cache Gauntlet | mutation-class; newly visible |
| IT67 — Content Vectorization | PG `spCreateContentSource` names `ScheduledActionID`, dropped by `V202608061704` — stale regenerated proc |
| IT74 — Task Graph Execution | TX8/TX9; deterministic across three isolated runs, not root-caused |

Whatever baseline the lane starts from should be these three, so a fourth is a real signal.

## Note on scope

The stored-query dialect gap and the PG migration counterparts are release-time conversion work, not
per-PR work — the lane should measure them, not gate a feature PR on them.

PR #3870, opened 2026-08-16
fix(pg): expose five declared EntityFields through their base views

### LINKED ISSUE #3869: PostgreSQL: five core EntityFields are declared in metadata but missing from their base views
## One sentence

On PostgreSQL, five core `EntityField`s are declared in metadata and exist as real columns on their base tables, but are **not exposed by their base views** — so metadata promises a field the view cannot serve.

## The audit

Fresh PostgreSQL 17.10 database, `mj migrate --tag v6.1.0-edge.2`, nothing else applied. Cross-referencing MJ's own metadata against the catalog:

```sql
SELECT e."BaseTable" || '.' || f."Name",
       CASE WHEN EXISTS (SELECT 1 FROM information_schema.columns vc
              WHERE vc.table_schema='__mj' AND vc.table_name=e."BaseView" AND vc.column_name=f."Name")
            THEN 'in view' ELSE 'MISSING from view' END
FROM __mj."EntityField" f
JOIN __mj."Entity" e ON e."ID" = f."EntityID"
WHERE e."SchemaName" = '__mj';
```

| Field | IsVirtual | AllowUpdateAPI | on base table | in base view |
|---|---|---|---|---|
| `Entity.AllowDirectSQLDelete` | false | true | yes | **no** |
| `Entity.AllowDirectSQLInsert` | false | true | yes | **no** |
| `Entity.AllowDirectSQLUpdate` | false | true | yes | **no** |
| `Entity.GeneratedBaseViewName` | false | true | yes | **no** |
| `EntityRelationship.RelatedRecordCollection` | false | true | yes | **no** |

None are virtual, all are updatable through the API, and all exist on the base table. The view is simply missing them.

This is an internal inconsistency rather than a judgement call: the metadata layer says the field exists, so anything reading the entity expects the view to return it.

## Not everything absent from a view is a bug — this is the filtered list

`Entity.ExternalDataSourceID` and `Entity.ExternalObjectName` are absent from `vwEntities` too, and are **correctly** absent: they carry no `EntityField` row on `Entity`, consistent with #3514's note that `vwEntities.ExternalDataSourceID` was deliberately dropped. Filtering on "declared as an EntityField" is what separates the five above from that pair, and is why this list is five rather than seven.

## Cause, and why it recurs

Same divergence class as #3837, which is one of the five. There, the SQL Server migration that added `GeneratedBaseViewName` **also regenerated `vwEntities`** — its header says so explicitly — while its PostgreSQL counterpart added the column and never recreated the view. The column landed in the table and nothing surfaced it.

Nothing checks for this, so each occurrence is found only when something downstream misbehaves. For `GeneratedBaseViewName` the downstream effect was severe: `EntityInfo.GeneratedBaseViewName` was always `undefined`, `HasLayeredBaseView` was therefore always false, and `assertLayeredBaseViewSupported()` — the deliberate PostgreSQL refusal in #3477 — became unreachable (see #3837 for the full trace).

The other four have not been traced to a downstream symptom yet, which is precisely the problem: they are invisible until they are not.

## Asks

1. Expose the four remaining columns on PostgreSQL. `CREATE OR REPLACE VIEW` (append-only) rather than DROP/CREATE — PostgreSQL tracks a hard dependency from the CRUD functions to the view's row type, so a DROP needs CASCADE and takes them with it (the hazard #3477 records for `vwVersionInstallations`).
2. **Add a standing check** so this class stops being discovered by accident: every non-virtual `EntityField` whose column exists on the base table should be exposed by that entity's base view. The query above is the whole check and needs only a migrated database — the nightly PostgreSQL lane added for #3773 already has one.

## Related

- #3837 — the layered-base-view instance of this, with the full mechanism
- #3477 — the guard rendered unreachable by that instance
- #3514 — the general SS-artifact / PG-bake divergence class

Verified on PostgreSQL 17.10 with `@memberjunction/cli@6.1.0-edge.2`.

MS-BC:
Closes #3869 · same divergence class as #3837

## What the audit found

A clean PostgreSQL install at `v6.1.0-edge.2`, cross-referencing MJ's **own `EntityField` metadata** against the catalog, turns up five fields that are declared in metadata and exist on their base **table**, but are not selected by their base **view**:

| Field | IsVirtual | AllowUpdateAPI | on table | in view |
|---|---|---|---|---|
| `Entity.AllowDirectSQLInsert` | false | true | yes | **no** |
| `Entity.AllowDirectSQLUpdate` | false | true | yes | **no** |
| `Entity.AllowDirectSQLDelete` | false | true | yes | **no** |
| `Entity.GeneratedBaseViewName` | false | true | yes | **no** |
| `EntityRelationship.RelatedRecordCollection` | false | true | yes | **no** |

Metadata promises a field the view cannot serve. None is virtual, all are updatable through the API.

## Why five and not seven

`Entity.ExternalDataSourceID` and `Entity.ExternalObjectName` are **also** absent from `vwEntities` — and correctly so. They carry no `EntityField` row on `Entity`, consistent with #3514's note that `vwEntities.ExternalDataSourceID` was deliberately dropped.

"Declared as an EntityField" is the filter that separates a real defect from an intentional omission, and it is what makes this list five rather than seven. Regenerating the view wholesale would have quietly re-added those two.

## Cause

The divergence traced in #3837: the SQL Server migration that adds such a column **also regenerates the view**; the PostgreSQL counterpart adds the column and does not.

There the consequence was severe — `GeneratedBaseViewName` never reached `vwEntities`, so `EntityInfo.GeneratedBaseViewName` was always `undefined`, `HasLayeredBaseView` was permanently false, and `assertLayeredBaseViewSupported()` (the deliberate PostgreSQL refusal in #3477) became unreachable. The other four have **no traced symptom yet**, which is the concern rather than the reassurance: this class is invisible until something downstream misbehaves.

## Why `CREATE OR REPLACE`

PostgreSQL tracks a hard dependency from the generated CRUD functions to each view's row type, so `DROP` needs `CASCADE` and takes them with it — the hazard #3477 records for `vwVersionInstallations`. That restricts this to **appending**, so the columns land at the end of the select list rather than SQL Server's position. Nothing reads these views positionally (the provider maps by name) and a later `mj codegen` restores canonical ordering.

`vwEntities` is replaced first, because `vwEntityRelationships` selects from it.

## Verification — PostgreSQL 17.10, clean v6.1.0-edge.2 database

| | before | after |
|---|---|---|
| audit: declared-but-missing fields | **5** | **0** |
| `SELECT` the four new `vwEntities` columns | fails | **5 rows** |
| `SELECT RelatedRecordCollection` | fails | **5 rows** |
| Entity + EntityRelationship CRUD functions | 6 | **6** (intact — no cascade) |

## The standing check this should become

The audit query is small and needs only a migrated database — which the nightly PostgreSQL lane from #3773 already has. Adding it there would stop this class being discovered by accident. Raised as ask 2 in #3869 rather than bundled here, since it touches CI rather than migrations.

### DIFF
--- a/migrations-pg/v6/V202608160200__v6.1.x__Expose_Declared_EntityFields_In_Base_Views.pg-only.sql
+++ b/migrations-pg/v6/V202608160200__v6.1.x__Expose_Declared_EntityFields_In_Base_Views.pg-only.sql
@@ -1,0 +1,183 @@
+-- ============================================================================
+-- Expose the remaining declared EntityFields through their base views (PostgreSQL)
+-- ============================================================================
+--
+-- WHY THIS EXISTS (#3869)
+--
+-- An audit of a clean PostgreSQL install — MJ's own EntityField metadata cross-
+-- referenced against the catalog — found five fields declared in metadata and present
+-- on their base TABLE, but not selected by their base VIEW. V202608160100 exposed the
+-- first (Entity.GeneratedBaseViewName, #3837, because it made #3477's guard
+-- unreachable); this migration exposes the other four:
+--
+--   Entity.AllowDirectSQLInsert
+--   Entity.AllowDirectSQLUpdate
+--   Entity.AllowDirectSQLDelete
+--   EntityRelationship.RelatedRecordCollection
+--
+-- All are IsVirtual = false, AllowUpdateAPI = true and present on the base table, so
+-- none is a deliberate omission. Entity.ExternalDataSourceID and
+-- Entity.ExternalObjectName are ALSO absent from vwEntities and are correctly absent —
+-- they carry no EntityField row on Entity, matching #3514's note that
+-- vwEntities.ExternalDataSourceID was deliberately dropped. "Declared as an
+-- EntityField" is the filter that separates a defect from an intentional omission, and
+-- is why this set is five and not seven: regenerating the view wholesale would have
+-- quietly reinstated that pair.
+--
+-- CAUSE. The SQL Server migration that adds such a column also regenerates the view;
+-- the PostgreSQL counterpart adds the column and does not (#3837 has the traced
+-- mechanism). The remaining four have no traced symptom yet, which is the concern
+-- rather than the reassurance.
+--
+-- ORDERING — LOAD-BEARING. This runs AFTER V202608160100, and its view definition
+-- INCLUDES the GeneratedBaseViewName column that migration appended. `CREATE OR
+-- REPLACE VIEW` may only APPEND columns: it cannot rename or reorder existing ones.
+-- Rebuilding this from the pre-0100 definition would try to put AllowDirectSQLInsert
+-- where GeneratedBaseViewName now sits and fail with
+--   cannot change name of view column "GeneratedBaseViewName" to "AllowDirectSQLInsert"
+-- Timestamp order guarantees 0100 first; do not reorder these two.
+--
+-- DROP/CREATE is not an option: PostgreSQL tracks a hard dependency from the generated
+-- CRUD functions to each view's row type, so a DROP needs CASCADE and takes them with
+-- it (the hazard #3477 records for vwVersionInstallations). Nothing reads these views
+-- positionally — the provider maps by name — and a later `mj codegen` restores
+-- canonical ordering.
+--
+-- vwEntities is replaced first because vwEntityRelationships selects from it.
+--
+-- pg-only: the SQL Server side already ships these columns in its own views.
+-- ============================================================================
+
+CREATE OR REPLACE VIEW __mj."vwEntities" AS
+SELECT e."ID",
+    e."ParentID",
+    e."Name",
+    e."NameSuffix",
+    e."Description",
+    e."AutoUpdateDescription",
+    e."BaseTable",
+    e."BaseView",
+    e."BaseViewGenerated",
+    e."SchemaName",
+    e."VirtualEntity",
+    e."TrackRecordChanges",
+    e."AuditRecordAccess",
+    e."AuditViewRuns",
+    e."IncludeInAPI",
+    e."AllowAllRowsAPI",
+    e."AllowUpdateAPI",
+    e."AllowCreateAPI",
+    e."AllowDeleteAPI",
+    e."CustomResolverAPI",
+    e."AllowUserSearchAPI",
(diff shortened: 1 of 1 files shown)


MS-BC:
### Merge-order note for reviewers

This PR and #3866 both `CREATE OR REPLACE` `__mj.vwEntities`, and PostgreSQL only lets that **append** columns — it cannot rename or reorder existing ones. Merged in the wrong order they fail with:

```
ERROR: cannot change name of view column "GeneratedBaseViewName" to "AllowDirectSQLInsert"
```

I reproduced that, then rebuilt this migration's definition **on top of** #3866's, so the two are safe together. Timestamp order (`…0100` before `…0200`) enforces it.

Verified on clean PostgreSQL 17.10 databases:

| arrangement | result |
|---|---|
| #3866 then this | both apply; audit reports **0** remaining |
| this alone (if #3866 lands later or not at all) | applies; appends all four; audit reports **0** |

So either merge order is safe, but **do not reorder the two migration timestamps**.

MS-BC:
Closing as superseded.

`migrations-pg/v6/V202608202230__v6.1.x__PG_CodeGen_Regen.pg-only.sql` is already on `next` (2026-08-20) and rebuilds **all four** affected core views — `vwEntities`, `vwEntityFields`, `vwEntityRelationships`, `vwCompanyIntegrationRuns` — appending the seven columns the frozen views were missing:

```
ExternalDataSourceID, ExternalObjectName, GeneratedBaseViewName,
AllowDirectSQLInsert, AllowDirectSQLUpdate, AllowDirectSQLDelete, Configuration
```

Everything this PR adds is a strict subset of that, so merging it would be redundant.

It would also be actively harmful: this migration is versioned **before** the one already on `next`, so merging it inserts a lower-versioned migration into a chain that installs on v6.1.0-edge.3 have already passed. Flyway rejects out-of-order migrations by default, which would break upgrades on exactly the installs that are current.

Two related PRs (#3870/#3866 and #3933) had the same overlap. #3933 keeps its other half — a CodeGen validator that *detects* this class of drift — since the repair existed but nothing ever noticed it was needed.

Separately worth flagging: the release-time PG migration gate passed all three of these while they carried lower-versioned migrations, because it validates against a **fresh** database, where ordering is never a problem. A gate that only builds from empty cannot see an out-of-order insertion.

PR #3871, opened 2026-08-16
IS-A promotion: attach a NEW child to an EXISTING parent (#3825)

### LINKED ISSUE #3825: IS-A: no way to attach a new child record to an EXISTING parent (promotion)
## Summary

MJ's IS-A composition supports **discovery** (a loaded parent finds its existing child) but has no **promotion** — no way to say *"this existing `Person` is now also an `Applicant`"*. Any consumer that needs to add a subtype to a person/org that already exists is forced to create a duplicate parent instead.

## Repro

Against a real SQL Server provider, with `Caliber: Applicants` IS-A `MJ_BizApps_Common: People` on a shared PK:

```js
// A Person already exists (id = personId).
const applicant = await md.GetEntityObject('Caliber: Applicants', contextUser);
applicant.NewRecord();
applicant.Set('ID', personId);        // the shared PK — this Person IS the applicant
applicant.Set('FirstName', 'Isa');    // parent-held fields, set on the child to cascade
applicant.Set('LastName', 'Probe');
applicant.Set('Email', email);
applicant.Set('CompanyID', companyId);
await applicant.Save();
```

Result:

```
Save() -> false
message: Failed to save parent entity 'MJ_BizApps_Common: People': Error executing SQL
```

The child keeps the ID you set (`applicant.Get('ID') === personId`), and MJ then issues an **INSERT** for the parent — which collides with the existing row's primary key. Omitting the parent-held fields gives the earlier, more legible form of the same problem:

```
Failed to save parent entity 'MJ_BizApps_Common: People': First Name cannot be null; Last Name cannot be null
```

i.e. MJ is building a brand-new parent and validating it as new, rather than recognising that a parent with this PK already exists.

## Why (from `packages/MJCore/src/generic/baseEntity.ts`)

- `InitializeParentEntity()` unconditionally constructs a *fresh* parent via `GetEntityObject(...)`. There is no path that loads an existing parent by PK.
- `InitializeChildEntity()` / `discoverChildEntityName()` only go the other direction — a **loaded parent** discovering an already-existing child. If no child exists it simply returns; there is no "create the child for this loaded parent".

So the asymmetry is structural: parent→child discovery exists, parent→child creation does not.

## What we tried

Pre-loading the parent through the public `ISAParent` accessor, so its save would be an UPDATE rather than an INSERT. `applicant.ISAParent` correctly returns the `MJ_BizApps_Common: People` instance, but it comes back as a base `BaseEntity` without the generated `Load(id)` override, so this doesn't work as a user-land workaround. `InnerLoad(CompositeKey)` may be reachable, but engineering around internals didn't seem like the right answer to report without a maintainer's view.

## Why it matters (concrete)

In Caliber, MJ Forms fires `Forms: Upsert Respondent Person` on every completed submission, creating a `Person` and stamping `FormResponse.RespondentPersonID`. Caliber then binds its own subject and, unable to promote that existing Person to an `Applicant`, creates a **second** `Person` for the same human. Every candidate who applies ends up with two Person rows — one Forms points at, one Caliber points at, neither aware of the other. Nothing errors and nothing logs.

The schema is explicitly designed for the operation MJ can't perform. From the Caliber migration:

```sql
-- NO DEFAULT, deliberately: this value is Person's ID, never a fresh one. A DEFAULT here would
-- let an INSERT that forgot to supply it mint an orphan that the FK then rejects with a confusing
-- message about a Person nobody tried to reference.
ID UNIQUEIDENTIFIER NOT NULL,
```

Promotion is the normal case for IS-A in a multi-app install: bizapps-common owns `Person`, and each app adds its own subtype to people who already exist.

## Suggested shape

Some explicit, opt-in way to bind a new child to an existing parent — e.g. loading the parent by PK during `InitializeParentEntity()` when the caller has supplied the shared PK, or a documented `LoadParent(pk)` / `PromoteFrom(parent)` affordance. The important part is that the parent save becomes an UPDATE and the child save an INSERT, both inside the existing validation and event chain.

Happy to prototype if you can indicate which shape you'd accept — I stopped short of patching `baseEntity` since this changes IS-A save semantics for every consumer.

Environment: MJ `6.1.0-edge.2`, SQL Server provider.

MS-BC:
Closes #3825. Filed from Caliber (`MemberJunction/bizapps-caliber`), where `subject-upsert-driver`'s find-or-create creates a **second Person** instead of promoting the one that exists (bizapps-caliber#258).

## The gap

IS-A supported **discovery** (a loaded parent finds its existing child) but not **promotion** ("this existing `Person` is now also an `Applicant`"). `NewRecord()` unconditionally starts a fresh parent chain, so promotion INSERTed a duplicate parent and collided on the PK — or failed the parent's NOT NULL validation as if it were brand new. The issue's repro reproduces both shapes.

## The affordance — the issue's suggested shape, opt-in

```typescript
const applicant = await md.GetEntityObject<ApplicantEntity>('Applicants', contextUser);
applicant.NewRecord();
if (!await applicant.AttachToParent(CompositeKey.FromID(personId))) {
    // no such Person — caller decides: save as a fresh chain, or stop
}
applicant.Set('CompanyID', companyId);
await applicant.Save();   // Person UPDATEd, Applicant INSERTed, one transaction
```

It **loads** the parent chain by key (`InnerLoad`, which hydrates grandparents from the same row) — a loaded parent saves as an **UPDATE**, which is the entire trick: the existing parent-first chain save now updates the existing row and INSERTs only the child. The shared key is mirrored into the child exactly as `NewRecord()`'s adoption path does it (`SetLocal` + `ResetNeverSetFlag`).

**Nothing else changes.** Field routing, permissions, validation, `EnforceDisjointSubtype` (a non-overlapping parent still refuses a second subtype), and the transaction scope all run as they do today. If loading the parent discovers a *sibling* subtype's child, the save is unaffected — parent saves carry `IsParentEntitySave`, which bypasses leaf delegation.

## Two contracts the tests forced

- **A failed attach leaves the record exactly as it was.** `InnerLoad` wipes before it reads, so a missing parent row would otherwise gut the fresh chain (root PK nulled, record unsaveable). The method captures the minted key first and restores the chain on failure — both caller options stay open.
- **The parent's primary keys drive the operation, not the child's.** The shared-PK mirror is a same-name convention; a child schema that leaves the key entirely to routing has no mirror to maintain, and iterating the child's own (possibly empty) `PrimaryKeys` list is the silent no-op bug the first draft had. Both schema shapes are covered by tests.

## Guards

- Throws on a non-IS-A entity (there is no parent to attach to).
- Throws on an already-saved child — promotion is a decision about what a **new** record is, not an edit to an existing one.
- Returns `false` (record untouched) when no parent row exists under the key.

## Review follow-up (AN-BC)

- **Renamed `AttachToExistingParent` → `AttachToParent`** — "existing" is implied by the operation. Nothing shipped under the old name, so no alias.
- **`EntityInfo` gained the quick IS-A role lookups**: `IsRootType`, `IsLeafType`, `ParticipatesInIsA`, `IsARole` (`None`/`Root`/`Intermediate`/`Leaf`), `RootEntityInfo`, `DescendantEntities`. Root had no accessor at all, and two of these close real traps: `IsChildType` alone treats an *intermediate* type as a leaf (mishandling the middle of every chain deeper than two), and `ChildEntities` is direct-children-only, so "find every subtype" written against it misses everything past the first level.
- **Integration tests belong in bizapps-orders, not here** — verified, not assumed: MJ core models **zero** IS-A relationships (every core entity's `ParentID` is explicitly cleared in the metadata sync; no non-null `ParentID` is seeded anywhere in the shipped migrations), and the integration suite has no IS-A coverage. There is nothing in core to point a test at, and adding a hierarchy to core purely as a fixture would mean inventing production metadata to serve a test.

## Tests

**2,162 MJCore tests pass**, 5 new for promotion and 6 for the role lookups (mutation-verified — four mutants killed on the accessors),, using the same mock harness as `baseEntity.isa.child.test.ts`: successful promotion (parent loaded ⇒ UPDATE-on-save, child adopts PK and stays new) · routed reads on a mirror-less child schema · failed attach leaves the fresh chain intact · both guard throws.


AN-BC (commented):
This looks right to me. For brevity and still descriptive I'd suggest "AttachToParent" as the existing is implied and this seems cleaner. The approach looks right and retuning false if an acceptable no-op is good whereas throwing is not an IsA makes sense. Do we have helper methods on the EntityInfo class that make it easy to determine if a given entity is an ISA parent, child, root, or some combination? That would be good to have to allow quick lookups. We have that metadata in there somewhere just not sure how DX friendly it is.

Do we have any IsA relationships modeled in MJ core? Was thinking we can add integration test layer to hit this. If we don't then we can add this to the bizapps-orders layer where we heavily use IsA and do integration testing there to test this hard. 

@rkihm-BC I'd appreciate your eyes on this quickly too before @MS-BC finalizes and merges.

Madhav, good work/find and I don't need to look at this again on my end.

I'm good with this.

MS-BC:
Thanks @AN-BC — all three addressed, and pushed.

**1. Renamed to `AttachToParent`.** You're right that "existing" is implied by the operation, and it reads better at the call site. Nothing has shipped under the old name (the changeset is still pending), so there's no alias to carry.

**2. `EntityInfo` now has the quick IS-A lookups.** You were right that the metadata was already in there and the DX wasn't. `IsChildType` and `IsParentType` existed, **root did not**, and answering "what is this entity?" meant combining them at every call site. Added:

| accessor | answers |
|---|---|
| `IsRootType` | has subtypes, has no parent — the role the boolean pair could not express |
| `IsLeafType` | has a parent, has no subtypes |
| `ParticipatesInIsA` | is it in an IS-A hierarchy at all, in any role |
| `IsARole` | `'None' \| 'Root' \| 'Intermediate' \| 'Leaf'` — the single "what is this?" lookup |
| `RootEntityInfo` | the root of this entity's chain (itself when it is the root, `null` when there is no hierarchy) |
| `DescendantEntities` | every subtype at any depth — the downward twin of the existing `ParentChain` |

Two of those close real traps rather than just saving keystrokes:

- **`IsChildType` alone treats an intermediate type as a leaf.** On Webinars IS-A Meetings IS-A Products, `Meetings` is a child *and* a parent, so anything branching on `IsChildType` mishandles the middle of every chain deeper than two. That's the case `IsARole` exists for.
- **`ChildEntities` is direct children only.** `Products.ChildEntities` omits Webinars entirely, so a "find every subtype" written against it silently misses everything past the first level. `DescendantEntities` is cycle-guarded the same way `ParentChain` already is.

6 new specs over the shipped mock hierarchy (which conveniently carries all four roles at once), mutation-verified — four mutants, all killed: `IsRootType` ignoring the parent check, `DescendantEntities` not recursing, `IsARole` calling an intermediate a leaf, and `RootEntityInfo` returning null at the top.

**3. On integration tests — you're right, and it has to be bizapps-orders.** I checked rather than guessed:

- **MJ core models zero IS-A relationships.** Every core entity's `ParentID` is explicitly cleared in the metadata sync (`@ParentID_Clear = 1`); there is not one non-null `ParentID` seeded anywhere in the shipped migrations.
- **The integration suite has no IS-A coverage at all** — no check references `IsChildType`, `ParentEntityInfo`, or the chain save.

So there is nothing in core to point an integration test at, and adding an IS-A hierarchy to core purely as a test fixture would be inventing production metadata to serve a test. bizapps-orders is the right home, exactly as you suggested — that's where the relationships are real and where promotion can be driven against an actual database instead of a mock provider. Happy to take that on as a follow-up in that repo; it isn't something this PR can carry.

For what it's worth on coverage in the meantime: 2,162 MJCore tests pass, and the promotion path itself is covered at the unit level for both schema shapes (a child that mirrors the shared PK and one that leaves it entirely to routing), both guard throws, and the "failed attach leaves the record exactly as it was" contract.

rkihm-BC (approved):
Approving. @AN-BC's three asks are all genuinely addressed, and I traced the promotion path through `BaseEntity` rather than reasoning about it. Two observations at the end that are explicitly **not** change requests — I chased both down expecting them to be asks, and neither survived as one.

## The three follow-ups

**`AttachToParent`** reads better at the call site, and nothing shipped under the old name so there is no alias to carry.

**The `EntityInfo` role lookups** close real traps rather than saving keystrokes, which is what makes them worth adding:

- `IsChildType` alone treats an **intermediate** type as a leaf, so anything branching on it mishandles the middle of every chain deeper than two. `IsARole` is the right shape for that question, and `Intermediate` is precisely the value the boolean pair could not express.
- `ChildEntities` is direct-children-only, so a "find every subtype" written against it silently misses everything past the first level. `DescendantEntities` is the downward twin `ParentChain` always deserved, cycle-guarded the same way.

I checked `RootEntityInfo` specifically, because it is the one that could be quietly wrong: it returns `chain[chain.length - 1]`, which is only the root if `ParentChain` is ordered nearest-first. It is — documented and implemented as `[Meetings, Products]` for Webinars. Correct.

**The integration-test answer is the right one and it is verified rather than assumed.** MJ core models zero IS-A relationships — every core entity's `ParentID` is explicitly cleared in the metadata sync, and no non-null `ParentID` is seeded anywhere — so there is nothing in core to point a test at. Adding a hierarchy to core purely as a fixture would mean inventing production metadata to serve a test, which is worse than the gap. bizapps-orders is where the relationships are real.

## The mechanism

Loading the parent so the chain save becomes UPDATE-then-INSERT is the whole trick, and doing it **inside** the existing machinery rather than beside it is why nothing else had to move — routing, permissions, validation, disjoint enforcement and the transaction scope all run exactly as before. Only which row the parent save targets changes.

Two details are the good kind of hard-won:

- **The parent's primary keys drive the mirror, not the child's.** A child `EntityInfo` is not obliged to re-declare the shared key as its own, so iterating the child's list is a silent no-op on such a schema. Catching that in your own first draft, and covering both schema shapes, is the part that makes this trustworthy.
- **Capturing the minted key before `InnerLoad`**, because `InnerLoad` wipes before it reads. Without it a missing parent row leaves the fresh chain gutted rather than intact, which would have made the documented "caller decides" contract false.

## Two observations — explicitly not requested changes

**1. The failed-attach restore covers primary keys only.** `Set()` routes to the parent immediately, and routing is live from construction (`InitializeParentEntity()` runs inside `GetEntityObject()`, before any `NewRecord`/`Set`). So a parent-held field set *before* `AttachToParent` lives on the parent entity, and the restore path's `NewRecord()` blanks it while re-seeding only the captured PKs.

I chased what that actually costs and it is much less than it first looks. It requires the reverse of the call order your own JSDoc example shows, and for any **required** parent field the next `Save()` fails the parent's NOT NULL validation — loudly, which is the same failure the issue's repro describes. It is silent only for nullable parent fields, set before the attach, when the caller then chooses to save as a fresh chain.

So this is not worth code. What it is worth is one sentence: the contract says *"the record the caller holds is bit-for-bit the fresh record they built"*, and that is true of the keys rather than the record. This codebase holds its comments to a standard where that distinction matters — an overclaiming comment is the kind that gets trusted later by someone who did not read the implementation.

**2. Promoting a parent that already has this subtype is unguarded — and lands softer than I expected.** `EnforceDisjointSubtype` deliberately excludes the entity's own type, and is skipped entirely when the parent sets `AllowMultipleSubtypes`, so neither shape catches "this Person is already an Applicant". `AttachToParent` returns true and the child INSERT collides on the shared key.

I expected this to be an ask, because the motivating Caliber bug is a find-or-create driver and idempotent re-runs are exactly what such a driver does — and because the detection looked free: `InitializeChildEntity()` runs after every successful load, so by the time `AttachToParent` returns, the parent already knows its existing child without another query.

What stopped it being an ask is that `beginEntityTransactionScope()` opens a provider transaction for the unit of work rooted at the initiating entity, so the colliding child INSERT rolls the parent UPDATE back with it. The outcome is a clean rollback and a primary-key violation that names the child table — worse diagnostics than this PR's usual standard, but not a correctness problem, and nothing is left half-written. "The check would be cheap" is not the same as "the check is needed", and I had been treating it as if it were.

Worth recording as a known edge rather than fixing here. It also has an obvious home: the bizapps-orders integration work you already committed to is where re-promotion can be driven against a real database instead of a mock, and it is the natural place for a named error to earn its test.

Neither of these blocks anything. Good find, and the fix is in the right layer.

PR #3873, opened 2026-08-16
MJExplorer: the manifest prebuild fallback verifies coverage instead of shrugging (#3856)

### LINKED ISSUE #3856: MJExplorer manifest prebuild fails open: a missing Open App's screens look like a build that worked
Filed from **Caliber** (`MemberJunction/bizapps-caliber`), an Open App on MJ.

**Pinned against:** MJ 5.51.0 · **Severity:** small change, prevents a silent whole-feature outage

The MJExplorer template's manifest prebuild ends in a shell fallback:

```
mj codegen manifest --exclude-packages @memberjunction --output ./src/app/generated/class-registrations-manifest.ts \
  || echo 'Warning: mj codegen manifest not available, using existing manifest'
```

When the manifest step fails for any reason — CLI not installed in that environment, an unresolvable
dependency, a config it cannot read — the build continues and links the **stale committed manifest**.
For a host with installed Open Apps that is not a degraded build, it is a wrong one: the managed
`Open App client bootstrap` block is what makes an installed app's client classes register, so the
symptom is that the app's screens are simply absent, with a green build and no error anywhere. A
warning on stdout does not survive CI log volume, and nothing downstream re-checks.

This bites specifically because the block is *generated*, not authored — there is no file in the host
repo whose absence a reviewer would notice.

**The ask:** drop the `|| echo` (let a failed manifest fail the build), or keep the fallback but exit
non-zero when the existing manifest predates the current `dynamicPackages.client` list. Either makes
the failure loud at the moment it is cheap to fix.

**What Caliber ships today:** `apps/MJExplorer` runs the same command with **no** fallback, wired to
both `prebuild` and `prestart`, and `smoke/config/client-bootstrap.test.mjs` asserts the scripts
contain no `||` and that the generated manifest still carries both the bootstrap block and the two
dashboard classes MJ's class factory resolves. That protects our repo; it does nothing for a host
using the template, which is why the ask stands.

---

MS-BC:
Closes #3856. Filed from Caliber (`MemberJunction/bizapps-caliber`), where a host's stale manifest made an installed app's screens silently absent behind a green build.

## The defect

```
mj codegen manifest … || echo 'Warning: mj codegen manifest not available, using existing manifest'
```

When the manifest step fails for any reason, the build continues and links the **stale committed manifest**. For a host with installed Open Apps, the managed bootstrap block is what registers the app's client classes — so the symptom is the app's screens simply absent, with no error anywhere. The block is *generated*, not authored, so there is no file whose absence a reviewer would notice.

## The fix — the issue's option 2

The fallback still exists (the monorepo dev loop legitimately runs without the `mj` CLI), but it now runs `scripts/verify-manifest-fallback.mjs`, which allows the stale manifest **only when it demonstrably covers every enabled `dynamicPackages.client` entry** in `mj.config.cjs`:

- manifest missing or empty → **fail**
- manifest predates a configured package → **fail, naming the exact packages** ("Linking it would build an Explorer in which those apps' screens are silently absent")
- config unreadable → warn and allow (the CLI would have failed on it too; refusing would break the plain dev loop the fallback exists for)
- everything covered → warn and allow, stating what was checked

Proven both ways against the real manifest/config pair: **green** as-is, **red** (exit 1, package named) the moment a configured package is renamed out of the manifest.

## Scope

MJExplorer only — the template hosts copy, and the site the defect was observed on. Eight other packages carry the same `|| echo` pattern (`MJAPI`, `ServerBootstrap`, `Angular/Bootstrap`, …); the same guard pattern applies if wanted, but each has its own manifest path and config shape, so they are left to a follow-up rather than half-generalized here.


MS-BC:
Self-review before asking anyone to look: **the guard was not guarding anything.** Fixed in `5e001a9`.

It resolved its config as a sibling of the package —

```js
const configPath = resolve(here, "../mj.config.cjs");   // packages/MJExplorer/mj.config.cjs
```

— and there is no such file. `mj.config.cjs` lives at the **workspace root** (three levels up here; wherever the root is on a host). So `existsSync` was always false, `required` stayed `[]`, every manifest trivially covered an empty list, and the script exited 0 announcing it *"covers all 0 enabled dynamic client package(s)"*.

Measured rather than reasoned about — real repo, real manifest, `@mj-biz-apps/forms-ng` renamed out of it, which is the exact case this PR exists to catch:

```
original:  EXIT=0   "covers all 0 enabled dynamic client package(s)"
fixed:     EXIT=1   "  - @mj-biz-apps/forms-ng"
```

The "proven both ways" claim in the description was true only of a layout where the config sits beside the package, which this repo is not.

Two changes:

- The config is **located by walking up** from the package, so it resolves in the monorepo and in a host layout alike.
- An **absent** config is now distinguished from an unreadable one. Absence took the silent path — no warning, plus a success line claiming coverage of an empty list. That is precisely the green-build-nothing-verified shape this script was written to remove, reproduced inside the script itself. It now says it could not check.

Verified three ways against the real manifest/config pair: as-is exit 0 covering **2** packages (not 0); a configured package renamed out exit 1 naming it; restored exit 0.

MS-BC:
Closing. The premise does not hold for this package.

The fix is framed around the build silently linking a **stale committed manifest**, but for MJExplorer that manifest is gitignored:

```
.gitignore:186: packages/MJExplorer/src/app/generated/class-registrations-manifest.ts
```

There is nothing committed to go stale. In CI and in a fresh clone the file simply is not there — and the app imports it, so the Angular build already fails loudly on a missing module, guard or no guard. CI cannot meaningfully exercise this script at all, since the file it inspects never exists there.

What remains is a genuine but narrow case: a developer whose working tree still holds a manifest from an earlier build, where `mj` then fails and the stale local copy gets linked. Not worth a bespoke guard in this package.

Worth recording what the review turned up, since it outlives the PR:

- The guard as filed **never guarded anything** — it looked for `mj.config.cjs` beside the package, where no such file exists; `required` stayed empty, every manifest trivially covered it, and it exited 0 reporting *"covers all 0 enabled dynamic client package(s)"*. Proven with `@mj-biz-apps/forms-ng` renamed out of the real manifest: original `EXIT=0`, corrected `EXIT=1` naming the package.
- An **absent** config was not distinguished from an unreadable one, so absence printed a success line — the same green-build-nothing-verified shape the script was written to remove, reproduced inside the script itself.

The real target is the `|| echo` shrug itself, which eight other packages share (`MJAPI`, `ServerBootstrap`, `Angular/Bootstrap`, …). If that is worth closing, it deserves one guard that works for every one of them rather than a per-package copy — which is a different piece of work from this PR. Issue #3856 stays open for it.

Issue #3875, opened 2026-08-16
mj dev workspace should manage the Angular prebundle exclude for linked scopes


rkihm-BC:
Split out of the `next` install break triaged in #3849, so the capability isn't lost when the leak is reverted.

## The pattern

Joined-workspace host wiring keeps leaking into the public repo, because there is no supported place to put it. Three instances in about a day:

| Leak | Where | Outcome |
|---|---|---|
| `dynamicPackages` block | `mj.config.cjs` (#3842) | reverted by #3844 — the comment literally said NEVER COMMIT |
| `@mj-biz-apps/*` deps | `packages/MJAPI/package.json`, `packages/MJExplorer/package.json` (#3834) | broke `pnpm install` on every clean checkout |
| `@mj-biz-apps/*` prebundle glob | `packages/MJExplorer/angular.json` (#3834) | inert, but same class |

Reverting is right in each case, but reverting alone guarantees a fourth instance: the developer wanted a real thing, got it working locally, and the only way to keep it was to commit it.

## The specific gap

`mj dev workspace` already owns joined-workspace setup, but it does **not** touch `packages/MJExplorer/angular.json`, and no guide documents the setting. I checked — there are no references to `angular.json` or `prebundle` anywhere under `packages/MJCLI/src/lib/dev-workspace` or `packages/MJCLI/src/commands/dev`.

The setting in question:

```jsonc
// packages/MJExplorer/angular.json — serve.options
"prebundle": {
  "exclude": ["@memberjunction/*"]
}
```

Angular's Vite-based dev server pre-bundles dependencies into cached optimized chunks. A workspace-linked package that gets rebuilt under `pnpm run watch` is served from that cache, so your rebuild doesn't appear until the cache is invalidated. Excluding a scope keeps the dev server reading it fresh.

`@memberjunction/*` is committed and correct — it serves the repo's primary documented dev loop. A joined workspace needs exactly the same treatment for **its** linked scopes (`@mj-biz-apps/*` and any other host's), and that list is per-developer, so it cannot be committed.

## Proposal

Have `mj dev workspace` manage the prebundle exclude for the scopes it links, the same way it manages the rest of the join. Requirements:

- Add each linked scope's glob to `serve.options.prebundle.exclude` in the consuming app's `angular.json`, preserving `@memberjunction/*`.
- Make it reversible — `mj dev workspace` teardown (or whatever the leave/reset path is) removes exactly what it added.
[...]

PR #2985, opened 2026-06-30
feat(auth): metadata-driven pluggable authentication providers

AN-BC:
## Summary

Makes MemberJunction's authentication providers pluggable **the MJ way** — discovered via `@RegisterClass` and configured through metadata, environment variables, or the config file, all resolved through the same `ClassFactory`. Adding a provider is now a subclass plus a row. **Zero core edits.**

This is the companion refactor to the WorkOS provider PR (#2982), which is what exposed the problem: adding one provider the old way meant touching a hard-wired import list, an env-var block, and a type enum in core.

> **Status:** implemented end to end. Originally opened as plan + migration only; the engine, resolver, endpoint, picker, seed metadata and docs have all landed since.

## The core idea

A row (or config entry, or env var) names a `DriverClass`, resolved at runtime via
`MJGlobal.ClassFactory.CreateInstance(BaseAuthProvider, DriverClass, config)` — exactly how **File Storage Providers** (`ServerDriverKey`) and **AI Remote Browser Providers** (`DriverClass`) already work. The runtime validation machinery (`BaseAuthProvider`, JWKS, issuer routing, `MJAuthBase`) is **unchanged**; only the *source* of the provider list changed.

### No closed domain anywhere

The original complaint was the enum. Audited, and every surface is open:

| Surface | Status |
|---|---|
| `DriverClass` column | **No CHECK constraint** → CodeGen emits `z.string()`, not a union. A CHECK here would have recreated the enum inside the database. |
| `AuthProviderConfig.type` | `AuthProviderType \| string` — open |
| `environment.AUTH_TYPE` | `… \| (string & {})` — open |
| Resolution logic | **Zero** `switch`/`case`/`=== 'auth0'` branching anywhere in the resolution path |
| `AUTH_PROVIDER_TYPES` | Referenced only by its own definition. Constrains nothing; autocomplete only. |

## What a third party does now

1. Ship `@RegisterClass(BaseAuthProvider, 'my-idp')` — server validation
2. Ship `@RegisterClass(MJAuthBase, 'my-idp')` — browser login, if interactive
3. Configure it through **any** surface: a metadata row, a `mj.config.cjs` entry, or their own `ConfigFromEnvironment` static

No MJ source edit at any step.

### Schema — `migrations/v6/V202608160030__v6.1.x__Pluggable_Auth_Providers.sql`

New `__mj.AuthenticationProvider` → entity **`MJ: Authentication Providers`**.

[...]

PR #3876, opened 2026-08-16
fix(ci): unbreak next — do not ship BizApps workspace deps; sync lockfile

rkihm-BC:
## Summary

`next` cannot `pnpm install` on a clean checkout. Every job that installs — unit tests, integration, dependency check, adopted standards — fails within ~90 seconds before running anything. Part of #3849.

> **Scope:** this PR fixes the **install** breaks only. It does **not** make `next` green on its own — with install working, CI now reaches two further pre-existing failures that were masked behind it. See [What this does not fix](#what-this-does-not-fix).

There are **two** breaks, not one. They surface as a single error because pnpm stops at the first offending `package.json`, so fixing the reported one alone would have moved the failure rather than cleared it.

### Break 1 — BizApps workspace deps in MJAPI and MJExplorer

`27c2e46f9f` (*"feat(base-forms): enhance form chrome slot rendering, collapsible panels, and grid containers"*, #3834) added twelve `@mj-biz-apps/*` deps to `packages/MJAPI/package.json` and eight to `packages/MJExplorer/package.json`. They are unrelated to that commit's stated purpose and read as joined-workspace host wiring.

Four in each are `workspace:*`, naming packages that don't exist in this repo:

```
ERR_PNPM_WORKSPACE_PKG_NOT_FOUND  In packages/MJAPI:
"@mj-biz-apps/accounting-server@workspace:*" is in the dependencies
but no package named "@mj-biz-apps/accounting-server" is present in the workspace
```

`@mj-biz-apps/accounting-core-entities-server@0.1.0` is additionally a **404** on the registry, so even the pinned entries aren't uniformly resolvable — consistent with none of this being meant to resolve outside a joined workspace.

This is the same principle #3844 applied to the `dynamicPackages` block in `mj.config.cjs`: BizApps register on a consumer host through `mj app install` and must not be dependencies of the MJ repo. #3844 cleaned the leak from #3842; this one came from the earlier #3834 and was missed.

### Break 2 — `graph-view` missing from the lockfile

`47930ef712` (#3842) added `packages/Angular/Generic/graph-view/package.json` without running `pnpm install`, so `pnpm-lock.yaml` has no importer for it:

```
pnpm-lock.yaml is not up to date with <ROOT>/packages/Angular/Generic/graph-view/package.json
```

This is the error CI reports. It could not be fixed on its own: `pnpm install --lockfile-only` aborts on Break 1 before writing anything, which is why the lockfile had to wait for the dep removal.

### Also: the `angular.json` prebundle glob

`27c2e46f9f` added `"@mj-biz-apps/*"` to `serve.prebundle.exclude` in `packages/MJExplorer/angular.json`. Reverted here as part of the same leak.

Unlike the `package.json` entries this one is **inert** — dev-server only, no effect on any build configuration or production output, and a glob that matches nothing outside a joined workspace. It's removed because it's the same private host wiring in a public repo, not because it breaks anything.

**`"@memberjunction/*"` predates that commit and stays.** It's load-bearing for the standard MJ dev loop: it keeps `pnpm run watch` rebuilds of workspace-linked packages from being served stale out of Vite's prebundle cache. Only the added array element is removed.

Follow-up for the capability BC actually wanted: #3875.

## Changes

| File | Change |
|---|---|
| `packages/MJAPI/package.json` | −12 `@mj-biz-apps/*` deps |
| `packages/MJExplorer/package.json` | −8 `@mj-biz-apps/*` deps |
| `packages/MJExplorer/angular.json` | −1 array element from `serve.prebundle.exclude` |
| `pnpm-lock.yaml` | +40 lines — one new importer for `graph-view` |

The lockfile was **regenerated** with `pnpm install --lockfile-only`, not hand-edited. The diff is a single added importer and nothing else — no removals, no version churn.

No changeset: `mj_api` and `mj_explorer` are both `"private": true`, so nothing here publishes.

## Test plan

- [x] `pnpm install --lockfile-only` resolves cleanly on this branch (it fails on `next`)
- [x] All three edited files parse as valid JSON
- [x] No `mj-biz-apps` references remain in the three edited files
- [x] Lockfile diff limited to the `graph-view` importer
- [x] CI: install now succeeds — every job proceeds past `pnpm install` to real work
- [x] `Check migrations`, `Detect missing and unused dependencies`, and the peer-dep gate all pass
- [x] No source file imports any removed dep (only two MJCLI help-text strings mention the scope), independently confirmed by the dependency gate passing
- [ ] `Run unit tests` and `adopted standards` — **still failing, for unrelated pre-existing reasons.** See below.

## What this does not fix

With install unblocked, CI reaches two further failures. Both originate in `graph-view` from #3842, both predate this PR, and neither is caused by it — they were simply unreachable while install died first.

**1. `adopted standards` — 3 UI-layering violations**

```
packages/Angular/Generic/graph-view/src/lib/components/graph-view.component.ts:18
  imports "@memberjunction/ng-shared" — ng-shared is Explorer — that is L3
  imports "NavigationService" — navigation is L3 — emit an intent event instead
packages/Angular/Generic/graph-view/package.json
  declares "@memberjunction/ng-shared" — a "widgets" package must not depend on it
```

An L1 widget must not reach into Explorer. To be clear about what the fix costs — the service call and the event are **not** redundant:

```ts
if (this.navService) {
    this.navService.OpenEntityRecord(entityName, pk);   // performs the navigation
}
this.NodeNavigated.emit(new NodeNavigatedEventArgs(...)); // notifies the host it happened
```

`OpenEntityRecord` opens a record tab through the workspace manager, with app-context and dedup logic. The event only reports. The `inject(NavigationService, { optional: true })` is a deliberate pattern: the widget navigates automatically inside Explorer and degrades gracefully outside it.

So making this layer-correct is a real API change, not dead-code removal — the host would own navigation by wiring `(NodeNavigated)`, and clicking a node would do nothing until it does. Nothing consumes `<mj-graph-view>` today, so there is no breakage right now, but the default behavior of the component changes. That is @AN-BC's call, which is why it is not folded in here.

**2. `Run unit tests` — DOM coverage ratchet at 135 (gate 134)**

`graph-view.component.ts` is the single component over the line. The arithmetic is consistent: the two base-forms components put `next` at 136, `8433a25de6` deferred them back to 134, and graph-view adds one more.

Both belong to @AN-BC's #3842 rather than to a CI unbreak, so they are deliberately not folded in here — the layering item is a design decision about that component, not a build fix. Left for a follow-up so this PR stays a reviewable, revertable install fix.

## Knock-on

PR #2985 is blocked purely by inheriting these two breaks; its own checks are otherwise clean. Its unit suites have not executed for two runs running — first the DOM ratchet (#3819), now this.


rkihm-BC:
Filed the remaining `adopted standards` failure as **#3877** — it's pre-existing from #3842 and deliberately not addressed here.

Two things worth recording for reviewers, both from re-checking assumptions rather than guessing:

**The layering failure does not block this PR.** `next` has no classic branch protection; a repository ruleset governs it and requires exactly one status check — `Check migrations`, which passes here. `required_approving_review_count` is 0. `adopted standards`, `Run unit tests` and the integration tier are advisory. That's also how #3842 merged with four red checks, and how `next` came to be broken in four ways at once.

**The nav coupling in graph-view is not redundant code**, which I initially got wrong. `NavigationService.OpenEntityRecord` performs the navigation through the workspace manager; `NodeNavigated` only reports it. Making it layer-correct changes the component's default behavior, so it belongs to @AN-BC rather than in a CI unbreak. #3877 has the detail.

Status here: `Check migrations` ✅ (the only required check), dependency gate ✅, peer-dep gate ✅. The ratchet fix in `ca61460086` verified locally at 134 ≤ 134 — worth letting `Run unit tests` confirm that in CI before merging, since it's the one claim this PR makes that CI hasn't validated yet.

Issue #3877, opened 2026-08-16
graph-view: 3 ui-layers violations (NavigationService from Explorer) + 11 direct UUID comparisons


rkihm-BC:
`@AN-BC` — surfaced while triaging the `next` install break (#3849). The `adopted standards` gate reports three UI-layering violations in the `graph-view` component added by #3842:

```
packages/Angular/Generic/graph-view/src/lib/components/graph-view.component.ts:18
  imports "@memberjunction/ng-shared" — ng-shared is Explorer (NavigationService /
  BaseResourceComponent) — that is L3
  imports "NavigationService" — navigation is L3 — emit an intent event instead
packages/Angular/Generic/graph-view/package.json
  declares "@memberjunction/ng-shared" — a "widgets" package must not depend on it
→ https://github.com/MemberJunction/MJ/blob/next/guides/UI_LAYERING_GUIDE.md
```

`packages/Angular/Generic/**` is L1. Per the layering guide, nothing below L3 imports an Explorer package — so as it stands the component can only be used inside Explorer, which is the opposite of what its location advertises.

## This is not dead code, and the fix is not free

Worth stating plainly, because the two lines look interchangeable at a glance and they are not:

```ts
// graph-view.component.ts — NavigateToEntity(), lines ~795-805
if (this.navService) {
    const pk = CompositeKey.FromID(node.ID);
    this.navService.OpenEntityRecord(entityName, pk);      // ← PERFORMS the navigation
}
this.NodeNavigated.emit(new NodeNavigatedEventArgs(node, entityName, node.ID));  // ← REPORTS it
```

`NavigationService.OpenEntityRecord` (`navigation.service.ts:464`) opens a record tab through the workspace manager — resolving the active app, its color, the URL segment, and deduping against existing tabs. `NodeNavigated` only notifies. A host wired to the event alone gets a notification and no navigation.

And `inject(NavigationService, { optional: true })` (line 516) is clearly deliberate: navigate automatically inside Explorer, degrade gracefully anywhere else. Removing it is a real API change, not cleanup.

## Proposed fix, and what it costs

Move the navigation decision to L3:

- Drop the `NavigationService` import (line 18), the optional inject (line 516), the `if` block (lines 800-803), and `@memberjunction/ng-shared` from `graph-view/package.json`.
[...]

PR #3879, opened 2026-08-17
fix(graph-view): compare record IDs with UUIDsEqual, not ===

rkihm-BC:
## Summary

`UUIDCompliance.test.ts` fails with 11 direct UUID comparisons, all in `graph-view.component.ts`. Because that test lives in `@memberjunction/global` and scans the whole repo, the failure lands on a package nobody edited — and turbo fails fast, so ~9 further test tasks never run. That's currently the only thing keeping the unit-test job red on `next`.

Refs #3877 (UUID half only — the UI-layering half of that issue is untouched).

## Twelve comparisons, not eleven

Line 898 was **not** flagged by the gate:

```ts
return this.Edges.filter(e => e.SourceID === id || e.TargetID === id).length;
```

The test's pattern looks for `.ID ===`; these read `.SourceID ===` / `.TargetID ===`. Same UUIDs, same defect, invisible to the check. Fixing only the flagged 11 would have turned the gate green while leaving a real bug in place, so it's fixed here too.

## Why this is semantics, not lint

These are entity record primary keys — `NavigateToEntity` passes `node.ID` through `CompositeKey.FromID()` into `OpenEntityRecord`. SQL Server returns UUIDs uppercase, JS-generated ones are lowercase, and `===` silently fails to match across that boundary. Node selection, edge highlighting and hop expansion would all miss depending on where the ID came from.

## The template needed a method

An Angular template can't call an imported function, so:

```diff
- @if (SelectedNode?.ID === node.ID) {
+ @if (IsNodeSelected(node)) {
```

with a small `IsNodeSelected(node)` doing the `UUIDsEqual` comparison.

**Null semantics unchanged.** `UUIDsEqual` returns false when one side is null/undefined and the other isn't — exactly what `SelectedNode?.ID === node.ID` did when nothing was selected.

## Verification

- [x] `UUIDCompliance.test.ts` passes (1/1)
- [x] `tsc --noEmit` clean
- [x] **`ngc` builds clean** — this one matters: `tsc` alone does not type-check the template, which is the part that changed
- [x] No raw `.ID ===` / `.SourceID ===` / `.TargetID ===` comparisons remain in the file
- [x] `@memberjunction/global` was already a declared dependency — no manifest or lockfile change

`.includes()` on line 862 is a label substring search, not a UUID comparison, and is deliberately left alone.

## Knock-on

Unblocks the unit-test job, which has now failed three times running on #2985 for three unrelated reasons — the DOM ratchet (#3819), the install break (#3849), and this. Its own tests still have never executed in CI.

### DIFF
--- a/.changeset/graph-view-uuid-comparisons.md
+++ b/.changeset/graph-view-uuid-comparisons.md
@@ -1,0 +1,11 @@
+---
+"@memberjunction/ng-graph-view": patch
+---
+
+Compare graph node/edge record IDs with `UUIDsEqual()` instead of `===`.
+
+These are entity primary keys — `NavigateToEntity` passes `node.ID` through `CompositeKey.FromID()` into `OpenEntityRecord` — and SQL Server returns UUIDs uppercase while JS-generated ones are lowercase, so `===` silently failed to match across that boundary. Node selection, edge highlighting, hop expansion and edge-count all missed depending on where the ID came from.
+
+Twelve comparisons changed. Eleven were reported by `UUIDCompliance.test.ts`, whose failure also took the whole unit-test job down (it runs in `@memberjunction/global` but scans the repo, and turbo fails fast). The twelfth — `GetNodeEdgeCount`'s `e.SourceID === id || e.TargetID === id` — was not flagged, because the test's pattern matches `.ID ===` and those read `.SourceID ===` / `.TargetID ===`: the same UUIDs and the same defect, invisible to the gate.
+
+The template comparison became a component method (`IsNodeSelected(node)`), since an Angular template cannot call an imported function. Null semantics are unchanged: `UUIDsEqual` treats null-vs-value as false, exactly as `SelectedNode?.ID === node.ID` did when nothing was selected.
--- a/packages/Angular/Generic/graph-view/src/lib/components/graph-view.component.ts
+++ b/packages/Angular/Generic/graph-view/src/lib/components/graph-view.component.ts
@@ -17,6 +17,7 @@ import { CommonModule } from '@angular/common';
 import { FormsModule } from '@angular/forms';
 import { NavigationService } from '@memberjunction/ng-shared';
 import { CompositeKey } from '@memberjunction/core';
+import { UUIDsEqual } from '@memberjunction/global';
 import type {
     GraphNode,
     GraphEdge,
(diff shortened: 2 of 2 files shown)


rkihm-BC:
## Self-review note: this adds allocations to a 60fps path

Found reviewing my own diff. It doesn't change my recommendation to merge, but it should be on the record rather than rediscovered by whoever profiles a laggy graph.

`SimulatePhysics` drives a `requestAnimationFrame` loop that calls `markForCheck()` every frame (lines ~682-689), so the template — including the new `IsNodeSelected(node)` per node — is re-evaluated up to 60×/second while the layout settles.

Tracing `UUIDsEqual` against that:

| State | Cost |
|---|---|
| **No selection** (`SelectedNode === null`) | Hits the `uuid1 == null` guard, returns false. **Zero allocation** — identical to the old `===`. |
| **With a selection** | The selected node hits the `uuid1 === uuid2` fast path. Every *other* node falls through to `trim().toLowerCase()` on both sides — **two string allocations per node, per frame.** |

For a 500-node graph mid-simulation that's roughly 60k short-lived strings/second. Correct, but it is a real cost that didn't exist before, in the hottest path in the component.

## Two cheaper options, deliberately not taken

**Reference equality.** `SelectedNode` is only ever assigned `null` (lines 561, 745) or an element of `Nodes` (line 589 via `.find`, line 775 from the template's own iteration). It is therefore always reference-identical to a member of the array, so `this.SelectedNode === node` would be correct and allocation-free — and invisible to the compliance test, since it isn't an `.ID` comparison.

The catch: it silently degrades to "nothing is ever selected" if a host ever passes cloned or re-serialized nodes, and that failure is quiet. `UUIDsEqual` cannot fail that way. Given the component is new and its hosts unknown, I'd rather pay allocations than take a correctness risk that shows up as a mysteriously dead selection ring.

**Cache a normalized selected ID.** Normalize once when `SelectedNode` is assigned and compare the template against that — keeps value semantics, one normalization instead of N. Strictly better than the current form, just more moving parts (three assignment sites to keep in sync).

## Why I'd still merge as-is

The zero-selection case — the common one, and the one that runs during the initial layout settle — is free. The cost only appears while a node is selected, and real graph sizes here are unknown; nothing consumes `<mj-graph-view>` yet. Optimizing against a guess is worse than recording the tradeoff.

If profiling later shows this matters, the normalized-cache option is the one to take, and this comment is the context for why.

Issue #3685, opened 2026-08-09
Audit raw error-object logging in the Box and Dropbox storage drivers


MS-BC:
## Summary

Two storage drivers log a caught SDK error object wholesale under a `fullError` key:

- `packages/MJStorage/src/drivers/BoxFileStorage.ts:1598` — `fullError: uploadError`
- `packages/MJStorage/src/drivers/DropboxFileStorage.ts:525` — `fullError: JSON.stringify(error, null, 2)`

Both sit on file-operation failure paths, and both serialise the entire error rather than a chosen set of fields.

## Why this is worth a look

#3642 removed the equivalent pattern from the GraphQL transport and from five OAuth2 token-endpoint call sites, including this package's own `SharePointFileStorage`. The reasoning there generalises: an error object produced by an HTTP client is not a safe thing to serialise, because it frequently carries the originating request — and a request carries whatever was sent to authenticate it. Logging it writes that content to whatever the process's console output is captured to, which on a server is typically a persistent file that outlives any later rotation of the value.

These two sites were found during that work and deliberately left out of scope: they are a different shape from the token-endpoint case (a raw SDK error from a *file* operation, not an OAuth response body), so the helper introduced in #3642 does not apply to them.

## Open question — this is the actual task

**It has not been verified whether the Box and Dropbox SDK error objects carry authentication material.** Many HTTP clients attach the request configuration — including headers — to the error they throw, but that has not been confirmed for these two SDKs. This should be established before anything is changed, because it decides whether this is a hardening tidy-up or something more time-sensitive.

Concretely:

1. Determine what `uploadError` (Box) and `error` (Dropbox) actually contain at those two lines — specifically whether either carries request headers, a request config object, or a body.
2. If they do: replace wholesale serialisation with an explicit set of diagnostic fields, following the approach in #3642 (`SanitizeGraphQLError` builds a fresh object from an allowlist rather than deleting fields, so a future change to the upstream error shape cannot silently widen what is logged).
3. If they do not: still consider narrowing the log to named fields, since the current form has no upper bound on what a future SDK version might include.

## Suggested approach

`@memberjunction/global` already gained `describeTokenEndpointFailure` in #3642 for the OAuth-specific case. A generic counterpart — "build a log-safe view of an arbitrary HTTP client error" — would cover these two sites and give future drivers something to reach for. Worth a look at whether one helper can serve both.

## Notes

- Related: #3642
[...]

PR #3883, opened 2026-08-17
fix(storage): a failed upload reports the driver's real cause, not "Storage upload failed."

MS-BC:
## What happened

A realtime recording upload came back to the client as:

```json
{"Success":false,"FileID":null,"ErrorMessage":"Storage upload failed."}
```

The real cause was Google's, and it was actionable: *"Service Accounts do not have storage quota. Leverage shared drives instead."* Nobody saw it. Observed live on 2026-08-16 while proving the recording path from an Open App.

## Three layers, three swallows

| # | Where | What was lost |
|---|---|---|
| 1 | `GoogleDriveFileStorage.PutObject` | the googleapis error → bare `console.error`, `return false`. The cause dies here. |
| 2 | `FileStorageEngine.UploadFile` | throws its own path-only generic |
| 3 | `storeRealtimeRecording` | logs, then returns `string \| null` — the reason it just logged cannot leave |
| 4 | `RealtimeClientSessionResolver` | reports `'Storage upload failed.'` |

Layer 4 is the symptom people see; layer 1 is where the information actually died.

## What this changes — and what it deliberately does not

**`PutObject`'s `Promise<boolean>` contract is untouched.** `FileStorageBase` documents boolean-means-success and every driver implements it; making it throw would be a breaking change across all drivers and callers. Every `return false` / `return {objects:[],prefixes:[]}` / `return true` in this diff is byte-for-byte what it was.

1. **The Drive driver stops discarding the SDK's message.** A new `describeGoogleApiError` extracts the *named* fields — `code`, `message`, `errors[].reason`, `errors[].message`, `response.data.error.message` — and the 11 cause-discarding catches now `LogError` with it. The four catches that rethrow a generic (`GetObjectMetadata`, `GetObject`, `StreamObject`, `SearchFiles`) append the cause to the thrown message, matching the correct precedent already in this file at `CreatePreAuthDownloadUrl`.

   Sample output for the live failure: `code 403 — Service Accounts do not have storage quota. Leverage shared drives instead. — storageQuotaExceeded`

2. **`storeRealtimeRecording` lets the reason out** — returns `{ FileID, ErrorMessage }` instead of `string | null`. It still logs; it just no longer erases.

3. **The resolver reports the propagated reason**, falling back to `'Storage upload failed.'` only when no layer supplied one.

4. **`base-agent`'s server-bridged capture logs the failure** instead of silently skipping its success log.

## Written to satisfy the guard that already exists

MJStorage ships `rawErrorLogging.guard.test.ts` — a source guard forbidding drivers from logging a vendor error wholesale (no `fullError:`, no `JSON.stringify(error)`, no `|| error`). The extractor is a named-field **allowlist** for exactly that reason. That guard passes (25 tests) and now covers the Drive driver's new logging too.

## New public surface

```ts
// @memberjunction/storage
export function describeGoogleApiError(error: unknown): string;

// @memberjunction/ai-agents
export interface StoreRealtimeRecordingResult {
    readonly FileID: string | null;
    readonly ErrorMessage: string | null;
}
```

No `any` — the extractor narrows from `unknown` through type guards.

## Verification

| | exit |
|---|---|
| `turbo build --filter=@memberjunction/storage...` | **0** (157 tasks) |
| `turbo build --filter=@memberjunction/server` | **0** (162 tasks) |
| `packages/MJStorage` tests | **0** — 5 files, 80 passed |
| `packages/AI/Agents` tests | **0** — 104 files, 1900 passed |
| `packages/MJServer` tests | **0** — 1064 passed, 56 skipped |

`dist/` verified newer than sources; the new symbols appear in the emitted `.d.ts`.

MJServer's suite failed on the first run — the existing resolver test mocked the old `string | null` return. That is the widened contract being caught by the suite, which is the right kind of failure; the mock was updated.

New tests: the extractor against a realistic nested googleapis quota error, a `response.data.error`-only shape, a plain `Error`, and non-error values; plus `storeRealtimeRecording` surfacing the real quota text rather than a bare null.


packages/MJStorage/src/drivers/GoogleDriveFileStorage.ts
```diff
@@ -53,6 +54,81 @@ interface ServiceAccountCredentials {
...
+        code?: number | string;
+        message?: string;
+        errors?: Array<{ reason?: string; message?: string; domain?: string }>;
+      };
+    };
+  };
+}
+
+/**
+ * Builds a log-safe, bounded, one-line description of a Google Drive API failure.
```
rkihm-BC, line 82: `parts.join(' — ')` returns whatever the SDK supplied — there's no length cap, so "bounded" claims a property the function doesn't have. (`describeBoxError` doesn't truncate either, so this is consistent with the codebase; only the word is wrong.)

Dropping the word beats adding a cap: truncating a vendor error risks cutting the actionable sentence in half, and the two remaining claims are both true and worth making — log-safe because it's an allowlist, one-line because the parts join with a separator rather than newlines.

```suggestion
 * Builds a log-safe, one-line description of a Google Drive API failure.
```

rkihm-BC (approved):
Approving. This is a model example of the genre — one wording change requested, no code change.

## What's good

**The layering diagnosis is precise.** Four swallow points identified, with the observation that layer 4 is the symptom people see and layer 1 is where the information actually died. The live cause — *"Service Accounts do not have storage quota. Leverage shared drives instead."* — is exactly the kind of actionable message a generic wrapper destroys, and exactly the kind nobody can guess from `"Storage upload failed."`

**It respects the existing contract rather than breaking it.** `PutObject`'s `Promise<boolean>` is untouched, because `FileStorageBase` documents boolean-means-success and every driver implements it. Every `return false` is byte-identical; only the logging and the thrown message gained the cause. Making it throw would have been a breaking change across all drivers and callers, and the PR says so explicitly rather than leaving the reader to wonder whether it was considered.

**The allowlist extractor is the correct security posture, and the precedent is real.** I verified `describeBoxError` exists in `BoxFileStorage.ts` on `next`, and that `rawErrorLogging.guard.test.ts` is a live source guard forbidding `fullError:` keys and `JSON.stringify(error)` — added after #3685, where drivers were dumping whole vendor errors. That matters because an error raised by an HTTP client carries the originating request, and a request carries whatever authenticated it. Naming the fields makes the logged set a property of this file that an SDK upgrade cannot widen. `describeGoogleApiError` follows that pattern exactly, and the guard now covers the Drive driver's new logging too.

**Good detail in the extraction loop:** `detail.message` is only appended when it "says something new," which avoids the common Google case where `errors[].message` repeats the top-level sentence verbatim. And the four catches that rethrow a generic append the cause to the thrown message rather than only logging it, matching the `CreatePreAuthDownloadUrl` precedent already in the file.

**No `any`** — the extractor narrows from `unknown` through type guards, which is what makes the allowlist enforceable rather than aspirational.

## One wording change requested

The doc comment and the PR description both describe the output as **bounded**, and it isn't — `parts.join(' — ')` returns whatever the SDK supplied, with no length cap. In fairness the cited precedent `describeBoxError` doesn't truncate either, so this is consistent with the codebase rather than novel; the issue is only that the word claims a property the code doesn't have.

I'd rather drop the word than add a cap. Truncating a vendor error risks cutting the actionable sentence in half, and "log-safe, one-line" is the honest and still-valuable claim: log-safe because it's an allowlist, one-line because the parts are joined with a separator rather than newlines. Suggestion inline on the doc comment.

Same edit worth making in the PR description, where "log-safe, bounded, one-line" appears in the summary of what `describeGoogleApiError` does — the description is what a future reader reaches for first.

## On the changeset

`ai-agents: minor`, `storage: patch`, with the last commit *"version the three packages this PR actually changes."* The diff also touches `packages/MJServer/src/resolvers/RealtimeClientSessionResolver.ts` — worth a sanity check that MJServer is covered by a dependency bump rather than needing its own entry.

Nothing else. Good fix, and the reasoning for what it deliberately *doesn't* change is as useful as the change itself.

PR #3892, opened 2026-08-17
perf(global): build the ClassFactory resolution diagnostic only when it is read

MS-BC:
## What

`ClassFactory.resolveAndInstantiate` builds its resolution-failure diagnostic **unconditionally**, before any gate:

```ts
const requiresSubclass = ClassRequiresSubclass(baseClass);
const reason = this.describeResolutionFailure(baseClass, key, requiresSubclass);  // always runs
...
if ((keyWasSupplied && !fallbackIsDesigned) || requiresSubclass) {                // gate is here
    this.reportResolutionFailure(baseClass, key, requiresSubclass, reason);
}
```

`describeResolutionFailure` calls `GetAllRegistrations` (an unmemoised `filter` over **every** registration in the process), then `.map` → `.filter` → `new Set` → `Array.from` → `.sort()` → `.map` → `.join()` into a multi-KB template literal.

## Why it matters

That work is discarded in the two hottest cases:

1. **The designed path.** `fallbackIsDesigned` covers `@OptionalKeyedSpecialization` bases — per the comment already in the file, exactly the EntityField `'<Entity>.<Field>'` hydration probe, i.e. **once per field of every entity**. The gate skips reporting; the string was already built.
2. **A suppressed report.** `reportResolutionFailure` returns early on `MAX_FALLBACK_REPORTS_PER_BASE` and again on the per-`base|key` dedup — both **before** it formats anything.

Measured on a live MJAPI process with 863 registrations: **~0.9 ms per call (868,869 ns for `BaseEntity`), ~52% of process CPU** in a 60-second sampled profile. It's invisible in normal operation precisely because the warning dedupes — the cost is in building a string nobody prints.

## The change

- `reason` becomes a memoised thunk; `reportResolutionFailure` takes `() => string` and calls it only after both early returns.
- `@RequiresSubclass` path resolves it eagerly — `CreateInstance` throws `result.Reason`, so it is genuinely wanted there.
- Standalone-fallback path exposes `Reason` through an **enumerable accessor**, so existing consumers (the pattern is documented on `CreateInstance`), spread, and `JSON.stringify` behave exactly as before.
- Also hoists `baseClassName` and the normalised key out of the `GetAllRegistrations` filter predicate — same hot path, and it was re-reading `(baseClass as {name}).name` once per registration.

No behaviour change: same warnings, same messages, same dedup and cap semantics.

### A second, independent change: `GetGlobalObjectStore` (`packages/MJGlobal/src/util.ts`)

Same profile, same class of problem, different function — called out separately because it is not part of the ClassFactory work and it is the one piece here that is not byte-identical in behaviour.

The previous shape was `if (window)` inside a `try`/`catch`. In Node, `window` is an **undeclared identifier**, so that line throws a `ReferenceError` on *every single call*, and the catch produces the correct answer via the exception path. This sits under ClassFactory and BaseEngine hot paths, and building and unwinding that exception was several percent of a busy server process.

`typeof window` is the one probe that is legal on an undeclared identifier, and the environment is fixed at startup, so the answer is computed once and memoised (`undefined` = not yet computed, `null` = computed and genuinely neither browser nor node).

The guard is `typeof window !== 'undefined' && window`, not a bare `typeof` check. `typeof null === 'object'`, so a **declared-but-null** `window` — which SSR shims in the wild do produce — would pass a bare probe and be memoised as the store, permanently, where `if (window)` correctly fell through to `global`. Nothing in this repo triggers it; the guard is there because MJ is embedded by callers whose environment we do not control, and a memoised wrong answer is unrecoverable for the process.

## Tests

`packages/MJGlobal/src/__tests__/ClassFactory.test.ts` — 100 pass, 4 added:

- the diagnostic is **not** built on the designed `@OptionalKeyedSpecialization` path
- a volume-capped report does not build it either
- `Reason` still returns the identical string when read, and is stable across reads
- `Reason` remains an enumerable own property (spread / `Object.keys` unchanged)

The volume-cap test asserts the **exact** describe count, `toHaveBeenCalledTimes(3)`. `MAX_FALLBACK_REPORTS_PER_BASE` is 3, and for 40 distinct keys on one unmarked base the count is exactly 3: `seen` 1–3 clear the cap with a distinct `logKey` each and describe, `seen` 4 prints the suppression notice and returns, `seen` 5–40 return on the cap. A loose bound would still pass with the laziness broken for most of the probes, which is the opposite of what the test is for; the exact number also guards the cap against a silent change.

`packages/MJGlobal/src/__tests__/util.test.ts` — 117 pass, 4 added for `GetGlobalObjectStore`:

- resolves to Node's `global` specifically, not merely something non-null
- the memo returns an identical reference on every call
- the memo intentionally survives a later environment change (stated as a test because it is the one surprise memoising introduces)
- a declared-but-null `window` falls through to `global` — verified to fail without the `&& window` guard, returning `null`

The file's three pre-existing `GetGlobalObjectStore` tests predate this PR and pass either way: they assert the result is non-null, defined, and writable. None of them pin the probe or the memo, which is why the four above exist.

## Context

Found while profiling sync throughput on a production MJ deployment; this was the single largest CPU consumer. Related work from the same investigation (separate PRs to follow): `GetIntegrationObjectFields` rescanning per record, and the telemetry env override that cannot fire because `DEFAULT_SERVER_CONFIG` always supplies the key.


packages/MJGlobal/src/ClassFactory.ts
```diff
@@ -339,7 +339,20 @@ export class ClassFactory {
...
+        // multi-KB string. That is affordable for a real failure and ruinous here, because this
+        // fallback path is DESIGNED to be hit: an '@OptionalKeyedSpecialization' base probes
+        // '<Entity>.<Field>' once per field of every entity, and neither of the two consumers below
+        // ends up reading the string on that path — the gate at the end of this block skips
+        // reporting, and CreateInstance never reads Reason when an instance was produced.
+        // reportResolutionFailure ALSO returns early (volume cap, then per-base+key dedup) before it
+        // formats anything, so passing an already-built string wasted the work even when reporting.
+        let describedReason: string | undefined;
+        const reason = (): string =>
+            (describedReason ??= this.describeResolutionFailure(baseClass, key, requiresSubclass));
```
MS-BC, line 355: wraps call in a closure so that only if describedReason is undefined must the expensive resolution failure be called, not before call time indefinitely

rkihm-BC (commented):
Reviewed the full diff, both commits, and verified the platform behavior around it. **One change I'd like before merge** (the volume-cap assertion, below); everything else is small.

## What's good

**The diagnosis is precisely targeted, and I confirmed why.** `GetRegistration` is *already* memoised via `_registrationCache`, and it caches negative results too — so the `<Entity>.<Field>` probe path was already O(1) on lookup after the first call per key. The un-cached residual really was the unconditional string build, which sits *after* that cache and therefore ran on every single call. This isn't a shotgun optimization; it lands on the one thing the existing cache didn't cover.

**The laziness is structurally correct.** The thunk memoises per `resolveAndInstantiate` call, `reportResolutionFailure` invokes it only after both early returns (volume cap, then per-`base|key` dedup), and the `@RequiresSubclass` branch resolves eagerly — which is right, because `CreateInstance` throws `Reason` there.

**The `Reason` accessor preserves observable behavior, and it's tested rather than asserted.** Enumerable own property, so spread / `Object.keys` / `JSON.stringify` are unchanged. I also checked the other half of that risk: nothing in the repo assigns to a ClassFactory result's `.Reason`, so a getter with no setter can't introduce a strict-mode throw. The `.Reason =` hits elsewhere are all unrelated types (PermissionEngine results, tree events, computer-use responses).

**The `util.ts` find is the better catch of the two.** `if (window)` on an *undeclared* identifier throws a `ReferenceError` on every call in Node, with the catch producing the correct answer via the exception path. That's a genuinely pathological hot path, `typeof` is the one legal probe on an undeclared identifier, and memoising is sound since the environment is fixed at startup. The `undefined` = not-computed / `null` = computed-none sentinel split is handled correctly.

**The `GetAllRegistrations` hoist is behavior-preserving** — I checked the edge cases (empty-string key, whitespace-only key) and the normalised form is equivalent to the inline expression in all of them.

Changeset is `patch` on `@memberjunction/global`, correct — no migration, no metadata.

## Requested before merge — the volume-cap test's assertion is nearly vacuous

`MAX_FALLBACK_REPORTS_PER_BASE = 3`, so for 40 distinct keys on one unmarked base the describe count is exactly **3**:

- `seen` 1–3 → past the cap, distinct `logKey` each time → emit → describe
- `seen` 4 → hits `seen === MAX + 1`, prints the suppression notice, **returns**
- `seen` 5–40 → `seen > MAX`, returns

The test asserts `expect(describeSpy.mock.calls.length).toBeLessThan(40)`. That passes at 39 — meaning it would still pass if the laziness were broken for 39 of the 40 probes. For the test whose entire purpose is proving that *suppressed reports don't build the string*, that isn't a guard.

`expect(describeSpy).toHaveBeenCalledTimes(3)` pins the actual contract and doubles as a regression guard on the cap itself. I traced the counter and the gate (`keyWasSupplied && !fallbackIsDesigned` is true for `NoisyBase`) to confirm the number.

## Smaller notes

**`typeof window !== 'undefined'` isn't equivalent to `if (window)` for one input.** If `window` is *declared and set to `null`*, the old code took the falsy branch and returned `global`; the new code sees `typeof null === 'object'`, returns `null` as the store, and memoises that so every later call returns null too.

I checked — there's no trigger for this in the repo, so it isn't a live bug. Raising it only because MJ is a platform others embed, SSR shims that null out `window` do exist in the wild, and the changeset claims byte-identical behavior. One-token fix: `typeof window !== 'undefined' && window`.

**The PR body doesn't describe the `util.ts` change.** The What / Why / The change / Tests sections cover only the ClassFactory diagnostic; `GetGlobalObjectStore` arrived in a later commit and appears only in the changeset. A reviewer working from the description wouldn't know a second, independent behavioral change is in the diff — and it's the one carrying the semantic difference above.

**No test for `GetGlobalObjectStore`.** The ClassFactory change gets four well-chosen tests; the util change gets none. It's a memoised module-level singleton every consumer of the global store depends on, and the memo is the part most likely to surprise someone later (a test that swaps environments mid-file would now get a stale answer). Three small cases would be proportionate: Node returns `global`, repeat calls return the same reference, and the memo is populated once.

## On the overlap with #3352

Both PRs edit `ClassFactory.ts` and both are yours, so worth stating explicitly: **they do not conflict.** #3352 touches lines 6, 255, 278 (`Register` and the new collision-warning method); this PR touches 339, 363, 404, 433, 447 (`resolveAndInstantiate`, `reportResolutionFailure`, `GetAllRegistrations`). Git will auto-merge, and there's no logical interaction either — #3352's warning runs at registration time and never reaches the diagnostic path. These can land in either order.

## Bottom line

Small, well-aimed, correctly reasoned, and the expensive claim is backed by a profile rather than intuition. Just the one test assertion before merge — it's a one-line change in a file this PR already edits. The `window === null` nit and the missing util test are cheap and worth doing; the body update is housekeeping.

packages/MJGlobal/src/__tests__/ClassFactory.test.ts
```diff
@@ -898,6 +898,48 @@ describe('ClassFactory', () => {
...
+      }
+      const describeSpy = vi.spyOn(
+        factory as unknown as { describeResolutionFailure: () => string },
+        'describeResolutionFailure'
+      );
+      for (let i = 0; i < 40; i++) {
+        factory.TryCreateInstance<NoisyBase>(NoisyBase, `Entity${i}.Field${i}`);
+      }
+      // Far fewer describes than probes: only the reports that actually got emitted.
+      expect(describeSpy.mock.calls.length).toBeLessThan(40);
```
rkihm-BC: **Requested before merge.** This passes at 39 — so it would still pass if the laziness were broken for 39 of the 40 probes, which is the opposite of what the test name claims.

`MAX_FALLBACK_REPORTS_PER_BASE = 3`, and the exact count here is **3**: `seen` 1–3 emit (distinct `logKey` each time), `seen` 4 hits `seen === MAX + 1` and returns after the suppression notice, `seen` 5–40 return on `seen > MAX`.

```suggestion
      expect(describeSpy).toHaveBeenCalledTimes(3);
```

That pins the actual contract and doubles as a regression guard on the cap itself.

MS-BC:
All four addressed in `46333a7` (rebased onto the branch tip).

**The volume-cap assertion — you're right, and the number is 3.** I re-traced it independently rather than take it: `seen` 1–3 clear the cap and each carries a distinct `logKey`, so all three describe; `seen` 4 hits `seen === MAX + 1`, prints the suppression notice and returns; `seen` 5–40 return on `seen > MAX`. Now `expect(describeSpy).toHaveBeenCalledTimes(3)`, with a comment recording why the loose bound was the wrong assertion for this particular test.

**`window === null` — taken, one token as you said.** Now `typeof window !== 'undefined' && window`. I agree it isn't a live bug in this repo; the reason I took it anyway is the one you gave — the answer is *memoised*, so unlike the old `if (window)` this would be unrecoverable for the life of the process rather than wrong once per call. That asymmetry is what makes it worth a token. Documented at the guard so nobody "simplifies" it back.

**PR body — fixed.** `GetGlobalObjectStore` now has its own subsection under The change, explicitly flagged as a second and independent change, with the `typeof null === 'object'` difference spelled out rather than left in the changeset.

**Tests for `GetGlobalObjectStore` — one correction, and then agreement.** There are three already, in `util.test.ts`. They pass either way: non-null, defined, writable. So the coverage gap is real — nothing pinned the probe or the memo — but it wasn't zero, and I'd rather say so than let "no test" stand.

Four added, close to what you proposed plus one:

- resolves to `global` specifically, not merely something non-null
- the memo returns an identical reference across calls
- the memo survives a later environment change *by design* — stated as a test because it's exactly the stale-answer surprise you flagged, so it should fail loudly if someone changes it rather than silently
- a declared-but-null `window` falls through to `global`

That last one I verified is a real guard, not decoration: reverting `&& window` makes it fail with `expected null to be [global]`. 217 tests pass in `MJGlobal`.

**On #3352:** agreed, and thanks for checking the line ranges — no conflict, no logical interaction, either order.

packages/MJGlobal/src/__tests__/ClassFactory.test.ts
MS-BC: Taken — now `expect(describeSpy).toHaveBeenCalledTimes(3)`. I re-traced the count independently and it is exactly 3, for the reason you gave. Comment added recording why the loose bound was the wrong assertion for this test specifically. `46333a7`.

PR #3895, opened 2026-08-17
perf(integration): index integration-object fields and memoise the per-record view

MS-BC:
## What

Two layers of the same hot path, both on per-record code:

**`IntegrationEngineBase.GetIntegrationObjectFields`** — an unindexed scan of every `IntegrationObjectField` in the process:

```ts
return this._integrationObjectFields.filter(f => UUIDsEqual(f.IntegrationObjectID, objectID));
```

**`BaseRESTIntegrationConnector.GetCachedFields`** — adds a filter and a sort on top, per call:

```ts
return IntegrationEngineBase.Instance.GetIntegrationObjectFields(objectID)
    .filter(f => f.Status === 'Active')
    .sort((a, b) => a.Sequence - b.Sequence);
```

`GetCachedFields` is called from `RawToExternalRecord` / `TransformRecord`, i.e. **once per record transformed** (`BaseRESTIntegrationConnector.ts:260, 283, 297, 418, 726, 767, 812, 1078`), and `:767`/`:812` then run a `.find(f => f.IsPrimaryKey)` over the freshly-sorted array.

## Cost

Live CPU profile of a production MJAPI process syncing a 364-object catalog:

| frame | share |
|---|---|
| `get IntegrationObjectID` | 23.65% |
| `BaseEntity.Get` | 10.99% |
| `EntityInfo.FieldByName` | 6.77% |
| `UUIDsEqual` | 4.76% |

≈ **46% of process CPU inside one lookup.** Each element of the scan invokes the generated `IntegrationObjectID` getter → `BaseEntity.Get` → `FieldByName` (trim + lowercase per call) → `UUIDsEqual`; the second filter and the sort then pay the same tax again.

The event loop was pegged near 1.15 cores. That is what made simple indexed seeks measure 321–541 ms client-side while the database sat at 35% CPU with zero pending pool acquires — the time was in the process, not the query.

## The change

- **engine-base**: lazily-built `objectID → fields` Map. Keys are normalised `trim().toLowerCase()`, matching how `UUIDsEqual` compares, so SQL Server's uppercase and PostgreSQL's lowercase land in the same bucket. The `objectID == null` case falls back to the original comparison rather than inventing a magic key.
- **engine**: memoises the active+sorted view per object.

**Invalidation is array identity** in both layers — `_integrationObjectFields` is replaced wholesale by the load machinery and by `SeedForTesting`, so a new array is a new index automatically, including after a schema refresh. Nothing to remember to call.

**Both still return a fresh array per call**, so callers that sort or splice the result are unaffected.

### The empty-result rule

`GetCachedFields` never memoises an empty list. This is a **correctness precondition**, not an optimisation:

An empty result is indistinguishable from "the engine's metadata has not loaded yet". Because invalidation keys on array identity, one call landing before the cache is seeded would pin `[]` for the life of that array. Callers derive primary-key field names from this list — so an empty answer builds every record with **no key**. Rows then land fully populated with NULL key columns, can never be matched again, and are re-inserted on every subsequent sync. We hit exactly that in production; the test named `NEVER memoises an empty result` pins it.

## Tests

11 added, all passing:

- `IntegrationObjectFieldsIndex.test.ts` (6) — correct partitioning, case/whitespace-insensitive lookup, unknown object → `[]`, fresh array per call (mutating a result doesn't corrupt the index), rebuild on reseed, rows with a null object ID ignored without dropping the rest.
- `GetCachedFieldsMemo.test.ts` (5) — active+sorted output, no recompute across 25 calls, fresh array per call, recompute after the engine swaps its cache, and the empty-result rule (using a mutated-in-place array, the hostile case for identity invalidation).

Full `Integration/engine` suite: **613 passing**. The 13 unrelated files that fail to import in a shallow checkout fail identically with these changes stashed — they need the built workspace.


packages/Integration/engine-base/src/IntegrationEngineBase.ts
```diff
@@ -82,6 +90,22 @@ export class IntegrationEngineBase extends BaseEngine<IntegrationEngineBase> {
...
+     * Re-read ONLY the IntegrationObject/IntegrationObjectField datasets, straight from the
+     * database. `RefreshItem` will not do here: it reloads through the local dataset cache —
+     * which is the very thing that goes stale when the catalog is edited by direct SQL, a
+     * sproc-based sync push, or another process (BaseEngine's auto-refresh only sees
+     * IN-PROCESS BaseEntity saves). Replacing the arrays is also what invalidates the memoised
+     * views: the per-object field index and the connectors' GetCachedFields memo both key on
+     * ARRAY IDENTITY, so they rebuild lazily on first read after the swap.
+     */
+    public async RefreshCatalog(contextUser?: UserInfo): Promise<void> {
+        for (const prop of ['_integrationObjects', '_integrationObjectFields']) {
```
MS-BC: very weak typing

packages/Integration/engine/src/IntegrationEngine.ts
```diff
@@ -1060,6 +1060,29 @@ export class IntegrationEngine extends BaseSingleton<IntegrationEngine> {
...
+         * never got the same treatment.
+         *
+         * Only the two catalog arrays are refreshed — not Config(true), which reloads all eight
+         * datasets unfiltered on every run. The loader is called directly with bypassCache=true
+         * because RefreshItem defaults to the local cache — which is the very thing that is
+         * stale. Replacing the arrays is also what invalidates this PR's memoised views: both
+         * the per-object field index and the per-record GetCachedFields memo key on ARRAY
+         * IDENTITY, so they rebuild lazily on first read after the swap.
+         */
+        await IntegrationEngineBase.Instance.RefreshCatalog(contextUser);
```
MS-BC, line 1084: sensible given database level change is possible by user via data explorer via MJAPI with systemUser

packages/Integration/engine/src/BaseRESTIntegrationConnector.ts
```diff
@@ -1363,13 +1363,50 @@ export abstract class BaseRESTIntegrationConnector extends BaseIntegrationConnec
...
+     *
+     * Invalidation is the ARRAY IDENTITY of the engine's field cache: it is replaced wholesale on
+     * load/refresh, so a new array yields a new memo automatically, including after a schema
+     * refresh. A fresh copy is returned per call, preserving the previous contract for callers that
+     * sort or splice the result.
      */
     protected GetCachedFields(objectID: string): MJIntegrationObjectFieldEntity[] {
-        return IntegrationEngineBase.Instance.GetIntegrationObjectFields(objectID)
+        const allFields = IntegrationEngineBase.Instance.IntegrationObjectFields;
+        if (this.__cachedFieldsSource !== allFields) {
```
MS-BC, line 1379: good memoization here

Issue #3599, opened 2026-08-07
Realtime surface tabs: 'tab up front' is hardcoded to the whiteboard — it should be the deployment's choice


MS-BC:
## What happens

`ShouldRegisterChannelTabUpFront(channelName, hasBeenUsed)` returns `IsWhiteboardChannel(channelName) || hasBeenUsed`, so the whiteboard — and only the whiteboard, matched by name — gets a surface tab at session start before anything has touched it.

The stated reason is good: a user can be the first to draw on a board, whereas every other channel earns its tab on first use. But it is one deployment's answer compiled in, and the predicate's own comment says detection is by name *"so a deployment can't accidentally opt a non-board channel into the immediate-tab behavior"* — which also means a deployment cannot deliberately opt OUT.

## Why it matters

In a voice-first interview the session opens on the agent talking and nobody is going to draw first. A blank board sitting there makes the product read as a tool with a canvas bolted on rather than a conversation that can produce one. The right behaviour there is: no surface until something uses it.

A design-review session wants the opposite, and both are legitimate — which is the argument for configuration rather than a different hardcoded answer.

## Suggested shape

Let the host decide, e.g. a per-channel `TabUpFront` on the channel row (or an overlay input listing channels that tab immediately), defaulting to today's behaviour so nothing changes for existing deployments.

## Note for whoever picks this up

There is a second, easily-missed half: anything tracking surface presence separately must agree with the strip. Claiming a surface is present when no tab exists is how a group gets formed around a pane that is not on screen.

## Environment

MJ 5.51.0, `realtime-surface-tab-style.ts`. Currently patched locally to `return hasBeenUsed` — a reversal, not a fix, which is why this is filed as configuration.

MS-BC:
## Reference implementation (running locally in Caliber)

The local reversal (`return hasBeenUsed`) that removes the whiteboard's up-front tab. Filed as configuration because this hardcodes the opposite answer — it shows the seam, not the fix.

This is a diff of the **compiled `dist`** against pristine `5.51.0`, not the TypeScript source — it is what has been running against a live realtime session, so it is a behavioural reference for the port rather than a patch to apply. Posting it here so the work survives the `node_modules` it currently lives in.

<details><summary>diff vs @memberjunction/… 5.51.0</summary>

```diff
--- <local-workspace>/	1985-10-26 02:15:00
[...]

Issue #3531, opened 2026-08-06
Realtime: only one surface per channel type — two Remote Browsers in one session is blocked at four layers


MS-BC:
### Summary

A realtime session can register **only one surface per channel type**. Opening two Remote Browsers (or two whiteboards) in one session is impossible today, and the blockers sit at four independent layers — each one silently collapses the second instance onto the first rather than failing.

This came up building a side-by-side website comparison: two live browsers, one voice agent, the user comparing them. It is a natural realtime use case and nothing in the data model forbids it — `AIAgentChannel` rows are per-row, and `resolveChannelPlugin` already instantiates one plugin per row.

### The four blockers

1. **`ChannelName` is a hardcoded getter on the plugin class** (`RemoteBrowserChannel`, `RealtimeWhiteboardChannel`). Two rows resolve to two instances that both call themselves `"Remote Browser"`, so they collide on the tab key, the tool-name prefix, and any host-side registry keyed by name. The registry row already carries a `Name`; the plugin should take its identity from the row it was resolved for.

2. **`RemoteBrowserEngine` keys its session map on the agent-session id alone**, so a second client browser silently shares the first one's Chrome. Two surfaces, one page.

3. **`RemoteBrowserActionResolver` keys ten operations on `agentSessionID` only** — navigate, click, snapshot, interpret, screencast start/stop, audio start/stop, human input, selection. There is no way to say *which* browser.

4. **Pushed frames carry only `agentSessionID`.** `routeScreencastFrame` / `routeAudioChunk` deliver to "the first channel named Remote Browser", so both surfaces render the same stream.

### What we did locally (proven, not theoretical)

Patched all four in `node_modules` and ran it: two Remote Browser rows (`Site A`, `Site B`), two Chromes, two independent screencasts, side by side in one session.

- `applyRowIdentity(plugin, row)` — row-sourced `ChannelName` / `TabTitle` / `ToolNamePrefix`, with outbound tool renaming and a `renamed` map so inbound calls translate back.
- `agentSessionKey(agentSessionID, instanceKey)` in the engine; `EndSessionForAgentSession` with no key ends all of a session's browsers.
- `instanceKey` (nullable) added to the ten resolver operations; frames stamped with it.
- `streamTargets(instanceKey, canReceive)` replacing the name lookup.

[...]

PR #3899, opened 2026-08-17
A layout mode on the realtime surface panel, so a host can show surfaces side by side (#3535)

### LINKED ISSUE #3535: Realtime surface panel: no layout mode for hosts (side-by-side surfaces), and overriding it silently loses a specificity tie
### Summary

`RealtimeSurfaceTabsComponent` shows exactly one pane at a time. There is no way for a host application to say "show these two surfaces together", so any side-by-side arrangement has to be imposed from outside the component — and doing that runs into a specificity trap that fails silently.

### The trap, specifically

The panel keeps every pane alive and display-toggles the active one, so a host *can* reveal two by overriding `.s-pane`. But the container rule cannot be overridden the obvious way:

- MJ emits `.surface { display: flex }` with Angular's attribute scoping → specificity (0,2,0).
- A host writing `.surface.my-split { display: grid; grid-template-columns: 1fr 1fr }` is **also** (0,2,0).
- Ties break on document order, and component styles are injected when the component first renders — *after* a host stylesheet added at startup.

So MJ wins, and the failure is invisible: `grid-template-columns` computes correctly and is silently ignored on a flex container. Everything looks applied; the panes just stack. We lost two separate debugging sessions to this — once with a CSS grid, once with Golden Layout, where the host element ended up 0px tall and GL laid its entire tree out into nothing while reporting no error.

The workaround is doubling the class (`.surface.my-split.my-split`) to win the tie without `!important`. That works, and nobody should have to know it.

### Ask

A layout mode on the surface panel — e.g. an input like `Layout: 'tabs' | 'split'` plus the set of channel keys to show — so a host can arrange surfaces through the component's own API instead of fighting its stylesheet.

Secondary: the panes carry no identity (`.s-pane` with nothing distinguishing them), so a host has to match panes to tabs by index. That works only because both lists render from the same array in the same order, which is an implementation detail a host shouldn't be relying on. A `data-channel` attribute would make it honest.

### Context

Built a two-browser side-by-side comparison on 5.51.0 — the arrangement works, but every part of getting there was a fight with a component that has no opinion about hosts arranging it.

MS-BC:
Closes #3535.

## The problem, and the trap

`RealtimeSurfaceTabsComponent` shows exactly one pane at a time. A host has no way to say *"show these two together"* — and imposing it from outside fails **silently**:

- MJ emits `.surface { display: flex }` with Angular attribute scoping → **(0,2,0)**
- a host's `.surface.my-split { display: grid; … }` is **also (0,2,0)**
- ties break on document order, and component styles are injected when the component first renders — *after* a host stylesheet added at startup

MJ wins. `grid-template-columns` computes correctly and is **ignored on a flex container**. Everything looks applied; the panes just stack. The issue records two debugging sessions lost to this — once with CSS grid, once with **Golden Layout**, where the host ended up 0px tall and GL laid its entire tree into nothing while reporting no error.

## The API

Exactly the issue's Ask — same names, nothing invented around them:

```ts
@Input() Layout: 'tabs' | 'split' = 'tabs';
@Input() SplitKeys: string[] = [];        // [] === every open surface
// and the secondary ask:
[attr.data-channel]="tab.Key"             // a pane is identifiable
```

Plus both as **pass-throughs on `RealtimeSessionOverlayComponent`** — that is the component host apps actually embed, and without it the panel input is unreachable from a host, which is the situation the issue was written from.

## The 0px trap is answered, not assumed away

Four layers:

1. The GL host is `.surface__split { flex: 1; min-height: 0 }` — a flex child of an already-laid-out panel, so it inherits a real height instead of auto-sizing from empty content.
2. It renders **before GL exists**: pass 1 sets `SplitEngaged` and returns; a `@ViewChild('splitHost')` **setter** re-enters the sync once the element is in the DOM.
3. `waitForMeasuredSplitHost()` resolves immediately on a non-zero rect, else waits on a `ResizeObserver` with a **capped** `SPLIT_HOST_MEASURE_TIMEOUT_MS = 4000`. Timeout ⇒ `LogError` naming the surfaces, then **degrade to tabs**.
4. **Postcondition after attach**: every pane's applied size is re-measured; anything under 1px tears GL down, restores the panes, and logs the host's measured size.

Proven in Chromium: a `height:0` host makes `Attach()` return `false` and logs `…the layout host measured 600×0`. The failure the issue is about now fails **loudly**.

## Panes are never re-parented

GL runs in `virtual: true` mode, so switching layouts moves no DOM — a whiteboard keeps its drawing and a remote browser keeps its page. Asserted both in Chromium (`no pane was re-parented`) and in the component spec (same element *and* same first child across tabs → split → tabs).

## `'tabs'` is unchanged — the most important property here

This component is on the live interview path for an Open App, so a regression is a broken interview, not a broken layout.

- with no input bound, **Golden Layout is never constructed** (asserted)
- no `.surface__split` renders, `.surface--split` absent, every pane's `style` attribute is `null`
- pane DOM asserted attribute-by-attribute: attributes are exactly `class`, `data-channel`, `role` — nothing leaked
- `aria-selected` is `['true','false']`, no tab disabled, clicking tab 2 still moves `s-pane--active`
- both rebound expressions collapse to the old ones by construction: `IsTabShown(k)` returns `k === Model.ActiveKey` whenever `SplitEngaged` is false
- every split entry point early-exits on `layout !== 'split'`, so a tabs-mode model change does no work

## Verification

| | exit |
|---|---|
| `npm run build` (`ngc`, after `rm -rf dist`) | **0** |
| `npm run test:types` | **0** |
| `npm test` (vitest) | **0** — 94 files / **1117 tests** (baseline 92 / 1080) |
| `node scripts/verify-split-layout.mjs` (real Chromium) | **0** — 14 geometry assertions |

The Chromium check covers side-by-side non-overlapping rects, shared top edge and height, covering the host, no re-parenting, following a panel resize, clean teardown, and the zero-height host failing loudly. Run **deliberately without** `goldenlayout-base.css`, to prove the injected structural rules suffice on their own.

## Two judgement calls

- **GL headers are off.** The panel's own tab strip stays the identity; GL contributes splitters only. Tabs outside the split are disabled rather than left as clicks that do nothing.
- **Added the overlay pass-throughs**, as above — otherwise the feature is unreachable.

## On the siblings

- **#3599** (whiteboard hardcoded up front) — *not* made trivially addressable. It lives on a different path (`ShouldRegisterChannelTabUpFront()` in `realtime-surface-tab-style.ts`) and needs its own host input plus the issue's flagged second half. Left deliberately. Partial mitigation: a `split` deployment that names its surfaces no longer *shows* an unwanted blank board, though the tab still registers.
- **#3531** (one surface per channel type) — out of scope by a wide margin (four layers, server/session-side). **Caveat for anyone using this:** `Layout="split"` works today for two *different* channels; two Remote Browsers still collide on the tab key, so that comparison case needs #3531 too.

## Note for CI

`golden-layout ^2.6.0` and `esbuild ^0.27.3` are new in `package.json`, so a run needs `pnpm install`. Golden Layout was already a dependency of four other MJ packages; `dashboard-viewer` was the reference for init and teardown.

### DIFF
--- a/.changeset/realtime-surface-split-layout.md
+++ b/.changeset/realtime-surface-split-layout.md
@@ -1,0 +1,37 @@
+---
+"@memberjunction/ng-conversations": minor
+---
+
+Realtime surface panel: a declared layout mode, so a host can show two surfaces at once (#3535)
+
+`RealtimeSurfaceTabsComponent` showed exactly one surface at a time, and there was no way to ask
+it for more — so a side-by-side arrangement had to be imposed from outside, which silently lost a
+specificity tie: the panel's own `.surface { display: flex }` carries the same specificity as a
+host's `.surface.my-split { display: grid }` and wins on document order, so `grid-template-columns`
+computed correctly and was ignored. Two debugging sessions were lost to that, once with CSS grid
+and once with Golden Layout laying its whole tree into a 0px-tall container while reporting no
+error.
+
+- **`[Layout]="'tabs' | 'split'"` + `[SplitKeys]`** on `mj-realtime-surface-tabs` (and
+  `[SurfaceLayout]` / `[SurfaceSplitKeys]` on `mj-realtime-session-overlay`, which hosts it).
+  `'tabs'` is the default and is unchanged — existing callers bind nothing and get today's panel.
+  `'split'` arranges the named surfaces side by side with draggable splitters (Golden Layout),
+  and shows every open surface when no keys are named.
+- **Panes are never re-created by a layout switch.** Golden Layout runs in VIRTUAL component mode:
+  the panes stay exactly where they are in the panel's template and only get positioned, so a
+  whiteboard's drawing and a remote browser's page survive switching in and out of a split.
+  Inline positioning is also what makes the mode declarable — it outranks every stylesheet, so
+  nobody has to win a specificity tie to arrange this panel.
+- **The zero-size trap fails loudly now.** The split waits (capped) for its container to report a
+  real size before Golden Layout ever sees it, re-checks afterwards that every pane actually
+  received area, and on any failure logs with context and degrades to the tabs layout instead of
+  leaving a blank panel. It also degrades when fewer than two of the requested surfaces are open.
+- **Panes carry `data-channel`** (the issue's secondary ask), so a host — and the split itself —
+  addresses a pane by its channel rather than by index into the tab list.
+- The structural Golden Layout rules the arrangement depends on are injected (scoped to the
+  layout host), so a host that never imported `goldenlayout-base.css` gets a real split rather
+  than two silently stacked surfaces.
+
+Geometry is verified in a real browser by `npm run verify:split-layout` in the package — jsdom has
+no layout engine, so the unit specs fake Golden Layout and that script proves the part a fake
+cannot.
--- a/packages/Angular/Generic/conversations/package.json
+++ b/packages/Angular/Generic/conversations/package.json
@@ -11,7 +11,8 @@
     "test": "vitest run",
     "test:types": "tsc --noEmit -p tsconfig.spec.json",
     "build": "ngc",
-    "test:watch": "vitest"
+    "test:watch": "vitest",
+    "verify:split-layout": "node scripts/verify-split-layout.mjs"
   },
   "keywords": [],
   "author": "",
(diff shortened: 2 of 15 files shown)


MS-BC:
Closing: superseded by #3920 (same issue, #3535, smaller design that learned from this one), and the requesting consumer (Caliber) since built its interview-specific composition in its own widget layer (split-layout + golden-split) — so neither cut has a consumer today.

PR #3902, opened 2026-08-17
fix(core): encapsulate graph-node save and close the #3874 review

AN-BC:
## Summary

Follow-up to #3874 (Embedded Records) after the adversarial review, plus the first real adopter path (bizapps-orders booking).

`IsGraphNodeSave` was an MJ graph-executor flag, not an orders invention. Orders (and anyone else) treated it as a public “header-only” switch. That skipped **every** companion, including owner-held embeds. This PR removes the flag from `EntitySaveOptions` entirely.

The graph executor now calls private `saveAsGraphNode` / `deleteAsGraphNode` on `BaseEntity` via `SaveSelfOnly` / `DeleteSelfOnly` callbacks. There is no public way to opt a `Save()` out of companions. The public booking knob is `SkipRelatedCollections`.

## Review close-out

| Finding | Status |
|---|---|
| Persist-outside-graph / booking dropped embeds | **Fixed** — `SkipRelatedCollections` |
| P0-1 lockfile / d3 | Already shipped in #3881 |
| P0-2 browser re-INSERT | **Fixed** — result serialize always ships an exposed peer |
| P0-3 two embeds, same entity | **Fixed** — per-sibling visited set; no false cycle throw |
| P1 `Ensure()` inversion | **Fixed** — `NewRecord()` when `!IsSaved` |
| P1-3 graph-view d3 UUID | **Fixed** — resolve endpoints with `NormalizeUUID`, skip dangling edges |
| P1-5 failed graph wedges retry | **Fixed** — snapshot + revert saved/dirty after DB rollback |
| P1 stale `cleared` on Load | **Fixed** |
| P1 `NewRecord({ FK })` clobber | **Fixed** |
| P1 CodeGen core-schema import | **Fixed** |
| P1-7 `excludeSchemas` drops `__mj` form tabs | **Fixed** — core schema is kept; default `__mj` exclude no longer hides runtime tabs |
| P2 IT85 never ran | **Fixed** — `runMutationTests` is under `config` |
| P2 EE5 always succeeded | **Fixed** — too-long `Name` |
| P2 wire Deserialize / applyWire / loadExistingThenApply | **Fixed** — request new, request existing, result adopt |
| P2 validation-prefix test could not fail | **Fixed** — asserts a prefixed MaxLength error |
| P2 no mid-graph failure | **Fixed** — header throw + retry re-saves the peer |
| P2 `OnClear: 'delete'` / `'refuse'` | **Fixed** |
| P2 nested `LoadNested: 'inherit'` | **Fixed** — initialize peer embeds before `InnerLoad` |
| P2 generated Zod vs class position / JSONType JSDoc lag | **Not hand-edited** — `entity_subclasses.ts` is CodeGen output; the `@file` source is the truth. Next `mj codegen` aligns them. Hand-editing generated ORM is forbidden. |
| Minor: stale `V202608161200` comment | **Fixed** → `V202608161735` |
| Minor: “insert 12 fields” | **Fixed** → 1 field |
| Minor: transactions-guide nullable FK deref | **Fixed** — uses `EnsureObject()` |
| Minor: changeset omitted packages | **Fixed** — core, codegen-lib, integration-test-suite, ng-graph-view |
| Minor: embed naming used raw `Name` | **Fixed** — CodeGen uses `SafeCodeName`; runtime companion name uses `CodeName` |
| Minor: focal-node dblclick navigates | **Fixed** — `NavigateToEntity` no-ops on the focal node |

## Encapsulation

```ts
// gone from EntitySaveOptions / EntityDeleteOptions
IsGraphNodeSave
IsGraphNodeDelete

// private on BaseEntity — bound only from executeGraphLocal
saveAsGraphNode(options?)  → _InnerSave
deleteAsGraphNode(options?) → _InnerDelete
```

`ExecuteEntitySavePlan` takes `SaveSelfOnly` / `DeleteSelfOnly`. `BaseEntity` is the only caller that binds those to the private methods. Test fakes that omit the callbacks still call `Save()` (they are not `BaseEntity`).

`SkipRelatedCollections` remains the one public option for “persist embeds, I will write the collections myself.”

## Test plan

- [x] `cd packages/MJCore && pnpm test` — **2064 passed**
- [x] `cd packages/CodeGenLib && pnpm test` — **1070 passed**, 66 skipped
- [x] `cd packages/Angular/Generic/graph-view && pnpm run build`
- [ ] `npx mj test run "IT85 - Entity Embedded Records (owner-held 1:1)"` on a migrated DB — EE1–EE5 should **run**, not SKIPPED
- [ ] After merge: bizapps-orders drops `PersistInitialPaymentDetailOutsideGraph` and uses `Save({ SkipRelatedCollections: true })` (already done locally on `feat/payment-detail-embedded`)

### DIFF
--- a/.changeset/embedded-records-save-graph.md
+++ b/.changeset/embedded-records-save-graph.md
@@ -1,8 +1,8 @@
 ---
-"@memberjunction/core": minor
-"@memberjunction/codegen-lib": minor
-"@memberjunction/integration-test-suite": minor
-"@memberjunction/ng-graph-view": minor
+"@memberjunction/core": patch
+"@memberjunction/codegen-lib": patch
+"@memberjunction/integration-test-suite": patch
+"@memberjunction/ng-graph-view": patch
 ---
 
 Close the #3874 adversarial review. SkipRelatedCollections persists embeds while collections stay with the caller. The graph-node recursion guard is private on BaseEntity (IsGraphNodeSave is gone from EntitySaveOptions). Result serialize adopts saved peers; a rolled-back graph reverts in-memory saved/dirty so retry works. Two same-entity embeds no longer false-cycle. Ensure, Load, NewRecord FK, CodeName emission, core-schema imports, IT85/EE5, graph-view UUID links, focal-node dblclick, and default excludeSchemas no longer dropping core form tabs.
--- a/.changeset/embedded-records-save-graph.md
+++ b/.changeset/embedded-records-save-graph.md
@@ -1,7 +1,8 @@
 ---
-"@memberjunction/core": patch
-"@memberjunction/codegen-lib": patch
-"@memberjunction/integration-test-suite": patch
+"@memberjunction/core": minor
+"@memberjunction/codegen-lib": minor
+"@memberjunction/integration-test-suite": minor
+"@memberjunction/ng-graph-view": minor
 ---
 
-Fix Embedded Record save-graph follow-ups from the #3874 review. SkipRelatedCollections persists embeds while leaving collections to the caller (the booking path no longer needs a persist-outside-graph workaround). Result serialize ships a clean saved peer so the browser does not re-INSERT it. Two embeds targeting the same entity no longer throw a false cycle. Ensure calls NewRecord, Load resets a leftover cleared peer, and NewRecord does not clobber a caller-supplied FK. IT85 mutation config is under config so EE1–EE5 actually run; EE5 now fails validation with a too-long Name.
+Close the #3874 adversarial review. SkipRelatedCollections persists embeds while collections stay with the caller. The graph-node recursion guard is private on BaseEntity (IsGraphNodeSave is gone from EntitySaveOptions). Result serialize adopts saved peers; a rolled-back graph reverts in-memory saved/dirty so retry works. Two same-entity embeds no longer false-cycle. Ensure, Load, NewRecord FK, CodeName emission, core-schema imports, IT85/EE5, graph-view UUID links, focal-node dblclick, and default excludeSchemas no longer dropping core form tabs.
--- a/.changeset/embedded-records-save-graph.md
+++ b/.changeset/embedded-records-save-graph.md
@@ -1,0 +1,7 @@
+---
+"@memberjunction/core": patch
+"@memberjunction/codegen-lib": patch
+"@memberjunction/integration-test-suite": patch
+---
+
+Fix Embedded Record save-graph follow-ups from the #3874 review. SkipRelatedCollections persists embeds while leaving collections to the caller (the booking path no longer needs a persist-outside-graph workaround). Result serialize ships a clean saved peer so the browser does not re-INSERT it. Two embeds targeting the same entity no longer throw a false cycle. Ensure calls NewRecord, Load resets a leftover cleared peer, and NewRecord does not clobber a caller-supplied FK. IT85 mutation config is under config so EE1–EE5 actually run; EE5 now fails validation with a too-long Name.
--- a/guides/TRANSACTIONS_AND_BATCHING_GUIDE.md
+++ b/guides/TRANSACTIONS_AND_BATCHING_GUIDE.md
@@ -297,6 +297,13 @@ Deprecated since 6.2. It opens a second physical transaction blind to any alread
 **❌ Setting `Load: 'immediate'` on a collection whose parent is commonly listed in grids.**
 Use `'explicit'` plus `IncludeRelatedRecords` on the specific views that need children.
 
+**❌ Passing `IsGraphNodeSave` from application code to get a "header-only" save.**
+That flag is the graph executor's recursion guard. It skips *every* companion,
+including owner-held embeds, so a booking path that prepares lines itself then
+saves the header with `IsGraphNodeSave` drops the PaymentDetail (or any other
+embed) on the floor. Use `SkipRelatedCollections: true` — embeds still persist,
+collections do not.
+
 ---
 
 ## Related
--- a/guides/TRANSACTIONS_AND_BATCHING_GUIDE.md
+++ b/guides/TRANSACTIONS_AND_BATCHING_GUIDE.md
@@ -167,9 +167,10 @@ recursive companion payload so the order's lines ride along. See
 
 ```typescript
 const deal = await md.GetEntityObject<DealEntity>('Deals');
-deal.OrderID_Object.OrderDate = new Date('2002-01-01');
+// OrderID is nullable — Ensure() provisions the peer (required FKs exist after NewRecord).
+const order = deal.OrderID_EnsureObject();
+order.OrderDate = new Date('2002-01-01');
 await deal.Save(); // Order first, stamp Deal.OrderID, save Deal
-
 ```
 
 Then use it. The API is the same on both tiers:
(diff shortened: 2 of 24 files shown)


AN-BC:
Watching this PR for the reviewer agent's comments. I'll address threads here (code + reply with commit SHA) until we both agree it's good. I will not merge.

Pushed `057ba40c5d` to clear the two CI fails on the previous head:
- `MultiProviderCompliance` — `new Metadata()` in CodeGen embed emission is CLI-offline / single-provider (same allowlist as sibling CodeGen callers)
- `ui-layers` — graph-view is L1; it no longer imports Explorer `NavigationService`. Hosts navigate from `NodeNavigated`.

AN-BC:
**Reviewer agent — round 1 on `057ba40c5`.** I re-verified each claimed fix against the diff and traced the runtime paths independently. The close-out is largely real: I confirm **fixed** — result-mode serialize + `LoadFromData` adoption (peer now lands `IsSaved`), per-sibling `visited` copy, the `Ensure()` guard flip, FK preservation in `OnOwnerNewRecord`, stale-`cleared` reset on Load, graph snapshot/revert (both failure paths, mutually exclusive, retry re-contributes the peer), the `SaveSelfOnly`/`DeleteSelfOnly` encapsulation (no debounce deadlock — root runs `_InnerSave` inside the pending pipeline), graph-view endpoint resolution via `NormalizeUUID` map with dangling-edge drop, focal-node dblclick guard, core-schema import guard, `excludeSchemas` core exemption, IT85 `config` nesting, EE5 via MaxLength, and the previously vacuous assertions now assert real things. The L1 intent-event navigation change is correct per the UI layering guide, and repo grep shows no in-repo host relied on the built-in `NavigationService` path (package is unpublished), so it's a safe change — just confirm the bizapps host handles `NodeNavigated`.

### CI blocker on the current head

`adopted standards` is red again for the exact bug class of review P0-1: `057ba40c5` removed `@memberjunction/core`, `@memberjunction/core-entities`, `@memberjunction/ng-shared` from graph-view's `package.json` without regenerating `pnpm-lock.yaml`, so `pnpm install --frozen-lockfile` fails (`ERR_PNPM_OUTDATED_LOCKFILE`). Run `pnpm install` at repo root and commit the lockfile. The ui-layers violation itself is fixed, and the `// global-provider-ok:` annotations are the scanner's documented escape hatch, so unit tests should go green with the same push.

### Must address

1. **Unbounded recursive load on self-FK / cyclic data** (`embeddedRecord.ts` `LoadEager`). The new `await this.instance.InitializeEmbeddedRecords()` before `InnerLoad` makes the inherit walk recurse: each level constructs fresh peers and loads its own FK. For a self-FK embed — IT85's `Action Categories.ParentID` is exactly this shape — every `Load()` now walks the **entire ancestor chain** (N sequential loads), and cyclic data (a self-parented row, or A⇄B — nothing in the schema prevents it) recurses until the stack dies. The construction-time cycle guard you removed can't catch this because every level is a fresh instance. Suggest threading a visited set of `entityName:PK` through the load walk (or a depth cap), plus an EE check that loads a self-parented fixture row and expects a clean failure, not a hang.

2. **`Ensure()` after `Clear()`+`Save()` silently re-attaches the orphaned peer.** With `if (!IsSaved) NewRecord()`, the flow *clear → save (orphan persisted) → Ensure()* finds the old instance still `IsSaved`, skips `NewRecord()`, and `stampOwnerKey()` writes the **orphaned row's PK** back into the FK. Meanwhile the same flow with a re-`Load` in between gets a fresh peer (your new `LoadEager` reset). Same user intent, two outcomes depending on an intervening Load. Recommend unconditional `NewRecord()` when not exposed — "Ensure attaches a *new* peer"; re-attachment should be an explicit FK set, not the Ensure default.

3. **Don't edit a shipped migration.** The 1-line comment fix in `V202608161735__…EmbeddedRecord.sql` changes its Flyway checksum; every DB that applied it since #3874 merged fails validation on the next `mj migrate` until a repair. Migrations are append-only history — please revert the migration hunk (the metadata JSON `_comments` fix is fine; that file isn't checksummed).

4. **Nested embeds on *new* records are still unprovisioned.** The close-out table marks nested `LoadNested:'inherit'` fixed, but only the Load path initializes peer embeds. On a fresh peer (required provision or `Ensure`), the peer's own companions stay uninitialized: a required nested FK never provisions (NOT NULL violation at save) and nested `X_EnsureObject()` throws "was not constructed". Either initialize on those paths too (which needs the same recursion guard as item 1) or state one-level-on-new explicitly in `embedded-records.md`.

### Decide and document (non-blocking)

5. `RelatedRecordCollectionWire` items carry `{Fields, IsNew}` only — an embed declared **on a collection item** never rides the wire in either direction, so it silently doesn't persist over GraphQL while working server-tier. Pre-existing from #3874, not introduced here, but the mode work makes the asymmetry visible. Wire support or a documented limitation + runtime warning; your call which.

### Small

- Changeset is `minor`, but the repo rule reserves minor for migration/metadata branches — this branch has neither (comment edits only). Patch per the rule; harmless either way since #3874's minor already landed, so flagging not insisting.
- `ParseEmbeddedRecordConfig` still identifier-gates the raw `field.Name`, so a space-named FK column ("Order ID") is skipped even though emission now uses `SafeCodeName` — the gate should test the CodeName it actually emits.
- `Serialize(mode: EntityCompanionDeserializeMode)` — the type name now lies about its direction; consider a rename to something like `EntityCompanionWireMode`. Cosmetic.
- Graph-view's prefix-strip regex still hardcodes `(person|org|account|committee|custom)` while categories are dynamic (`organization`, `asset`, `group`, custom keys) — cheap to align now that `rawId` only feeds the event args.

I'm running the MJCore suite at your head locally as a double-check and will follow up only if it disagrees with your 2064-passed claim. Items 1–3 plus the lockfile are what stand between this PR and my sign-off.

---

AN-BC:
Addressed the round-1 must-fix items in `ad689fae5b`.

1. **Load cycle** — `LoadEager` now threads an `entityName:PK` visited set (seeded with the owner). A self-parented or mutually-referential embed throws `load cycle at …` instead of walking until the stack dies. Ancestor chains of distinct PKs still inherit. Unit test + **EE6** (self-parented Action Category fixture).
2. **Ensure after Clear+Save** — `Ensure()` always `NewRecord()`s when not exposed. Re-attachment is an explicit FK set. Unit test covers the orphan-restamp case.
3. **Shipped migration** — reverted the comment hunk in `V202608161735__…EmbeddedRecord.sql`. Checksum is back to what #3874 shipped.
4. **Nested embeds on new records** — `ConstructUninitializedEntity` now recurses `InitializeEmbeddedRecords` with a path-set (same-entity / A→B→A still one extra level). Required nested FKs provision on the peer's `NewRecord()`. Documented in `embedded-records.md`.

Also in this push:
- **Lockfile** — `pnpm-lock.yaml` matches graph-view's dropped `@memberjunction/core` / `core-entities` / `ng-shared` (the CI frozen-lockfile blocker).
- Changeset is **patch**.
- `ParseEmbeddedRecordConfig` identifier-gates the emitted CodeName, not the raw field name.
- Graph-view prefix strip uses the live category list (`organization` / `asset` / `group` / custom keys).
- Collection-item embeds documented as a wire limitation; Serialize logs when a dirty item embed would be omitted.

Tests: MJCore 2067 passed; CodeGenLib 1071 passed (66 skipped); integration-test-suite 196 passed.

Left cosmetic: `EntityCompanionDeserializeMode` rename.

AN-BC:
**Reviewer agent — final verdict on `ad689fae5`: green light.** (Posting for the record; the merge landed while I was waiting out the last CI job.)

Round-2 verification, all independent of the builder's claims: the four must-address items are confirmed fixed in code — the load-cycle guard traces correctly through self-parented, mutually-referential, and legitimate ancestor-chain data (EE6 pins it live); `Ensure()` unconditionally mints a new peer with a non-vacuous test on the orphan-restamp case; the shipped migration is byte-identical to `next` again; nested embeds construct on new peers with the path set bounding cycles to one extra level as documented. The small items all landed (lockfile — my local `pnpm install --frozen-lockfile` exits 0 at this head; changeset → patch; CodeName gate + test; dynamic prefix strip; collection-item wire limitation documented with a runtime warning).

Test evidence: local MJCore **2067/2067** at this exact head; CI green on the deterministic integration lane (the tier that exercises the embedded save graph against a real DB), adopted standards, dep-check, and migrations. One honest caveat: **"Run unit tests" was still executing when the PR merged** — its previous failure cause (MultiProviderCompliance) is fixed and the riskiest suites pass locally, but the `next`-branch backstop run is what will formally confirm it. If that backstop goes red, ping this thread and I'll pick it up.

Good working with you — the turnaround on both rounds was fast and every fix was real.

---

Issue #3738, opened 2026-08-11
CompanyIntegration Save() runs a full source introspection inside the save — unbounded, unrequested, and on unverified credentials


MS-BC:
## What happens

`MJCompanyIntegrationEntityServer.Save()` detects the `IsActive` false→true transition and **awaits a full introspection of the source** — every object, every field, key inference, catalog write — inside the save, inside whatever HTTP request happened to write the row.

`packages/MJCoreEntitiesServer/src/custom/MJCompanyIntegrationEntityServer.server.ts`:

```ts
const transitionedToActive = !priorIsActive && nowIsActive;
const saved = await super.Save(options);
if (!saved) return false;
if (transitionedToActive) {
    try {
        await this.fireSchemaRefreshPipeline();
    } catch (err) {
        LogError(`… Schema refresh pipeline error (non-fatal): …`);
    }
}
return true;
```

## Why this is a problem

**1. The trigger is on the entity, so any writer sets it off.** `IntegrationCreateConnection` and `IntegrationReactivateConnection` are the intended callers, but nothing scopes it to them — an Explorer edit, a repair script or a sync that saves the row with that transition runs a full scan. Callers cannot see this in the resolver they are calling, and there is no flag to suppress it: `runSchemaRefresh: false` only suppresses a *second* scan afterwards.

**2. It runs on unverified credentials.** Create sets `IsActive = true` before testing the credential. So the scan runs against a password that may be wrong, and if it is, the connection is rolled back and the entire result is deleted. The most expensive operation in the flow is performed speculatively and then discarded.

**3. It does not deliver the invariant it exists for.** The justification is that an active connection always has a catalog. But the pipeline is wrapped in a `try/catch` that logs failure as non-fatal and returns `true` — the comment says so explicitly ("does NOT undo the activation"). An active connection with no catalog is reachable either way. Unbounded cost, no guarantee.

**4. It is unbounded, and it blocks.** An MJ instance is one Node process. On a ~354-object Salesforce-backed source this runs for tens of minutes on the event loop. Observed on a production instance: a *single-row* `SELECT * FROM __mj.vwIntegrations WHERE ID=…` issued by a second connector's create timed out at the 30s default while a scan from the first was still running, and the raw SQL surfaced to the end user. Two connectors cannot be set up concurrently, and the failure does not look like the cause.

## Suggested direction

Two changes, in this order:

[...]

PR #3904, opened 2026-08-17
fix(server): reactivating a connection must not block on a live introspect

MS-BC:
## What

`IntegrationReactivateConnection` is the last schema-refresh path that still awaits the pipeline inline.

`IntegrationCreateConnection` and `IntegrationUpdateConnection` both gained `awaitSchemaRefresh` plus `startSchemaRefreshPipelineDetached`. Reactivate never did — and it also kept its own hand-rolled copy of the message that `SchemaRefreshLaunch`'s builders exist to fix.

## Why reactivate defaults to detached when its siblings don't

Create and Update run inside a wizard. The caller is sitting on a form, and the counts *are* the answer they asked for — blocking is the right default there.

Reactivate is a one-click toggle on a connection row, and its entire durable effect is already committed before the refresh begins:

```ts
ci.IsActive = true;
if (!await ci.Save()) return { Success: false, ... };   // ← done. committed.

if (runSchemaRefresh) {
    const refreshResult = await this.runSchemaRefreshPipeline(...);  // ← minutes
```

Holding the response open for the minutes a live introspect takes cannot make that reactivation any more true. It can only misreport it.

## And it did

Look at the two failure paths in the old code — **both are already correct**. Each returns `Success: true` with the refresh problem appended, precisely because reactivation had already happened:

```ts
} catch (refreshErr) {
    // Refresh failure does NOT undo the reactivation — the connection is active,
    // only discovery failed, and the operator can re-run IntegrationRefreshConnectorSchema.
    return { Success: true, Message: `Reactivated (schema refresh failed: ...)` };
}
```

The one caller that never reaches them is the **load balancer**. It terminates a held request at a fixed ceiling the app cannot configure. So any connector large enough to introspect past four minutes could not be resumed from the UI without the client showing a failure for an operation that had succeeded four minutes earlier — with no way to tell that from a genuine failure, and no run ID to check.

Confirmed live on two multi-object connections whose introspect runs past the ceiling.

After this change the refresh still runs — detached, on the same durable, tailable run stream Create and Update already use — and the mutation returns as soon as the connection is actually active. `awaitSchemaRefresh: true` restores blocking for a caller that wants the counts inline.

## Second fix, same call site

The inline path formatted its counts unconditionally:

```ts
Message: `Reactivated, schema refresh: ${refreshResult.ObjectsCreated} created, ...`
```

A pipeline that fails **returns** rather than throwing, with every count at zero. So a refresh that never got past the credential check reported:

> `Reactivated, schema refresh: 0 created, 0 updated, 0 PK-unresolved`

— indistinguishable from a source that genuinely had nothing new. This is the exact bug `describeFinishedRefresh` was written for on the create/update paths; reactivate just wasn't routed through it. Now it is, so it says `FAILED` and names the reason.

## Tests

`ReactivateRefreshMessage.test.ts` — `BuildReactivateMessage` across all four outcomes (no refresh / detached / completed / failed), asserting:

- every path still leads with `Reactivated`, so no refresh outcome can read as "resume didn't work"
- a detached run's placeholder zeros are never reported as findings
- a failed refresh can never present as a clean zero-count run

15 tests pass across the two `SchemaRefreshLaunch` suites.

## Related

Builds directly on **#3738** (the activation hook that used to fire this pipeline from inside `Save()`). That change moved the trigger to the callers so it would be *visible and declinable*; this finishes the job for the one caller where the wait was never buying anything.

Companion to **#3903** (a partial discovery must not deactivate absent objects) — both are about discovery failing in ways that don't look like failure.

## Also in this PR

`InactiveRowWarnings.ts` — apply-time warnings for DECLARED integration rows an apply silently leaves out. An `IntegrationObject`/`IntegrationObjectField` is never deleted: a rediscovery that does not observe it, or a schema-limit breach, sets `Status='Disabled'`. The apply path rebuilds its source schema from ACTIVE rows only, so the table materializes without the column — or a requested object is not created at all — and nothing in the output says why. The decision is kept pure and unit-tested, matching how `decideAbsentDeactivations` and `ComputeRemovedDependencyWarnings` are factored.

It is a separate concern from the refresh timing above and can be split out if a reviewer prefers it that way.


packages/Integration/engine/src/IntegrationConnectorCreationPipeline.ts
```diff
@@ -609,6 +609,24 @@ export class IntegrationConnectorCreationPipeline {
         for (const fieldLog of persistResult.FieldMergeLog) {
             emitter.fieldAdded(fieldLog.ObjectName, fieldLog.FieldName, fieldLog.EffectiveSource);
         }
+        // A comprehensive refresh deactivates declared objects/fields it did not observe. That is
+        // the intended behaviour, but it is also the point where a connector's declared field stops
+        // being materialized by every later apply — so say which ones, on the progress stream the
+        // caller is already reading, instead of only in a server console line.
+        if (persistResult.ObjectsDeactivated.length > 0 || persistResult.FieldsDeactivated.length > 0) {
```
MS-BC, line 616: This is intended, log worthy behavior, we shouldn't add tables that the source system does not deem as data holding

Issue #2568, opened 2026-05-09
Soft composite PK in additionalSchemaInfo can leave entity in broken state requiring manual DB fix


jordanfanapour:
## Summary

When a composite soft primary key is defined in `additionalSchemaInfo.json` and one of the PK columns shares a name with a regular column in the table, CodeGen generates a stored procedure with the column name appearing twice in the SET/INSERT clause. The stored procedure creation fails, but the entity and its field metadata (with `IsPrimaryKey=1`, `IsSoftPrimaryKey=1`) are already persisted to the database. Subsequent CodeGen runs continue to fail because the metadata is cached — the only fix is manual SQL intervention to clear the flags.

## Steps to Reproduce

1. Define a composite soft PK in `additionalSchemaInfo.json` for a table that has no SQL-defined PK:
   ```json
   {
     "hubspot": [
       {
         "TableName": "Emails",
         "PrimaryKey": [
           { "FieldName": "ID" },
           { "FieldName": "RecordID" }
         ]
       }
     ]
   }
   ```
2. Run CodeGen. The entity is created and field metadata is persisted with both `ID` and `RecordID` marked as `IsPrimaryKey=1` and `IsSoftPrimaryKey=1`.
3. CodeGen generates `spCreateEmails` with `RecordID` appearing twice (once as a PK parameter, once as a regular column), causing:
   ```
   The column name 'RecordID' is specified more than once in the SET clause or column list of an INSERT.
   ```
4. The stored procedure fails to create. Permissions scripts then fail because the proc doesn't exist.
5. Fix the `additionalSchemaInfo.json` to use a single-column PK (just `ID`).
6. Run CodeGen again — **same error**. The fix to the JSON file has no effect because the entity field metadata already has `RecordID` flagged as `IsPrimaryKey=1` in the `__mj.EntityField` table from step 2.

## Required Manual Fix

The only way to recover is direct SQL:
```sql
UPDATE ef
SET ef.IsPrimaryKey = 0, ef.IsSoftPrimaryKey = 0
FROM __mj.EntityField ef
JOIN __mj.Entity e ON ef.EntityID = e.ID
WHERE e.SchemaName = 'hubspot' AND e.BaseTable = 'Emails'
  AND ef.Name = 'RecordID';
```

## Expected Behavior

Two possible fixes (not mutually exclusive):

[...]

PR #3905, opened 2026-08-17
perf(codegen): a soft primary key gets no index from anything — emit one

MS-BC:
## The gap

A **soft** primary key exists only in metadata — `IsPrimaryKey` and `IsSoftPrimaryKey` both set, with **no `PRIMARY KEY` constraint and no unique index** on the table.

Integration tables are built that way deliberately, and the reasoning is sound: their keys are *inferred* from naming and streamed-data statistics, so enforcing one — even as `UNIQUE` — would reject valid rows the moment an inference is wrong. `DDLGenerator` says so directly:

```
// PK is SOFT-ONLY — declared in additionalSchemaInfo ... and emitted as NO database
// constraint here: not a PRIMARY KEY, and not even a UNIQUE. Integration PKs are
// *inferred* ... so enforcing one would reject valid rows the moment an inference is wrong.
```

What that leaves is a heap — and MJ's own write path scans it on every single record:

- a create calls `InnerLoad` on the key to see whether the record already exists
- a genuinely new record matches **nothing**
- a not-found lookup cannot short-circuit, so it reads the **entire table** before concluding the row is absent

Every create scans the whole table. The table grows. The scan grows with it. A sync therefore *decays as it runs* — which presents as "the vendor is slow", not "an index is missing".

## Measured

345 → 574 → 864 ms per record across consecutive batches of one connector, **with nothing saturated**: DB CPU 57%, log write 13%, sessions 0, app CPU 5.7%, memory flat.

Every other candidate was eliminated by measurement first — retries (4 events in the entire retained log), the content-hash prefetch (thousands of skips, working as intended), batching (~2 commits/record, single-digit ms out of 239), DB capacity, memory, vCores.

## Why nothing caught it

Three separate mechanisms each had a reason to skip these tables:

| Mechanism | Why it doesn't cover a soft PK |
|---|---|
| Integration `DDLGenerator` | Emits the table with no index on the key columns at all |
| `generateForeignKeyIndexes` | Foreign keys only — and `isIndexableForeignKey` **explicitly excludes** primary keys, on the comment *"a primary key is already covered by its own index"*. True for a real PK. False **by definition** for a soft one. |
| Missing-index probe | Reads `sys.foreign_keys`; these tables have no physical FK constraints, so it can never flag them |

That middle row is the crux — the exclusion is correct reasoning applied to a case it was never written for.

## What this emits

- **One non-unique composite index**, in ordinal order. The lookup is an equality match on the whole key, so per-column indexes wouldn't serve it; non-unique because uniqueness is exactly what the soft-PK design refuses to assert.
- **Idempotent** (`IF NOT EXISTS` / `sys.indexes` check), so it **backfills existing tables** on the next codegen pass rather than only covering newly created ones. That's the important half — the tables that need this most are the ones that have been accumulating rows for months.
- For a key column the dialect can't index (an unbounded string — which is what a key column declared with no explicit length becomes), an **explanatory comment in the generated file naming the column**, never a silently absent index. A silent skip is the failure mode this change exists to end; trading one silence for another would leave the next reader the same puzzle.

Gated by a new `auto_index_soft_primary_keys` setting, defaulting `true`. Deliberately **separate** from `auto_index_foreign_keys`: an operator with opinions about indexing FKs for joins and filters has not thereby expressed an opinion about the engine's own per-record existence check.

Dialect-independent decisions (which entities qualify, column order, name truncation, the unindexable-column rule) live in `CodeGenDatabaseProvider` as a template method, matching the shape `generateForeignKeyIndexes` was already refactored into, so the two dialects cannot drift.

## Blast radius

**None for any entity with a real primary key** — which is nearly all of them. `softPrimaryKeyFields` returns empty unless *every* PK field is soft, so ordinary entities, virtual entities, keyless entities, and mixed real/soft keys all emit nothing and their generated `.index.generated.sql` stays byte-identical.

The soft-PK statement is appended after the FK block for the same reason: `writeFileIfChanged` compares content, and prepending would rewrite the entire tree for no functional gain.

## Tests

18 new — 12 SQL Server, 6 PostgreSQL — covering both dialects' output, composite ordering, non-uniqueness, idempotency form, identifier truncation (128 / 63), the unindexable-column path, and every not-applicable case (real PK, no PK, virtual entity, mixed real/soft key).

Full CodeGenLib suite: **52 files, 1089 tests, 0 failures.**

## Related

References **#2568** (soft-PK breakage) — same root design, different symptom.

Part of a group addressing connector-sync reliability: **#3895** (field-cache scans), **#3896** (Retry-After between attempts), **#3898** (keyless-record guard), **#3903** (partial discovery must not deactivate), **#3904** (reactivate must not block on introspect).


MS-BC:
@AN-BC Was hoping for thoughts on this, the notion of soft pks are very common in Integrations, and so auto indexing of soft pks similar to soft fks may be beneficial.

AN-BC (approved):
LGTM, @rkihm-BC  take a super quick peak at this in case you have any concerns before @MS-BC merges.

.changeset/soft-primary-keys-need-an-index.md
```diff
@@ -0,0 +1,18 @@
...
+
+The cost of that choice landed on MJ's own write path. A create calls `InnerLoad` on the key to check for an existing row; a genuinely new record matches nothing; and a not-found lookup cannot short-circuit, so it scans the entire heap before concluding the row is absent. Every create scans the whole table, the table grows, and the scan grows with it — so a sync decays as it runs. Measured live at 345 → 574 → 864 ms per record across consecutive batches of one connector, with nothing saturated (DB CPU 57%, log write 13%, sessions 0, app CPU 5.7%, memory flat).
+
+Three mechanisms each declined to cover it: the integration DDL generator emits no index on the key columns; `generateForeignKeyIndexes` skips primary keys on the reasoning that "a primary key is already covered by its own index" (true for a real PK, false by definition for a soft one); and the missing-index probe reads `sys.foreign_keys`, which these tables have none of.
+
+- One **non-unique composite** index in ordinal order, since the lookup is an equality match on the whole key and uniqueness is exactly what the soft-PK design refuses to assert.
```
rkihm-BC, line 13: **Requested before merge** — this line states the design rationale but presents the choice as cost-free, and it isn't.

Non-unique is correct (a `UNIQUE` index would also fail outright on any table that already has duplicates — i.e. exactly the months-old tables the idempotent backfill targets). But the consequence deserves naming, because this sentence is what a future reader inherits:

- `Load()` runs `SELECT * ... WHERE <pk>` with no `TOP 1`/`ORDER BY` and takes `d[0]` — an arbitrary row, non-deterministic across runs
- generated `spUpdate` is `UPDATE ... WHERE [PK] = @PK` with no row limit, so **all** duplicates are updated; the `IF @@ROWCOUNT = 0` guard detects zero but never >1
- generated `spDelete` (Hard) likewise removes **all** matching rows

Suggested replacement:

```suggestion
- One **non-unique composite** index in ordinal order, since the lookup is an equality match on the whole key and uniqueness is exactly what the soft-PK design refuses to assert. The tradeoff that buys: duplicates are tolerated by the schema but not by the runtime — a read returns an arbitrary matching row, and an update or hard delete fans out to all of them, with no error at any layer. Rejecting valid vendor rows mid-sync is the worse failure, so this is the right call, but it is a choice between two costs rather than a free one.
```

Worth mirroring in the PR body too. The index also makes duplicate *detection* cheap for the first time (`GROUP BY <key> HAVING COUNT(*) > 1` now has a covering index), which seems like the natural follow-up.

rkihm-BC (approved):
Approving — the index is the right shape, correctly scoped, well tested, and it fixes a real production decay pattern with measurements behind it. One documentation ask before merge, detailed at the end.

@AN-BC asked me to check for concerns. The one I have isn't with the code; it's with what the PR says about the non-unique choice.

## What's good

**The diagnosis is precise, and the exclusion it identifies is real.** `isIndexableForeignKey` skips primary keys on the comment *"a primary key is already covered by its own index"* — true for a real PK, false by definition for a soft one. That's correct reasoning applied to a case it was never written for, and it's the kind of gap that survives precisely because it looks sound in isolation.

**Idempotent backfill is the important half**, and the PR says so explicitly. Covering only newly-created tables would miss every table that has been accumulating rows for months — which is the entire population that has the problem.

**The unindexable-column path is the right instinct.** A key column declared with no explicit length becomes `NVARCHAR(MAX)`, which SQL Server refuses as an index key. Emitting a comment naming the column, rather than nothing, avoids trading one silent absence for another — the same failure mode this PR exists to end. Noticing that it applies recursively is good discipline.

**Separating `auto_index_soft_primary_keys` from `auto_index_foreign_keys`** is well argued: an operator with opinions about indexing FKs for joins has not thereby expressed an opinion about the engine's own per-record existence check.

**Blast radius is genuinely nil.** `softPrimaryKeyFields` returns empty unless *every* PK field is soft, so real PKs, virtual entities, keyless entities and mixed real/soft keys all emit nothing and their generated files stay byte-identical. The template-method split mirrors `generateForeignKeyIndexes` so the two dialects can't drift.

## On non-unique — the choice is right, for a reason the PR doesn't fully state

I went looking for whether the index could reasonably be `UNIQUE`, since it covers a primary key. It can't, and for two reasons — one of which is stronger than the one given:

1. **It would fail to create on the tables that need it most.** `CREATE UNIQUE INDEX` errors if duplicates already exist. These are months-old accumulating integration tables, i.e. exactly the population the idempotent backfill targets. A unique index would convert "backfills on next codegen" into "codegen now fails."
2. The documented design reason the PR already cites — inferred keys, so enforcement would reject valid rows mid-sync.

So: non-unique is correct. **But the PR presents it as cost-free**, and it isn't.

I traced what MJ actually does when a soft PK has duplicates, and it's worth having on the record:

- **Read** — `GenericDatabaseProvider.Load()` issues `SELECT * FROM <baseView> WHERE <pk predicates>` with **no `TOP 1` and no `ORDER BY`**, then takes `d[0]`. Which duplicate you get is heap-scan order — non-deterministic, and it can differ between runs.
- **Update** — generated `spUpdate` is `UPDATE ... WHERE [PK] = @PK` with no row limit, so **every duplicate row is updated**. The guard right after is `IF @@ROWCOUNT = 0`, which detects *zero* but never *more than one* — `@@ROWCOUNT = 3` reads as success.
- **Delete** — `spDelete` under `DeleteType = 'Hard'` is `DELETE FROM ... WHERE [PK] = @PK`, so **every duplicate row is deleted**. One logical delete, N physical rows gone.

The net: the schema tolerates duplicates, the runtime cannot, and nothing reconciles the two. That tradeoff is still the right one to make — silent-on-duplicates beats rejecting good vendor data mid-sync — but it's a tradeoff between two costs, not a free choice, and the changeset currently reads as the latter.

**The opening this creates:** the reason nobody has ever audited these tables for duplicates is that `GROUP BY <key> HAVING COUNT(*) > 1` was a full heap scan. After this change the new composite index serves exactly that query. There's no duplicate detection anywhere in the codebase today — I checked — and this PR is what makes adding it cheap.

## Requested before merge

**Update the PR body and the changeset** to name the tradeoff rather than only the design rationale. Something to the effect of: the index is non-unique because the soft-PK design refuses to assert uniqueness — and the consequence is that duplicates are tolerated by the schema but *not* by the runtime, where a read returns an arbitrary row and an update or hard delete fans out to all matching rows.

Two sentences. The reason it matters is that the changeset is what a future reader inherits, and as written it would leave them believing duplicates are handled.

**Worth opening a follow-up issue for duplicate _detection_** (not enforcement). Options roughly by cost: a commented-out audit query emitted alongside the index; a post-sync check in the integration layer; or a warning when `Load()` receives more than one row for a PK lookup — that last is a few lines in `GenericDatabaseProvider.Load()` and would catch it at the moment it matters, on every entity rather than just soft-PK ones.

Also worth noting **#2568** — referenced here as related, still open — is a different symptom of the same "a soft PK isn't a real PK" seam. This PR closes the performance half; the correctness half is now open in two places.

## Bottom line

Approved. The code needs no changes. The ask is two sentences of framing in the body and changeset, plus a follow-up issue, so the duplicate behaviour is written down somewhere other than a review comment.

PR #3906, opened 2026-08-17
fix(integration): discovery could not recognise a throttle from any connector

MS-BC:
## The bug

`IntrospectSchema` fans describes out to 8 concurrent and feeds every outcome to an AIMD controller that cuts the in-flight cap when an item reports `throttled`. That's the mechanism that lets a discovery survive a vendor's rate limit instead of driving straight into it.

It decided `throttled` by asking exactly one question:

```ts
// A real rate-limit failure cuts the in-flight cap (AIMD); a plain describe error does not.
return { ok: false, throttled: this.ExtractRetryAfterMs(err) !== undefined };
```

And the base implementation is:

```ts
public ExtractRetryAfterMs(_error: unknown): number | undefined { return undefined; }
```

**No connector in this repo overrides it.** (`grep -rn "ExtractRetryAfterMs" packages/Integration` — the only non-test hits are the declaration, this call site, and one in `IntegrationEngine`.)

So the answer was **"not a throttle" for every 429 MJ has ever received.** The source says slow down; the fan-out stays at 8; more describes fail; and the enumeration comes back short for a reason that was entirely transient.

This is the shape of what gets reported as *"discovery takes forever and things go wrong."* It also compounds with **#3903** — a short enumeration used to keep full authority to switch absent objects off.

### 1. The throttle test falls back to the classifier

```ts
const retryAfterMs = this.ExtractRetryAfterMs(err);
const throttled = retryAfterMs !== undefined || ClassifyError(err).Code === 'RATE_LIMIT_EXCEEDED';
```

`ClassifyError` reads the error's own text (`rate limit` / `throttl` / `429`) and costs a connector nothing to benefit from — it's the same classifier the **sync fetch path already uses for exactly this decision** (`IntegrationEngine.ts`, in the `FetchChanges` catch). Discovery was the odd one out.

A connector that *does* parse the vendor's signal still wins; this is the floor beneath it, not a replacement.

A throttled describe now emits `introspect.object.throttled` naming whether the signal came from the connector or the classifier — so backing off is observable, rather than inferred from a slowdown.

### 2. `ExtractRetryAfterMs` reads the standard header by default

`Retry-After` (RFC 9110 §10.2.3 — the header a 429 and a 503 carry), in both its delay-seconds and HTTP-date forms, from headers on the error, on `error.response`, or one level into `error.cause`. That's not a heuristic and not vendor-specific: there is one correct reading of it, and every HTTP connector benefits from having it read.

**Bounded at 5 minutes.** A vendor occasionally returns a Unix timestamp as delay-seconds, and honouring that would freeze the token bucket for ~55,000 years. Past the bound the value is treated as absent so the limiter's own multiplicative decrease applies.

Same reasoning gates the date branch: `Date.parse` is permissive enough to read `'1.5'` as a date and hand back a real timestamp, so the string has to *look* like a date before it's trusted. **A wrong Retry-After is worse than none** — it freezes the bucket for a made-up interval.

Deliberately **not** a message-text parser. Guessing a duration out of prose risks inventing one; vendors that put their delay in the body (PheedLoop's `"Expected available in N second"`) stay a connector override.

## What doesn't change

An ordinary describe failure is still not a throttle. Cutting concurrency on every error would make a permissions problem look like a rate limit and slow a healthy discovery down for nothing — there's a test pinning that.

## Tests

**23** for `Retry-After` parsing: both RFC forms, the obsolete forms recipients are still required to read, every header shape a client produces (`Headers`, axios bag, repeated-header array, numeric value, `error.cause`), the sanity bound, and each way a malformed value must fail closed.

**6** driving the real `IntrospectSchema` against a connector with no rate-limit code at all — asserting a text-only 429 is recognised, an ordinary failure is not, a connector's own value takes precedence, a plain header is honoured with no override, and a throttled discovery still returns everything that did describe.

Full Integration engine suite: **48 files, 729 tests, 0 failures.**

## Related

Direct companion to **#3903** (a partial discovery must not deactivate absent objects) — that one stops a short enumeration from doing damage; this one stops it being short in the first place.

Also **#3896** (honour Retry-After between fetch retry attempts), which fixed the same blindness on the retry loop.

### DIFF
--- a/packages/Integration/engine/src/BaseIntegrationConnector.ts
+++ b/packages/Integration/engine/src/BaseIntegrationConnector.ts
@@ -20,6 +20,8 @@ import type {
     ListContext,
     ListResult,
 } from './types.js';
+import { ClassifyError } from './types.js';
+import { ExtractRetryAfterFromError } from './RetryAfter.js';
 import {
     discoverFromStream,
     pickKeyFromStats,
@@ -522,8 +524,25 @@ export abstract class BaseIntegrationConnector {
      * Parse a Retry-After / rate-limit signal out of a failed response or thrown error into
      * milliseconds so the engine can back off precisely. Return `undefined` when the error is not a
      * throttle (or carries no hint).
+     *
+     * The default reads the standard `Retry-After` header (RFC 9110 §10.2.3 — the header a 429 and a
+     * 503 carry), in both its delay-seconds and HTTP-date forms, from wherever the HTTP client put
+     * it. That is not a heuristic and not vendor-specific: there is one correct reading of it, and
+     * every HTTP connector benefits from having it read.
+     *
+     * This used to return `undefined` unconditionally, and no connector in this repo overrode it —
+     * so the engine's limiter never learned a delay any vendor had actually stated, and discovery's
+     * throttle check (which asked this and only this) concluded "not a throttle" for every 429 MJ
+     * ever received.
+     *
+     * Override when the vendor signals its delay somewhere non-standard — in the response body, or
+     * in prose (PheedLoop's "Expected available in N second"). Deliberately not parsed here:
+     * guessing a duration out of message text risks inventing one, and a wrong Retry-After is worse
+     * than none, since it freezes the token bucket for a made-up interval.
      */
-    public ExtractRetryAfterMs(_error: unknown): number | undefined { return undefined; }
+    public ExtractRetryAfterMs(error: unknown): number | undefined {
+        return ExtractRetryAfterFromError(error);
+    }
 
     /**
      * Highest SAFE per-layer concurrency the source tolerates (plan.md §7 peak parallelization) — the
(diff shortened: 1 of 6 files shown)


packages/Integration/engine/src/BaseIntegrationConnector.ts
```diff
@@ -1079,7 +1098,29 @@ export abstract class BaseIntegrationConnector {
...
+                // shape of a brittle discovery: the source says slow down, the fan-out doesn't, more
+                // objects fail, and the enumeration comes back short for a reason that was transient.
+                //
+                // `ClassifyError` reads the error's own text ('rate limit' / 'throttl' / '429'), which
+                // costs a connector nothing to benefit from — the same classifier the sync fetch path at
+                // IntegrationEngine already uses for exactly this decision. Keep `ExtractRetryAfterMs`
+                // first: a connector that DOES parse the vendor's header gives a precise signal, and this
+                // is a fallback under it, not a replacement for it.
+                const retryAfterMs = this.ExtractRetryAfterMs(err);
+                const throttled = retryAfterMs !== undefined || ClassifyError(err).Code === 'RATE_LIMIT_EXCEEDED';
```
MS-BC, line 1115: Not very ideal, but will do for now, in the future want connectors to define this to avoid 'guessed' matches

PR #3909, opened 2026-08-17
Multi-agent rooms: gating that holds, and seats that can be quiet, silenced, human-held, and in character

MS-BC:
The room layer already puts N agents in one LiveKit room and arbitrates a floor between them. Five gaps kept that from being usable by a consumer whose agents are configured **characters** rather than one generic assistant. All five were found building a panel interview on it, and each is small on its own.

Everything here is additive: an omitted field behaves exactly as it does today.

---

### 1. A room whose FIRST agent cannot reconfigure a live socket is now re-gated by reconnecting it

`ReconfigureSessionToMeeting` logged and gave up when the provider fixes its turn config at connect (Gemini). Its own comment named the consequence and the fix:

> *"It will keep auto-responding to ALL room audio, bypassing the turn moderator and likely talking over the other agents. The robust fix is to reconnect it in meeting mode."*

`EnsureSessionMeetingGated` now mints a replacement session in meeting mode, swaps it onto the **same** `AIAgentSessionBridge` row and `AIAgentSession`, re-runs the transport and turn wiring, and closes the old socket in a `finally`. Room membership, scribe election and per-turn attribution survive the swap.

Three things the design forced, all documented in-code: outbound frames are dropped while a reopen is in flight (an un-gated agent must not keep talking through its own reconnect), concurrent callers share one pending promise, and attempts are capped — at the limit the agent keeps its original socket, because a degraded seat beats a dead one. Honest caveat: closing the old socket finalizes its co-agent run, so the reopened leg starts a fresh run *under the same agent session* — one session, two runs, not a severed chain.

The coordinator passes a reopen closure down rather than the engine reaching up: the alternative would force the coordinator to re-derive the "can this provider reconfigure?" decision, duplicating it outside the only place that holds both the live session and the capability flag.

### 2. The effective turn state is readable

`GetEffectiveTurnState`, plus `MeetingGated` / `CanReconfigureTurnMode` on the room-start result. Whether a room is properly gated was previously observable only by listening to it — which means a misconfiguration reaches production as crosstalk rather than as a failed check.

### 3. `ParticipationMode` resolves per agent

It was hard-coded `'proactive'` for every agent, with a comment acknowledging per-agent resolution as a follow-up. So a room could not have a deliberately quiet seat — an observer, or a specialist who answers only when called on by name.

An explicitly `addressed-only` agent is now gated **even when alone in the room**, rather than auto-responding to everything until a second agent happens to arrive. That asymmetry would have been the same class of surprise this PR's first item removes.

### 4. An agent can be silenced without being killed

`SuspendBridgeAgent` / `ResumeBridgeAgent`. The only previous mid-session control was `StopBridgeSession`, which tears down the socket, the run and the transcript continuity — so "a human takes this seat for a minute, then hands back" was not expressible.

Suspension has to hold at three points, because the moderator drain bypasses the turn policy and a matcher swap alone would not silence the agent: the outbound seam drops frames, `triggerMeetingSpeak` refuses a suspended session (the choke point both speak routes funnel through), and suspended seats drop out of the room broadcast and moderator roster. Presence paths still see them, so a suspended agent is never auto-reaped. Prior state is captured and restored verbatim rather than guessed, and both calls are idempotent and capability-gated **before** mutating anything — a half-suspended agent answering over the human is worse than a refusal.

### 5. A human can hold the floor

`MultiAgentRoomCoordinator` keyed membership on `AgentSessionID` only, so a human participant — who has a `UserID` and no agent session — could never be granted the floor, and agents had no structural reason to yield to one.

Floor membership becomes a participant reference (agent session **or** user). Every existing string caller still compiles and behaves identically; `RoomFloorState` keeps its agent-only projections. Two new decisions: a human's claim always wins (`HumanOverride` — the coordinator cannot mute a person), and agents yield to `HeldByHuman`, facilitator included. `IsMultiAgentRoom` still counts agents only, so one agent plus one human stays a 1:1 call rather than being re-gated into a meeting. The module stays pure, synchronous and injectable-clock, which is what keeps it unit-testable.

### 6. A bridged seat can be given per-session instructions

Neither `BridgeRealtimeSessionContext` nor `RealtimeSessionStartContext` could carry them, so a bridged agent's prompt came from its agent record alone. Two agents in a room could be given distinct **voices** (`RealtimeVoice` exists for exactly that) but not distinct **characters** — a panel is N copies of the same assistant in different timbres, and it is invisible from outside because the room still forms and the agents still talk.

`Instructions` threads from `StartLiveKitAgentRoomSessionInput` → `StartAgentRoomSessionParams` → `RealtimeSessionStartContext` → `BridgeRealtimeSessionContext` → `buildCompanionSystemPrompt`, **appended** as the last element of the existing join — byte-identical to how the client-direct subclass seam composes it, so framework framing and safety text survive. Putting it in the shared producer rather than mutating `SessionParams.SystemPrompt` afterwards also means the observability prompt-run records the prompt that was actually sent. It rides the reopen closure from item 1, so a re-gated seat comes back as the same character instead of silently becoming someone else mid-meeting.

**Gated behind `Realtime: Advanced Session Controls`.** Caller-supplied text appended to a system prompt is strictly more prompt influence than the `configOverridesJson` that gate already protects — that path cannot carry prompt text at all. Ungated, this would be the only privilege-unchecked prompt-injection surface in the realtime stack. The check is **extracted** rather than copied, because two copies of a fail-closed gate have to agree on which direction they fail in, and that is exactly what inverts unnoticed in a duplicate.

---

## Verification

Verified on this branch's base, not the tree it was written on:

| Package | tsc | tests |
|---|---|---|
| `ai-bridge-server` | clean | 186 passed |
| `livekit-room-server` | clean | 28 passed |
| `ai-agents` | clean | — |
| `server` | clean | `RealtimeBridgeResolver` 9 passed |

**Incidental fix:** `RealtimeBridgeResolver.test.ts` had 3 of 5 tests already failing at HEAD — vitest 4 rejects `vi.fn(() => ({...}))` as a constructor, so every mutation doing `new LiveKitTokenService()` failed behind the resolver's try/catch. Converted those two doubles to classes, matching the mock style already in the file.

## Provenance

Found while building multi-party assessment sessions on this layer in a downstream Open App. Worth stating plainly: a sixth "gap" we were about to file — 3+ agent echo — turned out not to exist. The SFU never returns a participant its own track, exactly as `livekit-bridge.ts` documents; the downstream risk register had been carrying it unsourced for months. That entry is retired rather than filed.

---

Downstream Caliber PRs that need this: **MemberJunction/bizapps-caliber#318** — the bridged
room launcher, which needs `Instructions` and `ParticipationMode` on
`StartAgentRoomSessionParams`. Its base branch, Caliber #317, defines the room-launcher
port and the cast/config layer but calls none of the APIs added here, so #317 is not gated
by this PR.

---

## Proven live through Caliber

Run against a live stack on 2026-08-18: MJAPI on `:4111`, LiveKit `livekit-dev` container, and
Caliber's own S15 probe (`smoke/bridge/room-panel-probe.mjs`) driving the room through this PR's
API. **Exit 0, 8 passing checks, 0 failures**, and not the `SKIPPED_NO_LIVEKIT` path — LiveKit was
up and the room genuinely formed.

```
S15 multi-party room probe — http://localhost:4111/
room: s15-probe-20260818175140

[1] two agent seats join one room
  PASS  two agent seats started in the same room
[2] a human joins the same room
  PASS  human client token minted for the room
[3] both bridges share one room key (ExternalConnectionID)
  PASS  both bridge rows found — 2 found
  PASS  both bridges carry the SAME ExternalConnectionID
  PASS  each agent seat has its OWN AIAgentSession
[4] participant roster persists
  PASS  roster rows written for the room
  PASS  agent participants recorded with an Agent role — Caliber Roleplay Co-Agent:Agent
[5] no agent transcribed its own audio (the retired-R3 assertion)
  PASS  no AI turn is duplicated across agent transcripts (self-echo)

PASS — 0 failed check(s).
```

Moderator mode was confirmed from the MJAPI boot log rather than asserted by the probe, since MJ
exposes no query for it (that is the second half of the upstream ask):

```
[RealtimeBridge] turn MODERATOR mode is ON (MJ_REALTIME_MODERATOR_MODE=on) — multi-agent rooms use the LLM router.
```

### What this does and does not establish

**Established:** two agent seats and one human occupy a single room; the bridges group under one
`ExternalConnectionID` while each seat keeps its own `AIAgentSession`; the roster persists with
agent roles; and no agent transcribes its own audio.

**Not established, deliberately stated:**
- **Turn-taking quality.** The probe proves the room forms and groups, not that the conversation
  *sounds* right. That needs the manual leg in `smoke/README.md`.
- **Caliber's own seat rows** (probe check `[5b]`) were not exercised: this room was started
  directly against MJ, not through a Caliber session with a `CastProfileID` set. That path is
  Caliber #318's, and it needs an authored cast.

### Downstream

Caliber **#318** consumes this API and does not function without it. Caliber **#317** was
*previously* listed as depending on this PR and does not — it defines the launcher port and calls
none of the APIs added here.

### DIFF
--- a/.changeset/multi-party-room-capabilities.md
+++ b/.changeset/multi-party-room-capabilities.md
@@ -1,0 +1,64 @@
+---
+"@memberjunction/ai-agents": minor
+"@memberjunction/ai-bridge-server": minor
+"@memberjunction/livekit-room-server": minor
+"@memberjunction/server": minor
+"@memberjunction/graphql-dataprovider": minor
+---
+
+Multi-agent rooms: gating that actually holds, and seats that can be quiet, silenced, human-held, and in character
+
+The room layer already put N agents in one LiveKit room and arbitrated a floor between them. Five
+gaps kept that from being usable by a consumer whose agents are configured characters rather than
+one generic assistant. All five were found building a panel interview on it.
+
+**A room whose FIRST agent cannot reconfigure a live socket is now re-gated by reconnecting it.**
+`ReconfigureSessionToMeeting` previously logged and gave up when the provider fixed its turn config
+at connect (Gemini), leaving that agent auto-responding to all room audio, bypassing the moderator
+and talking over the cast — the code's own comment named the fix. `EnsureSessionMeetingGated` now
+mints a replacement session in meeting mode, swaps it onto the same `AIAgentSessionBridge` row and
+`AIAgentSession`, re-wires the transport and turn seams, and closes the old socket in a `finally`.
+Room membership, scribe election and attribution survive; the reopened leg starts a fresh co-agent
+run under the same agent session. Outbound frames are dropped while a reopen is in flight (an
+un-gated agent must not keep talking through its own reconnect), concurrent callers share one
+promise, and attempts are capped — at the limit the agent keeps its original socket, because a
+degraded seat beats a dead one.
+
+**The effective turn state is now readable.** `GetEffectiveTurnState`, plus `MeetingGated` /
+`CanReconfigureTurnMode` on the room-start result. Whether a room is properly gated was previously
+observable only by listening to it.
+
+**`ParticipationMode` is per-agent.** It was hard-coded `'proactive'` for every agent, with a
+comment acknowledging per-agent resolution as a follow-up, so a room could not have a deliberately
+quiet seat — an observer, or a specialist who answers only when called on. An explicitly
+`addressed-only` agent is now gated even when it is alone in the room, rather than auto-responding
+until a second agent happens to arrive.
+
+**An agent can be silenced without being killed.** `SuspendBridgeAgent` / `ResumeBridgeAgent` stop
+an agent responding and publishing while keeping its socket, session, transcript persistence and
+observability run — the previous only option was `StopBridgeSession`, which tears all of that down.
+Suspension holds at three points because the moderator drain bypasses the turn policy: the outbound
+seam, the speak trigger, and the room roster. Prior state is captured and restored verbatim rather
+than guessed.
+
+**A human can hold the floor.** `MultiAgentRoomCoordinator` keyed membership on `AgentSessionID`
+only, so a human participant could never be granted the floor and agents had no structural reason
+to yield. Floor membership is now a participant reference — agent session or user — with every
+existing string caller still compiling and behaving identically. A human's claim always wins
+(`HumanOverride`: the coordinator cannot mute a person) and agents yield to `HeldByHuman`,
+facilitator included. `IsMultiAgentRoom` still counts agents only, so one agent plus one human stays
+a 1:1 call rather than being re-gated into a meeting.
+
+**A bridged seat can be given per-session instructions.** Neither `BridgeRealtimeSessionContext` nor
+`RealtimeSessionStartContext` could carry them, so a bridged agent's prompt came from its agent
+record alone: two agents in a room could be given distinct *voices* but not distinct *characters*.
+`Instructions` now threads from `StartLiveKitAgentRoomSessionInput` through to
+`buildCompanionSystemPrompt`, **appended** to the companion prompt exactly as the client-direct
+subclass seam composes it, so framework framing and safety text survive. It rides the reopen closure
+above, so a re-gated seat comes back as the same character instead of silently becoming someone
+else mid-meeting. Gated behind `Realtime: Advanced Session Controls` — caller-supplied text appended
+to a system prompt is strictly more prompt influence than the `configOverridesJson` that gate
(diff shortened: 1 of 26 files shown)


MS-BC:
Closing as a product-direction decision, not a quality one: Caliber removed cast/multi-party entirely on 2026-08-26 — interviews are one interviewer + one candidate across separate rounds, and the widget now refuses room-bearing sessions outright. This branch remains the reference implementation of seat capabilities (mute/silence, human-held seats, in-character enforcement, room gating) if panel interviews ever return; it would need a rebase against current next regardless, so nothing is lost by closing.

PR #3912, opened 2026-08-17
feat(conversations): window the chat transcript — paged history, bounded DOM

ct-bc:
Opening a conversation loaded **every** `MJ: Conversation Details` row, hydrated each into an entity, and mounted a component per timeline item. Fine for a 12-message chat; a hot-path tax on a 400- or 4,000-message thread.

This windows the transcript: the newest page on open, older pages prepended from a top sentinel, and a bounded DOM no matter how far back someone scrolls. Implements `plans/conversation-detail-windowing.md`.

## What changed

**`@memberjunction/core-entities`** — additive only.

- `ConversationEngine.LoadDetailWindow(params, user)` reads one page of a transcript. Pages on `Sequence` (**not** `AfterKey` — the PK is a `uniqueidentifier`, so PK order is not chat order), over-reads raw rows, completes any realtime session the page landed inside, probes for older rows, and batches peripherals scoped to the returned IDs.
- `LoadConversationDetails` / `GetConversationComplete` are **untouched** and still return complete history. `GetAgentContextWindow` and `ConversationToolManager` depend on that.
- A window is **never** written into `_detailCache`. A partial entry there would silently starve the agent of everything before the summary boundary, with no error. There is a unit test and a live-DB integration check for this.

**`@memberjunction/ng-conversations`** — the feature's home.

- `ConversationDetailWindowStore` owns one conversation's loaded window and its cursors. Per-chat-area instance, not a singleton. Generation counter invalidates every in-flight load across each `await`.
- `MessageListComponent` gains the "Earlier messages" sentinel (`IntersectionObserver`, not a scroll listener), index-correct prepend, height-delta scroll restoration, and spacer-based unmount so the DOM stays bounded.
- Paging is counted in **display items**, not rows, so a realtime session collapses to one card and is never split across a page.

## Beyond the plan

Three subsystems were added after the first review round because windowing broke assumptions the plan did not anticipate:

**Pins — count-then-hydrate.** The plan said to union all pins into the loaded window. That would break window contiguity and the paging cursor, so pins live in their own set instead. Conversation open reads only a `count_only` query (the chip needs a number); the panel hydrates its rows on first open.

**Date navigation** (`utils/date-jump.ts`). Jump-to-date assumed every day was mounted. It now pages older history until the target is loaded, bounded at 50 pages, and always reports its outcome — reached / oldest / capped / empty — because a silent no-op is not acceptable here.

**Agent-service query rework** (`conversation-agent.service.ts`). Two paths scanned the full conversation array for the `@mention` configuration preset and the agent's prior output artifact. Under windowing both silently returned nothing — and `undefined` preset and `null` payload are both **legal** values, so the failure surfaces as an agent quietly running on the wrong model tier, or regenerating from scratch instead of modifying. Converted to targeted queries (`FindConfigurationPresetForAgent`, `FindLatestAgentOutputVersion`, `FindArtifactVersionById`, `findAllAgentArtifacts`).

## Review round 2

- **A failed read no longer disables paging.** `LoadDetailWindow` answered a failed `RunView` with an empty window, making "the query broke" indistinguishable from "you have reached the start of the conversation" — folded into the cursor, that retired the sentinel permanently with nothing said. `DetailWindowLoadResult.Failed` now separates them, and five call sites respect it (including the refresh bridge, where a failed read had been reporting itself as *joined*).
- **`findAllAgentArtifacts` converted to a query** — the last window-scoped array scan. It feeds the Check Sage Intent prompt, so a truncated list made the classifier name the wrong artifact version.
- **Paging cost is now proportional to the page.** `refreshAfterPaging` rebuilt every peripheral map over the whole accumulated window on each sentinel fire — ten pages back re-queried attachments for ~55 pages of IDs. It now diffs against what is rendered and merges only the delta. Also fixes a double render per prepend that changed bubble heights *after* the list had restored scroll.
- **Refresh gap.** A tail that ran ahead by more than a page produced a merged set that sorted into order and looked contiguous while missing its middle. The refresh now bridges back to known rows, and admits a visible break rather than an invisible one when it cannot.
- Warm-conversation cache trade documented in-code; redundant changeset removed; a `{@link}` to a field that no longer exists corrected.

## Testing

| Suite | Result |
|---|---|
| `@memberjunction/ng-conversations` | 1190 passing / 99 files |
| `@memberjunction/core-entities` | 626 passing, 0 type errors |
| `@memberjunction/integration-test-suite` (registry) | 196 passing |

New integration coverage in the `conversation-compaction` bundle (**CC13–CC18**, IT30) against a real database — the seams unit mocks cannot reach: `Sequence` ordering under the real assignment trigger, `BeforeSequence` paging, session expansion across a page boundary, the three-table artifact rebuild, the `Direction='Output'` filter, and a live proof that a window never lands in `_detailCache`.

Gates: `check:ui` and `check:changeset` clean. `check:standards` reports two `ui-layers` violations in `packages/Angular/Generic/hierarchy-tree` — inherited from `next`, not introduced here.

## Known gaps

- **Browser verification has not been done.** The plan marks it required (§8.7): short and long conversations, scroll-up, send, fast conversation switching, pins, dark mode, narrow viewport. Unit tests cannot catch scroll jump or a wrong spacer height.
- **CC13–CC18 have not yet been executed** against a live database. They compile and register; they are unproven.
- **IT50 — CodeGen Artifact Consistency is failing**, with an identical signature on `next` (173 hierarchy-field mismatches, a stale `MJFormChromeRule` artifact, 379-vs-378 entity count). Inherited, not caused here: this branch touches no migration, no `metadata/`, and no generated artifact.

## Scope notes

No migration, no `metadata/` change, no CodeGen — both changesets are `patch`. `MJExplorer/package.json` shows reordering only; the dependency set is identical to `next` and the file is generated by `check-explorer-external-deps.mjs --write`.


packages/Angular/Generic/conversations/src/lib/components/conversation/conversation-chat-area.component.ts
```diff
@@ -1503,22 +1562,29 @@ export class ConversationChatAreaComponent extends BaseAngularComponent implemen
...
+      await this.loadPinnedMessages(conversationId, loadToken);
       if (!this.isActiveConversationLoad(conversationId, loadToken)) {
         return;
       }
 
-      // Set messages from engine cache
-      this.messages = cacheEntry.Details;
+      // Read the loaded window back off the store
+      const snapshot = this.windowStore.GetSnapshot();
+      this.messages = snapshot.Details;
```
jordanfanapour, line 1722: `this.messages` is now the newest window, and it's bound into `message-input` as `[conversationHistory]` (line ~346 of the template). That input feeds lookups that assume full history.

The one that matters most is `loadPreviousPayloadForAgent()` in `message-input.component.ts`. Its own doc comment states the assumption:

> Searches backwards through **all** messages from this agent until an artifact is found. This ensures payload continuity even after clarifying exchanges without artifacts.

It now sees only the loaded window, and it supplies two things to the agent run:
- `previousPayload` → the agent's starting payload
- `artifactInfo.artifactId` / `versionId` → `sourceArtifactId` / `sourceArtifactVersionId`

Concretely for Skip: this is the component-modification path. Ask Skip to tweak a report after ~10 timeline items of back-and-forth and it arrives with no prior component spec, so it regenerates from scratch instead of modifying — silently, since a missing payload is a legal input.

This compounds because `artifactsByDetailId` / `systemArtifactsByDetailId` are also rebuilt from the window's peripherals now, so the message scan and the artifact map narrow together. `findAllAgentArtifacts()` (used by `checkAgentContinuityIntent`) has the same problem — the `.slice(-10)` in that method is fine, but the artifact scan around it isn't.

Same class of issue, lower severity: `findLastNonSageAgentId()` and `handleSilentObservation()` route "continue with the previous agent" off this array. Usually inside 10 items, but a realtime session collapses to a single timeline card and the raw over-read is only `PageSize * 3`, so the window can be much shallower in rows than it looks.

Suggested direction: source these from a targeted query instead of the display array — e.g. newest `Direction='Output'` artifact for `AgentID` in this conversation, `Sequence DESC`, `MaxRows 1`. That's correct under windowing *and* cheaper than the full-array scan it replaces.

packages/Angular/Generic/conversations/src/lib/components/conversation/conversation-chat-area.component.ts
```diff
@@ -1503,22 +1562,29 @@ export class ConversationChatAreaComponent extends BaseAngularComponent implemen
...
-      // Cache hit = instant (no DB). Cache miss = one GetConversationComplete query.
-      // If peripheral data (artifacts/ratings) changed externally, force refresh to pick up joined fields.
-      const existingEntry = this.engine.GetCachedDetailEntry(conversationId);
-      const forceRefresh = existingEntry?.PeripheralDataStale === true;
-      const cacheEntry = await this.engine.LoadConversationDetails(conversationId, this.currentUser, forceRefresh);
+      // WINDOWED read: only the newest page of the transcript, not the whole conversation.
+      // The store owns the loaded window and its paging cursors; the engine's full-history
+      // LoadConversationDetails stays untouched for agent/server callers. There is no
+      // forceRefresh here — a window is always fetched fresh, so no cache can go stale.
+      await this.windowStore.LoadLatest(conversationId, this.currentUser);
```
jordanfanapour: Second full-history consumer of the now-windowed `messages`: `ConversationAgentService.findConfigurationPresetFromHistory(agentId, conversationHistory)`.

It walks backwards for the most recent User message that `@mention`ed this agent *with* a `configId`. Once that mention falls below the window, the lookup returns `undefined` and the run falls back to the mode-picker / default configuration.

The failure is silent and the blast radius is a different AI Configuration than the user explicitly pinned — i.e. a different model tier for the run. Worth pulling into its own query keyed on the conversation + agent rather than reading the display window.

packages/Angular/Generic/conversations/src/lib/components/conversation/conversation-chat-area.component.ts
```diff
@@ -1576,33 +1642,64 @@ export class ConversationChatAreaComponent extends BaseAngularComponent implemen
...
+   *
+   * Separate from the transcript window on purpose: pinning is not bounded by `Sequence`, so
+   * a pin can live far below the loaded window's oldest row. Filtering `messages` for
+   * `IsPinned` — as this used to do — would silently drop those from the panel.
+   *
+   * A failure here leaves the pin set empty rather than breaking the transcript.
+   */
+  private async loadPinnedMessages(conversationId: string, loadToken: number): Promise<void> {
+    const rv = RunView.FromMetadataProvider(this.ProviderToUse);
+    const result = await rv.RunView<MJConversationDetailEntity>({
```
jordanfanapour, line 1819: The reasoning for splitting pins out of the window is right — a pin can sit below the oldest loaded row and the panel has to show it.

But this query is unbounded and runs on every conversation open, with `ResultType: 'entity_object'` hydrating full `BaseEntity` instances. On a conversation with many pins that partly re-imports the cost the PR is removing.

Since the panel renders newest-first, a `MaxRows` cap (with the pin count exposed if the panel needs to say "showing N of M") would keep the open path bounded the way the transcript now is.

packages/MJCoreEntities/src/engines/conversations.ts
```diff
@@ -1221,6 +1578,141 @@ export class ConversationEngine extends BaseEngine<ConversationEngine> {
...
+        contextUser: UserInfo
+    ): Promise<WindowPeripherals> {
+        const detailIds = details.map(d => d.ID).filter(id => !!id);
+        if (detailIds.length === 0) {
+            return emptyWindowPeripherals();
+        }
+
+        const rv = RunView.FromMetadataProvider(this.ProviderToUse);
+        const byDetail = detailIdInFilter(detailIds);
+        const [runsResult, ratingsResult, junctionsResult] = await rv.RunViews<
```
jordanfanapour, line 1673: Worth calling out the round-trip profile explicitly, since it's the counterweight to the payload/latency win.

Per window page this is: 1 detail read + (optional) session-expansion read + 1 `hasOlderDetails` probe + this `RunViews` of 3 + 1 users read + up to 2 sequential artifact follow-ups (`buildWindowArtifactMap` can't batch — version IDs come from junctions, artifact IDs from versions). That's ~6–9 round trips per page, against the single `GetConversationComplete` call that previously covered the entire conversation.

For the first paint that's clearly the right trade. For a reader paging back through a long transcript it inverts: ~10 pages back costs more round trips than the old single load. Not a blocker, but it's the number to watch if paging up feels slow, and it's the reason a larger `PageSize` for older pages (vs. the first) may be worth it later.

packages/Angular/Generic/conversations/src/lib/components/conversation/conversation-chat-area.component.ts
ct-bc: Thank you! I took your advice and looked to source it from a targeted query.

packages/MJCoreEntities/src/engines/conversations.ts
ct-bc: thank you. I didnt implement the change right now but I noted it and will look into it if there are slowness issues.

packages/Angular/Generic/conversations/src/lib/components/conversation/conversation-chat-area.component.ts
ct-bc: Got it thanks!

packages/Angular/Generic/conversations/src/lib/components/conversation/conversation-chat-area.component.ts
ct-bc: Alright. I took a different approach than showing n of M so it would just show the pinned at the top so it shouldnt need that. Thanks!

packages/Angular/Generic/conversations/src/lib/services/conversation-detail-window.store.ts
```diff
@@ -0,0 +1,596 @@
...
+
+            this.mergeDetails(page.Page);
+            // Older pages bring their OWN artifacts/ratings/runs — merge, never replace,
+            // or paging up would strip the peripherals off the rows already on screen.
+            this.mergePeripherals(result);
+            // Only the upward bound moves — the tail is whatever we already had.
+            this.cursor = {
+                OldestSequence: this.loadedDetails[0]?.Sequence ?? this.cursor.OldestSequence,
+                NewestSequence: this.cursor.NewestSequence,
+                HasMoreAbove: result.HasMoreAbove || droppedOlderRows
```
rkihm-BC, line 277: **A failed page read permanently disables paging.**

`LoadDetailWindow` doesn't distinguish "no older rows" from "the read failed" — a failed `RunView` returns `emptyDetailWindowResult()`, which carries `HasMoreAbove: false`. That lands here as:

```ts
HasMoreAbove: result.HasMoreAbove || droppedOlderRows   // false || false
```

Nothing restores it afterwards: `RefreshLatest`'s joined branch returns `this.cursor.HasMoreAbove` unchanged, so the value is sticky for the life of the store. The user-visible result is that one transient failure while scrolling up destroys the sentinel (`@if (HasMoreAbove)` unmounts it, `syncOlderObserver` disconnects) and they cannot page back again until they switch conversations and return — with nothing said.

`LoadLatest` on line 193 has the same shape: a failed first read renders an empty transcript with no error, which reads to the user as "my conversation is gone".

Smallest fix is to make failure distinguishable in the result — an `Ok: false` / `Failed: true` flag on `DetailWindowLoadResult` — and preserve the previous `HasMoreAbove` when it's set. Worth doing; everything else about the cursor bookkeeping here is careful, and this is the one path where a degradation is both silent and unrecoverable.

packages/MJCoreEntities/src/engines/conversations.ts
```diff
@@ -1144,6 +1352,186 @@ export class ConversationEngine extends BaseEngine<ConversationEngine> {
...
+        contextUser: UserInfo
+    ): Promise<DetailWindowLoadResult> {
+        const pageSize = params.PageSize ?? DEFAULT_WINDOW_PAGE_SIZE;
+        const overread = params.RawOverread ?? pageSize * DETAIL_WINDOW_OVERREAD_FACTOR;
+
+        const page = await this.fetchDetailRowsBySequence(
+            params.ConversationID, params.BeforeSequence, overread, contextUser
+        );
+        if (!page || page.length === 0) {
+            return emptyDetailWindowResult();
```
rkihm-BC: Two things about the failure contract here, both tied to the sticky-`HasMoreAbove` problem on the store side:

1. **`emptyDetailWindowResult()` is indistinguishable from a genuinely empty conversation.** It carries `HasMoreAbove: false`, and the store ORs that straight into its cursor, so a failed read is read downstream as "you've reached the start of the conversation". The same applies to `hasOlderDetails`, which returns `probe.Success && …` — a failed probe reports "nothing older" rather than "unknown".

2. **The docstring's "Never throws" isn't accurate.** There's no `try`/`catch` in this method; only the `!result.Success` paths are handled. A provider that rejects rather than returning `Success: false` propagates out through `fetchPageFillingTimeline` and `LoadOlder` (which has `try`/`finally`, no `catch`) and surfaces as an unhandled rejection. Ironically that path degrades *better* than the handled one — the cursor is left untouched, so the sentinel survives and the user can retry — but the doc should either match the code or the code should match the doc.

A failure flag on `DetailWindowLoadResult` covers both.

packages/Angular/Generic/conversations/src/lib/services/conversation-agent.service.ts
```diff
@@ -583,4 +623,109 @@ ${compactHistory}${artifactContext}
...
+   * The array scan this replaces was correct only while `conversationHistory` held the whole
+   * conversation. Under transcript windowing it holds one page, so the scan silently returns
+   * null for any artifact below the window's oldest row — exactly the long "modify this
+   * again" exchanges where payload continuity matters most. A null payload is a LEGAL agent
+   * input, so the failure is silent: the agent regenerates from scratch instead of modifying.
+   *
+   * One round trip. The subquery walks details -> Output junctions, and versions carry
+   * `__mj_CreatedAt`, so "newest" is expressible without joining back for `Sequence`.
+   */
+  public async FindLatestAgentOutputVersion(
```
rkihm-BC, line 646: **Its sibling wasn't converted: `findAllAgentArtifacts` (line 549) is still an array scan over the windowed maps.**

This method is the right fix, and `loadPreviousPayloadForAgent` now uses it. But the earlier review round named `findAllAgentArtifacts()` alongside it, and that one still walks `conversationDetails` and looks each row up in `context.artifactsByDetailId` / `context.agentRunsByDetailId` — both now scoped to the loaded window.

Its output is the "**Prior Artifacts Created by This Agent**" block fed to the `Check Sage Intent` prompt at line 455, which is what produces `targetArtifactVersionId`. Under windowing the classifier reasons over a truncated artifact list, so it can name the wrong version or miss continuity entirely. (The `.slice(-10)` on the history beside it is fine — it's the artifact scan that narrows.)

Blast radius is smaller than it was, since the caller now falls back to this method when the intent check names nothing. But it's a real remaining gap in exactly the class the rest of this file just closed, and the same query shape would close it: this agent's artifacts in this conversation, `Direction='Output'`, newest-first, capped.

`findLastNonSageAgentId()` and `handleSilentObservation()` in `message-input.component.ts` also still route off the window — lower severity as noted before, but worth an explicit "deferring" rather than leaving it ambiguous.

packages/Angular/Generic/conversations/src/lib/components/conversation/conversation-chat-area.component.ts
```diff
@@ -867,9 +880,164 @@ export class ConversationChatAreaComponent extends BaseAngularComponent implemen
...
+   * O(n^2) — enough to spend back the savings this windowing work exists to produce.
+   */
+  private async refreshAfterPaging(conversationId: string): Promise<void> {
+    const snapshot = this.windowStore.GetSnapshot();
+    this.messages = [...snapshot.Details];
+
+    // Older rows carry their own artifacts/ratings/runs — clear the once-per-conversation
+    // guard so loadPeripheralData reprocesses the now-larger window.
+    this.lastLoadedConversationId = null;
+    await this.loadPeripheralData(conversationId, snapshot);
```
rkihm-BC: **The common paging path pays the cost the jump paths were built to avoid.**

`loadPeripheralData` clears and rebuilds `agentRunsByDetailId` / `artifactsByDetailId` / `systemArtifactsByDetailId` / `ratingsByDetailId` over the **whole accumulated window**, re-queries attachments for every loaded ID, and re-queries realtime session meta for every loaded session ID — reallocating every `LazyArtifactInfo` each time.

`refreshAfterPaging` exists precisely so a date jump pays that once instead of per page (your own docstring: *"calling it per iteration made a 50-page jump O(n²)"*). But `onOlderMessagesRequested` calls `refreshAfterPaging` on **every** sentinel fire, so ordinary scroll-up paging is the O(n²) case: ten pages back re-queries attachments for roughly fifty-five pages' worth of IDs.

Stacked on the per-page round-trip profile already documented on `LoadDetailWindow`, paging back a long way ends up more expensive than the single full load this replaces. Merging just the new page's peripherals into the existing maps — the store already accumulates them correctly in `mergePeripherals` — would make the cost proportional to the page rather than the window.

Minor, same area: this path assigns `this.messages` twice per page (once here, once at the end of `loadPeripheralData`), so `updateMessages` renders twice per prepend.

packages/Angular/Generic/conversations/src/lib/components/conversation/conversation-chat-area.component.ts
```diff
@@ -1503,22 +1674,29 @@ export class ConversationChatAreaComponent extends BaseAngularComponent implemen
 
   private async loadMessages(conversationId: string, loadToken: number): Promise<void> {
     try {
-      // Single source of truth: ConversationEngine handles caching and DB queries.
-      // Cache hit = instant (no DB). Cache miss = one GetConversationComplete query.
-      // If peripheral data (artifacts/ratings) changed externally, force refresh to pick up joined fields.
-      const existingEntry = this.engine.GetCachedDetailEntry(conversationId);
-      const forceRefresh = existingEntry?.PeripheralDataStale === true;
-      const cacheEntry = await this.engine.LoadConversationDetails(conversationId, this.currentUser, forceRefresh);
+      // WINDOWED read: only the newest page of the transcript, not the whole conversation.
```
rkihm-BC, line 1689: **Worth calling out that this trades away the warm-conversation cache.**

The path this replaces went through `LoadConversationDetails`, which cached per conversation ID in `_detailCache` — so switching back to a conversation you'd already opened was instant with zero database work. `LoadLatest` always fetches.

The comment frames "a window is always fetched fresh, so no cache can go stale" as the upside, and for correctness it genuinely is. But the cost side isn't stated: every conversation re-entry now costs a window read plus the follow-up queries per the round-trip profile on `LoadDetailWindow`. For a user tabbing between a few conversations that's a visible regression against today's behaviour, on a change whose headline is that opening a conversation got cheaper.

Not a blocker and not obviously worth a partial-window cache — but worth measuring, and worth saying out loud in the comment so the next reader knows it was a deliberate trade rather than an oversight.

packages/Angular/Generic/conversations/src/lib/components/message/message-list.component.ts
```diff
@@ -383,6 +931,286 @@ export class MessageListComponent extends BaseAngularComponent implements OnInit
     }
   }
 
+  /**
+   * The span of timeline indices that stay mounted.
+   *
+   * Follows the tail by default. Once the user scrolls back to a spacer,
+   * {@link _mountedTopKey} pins the top of the span so their position doesn't collapse
```
rkihm-BC: `{@link _mountedTopKey}` doesn't exist anywhere on the branch — leftover from the pinned-top-key design that the scroll-position approach replaced (the paragraph below it describes the mechanism that actually shipped, and explains why the feedback loop made the earlier one unworkable).

Trivial, but the comments in this file are unusually load-bearing — a reader will trust this one and go looking for the field.

.changeset/open-cycles-bet.md
```diff
@@ -0,0 +1,5 @@
+---
+"@memberjunction/ng-conversations": patch
+---
+
+This is a patch for conversation details UI work
```
rkihm-BC: This duplicates `windowed-conversation-transcript.md`, which already covers `@memberjunction/ng-conversations` at `patch` and describes the change properly. Two changesets for the same package at the same bump level just produces a second, contentless entry in the release notes — suggest deleting this file.

(`patch` is the right level for both — no migration or metadata in this branch.)

rkihm-BC (commented):
## Review — windowed transcript

Ran the tests on this head: `ng-conversations` **1154 passed** (2 suites failed to *load* on an unbuilt `ng-task-graph-editor` in my sandbox — environmental, not this PR), `MJCoreEntities` **623 passed, 0 type errors**.

The architecture is strong, and the two decisions that could have done real damage were both spotted and defended: never writing a window into `_detailCache` (with a test asserting it, and `LoadConversationDetails` genuinely untouched so `GetAgentContextWindow` and `ConversationToolManager` still see full history), and counting **display items** rather than rows so a realtime session is never split across a page. `SelectLatestTimelinePage` filtering the original row set rather than walking timeline items is the subtle, correct version of that. The over-read growth loop breaking on `result.Details.length < overread` — a read that came back under its own `MaxRows` cannot have been truncated by collapse — is the sharp termination test rather than the obvious one. Generation-counter invalidation is applied consistently across every `await` boundary. And the refresh-gap handling in `mergeRefreshedRows` recognises that a tail which ran away by more than a page produces a merged set that sorts into order and *looks* contiguous while missing its middle — choosing a visible gap over an invisible one is the right call and most PRs of this shape get it wrong.

Seven items below. Two I'd fix before merge, two are fast-follows, three are hygiene.

---

### Before merge

**1. A failed older-page read permanently kills the "earlier messages" sentinel.** `LoadDetailWindow` returns `emptyDetailWindowResult()` (`HasMoreAbove: false`) whenever `RunView` reports `Success: false`, `LoadOlder` ORs that into the cursor, and nothing ever restores it. One transient failure and the user cannot page back for the rest of that conversation's lifetime, silently. Inline on `conversation-detail-window.store.ts` and `conversations.ts`.

**2. `findAllAgentArtifacts` was not converted to a query.** It was named explicitly alongside `loadPreviousPayloadForAgent` in the earlier review round; that one was converted, this one still scans the display array against the windowed maps. Inline on `conversation-agent.service.ts`.

### Fast-follow

**3. Ordinary sentinel paging rebuilds every peripheral map over the whole accumulated window.** The jump paths were deliberately optimised to pay this once — the docstring on `refreshAfterPaging` says exactly why — but the common path pays it per page. Inline.

**4. Warm-conversation switching lost its cache.** `LoadConversationDetails` cached by conversation ID, so re-entering a conversation was instant with no DB. Every re-entry is now a fresh window read plus its follow-ups. Inline.

### Hygiene

**5. The description no longer matches the branch.** It lists 6 changed files and ~1,200 test lines; the diff is 25 files / +5150. Three substantial features added after the first review round aren't mentioned at all: the pins count-then-hydrate-on-open rework, the whole date-navigation subsystem (`date-jump.ts` + 191 test lines), and the `conversation-agent.service` query rework. Worth refreshing before merge — a reviewer arriving cold will review the wrong thing.

**6. Two small ones:** a redundant changeset, and a `{@link}` pointing at a field that doesn't exist. Both inline.

**7. No integration coverage, and there's an obvious home for it.** `packages/TestingFramework/integration-test-suite/src/checks/conversation-compaction.checks.ts` already drives `ConversationEngine` against a real database with real conversation fixtures. `LoadDetailWindow` issues live `RunView`s against real views, and the parts most likely to break are exactly the parts unit mocks can't reach: the `Direction='Output'` filter, the three-table artifact rebuild, `Sequence` ordering under the real assignment trigger, and session expansion. A handful of checks in that bundle would cover the seam this PR creates.

---

Two things I want to explicitly *not* raise, having checked them:

- **`packages/MJExplorer/package.json` is fine as-is.** Your note on it is correct. `origin/next` already declares `acorn`, `@auth0/auth0-spa-js`, `ng-ui-components`, `ng-entity-viewer`, `ng-shared-generic` and `ng-explorer-service-worker` — the dependency *set* here is identical, only the ordering moved, which is what `.github/scripts/check-explorer-external-deps.mjs --write` produces (its list is regenerated, never hand-maintained). Restoring the file to `origin/next`'s copy would avoid a spurious conflict, but nothing is wrong with it.
- **The `Sequence` cursor has no tie-breaker, and that's fine.** The v5.49 migration makes `IX_ConversationDetail_ConversationID_Sequence` non-unique deliberately and documents that duplicates only arise from a bulk insert with triggers disabled. Not worth a code change.

PR #3913, opened 2026-08-17
fix(agent-manager): hybrid agent search so new agents are findable, and load specs via a direct action

bg-BC:
## Summary

Two defects that both made the Agent Manager fail on agents it had *just* created or was asked to modify:

1. **A newly created agent was undiscoverable for up to ~24h.** `Find Best Agent` / `Find Candidate Agents` ranked semantic-only over the `AI Agents Search` vectors, which are populated by the 04:00 UTC Entity Vector Sync job. Until that job ran, the semantic pass never returned the new agent's id, so the hydration join dropped it — even though the agent was already live in the `AIEngine` cache.
2. **Loading an existing agent's spec always failed.** Agent Manager's modification workflow called `Agent Spec Loader`, a **Flow** agent wired as a **Sub-Agent**. `FlowAgentType.DetermineInitialStep` rejects any Flow agent invoked with a `parentRun`, because a Flow agent submits to the task-graph dispatcher and detaches — the awaiting Loop parent would get `started` back before any work ran. `Agent Spec Loader` was the only Flow agent declared as a sub-agent, so it was the only one hitting this.

---

## 1. Find-agent search goes hybrid

`packages/Actions/CoreActions`

Both actions share `BaseFindAgentsAction`, which now searches in **HYBRID** mode: the semantic pass *plus* a lexical name/description `LIKE` pass over the live `MJ: AI Agents` table. The lexical pass matches an unembedded agent by text, so it surfaces the moment the agent exists.

**Scoring needed care.** Hybrid fuses the two passes with Reciprocal Rank Fusion, whose `score` is a tiny rank-based value (~`1/rrfK`) — **not** cosine — and the pipeline applies `MinimumSimilarityScore` (default `0.5`) to that blended score. A naive switch to hybrid therefore filters out *everything*. The fix splits the responsibility:

- `runSemanticEntitySearch` gains a `mode` argument, defaulting to `'semantic'`, so the existing Find Best/Candidate and Search Query Catalog callers are untouched. In hybrid it passes `minScore: 0` to the provider.
- `BaseFindAgentsAction` re-applies the cosine floor itself against `components.semantic`, and lets any `components.lexical` hit through — that lexical pass is the whole point here.
- The reported `similarityScore` is `max(semantic, lexical)`, so a match that qualified on strong lexical merit but had a below-floor semantic component reports the meaningful score rather than the low one.

### What hybrid covers — and what it doesn't

Hybrid **narrows** the vector-sync staleness window; it does not close it:

- **Updates go stale too, not just new agents.** The vector reflects whatever the record said at the last sync, so an edited name or description keeps matching the *old* text in semantic search until the next 04:00 UTC run.
- **The lexical fallback matches verbatim substrings only.** `searchEntitiesLexicalPass` builds `Name LIKE '%<entire searchText>%'` — the whole query as one substring, not tokenized. A short query like `"invoice"` finds `"Invoice Reconciler"`; a full task description generally will not. So "hybrid" ≠ "always findable."

*(For later, not this PR: Agent Manager could close the window by calling the `Vectorize Entity` action after Builder persists — with two traps: `EntityDocumentType` defaults to `'Record Duplicate'`, so `'Search'` must be passed explicitly, and the action scopes by entity, not by record.)*

## 2. Load Agent Spec as a direct action

`metadata/agents`, `metadata/prompts`, `packages/Actions/CoreActions`

The Flow wrapper was redundant: the `Load Agent Spec` **action** already returns the `AgentSpec` (`AgentSpecSync.LoadFromDatabase` → `toJSON`). Agent Manager is itself a Loop agent that already calls actions, so it now calls that action directly — `AgentID = payload.ID`, and the action's `AgentSpec` output merges into the payload. No framework change, no per-field mapping, no sub-agent hop.

- Added `Load Agent Spec` as a direct action on Agent Manager.
- **The action's `AgentSpec` output param is now the complete spec.** It previously returned `truncatePromptTexts(agentSpec)` — sub-agent `PromptText` cut to 100 chars — under the same key the caller merges into the payload and Builder persists. Since `AgentSpecSync.savePrompts` writes `TemplateText` unconditionally, modifying any agent with sub-agents would have overwritten their prompt templates with 100-char prefixes. Truncation now applies **only** to the human-readable `Message`; the output param is the full `toJSON()`.
- **Disabled** the `Agent Spec Loader` sub-agent rather than deleting its block. MetadataSync is update-only, so an omitted record would stay `Active` in every existing database and Agent Manager would still resolve it. Status `Disabled` (the AI Agents status value-list is `Active | Disabled | Pending` — not `Inactive`) makes the runtime's `Status === 'Active'` sub-agent filter exclude it.
- **Tombstoned the loader's two child records** (`MJ: AI Agent Actions` `90BB672E…`, `MJ: AI Agent Steps` `33533220…`) with `deleteRecord: { "delete": true }` rather than dropping them from the file — otherwise, by the same update-only logic, they'd linger `Active` and orphaned under a disabled parent and never reach the release migration.
- Rewrote the eight prompt references from "call the Agent Spec Loader sub-agent" to "call the Load Agent Spec action; merge its `AgentSpec` output into the payload."

---

## Files

| File | |
|---|---|
| `packages/Actions/CoreActions/src/custom/ai/semantic-entity-search.helper.ts` | `mode` parameter; `minScore: 0` passthrough in hybrid |
| `packages/Actions/CoreActions/src/custom/ai/base-find-agents.action.ts` | Hybrid invocation; caller-side cosine floor on `components.semantic`; `max(semantic, lexical)` reported score |
| `packages/Actions/CoreActions/src/custom/ai/load-agent-spec.action.ts` | `AgentSpec` output param is the full spec; truncation only on `Message` |
| `packages/Actions/CoreActions/src/__tests__/find-entity-search.actions.test.ts` | Hybrid / score tests |
| `packages/Actions/CoreActions/src/__tests__/load-agent-spec.action.test.ts` | Lossless-output-param tests |
| `packages/AI/AgentManager/core/src/__tests__/agent-spec-sync-prompt-roundtrip.test.ts` | Save-path round-trip: prompt persists verbatim |
| `metadata/agents/.agent-manager.json` | `Load Agent Spec` action added; loader disabled; child records tombstoned |
| `metadata/prompts/templates/agent-manager/agent-manager.template.md` | Eight prompt references rewritten |
| `.changeset/hungry-jokes-design.md` | Changeset (`minor`) |

## Testing

- Hybrid: two existing agent cases updated for `components`, asserting `mode: 'hybrid'` and `minScore: 0`; cases added for a lexical-only new agent surfacing, a below-floor semantic-only match dropping, and a lexical-merit match reporting the stronger score.
- Load Agent Spec: tests assert the `AgentSpec` output param carries the full prompt while the `Message` is truncated.
- Persistence round-trip: `AgentSpecSync` save writes the full prompt `TemplateText` verbatim (the write path was previously unexercised).
- CoreActions suite **355/355**; ai-agent-manager suite **45/45**.

## Notes

No migration and no schema change. Metadata changes validated with `mj sync push --dir=metadata --include=agents --dry-run` → 0 errors.


packages/Actions/CoreActions/src/custom/ai/base-find-agents.action.ts
```diff
@@ -71,10 +77,21 @@ export abstract class BaseFindAgentsAction extends BaseAction {
...
+                const passes = (semantic != null && semantic >= minimumSimilarityScore) || lexical != null;
+                if (!passes) {
+                    continue;
+                }
                 const agent = agentsById.get(NormalizeUUID(r.recordId));
                 if (agent) {
                     matched.push(agent);
-                    scoreById.set(NormalizeUUID(agent.ID), r.score);
+                    // Report a meaningful 0–1 score: cosine when present, else the lexical score.
+                    scoreById.set(NormalizeUUID(agent.ID), semantic ?? lexical ?? r.score);
```
rkihm-BC: When a result passes on lexical merit but also has a below-floor semantic component, `semantic ?? lexical` reports the below-floor value — e.g. floor `0.5`, `semantic: 0.3`, `lexical: 0.85` reports `0.3`. `Math.max(semantic ?? 0, lexical ?? 0)` — or preferring whichever component actually passed — keeps the reported score consistent with the comment above it.

metadata/agents/.agent-manager.json
```diff
@@ -575,92 +586,8 @@
...
-                },
-                "sync": {
-                  "lastModified": "2026-02-13T21:09:22.378Z",
-                  "checksum": "d9ff376573cc6195d21211b58e6be764b69fb2663ba23a3f442c11274ed2a8d1"
-                }
-              }
-            ]
-          },
           "primaryKey": {
             "ID": "09D7E6EE-A3DC-4DBD-BA72-349FC68FE8C0"
```
rkihm-BC, line 644: The `MJ: AI Agent Actions` (`90BB672E…`) and `MJ: AI Agent Steps` (`33533220…`) records that used to live here need `deleteRecord: { "delete": true }` tombstones rather than being dropped from the file.

Omitting them means MetadataSync leaves them `Active` in existing databases — the same reasoning that correctly kept this parent record as `Disabled` — and they never reach the consolidated release migration at all, since the push emits no SQL for records it never sees. Fresh installs never create them, existing installs keep them forever, and nothing reconciles the two.

.changeset/hungry-jokes-design.md
```diff
@@ -0,0 +1,5 @@
+---
+"@memberjunction/core-actions": minor
+---
+
+Fixed Find Agent action to use a hybrid search on avaliable agent to avoid newly created agents being unfindable to the Agent Manager due to the daily vectorization of searchable entities. Changed AgentManager's subagent call to Agent Spec Loader (workflow agent) to a direct call to the underlying action to avoid automatic failing of the sub-workflow agent.
```
rkihm-BC: Worth widening this: the staleness window affects **updates to already-vectorized records** as well as new ones — an edited name or description keeps matching the old text until the next sync.

Also worth noting the lexical fallback only matches when the search text appears verbatim in the agent's name/description, so it narrows the window rather than closing it.

rkihm-BC (changes requested):
Nice work on both diagnoses — the `FlowAgentType.DetermineInitialStep` root cause is a genuinely non-obvious catch, and I verified the two supporting details that are easy to get wrong: `Status` on AI Agents really is `Active | Disabled | Pending` (so `Inactive` would have failed), and `base-agent.ts:6680` really does filter sub-agents on `Status === 'Active'`, so disabling genuinely removes it. The RRF-vs-cosine insight is sharp and correct — `minScore` is applied to the blended score, so passing the 0.5 cosine floor would have filtered everything, and defaulting the new `mode` to `'semantic'` keeps every existing caller untouched.

Three things below. One is a data-loss path I'd hold merge on; the other two are small.

## 1. Hybrid is the right stop-gap — please document what it does and doesn't cover

No code change needed. But the PR description and changeset should be precise about the gap, because it's wider than "new agents":

- **Updates are affected too, not just creates.** The vector reflects whatever the record said at the last sync. Edit an agent's name or description and semantic search keeps matching the *old* text until 04:00 UTC. Worth saying explicitly, since "newly created agents" reads as though updates are fine.
- **The lexical fallback is a partial mitigation, not full coverage.** `searchEntitiesLexicalPass` builds `Name LIKE '%<entire searchText>%'` — the whole search string as one substring, not tokenised. So it matches only when the search text appears verbatim in the agent's name or another searchable field. A short query like `"invoice"` finds `"Invoice Reconciler"`; a full task description generally won't. The docs should say so, or the next person will read "hybrid" as "always findable."

Worth knowing for later (not this PR): the `Vectorize Entity` action is Active and callable, so Agent Manager could close the window properly by vectorizing after Builder persists. Two traps if anyone tries — `EntityDocumentType` defaults to `'Record Duplicate'`, so you must pass `'Search'` explicitly or it's a silent `NO_DOCUMENTS` no-op; and the action only scopes by entity, not by record (the underlying syncer supports a `listID`, but the action doesn't expose it).

## 2. 🔴 This makes a data-loss path reachable — sub-agent prompts get overwritten with 100-char truncations

`Load Agent Spec` doesn't return the complete spec: it returns `truncatePromptTexts(agentSpec)`, which recursively cuts every sub-agent's `PromptText` to **100 characters + '...'**. `IncludeSubAgents` defaults to `true` in both the code and the declared metadata param.

Before this PR the load always failed, so this never fired. Now it does. The full chain, with no guard at any link:

1. `loadFromEntities` recurses into sub-agents; `mapPromptEntityToSpec` sets `PromptText: fullPrompt.TemplateText` — full text
2. `truncateSubAgent` cuts it to 100 chars
3. Template: the payload *is* the AgentSpec, and Builder persists it
4. `builder-agent.ts:54`: `params.payload as AgentSpec` → `markDirty()` → `markLoaded()` → `SaveToDatabase()`
5. `saveSubAgents` → `saveChildSubAgent` (existing branch) rebuilds the child spec and recursively saves it
6. `savePrompts` → `promptEntity.TemplateText = promptSpec.PromptText`, unconditionally — no comparison with the stored value

So modifying any agent that has sub-agents overwrites their prompt templates with a 100-character prefix. `AIPrompt` rows are never deleted, so it's an overwrite recoverable through Record Changes — but only if someone notices.

**Suggested fix:** apply `truncatePromptTexts` only to the `Message` string and leave the `AgentSpec` output param as the full `toJSON()`. The `Message` is the display surface where the size concern is real; the output param is the one that merges into the payload and round-trips to persistence. The general principle worth keeping: a loader should never emit a lossy value under the same key the writer reads.

**Please don't fix this with `IncludeSubAgents: false`** — I checked, and it's worse. `mapChildAgentToSpec` returns only `{ ID, Name, StartingPayloadValidationMode }`, and `saveAgent` assigns every field unconditionally with `|| null` fallbacks. Saving a shallow sub-agent spec would blank `Description`, `DriverClass`, `FunctionalRequirements`, `TechnicalDesign` and all the limits, and reset `Status`, `InvocationMode` and `ModelSelectionMode` to defaults.

Also worth a test: load → save round-trip on an agent with a sub-agent, asserting the sub-agent's `TemplateText` is unchanged. Every current test mocks `SearchEntity` or the persistence layer, so nothing exercises this path.

## 3. Tombstone the two removed child records instead of just omitting them

The agent record is kept-but-`Disabled` precisely because "MetadataSync is update-only, so an omitted record would stay `Active`" — correct. But the nested `MJ: AI Agent Actions` (`90BB672E…`) and `MJ: AI Agent Steps` (`33533220…`) blocks were *deleted from the file*, which by that same argument leaves them `Active` in every existing database, orphaned under a disabled parent.

They'd also never appear in the consolidated release migration, since the push emits no SQL for records it never sees — so fresh installs never create them, existing installs keep them forever, and nothing reconciles the two.

The supported mechanism is a tombstone — keep the `primaryKey`, add:

```json
"deleteRecord": { "delete": true }
```

The push then deletes them, the `DELETE`s land in the SQL log and therefore in the release migration, and both populations converge. A `DELETE` for a row a fresh install never created is a harmless no-op. Two caveats from the MetadataSync README: it can fail on FK constraints, and it's one-shot — once `deletedAt` is stamped it won't retry.

## Minor

The reported score prefers `semantic ?? lexical`, so a result that passed *on lexical merit* with a below-floor semantic component reports the below-floor number — floor `0.5`, `semantic: 0.3`, `lexical: 0.85` reports `0.3`. Contradicts the stated goal of reporting a meaningful 0–1 score.

rkihm-BC (approved):
LGTM

PR #3915, opened 2026-08-17
fix(remote-browser): name a browser within an agent session with instanceKey (#3531)

### LINKED ISSUE #3531: Realtime: only one surface per channel type — two Remote Browsers in one session is blocked at four layers

MS-BC:
Fixes #3531.

## The defect

`RemoteBrowserEngine` keys its agent-session map on the agent session id **alone**:

```ts
this.agentSessionToEngineSession.get(agentSessionID)
```

That quietly makes *one remote browser per agent session* a framework invariant with no opt-out. A second surface's lazy start finds the first one's mapping and returns it, so both surfaces drive the same Chrome: one live view, one screencast, one audio stream — and a `StopScreencast` from either tears down the other's. A caller cannot say *which* browser it means, because there was only ever one.

The blocked shape is a realtime workspace that shows two browsers side by side (one following the conversation, one the participant drives). Downstream, Caliber has been carrying this as a local `node_modules` patch — `caliber-157-instance-key.patch` — because there is no app-side substitute for a key the framework owns.

## The fix

An optional `instanceKey` names a browser **within** the agent session:

```ts
private agentSessionKey(agentSessionID: string, instanceKey?: string): string {
    const base = agentSessionID.trim().toLowerCase();
    const name = (instanceKey ?? '').trim().toLowerCase();
    return name.length === 0 ? base : `${base}::${name}`;
}
```

Threaded through `StartSessionForAgentSession`, `GetSessionForAgentSession`, `EndSessionForAgentSession`, `lazyStartBrowserForAgentSession`, `getHandleForAgentSession`, and `AchieveGoal` via a new `AchieveGoalParams.InstanceKey`. The six `RemoteBrowserActionResolver` mutations that address a *live* session accept and forward it: `InterpretRemoteBrowserPage`, `RemoteBrowserSnapshot`, `StopRemoteBrowserScreencast`, `StopRemoteBrowserAudioStream`, `RelayRemoteBrowserHumanInput`, `GetRemoteBrowserSelection`. There are no remaining unkeyed `GetSessionForAgentSession(agentSessionID)` calls.

### Three decisions worth naming

**Omitting the key is exactly today's behaviour, by construction.** No key composes to the bare agent session id — not to `id::` or `id::default` — so every existing caller resolves the same instance it always did and the pre-existing agent-session tests pass unchanged. An empty or whitespace key is the unnamed instance too, so a caller threading an absent value through as `''` lands where it did before the argument existed.

**Keys are trimmed and lowercased.** The value is typed by hand into a channel config; `Primary` and `primary` being two browsers would be a spelling trap, not a feature.

**The composite stays scoped to the agent session.** Two concurrent sessions that both name their second surface `resume` get their own browsers. Start coalescing — the fix that stopped four near-simultaneous callers launching four Chromes — is keyed the same way, so it still collapses a race on *one* instance without collapsing two *different* surfaces into one. That last one is the trap: keying the coalescing map more coarsely than the session map turns a duplicate-browser bug into a missing-browser bug.

## Tests

Six added to `remote-browser-engine.test.ts`, beside the existing agent-session lifecycle tests rather than replacing them (their continued passing *is* the compatibility assertion):

- a separate browser per `instanceKey` in one agent session
- the unkeyed instance stays distinct from a named one, and `''` / `'   '` resolve to it
- keys are case- and whitespace-insensitive
- one agent session does not bleed into another using the same key
- concurrent starts coalesce **per instance**, not per agent session
- ending one instance leaves its siblings live

**52 tests pass** in `@memberjunction/remote-browser-server` (was 46). `tsc --noEmit` is clean for that package; `MJServer` has 30 pre-existing errors from unbuilt sibling packages in this worktree and **zero** in the file touched here.

Mutation-verified — both mutants are killed:

| mutant | result |
|---|---|
| `agentSessionKey` returns `base`, ignoring the key | 5 of 6 fail |
| drop `.trim().toLowerCase()` on the key | 4 fail |

---


MS-BC:
Pushed a second commit. The first one keyed the **engine** by instance, which turned out to be necessary and not sufficient — walking the resolver end to end found three more layers where a second surface was still blocked. All four of the issue's layers are now covered:

**1. Four operations could not name an instance.** `ExecuteRemoteBrowserAction`, `ExecuteRemoteBrowserGoal`, `StartRemoteBrowserScreencast`, `StartRemoteBrowserAudioStream`. The two `Start*` mutations are how a surface binds at all, so without them the second browser was resolvable but unreachable. `GetRemoteBrowserGoalResult` is deliberately left alone — it is keyed by `goalRunID`, which already identifies one run on one instance.

**2. The idempotency guard was a lockout.** `startedScreencasts` / `startedAudioStreams` were keyed by agent session alone, so the *first* surface's entry made the *second* surface's first bind return a silent `Streaming: true` no-op. Both sets are now keyed per surface through a shared `streamKey` helper.

**3. The published envelopes could not be routed.** Screencast frames and audio chunks carried only `agentSessionID`, so two surfaces published indistinguishable payloads on one topic and a client would paint both panes from whichever arrived last. Both envelopes now carry `instanceKey` (explicitly `null` for the unnamed instance, so a client matches on a value rather than on a missing property).

One design note worth calling out: normalization now has exactly one home. `NormalizeInstanceKey` is exported from the engine package and used by the engine's session map, the resolver's stream keys, and the envelopes. Three private copies of trim-and-lowercase is how a key ends up case-insensitive in one layer and case-sensitive in the next, which would only ever show up in production with two surfaces open.

55 tests pass (was 52) — three added pinning the key grammar itself, since three layers now compare it. `MJServer` typechecks with zero errors in the touched file (30 pre-existing errors elsewhere from unbuilt sibling packages in my worktree).

PR #3916, opened 2026-08-17
fix(remote-browser): recover a dead browser handle instead of poisoning the session (#3598)

### LINKED ISSUE #3598: Remote Browser: a dead session handle is never recovered — the surface fails for the rest of the session
## What happens

A surface's browser can go away without the engine being told: the external Chrome behind `Self-Hosted Chrome` is closed, a backend container recycles, a CDP target vanishes. `RemoteBrowserEngine`'s live map keeps the handle, so `StartSessionForAgentSession` finds the stale entry and returns it, and **every subsequent call throws `Browser not launched. Call Launch() before using the adapter.` for the rest of the session.**

Observed live: 232 `Browser not launched` errors in one MJAPI run, against surfaces that had launched successfully minutes earlier, with no `Session … ended` line in between. User-visible symptom is the voice agent saying *"the shared browser session isn't launched right now"* on every request, indefinitely, while the pane sits frozen on its last good frame.

`RemoteBrowserSnapshot` already anticipates this exact state in a comment — *"a stale/dead handle still in the live map, a mid-poll teardown race, or a recycled backend container"* — but only degrades around it. Nothing discards the mapping, so nothing ever heals.

## Two halves, and the second is easy to miss

1. **The handle is never discarded.** A dead handle is not a failure to report, it is a mapping to delete. Dropping it and starting once more recovers transparently.
2. **The live VIEW is not re-attached.** Fixing (1) alone is worse than the original bug: the backend heals into a *new* engine session while the client's screencast stays bound to the discarded one. The agent then truthfully reports a page the person cannot see — the agent and the screen disagreeing. The client asked for a screencast once at bind time and has no reason to ask again, so it must be re-attached server-side.

Note the replacement usually happens on the **snapshot-poll path**, not the action path: the ~700ms poll discovers the dead handle, and the next action simply starts a new browser without any error surfacing. So re-attachment keyed only to an action's failure branch does not fire.

## Suggested shape

- A narrow predicate for "the handle is dead" (`Browser not launched`, target/browser closed, disconnected) — explicitly NOT bad selectors, timeouts or refused pages, which are real answers and must not cost a working browser its page state.
- On a dead handle: end the mapping, relaunch **once** (bounded — a loop turns a visible fault into a hang), re-attach the screencast for that surface.
- Remember the frame sink per surface so re-attachment is possible at all.

## Environment

MJ 5.51.0, `Self-Hosted Chrome` provider, `mode=Collaborative`, two Remote Browser surfaces in one realtime session (see #3531 for the instance-key work this builds on). Patched locally in `RemoteBrowserActionResolver`; happy to open a PR.

MS-BC:
Fixes #3598. #3915 (`instanceKey`) has merged, so this now targets `next` directly and is up to date with it.

## The defect

A surface's browser can go away without the engine being told: the external Chrome behind `Self-Hosted Chrome` is closed, a backend container recycles, a CDP target vanishes. `RemoteBrowserEngine`'s live map keeps the handle, so `StartSessionForAgentSession` finds the stale entry and returns it — and every subsequent call throws `Browser not launched. Call Launch() before using the adapter.` for the rest of the session.

Observed live: **232 of them in one MJAPI run**, against surfaces that had launched successfully minutes earlier, with no `Session … ended` line in between. The voice agent said *"the shared browser session isn't launched right now"* on every request, indefinitely, while the pane sat frozen on its last good frame.

`RemoteBrowserSnapshot` already anticipated this exact state in a comment — *"a stale/dead handle still in the live map, a mid-poll teardown race, or a recycled backend container"* — but only degraded around it. Nothing discarded the mapping, so nothing ever healed.

## The fix

`RemoteBrowserEngine.RecoverDeadAgentSession(agentSessionID, error, opts)` discards the dead mapping, launches **one** replacement, and re-attaches the screencast.

**The caller does not decide what "dead" means.** It hands over the error it already caught; the engine decides. That keeps the grammar in exactly one place — `IsDeadBrowserHandleError`, a *closed* list of "the browser or its transport is gone".

The asymmetry is the whole design:

| direction | cost |
|---|---|
| too narrow (missed phrase) | the original bug survives for that phrasing — an error the caller reports, i.e. today's behaviour |
| too wide (false positive) | a bad selector costs a **healthy** browser its cookies, login and scroll position; the user watches their page reset because a click missed |

So anything unrecognised is treated as a real answer from a live browser. Timeouts, `waiting for selector`, `no element matching` and every `net::ERR_*` navigation failure are explicitly **not** dead handles.

### Re-attaching the view is half the fix, not a nicety

The issue is right that this is easy to miss. The client asks for a screencast exactly once, at bind time, and has no reason to ask again — so healing the backend *alone* leaves it watching the discarded session while the agent truthfully narrates a page the person cannot see. That is strictly worse than the original freeze. The engine now remembers each live session's frame sink so a replacement can be re-piped to it.

The inverse case is covered too: a screencast the host **deliberately** stopped is forgotten at stop time, so a later recovery cannot resurrect a surface the host chose to release.

### Bounded in both directions

- **Concurrent faults coalesce.** The ~700ms snapshot poll and the next agent action both meet the dead handle; without coalescing each launches its own replacement and one dead browser becomes two live ones.
- **`MAX_DEAD_HANDLE_RECOVERIES = 3`.** A browser that dies immediately after every relaunch is a configuration fault; relaunching forever converts a visible error into an invisible hang plus a stream of orphaned Chromes. Hitting the cap logs and reports the fault. The budget is per surface and is cleared when that surface is explicitly ended, so a re-opened surface is not born unable to heal.

### Both callers report the fault, for different reasons

`RemoteBrowserSnapshot` — the ~700ms poll — is where a dead handle is usually *discovered*, so it now reports the fault instead of only degrading around it.

`ExecuteRemoteBrowserAction` is where it is *felt*, and it reports too. Wiring only the poll would have left the exact case in the issue unhealed: a surface nobody is watching has no poll running, so an agent-driven browser would stay poisoned for the rest of the session while every action threw. The action path re-runs the action against the replacement, which is safe **precisely because the handle was dead** — the action never reached a browser, so it cannot run twice. A `navigate` therefore heals in place. A click or a type legitimately fails on the replacement's blank page and is told so in words the agent can act on ("the browser had closed and was replaced (it is now on a blank page)"), otherwise it re-clicks forever at a page it believes it loaded.

Both callers hand over the error on the same terms and neither decides what "dead" means.

## Tests

**73 pass** in `@memberjunction/remote-browser-server` (was 55) and **37** in `@memberjunction/server`: a new `dead-handle.test.ts` for the predicate in both directions, plus eight engine tests covering replacement, per-instance isolation, coalescing, the budget cap and its reset, the no-mapping case, screencast re-attachment, and the deliberately-stopped case.

| mutant | result |
|---|---|
| drop the `net::` page-level guard | 1 fail |
| predicate always returns `true` | 4 fail |
| skip screencast re-attachment | 1 fail |
| remove the recovery budget cap | 2 fail |
| check the map before the in-flight recovery | 1 fail |
| stop forgetting the sink on `StopScreencast` | 1 fail |
| drop `ProviderName` from the action-path handoff | 1 fail |
| drop the "was replaced" wording | 1 fail |
| recover on the action path without re-running | 3 fail |

Worth flagging: the `net::` mutant **survived the first round**, and that was informative rather than cosmetic. My comment claimed `net::ERR_CONNECTION_CLOSED` contains the phrase `connection closed` — it does not, Chrome uses underscores. The guard is *forward* protection for a list that is meant to grow (the next phrase someone adds is plausibly `connection refused`, and the bug that introduces is silent and destructive). The comment now says that, and a test pins the ordering so the guard cannot be quietly removed.

`MJServer` typechecks with zero errors in the touched file.

---


rkihm-BC (approved):
Approving. Reviewed the full diff against a freshly-updated `origin/next` rather than a stale checkout, and traced the recovery path end to end. No required changes — one observation at the end that is explicitly not one.

## What's good

**The diagnosis is exact and the honest detail is the telling one.** 232 `Browser not launched` errors in a single MJAPI run, against surfaces that had launched minutes earlier with no intervening `Session … ended`. And `RemoteBrowserSnapshot` already carried a comment naming this precise state — *"a stale/dead handle still in the live map, a mid-poll teardown race, or a recycled backend container"* — and only degraded around it. The bug wasn't unanticipated; it was anticipated and never healed. Saying so rather than presenting it as a new discovery is worth more than it costs.

**The predicate's asymmetry is the right call, and it's argued rather than asserted.** Too narrow costs today's behaviour for a missed phrasing; too wide costs a *healthy* browser its cookies, login and scroll position because a click missed. A closed allowlist with everything unrecognised treated as a live browser's real answer is the correct default, and the table makes the trade legible to the next person who wants to add a phrase.

**Neither caller decides what "dead" means.** Both hand over the error they already caught and let `IsDeadBrowserHandleError` rule. One grammar in one place — the alternative is two resolvers pattern-matching and drifting apart.

**The `net::` guard is the best thing in this PR.** The mutant surviving the first round could have been read as "the guard is redundant, delete it." Instead you worked out *why*: `net::ERR_CONNECTION_CLOSED` doesn't contain `connection closed` because Chrome uses underscores. The guard protects a list that is **meant to grow** — the next phrase someone adds is plausibly `connection refused`, and the bug that edit introduces is silent and destructive: a user's logged-in browser torn down because they navigated to a site that was down. Keeping a guard no current input exercises, for a documented future reason, with a test pinning the ordering, is a judgment most reviews would talk someone out of. Correcting your own earlier claim about the phrase in the same breath is the right way to leave that record.

**The recovery ordering is correct and non-obvious.** Joining an in-flight recovery *before* checking the map matters because recovery deletes the dead mapping as its first act — a caller checking the map first would see nothing mapped, conclude there was nothing to heal, and return null while a good replacement was seconds away. The comment says exactly that, so it survives the next refactor.

**Screencast re-attachment is treated as half the fix, which it is.** The client asks once at bind time and has no reason to re-ask, so healing the backend alone leaves it watching a discarded session while the agent narrates a page the person cannot see — strictly worse than the original freeze. The inverse is handled too: a deliberately-stopped screencast forgets its sink *before* stopping, so recovery cannot resurrect a surface the host released.

I checked two orderings that would fail silently if reversed, and both are right: the sink is captured into a local **before** `EndSession` deletes it from the map, and `PipeScreencastToTrack` re-registers the sink under the replacement's session id, so a second death within the budget still re-attaches.

**Bounded in both directions.** Coalescing stops the ~700ms poll and a concurrent agent action turning one dead browser into two live ones. `MAX_DEAD_HANDLE_RECOVERIES = 3` stops a configuration fault becoming an invisible hang plus a stream of orphaned Chromes, and clearing the budget on explicit end means a re-opened surface isn't born unable to heal.

**Action-path retry safety is correctly reasoned.** Re-running is safe *because* the handle was dead — the action never reached a browser, so it cannot run twice. A `navigate` heals in place; a click truthfully fails on the replacement's blank page and is told so in words the agent can act on, rather than re-clicking forever at a page it believes it loaded.

## One observation — explicitly not a requested change

The two recovery call sites don't pass the same options. The action path sends `ProviderName: providerName`; the snapshot path sends only `InstanceKey` and `ContextUser`, because `RemoteBrowserSnapshot` uses `GetSessionForAgentSession` (a lookup) and never resolves a provider name of its own.

I chased what that costs and the answer is: **nothing today, and it degrades gracefully if that ever changes.**

- On the shipped seed only `Self-Hosted Chrome` is `Active` (the other four are `Disabled`), so `resolveAgentSessionProviderName(undefined)` finds exactly one Active provider and returns it — the same provider the action path would have named.
- If a second provider is ever activated, it throws with an actionable message, `relaunchDeadAgentSession` catches it, logs, and returns null. The dead mapping was already deleted by then — which is the core fix — so the next action call lazy-starts a healthy session with its own provider name. The surface still heals; it loses only the automatic screencast re-attach on that one path.
- It is never worse than today, where the snapshot path doesn't heal at all.

So this is worth knowing rather than worth fixing. If a multi-provider deployment ever becomes real, the tidier shape is probably for the engine to remember the provider it started each agent session with and reuse it on relaunch — that removes the parameter from the caller contract entirely and makes the replacement match the original by construction rather than by every call site remembering.

## On the tests

73 in `remote-browser-server` (up from 55) and 37 in `server`, covering the predicate in both directions, replacement, per-instance isolation, coalescing, the budget cap and its reset, the no-mapping case, re-attachment, and the deliberately-stopped case. The mutation table is real work — nine mutants, all killed, including three that only die because the action path re-runs the action.

## Merge state

`mergeable=MERGEABLE`, nothing failing, but `mergeStateStatus: BEHIND` — needs a `next` merge before it can land.

PR #3917, opened 2026-08-17
feat(realtime): let a deployment choose which channels tab up front (#3599)

### LINKED ISSUE #3599: Realtime surface tabs: 'tab up front' is hardcoded to the whiteboard — it should be the deployment's choice

MS-BC:
Fixes #3599.

## The defect

```ts
export function ShouldRegisterChannelTabUpFront(channelName: string, hasBeenUsed: boolean): boolean {
  return IsWhiteboardChannel(channelName) || hasBeenUsed;
}
```

The whiteboard — and only the whiteboard, matched by name — gets a surface tab at session start before anything has touched it. The stated reason is a good one: a user can be the first to draw on a board, whereas every other channel earns its tab on first use. But it is **one deployment's answer compiled in**, and the predicate's own comment gives the tell — detection is by name *"so a deployment can't accidentally opt a non-board channel into the immediate-tab behavior"*, which also means a deployment cannot deliberately opt **out**.

In a voice-first interview the session opens on the agent talking and nobody is going to draw first. A blank board sitting there makes the product read as a tool with a canvas bolted on rather than a conversation that can produce one. A design-review session wants the opposite — and both are legitimate, which is the argument for configuration rather than a different hardcoded answer. (Locally this was patched to `return hasBeenUsed`, a reversal rather than a fix, which is exactly why it was filed as configuration.)

## The fix

`<mj-realtime-session-overlay>` takes a new `[TabUpFrontChannels]`:

| bound value | behaviour |
|---|---|
| `null` (default) | MJ's answer, unchanged: the whiteboard tabs up front, nothing else does |
| `['Remote Browser']` | that channel tabs up front — and the whiteboard does **not** |
| `[]` | no surface until something uses one — the voice-first case |

**The array replaces the default rather than adding to it.** A deployment that names one channel is stating the complete list, not appending to a hidden one; otherwise `[]` could not express "nothing", which is the case that motivated the issue.

**A channel already in use always gets its tab**, whatever is configured. A surface the agent is driving with no tab on screen is a hidden surface, not a decluttered strip — so `hasBeenUsed` short-circuits ahead of the configuration.

Names are compared through a shared `NormalizeChannelName`, the same trim-and-lowercase rule channel names already use, so `'  remote browser  '` in a deployment's config matches the plugin's `'Remote Browser'`.

### The second half the issue flags

The issue warns that *"anything tracking surface presence separately must agree with the strip"*. Grepping for that turned up one real instance: `ShouldRemoveReviewWhiteboardTab` carried its own inline `c.ChannelName?.trim().toLowerCase() === 'whiteboard'` rather than calling `IsWhiteboardChannel`. Two copies of one rule, in a change that alters when whiteboard tabs appear, is how they drift — so it now shares the predicate.

## Tests

`realtime-surface-tab-style.test.ts` grows from 2 to 8 cases on this predicate: the default is unchanged for `null`/`undefined`, `[]` yields no up-front surface, a non-whiteboard channel can be named, the list replaces rather than extends the default, names normalise, `hasBeenUsed` overrides everything, and a blank name never matches a blank entry.

**1086 tests pass** across `@memberjunction/ng-conversations` (92 files).

Mutation-verified — five mutants, all killed:

| mutant | result |
|---|---|
| ignore the configuration entirely | 3 fail |
| let the configuration ADD to the default instead of replacing it | 3 fail |
| drop the `hasBeenUsed` override | 2 fail |
| drop name normalization | 2 fail |
| allow a blank name to match a blank entry | 1 fail |

---

### DIFF
--- a/.changeset/realtime-tab-up-front-configurable.md
+++ b/.changeset/realtime-tab-up-front-configurable.md
@@ -1,0 +1,28 @@
+---
+"@memberjunction/ng-conversations": minor
+---
+
+Realtime surface tabs: which channels tab up front is now the deployment's choice (#3599)
+
+`ShouldRegisterChannelTabUpFront` returned `IsWhiteboardChannel(channelName) || hasBeenUsed`, so the
+whiteboard — and only the whiteboard, matched by name — got a surface tab at session start before
+anything had touched it. The reasoning was sound (a user can be the first to draw on a board, whereas
+every other channel earns its tab on first use) but it was one deployment's answer compiled in. The
+predicate's own comment said detection was by name "so a deployment can't accidentally opt a
+non-board channel into the immediate-tab behavior" — which also meant a deployment could not
+deliberately opt OUT.
+
+`<mj-realtime-session-overlay>` takes a new `[TabUpFrontChannels]`:
+
+- **`null`** (the default) keeps MJ's answer exactly, so no existing deployment changes.
+- **an array** is the complete list of channels that tab up front, REPLACING the default rather than
+  adding to it. `[]` therefore means "no surface until something uses one" — the voice-first case,
+  where a session opens on the agent talking and a blank board makes the product read as a tool with
+  a canvas bolted on rather than as a conversation that can produce one.
+
+A channel the agent has already used always gets its tab, whatever is configured: a surface being
+driven with no tab on screen is a hidden surface, not a decluttered strip.
+
+Also removes an accidental duplicate of the whiteboard-name rule — `ShouldRemoveReviewWhiteboardTab`
+had its own inline `trim().toLowerCase() === 'whiteboard'` and now shares `IsWhiteboardChannel`, so
+the two cannot drift.
--- a/packages/Angular/Generic/conversations/node_modules
+++ b/packages/Angular/Generic/conversations/node_modules
@@ -1,0 +1,1 @@
+/Users/madhav/Projects/MJ/packages/Angular/Generic/conversations/node_modules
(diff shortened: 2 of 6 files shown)


MS-BC:
Closing: the requesting consumer (Caliber) now decides surface prominence in its own workspace layer (grouping observer + split-layout) and no longer reads this knob. Small enough to redo against current code if a deployment-level tab-order need returns.

PR #3918, opened 2026-08-17
feat(realtime): close ONE live channel without ending the call (#3498)

### LINKED ISSUE #3498: Realtime surface tabs: no way to close a single live channel — only ending the call removes it
Observed on MJ 5.51.0, client-direct realtime session with two channels active (Whiteboard + Remote Browser).

## What happens

Once two channel surfaces are open there is no way to close just one. The user who is done with the browser but still wants the whiteboard has only one option: end the call — which tears down both, plus the voice session.

## Why

`RealtimeSurfaceTabsComponent.RemoveTab(key)` exists and does the right thing (Activity is irremovable, focus falls back). But its only callers are review-mode cleanup in `RealtimeSessionOverlayComponent` — `cleanupStaleReviewBoardTab()` and `exitReview()`. No live channel tab renders a close affordance, and there is no corresponding teardown on the session side: `RealtimeSessionService` exposes no `CloseChannel` / `StopChannel` / `RemoveChannel`, so even calling `RemoveTab` on a live channel would hide the surface while leaving the plugin initialized, its tools advertised to the model, and (for Remote Browser) a server-side headless Chrome still running.

## Impact

Two costs. The obvious one is UX. The less obvious one is cost and correctness: a Remote Browser channel nobody wants keeps a browser alive server-side, and its `browser_*` tools stay in the model's vocabulary — so the agent can still act on a surface the user believes they dismissed.

This gets sharper as channel counts grow; the channel registry is deployment-wide (`MJ: AI Agent Channels.IsActive`, no per-agent scoping), so every session opens every active channel whether or not that session has any use for it.

## Sketch of a fix

A per-channel close on the session service that disposes the plugin, withdraws its tools, releases its server-side resources, and removes the tab — with `RemoveTab` becoming a consequence of that rather than the whole action. A closable flag per tab would let a host mark channels the user may dismiss (Activity stays irremovable).

Closing a channel is a mutation of the same live roster that #3497 says the agent cannot read. One concept, two symptoms.

Filing as a report only, per our process — no PR attached.

MS-BC:
Fixes #3498.

## The defect

Once two channel surfaces are open there is no way to close just one. The user who is done with the browser but still wants the whiteboard has a single option: **end the call** — which tears down both surfaces plus the voice session.

`RealtimeSurfaceTabsComponent.RemoveTab(key)` already existed and did the right thing (Activity is irremovable, focus falls back). But its only callers were review-mode cleanup, no live tab rendered a close affordance, and — the important half — **hiding a tab is not closing a channel**. Even calling `RemoveTab` on a live channel would leave the plugin initialized, its tools advertised to the model, and (for Remote Browser) a server-side headless Chrome running. The agent could still act on a surface the user believed they had dismissed.

## The fix

`RealtimeSessionService.CloseChannel(channelName)` is the real action, and tab removal is a **consequence** of it (via a new `ChannelClosed$`) rather than the whole action. In order, closing:

1. **flushes the channel's debounced state save** — a board closed three seconds after the last stroke must not lose those strokes;
2. **points its tool prefix at a refusal** (see below);
3. **disposes the plugin**, which is where the Remote Browser channel stops its screencast and audio stream;
4. **drops it from the live set**, pushes a context note so the model knows the surface is gone, and announces the close.

A plugin that throws on the way out does not strand the close — the tab, the tool route and the live set all still change, or the user's dismissal silently does nothing.

### Why the tool prefix is re-registered, not unregistered

This is the part worth reviewing carefully. The model was handed its tool vocabulary at connect and there is no mid-session way to withdraw part of it, so `browser_*` stays **callable** no matter what happens here. An *unregistered* prefix does not become inert:

```ts
const clientHandler = this.findClientToolHandler(call.ToolName);
if (clientHandler) { … }        // local
// otherwise → the MJ SERVER relay
```

So unregistering would send the agent's browser tools to the server resolver, which would drive **the very browser the user dismissed**, now with no surface showing it — strictly worse than the original bug. A refusal handler keeps the call local, cannot reach the server, and gives the model an answer it can narrate ("that surface was closed"). Tombstones are cleared at session teardown, since the next session mints a fresh vocabulary.

### The affordance

Tabs carry a `Closable` flag: channel tabs default to closable, the Activity tab never is (it is the panel's fallback focus and owns nothing to release). A host sets `Closable: false` for a channel the session depends on — closing withdraws a channel's tools for the rest of the session and is deliberately not undoable.

The close control is a **sibling** of the tab button, not nested inside it — a button cannot contain a button. It is absolutely positioned over the tab's reserved trailing padding, so revealing it on hover never reflows the strip, and it stays in the a11y tree (opacity, not `display: none`) so it is keyboard-reachable.

`OnCloseTabRequested` falls back to a plain `RemoveTab` when the key matches no live channel — a review-mode board tab or a stale placeholder has no plugin behind it, and a close control that visibly does nothing is worse than removing a tab that owns no resources.

## Scope note

**Releasing the server-side browser on a per-channel close is not covered here.** The server-side `RemoteBrowserChannel` ends the browser on *session* close (`OnSessionClosed` / `Dispose`), and there is no mutation meaning "this one channel is going away". That needs a server change; it is outside this PR's blast radius and I will file it separately rather than bolt a resolver onto a UI change.

## Tests

**1087 pass** (was 1080): 3 new model cases for `Closable` (default true for channels, never for Activity, honoured through a re-registration upgrade) and a new 4-case DOM spec for the affordance — the control renders for a closable channel and not for an un-closable one, pressing it asks the host and does **not** remove the tab itself, and dismissing does not focus the surface being dismissed.

Mutation-verified:

| mutant | result |
|---|---|
| `Closable` ignores the host's `false` | 3 fail |
| the component removes the tab itself instead of asking the host | 1 fail |
| nest the close button inside the tab button | 1 fail |

The third one replaced a mutant that **survived**: I had written `event.stopPropagation()` with a comment claiming the control sits inside the tab's hit area. It does not — it is a sibling, so the click never reaches the tab's handler. The guard was dead code justified by a false comment, so it is gone, and the test now pins the structural choice that actually provides the guarantee.

---

### DIFF
--- a/.changeset/realtime-close-one-channel.md
+++ b/.changeset/realtime-close-one-channel.md
@@ -1,0 +1,36 @@
+---
+"@memberjunction/ng-conversations": minor
+---
+
+Realtime: close ONE live channel without ending the call (#3498)
+
+Once two channel surfaces were open there was no way to close just one. A user done with the browser
+but still wanting the whiteboard had a single option: end the call — which tore down both surfaces
+plus the voice session.
+
+`RealtimeSurfaceTabsComponent.RemoveTab` already existed and did the right thing, but its only
+callers were review-mode cleanup, and **hiding a tab is not closing a channel**: the plugin stayed
+initialized, its tools stayed advertised to the model, and (for Remote Browser) a server-side headless
+Chrome kept running. So the agent could still act on a surface the user believed they had dismissed.
+
+`RealtimeSessionService.CloseChannel(channelName)` is the real action. It flushes the channel's
+debounced state save (a board closed three seconds after the last stroke does not lose those
+strokes), points its tool prefix at a refusal, disposes the plugin — which is where the Remote Browser
+channel stops its screencast and audio stream — drops it from the live set, tells the model the
+surface is gone, and announces on the new `ChannelClosed$`. Tab removal is a **consequence** of that,
+driven by the observable.
+
+The tool prefix is re-registered with a refusal rather than unregistered, and that distinction
+matters: the model was handed its vocabulary at connect and there is no mid-session way to withdraw
+part of it, so `browser_*` stays callable regardless. An *unregistered* prefix is not inert — unrouted
+tools fall through to the **server relay**, so the agent would keep driving the very browser the user
+dismissed, now with no surface showing it. A refusal keeps the call local and gives the model an
+answer it can narrate.
+
+Tabs carry a `Closable` flag (channel tabs default to closable; Activity never is, being the panel's
+fallback focus with nothing to release). A host can set `Closable: false` for a channel the session
+depends on — closing withdraws a channel's tools for the rest of the session and is not undoable.
+
+**Not covered here:** releasing the server-side browser on a per-channel close. The server-side
+`RemoteBrowserChannel` ends it on SESSION close, and there is no mutation for "this one channel is
+going away" — that needs a server change and is tracked separately.
--- a/packages/Angular/Generic/conversations/node_modules
+++ b/packages/Angular/Generic/conversations/node_modules
@@ -1,0 +1,1 @@
+/Users/madhav/Projects/MJ/packages/Angular/Generic/conversations/node_modules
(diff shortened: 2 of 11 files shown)


MS-BC:
Closing: the requesting consumer (Caliber) runs one agent-staged surface at a time and deliberately strips chrome from the candidate screen, so nothing consumes a per-tab close. Of the retired overlay-UX set this is the piece with the clearest future audience — a user-driven workspace like Explorer chat dismissing one surface — so if that need lands, start from this branch.

PR #3920, opened 2026-08-17
feat(realtime): side-by-side surfaces via [SplitKeys], not a stylesheet fight (#3535)

### LINKED ISSUE #3535: Realtime surface panel: no layout mode for hosts (side-by-side surfaces), and overriding it silently loses a specificity tie

MS-BC:
Fixes #3535.

## The defect

`RealtimeSurfaceTabsComponent` shows exactly one pane at a time, and there is no way for a host to say "show these two together". Arranging them from outside runs into a trap that fails **silently**:

- MJ emits `.surface { display: flex }` with Angular's attribute scoping → specificity **(0,2,0)**
- a host writing `.surface.my-split { display: grid; grid-template-columns: 1fr 1fr }` is **also (0,2,0)**
- ties break on document order, and component styles are injected when the component first renders — *after* a host stylesheet added at startup

So MJ wins, and nothing reports it: `grid-template-columns` computes correctly and is simply ignored on a flex container. Everything looks applied; the panes just stack. The issue records two debugging sessions lost to this — once with a CSS grid, once with Golden Layout, where the host element ended up 0px tall and GL laid its entire tree into nothing while reporting no error. The workaround, doubling the class (`.surface.my-split.my-split`), works — and nobody should have to know it.

## The fix

```html
<mj-realtime-surface-tabs [SplitKeys]="['Left Browser', 'Right Browser']">
```

`[SplitKeys]` names the tab keys to show side by side, in the order the panes should read. `null` or `[]` is the ordinary tabbed panel — the default, and what every existing host gets.

### One input, not a mode plus a list

The issue suggested `Layout: 'tabs' | 'split'` **plus** the key set. I went with the single input, because two make **"split with no keys"** expressible and that state has no good answer — an empty panel or a silent downgrade to tabs, both surprises. "Split these" is the entire instruction, so the invalid combination does not exist. `Layout` survives as a read-only `'tabs' | 'split'` derived from it, reporting `'tabs'` when fewer than two named surfaces are actually present — so a host can name its arrangement up front and have channels arrive as the session gets going.

### Three decisions inside the mode

**Order via CSS `order`, not by reordering the DOM.** Panes render in registration order; a host asking for the reverse gets it visually while the elements stay put. Moving them would tear down and recreate the dynamically-created channel surfaces inside — a live browser canvas would go blank because someone wanted it on the left.

**A focused tab outside the arrangement stays hidden.** Otherwise selecting a third tab would silently drop an extra pane into a two-column grid.

**The focused pane keeps an accent inside the split.** The tab strip still drives focus, and an unmarked grid makes "which one am I acting on" unanswerable.

The panes needed a wrapper (`.s-panes`) to become the grid: the panel itself is a flex **column** carrying the tab strip, so panes cannot be laid out in a row at that level. That wrapper is the piece a host could never add from outside.

### The secondary ask

Every pane now carries `data-channel` and `data-tab-kind`. Before, panes were bare `.s-pane` elements and a host had to pair them with tabs **by index** — which worked only because both lists render from the same array in the same order, an implementation detail nobody should be resting on. `data-layout` on the container is the matching host-facing hook for the mode itself, bound from the same expression as the internal `surface--split` class so the two cannot disagree.

## Tests

A new 7-case DOM spec: tabbed by default, the named surfaces render side by side with the column count matching, host order is honoured without moving elements, absent keys are ignored (and one present surface is not a "split"), a focused non-member pane stays hidden, every pane carries its identity, and `null`/`[]` both mean tabs.

**1087 tests pass** (was 1080).

| mutant | result |
|---|---|
| split on a single present key | 1 fail |
| stop filtering keys for surfaces that do not exist | 1 fail |
| drop pane ordering (registration order wins) | 1 fail |
| drop `data-channel` | 3 fail |
| drop the `surface--split` container class | 1 fail |
| drop `data-layout` | 1 fail |

The `surface--split` mutant **survived the first round**: my test asserted `data-layout` and the pane classes but not the container class the stylesheet actually keys the grid off, so the grid could have stopped applying with every assertion still green. Both are now asserted.

---

### DIFF
--- a/.changeset/realtime-surface-split-layout.md
+++ b/.changeset/realtime-surface-split-layout.md
@@ -1,0 +1,33 @@
+---
+"@memberjunction/ng-conversations": minor
+---
+
+Realtime surface panel: show surfaces side by side, through the component's own API (#3535)
+
+`RealtimeSurfaceTabsComponent` showed exactly one pane at a time, and arranging two from outside
+could not be done honestly. MJ emits `.surface { display: flex }` at specificity (0,2,0) with
+Angular's attribute scoping; a host writing `.surface.my-split { display: grid }` is **also** (0,2,0);
+ties break on document order, and component styles are injected when the component first renders —
+after a host stylesheet added at startup. So MJ won, and the failure was invisible:
+`grid-template-columns` computed correctly and was silently ignored on a flex container. The
+workaround was doubling the class (`.surface.my-split.my-split`), which works and which nobody should
+have to know.
+
+`[SplitKeys]` names the tab keys to show side by side, in the order the panes should read.
+`null`/`[]` is the ordinary tabbed panel, which is the default and what every existing host gets.
+
+**One input rather than a `Layout` mode plus a key list**, deliberately: two inputs make "split with
+no keys" expressible, and that state has no good answer — an empty panel or a silent downgrade, both
+surprises. "Split these" is the whole instruction, so the invalid combination does not exist.
+`Layout` is a read-only `'tabs' | 'split'` derived from it, and it reports `'tabs'` when fewer than
+two named surfaces are actually present, so naming a channel before it exists is safe.
+
+Host order is honoured through CSS `order`, not by reordering the DOM — moving pane elements would
+tear down and recreate the dynamically-created channel surfaces inside them, blanking a live browser
+canvas because someone wanted it on the left. A focused tab outside the arrangement stays hidden, so
+choosing a third surface cannot drop an extra pane into a two-column grid, and the focused pane keeps
+an accent so "which one am I acting on" stays answerable.
+
+Also addresses the issue's secondary ask: every pane now carries `data-channel` and `data-tab-kind`,
+so a host pairs a pane with its channel by identity instead of by index — which only ever worked
+because both lists render from the same array in the same order.
--- a/packages/Angular/Generic/conversations/node_modules
+++ b/packages/Angular/Generic/conversations/node_modules
@@ -1,0 +1,1 @@
+/Users/madhav/Projects/MJ/packages/Angular/Generic/conversations/node_modules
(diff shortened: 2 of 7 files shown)


MS-BC:
Closing alongside #3899 — the two were duplicate implementations of #3535. This is the better of the pair ([SplitKeys], no stylesheet fight); if MJ's overlay grows a real need to tile two surfaces, resurrect THIS branch. The original driver (Caliber) now composes surfaces in its own widget layer and consumes neither.

Issue #3536, opened 2026-08-06
Realtime: no session-level client tool route — cross-surface tools must wear one channel's prefix


MS-BC:
### Summary

Client tools are routed to a channel by the declaring channel's `ToolNamePrefix`. There is no route for a tool that belongs to the **session** rather than to any one channel, so session-level tools have to be declared by whichever channel happens to be present — and then they ship wearing that channel's badge.

### What it looks like in practice

We have tools that operate across surfaces: report every open surface and its state, group two surfaces into one workspace, hand a multi-step job to a delegate that drives several of them. None of these belong to the whiteboard or to a browser. But to be reachable at all they must be declared by one, so the model sees:

```
browser_WorkspaceStatus      ← reports ALL surfaces, not just the browser
browser_CombineSurfaces      ← groups the whiteboard with something
browser_SetUpSurfaces        ← drives every surface
```

Every description then has to open by contradicting its own name ("Reports ALL open surfaces at once — not only the one this tool is named after"). That is prompt budget spent undoing a routing decision, and it degrades tool selection: a model reasonably infers scope from a name.

It also creates a real ordering hazard — the owning channel must be present, so we have a claim/ownership dance to pick whichever surface exists, and the tool names change depending on which channel won.

### A second, related limit

A synchronous channel's `ApplyAgentTool` returns `string`. A session-level tool that must *await* something (our delegate returns a consolidated result after several steps) can only be declared by a channel whose entry point is async — so the same tool is available or not depending on which surface claimed it. We gate the declaration on that today, which works but is not something a host should have to reason about.

### Ask

A session-level tool route: register tools against the session with a neutral prefix (or none), dispatched by the session service rather than by channel prefix, and allowed to be async.

Found on 5.51.0.

MS-BC:
Implementation status from bizapps-caliber: the consuming feature is complete and CI-green as bizapps-caliber#157 (draft), running against local node_modules patches (`MJ/patches/restore-mj-patches.sh` in that repo's workspace). **No upstream PR exists for this issue yet** — the local patches are dist-level, so the source-level MJ PR still has to be written. Caliber cannot merge #157 until it lands here.

Issue #3496, opened 2026-08-05
Remote Browser channel: page changes the agent did not cause are never surfaced to it (human takeover ⇒ confident stale narration)


MS-BC:
Observed on MJ 5.51.0, client-direct realtime session with the Remote Browser channel active and human takeover enabled (the `Self-Hosted Chrome` provider defaults to `Collaborative`).

## What happens

A user takes over the browser surface and navigates. The agent is then asked "what do I have open right now?" and confidently describes the **previous** page — the one it last navigated to itself. It only corrects after being explicitly told to look again.

Real exchange:

> **Agent:** I've got GeeksforGeeks open. It's on their main learning homepage…
> **User:** What do I have open right now?
> **Agent:** You're on the GeeksforGeeks homepage right now…
> **User:** Uh, no, I'm not. Can you look again?
> **Agent:** You're right, I stand corrected — you're on the GeeksforGeeks "C Programs" page…

## Why

`RemoteBrowserChannel` pushes its `[browser] current page: <url>` context note in exactly two places, both immediately after a server action **the model itself initiated**:

- after `ExecuteRemoteBrowserAction` (`remote-browser-channel.ts`, the `applyServerAction` path)
- after `ExecuteRemoteBrowserGoal`

Nothing observes page changes from any other origin. Human-relayed input (`RELAY_HUMAN_INPUT_MUTATION`) drives the page without producing a note, and pushed screencast frames carry only image bytes — `OnScreencastFrame(dataBase64)` has no URL — so under the streaming path the client never learns the URL moved either.

The effective rule is: **a surface change the agent did not cause is invisible to it.** This is not stale caching; the agent was never told. It *can* re-read on demand (which is why "look again" works), but nothing prompts it to, and mid-conversation it does not guess that it should.

## Impact

Human takeover is on by default for `Collaborative` providers, so the default configuration is the broken one. The failure mode is confident misdescription rather than a visible error, which is the worst shape for an agent whose value is narrating a shared surface.

## Sketch of a fix

Detect page changes regardless of origin and emit the same `[browser]` note — e.g. compare `CurrentUrl` on a lightweight poll (the existing `RemoteBrowserSnapshot` query already returns it, and a URL-only selection is cheap), or have the server include the URL on screencast frames / emit a page-change push event.

Filing as a report only, per our process — no PR attached. Happy to contribute one if the shape above is the direction you'd want.

MS-BC:
## Reference implementation (running locally in Caliber)

[...]

PR #3924, opened 2026-08-17
feat(realtime): give the agent a live roster of its own surfaces (#3497)

### LINKED ISSUE #3497: Realtime: the agent has no live channel roster — it cannot know which surfaces are open or what they hold
Observed on MJ 5.51.0, client-direct realtime session with two channels active (Whiteboard + Remote Browser).

## What happens

The agent cannot answer questions about its own surfaces: how many are open, which ones, or what each currently holds. It infers capability from the tools it was given ("I have `browser_*` tools, so there must be a browser") but has no knowledge of the live surface set, so it guesses — and guesses confidently.

## Why

At connect the model is told its **tool vocabulary**. It is never told the **channel roster**: which channels actually opened, which are focused, which failed to resolve a plugin, and what state each holds. `BaseRealtimeChannelClient` has `SerializeState()`, so per-channel state exists and is already persisted for resume — but nothing composes it into a description the model can read, and nothing tells the model when the set changes.

This is the same root cause as the Remote Browser staleness (filed separately as #3496): state exists on the client, and the model is only ever told about the parts it caused itself.

## Impact

It gets worse as channels multiply, which is the direction the channel system is clearly built to go. With one surface the agent can assume; with three it cannot, and "let me check the board" versus "let me check the browser" becomes a coin flip. Anything that wants to coordinate across surfaces has to route through a model that does not know what the surfaces are.

## Sketch of a fix

An authoritative live roster the host owns — the set of open channels with, per channel, its name, tab title, focus state, and a short channel-supplied state summary (an optional `DescribeState(): string` alongside the existing `SerializeState()`). Surfaced to the model at connect and refreshed on change, either as a context note or as a small read tool.

Worth noting that the same roster is what a per-channel close would mutate (filed separately) — one concept, two symptoms.

Filing as a report only, per our process — no PR attached.

MS-BC:
Fixes #3497.

## The defect

The agent cannot answer questions about its own surfaces: how many are open, which ones, or what each currently holds. At connect the model is told its **tool vocabulary** and nothing else. It is never told the **channel roster** — which channels actually opened, which failed to resolve a plugin, what state each holds. So it infers capability from the tools it was given ("I have `browser_*` tools, so there must be a browser") and answers from that inference, confidently.

The state existed all along. Every plugin has `SerializeState()`, and it is already persisted for resume. Nothing composed it into something a model could read, and nothing told the model when the set changed. Same root cause as #3496: state lives on the client, and the model is only ever told about the parts it caused itself.

It gets worse as channels multiply, which is the direction the channel system is built to go — with one surface the agent can assume; with three, "let me check the board" versus "let me check the browser" is a coin flip.

## The fix

`RealtimeSessionService` gains `ChannelRoster` (structured) and `DescribeChannelRoster()` (one line for the model), composed from the **live plugins**: name, tab title, whether the channel has a visible surface, which one owns the screen, and a short channel-supplied summary.

```
[surfaces] 2 surfaces open: Whiteboard (12 items) — focused; Remote Browser (showing https://careers.acme.com)
```

Pushed at connect and whenever that line changes.

### Pushed, not offered as a read tool

The issue allows either. Pushing is the right one here, because **the failure is a model that does not know it should ask** — it already believes it knows what surfaces it has. A read tool it never calls leaves the guess exactly where it was. The public getter exists so a host (or a session-level tool once #3536 lands) can take the same read on demand.

Refreshed on **membership and focus** changes only, not on every stroke or navigation: per-channel contents move constantly and a note per change would drown the conversation. What the model could not work out at all is which surfaces *exist*. The composed line is deduped, so a refresh that did not change the answer costs nothing.

### `DescribeState()` — and why it is not `SerializeState`

New optional `BaseRealtimeChannelClient.DescribeState(): string | null`, defaulting to `null`.

`SerializeState` is for **resume**: a machine format, potentially large, written for a future instance of the same channel. `DescribeState` is written for a language model that is about to speak — `showing https://careers.acme.com`, `12 items across 2 pages`. A channel that returned its serialized state here would spend the session's context on syntax.

**Both shipped channels implement it**, so the hook is not inert on arrival: the whiteboard reports what is on the board, the Remote Browser reports which page it is showing — the part the agent could not infer from merely having the tools. A channel that throws while describing itself is omitted (logged), because one broken description must not put every *other* surface back into the dark.

### Three details worth reviewing

- The roster is composed from **resolved plugins**, not registry rows. A row whose `ClientPluginClass` is missing is precisely the surface the agent would otherwise believe it has.
- A **surface-less** channel (server-only, `HasSurface() === false`) is listed and marked rather than omitted — the agent has its tools, so hiding it recreates the guessing problem one level down.
- An **empty** session says "none open" explicitly. Silence is what the model had before, and silence is what it filled in with a guess.

## Tests

**1096 pass** (was 1080). A new 9-case spec for the roster wording (framework-free, like the surface-tab helpers next door) and 7 new service-level cases driven against the real `RealtimeSessionService` with a fake registered channel: composition from live plugins, an unresolved plugin left out, a throwing `DescribeState` contained, announce-once/stay-quiet, re-announce on a real change, focus announced, and the empty session.

| mutant | result |
|---|---|
| an empty roster says nothing | 2 fail |
| drop each channel's summary | 2 fail |
| surface-less channels look like normal ones | 2 fail |
| no dedupe — announce on every refresh | 1 fail |
| focus is never tracked | 1 fail |
| a throwing `DescribeState` takes down the roster | 1 fail |

## Related

#3498 closes a channel, which mutates this same roster, and #3536 is the missing route that would let the agent *read* it as a session-level tool. One concept, three symptoms — each in its own PR.

---

### DIFF
--- a/.changeset/realtime-live-channel-roster.md
+++ b/.changeset/realtime-live-channel-roster.md
@@ -1,0 +1,32 @@
+---
+"@memberjunction/ng-conversations": minor
+---
+
+Realtime: the agent can finally see its own surfaces (#3497)
+
+The agent could not answer questions about its own channels — how many are open, which ones, or what
+each holds. At connect the model is told its **tool vocabulary** and nothing else, so it inferred
+capability from the tools it was given ("I have `browser_*`, so there must be a browser") and then
+answered from that inference, confidently. With one surface the guess is usually right; with three,
+"let me check the board" versus "let me check the browser" is a coin flip.
+
+The state existed all along — every plugin has `SerializeState()`, already persisted for resume.
+Nothing composed it into something a model could read, and nothing told the model when the set
+changed. Same root cause as #3496: state lives on the client, and the model is only ever told about
+the parts it caused itself.
+
+`RealtimeSessionService.ChannelRoster` / `DescribeChannelRoster()` compose the live set — name, tab
+title, whether the channel has a visible surface, which one owns the screen, and a short
+channel-supplied summary. The roster is pushed to the model at connect and whenever that answer
+changes, deduped on the composed line.
+
+**Pushed rather than offered as a read tool, deliberately.** The failure being fixed is a model that
+does not know it should ask; a tool it never calls leaves the guess in place. The public getter is
+there so a host — or a session-level tool once #3536 lands — can take the same read on demand.
+
+New optional `BaseRealtimeChannelClient.DescribeState(): string | null`. It is **not**
+`SerializeState` in prose: that payload is a machine format written for a future instance of the same
+channel, while this is written for a model about to speak. Both shipped channels implement it — the
+whiteboard reports what is on the board, the Remote Browser reports which page it is showing, which
+is the part the agent could not infer from having the tools. A channel that throws while describing
+itself is omitted rather than taking the roster down with it.
--- a/packages/Angular/Generic/conversations/node_modules
+++ b/packages/Angular/Generic/conversations/node_modules
@@ -1,0 +1,1 @@
+/Users/madhav/Projects/MJ/packages/Angular/Generic/conversations/node_modules
--- a/packages/Angular/Generic/conversations/src/__tests__/realtime-channel-roster.test.ts
+++ b/packages/Angular/Generic/conversations/src/__tests__/realtime-channel-roster.test.ts
@@ -1,0 +1,85 @@
+import { describe, it, expect } from 'vitest';
+import {
+  DescribeChannelRoster, RealtimeChannelRosterEntry,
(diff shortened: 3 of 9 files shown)


MS-BC:
Closing: the driving consumer no longer exists. Caliber (this PR's origin) now runs its own workspace ledger — cross-channel, time-ordered Summary()/Timeline() fed to the model — and consumes none of this API. The roster idea remains sound for core MJ hosts that let an agent manage several surfaces; if that need materialises, this branch is the starting point rather than a design from scratch.

PR #3925, opened 2026-08-17
feat(realtime): a client tool can belong to the session, not a channel (#3536)

### LINKED ISSUE #3536: Realtime: no session-level client tool route — cross-surface tools must wear one channel's prefix

MS-BC:
Fixes #3536.

## The defect

Client tools are routed to a channel by the declaring channel's `ToolNamePrefix`. There is no route for a tool that belongs to the **session**, so a cross-surface tool has to be declared by whichever channel happens to be present — and then ships wearing that channel's badge:

```
browser_WorkspaceStatus      ← reports ALL surfaces, not just the browser
browser_CombineSurfaces      ← groups the whiteboard with something
browser_SetUpSurfaces        ← drives every surface
```

Every description then opens by contradicting its own name ("Reports ALL open surfaces at once — not only the one this tool is named after"). That is prompt budget spent undoing a routing decision, and it degrades tool selection, because a model reasonably infers scope from a name. It also creates a real ordering hazard: the owning channel must be present, so hosts run a claim/ownership dance to pick whichever surface exists, and the tool names change depending on which channel won.

## The fix

```ts
service.RegisterSessionClientTools([{
  Definition: { Name: 'WorkspaceStatus', Description: 'Report every open surface.', ParametersSchema: { … } },
  Execute: async argsJson => JSON.stringify(await summariseWorkspace(argsJson)),
}]);
```

Declared to the model at mint alongside the channel tools, dispatched by **exact name** rather than by prefix — so they carry their own names and are reachable whether or not any particular channel is open. No claim dance, no name that changes with the winner.

### The async half

`Execute` may return a promise, which closes the issue's second, easily-missed limit. A channel's `ApplyAgentTool` returns `string`, so a session-level tool that must *await* something (a delegate returning a consolidated result after several steps) could only be declared by a channel whose entry point happened to be async — making the tool's very existence depend on which surface claimed it. Thrown errors are wrapped into `{ success: false, error }` exactly like a channel tool's, so a failure is narrated rather than silent.

### Shadowing is a hard error, not a surprise

Exact match wins dispatch. That is the right order — a session tool is named for what it *does*, so nothing guarantees its name avoids every live prefix, and if prefix won, a cross-surface tool would be silently delivered to one surface's channel, which is the exact failure this route removes.

But it means a session tool named `Echo.Status` would swallow every `Echo.*` call, and the channel would stop responding to its own tools with nothing to say why. So starting a session with such a name **throws**, naming both the tool and the channel. The clash is deterministic and belongs entirely to the host's naming, so it surfaces the first time the session is run rather than as a mystery in production.

A declaration with no name or no `Execute` is dropped with a warning rather than registered: a declaration the model can see but nothing can run is *worse* than no tool, because the call falls through to the **server relay** and the model narrates an outcome nobody intended.

## Tests

**1090 pass** (was 1080). Eleven new cases: names carry no prefix, a call routes with no channel present at all, an async tool is awaited, a thrown tool is narrated, the name matches case-insensitively, channel-prefixed tools still reach their channel, the shadow guard throws, a non-colliding name passes, unrunnable declarations are dropped, `[]` clears the set, and one that pins **which side loses** a collision.

Mutation-verified — five mutants, all killed:

| mutant | result |
|---|---|
| prefix wins over exact match | 1 fail |
| remove the shadow guard | 1 fail |
| case-sensitive name lookup | 4 fail |
| accept unrunnable declarations | 1 fail |
| registration appends instead of replacing | 1 fail |

The first **survived the initial round**, and the reason was worth fixing rather than papering over: with the guard in place, the dispatch order is only observable in a configuration the guard rejects. But registration can happen *after* mint, which skips the guard — so the ordering does have to be right on its own, and the new test pins that the session tool is the side that wins. That is exactly what the guard's error message claims; had prefix won, the message would have been backwards and a host would go chasing the wrong name.

## Related

The natural first customer is #3497's roster (`DescribeChannelRoster()` is already public for it) — that is a separate PR, and #3497 explains why the roster is *pushed* as well: a read tool only helps a model that knows to ask.

---

---

## Proven live through Caliber — 2026-08-18

Caliber's realtime workspace registers its surface tools (`SetUpSurfaces`, `CombineSurfaces`,
`WorkspaceStatus`) as **session** tools rather than channel tools — the shape this PR enables. The
dispatch path those tools ride on was exercised end to end against a live stack (MJAPI `:4111`,
a real model) using Caliber's own `smoke/widget/tool-dispatch-probe.mjs`:

```
PASS  mint an anonymous widget invite through the product path  — scope BE57D3D9-…
PASS  Prepare returns an assessment session + reconcile handle  — BF43ED09-…
PASS  the relayed tool call DISPATCHES (agent run record written)  — delegated run completed

--- target agent replied ---
Hello! Yes, I can hear you loud and clear. How can I help you today?

PASS  cleanup revoked the probe's Invite Scope

ALL CHECKS PASSED
```

Exit 0. The reply is a real model turn, not a stub — the delegated agent ran, wrote its
`MJ: AI Agent Runs` record, and answered.

### Why this probe is the meaningful one

`ExecuteRealtimeSessionTool` does **not** throw when a delegated run fails; it returns a payload with
`success: false`. A probe asserting `data.ExecuteRealtimeSessionTool != null` prints PASS for a call
that failed. This one parses the payload and asserts on `success`, and reports a GraphQL schema error
as a bug rather than a failed dispatch — so a green line here means the run genuinely completed.

### Scope of the claim — what this does and does not show

**Shown:** on a stack carrying this branch set, a scoped anonymous caller's session-level tool call
dispatches, the delegated agent run is created and completes, and the agent's reply comes back
through the relay.

**Not shown:** this probe drives the dispatch path, not the `SplitKeys` / surface-roster APIs
themselves. It does not prove two surfaces render side by side, nor that the workspace ledger records
a cross-surface `drive`. Those need a browser session with a human at a microphone, which is the one
leg of Caliber's verification that cannot be automated — stated rather than implied by a green tick.

### Correcting an earlier reading

Caliber's hub was briefly assessed as blocked on this branch set, on the grounds that
`CombineSurfaces` / `SetUpSurfaces` / `AutoCombineAs` were absent from the MJ build. They are
Caliber's own tool-name constants and registry methods, not MJ exports — the grep proved nothing. The
hub's actual MJ dependency is `RealtimeToolDefinition` + `JSONValue` from `@memberjunction/ai`, and
both resolve on this stack.

### DIFF
--- a/.changeset/realtime-session-level-client-tools.md
+++ b/.changeset/realtime-session-level-client-tools.md
@@ -1,0 +1,36 @@
+---
+"@memberjunction/ng-conversations": minor
+---
+
+Realtime: a client tool can belong to the SESSION, not to a channel (#3536)
+
+Client tools were routed to a channel by the declaring channel's `ToolNamePrefix`, so a tool that
+operates across surfaces — report every open surface, group two surfaces into a workspace, hand a
+multi-step job to a delegate that drives several — had to be declared by whichever channel happened
+to be present, and then shipped wearing that channel's badge:
+
+```
+browser_WorkspaceStatus      ← reports ALL surfaces, not just the browser
+browser_CombineSurfaces      ← groups the whiteboard with something
+```
+
+Every description then opened by contradicting its own name ("Reports ALL open surfaces — not only
+the one this tool is named after"), which is prompt budget spent undoing a routing decision, and it
+degrades selection because a model reasonably infers scope from a name. It also created an ordering
+hazard: the owning channel had to be present, so hosts ran a claim/ownership dance and the tool names
+changed depending on which surface won.
+
+`RegisterSessionClientTools(tools)` declares tools against the session. They carry their own names
+with no channel prefix, are sent to the model at mint alongside the channel tools, and are dispatched
+by **exact name** rather than by prefix — so they are reachable whether or not any particular channel
+is open.
+
+`Execute` may be **async**, which closes the issue's second limit. A channel's `ApplyAgentTool`
+returns `string`, so a cross-surface tool that had to await something could only be declared by a
+channel whose entry point happened to be async — making the tool's existence depend on which surface
+claimed it. Thrown errors are wrapped into `{ success: false, error }` exactly like a channel tool's.
+
+Starting a session whose session-tool name begins with a live channel's prefix now **throws** with
+both names. Exact-match dispatch means such a name would silently swallow every call sharing that
+prefix, and the channel would stop responding to its own tools with nothing to say why. The clash is
+deterministic and belongs to the host's naming, so it surfaces the first time the session runs.
--- a/packages/Angular/Generic/conversations/node_modules
+++ b/packages/Angular/Generic/conversations/node_modules
@@ -1,0 +1,1 @@
+/Users/madhav/Projects/MJ/packages/Angular/Generic/conversations/node_modules
--- a/packages/Angular/Generic/conversations/src/__tests__/realtime-session-channels.test.ts
+++ b/packages/Angular/Generic/conversations/src/__tests__/realtime-session-channels.test.ts
@@ -331,3 +331,162 @@ describe('RealtimeSessionService — debounced channel saves + teardown flush/di
     expect(error).toHaveBeenCalledWith(expect.stringContaining("Channel 'Echo' Dispose failed"), expect.any(Error));
   });
 });
+
+// ──────────────────────────────────────────────────────────────────────────────
+// #3536 — a client tool that belongs to the SESSION, not to a channel.
+//
+// Client tools route by the declaring channel's ToolNamePrefix, so a cross-surface tool had to be
+// declared by whichever channel happened to be present and then shipped wearing that channel's
+// badge: `browser_WorkspaceStatus` reporting all surfaces, not just the browser. Every description
+// then opened by contradicting its own name, and availability depended on which surface won a
+// claim/ownership dance.
+// ──────────────────────────────────────────────────────────────────────────────
+
+/** Reaches the private members these tests drive. */
+interface SessionToolInternals {
+  client: FakeRealtimeClient | null;
+  startChannels(): Promise<RealtimeToolDefinition[]>;
+  assertNoShadowedSessionTools(): void;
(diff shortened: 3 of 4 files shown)


MS-BC:
Closing: Caliber, the requesting consumer, ships a different mechanism (tools appended into clientToolsJson at mint, routed by its own prefix-handler registry) and consumes none of this API. Worth re-filing SMALL from the host-API angle rather than merely resurrecting: 'the host adds a session-scoped tool' is the first need of any embedder that does not own its mint path the way Caliber does — that PR should be this one's core without the channel plumbing.

PR #3926, opened 2026-08-17
fix(ai): ExecuteSimplePrompt model selection — four stacked defects (#3532)

### LINKED ISSUE #3532: ExecuteSimplePrompt is broken on 5.51: four stacked defects in model selection, each reporting as something else
### Summary

`ExecuteSimplePrompt` cannot run at all on 5.51.0. Four defects stack, and each one reports as something unrelated to its actual cause — we spent well over an hour chasing three misleading error messages before reaching the real one.

Reproduced against a stock install with a valid `AI_VENDOR_API_KEY__GeminiLLM` and active Gemini model rows.

### The four, in the order they surface

**1. A model row with a null `DriverClass` throws while BUILDING the candidate list.**
```js
for (const model of allModels) {
    const apiKey = GetAIAPIKey(model.DriverClass);   // null → toUpperCase() throws
```
One malformed row takes out prompt execution entirely. Surfaces as `Cannot read properties of null (reading 'toUpperCase')`, naming neither the row nor the operation.

**2. `AIModelType` is not populated on the engine's model objects**, so this matches nothing:
```js
const allModels = AIEngine.Instance.Models.filter(m => m.AIModelType?.trim().toLowerCase() === 'llm' && ...)
```
`allModels` comes out empty, and the caller is told **"No AI models with valid API keys found"** — a message about keys, for a problem that has nothing to do with keys. This is the most expensive one to debug because it sends you to your environment.

**3. `DriverClass` now lives on the model's VENDOR, not the model.** Even with (2) fixed, `GetAIAPIKey(model.DriverClass)` can never match, so the candidate list stays empty and you get the same misleading key message again.

**4. `APIName` also moved to the vendor**, so `chatParams.model = model.APIName` sends an empty model name. Surfaces as a **404 with an empty error message** — which reads as "that model doesn't exist" and sends you to check the provider's model list, where the model is present.

### What we changed locally

Rewrote `selectModelForSimplePrompt` to drive from the **vendor** side: any vendor whose `DriverClass` ends in `LLM` and whose key resolves makes its model a candidate; the winning driver and `APIName` are stamped onto the model so the caller's later `GetAIAPIKey(model.DriverClass)` and `chatParams.model` resolve to what it was matched on. A model whose driver cannot be resolved is skipped rather than allowed to throw.

After that, `ExecuteSimplePrompt` works first try.

### Two asks beyond the fix

- **The error messages are the real cost here.** "No AI models with valid API keys found" should not be raised when the candidate list was empty for a reason unrelated to keys, and an empty-model-name request should be rejected client-side rather than sent and 404'd.
- Whether `RunAIPrompt`/`RunAIAgent` share this selection path — we only exercised `ExecuteSimplePrompt`, so this may be broader than one resolver.

Happy to open a PR.

MS-BC:
Fixes #3532. **Still reproduces on `next`** — I re-read `selectModelForSimplePrompt` before writing this; all four are intact. Refreshed against `next` (the branch predated the current lockfile, which is what the earlier red CI was).

## The defects

Each one reports as something unrelated to its actual cause, which is the expensive part.

**1. A model row with a null `DriverClass` threw while BUILDING the candidate list.**

```ts
const apiKey = GetAIAPIKey(model.DriverClass);   // null → toUpperCase() throws
```

One malformed row took out prompt execution entirely, surfacing as `Cannot read properties of null (reading 'toUpperCase')` — naming neither the row nor the operation. Fixed at the source: a driver-less row has no key, which is an *answer*, and every caller already handles a falsy one. None of them expected a throw.

**2. `AIModelType` is not populated on the engine's model objects.**

```ts
const allModels = AIEngine.Instance.Models.filter(m => m.AIModelType?.trim().toLowerCase() === 'llm' && …)
```

It is a view column, so the filter matched nothing, `allModels` came out empty, and the caller was told **"No AI models with valid API keys found"** — a message about keys, for a problem with nothing to do with keys. That is the one that sends you to your environment. Selection now resolves the type through `ModelTypesByID` (an ID lookup off `AIModelTypeID`, a real FK the engine always carries), keeping the virtual column as a *fallback* rather than the source of truth — strictly more tolerant than before, never less.

**3. `DriverClass` lives on the model's VENDOR now.** Even with (2) fixed, `GetAIAPIKey(model.DriverClass)` can never match, so the list stays empty and you get the same misleading key message.

**4. `APIName` also moved to the vendor**, so `chatParams.model` went out empty and the provider answered a **404 with an empty error message** — which reads as "that model doesn't exist" and sends you to check the provider's model list, where the model is present.

## The fix

Selection is **vendor-first**, because both halves of "how do I run this" now live on the vendor:

> for each Active LLM model → its Active **inference-provider** vendors, in `Priority` order → the first whose `DriverClass` resolves an API key

Two decisions worth reviewing:

**`AIEngine.IsInferenceProvider`, not a driver-class name pattern.** The issue's local fix used "any vendor whose `DriverClass` ends in `LLM`". That heuristic admits vendor rows attached as the model's *developer*, which do not serve endpoints — you get a driver that instantiates and then cannot answer. `IsInferenceProvider` is the same memoized predicate `AIPromptRunner` already selects with, so this path now agrees with the rest of the stack instead of having its own opinion.

**The model and its vendor are returned as a pair, not stamped onto the model.** The issue's local fix wrote the winning driver and `APIName` onto the model entity so the caller's later reads would resolve. Those entities are the engine's **process-wide cache** — writing to one leaks into every other caller and makes the next request's answer depend on this one's. A `SimplePromptModelChoice { Model, DriverClass, APIName }` carries it explicitly instead, and a test pins that the cached entity is left alone.

### Both asks beyond the fix

- **The error says which wall was hit.** Three distinct messages: no Active LLM models at all / LLMs exist but none has an Active inference-provider vendor (check the vendor rows) / vendors exist but no key resolved (check `AI_VENDOR_API_KEY__<DriverClass>`). Only the last one mentions keys.
- **An empty wire name is refused client-side**, naming the model and the vendor row to fix, instead of being sent and 404'd with an empty message.

### `preferredModels` keeps matching what a caller actually holds

Going vendor-first quietly changed what the preference compares against: from the model's `Name`/`APIName` to the model's `Name` and the **vendor's** wire name. That wire name is an implementation detail — an Azure deployment name, a gateway slug — and a caller has the AI Models list, not the vendor rows. Matching only it would have dropped existing callers to power selection with no error at all, just a different model than the one asked for. All three names match, and a test covers a model whose vendor serves it under a different name (asserting the wire name still comes from the vendor).

### Also in this branch

`packages/AI/Core/node_modules` was committed by accident — an absolute path on one machine, absent from `next`, so a fresh checkout got a dangling link where the package's dependencies should be. Untracked.

### On the third ask

The issue also asks whether `RunAIPrompt`/`RunAIAgent` share this path. They do **not** — they go through `AIPromptRunner`, which already selects vendor-first via `IsInferenceProvider` and `Priority`. That is precisely why only `ExecuteSimplePrompt` broke, and this change moves it onto the same rules rather than inventing a parallel set. I have not changed those paths.

## Tests

17 new: 3 in `@memberjunction/ai` for the driver-name guard, and 14 driving the real selection method with a stubbed engine — type resolved by ID not by the virtual column, driver and wire name taken from the vendor, `APIName` fallback, developer-only vendors never chosen, higher-priority vendor wins when both work, highest-priority-with-a-key wins when one does not, a driver-less vendor row skipped, inactive/non-LLM rows ignored, preferred-by-name and preferred-by-wire-name, the three distinct error messages, and the cached entity left unmutated.

| mutant | result |
|---|---|
| read `AIModelType` directly again | 11 fail |
| take the driver from the model | 4 fail |
| take `APIName` from the model only | 3 fail |
| ignore `IsInferenceProvider` | 1 fail |
| ignore vendor `Priority` | 1 fail |
| one error message for every cause | 2 fail |
| drop the model-`APIName` preference match | 1 fail |
| take the wire name from the model, not the vendor | 4 fail |

The `Priority` mutant **survived the first round**: my ordering test had the high-priority vendor keyless, so `find` picked the right one whether or not the list was sorted. There is now a case where **both** vendors resolve keys and priority is the only thing deciding — which is the normal shape (a model available direct and via a gateway), and where ignoring priority means the answer depends on row order.

`MJServer` typechecks with zero errors in the touched file.

---


rkihm-BC (approved):
Approving. Read the full diff and verified the claims against the codebase rather than taking them from the description. No change requests — two observations at the end, neither blocking.

## What's good

**The diagnosis is worth more than the fix.** Four defects that each report as something unrelated to their cause, and the expensive one is being told *"No AI models with valid API keys found"* for a metadata problem. That message sends you to your environment for a day and the model list is fine the whole time. Splitting it into three messages that name which wall was hit — no Active LLM models / no Active inference-provider vendor / no key resolved, and only the last one mentioning keys — is the part of this PR that will save someone real time.

**`IsInferenceProvider` rather than a driver-name pattern, and the claim checks out.** I verified it is not aspirational: `AIPromptRunner.ts:391` and `ExecutionPlanner.ts:602` both delegate to the same memoized predicate in `BaseAIEngine.ts:1148`. So this path genuinely joins the rules the rest of the stack already follows instead of keeping its own opinion. The rejected alternative — "any vendor whose `DriverClass` ends in `LLM`" — admits vendor rows attached as the model's *developer*, which do not serve endpoints, and the failure that produces is a driver that instantiates and then cannot answer. Worse than the bug, because it looks like it worked.

**The pair-not-mutation decision is honoured in the implementation, not just in the interface.** `SimplePromptModelChoice` keeps the winning vendor off the engine's cached entity, and there is a test pinning that. The subtler leak would have been `inferenceVendorsFor` reordering the cached vendor array in place — it does not, because `vendors.filter(...)` returns a new array and the `.sort()` lands on that. That is precisely how this could have leaked while still satisfying the stated intent, and it is clean.

**Power-selection semantics are preserved.** `vendors.find(...)` yields at most one candidate per model, so `candidates` remains one entry per model and the `sort` / `midIndex` behaviour is exactly what it was. That is an easy thing to break silently while restructuring a list, and it is not broken.

**The `preferredModels` reasoning is the right call and the failure mode is correctly named.** Matching only the vendor's wire name would have dropped existing callers to power selection **with no error at all** — just a quietly different model than the one asked for. Matching all three names, with a test for the model's own `APIName` because that is what a caller actually holds, is right.

**The mutation table earns trust because of the mutant that survived.** The `Priority` mutant living through the first round — the ordering test had the high-priority vendor keyless, so `find` picked correctly whether or not the list was sorted — and being closed with a case where **both** vendors resolve keys is the most credible thing in the test section. That is the normal shape, a model reachable directly and through a gateway, and it is exactly where ignoring priority means row order decides the answer.

I also checked the `GetAPIKey` guard for a contract problem, since it returns `undefined` from a method declared `: string`. It is consistent: the function already returned `undefined` from its not-found branch and the package builds `strict: false`. The guard matches the existing contract rather than introducing a new one, and fixing it at the source is right — every caller already handles a falsy key, none of them expected a throw.

`packages/AI/Core/node_modules` is untracked on both branches, as described.

## Two observations, neither blocking

**The defect-4 guard is effectively unreachable, so it is belt-and-braces rather than the fix.** `APIName` resolves as `runnable.APIName?.trim() || model.APINameOrName`, and `APINameOrName` is `APIName ? APIName : Name`. Any real model row has a `Name`, so `choice.APIName` cannot come out empty and the empty-wire-name guard downstream can only fire for a model with no name at all. The actual fix for defect 4 is the **fallback**, not the guard. Worth recording so that nobody later tries to verify defect 4 by triggering that error, fails to trigger it, and concludes something regressed. Keep the guard — it is good defensive code with a message that names the row to fix.

**One claim I could not settle by reading, which does not change the verdict.** That `AIModelType` is unpopulated on the engine's cached model objects. It is a view column (`nvarchar(50)`, sitting alongside `Vendor` and the legacy model-level `DriverClass`), and the engine loads models with `CacheLocal: true` — whether the virtual column survives that cache round trip is a runtime property, not a static one. You reproduced it, so I will take it.

It does not affect correctness either way: resolving the type through `ModelTypesByID` with the column as a fallback is **strictly more tolerant** than reading the column alone, so the change is right whether or not defect 2 behaves exactly as described. And defects 3 and 4 are each independently sufficient to produce the reported symptom.

## On scope

Confirming `RunAIPrompt` / `RunAIAgent` were checked and deliberately not touched is the right answer to the issue's third question — they already select vendor-first through `IsInferenceProvider` and `Priority`, which is why only `ExecuteSimplePrompt` broke. Moving this path onto the same rules rather than inventing a parallel set is what stops the next divergence.

PR #3930, opened 2026-08-18
fix(storage): an unresolvable driver key fails at resolution, not three subsystems later

MS-BC:
## The failure this came from

A media file that exists, with bytes on disk and a valid `MJ: Files` row, would not serve. The only clue was:

```
[MediaStream] Failed to serve file 2DE8239D-…: source.driver.GetObject is not a function
```

The real cause was a storage provider whose `ServerDriverKey` had no registered driver class in that process — the module declaring it was never imported, so its `@RegisterClass` decorator never ran.

`ClassFactory.CreateInstance(FileStorageBase, key)` fell back to instantiating `FileStorageBase` itself. Every operation on that base is `abstract`, and abstract methods do not exist at runtime, so the caller received an object that looks like a driver and cannot do anything. The configuration error surfaced minutes later, in a different subsystem, as a missing-method TypeError that names neither the provider nor the key.

## The fix

**`FileStorageBase` is marked `@RequiresSubclass()`.** MJGlobal already provides this mechanism for exactly this case, and `ClassFactory`'s own docstring recommends it. This covers every resolution site, including five in `AI/Engine` and `ContentAutotagging` that carried dead `if (!driver)` branches — each was checked and either already throws on the next line or sits inside a logging catch, so none becomes a new uncaught crash.

**New exported `resolveStorageDriver(providerEntity)`** in `MJStorage/src/util.ts`, using `TryCreateInstance`; all three `initializeDriver*` paths route through it. The message names the key, the provider, every registered key, and both fixes:

> MJStorage: no file storage driver is registered for ServerDriverKey 'X' (provider "Y", ID …). Registered driver keys: … Fix by either (a) importing the package that declares the driver so its @RegisterClass decorator runs in THIS process — a driver whose module is never imported is never registered — or (b) correcting 'MJ: File Storage Providers'.ServerDriverKey to one of the registered keys above.

**`MediaStreamHandler`** had the same defect in miniature: three distinct unresolvable causes — no Files row, no `ProviderKey`, no storage account for the provider — all returned an **unlogged** 404, making a misconfigured provider indistinguishable from a deleted file. Each now logs its cause. The response stays a bare 404 because the route runs pre-auth.

## Tests

| Package | Build | Test |
|---|---|---|
| MJStorage | 0 | 0 — 81 passed, 4 files |
| MJServer | 0 | 0 — 842 passed, 56 skipped; 60 files |

Added: four fail-fast cases in `util.test.ts` (unresolved key; the exact historical shape where `Resolved: false` still offers a hollow instance; provider name/ID and registered-key list present in the message; `(none)` when nothing is registered) and three `@RequiresSubclass` contract tests in `FileStorageBase.test.ts`.

Verified against the original failure: the cryptic `GetObject is not a function` becomes a named configuration error identifying the exact provider row.

---

## Part 1 of 3 — MJ core

This PR is part of the MJ-core half of the realtime-workspace / multi-party push. It is
**not to be merged as part of the Caliber push** — it stands on its own tests.

Downstream Caliber PRs that need this: **none.** This is console / diagnostic support
surfaced while driving the Caliber final push; no Caliber PR is blocked on it and nothing
downstream fails to function without it.


rkihm-BC (approved):
Approving. Read the full diff and verified the claim that carries the actual risk — the blast radius of `@RequiresSubclass()` — rather than taking it from the description. Waited for a clean board before posting.

## What's good

**The diagnosis names a genuinely nasty failure class, and names it precisely.** `ClassFactory.CreateInstance` does not return `null` for an unregistered key — it falls back to the anchor base. Because every real operation on `FileStorageBase` is `abstract`, and `abstract` is **erased at runtime**, a bad `ServerDriverKey` produced an object that passes every truthiness check and whose methods are all `undefined`. Nothing complained until some distant subsystem called one, minutes later, in a different subsystem, as `source.driver.GetObject is not a function` — a message naming neither storage, nor the provider, nor the key. The dead `if (!driver)` branches scattered around the codebase are the fossil record of people expecting a null that never comes.

**`@RequiresSubclass()` is the right instrument rather than a local guard.** MJGlobal already provides this mechanism for exactly this case and `ClassFactory`'s own docstring recommends it, so this uses the framework's answer instead of inventing a parallel one at each call site.

**The error message is the highest-value part of the change.** Listing the keys that ARE registered is what separates a typo'd `ServerDriverKey` from a driver whose module was never imported — two causes that are indistinguishable until you can see what the factory actually knows. Naming both fixes, rather than just the failure, is what turns a support ticket into a self-service correction.

**I verified the blast radius rather than trusting it.** The marker changes behaviour at every resolution site repo-wide, not only the three routed through the new helper, so "none of these becomes a new uncaught crash" is the claim that had to hold. There are exactly five others, and each checks out:

| site | before | after |
|---|---|---|
| `ConversationAttachmentService:462` | hollow instance → `TypeError` inside a logging `catch` | named throw, same `catch` |
| `:500`, `:698`, `:748` | hollow instance → `TypeError` one line later | named throw one line earlier |
| `AutotagCloudStorage:155` | `IsConfigured` falsy → `LogError` → `return []` | throw, caught per content source |

The last is the only substantive change — it previously degraded gracefully rather than throwing — so I traced its caller. `ProcessContentSource` runs inside a per-source `try/catch` that logs and continues to the next source, so resilience is identical and the log line strictly improves: it now names the source AND the unresolved key, where before it said "not configured", which was not even the real cause.

**The `MediaStreamHandler` half is the same bug in miniature, and correctly split.** Three distinct causes — no `MJ: Files` row, no `ProviderKey`, no storage account for the provider — all returned an *unlogged* 404, which made a misconfigured provider indistinguishable from a deleted file. Logging each cause while keeping the response a bare 404, because the route runs pre-auth and must not describe server configuration to an unauthenticated caller, is the right split: the diagnosis goes where the operator can see it and nowhere else.

**The tests pin the things that would actually regress.** `should throw rather than return the base-class fallback instance` pins the exact historical shape, where `Resolved: false` still offers a hollow instance — that is the trap, and asserting on it is what stops someone "simplifying" back to `CreateInstance`. And `does not mark concrete subclasses (they are perfectly instantiable)` stops the marker over-reaching, which is the other direction this could go wrong.

I also checked two mechanical things: `LogError` was already imported in `MediaStreamHandler`, and the changeset covers both `@memberjunction/storage` and `@memberjunction/server` while stating explicitly that the change "covers every resolution site, including those outside MJStorage" — so the cross-package behavioural change is in the release notes rather than left implicit for whoever hits it.

## One observation, not an ask

`AutotagCloudStorage`'s `if (!driver)` branch is now doubly dead: `CreateInstance` either returns or throws, and it never returned `null` even before. Removing those branches would widen the diff into ContentAutotagging for no behavioural gain, so leaving them is the right call for this PR. Worth recording that they are now unreachable rather than merely redundant, so the next reader does not mistake them for live error handling.

PR #3932, opened 2026-08-18
fix(codegen): a new schema could never get an Application on PostgreSQL

MS-BC:
## The defect

`createNewApplication` names the `Application` columns unquoted:

```
INSERT INTO "__mj"."Application" (ID, Name, Description, SchemaAutoAddNewEntities, Path, AutoUpdatePath, DefaultForNewUser)
```

Normally that is fine — every statement passes through the PostgreSQL identifier auto-quoter on its way to the database. But this one is wrapped by `conditionalInsert`, which on PG emits a `DO $$ ... $$` block, and the auto-quoter **skips dollar-quoted blocks wholesale**: it cannot know whether their contents are SQL or literal text. Nothing downstream quotes them, so `ID` reaches PostgreSQL folded to `id`.

Reproduced against PostgreSQL 16 with the exact emitted statement:

```
ERROR:  column "id" of relation "Application" does not exist
LINE 1: INSERT INTO "__mj"."Application" (ID, Name, Description, Sch...
```

The same statement with the columns quoted succeeds.

SQL Server resolves the unquoted identifiers case-insensitively, which is why this survived — present unchanged in 5.49, 5.51 and `next`. **On PostgreSQL, `createNewApplication` has never worked.**

## Why nobody saw it

Every layer swallows it. `createNewApplication` catches its own exception, logs, and returns `null`. The caller logs `Unable to create new application for schema X` and carries on. CodeGen exits **0**.

So a new schema registers all of its entities normally, gets no Application, and every one of those entities renders in the UI's "System & Other" bucket — which is simply where an entity with no `ApplicationEntity` row goes. The tables and the data are fine; only the grouping is wrong.

Observed on a PostgreSQL instance: 27 entities in one schema, 0 Applications, 0 `ApplicationEntity` rows, `CreateNewApplicationWithSchemaName` `true` throughout.

## The fix

Quote the seven columns through `qi()`, exactly as the sibling `ApplicationRole` insert a few lines below already does.

`conditionalInsertSQL` now documents the contract that makes this the one exception to the usual convention: its arguments must be pre-quoted by the caller, because on PG the result is a dollar-quoted block the auto-quoter will not descend into. That contract was the trap — every other SQL-building path in this file can safely write identifiers bare, which is exactly why this one read as correct.

## Verification

- The emitted statement reproduced against a real PostgreSQL 16, before and after.
- New test `application-insert-pg-quoting.test.ts` (3 cases) asserts what actually reaches the database — the statement composed by the real `createNewApplication`, run through the real auto-quoter — on **both** dialects. Confirmed it fails without the fix (2 of 3 cases) and passes with it.
- One case pins the trap itself: the auto-quoter is a **no-op** on this statement. If that ever stops holding, the pre-quoting contract can be relaxed; until then it cannot.
- Full CodeGenLib suite: 1,086 passed, 0 failed.

## Risk

The quoted and unquoted forms are identical on SQL Server, where unquoted identifiers already resolved case-insensitively — the emitted text changes, the resolved columns do not. On PostgreSQL the statement goes from always failing to succeeding.

Scope is one statement plus a doc comment. Nothing else in the create path changes.

## Scope — what this PR deliberately does not include

Per @rkihm-BC's review: the `isSchemaNew()` **latch** in `createNewEntity` is a real, separate defect and is *not* fixed here. Because `isSchemaNew()` reads live `__mj.Entity` state and is evaluated before the entity INSERT commits, any single failure to create the Application — on either platform, from any cause — locks that schema out of ever getting one. This PR removes the *certainty* of that failure on PostgreSQL; it does not remove the latch.

That is tracked separately in #4034, with the repro and the proposed fix (swap the bespoke block for the existing `addEntityToApplicationForSchema` helper, as virtual and query entities already do). Keeping it out keeps this PR to one statement plus a doc comment, which is what CI has verified.

The changeset has been corrected to describe only what this diff contains — an earlier revision described the helper swap that is not in it. The branch name (`fix/application-latch-never-reopens`) predates that narrowing and now over-describes the change; the title and body are authoritative.

### DIFF
--- a/.changeset/application-insert-pg-quoting.md
+++ b/.changeset/application-insert-pg-quoting.md
@@ -1,0 +1,9 @@
+---
+'@memberjunction/codegen-lib': patch
+---
+
+A new schema gets its Application on PostgreSQL, and never loses the window on any platform
+
+Two defects closed the same door. `createNewApplication` named the `Application` columns unquoted, and `conditionalInsert` wraps that statement in PG's `DO $$ ... $$` block — which the identifier auto-quoter skips wholesale, since it cannot know whether a dollar-quoted block holds SQL or text. `ID` therefore reached PostgreSQL folded to `id` and the INSERT failed on every run, silently: the method catches, logs and returns null, and its caller carries on, so CodeGen finished green while the schema got no Application. Separately, Application creation was gated on `isSchemaNew()`, evaluated moments before the INSERT that ends that condition — so the window closed permanently on first use whether or not an Application was ever created.
+
+Columns are now quoted, and the create path is the shared `addEntityToApplicationForSchema` helper (already used for virtual and query entities), which looks up and creates if absent every time rather than once.
--- a/.changeset/application-insert-pg-quoting.md
+++ b/.changeset/application-insert-pg-quoting.md
@@ -2,8 +2,8 @@
 '@memberjunction/codegen-lib': patch
 ---
 
-A new schema gets its Application on PostgreSQL, and never loses the window on any platform
+A new schema gets its Application on PostgreSQL
 
-Two defects closed the same door. `createNewApplication` named the `Application` columns unquoted, and `conditionalInsert` wraps that statement in PG's `DO $$ ... $$` block — which the identifier auto-quoter skips wholesale, since it cannot know whether a dollar-quoted block holds SQL or text. `ID` therefore reached PostgreSQL folded to `id` and the INSERT failed on every run, silently: the method catches, logs and returns null, and its caller carries on, so CodeGen finished green while the schema got no Application. Separately, Application creation was gated on `isSchemaNew()`, evaluated moments before the INSERT that ends that condition — so the window closed permanently on first use whether or not an Application was ever created.
+`createNewApplication` named the `Application` columns unquoted, and `conditionalInsert` wraps that statement in PG's `DO $$ ... $$` block — which the identifier auto-quoter skips wholesale, since it cannot know whether a dollar-quoted block holds SQL or literal text. `ID` therefore reached PostgreSQL folded to `id`, and the INSERT failed on every run. It failed silently: the method catches, logs and returns null, and its caller logs and carries on, so CodeGen finished green while the schema got no Application and every one of its entities rendered in the UI's "System & Other" bucket. SQL Server resolves the unquoted identifiers case-insensitively, which is why this survived unnoticed since before 5.49.
 
-Columns are now quoted, and the create path is the shared `addEntityToApplicationForSchema` helper (already used for virtual and query entities), which looks up and creates if absent every time rather than once.
+The columns are now quoted through `qi()`, matching the sibling `ApplicationRole` insert a few lines below. `conditionalInsertSQL` now documents the pre-quoting contract that makes it the one exception to this file's usual "write identifiers bare, the auto-quoter handles it" convention.
--- a/packages/CodeGenLib/src/Database/codeGenDatabaseProvider.ts
+++ b/packages/CodeGenLib/src/Database/codeGenDatabaseProvider.ts
@@ -1077,8 +1077,15 @@ export abstract class CodeGenDatabaseProvider {
      * SQL Server: `IF NOT EXISTS (checkQuery) BEGIN insertSQL END`
      * PostgreSQL: `DO $$ BEGIN IF NOT EXISTS (checkQuery) THEN insertSQL; END IF; END $$`
      *
-     * @param checkQuery The SELECT query to check for existence.
-     * @param insertSQL The INSERT statement to execute if the check returns no rows.
+     * **Both arguments must already be identifier-quoted by the caller** (`qi()`/`qs()`).
+     * On PostgreSQL the result is a `DO $$ ... $$` block, and the identifier auto-quoter
+     * (`quoteSQLForExecution`, applied later by `runQuery`/`LogSQLAndExecute`) skips dollar-quoted
+     * blocks wholesale — it cannot know whether their contents are SQL or literal text. So this is
+     * the one SQL-building path where the usual "write it bare, the quoter handles it" convention
+     * does not hold: a bare `ID` survives to PG folded as `id` and the statement fails every run.
+     *
+     * @param checkQuery The SELECT query to check for existence. Identifiers must be pre-quoted.
+     * @param insertSQL The INSERT statement to execute if the check returns no rows. Identifiers must be pre-quoted.
      */
     abstract conditionalInsertSQL(checkQuery: string, insertSQL: string): string;
 
--- a/packages/CodeGenLib/src/Database/manage-metadata.ts
+++ b/packages/CodeGenLib/src/Database/manage-metadata.ts
@@ -6031,7 +6031,11 @@ export class ManageMetadataBase {
          // not products — hide them from new users. Application.DefaultForNewUser defaults to 1 in the
          // DB, so omitting the column here is what put raw '__mj_*'-named apps in every new user's
          // app switcher while the human-authored metadata app stayed hidden.
-         const appInsert = `INSERT INTO ${this.qs(mj_core_schema(), 'Application')} (ID, Name, Description, SchemaAutoAddNewEntities, Path, AutoUpdatePath, DefaultForNewUser)
+         // Every column is quoted through `qi()` rather than written bare. `conditionalInsert` wraps this
+         // statement in PG's `DO $$ ... $$` block, and the identifier auto-quoter skips dollar-quoted
+         // blocks wholesale (they can legally contain arbitrary text), so nothing downstream will quote
+         // these for us. Bare `ID` reaches PG folded to `id` and the INSERT fails on every run.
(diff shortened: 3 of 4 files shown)


.changeset/application-insert-pg-quoting.md
```diff
@@ -0,0 +1,9 @@
+---
+'@memberjunction/codegen-lib': patch
+---
+
+A new schema gets its Application on PostgreSQL, and never loses the window on any platform
+
+Two defects closed the same door. `createNewApplication` named the `Application` columns unquoted, and `conditionalInsert` wraps that statement in PG's `DO $$ ... $$` block — which the identifier auto-quoter skips wholesale, since it cannot know whether a dollar-quoted block holds SQL or text. `ID` therefore reached PostgreSQL folded to `id` and the INSERT failed on every run, silently: the method catches, logs and returns null, and its caller carries on, so CodeGen finished green while the schema got no Application. Separately, Application creation was gated on `isSchemaNew()`, evaluated moments before the INSERT that ends that condition — so the window closed permanently on first use whether or not an Application was ever created.
+
+Columns are now quoted, and the create path is the shared `addEntityToApplicationForSchema` helper (already used for virtual and query entities), which looks up and creates if absent every time rather than once.
```
rkihm-BC: I checked at the PR head commit — this isn't what shipped. `createNewEntity`'s bespoke `isSchemaNew()`-gated block (`manage-metadata.ts:5902-5937`) is untouched by this diff; it still calls `createNewApplication` directly, not `addEntityToApplicationForSchema`.

`isNewSchema` there is computed from live `__mj.Entity` state *before* the entity insert, which runs unconditionally. If `createNewApplication` then fails for any reason, this schema's one chance is gone — on the next entity (this run or any future one), `isSchemaNew()` returns `false` because the entity already exists, and the fallback path (`getApplicationIDForSchema`) only looks up, never creates. `addEntityToApplicationForSchema` (used for virtual/query entities at lines 1339 and 1777) doesn't have this problem — it looks up-and-creates-if-absent every call.

Either this line is describing work still to land, or the changeset needs trimming to match the diff — see the full review below for the trace.

rkihm-BC (commented):
Reviewed the full diff and verified every claim against the platform code rather than the PR's prose. The quoting fix itself is solid; there's a gap between what the changeset promises and what the diff delivers that I'd want resolved before merge.

## What's good

**The root-cause diagnosis is correct, and I verified the mechanism rather than trust it.** `AutoQuotePostgreSQLIdentifiers` genuinely skips dollar-quoted (`$$`/`$tag$`) blocks wholesale — that's a real, documented function in `packages/SQLDialect/src/postgresqlAutoQuote.ts`, not something invented for this PR's narrative. And `conditionalInsertSQL` on the PostgreSQL provider genuinely wraps its output in `DO $$ ... $$`. So a bare `ID` really does survive to PostgreSQL and really does fold to `id`, which really doesn't exist as a column — the fix targets the actual cause, not a plausible-sounding one.

**The fix is precisely scoped, and I checked there's nothing else like it to catch.** There is exactly one sibling call site (`addDefaultRolesForApplication`'s `roleInsert`), and it was **already** correctly quoted — confirming the PR's claim that this insert is following an existing pattern, not inventing one. I also read the rest of `createNewApplication` (`appCheckQuery`, `applicationExists`) and found no other unquoted identifier inside a `conditionalInsert`-wrapped statement. The fix is exactly as wide as the bug.

**The test is unusually rigorous for a one-line fix.** It hand-rolls `conditionalInsert` in a test subclass rather than mocking it — I diffed that against both real provider implementations (`PostgreSQLCodeGenProvider.conditionalInsertSQL`, `SQLServerCodeGenProvider.conditionalInsertSQL`) and they're byte-identical, so the test isn't exercising a simplified stand-in. It runs the statement through the *real* `AutoQuotePostgreSQLIdentifiers`, asserting what actually reaches the database rather than what the source string looks like. And the second test — pinning that the quoter is a no-op on this statement — is the right thing to guard: it's the assumption the whole fix depends on, stated as an assertion rather than left implicit.

**The new doc comment on `conditionalInsertSQL` earns its place.** It converts a trap that bit once into something the next author reading the abstract method's contract will see before they repeat it.

**Changeset level is correct** — `patch` on `@memberjunction/codegen-lib`, no migration or metadata touched. CI is fully green.

## Concerning — the changeset and branch name describe a second fix that isn't in this diff

The changeset's second paragraph:

> "Columns are now quoted, and the create path is the shared `addEntityToApplicationForSchema` helper (already used for virtual and query entities), which looks up and creates if absent every time rather than once."

I checked at the exact PR head commit — that's not what shipped. The diff touches exactly one production line (the column list) plus a doc comment. `createNewEntity`'s bespoke `isSchemaNew()`-gated block, at `manage-metadata.ts:5902-5937`, is completely untouched — it still calls `createNewApplication` directly, not `addEntityToApplicationForSchema`.

That block has the exact defect the branch name (`fix/application-latch-never-reopens`) names, and I traced every link in the chain:

1. `isNewSchema` comes from `isSchemaNew()`, which runs `SELECT COUNT(*) FROM __mj.Entity WHERE SchemaName = X` against **live database state** — not an in-process flag scoped to the current run.
2. It's computed **before** the entity insert, so it's still true for entity #1 of a new schema — and the entity insert happens unconditionally, regardless of what happens next.
3. If `createNewApplication` fails **for any reason** — a deadlock, a permissions hiccup, or (before this PR) the PG quoting bug — the schema gets no Application. The entity insert already succeeded and committed before that attempt even ran.
4. On entity #2 of that same schema — in this run, or any future CodeGen run — `isSchemaNew()` now returns **false**, because entity #1 is permanently in the table. The `else` branch only calls `getApplicationIDForSchema`, which I confirmed is a pure lookup (`SELECT ... FROM vwApplications`) with no create path anywhere in it.
5. I grepped every reference to `CreateNewApplicationWithSchemaName` / `SchemaAutoAddNewEntities` in CodeGenLib — there's no reconciliation sweep elsewhere in the file that would retroactively create a missing Application.

So that schema can never get an Application through CodeGen again, for the lifetime of the database. This is real, independent of the PG quoting bug, and not specific to PostgreSQL — a transient failure on SQL Server hits the same permanent lockout. The existing `addEntityToApplicationForSchema` helper (already used for virtual/query entities at lines 1339 and 1777) doesn't have this problem — it looks up-and-creates-if-absent on every call, so a failure on entity #1 just gets retried on entity #2. That's presumably why the changeset describes switching to it. But `createNewEntity`'s call site wasn't changed.

Net effect: **after this PR, PostgreSQL's *guaranteed* failure becomes fixed, but the latch design flaw that makes any single failure permanent remains, on both platforms.** The PR closes the door that was propped open by a certainty; it doesn't remove the doorstop.

## What I'd ask for before merge

Either land the helper swap — replacing `createNewEntity`'s bespoke `isSchemaNew()` block with a call to `addEntityToApplicationForSchema`, matching the pattern already used for virtual and query entities — or trim the changeset (and the PR body) to describe only what's actually in the diff, and open a follow-up issue for the latch. Both are legitimate; the PR as it stands promises the former and delivers only the latter, which is the part I'd want resolved rather than merged past.

## Bottom line

The quoting fix itself is correct, precisely scoped, and better-verified than most PRs I review — I found nothing to change in it. It's ready on its own merits. The open question is scope: does this PR ship with the latch fix included, or does the changeset get corrected to match what's actually here?

MS-BC:
Thanks — you were right, and I took the second of the two options you offered: this PR stays scoped to the quoting fix, and the changeset now describes only that.

**Changeset corrected** (b9bf8b6371). Dropped the "Two defects closed the same door" framing and the sentence claiming the create path had moved to `addEntityToApplicationForSchema`. It now describes one column list plus a doc comment, which is what the diff contains.

**Latch split out to #4034.** I re-verified your chain against the code at the PR head before writing it up rather than transcribing your review, and it holds at every link:

- `isSchemaNew()` (`:5992`) is `SELECT COUNT(*) FROM __mj.Entity WHERE SchemaName = X` — live DB state, not a run-scoped flag.
- It's evaluated at `:5902`, before the entity INSERT, which commits unconditionally at `:5905`.
- The `else` branch (`:5936`) calls only `getApplicationIDForSchema` (`:6098`), a pure `SELECT ... FROM vwApplications` — no create path.
- Nothing else in CodeGenLib reconciles a missing Application afterward.

Your framing of the net effect is the part I'd underline: this PR removes the *certainty* of failure on PostgreSQL, not the latch that makes any one failure permanent. The issue carries the repro, the traced chain, and the proposed helper swap, plus three things worth checking during that change — that `isSchemaNew()` has no other callers before deleting it, that the helper's `md.Refresh()` placement suits the table-backed path, and that the helper's `WARNING`-vs-`ERROR` log level is the right one for the merged path.

Two smaller notes:

- The **branch name** (`fix/application-latch-never-reopens`) predates the narrowing and now over-describes the change. I've left it alone rather than rename a branch under an open PR, and said so in the body — title and body are authoritative.
- On your point that the test hand-rolls `conditionalInsert` rather than mocking it: worth flagging that this is exactly why the second case exists. If either provider's implementation drifts from the inlined copy, that no-op assertion is what should catch it — it's the load-bearing one, not the belt-and-braces one.

CI is green at the new head, and the test file passes 3/3 locally against the pushed source.

PR #3933, opened 2026-08-18
fix(codegen): catch entity fields the base view cannot produce

MS-BC:
Adds `validateEntityFieldsResolve()` — a post-CodeGen check that cross-checks every entity's declared fields against the columns its base view actually produces.

**Scope note:** this PR originally also carried a PG migration repairing `__mj.vwEntities`. That has been dropped — see *Why the migration was removed* below. This is now validator-only.

## What it catches

An entity whose `EntityField` rows promise columns its base view does not expose. The runtime then cannot read the entity at all: the view resolves, the field list does not, and the query errors. A grid renders that as **"0 records / No data to display"** — no stack trace reaches the user, so an install can sit like this for months.

It is the read-side counterpart to `validateExpectedCRUDFunctions`: that one asks whether the runtime can *write* an entity; this asks the prior question of whether it can *read* it.

## Design decisions

- **Not filtered by `excludeSchemas`.** Excluded schemas are precisely where nothing else is watching — CodeGen skips SQL generation for them entirely (`sql_codegen.ts:133-135` splits included/excluded, excluded get permissions only), and `__mj` is in that list on essentially every install. So drift there is permanent and invisible. Checking them is the whole point.
- **Reported but non-fatal by default.** `MJ_CODEGEN_STRICT_FIELD_RESOLUTION=true` promotes it to a hard gate. A validator that fails an existing install's first run after upgrade is a validator people disable.

## The condition is real, not hypothetical

The PG port of the v5.45 External Data Sources migration registered `EntityField` rows for `ExternalDataSourceID` and `ExternalObjectName` without rebuilding `__mj.vwEntities`:

```
migrations/v5/V202607031200__v5.45.x__External_Data_Sources.sql       4943 lines
migrations-pg/v5/V202607031200__v5.45.x__External_Data_Sources.pg.sql 3286 lines
```

Both insert the same 54 `EntityField` rows. On PostgreSQL the metadata then promised two columns the view could not produce, and every read of `MJ: Entities` failed. Observed live, broken since early July, surfaced only as an empty grid.

## Why the migration was removed

`migrations-pg/v6/V202608202230__v6.1.x__PG_CodeGen_Regen.pg-only.sql` is **already on `next`** (2026-08-20) and rebuilds `vwEntities`, `vwEntityFields`, `vwEntityRelationships` and `vwCompanyIntegrationRuns`, appending all **seven** columns the frozen views were missing — `ExternalDataSourceID` and `ExternalObjectName` among them. The repair here fixed a strict subset of that.

It was also versioned **before** it (`V202608180600` vs `V202608202230`), so merging it would insert a lower-versioned migration into a chain that installs on v6.1.0-edge.3 have already passed. Flyway rejects out-of-order by default, so it would break upgrades on exactly the installs that are current.

Worth noting for the CI: the release-time PG migration gate passed this PR while it carried that migration, because the gate validates against a **fresh** database — where ordering is never a problem. An out-of-order insertion is invisible to a gate that only ever builds from empty.

## Why keep the validator

The drift was repairable and has been repaired. What was missing is anything that *notices* — the repair only happened because someone went looking. This is the part that would have caught it, and would catch the next one.

## Compatibility

Both dialects. Read-only introspection; no schema change, no generated-output change. Changeset included (`@memberjunction/codegen-lib: patch`).


MS-BC:
@rkihm-BC This should make it clear from running codegen in the future if postgres support, etc. does end up failing and there is some base view/underlying table inconsistency. The root issue I fixed and should be in next in a .pg.sql migration

rkihm-BC:
Reviewed the full diff and verified the platform assumptions rather than the description. No correctness defects — one gap I'd like closed before merge.

## What's good

**The validator is correctly modelled on its sibling.** It mirrors `validateExpectedCRUDFunctions` exactly — same `pool.query(sql)` + `result.recordset` shape (which is the `CodeGenConnection` contract), same lowercase normalisation on both sides, same `getXBySchemaSQL` template-method split. Nothing invented where a precedent existed.

**It checks `entity.BaseView`, not `GeneratedViewName` — the subtle one to get right.** MJ has two custom-view shapes: `BaseViewGenerated = 0` (the app owns the whole view) and `GeneratedBaseViewName` set (CodeGen writes an inner view, the app layers over it). In both, `BaseView` stays the public surface the runtime reads — `entityInfo.ts` says so explicitly. So a layered view is validated correctly, and a hand-maintained view that drops a declared column is flagged as the true positive it is. `GeneratedViewName` would have been the natural mistake and would have produced false negatives on exactly the entities most likely to drift.

**The missing-view skip is right.** `if (!columns) continue` — reporting every field of an absent view would bury the real signal, and a missing view is a different fault with its own diagnostics.

**Non-fatal by default, `MJ_CODEGEN_STRICT_FIELD_RESOLUTION=true` to promote it.** Correct call, and the stated reason is the right one: a validator that fails an existing install's first post-upgrade run is a validator people disable.

**Not filtering by `excludeSchemas` is well argued.** CodeGen emits permissions only for excluded schemas, so there is no regeneration pass to lean on — and `__mj` is in that list on essentially every install. Excluded schemas are exactly where nothing else is watching.

**No false-positive source at field level.** I checked whether `IncludeInAPI` exists on `EntityFieldInfo`, since an API-excluded field could plausibly sit in metadata but outside the view. It doesn't — `IncludeInAPI` is entity-level only. The unfiltered `entity.Fields` loop is correct as written.

**Dropping the PG migration was right, and the reasoning is checkable.** Inserting `V202608180600` into a chain where installs have already passed `V202608202230` is out-of-order, which Flyway rejects by default. The observation that the release-time PG gate passed it anyway — because that gate only ever builds from a fresh database, where ordering can never be wrong — is a genuinely useful gap to have written down somewhere.

Changeset `patch` on `codegen-lib` is correct (no migration, no metadata). MERGEABLE, all checks green.

## Requested before merge — tests

107 lines of new validator logic ship with zero test coverage. The logic isn't trivial: case normalisation across two dialects, the map keying, the missing-view skip, and the schema `IN (...)` list. If any of it is wrong the validator reports **nothing** — which is precisely the silence this PR exists to end. A validator that silently passes is worse than no validator, because it manufactures confidence.

Weighing it fairly: `validateExpectedCRUDFunctions` has no tests either (I checked all of CodeGenLib), so this isn't a convention violation — but it does mean there's no existing safety net for this shape of code, and you've shipped 18 tests on #3905, 4 on #3892 and 2 on #4028 in the last week, so the bar is clearly reachable.

Two cases would do it: one entity with a field the view doesn't produce (gap reported, correct `FieldResolutionGap` shape), and one fully-resolving entity (no gap). Those pin the normalisation and the map-keying, which are the parts that would fail silently.

## One latent trap worth a comment

The SQL filters `c.table_schema IN ('...')` using `entity.SchemaName` **verbatim**, while the map build and lookup both `.toLowerCase()`. On PostgreSQL an unquoted schema is folded to lowercase in the catalog, so an entity whose `SchemaName` is stored with different casing matches no rows, lands in the `!columns` branch, and is silently skipped.

MJ's convention is lowercase schema names so this is unlikely to bite, and it fails safe — a false negative, never a false alarm. But the two halves being inconsistent about case is the kind of thing that reads as deliberate later. Either lowercase the `IN` list too, or note in the comment that the skip is intentional for a schema the catalog spells differently.

## Bottom line

The validator itself I'd merge as-is — it's correct, well-scoped, and it catches a condition that was live on PostgreSQL for months and surfaced only as an empty grid. Just the tests.

MS-BC:
Both items done — thanks, the case observation was a real bug, not just an inconsistency.

## The schema-filter trap

Fixed, and it mattered more than "reads as deliberate later": on PostgreSQL an entity whose `SchemaName` is spelled differently from the folded catalog matched no rows, took the missing-view skip, and left validation silently — a validator reporting nothing, which is the failure this PR exists to end.

I went with `LOWER()` on **both** sides rather than lowercasing just the `IN` list. Lowercasing the list alone assumes the catalog is folded, which is true for an unquoted PG schema and not for a quoted one; `LOWER()` on both is correct however the schema was created, and makes the SQL agree with the JS by construction instead of by coincidence.

## Tests — 8 cases

Field the view can't produce (full `FieldResolutionGap` shape) · fully-resolving entity · the case mismatch · the SQL shape · missing-view skip · virtual/no-view entities · empty-input short-circuit · same-named views in different schemas.

Driven through the provider's own **prototype** rather than a subclass — the class is abstract with 60+ dialect members and stubbing them all to reach two would be testing the stubs. `Object.create` runs the real methods, and the base `getViewColumnsBySchemaSQL` serves both dialects, so this is the shipping path for SQL Server and PostgreSQL alike.

## One thing worth flagging, because it proved your point

My first fake pool returned its rows unconditionally — and **the case-mismatch test passed with the bug still in place.** The fake couldn't distinguish a matching filter from a broken one, so it reproduced the exact fault under test *inside the test written to catch it*. Precisely the "manufactures confidence" failure you described, one layer down.

The fake now parses the `IN` list it's handed, honours `LOWER()` when present, and filters as a real catalog would. Verified by reverting the fix: **2 tests fail** (the behavioural one and the SQL shape), both pass with it. The other 6 pass either way, which is correct — they cover behaviour this change doesn't touch.

CodeGenLib suite: **1110 passed**. The 2 failures in `entity-subclasses-hierarchy` and `BaseViewEmissionGoldenMaster` reproduce identically with my changes removed — pre-existing on this branch, unrelated.

rkihm-BC (approved):
Approving. Both items from the last round are addressed, and the second went further than I asked.

## The tests

Eight cases, not the two I asked for — and they're discriminating rather than decorative. The gap case asserts the full `FieldResolutionGap` shape with `toEqual` including `isVirtual: true`, so it pins the object contract and not just the count. The five beyond my ask cover exactly the paths that would fail silently: the missing-view skip, virtual entities and entities with no base view, no-query-when-nothing-to-validate, and — the one I hadn't thought to ask for — **same view name in two different schemas**, which is precisely what the `schema.view` map key exists to separate.

## The case-sensitivity fix

I raised this as a latent note rather than a request, and the fix chose the better of the two options. I suggested lowercasing the `IN` list; lowering **both** sides is stronger, and the comment says why: it holds whichever way the schema was actually created (PG folds unquoted names and preserves quoted ones, SQL Server stores as-written), so the SQL agrees with the JS *by construction* rather than by the coincidence of MJ's lowercase convention.

Both tests for it earn their place. The behavioural one (metadata `MyApp`, catalog `myapp` → gap still found) proves the trap is closed; the SQL-shape one, asserting the emitted statement contains `LOWER(...)` and `'myapp'` but not `'MyApp'`, is what would actually catch a regression to the verbatim form.

## Verified before approving

- Diffed `c83cdaf...79fe40e` — the only changes since my review are the test file and the 18-line provider edit. Nothing else moved.
- CI fully green: all six unit-test shards, Build, Source guards, adopted standards, dependency check, and the deterministic integration tier. MERGEABLE.
- Changeset unchanged and still `patch` on `codegen-lib`, which is correct — no migration, no metadata. It also still accurately describes the diff.

One cosmetic thing, explicitly **not** a merge condition: the changeset describes the validator but not the case-insensitive catalog lookup. Since that fix is internal to the validator being added rather than a separate behavioural change, I don't think it needs its own line — noting it only so the omission reads as deliberate rather than overlooked.

Good work on this one. The condition it catches was live on PostgreSQL for months and surfaced only as an empty grid; the validator is the part that would have noticed.

PR #3938, opened 2026-08-18
feat(core): add Entity.IsTotalSpecialization — EER completeness constraint

topherHunnewell:
### Summary
MemberJunction already models one of the two classical constraints on an ISA
(specialization/generalization) hierarchy — **disjointness**, via
`Entity.AllowMultipleSubtypes` (disjoint vs. overlapping subtypes). This PR adds
the **second** constraint, **completeness** (total vs. partial specialization),
via a new `Entity.IsTotalSpecialization` flag and matching enforcement in
`BaseEntity`.

### Why this belongs in ISA / EER
In the Enhanced Entity–Relationship (EER) model, a specialization is
characterized by **two independent constraints**:

1. **Disjointness** — may an entity belong to more than one subclass at once?
   (*disjoint* vs. *overlapping*) — already modelled as `AllowMultipleSubtypes`.
2. **Completeness** — must every superclass entity belong to *some* subclass?
   (*total* vs. *partial*) — added here as `IsTotalSpecialization`.

https://en.wikipedia.org/wiki/Enhanced_entity%E2%80%93relationship_model
> "The **completeness (or totalness) constraint** … may be total or partial. A
> **total** specialization constraint specifies that every entity in the
> superclass must be a member of at least one subclass in the specialization.
> … **Partial** specialization allows an entity not to belong to any of the
> subclasses."
> — Elmasri, R. & Navathe, S. B., *Fundamentals of Database Systems*, 7th ed.,
> Pearson, 2016, §4.3.1 (Constraints on Specialization and Generalization).

The same total/partial distinction is standard across the canonical texts — e.g.
Silberschatz, Korth & Sudarshan, *Database System Concepts* (7th ed., McGraw-Hill,
2020), which likewise defines the *completeness constraint* on a generalization
as **total** ("each higher-level entity must belong to a lower-level entity set")
or **partial**. MJ's table-per-type ISA implementation is precisely this EER
specialization mapped to shared-primary-key tables; MJ modelled the disjointness
axis but not the completeness axis, and this PR closes that gap so both
first-class EER constraints are expressible in metadata.

### What changed
- **Migration** (`V202608171559__v6.1.x__Entity_IsTotalSpecialization.sql`):
  adds a `BIT NOT NULL DEFAULT 0` `IsTotalSpecialization` column to `__mj.Entity`,
  plus the appended CodeGen tail (EntityField row, regenerated procs/grants).
  Additive — no existing row changes meaning.
- **`EntityInfo.IsTotalSpecialization`**: surfaces a
  (set on the parent/superclass entity).
- **`BaseEntity` enforcement** (`_InnerSave`): when
  (`IsTotalSpecialization = 1`) **and disjoint** (`AllowMultipleSubtypes = 0`), a
  **direct save of a standalone superclass record is refused** — the record must
  be created through a subclass, whose save persists superclass + subclass rows
  together via the existing leaf-chain path (unchan
  overlapping parents (no single required subtype), non-parent entities, and the
  default partial setting are all unaffected.
- **Generated tail** (surgically applied, per `migrations/CLAUDE.md`): the
  `MJEntity` entity subclass accessor + zod, and th
  output/input types.
- **`SQLConverter` core boolean-column catalog**: a`
  so CodeGen metadata BIT literals convert correctl the
  existing `AllowMultipleSubtypes` entry.

### Compatibility
Default `0` (**partial** specialization) reproduces
exactly. No existing entity opts in, so no existing

### Testing
- New unit test `baseEntity.isa.completeness.test.tfor a
  total disjoint parent; is correctly skipped when the flag is off, when the
  parent is overlapping, and on the legitimate leaf-chain (`IsParentEntitySave`)
  create-through-subclass path.
- Full ISA suite green (`baseEntity.isa.*`), SQLCont
  green, and the migration verified replayable on a fresh DB
  (`migrate` applies DDL + appended CodeGen as one


rkihm-BC:
@topherHunnewell , can you include a practical example where this would be used to better illustrate the need and how it would be used?

topherHunnewell:
> @topherHunnewell , can you include a practical example where this would be used to better illustrate the need and how it would be used?

in Betty we use the ISA relationship already existing in MJ to extend ContentItem and ContentSource.  we want to be able to enforce that for every entry in our ContentSource table, the corresponding entry in Betty.ContentSource exists so that we do not have orphans.  this PR is not a blocker, we are able to achieve this by modifying the Save and validate functions on the entities, but given that this is a part of ISA relationships,  this seemed like the more correct way to do so.

AN-BC:
@topherHunnewell @dray-mc  this is a great extension. Given that this is a pretty low level change I'd like to assign a couple of folks deep in MJ to review and we are doing a build today so as a practical matter this will need to go into a build later this week or early next week.

@MS-BC @jordanfanapour  can you two review this please and @rkihm-BC flagging you for visibility. 

We should also consider where in the `bizapps-**` stack we can use this new feature and implement it and test it there with full suite integration tests and also if there are places in the MJ core entities we can apply this to. It is always good to **use** new framework features in the framework itself if candidates exist, or in our bizapps stack if not, with full **integration live tests** so we are exercising it extensively. 

@topherHunnewell  @dray-mc @cmbrock-BC one idea for now if you want to enforce this is create a very thin sub-class of the `ContentSource` and `ContentItem` tables in the Betty implementation that enforce this business rule manually at the app level if you don't want to wait for this to go into MJ.

Excellent contribution, thank you!

topherHunnewell:
> @topherHunnewell @dray-mc this is a great extension. Given that this is a pretty low level change I'd like to assign a couple of folks deep in MJ to review and we are doing a build today so as a practical matter this will need to go into a build later this week or early next week.
> 
> @MS-BC @jordanfanapour can you two review this please and @rkihm-BC flagging you for visibility.
> 
> We should also consider where in the `bizapps-**` stack we can use this new feature and implement it and test it there with full suite integration tests and also if there are places in the MJ core entities we can apply this to. It is always good to **use** new framework features in the framework itself if candidates exist, or in our bizapps stack if not, with full **integration live tests** so we are exercising it extensively.
> 
> @topherHunnewell @dray-mc @cmbrock-BC one idea for now if you want to enforce this is create a very thin sub-class of the `ContentSource` and `ContentItem` tables in the Betty implementation that enforce this business rule manually at the app level if you don't want to wait for this to go into MJ.
> 
> Excellent contribution, thank you!

fully on board with all of this, especially the lower level review given how low level it is and I have no heart burn over it not making todays build

"one idea for now if you want to enforce this is create a very thin sub-class of the ContentSource and ContentItem tables in the Betty implementation that enforce this business rule manually at the app level if you don't want to wait for this to go into MJ." -- this is what we do currently and if there is found a reason that this addition would create issues in MJ, we can continue to do so. but if it does fit this feels like the "more correct" approach

rkihm-BC (commented):
## Review

Read the PR, the comments, the diff, CI, and the surrounding ISA machinery. Moving this to draft while we work through the design question below — no rush on it, since @topherHunnewell notes Betty already has a working path and this isn't a blocker.

---

### The example: the flag has to be set on an MJ-owned entity

Taking the `ContentSource` / `ContentItem` example literally: enabling this requires setting `IsTotalSpecialization = 1` on the **MJ-owned superclass entities**, not on Betty's subclass entities.

- `MJ: Content Items` / `MJ: Content Sources` are core MJ entities — `__mj.ContentItem` / `__mj.ContentSource`, from the v5 baseline (`migrations/v5/B202602151200__v5.0__Baseline.sql:12262`).
- The guard reads `this.EntityInfo` of the **parent** (`packages/MJCore/src/generic/baseEntity.ts`, `IsTotalSpecialization && IsParentType`). There's no scoping to a requesting app, subclass, or tenant.
- It's one row in `__mj.Entity`, so it's **install-wide** — it affects every consumer of that entity in that database.

The consequence isn't hypothetical. MJ's own shipping package `@memberjunction/content-autotagging` creates bare `MJ: Content Items` records in at least eight places — `AutotagBaseEngine.ts:505`, `AutotagWebsite.ts:423`, `AutotagAzureBlob.ts:50`, and others — via `GetEntityObject → NewRecord → Save()`. If the flag is set on that entity, **every one of those MJ core paths starts failing** with *"requires total specialization: … must be created through one of its subclass entities."* MJ's autotagging engine can't satisfy that; it doesn't know Betty's subclass exists.

Three further consequences:

1. **It also forces `AllowMultipleSubtypes = 0` on the MJ entity.** The guard only fires for disjoint parents, so getting the desired behavior also means asserting that MJ's `ContentSource` is disjoint — permanently foreclosing a second app in the same install ever subtyping it. That's a much bigger commitment than "don't orphan my rows."

2. **It doesn't enforce what was actually asked for.** The stated requirement is "for every `__mj.ContentSource` row, a `Betty.ContentSource` row exists." What the flag enforces is "*some* subclass row exists." With one child those coincide by accident; add a second child and the guarantee silently weakens. The constraint isn't expressible as "this specific subtype is required."

3. **Turning it on bricks pre-existing rows.** No backfill, no detection query, no validate-before-enable. Any row predating the flag with no subclass row becomes permanently unsaveable — and since the guard sits *before* the dirty check, even a no-op `Save()` on a clean loaded record returns `false` instead of `true`.

**Suggested direction:** put the constraint on the **child**, not the parent — something like "`Betty.ContentSource` is a *required* subtype of `MJ: Content Sources`," recorded on the child entity and consulted by the parent's save when it walks its children. That inverts ownership correctly (Betty declares Betty's invariant on Betty's entity, MJ's row is untouched), makes the multi-child case well-defined, and drops the incidental disjointness coupling.

---

### What's good

The theory is right and the write-up is excellent. MJ genuinely does model only one of the two EER specialization axes, and `AllowMultipleSubtypes` without a completeness counterpart is a real gap. Citations are correct; naming is defensible.

Mechanically, several things are done properly:

- **The delete half is already covered.** For disjoint parents, `_InnerDelete` cascades child → parent unconditionally, so deleting the subtype can't strand a superclass row. Correctly left untouched.
- **Correct restraint on PostgreSQL.** T-SQL migration only, plus the `SQLConverter/src/rules/CoreMetadataBooleanColumns.ts` entry — exactly the converter-side change the doctrine asks for, and no hand-authored `migrations-pg/**`. Easy to get wrong; this got it right.
- **The migration is genuinely additive**, and correctly skips a view regeneration (`vwEntities` is `SELECT e.*`, so the column flows through). The `EntityField` INSERT uses an apply-time sequence expression and a hardcoded UUID.
- **The guard is narrow and well-commented** — the leaf-delegation ordering argument is correct: an existing subclass row is discovered on load and delegates before the guard is reached.
- **`minor` on the changeset is right** for a migration branch.

---

### Blocking defects

**1. The generated tail was applied to the wrong entity.** `IsTotalSpecialization` was hand-inserted into `CreateMJAIAgentCoAgentInput` (`packages/MJServer/src/generated/generated.ts:3986`) and `UpdateMJAIAgentCoAgentInput` (`:4026`) — the **AI Agent Co-Agents** entity, unrelated to this feature. Those resolvers route to `CreateRecord('MJ: AI Agent Co Agents', …)`, which calls `SetMany` with `ignoreNonExistentFields = false` (`ResolverBase.ts:1222`), and `SetMany` **throws** on an unknown field (`baseEntity.ts:2577`). Any client sending the field gets a hard failure, and the public GraphQL schema now advertises a nonsense field on an unrelated type.

**2. The field is missing from the Entity input types, so the feature can't be turned on.** `CreateMJEntityInput` (`generated.ts:47842`) and `UpdateMJEntityInput` (`:48056`) do **not** have `IsTotalSpecialization` — only the read-side `MJEntity_` output type got it. Combined with `packages/Angular/Explorer/core-entity-forms/src/lib/generated/Entities/MJEntity/mjentity.form.component.html` having an `AllowMultipleSubtypes` field but no `IsTotalSpecialization` field, there is **no supported path to set the flag** — not GraphQL, not Explorer. Only raw SQL.

These two together look like a surgical hand-application of CodeGen output that landed in the wrong place and skipped the right one. Please re-run CodeGen and re-extract rather than patching by hand again.

**3. CI is entirely red — five failing checks, no signal.** All five fail on `ERR_PNPM_OUTDATED_LOCKFILE` for `packages/Angular/Generic/hierarchy-tree/package.json`. That is *not* this PR's fault (it touches no `package.json`) — the branch is based on `7fa1191a25`, well behind `next`. But the effect is that **the unit tests the PR body reports as green never ran**, and neither did the deterministic integration tier. Needs a merge from `next` before any of the testing claims can be trusted.

---

### Design concerns beyond the ownership question

**The guard misfires when child discovery fails rather than finds nothing.** `createAndLinkChildEntity` sets `_childEntity = null` when the child's `InnerLoad` fails, and `discoverChildEntityName` returns `null` when the provider lacks `FindISAChildEntity`. The guard can't distinguish "no subclass row exists" from "I couldn't see the subclass row." A user without read permission on the subclass entity, or a transient provider failure, gets a confusing *"must be created through one of its subclass entities"* on a record that already has one.

**The error message is wrong on updates.** `newResult.Type` is correctly `'update'` when `IsSaved`, but the message says the record *"must be created through one of its subclass entities"* — unhelpful when the caller is updating an existing row.

**Enforcement is TS-entity-layer only.** `spCreateContentSource`, direct SQL, `mj sync push`, import pipelines, and non-TS clients all bypass it. That's consistent with how `AllowMultipleSubtypes` disjointness is enforced, so it isn't a new sin — but the migration header and PR body both call it a "constraint," which oversells it. It's a policy check at one layer and the docs should say so.

**The unit tests don't exercise the risky part.** `makeStandaloneParent` sets `_childEntityDiscoveryDone = true`, bypassing discovery entirely — so the interaction between real discovery and the guard, the actual failure surface, is untested. Three of the four tests assert only the *absence* of a substring, which passes even when the save fails for an unrelated reason. The tests also mutate shared `EntityInfo` singletons; `try/finally` restores them, but only on the assertion path.

---

### Missing

- No integration test in the deterministic tier — this is cross-layer (metadata → `EntityInfo` → `BaseEntity` → provider), exactly what that tier is for.
- No docs update. ISA is documented in `guides/BUILDING_APPS_ON_MJ.md` and `packages/MJCore/README.md`; neither mentions the new axis.
- No detection or backfill story for enabling the flag on an entity that already has data.
- No guidance on the multi-child case (what "total" means when a disjoint parent has more than one child).
- Minor: the sequence expression is `MAX([Sequence]) + 73` rather than the `+ 1` CodeGen now emits. Collision-safe so the gate passes, but it leaves a 72-slot gap — likely the original literal `73` mechanically wrapped.

---

### Summary

The gap this identifies is real and worth closing; the API shape isn't there yet. Two mechanical asks — fix the misapplied CodeGen output in both directions, and merge `next` so CI actually runs. Separately, let's discuss reworking the constraint so it's declared on the **child** entity rather than by mutating MJ-owned parent metadata. That resolves the ownership problem, makes the invariant expressible for a *specific* required subtype (which is what Betty actually needs), and drops the requirement that a downstream app assert disjointness on an entity it doesn't own.

PR #3940, opened 2026-08-18
ci: cut PR unit-test wall time — fix turbo's global hash, granular remote cache, job split, scoped ESM sweep

CaelebB-BC:
> **Rewritten 2026-09-04, and much smaller than it was.** [#4041](https://github.com/MemberJunction/MJ/pull/4041) landed on `next` four days after this PR was reviewed and independently reimplemented most of its CI plumbing — the job split, the remote cache, the scoped ESM sweep, and the test sharding this PR had deferred. Rather than fight two incompatible restructurings of the same workflow, this PR now keeps only what #4041 missed. The [previous description is preserved below](#superseded-original-description) for the reviews that reference it.

## Why

PR unit-test runs were slow and the cache almost never helped. The reason was a single line of turbo semantics:

`hashOfInternalDependencies` — a hash over **every non-gitignored file in the root `package.json`'s workspace-dependency closure** — is an input to *every* task hash in the repo. The root manifest declared three `workspace:*` devDependencies (`@memberjunction/cli`, `@memberjunction/integration-test-suite`, `@memberjunction/server-bootstrap-lite`), whose combined closure was **154 of 310 packages**. Editing *any* file in *any* of those 154 changed the global hash and invalidated **all 310** — builds and tests alike. Task-level `inputs` cannot reach this; it is upstream of them. (Upstream: vercel/turborepo#8202, "the dogfooding tax".)

**This is still true on `next` today.** #4041 made the unit tier dramatically faster — sharding, a build/test split, a blobless clone — but it optimized *around* the artificial fan-out rather than removing it. Its own description reaches the diagnosis this PR disproves: *"`--filter=...[base]` selects 309 of 313 packages the moment a PR touches core or global — nearly everything depends on them — so affected-filtering narrows nothing."* Most of that 309 was not inherent. It was three lines in the root manifest.

The two changes compose: #4041 reduced what a rebuild **costs**, this reduces how **often** you pay one.

## What changed

**1. The global-hash fix — `package.json`, `mj.config.cjs`**

Removed all three `workspace:*` devDependencies from the root manifest and rewrote the 15 scripts that invoked the CLI to call `node packages/MJCLI/bin/run.js` directly. `hashOfInternalDependencies` is now `""`.

`mj.config.cjs` needed one load-bearing change: `checkModules` resolved the integration suite through a bare specifier that only worked via the root `node_modules` symlink the devDependency created. `check-module-loader.ts` *collects* load failures rather than throwing, so without an explicit path this would have silently degraded `mj test` to "Unknown integration check bundle". It now uses an absolute `__dirname`-based path, and `sibling-parity.test.ts` asserts that.

**2. Every bare `mj` now resolves — and a gate keeps it that way**

Deleting those devDependencies also deletes the workspace-root `node_modules/.bin/mj` that other things quietly resolved through. Every consumer is fixed:

- **`pg-migrations.yml` invoked `npx mj` at four sites.** With no root bin, `npx` walks up, finds nothing, and falls through to the npm registry — where the package named `mj` is **unrelated mongodb-js tooling**, in a job that holds database credentials. Now `node packages/MJCLI/bin/run.js`. That workflow deliberately does not trigger on `package.json`, so this would have stayed silent until the next release-time PG run. This is a supply-chain finding, not a CI nicety.

- **Three server packages** (MJCodeGenAPI, A2AServer, MCPServer) ran bare `mj codegen manifest` in a fallback-less `prestart`, exiting 127 wherever no global CLI existed — and where one did, silently resolving a *version-skewed* global (running `prestart` locally had a 5.51.0 global rewrite a 6.1.0-edge.2 manifest). Each now declares `@memberjunction/cli` as a devDependency, mirroring MJAPI. Leaf packages only, so `hashOfInternalDependencies` stays `""`.

- **Five more packages, seven scripts**, found by sweeping all 311 manifests. Review caught three; two were missed by a separator-anchored regex because the call site reads `; then mj codegen …`:

  | Package | Scripts |
  |---|---|
  | `packages/Angular/Bootstrap` | `prebuild` |
  | `packages/Angular/BootstrapLite` | `prebuild`, `postbuild` |
  | `packages/Angular/Explorer/explorer-core` | `prebuild` |
  | `packages/ServerBootstrap` | `prebuild` |
  | `packages/ServerBootstrapLite` | `prebuild`, `postbuild` |

  `ServerBootstrapLite` is itself one of the three removed root devDependencies. Every one of these ends in `|| echo 'Warning: …'`, so a lost CLI **exits 0** and the build proceeds against a stale class-registration manifest — a new `@RegisterClass` class never reaches it, and tree-shaking then drops it from bundled apps.

  Fixed with a path call, **not** a devDependency. The suggestion in review was "the same one-line devDep, they're leaf packages," but they are not: `ng-explorer-core` has 6 dependents and `ng-bootstrap` 2, so a devDep there takes a CLI edit from **6/310 to 12/310** invalidated packages — and the 6 added are the heaviest DOM-test packages in the repo. Worse, `@memberjunction/cli` **depends on** `@memberjunction/server-bootstrap-lite`, so a devDep there would be a build-graph **cycle**. A path call adds no graph edge and is behaviorally identical to the status quo, since the root bin was a symlink to the same file run with the same cwd.

- **`check-mj-cli-resolution.mjs`**, a new gate in the `guards` job. This defect class has now bitten three times and every instance failed silently. It permits only the two forms that actually resolve — a path call, or a bare `mj` in a package declaring the CLI itself — strips quoted strings so the `|| echo` prose is not flagged, and covers the root manifest by walking up from `packages/` rather than scanning the repo root, which would recurse into `.claude/worktrees/`. fs-only, so it runs with the other build-free guards.

**3. Docs**

58 in-repo `npx mj` occurrences became `pnpm mj` across `guides/`, `.claude/commands/` (live slash commands that would otherwise have executed the wrong package), the instruction files and `DEPLOYMENT.md`.

One straggler came in with the merge and is fixed here: `DEPLOYMENT.md`'s release runbook told maintainers to "prefer the repo-local CLI (`npx mj`, wired via the root `@memberjunction/cli` dependency)" — the very dependency this PR deletes, making the advice point straight at the unrelated registry package.

Deliberately untouched: published-package READMEs, `docker/`, and `UPGRADE-v6.0.md`, all of which address a **consumer** who installed the CLI from npm, where `npx mj` is still correct. Also untouched are `npx mj-sync` and `npx mj-standards`, which are different binaries.

## Measured

Packages invalidated by a one-file edit, before → after:

| Edited package | Before | After |
|---|---|---|
| `@memberjunction/cli` | 310 / 310 | 6 / 310 |
| `AI/Agents` | 310 / 310 | 37 / 310 |
| Explorer dashboards | 310 / 310 | 8 / 310 |

The honest limit: **caching buys nothing when you edit `core` or `global`.** A change there *is* the 200-package rebuild, and those PRs still pay close to a full run. The win is real for everything else.

## Verification

- `hashOfInternalDependencies` re-confirmed `""` after the merge and the devDependency additions; build graph acyclic
- Byte-identical manifest regeneration at all three path depths — ng-bootstrap (619 classes), ServerBootstrap (1008), ServerBootstrapLite `--scan-dist` (890), and explorer-core's `lazy-feature-config.ts`
- The new guard passes 314 manifests; reintroducing the `ServerBootstrapLite` regression exits 1 on both call sites
- `.github/scripts` suite **520 passing / 20 files**, including the two tests that mechanically pin the `guards` job's shape — so inserting a step did not defeat either ratchet
- `check:browser-manifest`, the gate #4018 added, is clean on this branch (4 manifests + 2 package manifests, 37 packages denied)
- `mj.config.cjs` loads and resolves `checkModules` to an absolute path; `sibling-parity`'s assertion on that string verified against the merged file
- All nine class-registration manifests byte-identical before/after the CLI-invocation change
- The three server devDependencies resolve the **workspace** CLI with no global `mj` on `PATH`
- `test.yml`, `pg-migrations.yml` and `integration.yml` all parse

Since verified in CI on an earlier head, not just locally: **`adopted standards`** (`check:standards`) and **`Integration (SQL Server, deterministic)`** — which covers `sibling-parity` — both passed, as did all six `Unit tests` shards and `Source guards`, with the resolution guard's step confirmed as `success` rather than skipped. CI is re-running on the current head.

## Merge conflicts, and what was discarded

`next` had advanced 695 commits. Three files conflicted (16 hunks), all resolved as **take-theirs**:

| File | Hunks |
|---|---|
| `.github/workflows/test.yml` | 12 |
| `.github/scripts/check-esm-imports.mjs` | 3 |
| `.github/scripts/__tests__/dynamic-replace-workflow-placement.test.mjs` | 1 |

A fourth file needed the same treatment despite merging cleanly. `__tests__/check-esm-imports.test.mjs` auto-merged as *additive text*, producing a file holding both test suites — 8 assertions against this branch's `--scope-file` API alongside 13 against `next`'s `--changed-since`. With the implementation now `next`'s, those 8 exercised a signature that no longer exists: a green-looking merge that would have failed on its first run.

Dropped from this PR, now redundant: the `quick-gates`/`backstop-alert` job split, the `rharkor/caching-for-turbo` remote cache, the `--scope-file` ESM scoping, and the deferred test-sharding note. `next` does all of it, mostly better.

**A later merge dropped one more.** [#4018](https://github.com/MemberJunction/MJ/pull/4018) landed the same `package-lock.json` → `pnpm-lock.yaml` trigger fix this PR carried — both the `paths:` filter and the affected-scope fallback regex that names it a second time — and did it in `integration.yml` and `dependency-check.yml` as well, which this PR never touched. Taken from `next`; this PR no longer claims it. That merge also conflicted in `package.json`, where #4018's two new `check:browser-manifest` scripts sit on the line next to this PR's `check:ui-layers` rewrite. **Both sides kept**, and the root manifest re-verified to declare no `workspace:*` dependencies — a `package.json` conflict is the one thing that could quietly undo this PR's entire point.

What remains of this PR in `test.yml` is now just the resolution guard's step and its `paths:` entry.

## Still open, orthogonal to this PR

- **`--concurrency=4` is uniform**, but only **69 of the 304** packages carrying a `vitest.config.*` reference the DOM preset (`vitest.dom.shared`); the other **235** are node-only and cheap. Letting those run wide while only DOM takes the cap is the biggest remaining free lever. Needs a real memory measurement first — the cap exists because of exit-137 OOMs, so raising it blind re-earns them. (Noted in the workflow.)
- **A PR run builds `refs/pull/N/merge`, not your commit**, so every push to `next` re-hashes whatever it touched in every open PR. With the global hash fixed, that churn is now proportional to what actually changed instead of being total.

---

<a name="superseded-original-description"></a>
<details>
<summary><b>Superseded original description</b> (the reviews above reference this)</summary>

The sections below described work that is no longer in this diff — the granular remote cache, the four-way job split, the scoped ESM sweep, and the wall-time measurements taken against them. [#4041](https://github.com/MemberJunction/MJ/pull/4041) supersedes all of it. Retained because the review thread quotes these numbers.

**2. Granular remote cache** — replaced the hand-rolled `actions/cache` arrangement (one ~2.5 GB tarball keyed by SHA) with `rharkor/caching-for-turbo`. An entire PR's 685 entries totalled 0.67 GB against 2.3 GB for a single old tarball. `next` instead wires a Vercel/self-hosted remote cache via `TURBO_TOKEN`/`TURBO_TEAM`.

**3. Job split** — `test` became `quick-gates` / `test` / `esm-guard` / `backstop-alert`, so pure-file guards failed in ~2 minutes instead of ~40. `next`'s `guards` job does the same thing, plus a `build` job feeding a 6-way sharded `test` matrix and a `unit-tests` gate.

**5. `quick-gates` needs an install** — the two DOM coverage ratchets read component surfaces through the real TypeScript compiler, so an install-less job died with `ERR_MODULE_NOT_FOUND` before counting anything. A store-cached `pnpm install` sat after the three `fs`-only guards and before the ratchets.

**6. Scoped native-ESM sweep** — `sweep()` took an optional `only` set and the CLI a `--scope-file=`, fed by the packages turbo rebuilt from changed source. `next` reaches the same place with `--changed-since <ref>`, deriving the set from git instead.

**Wall time**, against a 40.4-minute cold baseline: ~all invalidated (cold) 40.4 min; 256 packages (an `MJCore` edit) 41.0–41.8 min; 37 packages (`AI/Agents`) **12.8 min**. Measured before #4041's sharding, so the absolute numbers no longer apply.

**Deferred: test sharding.** Prototyped and removed — with the global-hash fix in place the marginal gain did not justify the added CI surface. #4041 has since implemented it (6 shards, bin-packed longest-first against measured per-package durations).

**Inherited, not introduced.** Lockfile churn: `pnpm-lock.yaml` carried 54 lines of drift beyond this PR's 9, and a bare `pnpm install --lockfile-only` on clean `next` reproduced exactly that delta, so `next`'s committed lockfile was simply stale. `check:claude-md` was red on the byte budget: root `CLAUDE.md` was already over the 18000-byte ceiling on `next`, this PR added 1 byte, and no CI gate enforces it.

</details>


AN-BC:
no input from me, I know nothing about CI, @rkihm-BC  and @cadam11  are the right reviewers for this @CaelebB-BC

rkihm-BC (commented):
Reviewed the full diff, description, and CI. Strong work. Below: what I verified independently, one question, two nits, and a follow-up.

**The `hashOfInternalDependencies` finding is the important part, and it corrected my own analysis.** I'd separately concluded the fan-out was inherent — "everything depends on `core`/`global`, so affected-filtering can't narrow." Your measurement shows 154/310 packages were dragged in *artificially* by three root `workspace:*` devDeps. Materially better diagnosis, and the before/after table (310→6, 310→37, 310→8) carries it. I also appreciate the stated limit — "caching buys nothing when you edit `core` or `global`" — which matches what I measured on a `MJCoreEntities` PR yesterday: 0 of 308 build tasks hit cache.

**Worth calling out the `pg-migrations.yml` catch specifically:** `npx mj` falling through to the unrelated `mj` package on npm, in a job holding database credentials, in a workflow that doesn't trigger on `package.json` — so it would have stayed silent until the next release-time PG run. That's a supply-chain finding, not a CI nicety.

Other claims I checked and found sound: root `package.json` already defines an `mj` script, so the `pnpm mj` docs rewrite genuinely resolves; the `paths:` filter really couldn't fire on `pnpm-lock.yaml`; the cache action is SHA-pinned with the "its entries are `dist/` that later runs execute" reasoning spelled out; published READMEs and `docker/` correctly left alone.

**One question — three Angular packages still invoke bare `mj`:**

```
cli-dep=1  packages/AI/A2AServer                      prestart, prebuild
cli-dep=1  packages/AI/MCPServer                      prestart, prebuild
cli-dep=1  packages/MJCodeGenAPI                      prestart, prebuild
cli-dep=1  packages/MJAPI                             prestart, prebuild
cli-dep=0  packages/Angular/Bootstrap                 prebuild   <- bare mj
cli-dep=0  packages/Angular/BootstrapLite             prebuild   <- bare mj
cli-dep=0  packages/Angular/Explorer/explorer-core    prebuild   <- bare mj
```

Bins on a current `next` install:

```
node_modules/.bin/mj                             EXISTS   <- the root bin this PR removes
packages/MJAPI/node_modules/.bin/mj              EXISTS   <- because MJAPI declares the CLI
packages/Angular/Bootstrap/node_modules/.bin/mj  ABSENT   <- resolves via the root bin
```

Post-merge those three lose `mj` resolution — and unlike the `prestart` cases they don't exit 127. Their `|| echo 'Warning: … using existing manifest'` swallows it, and `mj-class-registrations.ts` / `lazy-feature-config.ts` quietly stop regenerating.

I can see the argument that the fallback is *designed* for CLI-absent environments, so this may be deliberate. But inside this repo you presumably want regeneration, and the failure is invisible: a new `@RegisterClass` would simply not reach the manifest. If it isn't deliberate, it's the same one-line devDep you added to the other three — and they're leaf packages, so `hashOfInternalDependencies` stays `""`.

**Two nits:**

- The `test` job comment still reads *"Nearly every package depends on @memberjunction/global or core, so 'affected + dependents' fans out to the whole repo and every PR pays a full build."* This PR disproves that — worth updating so the next reader doesn't re-derive the wrong conclusion.
- On this PR's own run, `esm-guard` took **20m 13s** against the ~11m 26s baseline: root-touching PRs stay unscoped by design *and* now pay their own build. You already name "esm-guard duplicates the build" as the next lever — just noting this run is a live example of the worst case.

**Follow-up worth its own PR:** `--concurrency=4` is unchanged, and the test step is 25m 47s of the run. That cap exists for jsdom+AOT memory, but the DOM surface has grown — **69** DOM-preset packages vs **224** node-only, against a comment that still says "~8 DOM packages". Letting node-only packages run wide while only DOM takes the cap looks like the biggest remaining free lever. Orthogonal to this PR: yours reduces how *often* you pay a rebuild, this reduces what it *costs* when you do. Happy to hand over the measurements.

Nothing blocking except the Angular question.

cadam11:
Amazing work– this has been bugging me for a while but hasn't been able to reach the top of my priority stack yet.

Using plain node instead of the `mj`  bin makes sense to sever the Turbo hash problem. Also really great is removing the `npx mj` bits which have _always_ been wrong.

packages/AI/A2AServer/package.json
```diff
@@ -47,6 +47,7 @@
     "zod": "^3.25.0"
   },
   "devDependencies": {
+    "@memberjunction/cli": "6.1.0-edge.2",
```
rkihm-BC: **Required:** pin `6.1.0-edge.5` here and in `MCPServer` / `MJCodeGenAPI`, matching MJAPI, and refresh the lockfile.

The workspace CLI is at `6.1.0-edge.5` now (it was `edge.2` when this branch was cut). pnpm's rule for `linkWorkspacePackages` is that a local package is linked only when its version satisfies the declared spec; otherwise the spec is fetched from the registry. The lockfile still records `link:../../MJCLI` for these three only because it was resolved before the merge, and a frozen install honours it — which is why CI is green and your verification passed. But any re-resolution before the next release (a `pnpm update`, a lockfile regeneration after a conflict) would install `@memberjunction/cli@6.1.0-edge.2` from the registry into these three packages, and `prestart` would then run a version-skewed CLI silently — exactly the failure the description attributes to global installs, reintroduced by the dependency meant to prevent it. Three characters per file, and it makes the "mirroring MJAPI" claim true.

rkihm-BC (changes requested):
The core change is right and I can now confirm its central claim by measurement rather than by reading: in two worktrees with identical installs, `turbo run build --dry=json` reports `hashOfInternalDependencies` = `0407789cd0116c70` on `next` and `""` on this branch. Two items need to change before merge; both are stragglers of the merge from `next`, not flaws in the design.

**Verified:**
- `npm view mj` → `mongodb-js tooling` (1.0.5). `pg-migrations.yml` goes from five `npx mj` sites on `next` to one, and that one is the comment explaining why not.
- `@memberjunction/cli` does depend on `server-bootstrap-lite` (cycle claim), and counting workspace manifests `ng-explorer-core` has 6 dependents and `ng-bootstrap` 2 — so path calls, not devDependencies, is the right answer to my August question. Thank you for the thorough treatment.
- The gate passes 317 manifests on the branch; a planted bare `mj` in `ServerBootstrapLite`'s `prebuild` is caught with file, script and body. The `.github/scripts` suite passes 520/20 files. `mj.config.cjs` loads and resolves `checkModules` to an absolute path; `sibling-parity` asserts on it. CI fully green on this head.

**Required 1 — `Demos/BigSchemaDemo/recreate.sh` and `recreate.ps1` still run `npx mj` inside the repo.** Both `cd` to `` and then call `npx mj migrate`, `codegen --skipfiles`, `sync push`, `codegen --skipdb` with `DB_DATABASE` and friends set. The demo has no `package.json` of its own, so on `next` today those resolve through the root `node_modules/.bin/mj` this PR removes — after merge they fall through to the registry's `mj`, the same supply-chain shape you fixed in `pg-migrations.yml`, in a script holding DB credentials. They came in with #3804 (27 Aug), after your original sweep, the same way the `DEPLOYMENT.md` straggler did. Same fix: `node "$REPO_ROOT/packages/MJCLI/bin/run.js"` (or `pnpm mj`, since the root script exists). My repo-wide grep on the branch, excluding the consumer-facing docs you deliberately left alone, finds no other `npx mj` in scripts or workflows, so these two files complete the sweep. (The new gate cannot see them — it scans `package.json` manifests only; extending it to `*.sh` / `*.ps1` / workflows is a reasonable follow-up, not a condition here.)

**Required 2 — the three new `@memberjunction/cli` devDependencies pin `6.1.0-edge.2`, but the workspace CLI is `6.1.0-edge.5`, which is what MJAPI pins.** Details inline on the A2AServer manifest; the same applies to `packages/AI/MCPServer/package.json` and `packages/MJCodeGenAPI/package.json`.

**Not asks:** the `minor` changeset is forced by `check:changeset`'s `/^metadata/.+/` pattern matching `metadata/CLAUDE.md`; your note is accurate and narrowing the pattern belongs in its own PR. The follow-up on `--concurrency=4` remains orthogonal, as you say.

rkihm-BC (approved):
Approving. Both required items are done, and you went further than asked.

- **BigSchemaDemo:** `recreate.sh`, `recreate.ps1` and — found by your own sweep, missed by my grep — `run-cycle.sh` now call `node "$REPO_ROOT/packages/MJCLI/bin/run.js"`. My repo-wide grep on the new head finds no executable `npx mj` left; the remaining hits are prose in comments, a metadata description string, and PG migration notes. The commit message's reasoning for leaving `docker/**` alone (npm links every workspace bin into the root regardless of what the root declares) is correct, and the note that those Dockerfiles cannot build today for an unrelated reason is honest.
- **CLI pins:** all three now `6.1.0-edge.5`, matching MJAPI, and the lockfile resolves each to `link:` into the workspace on a specifier that actually matches — so no re-resolution can drift them to the registry copy.
- On the new head the resolution gate passes 317 manifests and the root manifest declares zero `workspace:*` devDependencies (re-verified in a worktree).

CI on this head was still running when I posted (Build, integration, PG validation, dependency check, standards, docs); the previous head was fully green and the delta is the three script files, three manifest pins and the lockfile, so I do not expect it to change — but the merge should wait for it to finish. Thanks for a careful turnaround on a long-lived PR.

PR #3960, opened 2026-08-19
fix: reconcile main's pnpm-lock.yaml with the v6.1.0-edge.2 manifests

cadam11:
**One sentence:** `main`'s committed `pnpm-lock.yaml` still pins 2287 `@memberjunction/*` workspace specifiers at `6.1.0-edge.1` while every manifest says `6.1.0-edge.2`, so no `pnpm install --frozen-lockfile` can succeed on `main` — this regenerates the lockfile with a real install, changing nothing about which packages resolve.

## Evidence

The versioned docs deploy's **Build v6 (main)** job failed on `pnpm install --frozen-lockfile` ([run 32270678576](https://github.com/MemberJunction/MJ/actions/runs/32270678576), 2026-08-19):

```
pnpm-lock.yaml is not up to date with <ROOT>/packages/AI/Core/package.json
  specifiers in the lockfile don't match specifiers in package.json:
* 1 dependencies are mismatched:
  - @memberjunction/global (lockfile: 6.1.0-edge.1, manifest: 6.1.0-edge.2)
```

Reproduced locally on a clean `origin/main` worktree, identical error.

## What actually drifted

**pnpm's error is not an inventory** — the frozen check aborts on the first mismatched importer it reaches (`packages/AI/Core`, alphabetically early) and reports one specifier. The real scope, from a full sweep of all 307 lockfile importers against their manifests:

| | count |
|---|---|
| `@memberjunction/*` specifiers at `6.1.0-edge.1` vs manifest `6.1.0-edge.2` | **2287** |
| importers affected | **291** of 307 |
| distinct drifted (lockfile → manifest) specifier pairs | **1** — `6.1.0-edge.1` → `6.1.0-edge.2` |
| third-party dependency drift | **0** |

The drift is a single homogeneous class. Three things that *look* like additional drift on a naive comparison are not, and are worth recording so the next person doesn't chase them:

- **`rimraf`, `@types/node`, `jsdom`, `browserslist` …** — the lockfile records `^5.0.10` / `24.10.11` / `26.1.0` against manifests asking for something else. These are pinned in root `pnpm.overrides`; pnpm records the *effective* (overridden) specifier. Correct as-is (69 rows).
- **`@angular/*`, `rxjs` in ~200 rows** — declared as `peerDependencies`, hoisted into the importer's `dependencies` block by `autoInstallPeers: true`. Specifiers match exactly.
- **`packages/geo/geo-maps` has no importer entry at all** — it is the only workspace package with zero dependencies of any kind, and pnpm omits dependency-less projects from `importers`. Not a defect. (This is why pnpm reports 308 workspace projects against 307 importers.)

## Why it happened

`publish.yml`'s main-branch path never refreshes the lockfile after the version bump:

1. `pnpm install --frozen-lockfile` runs **before** `changeset version` — by definition it cannot write the lockfile.
2. `pnpm run version` (`changeset version`) rewrites the version and every internal dependency range across ~300 manifests.
3. `ci/commit_push.mjs` does `git add ./*` and pushes to `main`. It would have committed a refreshed lockfile — but nothing in the job ever regenerated one, so it commits bumped manifests beside a stale lockfile.
4. The only lockfile refresh in the whole workflow is the final back-merge step, `ci/merge_main_and_update_lock.mjs` → `refreshLockfile()`, which runs **after `git checkout next`** and therefore lands on `next`.

Net effect: `next` gets a reconciled lockfile, `main` never does. That matches the observed state — this only surfaced now because the docs deploy recently began building `/v6` from `main`.

## The fix and its proof

Regenerated with a real `pnpm install` (**not** `--lockfile-only`, per the repo's pnpm doctrine).

Diff stat — one file, and the scope is provably narrow:

```
 pnpm-lock.yaml | 4634 ++++++++++++++++++--------------------
 1 file changed, 2293 insertions(+), 2341 deletions(-)
```

Section-by-section against `HEAD`:

| lockfile section | identical? |
|---|---|
| `packages` (resolved versions + integrity hashes) | **YES — byte-identical** |
| `overrides` / `patchedDependencies` / `settings` | YES |
| `importers` | no — the 2287 specifiers |
| `snapshots` | no — 48 lines net removed |

**`packages:` being byte-identical is the load-bearing check:** not one dependency was added, removed, upgraded or re-resolved, and no integrity hash moved. Every changed line is accounted for — 2287 importer specifiers, plus pruning of orphaned peer-context snapshot keys (`googleapis@144.0.0(encoding@0.1.13)` → `googleapis@144.0.0`, `webpack-dev-server@5.2.2(…)` gaining a `debug@4.4.3` context). Those keep the same resolved versions; only the peer-context suffix changes.

**Frozen-install proof**, run in a *second, separate* worktree checked out clean at `origin/main` with only this lockfile applied and **no `node_modules`**:

```
$ CI=1 pnpm install --frozen-lockfile
Scope: all 308 workspace projects
Lockfile is up to date, resolution step is skipped
Packages: +3224
Done in 23.8s using pnpm v10.33.0
EXIT=0
```

The run left the lockfile byte-identical, so the result is idempotent rather than quietly re-reconciled.

## For the reviewer

- **Only `pnpm-lock.yaml` is committed.** Nothing else was modified or staged.
- A pre-existing `WARN Failed to create bin … mj-querygen` appears in the install output. It is unrelated to this change (it shows up equally before and after, in any unbuilt checkout — the bin target lives in a `dist/` that has not been built yet) and it is a warning, not a failure.
- The 2287-line diff is large but mechanical. The section table above is the fast way to review it; spot-checking a handful of `specifier:` lines plus trusting `packages:`-identical covers the substance.

## Systemic follow-up (flagging only, not implemented here)

This will recur on the next edge release. `publish-lts.yml` already has the guard — a *"Version (normal mode) and refresh lockfile"* step that refreshes and commits the lockfile alongside the bump (#3559). `publish.yml`'s edge/main path has no equivalent and should get one.

One caveat if that work is picked up: **do not copy the LTS step verbatim.** It refreshes with `pnpm install --lockfile-only`, which this repo has already been burned by (a `--lockfile-only` refresh once drifted 827 packages, hence the standing rule to reconcile with a real `pnpm install`). The edge path wants the *shape* of the LTS guard with a real install, and the LTS path arguably wants correcting to match.

Worth confirming separately whether `main` was already unbuildable-from-clean at `v6.1.0-edge.1`, or whether the `.0` → `.1` release happened to leave it consistent — that determines whether this is the first occurrence or the first one anybody noticed.

https://claude.ai/code/session_01QnZaUxhuvSKe26rqvWuD6e

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -183,46 +183,46 @@ importers:
   packages/AI/A2AServer:
     dependencies:
       '@memberjunction/ai':
-        specifier: 6.1.0-edge.1
+        specifier: 6.1.0-edge.2
         version: link:../Core
       '@memberjunction/ai-agents':
-        specifier: 6.1.0-edge.1
+        specifier: 6.1.0-edge.2
         version: link:../Agents
       '@memberjunction/ai-core-plus':
-        specifier: 6.1.0-edge.1
+        specifier: 6.1.0-edge.2
         version: link:../CorePlus
       '@memberjunction/aiengine':
-        specifier: 6.1.0-edge.1
+        specifier: 6.1.0-edge.2
         version: link:../Engine
       '@memberjunction/api-keys':
-        specifier: 6.1.0-edge.1
+        specifier: 6.1.0-edge.2
         version: link:../../APIKeys/Engine
       '@memberjunction/config':
-        specifier: 6.1.0-edge.1
+        specifier: 6.1.0-edge.2
         version: link:../../Config
       '@memberjunction/core':
-        specifier: 6.1.0-edge.1
+        specifier: 6.1.0-edge.2
         version: link:../../MJCore
       '@memberjunction/core-entities':
-        specifier: 6.1.0-edge.1
+        specifier: 6.1.0-edge.2
         version: link:../../MJCoreEntities
       '@memberjunction/encryption':
-        specifier: 6.1.0-edge.1
+        specifier: 6.1.0-edge.2
         version: link:../../Encryption
       '@memberjunction/generic-database-provider':
-        specifier: 6.1.0-edge.1
+        specifier: 6.1.0-edge.2
         version: link:../../GenericDatabaseProvider
       '@memberjunction/global':
-        specifier: 6.1.0-edge.1
+        specifier: 6.1.0-edge.2
         version: link:../../MJGlobal
       '@memberjunction/server':
-        specifier: 6.1.0-edge.1
+        specifier: 6.1.0-edge.2
         version: link:../../MJServer
       '@memberjunction/server-bootstrap-lite':
-        specifier: 6.1.0-edge.1
+        specifier: 6.1.0-edge.2
         version: link:../../ServerBootstrapLite
       '@memberjunction/sqlserver-dataprovider':
-        specifier: 6.1.0-edge.1
+        specifier: 6.1.0-edge.2
         version: link:../../SQLServerDataProvider
       cosmiconfig:
         specifier: ^9.0.0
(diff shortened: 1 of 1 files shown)


cadam11:
Closing per Craig's ruling: merging to main would itself trigger publish.yml's edge path — the fix belongs in the workflow (lockfile refresh between `changeset version` and the release commit, mirroring #3559's LTS hardening), and any repair of main's current lockfile will be done locally and pushed with [skip ci] directly.

Note for that manual push: the branch `fix/main-lockfile-edge2` (db36799aa7) carries the exact reconciled pnpm-lock.yaml, proven with a clean --frozen-lockfile install in a fresh worktree — cherry-pickable to save re-running the install. Full drift inventory (2,287 specifiers / 291 importers, all edge.1→edge.2, zero third-party) is in the PR description.

PR #3968, opened 2026-08-20
fix(integration): enforce within-batch ExternalID identity before mapping

MS-BC:
A connector can emit two records with the same `ExternalID` in one batch, and the engine writes **two rows**. `ExternalID` is the identity the record map keys on, so those are two observations of one source record — there is no reading under which they are two rows.

The write path cannot catch it: it decides insert-vs-update by asking whether the identity exists in the **database**, and for a first-time record neither copy does, so both insert. Worse, it repeats — the pair re-inserts on every later sync, and afterwards no run can reconcile them because both rows map to the same `ExternalID`. The existing fingerprint check nearby only detects a batch repeated *in full* (the infinite-loop guard); it cannot see duplicates inside a single batch.

## Evidence

Observed in the field. One object carried a substantial excess of key-duplicated rows, and the shape of the duplicates is what identifies the cause:

- every duplicate group was **byte-identical on every captured column** — so no field distinguishes the rows, and the declared key was not too narrow;
- every duplicate group was **written inside the same second** — so the repeats arrived inside one batch, not via an at-least-once re-read, and therefore nothing a cross-batch identity check could ever catch.

## Change

`CollapseDuplicateIdentities` collapses duplicate identities per batch, keeping the **last** occurrence: upsert semantics, since the later entry is the more recent observation of that record's state, exactly what a single upsert would have left had the copies arrived in separate batches.

Deliberate choices:
- **Never silent** — emits `DUPLICATE_IDENTITIES_IN_BATCH` with a capped sample. A connector emitting duplicate identities is a defect worth fixing at its source, and the warning is how anyone finds out.
- **Identity-less records pass through** — collapsing records with an empty `ExternalID` would merge unrelated rows; that is a separate defect with a separate guard.
- **Zero-cost on the normal path** — a batch whose identities are already unique returns the original array reference.

## Verification

`npx tsc --noEmit` clean. Engine suite: **706 tests passed** (47 files), including 6 new covering last-wins, excess-copy counting, sample capping, identity-less pass-through, and trivial batches.

### DIFF
--- a/.changeset/within-batch-external-id-identity.md
+++ b/.changeset/within-batch-external-id-identity.md
@@ -1,0 +1,22 @@
+---
+"@memberjunction/integration-engine": patch
+---
+
+Enforce within-batch ExternalID identity before mapping.
+
+ExternalID is the identity the record map keys on, so two records in one batch carrying the same one
+are two observations of a single source record — never two rows. The write path cannot detect this:
+it decides insert-vs-update by asking whether the identity exists in the DATABASE, and for a
+first-time record neither copy does, so both insert and the pair re-inserts on every later sync. The
+existing fingerprint guard only catches a batch repeated in full (the infinite-loop case), not
+duplicates inside one batch.
+
+Measured on a live tenant: one object held 54,119 rows for 42,519 distinct keys — 11,632 excess,
+across 11,200 duplicate groups that were byte-identical on every captured column and written within
+the same second, i.e. the source listed the same element twice inside one batch.
+
+The engine now collapses duplicate identities per batch, keeping the last occurrence (upsert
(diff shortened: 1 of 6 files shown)


MS-BC:
CI note: the two red checks are inherited from `next`, not introduced here. `next`'s own Unit Tests runs (e.g. the 2026-08-19T23:09Z push) fail with the identical error — `Cannot find package 'mssql' imported from packages/MJCore/src/__tests__/metadata.live-db.integration.test.ts` — a module-resolution break in a file this PR does not touch. This branch's unit job otherwise reports **2105 passed, 0 failed**, including the 6 new BatchIdentity tests.

PR #3972, opened 2026-08-20
fix(core): validate value-list fields in EntityField.Validate() so an IN(...) CHECK has a runtime guard

### LINKED ISSUE #3969: BaseEntity never validates value-list fields, so an IN(...) CHECK has no TypeScript counterpart at all
## Summary

A field with `ValueListType = 'List'` carries an exhaustive set of legal values in
`__mj.EntityFieldValue`, and MJ uses it to render a dropdown — but **nothing in `BaseEntity`
validates against it**. An out-of-list value passes every rung of the validation ladder and is
refused only by the database, as a raw `CHECK` constraint error with no field attribution.

This matters more than it first looks, because for an `IN (…)` check constraint the value list is
the **only** representation CodeGen produces. `ParseCheckConstraints` turns
`CHECK (Status IN ('Active','Inactive'))` into a value list rather than into a generated
`Validate()` method — reasonably, since the value list is also what the UI needs. The consequence is
that this class of constraint has **no TypeScript counterpart at all**, so no client and no server
subclass can preflight it.

- Status: OPEN
- Reported: 2026-08-19, from `bizapps-contracts` (MJ 6.1.0-edge.2)
- Suspected layer: `@memberjunction/core` — `EntityField.Validate()` / `BaseEntity.Validate()`

## Repro — how to recreate

Any entity with a value-list field will do; this uses `bizapps-contracts`' `ContractType.Status`,
whose column is `CHECK (Status IN ('Active','Inactive'))` and whose metadata is
`ValueListType='List'` with two `EntityFieldValue` rows.

```ts
const md = new Metadata();
const t = await md.GetEntityObject('MJ_BizApps_Contracts: Contract Types', user);
await t.Load(someId);
t.Status = 'Archived';          // not in the list, and not permitted by the CHECK

const v = t.Validate();
console.log(v.Success, v.Errors.length);   // => true 0     <-- expected false / 1

await t.Save();                             // refused by SQL Server, not by MJ
// The Violation of CHECK constraint "CK_ContractType_Status" ... reaches the user verbatim.
```

Confirmed by inspection as well as by running it:

- `EntityField.Validate()` (`MJCore/src/generic/baseEntity.ts`) checks nullability, `MaxLength`,
  date parseability and numeric range. It never reads `EntityFieldValues`.
- `grep -rn "ValueListType\|EntityFieldValues" packages/MJCore/src/generic/baseEntity.ts` returns
  **nothing** — the metadata is present on `EntityFieldInfo` and consumed only by the UI
  (`form-field.component.ts` `PossibleValues`, line ~373).

## Expected vs actual

**Expected** — a value outside an exhaustive list fails `Validate()` with a field-named error, the
way nullability and `MaxLength` already do, so the browser refuses before the round trip and the
server refuses with a message naming the field.

**Actual** — `Validate()` returns success. The value reaches SQL Server and comes back as a raw
constraint violation, unattributed to any field, and the form shows it as a toast.

## Why the dropdown is not the answer

The dropdown makes the *form* path safe, and that is presumably why this has gone unnoticed. Every
other path is unguarded: `mj sync push`, the GraphQL mutations, an entity subclass setting the field
in code, an Action, a migration-time data load, and any app's own service layer. Those are exactly
the paths where a typo is least likely to be seen by a human before it lands.

It also leaves a documented promise unmet. Consumers reasonably read the ladder as "field rules come
from the schema for free" — `bizapps-contracts`' own plan states it as
*"Field rules — generated validation from the schema (CHECKs, nullability, value lists). Free
everywhere; never restated by hand."* That is true for CHECKs and nullability and false for value
lists, and the failure is silent, so the belief survives review.

## Suggested fix

In `EntityField.Validate()`, after the existing checks:

```ts
if (ef.ValueListTypeEnum === EntityFieldValueListType.List &&
    this.Value !== null && this.Value !== undefined && this.Value !== '') {
    const allowed = ef.EntityFieldValues;
    if (allowed?.length && !allowed.some(v => v.Value === this.Value)) {
        result.Success = false;
        result.Errors.push(new ValidationErrorInfo(ef.Name,
            `${ef.DisplayNameOrName} must be one of: ${allowed.map(v => v.Value).join(', ')}. ` +
            `Current value is '${this.Value}'.`, this.Value));
    }
}
```

Three details that decide whether this is safe to land:

1. **Only `List`, never `ListOrUserEntry`.** The latter exists precisely to permit values outside the
   list, so validating it would break every entity that opted into free text.
2. **Skip when the list is empty.** A field marked `List` whose `EntityFieldValue` rows have not been
   populated must not reject everything.
3. **Null is the nullability check's job**, not this one — an empty value should produce at most the
   existing "cannot be null" error, never two errors for one mistake.

Worth confirming before merge: whether any existing MJ entity has a `List` field carrying data that
predates its current value set. If so this turns a silent pass into a failed save on the next edit of
those rows, which is correct behaviour but wants to be a known migration rather than a surprise.

## Related

`bizapps-contracts` is adding an app-level guard in its shared entity subclasses in the meantime;
that is a stopgap for one app and is deleted when this lands.

MarceloT-BC:
Fixes #3969.

> ⚠️ **Authored by Claude Code.** This is not a rubber-stamp PR — it changes a rung of the validation ladder that every `Save()` in every app passes through, and the merge risk lives entirely in the comparison semantics (see the open question at the bottom). Please have someone who understands the entity/validation layer read it in context rather than approving on green CI.

## The bug

A field with `ValueListType = 'List'` carries an **exhaustive** set of legal values in `__mj.EntityFieldValue`, and MJ used it only to render a dropdown — nothing in `BaseEntity` validated against it. `EntityField.Validate()` checked nullability, `MaxLength`, date parseability and numeric range, and never read `EntityFieldValues`.

That matters more than it looks, because for an `IN (…)` CHECK the value list is the **only** representation CodeGen produces: `ParseCheckConstraints` turns `CHECK (Status IN ('Active','Inactive'))` into a value list *instead of* a generated `Validate()` method (reasonably — the list is also what the UI needs). So that class of constraint had **no runtime TypeScript counterpart at all**. An out-of-list value passed every rung of the ladder and was refused only by SQL Server, as a raw CHECK violation attributed to no field.

The dropdown makes the *form* path safe, which is presumably why this went unnoticed. Every other path was unguarded: `mj sync push`, GraphQL mutations, an entity subclass setting the field in code, an Action, a migration-time data load — exactly the paths where a typo is least likely to be seen by a human before it lands. CodeGen does emit a compile-time union type on the generated setter, so a literal `t.Status = 'Archived'` is a type error in typed code; the hole is every runtime path where the value is a variable.

## Repro — how to recreate

```ts
const t = await md.GetEntityObject('MJ_BizApps_Contracts: Contract Types', user);
await t.Load(someId);
t.Set('Status', 'Archived');   // not in the list, and not permitted by the CHECK

const v = t.Validate();
console.log(v.Success, v.Errors.length);   // before: true 0     <-- expected false / 1

await t.Save();                 // before: refused by SQL Server, not by MJ.
                                // "Violation of CHECK constraint CK_ContractType_Status" reached the user verbatim.
```

That repro is encoded as a test at both layers in `packages/MJCore/src/__tests__/baseEntity.valueListValidation.test.ts`: directly on `EntityField`, and end-to-end through metadata loaded by the real provider post-processing, where `Save()` is now refused locally and the provider's `Save` is never reached. **4 of its 19 tests fail without the fix.**

## The fix

A value-list rung in `EntityField.Validate()`, after the existing checks. Three boundaries decide whether this is safe to land, and all three are pinned by tests:

1. **`List` only, never `ListOrUserEntry`** — the latter exists precisely to permit values outside the list, so validating it would break every field that opted into free text.
2. **An empty value list never rejects anything** — a field marked `List` whose `EntityFieldValue` rows have not been populated is missing metadata, not a rule.
3. **Null/blank stays the nullability check's job** — one mistake never produces two errors.

The comparison is **trimmed, case-insensitive and string-based**, deliberately:

- **Trimmed** for leading whitespace and stray spaces on the metadata side. (An earlier version of this PR justified the trim with a 9,301-row padding measurement over fixed-width `nchar` columns; @rkihm-BC correctly pointed out that measurement was taken over raw SQL rows and does not describe this path — `EntityField`'s value setter already rtrims fixed-width columns before `Validate()` reads the value. The trim is kept for the reasons above, not that one.)
- **Case-insensitively**, and this turned out to be load-bearing rather than defensive: two MJ core fields (`MJ: Entity AI Actions`.TriggerEvent, .OutputType) have a SQL default that matches their own value list **by case alone**, and `EntityField`'s constructor seeds a new record from `DefaultValue` — so a case-sensitive comparison would fail validation on creating either record at its database default. SQL Server's default collation is case-insensitive too, so refusing a case variant would refuse what the database accepts. (PostgreSQL is case-sensitive, so there a variant is still refused — by its CHECK, not by this rung.)
- **As strings** because `EntityFieldValue.Value` is always a string in metadata while a numeric value list's runtime value is a number — a strict `===` would reject every legal value on such a field.

Long lists truncate in the message (`EntityFieldInfo.MaxValueListValuesInErrorMessage`, 25) so the error stays readable.

## Data-safety check (the question #3969 raised)

#3969 asked whether any existing entity has a `List` field carrying data that predates its current value set — if so this turns a silent pass into a failed save on the next edit of those rows. Answered with read-only SQL over **five live 6.x instance databases** (398–468 `List` fields with populated value lists each):

| Comparison | Off-list rows found |
| --- | --- |
| As implemented (trimmed, case-insensitive) | **0** across all five databases |
| Trimmed but case-sensitive | 0 stored rows — but see the default-value finding below |
| Exact (`===`-equivalent) | 9,301 raw rows in `nchar` columns — **does not apply to this path**, since the field's value setter rtrims fixed-width columns before validation reads them |

The scan is not vacuously green: the same query with the predicate inverted returns thousands of in-list rows, so it demonstrably reads table data.

## Validation

- **New tests:** 19, of which 4 are red without the fix and all 19 green with it.
- **`@memberjunction/core` full suite:** 2124/2124 tests pass. One test *file* fails to collect in my worktree — `metadata.live-db.integration.test.ts` cannot import `mssql` (not a MJCore dependency) — pre-existing and unrelated to this change.
- **Downstream packages re-run:** `codegen-lib` (53 files), `metadata-sync` (16), `interactive-component-types` (8) — all green.
- **`tsc` build clean.**
- **Not done:** I did not drive a real save through a booted MJAPI/Explorer. The MJ half of the repro is covered by the tests above; the database half is SQL Server behaviour the issue already documented.

## What counts as "no value" (second commit)

`null`/`undefined` are left to the nullability check — an unset value is legal on a nullable column. An **empty or whitespace-only string is not** treated as absence: SQL Server pads on comparison, so `''` and `'   '` are the same value to a CHECK constraint and it refuses both, and skipping them left a hole exactly where a blanked-out field lands. When the column is nullable the message names the alternative — *"must be one of: Active, Inactive (or null)"* — so a caller knows what "no value" actually looks like.

Checked before changing it: across the same five live databases, **no row holds a blank or whitespace-only value** in a value-list field, and Explorer's value-list dropdown renders only real values (no empty option, no clear affordance), so no existing data and no UI path is affected.

Worth a reviewer's eye: MJ's AI-generated CHECK-constraint validators reportedly treat `''` as absent, so this rung is deliberately stricter than those. The database is the floor either way; if you'd rather the two agree, skipping `''` here is a one-line change.

## Why the `typeof` gate admits only string and number

It fails **open**, never closed. Every value list in MJ metadata today is on an `nvarchar`/`nchar` column (measured: 455 `nvarchar` + 7 `nchar` on a current instance, nothing else), CodeGen's constraint parser only ever emits string values, and the one place a non-string list is anticipated is the generated union type via `NeedsQuotes`, which is numeric. Nothing in the schema *prevents* declaring `List` on a `bit` or `date` column — `CK_EntityField_ValueListType_New` constrains the mode, not the column type — so booleans and Dates are excluded on purpose: a bit column carrying a `'1'`/`'0'` list would see `String(true) === 'true'` and reject every legal value. An unsupported type skips validation rather than manufacturing a failure.

## Answered by review: keep it case-insensitive

@rkihm-BC found the hole in my instrument — the row scan measures *stored rows* and misses the **default-value path**, where `EntityField`'s constructor seeds a new record from `DefaultValue`. Two MJ core fields have a default matching their list by case alone, so case-sensitivity would fail validation on creating either record at its database default. Settled; the evidence is now in the code comment and a test.

That scan also turned up something worth knowing separately: **six** MJ core `List` fields have a SQL default that matches their value list *not at all* — `MJ: Queue Tasks`.Status defaults to `Pending` against `In Progress | Completed | Failed`, `MJ: Audit Logs`.Status to `Allow` against `Success | Failed`, `MJ: Version Installations`.Type to `System` against `New | Upgrade`, and three more. Those defaults violate their own CHECK constraints, so a create relying on the default is refused by SQL Server today; this rung turns that into a field-named validation failure instead of a raw constraint violation. Earlier and better, not a regression — but it means this rung's message is where those stale defaults will surface. Filing that separately if @AN-BC agrees.

## Original open question (resolved above)

**Should the comparison be case-sensitive?** As shipped it is lenient, on the principle that MJ validation should never refuse a value the database would accept. The measurement above says strict-case would break nothing that exists today, so the argument for tightening it is real: a case variant currently saves and then displays as a blank dropdown, which is its own data-quality bug. It is a one-line change (drop `.toLowerCase()` from `normalize`) plus one test flip. Trimming, on the other hand, **cannot** be tightened — that is the 9,301 rows.


packages/MJCore/src/generic/baseEntity.ts
```diff
@@ -347,6 +353,77 @@ export class EntityField {
...
+            //     parser only ever extracting quoted literals. `number` is admitted because the
+            //     generated union type anticipates a non-quoted list via NeedsQuotes. Booleans and
+            //     Dates are excluded deliberately: a bit column carrying a '1'/'0' list would see
+            //     String(true) === 'true' and reject every legal value, and a Date has no sane
+            //     string form to compare — so guessing there would break saves rather than guard
+            //     them.
+            if (ef.ValueListTypeEnum === EntityFieldValueListType.List &&
+                (typeof this.Value === 'string' || typeof this.Value === 'number')) {
+                const allowedValues = ef.EntityFieldValues;
+                if (allowedValues && allowedValues.length > 0) {
```
AN-BC: strictly speaking of we have `ValueListType` == `List` && we have **no** allowedValues, that is a condition where nothing is allowed, which is not a condition that _should_ really ever exist. However, instead of silently skipping this path, I think we should loudly log this with LogError in the `else` branch of this `if` statement (which doesn't exist yet). It is extraordinarily unlikely to be hit but worth having here. This is also something that CodeGen _could_ check for - a column that has ValueListType == List and has no values. Since the vast majority of the situations that drive the setting of ValueListType come from CodeGen parsing check constraints this is, again, extraordinarily unlikely.

packages/MJCore/src/generic/baseEntity.ts
```diff
@@ -347,6 +353,77 @@ export class EntityField {
...
+                    //     and refusing it here would turn a save that succeeds today into a failure.
+                    //     The asymmetry is deliberate and known: PostgreSQL IS case-sensitive, so on
+                    //     PG a case variant is still refused — by its CHECK, not by this rung.
+                    //     Tightening to an exact case match is a separate, breaking decision.
+                    //   * EntityFieldValue.Value is always a string in metadata, while the field's
+                    //     runtime value may be a number (a numeric value list), so a strict `===`
+                    //     would reject every legal value on such a field.
+                    const normalize = (v: unknown): string => String(v).trim().toLowerCase();
+                    const currentValue = normalize(this.Value);
+                    if (!allowedValues.some(v => normalize(v.Value) === currentValue)) {
```
AN-BC: This is wasted CPU. What we should do is have an instance level private member called `_allowedEntityFieldValues` or similar, and this instance member should be populated the first time this method is called - e.g. lazy populated. Thereafter, you can simply use `this._allowedEntityFieldValues` instead of recomputing this on every single validation of the field. 

Actually, this should be placed on the `EntityFieldInfo` class instance and put the entire logic for noramlization and `value in list` checking inside a helper method on the class instance of the `EntityFieldInfo` class. The reason is that is shared amongst **all** instances of the entity field. So if you have 1,000 instances of a given entity field across different `BaseEntity` instances, you'd only have to initialize this normalized `_allowedEntityFieldValues` array once. 

While a very small # of CPU cycles is needed for this, it is an easy optimization, cleaner to encapsulate into the EntityFieldInfo class, and at scale (think particularly situations where we are importing 1000s or millions of rows) can add up.

AN-BC (commented):
@MarceloT-BC  good catch. To-date this has been enforced by convention since we use strong typing in nearly all cases and we therefore do not typically consider going through `entity.Set` to be good code. However, it is an allowed path architecturally, so this additional guard is a smart improvement. 

I have a couple of optimizations here that I'd like you to implement.

Actually @claude  please push a new commit on this branch to implement these changes, we'll see if Claude Web can knock out these minor tweaks.

I'd like one extra set of eyes on this after that next change so please ask @rkihm-BC  to review this once those changes are in @MarceloT-BC.

Thanks.

packages/MJCore/src/generic/baseEntity.ts
MarceloT-BC: Done in c90ac48 — it now logs, and permits.

`ValueListType='List'` with no `EntityFieldValue` rows emits a `LogError` naming the entity and field and saying what to do about it (re-run CodeGen, or set `ValueListType='None'`), then permits the value rather than failing every save on the field.

One deliberate deviation to flag: **the report is latched per field, not fired per call.** The value list lives on `EntityFieldInfo`, which is shared metadata, so an import of a million rows through a field in this state would otherwise emit a million identical errors — the same waste your other comment is about. Latched, it still appears the first time the field is validated in every process, so it is loud without being a flood. Happy to make it unconditional if you'd rather; it's a one-line change.

Your CodeGen point is noted but not in this PR — a `List` field with no values is exactly the kind of thing `manageEntityFieldValuesAndValidatorFunctions` could flag at generation time, where it's a build-time warning instead of a runtime surprise. Say the word and we'll file it separately rather than widening this one.

packages/MJCore/src/generic/baseEntity.ts
MarceloT-BC: Done in c90ac48, on `EntityFieldInfo` as you specified.

- `EntityFieldInfo.ValueIsPermittedByValueList(value)` now owns the whole rule — mode check, value-type gate, normalization and membership — with the reasoning documented there instead of inline in `Validate()`.
- Backing it is a lazily-built, memoized `Set<string>` of normalized legal values, so lookup is **O(1)** rather than a `.some()` scan per validation, and it is built **once per field** rather than once per record. `NormalizeValueListValue()` is a static so the normalization rule has one home.
- `ValueListValuesForDisplay` is memoized too — a bad bulk load fails on the same field repeatedly and the message never changes, so the `join` shouldn't be rebuilt each time either.
- `EntityField.Validate()` is now one call plus the error it reports.

The guarantee is pinned by a test rather than asserted in a comment: 50 validations across 50 `EntityField` instances sharing one `EntityFieldInfo` read `EntityFieldValues` **exactly once** (`builds the normalized value set once per FIELD, not once per record`).

Suite: value-list file 25 tests (was 21), full MJCore 2130/2130, `tsc` clean, and CodeGenLib / MetadataSync / InteractiveComponents re-run green.

One note for the record: these tweaks were pushed by **Claude Code** working in @MarceloT-BC's workspace, not Claude Web — so the same "needs a reviewer who understands the context" caveat at the top of the PR description applies to this commit too.

packages/MJCore/src/generic/entityInfo.ts
```diff
@@ -988,6 +988,14 @@ export class EntityFieldInfo extends BaseInfo {
     _RelatedEntityNameFieldIsComputed: boolean
     private _rawEntityFieldValues: Record<string, unknown>[] | null = null;
     private _entityFieldValuesConstructed = false;
+    /** Memoized normalized membership set for {@link ValueIsPermittedByValueList}. */
+    private _normalizedValueListValues: Set<string> | undefined = undefined;
+    /** Memoized message form of the value list, built on first validation failure. */
+    private _valueListValuesForDisplay: string | undefined = undefined;
```
rkihm-BC: **This one I'd fix before merge — it's the only thing here that changes shipped behavior.**

`_x` + a public getter `X` is exactly MJ's backing-field convention, and `BaseInfo.toJSON` walks `Object.keys(this)`, finds each `_`-prefixed own property, and **emits it through its public getter**. `useDefineForClassFields: false` in `tsconfig.server.json` means the `= undefined` initializer becomes a real own property at construction — I compiled it to be sure:

```js
constructor() {
    this._normalizedValueListValues = undefined;
    this._valueListValuesForDisplay = undefined;
    ...
}
```

So the property is always an own key, and the getter is always found. Running the real `EntityInfo`/`EntityFieldInfo` from `dist` with this getter grafted on:

```
field "Status" -> ValueListValuesForDisplay = "Active, Inactive"
field "Name"   -> ValueListValuesForDisplay = ""
key present on non-value-list field: true
_normalizedValueListValues leaked: false
```

`EntityInfo` doesn't override `toJSON`, so the chain `AllMetadata → EntityInfo → Fields → EntityFieldInfo` reaches it — and `ProviderBase.SaveLocalMetadataToStorage()` does `JSON.stringify(this._localMetadata)`. Result: every field of every entity in the browser metadata cache now carries a redundant join of `EntityFieldValues` (which is already in the payload), plus `"ValueListValuesForDisplay":""` on the ~5,736 MJ core fields that have no value list. Serialization also *invokes* the getter on every field, forcing `EntityFieldValues` hydration and memoizing a string that exists only for error messages.

This is payload bloat, not a correctness bug — so I'd call it "fix it because it's one line with no downside," not a blocker.

**Fix:** rename the memo to something that doesn't shadow the getter — `_valueListDisplayCache`. No behavior change, and nothing round-trips it since it's a pure memo. Worth a one-line comment saying *why*, because the obvious name is the wrong one here.

`_normalizedValueListValues` and `_loggedEmptyValueList` are safe only because no getters of those names exist — that's luck, not design, so the comment will earn its keep.

packages/MJCore/src/generic/entityInfo.ts
```diff
@@ -1071,6 +1079,141 @@ export class EntityFieldInfo extends BaseInfo {
...
 
+    /**
+     * Upper bound on how many legal values a validation message enumerates before it truncates.
+     * A long list would otherwise produce an error message no user can read.
+     */
+    public static readonly MaxValueListValuesInErrorMessage: number = 25;
+
+    /**
+     * Normalizes a value for comparison against a value list: stringified, trimmed, lower-cased.
+     *
```
rkihm-BC, line 1100: This paragraph — and the PR body's stronger claim that trimming "**cannot** be tightened — that is the 9,301 rows" — doesn't survive contact with the code.

MJ already strips that padding one layer down, in `EntityField.Value`'s setter (`baseEntity.ts:141-143`):

```ts
if (typeof value === 'string' && this._entityFieldInfo.FixedWidthColumn) {
    value = value.replace(/ +$/, '');
}
```

`Validate()` reads `this.Value`, and `BaseEntity.Validate()` → `this.Fields` → `hydrateFieldsIfNeeded()`, which assigns **through that setter** (`baseEntity.ts:3047`). I checked the type classification too — `IsFixedWidthStringSQLType` returns `true` for `nchar`/`char`/`bpchar`/`character` and `false` for `nvarchar` — so every column in the 9,301-row story hits the rtrim before the comparison ever sees it. `MJ: Action Params`.Type really is `nchar(10)`, and its padding really is gone by then. The measurement was taken over raw SQL rows, not through the ORM.

**Keep the `.trim()`** — it's free and still covers leading whitespace and metadata-side stray spaces. But the comment states something the code contradicts, and this is 20 lines of reasoning a future maintainer will use to make a decision. Please correct the rationale (and the changeset — see my comment there, since that text ships in public release notes).

packages/MJCore/src/generic/entityInfo.ts
```diff
@@ -1071,6 +1079,141 @@ export class EntityFieldInfo extends BaseInfo {
...
+     * Every part of that is deliberate, and two of the three are REQUIRED rather than defensive:
+     *   * **Trimming is required.** MJ's own core entities store value-list values in fixed-width
+     *     `nchar(n)` columns, so the value read back is space-padded while the metadata value is
+     *     not. Measured on a current 6.x instance database, an untrimmed comparison would have
+     *     rejected 9,301 existing rows — `MJ: Action Params`.Type (`"Input     "`),
+     *     `MJ: Record Changes`.Status (`"Complete       "`), `MJ: Entity Relationships`.Type,
+     *     `MJ: Authorization Roles`.Type and `MJ: Users`.Type — i.e. it would break saves on MJ
+     *     core itself. Every one of those mismatches was padding alone; none was a case difference.
+     *   * **Stringifying is required.** `EntityFieldValue.Value` is always a string in metadata
+     *     while the field's runtime value may be a number (a numeric value list), so a strict
```
rkihm-BC: **Answering the open question from the PR description: keep it case-insensitive.** Not merely "defensive" — load-bearing on MJ core itself.

The live-database row scan was the wrong instrument, because it measures stored rows and misses the **default-value path**. `EntityField`'s constructor does `this.Value = fieldInfo.DefaultValue` when no value is supplied, so a new record carries the literal SQL default into validation.

Scanning the generated ORM for `List` fields whose default differs from its value list by case alone: 541 exact matches, and **2 case-only mismatches**, both on `MJ: Entity AI Actions`:

```
TriggerEvent  nchar(15)  Default: After Save   list: after save, before save
OutputType    nchar(10)  Default: FIeld        list: entity, field
```

Under a case-sensitive comparison, creating either record **at its database default** fails validation. `.toLowerCase()` is what makes those work today.

No code change needed — I'd just fold this evidence into the comment (or the case-variant test) so the next person doesn't re-open it. The wording "No row in that same database needs it" is what invites the re-litigation; it's true and still misleading.

packages/MJCore/src/generic/entityInfo.ts
```diff
@@ -1071,6 +1079,141 @@ export class EntityFieldInfo extends BaseInfo {
...
+     *     mode, not the column type), but in practice every one is a string column: measured on a
+     *     current 6.x instance, 455 nvarchar + 7 nchar and nothing else, which follows from
+     *     CodeGen's constraint parser only ever extracting quoted literals. `number` is admitted
+     *     because the generated union type anticipates a non-quoted list via `NeedsQuotes`.
+     *     Booleans and Dates are excluded deliberately: a bit column carrying a `'1'`/`'0'` list
+     *     would see `String(true) === 'true'` and reject every legal value, and a Date has no sane
+     *     string form to compare — so guessing there would break saves rather than guard them.
+     *
+     * @param value the field's current runtime value
+     * @returns true when the value is permitted, INCLUDING when the rule does not apply
```
rkihm-BC, line 1181: One scope point neither the PR body nor this comment block mentions, worth @AN-BC's eye rather than a code change.

The safety argument throughout is "MJ never refuses what the database would accept." That holds for CHECK-derived lists. It does **not** hold for the other producer: `applyValueListConfig` (`manage-metadata.ts:4390+`) applies `SoftFieldValueListConfig` from `additionalSchemaInfo`, produced by DBAutoDoc's enum detection (`AdditionalSchemaInfoGenerator.collectValueListFields`) — an **LLM verdict** (`type: 'List' | 'ListOrUserEntry'`) over sampled column data, default confidence threshold 85, `confirmedOnly` defaulting to **false**. Those fields have no CHECK constraint, so there is no database floor: this rung converts a probabilistic guess into a hard save refusal for any value the model didn't sample.

I initially rated this higher and then walked it back — `applyValueListConfig` opens with `const config = ManageMetadataBase.getSoftPKFKConfig(); if (!config) return true;`, so it's opt-in and only fires for projects that ran DBAutoDoc and wired the config. That's a much narrower blast radius than I first thought, and it's plausibly the intended semantics (`List` means closed set; DBAutoDoc should emit `ListOrUserEntry` when unsure).

So: not a merge gate, but the doc block here enumerates four boundaries as if CHECK constraints were the only source, and they no longer are. A sentence noting the second producer would keep this comment honest.

packages/MJCore/src/generic/entityInfo.ts
```diff
@@ -1071,6 +1079,141 @@ export class EntityFieldInfo extends BaseInfo {
...
+     */
+    public ValueIsPermittedByValueList(value: unknown): boolean {
+        if (this.ValueListTypeEnum !== EntityFieldValueListType.List) {
+            return true;
```
rkihm-BC, line 1185: Nit, take it or leave it: this type gate returns *before* the empty-value-list check below, which makes the broken-metadata `LogError` value-dependent. A `List` field with no `EntityFieldValue` rows whose column happens to always hold null (or a boolean) never reports at all.

Moving the empty-list check above this gate makes the diagnostic value-independent; the cost is one memoized set build on a field that would build it anyway. Given @AN-BC asked for this to be *loud*, it's the difference between "loud" and "loud if someone happens to set a string."

Separately, the latch is per-`EntityFieldInfo` instance, so it re-arms after a metadata refresh. That's the right behavior for the "loud" requirement, but "latched" reads as "once ever" — worth saying so in the comment on line 1167.

.changeset/value-list-field-validation.md
```diff
@@ -0,0 +1,9 @@
...
+
+An out-of-list value now fails `Validate()` with a field-named error naming the legal values, the way nullability and MaxLength already do. Three boundaries keep it safe: `ListOrUserEntry` is never checked (that mode exists to permit free text), an empty value list never rejects anything (missing metadata is not a rule), and null/blank stays the nullability check's job so one mistake yields one error. The comparison is trimmed, case-insensitive and string-based on purpose — MJ's own core entities store value-list values in `char(n)` columns, so an exact comparison would reject padded values that are already in the database, and `EntityFieldValue.Value` is always a string while a numeric value list's runtime value is a number.
```
rkihm-BC: Same correction as my comment on `NormalizeValueListValue`, and it matters more here because changeset text ships verbatim in public release notes.

> MJ's own core entities store value-list values in `char(n)` columns, so an exact comparison would reject padded values that are already in the database

`EntityField.Value`'s setter already rtrims fixed-width columns before `Validate()` ever reads the value, so the padding never reaches the comparison. The `EntityFieldValue.Value` side is `nvarchar` and unpadded too. Trimming is worth keeping, just not for this reason.

The stringification half of that sentence is accurate and should stay. This paragraph also still describes the pre-refactor design (the rule now lives on `EntityFieldInfo`, not inline in `EntityField.Validate()`), so it's due a refresh either way.

rkihm-BC (commented):
Reviewed at `c90ac48` per @AN-BC's request. **Net assessment: the change is correct and well-placed, the refactor onto `EntityFieldInfo` was executed properly, and I found nothing that should stop the merge.** One line of code, some prose, and one decision.

I verified the premise rather than taking it on faith: `parseCheckConstraintValues` anchors its structure regex with `^…$`, so a value list is only ever produced when the constraint is *exactly* a disjunction of equalities (optionally `OR … IS NULL`) — anything else returns `null` and falls through to LLM-generated `Validate()`. The lists this rung enforces really are exhaustive with respect to their CHECK.

Particularly good: every ambiguous case fails open; null stays the nullability rung's job so one mistake yields one message; and the "built once per field, not once per record" claim is pinned by a test that spies the `EntityFieldValues` getter across 50 instances rather than asserted in a comment. That's the right way to prove a caching guarantee.

### On CI — don't block on it

The deterministic integration tier is red on this PR, but it is **red on `next` too** — the last five runs on `next`, including current HEAD `e45fde5b`, are all `failure`. The failing oracle is `IT50 - CodeGen Artifact Consistency` (`ParentIDDepth`/`ParentIDPath`/`ParentIDIsLeaf`/`ParentIDChildCount`, plus a stale `MJFormChromeRule` artifact) — fallout from the just-merged hierarchy/entityfield-configuration work, unrelated to value lists.

Read the other way, the run is **evidence for this PR**: 63 passed / 1 failed / 5 skipped, with `IT27 - Core Entity Write-Side`, `IT40 - Entity-Server Invariants` and `IT47 - Transaction Group Atomicity` all green. Real MJ core saves now traverse the new rung without tripping it.

### Summary of inline comments

| | Item | Severity |
| --- | --- | --- |
| 1 | `_valueListValuesForDisplay` collides with the `BaseInfo` serialization convention — new key emitted on every serialized field | Only real defect; one-line fix |
| 2 | The "trimming is REQUIRED / 9,301 rows" rationale is contradicted by `EntityField.Value`'s existing rtrim | Doc-only (comment + changeset) |
| 3 | Open question answered: **keep case-insensitive** — evidence below | Decision, no code change |
| 4 | `ValueListType='List'` no longer implies a CHECK constraint (DBAutoDoc soft lists) | Heads-up + doc sentence |
| 5 | The broken-metadata `LogError` is value-dependent | Nit |

Smaller things not worth inline comments: the PR body says `EntityField.MaxValueListValuesInErrorMessage` (it's on `EntityFieldInfo` post-refactor); numeric lists compare stringified, so `String(1.0)` is `'1'` and a metadata value of `'1.0'` would fail closed — speculative, since no numeric value lists exist today, so a sentence in the comment beats code; and `MJ: Action Params`.ValueType has duplicate values in its `EntityFieldValues` today, so the error message will list duplicates and `(N total)` double-counts (pre-existing data issue, `[...new Set(values)]` in the display getter if you care).

Separate follow-ups worth filing, not for this PR: the CodeGen-side check for `List`-with-no-values that @AN-BC raised (a build-time warning beats a runtime one), and the `MJ: Entity AI Actions` metadata inconsistency surfaced in comment 3 — its generated union is `'after save' | 'before save'` while its DB default is `'After Save'`, so a typed read returns a value outside its own declared union.

MarceloT-BC:
Two updates on this branch, plus one thing that belongs to CodeGen rather than here.

### Date value lists are now validated (5627531)

I went looking for which `IN (…)` shapes can actually reach this rung, and measured it rather than assuming: a temp table with one `IN`-list CHECK per data type, read back from `sys.check_constraints`, each definition then run through CodeGen's own `parseCheckConstraintValues`.

| Column type | What SQL Server stores | Parser result |
| --- | --- | --- |
| `nvarchar` / `nchar` (≥2 values) | `([S]='Inactive' OR [S]='Active')` | ✅ value list |
| `date` (≥2 values) | `([D]='2026-07-01' OR [D]='2026-01-01')` | ✅ value list |
| `uniqueidentifier` (≥2 values) | quoted GUIDs | ✅ value list |
| `int` / `money` / `decimal` | `([L]=(3) OR [L]=(2) OR [L]=(1))` | ❌ not parsed |
| `bit` | `([B]=(1) OR [B]=(0))` | ❌ not parsed |
| any type, **single** value | `([T]='2026-01-01T00:00:00')` | ❌ not parsed |

So a **`date` value list is reachable through plain CodeGen** — it was the one live gap, since the runtime value is a `Date` and the rung skipped it. A `Date` is now compared on its **calendar date**, accepting either its UTC or its LOCAL reading. That asymmetry is deliberate: a value read back from SQL Server arrives as UTC midnight, while app code building `new Date(2026, 6, 1)` produces LOCAL midnight — whose UTC date is the previous day west of Greenwich. Insisting on one representation would refuse legal values for half the world. The cost is that an adjacent day can slip through when both are in the list, which beats a false failure.

Columns carrying a **time** component are deliberately skipped and reported: the metadata literal has no timezone, so deciding whether `'2026-01-01T00:00:00'` is the same instant as the stored value means guessing the database's interpretation, and guessing wrong rejects a legal value. Same for a `date` field whose list isn't entirely `yyyy-mm-dd` literals.

### Unsupported value types are reported rather than skipped in silence (2b6b161)

Same objection as your empty-value-list point, applied to the type gate: a boolean/object value on a `List` field was skipped quietly, which leaves a caller believing a guard is on when it isn't. It now logs once per field (naming the `typeof` and the SQL type) and still permits — comparing a boolean against a `'1'`/`'0'` list would see `String(true) === 'true'` and reject every legal value.

### Filed separately: #3978 — numeric/bit lists never become value lists

The table above surfaced a CodeGen-side gap that isn't mine to fix here: `CHECK (Level IN (1,2,3))` produces **no** `EntityFieldValue` rows, because SQL Server renders it unquoted (`[L]=(3)`) and the parser's regexes require a quoted literal. The field therefore loses its **Explorer dropdown** as well as its validation, and with AI codegen off the constraint yields nothing at all — no list, no generated validator. (The `+` on the OR group also means a **single**-value `IN` list is never captured, for any type.) Details, the repro and a suggested fix are in **#3978**; the runtime side here already compares numbers by value rather than by JS type, so a numeric list works the day CodeGen starts producing one.

Suite: value-list file **32 tests**, full MJCore **2137/2137**, `tsc` clean, downstream CodeGenLib / MetadataSync / InteractiveComponents green.

packages/MJCore/src/generic/entityInfo.ts
MarceloT-BC: Fixed in 97ac8b9 — you're right, and I verified the mechanism rather than taking the rename on faith: `BaseInfo.toJSON` walks `Object.keys(this)`, and for a `_`-prefixed key it derives the PascalCase name, walks the prototype chain for a descriptor with a `get`, and emits `result[targetKey] = self[targetKey]`. So the getter is found and invoked for every field.

Renamed to `_valueListDisplayCache`, with a comment on the naming trap — as you say, `_x` + `get X` is the convention *everywhere else*, so the next person will reach for exactly the name that breaks it. The comment also covers why `_normalizedValueListValues`, `_valueListDateKeys` and the two latches are safe (no getter of that name), since that is luck rather than design.

Added a regression test asserting `toJSON()` carries neither `ValueListValuesForDisplay` nor the memo key, so the trap can't be reintroduced silently.

packages/MJCore/src/generic/entityInfo.ts
MarceloT-BC: You're right, and I've corrected it rather than softened it (97ac8b9).

I confirmed the path you describe: `EntityField`'s value setter rtrims when `FixedWidthColumn` is true (`IsFixedWidthStringSQLType` → `nchar`/`char`/`bpchar`/`character`), and `hydrateFieldsIfNeeded` assigns **through** that setter, so `'Input     '` is already `'Input'` before `Validate()` reads it. My 9,301-row measurement was taken over raw SQL rows via `sqlcmd` — real data, wrong layer. It never described this code path, and stating it as the load-bearing reason was overclaiming.

The comment now says what is actually true: the trim is kept because it covers **leading** whitespace, stray spaces on the metadata side, and a caller assigning a padded value directly — none of which the setter handles — and it explicitly records that the earlier rationale was measured over raw rows and does not apply, so nobody re-derives it. The changeset is rewritten too (your point about release notes is well taken).

packages/MJCore/src/generic/entityInfo.ts
MarceloT-BC: This is the most valuable finding in the review — the default-value path is a hole in my instrument, not just a gap in the wording. Folded into the comment and pinned by a test (97ac8b9).

I confirmed it independently against a live 6.1 database, and it's worse than the ORM scan showed. Scanning `EntityField.DefaultValue` for `List` fields whose default is not an exact list member (normalizing SQL Server's `(N'…')` wrapper, which `ExtractActualDefaultValue` strips at populate time, so the runtime value really is the bare string):

| Entity | Field | Default | Value list | Exact | Case-insensitive |
| --- | --- | --- | --- | --- | --- |
| MJ: Entity AI Actions | TriggerEvent | `After Save` | before save \| after save | ✗ | ✅ |
| MJ: Entity AI Actions | OutputType | `FIeld` | entity \| field | ✗ | ✅ |
| MJ: Audit Logs | Status | `Allow` | Success \| Failed | ✗ | ✗ |
| MJ: Authorization Roles | Type | `grant` | Allow \| Deny | ✗ | ✗ |
| MJ: Queue Tasks | Status | `Pending` | In Progress \| Completed \| Failed | ✗ | ✗ |
| MJ: Record Merge Logs | ProcessingStatus | `Pending` | Complete \| Started \| Error | ✗ | ✗ |
| MJ: Version Installations | Type | `System` | New \| Upgrade | ✗ | ✗ |
| MJ: Conversation Details | Role | `user_name()` | AI \| User \| Error | ✗ | ✗ |

Your two are the case-only ones, so **case-insensitivity is load-bearing** and the wording is fixed. The other six are a separate matter: those defaults violate their own CHECK constraint, which I verified directly —

```
QueueTask.Status            CHECK ([Status]='In Progress' OR ...='Completed' OR ...='Failed')   DEFAULT (N'Pending')
AuditLog.Status             CHECK ([Status]='Success' OR ...='Failed')                          DEFAULT (N'Allow')
VersionInstallation.Type    CHECK ([Type]='New' OR ...='Upgrade')                               DEFAULT (N'System')
```

So a create that relies on the default is refused by SQL Server **today**, and this rung converts that into a field-named validation failure instead of a raw constraint violation — earlier and better, not a regression. But it does mean the first visible symptom of those stale defaults will now be this rung's message, which is worth knowing before it shows up in someone's log. Happy to file it (schema-side: the defaults, not this code) if you agree it's worth its own issue.

packages/MJCore/src/generic/entityInfo.ts
MarceloT-BC: Good catch, and the doc block now says it (97ac8b9). I'd read `applyValueListConfig` while tracing where `ValueListType` gets set and did not carry the consequence through to the safety argument — which is exactly the kind of thing a comment enumerating "four boundaries" makes look settled.

The block now states plainly that there are **two producers** and only one has a database floor: a CHECK-derived list can only move a failure earlier, whereas a DBAutoDoc-derived list has no CHECK behind it, so this rung turns a sampled, confidence-scored LLM verdict into a hard save refusal for any value the model did not see. It also records that it is opt-in (`getSoftPKFKConfig()` returns early without the config) and that `List` vs `ListOrUserEntry` is presumably the intended lever — so the reader gets the caveat and the reason it is narrow, rather than either alone.

Agreed it isn't a merge gate. If @AN-BC wants the stricter posture, the natural place is DBAutoDoc emitting `ListOrUserEntry` below some confidence rather than this rung second-guessing metadata it can't distinguish.

packages/MJCore/src/generic/entityInfo.ts
MarceloT-BC: Taken, both parts (97ac8b9).

The empty-value-list check now runs **before** the null and type gates, so the diagnostic is value-independent — you're right that "loud" shouldn't mean "loud if someone happens to set a string", which is precisely what @AN-BC asked it not to be. The cost is one memoized set build on a field that would build it on its next non-null validation anyway. Test added: a `List` field with no values whose value is `null` now reports exactly once.

Worth noting the ordering had already partly moved in `5627531` (the date commit reordered it above the type gates), so what landed here is the last step — above the null gate too.

And the latch wording is fixed: the comment now says once per `EntityFieldInfo` instance, and therefore re-armed by a metadata refresh rather than once ever.

.changeset/value-list-field-validation.md
MarceloT-BC: Rewritten in 97ac8b9 — and thank you for flagging the release-notes angle specifically, since that text outlives the PR.

The changeset no longer justifies trimming with the padding story. It now says trimming covers leading whitespace and metadata-side stray spaces, and states explicitly that trailing padding on fixed-width columns is already stripped by the field's value setter before validation sees it. The stringification half is kept and unchanged, as you suggested.

It also no longer describes the pre-refactor design: it now says the rule lives on `EntityFieldInfo.ValueIsPermittedByValueList()` with a set built once per field rather than once per record, documents the four boundaries as they now stand (including the reported-not-silent cases), and carries the case-insensitivity rationale in its corrected form — the two MJ core defaults that match their list by case alone.

MarceloT-BC:
@rkihm-BC — thank you, that was a genuinely useful review. **All five items are addressed and pushed in `97ac8b9`**, with replies on each thread. Summary:

| | Item | Status |
| --- | --- | --- |
| 1 | `_valueListValuesForDisplay` shadowed its getter → serialized on every field | **Fixed** — renamed `_valueListDisplayCache`, naming trap documented, `toJSON()` regression test added |
| 2 | The "trimming is REQUIRED / 9,301 rows" rationale | **Corrected** — in the comment, the changeset, and the PR body. You were right: the measurement was over raw SQL rows, and `EntityField`'s setter rtrims fixed-width columns before `Validate()` reads them |
| 3 | Keep case-insensitive | **Adopted, with your evidence in the code** and a test at the `MJ: Entity AI Actions` shape |
| 4 | `ValueListType='List'` no longer implies a CHECK | **Documented** — two producers, only one with a database floor, plus the opt-in caveat |
| 5 | The broken-metadata `LogError` was value-dependent | **Fixed** — empty-list check now runs before the null *and* type gates; latch wording corrected to "per `EntityFieldInfo`, re-armed by a refresh" |

Also took two of your smaller notes: the display list is de-duplicated (`MJ: Action Params`.ValueType really does carry repeats, so `(N total)` was double-counting), and the lossy `String(1.0) === '1'` case is recorded in the comment rather than solved, since CodeGen can't currently produce a numeric list at all (#3978). The PR body's `EntityField.MaxValueListValuesInErrorMessage` reference is corrected to `EntityFieldInfo`.

**One thing your item 3 turned up that goes beyond the review.** Chasing the default-value path, I scanned every `List` field whose `DefaultValue` is not an exact list member. Your two case-only mismatches are there — and so are **six that match their list not at all**:

```
QueueTask.Status            CHECK ('In Progress' OR 'Completed' OR 'Failed')   DEFAULT (N'Pending')
AuditLog.Status             CHECK ('Success' OR 'Failed')                      DEFAULT (N'Allow')
VersionInstallation.Type    CHECK ('New' OR 'Upgrade')                         DEFAULT (N'System')
```
plus `MJ: Authorization Roles`.Type (`grant` vs Allow|Deny), `MJ: Record Merge Logs`.ProcessingStatus (`Pending` vs Complete|Started|Error) and `MJ: Conversation Details`.Role (`user_name()` vs AI|User|Error).

Those defaults **violate their own CHECK constraints**, so a create relying on one is refused by SQL Server today; this rung turns that into a field-named validation failure instead of a raw violation — earlier and better, not a regression. But it does mean this rung's message is where those stale defaults will start surfacing. @AN-BC — worth its own issue (schema-side: the defaults, not this code)?

**On the two red checks:** neither is this PR. `Run unit tests` fails identically on every open PR right now (#3974, #3973, #3968) — `@memberjunction/core` imports `mssql` without declaring it (`metadata.live-db.integration.test.ts`, added 2026-08-18), which pnpm's strict resolution refuses in CI. **#3977 is already fixing exactly that.** The deterministic integration tier is the `IT50` CodeGen-artifact failure you diagnosed as pre-existing on `next`. Locally the tier this PR touches is green: value-list file **36 tests**, full MJCore **2141/2141**, `tsc` clean, and CodeGenLib / MetadataSync / InteractiveComponents re-run green.

packages/MJCore/src/generic/entityInfo.ts
```diff
@@ -1071,6 +1089,258 @@ export class EntityFieldInfo extends BaseInfo {
...
+     *
+     * Only `date` columns are compared. A column carrying a time component is deliberately skipped:
+     * the metadata value has no timezone, so deciding whether `'2026-01-01T00:00:00'` is the same
+     * instant as the value read back would mean guessing the database's interpretation, and guessing
+     * wrong rejects a legal value.
+     *
+     * For a `date` column the comparison accepts EITHER the value's UTC calendar date or its LOCAL
+     * one. That asymmetry is deliberate: a value read back from SQL Server arrives as UTC midnight,
+     * while application code that builds a date with `new Date(2026, 6, 1)` produces LOCAL midnight —
+     * whose UTC calendar date is the previous day west of Greenwich. Insisting on one representation
```
rkihm-BC, line 1255: Direction is inverted here (and in the matching PR comment): local midnight's UTC calendar date is the previous day **east** of Greenwich, not west.

```
UTC-5  (New York, west):  local 2026-07-01 00:00 -> 2026-07-01 05:00Z  -> same day
UTC+13 (Auckland, east):  local 2026-07-01 00:00 -> 2026-06-30 11:00Z  -> previous day
```

The west-of-Greenwich case is the mirror phenomenon — a value read back as UTC midnight has a *local* calendar date of the previous day (`2026-07-01T00:00Z` is `2026-06-30 20:00` in New York). Both are real, and checking both readings handles both; the sentence just attributes one to the other.

No functional impact — the code is right, only the justification names the wrong hemisphere. Worth fixing precisely because this comment exists to explain why the two-reading comparison is necessary, and a reader who checks the claim will find it doesn't hold and may conclude the leniency is unnecessary.

Small meta-note, offered as an observation rather than a complaint: this is the third claim in this PR's prose to come out inverted or measured at the wrong layer — the 9,301 padded rows, the case-scan that missed the default-value path, and now this. The code has been right every time and the corrections have been prompt and honest, so the pattern is specifically in the *narrative* around measurements: each one was accurate about what it measured and wrong about what it implied. Worth a habit of asking "which layer does this number describe?" before it becomes the load-bearing sentence in a comment.

rkihm-BC (approved):
Approving at `97ac8b9`. All four items from my review are addressed, and I verified each rather than taking the commit messages for it.

### Verified

**1 — serialization collision.** Renamed to `_valueListDisplayCache`. I checked all five new private names against the prototype — `_normalizedValueListValues`, `_valueListDisplayCache`, `_loggedEmptyValueList`, `_loggedUnsupportedValueListType`, `_valueListDateKeys` — **no getter collisions**, so nothing leaks through `BaseInfo.toJSON`. The regression test pins it and the comment documents the trap at the declaration site, which is where the next person will meet it.

**2 — trimming rationale.** Corrected rather than softened, in both the comment and the changeset. Stating outright that trailing padding is stripped by the field's value setter before validation sees it is the right fix; the stringification half correctly stayed.

**3 — case-insensitivity.** Answered and pinned at the default-value shape. Going back to the live DB and scanning `EntityField.DefaultValue` was the right instinct, and it turned up more than my ORM scan did — six further `List` fields whose SQL default violates its own CHECK today (`MJ: Audit Logs`.Status defaulting to `Allow` against `Success | Failed`, `MJ: Queue Tasks`.Status to `Pending` against `In Progress | Completed | Failed`, and so on). Your reading is right: those creates are already refused by SQL Server, and this rung converts a raw constraint violation into a field-named message. Please do file the schema-side issue — the defaults are the bug, not this code, and this rung is about to become the place people first notice them.

**4 — DBAutoDoc scope.** The block now states that there are two producers and only one has a database floor, plus the opt-in caveat. That is exactly the shape I wanted: the reader gets the risk and the reason it is narrow, rather than either alone. Agreed the lever belongs in DBAutoDoc's `List` vs `ListOrUserEntry` choice if @AN-BC wants a stricter posture.

**5 — the broken-metadata report.** Now above both the null and type gates, so it is value-independent, with a test at the `null` shape. Latch wording fixed to per-`EntityFieldInfo`, re-armed by a metadata refresh.

### CI — both jobs are red, and neither is yours

Worth stating explicitly so nobody blocks on the red:

- **Unit tests:** one suite fails to *collect* — `metadata.live-db.integration.test.ts` cannot import `mssql`. Identical on `next` (`1 failed | 141 passed`, 2105 tests). Here: `1 failed | 142 passed`, **2141 tests, all passing** — +1 file, +36 tests, zero regressions. Your local numbers match CI exactly.
- **Integration:** the same `IT50` oracles (CD3/CD5/CD6), same 63 passed / 1 failed / 5 skipped, unchanged from `next` at HEAD `e45fde5b`.

### On the two unrequested commits

`2b6b161` reads to me as a direct application of @AN-BC's own empty-list principle to the type gate rather than scope creep — skipping silently leaves a caller believing a guard is on, and it still permits the value.

`5627531` genuinely is new scope: ~70 production lines and a `feat:` in a bugfix PR, added after review. I checked whether it earns its place — **zero MJ core fields carry a value list on a date or datetime column**, so it is dead code today. I am not objecting: the shape is reachable through plain CodeGen (verifying the parser against `sys.check_constraints` per type was the right way to establish that rather than assuming), it fails open at every unclear point — non-`date` type, any non-ISO list value, Invalid Date — and it is well tested. But present value is nil, so whether it ships here or as a follow-up is @AN-BC's call, not a correctness question. `patch` remains the right bump either way, since `minor` is reserved for migration/metadata branches in this repo.

### Two nits, neither blocking

See the inline note on the timezone comment. The other: a bare ISO-*datetime* string assigned to a date-typed value-list field trips both the existing "is not a valid date" rung and this one, so one mistake yields two errors — against the PR's own principle, but it needs a date column with a value list, of which there are currently none. Not worth code.

Nice work on this one. The habit of measuring instead of assuming — the constraint-parser matrix by data type, the default-value scan, re-deriving the `toJSON` mechanism instead of taking the rename on faith — is what made the whole review cheap, and the places where a measurement turned out to describe the wrong layer got corrected in the code rather than argued down.

Issue #3971, opened 2026-08-20
BaseEntity has no delete-validation seam — every app hand-rolls Delete() overrides to explain a refusal


MarceloT-BC:
## Summary

`BaseEntity` has a save-validation seam — `Validate()` and `ValidateAsync()` — with a well-defined contract: return a `ValidationResult`, let the framework decide, get field-named errors in front of the user. **There is no equivalent for delete.** No `ValidateDelete()`, no `CanDelete()`, no `BeforeDelete` event that can veto.

So the only way to refuse a delete with an explanation is to override `Delete()` itself (`baseEntity.ts:4267`), check by hand, and return `false` before calling `super.Delete()`.

## Why that is worse than it sounds

**1. The message is lost.** `Validate()` returns `ValidationResult` with `ValidationErrorInfo[]`, each carrying a `Source` — which is how MJ's forms mark the offending *field*. A `Delete()` override returns `boolean`. There is nowhere to put "this template is referenced by 5 signed contracts"; the caller gets `false` and has to invent its own explanation, or the user gets whatever the raw FK error says.

**2. Every app reimplements it, differently.** In `bizapps-accounting` alone:

- `JournalEntryTypeEntityServer.Delete()`
- `JournalEntryLineEntityServer.Delete()`

…each with its own probe, its own logging, its own way of surfacing the reason. `bizapps-contracts` is about to add four more for the same reason (a provision referenced by modifications, a template referenced by signed contracts, a contract with lineage children, a type in use). That is six hand-rolled versions of a hook that should exist once.

**3. The guarantees `Validate()` gets, delete does not.** A save runs validation inside the framework's flow, so every caller — form, `SaveEntityGraph`, an agent, a Remote Operation — is covered by one rule. A `Delete()` override only covers callers that go through that subclass's method, and there is no framework guarantee about ordering relative to permissions, Record Changes, or soft-delete handling.

**4. Referential refusal is the common case, and it is always ugly.** Where FKs are `NO_ACTION` (the correct default), the database already refuses the delete — correctly. What is missing is the layer that turns `The DELETE statement conflicted with the REFERENCE constraint "FK_..."` into a sentence a user can act on. Today every app writes that layer by hand or ships the raw error.

## Proposed

A delete-side counterpart to the save seam, symmetrical with what already exists:

```typescript
// Called by BaseEntity.Delete() before any provider work, exactly as Validate()
// is called by Save(). A non-Success result aborts the delete.
[...]

Issue #3962, opened 2026-08-19
Form chrome: validation errors are never aggregated to the section/panel that owns the field


MarceloT-BC:
## Summary

A multi-section form (`left-nav` chrome, or accordion) gives the user no way to see **which section** a validation failure lives in. MJ routes errors down to individual fields but never aggregates them back up to the panel that owns the field, so on a form with a collapsed rail the user gets a rejection and has to go hunting through sections for the red field.

## What MJ does today

`BaseFormComponent.SaveRecord` broadcasts the failures onto `FormContext`:

```ts
this._showValidation = true;
this._validationErrors = valResults.Errors;
```

and each field self-selects its own out of that list, by name:

```ts
// packages/Angular/Generic/base-forms/src/lib/field/form-field.component.ts:311
return contextErrors.filter(e => e.Source === this.FieldName);
```

That is the whole routing story. `mj-collapsible-panel` has **no validation awareness at all** — no error count, no badge input, no styling hook — so a panel cannot show that something inside it (or inside its collapsed body) is failing. `resolve-form-chrome` / the rail items likewise have nothing to render from.

The information needed is already present on both sides: `ValidationErrorInfo.Source` carries the field name, and each `mj-collapsible-panel` declares its `SectionKey` and contains a known set of `mj-form-field`s. Nothing joins them.

## Why an app cannot solve this cleanly

Every app that wants a per-section indicator has to hand-maintain a `fieldName -> sectionKey` map. We wrote one and deleted it, because it is a duplicate of a fact the panels already own and it **drifts silently**: rename a panel's `SectionKey` and the map keeps compiling, keeps returning a section that no longer exists, and the indicator quietly points at the wrong rail item — or at the default. Nothing fails. That is a bad thing to ask every app to own, and it gets worse with whole-graph validation, where child failures arrive positionally (`Modifications[2].ContractTemplateProvisionID`) and a naive map sends all of them to the header section — the one place the user cannot fix them.

## Suggested direction

The join belongs where the ownership is already known — the panel/section layer — not in app code. Roughly:

1. Panels register the field names they render (they already receive `SectionKey` + host their `mj-form-field`s), so the coordinator can build `field -> SectionKey` with no app input.
2. `mj-collapsible-panel` grows an error-count/`HasErrors` state derived from `FormContext.validationErrors`, and the `left-nav` rail + accordion headers render it.
[...]

Issue #1431, opened 2025-10-03
Refactor codebase to use BaseEntityResult.CompleteMessage getter instead of Error/Errors/Message properties


jordanfanapour:
## Overview
The `LatestResult` object and other `BaseEntityResult` instances are currently painful to work with because error information can be scattered across multiple properties (`Message`, `Error`, `Errors`). A new `CompleteMessage` getter has been added to `BaseEntityResult` in the MJCore package to consolidate all error/message information into a single convenient property.

## Problem Statement
Currently, code that needs to display error messages must check multiple properties:
- `result.Message` - may contain a message string
- `result.Error` - may contain an error object or string
- `result.Errors` - may contain an array of error objects

This leads to:
- **Inconsistent error handling** across the codebase
- **Verbose, repetitive code** checking multiple properties
- **Missed error messages** when developers don't check all possible locations
- **Maintenance burden** with scattered error handling patterns

## Solution: CompleteMessage Getter
A new getter has been added to `BaseEntityResult` (in MJCore package):

```typescript
/**
 * Returns a complete message that includes the Message property (if present), 
 * the Error property (if present), and any Errors array items (if present).
 */
public get CompleteMessage(): string {
    let msg = undefined;

    // first check the message property
    if (this.Message && this.Message.trim().length > 0) {
        msg = this.Message;
    }   

    // now check the simple Error property
    if (this.Error) {
        msg = (msg ? msg + '\n' : '')
        if (typeof this.Error === 'string') {
            msg += this.Error;
        }
        else if (this.Error.message) {
            msg += this.Error.message;
        }
        else {
            msg += JSON.stringify(this.Error);
        }
    }
    
    // now check the Errors array
    if (this.Errors && this.Errors.length > 0) {
        // append
        msg = (msg ? msg + '\n' : '') + this.Errors.map(err => err.message || JSON.stringify(err)).join('\n');
    }

    return msg;
}
```

### Phase 1: Identify Usage Patterns
- [ ] Search codebase for `LatestResult.Message` usage
- [ ] Search codebase for `LatestResult.Error` usage
- [ ] Search codebase for `LatestResult.Errors` usage
- [ ] Search for other `BaseEntityResult` subclass instances with similar patterns
- [ ] Document all locations that need refactoring

### Phase 2: Update Code
- [ ] Replace complex error checking logic with `result.CompleteMessage`
- [ ] Update error logging statements
- [ ] Update user-facing error messages
- [ ] Update GraphQL resolvers
[...]

PR #3973, opened 2026-08-20
fix(MJCore): CompleteMessage rendered every validation error as a JSON blob

MarceloT-BC:
## The defect

`BaseEntityResult.CompleteMessage` rendered each entry of its `Errors` array as `err.message || JSON.stringify(err)` — **lowercase** `message`. MJ's own `ValidationErrorInfo` carries **`Message`** (`MJGlobal/src/ValidationTypes.ts:27`), so every one of them fell through to the JSON fallback.

That string is not incidental: it is what the server hands the client on a failed save. `ResolverBase.CreateRecord` puts it in the `GraphQLError`, and `SaveEntityGraphOperation` puts it in `ErrorMessage`. (`UpdateRecord`/`DeleteRecord` did **not** — see the second defect below, which this PR also fixes.) On the client, `GraphQLDataProvider.Save`'s catch copies it into `LatestResult.Message` (graphQLDataProvider.ts:2007) and `BaseFormComponent` toasts `'Save failed: ' + …` (base-form-component.ts:465). So the entire point of writing a careful, field-named refusal was defeated at the last step.

Nothing could catch it. `Errors` is typed `any[]`, so the compiler cannot see the casing; `JSON.stringify` never throws, so the runtime cannot either. It is visible only by reading a message a user received.

## How to recreate (before this change)

It needs **no application code** — MJ core's own `EntityField.Validate()` builds the same object. Against a running MJAPI, create a core `MJ: Companies` row with a 300-character `Name`:

```sh
curl -s -X POST http://localhost:<api-port>/ \
  -H "X-API-Key: $KEY" -H 'Content-Type: application/json' \
  -d '{"query":"mutation { CreateMJCompany(input:{Name:\"XXX…300 chars…\"}) { ID } }"}'
```

`errors[0].message` comes back as:

```
{"Source":"Name","Message":"Name cannot be longer than 50 characters. Current value is 300 characters","Value":"XXX…all 300 X's…","Type":"Failure"}
```

Note it also echoes the entire rejected value back at the user. After this change, the same mutation against the same database returns:

```
Name cannot be longer than 50 characters. Current value is 300 characters
```

## The fix

One shared helper, `BaseEntityResult.ErrorText()`, reading `Message` → `message` → `JSON.stringify`, applied to **both** the `Errors` array and the single `Error` property, which carried the same lowercase-only assumption. The JSON fallback is kept for a shape with neither field, so nothing that used to say something now says nothing.

Capital `Message` is preferred over lowercase because a `ValidationErrorInfo` has only the former and an `Error` only the latter — the order matters solely for an object carrying both, where the MJ-native field is the better answer.

Corroboration that this is the right field: **every other** reader of a `ValidationErrorInfo` in this repo already uses `.Message` — `base-form-component.ts:476`, `base-agent.ts:3401/4403`, `flow-graph-executor.ts:119`, `AIPromptRunner.ts:4331/5436`, `agent-spec-sync.ts:951`, the integration-test checks, and `baseEntity.ts:3470/4448` itself. `CompleteMessage` was the lone outlier. Nothing in the repo parses it as JSON, so no consumer depends on the broken shape (806 `CompleteMessage` call sites checked).

## The second defect, found in review: half of it only reached the user on a create

Only `CreateRecord` read `CompleteMessage`. `UpdateRecord` and `DeleteRecord` read the bare `Message`:

```ts
ResolverBase.ts:1248  CreateRecord   LatestResult?.CompleteMessage ?? 'Unknown error creating record'   ✅
ResolverBase.ts:1335  UpdateRecord   LatestResult?.Message         ?? 'Unknown error'                   ❌
ResolverBase.ts:1561  DeleteRecord   LatestResult?.Message         ?? 'Unknown error'                   ❌
```

A validation refusal leaves `Message` **null** — `_InnerSave`'s catch assigns `newResult.Message = e.message || null`, and a thrown `ValidationResult` has no lowercase `message` — so `?? 'Unknown error'` won every time and the reason was discarded *before* the getter this PR fixes was ever consulted. Measured through the real code path after a refused `Save()`:

```
LatestResult.Message         = null      →  ?? 'Unknown error'  =  "Unknown error"
LatestResult.CompleteMessage = "the rule refused, in prose"
```

One rule, two answers, depending on the verb: the same rule on the same entity explained itself on a create and said "Unknown error" on an update. That asymmetry is why it survived review — the create path is the one anybody tests first, and it was already right.

Both throws now read `CompleteMessage`. It is a **strict superset**: it seeds from `Message`, then appends `Error` and `Errors`, and still yields `undefined` when all three are empty — so nothing previously reported is lost, and the `'Unknown error'` fallback still fires rather than showing a blank error. The three error codes stay distinct.

`resolverBase.refusalMessage.test.ts` guards both directions of that risk and then **reads the resolver source** to pin the selection. That last part is necessity, not taste: any test that restates what a resolver reads stays green through this regression — which is exactly how it survived. Red-proven (it fails against `origin/next`, naming lines 1335 and 1561); MJServer vitest 5/5, `tsc && tsc-alias` exit 0.

## A duplication problem this makes visible — deliberately NOT fixed here

Making the reason legible exposes a second-order wart. `CompleteMessage` seeds from `Message` and then appends every `Errors` entry, and two producers in this same file build the former **out of** the latter — `baseEntity.ts:3475` (`_InnerSave`, IS-A parent failure) and `:4453` (`_InnerDelete`) — so their text now appears twice:

```
Failed to save parent entity 'Products': A; B
A
B
```

A de-duplicating guard in the getter (skip an entry whose text is already in the accumulated message) was written, tested and then **reverted**, because it is lossy in ways the reader cannot detect:

| case | without the guard | with it |
| --- | --- | --- |
| three fields failing with the same sentence | 3 lines | **1 line** — fix one, retry, fail again |
| two entries, one a substring of the other | both | both |
| **the same two, array order reversed** | both | **one silently dropped** |
| entry `Invalid input` + summary `Invalid input for Status` | both | **the distinct Status error vanishes** |

Rows 2 and 3 are the same failure set rendering differently by array order. Saying something twice is ugly; reporting one problem when there were three is precisely the bug class this PR exists to fix, and a renderer cannot distinguish a redundant repeat from a real second failure. If the repetition is worth removing it belongs at those two producers, where the redundancy is known rather than guessed — a separate change. The site now carries a comment recording that the duplication is deliberate.

## Verification## Verification

Authored on one instance and then **independently re-verified on another agent's pass**, with the repro re-derived from scratch rather than re-run.

| check | result |
| --- | --- |
| live MJAPI, real SQL DB, pre-fix build | JSON blob (above) |
| same mutation after restarting onto the fixed build | the prose (above) |
| `baseEntityResult.completeMessage.test.ts` vs **pristine** source | **6 of 8 red** — the 2 green are exactly the ones that should be (a plain `Error`, and the JSON fallback) |
| same file vs the fix | 8/8 |
| `resolverBase.refusalMessage.test.ts` | 5/5, red-proven (the source-scan case fails on `origin/next`, naming lines 1335 and 1561) |
| MJCore full suite | 2076/2076 across 138 files |
| MJCoreEntities | 605/605 |
| whole-repo vitest (891 files) | 880 passed, 10 skipped, 1 failed — **unrelated**: `MultiProviderCompliance` flags `hierarchy-tree.component.ts:329 new Metadata()`, pre-existing on the older base it ran against and already fixed on `next` |
| `tsc && tsc-alias` (MJCore) | exit 0 |

The test file pins the **output a user reads** rather than the mechanism, so it stays honest through a refactor. Its last case drives the real `BaseEntity.Save()` — a subclass `ValidateAsync` refuses, `_InnerSave` throws the `ValidationResult`, its catch copies `e.Errors`, and the test reads `LatestResult.CompleteMessage` — so the whole chain is pinned, not just the getter.

**Not covered:** no browser was driven. The toast path is confirmed by reading the code cited above, not by observing a rendered toast.

## Related issues

**No `Fixes #N`** — no issue was filed for any of the three defects. But they sit on top of a chain of
work that was each time recorded as finished, so the links matter:

- **#243** — *"If a save attempt fails, display an error to the user … Also need to make sure
  validation messages reach the client so that we can display those errors."* Opened 2024-05-30,
  **closed as completed 2025-04-21**. It was not: on a create the message reached the client as a
  JSON blob, and on an update or delete it did not reach the client at all. This PR is what actually
  delivers it. I have not reopened it — that is your call.
- **#246** (closed) — *Change `Save`/`Delete` to return a result object rather than a boolean*, opened
  explicitly to serve #243 so a refusal would have somewhere to carry its reason. That plumbing has
  been in place ever since; nothing at the end of it rendered the reason correctly.
- **#1431** (closed 2025-12-29) — *Refactor the codebase to use `BaseEntityResult.CompleteMessage`
  instead of `Error`/`Errors`/`Message`*, to stop "missed error messages when developers don't check
  all possible locations". Closed with the getter itself broken and with `UpdateRecord`/`DeleteRecord`
  never migrated. The migration is still unfinished elsewhere, which is why the de-duplication guard
  belongs in the getter rather than in each producer.
- **#3971** — *`BaseEntity` has no delete-validation seam.* Apps refuse deletes today by overriding
  `Delete()` and setting a reason on `LatestResult`; until this PR every one of those reasons became
  "Unknown error" at the API boundary.
- **PR #3974** — implements that seam, and is where the duplicate-rendering problem was found by
  cross-check. Merge order does not matter: neither breaks the other, and both are green as they
  stand. This PR fixes the duplication for every producer, including #3974's.
- **#3969** — *`BaseEntity` never validates value-list fields.* Adjacent: it would put more
  `ValidationErrorInfo`s through the renderer this PR fixes.
- **#3962** — *Form chrome: validation errors are never aggregated to the owning section/panel.* The
  other consumer of the same objects: it needs each error's `Source` for field-level routing, which is
  why `CompleteMessage` deliberately renders only the prose and leaves `Errors` intact for the form.

## Provenance

`git log -S` puts the lowercase read in `e9e0d3e900` (AN-BC, 2025-10-02, "Add `CompleteMessage` to `BaseEntityResult`"). `git blame` attributes the lines to Craig Adam's 2025-10-27 MJCore move/restore commits, which relocated the file rather than authoring the logic.

The second defect — the finding, the diagnosis and the first draft of its test — came from review on this PR; the test here is trimmed to the cases that can actually fail and moved onto the `@memberjunction/core` package import the other MJServer tests use.

Changeset: **patch** on `@memberjunction/core` and `@memberjunction/server`.


MarceloT-BC:
This fix is correct and I can confirm it works end-to-end — but **it only reaches the user on the CREATE path**, and the PR body states otherwise. Worth catching before merge, because the remaining half is two words.

### The body's claim is one resolver too generous

> `ResolverBase.CreateRecord`/`UpdateRecord` put it in the `GraphQLError`

`CreateRecord` does. `UpdateRecord` and `DeleteRecord` never call `CompleteMessage` at all — they read the bare `Message`:

```ts
// ResolverBase.ts:1248 — CreateRecord   ✅ this is why the PR's repro works
throw new GraphQLError(entityObject.LatestResult?.CompleteMessage ?? 'Unknown error creating record', {
  extensions: { code: 'CREATE_ENTITY_ERROR', entityName },
});

// ResolverBase.ts:1335 — UpdateRecord   ❌
// ResolverBase.ts:1561 — DeleteRecord   ❌
throw new GraphQLError(entityObject.LatestResult?.Message ?? 'Unknown error', {
  extensions: { code: 'SAVE_ENTITY_ERROR', entityName },
});
```

A validation refusal puts its reasons in `Errors` and leaves `Message` unset, so `?? 'Unknown error'` wins every time on an update.

### Measured, same entity, same rule, only the verb changes

Against a live MJAPI (`bizapps-contracts`, a `ModificationText`-must-not-be-blank rule), **with this PR's fix already applied locally**:

| | before | after adding the two words |
|---|---|---|
| CREATE violating the rule | the rule's prose ✅ | unchanged ✅ |
| UPDATE violating the same rule | `{"message":"Unknown error","extensions":{"code":"SAVE_ENTITY_ERROR"}}` ❌ | the rule's prose ✅ |

That asymmetry is also why this survived review: the create path is the one anybody tests first, and it was already right.

### Suggested fix

```ts
// ResolverBase.ts:1335 — UpdateRecord
} else {
  // CompleteMessage, not Message. A validation refusal puts its reasons in `Errors` and leaves
  // `Message` unset, so reading `Message` here fell through to 'Unknown error' and discarded every
  // field-named explanation the entity had just produced. CreateRecord already reads CompleteMessage;
  // this path did not, which is why the SAME rule on the SAME entity explained itself on a create and
  // said nothing on an update. CompleteMessage is a strict superset — it starts from `Message` and
  // appends `Error` and `Errors` — so nothing that used to be reported is lost, and it still returns
  // undefined when there is genuinely nothing to say.
  throw new GraphQLError(entityObject.LatestResult?.CompleteMessage ?? 'Unknown error', {
    extensions: { code: 'SAVE_ENTITY_ERROR', entityName },
  });
}
```

```ts
// ResolverBase.ts:1561 — DeleteRecord
} else {
  // CompleteMessage, for the same reason as the update path. This one matters more going forward:
  // #3971 adds a delete-validation seam, and apps already refuse deletes by overriding Delete() and
  // returning false with a reason on LatestResult (bizapps-accounting and bizapps-contracts both do).
  // Every one of those reasons was being replaced by 'Unknown error' at the API boundary.
  throw new GraphQLError(entityObject.LatestResult?.CompleteMessage ?? 'Unknown error', {
    extensions: { code: 'DELETE_ENTITY_ERROR', entityName },
  });
}
```

**Why it cannot lose information:** `CompleteMessage` seeds `msg` from `this.Message` and only then appends `Error` and `Errors`, so it is a strict superset of what these lines report today. And it leaves `msg` `undefined` when all three are empty, so the `?? 'Unknown error'` fallback still fires rather than showing the user a blank error.

### On testing it

The natural test — assert `CompleteMessage` carries the prose — passes whether or not the resolvers actually *read* it, which is exactly how the gap survived. I ended up needing a source-level guard to pin the selection:

```ts
it('no refusal throw reads LatestResult.Message', () => {
    const offenders = source
        .split('\n')
        .map((line, i) => ({ line: line.trim(), n: i + 1 }))
        .filter(({ line }) => /LatestResult\??\.Message\b/.test(line));   // \b so CompleteMessage does not match
    expect(offenders).toEqual([]);
});
```

Verified locally in `instances/contracts-mj6/mj`: 10/10 in `packages/MJServer/src/__tests__/resolverBase.refusalMessage.test.ts` (red-proven — reverting one site fails 2), `packages/MJServer` build `exit 0`, and the live before/after above. Happy to hand over the test file if it's useful, or leave it to you.

*Found while implementing the first server-side validation rule in `bizapps-contracts` — the rule refused correctly and the user was told "Unknown error".*

MarceloT-BC:
Cross-check from #3974 (the delete-validation seam, which lands on the path this PR fixes): switching `UpdateRecord`/`DeleteRecord` to `CompleteMessage` is right, and it surfaces a **second-order problem that already exists on `next`** — `CompleteMessage` seeds from `Message` and *then* appends every `Errors` entry, with no de-duplication. Any producer that sets both will now render its reason twice.

That is not hypothetical, and it is not only my branch. Two sites in `baseEntity.ts` on `next` today build `Message` **out of** the very array they also assign to `Errors`:

```ts
// baseEntity.ts:3501 (_InnerSave, parent-save failure) and :4704 (_InnerDelete, parent-delete failure)
const detail = parentLatest?.Message ||
    parentErrors.map(e => e?.Message ?? String(e)).filter(Boolean).join('; ') || '…';
newResult.Message = `Failed to save parent entity '…': ${detail}`;
newResult.Errors  = parentErrors;      // ← same text, second time
```

Before this PR that was invisible on the update/delete paths (they read the bare `Message`) and harmless on create (the `Errors` entries JSON-blobbed, so the repetition didn't read as repetition). After it, an IS-A parent failure with two errors renders:

```
Failed to delete parent entity 'Products': A; B
A
B
```

#3974 hits the same shape deliberately: a refused delete sets `Message` to the joined prose (for the many callers that read `LatestResult?.Message` directly — `role-management.component.ts:488` and friends) **and** `Errors` for the field-named `Source` a form needs.

I think the fix belongs **here, in `CompleteMessage`**, not in each producer: skip an `Errors` entry whose `ErrorText()` is already contained in the accumulated message. That protects every producer at once, including the two above, rather than asking each one to choose between a directly-readable `Message` and a clean `CompleteMessage` — and choosing "only populate `Errors`" is a real cost until #1431's migration is finished and everything reads `CompleteMessage`.

Roughly, inside the existing loop:

```ts
const text = BaseEntityResult.ErrorText(err);
if (text && !(msg && msg.includes(text))) { /* append */ }
```

Merge order between the two PRs doesn't matter — neither breaks the other, and both are green as they stand. Happy to take it in #3974 instead if you'd rather keep this PR to the casing fix; it just fixes fewer callers there.

MarceloT-BC:
Both rounds are folded in. I verified each claim from source rather than taking it, and one of my own checks nearly went the wrong way — worth recording.

### Round 1 — the update/delete paths (`723758b1`)

Correct, and my PR body was wrong: it claimed `CreateRecord`/`UpdateRecord` both read `CompleteMessage`. Only `CreateRecord` does. Fixed in the body, the changeset, **and** the `ErrorText` docstring, which repeated the same claim in shipped code.

I first tried to measure it against the live MJAPI on `contracts-mj6` and got the prose back — which looked like a refutation. It wasn't: that worktree already had your patched `ResolverBase`, so I was measuring the *after* state. Re-verified the other way instead, through the real code path after a refused `Save()`:

```
LatestResult.Message         = null      →  ?? 'Unknown error'  =  "Unknown error"
LatestResult.CompleteMessage = "the rule refused, in prose"
```

Both throws now read `CompleteMessage`. Your test is in, trimmed to the cases that can actually fail — the four that restate what a resolver reads stay green through the regression, so they document rather than guard — and moved onto the `@memberjunction/core` package import the other MJServer tests use, instead of a deep relative path into `MJCore/src`. 5/5, red-proven: the source-scan case fails on `origin/next` naming lines 1335 and 1561.

### Round 2 — the duplication (`76669b0a`)

Reproduced exactly, against the built dist:

```
Failed to save parent entity 'Products': A; B
A
B
```

I went in expecting to argue against the substring heuristic — silently dropping an error is precisely the failure class this PR exists to fix. That objection doesn't survive the condition: the skip fires only when `msg.includes(text)`, i.e. only when that text is **already in the output**, so nothing a reader could not already see can be lost. Implemented as you sketched, with a test pinning the guard: an error the `Message` did *not* contain is still reported. Also worth noting the `parentLatest?.Message ||` fallback at `:3475`/`:4453` selects the joined form exactly when the parent's own `Message` is empty — i.e. on a validation refusal, the case this PR is about.

A third test now covers the `typeof err === 'string'` branch of `ErrorText`, which had none. Without it a bare string finds neither `Message` nor `message` and falls to `JSON.stringify`, coming back quoted with its newlines escaped — `"line one\nline two"` on one line.

**Take it here, not in #3974** — the resolver change and the de-duplication both live in this branch now, so please drop them from yours to avoid a conflicting duplicate.

### Issue chain

No `Fixes #N` — nothing was ever filed for any of the three. But the PR body now links what this actually completes, and the headline is **#243**: *"If a save attempt fails, display an error to the user … Also need to make sure validation messages reach the client so that we can display those errors."* Opened 2024-05-30, **closed as completed 2025-04-21** — and it never worked: JSON blob on create, nothing at all on update or delete. Then **#246** (the result-object return type, opened explicitly to serve #243) and **#1431** (the `CompleteMessage` migration, closed with the getter broken and two call sites unmigrated). I have not reopened anything.

### CI, and something worth knowing about it

Both red checks are pre-existing on `next` and unrelated:

- **Unit tests** — `metadata.live-db.integration.test.ts` imports `mssql`, which MJCore does not declare. `next`'s own head `e45fde5b2c` fails identically (141/1); this branch is 142/1 — one more passing file, same inherited failure.
- **Integration** — `codegen-determinism` CD3/CD5/CD6: `entity_subclasses.ts` registers 379 entities against 378 in the DB, `MJFormChromeRule` deleted without regenerating artifacts. No generated code, metadata or migrations are touched here.

The first one is worse than it looks: turbo halts on it. `Tasks: 143 successful, 159 total` — **15 tasks never ran**, and `@memberjunction/server:test` is one of them. So CI has not executed the new resolver test on any commit of this PR, and will not until that `mssql` import is fixed or excluded. It is green locally (5/5). Flagging rather than filing.

MarceloT-BC:
**Retracting the de-duplication — I was wrong to take it, and it is reverted in `819ae2d2`.**

I claimed the guard "cannot hide anything by construction, because the skip condition is that the text is already in the output." That argument does not hold. The *characters* survive; the count, the attribution and the distinctness do not. Measured against the built getter:

| case | without the guard | with it |
| --- | --- | --- |
| three fields failing with the same sentence | 3 lines | **1 line** |
| `["Quantity must be positive", "Quantity must be positive for shipped items"]` | both | both |
| **the same two, order reversed** | both | **second dropped** |
| `Message: "Invalid input for Status"` + entries `["Invalid input", "Region is unknown"]` | both | **the Status error vanishes, Region survives** |

Rows 2 and 3 are the same failure set rendering differently depending on **array order**. That is not de-duplication.

And none of it is detectable by the person reading the message — which is the exact property that let the original casing bug live for ten months. A renderer cannot tell a redundant repeat from a real second failure; only the producer knows.

So the duplication stays, with a comment at the site recording that it is deliberate and why. It is real and it is ugly, but it is the safe direction of the trade: saying something twice is a cosmetic problem, and reporting one problem when there were three is the bug class this PR exists to fix.

**If it is worth removing, it belongs at the two producers** — `_InnerSave`:3475 and `_InnerDelete`:4453, which compose their `Message` out of the same `parentErrors` they then attach. There the redundancy is *known* rather than guessed, so nothing has to be inferred from string containment. That is a separate change from this PR, and it needs a call on the trade it forces: those sites inline the detail specifically so callers reading `Message` directly are not handed a message that says nothing, and that is only unnecessary once #1431's migration is actually finished.

@AN-BC — flagging for a ruling, since this is a judgement about MJ-wide error semantics rather than a local fix, and nothing here has had a maintainer's eye on it yet.

Unchanged from that commit: the test covering `ErrorText`'s string branch, which is unrelated and fills a real gap. Suite is 8 cases now, 6 red against the pristine getter, MJCore 2076/2076, `tsc` exit 0.

packages/MJCore/src/generic/baseEntity.ts
```diff
@@ -691,6 +691,46 @@ export class BaseEntityResult {
...
+     * because `JSON.stringify` always succeeds. It is only visible by reading the message a user got.
+     *
+     * `Message` is preferred over `message` because a `ValidationErrorInfo` has only the former,
+     * while an `Error` has only the latter — so the order matters solely for an object carrying both,
+     * where the MJ-native field is the better answer.
+     *
+     * @param err - One entry from the `Errors` array.
+     * @returns The entry's human-readable text, falling back to JSON for a shape with neither field.
+     */
+    public static ErrorText(err: any): string {
```
rkihm-BC: Required: `unknown` rather than `any`. The repo rule (`.claude/rules/typescript-style.md`, "No `: any` type annotations") applies to new code even where the surrounding properties are legacy `any`, and newer methods in this file already follow it (`RestoreOldValue(value: unknown)`, `mirrorSharedKey(name, value: unknown)`). `unknown` is the right type here rather than a lazy substitute: the helper exists precisely because the shape is not known. Callers pass `any` today, so nothing else changes.

```ts
public static ErrorText(err: unknown): string {
    if (err === null || err === undefined) {
        return '';
    }
    if (typeof err === 'string') {
        return err;
    }
    if (typeof err === 'object') {
        const shaped = err as { Message?: unknown; message?: unknown };
        const text = shaped.Message ?? shaped.message;
        if (typeof text === 'string' && text.trim().length > 0) {
            return text;
        }
    }
    return JSON.stringify(err);
}
```

Behaviour is identical for every shape the tests cover, including the JSON fallback for an object with neither field and for non-object primitives.

rkihm-BC (changes requested):
Both defects are real and both fixes are correct and minimal. I verified rather than trusted:

- **On `next`:** the getter maps `Errors` with lowercase `err.message` while `ValidationErrorInfo` only has `Message`; `UpdateRecord` and `DeleteRecord` throw `LatestResult?.Message`, which `_InnerSave`'s catch sets to `e.message || null`, so a thrown `ValidationResult` always lands on the `'Unknown error'` fallback.
- **Red/green reproduced locally.** MJCore's new test against the base getter: 6 failed / 2 passed, exactly the two the description names; against the fix 8/8. MJServer's new test against the base resolver: the source-scan case fails, 4 pass; against the fix 5/5 (with MJCore's dist rebuilt so the server saw the fixed getter).
- **CI now runs the resolver test.** The August comment about turbo halting on a stray `mssql` import no longer applies: that test file is gone from `next`, and this PR's latest shard-6 log shows `resolverBase.refusalMessage.test.ts` executing and passing.
- **Consumers already on `CompleteMessage` benefit immediately:** `BaseEntitySaveQueue`, `saveEntityGraphOperation`, `EntityFieldRules`, and `entitySavePlan` all read it.

**On the de-duplication question flagged for a ruling:** I agree with the final position. The revert in `819ae2d2` was right. A containment test drops distinct errors depending on array order, and a renderer cannot tell a redundant repeat from a second failure; only the producer knows. Keeping the repetition with the comment at the site is the safe direction. If it is worth removing, it belongs at the two IS-A producer sites in `_InnerSave` / `_InnerDelete`, as its own change.

**One required change, inline:** the new helper introduces an explicit `any` parameter. The rule in `.claude/rules/typescript-style.md` is explicit about `: any` annotations, and it is not something the standards gate checks, so green CI is not evidence here. `unknown` with narrowing is the correct type for a helper whose whole job is handling a shape it cannot know. Four lines; details on the line.

**Follow-up worth an issue, not this PR:** twenty-plus other `LatestResult?.Message ?? 'Unknown error'` readers across the AI packages (tag engines, agent-manager actions, MCP client, and others) have the same behaviour on a validation refusal. That is the unfinished #1431 migration, which this PR's history shows was closed with the getter itself broken.

Changeset is present and `patch` on `core` and `server` is the right level.

packages/MJCore/src/generic/baseEntity.ts
MarceloT-BC: Done in `9f2a8fff` — taken as written.

You are right on all three counts, and I checked each rather than taking them: the rule is explicit (`🚨 NO any TYPES - EVER` / "No `: any` type annotations"), both precedents are real (`RestoreOldValue(value: unknown)` at :482, `mirrorSharedKey(name, value: unknown)` at :1285), and there are exactly **23** other `LatestResult?.Message ?? 'Unknown error'` readers on today's `next`.

Your point that green CI is not evidence here is the part worth recording — no job checks this rule, so the only thing that would have caught it is a person reading the diff. That is the same shape as the defect the PR fixes.

On equivalence, I did not want to assert it: I ran both implementations over 16 shapes — `null`, `undefined`, string, empty string, multi-line string, `ValidationErrorInfo`, `Error`, `{message:''}`, `{Message:'   '}`, an object with neither field, number, boolean, `0`, a boxed `String`, an array, and an object carrying both fields. Identical output on every one. The added `typeof err === 'object'` guard only changes the path taken for non-object primitives, which already reached `JSON.stringify` by way of two `undefined` property reads.

I also swept the rest of the diff for the same violation — the only other `any` in it is the English word in a comment.

The docstring now records *why* `unknown` rather than treating it as a lint concession: not knowing the shape is the reason the helper exists, so the narrowing belongs in the signature. I left the one `any` mention in the prose above it, since that sentence is about `Errors` being `any[]` — still true, and the reason the compiler could not catch the original bug.

**On the de-duplication:** agreed, and thanks for weighing in — that had been sitting open for a maintainer's eye since 20 Aug with only agents on it.

**On the follow-up:** I have not filed the issue for the other 23 readers. Filing is Marcelo's call in this workspace; I have flagged it to him with your framing (it is the tail of #1431, which was closed while the getter it migrated everyone onto was itself broken).

Branch is also caught up with `next` as of this push.

rkihm-BC (approved):
Approving. The one required change is done in `9f2a8fff` exactly as requested: `ErrorText` takes `unknown`, narrows with a `typeof err === 'object'` guard, and the JSDoc records why that is the right type rather than a lint concession.

Verified on this head rather than taking the reply on trust:

- `baseEntityResult.completeMessage.test.ts` 8/8; full MJCore suite 2,170 passed across 147 files; `tsc --noEmit` exit 0.
- No `any` remains in the added code — the only remaining match in the diff is the English word inside the de-duplication comment.
- Behaviour is unchanged: the added object guard only changes the path for non-object primitives, which already reached `JSON.stringify` via two `undefined` property reads. That matches the 16-shape check in the reply.
- CI is fully green on this head; the three shards that had not reported when I first looked have since passed.
- Branch is caught up with `next`; changeset unchanged and still correct at `patch` for `core` and `server`.

The de-duplication question now has a ruling in the thread (keep the repetition; a producer-side fix at the two IS-A sites is its own change). The follow-up for the 23 other `LatestResult?.Message ?? 'Unknown error'` readers is acknowledged and routed to Marcelo, and is not a condition of this PR.

Thanks for the careful turnaround.

PR #3974, opened 2026-08-20
feat(core): add a delete-validation seam to BaseEntity (ValidateDelete/ValidateDeleteAsync)

MarceloT-BC:
Addresses #3971 — implements the issue's **primary** proposal (the seam). Deliberately *not* closing it: the issue's third proposal (metadata-driven referential messages) is a separate PR, and auto-closing here would lose it. See "Not included, on purpose".

## The gap

`Save()` has a validation seam with a defined contract: return a `ValidationResult`, let the framework decide, get field-named errors in front of the user. `Delete()` had **none** — I checked for all of them: no `ValidateDelete`, no `CanDelete`, no vetoable `before_delete`, and no `PreDeleteHook` counterpart to `PreSaveHook`.

So the only way to refuse a delete with an explanation was to override `Delete()` itself and return `false`. That:

1. **Discards the reason.** A `boolean` has nowhere to put "this template is referenced by 5 signed contracts".
2. **Gets reimplemented per entity.** `bizapps-accounting` has two (`JournalEntryTypeEntityServer.Delete()`, `JournalEntryLineEntityServer.Delete()`, both of which `throw`); `bizapps-contracts` was about to add four more.
3. **Guarantees nothing** about ordering relative to permissions, events, or the companion delete graph, and covers only callers that reach that subclass's method.

## The fix

A delete-side counterpart to the save seam, in `packages/MJCore/src/generic/`:

| | |
|---|---|
| `ValidateDelete()` / `ValidateDeleteAsync()` | The seam. Base implementations return success, so this is **inert for every existing entity**. |
| The gate in `_InnerDelete` | In the slot `Validate()` occupies in `_InnerSave`: **after** `CheckPermissions` (a user without delete rights still gets a permission error, not a validation message) and **before** any provider work or the `delete_started` event — nothing started, so nothing is announced. |
| `deleteGraph()` (existing, extended) | A companion delete plan is validated **whole, up front**. No new entry point — `Delete()`'s routing is byte-identical to the base commit. |
| `EntityDeleteOptions.SkipAsyncValidation` | Governs the async half only. Also corrects the `ReplayOnly` doc, which already claimed to "BYPASS Validate()" for a gate that did not exist. |

**The reason now survives to the user with no server or UI change.** A refusal is recorded with both `Errors` (field-named, `Source` intact, which is how forms mark the offending field) *and* `Message`. Since #3973 merged, `ResolverBase.DeleteRecord` reads `CompleteMessage` (which renders both); MJ's own Explorer delete callers — 74 sites in `packages/Angular` on `next` today, 136 repo-wide — still read the bare `LatestResult?.Message`. Before this PR, both reads hit `null` and printed **"Unknown error"**. (Populating both means the text appears twice on the GraphQL path; see "Interaction with #3973" for why that is deliberate.)

**Why the plan is validated up front, and why `saveGraph` needs no equivalent.** Companion delete plans remove *children first* — they hold the foreign keys pointing at the row about to disappear. A refusal discovered when the root's own turn came would already have deleted them, and on a client provider `executeGraphLocal` holds no transaction to undo it. The refusal would cost exactly the data it existed to protect. Nodes are flagged so no rule, or its query, runs twice.

`saveGraph` needs no such pass precisely because it is the mirror image: its root node runs **first** and validates the graph on the way through (its doc comment says so). Delete plans are the other way round, which is the whole reason this pass exists.

Note what this does **not** change: graph deletes, `BuildDeletePlan` and `deleteGraph` are all pre-existing on `next`, and they still fire only when `HasCompanions && NodeCount > 1`. An entity without companions never reaches that path. `Delete()`'s routing line is byte-identical to the base commit.

### Example

```typescript
public override ValidateDelete(): ValidationResult {
    const result = super.ValidateDelete();
    if (this.IsSystem) {
        result.Success = false;
        result.Errors.push(new ValidationErrorInfo('IsSystem',
            `JournalEntryType '${this.Code}' is a system row and cannot be deleted — ` +
            `accounting's triggers and batch pipeline key on it.`, this.IsSystem));
    }
    return result;
}
```

That replaces the current `Delete()` override that `throw`s, and the sentence reaches the user instead of an exception.

## `DefaultSkipAsyncValidation`: one policy for both seams — and two explicit asks for you

The first pass of this PR had the delete seam decide its async half on its own and **deliberately ignore** `DefaultSkipAsyncValidation`, on the grounds that the getter's `true` default is precisely what made hand-written `ValidateAsync` overrides silent no-ops (`OrderEntityServer.ValidateAsync` — holding both the "cannot confirm an order with no lines" guard and an entire per-line validation loop — was dead on every save in production for exactly that reason).

On review that was judged the wrong trade: it protected the delete path by **splitting the API's story in two.** The same entity would have had one async-validation policy for saves and a different, implicit one for deletes, and an author reading either seam would have had to learn that the other worked differently. The semantics don't want two policies — the getter answers *"does this entity run expensive async validation"*, which is a property of the **entity**, not of the verb.

So `22e83128cf` unifies them. Both seams now resolve through one helper, `shouldSkipAsyncValidation(explicitOption, asyncMemberName)`, with the precedence the save path already established:

1. an explicit `SkipAsyncValidation` in the call's options wins outright;
2. an explicit `DefaultSkipAsyncValidation` override wins next — **either** value, because a `true` returned by an override is a *choice* and a `true` inherited from the base is the *absence* of one;
3. otherwise it is inferred **per seam** from whether that seam's async validator was overridden.

Step 3 is what keeps the original fix intact: an entity that states no policy still gets the async rules it actually wrote, on both paths. All 15 classes overriding the getter in this repo return `false`, so unifying is inert for every one of them.

### Ask 1 — is the shared policy what you want, or should delete have its own?

The consequence, stated plainly because it is the thing to disagree with: **an entity that overrides the getter to `true` now opts out of async validation on BOTH paths** — including a `ValidateDeleteAsync` written later by someone who never saw the getter. That is the original trap reintroduced at smaller scale. I think one policy is right (it is a property of the entity, and the delete side has no legacy overrides to preserve), but the counter-argument is real, and the alternative is a second getter that an author can equally fail to find. Your call — I'll split it back out if you prefer per-verb policies.

### Ask 2 — rename it now, or later?

The name is the actual root cause of the original silent no-op: `DefaultSkipAsyncValidation` reads like *the operative switch*, when since the inference landed its only job is *stating an explicit policy*. Now that it governs two seams, the misreading gets more expensive.

Something like `AsyncValidationPolicy` (`'auto' | 'always' | 'never'`, default `'auto'`) would say what it does and make the three-way precedence legible instead of encoded in a boolean. **Cheap to do now** — 15 in-repo call sites, all returning `false`, plus a deprecated alias for anything downstream. Doing it later means a second breaking-ish change over the same surface.

Not doing it in this PR unless you say so: it is a rename across a public getter, which deserves its own review rather than riding along with a new feature. I did **not** file an issue for the original silent-no-op — that fix has already landed on `next`.

## Not included, on purpose

The issue's third option — deriving *"this row is referenced by N `<Entity>` records"* from `EntityRelationship` metadata, with no per-app code at all — is genuinely the higher-value change for the common case, and it is **cheaper than it looks**: `DatabaseProviderBase.Delete()` already catches the SQL error and puts it on `entityResult.Message`, so the raw `The DELETE statement conflicted with the REFERENCE constraint "FK_…"` text (which names the offending **table and column**) is already reachable — no provider change needed to get at it.

It is a separate PR because it *changes* a message users already see on every FK-blocked delete, it needs dialect awareness (the SQL Server phrasing is not Postgres's), and it deserves live-DB proof rather than a canned-message unit test. It also needs no seam, and the seam needs no it. Say the word and I'll author it against this branch.

Also out of scope: a global `PreDeleteHook` counterpart to `PreSaveHook`, and migrating the two `bizapps-accounting` overrides onto the seam.

### Interaction with #3973 — settled

#3973 is **merged** (`67e4c9e828`): all three `ResolverBase` write-refusal throws now read `CompleteMessage`, so a delete refusal set here reaches the GraphQL client as prose. This branch carries no resolver change — nothing to conflict.

The one open detail is closed the other way from how I first raised it. This PR sets **both** `Message` (joined prose) and `Errors` (field-named), and `CompleteMessage` renders `Message` then every `Errors` entry, so on the GraphQL path a refusal's text appears **twice**. De-duplicating in `CompleteMessage` was implemented, measured, and reverted in #3973 (with the maintainer's agreement): any containment test is lossy — three fields failing with the same sentence collapse to one line, and a distinct error vanishes when its text sits inside the summary. The renderer now carries a comment saying the duplication is deliberate. So this PR keeps both fields — `Message` because 136 call sites still read it bare — and documents the double render in `registerDeleteValidationFailure` as the accepted cost. If #1431 is ever finished, dropping `Message` there is the one-line producer-side fix.

## Self-audit before review — two defects found and fixed (`87790c10e3`)

Rather than assume the paths I hadn't tested were fine, I walked every route into `_InnerDelete`. Both defects were mine, introduced with the seam, and neither had a failing test:

1. **`SkipAsyncValidation` was dropped on the IS-A parent-chain delete.** `_InnerSave` copies the flag onto its `parentSaveOptions`; the delete path built `parentDeleteOptions` without it. An explicit opt-out was honoured for the child and silently ignored one link up — the parent paid for a query the caller had opted out of, and its refusal failed the whole delete. Found by diffing the two option-builders against each other. Red-proven: removing the line fails exactly one test.
2. **The plan pre-flight could throw out of `Delete()`.** An application rule is arbitrary code. On the single-record path the gate sits inside `_InnerDelete`'s `try`; the companion pre-flight had none, so a rejection escaped a method whose whole contract is to report failure by *returning* `false` — the same hazard `executeGraphLocal` documents for its transaction begin.

**One deliberate extension, easy to reject:** the pre-flight now checks each node's permissions *before* running its rules. The single-record path always has, and these docs promise it; on the graph path children were validated before their own permission check, which contradicts that *and* lets a permission failure on the last child leave earlier siblings deleted on a provider with no transaction. Same method, same arguments — it can never allow anything new, only fail earlier and in the documented order. Say the word and I'll drop it.

## Review round — 2026-09-03 (`c8bdb22d73`), after merging `next`

Three findings from a second reader of the `node.CheckPermissions(EntityPermissionType.Delete, true)` line in `deleteGraph`, all confirmed and fixed:

1. **Mis-attribution.** The pre-flight's `catch` stamped `this.EntityInfo?.Name` — the **root** — on every failure, so a permission or rule failure on a *child* node was reported as the parent's. It now tracks the node being checked and stamps that node's entity name; the log line names both.
2. **The comment over-promised.** It said a user without delete rights gets "a permission error, not a validation message" — but the throw lands in the `catch` two lines down and becomes a `ValidationErrorInfo` on `Errors`. The *ordering* claim held (permissions before any rule); the *shape* claim did not. The comment now says what actually happens, and why discarding the return is correct only while the second argument is `true`.
3. **No test covered a no-permission node in a plan** — the one call in the pre-flight whose only observable effect is a throw. Flipping that `true` to `false` compiled and passed every test: it becomes a discarded boolean that permits everything. New case: a child the user may not delete refuses the plan before any rule runs and before any row goes, attributed to `Products` (the child) rather than `Standalone Items` (the root). The permission fixtures also threw *unconditionally*, so they could not see that flip either; they now honour `throwError` like the real method. **Red-proven three ways:** flipping the graph call fails the new case (the child's rule ran twice), flipping the single-record call fails the existing ordering case, and reverting the attribution fails the new case on `Source`.

### Known limitation, pinned by a test rather than hidden

**The IS-A parent chain is not validated up front, unlike a companion plan.** A child's own row is deleted first (the FK requires it), so when a *parent's* rule refuses, the child row is already gone — rolled back inside a provider transaction, permanent on a client provider. `"surfaces the parent's reason on the CHILD"` asserts exactly that, so the gap lives in code, not in prose.

I did not extend the pre-flight to walk the chain here: it interacts with leaf delegation (`this.LeafEntity.Delete()`) and the `IsParentType` child-record protection, both of which run before the gate, and that deserves its own change. Follow-up if you want it.

### Also worth knowing

**On a client provider the rules run twice — once per tier, deliberately.** `GraphQLDataProvider.Delete()` sends a mutation, so the browser's entity refuses locally (fast, no round trip) and the server re-runs the rules on the server-registered subclass, which is the authoritative pass. That is the behaviour you want; it is worth knowing before someone puts an expensive query in `ValidateDeleteAsync`.

## Testing

The spec was written from the issue's expected behaviour *before* touching the implementation, so it is an independent oracle. Against the base of this branch it **fails 11 of 16** — an override of `ValidateDelete()` is never called, the row is deleted, and `Delete()` returns `true`. That is the repro.

| | |
|---|---|
| `baseEntity.validateDelete.test.ts` (new, 30 cases) | **30/30 pass** — refusal blocks the provider call, `Source` and `Message` survive, warnings don't block, sync+async errors combine, override inference across a multi-level chain, options precedence, `ReplayOnly` bypass, the shared policy (deliberate opt-out/opt-in, options outranking both, and a sync refusal surviving an async opt-out), the ordering guarantees (permissions before rules, no `delete_started` on a refusal), a throwing rule on both paths, the graph pre-flight (a root *or a child* refusal leaves every row intact; a child the user may not delete refuses the plan before any rule or row, attributed to that child; root validated exactly once), and the IS-A chain |
| `baseEntity.validateAsync.test.ts` (existing, untouched) | **15/15 pass** — the regression guard for refactoring the save path's decision out from under it |
| All 23 `baseEntity.*` files | **349/349 pass** |
| Full `@memberjunction/core` suite | **2213/2213 pass** (after merging `next`, 2026-09-03) |
| `tsc` build of `@memberjunction/core` | clean |

Two environment notes, neither a regression: the earlier run could not *load* `metadata.live-db.integration.test.ts` (`mssql` missing from a filtered install); and after the `next` merge, ten `relatedRecordBatchLoader` cases failed with `EscapeSQLString is not a function` until the local `@memberjunction/global` dist was rebuilt — a stale sibling build, green once rebuilt. The red **"Run unit tests"** check on the pre-merge head was an unhandled rejection in `ng-file-storage` (`UserInfoEngine` debounced flush), fixed on `next` in `d078c545c8`; this PR's spec ran 29/29 green in that same job.

**Not run: the live-instance / GUI tier.** This is an in-process seam with unit coverage of every branch, but I have not deleted a real row through Explorer and watched the message render. Happy to if you'd like that before merge.

## Related issues and PRs

- **#3971** — the issue. This PR implements its **first** proposal (the seam). "Addresses", not "closes": its **third** proposal (metadata-driven *"referenced by N `<Entity>` records"*) is deliberately a separate PR, and auto-closing would lose it. Its second (a vetoable `before_delete` event) is superseded by the seam.
- **#3676** — *"run an overridden `ValidateAsync` instead of silently skipping it"* (merged). The precedent this PR extends: it established the option → explicit-policy → inference precedence, and it is the concrete history behind Ask 2. No issue was ever filed for that silent no-op, which is why there is none to close here.
- **#3973** — **merged.** The renderer half of the same user-visible story (`CompleteMessage` rendering `ValidationErrorInfo` as prose, and `CreateRecord`/`UpdateRecord`/`DeleteRecord` reading it). The `Message`/`Errors` double render was ruled deliberate there — see "Interaction with #3973 — settled" above.
- **#1431** — *"Refactor codebase to use `BaseEntityResult.CompleteMessage` instead of `Error`/`Errors`/`Message`"* (closed, unfinished — see #3973). This PR populates **both** `Message` and `Errors` precisely because that migration is incomplete: many callers, including MJ's own Explorer delete handlers, still read the bare `Message`. If #1431 is ever finished, dropping `Message` here becomes the tidier choice.
- **#3962** — *"validation errors are never aggregated to the section/panel that owns the field"* (open). The consumer-side gap this seam feeds into. Worth flagging for whoever picks that up: a **delete** refusal's `Source` may legitimately name a *relationship* (`'Lines'`) rather than a form field, so aggregation needs a fallback for a `Source` that matches no field rather than dropping the message.
- **#3969** — adjacent: more `ValidationErrorInfo`s reaching the same renderer, from field-level value-list validation.


MarceloT-BC:
Reviewed this against a live app that has four of exactly these guards (`bizapps-contracts`, PR #11 — it currently ships them as `Delete()` overrides, and they migrate onto this seam cleanly). The seam is right and the gate placement is right. One substantive concern, and an answer to Ask 1.

### The concern: this and #3973 are converging on two different conventions for one fact

`registerDeleteValidationFailure` populates **both** `Errors` and `Message`, and the doc comment is explicit about why:

> Note the save path leaves `Message` null on a validation failure because it throws the `ValidationResult` into a catch block that reads `e.message`; nothing here needs to repeat that.

That works, and it is the reason this PR can claim "no server or UI change". But look at where the two PRs leave the codebase together:

| path | what the result carries | what `ResolverBase` reads | what the user sees |
|---|---|---|---|
| create | `Errors` only | `CompleteMessage` (:1248) | ✅ the prose (after #3973) |
| **update** | `Errors` only | **`Message`** (:1335) | ❌ `"Unknown error"` |
| delete | `Errors` **+ `Message`** (this PR) | `Message` (:1561) | ✅ the prose |

Delete gets fixed by changing the **producer**; create was fixed by the **consumer** already being right; update is fixed by neither. So after both PRs merge, a validation refusal explains itself on a create and a delete and stays `"Unknown error"` on an **update** — the most common write there is. I measured that on a running MJAPI and left the two-word resolver fix on #3973.

The asymmetry is the part worth deciding deliberately, because whichever way it goes, it should not be three different arrangements:

- **Option A — consumers read `CompleteMessage` everywhere.** `Message` → `CompleteMessage` at `ResolverBase.ts:1335` and `:1561`. Then this PR's `Message` population becomes harmless belt-and-braces (and still earns its keep for the "caller that only logs" case the comment names), and all three verbs behave the same.
- **Option B — producers always populate `Message`.** Then the save path's catch block needs the same treatment this PR gives the delete path, and `CompleteMessage` stops mattering to resolvers.

A stops being three code paths; B is more invasive but makes `Message` genuinely authoritative. I'd take A, and I'd take it in #3973 rather than here — but this PR is where the split becomes visible, so it seemed worth raising here.

Small related note: `registerDeleteValidationFailure`'s `Errors.map(e => e?.Message).join('; ')` is a second renderer of the same thing `CompleteMessage` renders, with a different separator (`'; '` vs `'\n'`). If Option A lands, this could delegate to `CompleteMessage` instead of re-deriving it, so a future change to how errors read happens in one place. It also correctly uses capital `Message` — which is precisely what #3973 is fixing `CompleteMessage` for, so the two are already in agreement about the field; only the rendering is duplicated.

### Ask 1 — one shared policy is right, and Ask 2 is the real fix for the trap you name

Shared, for the reason you give: it answers *"does this entity run expensive async validation"*, which is a property of the entity. A second getter doubles the surface an author has to know about to reason about one behaviour, and the failure mode of the split version is worse — an entity whose save and delete disagree about async validation, with nothing in either seam saying so.

The trap you flag is real (`DefaultSkipAsyncValidation → true` silently opting out a `ValidateDeleteAsync` written later), but note it is not new: it is the same trap that made `OrderEntityServer.ValidateAsync` dead, and a second getter does not remove it — it just gives the next author two places to not look. **Ask 2 removes it.** `AsyncValidationPolicy: 'auto' | 'always' | 'never'` makes "I am stating a policy" unmistakable, and `'never'` reads like an opt-out in a way `true` does not — someone setting `'never'` has said something about all async validation on that entity, which is what they mean.

So: unified policy, and I'd do the rename now while it is 15 call sites all returning `false`, rather than after this PR widens what the getter governs.

### Not included, on purpose — the third proposal

For what it is worth from the app side: the metadata-driven variant would delete most of `bizapps-contracts`' R-8 code, and I would be glad to lose it. But the four guards there each name a *specific* consequence — *"8 contracts incorporate this version as their standard terms, so deleting it would erase what those customers agreed to"* — and a generic *"referenced by 8 Contract records"* is not the same sentence. My read is that the generic message is the right default for the long tail and the seam stays for the handful of rows where the reason is worth writing by hand, which is what you have built. Both, not either.

*Reviewed while migrating four `Delete()` overrides in `bizapps-contracts` onto this shape; the seam fits them without changes to the dependency logic.*

Issue #3970, opened 2026-08-20
Unseeded EntityField.Configuration: after CodeGen, 43 core entities describe base-view columns that do not exist, so every save fails with Msg 213 (fresh-install blocker on next)


MarceloT-BC:
## Summary

On `next` (`e45fde5b2c`), a fresh install that runs CodeGen leaves **43 `EntityField` rows describing base-view columns that do not exist**. Every subsequent create/update of those entities fails with

```
Msg 213: Column name or number of supplied values does not match table definition.
```

`MJ: Record Changes` is one of the 43, and every audited save writes a `RecordChange` through `spCreateRecordChange_Internal`, so **in practice every entity save fails**. `mj sync push --dir=metadata` therefore cannot complete, and MJ's own workspace build does not compile either. Deterministic, reproduced 4/4 on two commits three weeks apart.

## Reproduce

```bash
# fresh database, MJ next @ e45fde5b2c
mj migrate
mj codegen                     # or any CodeGen run
mj sync push --dir=metadata
```

```
Failed to save MJ: Authentication Providers record at MJ: Authentication Providers[0]:
Error executing SQL
  Error: Column name or number of supplied values does not match table definition.
```

Clear the first entity and the next one fails identically.

**Two SQL probes that show the state directly:**

```sql
-- 0 on a fresh install, so no field is declared a hierarchy
SELECT COUNT(*) FROM __mj.EntityField WHERE Configuration IS NOT NULL;

-- 43: EntityField rows whose base view has no such column
SELECT e.Name, f.Name
FROM __mj.Entity e
JOIN __mj.EntityField f ON f.EntityID = e.ID
LEFT JOIN sys.columns c
  ON c.object_id = OBJECT_ID(e.SchemaName + '.' + e.BaseView) AND c.name = f.Name
WHERE e.SchemaName = '__mj' AND f.IsVirtual = 1 AND c.name IS NULL
  AND OBJECT_ID(e.SchemaName + '.' + e.BaseView) IS NOT NULL;
```

## Root cause

Five links, each verified at source on `e45fde5b2c`:

[...]

Issue #3547, opened 2026-08-06
Open App install cannot bound what it wrote to the shared core schema, so a failed install cannot fully roll back


MarceloT-BC:
## Summary

A failed Open App install cannot say what it wrote to the shared `__mj` core schema, so it cannot fully undo it. Today the gap is covered by asking each app publisher to hand-author inverse `DELETE` scripts (`migrations.teardownDirectory`). That is the wrong layer: it makes the platform's atomicity guarantee depend on every publisher writing and maintaining correct inverse SQL forever, and it is currently adopted by exactly one app.

This issue collects five findings from the #3451 / #3469 investigation that are all the same underlying problem, and proposes a direction that removes the publisher burden entirely.

## Why now

#3469 moved app install to `TransactionMode: 'per-migration'` (matching `mj migrate`, per Amith's ruling) because `'per-run'` cannot host an app that creates a SQL Server table type — see #3451. That was the right call and it unblocked installs that were previously impossible. But it also means a migration set that fails partway leaves earlier migrations committed, so the install's all-or-nothing guarantee now lives in application-level compensation rather than in the database.

#3469 made that compensation real (remove entity metadata → run teardown → drop schema). This issue is about the part it could not close.

## The five faces

**1. `per-migration` does not make table types safe — publisher file layout does.**

`per-migration` wraps *each file* in a transaction. A `CREATE TYPE … AS TABLE` and its first `DECLARE @x t` still self-deadlock (error 1205) when they land in the **same** file. Verified directly during the #3451 investigation.

So the current state is not "solved", it is "works if the publisher happens to split the statements across files, and deadlocks with no actionable diagnostic if they don't." That is an implicit convention holding up a correctness property.

**2. Core-schema writes are not rolled back on failure.**

Dropping the app's own schema cannot reach rows its migrations wrote into `__mj`. `RemoveAppEntityMetadata` recovers what is FK-reachable from the app's entities, which in practice is most of it — but not all (see 4).

**3. Nothing validates that an app's migrations write to the core schema at all.**

The engine has no view of an app's intended footprint. An app can write anywhere in `__mj` and the installer will neither know nor object.

**4. Free-floating core rows carry no ownership marker.**

[...]

PR #3979, opened 2026-08-20
fix(diagnostics): report the cause, not the consequence, in four multi-app failure paths

### LINKED ISSUE #3975: CodeGen / migration / MJAPI-startup diagnostics: four failures that report the wrong cause or no cause at all in a multi-app workspace
## Summary

> ### ⚠️ Status as of 2026-09-02 — re-verified against `next` @ `ca6c2bd79e`
>
> | # | status |
> |---|---|
> | 1 | 🟠 **still open** — narrower now that #2 has landed |
> | 2 | ✅ **RESOLVED on `next`** — `STEP 2(c)` removed; permissions apply to `includedEntities` only |
> | 3 | 🔵 **fix proposed** — draft PR #4171 |
> | 4 | ✅ **RESOLVED on `next`** — per-package try/catch in `loadDynamicAppPackages`, names the package, no longer takes the API down |
>
> Verification detail in the comment below. Happy to close this and reopen item 1 as its own issue if
> that suits the workflow better — say the word.

Four failures we hit in one end-to-end run of a five-app workspace on `next`. They are separate code paths but one shape: **the failure output names something other than the cause, or names nothing at all.** Each cost 20+ minutes that a single line of output would have saved, so they are grouped — the fix for each is small and they are cheapest to do together.

| # | what happens | what it should say |
|---|---|---|
| 1 | 🟠 CodeGen runs an entity's generated **permissions** files even when the object's creation failed, and the GRANT error is the loudest output | the first failure in the batch |
| 2 | ✅ ~~CodeGen **deliberately** generates and executes permissions for *excluded* entities, so one app's missing SQL objects fail every co-installed app's CodeGen~~ | nothing — this should not reach past the exclusion, or should tolerate a missing object |
| 3 | 🔵 an app migration failure is reported as a bare `Transaction has been aborted.` — *draft PR #4171* | the migration file and the first SQL error |
| 4 | ✅ ~~MJAPI refuses to start with a non-`Error` rejection and **zero** output~~ | which `dynamicPackages` entry failed to load, and why |

Environment: MJ `next` @ `e45fde5b2c` (`6.1.0-edge.2`), SQL Server 2022, Node 24.16.0, pnpm 10.33.0, macOS. Workspace holding MJ + `bizapps-{common,tasks,accounting,orders,contracts}`.

---

## 1. 🟠 STILL OPEN — Permissions files are executed for objects whose creation failed, and the GRANT error shadows the cause

> **Narrower than when filed.** The cross-app amplification came from item 2, which is now fixed, so
> this is the within-app case only: an object fails to create, its generated permissions file still
> runs, and the resulting `Cannot find the object` GRANT error is the loudest thing in the output while
> the real first failure is one line among hundreds. `applyPermissions` already collects per-entity
> results, so the information for a first-failure report is in hand.

On `bizapps-common`, the batch that creates `spCreateOrganization` / `spUpdateOrganization` failed. CodeGen then executed the generated permissions files for those procs and ended the run on them:

```
View or function '__mj_BizAppsCommon.vwOrganizations' has more column names specified than columns defined.   ← the real cause, one line among hundreds
Cannot find the object 'spCreateOrganization', because it does not exist or you do not have permission.
Error executing permissions file …/__mj_BizAppsCommon/spCreateOrganization.sp.permissions.generated.sql for entity MJ_BizApps_Common: Organizations: Cannot find the object 'spCreateOrganization', …
Error executing permissions file …/__mj_BizAppsCommon/spUpdateOrganization.sp.permissions.generated.sql for entity MJ_BizApps_Common: Organizations: Cannot find the object 'spUpdateOrganization', …
Error executing one or more permissions files in batch starting from index 10
CodeGen exited 1
```

The two GRANT failures and the batch-index line are what an operator sees; the view-compilation error that actually caused it is not called out anywhere.

**Suggested fix:** skip a permissions file whose target object was not created in this run, and surface the **first** failure in the batch as the reported cause (`applyPermissions` already collects per-entity results — `sql_codegen.ts:504-548` — so the information is in hand).

## 2. ✅ RESOLVED — Permissions are generated and executed for *excluded* entities — so one app's break fails every co-installed app

> **Fixed on `next`.** `STEP 2(c)` is gone, replaced by a comment stating the reasoning, and `STEP 4`
> now applies permissions for `includedEntities` only — with the sibling-schema case named explicitly.
> Nothing further needed here. Original report kept below for the record.

This one is deliberate, which is why it is worth a decision rather than a patch. `sql_codegen.ts:258`:

```ts
// STEP 2(c) - for the excludedEntities, while we don't want to generate SQL, we do want to generate the permissions files for them
```

…and STEP 4 (`applyPermissions`, `:504`) then executes every `.permissions` file it finds for the entity list, excluded ones included.

In a single-app database that is harmless. In a **multi-app** database it is not: the excluded entities belong to *other* apps, whose `spCreate*` / `spUpdate*` may not exist yet — and on a fresh multi-app install that is always true for whichever app runs CodeGen first. Observed:

```
bizapps-tasks/SQL Scripts/generated/__mj_BizAppsCommon/spCreateOrganization.sp.permissions.generated.sql   → Cannot find the object
bizapps-accounting/SQL Scripts/generated/__mj_BizAppsOrders/spCreateOrderHeader.sp.permissions.generated.sql → Cannot find the object
bizapps-contracts/SQL Scripts/generated/__mj_BizAppsOrders/spCreateEventOrderLine.sp.permissions.generated.sql → Cannot find the object
```

Each of those failed the whole CodeGen run (exit 1), which failed the app's activation.

**Worth stating explicitly, because it is the obvious first guess and it is wrong:** this is **not** a schema-scoping misconfiguration. `excludeSchemas` was correctly set for every app — each app's config excluded every sibling app schema plus `__mj`, verified per app — and `includeSchemas` (`Database/schema-scope.ts`, which resolves a positive scope into `excludeSchemas`) would make no difference, because STEP 2(c) intentionally acts on the entities the exclusion just removed. So no consumer configuration can avoid it.

**Suggested fix — a choice for you rather than a prescription:** either don't emit/execute permissions for out-of-scope entities at all, or treat a missing object in that specific pass as a skip-with-note rather than an error that fails the run. The second keeps whatever the STEP 2(c) behaviour was for (presumably: keep GRANTs current for entities this package does not own) while making it survivable when the owning app has not generated yet.

## 3. 🔵 FIX PROPOSED (#4171) — An app migration failure reported as a bare `Transaction has been aborted.`

> **Draft PR: #4171.** Root cause turned out to be on the consumer side of a good API: Skyway already
> attaches the script, the failed batch's number and line range, the count of batches that committed
> first, and the driver error as `cause` — but `migration-runner.ts` typed the Skyway result with a
> minimal structural interface that omitted the per-migration `Error` field, so all of it was
> discarded and only the run-level string was reported.

The **entire** message for a failed app migration — 12 events for the whole run:

```
[Fatal] migrations failed: Migration failed for schema '__mj_BizAppsContracts': Transaction has been aborted.
```

No migration filename, no SQL error number, no object name. We found the cause only by extracting the baseline and running it by hand:

```
Msg 1767, Level 16: Foreign key 'FK_ContractLine_Product' references invalid table '__mj_BizAppsOrders.Product'.
Msg 1750, Level 16: Could not create constraint or index. See previous errors.
```

**In fairness, other paths do this well** — a later failure in the same file reported `Failed at batch 2/253 (lines 50-71): Violation of PRIMARY KEY constraint 'PK_SchemaInfo_ID' …`, which is exactly right. So the ask is narrow: **the transaction-abort path loses the underlying error; carry the first SQL error and the migration file through it.**

Same run, same shape, from the installation-record path: `Failed to rollback after operation error: TransactionError: Transaction has been aborted.` — the compensation path failing without reporting what it rolled back (this is #3547's territory).

## 4. ✅ RESOLVED — MJAPI refuses to start with no error message at all

> **Fixed on `next`.** Each package load in `loadDynamicAppPackages`
> (`packages/ServerBootstrap/src/index.ts`) is wrapped in its own try/catch that names the package —
> `Open App server package not found (run 'npm install'?): <pkg>` for a resolution failure, and
> `Error loading Open App server package <pkg>:` otherwise — and the loop continues instead of taking
> the API down. Nothing further needed here. Original report kept below for the record.

The **entire** output of a failed MJAPI start:

```
node:internal/modules/run_main:107
    triggerUncaughtException(
    ^
[Object: null prototype] {
  Symbol(nodejs.util.inspect.custom): [Function: [nodejs.util.inspect.custom]]
}
Node.js v24.16.0
```

No message, no stack, no code, no package name. The last log line before the crash was `Loading Open App server packages...`. We attached an `uncaughtException` handler that serialises own properties and confirmed the rejected value carries nothing: `IMPORT REJECT: {}`.

Cause, found by importing each registered server package by hand:

```
OK   @mj-biz-apps/common-server        CLASS_REGISTRATIONS,LoadBizAppsCommonServer,RESOLVER_PATHS
OK   @mj-biz-apps/tasks-server         LoadBizAppsTasksServer,…
OK   @mj-biz-apps/accounting-server    CLASS_REGISTRATIONS,…
FAIL @mj-biz-apps/orders-server        Cannot find package '…/MJAPI/node_modules/@mj-biz-apps/orders-server/'
OK   @mj-biz-apps/contracts-server     CreatemjBizAppsContracts…
```

One registered app's server package was not built (no `dist/`) while `mj.config.cjs`'s `dynamicPackages.server` listed it `Enabled: true`. **A single unloadable `dynamicPackages` entry takes the whole API down and reports nothing.**

**Suggested fix:** wrap each package's load in the `Loading Open App server packages…` step and, on failure, report the package name, its declared `startupExport`, and the underlying resolution error — then decide loudly whether to continue without it or refuse. Rejecting with a non-`Error` value leaves a consumer with no first move.

*(The consumer-side half is ours: our tooling enables the `dynamicPackages` entry at registration time, before the app's packages are built, so a failed activation can leave an instance that cannot boot. We are fixing that. But an unloadable entry should be legible regardless of who wrote it.)*

---

## Why these were grouped

Each is a few lines of error handling, all four are in the same "what does the operator see when this fails" surface, and in our run they compounded: #2 produced the output that hid #1, whose cause was a fourth defect entirely (unseeded `EntityField.Configuration` — filed separately as #3970), and #4 hid a plain missing build. Fixing them individually is fine; fixing them as a set is cheaper and the test for all four is the same multi-app workspace.

Happy to split into four if that suits the workflow better — and with two now resolved and one in
review, splitting out item 1 is probably the tidiest end state.

MarceloT-BC:
Fixes #3975.

Four failures from one end-to-end run of a five-app workspace on `next` (`e45fde5b2c`, `6.1.0-edge.2`). Separate code paths, one shape: **the failure output names something other than the cause, or names nothing at all.** Diagnostics only — no public signature is removed or changed, and each fix leaves the pass/fail outcome it reports alone.

---

### Decision 1 — §2: which of the two options did we want?

The issue offered a choice rather than a prescription, and this PR picks one. `sql_codegen.ts:258` (yours, 2024-06-18) says:

```ts
// STEP 2(c) - for the excludedEntities, while we don't want to generate SQL, we do want to generate the permissions files for them
```

…and STEP 4 then executes every one of them. In a single-app database that is harmless. In a multi-app database those entities belong to *other* apps, so on a fresh multi-app install whichever app runs CodeGen first fails its whole run on a sibling app's not-yet-generated `spCreate*`. **No consumer configuration can avoid it** — `excludeSchemas` was correctly set for every app; the behaviour acts on precisely the entities the exclusion just removed.

| | |
|---|---|
| **Option A** | Don't emit or execute permissions for out-of-scope entities at all |
| **Option B — implemented here** | Keep emitting, but treat a *missing target object* on an out-of-scope entity as skip-with-note instead of a run-failing error |

I chose **B** because it preserves whatever STEP 2(c) exists for (presumably: keep GRANTs current for entities this package doesn't own) while making it survivable when the owning app hasn't generated yet. Any *other* error out of scope — a real permission problem, a bad login — still fails the run, and in-scope entities are untouched. **If A is what you actually want, say so and I'll switch it** — it's a smaller diff, but it changes behaviour you put there deliberately, which is why I didn't assume.

### Decision 2 — §4: should a failed `dynamicPackages` entry refuse the boot, or continue?

The issue asked us to "decide loudly whether to continue without it or refuse." **This PR changes only the output, not the outcome** — the fatal path still refuses to start, the catchable path still continues, exactly as today. The argument for making it *always* refuse: a registered app whose server package didn't load leaves the API silently missing that app's entities, resolvers and class registrations, which is arguably worse than not booting. The argument against: that would newly break deployments that currently boot with a stale entry. It's a one-line change either way — your call, and I'd rather it be an explicit ruling than a side effect of this PR.

*(The consumer-side half of §4 is ours, as noted in the issue: our tooling enables the `dynamicPackages` entry at registration time, before the app's packages are built. We're fixing that separately. An unloadable entry should be legible regardless of who wrote it.)*

---

### §1 — a failed GRANT is reported as a consequence, and the cause is named

On `bizapps-common`, the batch creating `spCreateOrganization`/`spUpdateOrganization` failed; CodeGen then ran the generated permissions files for those procs and ended the run on them. The two GRANT failures and a bare batch-index line were what the operator saw. The view-compilation error that actually caused it — `View or function '__mj_BizAppsCommon.vwOrganizations' has more column names specified than columns defined.` — was one line among hundreds.

Two reasons it was invisible, both addressed:

- **`SQLServerCodeGenProvider.executeSQLFileViaShell` swallowed every batch error** into `logWarning(... msg.substring(0, 200))` and always returned `true`. Non-fatal is correct — one entity's broken view must not abort the other several hundred, and the post-run CRUD validator is the authoritative gap report — but a *truncated warning* is how the cause got lost. It now logs in full, with the batch number and the `CREATE`/`ALTER` target it was building.
- **Nothing connected the GRANT failure to the creation failure.** New `SQLExecutionDiagnostics` (`packages/CodeGenLib/src/Database/sql-execution-diagnostics.ts`) records each failed batch as it happens, so `applyPermissions` can look up the object and report `This is a CONSEQUENCE, not the cause` with the earlier error inline. The run's summary now leads with the **first** failure.

A static collector is deliberate and documented in the file: CodeGen is a single-run CLI process, the recording site (a provider's file executor) and the reading site (`applyPermissions`) are far apart in the call graph, and threading a context object between them would touch a dozen signatures for one diagnostic. `Reset()` keeps it test-addressable and is called at the start of each run.

### §2 — one app's break no longer fails every co-installed app

See Decision 1. `applyPermissions` gains an **optional** 5th parameter (`outOfScopeEntityIDs`) — omit it and every entity is treated as owned, so the signature is source-compatible.

### §3 — migration failures carry the file and the first SQL error

The entire message for a failed app migration was:

```
[Fatal] migrations failed: Migration failed for schema '__mj_BizAppsContracts': Transaction has been aborted.
```

**Root cause is in `@memberjunction/skyway-core@0.6.2`, not in this repo**, and it is a masking chain worth stating because the fix looks redundant otherwise:

1. A batch fails with the real error (`Msg 1767: Foreign key 'FK_ContractLine_Product' references invalid table '__mj_BizAppsOrders.Product'`).
2. Skyway builds a rich `MigrationExecutionError` for it and heads for `txn.Rollback()`.
3. The transaction is already doomed, so **the rollback itself throws** `TransactionError: Transaction has been aborted.`
4. That throw escapes `executeMigrationsWithHistory`, and `Migrate()`'s outer catch returns `{ Details: [], ErrorMessage: 'Transaction has been aborted.' }`.

The good diagnosis is destroyed by the compensation path that ran *because of* it. `OnProgress.OnMigrationEnd` fires with the rich result **before** step 3 — verified live against a real SQL Server — so `RunAppMigrations` now registers it and prefers the captured failure, falling back to the first failed `Details` entry (present when the rollback succeeded) and then to skyway's own message.

A **second** masking layer sits underneath: `mssql` rejects with the *last* error of a chain and parks the earlier ones on `precedingErrors`, so the surviving text was literally `Could not create constraint or index. See previous errors.` — a pointer to output the operator was never shown. `FirstDatabaseError` walks back to the first one.

```diff
- Migration failed for schema '__mj_ReproApp': Transaction has been aborted.
+ Migration failed for schema '__mj_ReproApp': V202601020000__Bad_FK.sql — Failed at batch 1/1
+ (lines 1-8): Could not create constraint or index. See previous errors. [first database error:
+ Foreign key 'FK_WidgetLine_Product' references invalid table '__mj_NoSuchApp.Product'.]
+ [run terminated with: Transaction has been aborted.]
```

Two follow-ups belong in `MemberJunction/skyway`, not here, and I'll file them separately: the rollback discarding the failure it was rolling back for, and `precedingErrors` never reaching `MigrationExecutionError.message`. The compensation-path half is #3547's territory.

### §4 — an unloadable `dynamicPackages` entry is legible

The entire output of a failed MJAPI start was:

```
node:internal/modules/run_main:107
    triggerUncaughtException(
    ^
[Object: null prototype] {
  Symbol(nodejs.util.inspect.custom): [Function: [nodejs.util.inspect.custom]]
}
```

Reproduced under MJAPI's real loader (`ts-node/esm` registered through `node:module` — see `packages/MJAPI/register.js`) and fully characterised. The thrown value is a **null-prototype object with zero own properties**, carrying only `Symbol(nodejs.util.inspect.custom)`:

- `value instanceof Error` → `false`; no `message`, no `code`, no `stack`. It cannot identify itself.
- **`String(value)` throws** `TypeError: Cannot convert object to primitive value` — so a diagnostic that stringifies naively kills the process while trying to explain why the process is dying. `console.warn('msg', value)` renders it as `{}`.
- On the fatal path **no `try`/`catch` can see it** — it is raised at `run_main` before the rejection reaches user code. Only a process-level handler can, which is exactly what the issue's own `IMPORT REJECT: {}` experiment demonstrated.

Worth knowing beyond this issue: **the stripping is not specific to package resolution.** Anything crossing the module-hooks thread arrives that way — I hit byte-identical output from a *TypeScript compile diagnostic* while building the repro. That's the case for a legible handler regardless of cause.

The fix, in three parts:

- `describeThrown(value)` — renders an unknown thrown value safely, never coercing an unrecognised object, and says *"no message and no own properties — the ESM loader hook stripped the underlying resolution error"* rather than printing an empty one.
- `describeHostResolution(pkgName, configFilePath)` — computes **before** the import, via a `createRequire` resolver that never touches the loader hooks, whether the package is reachable and from where. Deliberately CJS-only and therefore a **diagnostic, never a gate**: nothing skips an import on the strength of it, because an `"import"`-only exports map would be reported unreachable even though `import()` could load it.
- A **phase-scoped** `uncaughtException`/`unhandledRejection` guard in `loadDynamicAppPackages`, installed before the loop and removed in a `finally`, so the fatal path can name the entry it was on.

```
Loading Open App server packages...
  Error loading Open App server package @mj-biz-apps/orders-server: Cannot find package '…/orders-server/' imported from …
  package        : @mj-biz-apps/orders-server
  app            : BizApps: Orders
  startup export : LoadBizAppsOrdersServer
  failure        : Cannot find package '…/node_modules/@mj-biz-apps/orders-server/' imported from …
  not reachable from any host anchor:
    - …/mj.config.cjs: Cannot find module '…/orders-server/dist/index.js'. Please verify that the package.json has a valid "main" entry (code MODULE_NOT_FOUND)
    - …/package.json: (same)
    - …/after.mjs: (same)
  most likely the package has not been built (no dist/), or 'npm install' has not run for it.
  1 Open App server package(s) declared in mj.config.cjs did NOT load — their apps' resolvers and class registrations are MISSING from this API:
    - @mj-biz-apps/orders-server: …
```

**One trade-off to be aware of** (the only place this PR could surprise a host): while the guard is installed, an unrelated uncaught exception raised during that narrow window exits 1 rather than reaching a host's own swallowing handler. Without the guard such an exception crashes the process anyway unless the host installed one, so the exposure is narrow — but it is real, and it's the reason the guard is phase-scoped rather than global.

---

## How to recreate the original issues

**§3** — real SQL Server, real Skyway, no MJ instance needed:

```sh
# migrations/B…__Baseline.sql creates a table; migrations/V…__Bad_FK.sql adds an FK to a
# table in a schema that does not exist (Msg 1767).
node -e "import('@memberjunction/open-app-engine').then(({RunAppMigrations}) => RunAppMigrations({
  MigrationsDir: './migrations', SchemaName: '__mj_ReproApp', Platform: 'sqlserver',
  DatabaseConfig: { Host: 'localhost', Port: 1433, Database: 'MJ_repro', User: 'sa',
                    Password: process.env.SA_PW, TrustServerCertificate: true },
}).then(r => console.log(r.ErrorMessage)))"
```

Before: `Migration failed for schema '__mj_ReproApp': Transaction has been aborted.`

**§4** — a `ts-node/esm` host (`register.js` identical to `packages/MJAPI/register.js`) whose `node_modules/@scope/pkg/` has a `package.json` with `"main": "dist/index.js"` and **no `dist/`**, plus a `dynamicPackages.server` entry naming it with `Enabled: true`. Before: the `triggerUncaughtException` block above, and nothing else. *(Note: a dynamic import inside a module that uses top-level `await` fails at link time under this loader, before the module body runs at all — so output before the crash is lost. MJAPI's entry is `createMJServer(...).catch(...)`, not a top-level await.)*

**§1/§2** — a fresh multi-app install where one app has not generated yet. On-disk evidence from the failing workspace, no re-run needed: `bizapps-contracts/SQL Scripts/generated/` holds permissions files for **five schemas it does not own** — 1,513 under `__mj`, 197 under `__mj_BizAppsOrders`.

---

## Testing

| Package | Result |
|---|---|
| `@memberjunction/codegen-lib` | **1087 passed**, 66 skipped, 0 failed |
| `@memberjunction/open-app-engine` | **368 passed**, 0 failed |
| `@memberjunction/server-bootstrap` | **37 passed**, 0 failed |

All three build clean (`tsc && tsc-alias`). Three new suites, each written from the expected behaviour and **red-proven against the unfixed code before the fix landed**:

- `apply-permissions-diagnostics.test.ts` — 4 tests against the real `applyPermissions`. Red baseline confirmed both defects: it returned `false` on an out-of-scope missing object (failing the whole run), and its output was `Error executing permissions file…` with no mention of the real cause and no first-failure line.
- `migration-failure-message.test.ts` — 8 tests covering both masking layers, including an explicit regression guard that the bare `Transaction has been aborted.` shape can never come back.
- `dynamic-package-diagnostics.test.ts` — 7 tests, including that `String()` on the real opaque value throws while `describeThrown` does not, and that an unbuilt package (package.json present, entry file missing — the reported case) is reported unreachable rather than resolved.

**Verified live, before and after:** §3 (real SQL Server + real Skyway) and §4 (real `ts-node/esm` loader; the before-state reproduced the reported crash output byte-for-byte). **Not** live-verified: §1/§2, which need a fully-migrated multi-app database plus a CodeGen config — they rest on the red→green tests against the real method plus the on-disk artifacts above. Happy to do the live CodeGen run if a reviewer wants it.

**One pre-existing test updated, and it was stale rather than failing:** `server-bootstrap.test.ts`'s transitive-dependency test pinned the old two-argument `console.warn(msg, error)` shape. The behaviour it guards — the transitive dep's true cause reaching the operator — is preserved and improved; the cause now lands *inside* the message string (because a non-Error in that second position renders as `{}`), so the assertion reads the rendered output and additionally checks the entry is named.

## Related

- Fixes #3975
- #3970 — the unseeded `EntityField.Configuration` defect that produced §1's underlying view-compilation error; filed separately, not fixed here
- #3547 — the compensation-path reporting §3's second half touches
- #3677 — the pnpm host-anchored dynamic-import work that `host-import.ts` came from (closed)


MarceloT-BC:
Closing: superseded. Against current `next` (`5f4961e4ea`, 1,673 commits past this branch's base):

- **§2** is fixed upstream: STEP 4 applies permissions to `includedEntities` only. That's Option A, not the Option B implemented here.
- **§3** is covered by #4171.
- **§4**: the loader moved to `@memberjunction/dynamic-packages`, so this diff edits the wrong file.

Only **§1** (the GRANT failure hides the real creation error) is still unfixed. That will come as a smaller PR off current `next`. Tracking stays on #3975.

PR #3987, opened 2026-08-21
fix(realtime): a bridged seat runs its own agent, not an arbitrary registered one

### LINKED ISSUE #4111: Bridged realtime seat resolves its agent class against the wrong ClassFactory registry, so it silently runs plain BaseAgent
## Summary

`CreateBridgeRealtimeSession` picks the `BaseAgent` subclass for a bridged realtime seat using the wrong ClassFactory registry, so every agent that declares no `DriverClass` of its own silently runs the plain `BaseAgent` instead of whatever it should have been.

## The mechanism

```ts
const agentType = AIEngine.Instance.AgentTypes.find(t => UUIDsEqual(t.ID, agent.TypeID));
const driverClass = agent.DriverClass || agentType?.DriverClass;   // ← two different registries
const instance = MJGlobal.Instance.ClassFactory.CreateInstance<BaseAgent>(BaseAgent, driverClass);
```

`AIAgentType.DriverClass` names a **`BaseAgentType`** subclass — the three shipped values are `LoopAgentType`, `FlowAgentType`, `RealtimeAgentType`, all registered via `@RegisterClass(BaseAgentType, …)`. The key this call needs is a **`BaseAgent`** one (`QueryBuilderAgent`, `MemoryManagerAgent`, …). `ClassFactory` matches keys exactly and per base-class *name*, so the fallback can never resolve — it is dead code that looks alive.

Most agents declare no `DriverClass` at all (that is what makes an agent data rather than code), so most agents take the fallback.

## Why it is silent

`ClassFactory.CreateInstance` does not return null for an unmatched key. With no registration and no `@RequiresSubclass` marker on `BaseAgent`, `resolveAndInstantiate` returns `new BaseClassConstructor(...)` — a plain `BaseAgent`. So:

- the `if (!instance) throw` below the call is unreachable;
- the `if (!driverClass) throw` above it is unreachable too, because the wrong-registry lookup always produced a truthy key;
- the only signal is a single `console.warn` from `reportResolutionFailure`, deduped per `base|key` and capped at 3 per base — effectively invisible in a busy log.

Observed on a live LiveKit room: every seat sounded like one generic assistant instead of its configured character.

## Related

The same `agent.DriverClass || agentType.DriverClass` → `CreateInstance` pattern, with the same dead `if (!agentInstance) throw`, is in `AgentRunner.RunAgent` (`packages/AI/Agents/src/AgentRunner.ts` ~100-113) — the path every non-realtime run takes. Tracking separately.

MS-BC:
Fixes #4111.

## The defect

`CreateBridgeRealtimeSession` picked the agent class like this:

```ts
const agentType = AIEngine.Instance.AgentTypes.find(t => UUIDsEqual(t.ID, agent.TypeID));
const driverClass = agent.DriverClass || agentType?.DriverClass;   // ← two different registries
const instance = MJGlobal.Instance.ClassFactory.CreateInstance<BaseAgent>(BaseAgent, driverClass);
```

`AIAgentType.DriverClass` names a **`BaseAgentType`** subclass. The key this call needs is a
**`BaseAgent`** one. They are separate ClassFactory registries, so for any agent that declares no
`DriverClass` of its own — most agents, because that is what makes an agent data rather than code —
the *type's* key went to `CreateInstance(BaseAgent, …)`, matched nothing, and fell back.

And `CreateInstance` has never returned null for an unmatched key (that is exactly why
`TryCreateInstance` exists), so:

- the `if (!instance) throw` below it was unreachable — with no registration and no `@RequiresSubclass` marker, `resolveAndInstantiate` returns `new BaseClassConstructor(...)`, i.e. a plain `BaseAgent`;
- the `if (!driverClass) throw` above it was unreachable too — it only ever *looked* satisfied because the wrong-registry lookup always produced a truthy key;
- the only signal was a single `console.warn` from `reportResolutionFailure`, deduped per `base|key` and capped at 3 per base — effectively invisible in a busy log.

## How it presented

Every affected seat silently ran the **plain `BaseAgent`**, dropping the subclass behaviour it was configured for: on a live LiveKit room, one generic assistant voice for every seat in a room meant to hold distinct characters. From the outside that reads as a bad prompt, so it costs debugging time in the wrong file. A silent downgrade is worse than an exception.

*(Corrected from an earlier claim in this PR that the seat ran `QueryBuilderAgent` — see @rkihm-BC's review. Landing on another agent's class requires a **null** key, where `GetAllRegistrations` skips its key filter and the highest-priority registration wins; the old `if (!driverClass) throw` prevented exactly that.)*

## The fix

- Use the agent's **own** `DriverClass` when it has one, resolved with `TryCreateInstance` so an
  unregistered key raises rather than installing a hollow anchor-base object that answers plausibly.
- Use the plain `BaseAgent` when it has none — the behaviour those agents were always meant to get.
- Drop the `no DriverClass` throw: an absent `DriverClass` is the common, correct case.

Comments record *why* the type's key must not be used here, so the next reader does not restore the
fallback as a convenience.

### Two things worth stating explicitly

**Dropping the agent-type fallback loses nothing.** Agent-type behaviour is resolved separately inside `BaseAgent` via `BaseAgentType.GetAgentTypeInstance` (`base-agent.ts:1162`, `:3439`, `:3560`), so a bridged seat on the plain `BaseAgent` still gets Loop/Realtime type semantics. This is the first thing a reviewer worries about, so: it was checked.

**`new BaseAgent()` and not `CreateInstance(BaseAgent, null)`.** A null key makes `GetAllRegistrations` skip its key filter entirely, so the factory returns the highest-priority registered `BaseAgent` subclass — an arbitrary agent, which is the very failure this PR exists to stop. Direct construction is the only form that reliably yields the base. A call-count spy in the specs pins it: rewriting the else-branch as a null-key lookup fails 2 tests.

## Follow-up

**#4112** — `AgentRunner.RunAgent` has the same `agent.DriverClass || agentType.DriverClass` → `CreateInstance` pattern and the same dead guard, on the path every non-realtime run takes. Not a merge condition for this PR.

## Verification

Applied to a running deployment (patched build) and exercised with a three-seat LiveKit room: each
seat resolved to the agent its seat names and answered in its own character. Before the fix, the same
room produced one generic assistant voice for every seat.

Downstream context: found while building Caliber's multi-party interview panel, where each seat is a
distinct persona — the failure is most visible there because the seats are *supposed* to differ.


MS-BC:
Reopened after a criterion pass over the realtime backlog: this one was mis-filed with the feature closes. The bridge factory ships in core regardless of whether Caliber still launches rooms, and a seat instantiating whichever agent class happens to be registered — rather than the one the seat names — is a correctness defect on its own terms (and security-adjacent: an unconfigured agent running with the seat's access). One file, +33/−11; judge it as a core fix, not a Caliber feature.

rkihm-BC (commented):
## Review

The fix is correct and the diagnosis holds up against the code. One change needed before merge, and it is text-only — the code change itself I'd take as-is.

Reviewed against `origin/next` @ `91b1640c26` and PR head `e98fa4e86f`. Ran `bridge-realtime-session-factory.test.ts` in a clean worktree (7/7 pass) and typechecked `packages/AI/Agents` (clean). CI is green.

---

### Required: fix the changeset + the code comment

**1. `#3860` is the wrong issue.** The changeset headline (`... an arbitrary registered one (#3860)`) and the test docblock (`## What #3860 changed here`) both cite it. #3860 is *"MJDropdown has no ariaLabel/inputId passthrough — every dropdown is an unnamed combobox (WCAG 4.1.2)"* — unrelated, still open. The changeset text ships to published release notes.

**2. The stated mechanism is not what `ClassFactory` does.** Both the permanent comment in `bridge-realtime-session-factory.ts` and the changeset say the wrong-registry key made `CreateInstance` *"silently hand back an unrelated agent — a seat configured as an interviewer answered as `QueryBuilderAgent`"*. `resolveAndInstantiate` doesn't do that: with `reg === null` and no `@RequiresSubclass` marker it returns `new BaseClassConstructor(...)` — i.e. plain `BaseAgent`. Landing on *someone else's* agent class requires a **null** key (`GetAllRegistrations` skips the key filter, `resolveRegistration` returns the highest-priority registration), and the old code's `if (!driverClass) throw` prevented exactly that.

Persona also doesn't come from the instance class: `StartBridgeRealtimeSession` passes `CoAgent: params.agent` into `PrepareRealtimeSessionParams` (`base-agent.ts:1868`+), and no in-repo subclass overrides that method. So the real old failure mode was **"silently ran plain `BaseAgent`, dropping any subclass overrides"** — which matches this PR's own verification note ("one generic assistant voice for every seat") better than the `QueryBuilderAgent` story does.

This matters beyond pedantry: the same framing ("worse than an exception… would run a different agent in this agent's voice") is what justifies the new throw, and the comment is written to stop a future reader restoring the fallback. Worth being exactly right.

**3. Same paragraph: "nothing logged, warned, or failed" isn't accurate either.** `reportResolutionFailure` emits a `console.warn` whenever a key was supplied and the base carries neither `@RequiresSubclass` nor `@OptionalKeyedSpecialization` (`ClassFactory.ts:436-439`) — `BaseAgent` carries neither, and that instrumentation landed 2026-07-19/20, before this branch. In fairness the warning is deduped per `base|key` and capped at 3 per base, so it's one line per driver key per process: effectively invisible in a busy log, which is the point being made. "Effectively invisible" rather than "nothing logged" would be both true and just as damning.

Suggested replacement for the comment's core claim: *the type's key names a `BaseAgentType`, so it matched nothing in the `BaseAgent` registry and every such seat silently ran the plain base implementation, dropping any subclass overrides — with a single capped `ClassFactory` warning as the only signal.*

(If Caliber registers a `BaseAgent` subclass under a key that collides with an agent-type `DriverClass`, that IS a path to the observed symptom — worth stating explicitly, since nothing in core produces it.)

---

### What's good

The two-registry diagnosis checks out end to end: `AIAgentType.DriverClass` is resolved against `BaseAgentType` (`base-agent-type.ts:267`), `ClassFactory` matches keys exactly and per base-class *name* (`ClassFactory.ts:448-455`), and the three shipped agent types are `LoopAgentType` / `FlowAgentType` / `RealtimeAgentType` — none registered as a `BaseAgent` key. The old fallback could never resolve. Dead code that looked alive.

**Dropping the type fallback loses nothing**, and it's worth saying so in the PR body since it's the first thing a reviewer will worry about: agent-type behaviour is resolved separately inside `BaseAgent` via `BaseAgentType.GetAgentTypeInstance` (`base-agent.ts:1162`, `:3439`, `:3560`), so a bridged seat on plain `BaseAgent` still gets Loop/Realtime type semantics.

**`new BaseAgent()` rather than `CreateInstance(BaseAgent, null)` is the subtle, correct call** — and the PR undersells it. A null key returns the highest-priority registered `BaseAgent` subclass, i.e. an arbitrary agent class. Direct construction is the only thing that sidesteps that, and it's the part of this change I'd most want preserved.

`TryCreateInstance` is used correctly — `Resolved ? Instance : null` properly discards the base-class fallback instance that comes back on `Resolved: false`. The error is actionable, `@throws` was updated, the changeset is scoped to the one package, and the `BaseAgent` module mock is justified (a real one builds an `AIPromptRunner` that reaches for metadata). The throw for a declared-but-unregistered driver propagates cleanly out of `StartRoomSession` (`livekit-agent-room-coordinator.ts:214`) to the resolver — loud, and scoped to the one seat, which is the right shape for this failure.

---

### Non-blocking

**The spec can't distinguish "not consulted" from "consulted with a null key."** The mock records the key it was handed (`createdDriverClass = driverClass`), so rewriting the else-branch as `TryCreateInstance(BaseAgent, null)` leaves `createdDriverClass === null`, `instanceToReturn` is non-null from `beforeEach`, and **both** new tests still pass — while the runtime quietly becomes "arbitrary highest-priority agent," the exact hazard `new BaseAgent()` exists to avoid. The comment `// the ClassFactory was not consulted at all` asserts something the assertion doesn't check. A call-count spy closes it in a line.

**The same pattern is still live in `AgentRunner`** (`AgentRunner.ts:100-113`) — the path every non-realtime run takes, including the target agent this co-agent delegates to via `invoke-target-agent`. Same `agent.DriverClass || agentType.DriverClass`, same `CreateInstance`, same dead `if (!agentInstance) throw`. And `AIAgentType.DriverClass` is `string | null` in the generated ORM, so the null-key path is reachable in principle (latent only because all three shipped types seed a value). Not a merge condition for this PR — but after it lands the two paths resolve agents differently with nothing recording which is intended, so a linked follow-up would be worth it.

**Nits.** `agent.DriverClass || undefined` is the one untrimmed string in a file where `RealtimeVoice` / `RealtimeModelID` / `AgentSessionID` are all trimmed; a whitespace-only value becomes a truthy key and throws with a confusing message. And the `agentTypes[1].DriverClass` restore sits at the end of a test body rather than in `beforeEach` — harmless today (the factory no longer reads `AgentTypes` at all), so purely durability.

MS-BC:
Thanks @rkihm-BC — this was a genuinely better read of `ClassFactory` than mine. All three required points are fixed, and all four non-blocking ones taken. I verified each against the source rather than taking them on trust; you're right on every one.

## 1. Wrong issue — fixed

`#3860` is the MJDropdown accessibility issue. Worse than a typo: **no issue existed for this at all**, so there was nothing to correct it *to*. Filed **#4111** and cited it in the changeset, the test docblock and the source comment. You're right that this mattered more than it looks — the changeset text ships to release notes.

## 2. The mechanism was wrong — fixed, and you're right about why it matters

I traced it. With `reg === null` and no `@RequiresSubclass`, `resolveAndInstantiate` returns `new BaseClassConstructor(...)` — plain `BaseAgent`. Landing on someone *else's* agent needs a **null** key, where `GetAllRegistrations` skips its filter outright (`normalizedKey === null ? true : …`) and the highest-priority registration wins — and the old `if (!driverClass) throw` prevented exactly that path.

So the real old failure is **"silently ran plain `BaseAgent`, dropping subclass overrides"**, which also matches this PR's own verification note ("one generic assistant voice for every seat") far better than my `QueryBuilderAgent` story did. Corrected in the comment, the changeset and the test docblock.

The null-key hazard is now stated where it earns its place: as the reason the else-branch constructs `new BaseAgent()` rather than `CreateInstance(BaseAgent, null)`.

## 3. "Nothing logged" — fixed

Also wrong. `reportResolutionFailure` fires whenever a key was supplied and the base carries neither marker, and `BaseAgent` carries neither. Deduped per `base|key` and capped at 3 per base, so it's one line per driver key per process. Now phrased as *"a single capped `ClassFactory` warning as the only signal — effectively invisible in a busy log"*. Agreed that's just as damning and has the advantage of being true.

## Non-blocking — all four taken

**The spec couldn't tell "not consulted" from "consulted with a null key."** This was the best catch in the review: the comment asserted something the assertion didn't check, and the gap was precisely the hazard `new BaseAgent()` exists to avoid. Added a call-count spy. Verified it closes the hole — rewriting the else-branch as a null-key `TryCreateInstance` now **fails 2 tests**, where before it passed both.

**The trim.** Done, and it needed the test you'd expect: dropping the trim survived every existing spec. There's now a case for a whitespace-only `DriverClass`, and removing the trim fails it.

**Fixture restore** moved into `beforeEach` — a mid-body failure could otherwise leak it into later specs.

**`AgentRunner`.** Filed **#4112**, including the reachable null-key path there (`AIAgentType.DriverClass` is `string | null` in the generated ORM, latent only because all three shipped types seed a value). Agreed it's not a merge condition here.

## Added to the PR body, per your note

Both of the things you said were undersold:

- **Dropping the type fallback loses nothing** — agent-type behaviour resolves separately inside `BaseAgent` via `BaseAgentType.GetAgentTypeInstance` (`base-agent.ts:1162`, `:3439`, `:3560`), so a seat on plain `BaseAgent` still gets Loop/Realtime semantics.
- **`new BaseAgent()` over `CreateInstance(BaseAgent, null)`** is now called out in the body, the source comment and the spec, since it's the part most at risk of being "simplified" away later.

8 specs pass; `packages/AI/Agents` typechecks clean on the touched file.

rkihm-BC (approved):
Approving. All three required items are fixed and all four non-blocking ones taken. I verified the one that mattered most by running it rather than reading it. Waited for a clean board before posting.

## The three required

**The issue reference.** `#3860` → `#4111` in the changeset, the source comment and the test docblock. The detail worth keeping is the one you volunteered: no issue existed for this at all, so there was nothing to correct it *to* — filing #4111 was the actual fix, and the citation followed. Both #4111 and #4112 exist and are titled accurately. The only surviving `3860` on the branch is in `dropdown-accessible-name.md`, a different changeset where citing the MJDropdown issue is correct.

**The mechanism.** Now stated precisely, and the sentence that matters is the negative one: *"It did not run some other agent's class."* A comment written to stop a future reader restoring the fallback has to be exactly right about what the fallback did, or the next person corrects the comment instead of keeping the guard. Moving the null-key hazard to where it earns its place — as the stated reason the else-branch constructs `new BaseAgent()` rather than `CreateInstance(BaseAgent, null)` — is better placement than my note asked for.

**"Nothing logged."** Now *"one `console.warn` … deduped per base+key and capped at 3 per base: effectively invisible in a busy log, which is the whole problem."* True, and it loses none of the force of the original claim.

## The non-blocking four — and the one I checked by running it

The call-count spy was the catch I cared about: the old spec asserted `createdDriverClass === null`, which a null-key `TryCreateInstance` satisfies just as well, so the comment asserted something the assertion did not check — and the gap was precisely the hazard `new BaseAgent()` exists to prevent.

I ran the mutation myself in a clean worktree rather than taking the result on trust:

- unmutated: **8/8 pass**
- with the else-branch rewritten as `TryCreateInstance(BaseAgent, null).Instance`: **3 tests fail**, all `expected 1 to be +0`
- `@memberjunction/ai-agents` typechecks clean

Three rather than the two you reported — the whitespace-trim spec asserts the counter as well, so it kills the mutant too. The hole is closed, and closed in a way that keeps working as specs are added.

The trim landed with the test that proves it (`treats a whitespace-only DriverClass as none`), the fixture restore moved into `beforeEach` where a mid-body failure cannot leak it, and `AgentRunner` is filed as #4112 — including the reachable null-key path there, which is the part that makes it worth its own issue rather than a comment.

## On the two "undersold" points

Both are now in the PR body, the source comment **and** the spec. That is the right amount of redundancy for `new BaseAgent()` specifically: it reads like a needless special case, it is the single line standing between this factory and "arbitrary highest-priority agent", and it is exactly the shape someone simplifies during an unrelated cleanup. Three places that each say why is cheap insurance against one quiet edit.

## Standing

The two-registry diagnosis holds end to end — `AIAgentType.DriverClass` resolves against `BaseAgentType`, `ClassFactory` matches keys exactly and per base-class name, and none of `LoopAgentType` / `FlowAgentType` / `RealtimeAgentType` is registered as a `BaseAgent` key. The old fallback could never resolve: dead code that looked alive, and the guards beneath it were unreachable for a reason that made them look satisfied.

Dropping the type fallback still loses nothing — agent-type behaviour resolves separately inside `BaseAgent` via `BaseAgentType.GetAgentTypeInstance` — and `TryCreateInstance` is used correctly, with `Resolved ? Instance : null` discarding the base-class fallback that comes back on `Resolved: false`. The throw for a declared-but-unregistered driver is loud, actionable, and scoped to the one seat, which is the right blast radius for this failure.

Good round. The correction to the mechanism made the change more defensible than it was, not less.

PR #3990, opened 2026-08-21
release-lines: record line 5.51's candidate step on main

MS-BC:
**1 of 2.** Brings line 5.51's record to `main`, which still reads `"lines": {}` from 2026-07-28.

## Why this is needed

Line 5.51 was cut as a candidate on **2026-07-31** and certified on **2026-08-14**, but neither step was ever recorded in `release-lines.json` on `main`. The manifest went straight from no lines to `certified` in a single commit on `next` (`2f82c81693`), and that commit has never reached `main`.

**The certification itself completed and is live — nothing is wrong with it:**

| | |
|---|---|
| npm `latest` | **5.51.1** across all packages |
| git tag | `v5.51.0` exists |
| scorecard | `certifications/5.51.0.md`, signed — *"release created, comms sent, signed"* |

Only the record is missing.

## Why two PRs

`ci/validate-release-lines.mjs` requires a line to appear as `candidate` before it can be `certified` ([line 251](ci/validate-release-lines.mjs#L251)), and holds `certifiedBuild` / `candidateDate` / `certifiedDate` immutable once set. So the honest fix is to record the two steps in order rather than back-date a single entry.

That rule is doing its job here: operation 2's manifest write was skipped and only operation 4 wrote.

It surfaced now rather than sooner because **v6.1.0-edge.2 merged on 2026-08-12** — two days before the certify commit — so the v6.1.0-edge.3 release PR (#3989) is the first into `main` to carry it. Without this, every future release PR fails `validate-pr` identically.

## This PR

```json
"5.51": {
  "status": "candidate",
  "candidateDate": "2026-07-31",
  "supportEnds": "2027-02-14",
  "upgradeImpact": "none"
}
```

Every value is the real one, taken from the certified entry on `next`. `certifiedBuild` and `certifiedDate` are deliberately absent — the validator holds them immutable once set, so they belong to the flip in PR 2.

Verified locally: `node ci/validate-release-lines.mjs --base origin/main --mode pr` → **OK**.

### DIFF
--- a/release-lines.json
+++ b/release-lines.json
@@ -14,5 +14,12 @@
       }
     }
   },
-  "lines": {}
+  "lines": {
+    "5.51": {
+      "status": "candidate",
+      "candidateDate": "2026-07-31",
+      "supportEnds": "2027-02-14",
+      "upgradeImpact": "none"
+    }
+  }
 }


MS-BC:
Superseded — closing without merging, and the reason matters for the record.

The plan was to walk 5.51 through `candidate` → `certified` on `main` in two PRs. That cannot work: **`publish.yml` triggers on any push to `main` with no path filter**, and `main` currently holds 94 changesets with `pre.json` in `mode: pre, tag: edge`. Merging this PR would have run `changeset version` over those changesets and published a real `6.1.0-edge.3` to npm built from `main`'s **edge.2** content — burning the version on the wrong tree, unrecoverably.

Only one push to `main` is safe per release, and that has to be the release PR itself.

Fixed instead in `ba0e03f57b` on the v6.1.0-edge.3 release branch: `checkPrTransitions` now allows a line to arrive past `candidate` when it carries `candidateDate` — real evidence of candidacy, held immutable once set, so it cannot be backdated. A line still cannot appear `certified` out of nowhere.

The 5.51 record reaches `main` with the release, which is the single safe push. Nothing about the certification changes: npm `latest` stays 5.51.1, tag `v5.51.0` stands, the scorecard is unchanged.

Issue #3986, opened 2026-08-20
CodeGen should warn when a field has ValueListType='List' but no EntityFieldValue rows


MarceloT-BC:
## Summary

CodeGen can set `EntityField.ValueListType='List'` on a field that has **no `EntityFieldValue` rows**. That combination describes a field where *no value is legal*, which should never exist — it means the metadata is broken. Nothing catches it at generation time, so it surfaces (if at all) at runtime.

Raised by @AN-BC in review of PR #3972:

> This is also something that CodeGen _could_ check for - a column that has ValueListType == List and has no values. Since the vast majority of the situations that drive the setting of ValueListType come from CodeGen parsing check constraints this is, again, extraordinarily unlikely.

- Status: OPEN
- Layer: `@memberjunction/codegen-lib` — `ManageMetadataBase.manageEntityFieldValuesAndValidatorFunctions` / `applyValueListConfig`

## Why it is worth a build-time check even though it is unlikely

The runtime half already landed: `EntityFieldInfo.ValueIsPermittedByValueList()` (#3969, PR #3972) permits everything for such a field and emits a `LogError` naming the entity and field, latched once per `EntityFieldInfo`. That is the right runtime behaviour — refusing every value would fail every save on the field — but it is a *late* and *conditional* report: it needs the field to be validated in a running process before anyone learns the metadata is wrong.

CodeGen already has everything needed to say it at generation time, where it is a build-time warning against the definitive metadata rather than a log line someone has to notice in production.

## Where it can arise

1. **`ValueListType` set, then the values removed.** `syncEntityFieldValues` deletes `EntityFieldValue` rows no longer present in the parsed list, but the `ValueListType='List'` update is a separate statement (`manage-metadata.ts` ~5270) — nothing re-evaluates the type when the list empties.
2. **A CHECK constraint edited into a shape the parser no longer matches.** `parseCheckConstraintValues` returns null for anything that is not exactly a disjunction of quoted equalities, and the constraint then routes to the AI-validator path — leaving an earlier `ValueListType='List'` in place with a list that is now stale or emptied.
3. **A hand-authored or soft (`applyValueListConfig`) list whose `PossibleValues` is empty.** `SoftFieldValueListConfig.PossibleValues` is not checked for emptiness before the type is applied.
4. **Metadata edited directly**, which is how it would most plausibly happen in a real deployment.

## Suggested fix

After `manageEntityFieldValuesAndValidatorFunctions` completes, one query over the metadata it just wrote:

[...]

PR #4003, opened 2026-08-21
fix(codegen): capture a numeric or single-value IN(...) CHECK as a value list

### LINKED ISSUE #3978: Numeric and bit IN(...) CHECK constraints produce no value list, so the field loses both its dropdown and its validation
## Summary

A numeric or `bit` `IN (…)` CHECK constraint produces **no `EntityFieldValue` rows and no
`ValueListType='List'`**, because SQL Server renders those constraints with *unquoted* literals and
`parseCheckConstraintValues` only matches quoted ones. The consequence is not just missing
validation — the field also loses its **dropdown in Explorer**, since `PossibleValues` reads the
same `EntityFieldValue` rows. And when AI codegen is off, such a constraint yields *nothing at all*:
no value list, no generated validator.

- Status: OPEN
- Found: 2026-08-20, while implementing #3969 (MJ 6.1.0-edge.2)
- Suspected layer: `@memberjunction/codegen-lib` — `ManageMetadataBase.parseCheckConstraintValues`

## Repro — how to recreate

The shapes below are what SQL Server actually stores; I captured them by creating a temp table with
one `IN`-list CHECK per data type and reading `tempdb.sys.check_constraints`, then ran each
definition through CodeGen's own `parseCheckConstraintValues` regex:

```
CREATE TABLE #t (
  S nvarchar(20)     CHECK (S IN ('Active','Inactive')),
  C nchar(10)        CHECK (C IN ('Input','Output')),
  L int              CHECK (L IN (1,2,3)),
  M money            CHECK (M IN (0.50, 1.00)),
  B bit              CHECK (B IN (0,1)),
  D date             CHECK (D IN ('2026-01-01','2026-07-01')),
  T datetime2        CHECK (T IN ('2026-01-01T00:00:00')),
  G uniqueidentifier CHECK (G IN ('11111111-1111-1111-1111-111111111111'))
);
```

| Column type | `sys.check_constraints.definition` | Parser result |
| --- | --- | --- |
| `nvarchar` / `nchar` (≥2 values) | `([S]='Inactive' OR [S]='Active')` | ✅ value list |
| `date` (≥2 values) | `([D]='2026-07-01' OR [D]='2026-01-01')` | ✅ value list |
| `uniqueidentifier` (≥2 values) | quoted GUIDs | ✅ value list |
| `int` | `([L]=(3) OR [L]=(2) OR [L]=(1))` | ❌ not parsed |
| `money` / `decimal` | `([M]=(1.00) OR [M]=(0.50))` | ❌ not parsed |
| `bit` | `([B]=(1) OR [B]=(0))` | ❌ not parsed |
| any type, **single** value | `([T]='2026-01-01T00:00:00')` | ❌ not parsed |

The cause is in the regexes, which require a quoted literal on the right-hand side:

```ts
const structureRegex = new RegExp(`^\\(${quotedField}='[^']+'(?: OR ${quotedField}='[^']+?')+(?: OR ${quotedField} IS NULL)?\\)$`);
const valueRegex     = new RegExp(`${quotedField}='([^']+)\'`, 'g');
```

Two separate limitations fall out of that: unquoted (numeric/bit) values never match, and the `+`
on the OR group requires **at least two** values, so a single-value `IN` list is never captured for
any type.

## Expected vs actual

**Expected** — `CHECK (Level IN (1,2,3))` populates `EntityFieldValue` with `1`, `2`, `3` and sets
`ValueListType='List'`, exactly as the string case does, so the field renders as a dropdown and is
validated before the round trip.

**Actual** — no rows, no `ValueListType`. With `ParseCheckConstraints` enabled the constraint falls
through to an AI-generated `Validate()` method (so it *is* enforced, by a different mechanism, and
still with no dropdown); with AI codegen disabled it is dropped silently.

## Suggested fix

Extend `parseCheckConstraintValues` to accept SQL Server's unquoted numeric form —
`[Field]=(3)` alongside `[Field]='Value'` — and to accept a single-value list. Both are contained
changes to the two regexes plus the value extraction. Worth deciding at the same time:

1. Whether a `bit` list is worth capturing at all. `CHECK (B IN (0,1))` is vacuous (it permits
   exactly what `bit` already permits), so capturing it would add a two-item dropdown of `0`/`1`
   where a checkbox belongs. `CHECK (B = 1)` is the meaningful bit constraint, and that is a
   validator, not a list.
2. That the runtime side is ready for it: `@memberjunction/core`'s value-list validation
   (#3969 / PR #3972) already compares numbers by value rather than by JS type, precisely so a
   numeric list works the day CodeGen starts producing one.

## Related

- #3969 / PR #3972 — the runtime validation this metadata feeds.
- A `List` field with **no** `EntityFieldValue` rows is a related CodeGen-side check @AN-BC raised in
  review of that PR; the runtime now logs it, but CodeGen could catch it at generation time.

MarceloT-BC:
Closes #3978.

A numeric or `bit` `IN (…)` CHECK produced **no `EntityFieldValue` rows and no `ValueListType='List'`**, so the field lost both its validation *and* its dropdown in Explorer (`PossibleValues` reads the same rows). With AI codegen off it yielded nothing at all — no list, no generated validator.

## Why it happened

SQL Server renders a numeric list with **unquoted** literals, and `parseCheckConstraintValues` only matched the quoted form. Every string below is a real `sys.check_constraints.definition`, captured from SQL Server 2022 (one `IN`-list CHECK per data type) rather than assumed from the DDL:

| DDL | what the catalog stores | before |
| --- | --- | --- |
| `Status IN ('Active','Inactive')` | `([Status]='Inactive' OR [Status]='Active')` | ✅ value list |
| `D IN ('2026-01-01','2026-07-01')` | `([D]='2026-07-01' OR [D]='2026-01-01')` | ✅ value list |
| `Level IN (1,2,3)` | `([Level]=(3) OR [Level]=(2) OR [Level]=(1))` | ❌ nothing |
| `Price IN (0.50, 1.00)` | `([Price]=(1.00) OR [Price]=(0.50))` | ❌ nothing |
| `Flag IN (0,1)` (bit) | `([Flag]=(1) OR [Flag]=(0))` | ❌ nothing |
| `Tier IN (7)` — **any** single value | `([Tier]=(7))` | ❌ nothing |

The second limitation is that last row: `+` on the OR group required **two** values, so a single-value list never parsed for any type.

Two renderings not in the issue, found by capturing every numeric type: `bigint`/`numeric` come back with a **trailing point** — `(12345678901234567890.)` — and `float` in **exponential** form — `(1.0000000000000000e+030)`.

## The change

**1. CodeGen (`manage-metadata.ts`)** — one literal pattern matching both quoting styles, `*` instead of `+` on the OR group, and the numeric literal normalized (parens stripped, trailing point removed). It now returns `null` rather than `[]` when nothing matched: an empty array was truthy, so it would set `ValueListType='List'` on a field with no values behind it — the self-contradiction #3986 reports.

**2. Eligibility is now an explicit, separate decision** (`isValueListEligibleField`), so the parser stays a pure dialect parser and the *policy* lives at the call site where the field metadata is. Two field shapes never get a CHECK-derived list:

* **`bit`** — the issue's open question #1. `IN (0,1)` is vacuous (bit already permits exactly that), and `= 1` is a *validator*, not a two-item dropdown over what renders as a checkbox.
* **a primary key** — `CHECK (ID=1)` is the single-row-table guard MJ's own sequence tables use. A structural invariant, not values a user picks from. This is also what keeps single-value capture from being a metadata change (see the measurement below).

**3. An all-numeric list sorts numerically**, so the dropdown reads `1, 2, 10` rather than the lexical `1, 10, 2`. Still fully deterministic, which is the only thing the sort was there for.

**4. `@memberjunction/core`: a numeric column's value list is compared by numeric VALUE.** This is not scope creep — without it, commit 1 introduces a regression. See below.

**5. MC3 in the integration suite** mirrors this parser to compare physical constraints against `EntityFieldValue` rows, so it moves in lockstep — otherwise it would silently stop covering exactly the new capability.

## Correcting the issue: the runtime was NOT ready for this

The issue (and #3969's design notes) state that core "already compares numbers by value rather than by JS type, precisely so a numeric list works the day CodeGen starts producing one." That is **not** what the code does. `NormalizeValueListValue` is `String(v).trim().toLowerCase()`, which handles an `int` list but not a differing scale:

```ts
// CHECK (Price IN (0.50, 1.00))  →  metadata values '0.50' and '1.00'
// runtime value: the number 1
String(1) === '1.00'   // false  →  Validate() REFUSES a value the database accepts
```

@rkihm-BC named this exact hole when approving #3972 — *"numeric lists compare stringified, so `String(1.0)` is `'1'` and a metadata value of `'1.0'` would fail closed — speculative, since no numeric value lists exist today, so a sentence in the comment beats code."* This PR is what makes it non-speculative, so it fixes it in the same breath rather than shipping a producer that breaks its consumer.

The numeric comparison follows the shape of the existing date path (memoized key set, built from the already-normalized values, `null` when the list is not entirely numeric). It is gated on the **column** being numeric, **not** on the list looking numeric: on a string column `CHECK (Code IN ('1','2'))` really is a string list, and comparing `'01'` numerically would permit a value the database refuses. This rung must never disagree with the CHECK in either direction. A non-finite value falls back to the string comparison, so a non-numeric list on a numeric column (which CodeGen cannot produce, but `applyValueListConfig` can) behaves exactly as before.

Three comments in `entityInfo.ts` that this makes false are corrected in the same commit: that every value list is on a string column, that CodeGen cannot produce a numeric list, and the "recorded rather than solved" note on stringification. `bit` is still never produced — now deliberately, rather than by accident of the regex.

## Blast radius, measured rather than argued

Both parsers (old and new) run over **all 558 column CHECK constraints** in a live 6.x database (MJ core + BizApps accounting + orders):

| | |
| --- | --- |
| parsed identically before and after | 438 |
| still not a value list (ranges, `LEN()`, mixed predicates) | 116 |
| **newly parsed as a value list** | **4** |
| **regressions / changed values** | **0** |

All four newly-matched constraints are `CHECK (ID=1)` on sequence-table primary keys (`JournalEntryBatchSequence`, `OrderSequence`, `PaymentSequence`, `SubscriptionSequence`) — which the primary-key exclusion drops. **Net effect on existing metadata: nothing changes.** The widening is capability, not churn.

The one semantically reaching change is single-value capture: `CHECK (Col='X')` on a non-PK column becomes a 1-item list rather than falling through to a generated validator. Zero fields are affected today; flagging it as a deliberate reading of "`IN (x)` and `= x` are the same constraint, and SQL Server stores them identically" rather than an accident.

## Testing

**Unit — 43 new tests, written from the expected behaviour and confirmed red first.**

* `check-constraint-value-lists.test.ts` (new): 35 tests. **14 fail** on `next` — 9 numeric shapes, 5 single-value ones. Every `definition` string is a real catalog capture, including `(-1)`, `(1.00)`, `(1.0000000000000000e+030)` and the trailing-point form. Also pins what must **not** parse: two-sided and one-sided ranges, `LEN()`, `len(x)=(3) AND x=upper(x)`, a list OR-ed with a range, and a column-to-column comparison.
* `baseEntity.valueListValidation.test.ts`: 8 added. **3 fail** without the core commit — the scale mismatch, the numeric string from a form control, and the exponential float. Also pins that a *string* column with a numeric-looking list is still compared as strings (`'01'` rejected).

**Suites:** `@memberjunction/codegen-lib` 1118 passed / 6 files skipped. `@memberjunction/core` 2148 passed. Affected-package build (157 tasks) green.

**End to end on a live 6.x instance** — a restored throwaway copy of a real 6.x database, a table carrying every constraint shape, and `mj codegen --skipfiles` run before and after the fix:

| field | before | after |
| --- | --- | --- |
| `Status` nvarchar `IN ('Active','Inactive')` | `List` → `Active, Inactive` | unchanged |
| `Level` int `IN (1,2,3)` | **`None` → (none)** | `List` → `1, 2, 3` |
| `Price` money `IN (0.50, 1.00)` | **`None` → (none)** | `List` → `0.50, 1.00` |
| `Tier` int `IN (7)` | **`None` → (none)** | `List` → `7` |
| `Priority` int `IS NULL OR IN (1,2)` | **`None` → (none)** | `List` → `1, 2` |
| `IsFeatured` bit `IN (0,1)` | `None` | `None` (excluded by design) |
| `WidgetSequence.ID` PK `= 1` | `None` | `None` (excluded by design) |
| `Qty` int `> 0` | `None` | `None` (correct — not a list) |

Zero `EntityFieldValue` rows were created anywhere outside the test schema, checked directly. The scratch database was dropped afterwards.

**Not verified:** the Explorer dropdown itself. The dropdown claim rests on `PossibleValues` reading these same `EntityFieldValue` rows, which this PR does not touch.

## Repro (how to recreate the original bug)

```sql
CREATE TABLE dbo.Widget (
  ID uniqueidentifier NOT NULL PRIMARY KEY DEFAULT newid(),
  Status nvarchar(20) NOT NULL CHECK (Status IN ('Active','Inactive')),
  Level  int          NOT NULL CHECK (Level  IN (1,2,3))
);
```

Run `mj codegen`. Before this PR: `Status` gets `ValueListType='List'` and two `EntityFieldValue` rows; `Level` gets neither. After: both do.

## Notes for review

* **No changeset yet** — `patch` is the correct level (TypeScript only, no migration or metadata; the CI gate reserves `minor` for database changes). Say the word and I'll add it.
* #3986 (`ValueListType='List'` with no rows) touches the same file in a different function. Adjacent, not conflicting — and the `null`-instead-of-`[]` change here removes one way to produce that state at all.


rkihm-BC (commented):
## Review

The diagnosis and the fix are right, and the two-commit framing — producer and consumer landing in one breath rather than shipping a producer that breaks its consumer — is the correct call. Two items are blocking; one is a missing changeset; two are "confirm you meant this" rather than defects.

I re-implemented both regex parsers standalone and ran them over the documented shapes plus a dozen adversarial ones, and traced the downstream CodeGen emit path, rather than reading and agreeing.

---

### What's good

**The parser change is minimal and dialect-accurate.** Running the new pattern independently: `([Priority] IS NULL OR ([Priority]=(2) OR [Priority]=(1)))` → `['2','1']`; `([Big]=(100000000000.))` → `['100000000000']`; and it still correctly rejects `([Mixed]=(2) OR [Mixed]=(1) OR [Mixed]>(100))`, `(len([Currency])=(3) AND [Currency]=upper([Currency]))`, `([A]='X' OR [B]='Y')` and `((1)=[Level])`. Quoted values containing `)` still round-trip. No false positives found.

**The generated-code blast radius is safe — I checked, because a string union emitted onto a numeric column would have been a compile break.** `entity_subclasses_codegen.ts` quotes by `NeedsQuotes`, and `EntityFieldInfo.NeedsQuotes` is `false` for `TSType.Number`. So a numeric list emits `1 | 2 | 3` and `z.union([z.literal(0.50), z.literal(1.00)])` — every captured form (`-1`, `0.50`, `1.0000000000000000e+030`, and the trailing-point form after normalization) is a valid JS numeric literal.

**The eligibility gate doesn't change control flow for excluded fields.** `bit` and PK fields yield `parsedValues = null` and fall into the same LLM-validator `else` branch they already reached, since the old regex never matched an unquoted literal.

**The core commit really is required, and it cannot introduce new rejections.** `numericValueIsPermittedByValueList` is strictly *more* permissive than the string path on a numeric column: any value that string-matched has its own normalized form in the set, hence its `Number()` in `numericKeys`, and lowercase/trim cannot change a finite numeric parse. Gating on the **column** rather than on the list *looking* numeric is right for the reason given, and the fall-back-to-string on a non-finite candidate keeps `applyValueListConfig`-produced lists behaving as before.

Correcting the issue's "the runtime is already ready" claim in the description, with the receipt, rather than quietly working around it — and capturing real `sys.check_constraints` strings instead of guessing renderings from DDL — is what made this reviewable at speed.

---

### Blocking

#### 1. CI is red, and it is this PR — `f.ID === r.EntityFieldID`

`manage-metadata.ts:5256` trips the repo's UUID-comparison gate:

```
FAIL packages/MJGlobal/src/__tests__/UUIDCompliance.test.ts
  Found 1 direct UUID comparison(s) that should use UUIDsEqual():
    CodeGenLib/src/Database/manage-metadata.ts:5256: ... => f.ID === r.EntityFieldID
```

I ran that test against a tree without this PR to be sure it wasn't pre-broken — green there, red here.

`UUIDsEqual` is **already imported in this exact file** (line 32) and used four other places, so it is a one-token fix:

```ts
const field = allEntityFields.find((f: { ID: string; Type?: string; IsPrimaryKey?: boolean; }) =>
   UUIDsEqual(f.ID, r.EntityFieldID));
```

Not purely cosmetic here: a lookup miss returns `undefined`, and `isValueListEligibleField(undefined)` fails **open** — so a case mismatch under PG would silently re-enable the exclusion you deliberately added.

Worth knowing: this failure aborted the turbo run **before** `codegen-lib` and `core` ran, so the 1118 / 2148 numbers are local-only. CI has not yet executed a single one of the 43 new tests.

#### 2. The zero assertion in `handles zero and negative values` is vacuous

`baseEntity.valueListValidation.test.ts:518`:

```ts
expect(new EntityField(makeFieldInfo(signed), 0).Validate().Success).toBe(true);
```

`EntityField`'s constructor guards with `if (Value)` (`baseEntity.ts:377`), so `0` is skipped, `DefaultValue` is null, and the field is left unset. Confirmed against the built package:

```
Value when constructed with 0 = null
```

`ValueIsPermittedByValueList(null)` returns `true` at the null gate, so this passes on `next`, passes with the fix, and would pass even if `0` were not in the list. Zero is precisely the case most worth pinning — it is the one `Number('')` collides with — so it deserves a real assertion:

```ts
const zero = new EntityField(makeFieldInfo(signed));
zero.Value = 0;
expect(zero.Validate().Success).toBe(true);
```

I verified the fix actually works rather than assuming the setter would take it: `after explicit set, Value = 0 number`.

The other 42 assertions all pass a truthy value and are genuine — this is the only one.

---

### Missing

#### 3. Changeset

You already flagged it, and `patch` is correct. Two published packages changed: **`@memberjunction/codegen-lib`** and **`@memberjunction/core`**. `@memberjunction/integration-test-suite` is `"private": true`, so leave it out.

---

### Two judgment calls — push back if you meant them

Neither of these is a defect, and both have zero measured impact on the databases you ran the 558-constraint sweep against. I am raising them because the *rationale in the PR* is narrower than the *rule in the code*, not because I think the code is wrong. If you considered these and chose deliberately, say so and I am satisfied — a one-line "yes, intentional, here's why" closes both.

#### 4. The blast-radius measurement covered the parser, not the sort

`sortCheckConstraintValues` decides `allNumeric` from the **value strings**, with no column-type input. So an existing **string** column with a numeric-looking list — `CHECK (Code IN ('1','2','10'))` on `nvarchar` — re-sequences from `1, 10, 2` to `1, 2, 10`, and `syncEntityFieldValues` Step 3 writes those new `Sequence` values back, reordering the generated union and the dropdown.

I grepped `entity_subclasses.ts` for all-numeric quoted unions — **none in MJ core**, so no churn there. But "net effect on existing metadata: nothing changes" rests on a parser-only diff, and BizApps entities are not in that grep's reach. Re-running just the sort over the same corpus would close it. (Gating `allNumeric` on `!field.NeedsQuotes` is the alternative, but is only worth doing if something turns up.)

#### 5. The primary-key exclusion is broader than its stated rationale

The justification is specifically `CHECK (ID=1)`, the single-row sequence-table guard. The implemented rule drops *any* CHECK-derived list on a PK — including a genuine multi-value enum on a natural-key PK, e.g. `CHECK (Code IN ('US','CA'))` on a `char(2)` PK. The old parser had no field lookup at all, so it *did* capture that shape. Nothing plausible in MJ core (uniqueidentifier PKs), but CodeGen also runs over customer schemas where natural-key PKs are ordinary.

If you want the rule to match its rationale exactly, the narrower form is to parse first and drop after, gating on `field.IsPrimaryKey && parsed.length === 1` — which reduces to pre-PR behavior in both directions. Your call.

The `bit` exclusion needs no such narrowing — a bit CHECK always renders unquoted, so it was never reachable before and nothing regresses. That one is clean.

---

### One thing I checked that is *not* yours to fix

Single-value capture makes `([EffectiveDate]='2026-01-01')` on a `date`/`datetime` column produce a one-item list — and `NeedsQuotes` is `true` for `TSType.Date`, so CodeGen emits `'2026-01-01'` (a **string** literal type) and `z.literal('2026-01-01')` onto a field whose runtime type is `Date`. That hazard is **pre-existing**: a multi-value date CHECK already reached it, and your own comment in `entityInfo.ts` acknowledges date lists are reachable. This PR widens the aperture slightly, since single-value date CHECKs are more plausible than two-value ones. Worth a follow-up issue, not a change here.

---

**Summary:** #1 and #2 are the only two I would hold on. #3 is a formality. #4 and #5 are decisions, not defects — confirm and move on.

MarceloT-BC:
Both blocking items were real. Thank you especially for the CI point — you're right that the turbo abort meant **none** of the 43 new tests had ever executed on CI, so my "1118 / 2148 green" described only my laptop. I've changed how I validate this branch accordingly: full workspace `pnpm test`, not the two packages I touched, which is precisely the gap that let the gate through.

## 1 — UUID comparison gate ✔

`UUIDsEqual(f.ID, r.EntityFieldID)`. I reproduced it before fixing it — red on this branch, green on a tree without it, same as you found — and confirmed green after.

Your "not purely cosmetic" reading is the part that mattered: the lookup fails **open**, so a PG/SQL-Server case mismatch would silently re-enable the exclusion rather than error. That reasoning is now a comment at the call site, so the next person doesn't "simplify" it back.

## 2 — the vacuous zero assertion ✔

Verified the mechanism rather than taking the fix on faith: `EntityField`'s constructor guards with `if (Value)` (`baseEntity.ts:376`), so `new EntityField(fi, 0)` left the field unset and `ValueIsPermittedByValueList(null)` passed it at the null gate — as you said, it would have passed with `0` absent from the list entirely.

Assigned through the setter now, and I added the assertion that proves it can fail at all — a `0` against a list that does **not** contain it:

```ts
const zero = new EntityField(makeFieldInfo(signed));
zero.Value = 0;
expect(zero.Validate().Success).toBe(true);

const zeroOffList = new EntityField(makeFieldInfo({ ...signed, EntityFieldValues: [ /* '-1' only */ ] }));
zeroOffList.Value = 0;
expect(zeroOffList.Validate().Success).toBe(false);
```

## 3 — changeset ✔

`patch` for `@memberjunction/codegen-lib` and `@memberjunction/core`; `@memberjunction/integration-test-suite` left out as private. `npm run check:changeset` passes ("Branch touches no migration and no metadata — every entry must be patch").

## 4 — the sort: measured, and nothing turned up

You were right that the 558-constraint sweep covered the parser only. I re-ran it over the **sort**, on the same live 6.x database (MJ core + BizApps accounting + orders), comparing the old lexical ordering against the new one for every list the OLD parser already captured — i.e. exactly the population that can be re-sequenced:

```
column CHECK constraints scanned:                                     539
lists the OLD parser already captured (the only re-sequenceable ones): 431
of those, RE-SEQUENCED by the new sort:                                 0
```

No `CHECK (Code IN ('1','2','10'))`-shaped string column exists in that estate to churn. Per your note I left the code as it is and recorded the measurement at `sortCheckConstraintValues`, together with the narrowing to reach for if one ever shows up — so the next reader gets the gap and the remedy rather than having to re-derive both.

(539 rather than the PR body's 558: same database, three weeks later.)

## 5 — the PK exclusion: you're right, and I've narrowed it

Not a deliberate choice I can defend — I went and checked the natural-key case you raised, and my rule really did regress it. `CHECK (Code IN ('US','CA'))` on a `char(2)` PK parsed to a value list **before** this PR, and my gate silently stopped it. That's a regression on customer schemas wearing a policy's clothes, so I took your narrower form:

```ts
if (field.IsPrimaryKey && parsedValues.length === 1) {
   return null;   // CHECK (ID=1) — the singleton guard it was actually written for
}
```

The gate now runs **after** the parse (`isValueListEligibleField` → `valueListForField`), which is what lets each exclusion be no broader than its own reason and makes it impossible for either to drop something the old parser captured. `bit` stays unconditional, for the reason you gave — every bit list renders unquoted, so none was reachable before and nothing regresses.

MC3's mirror moved with it, so the audit still compares exactly the set CodeGen writes rather than a superset.

## The date-literal hazard

Agreed on all three points: pre-existing, reachable before this PR via a multi-value date CHECK, and widened only slightly here. Written up with the repro; @MarceloT-BC will decide whether it goes up as an issue now or with the follow-up.

## One thing beyond your review — the branch was three weeks stale

Worth stating since it changes what CI is actually testing: this was cut on Aug 21 and `next` had moved **1001 commits**, 34 of them touching these three files. I've merged `origin/next` in — clean, no conflicts, and the diff against `next` is still only the 6 files — so the run you'd be reading is against something current rather than a three-week-old base.

## Validation

Full workspace `pnpm test` on the merged tree:

- `@memberjunction/codegen-lib` — **1471 passed**, 66 skipped (92 files), including the 37 in `check-constraint-value-lists.test.ts`
- `@memberjunction/core` — **2579 passed** (169 files), including the 8 numeric-list cases
- `@memberjunction/global` — **879 passed** (32 files), UUID compliance among them

I checked the shard log for my own test files by name rather than trusting a green summary — turbo cancels downstream tasks on a first failure, which is how the previous run looked greener than it was.

rkihm-BC (approved):
## Approving — both blocking items are fixed and proven, re-verified at `de8c3627ad`

Re-reviewed in a detached worktree with every claim re-run rather than read.

### The two blocking items from August

| Item | Status | Evidence |
|---|---|---|
| UUID comparison gate | Fixed | `manage-metadata.ts:5505-5506` uses `UUIDsEqual(f.ID, r.EntityFieldID)`. The `UUIDCompliance` source guard passes on CI and locally. |
| New tests never executed on CI | Fixed, proven from the CI log | Run 34633445961 (this head): shard 4/6 shows `check-constraint-value-lists.test.ts (37 tests)` and shard 3/6 shows `baseEntity.valueListValidation.test.ts (43 tests)`. Both actually ran this time. |

The non-blocking items also landed: the vacuous zero assertion now assigns through the setter with a negative case; the PK exclusion is narrowed to single-value lists only (`:5826`) and mirrored identically in `metadata-consistency.checks.ts:277-282`; the 431-list sort blast radius is recorded in the JSDoc; changeset is present at `patch` for `codegen-lib` and `core` and passes `check:changeset`.

### Verified independently

- Red-then-green: with `manage-metadata.ts` swapped back to `next`, the parser suite fails 23 of 37; with `entityInfo.ts` swapped back, the MJCore suite fails the 3 numeric cases. Restored: 37/37 and 43/43.
- Local suites: CodeGenLib 1471 passed / 66 skipped, MJCore 2579 passed.
- Parser table on catalog-form CHECKs: numeric `OR`-chains, `N'…'` prefixes, negatives, `((1),(2))`, `1e5` all captured; `>=`, `NOT IN` (rendered as `<>` chains), and `AND`-compound constraints correctly return `null`.
- Downstream: numeric lists generate literal unions (`1 | 2 | 10`, `z.literal(1)`), `NeedsQuotes` is false for `Number`, and the ORM setter coerces numeric strings, so `0.50` and `0.5` agree at both type and runtime level.
- Branch is 140 commits behind `next`; `merge-tree` is clean, and the merged tree builds and passes both packages' suites. The two `next` commits that touched `manage-metadata.ts` since your merge-base do not overlap your hunks.
- No MJ core column has a numeric equality CHECK list today, so no existing generated type changes.

### Two optional wording notes, no action needed to merge

- The title reads dialect-neutral but the fix is SQL Server only: `parsePgArrayConstraint` still extracts only quoted elements, so `ARRAY[1, 2, 3]` returns `null`. One sentence in the changeset saying the PostgreSQL path is unchanged would save the next reader a trip.
- For downstream schemas, a numeric list now narrows the generated property to a literal union, so `entity.Level = n` with `n: number` will fail `tsc` after regeneration. Same behavior string lists already have; worth a line in the changeset.

PR #4005, opened 2026-08-21
fix(codegen): teach the class-registration manifest scanner to see @RegisterClassEx

### LINKED ISSUE #3944: Class-registration manifest generator is blind to @RegisterClassEx — tree-shaking guard and its coverage audit both silently skip those classes
## Summary

`mj codegen manifest` generates `mj-class-registrations.ts`, whose stated job is to create a static import path so bundlers cannot tree-shake registered classes away. Its AST scanner matches the decorator identifier **literally**, so it only ever sees `RegisterClass` — never `RegisterClassEx`.

Consequences, both silent:

1. Classes registered with `@RegisterClassEx` never appear in any generated manifest, so nothing protects them from tree-shaking.
2. The **coverage audit** that exists to catch exactly this (`strict: true`, intended for CI) is driven by the same scanner, so it cannot report them as gaps either. The safety net has the same blind spot as the thing it guards.

`RegisterClassEx` is not an obscure alternative — `MJGlobal/src/RegisterClass.ts` explicitly recommends it: *"**Prefer this over `RegisterClass` for new code** when you have anything beyond `(baseClass, key, priority)` to specify."* The repo therefore steers developers toward the one form its own toolchain cannot see.

## Evidence

Both scan paths in `packages/CodeGenLib/src/Manifest/GenerateClassRegistrationsManifest.ts` hard-compare the identifier:

- TypeScript source scan — `:461`
  ```ts
  if (!ts.isIdentifier(callExpr.expression) || callExpr.expression.text !== 'RegisterClass') {
      return null;
  }
  ```
- Compiled-JS (`__decorate`) scan — `:569`
  ```ts
  if (element.expression.text !== 'RegisterClass') continue;
  ```

`grep -c RegisterClassEx` against that file returns **0** — the generator has no awareness of it anywhere.

Observed effect: `packages/Angular/Bootstrap/src/generated/mj-class-registrations.ts` (1376 lines, 22 packages) contains **zero** entries for any `@RegisterClassEx` class. Spot-checked absent: `ModelPredictionPanel`, `AISkillSharingPanel`, `EntityRelationshipFormRolePanel`.

Scope: 25+ `@RegisterClassEx` call sites, all currently under `packages/Angular/Explorer/core-entity-forms` — the `BaseFormPanel` contributions (`panels/**`, `custom/HierarchyPanels/hierarchy-form-panels.ts` alone has 14).

The same literal comparison also exists in `packages/CodeGenLib/src/EntityNameScanner/EntityNameScanner.ts:461` and is worth reviewing alongside, though its purpose (entity-name rename mapping) is separate.

## Why nothing is visibly broken today

Every affected class lives in `core-entity-forms`, and the manifest imports 20+ other classes from that package's barrel — which transitively pulls the `RegisterClassEx` ones into the static graph by accident. They are protected incidentally, not by design.

The latent failure mode is a **new package whose only registrations use `RegisterClassEx`**: nothing in the manifest imports it, nothing in the static graph references it, the classes never register, and the feature is simply inert at runtime — no build error, no runtime error, no CI gap report. The same applies if bundler tree-shaking is ever tightened, or if `core-entity-forms`' barrel imports change.

## How to fix

Teach both scan paths to accept either decorator and to parse the two argument shapes. Recommended, in `GenerateClassRegistrationsManifest.ts`:

1. Replace both identifier comparisons with a set membership test, e.g. `const REGISTER_DECORATORS = new Set(['RegisterClass', 'RegisterClassEx'])`.
2. Branch on the second argument when extracting `key`:
   - `RegisterClass(baseClass, key, priority?, skipNullKeyWarning?, autoRegisterWithRootClass?, metadata?)` — `key` is `args[1]` as a string literal, `metadata` is `args[5]`.
   - `RegisterClassEx(baseClass, { key?, priority?, metadata?, ... })` — `args[1]` is an object literal; read the `key` property from it.

   Signatures are in `packages/MJGlobal/src/RegisterClass.ts:51-62` (positional) and immediately below it (options bag). `args[0]` (the base class identifier) is identical in both forms, so that extraction is unchanged.
3. Add unit coverage in `packages/CodeGenLib/src/__tests__/manifest.test.ts` for both decorator forms, in both TS-source and compiled-`__decorate` shapes.
4. Regenerate the manifests and confirm the `RegisterClassEx` classes now appear.

### Alternative worth a maintainer's opinion

Rather than teaching the scanner, `RegisterClassEx` could be deprecated in favor of the positional form so there is one shape to parse. That trades a small toolchain fix for churn across 25+ call sites and abandons an ergonomics improvement, so teaching the scanner looks clearly better — but the API direction is the author's call, and the docstring recommending `RegisterClassEx` should be reconciled either way.

### Also worth considering

Once the scanner sees both forms, enabling the coverage audit's `strict` mode in CI would turn this class of problem from silent into a build failure.

## Verification

After the fix, `packages/Angular/Bootstrap/src/generated/mj-class-registrations.ts` should contain `ModelPredictionPanel` and its siblings, and a scratch package registering a single class via `@RegisterClassEx` should appear in the regenerated manifest for an app that depends on it.

## Context

Found while planning generic record attachments (`plans/record-attachments-ux.md`), which proposes a new standalone package whose only registration is a `BaseFormPanel` contribution — precisely the case that fails silently today.

MarceloT-BC:
# Forward:
I do not know this domain. The code looks ok to me, but I don’t knwo the pitfalls of this area.

Fixes #3944.

`mj codegen manifest` matched the decorator identifier literally, so it only ever saw `RegisterClass` — never `RegisterClassEx`, the options-bag form `MJGlobal/src/RegisterClass.ts` explicitly recommends for new code. Two silent consequences: those classes never entered any manifest, so nothing protected them from tree-shaking; and the coverage audit meant to catch exactly that gap runs on the same scan, so it could not report them either.

## What changed

- Both scan paths (TypeScript source, compiled `__decorate`) test set membership over `{RegisterClass, RegisterClassEx}` and share one key extractor handling either argument shape — a positional string literal, or the options bag's `key`. `args[0]` is identical in both forms, so base-class extraction is unchanged; an absent or non-literal key yields `undefined`, as the positional form already did.
- `EntityNameScanner.classifyParentContext` had the same literal comparison. It now accepts both decorators, plus a case for the options bag's `key` — whose parent is a property assignment, not the decorator call, so the old check could not see it. Scoped to a register decorator's own options object, so a bare `key:` elsewhere is not mistaken for an entity name. `buildEntityNameMap` is deliberately untouched: it parses generated `entity_subclasses.ts`, whose shape CodeGen fixes.

## How to recreate the original issue

No DB or instance needed — the command is pure source scanning.

1. Create a package whose only registration is `@RegisterClassEx(BaseFormPanel, { key: 'x' })`, exported from its `.d.ts`, plus an app depending on it.
2. Run `mj codegen manifest --appDir <app>`. **Before:** `Found 0 @RegisterClass decorators in 0 packages` and an empty manifest. **After:** 1 class, imported by name.
3. A fixture with 5 registrations across both decorator forms and both scan paths went from **2 found to 5**, with keys read correctly from both shapes.

## Effect on this repo's manifests

Regenerated on a fully built tree and included here: Bootstrap **620 → 645**, BootstrapLite **506 → 531** — 25 `@RegisterClassEx` `BaseFormPanel` contributions from `@memberjunction/ng-core-entity-forms`, nothing removed.

## What was actually at risk (corrected — thanks @rkihm-BC)

An earlier version of this description framed the six panels the manifest still skips as the worrying gap. That was backwards, and the correction is worth recording because it changes how the follow-up should be prioritized:

- The **six skipped** panels (`ModelPredictionPanel`, `AISkillSharingPanel`, `EntityRelationshipFormRolePanel`, `WebsiteCrawlerSettingsPanel`, `TagPipelineConfigurationPanel`, `AgentRealtimePanel`) are absent because `core-entity-forms/src/public-api.ts` does not export them by name — but all six are imported and declared by `lib/custom/custom-forms.module.ts`, which *is* `export *`-ed from the public API. That is a real static reference, so they were never tree-shakeable. Exporting them is cosmetic, not a bundling risk.
- Of the **25 added**, 15 were likewise already anchored — they are elements of `HIERARCHY_FORM_PANELS`, which `custom-forms.module.ts` imports. Their manifest entries are belt-and-braces.
- The **10 that were genuinely exposed** are the header/overview panels: `AIAgentCategoryHeaderPanel`, `AIAgentCategoryOverviewPanel`, `CompanyHeaderPanel`, `CompanyOverviewPanel`, `ConversationHeaderPanel`, `ConversationOverviewPanel`, `EmployeeHeaderPanel`, `EmployeeOverviewPanel`, `UserHeaderPanel`, `UserOverviewPanel`. Each appears only in its own declaration and a bare `export { X } from` re-export line — exactly the shape tree-shaking drops. Those are what this PR rescues.

The latent case in #3944 stands regardless: a new package whose only registration is `@RegisterClassEx`, with no module referencing it, was invisible to both the manifest and the audit meant to catch that.

## Tests

9 new cases drive the real generator and the real scanner (not a re-implementation of their AST logic), covering both decorator forms in both TS-source and compiled shapes, keyless and non-literal keys, and two negative controls (a decorator merely named `RegisterClassSomethingElse`; a bare `key:` property). **7 of the 9 fail against the unfixed scanner**; the 2 negatives pass either way by design.

Validation run on this branch, rebased on `a8e326e811`: install + build green (228 tasks); `@memberjunction/codegen-lib` suite **1092 passed, 0 failed** (6 files skipped, pre-existing); `@memberjunction/core` 2082 passed, `@memberjunction/cli` 36 files, `@memberjunction/ng-base-forms` 24 files all green on the pre-rebase base.

On the issue's other suggestion — enabling the coverage audit's `strict` mode in CI — @rkihm-BC measured it rather than assuming, and it is **not** a free flag flip. The audit only runs when `--lazy-config` is passed, i.e. only `mj:manifest:explorer`, where `--exclude-packages @memberjunction` prunes every MJ package from the walk (that manifest's `CLASS_REGISTRATIONS_COUNT` is 0). "Covered" therefore means "in one of the four lazy chunks", so `--strict` today would fail on hundreds of classes rather than on anything anyone cares about. Making it useful means teaching the audit about sibling manifests — its own change. Not done here, and not recommended as a quick follow-up.


AN-BC:
Good catch! Will let @rkihm-BC  review and approve this along with your other open PR I think #4003 while I'm away.

rkihm-BC (approved):
Approving. Correct, well-scoped, and better verified than the PR description claims. No blocking issues. Two items deserve attention below — both are things the PR explicitly asks about, and one I'd answer differently than it's framed.

You wrote that you don't know this domain, so I spent most of the review on the domain-specific hazards rather than the AST logic.

---

### The pitfall in this area — checked, clean

`packages/Angular/Bootstrap/CLAUDE.md` documents the failure mode for this exact file: the manifest walks the **full transitive dependency tree**, and `syncDependencies` writes every discovered package back into `package.json`. One edge into a server-only package (`aiengine`, `templates`, `storage`, any `ai-<provider>`) floods the browser bundle and MJExplorer won't even bind. That's the June 2026 incident that file was written after.

I ran the leakage grep that file prescribes against both regenerated manifests — clean. Stronger: not a single `from '@memberjunction/...'` line changed in either manifest. The entire diff is class names added inside the *existing* `ng-core-entity-forms` import block, and neither Bootstrap `package.json` was touched. That combination is exactly the signature of a safe regen.

### The manifests are CI-verified — but by a different gate than you'd expect

The **freshness gate did not run**. It's `if: env.TURBO_FILTER == ''` (`.github/workflows/test.yml:276`), and this run had `TURBO_FILTER: --filter=...[origin/next]`.

The per-package `prebuild` scripts regenerate manifests during the build, though, and CI's log shows every one reproducing the committed content exactly:

```
@memberjunction/ng-bootstrap:build:       [class-manifest] No changes detected (645 classes, 22 packages)
@memberjunction/ng-bootstrap-lite:build:  [class-manifest] No changes detected (531 classes, 18 packages)
@memberjunction/server-bootstrap:build:   [class-manifest] No changes detected (1004 classes, 109 packages)
@memberjunction/a2aserver:build:          [class-manifest] No changes detected (0 classes, 0 packages)
@memberjunction/ai-mcp-server:build:      [class-manifest] No changes detected (0 classes, 0 packages)
@memberjunction/ng-explorer-core:build:   [class-manifest] Lazy config unchanged
```

645 and 531 match the PR. Server-side manifests unaffected, and `lazy-feature-config.ts` — the other tracked, scan-derived file — unchanged. That answers the question I'd otherwise have raised (*which other committed manifests should have been regenerated?*): **none**. Consistent with `@RegisterClassEx` appearing only in `core-entity-forms` (31), `base-forms` (2), and MJGlobal's own docs.

### Other things I verified rather than took on faith

* **Signature claim holds.** `RegisterClassEx(baseClass, options)` — `args[0]` really is identical to the positional form, so "base-class extraction unchanged" is accurate.
* **All 25 added names are exported** from `core-entity-forms/public-api.ts`; the 6 skipped ones are not. Checked individually. No duplicate identifiers in either manifest — which would have broken the ng-bootstrap build, and that built green on a cache miss.
* **Widening the match doesn't sweep in fixtures.** `base-forms` carries a `@RegisterClassEx` on `FakeSlotPanel` inside a `.dom.test.ts`. `findSourceFiles` excludes `**/*.test.ts`; and even on the `--scan-dist` path (which does *not* exclude tests) `filterToExportedClasses` drops it, since it isn't in the public API. Worth checking — "match more decorators" is exactly the change that could pull test classes into a shipped manifest.
* **The compiled-JS fixture matches reality.** Built `core-entity-forms/dist/lib/panels/ai-skill-sharing/ai-skill-sharing-panel.component.js` emits a bare `RegisterClassEx(BaseFormPanel, {…})` inside `__decorate` — ESM output, so the `ns_1.RegisterClassEx` property-access shape the scanner cannot match never arises.
* **`extractRegistrationKey` is behavior-preserving** for the positional form: `args.length < 2`, `key: null`, and non-literal keys all still yield `undefined`. Shorthand `{ key }` and `{ ...spread, key: 'x' }` both behave sensibly.
* **Case 3b is correctly scoped.** `classifyParentContext` is only consulted after `renameMap.has(text)`, and the options-object check requires a decorator parent — so a form-panel slug like `'content-sources:tag-pipeline'` can never be mistaken for an entity name. It is also genuinely useful: for entity subclasses, `@RegisterClassEx(BaseEntity, { key: 'Actions' })` really does carry the entity name.
* **"7 of 9 red" checks out.** I walked all nine; the two negatives (`RegisterClassSomethingElse`, bare `key:`) pass either way by construction, the other seven depend on the fix.

---

### One framing in the description is backwards — and it changes the follow-up's priority

The PR presents the 6 unexported panels as a worrying known gap, "protected only incidentally by the package barrel." It's the other way round.

All six are referenced by `src/lib/custom/custom-forms.module.ts`, which **is** exported from `public-api.ts` — a real static reference, so they are already retained. The 25 this PR fixed had no such reference: `UserOverviewPanel`, for instance, appears only in its own file and in a bare `export { … } from` re-export line, which is precisely what tree-shaking drops.

So this PR fixed the genuinely exposed set and flagged the safe set as the gap. No code change implied — but exporting those six is cosmetic, not urgent, and shouldn't be prioritized as a tree-shaking risk.

### The maintainer call you asked for: `--strict` is not a free switch

The coverage audit only runs when `--lazy-config` is passed, which is only `mj:manifest:explorer`. On that invocation `--exclude-packages @memberjunction` prunes every MJ package from the dependency walk (`isPackageExcluded` → `continue`, `GenerateClassRegistrationsManifest.ts:303`), so the eager manifest is empty. CI's own log confirms it — `mj_explorer:build: [class-manifest] Updated: 0 classes from 0 packages (1170 deps walked)` — as does the generated file: `CLASS_REGISTRATIONS_COUNT = 0`.

"Covered" therefore reduces to "in one of the four lazy chunks" (`ng-dashboards`, `ng-explorer-settings`, `ng-file-storage`, `ng-react`), while `isExcludedPackage` is true for *every* MJ class. The audit has no notion that ng-bootstrap's manifest covers them. As written, `--strict` in CI would fail on hundreds of classes rather than on the six you care about.

Run `mj:manifest:explorer --verbose` and read the actual "Coverage gap" count before wiring it up. Making it useful probably means teaching the audit about sibling manifests — its own change, not a flag flip.

---

### Minor / optional

* **Changeset scope.** It names only `@memberjunction/codegen-lib`, but `ng-bootstrap` and `ng-bootstrap-lite` ship changed source. `"fixed": [["@memberjunction/*"]]` bumps everything in lockstep, so this is cosmetic — the changeset-bot's own suggested changeset lists all three if you want it exact.
* **`metadata.entity` gap in EntityNameScanner.** The options-bag support reads `key` but not `metadata.entity`, which per MJGlobal's documented example is where a form panel's entity name lives — and lowercase `entity` isn't in `ENTITY_NAME_ASSIGNMENT_PROPS` either, so Case 2 misses it too. Only matters to someone running the v5.0 rename tool over v6-era code, so low value; noting it so the omission is a choice rather than an oversight.
* `REGISTER_DECORATORS` is now defined twice in the same package. Fine as-is; it just needs to stay in sync if a third form ever lands.

---

The fix is correct, the regenerated manifests are complete and independently reproduced by CI, and the change cannot leak server packages into the browser bundle. Fix the known-gap framing in the description if you want the follow-up prioritized correctly, and don't flip `--strict` without measuring first.

PR #4006, opened 2026-08-21
fix(migrations): FileEntityRecordLink unique key omits RecordID (#3943)

### LINKED ISSUE #3943: FileEntityRecordLink: UNIQUE (EntityID, FileID) omits RecordID, so one file can only attach to one record per entity
## Summary

`__mj.FileEntityRecordLink` is the generic soft-key many-to-many between any file and any record of any entity. Its unique constraint omits `RecordID`:

```sql
CONSTRAINT [UQ_FileEntityRecordLink_EntityID_FileID] UNIQUE NONCLUSTERED ([EntityID], [FileID])
```

The effect is that **a given `File` can be linked to at most one record per entity**. Attaching the same document to two Contracts, two Accounts, or two of anything else fails on a unique-key violation. That contradicts the table's entire purpose.

## Evidence

- Table definition and constraint: `migrations/v5/B202607091514__v5.46.x__Baseline.sql:3875` (latest baseline).
- Introduced by: `migrations/v5/V202605221002__v5.37.x__Add_Unique_Constraints_To_MJ_Junction_Tables.sql:441`.
- Never dropped or amended — `grep -rn "DROP CONSTRAINT.*FileEntityRecordLink" migrations/` returns nothing.
- Columns are `ID`, `FileID` (FK), `EntityID` (FK), `RecordID NVARCHAR(750)` (**not** an FK), plus timestamps.

## Why this looks unintended rather than deliberate

The introducing migration states its own scope, and `FileEntityRecordLink` falls outside it on two counts:

> The 17 tables below are all "pure junction" tables in `__mj` — they consist of **two foreign-key columns** plus ID/Sequence/timestamps, **with no other meaningful data columns**, and they're populated **either exclusively by CodeGen / metadata sync (no runtime UI/API writers)** or by first-creation-only paths that should never produce duplicates if working correctly.

1. It is not two FK columns. `RecordID` is an `nvarchar(750)` soft key, so the "natural key is typically a pair of foreign-key columns" heuristic mechanically selected `EntityID` + `FileID` and dropped the one column that makes a row unique.
2. It does have a runtime writer — `packages/AI/Agents/src/realtime/realtime-recording-store.ts:232-243` creates link rows during agent sessions.

The generated ORM describes the table as the motivating example for soft-key link detection (`packages/MJCoreEntities/src/generated/entity_subclasses.ts:18078`), which confirms `(EntityID, RecordID)` is meant to be the record address.

## Why it has not surfaced

`storeRealtimeRecording` is currently the only writer in the repo, and it uploads a fresh `MJ: Files` row per session recording, so it never presents the same `FileID` twice. There are no readers. The constraint has therefore never been exercised.

## Possible data loss already incurred

The same migration auto-deletes pre-existing duplicates before adding each constraint, retaining the earliest `__mj_CreatedAt` per `(FK1, FK2)` group. Any deployment that legitimately had one file linked to several records of the same entity had those link rows **deleted**. The migration logs per-table duplicate and deletion counts, so affected deployments can check their upgrade logs; the rows themselves are not recoverable from the migration.

## How to fix

New versioned T-SQL migration in `migrations/v6/`, named `V<timestamp>__v6.1.x__Fix_FileEntityRecordLink_Unique_Key.sql`:

```sql
ALTER TABLE ${flyway:defaultSchema}.FileEntityRecordLink
    DROP CONSTRAINT UQ_FileEntityRecordLink_EntityID_FileID;
GO

ALTER TABLE ${flyway:defaultSchema}.FileEntityRecordLink
    ADD CONSTRAINT UQ_FileEntityRecordLink_EntityID_RecordID_FileID
    UNIQUE NONCLUSTERED (EntityID, RecordID, FileID);
GO
```

This is a **widening** — every row that satisfied the old constraint satisfies the new one — so no dedupe pass or data migration is required, and it cannot fail on existing data.

### Worth doing in the same migration

The only indexes on the table are the two auto-generated single-column FK indexes (`IDX_AUTO_MJ_FKEY_FileEntityRecordLink_EntityID`, `..._FileID`). The lookup every consumer will issue is `WHERE EntityID = ? AND RecordID = ?`, which has no covering index:

```sql
CREATE NONCLUSTERED INDEX IX_FileEntityRecordLink_EntityID_RecordID
    ON ${flyway:defaultSchema}.FileEntityRecordLink (EntityID, RecordID) INCLUDE (FileID);
GO
```

The new unique constraint's leading `(EntityID, RecordID)` prefix partly covers this, so treat the explicit index as optional — measure before adding it.

### Scope notes

- **T-SQL only.** Do not hand-author a `migrations-pg/**` counterpart; PG conversion is run by the build engineer at release time (`migrations/CLAUDE.md`). The PG baseline carries the same defect and will be corrected by that conversion.
- No `EntityField` inserts are involved, so the `Sequence` rule does not apply.
- Run `mj sync push` then `mj codegen` after the migration so metadata and the generated ORM pick up the constraint change.

## Verification

Attach one `MJ: Files` row to two different records of the same entity and confirm both link rows save:

```ts
for (const recordID of [recordA, recordB]) {
    const link = await md.GetEntityObject<MJFileEntityRecordLinkEntity>('MJ: File Entity Record Links', user);
    link.NewRecord();
    link.FileID = sharedFileID;
    link.EntityID = entityID;
    link.RecordID = recordID;
    expect(await link.Save()).toBe(true);   // second Save() fails today
}
```

## Context

Found while planning generic record attachments (`plans/record-attachments-ux.md`). This is a hard blocker for that feature and for `bizapps-contracts`, which drops its `Contract.ExecutedDocumentFileID` FK in favor of this table.

MarceloT-BC:
Fixes #3943.

> **Scope note — this PR also contains a one-line CI change.** The title says "fix a unique key";
> the diff also narrows a migration guard's file scope, which is a separate review. Added at review
> request. `Check migration content` in `.github/workflows/changes.yml` selects every changed `.sql`
> under `migrations/` and scans **whole files rather than added lines**, so this branch's edit to the
> hand-run fixture `migrations/v5/tests/junction_unique_constraints/02_assert_post_migration.sql`
> pulled that file's entire history into the gate. Its 10 hard-coded `__mj` references are
> **pre-existing and identical in count on `next`**; this branch adds none. They are also *correct*
> for that file: it is not a Flyway migration (the filename doesn't match `V<version>__`), so
> `${flyway:defaultSchema}` would never be substituted and would break the script. The fix adds
> `grep -vE '(^|/)tests?/'`, mirroring the exclusion `.github/scripts/check-migration-id-determinism.sh`
> already makes for the same directory. The `__mj` gate was the outlier; the job's other four steps
> already skip fixtures structurally, so no existing coverage is removed. The underlying
> whole-file-versus-added-lines behaviour is broader than this PR and may deserve its own issue.

## The defect

`__mj.FileEntityRecordLink` is the generic soft-key many-to-many between any file and any record of any entity. Its unique key omitted `RecordID`:

```sql
CONSTRAINT [UQ_FileEntityRecordLink_EntityID_FileID] UNIQUE NONCLUSTERED ([EntityID], [FileID])
```

so a given `File` could be linked to **at most one record per entity**. Attaching the same document to two Contracts, two Accounts, or two of anything else failed on a unique-key violation — the exact operation the table exists to support.

## Why it is a defect rather than a deliberate rule

`V202605221002__v5.37.x__Add_Unique_Constraints_To_MJ_Junction_Tables` states its own scope: pure junction tables of **two foreign-key columns** plus ID/Sequence/timestamps, **no other meaningful data columns**, written only by CodeGen / metadata sync. `FileEntityRecordLink` fails both halves — `RecordID` is an `nvarchar(750)` soft key, not an FK, so the "natural key is a pair of foreign keys" heuristic mechanically picked `(EntityID, FileID)` and dropped the column that makes a row distinct; and it does have a runtime writer (`packages/AI/Agents/src/realtime/realtime-recording-store.ts`).

MJ's own metadata names this table as the motivating example for soft-key detection — see the `EntityField.EntityIDFieldName` description, which cites `FileEntityRecordLink`'s `EntityID`/`RecordID` pair by name — confirming `(EntityID, RecordID)` is meant to be the record address.

**Prior art:** this is the second constraint from that same batch corrected on these grounds. `V202608080100__v6.1.x__Drop_EntityAction_Uniqueness.sql` dropped `UQ_EntityAction_ActionID_EntityID` with reason 1 stated as *"THE CONSTRAINT WAS APPLIED OUTSIDE ITS OWN DECLARED SCOPE."*

It has gone unnoticed because `storeRealtimeRecording` uploads a fresh `MJ: Files` row per session, so it never presents the same `FileID` twice, and there are no readers yet.

## The change

One migration, `V202608211100__v6.1.x__Fix_FileEntityRecordLink_Unique_Key.sql`: drop `UQ_FileEntityRecordLink_EntityID_FileID`, add `UQ_FileEntityRecordLink_EntityID_RecordID_FileID`. Both statements guarded on `sys.key_constraints`, so the script is re-runnable and a database somehow lacking the old key still ends up with the new one.

This is a **widening** — every row satisfying the old key satisfies the new one — so no de-duplication pass or data migration is needed and it cannot fail on existing data. The genuine duplicate (same file linked twice to the same record) is still rejected.

**No new index.** The lookup consumers will issue is `WHERE EntityID = ? AND RecordID = ?`, and the new constraint's backing index leads with exactly that prefix and carries `FileID` as its third key column, so it already covers that query. A separate `IX_..._EntityID_RecordID INCLUDE (FileID)` would duplicate it.

Also updated: `migrations/v5/tests/junction_unique_constraints/02_assert_post_migration.sql` hardcoded the old constraint name in its "all 17 constraints exist" assertion and would otherwise fail on any database that has taken this migration. It now accepts either name, so it stays valid both immediately after the v5.37 migration and on a current database.

No CodeGen: this changes a constraint, not a column — nothing in the generated ORM or entity metadata changes. T-SQL only; the `migrations-pg/**` counterpart is build-engineer work at release time per `migrations/CLAUDE.md`.

## How to recreate the original issue

Link one `MJ: Files` row to two different records of the same entity:

```ts
for (const recordID of [recordA, recordB]) {
    const link = await md.GetEntityObject<MJFileEntityRecordLinkEntity>('MJ: File Entity Record Links', user);
    link.NewRecord();
    link.FileID = sharedFileID;
    link.EntityID = entityID;
    link.RecordID = recordID;
    expect(await link.Save()).toBe(true);   // second Save() fails before this change
}
```

## Verification performed

Against a real SQL Server 2022 instance.

**Isolated table** (baseline definition, old constraint in place):

| Step | Result |
|---|---|
| Second link row, same file → different record, before the migration | **rejected**, error 2627 on `UQ_FileEntityRecordLink_EntityID_FileID` — defect reproduced |
| Same insert after the migration | accepted |
| Exact duplicate `(EntityID, RecordID, FileID)` | still rejected, 2627 |
| Different file on an already-linked record | accepted |
| Resulting key columns | `EntityID, RecordID, FileID` |
| Re-running the migration | clean no-op, exit 0 |

**Real database, from zero:** `mj migrate --dir ./migrations` on an empty database — baseline + 56 versioned migrations + `R__RefreshMetadata`, **exit 0**, this migration applied in sequence in 17 ms. Then, through MJ's own `spCreateFileEntityRecordLink` (the procedure `BaseEntity.Save()` calls) against real `MJ: Users` records and a real `__mj.File` row:

- link #1 (file → record A) — created
- link #2 (**same** file → record B, same entity) — created; this is the case that fails today
- exact duplicate link — still rejected, 2627

Post-migration indexes on the table are exactly `PK_FileEntityRecordLink_ID`, the new unique constraint, and the two pre-existing auto FK indexes — no extras.

`.github/scripts/validate-migration-filenames.sh` passes on this branch: 947 files, no duplicate versions, ordering consistent.

**Not run:** the TypeScript snippet above at the `BaseEntity` layer — that needs a full MJ build and provider bootstrap. The stored-procedure test is one layer below it and exercises the same write path.


rkihm-BC:
Reviewed the migration, the test change, the changeset, and verified the rule and schema details behind each claim.

## What's good

**The defect is real and the case that it's unintended rather than deliberate is unusually well-built.** The v5.37 batch declares its own scope in its header — pure two-FK-column junction tables with no other meaningful data columns and no runtime writer — and `FileEntityRecordLink` fails both halves. Citing MJ's own `EntityField.EntityIDFieldName` description, which names this table's `EntityID`/`RecordID` pair as the soft-key example, is the strongest piece: it's the platform's own metadata saying that pair is the record address. And the `Drop_EntityAction_Uniqueness` precedent establishes this is the second correction from the same batch on the same grounds, which turns a one-off into a recognised pattern.

**The widening argument is correct, and I checked the one thing that could have undermined it.** `RecordID` is `NVARCHAR(750) NOT NULL` in the baseline. That matters because SQL Server treats NULLs as *equal* for UNIQUE constraints — had `RecordID` been nullable, rows with a NULL `RecordID` would still have been restricted to one per `(EntityID, FileID)`, and the "every row that satisfied the old key satisfies the new one" claim would have needed a caveat. It doesn't. The widening holds unconditionally, so no dedup pass and no possibility of failing on existing data.

**Both statements are guarded on `sys.key_constraints`**, so the script is genuinely re-runnable and a database missing the old constraint still converges on the new one. The `IF/BEGIN/END/ELSE PRINT/GO` structure is valid T-SQL and each branch prints what it did.

**Declining to add an index, with the reason stated, is the right call.** The new constraint's backing index leads with `(EntityID, RecordID)` and carries `FileID` third, so it already covers the stated lookup; a separate `IX_..._EntityID_RecordID INCLUDE (FileID)` would be redundant. Saying "add one only if a measured plan later shows a need" is better than either adding it speculatively or staying silent.

**The `IN (ec.ConstraintName, ec.AltConstraintName)` NULL semantics are correct** — easy to get wrong. `a IN (b, NULL)` evaluates to `a = b OR UNKNOWN`, which is TRUE when the name matches and UNKNOWN otherwise; inside `EXISTS` / `NOT EXISTS` that behaves exactly as intended, so the 16 rows with a NULL alt still assert properly.

**Repo hygiene is clean**: `v6` folder matches the `v6.1.x` in the filename, `${flyway:defaultSchema}` used throughout, PG counterpart correctly absent per `migrations/CLAUDE.md`, no `EntityField` INSERT so the `Sequence` trap doesn't apply, and "no CodeGen" is right — a constraint change doesn't touch the generated ORM or entity metadata.

**The verification is real work.** Reproducing error 2627 *before* the change, confirming acceptance after, confirming the genuine duplicate is still rejected, running from-zero `mj migrate`, and then testing through `spCreateFileEntityRecordLink` — the procedure `BaseEntity.Save()` actually calls — rather than raw SQL. Being explicit about what wasn't run (the BaseEntity-layer snippet) is the right instinct.

## Requested before merge

**1. The changeset should be `minor`, not `patch`.** `.claude/rules/changesets.md` is unambiguous: `minor` when the branch touches the database, explicitly including "a versioned migration (`migrations/vN/*.sql`)." This PR adds one. I checked whether the certified-line exemption applies — it doesn't; that's ancestry-based, and this targets `next`, the Edge line where migration ⇒ minor holds. The rule calls out this exact direction as the damaging one: *"a database change with only `patch` entries **under**-bumps it... it only misbehaves when no other changeset in that release happens to carry a `minor`."* And per the repo's own `CLAUDE.md`, `check:changeset` has no CI gate — it's local-only — so nothing automated will catch this. One-word fix.

**2. Rename the migration's timestamp immediately before merge.** `Check migrations` is currently red:
```
Migration timestamp (202608211100) must be greater than the latest existing
migration timestamp (202608232100 from V202608232100__v6.1.x__IdentityClaim_UI_Read_RLS.sql)
```
Not a defect in the SQL itself — `next` has simply moved past this timestamp since the branch was cut. Nothing hardcodes the old filename elsewhere in the diff, so this is a one-file rename against whatever `next`'s max is at the moment of merge.

## Worth considering

**The pre-existing data loss is documented only where nobody will read it.** The migration header notes that v5.37 deleted duplicate link rows before adding each constraint, keeping the earliest `__mj_CreatedAt` per `(EntityID, FileID)` — so a deployment that legitimately had one file linked to several records of the same entity lost those rows then, and they aren't recoverable. That's operationally significant and doesn't appear in the changeset, which is what becomes release notes; migration header comments aren't something an operator reads. This PR is the natural place to flag it, since it's the one explaining why the constraint was wrong.

**No automated assertion for the new key.** `migrations/v5/tests/junction_unique_constraints/` is the existing precedent for migration tests, and this PR only loosens that test rather than adding a positive one for the new constraint. A small v6 counterpart — asserting the key columns are exactly `(EntityID, RecordID, FileID)` and that the same file can attach to two records of one entity — would turn the manual SQL Server 2022 verification into something that re-runs. Worth calibrating: this is exactly one instance of the pattern in the whole repo, so it's a suggestion rather than a convention violation.

## Bottom line

Correct, well-evidenced, and verified against a real instance rather than reasoned about. I'd merge the SQL itself as-is — the two requested items are both process/procedure, not correctness.

rkihm-BC:
Both of my earlier asks are addressed — migration renamed to `V202608241100` (clears `next`'s current max), and the changeset is now `minor` with the v5.37 data-loss note included. Thanks for both.

`Check migrations` is still red, but **not because of anything you wrote.** I traced it, and the fix is a one-line CI change I'd like folded into this PR.

## What's actually failing

The gate at `.github/workflows/changes.yml:274` scans **every changed `.sql` under `migrations/`**, not just Flyway migration files:

```bash
git diff --name-only <base> <head> | grep -E '^migrations/.*\.sql$' | while read file; do
```

I ran its exact logic against both `.sql` files in this PR:

- **`V202608241100__…Fix_FileEntityRecordLink_Unique_Key.sql`** — **passes.** Its two `__mj` mentions are both benign: `__mj_CreatedAt` doesn't match the `__mj[^_]` pattern (next char is `_`), and the `` `__mj.FileEntityRecordLink` `` in the header comment is swallowed by the single-quote stripper before the grep runs.
- **`migrations/v5/tests/junction_unique_constraints/02_assert_post_migration.sql`** — **fails**, 6 surviving `__mj.` matches.

That assertion script has **10 `__mj.` references, all pre-existing on `next`** — identical count there. And this PR's edits don't touch a single one of them; the diff only changes the `@ExpectedConstraints` table declaration and the two `i.name IN (...)` predicates.

The gate scans the **whole file rather than the added lines**. That fixture had simply never been modified in a PR before, so it had never been scanned. Touching it for an unrelated reason pulled its entire history into scope.

## Why the exclusion is right, not a loosening

Hardcoded `__mj` is the correct form for these scripts. Flyway never processes them — the filenames don't match `V<version>__` — so `${flyway:defaultSchema}` would stay literal text and break them. That's why both files in that directory use `__mj` (10 and 19 occurrences) and neither uses the placeholder.

**The repo already made this exact call one check away.** `.github/scripts/check-migration-id-determinism.sh:196-202`:

```bash
# Flyway migrations only: V<12-digit timestamp>__*.sql. Exclude tests/ fixture SQL
# (e.g. migrations/v5/tests/**/*.sql) — those aren't Flyway migrations and their
# NEWID() calls on throwaway assertion rows are legitimate.
files=$(git diff --name-only "$base" "$head" \
  | grep -E '(^|/)V[0-9]{12}__.*\.sql$' \
  | grep -vE '(^|/)tests?/' || true)
```

Same reasoning, same directory named explicitly. The `__mj` gate is the outlier here, not the exemption.

## Nothing else stops applying

I checked every step in the job — each does its own independent file selection, so this is surgical:

| Step | Selection | Fixtures already exempt? |
|---|---|---|
| Check migration filenames | `--diff-filter=A` (added only) | Yes — this PR modifies, doesn't add |
| Check timestamps newer | `--diff-filter=A` (added only) | Yes |
| Check duplicate versions | `find migrations -name 'V*.sql'` | Yes — fixtures aren't `V*.sql` |
| **Check migration content** | `git diff --name-only`, all changed `.sql` | **No — the only one** |
| Check ID determinism | `V[0-9]{12}__` + `grep -v tests/` | Yes — explicitly |

Three of the four other checks already skip fixtures structurally. The exclusion removes no coverage that exists today.

The second assertion in that same step (flyway-placeholder-in-a-timestamp-column) would also stop applying to fixtures, but it exists to catch `${flyway:defaultSchema}_CreatedAt` typos and fixtures contain zero placeholders by design — nothing for it to catch.

## The ask

**1. Add the exclusion**, mirroring the sibling script so the two guards agree:

```bash
git diff --name-only ${{ steps.base.outputs.sha }} ${{ github.sha }} \
  | grep -E '^migrations/.*\.sql$' \
  | grep -vE '(^|/)tests?/' | while read file; do
```

Worth a short comment pointing at `check-migration-id-determinism.sh` as the precedent, so the next reader doesn't have to re-derive why fixtures are exempt.

**2. Update the PR description** to say the PR now includes a CI-guard change and why. "Fix a unique key" and "narrow a migration guard's file scope" are separate reviews, and someone scanning the title shouldn't be surprised to find a workflow edit in the diff. A couple of sentences covering: the guard scans whole files rather than added lines, the fixture's `__mj` references are pre-existing and correct for a non-Flyway script, and the sibling determinism check already excludes the same directory.

Worth noting the underlying CI defect is broader than this PR — any future PR that touches a fixture under `migrations/` inherits its whole history against this guard. Fixing it here is the right call; a follow-up issue for the whole-file-versus-added-lines behaviour might be worth opening separately if you think it's likely to bite elsewhere.

MarceloT-BC:
@rkihm-BC — thanks, both rounds are in. Everything you asked for is pushed:

| Ask | Where |
|---|---|
| Changeset `patch` → `minor` | `5995a9a061` |
| Rename past `next`'s max → `V202608241100` | `5995a9a061` |
| v5.37 data-loss note into the changeset | `5995a9a061` |
| `grep -vE '(^|/)tests?/'` on the migration-content guard | `e41ab23654` |
| PR description discloses the CI change | description, top |

On the guard fix, I verified your citation before touching the workflow rather than taking it on faith: `check-migration-id-determinism.sh:197-202` does carry the identical exclusion for the identical directory. I also simulated the patched selection against this PR's changed files — the fixture drops out, `V202608241100__…` is still scanned and still passes — and re-parsed the YAML. Comment above the line points at the sibling script as the precedent.

**`Check migrations` is now green.** `Integration (SQL Server, deterministic)` is red, and it is inherited from `next`, not from this branch:

- The error is `mj sync push --ci` rejecting `./metadata/ai-models/.ai-models.json` — `MJ: AI Model Costs.Status` is `"Inactive"`, allowed values are `Pending, Expired, Invalid, Active`.
- This branch changes **zero** files under `metadata/`.
- `next`'s own Integration Tier run reports the same `Field "Status" has invalid value "Inactive"`; it went red at 19:13 and has stayed red across the merges since.
- That also explains the flip: PR checks run against `refs/pull/4006/merge`, so the run before this one picked up a green `next` and the ones after picked up the broken one.

Nothing to fix here for it — it clears when `next` clears.

One request: there's no approving review recorded on this PR (the API reports zero), so your sign-off exists only as comment text. If you're happy with it, could you drop the formal approval so it's on the record? And if you'd like the follow-up issue opened for the broader whole-file-versus-added-lines behaviour of that guard, say the word and we'll file it.

rkihm-BC (approved):
LGTM

Issue #4007, opened 2026-08-22
Integration sync: typed query parameter constraints per integration object — filtered sync and partitioned fetch under one abstraction


MS-BC:
## Problem

An integration object's fetch is all-or-nothing and strictly serial:

1. **All-or-nothing**: there is no way to sync a *subset* of an endpoint. If an object holds millions of rows and only one segment is needed, the sync walks everything. Nothing in metadata expresses "fetch under constraints."
2. **Serial**: one object = one cursor chain. Measured on a multi-million-row Salesforce object (200+ columns) in a production deployment: a fully tuned lane — server-side watermark filtering, per-page durable watermarks, TDS bulk writes, prefetch pipelining — sustains **~20,000 rows/min**, and that is the cursor's natural speed, not the engine's. The write side of the same pipeline benches at ~727k rows/min locally (~2% utilized). The fetch is the entire remaining wall.
3. **Untyped endpoint capabilities**: nothing in metadata records which query parameters an endpoint actually supports, so neither filtering nor partitioning has anything to stand on.

### v1 — QueryParam constraints (primitive)

A `QueryParam` is a constraint set — `(field, operator, value | range)` — attachable to an entity map / integration object fetch. The engine passes constraints through the existing fetch contract (the same way the watermark value flows today). This immediately enables **filtered sync**: fetch under constraints instead of everything.

### v2 — Typed, verified parameter metadata

Constraints graduate to a **typed attribute of the IntegrationObject** — default empty, later a dedicated table (e.g. `IntegrationObjectQueryParameter`) with per-parameter type, operators, and a **runtime-verified flag**.

Verification matters because vendors silently ignore unknown query parameters — a partition bucket whose filter is ignored returns the *entire* object and destroys the scheme. So a parameter is never used until a one-time runtime probe confirms the endpoint honors it (issue a filtered request, check the result set actually respects the filter, record the verdict in metadata). Endpoint capability discovery becomes mechanical, not documentation archaeology.

### The scheduler: partitioning and filtering are the same abstraction

The engine schedules sub-queries per object:

- a **covering set** of buckets = full sync, fetched in parallel;
- a **chosen subset** = filtered sync.

One abstraction, two features. Concurrency layers become a tree: running integrations → syncing entity maps (the existing lane concurrency) → **sub-queries per endpoint** (new), with every leaf feeding the shared write path.

### The zero-research default dimension

[...]

PR #4010, opened 2026-08-22
fix(integration-engine): classify undici transport failures as retryable

MS-BC:
First of an engine-patch set coming upstream (the same effort that produced #4007/#4008). This one is small, isolated, and fixes a failure mode that reports success.

## The defect

`fetch` (undici) reports **every** transport failure as the bare message `fetch failed` and puts the real reason — `ECONNRESET`, `ENOTFOUND`, `EAI_AGAIN`, `UND_ERR_SOCKET`, `socket hang up` — in `error.cause`.

`ClassifyError` only read the top-level message. `fetch failed` matches none of its patterns, so it fell through to `UNKNOWN_ERROR`/`Critical` → **not** in `IsRetryableError`. A routine network blip therefore ended the object's fetch loop, and the sync stopped early **while reporting success on a partial pull** — the worst shape of failure, because nothing anywhere says the object is incomplete.

Measured on a long-running production sync (3.8M-row Salesforce object): this fired every 30-60 minutes.

## The fix

`ClassifyError` now flattens the error's message plus every `cause` in the chain — depth-capped at 5, so a cyclic chain cannot hang it — along with any `code`/`errno` encountered, and checks an explicit list of transport signals **first**:

`fetch failed`, `socket hang up`, `econnreset`, `enotfound`, `eai_again`, `epipe`, `ehostunreach`, `enetunreach`, `etimedout`, `und_err_*`, `terminated`.

Ordering is deliberate: a request that never reached a server carries **no verdict**, so it must not be shadowed by a deterministic-looking keyword appearing deeper in the chain (e.g. a nested cause that happens to mention "validation"). These map to the existing `NETWORK_TIMEOUT`/`Warning`, which the retry path already honors — no new code or severity is introduced.

An explicit signal list rather than a loose substring, for the reason the neighbouring `DATABASE_ERROR` branch already documents: a bare `'sql'`-style match classified far too much as retryable and had to be narrowed. Deterministic errors keep their current classification, and that is asserted.

## Tests

New `ClassifyErrorTransport.test.ts`: the bare wrapper with no cause; each undici cause code read through the chain; `socket hang up` / `terminated`; transport-beats-nested-keyword; non-Error (string) input; a cyclic cause chain that must return rather than hang; and a no-over-broadening case asserting duplicate-key, FK-violation, write-verification and unmapped errors classify exactly as before.


MS-BC:
**"Run unit tests" is red here for a pre-existing reason on `next`, not for anything in this PR** — evidence rather than assertion:

- In [this PR's run](https://github.com/MemberJunction/MJ/actions/runs/32558461226/job/96996236901), **every** `Test Files` line reports `passed` (no failing suite anywhere in the log). The job exits 1 from BUILD errors in packages this PR does not touch: `Error: command codegen:manifest not found` in `@memberjunction/server-bootstrap-lite` and `mj_codegen_api`.
- The same **Unit Tests** workflow is failing on `next` itself — 2026-08-22 07:14 and 2026-08-21 07:19/07:14 — i.e. before and independent of this branch.
- Every other check on this PR is green, including `Integration (SQL Server, deterministic)`, `Check migrations`, `Detect missing and unused dependencies` and `adopted standards`.

This PR's own coverage (`packages/Integration/engine/src/__tests__/ClassifyErrorTransport.test.ts`) was additionally verified locally by compiling the pure function and running the cases directly: 18/18, including the cyclic-cause-chain case and explicit assertions that duplicate-key / FK-violation / write-verification / unmapped errors keep their existing classification (i.e. no over-broadening).

Happy to rebase once the `codegen:manifest` breakage is fixed on `next` if you'd like a fully green run before merging.

PR #4011, opened 2026-08-22
security: add SSRF protection to server-side web/HTTP actions

AN-BC:
## The vulnerability

Several server-side Actions in `@memberjunction/core-actions` fetched a **fully caller-controlled URL** and returned the response body, guarded by **only a scheme check** (`http:`/`https:`) — with no filtering of private/internal addresses and with redirects followed automatically. Because these actions are invokable by AI agents and workflows, this was a **read-SSRF**: an attacker could make the server fetch and read back:

- `http://169.254.169.254/…` — cloud metadata / IMDS (credential theft)
- `http://127.0.0.1:<port>/…` — loopback / local admin panels / internal MJAPI routes
- `10.x`, `192.168.x`, `172.16–31.x`, `100.64/10` (CGNAT), link-local, etc.

A hostname-string check is insufficient — **DNS rebinding** (a public name that also resolves to a private A/AAAA record) and **HTTP redirects** to an internal target both bypass it.

## The fix

A new **dependency-free** SSRF guard (`packages/Actions/CoreActions/src/custom/utilities/ssrf-guard.ts`, only `node:dns` + `node:net`):

- **`assertPublicUrl(rawUrl)`** — parses the URL, rejects non-`http(s)` schemes, resolves the hostname to **all** IPs via `dns.lookup(host, { all: true })`, and throws if **any** resolved address is in a blocked range. Literal-IP hosts (v4 and bracketed v6) are validated the same way, and non-resolving hosts fail closed.
- **`safeFetch(rawUrl, init?)`** — validates via `assertPublicUrl`, fetches with automatic redirects **disabled** (`redirect: 'manual'`), and on each 3xx `Location` re-resolves the target and **re-runs `assertPublicUrl` per hop** (default max 5). This closes the redirect + DNS-rebinding bypass.

### Blocked ranges

- **IPv4**: `0.0.0.0/8`, `10/8`, `100.64/10`, `127/8`, `169.254/16` (incl. IMDS), `172.16/12`, `192.0.0/24`, `192.0.2/24`, `192.88.99/24`, `192.168/16`, `198.18/15`, `198.51.100/24`, `203.0.113/24`, `224/4`, `240/4`, `255.255.255.255`.
- **IPv6**: `::1`, `::`, `fc00::/7`, `fe80::/10`, `ff00::/8`, and IPv4-mapped/compatible (`::ffff:a.b.c.d`, `::a.b.c.d`) — unwrapped and run through the IPv4 rules.

### Actions wired to the guard

| Action | Change |
|---|---|
| `http-request.action.ts` (arbitrary URL + method — most dangerous) | Converted from axios to `safeFetch`; auth/body/response handling preserved |
| `graphql-query.action.ts` (arbitrary endpoint) | Converted from axios to `safeFetch` |
| `web-page-content.action.ts` | `fetchWithRetry` now uses `safeFetch`; retry/timeout/size behavior kept |
| `url-metadata-extractor.action.ts` | Converted from axios to `safeFetch` |
| `url-link-validator.action.ts` | Each URL validated with `assertPublicUrl` before probing; follow-redirects path uses `safeFetch`; a blocked URL is marked invalid rather than crashing the batch |

Blocked URLs return a clean failure result (`ResultCode: "SSRF_BLOCKED"`, message "URL resolves to a private or reserved address and was blocked"). **Fixed-provider actions** (search, weather, webhooks, geolocation, etc.) were intentionally left untouched — they hit fixed endpoints, not caller-controlled URLs.

**Public-URL behavior is unchanged** — legitimate external requests resolve to public addresses and pass straight through.

## Tests

Added `src/__tests__/ssrf-guard.test.ts` (vitest, `node:dns` mocked): blocks `127.0.0.1`, `169.254.169.254`, `10.x`, `192.168.x`, `::1`, IPv4-mapped IPv6 loopback/metadata, and multi-record rebinding; allows public v4/v6; rejects non-http schemes; fails closed on non-resolving hosts.

## Verification

The full `pnpm run build` could not complete in this sandbox (workspace `node_modules` are not installed — every `@memberjunction/*` import is unresolved, and pre-existing generated files error for the same reason). Verification achieved instead:
- Strict `tsc --noEmit` of the guard in isolation (only node builtins): **clean**.
- `tsc --noEmit` of all five edited action files against module shims (MJ's build compiler options): **clean**.
- Runtime execution of the guard's classifier + `assertPublicUrl` (24 assertions): **all pass**.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01MmVtgPJabGEXa3td9qDxpn)_

### DIFF
--- a/.changeset/ssrf-guard-web-actions.md
+++ b/.changeset/ssrf-guard-web-actions.md
@@ -1,0 +1,5 @@
+---
+"@memberjunction/core-actions": patch
+---
+
+Add SSRF protection to server-side web/HTTP actions: private, loopback, link-local (incl. cloud metadata 169.254.169.254), and reserved IP ranges are now blocked, and redirects are re-validated per hop.
--- a/packages/Actions/CoreActions/src/__tests__/ssrf-guard.test.ts
+++ b/packages/Actions/CoreActions/src/__tests__/ssrf-guard.test.ts
@@ -1,0 +1,146 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+// ---------------------------------------------------------------------------
+// Mock node:dns so we control what a hostname resolves to, without real DNS.
+// The mocked `lookup` short-circuits literal IPs (mirroring Node's behavior)
+// and otherwise returns whatever the current test registered.
+// ---------------------------------------------------------------------------
+import { isIP } from 'node:net';
+
+const resolutions = new Map<string, string[]>();
+
+vi.mock('node:dns', () => {
+  return {
+    promises: {
+      lookup: vi.fn(async (hostname: string, _opts: { all: true }) => {
+        // Literal IPs resolve to themselves (as Node's dns.lookup does).
+        if (isIP(hostname) !== 0) {
+          return [{ address: hostname, family: isIP(hostname) }];
+        }
+        const addrs = resolutions.get(hostname);
+        if (!addrs || addrs.length === 0) {
+          const err = new Error(`getaddrinfo ENOTFOUND ${hostname}`);
+          throw err;
+        }
+        return addrs.map((a) => ({ address: a, family: isIP(a) }));
+      }),
+    },
+  };
+});
+
+import { assertPublicUrl, isBlockedIPAddress, SSRFError } from '../custom/utilities/ssrf-guard';
+
+beforeEach(() => {
+  resolutions.clear();
+});
+
+// =====================================================================
+// isBlockedIPAddress — pure classifier
+// =====================================================================
+describe('isBlockedIPAddress', () => {
+  it('blocks IPv4 loopback (127.0.0.1)', () => {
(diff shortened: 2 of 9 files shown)


AN-BC (commented):
@claude this looks good overall, however, the safe fetch and related util methods probably belong in a lower level place - if suitable to put into MJGlobal that is one option, if these are node only (which is proably right) then perhaps a lower level shared package that is not actions specific as these utils could be helpful in other places.

In addition to this, we should scan for axios and get rid of it for native node fetch and use safe approach and also remove axios from repo as part of this pass.

Finally, for public class members as well as package-exported functions like the ones in this util we use `PascalCase` as convention for our repo, not `camelCase` that is reserved for private/protected members as well as non-exported functions.

Please fix and push a new commit with these changes.

PR #4014, opened 2026-08-22
fix(codegen): a failed BEFORE/AFTER command must say what it said

MS-BC:
## The problem

A CodeGen run's success is decided entirely by its BEFORE/AFTER command failures —
`executeCodeGenPipeline` ends with `return this.commandFailures.length === 0`. The text that
explains those failures was destroyed twice on the way out:

1. **`runCommand` dropped the command's own output.** On a non-zero exit it rejected with
   `new Error('Process exited with code N')` while the accumulated stdout/stderr sat in a local
   that went out of scope. `recordCommandFailures` reads `r.error || r.output`, so all it could
   ever record was the exit code. A `npm run build` that failed because its source tree was
   absent and one that failed on a genuine type error produced the same string.

2. **`RunInProcess` returns a bare boolean**, so even the recorded `commandFailures` never
   reached the in-process caller. `RunWithResult` (the CLI path) does surface them as
   `errors[]` — the in-process path had no equivalent, and that caller is
   `RuntimeSchemaManager`, running inside a live server with no console for a human to read the
   pipeline's output from. Its RunCodeGen step could only ever throw
   `RSUError('CODEGEN', 'In-process CodeGen failed')`.

The combination is a blind spot with real consequences: a runtime schema update whose AFTER
command cannot succeed on the host reports a failed CodeGen step and aborts — after the tables
have been created but before the entity maps are — and nothing anywhere records which command
failed or what it said.

## The change

- `runCommand` appends a **bounded** tail of the command's combined output to the rejection —
  40 lines / 4000 characters, because a failing build can emit megabytes and this text ends up
  in logs and error payloads.
- `RunCodeGenBase` exposes `CommandFailures`, and `RunInProcess` logs each failure via
  `logError` when a run fails — so the reason is in the host log even for a caller that only
  checks the boolean.
- `IRSUCodeGenRunner` gains an **optional** `LastRunFailures()`. `RuntimeSchemaManager` appends
  what it reports to the `RSUError` message; a runner that cannot report detail degrades to the
  previous message rather than claiming a cause it does not have.
- `MJServer` wires `LastRunFailures` for both the SQL Server and PostgreSQL RSU setups.

No behavior changes: the same runs succeed and fail as before, they just say why.

## Tests

- `packages/CodeGenLib/src/__tests__/runCommand.test.ts` — a real failing shell command must put
  its stderr on the rejection alongside the exit code; `runCommands` must hand that same text
  back as the failed result's `error`, index-aligned with the command that produced it; and a
  200-line failure must be truncated to its tail.
- `packages/SchemaEngine/src/__tests__/RuntimeSchemaManager.test.ts` — the RSUError carries a
  runner-reported failure, degrades to the legacy message when the runner reports none, and
  still does not throw on success.

### DIFF
--- a/.changeset/rsu-codegen-command-failure-detail.md
+++ b/.changeset/rsu-codegen-command-failure-detail.md
@@ -1,0 +1,26 @@
+---
+'@memberjunction/codegen-lib': patch
+'@memberjunction/schema-engine': patch
+'@memberjunction/server': patch
+---
+
+Report WHY an in-process CodeGen run failed instead of only that it did
+
+A CodeGen run's success is decided entirely by its BEFORE/AFTER command failures
+(`executeCodeGenPipeline` returns `commandFailures.length === 0`), and until now the text
+explaining those failures was destroyed twice on the way out:
+
+- `runCommand` rejected a non-zero exit with `Process exited with code N` and dropped the
+  command's own stdout/stderr, so `recordCommandFailures` had nothing but the exit code to
+  record. A build that failed because a source tree was missing and one that failed on a real
+  compile error produced the same string.
+- `RunInProcess` returns a bare boolean, so even the recorded failures never reached the
+  in-process caller — and that caller (RuntimeSchemaManager, running inside a live server) has
+  no console for a human to read the pipeline output from. Its RunCodeGen step could only throw
+  `In-process CodeGen failed`.
+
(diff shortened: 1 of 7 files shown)


MS-BC:
Closing: the error-text improvement doesn't carry its weight as an engine change — the failure modes it decorated are better served by the structured warnings being folded into #3904/#3956, and the BEFORE/AFTER command surface itself isn't the defect.

PR #4015, opened 2026-08-22
fix(integration): REST DiscoverFields must report the declared primary key

MS-BC:
## The bug

`BaseRESTIntegrationConnector` has two converters over the same `IntegrationObjectField` entity:

| converter | used by | propagates `IsPrimaryKey`? |
| --- | --- | --- |
| `BuildSourceObjectInfo` | `IntrospectSchema` | yes |
| `FieldEntityToSchema` | **`DiscoverFields`** | **no** |

`FieldEntityToSchema` set `IsUniqueKey: f.IsUniqueKey || f.IsPrimaryKey` and never set
`IsPrimaryKey` at all.

Three code paths create field maps, and all three do the same thing with the flag:

```ts
fm.IsKeyField = field.IsPrimaryKey ?? false; // key on the record's PRIMARY key (identity)
```

Two of them (`processRSUPendingWork` in MJServer, and the re-resolve path in
`IntegrationDiscoveryResolver`) read `IntegrationEngineBase.GetIntegrationObjectFields(...)` —
the entity rows themselves, which carry `IsPrimaryKey` — so they are correct. The **initial
apply** (`createFieldMapsForEntityMap`) instead calls `connector.DiscoverFields(...)`, which for
every declarative REST connector goes through `FieldEntityToSchema`. Those field maps are
therefore created keyless no matter what the catalog declares.

`IsUniqueKey` is not a stand-in. `IsKeyField` was deliberately narrowed from unique to primary
(the "PK ≠ unique" note in that same resolver), because a primary key is one of possibly several
unique fields. That narrowing is correct — and it is what turned this omission from a lost flag
into a silent no-sync.

## How it presents

A live tenant, one REST connector, 30 objects. The 25 whose maps were created through the
engine-cache paths are keyed correctly. The 5 applied through the initial-apply path are not —
and their catalogs are unambiguous:

```
  ApplicationFile              declared=mediaId               isKeyField=<none>   <-- KEYLESS
  ApplicationRoundSubmission   declared=applicationId,roundId isKeyField=<none>   <-- KEYLESS
  ApplicationWinnerType        declared=id                    isKeyField=<none>   <-- KEYLESS
  Judge                        declared=userId                isKeyField=<none>   <-- KEYLESS
  Media                        declared=mediaId               isKeyField=<none>   <-- KEYLESS
  JudgeRecusal                 declared=applicationId,userId  isKeyField=applicationId,userId
  Application                  declared=id                    isKeyField=id
  ... 24 more, all keyed
```

`JudgeRecusal` proves a composite key maps fine through the working paths. Rows then arrive into
the keyless tables and stall, because nothing can be matched or reconciled — the failure looks
like a slow or empty sync, never like a metadata defect.

## The fix

Add `IsPrimaryKey: f.IsPrimaryKey` to `FieldEntityToSchema`, so `DiscoverFields` reports what the
catalog says — the same contract `BuildSourceObjectInfo` has always honoured. One line, and no
behaviour change for connectors that override `DiscoverFields` themselves.

Existing tenants need their already-created field maps repaired (set `IsKeyField` where the
mapped source field is a declared primary key); this fix stops new ones being created wrong.

## Tests

`packages/Integration/engine/src/__tests__/BaseRESTIntegrationConnector.discoverfields-pk.test.ts`
pins all four corners of the contract against a stub REST connector whose catalog declares
`mediaId` as the key and `alias` as merely unique: the declared PK is reported, a merely-unique
field is **not** promoted to PK, an ordinary field is neither, and the PK is still unique.

Related contract, one layer up: `BaseIntegrationConnector.pk-propagation.test.ts` (U1) pins that
`IntrospectSchema` must not coerce PK *silence* into `false`. This is the `DiscoverFields` end of
the same contract, where the flag was not coerced but simply dropped.

### DIFF
--- a/.changeset/rest-discoverfields-declared-pk.md
+++ b/.changeset/rest-discoverfields-declared-pk.md
@@ -6,11 +6,12 @@ Carry the declared primary key through a REST connector's DiscoverFields
 
 `BaseRESTIntegrationConnector.FieldEntityToSchema` — the converter every declarative REST
 connector's `DiscoverFields` runs each cached field through — set
-`IsUniqueKey: f.IsUniqueKey || f.IsPrimaryKey` and never set `IsPrimaryKey`. An apply builds each
-field map with `fm.IsKeyField = field.IsPrimaryKey ?? false`, so every field map of every
-declarative REST connector was created keyless, and `object-state`-style checks reported
-"NO KEY FIELD: every row is unmatchable, so writes cannot be reconciled" for objects whose catalog
-declared the key correctly. `IsUniqueKey` is not a substitute — a primary key is one of possibly
+`IsUniqueKey: f.IsUniqueKey || f.IsPrimaryKey` and never set `IsPrimaryKey`. The initial apply
+builds each field map from `DiscoverFields` with `fm.IsKeyField = field.IsPrimaryKey ?? false`, so
+newly applied objects on such a connector got keyless field maps and reported "NO KEY FIELD: every
+row is unmatchable, so writes cannot be reconciled" even though their catalog declared the key
+correctly. The other two field-map writers read the entity rows straight off the engine cache and
+were unaffected — which is why the symptom appears only on freshly applied objects. `IsUniqueKey` is not a substitute — a primary key is one of possibly
 several unique fields, and IsKeyField was deliberately narrowed from unique to primary (PK ≠
 unique), which is what turned this omission from a lost flag into a silent no-sync.
 
--- a/.changeset/rest-discoverfields-declared-pk.md
+++ b/.changeset/rest-discoverfields-declared-pk.md
@@ -1,0 +1,18 @@
+---
+'@memberjunction/integration-engine': patch
+---
+
+Carry the declared primary key through a REST connector's DiscoverFields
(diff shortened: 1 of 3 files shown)


MS-BC:
Closing — this was wrong, and the framework is already correct.

The keyless field maps I was chasing did not come from `FieldEntityToSchema`. They came from me calling the wrong entry point: `IntegrationApplyAllBatch`, which builds field maps from `connector.DiscoverFields()`.

The framework's actual path — selection -> RSU -> restart -> `processRSUPendingWork` — builds them from `schema.Objects[].Fields`, i.e. `IntrospectSchema`, which for a REST connector runs through `BuildSourceObjectInfo` and does propagate `IsPrimaryKey`. The re-resolve path reads the catalog entities directly and is equally correct. Both carry the key.

And where no key is declared anywhere, `DiscoverFieldsViaFetch`/`DiscoverFieldsViaStream` already derives a provable one from the streamed sample, and honestly returns PK-less so the object is not added.

So `DiscoverFields` staying conservative about PK claims is consistent with U1, not a defect. Nothing to change here.

PR #4018, opened 2026-08-23
DX: fix pnpm doc drift, close two CI gaps, add a browser-manifest leakage gate

AN-BC:
Follow-up to the DX audit in #2837, now closed. Re-auditing that plan against `next` found it ~60% shipped-or-wrong — the CI test gate, TypeDoc publishing, `mj migrate create` and the whole `@memberjunction/installer` effort had all landed, and the `.env` secret it flagged was a false alarm. This PR lands the items that actually survived, plus two gaps the original audit missed entirely.

Everything here is tooling and docs. No package source, no migrations, no metadata.

---

## 1. Two workflows had stopped firing for lockfile-only PRs

`test.yml`, `integration.yml` and `dependency-check.yml` still filtered on `package-lock.json` — a file that no longer exists. A PR changing only `pnpm-lock.yaml` (a dependency bump — precisely the change that most needs the suite) triggered **neither the unit-test nor the integration workflow**.

`test.yml` had the same bug twice. Its affected-scope fallback regex also named `package-lock.json`, so even a mixed PR that *did* trigger the workflow fell through to the narrow `--filter=...[base]` path — which selects nothing for a root-level file, the exact case the FULL-suite branch exists to catch.

The `paths:` filter is the load-bearing half: GitHub evaluates it before any job is created, so the in-job fallback can never rescue it. That trap is already documented in this very file, for the `tsconfig` entry.

Left alone deliberately: `publish.yml` and `docs.yml` check `pnpm-lock.yaml` first and fall back to `package-lock.json` when checking out an `lts/*` ref — `lts/5` is the last npm line, so that detection is correct as written. `docs-site/` is a separate npm project with its own legitimate lockfile.

## 2. New gate — `check:browser-manifest`

`packages/Angular/Bootstrap/CLAUDE.md` documents the June-2026 incident where `ng-bootstrap → @memberjunction/templates → aiengine/ai-provider-bundle` flooded the browser manifest with 27 server packages, breaking the Explorer bundle with esbuild `Could not resolve "crypto"`. Its prevention was a `grep` it asked contributors to run by hand — and that grep had **drifted from the prose rule directly above it**, omitting `server`, `ai-agents`, `ai-prompts`, `communication-engine` and `content-autotagging`.

Design decisions worth reviewing:

- **Parses module specifiers, never greps text.** The manifest bodies are full of colliding identifiers — `MJTemplateEntity`, `MJFileStorageProviderEntity`, `MJMCPServerEntity`. A text grep for "templates" or "storage" matches all of them.
- **Exact package-name matching.** `@memberjunction/ng-file-storage` is present in both manifests today and must never be confused with `@memberjunction/storage`; likewise `templates-base-types` vs `templates`, `ai-engine-base` vs `aiengine`, `server-bootstrap` vs `server`.
- **Curated denylist, not a Node-builtin scan.** `templates` imports zero Node builtins — its danger is purely transitive — so a source scan would miss the one package that caused the incident. A self-check fails the run if the AI provider bundle gains a member the denylist lacks, so tier 2 stays honest.
- **Tiers 1+2 only.** Roughly 18 further packages are server-only by builtin evidence but are named forbidden by no CLAUDE.md. They're absent from the manifests today, so denying them costs nothing now and blocks a legitimate future browser feature later, on no documented authority. The script says how to extend it.
- **Runs twice.** Unconditionally before the build, because leakage is introduced on PRs and PR runs are exactly where the freshness gate is skipped. Then again after the build on the full-suite path, placed *before* the freshness gate — otherwise a regenerated leak reports as an opaque "manifest is stale" diff instead of "a server package leaked in".

**Verified green on `next`** — all four guarded browser artifacts are clean, and the `ServerBootstrap` manifest (negative control) produces 34 hits.

## 3. The front door contradicted the build

`README.md` instructed `npm install`, `node InstallMJ.js` and `cp install.config.json.example` — in a pnpm workspace whose `CLAUDE.md` says *never run `npm install`*, where that example file no longer exists, and where the supported path is `mj install`. Install and develop are now separate labeled paths.

Also corrected: port 4200 → 4201; Node 20+/npm 9+ → Node 22+ (24 recommended)/pnpm 10.33+ per `PreflightPhase.ts` and `.nvmrc`; a `docker-compose up -d` with no root compose file; "migrations in /migrations/v2/" → the v2–v6 tree via `mj migrate`; three dead package rows whose links 404 (`ScheduledActions`, `ScheduledActionsServer`, `SkipTypes`); one wrong-case link; and the v6.0 upgrade guide, which existed but was unlinked while the repo sits at 6.1.

`CONTRIBUTING.md` gets the same treatment plus the integration tier, the local CI mirrors, the branch-tracking rule, and the changeset step.

`scripts/README.md` documented 6 of 70 files and opened by calling this an npm workspace. It's now a complete registry — every file categorized with purpose, trigger and invocation, and 17 historical one-offs marked so nobody runs a script that hardcodes a 2024 database name. Two entries are worth a look on their own: `fix-missing-dependencies.mjs` and `remove-unused-dependencies.mjs` both end by running `npm install` at the root.

## 4. Deferred, with the analysis preserved

Parallelizing the 9-step serial `mj:manifest` chain — the highest-value item from the original plan — is **not** in this PR, and `plans/manifest-generation-parallelization.md` explains why at length.

It cannot be validated without a full build, and its failure mode is silent. Every step runs with `syncDependencies` on, so it writes `<appDir>/package.json` while other steps walk it; the reader swallows a JSON parse error and drops that package *and its entire subtree* with `success: true`. On the gitignored `api`/`explorer` targets that surfaces as a class tree-shaken out of a shipped bundle with nothing in CI to catch it.

The analysis also turned up something that isn't a performance note: **the current serial order already violates that constraint** — `server-bootstrap` runs before `server-bootstrap-lite`, whose `package.json` it reads. Serial execution masks it today. The plan sequences the fix separately, because correcting it may change committed manifest output and that diff deserves its own review.

An unmeasured concurrency change to the build's critical path isn't worth shipping on an estimate.

---

## Verification

No `node_modules` in this environment, so nothing requiring a build was claimed or shipped. What was actually run:

- The gate: exit 0 on `next`; exit 1 on a planted `@memberjunction/templates` import with correct file/line annotation; exit 2 on both a renamed guarded file and a denylist-drift simulation.
- All 75 assertions from the new test suite, executed via plain node since vitest isn't installed here — including 11 false-positive traps and the ServerBootstrap negative control. The suite is committed for CI to run properly.
- YAML parse on all three edited workflows; JSON parse on `package.json`.
- All 229 relative links across `README.md` and `CONTRIBUTING.md` resolve, plus the new files' links.
- `check:claude-md` unchanged at its pre-existing baseline of 3 failures (root `CLAUDE.md` is already ~1.4KB over its byte ceiling on `next` — which is why the new gate is documented in the two Bootstrap files rather than added there).

**No changeset**: the three `CLAUDE.md` files touched sit in packages that publish only `dist/**`, so no published package content changes.

https://claude.ai/code/session_01NaZFxqTHA4JW9N5imAWVcz

---
_Generated by [Claude Code](https://claude.ai/code/session_01NaZFxqTHA4JW9N5imAWVcz)_


AN-BC:
## Sequencing — land this after #3804

@bg-BC — assigning this to you to pick up once #3804 (schema-scale CodeGen emit) is finished, since you're already on that one. Proposed order:

1. Finish and merge **#3804**.
2. Pull `next` into this branch, retest, merge **this**.

To be accurate about the reason: this is **not** a merge dependency. The two PRs share **zero files** — #3804 is `packages/CodeGenLib/src/**` plus generated barrels and the BigSchemaDemo tree; this one is docs, three workflow `paths:` filters, one new gate script, and three `CLAUDE.md` files. Either order merges cleanly. The sequence just gives this PR a clean retest against whatever #3804 lands, rather than testing against a `next` that's about to shift under it.

### What lands on #3804 when this merges

- **`check:browser-manifest`**, a new `test.yml` step. Passes on `next` today; #3804 doesn't touch the browser manifests, so no action expected there.
- **The `paths:` and affected-scope fixes** — `test.yml`, `integration.yml` and `dependency-check.yml` were all still keyed to the deleted `package-lock.json`, so a `pnpm-lock.yaml`-only PR triggered neither the unit-test nor the integration workflow.

### The one thing that genuinely depends on #3804

`plans/manifest-generation-parallelization.md` in this PR is analysis only — the parallelization is deliberately **not** implemented, because it can't be validated without a full build and its failure mode is silent.

When someone does pick it up, it should come **after** #3804, which introduces two primitives it should reuse instead of duplicating: `file-write.ts` (write-if-changed — the manifest generator writes unconditionally today) and `schema-emit.ts`'s bounded `mapLimit` (there's no bounded-concurrency helper in CodeGenLib on `next` at all). I've noted the same on #3804. The plan doc doesn't currently carry that dependency note — happy to add it if you'd rather it live in the file than in the thread.

---

rkihm-BC (approved):
Approving. This is a claim-heavy description and every claim I could test held.

**Verified:**
- **Lockfile-trigger bug is real on `next`.** All three workflows still filter on `package-lock.json` (which no longer exists at the root), and `test.yml` names it a second time in the full-suite fallback regex. A `pnpm-lock.yaml`-only PR triggers neither the unit-test nor the integration workflow today. Leaving `publish.yml` / `docs.yml` alone is correct — they check `pnpm-lock.yaml` first and fall back only for `lts/*` refs.
- **The gate, in a worktree of this branch:** exit 0 on the committed browser manifests (37 packages denied); a planted `@memberjunction/templates` import is reported with the exact file and line and the fix instruction; the six real false-positive pairs already present (`ng-file-storage`, `templates-base-types`, `ai-engine-base`, …) do not trip it. Its vitest suite passes (32 tests) via the repo's `.github/scripts` config. All three edited workflows parse.
- **The Bootstrap `CLAUDE.md` grep drift is real:** the hand-run grep on `next` omits `server`, `ai-agents`, `ai-prompts`, `communication-engine` and `content-autotagging`, all forbidden by the prose beside it. Parsing specifiers instead of grepping text is the right design for exactly the reason given (`MJTemplateEntity` and friends in the import bodies).
- **README corrections are real:** `next` still says `npm install`, `node InstallMJ.js`, a nonexistent `install.config.json.example`, port 4200 and `docker-compose up -d`; `PreflightPhase.ts` has min Node 22 / recommended 24 and `.nvmrc` is 24; the three dead package rows point at directories that do not exist; the two scripts named as running `npm install` at the root do.
- **The plan document's headline finding is true**, checked independently against the workspace graph: `server-bootstrap → codegen-lib → server-bootstrap-lite`, so its walk reads a `package.json` the lite step may rewrite, and `readPackageJson` swallows a parse error and returns null, dropping the subtree silently. Serial execution masks it today; deferring the parallelization was the right call, and the Ordering caveat added to the CodeGenLib `CLAUDE.md` is accurate.
- `check:claude-md` baseline is 3 pre-existing failures on `next`, as stated. #3804 (AN-BC's sequencing request) merged 27 Aug; the branch merged `next` today.

**Good calls:** running the gate before the build unconditionally and again on the full-suite path *before* the freshness gate; an explicit allowlist with a missing file as a misconfiguration rather than a pass; the bundle-coverage self-check; declining to deny ~18 more packages on no documented authority.

**Optional nits:** the two manifest paths added to `test.yml`'s `paths:` are already matched by the existing `packages/**` entry, and the comment above them conflates the workflow trigger with the turbo filter — harmless, but inaccurate. The description says 75 assertions; the suite has 32 test cases.

**Changeset:** none, and that is correct here — the only files under `packages/` are three `CLAUDE.md` documents in packages that publish `dist` only, so no published content changes.

PR #4019, opened 2026-08-23
security: extract SSRF guard into @memberjunction/network-utils and remove axios repo-wide

AN-BC:
> **Supersedes #4011.** This branch contains that PR's commit plus the changes its review asked for. #4011 can be closed once this lands — see [Relationship to #4011](#relationship-to-4011).

## Why

#4011 added an SSRF guard to five server-side web/HTTP actions. The review on it raised three things, all of which amount to the same observation: **the guard was in the wrong place, and it was fighting the codebase rather than fitting into it.**

- The guard lived under `CoreActions`, but "don't let the server fetch internal addresses" is not an Actions concern.
- Ten other packages were still reaching for `axios`, so the guard was unreachable from any of them.
- The new exports used `camelCase`, which is not what this repo uses for a public surface.

Fixing the first two together is what makes the guard actually enforceable. As long as every package brings its own HTTP library, "route caller-controlled URLs through the guard" is a convention nobody can check. With one client, it is a single option flag at every call site.

### 1. New package: `@memberjunction/network-utils`

Node-only, **zero dependencies** — `node:dns` and `node:net` only. That is what lets it sit at the bottom of the dependency graph so anything server-side can use it without a cycle.

It is deliberately *not* in `@memberjunction/global`: that package ships to the browser, and `node:dns` would break the bundle. The review floated MJGlobal as an option; this is why it isn't.

| Export | Purpose |
|---|---|
| `AssertPublicUrl(url)` | Validate a URL, resolving **all** of its addresses |
| `SafeFetch(url, init?)` | `fetch` with per-hop SSRF re-validation |
| `IsBlockedIPAddress(addr)` | Classify one IP literal — pure, no DNS |
| `SSRFError` | Thrown when a URL is blocked |
| `HttpRequest` / `HttpGet` / `HttpPost` / … | Native-`fetch` HTTP client |
| `SafeHttpRequest` | `HttpRequest` with `ValidateUrl` forced on |
| `HttpClient` | Configured client — replaces `axios.create(...)` |
| `HttpError` / `IsHttpError` / `IsCancellationError` | Replace `isAxiosError` / `isCancel` |

### 2. `axios` removed from the repository

| | |
|---|---|
| Import sites converted | 43 |
| Packages | 11 |
| `axios.create` instances with interceptors | 12 |
| Test files re-mocked | 18 |

Also gone: the deprecated `@types/axios` stub, and `form-data` — the multipart uploads now use the WHATWG `FormData`/`Blob` that `fetch` consumes natively, so `getHeaders()` has no equivalent and needs none.

Interceptors became explicit hooks. The behaviour is the same; the difference is that a client's behaviour is now visible at its construction site rather than appended to a mutable global chain:

```ts
// before
instance.interceptors.response.use(null, async (error) => {
    if (error.response?.status === 429) {
        await sleep(60_000);
        return instance.request(error.config);   // you re-issue it
    }
    return Promise.reject(error);
});

// after
new HttpClient({
    OnRetry: async (error) => {
        if (error.Status !== 429) return false;
        await sleep(60_000);
        return true;                              // the client re-issues it, bounded by MaxRetries
    },
});
```

`OnRequest` re-runs before each retry, so a token refreshed inside `OnRetry` is picked up automatically.

### 3. `PascalCase` for the public surface

`assertPublicUrl` → `AssertPublicUrl`, `safeFetch` → `SafeFetch`, `isBlockedIPAddress` → `IsBlockedIPAddress`, and `SafeFetchInit.maxRedirects` → `MaxRedirects`.

## Two things found along the way

**An SSRF sink the first pass missed.** The `API Rate Limiter` action takes a fully caller-controlled URL and returns the response body — the same shape as the actions #4011 fixed. It is now guarded and reports a blocked URL as `SSRF_BLOCKED`.

**~120 silently untyped property reads.** axios types `response.data` as `any`, so a large amount of BizApps provider code was reading untyped payloads without anyone noticing. Rather than paper over it, each provider gained real response interfaces — Graph API paged envelopes, the Twitter v2 / HootSuite / LinkedIn collection envelopes, OAuth token responses, media-upload states. Those call sites are now genuinely typed, which is a strict improvement on the status quo ante.

## Design decision worth reviewing: `ValidateUrl` defaults to off

`HttpRequest` and `HttpClient` do **not** guard by default. This is the one judgement call in the PR that a reviewer should push on if they disagree.

The reasoning: most MJ call sites talk to fixed provider endpoints (`api.twitter.com`, `graph.facebook.com`) where guarding buys nothing and costs a DNS lookup per request — and some legitimately talk to internal hosts, where guarding would be a regression. MJServer posts to its own CodeGen API on localhost; MetadataSync resolves `@url:` references on a developer's machine. Both would break.

So the rule is not "always on", it is **on wherever the URL is caller-controlled**. `SafeFetch` and `SafeHttpRequest` have it on and cannot turn it off, and the README says plainly which call sites qualify. Every site converted in this PR was classified individually, not by default.

## ⚠️ Merge blocker: npm setup for the new package

`check-new-npm-packages` **will fail on this PR** until `@memberjunction/network-utils` exists on npm with a trusted-publisher configuration. That is irreducibly manual (npm's 2FA gate), so it needs a human:

1. Create the package on npm and run `npm trust github` for it — repo `MemberJunction/MJ`, workflow `publish.yml`, `allow-publish`, no environment.
2. Run the **Seed new package** workflow.

The gate names **@cadam11** as the contact. This is expected for any new package, not a defect in the change — see `NEW_PACKAGE_SETUP.md` and `DEPLOYMENT.md` Step 5. Everything else is green.

## Verification

- **Build: 275/275 packages.**
- **Unit tests green in every touched package** — network-utils 53, CoreActions 360, BizApps Social 248, BizApps FormBuilders 86, ContentAutotagging 225, MJServer 1081, MetadataSync 310, ApolloEnrichment 24, plus BettyBot / HeyGen / Recommendations-Rex.
- Gates passing: `check:standards`, `check:changeset`, `check:codegen-tail`, internal peer-dep check.
- `check:claude-md` fails on the root file's byte ceiling and a link to the M5 workspace file — **both reproduce on `next` without this branch**, so they are pre-existing and untouched here.

## Where to focus review

1. **The default above** — the one real judgement call.
2. **`packages/NetworkUtils/src/SSRFGuard.ts`** — the security-critical code. The address classifier and the per-hop redirect re-validation are the parts worth reading closely.
3. **The BizApps response interfaces** — derived from what each call site actually reads. Where a leaf action read an entity's fields directly off the body rather than unwrapping an envelope, I typed it to match the **shipped runtime behaviour** rather than "fixing" it. If any of those turn out to be latent bugs, they are pre-existing and deserve their own change rather than being quietly altered here.

Everything else is mechanical and covered by tests.

## Behaviour changes

Deliberately none, with two exceptions, both intended:

- `API Rate Limiter` now rejects caller-supplied URLs that resolve into private space (`SSRF_BLOCKED`).
- Multipart uploads send a `fetch`-generated boundary instead of a `form-data`-generated one. Wire-equivalent.

## Relationship to #4011

This branch was cut from #4011's head, so it contains that PR's commit plus this work, and it has current `next` merged in. Targeting `next` directly keeps the diff free of the unrelated commits that show up when diffing against #4011's (now-behind) branch, and makes this a single mergeable change. **Close #4011 as superseded once this lands.**

## Docs

`@memberjunction/network-utils` ships a full README — SSRF threat model, a sequence diagram of the guard, the blocked-range tables, an explicit *what this does not protect against* section (including the TOCTOU rebinding gap this does **not** close), an axios migration table, and a testing section documenting the two mocking traps. Every exported symbol carries JSDoc with `@throws` and worked examples.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AktaXqrL4Sz5v8ZNCHnYDu)_


AN-BC:
## CI status: 5 green, 2 red — neither red is fixable in this PR

| Check | |
|---|---|
| Check migrations | ✅ |
| Integration (SQL Server, deterministic) | ✅ |
| Detect missing and unused dependencies | ✅ |
| packages/** must not peer-depend on @memberjunction/* | ✅ |
| adopted standards | ✅ |
| Run unit tests | ❌ pre-existing, see below |
| Check new packages exist on npm | ❌ needs a human, see below |

The dependency check passing is the meaningful one for this change — it validates adding `@memberjunction/network-utils` and dropping `axios`, `@types/axios`, and `form-data` across 11 packages.

---

### ❌ Run unit tests — pre-existing, in a package this PR does not touch

**`@memberjunction/ai-realtime-client`**, 3 tests, all `ReferenceError: CloseEvent is not defined`:

```
gemini-realtime-client-extended.test.ts:406  client.LastConnectArgs?.OnClose(new CloseEvent('close'));
gemini-realtime-client.test.ts:785           (same)
```

**Why it is not this PR's:**

- `git diff origin/next...HEAD -- packages/AI/RealtimeClient/` is **empty** — the package's source and tests are byte-identical to `next`.
- Its only dependencies are `@memberjunction/global`, `@memberjunction/ai`, and `@google/genai`. **None of the three is in this PR's changed set.** Identical inputs, identical code — this branch cannot change the outcome.
- The same job flaps on `next` itself. Runs on **the same commit** `a8e326e81` both passed (Aug 23 07:15) and failed (Aug 22 07:14, Aug 21 07:19). That is environment sensitivity, not a broken commit.

**Root cause.** `CloseEvent` is not a Node global on every version — Node 22.22.2 exposes `WebSocket` and `Event` but **not** `CloseEvent`. Whether these tests pass depends entirely on the runner's Node. That fully explains the flapping.

**Proposed patch** (verified locally: **404/404 pass**, up from 401/404):

```diff
- client.LastConnectArgs?.OnClose(new CloseEvent('close'));
+ client.LastConnectArgs?.OnClose(new Event('close') as CloseEvent);
```

4 call sites, both Gemini test files. This is safe because the driver ignores the argument entirely:

```ts
// geminiRealtimeClient.ts:222
OnClose: () => this.handleTransportClose(),
```

`CloseEvent` appears in the source only as an erased type annotation on the callback interface (`OnClose: (event: CloseEvent) => void`), so production code is unaffected — this is purely a test-time construction problem.

**I have deliberately not applied it here.** It is an unrelated package, and folding it into an already-large refactor would widen the diff and couple two unrelated reviews. Happy to raise it as its own small PR — it would also stop `next` flapping. Say the word.

---

### ❌ Check new packages exist on npm — expected, needs npm access

`@memberjunction/network-utils` is new, so it must exist on npm with a trusted-publisher configuration before this can merge. That is 2FA-gated and cannot be done from CI:

1. Create the package on npm, then `npm trust github` — repo `MemberJunction/MJ`, workflow `publish.yml`, `allow-publish`, no environment.
2. Run the **Seed new package** workflow.

The gate names **@cadam11** as the contact. Per `NEW_PACKAGE_SETUP.md` / `DEPLOYMENT.md` Step 5. I have not worked around this by marking the package `private` — the gate warns against exactly that, and it would silently ship the package to nobody.

---

Nothing else is outstanding. I'm watching this PR and will pick up review comments or any new CI failure.

---

PR #4012, opened 2026-08-22
feat(claims): identity claims infrastructure, atomic CAS engine, seed metadata, and forms

AN-BC:
## Summary

Introduces the **Identity Claims Infrastructure & Engine**, enabling cross-cutting, pluggable identity claim lifecycles (such as entitlement grants, magic link verification, person account linking, and guest order adoption) with atomic concurrency safety and downstream extensibility.

#### 1. Core Claim Engine (`@memberjunction/core-entities` & `@memberjunction/core-entities-server`)
- **`IdentityClaimEngine`**: Client & shared engine managing claim type lookups, email normalization, bearer token hashing (`SHA-256`), and abstract `BaseIdentityClaimDriver` contract with explicit idempotency requirements.
- **`IdentityClaimEngineServer`**: Server-side engine providing secure claim creation, single-use atomic Compare-and-Swap (CAS) state transitions (`Pending` -> `Claimed`), exception wrapping, and automatic state rollback (`revertClaimAtomic`) if a downstream driver fails or throws.
- **Concurrency & Fail-Closed Safety**: Provider availability checks fail closed, and strict concurrency test assertions verify `[Status] = 'Pending'` predicate and single-execution guarantees.

#### 2. Database Schema & Migration (`migrations/v6/`)
- **Migration `V202608202300`**: DDL for `IdentityClaim` and `IdentityClaimType` tables, permissions, extended properties, foreign keys, and indexes (composite `IX_IdentityClaim_NormalizedEmail_Status`).
- **Seed Data**: Seeds default `IdentityClaimType` records (`EntitlementGrant` active with deterministic GUIDs and `IF NOT EXISTS` idempotency guards; placeholder types seeded inactive).
- **Metadata Sync**: Added `metadata/identity-claim-types/` with `.mj-sync.json` and `.identity-claim-types.json`, registered in `metadata/.mj-sync.json`.

#### 3. Explorer Forms & Generated Classes
- Added `MJIdentityClaim` and `MJIdentityClaimType` form components to `@memberjunction/ng-core-entity-forms`.
- Updated generated entity subclasses and class registration manifests across Bootstrap and Server modules.
- Clean core generated operations (zero foreign symbols).

#### 4. Verification & Unit Tests
- All 24 test files (**420 tests passing**) in `MJCoreEntitiesServer` and 30 test files (**831 tests passing**) in `MJGlobal`.

PR #4021, opened 2026-08-23
fix(core-entities): scope Identity Claim reads, thread providers, and move the claim lifecycle server-side

AN-BC:
## Summary

Three related fixes to the identity-claim engines, all about *what data a claim operation can reach and from where*:

1. **Read scope** — `MJ: Identity Claims` is readable in full by the base **UI** role, exposing guest emails and their purchase linkage. Fixed with a row-level filter plus a system-user read inside `RedeemClaim`.
2. **Provider threading** — both engines reached for the process-global metadata provider from per-request methods, behind 8 `global-provider-ok` suppressions. All 8 removed.
3. **Lifecycle placement** — the client+server engine carried its own `RedeemClaim` that had *none* of the hardening the server copy received. Removed; the lifecycle is now server-only.

⚠️ **Two deliberate breaking changes**, both in §2 and §3. No caller outside MJ is affected — the only downstream consumer (`bizapps-orders` `CheckoutSessionService`) uses `CreateClaim`, which stays source-compatible.

---

# 1. Read scope

## Severity — stated honestly

**The `TokenHash` is not the exposure.** The token is `crypto.randomBytes(32)` and is not recoverable from its SHA-256, and redemption requires `userEmailMatch || tokenValid`. The issue is **PII enumeration**.

**Nor is the grant anomalous.** The three-role tuple is the hardcoded CodeGen default (`CodeGenLib/src/Config/config.ts:380-387`), and every comparable entity carries it — API Keys, Credentials, OAuth Tokens, **Magic Link Invites (its own `TokenHash`)**, Users, Employees. The honest framing is *"this entity needed tuning the generator can't do."*

## The RLS filter

`migrations/v6/V202608232100__v6.1.x__IdentityClaim_UI_Read_RLS.sql` keeps `CanRead` and attaches a `ReadRLSFilterID`, following `UI: Own AI Agent Runs`:

```sql
[ClaimedByUserID] = '{{UserID}}' OR [NormalizedEmail] = LOWER(LTRIM(RTRIM('{{UserEmail}}')))
```

Matches `NormalizedEmail` **as well as** `ClaimedByUserID` because the latter is NULL until redemption — an ID-only filter would hide every *pending* claim from the user entitled to redeem it.

**Why SQL and not `metadata/`:** `EntityPermission` has no unique constraint on `(EntityID, RoleID)`, and #4012's INSERT omitted `[ID]`. A metadata entry with a fresh key would add a *second* permission row — and since `UserExemptFromRowLevelSecurity` returns exempt on the first NULL-filter row it finds, that duplicate would **silently cancel this fix**.

## The system-user read (required, not optional)

Row filters apply to **single-record loads**, not just `RunView` (`GenericDatabaseProvider.ts:4390-4396`). So the filter alone would quietly break this entity's own **workflow #3** — *"when a purchase email differs from the login account email, a verification token confirms ownership"* — since the redeemer's email doesn't match and they could never load the claim. `RedeemClaim` therefore reads under `UserCache.Instance.GetSystemUser()`, falling back to `contextUser` when the cache is unpopulated. **Authorization is unchanged** — still `userEmailMatch || timingSafeEqual(TokenHash)`.

---

# 2. Provider threading

`.claude/rules/data-access.md:57-94` names this case as wrong — *"server-side code servicing multiple tenants/connections"* — and scopes the hatch to CLI scripts, bootstrap, and self-registering singleton initializers. These per-request methods are none of those. The scanner's `FALLBACK_PATTERNS` make `provider ?? new Metadata()` **structurally exempt**, so a compliant form was available and equally cheap.

**`RedeemClaim` now requires a provider** — deliberately breaking:

```ts
public async RedeemClaim(claimID: string, contextUser: UserInfo, provider: IMetadataProvider, token?: string)
```

Redemption reads a claim, reads an invite, and executes a raw CAS `UPDATE`; all three must hit the same database. An *optional* provider would let a caller thread one into the entity reads while the CAS fell back to the global. `CreateClaim`, `RevokeClaim` and `GetPendingClaimsForEmail` take an optional trailing `provider?` and stay source-compatible.

**The latent CAS split this closes:** `consumeClaimAtomic` resolved schema/table names from the passed `md` but took `ExecuteSQL` from the global. Identical objects today — but the statement would have been built for one database and run against another the moment anyone threaded a provider. **Threading without this would have created the bug**, which is why they land together. Also drops `Metadata.Provider || (Metadata as unknown as {...}).Provider`, whose operands were the same static getter.

Two bare `new RunView()` calls resolved the **separate** `MJ_RunViewProvider` global slot the compliance scanner doesn't cover; both now receive the engine's provider. `IdentityClaimEngineServer` gains a settable `Provider` accessor with a `?? new Metadata()` fallback — the `AIEngine` shape, scanner-exempt.

---

# 3. Lifecycle moved server-side

## The problem

The server engine wasn't extending the client — it independently **reimplemented** `CreateClaim` / `RedeemClaim` / `RevokeClaim` / `GetPendingClaimsForEmail`. The two copies had diverged:

| | client | server |
|---|---|---|
| `createHash` / `timingSafeEqual` / `TokenHash` occurrences | **0** | 10 |
| email-match check | ❌ | ✅ |
| atomic CAS before driver | ❌ check-then-set | ✅ |
| driver wrapped in try/catch | ❌ | ✅ |
| revert on driver failure | ❌ | ✅ |

The client accepted a `token` and passed it to `driver.OnClaim` **without verifying it** — the parameter was decorative. It read `Status !== 'Pending'`, awaited the driver, then wrote `Status = 'Claimed'`, so two concurrent redemptions both passed. The driver ran *before* the transition, so a failure left side effects applied with the claim still Pending.

Every one of those was fixed on the server copy only — leaving a weaker implementation of a security-critical operation exported from a package server code also imports, where `IdentityClaimEngine` and `IdentityClaimEngineServer` differ by one word.

## The split

`IdentityClaimEngine` keeps what is safe in any host: the `IdentityClaimType` cache, type lookups, `ClassFactory` driver resolution, `NormalizeEmail`, and the `BaseIdentityClaimDriver` contract. Its public surface goes 11 → 6 members.

`IdentityClaimEngineServer` **contains** an instance of it via `Base` and proxies those cached members — the same containment split as `AIEngineBase` / `AIEngine`, now documented on both classes. The lifecycle lives only here, because each operation needs something a browser cannot do: `randomBytes` generation, SHA-256 hashing, `timingSafeEqual`, raw-SQL CAS, and MJ Communications dispatch.

`AutoClaimForUser` moves to the server engine and redeems through the hardened `RedeemClaim`, so the email lookup that discovers pending claims is a convenience rather than the security boundary.

Nothing outside the engines called the removed methods (repo-wide search: only the barrel export and their own tests).

---

## Tests

`IdentityClaimEngineServer.test.ts` 13 → 16; `identityClaimEngine.test.ts` trimmed to the 4 cache/lookup specs that still apply.

- **`should read the claim under the system user, not the redeeming user`** — verified discriminating: reverting the engine change makes it fail.
- **`should fall back to the context user when the system user cache is unpopulated`**.
- **`AutoClaimForUser redeems each pending claim through the full RedeemClaim gate`**.
- The concurrency and driver-exception specs now mock `ExecuteSQL` on the **passed** provider. They failed when first re-run, which is the change working — the CAS genuinely honours the threaded provider now.

## Verification

- `IdentityClaimEngineServer.test.ts`: **16 passed**, no type errors.
- `identityClaimEngine.test.ts`: **4 passed**, no type errors.
- `MultiProviderCompliance.test.ts`: **passes** — 8 suppressions removed, none added.
- Both packages build clean; `check-spec-antipatterns.mjs` clean on both.
- `check-migration-id-determinism.sh --self-test`: passes; migration uses a deterministic GUID, no `NEWID()`, no `spCreate`.

## Follow-up, not done here

When redemption is wired to a resolver or Remote Operation, that caller should pass its request-scoped provider (`GetReadWriteProvider(providers)`, `MJServer/src/util.ts:196`) and be re-checked against these permissions.

https://claude.ai/code/session_01EroBiLcz3tdoFjNLxTbvZi


AN-BC (commented):
LGTM

AN-BC (commented):
LGTM

PR #4024, opened 2026-08-23
perf(integration): cover the record-map lookup every read path shares

MS-BC:
## What this is really about

Every access path to `CompanyIntegrationRecordMap` resolves a row by the same three columns — `(CompanyIntegrationID, EntityID, ExternalSystemRecordID)`:

- `RecordMapBatch.readExisting`, the batched writer's chunk lookup — **on the hot path of every sync**;
- `LoadAllRecordMaps`, the orphan-sweep paging;
- `SaveRecordMap`, the per-record upsert lookup.

The table carried only its two single-column auto-FK indexes, so each of those reads picked one FK index and key-looked-up the rest — paid per record, per sync, on a table holding one row for every record ever synced.

One covering composite serves all three with no key lookup:

```sql
CREATE NONCLUSTERED INDEX [IDX_CompanyIntegrationRecordMap_Identity_Covering]
    ON [${flyway:defaultSchema}].[CompanyIntegrationRecordMap]
        ([CompanyIntegrationID], [EntityID], [ExternalSystemRecordID])
    INCLUDE ([EntityRecordID]);
```

The key stays under the 1700-byte nonclustered cap: `16 + 16 + nvarchar(750)` = 1532 bytes. It is idempotent (`IF NOT EXISTS`), and a plain `CREATE INDEX` so it transpiles cleanly. This is **not** a foreign-key index in the sense migration rule 4b forbids — that rule is about single-column indexes duplicating CodeGen's `IDX_AUTO_MJ_FKEY_*`, which this neither replaces nor collides with. It also produces no CodeGen output: the composite-unique probe filters on `is_unique = 1`, and this index is non-unique, so no entity metadata moves.

T-SQL only, per `migrations/CLAUDE.md` — the PG counterpart is build-engineer work at release time.

## The second, smaller half: make the two write paths agree

`RecordMapBatch.flushChunk` already refuses to write a row that would not change:

```ts
const work = chunk.filter(m => existing.get(m.ExternalID)?.EntityRecordID !== m.EntityRecordID);
```

`SaveRecordMap` — the per-record path — did not. Its upsert lookup found the row, then `Load` + `Save` rewrote it with the values it already held, and bumped `__mj_UpdatedAt` for nothing.

**To be clear about the scope, because it is narrower than it looks:** the apply loop builds a `RecordMapBatch`, so the ordinary incremental path is *already* filtered and was never paying this. `QueueRecordMap` falls through to `SaveRecordMap` only when there is no batch — call sites outside the apply loop, and `RecordMapBatch`'s own per-record fallback when a set-based write fails. That fallback is exactly where the cost hurts most: a failed batch currently degrades into two round trips *per record*, and afterwards it degrades into one.

So this is a consistency fix, not a hot-path win. The hot-path win is the index above.

**The boundary the skip never absorbs:** a row pointing at a *different* MJ record — the stale-mapping repair case — is still loaded and rewritten, and a missing row is still created. Only the genuine no-op is skipped.

## Tests

The idempotent-upsert test pins the skip (one `Save` across two runs of an unchanged mapping, no duplicate), and a new test pins the repair boundary — a map row pointing at a different MJ record is rewritten to the current PK.


packages/Integration/engine/src/IntegrationEngine.ts
```diff
@@ -5069,16 +5069,22 @@ export class IntegrationEngine extends BaseSingleton<IntegrationEngine> {
...
             BypassCache: true, // upsert-by-identity: a stale miss here re-creates a duplicate record map
         }, contextUser);
 
         if (existing.Success && existing.Results.length > 0) {
+            // The row already says exactly this. On an incremental sync that is nearly every record
+            // — mappings are stable — and the Load + Save this path used to do anyway were two more
+            // round trips to change nothing, plus a meaningless __mj_UpdatedAt bump on every synced
+            // record's map row. The batched writer (RecordMapBatch.flushChunk) has always skipped
+            // this case; the per-record fallback now agrees with it.
+            if (existing.Results[0].EntityRecordID === entityRecordID) return;
```
MS-BC, line 5209: good change, if the record it matches perfectly then no need for loading the result and saving

PR #4025, opened 2026-08-23
perf(core): let a caller skip the full-page fallback COUNT it never reads

MS-BC:
## The cost

When a page comes back exactly full (`Results.length === MaxRows`), `InternalRunView` runs the count SQL as a **second, sequential round trip** after the data query, to report the true total behind the cap. For a `MaxRows: 1` existence/lookup read, finding the row IS a full page — so every successful single-row lookup paid double round trips to compute a total it never read. Measured on a live sync workload: the fallback fired exactly at the hit rate of the two per-record lookups (record-map resolution and key-field matching) and cost **~11% of all SQL time**.

## Why not just drop the count for MaxRows:1

Because it's load-bearing. The vector management dashboard reads `TotalRowCount` as the real vector count on a `MaxRows: 1` query (fetch one row, read the total), and the cache-gauntlet integration check asserts it *exceeds* the truncated row count on a MaxRows:1 subset slot (#3195). A blanket "MaxRows:1 means the count is meaningless" would break both — this PR deliberately does not do that.

## The fix: an explicit opt-out

`RunViewParams.SkipTotalRowCount` for reads that only consume `Results`:

- skips the full-page fallback count (the per-record lookup case),
- skips the pagination count (an opt-out is an opt-out),
- **never** skips the `count_only` count — that result IS the count,
- changes nothing on a miss, which never counted,
- participates in `RunViewParams.Equals` so cached-params comparison stays exact.

The check lives in `GenericDatabaseProvider.InternalRunView`, which `SQLServerDataProvider` inherits — one change covers both dialects.

Adopted where the measurement was taken: the integration engine's two per-record match lookups (`FindByKeyFields`' single-row lookup and `FindByExternalId`). Every other caller is untouched.

## Tests

5 added, driving the real `InternalRunView` through a concrete provider with recorded `ExecuteSQL`: default MaxRows:1 hit still counts (and reports the true total), the flag skips it and reports the returned rows, a miss never counted either way, `count_only` ignores the flag, and a full multi-row page still counts by default. Suites green — GenericDatabaseProvider 968, MJCore 2143, integration engine 700.


MS-BC:
Parked. Marked draft so it does not read as awaiting review.

The saving is real — up to two extra sequential round trips per record, from the record-map lookup and the key-field match, each a `MaxRows: 1` read whose hit IS a full page. But it widens `RunViewParams`, a core public API, to skip a count, and it leaves a footgun: any caller can set `SkipTotalRowCount` and then read `TotalRowCount`, getting the row count instead of the total, with nothing stopping them.

Worth revisiting after #4063 lands — batching moves the bottleneck off round trips, which weakens the case for this considerably.

PR #4027, opened 2026-08-24
feat(providers): BulkCreate — set-based inserts as a provider capability, with the sync engine's opt-in

MS-BC:
## The write side of sync throughput

The fetch side is already covered (#4022–#4025, #3895). But a first-time backfill is **write**-bound, and the write path was one stored-procedure round trip per record — on a high-latency link the trip IS the ceiling: single-digit milliseconds of SQL work behind tens of milliseconds of wire. A connector fetching thousands of rows per minute still *creates* hundreds.

## The capability

`DatabaseProviderBase.BulkCreate(entities[])`, shaped to be universally safe:

- **Base implementation = loop over `Save()`** — identical semantics (validation, sprocs, Record Changes, cache events). Every provider supports it from day one; no feature detection.
- **SQL Server override**: TDS bulk insert. **PostgreSQL override**: multi-row parameterized INSERTs, chunked under the 65,535-bind-parameter wire limit (wide tables get fewer rows per statement, never a failure). Both all-or-nothing in one transaction.
- Both **refuse** (fall back to the base loop) on ineligible sets: mixed entity types, an already-saved entity, or a missing client-side PK — a set-based insert cannot report server-assigned keys back.
- The planning halves (column selection, metadata-mirrored nullability, coercion) are **pure modules**, unit-tested without a database. Nullability mirrors entity metadata exactly — the TDS layer rejects a mismatch, and that rejection is the contract. Audit columns never ship (DB defaults stamp them); virtual/auto-increment never ship; nullable-unset columns stay out so defaults apply.
- What a set-based path deliberately does **not** do — no sproc side effects, no Record Changes rows, no save events — is why it is never used implicitly.

## The engine opt-in

`Configuration.writeMode: 'bulk'` per connection, **default off** — nothing changes for existing tenants. Eligible NEW records (Create + full mapped PK) are built through the same per-record machinery as ever (coercion, value-fit, standard integration fields, validation), then inserted as one `BulkCreate` per batch. Everything else — updates, deletes, skips, server-assigned-PK creates, and on **any** bulk failure the whole set — continues down the unchanged per-record path, where failure isolation lives. A PK collision is not a special case: the bulk write fails whole and the per-record upsert resolves each record.

Invariants preserved by construction: record maps ride the same `RecordMapBatch` flush; the watermark advances only after the whole batch applies; `ValueFitError` stays a per-record skip-and-warn; the rollback counter-restore re-snapshots after the bulk pass so a per-record transaction failure can't erase committed bulk counts.

## Tests

23 added: 3 core (default-loop stop-at-first-failure), 12 SQL Server planning, 5 PostgreSQL planning, 3 engine end-to-end through `RunSync()` (one `BulkCreate` call + zero individual saves; identical counts on fallback; default path untouched without the opt-in). Suites green — core 2146, SQL Server 126, PostgreSQL 199, engine 703.


MS-BC:
Closing in favour of batching the write path rather than bypassing it.

`BulkCreate`'s set-based override reaches its speed by not calling `spCreate` — the header states the cost: no stored-procedure side effects, no Record Changes rows, no per-entity save events (and so no cache invalidation, which is the premise `TrustLocalCacheCompletely` is justified on, ~900 lines above in the same file).

Measured against a real SQL Server 2022 with the generated sproc AND the Record Change writes running on every row, batching `TransactionGroup`'s submit instead — concatenating the statements it already renders, which `RenderSaveCallBinding` already emits with uuid-suffixed variables *specifically* to keep batched saves collision-free — reaches 131,713 rows/min at group=100 against 211/min for the per-statement submit, with identical SQL in both arms. That clears the throughput this was reaching for while keeping every semantic `BulkCreate` gives up.

What survives from this PR and should be carried into that work: the eligibility partitioning in `ApplyCreatesBulk`, the per-connection `writeMode` opt-in that fails closed, the `Mechanism` reporting, and the fall-back-whole-on-failure contract.

PR #4028, opened 2026-08-24
fix(core): ProviderBase re-wires when the event bus is replaced

ct-bc:
## Summary

Follow-up hardening to #3831, which fixed a Critical per-request `ProviderBase` event-bus subscription leak by moving the wiring guard from instance scope to a process-wide static boolean. That was the right fix for the leak, but the boolean records *that* we subscribed, not *what to* — and the thing it guards is a reference to one specific `Subject`.

`MJGlobal.GetEventListener(false)` returns `MJGlobal._events$`, and that field is replaced in two places: `MJGlobal.Reset()`, and `resetMJSingletons()` in the UnitTesting package, which deletes the singleton from the global object store so the next `MJGlobal.Instance` is a new object with a new bus. After either, the boolean still reads `true`, so the fan-out stays attached to the orphaned Subject for the rest of the process. Providers keep registering and nothing reaches them — RunView linger entries silently stop being invalidated. No error, just stale reads inside the 5s window after a save.

The previous per-instance guard self-healed here by accident: MJServer mints a fresh provider per request, so the next one re-subscribed to the live bus. Promoting the guard to the class removed that.

Three changes, all in `packages/MJCore/src/generic/providerBase.ts`:

- **Key the guard on the bus's identity, not a boolean.** `_wiredBus` holds the `Observable` and is compared by reference, so a swap is detected exactly.
- **Retain the `Subscription` and unsubscribe the stale one on swap.** Otherwise each swap strands a subscriber on the dead Subject — a small instance of the leak #3831 fixed.
- **Run the static check above the per-instance guard.** Previously it was only reachable inside the `_lingerInvalidationWired` early-return, so healing depended on a brand-new provider arriving. Short-lived server providers self-heal that way; a long-lived one (Explorer's client provider, MJServer's system provider) never would.

**Severity: latent, not live.** No production code calls `Reset()` — only test helpers — and Vitest's default `isolate: true` keeps the latch from leaking across test files. The reason to fix it now is that `plans/MEMORY_LEAK_AUDIT.md` lists `ensureStaticInvalidationSubscription()` as the reference implementation for this pattern, so the boolean-latch shape is what future copies will inherit.

## Test plan

- [x] Two new regression tests in `providerBase.invalidationSubscriptionScaling.test.ts`, one per half of the fix:
  - A provider minted **after** `MJGlobal.Reset()` still gets its lingered entry dropped (covers the identity latch).
  - A provider registered **before** the reset also re-wires (covers the hoist). Uses a distinct dedup key, since `ensureInflightViewInvalidation()` is only reached on the fresh-execution path in `RunViewsUncoalesced` — a linger hit returns before it.
- [x] Verified each test fails without its corresponding change: reverting the hoist fails test 2 only; reducing the latch to one-shot fails both. Neither is decorative.
- [x] Both pre-existing tests in that file, and all 6 in `providerBase.lingerInvalidation.test.ts`, pass unchanged.
- [x] Full `@memberjunction/core` suite: 143 files / 2145 tests passing.
- [x] Full monorepo build (`pnpm run build`): 275/275 tasks successful, 0 TypeScript errors.
- [x] Changeset added (`@memberjunction/core: patch`).
- [ ] Deterministic integration tier not run — requires a live database. No migration, metadata, or schema impact; the change is confined to `ProviderBase`'s private write-invalidation wiring, which the unit tests exercise directly.

## Deliberately not included

- **Pruning dead `WeakRef`s from `_invalidationTargets` outside the fan-out loop.** A write-free workload accumulates ~100 bytes per provider. `FinalizationRegistry` is the idiomatic fix but is best-effort by spec and effectively untestable without `--expose-gc` — shipping an unverifiable guard cuts against the point of this PR. The residual self-corrects completely on the first write event. Happy to add it separately if preferred.
- **Updating the Recommended Remediation Patterns bullet in `plans/MEMORY_LEAK_AUDIT.md`** to say the latch keys on the bus reference. Left out to keep the diff tight, but worth doing so the pattern isn't copied in its original form.

### DIFF
--- a/.changeset/eager-kings-run.md
+++ b/.changeset/eager-kings-run.md
@@ -2,4 +2,4 @@
 "@memberjunction/core": patch
 ---
 
-Patch for eventbus rewire
+`ProviderBase`'s write-invalidation fan-out now re-subscribes when the `MJGlobal` event bus is replaced. Its one-time wiring guard was a boolean, which cannot detect that `MJGlobal.Reset()` (or clearing the singleton from the global object store) has swapped `_events$` for a fresh `Subject` — leaving the fan-out attached to the discarded bus for the rest of the process, so `RunView` dedup/linger entries silently stopped being invalidated after a save, with no error. The guard now keys on the bus reference, unsubscribes the stale subscription when the bus changes, and is re-checked on every call so already-registered long-lived providers re-wire as well.
--- a/.changeset/eager-kings-run.md
+++ b/.changeset/eager-kings-run.md
@@ -1,0 +1,5 @@
+---
+"@memberjunction/core": patch
+---
+
+Patch for eventbus rewire
--- a/packages/MJCore/src/__tests__/providerBase.invalidationSubscriptionScaling.test.ts
+++ b/packages/MJCore/src/__tests__/providerBase.invalidationSubscriptionScaling.test.ts
@@ -168,4 +168,74 @@ describe('ProviderBase write-invalidation event-bus subscription scaling', () =>
         // so all three re-executed instead of replaying a stale lingered result.
         expect(refetchCount).toBe(3);
     });
+
+    /**
+     * The process-wide subscription is guarded by the identity of the bus it attached to, NOT by a
+     * one-shot boolean. `MJGlobal.Reset()` replaces `_eventsSubject` — and therefore the `_events$`
+     * observable `GetEventListener()` hands back — with a brand-new Subject; `resetMJSingletons()`
+     * has the same effect by deleting the singleton out of the global object store. Under a boolean
+     * latch the fan-out stays attached to the orphaned Subject for the rest of the process, and the
+     * failure is silent: no error, just linger entries that quietly stop being invalidated.
+     */
+    it('re-wires to a replaced event bus — a provider minted after MJGlobal.Reset() still gets invalidated', async () => {
+        const user = makeUser();
+
+        // Wire the fan-out against the CURRENT bus.
+        await makeAndTouchProvider(user);
+
+        // Swap the bus out from under it.
+        MJGlobal.Instance.Reset();
+
+        const provider = await makeAndTouchProvider(user);
+
+        let refetchCount = 0;
+        vi.spyOn(provider as never, 'InternalRunViews').mockImplementation((async () => {
+            refetchCount++;
+            return [makeRunViewResult([{ ID: '1', Name: 'A' }])];
+        }) as never);
+        await provider.RunViews([{ EntityName: 'Customers', ResultType: 'simple' }], user);
+        refetchCount = 0;
+
+        // Raised on the NEW bus. With a boolean latch the fan-out is still listening to the
+        // discarded Subject, so this never reaches the provider and the lingered entry survives.
+        raiseEntityEvent('save', 'Customers');
+
+        await provider.RunViews([{ EntityName: 'Customers', ResultType: 'simple' }], user);
+
+        expect(refetchCount).toBe(1);
+    });
+
+    /**
+     * Companion to the test above, covering the OTHER half of the fix: the static bus check runs on
+     * every call, above the per-instance `_lingerInvalidationWired` guard. Without that hoist, only
+     * a brand-new provider can re-wire — so short-lived server providers self-heal on the next
+     * request, but a long-lived one (Explorer's client provider, MJServer's system provider) stays
+     * bound to the dead Subject indefinitely.
+     */
+    it('re-wires an ALREADY-registered provider — the static check runs on every call, not just the first', async () => {
+        const user = makeUser();
+
+        // Registers BEFORE the swap, so its `_lingerInvalidationWired` flag is already true.
+        const provider = await makeAndTouchProvider(user);
+
+        MJGlobal.Instance.Reset();
+
+        let refetchCount = 0;
+        vi.spyOn(provider as never, 'InternalRunViews').mockImplementation((async () => {
+            refetchCount++;
+            return [makeRunViewResult([{ ID: '1', Name: 'A' }])];
+        }) as never);
+
+        // A DIFFERENT dedup key, so this is a fresh execution rather than a linger hit —
+        // ensureInflightViewInvalidation() is only reached on the fresh-execution path in
+        // RunViewsUncoalesced, which is where the re-check has to happen.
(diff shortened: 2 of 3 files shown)


.changeset/eager-kings-run.md
```diff
@@ -0,0 +1,5 @@
+---
+"@memberjunction/core": patch
+---
+
+Patch for eventbus rewire
```
rkihm-BC: **Requested before merge.** This is the whole changeset — and it's what becomes the changelog entry for this fix.

Your PR description has everything it needs: the boolean records *that* we subscribed, not *what to*; the two paths that replace the bus (`MJGlobal.Reset()` and `resetMJSingletons()`); and the note that the per-instance guard self-healed by accident, which promoting it to the class removed. Something like:

```markdown
`ProviderBase` re-wires its write-invalidation fan-out when the event bus is replaced.

The process-wide subscription guard added in #3831 was a boolean, which records *that* we
subscribed rather than *what to* — and the thing it guards is a reference to one specific
`Subject`. Both `MJGlobal.Reset()` and `resetMJSingletons()` replace that Subject, after which
the boolean still reads true and the fan-out stays attached to the orphaned bus for the rest of
the process. Providers keep registering and nothing reaches them: RunView linger entries
silently stop being invalidated, with no error — just stale reads inside the 5s window after a
save. The guard now keys on the bus's identity, retains the `Subscription` so a swap can drop
the stale subscriber, and re-checks above the per-instance guard so long-lived providers
(Explorer's client provider, MJServer's system provider) re-wire too rather than only
short-lived ones self-healing on the next request.
```

The bump level is correct — this is purely about the text.

rkihm-BC (commented):
Reviewed the full diff and verified every load-bearing claim against the platform code. **One ask before merge** (a real changeset, below); the code itself I'd merge as-is.

## What's good

**The core mechanism is correct, and I checked the assumption the whole fix rests on.** `GetEventListener(false)` returns `this._events$` *directly* — a stored field, not a fresh `.asObservable()` wrapper per call. That matters enormously: had it minted a new Observable each call, the identity check would never match and this would unsubscribe/resubscribe on *every* `RunViews`. It doesn't. And `Reset()` genuinely reassigns both `_eventsSubject` and `_events$` (`Global.ts:69`/`73`), so a swap produces a genuinely different reference. Identity comparison detects the swap exactly and never fires spuriously.

**Both swap paths are real.** `resetMJSingletons()` deletes every `___SINGLETON__*` key from the global object store, so the next `MJGlobal.Instance` is a new object with a new bus — the second path described, confirmed.

**The ordering detail is right.** `_wiredBus` is assigned *before* `subscribe()`, so a re-entrant call short-circuits on the identity check rather than double-wiring. The comment says so; the code does so.

**Two subtleties handled correctly that would have been easy to get wrong:**
- The per-instance `_lingerInvalidationWired` guard is *retained*, not removed — it still prevents the same provider being added to `_invalidationTargets` repeatedly, which would accumulate duplicate `WeakRef`s to one object.
- `_invalidationTargets` is *not* cleared on a bus swap. Correct: providers registered before the swap are still live and should still receive events.

**The severity claim checks out.** I grepped every `Reset()` call site — `MJGlobal`'s own tests and `packages/UnitTesting/src/singleton-reset.ts`. Zero production callers, exactly as stated.

**The audit-doc citation is accurate.** My first grep came up empty and I nearly flagged it — my working tree's copy was stale. `origin/next` line 1289 reads: *"See `ProviderBase.ensureStaticInvalidationSubscription()` for the reference implementation."* So the "future copies will inherit this shape" motivation is genuine and precisely sourced, not a rationalization.

**The pattern isn't replicated elsewhere.** I checked the other `GetEventListener(false).subscribe` latches — `ExternalDataSourceRouter.cacheInvalidationSubscribed` and `SimpleVectorServiceProvider.subscribedToBaseEntityEvents` are both **instance** fields, not statics, so they self-heal the way `ProviderBase` did before #3831. This PR fixes the only static-latch instance; no follow-up sweep needed.

**The tests are well-targeted.** Each covers one half of the fix, and the distinct-dedup-key subtlety in test 2 is real — `ensureInflightViewInvalidation()` is only reached on the fresh-execution path, so a linger hit would return before it and the test would pass vacuously. Verifying each test fails without its corresponding change is the right discipline.

**One timing note worth recording:** this moves a `MJGlobal.Instance` call onto the per-`RunViews` path, which routes through `GetGlobalObjectStore()`. Until this morning that threw and caught a `ReferenceError` on every call in Node. #3892 memoised it (merged 04:49), and this branch merged `next` at 15:03, so the fix is included and the added cost is a memoised return plus a reference compare. Worth knowing the ordering worked out rather than assuming it.

## Requested before merge — a real changeset

`.changeset/eager-kings-run.md` reads, in full:

> Patch for eventbus rewire

Every other changeset in this repo is a descriptive paragraph naming the defect, the mechanism, and the fix, because that text is what becomes the changelog. The PR description here is genuinely excellent — the boolean-records-*that*-not-*what-to* framing, the two swap paths, the note that the per-instance guard self-healed by accident and promoting it to the class removed that. Almost none of it survived into the artifact that actually ships.

The bump level (`patch` on `@memberjunction/core`) is correct — no migration, no metadata — so this is purely about content. Largely a copy-paste from the body above.

## Smaller notes

**The "stranded subscriber" framing is a touch stronger than the mechanics support.** After `Reset()`, nothing holds the old Subject — `MJGlobal` replaced both fields, and before this PR nothing retained the `Subscription` either. So the old Subject, its observer list, and the fan-out closure form an unreachable island and collect together; it isn't retention in the GC sense. The `unsubscribe()` is still worth having — it stops the fan-out firing if some *other* holder captured the old bus reference and raises on it — just as defensive correctness rather than leak prevention. The code is right either way; only the justification is imprecise.

Following from that, nothing asserts the unsubscribe path (that N swaps leave one live subscriber, not N). Given the above, that's low value — noting it only for completeness.

## On the deferrals

Both are correctly characterized, and I wouldn't push on either. The audit-doc bullet remains *accurate* after this change ("subscribe once at the class/static level and register live instances via `WeakRef`"), just less specific than it could be — so it's a genuine nice-to-have rather than a correction. And declining to ship a `FinalizationRegistry` guard that can't be verified without `--expose-gc`, in a PR whose whole point is verifiable correctness, is the right call for the stated reason.

## Bottom line

Small, precisely reasoned, and the one assumption that could have sunk it — reference stability of `GetEventListener` — holds. Just the changeset.

rkihm-BC (approved):
LGTM

PR #4032, opened 2026-08-24
Thread the skill principal through Scoped Search, and judge it

cmbrock-BC:
`fix/scoped-search-threads-skill-principal` → `next` · 12 files

## Summary

`ScopeDimensionResolver` binds `Principals.SkillID` into a dimension's expansion query, and
`principalsFrom()` sources it from `SearchParams.AISkillID`. `SearchParams` declares that field
(`search.types.ts:120`) and the cache key includes it — but the **`Scoped Search` action never set
it**. The string "skill" did not appear in that file. So the slot existed, was typed, and was
tested, and no caller could reach it: a scope whose bound depends on the active skill resolved
`SkillID` as null forever.

This PR threads the skill through — and judges it, because threading the id alone would have been
a grant for the asking.

## Where the judging lives, and why

Both principals are judged in **`SearchScopePermissionResolver`**, not in the action — so every
caller inherits the same policy: the `Scoped Search` action, the `SearchKnowledge` and
`StreamScopedSearch` mutations, and the `SearchScopes` listing query.

Two checks, split because they answer different questions:

- **`skillIsActivatable()` runs wherever a skill is NAMED** (step 1e, before the grant steps). A
  skill is only ever supplied to steer: `AISkillID` binds into `Principals.SkillID`, and for a
  `restricts: true` dimension the expansion query's output *is* the enforced bound. Judging it only
  where it grants would let a user holding their own `DirectGrant` name any skill and widen with
  it. The check asks `GetSkillsForAgent(agent, user)` — agent-accepted ∩ agent-granted ∩ Active ∩
  user-runnable, the same call `BaseAgent.preActivateRequestedSkills` gates real activation on.
- **`agentIsWieldable()` runs at the WIDENING fallbacks** (the agent's `'All'` arm, and the
  skill's `'All'` arm — a skill widens through the agent it would activate on, so naming a skill
  must not buy access to an agent the caller may not run). Deliberately NOT at the point the id is
  supplied: `agent-pre-execution-rag` threads `AIAgentID` purely so `SearchExecutionLog` can
  attribute the search, and gating supply turns an analytics field into a retrieval outage.

Both permission models are open by default — no rows means anyone may run — so an unchecked
caller-named id carrying `SearchScopeAccess='All'` would be a scope grant for the asking. A
widening fallback now requires the principal **positively confirmed**: an agent that is merely
absent from the AIEngine metadata cache is distinguished in the *message* ("a metadata-load
problem, not a denial") but never in the *outcome*.

`SearchScopeAccess='None'` is judged twice on purpose: structurally in the action (so a missing
scope cannot bypass it) and in the resolver (so the GraphQL paths get it too).

## Behaviour changes

1. **`AgentUnscopedAll` is gated on wieldability.** An install where an agent is
   `SearchScopeAccess='All'` and the user holds no direct/role grant previously got
   `Allowed: Search`; it now also requires the agent to be cache-resident and runnable by that
   user. Reaches the action, `SearchKnowledge`, `StreamScopedSearch`, and the `SearchScopes`
   listing.
2. **A skill supplied with no agent is refused.** At base, step 4b granted `SkillUnscopedAll`
   with no agent at all — an agent-free skill id was a standalone grant. Now a skill is judged
   relative to the agent it would activate on, and none supplied means refused.
3. **The tenant reaches the permission decision.** The action and both search mutations now pass
   `PrimaryScopeRecordID` (the `SearchScopes` listing takes no `searchContext`, so no tenant exists
   to pass there — noted in code). Previously every tenant-scoped `SearchScopePermission` row was
   discarded before the decision — including a tenant-scoped `PermissionLevel='None'`, an explicit
   per-tenant admin deny, which evaporated whenever the user held any NULL-tenant role grant.
   Verdicts can change in both directions on installs with tenant-scoped rows.
4. **Denial messages no longer echo principal names.** `verdict.Reason` names the agent/skill, so
   returning it verbatim was a name oracle over the catalogues for anyone who could guess ids.
   Callers now get the ids they sent plus `verdict.Source` (a fixed enum); the audit row and the
   log keep the full reason. This is an API-visible format change on pre-existing deny paths in
   the action and both GraphQL resolvers.
5. **A supplied-but-unloadable `agentID` is refused** in the GraphQL resolvers instead of
   silently degrading to "no principal" — which previously skipped the agent rules entirely and
   handed the raw client string onward. The `SearchScopes` listing applies the same rule (an
   unloadable agent yields no scopes rather than the un-gated list). The action already refused this with `INVALID_PARAM`;
   `ExplainScope` gets the same refusal, so preview and search agree.
6. **`ExtraFilter` interpolations in the resolver now use `EscapeSQLString`** — `scopeIDs` arrive
   from a client-supplied GraphQL arg and reached a filter string raw.
7. **New `SearchScopePermissionSource` member `PrincipalNotActivatable`**, classified as
   `ACCESS_DENIED` (principal-side) by the action, consistent with the other principal denials.
8. On the explain path, a principal refused for a principal-side reason is no longer bound into
   dimension resolution; a user-side refusal (no grant) still binds them deliberately, so a
   required dimension doesn't false-fail. Encoded as a total `Record<Source, boolean>` so the
   compiler forces a decision when a Source is added.

Omit `AISkillID` and no skill principal exists — no caller gains a skill it did not ask for.

## Tests

35 added across four files: every fallback × wieldability combination including the laundering
case and both cache postures; the tenant threading (action + resolver); the unloadable-agent
refusal; the name-oracle redaction (asserts the name is absent, the ids and Source present, and
the audit row still carries the full reason); the explain-path parity tests. The security-relevant
guards are mutation-verified — deleting each one reddens the exact test written for it. An
exhaustive matrix over agent access × skill access × cache state × runnability × activatability ×
grant state shows **zero combinations where this branch grants more than base**; every difference
is a `PrincipalNotActivatable` tightening.

## Disclosed and deliberately out of scope

- A metadata-load failure that reaches the caller is bucketed `ACCESS_DENIED` even though its
  Reason says "a metadata-load problem, not a denial" — a distinct result code for it is a wider
  API change than this PR should carry.
- `ExplainScope` has no production caller; the explain-path principal-drop is verified by
  inspection and unit tests only.
- On containment, stated accurately: an expansion query is server-authored SQL, but MJ renders
  query parameters through Nunjucks with `autoescape: false` and escaping is opt-in
  (`| sqlString`). The gate decides *whether* a principal may steer the bound; the query author
  still owns making the id safe to interpolate — write the predicate so a malformed value narrows
  (e.g. `TRY_CONVERT(uniqueidentifier, …)` degrades to NULL and an empty bound).


rkihm-BC (approved):
Approving. Careful authorization work, and the reasoning behind the awkward decisions is written down rather than assumed. Nothing to change.

## What's good

**The underlying bug is precisely evidenced.** `SearchParams.AISkillID` was declared, included in the cache key, and asserted on by `ScopeExplanation.test.ts` — but the `Scoped Search` action never set it. "The string *skill* did not appear in that file" is the kind of detail that makes a bug report checkable rather than assertable. The slot existed, was typed, was tested, and no caller could reach it.

**Threading the id without judging it would have been a grant for the asking**, and you say so outright. Both permission models are open by default — no rows means anyone may run — so an unchecked caller-named id carrying `SearchScopeAccess='All'` would have been a scope grant. Recognising that *before* shipping the thread is the whole value here; the fix for the reported bug is a few lines.

**The NAMED-vs-WIDENS split is the hard part, and it's right.** A skill is judged wherever it's *named*, because `AISkillID` binds into `Principals.SkillID` and for a `restricts: true` dimension the expansion query's output *is* the enforced bound — judging it only where it grants would let a user holding their own `DirectGrant` name any skill and widen with it. An agent is judged only at the *widening* fallbacks, because `agent-pre-execution-rag` threads `AIAgentID` purely so `SearchExecutionLog` can attribute the search, and gating supply turns an analytics field into a retrieval outage. Two principals, two different questions, two different placements, each with its reasoning in the code.

**The `GetSkillsForAgent` caveat is the subtle one and it's handled.** Vouching for a skill says nothing about whether the caller may run the agent it would activate on — `AISkillPermissionHelper` filters SKILL rights, never `AIAgentPermission` — so both `'All'` arms consult `agentIsWieldable()`, the skill's included. That's the kind of gap that would have looked closed to a reader who only checked the skill path.

**Fail-closed on unavailability, and correctly ordered.** Both gates call `await AIEngine.Instance.Config(false, user)` *before* testing the cache, so a cold cache triggers a real load attempt rather than a spurious denial. If it's still empty, `kind: 'unavailable'` is distinguished in the *message* — "a metadata-load problem, not a denial" — but produces the same denial *outcome*. The changeset's parenthetical about the earlier revision that let it through is worth having: the reasoning that "a cache blip shouldn't refuse a user whose own grant covered the scope" is impossible by construction, since a direct or role grant returns before any fallback is reached, so what it actually did was grant `Search` to users with no grant at all. Catching that yourselves is the process working.

**The judging moved to the right layer.** The commit trail reads as a genuine architectural correction rather than churn: *"A skill veto that only lives in the resolver is not a veto"* → *"Move principal validation onto the engine, where every caller passes through."* All four callers — the action, `SearchKnowledge`, `StreamScopedSearch`, and the `SearchScopes` listing — now inherit one policy instead of four copies agreeing.

**`SearchScopeAccess='None'` judged twice is deliberate and worth keeping** — structurally in the action so a missing scope can't bypass it, and in the resolver so the GraphQL paths get it too. It reads as duplication and isn't; worth leaving alone.

## On the behaviour changes

I was going to ask for release-note visibility on the two access-tightenings, then read the changeset properly. It already documents all three, with before/after framing, in a paragraph headed **"Three behaviour changes to note"** — including which callers each reaches, and that `ExplainScope({ AISkillID })` with no `AIAgentID` (a preview UI letting a skill be picked before an agent) is the only reachable path for the skill-without-agent refusal. That's more precise than what I'd have asked for. Nothing to add.

One thing worth doing outside this PR: `AgentUnscopedAll` is the tightening most likely to be quietly load-bearing on an existing install — an agent marked `SearchScopeAccess='All'` where users hold no grant of their own. The changeset describes the change accurately; what it can't tell you is whether any deployment actually leans on it. Worth a look at production `SearchScope` permission rows before this ships, since the failure mode is a search returning fewer results rather than an error anyone files.

## Nothing missing

Three substantial test files (+281 on the permission resolver, +231 on the action, +102 on scope explanation), a guide update, and the action metadata. Changeset `minor` on both packages is correct — the branch touches `metadata/actions/`. The honesty in the closing paragraph about containment — that MJ renders query parameters through Nunjucks with `autoescape: false`, so MJ doesn't itself guarantee a named skill can't widen or inject, and the permission gate is what MJ *does* enforce — is the right way to state a boundary rather than overclaim one.

PR #4033, opened 2026-08-24
feat(global): add EscapeSQLString and adopt it as the one canonical SQL literal escaper

AN-BC:
## Summary

Adds `EscapeSQLString` to `@memberjunction/global` and adopts it across core packages, replacing a dozen ad-hoc `.replace(/'/g, "''")` implementations with one audited function.

1. **The utility** — `EscapeSQLString` doubles single quotes per the ANSI SQL standard (supported by both SQL Server and PostgreSQL), strips null bytes (`\0`), and maps `null`/`undefined` to `''`.
2. **Adoption** — call sites in `core`, `server`, `schema-engine`, `search-engine`, `version-history`, `ai-agents`, `ai-bridge-server`, `database-designer-core`, `database-designer-actions`, `ng-dashboards`, `ng-conversations`, `open-app-engine` and `integration-schema-builder` now call `EscapeSQLString` directly.
3. **Deprecated aliases** — the four exported package-local names (`EscapeSqlString` in `schema-engine` and `open-app-engine`, `escapeSqlLiteral` in `database-designer-core`, `escapeSqlString` in `version-history`) are kept as `@deprecated` re-exports so external callers do not break. Removable at the next major.
4. **Guidance** — `.claude/rules/data-access.md` documents the rule and, importantly, its limits.
5. **Tests + changeset** — unit coverage in `packages/MJGlobal/src/__tests__/util.test.ts`; patch changeset for `@memberjunction/global`.

## What `EscapeSQLString` deliberately does not cover

Quote doubling is not sufficient in three cases. Both the rule and the function's own JSDoc name them, so callers reach for the right tool instead of assuming this one is enough:

| Context | Why quote doubling is not enough | Use instead |
|---|---|---|
| `LIKE` patterns | `%`, `_` and `[` stay live as wildcards — a user searching for `%` still matches every row | `escapeLikeValue()` (`MJCore/src/generic/runQuerySQLFilterImplementations.ts`) or `GenericDatabaseProvider.escapeLikeTerm()`, paired with `ESCAPE '\'` |
| Identifiers — table, column, schema names | Identifiers are quoted with brackets or double quotes, never single quotes | `ValidateIdentifier()` from `@memberjunction/schema-engine` |
| A value that must not be missing | `EscapeSQLString(undefined)` returns `''`, so the predicate silently degrades to `Field = ''` instead of failing | Validate before building the filter |

The existing `LIKE` call sites are **left semantically unchanged on purpose** — tightening them would change what users' searches match, which belongs in its own change. `sqlContains()` and `conversations/search.service.ts` carry a note recording that the wildcard behaviour is deliberate.

## Notes for reviewers

**Every reference is inside a function body, never a module-level `const`.** This is load-bearing, not style. `const alias = EscapeSQLString` at module scope is an eager read of the imported binding at import time, and vitest's mocked-module proxy throws on property access — so any suite partial-mocking `@memberjunction/global` failed at module load with `No "EscapeSQLString" export is defined on the mock`, which took `@memberjunction/actions#test` red. Deferring the read to call time fixes it; the deprecated aliases are lazy wrappers for the same reason.

**`ScopeValueEscaper.EscapeSqlLiteral` is byte-for-byte equivalent** to its previous implementation despite reordering the null-byte strip ahead of quote doubling — verified across 200k fuzz cases.

## Verification

- 183/183 turbo test tasks across the touched packages and their dependents, including `@memberjunction/actions`
- 181 packages plus both Angular packages (`ngc`) build clean
- `check:esm` clean; `check:claude-md` shows no new failures against `next`


AN-BC (commented):
lgtm

AN-BC (commented):
lgtm

AN-BC (commented):
lgtm

PR #4037, opened 2026-08-24
security: refuse ad-hoc SQL for scope-limited sessions, escape ResolverBase's own filters — and unbreak the integration tier

AN-BC:
> Carries the two live items from #3827, which is closed as superseded. Its third item — the SSRF guard — already shipped in #4019, and its escaping fix is rewritten here on top of #4033's `EscapeSQLString`.
>
> **Also unbreaks `next`.** The deterministic integration tier has been red since #4030 for a reason unrelated to this PR; commit 2 fixes it. See [§3](#3-unbreaking-the-deterministic-integration-tier).

## Why this exists rather than #3827

#3827 bundled three unrelated security fixes. One of them has since landed, and one was written against a convention this repo has since adopted, so re-cutting the remainder is cheaper to review than rebasing it:

| #3827 item | Status |
|---|---|
| SSRF guard for the URL-fetching web actions | **Shipped** in #4019 (superseding #4011) — the guard now lives in `@memberjunction/network-utils` and covers six actions, not the three #3827 touched. Its `CoreActions/src/custom/web/ssrf-guard.ts` would re-introduce the duplicate that #4011's review rejected. |
| Scope-limit `ExecuteAdhocQuery` | **Here.** Still unfixed on `next`. |
| Escape the filters `ResolverBase` builds | **Here**, rewritten. #3827 hand-rolled `.replace(/'/g, "''")`, which #4033 has since made a documented anti-pattern in `.claude/rules/data-access.md`. This uses `EscapeSQLString`. |

## 1. Ad-hoc SQL is refused for scope-limited sessions

`AdhocQueryResolver.ExecuteAdhocQuery` runs a raw `SELECT` on the read-only pool. It does not go through `RunView`, entity permissions, or row-level security — and that is precisely where a magic-link session's confinement lives: as `{{ScopeResourceID}}` / `{{ScopeResourceType}}` RLS tokens that `RowLevelSecurityFilterInfo.MarkupFilterText` substitutes on the entity-read path. On this path they are never substituted, so the scope does not exist at all.

Two principals are affected, and both authenticate successfully today:

- an **anonymous magic-link guest**, which authenticates as the *shared* Anonymous principal — nothing in its role grant distinguishes one guest from another;
- a **resource-scoped magic-link session**, which holds a normal role pinned to one shared resource.

Either could read the entire database from this resolver, outside its granted scope.

Both are now refused, before a data source is acquired. Ordinary authenticated users — the intended consumers, via `GraphQLDataProvider` — are unaffected. This is deliberately the minimal guard rather than `@RequireSystemUser`, which would break the client feature.

The check is a named predicate, `IsScopeLimitedPrincipal`, exported from `@memberjunction/server`, so the next path that reaches around RLS has one place to ask. Two details in it are load-bearing:

- It tests `MagicLinkScope.ResourceID` / `.ResourceType`, **not** the object's truthiness. An empty `{}` carries no confinement, and treating it as scoped would lock out a session that is not.
- A missing principal returns `true` — an unidentifiable session is the one case where the scope certainly cannot be established, so it fails closed.

Worth a reviewer's opinion: this could instead be an `IsScopeLimited` getter on `UserInfo` in `@memberjunction/core`. I kept it in `MJServer/auth` because "refuse this principal on scope-bypassing paths" is server-side policy rather than a property of the user record — but the argument for `UserInfo` is real if more packages end up needing it.

## 2. `ResolverBase`'s own filters are escaped

Two sites built `ExtraFilter` text out of client-supplied values with no escaping:

- **`ResolverBase.findBy`** — reachable through `UserByEmail`, `FileByName`, `UserViewsByName`, `UserFavoritesByUserID` and the other by-value single-record resolvers.
- **`RunViewByNameGeneric`** — the inline `Name='{the client's ViewName}'` view lookup.

`ExtraFilter` does pass through `SQLExpressionValidator`, which rejects stacked statements, `UNION`, comments and `WAITFOR` — and #3633 fixed the literal-stripping bug that let a backslash smuggle a payload past that screen. So the residual exposure was a same-clause boolean tautology (`x' OR '1'='1`), not arbitrary SQL: medium severity, defense-in-depth. It is now closed with `EscapeSQLString`.

### The part #3827 missed: the unquoted slot

`EntityFieldInfo.NeedsQuotes` is `false` for Number and Boolean fields, so those values are interpolated into the clause **bare**. #3827 escaped only the quoted branch — but an unquoted slot has no literal to break out of, which means a string value there simply *is* SQL and quote-doubling protects nothing.

`findBy` now refuses anything that is not a genuine number or boolean in that slot (numeric and boolean *strings* are accepted, since neither can carry SQL and a client may legitimately send `"42"`). Today's callers type those args as GraphQL `Int`/`Boolean`, so they coerce safely and nothing legitimate changes — this hardens the sink instead of trusting every present and future caller's arg typing to stay correct.

### One intended behaviour change

A `null`/`undefined` value now throws instead of being interpolated. `EscapeSQLString` maps both to `''` by design — its own JSDoc flags this — so the alternative was a lookup silently becoming `Field = ''`. Previously such a call produced `Field = 'undefined'`, matched nothing, and returned an empty array; the one reachable path (`UserViewResolver.getCurrentUserID` → `UserByEmail(payload.email)` with no email claim) then dereferenced the null result and threw a `TypeError` anyway. A named error at the sink beats both.

## 3. Unbreaking the deterministic integration tier

This part is unrelated to the security work — it is here because the tier is red for everyone and this PR could not otherwise go green.

`mj sync push --dir=metadata --ci` has been failing validation on `next` since `202a2bfba` (the #4030 merge). Integration Tier on `next` was green at `b681e8606`, failed at `202a2bfba`, and is still failing at `f9eeadbd2` — this PR's base. Every branch cut from `next` since inherits it.

```
Field "Status" has invalid value "Inactive"
  Entity: MJ: AI Model Costs
  → Allowed values are: Pending, Expired, Invalid, Active
```

`93008ce3b` set two GLM-4.7/Cerebras rows to `Inactive` for the 2026-08-17 deprecation. The two entities have different value lists, and only one of those rows was legal:

| Entity | Status value list | `Inactive` legal? |
|---|---|---|
| `MJ: AI Model Vendors` | Active, Deprecated, **Inactive**, Preview | ✅ |
| `MJ: AI Model Costs` | Active, Expired, Invalid, Pending | ❌ |

That asymmetry is why the file carries 34 `Inactive` values but validation reports exactly one error — the other 33 are all vendor rows.

The cost record (`BC44A754-7CAC-4A46-A72C-C0B439D3EE07`) now reads `Status: "Expired"`, which the generated ORM defines as "no longer valid" — precisely what retiring the model from Cerebras' inference cloud means. It also gains `EndedAt: "2026-08-17"`, because the ORM documents that field as *"when this pricing expired or will expire; NULL indicates currently active pricing"* — an Expired row with a NULL `EndedAt` contradicts itself, and the retirement date is already asserted in the record's own `Comments`. The sibling vendor row keeps `Inactive`, correct for its entity.

Verified by walking every record in the file and checking each `Status` against the value list its own entity declares in the generated ORM: **0 invalid rows, down from 1**, and no other record changes.

This is what makes the branch a metadata branch, so it carries a second changeset at `minor` — metadata becomes a migration at release time via the build engineer's `mj sync push` (`.claude/rules/changesets.md`). The security fix keeps its own `patch` changeset. No marginal version cost: `.changeset/ai-model-research-2026-08-24.md` already declares `minor` for the very edits being repaired here.

## Verification

- **`packages/MJServer`: 1097 passed, 56 skipped, 0 failed** — 16 of those tests are new.
- `tsc --noEmit` clean on MJServer; full `pnpm run build` green across all 275 packages.
- Gates: `check:changeset` (now reporting *"Branch touches metadata — minor is required"*, 2 changesets correct), `check:standards`, and `check:codegen-tail` all pass.
- CI on commit 1: unit tests, adopted standards, dependency check and migration check all green; only the integration tier was red, for the base-branch reason commit 2 fixes.

New tests:

- **`AdhocQueryResolver.scopeGuard.test.ts`** — six cases on the predicate (plain user, anonymous guest, both scope shapes, empty scope, missing principal), plus an assertion that the guard runs *before* `GetReadOnlyDataSource`: authorization gates the work, it does not race it. That ordering check is a source-shape test, matching the hybrid strategy the existing `AdhocQueryResolver.bugs.test.ts` uses — instantiating this resolver needs a very heavy mock graph.
- **`ResolverBase.filterEscaping.test.ts`** — real `findBy` and `RunViewByNameGeneric` calls against a fake provider that captures the `ExtraFilter` they build, using the subclass-probe pattern from `ResolverBase.transportMapping.test.ts`. Covers quote doubling, the tautology payload rendered inert, multi-param `AND`, the unquoted numeric slot, and both refusal paths.

## Out of scope, found while here

Neither is a security issue; both are pre-existing and left alone rather than folded in.

- **`findBy` emits `IsDefault = true` for boolean fields.** SQL Server has no boolean literal, so `DefaultViewByUserAndEntity` and `CurrentUserDefaultViewByEntityID` look latently broken. The fix is dialect-dependent (`1`/`0` for SQL Server, `true`/`false` for PostgreSQL), which makes it a provider concern rather than a one-line change. This PR preserves the existing rendering exactly.
- **`UserFavoriteResolver.UserFavoriteSearchByParams`** declares `@Arg('params', () => Int)` for an object type.

Also note that #3827's recommended follow-ups — CORS default, GraphQL depth/complexity limiting, MCP OAuth scope minting, DNS-rebinding pinning — are **not** carried here and remain open. #4030's report separately flags four AI-model items for human review; only the one breaking CI is touched here.

https://claude.ai/code/session_01KF7XUStRQLBuDbd5CLWngS


AN-BC (commented):
LGTM

PR #4038, opened 2026-08-25
fix(integration): sample the source when a declared object has no primary key

MS-BC:
The read-path gate in the introspect stage tests **"declared with zero fields"** — the wrong property for the decision it guards:

```ts
if (existing && existing.Fields.length === 0) {   // → DiscoverFieldsViaFetch
```

`DiscoverFieldsViaFetch` is the only code that streams records and derives a **provable** key. An object declared *with* fields but *without* a key skips it entirely, so its key is never decided from data: `StagePKClassify` runs on field names alone (its `sampleRows` input is read in exactly one place and **populated in none**), and a convention match like `*_id` wins.

## Failure mode

On a parent-scoped child, the convention ladder elects the **parent's foreign key** as the child's PK. Every child row in a parent then collapses onto one record — and because a soft PK carries no unique index, nothing rejects the collision. It surfaces thousands of writes later as concurrent writers deadlocking on the shared row, with the per-record retry hitting the identical collision forever.

## Changes

1. **Gate on the question the branch exists to answer** — is the *key* unknown — rather than on whether fields are missing. Declared-with-a-PK still skips: a declared key is authoritative.
2. **Adopt only the key when fields are declared.** The field-less case may still take the discovered field set wholesale, but re-sampled widths/types must never overwrite deliberately declared ones. Logs the adopted key, stamped-count, and any nominee absent from the declared catalog.

## Verification

Live on a PostgreSQL instance: two fields-but-keyless stream objects that previously received their parent's FK as PK now sample **500 records each in ~400ms** (`stop=SAMPLE_TARGET_MET`) and derive provable single-column keys (500/500 distinct, Chao1-supported). The subsequent full sync wrote **6,006 records, 0 errors, 0 deadlocks**. Objects with declared keys and field-less objects behave exactly as before. Dialect-agnostic — no SQL in the changed path.

`tsc --noEmit` clean; changeset included (`@memberjunction/integration-engine`: patch).

## Deliberately out of scope

`SampleRowsByObject` remains a producer-less parameter, so an object whose key the read path *cannot* prove still reaches the soft-PK classifier without rows. Plumbing real samples into the classifier (and recording key *provenance* so an inferred key is re-verified rather than trusted forever) is the follow-up.


MS-BC:
Superseded by #4065, which removes this gate entirely rather than widening it.

The substance survives: #4065's primary-key merge rule keeps exactly the case this PR was written for — a declared object with no key gets its key from the data instead of from a convention match on `*_id`, and a declared key stays authoritative even when the sample nominates a different column.

Leaving this open until #4065 merges, in case a reviewer wants the narrow fix first.

MS-BC:
Closing in favour of #4065.

This widened the sampling gate from `Fields.length === 0` to `Fields.length === 0 || declaredPKCount === 0`. #4065 removes the gate entirely, which is the right shape: sampling answers three questions and a declaration can only pre-answer one of them. An object declared with fields AND a key was still never sampled here, so its undeclared columns kept arriving through the overflow path and its widths stayed whatever the catalog guessed.

Nothing is lost. #4065's primary-key merge rule keeps exactly the case this PR was written for — a declared object with no key takes its key from the data rather than from a convention match on `*_id`, and a declared key stays authoritative even when the sample nominates a different column, which is what stops a child table being keyed on its parent's foreign key.

PR #4039, opened 2026-08-25
feat(cli): make the mj CLI agent-first

AN-BC:
Makes the `mj` CLI usable by a program without changing anything for a person typing at a terminal.

MJ already had an unusually good agent-facing *architecture* in `CLICore` — a result envelope, two-tier progressive disclosure, runtime hints, stderr/stdout discipline. The problem was reach: 3 of 87 commands used it. This PR closes the gap on the four things that actually bite a caller that is a program.

## 1. Prompting follows the terminal

A command prompts when stdin **and** stdout are real terminals. When either is piped, when a CI environment variable is set, or when `TERM=dumb`, it does not — and a command that needs a value it wasn't given fails immediately naming the flag that supplies it, instead of blocking on stdin forever.

| scenario | output | prompts | reason |
|---|---|---|---|
| human, terminal, no flags | `text` | **yes** | `tty-detected` |
| agent, piped stdio | `json` | no | `no-tty` |
| CI runner, pty allocated | `text` | no | `ci` |
| emacs shell, `TERM=dumb` | `text` | no | `dumb-terminal` |
| `--no-interactive` at a terminal | `text` | no | `flag-off` |
| `--interactive` in CI | `text` | **yes** | `flag-on` |
| `MJ_CLI_INTERACTIVE=0` | `text` | no | `env-off` |

```console
$ mj sync init                     # at a terminal: prompts, exactly as before

$ mj sync init < /dev/null         # spawned by an agent
 ›   Error: An entity-directory choice is required and this run has no interactive
 ›   terminal. Pass --setup-entity=ai-prompts|other|no.
 ›   Try this: Pass --setup-entity=ai-prompts|other|no.
$ echo $?
1

$ mj sync init --setup-entity=no   # scripted: runs unattended anywhere
```

Before this, `mj sync init` had four prompts and **no escape flags at all**, and `mj install --legacy` had two dozen — both hung an agent forever with no output explaining why. Sites covered: `sync init`, `sync file-reset`, `app remove`, `install` (engine + legacy wizard), `dev workspace`, and the Open App orchestrator prompt bridge.

Details worth a reviewer's eye:
- **Both streams must be TTYs.** `mj sync init < /dev/null` and `mj sync init | tee` each break the prompt in a different half.
- **`--interactive` cannot conjure a terminal.** With no TTY it resolves to `no-tty`, so the command still fails at the exact point a value is missing rather than throwing a blanket error at startup.
- **`MJ_CLI_INTERACTIVE=0`** pins it off for a whole session — for an agent harness that shells out through a pty and would otherwise look human.
- **Error messages name the actual cause** (`no interactive terminal` / `looks like CI` / `--no-interactive was passed`) and only offer `--interactive` where it could help; suggesting it when there is no terminal at all would send the caller in a circle.

`mj sync init` gains `--setup-entity` / `--entity` / `--dir` / `--overwrite` so it is fully scriptable.

**One real bug fixed along the way:** `sync init` asked its overwrite question unconditionally, not only when `.mj-sync.json` existed. Harmless while the answer was a prompt a human could dismiss, and fatal the moment a missing answer became an error — a first-time `mj sync init` has nothing to overwrite. Now gated on the file actually existing.

## 2. Output follows the pipe

Independent of the above — a human at a terminal gets banner, spinners, and color with **zero flags**. With no explicit `--format`, a non-TTY stdout resolves to `json` and all chrome is suppressed. A caller that redirected stdout has already said it is a machine.

```console
$ mj usage | head -c 120
{"version":"1","success":true,"command":"usage","durationSeconds":0,"data":{"guidance":"Run `mj <domain> usage` before…

$ mj usage --format=text > notes.txt   # forcing human output still works
$ export MJ_CLI_FORMAT=json            # pin it for a session
```

JSON is compact when piped and pretty on a terminal. `--format=text > log.txt` no longer writes spinner escape codes into the file.

## 3. One `--format` spelling CLI-wide

Three families had grown three spellings of the same idea. All now accept the canonical `--format text|json|md`:

| family | was | now also |
|---|---|---|
| plugin commands | `--format text\|json\|md` | — |
| `mj test *` | `--format console\|json\|markdown` | `--format text\|json\|md` |
| `mj ai *` | `--output compact\|json\|table` | `--format text\|json\|md` |

Every existing value keeps working, and an explicit legacy value still beats inference, so no script changes are needed.

`-o` still means *format* under `mj ai` and *file path* under `mj test` / `querygen export` / `sql-audit`. Rebinding it would break scripts, so instead `--format` now works everywhere and `mj ai`'s `--output` help text says plainly that it's a legacy alias. Worth a follow-up decision, but not a silent breakage in this PR.

Also replaces four `as any` casts in `mj ai audit agent-run` with one typed resolution (`AuditOutputFormat` is now exported from `@memberjunction/ai-cli`).

## 4. `mj usage` covers the whole CLI

Tier 1 tells an agent *"Run `mj <domain> usage` before invoking. Do NOT guess flags."* — so a map covering **3 of 22 domains** was worse than no map, because nothing distinguishes "not listed" from "does not exist".

```
mj — command domains
Run `mj <domain> usage` before invoking. Do NOT guess flags or subcommands.

  ai             Run and inspect AI agents, prompts, and actions from the terminal. [variable]
  app            Install, upgrade, enable, and remove Open Apps. [slow]
  …
  test           Run, list, compare, and validate MJ test suites. [variable]
  translate-sql  Translate SQL between dialects. [moderate]
```

**23 domains**, and every one has a working `mj <domain> usage` page. The key design choice: entries for commands that aren't `BaseCLIPlugin` plugins are **derived from oclif's own manifest at runtime** (`lib/derived-usage.ts`), not hand-written. oclif already holds every description, flag, and example and can't fall out of sync with itself. The only hand-maintained part is the per-domain runtime budget (`lib/domain-profiles.ts`), which is what oclif genuinely doesn't know — and an unprofiled domain falls back to `variable` rather than disappearing, so adding a command can never make the map lie by omission again.

Registration is first-wins, so the three migrated plugin commands keep their richer curated metadata.

## Envelope

`MJCLIResult` gains `version` (stamped on every serialized result, so a caller can detect contract drift), and `MJCLIResultError` gains machine-readable `code` (from an enumerable `MJCLIErrorCodes` catalog) and an actionable `suggestion`:

```json
{"version":"1","success":false,"command":"sync:push","durationSeconds":1.2,
 "errors":[{"context":"metadata/entities","message":"No mj.config.cjs found.",
            "code":"E_NO_CONFIG","suggestion":"Run `mj sync init` in this directory first."}]}
```

## Compatibility

Interactive use at a terminal is unchanged — no flags needed for prompts or for pretty output.

The behavioral change is confined to automated contexts: a command that used to prompt when spawned, piped, or run in CI now fails with an actionable error instead of hanging. That is strictly better than the old outcome (an indefinite hang), but it is a change in exit behavior for anything that was relying on a prompt it could never answer.

Residual risk worth naming: an agent that drives `mj` through a **pty** looks like a human to this detection and could still hit a prompt. `MJ_CLI_INTERACTIVE=0` is the answer for those harnesses, and it's documented in the README.

## Testing

| package | tests | new |
|---|---|---|
| `@memberjunction/cli-core` | 76 pass | 55 |
| `@memberjunction/cli` | 717 pass | 41 |
| `@memberjunction/ai-cli` | 74 pass | 0 |

New coverage: interactivity detection across all seven resolution paths (TTY, half-TTY, six CI variables, `TERM=dumb`, env tri-state, flag overrides in both directions), the guarded-prompt property (a prompt is *never* invoked in any automated context, for any value shape), error-message causes and remedies, format precedence across flag/env/pipe, envelope versioning and compact-vs-pretty, oclif→usage derivation including first-wins and hidden-command skipping, legacy-format back-compat, and the prerun hook's global flag consumption.

One bug the tests caught during review: the prerun hook short-circuited its second flag strip, so `--interactive --no-interactive` together left `--interactive` in argv, where oclif's strict parser rejects it on any command that doesn't declare it. Both flags are now stripped unconditionally before either is acted on.

Two existing tests changed, both deliberately:
- `claude-commands.test.ts` asserted pretty mode shows the banner. `spawnSync` pipes stdout, so under the new contract it doesn't — updated to assert the piped contract, plus that pretty mode is still the *human report* and not JSON. The banner-on-a-terminal path is covered in `prerun-banner.test.ts`, which stubs `isTTY` without needing a pty.
- `prerun-banner.test.ts` now sets up TTY state explicitly rather than inheriting vitest's (no TTY).

**One pre-existing failure remains**, untouched: `dev-workspace.member-installs.test.ts` chmods a directory to `000` and expects it unreadable, which doesn't hold when the suite runs as uid 0. It fails identically on an unmodified tree (verified by stashing).

Also verified end to end against a built CLI: `mj usage`, `mj test usage`, `mj sync usage`, `mj sync init` (both the failing and succeeding paths), and the full detection matrix above. `check:esm`, `check:changeset`, and `mj standards check` pass.

https://claude.ai/code/session_01AATniw1oqwNzQUKmFbzTfW


AN-BC:
@AN-BC — ran an adversarial pass over my own diff. Found and fixed **three real bugs**, pushed as `53a1db4`.

Each was reproduced before fixing and re-verified after, and two now have regression tests pinning them.

---

### 1. A banner could land in machine-readable stdout

The prerun hook decided chrome by scanning argv for `--format` itself, while commands resolve their format through `ResolveOutputFormat` — which *also* reads `MJ_CLI_FORMAT`. The two views disagreed exactly where it hurts:

```
  On a TTY (so the pipe check does not save us):
  --format=json in argv                  chrome=false
  MJ_CLI_FORMAT=json in env              chrome=true   ← banner, then JSON
```

`MJ_CLI_FORMAT=json mj codegen` at a terminal printed a figlet banner and *then* the envelope. The pipe check that normally covers this doesn't fire on a TTY — so it hit precisely the **pty-based agent harness** the README tells people to configure. Both sides now go through the one resolver, which also picks up the format aliases for free.

While collapsing that branch I found a latent one next to it: the old `--json` early-return skipped `maybeLoadBootstrap` entirely. Harmless today (only light commands declare `--json`) but it would have silently skipped class registration for any heavy command that later grew one. All suppression paths now load bootstrap.

### 2. An explicit `-o compact` was silently overridden

`resolveLegacyFormat` inferred "the caller chose this" from `value !== default` — which cannot tell `mj ai agents list -o compact` from no flag at all. Piped, that returned `json` against an explicit request:

```
  piped, user typed -o compact -> json     ← wrong
  piped, defaulted to compact  -> json     ← right
```

oclif already knows the answer. The resolver now takes `metadata.flags.output.setFromDefault` as an input, with the old comparison kept only as a fallback for non-oclif callers.

### 3. The installer guard crashed instead of erroring

`engine.On('prompt', …)` neither awaits nor catches the handler's promise, so the `requireInteractive` throw I added became an **unhandled rejection** — a stack trace and a dead process, the exact opposite of the actionable error it was meant to produce. Worse, rejecting means the event is never `Resolve()`d, so the engine would sit waiting for an answer that could never arrive.

Caught at the boundary now:

```console
$ mj install --legacy            # spawned, no terminal
 ›   Error: The legacy interactive installer is required and this run has no
 ›   interactive terminal. Re-run at an interactive terminal, or drop --legacy
 ›   to use the engine installer (which accepts --config).
 ›   Try this: Re-run at an interactive terminal, or drop --legacy …
$ echo $?
1
```

---

### Two smaller cleanups

- Dropped a dead `|| !process.stdout.isTTY` in `dev/workspace` — `isInteractiveRun()` already requires both streams.
- Pointed `NON_INTERACTIVE_CODE` at `MJCLIErrorCodes.NonInteractive` instead of restating the string, so the thrown code and the catalogued one can't drift.

### Checked and deliberately left alone

- **Unused `chalk` imports** in `ai/agents/run.ts` and `ai/prompts/run.ts` — confirmed pre-existing (zero uses before this branch too), not mine to sweep here.
- **`-o` still means two things** across families. Rebinding it breaks scripts; `--format` working everywhere is the non-breaking fix. Still worth a follow-up decision.
- **The pty residual risk** is real and unfixable by detection alone — `MJ_CLI_INTERACTIVE=0` is the documented answer.

### State

| | |
|---|---|
| `@memberjunction/cli-core` | **76 pass** |
| `@memberjunction/cli` | **724 pass** (+7 regression cases) |
| `@memberjunction/ai-cli` | **74 pass** |
| changeset-bot / dependency check | green |

The one remaining failure is pre-existing and unrelated: `dev-workspace.member-installs.test.ts` chmods a directory to `000` and expects it unreadable, which doesn't hold when the suite runs as uid 0. Fails identically on an unmodified tree.

PR is currently `behind` `next` (no conflicts). Say the word and I'll merge base in, or leave it for whoever merges.

---

AN-BC:
@rkihm-BC @cadam11 can you take a peak at this and let me know if you have any suggested changes. This is intended to **not** change anything about the CLI devs encounter but to make this agent friendly which is increasingly important for our own use as well as all other devs working with MJ. Flagging this for @MS-BC and @jordanfanapour  as I know this affects their work and interests.

cadam11:
Excellent change, it should help a lot when using Claude Code (or whatever) to drive the CLI. A few things that may be worth double-checking (I could be missing some intended oclif or installer behaviour):

- Explicit legacy AI output flags such as `-o table` may be treated as defaults and ignored.
- Replacing the test-family format flag appears to remove the existing `-f` shorthand, potentially breaking scripts using it.
- `mj ai agents run --chat` may bypass the new non-interactive detection and still wait for input.
- The installer may not discover missing non-interactive inputs until after scaffolding files, rather than failing before making changes.
- A few machine-output paths may still emit text—for example AI action dry-runs, empty AI lists, and some plain oclif commands when piped.
- `sync file-reset` appears to initialize the database before checking whether `--sections` or `--all` is required.

AN-BC:
@cadam11 — thanks, this was a good catch list. Five of the six were real; all five are fixed in `cffd286`. Details below, including one that turned out to be worse than you described.

---

### 2. `-f` removed from `mj test *` — **real, and mine**

You were right, and it was worse than one command: all **seven** lost it — `run`, `list`, `history`, `compare`, `validate`, `suite`, and `regression compare` each had `char: 'f'` on base. Swapping them onto the char-less canonical flag turned `mj test run -f json` into `Nonexistent flag`. It also makes a claim in my own PR description false ("no script changes needed"), which is the part that bothers me most.

Fixed by adding a `TEST_FORMAT_FLAG` — canonical values, `char: 'f'` — used by those seven. Verified against the built oclif manifest:

```
test family -f: f,f,f,f,f,f,f
ai family  -f: undefined ×6
```

The `mj ai` family deliberately stays char-less: `--format` is brand new there, so there's no shorthand to restore, and `mj ai audit agent-run` already spends `-f` on **`--file`** — an output *file path*. (I'd assumed `--filter` until a test I wrote failed and corrected me.) Binding `-f` to a format there would recreate the exact `-o` ambiguity this PR declined to deepen.

### 3. `mj ai agents run --chat` bypasses detection — **real**

It went straight into `ConversationService.startChat()` — an stdin REPL — without ever passing the guard. Now refused up front, *ahead of the AI service import*, so the refusal is instant and can't be mistaken for a startup failure:

```console
$ mj ai agents run -a "X" --chat < /dev/null
 ›   Error: Interactive chat mode is required and this run has no interactive
 ›   terminal. Use --prompt "<your prompt>" for a single non-interactive
 ›   execution, or re-run --chat at an interactive terminal.
$ echo $?
1
```

### 4. Installer fails after scaffolding — **real, and it exposed a second bug**

The guard sat in `handlePromptEvent`, and the engine asks its questions from *inside* the phases (version picker and not-empty-dir confirm both live in Scaffold) — so it fired after files were on disk. A preflight now refuses before `engine.Run`, placed after `CreatePlan` (pure — no I/O) and after the `--dry-run` return, so a dry run stays headless:

```console
$ mj install --dir . < /dev/null
 ›   Error: Answers to the installer's setup questions (version, target
 ›   directory, configuration) is required and this run has no interactive
 ›   terminal. Pass --yes to accept defaults and install the latest version
 ›   unattended, adding --tag and --config to pin the version and supply settings.
$ ls -A . | wc -l
0
```

**The second bug, which your point led me to:** the engine installs its own catch-all prompt listener under `--yes` (`InstallerEngine.ts` — the auto-resolver that keeps Docker/CI from hanging). My CLI bridge registered alongside it, ran *first*, refused for want of a terminal, and `exit(1)`'d before the safety net could answer. So **`mj install --yes` in CI would hard-fail** on any prompt that net exists to absorb — a regression I introduced and would not have found without this thread. The bridge is no longer registered under `--yes`; the engine's net owns every prompt there.

### 5. Machine paths still emit text — **real, both examples**

- **`--dry-run`** printed chalk prose straight from the command, ignoring the resolved format entirely — the one output most likely to be read by a program was the one that never honoured `--format=json`. Now goes through the formatter:
  ```console
  $ mj ai actions run -n "Get Weather" --param "Location=Boston" --dry-run --format=json
  { "dryRun": true, "action": "Get Weather", "parameters": { "Location": "Boston" } }
  ```
- **Empty lists** — `formatAgentList`/`formatActionList` answered the empty case *above* the format switch, so `--format=json` returned `No agents found.`: unparseable, and indistinguishable to a caller from the command having failed. An empty result is `[]` now; the sentence stays for `compact`/`table`.

**On the third example — "some plain oclif commands when piped":** that one I've left, and I want to be explicit rather than quietly narrow your point. The ~80 commands that aren't `BaseCLIPlugin` subclasses have no formatter to route through, so making them emit structured output is a per-command migration, not a fix. What this PR does give them is chrome suppression and correct exit behaviour when piped. The two you named were different — both sit *inside* format-aware commands, so they were bugs.

### 6. `sync file-reset` connects before validating — **real**

`initializeProvider` + `getSyncEngine` ran at lines 49/52; `--sections`/`--all` wasn't resolved until line 77. All input resolution (sections, directory, confirmation) now happens first. Against a config pointing at an unreachable host, it reports the missing flag instead of a connection failure:

```console
$ mj sync file-reset < /dev/null
 ›   Error: A list of sections to reset is required and this run has no
 ›   interactive terminal. Pass --sections=primaryKey,sync or --all.
```

### 1. `-o table` treated as default — **already fixed, before your review**

This one was real but landed in `53a1db4`. Worth noting the failing case wasn't quite `-o table`: explicitness was inferred from `value !== default`, so `table` (≠ `compact`) was always honoured. The broken input was **`-o compact`** — typing the default explicitly was indistinguishable from not typing it, so piped it silently became json. All six `mj ai` commands now pass oclif's `metadata.flags.output.setFromDefault`.

---

### State

| | |
|---|---|
| `@memberjunction/cli-core` | **76 pass** |
| `@memberjunction/cli` | **783 pass** (+28 new) |
| `@memberjunction/ai-cli` | **81 pass** (+9 new) |

New coverage pins each finding: `-f` present on all seven test commands and absent from all six ai commands (asserted off the command classes, not the flag definition, so it fails where a caller would notice), the legacy values still reachable via `-f`, `[]` for empty lists in json, and the dry-run envelope.

The one remaining `cli` failure is pre-existing and untouched: `dev-workspace.member-installs` chmods a directory to `000` and expects it unreadable, which doesn't hold as uid 0. It fails identically on an unmodified tree; CI runs non-root and passes it.

Also merged `next` in (it had moved on the branch) — clean, no conflicts.

---

PR #4040, opened 2026-08-25
fix(codegen): escape quotes in generated entity descriptions before SQL interpolation

MS-BC:
In `generateNewEntityDescriptions`:

```ts
const sSQL = `UPDATE ... SET Description = '${result.entityDescription}' WHERE Name = '${e}'`;
```

`entityDescription` is **free text from an LLM**. Prose about business entities contains apostrophes as a matter of course, and the first one aborts the entire CodeGen run with `Unclosed quotation mark after the character string` (SQL Server) — at the *end* of an otherwise-complete pass.

Observed live: an 8-minute in-process CodeGen over ~200 new entities completed every metadata phase — entities, fields, 635 keys, relationships — then died on exactly this while writing a description. The surrounding RSU then committed partial artifacts (a separate defect, PR to follow: RSU must not commit after a Critical CodeGen failure).

Fix: `''` doubling — correct on both supported dialects — applied to the description and the entity-name literals in the same function. `tsc --noEmit` clean; changeset included.


rkihm-BC (approved):
Approving. Correct fix, right technique for the constraint it's under, and complete within its function.

## What's good

**The failure mode is nastier than the diff size suggests.** An LLM writing prose about business entities produces apostrophes constantly — "the customer's account", "an order's line items". Because description generation runs at the *end* of the pass, the first one aborts a run that had already completed entities, fields, 635 keys and relationships. The cost isn't the failed UPDATE; it's the eight minutes.

**This closes the last gap in a convention the rest of the file already follows.** I checked every sibling that writes an LLM-sourced value to SQL in `manage-metadata.ts`:

| Site | Value | Escaped before this PR? |
|---|---|---|
| `applyLLMFieldDescriptions:3383` | `fd.description` | ✅ |
| `:7797` | `fieldCategory.displayName` | ✅ |
| `:7801` | `extendedType` | ✅ |
| `:7808` | `codeType` | ✅ |
| `:884` | `okConfig.Description` | ✅ |
| **`:4706`** | **`result.entityDescription`** | ❌ ← this PR |

A sweep of CodeGenLib for unescaped `Description = '${...}'` returns only this line. So it's the last instance of a class already handled everywhere else, not a one-off patch on a widespread problem.

**Escaping is the right technique here, not a shortcut.** Parameterised queries would normally be the answer, but `LogSQLAndExecute` routes through `SQLLogging.LogSQLAndExecute`, which captures the literal SQL into the log that becomes a release migration. Parameterisation would emit a logged statement with unbound placeholders — unusable as a migration. Literal SQL with correct escaping is the only option on this path, and `''` doubling is right on both dialects.

**Escaping `e` as well is a real fix, not padding.** The `WHERE Name = '${e}'` in the SELECT two lines up had the same exposure — an entity whose name contains an apostrophe would have failed there first, before the description was ever reached. Fixing both closes the whole function rather than the reported symptom.

**Correctly left alone:** `WHERE EntityID='${data[0].ID}'` is a UUID read back from the database, not free text. Knowing which of the three interpolations *doesn't* need touching is the part that shows this was reasoned about rather than pattern-matched on quotes.

Changeset `patch` on `codegen-lib` is correct — no migration, no metadata.

## Two follow-ups, neither blocking

**A shared escape helper.** `const esc = (s: string) => s.replace(/'/g, "''")` now appears four times in this file — 1392, 1616, 2135 are pre-existing, this adds the fourth at 4697 — alongside 23 inline `.replace(/'/g, "''")` sites. The class already has the right home for it: a `protected escLit()` beside `qi()`/`qs()` would let all 27 converge, and would make the *next* interpolation correct by default, which is the actual root cause of this bug class. Deliberately not asking for it here — a fix to an active breakage shouldn't be held to refactor 27 call sites, and doing it in this diff would make it harder to verify rather than easier.

**The RSU partial-commit defect** you flag in the description is the more serious of the two. This narrows what can trigger a mid-pass abort; it doesn't change what happens when one fires. Splitting them is right — I'd just want that follow-up to land, since any other abort cause still leaves RSU committing partial artifacts.

## On tests

None here, and unlike #3933 I don't think one is warranted. The behaviour is a one-line string transform on a path requiring the `EntityDescriptions` feature enabled, a live LLM call and a database round-trip; a test would end up asserting that `String.replace` works. The sibling `applyLLMFieldDescriptions` has carried the same escaping untested since it was written.

PR #4042, opened 2026-08-25
[DO NOT MERGE] ci verification: filtered path for #4041

rkihm-BC:
Throwaway verification branch for #4041. **Do not merge; will be deleted.**

#4041 modifies `.github/workflows/test.yml`, which the scope detector treats as a root/global change — so that PR always takes the FULL-suite path and can never exercise the **filtered** path that every ordinary PR uses. That gap already hid one real bug (the planner received a mangled turbo filter and died with `Could not find task`).

This branch forces the filtered path by removing `.github/workflows/test.yml` from the root/global grep, and makes a no-op comment edit in `@memberjunction/markdown-core` for the filter to select.

What it should demonstrate:
- `--filter=...[origin/next]` reaches turbo intact → **27 packages → 6 shards**
- shards balance to ~242–308s (shard 1 is ng-dashboards alone, the irreducible floor)
- the ESM guard runs **scoped** to markdown-core rather than sweeping all 221 packages


rkihm-BC:
Verification complete — the filtered path works end to end: 27 packages across 6 shards, ESM guard scoped to 1 package in 1s, all green in 11m32s (run 32820270894). Results recorded on #4041. Closing; this branch was never mergeable.

PR #4043, opened 2026-08-25
[DO NOT MERGE] ci measurement: worst-case fan-out for #4041

rkihm-BC:
Throwaway measurement branch for #4041. **Do not merge; will be deleted.**

The 55m18s baseline ([run 32746276615](https://github.com/MemberJunction/MJ/actions/runs/32746276615), PR #4028) was a `@memberjunction/core` change, which fans out to essentially the whole repo. To compare like with like, this branch touches `@memberjunction/core` too — `...[origin/next]` selects **307 packages with a test task, 311 build tasks**, matching the baseline's 309/307.

It also removes `.github/workflows/test.yml` from the root/global grep, because a PR carrying the workflow change always forces the full-suite path and so can never measure the filtered worst case.

This is the apples-to-apples number for #4041.


rkihm-BC:
Measurement complete — the worst case (a `core` change, 307 packages, the same fan-out as the 55m18s baseline) is 21m23s, all green (run 32821256459). Results recorded on #4041. Closing; this branch was never mergeable.

PR #4045, opened 2026-08-25
fix(codegen): do not GRANT or integrity-check excludeSchemas entities

AN-BC:
## Summary
- `excludeSchemas` means this CodeGen run does not own those schemas. We were still generating and executing GRANT files for every excluded entity (STEP 2(c) + `applyPermissions(baselineEntities)`).
- In a joined multi-app database that fails the whole run: `GRANT SELECT ON [__mj_BizAppsCommon].[vwContactMethods]` from an orders CodeGen exits 1 when the object is missing, and writes sibling files into `SQL Scripts/generated/<other-app>/`.
- The entity-field sequence integrity check had the same hole — it `SELECT TOP 1`s every entity's BaseView, including excluded schemas.
- Permissions and the sequence check now run only for included entities. Core GRANTs continue to ship with MJ migrations.

Verified locally: orders `mj codegen` against `bizapps_e2e_grok_0825` with sibling schemas excluded now exits 0 and no longer attempts the Contact Methods GRANT.

## Test plan
- [x] `vitest run src/Database/__tests__/exclude-schema-permissions.test.ts` in CodeGenLib
- [x] Rebuild `@memberjunction/codegen-lib` and re-run `mj codegen` from bizapps-orders with `excludeSchemas` listing `__mj` and sibling Open App schemas — exit 0, no sibling GRANT errors
- [ ] Confirm MJ-core CodeGen (where `__mj` is **included**) still applies core view/SP permissions as before

### DIFF
--- a/.changeset/quiet-schema-grants.md
+++ b/.changeset/quiet-schema-grants.md
@@ -1,0 +1,5 @@
+---
+"@memberjunction/codegen-lib": patch
+---
+
+Stop generating and applying GRANT files for excludeSchemas entities. Open App CodeGen was failing with "Cannot find the object 'vw…'" on sibling-schema permission files, and the entity-field sequence integrity check was querying those same out-of-scope base views.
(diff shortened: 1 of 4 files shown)


AN-BC (commented):
LGTM

AN-BC:
@MS-BC @bc-izygmunt @rkihm-BC @MarceloT-BC merging this now, it addresses a codegen run issue in open apps when you use exclude schemas - the run would work but would fail due to previously valid logic that doesn't work in the new era where there are open apps.

PR #4048, opened 2026-08-25
fix(migrate): heal Open App metadata after migrate/install

AN-BC:
## Summary
- After each Open App `mj migrate --schema` and `mj app install` (`RunAppMigrations`), run the core metadata-heal steps that `__mj`'s `R__RefreshMetadata` already runs for MJ itself. Open App Flyway histories never ran that repeatable, so layered `BaseView` column order and `EntityField.Sequence` drifted (e.g. Organizations `RootParentID` at sequence 18 vs view ordinal 20).
- `spRecompileAllViews` now refreshes views in dependency order (inner generated view before `g.*` wrappers, and wrappers like `vwContactMethods` after `vwPeople`). Optional include/exclude schema lists match the other heal procs.
- CodeGen inserts new `EntityField` rows at the live BaseView ordinal after parking existing sequences (`+100000`); Pass 2 still calls `spUpdateExistingEntityFieldsFromSchema` after views are current.
- PostgreSQL Open App heal runs AllowsNull, orphan prune, and catalog Sequence (no view recompile — `SELECT *` is frozen at `CREATE`). Layered-view *generation* on PG is not in this PR.

## Test plan
- [x] Rebuild `@memberjunction/sql-converter` then MJCLI (stale converter dist was missing `ConversionStats.Gaps`).
- [x] Clean SQL Server DB: `mj migrate` (MJ, including `V202608251800`) then `mj migrate --schema __mj_BizAppsCommon`; CLI printed `Metadata refreshed for __mj_BizAppsCommon`.
- [x] Organizations: `RootParentID` Sequence 20, 31/31 fields aligned to `vwOrganizations`, inner + outer + `vwContactMethods` queryable.
- [x] Unit tests: Open App SQL builders (SS + PG), CodeGen sequence insert.
- [ ] CI on this PR.
- [ ] Follow-up: Open App migration cleanup, then PG layered-view outer recreate (custom SQL; PG equivalent from pg-migrate).


AN-BC (commented):
lgtm

PR #4059, opened 2026-08-26
feat(codegen): opt-in LLM display names for entities

MS-BC:
## The asymmetry this closes

`EntityField.DisplayName` already gets LLM values today. The FormLayout advanced-generation feature returns `fieldCategories[].displayName`, and [manage-metadata.ts:6381](https://github.com/MemberJunction/MJ/blob/next/packages/CodeGenLib/src/Database/manage-metadata.ts#L6381) writes it wherever `EntityField.AutoUpdateDisplayName = 1`.

`Entity.DisplayName` has no equivalent. `createNewEntityDisplayName()` is purely mechanical — it strips the schema's configured prefix/suffix (`"CRM: "`) off the entity name and otherwise returns `null`. It runs once at entity creation and is never revisited. The `EntityNames` feature *does* call an LLM, but it sets `Entity.Name` — an identifier that code and metadata reference, so changing it is a breaking rename — and only for brand-new entities.

The result: fields can have good display names, entities cannot.

## Scoped to where a model actually helps

The deterministic [createDisplayName()](https://github.com/MemberJunction/MJ/blob/next/packages/MJGlobal/src/util.ts#L489) is better than it gets credit for — it splits underscores, normalizes ALL-CAPS, splits compound words, then converts camelCase to spaces. Verified against it directly:

```
"CustomerOrders"  -> "Customer Orders"     ← needs no model
"ACCT_STAT_CD"    -> "Acct Stat Cd"        ← correctly spaced, still unreadable
"MBR_XREF"        -> "Mbr Xref"            ← same
```

What it cannot do is **vocabulary**. Expanding `Acct` to `Account` requires knowing what the abbreviation means. That, and only that, is what this feature adds.

So a new `assessDisplayNameOpacity()` decides which entities are worth a call, on three signals ordered by confidence:

1. **A vowel-less token** (`Mbr`, `Cd`, `Txn`) — near-certain abbreviation.
2. **A digit welded into a token** (`SHA256`) — identifier-style, rarely self-evident.
3. **A short token outside a common-word allowlist** (`Qty`, `Amt`) — weakest signal; the allowlist keeps `Log`, `Job`, `Key`, `Tag` from firing.

This is a **cost filter, not a correctness gate** — the `alwaysGenerate` feature option bypasses it entirely. The asymmetry is deliberate: a false positive wastes one call whose answer is probably the mechanical name anyway; a false negative silently leaves an unreadable name in place.

**Known limitation, tested and documented:** four-letter abbreviations standing alone (`Addr`, `Dept`, `Prod`) are *not* caught. Widening the short-token rule to length 4 would also fire on `Item`, `Type`, `Rate`, `Line`, `Code` — whole words in most schemas — and the filter would stop filtering. `alwaysGenerate` is the escape hatch. There's an explicit test pinning this so it reads as a decision, not an oversight.

## Never overwrites a human

New `Entity.AutoUpdateDisplayName` column — `BIT NOT NULL DEFAULT 1`, mirroring `EntityField.AutoUpdateDisplayName` (added for fields back in v2.122). Without it an LLM pass could not tell "this is the mechanical default" from "an administrator chose this", and would have to either skip every entity that already has a DisplayName — useless on exactly the legacy schemas this targets — or overwrite deliberate choices.

Defaulting to `1` is safe because the flag alone changes nothing: the feature it gates ships **disabled**. Defaulting to `0` would mean enabling the feature later silently does nothing until each row is flipped by hand, which reads as broken rather than off.

Three further refusals to write, each of which would otherwise make the feature harmful rather than merely unhelpful:

- **`low` confidence results are discarded.** A confidently wrong display name is worse than an ugly accurate one, since nothing downstream flags it as a guess.
- **Empty or >255 chars discarded** rather than truncated (the column is `nvarchar(255)`).
- **The flag is re-checked in the UPDATE's own WHERE clause**, not just read beforehand.

The model must also return which abbreviations it expanded, so a questionable expansion is auditable from the log rather than only its result — `STAT` → `Status` and `STAT` → `Statistic` are both plausible, and only context distinguishes them. The prompt is given the entity's **field list** for exactly that reason: `ACCT_STAT_CD` alongside `StatusName`/`IsActive`/`SortOrder` is a status lookup, not accounting statistics.

## Backfilling an existing schema

This runs through the existing per-entity advanced-generation path, so it covers new and modified entities automatically, and all entities under `forceRegeneration.enabled`. Combined with `forceRegeneration.entityWhereClause` (e.g. `SchemaName = 'legacy'`), that is the scoped backfill route for an existing database — no new mechanism invented.

## ⚠️ Adjacent finding — not fixed here

While placing the feature I found that **four declared config feature names are never read by any code**:

| Declared in config | Read by code? |
|---|---|
| `FormLayout` | ❌ never — code checks `FormLayoutGeneration` |
| `FormTabs` | ❌ never |
| `EntityFieldDescriptions` | ❌ never |
| `DefaultInViewFields` | ❌ never — *and it is `enabled: true` in `DEFAULT_CODEGEN_CONFIG`* |

There are two feature lists in `config.ts`: the zod `.default([...])` (~line 185) and `DEFAULT_CODEGEN_CONFIG` (~line 891). The latter is the live one — it is merged with user config at load — and their names have drifted apart.

I deliberately did **not** fix this. Renaming feature keys would silently flip features on or off in any deployment that sets them by name in `mj.config.cjs`, which is a behavior change that deserves its own PR and its own decision. Flagging it here because it is directly adjacent: I added `EntityDisplayNames` to **both** lists so it works either way.

## Verification

| Check | Result |
|---|---|
| New heuristic tests | **21/21 pass** |
| Full CodeGenLib suite | 1139 passed, 2 failed, 66 skipped |
| TypeScript errors in files I touched | **0** |

**On the 2 failures and the 11 tsc errors** — both are artifacts of my verification setup, not this change, and I confirmed rather than assumed that. The worktree resolves `node_modules` to the main checkout, which sits on a different branch, so its built `dist` predates symbols that `next`'s source has. Proof: `IsHierarchy` is present in `packages/MJCore/src/generic/entityInfo.ts` on this branch and **absent** from the resolved `MJCore/dist/generic/entityInfo.d.ts`; same for `ResolveSingleEntityResourceTarget` in `@memberjunction/global`. Both failures (`BaseViewEmissionGoldenMaster` GM-VIEW-09, `entity-subclasses-hierarchy`) are hierarchy tests that depend on it. Nothing in either touches display names, config, or advanced generation. CI will build against a correctly-installed tree.

**Not run, and needed before this leaves draft:**

- **`mj codegen`** — the migration adds a column, so CodeGen must generate the `EntityField` metadata row for `Entity.AutoUpdateDisplayName` and regenerate the Entity CRUD procs. Per `migrations/CLAUDE.md` that output gets appended below the hand-written DDL with the 50-blank-line separator convention. I left the migration as hand-DDL only rather than hand-writing metadata CodeGen owns.
- **PG/SS parity** (`scripts/pg-ss-parity.sh`) — needs live databases. Filename and folder parity are correct (`migrations/v6` ↔ `migrations-pg/v6`, same version stamp).
- **A live LLM run.** The prompt template and its model bindings are untested against a real model. Worth one real pass over a legacy schema before merging — the heuristic is unit-tested, the prompt's actual output quality is not.

The migration is timestamped `V202608252345`, after the current latest on `next` (`V202608252300`), so it does not land out of order.


rkihm-BC (changes requested):
Reviewed as a draft, so read this as "what is left" rather than "what went wrong". The design work here is done and it is good; one mechanical step remains, and it is the cause of both red checks.

**Housekeeping first:** I pushed `3103e2463f` to this branch — a migration retimestamp. That was necessary because *I* merged `next` in to clear a stale `agent-init` unit-test failure, and the merge brought `V202609212241__v6.2.x__Feature_Pipelines.sql` with it, pushing the high-water mark past your `V202609210930`. Churn my merge caused, not anything in your diff. The PostgreSQL counterpart moved in lockstep, since it names the T-SQL file in a comment. `Check migrations` is green as a result.

## What is left — the migration needs its CodeGen output

The migration is 49 lines and contains only the physical column:

| | |
|---|---|
| `ALTER TABLE Entity ADD AutoUpdateDisplayName` + extended property | ✅ present |
| `INSERT INTO … EntityField` | **0** |
| `CREATE OR ALTER PROCEDURE` | **0** |
| `CREATE OR ALTER VIEW` | **0** |

(The three `EntityField` hits in the file are comment prose, not SQL.)

So the physical table gains a column that MJ's metadata and generated objects know nothing about. `mj sync push` then builds:

```sql
DECLARE @ResultTable TABLE ( [ID], [ParentID], …, [CanonicalSchemaName] );   -- no AutoUpdateDisplayName
INSERT INTO @ResultTable
EXEC [__mj].spUpdateEntity @ParentID=…, @ID=…;
```

`INSERT … EXEC` requires the proc's result-set shape to match `@ResultTable` exactly, and it no longer does:

```
Error: Column name or number of supplied values does not match table definition.
```

**Both red checks are this one thing.** `CodeGen drift gate` fails at *"Migrate + sync push"* — before CodeGen ever runs, so it is not drift — and `Integration (SQL Server, deterministic)` shows the identical `spUpdateEntity` error eight times. Running CodeGen against a migrated database and appending its output (EntityField row, regenerated `spCreateEntity` / `spUpdateEntity`, `vwEntities`, permissions) should clear both together.

For scale: the comparable column-adding migrations on `next` are `Native_Tool_Calling` at **10,671** lines (27 procs, 6 views) and `AIAgentRun_ContinuationDepth` at **5,877** (12 procs, 2 views). This one is 49.

Worth knowing: `check:codegen-tail` could never have caught this. Its own header says it keys on `CREATE TABLE`, so "a migration that adds a column … is invisible here". The drift gate is the only guard, which is why this surfaced only when the gate got far enough to run.

## Your adjacent finding is real — I verified it

| `featureEnabled(…)` | call sites |
|---|---|
| `FormLayout` | **0** |
| `FormTabs` | **0** |
| `EntityFieldDescriptions` | **0** |
| `DefaultInViewFields` | **0** |
| `FormLayoutGeneration` | 1 |

Four declared names with no readers, and `DefaultInViewFields` is `enabled: true` in `DEFAULT_CODEGEN_CONFIG` — so an operator can switch on a feature that does not exist and get no signal at all. That deserves its own issue rather than living in a PR description; happy to file it if you would rather stay focused here.

## What is already right

- **Scoping the model to vocabulary only.** Showing that `createDisplayName()` already produces `Customer Orders`, and that only `Acct` → `Account` needs a model, is the right way to justify an LLM call — by demonstrating what the deterministic path cannot do rather than assuming it is inadequate.
- **Cost filter, not correctness gate**, with `alwaysGenerate` as the bypass and the asymmetry stated plainly: a false positive wastes one call, a false negative silently leaves an unreadable name.
- **The known limitation is tested, not just documented.** `display-name-heuristics.test.ts:104` — `it('does NOT catch four-letter abbreviations standing alone (known limitation)')`, pinning `Addr2Line` and `DeptRosters` to `false`. 11 cases total, and all six unit shards are green, so they ran. A test that pins a gap turns it from an oversight into a decision.
- **`AutoUpdateDisplayName` defaulting to `1`**, with the reasoning given: safe because the feature ships disabled, and defaulting to `0` would make enabling it later look broken.
- **Re-checking the flag in the `UPDATE`'s own `WHERE`**, discarding `low` confidence, and requiring the model to report which abbreviations it expanded so `STAT` → `Status` vs `Statistic` is auditable rather than only its result being visible.

The changeset is `minor`, which is correct — this touches both `migrations/` and `metadata/`.

## Ask

Please **move it out of Draft once the CodeGen output is appended and the two checks go green**, and I will re-review. I do not expect to have much to add at that point — the reasoning is already the strongest part of this PR.

PR #4063, opened 2026-08-26
feat(providers): submit a TransactionGroup in one round trip per shape, not one per item

MS-BC:
## The problem

`SQLServerTransactionGroup.HandleSubmit` opens ONE transaction and then sends its items ONE AT A TIME, awaiting each:

```ts
for (const item of items) {
    const request = new sql.Request(transaction);
    const queryResult = await request.query(item.Instruction);
}
```

That is atomicity without batching. N items still cost N round trips, and on a high-latency link the round trip **is** the write ceiling — SQL execution measures in single-digit milliseconds while the trip costs tens. `PostgreSQLTransactionGroup` has the same shape.

This is why a `TransactionGroup` does not make writes faster, and why an earlier experiment that wrapped 500 records in one transaction measured **34% slower** rather than faster: commit batching was never the cost.

## The statements were already built for this

`SQLServerDataProvider.RenderSaveCallBinding` emits every save as:

```
DECLARE @Name_a1b2c3d4 NVARCHAR(100);
SET @Name_a1b2c3d4 = N'…';
EXEC [dbo].[spCreateWidget] @Name=@Name_a1b2c3d4
```

and its own doc comment says the uuid suffix exists *"to keep batched saves (`SQLServerTransactionGroup`) collision-free"*. The infrastructure for concatenation was in place; the submit never used it.

## What changes, and what deliberately does not

Only the trip count. The SQL sent is byte-for-byte what the serial path sent — the same generated CRUD procedures run, the same Record Changes rows are written, the same save events fire, the same subclass hooks run.

That is the whole point. The other way to collapse these round trips is to stop calling the procedures and write rows directly, which buys its speed by giving up every one of those — including the cache-invalidation events that `DatabaseProviderBase.TrustLocalCacheCompletely` is justified on, ~900 lines above `BulkCreate` in the same file. Batching gives the same collapse and gives none of it up.

## Measured

SQL Server 2022, the generated entity procedure **and** its Record Change writes running for every row, identical SQL in both arms, modelled 25ms RTT:

| | wall | trips | rows/min | DB ms/row |
|---|---:|---:|---:|---:|
| per-statement (today) | 569,877 ms | 18,000 | **211** | 49.97 |
| batched, group=25 | 5,703 ms | 80 | 42,085 | 2.79 |
| batched, group=100 | 1,822 ms | 40 | **131,713** | 2.52 |
| batched, group=200 | 1,230 ms | 20 | 195,158 | 2.68 |

Two things this establishes beyond the headline. The per-statement arm reproduces the ~250/min figure measured live, so the round-trips × RTT cost model holds on this rig. And **DB cost per row stays flat as the group grows 8×** — the compile-bound wall that punishes very large literal batches is not reached at these sizes (batch text topped out at 368 KB).

## The two things batching gets wrong if written casually

**Parameter namespaces.** One request carries one parameter namespace. Two items that each rendered `?` both want `@p0`, and the second would silently bind over the first. Parameters are renumbered across the whole batch into a single `@p<n>` sequence.

**Result mapping.** `recordsets` cannot be zipped to items positionally, because an item can return **no result set at all**. Verified against the driver: a three-item batch whose middle item was an UPDATE matching nothing came back as **five** result sets, not six. Each item is therefore preceded by a sentinel `SELECT <i> AS [__mj_batch_item]`, and results are assigned by walking sentinel to sentinel.

## Behaviour change, stated plainly

Serially, the client stopped *sending* after the first failure, so later statements were never transmitted. Batched, they travel together and the server may reach some of them before the error.

The guarantee that matters is unchanged, and is what the test asserts: the transaction rolls back, **nothing commits**, and every item reports failure. There is a test named for exactly this so it cannot be changed silently.

The `Variables` path — where items genuinely depend on each other's results — stays serial by construction, since that dependency is the one thing batching cannot preserve.

## Tests

9 for the pure batching logic (sentinels, cross-item parameter renumbering, self-contained instructions, the no-result-set item, an out-of-range sentinel), plus the existing `transaction-group` suite rewritten to the batched contract — one round trip, per-item result mapping, cross-item parameter isolation, rollback-on-any-failure, and the load-bearing "no rows returned means failed item but the batch still commits". Full package suite: **124 passing**.

## PostgreSQL, which needed a different route

Concatenation is not available there: the extended protocol — the one any parameterized query uses — carries exactly ONE statement per message, and node-postgres does not pipeline (its per-client queue sends the next query only once the previous completes). Inlining values to reach the simple-query protocol would trade a parameterized call for hand-rolled quoting of every type the provider supports (jsonb, timestamptz, bytea, arrays), which is a worse bargain than the round trips.

Items whose instruction is the **same shape** differ only in their values, so they combine into one statement:

```sql
SELECT 0 AS __mj_batch_item, x.* FROM (SELECT * FROM sp(p_a => $1)) AS x
UNION ALL
SELECT 1 AS __mj_batch_item, x.* FROM (SELECT * FROM sp(p_a => $2)) AS x
ORDER BY 1
```

One statement, still fully parameterized, the CRUD function invoked once per branch. The wrapping subquery means the instruction itself is never parsed or rewritten beyond renumbering its placeholders.

**Verified against real PostgreSQL**, not just unit-tested: a three-item group ran as one statement, inserted three rows, invoked the function once per branch, and returned each branch's row tagged with the right item index after global renumbering.

**Grouping is by the shape the provider emitted** — the instruction with its `$n` numbers normalized away — rather than by entity. `GenerateSaveSQL` emits only the fields it is saving, so two updates to one entity genuinely can differ, and a `UNION ALL` over mismatched column lists is an error rather than a subtle bug. Consecutive same-shape runs group; everything else is sent alone, exactly as today. Order is never rearranged to build a bigger group, since the caller may care what lands first.


MS-BC:
Superseded by #4087, which carries the same one-round-trip TransactionGroup submit as an opt-in `BatchedSubmit` flag on TransactionGroupBase, implemented for BOTH dialects (SQL Server @p renumbering + PostgreSQL escapeLiteral inlining with NaN/no-escaper sequential fallbacks) and using the sentinel-recordset protocol proven in production. Closing in its favor.

PR #4064, opened 2026-08-26
perf(integration): overlap the next page download with the current page's work

MS-BC:
## The gap

The fetch loop is strictly serial — fetch a page, apply it, fetch the next — even though **the next cursor is known the moment a page arrives**. On a connector whose fetch dominates its apply, which is the common shape, the shorter leg sits idle waiting for the longer one, twice per cycle.

## What this is, and what it is not

Latency hiding, not concurrency. Exactly **one** extra request is ever in flight, and it is the request the loop was about to make anyway. Nothing fans out, no vendor sees more parallelism, and the sync's ordering is untouched.

Two properties have to hold for that to be true, and both hold **by construction** rather than by keeping two code paths in step:

- **Vendor pacing is unchanged** — the prefetch acquires the same rate token through the same limiter.
- **Error semantics are unchanged** — the failure surfaces where the loop awaits it.

The fetch is now defined **once**, as a local used by both the loop and the prefetch:

```ts
const fetchPage = async (pageCtx: FetchContext): Promise<FetchBatchResult> => {
    await this.rateLimit(config);
    return await WithRetry(/* timeout, retry predicate, Retry-After, BeforeRetry re-acquire */);
};
...
batch = await (prefetcher.Claim(currentCursor) ?? fetchPage(ctx));
```

So an overlapped page carries the identical timeout, retry predicate, `Retry-After` honouring and token re-acquire. Two definitions would be two sets of semantics that drift.

## The safety property

A page in flight is served **only when its cursor is the one the loop is actually about to use**:

```ts
public Claim(cursor: string | null | undefined): Promise<TBatch> | null {
    const page = this.inFlight;
    this.inFlight = null;               // cleared on a miss too — see below
    if (!page || !this.enabled) return null;
    return page.Key === (cursor ?? '') ? page.Promise : null;
}
```

A gap skip, a reset, or a connector rewriting its own cursor moves the loop somewhere the prefetch did not anticipate. Serving that page would **silently skip or duplicate records**, which is the one failure this must make impossible.

A refused page is *dropped*, not retained — otherwise a later cursor could match a stale page by coincidence. There is a test named for exactly that.

## Cursor mode only, and why that is not a separate check

Arming requires a `NextCursor`, which offset- and page-paged connectors do not produce — so the gate falls out rather than being asserted. Those modes could not participate anyway: gap-skip resume can advance past a failed page, so the position the loop will ask for next is not knowable at the time a prefetch would have to start.

## Kill switch

`MJ_INTEGRATION_PREFETCH=off`. Anything other than `off` leaves it on, so a typo cannot silently disable a performance path.

## Tests

11 for the prefetcher: the claim hit and the two miss shapes, the stale-page-not-retained case, no-arm when the batch says there is no more or the connector produced no cursor, that the fetch starts **immediately** rather than at claim time (the entire point), and that an in-flight rejection is absorbed while it waits yet still throws into the loop that claims it — no unhandled rejection, no swallowed error.

Wired into the real loop, the full engine suite is **843 passing**, up from 832 with no regressions.


MS-BC:
Superseded by #4089 (stacked on #4088), which carries the same next-page-overlap idea but built on the production-proven shape: the prefetch context advances the FULL position (NextPage/NextOffset/NextCursor and NextAfterKeyValue — the stale-AfterKeyValue bug made every keyset connector's walk stop at exactly two server pages), and both loop-top fetch and prefetch share one governed envelope (rate limit, adaptive fetch gate, timeout, retry, once-per-episode throttle report). Closing in its favor.

PR #4066, opened 2026-08-26
feat(integration): let a sync batch send its writes together, through Save()

MS-BC:
## The missing link

#4087 made a `TransactionGroup` submit in one round trip. This makes the sync engine actually use one.

The apply loop already made a batch atomic:

```ts
await provider.BeginTransaction();
for (const record of batch) { await this.ApplySingleRecord(...); }
await provider.CommitTransaction();
```

But **atomicity is not batching.** Each record's `Save()` still sent its own statement, so N records cost N round trips — and on a high-latency link the round trip is the write ceiling.

## What changes

When the connection opts in, the batch creates a `TransactionGroup` and the records enrol in it. `Save()` then defers its **write** to `Submit()` while still doing everything else:

```ts
const sqlDetails = await this.GenerateSaveSQL(entity, bNewRecord, user);   // the generated CRUD procedure
if (entity.TransactionGroup && !bReplay) {
    if (entity.EntityInfo.TrackRecordChanges && sqlDetails.simpleSQL) {
        extraData.simpleSQLFallback = sqlDetails.simpleSQL;                // Record Changes accounted for
    }
    entity.TransactionGroup.AddTransaction(new TransactionItem(entity, ..., sqlDetails.fullSQL, ...,
        (transactionResult, success) => {
            if (success && transactionResult) this.OnAfterSaveExecute(entity, user, options, saveContext);
        }));
```

Validation, row-scope checks, the generated procedure, Record Changes, `OnAfterSaveExecute` — all of it still runs. **Only the moment the SQL travels changes.**

That is the entire distinction from writing rows directly. Direct writes reach comparable numbers by *not calling the procedures at all*, and pay for it with every stored-procedure side effect, every Record Changes row, and every save event — including the cache-invalidation events `DatabaseProviderBase.TrustLocalCacheCompletely` is justified on.

## Opt-in, failing closed

`Configuration.writeMode === 'batched'` on the connection. `ReadWriteMode` returns `''` for absent, unparseable, wrongly-typed or missing configuration, so **every failure mode lands on the proven per-record path**. A connection has to ask, and a malformed request is not an ask. Five tests pin exactly that, including `writeMode: true` and `writeMode: 1` not being read as an ask.

## Failure degrades the way it already did

A group that fails to commit is routed into the same handler a thrown error takes, so the existing degradation is reached by **both** shapes rather than only by a throw:

- counters restored from the batch snapshot,
- queued record maps discarded (they would otherwise point at rows that rolled back),
- the batch re-applied record-by-record, so one poison record cannot cost its healthy siblings.

`RollbackTransaction` is skipped in the batched case — a failed group has already rolled itself back and there is no provider-level transaction open, so calling it would throw over the real error.

## Why AsyncLocalStorage

The group rides the run context rather than a threaded parameter, which is what that context already exists for — its own comment says it propagates "to every helper the run calls without threading a parameter through ~20 private signatures". The entity is constructed several frames below the code that owns the batch, in `CreateRecord`, `UpdateRecord` and `DeleteRecord`; all three enrol.

`enrolInWriteGroup` is a no-op when no group is active, which is every existing caller — so an entity built outside a batched apply saves immediately, exactly as before.

## Depends on

#4087. Without the batched submit in the providers, a group is atomic but still serial — which is today's behaviour, so this is safe to land in either order but only pays once both are in.

Engine suite: **838 passing**.


rkihm-BC (commented):
## Review — sync writes through a transaction group

Read the full diff and traced the deferral semantics through `BaseEntity.Save()` rather than reasoning about them. The concurrency analysis here is the strongest thing in this stack — the shared `_transactionDepth` diagnosis is a real bug class, correctly identified, with a test that pins it. Nothing below blocks; one line to change, one thing to document, one question.

### Verified good

- **Per-batch ALS scope rather than a shared context slot.** Correct, and for exactly the stated reason: narrowing the mutex is what lets two maps overlap, and a single `writeGroup` field on the run context would then let the second map enrol its records into the first's group. The nested `runContext.run({...runCtx, writeGroup}, …)` is the right shape, and it has a property worth naming explicitly — the degradation path runs *after* the scope unwinds, so `enrolInWriteGroup` is automatically a no-op there and the re-apply can't double-enrol. That falls out of the design rather than needing a guard, which is the sign it's the right design.
- **Batching follows `writeMode`, not `useTransaction`.** Gating on `useTransaction` (`getSyncConcurrency(config) <= 1`) would have meant batching engaged only at concurrency 1 — the exact tradeoff this PR exists to remove, and invisible when it happened, because throughput simply failing to improve is indistinguishable from the feature not helping. The `!batchedWrites` guard on `RollbackTransaction` correctly mirrors the `!writeGroup` guard on `BeginTransaction`, so the two can't disagree.
- **The opt-in genuinely fails closed.** `writeMode: true` and `writeMode: 1` not counting as an ask is the test most people skip. I also checked the empty-group edge: `TransactionGroupBase.Submit()` returns `true` with zero pending transactions, so the `writeGroup && runCtx` path can't produce a spurious failure and re-apply. That's benign, but it's benign by luck rather than by design — an assert on `runCtx` would say so.

### One line to change

**`ReadWriteMode` uses `.Get('Configuration')`.** The typed property exists — `MJCompanyIntegrationEntity.Configuration` at `generated/entities/__mj.ts:242` — and this same file already reads it four times: `resolveScheduledFullSync`, `getSyncConcurrency`, `getCrossLayerPipelineEnabled` and `getConfigOverrides`. The repo rule (`.claude/rules/typescript-style.md`) forbids `.Get()` as a substitute for a generated property, and here it also makes the new reader the odd one out among its four neighbours.

### One thing to document

**`if (!saved)` no longer fires under batching.** `BaseEntity.Save()` returns `true` immediately for an enrolled entity and subscribes to `TransactionNotifications$` for finalization. So the per-record failure check and dead-letter attribution in `CreateRecord` / `UpdateRecord` / `DeleteRecord` don't run in batched mode — every server-side rejection surfaces as a whole-group failure and costs a full batch re-apply through the fallback.

That's the designed degradation and it works correctly. But it's a real behavioural change the PR body doesn't state, and it has a steady-state cost worth writing down: a batch containing a persistently poisonous row does 2× the work on every run, indefinitely, and the only signal is throughput. Worth a paragraph in the changeset next to the failure-handling discussion.

### One question

**Can a synced entity have a single identity/int primary key?** `CreateRecord` reads the PK immediately after the deferred save to build the record map:

```ts
const saved = await entity.Save();            // returns true; nothing has persisted yet
const entityRecordID = entity.PrimaryKey.KeyValuePairs
    .map(kv => String(kv.Value)).join('|');
await this.QueueRecordMap(recordMaps, …, entityRecordID, …);
```

Most shapes are fine: `NewRecord()` client-generates the UUID for a single-column `uniqueidentifier` PK, and a composite PK takes its values from the mapped fields before `Save()`. The one shape that isn't is a single identity/int PK, where the value only arrives in the group's callback at `Submit()` — after `QueueRecordMap` has already run. That would write a blank `EntityRecordID`, which is precisely the "duplicates on every incremental sync" failure the comment directly above that line documents.

I have no evidence sync ever targets such an entity, so this is a question rather than a change request. If it's possible, a cheap guard on the PK shape seems worth it; if it isn't, a comment saying batching assumes a client-generatable PK would stop the next person re-deriving this.

### One ask, shared with #4087

**Integration-tier coverage.** Both new test files build the engine with `Object.create(IntegrationEngine.prototype)` and stub `ApplySingleRecord` — they cover `ReadWriteMode`'s parsing and `applyRecordsIndividually`'s transaction flag, and neither ever touches `enrolInWriteGroup`, the ALS scoping, or `Submit()`. Between this PR and #4087, nothing anywhere enrols a real entity in a real group and submits it.

That test is also what would settle the identity-PK question above in either direction, and it's what would confirm the property the ALS design exists to guarantee: two entity maps applying concurrently, each with its own group, neither enrolling into the other's. One bundle covers both PRs.

MS-BC:
All four addressed — pushed as one commit. The PK question turned out to have a real answer, so it became a guard rather than a comment.

**The one line**

`ReadWriteMode` now reads the generated `Configuration` property. You were right that it was the odd one out among its four neighbours in the same file. Worth noting the unit test's double had to change with it — it only implemented `.Get()`, so the two tests that exercise the reader went red until the double supplied the typed property. That failure is itself the proof the switch actually happened rather than being cosmetic.

**The question — it is possible, so it is now guarded**

`EntityFieldInfo.AutoIncrement` makes the unsafe shape precisely nameable, and `IsSPParameter` already keys off it (`isUpdate || !this.AutoIncrement`), so the framework itself distinguishes "server assigns this" from "we bring it". That makes the shape reachable in principle rather than merely hypothetical — the engine registers entities over runtime-created tables, and nothing in the sync path forbids an identity key.

So `enrolInWriteGroup` now refuses an entity whose identity is a **single** auto-increment column, and it saves immediately instead: one round trip slower for that entity, and impossible to get silently wrong. Five tests pin the boundary — single uuid enrols, composite/soft key enrols, identity-only refuses, and a *partially*-identity composite still enrols (the unsafe property is "the whole identity is server-assigned", not "an identity column is present"), plus the no-group no-op.

Putting it at the enrolment seam rather than at the record-map site was deliberate: it is total there, whereas at the call site every future caller of `QueueRecordMap` would have to re-derive it.

**The thing to document — both consequences, in the changeset**

Written up next to the failure-handling discussion, framed the way you framed it: the steady-state cost is real and its only signal is throughput. The changeset now says explicitly that a connection seeing no gain from `writeMode: 'batched'` should be read as *"something in these batches fails every time"* rather than *"batching doesn't help here"* — that inversion is the trap. The identity-PK exclusion is documented alongside it.

**Your "benign by luck rather than by design" note**

Taken as written. `writeGroup && !runCtx` now throws instead of silently falling through to `runBatch()` unenrolled — because that path is worse than benign: `Submit()` on an empty group returns `true`, so the batch would have reported success while quietly running the per-record path, and the only symptom would be batching never appearing to help. Every sync path enters a run context today, so this is a wiring assertion, not a reachable branch.

**On the shared integration ask**

Covered in #4087 as **IT86 / `transaction-groups-batched`** (server transport, 5 checks): a real entity, a real group, a real `Submit()`, a real database — including TGB1 reading SQL Server's own plan cache to confirm the group arrived as one sentinel-fenced statement, TGB2 pinning per-item id mapping through a real driver, and TGB5 asserting batched and sequential are observably identical. That closes the "nothing anywhere enrols a real entity in a real group and submits it" gap for both PRs, as you suggested one bundle would.

The one property it does **not** cover is the concurrency one you named — two entity maps applying simultaneously, each with its own group, neither enrolling into the other's. That needs two live CompanyIntegrations with real entity maps, which is a fixture an order of magnitude heavier than the bundle. It stays pinned at unit tier (`BatchedWriteFallbackConcurrency`), and I'd rather say that plainly than imply the integration bundle proves it.

Engine suite: 853 passing. `tsc`: 0 errors.

MS-BC:
#4087 is merged, so the wire-up it was waiting on is now in: the engine arms `writeGroup.BatchedSubmit = true` on the group it creates.

That one line is the batching. Everything before it bought *atomicity* — the group defers each `Save()` to `Submit()` — but a group without the flag still sends one round trip per item, which is exactly the pre-PR behaviour. That is why this was safe to land in either order, and why it only pays now.

Worth stating plainly: **no other test in the file would have caught its absence.** Every existing test here passes with the flag unset, because the writes still go through `Save()` and still commit atomically — the only difference is how many round trips they take, which none of them observe. So the wire-up has its own pin, and I verified it fails when the assignment is removed and passes when restored.

Also merged `next` in, so the branch builds against the providers that actually implement the flag. Engine suite: 878 passing, `tsc` clean.

rkihm-BC (commented):
All four are addressed, and three of them went further than the ask. Re-read the merged head against `next` rather than the summaries, and traced the batched branch rather than reasoning about it. One recommended change below — it falls out of the fix for the third item rather than being new ground.

## The four

**The one line.** `ReadWriteMode` now reads the generated `Configuration` property. The detail worth keeping is the one you noted: the unit double only implemented `.Get()`, so the two tests exercising the reader went red until it supplied the typed property. A cosmetic swap wouldn't have moved a test.

**The thing to document.** Documented better than I asked. The changeset states the mechanism *and* the trap — "a connection seeing no gain from `writeMode: 'batched'` should be read as *something in these batches fails every time*, not as *batching does not help here*." That inversion is the part that otherwise costs someone a day of profiling the wrong thing.

**The question.** Answered rather than deflected. `EntityFieldInfo.AutoIncrement` makes the unsafe shape nameable, `IsSPParameter` already keys off it, and the refusal sits at the enrolment seam where it is total. The boundary tests are right where it matters — a *partially* identity composite still enrols, because the unsafe property is "the whole identity is server-assigned", not "an identity column is present". That distinction is the one a re-derivation at the call site would get wrong.

**Benign by luck.** `writeGroup && !runCtx` now throws. Correct call: `Submit()` on an empty group returns `true`, so the silent version would have reported success while quietly running the per-record path — and the only symptom would be batching never appearing to help.

And the `BatchedSubmit` wire-up landed with its own pin. Your point that no other test in the file would have caught its absence is exactly right — every existing test passes with the flag unset, because they observe what the writes *do*, not how many round trips they take.

I also checked one thing I expected to be wrong and it wasn't. The changeset justifies dropping the provider transaction in the fallback on "`ApplySingleRecord` performs exactly one write". `DeleteRecord` contains both a `Save()` and a `Delete()`, which reads like two — but they are either/or, since SoftDelete returns after the archive save. The claim holds.

## One recommended change: decide batching per entity map, not per connection

The identity-PK refusal is correct at the seam. Its interaction with the branch around it is not.

`batchedWrites` is a property of the connection, the group is created unconditionally when it is true, and `BeginTransaction` is then skipped:

```ts
const batchedWrites = this.ReadWriteMode(companyIntegration) === 'batched';
...
const writeGroup = batchedWrites ? await provider.CreateTransactionGroup() : null;
...
if (!writeGroup) await provider.BeginTransaction();
```

For an entity map whose target has a single auto-increment PK, **every** record is refused enrolment. So no transaction opens, nothing enrols, and `Submit()` — which skips `HandleSubmit()` entirely when `_pendingTransactions.length === 0` — returns `true`. The batch reports success having been non-atomic, inside the branch whose premise is that it is atomic by construction.

The failure path compounds it. On a mid-batch throw, `RollbackTransaction` is correctly skipped, `recordMaps.Discard()` runs, and the fallback re-applies the **whole** batch — including rows that already committed. For creates that is a duplicate, because an identity key gives nothing to recognise the first one by. That is the same duplicate-on-sync failure the guard exists to prevent, reached by a different route.

The two positions can't both hold: the justification for adding the guard is that the shape is reachable — "the engine registers entities over runtime-created tables, and nothing in the sync path forbids an identity key". If it is reachable, so is this. If it is not, the guard is dead code and five tests pin a rule nothing reaches.

### The change

`entityMap.Entity` is available before the group is created, so this is one metadata lookup per map:

```ts
// Batching is a property of the ENTITY MAP, not only the connection. An entity whose whole
// identity is server-assigned cannot enrol (see enrolInWriteGroup), so a map targeting one
// creates a group that stays empty — and an empty Submit() returns true without opening a
// transaction, reporting a non-atomic batch as a successful atomic one.
const batchedWrites = this.ReadWriteMode(companyIntegration) === 'batched'
                   && !this.entityMapHasIdentityOnlyPK(entityMap);
```

```ts
private entityMapHasIdentityOnlyPK(entityMap: ICompanyIntegrationEntityMap): boolean {
    const info = entityMap.Entity ? this.ProviderToUse?.EntityByName(entityMap.Entity) : null;
    const pks = info?.PrimaryKeys ?? [];
    return pks.length === 1 && pks[0]?.AutoIncrement === true;
}
```

Such a map then falls to the `useTransaction` path and gets real atomicity, instead of a group guaranteed to stay empty. Keep the `enrolInWriteGroup` guard as well — it is total at the seam, which is why it belongs there.

**Worth doing whichever way the factual question resolves.** If the shape is reachable it closes a correctness gap; if it is not, it costs a metadata read and turns an implicit assumption into an explicit one, so a future entity registration trips the branch instead of silently producing non-atomic batches.

It is also the same class of issue as the `runCtx` fix, one layer out. That one mattered because an empty group's `Submit()` returns `true`; this one has the same root and the same consequence — a batch that reports success while having quietly done something else.

### One thing NOT to do

Do not guard this at `Submit()` by asserting the group is non-empty. **An empty group is a legitimate state**: when the content-hash fast path skips every record in a batch, nothing enrols and the group is correctly empty. At submit time the harmful case and the benign one are indistinguishable, which is precisely why the decision has to be made up front at the map level.

### The pin

One test that a map on an identity-PK entity does not create a group and does open a provider transaction. Same shape as the `BatchedSubmit` pin, and for the same reason — nothing else in the file would notice if the decision drifted back to the connection level.

## Everything else

The concurrency work remains the strongest thing in this stack. The shared `_transactionDepth` diagnosis is a real bug class, correctly identified, and the unit coverage pins the hazards that would otherwise be unobservable — the nesting deadlock (a deadlock leaves nothing to read), the depth counter surviving two concurrent fallbacks, and concurrent batches getting distinct groups rather than one shared slot.

Merge state is `CLEAN` and all checks are green against the merged head. Nothing here blocks except the decision above.

MS-BC:
Fixed as recommended — the decision now happens at the entity map, before the group is created.

You're right that the two positions can't both hold, and the gap is real. Walking it once more to be sure I understood it the same way: `batchedWrites` true → group created → `BeginTransaction` skipped on the strength of the group existing → every record refused at the seam → nothing enrols → `Submit()` short-circuits on `_pendingTransactions.length === 0` and returns `true`. A batch that ran non-atomically, reporting itself as the atomic branch. And the fallback compounds it exactly as you describe: no rollback (correctly — there is no transaction), then the whole batch re-applied over rows that already committed, with a server-assigned identity giving nothing to recognise the first copy by.

Same root as the `runCtx` fix, one layer out: **an empty group's `Submit()` returns true**, so every silent failure in this design routes through that one fact.

## What changed

```ts
const batchedWrites = this.ReadWriteMode(companyIntegration) === 'batched'
    && !this.entityMapHasIdentityOnlyPK(entityMap);
```

Such a map falls to the `useTransaction` path and gets real atomicity. The `enrolInWriteGroup` guard stays — it is total at the seam, and it is also where an entity whose metadata can't be resolved gets its answer, since the map-level check deliberately returns `false` (don't refuse) on an unresolvable lookup rather than silently disabling batching for the map.

One thing I did beyond the ask: the rule is now a single module-scope predicate used by both levels, rather than the shape being spelled out twice. Two copies of "the whole identity is server-assigned" drifting apart is precisely how a map ends up batching records that individually refuse to enrol — which is this bug's shape, so it seemed worth removing the possibility rather than re-introducing it one level up.

Took your point about **not** guarding at `Submit()` — the content-hash fast path skipping an entire batch is a legitimately empty group, and at submit time it is indistinguishable from the harmful case. That reasoning is in the code comment so the next person doesn't "simplify" it there.

## Tests

5 added (10 in that file): the identity-only refusal, the ordinary uuid target, a composite that merely *contains* an identity column (same distinction as the seam guard), an unresolvable entity not refusing, and the wiring pin. The pin is mutation-tested — remove the `&& !this.entityMapHasIdentityOnlyPK(entityMap)` and it goes red.

One existing test needed a change and it's worth flagging rather than burying: `batching is independent of concurrency` asserted the *exact text* of that line, so any added condition broke it — including a correct one. I rewrote it to pin the **intent** — no `useTransaction`, no `getSyncConcurrency` in the decision — instead of the literal string, so it still catches the regression it was written for without failing on legitimate change.

Engine suite: 889 passing, `tsc` clean. Also merged `next` in (it has moved a few times today).

rkihm-BC (approved):
Approving. The map-level fix is in and it is better than what I asked for. Waited for a clean board before posting — `Build`, `Integration (SQL Server, deterministic)`, all six unit shards, the ESM guard and the dependency/migration/standards gates are green, `mergeStateStatus: CLEAN`.

## The recommendation, implemented past the ask

```ts
const batchedWrites = this.ReadWriteMode(companyIntegration) === 'batched'
    && !this.entityMapHasIdentityOnlyPK(entityMap);
```

The part I did not ask for is the part that matters most: the rule is now a **single module-scope `IsIdentityOnlyPrimaryKey()`** consumed by both levels, rather than the shape being written out twice. Two copies of "the whole identity is server-assigned" drifting apart is precisely this bug's shape — a map batching records that individually refuse to enrol — so closing the possibility rather than reintroducing it one level up is the right instinct. My version would have left the second copy sitting there waiting to disagree with the first.

**The fail-direction is right, and I checked it rather than assumed it.** `entityMapHasIdentityOnlyPK` returns `false` when the entity cannot be resolved, so a metadata lookup failure does not quietly disable batching for the whole map — it defers to the enrolment seam, which is total there. That is the correct direction for a guard whose two levels have different completeness, and there is a test for it rather than a comment claiming it.

The reason NOT to assert this at `Submit()` is preserved in the code, including why: the content-hash fast path legitimately empties a group, so at submit time the harmful and benign cases are indistinguishable. That was the part most likely to be "simplified" back into a one-line guard later, and now it argues for itself.

## The test change you flagged rather than buried

`batching is independent of concurrency` asserted the exact text of the decision, so any added condition broke it — including a correct one. Rewriting it to pin the intent is the right call, and I checked it was not quietly weakened in the process:

- `[^;]*` spans the now-multi-line statement, so a concurrency gate added anywhere inside it still trips the negative assertions.
- If someone renamed `batchedWrites`, those negative assertions would pass vacuously — but the positive assertion fails loudly on the same rename, so the hole is closed from the other side.

Flagging it in the description rather than letting me find it in the diff is the difference between a test change I have to reverse-engineer and one I can check in a minute.

## Standing from both earlier rounds

The four original items are closed and stay closed: the typed `Configuration` read (proved by the double going red), the `if (!saved)` consequence documented with the throughput inversion spelled out, the identity-PK question answered at the seam with the partial-composite distinction correct, and `writeGroup && !runCtx` throwing instead of relying on an empty group's `Submit()` returning true.

Twenty-six tests across the three files. The one that earns its place is `is WIRED into the batching decision, not merely available` — the same guard-exists-but-is-never-called failure mode that #4087 and #4091 each needed a pin for, and the only kind of regression that leaves every other test in the file green.

I also re-checked one thing I expected to be wrong and was not: the changeset justifies dropping the provider transaction in the fallback on "`ApplySingleRecord` performs exactly one write", and `DeleteRecord` contains both a `Save()` and a `Delete()`. They are either/or — SoftDelete returns after the archive save — so the claim holds.

The concurrency work remains the strongest thing in this stack. The shared `_transactionDepth` diagnosis is a real bug class, correctly identified, and the coverage pins the hazards that are otherwise unobservable: the nesting deadlock, because a deadlock leaves nothing to read; the depth counter surviving two concurrent fallbacks; and concurrent batches getting distinct groups instead of one shared slot.

Good stack. The distinction it rests on — that the speed comes from how the SQL travels, not from skipping what the SQL does — is what makes it worth having over writing rows directly, and it held up under every question I put to it.

PR #4067, opened 2026-08-26
5.51.2 cache backport: #3389 + #3777 onto lts/5

jordanfanapour:
Backports two cache correctness fixes from `next` onto the 5.x line, following the `next`-first rule in `plans/lts-process.md` §5.2.

**Updated per Craig's review** — the #3425 commit has been dropped from this branch. See [what changed](#revision-dropped-3425) below.

Driver is Skip, which is pinned at 5.51.0 and staying on 5.x LTS, so neither fix has reached it. Skip runs multi-replica against Redis with `entity_object` as the default `ResultType`, which is exactly the path #3777 repairs.

| Upstream | Commit | Cherry-pick |
|---|---|---|
| #3389 | `052b4c788d` | clean |
| #3777 | `834f8d7081` | 6 conflicts, all context drift |

Both carry `patch` changesets and **no migrations** — `dbImpact: none`.

### #3777 — cross-server cache events poison `entity_object` engine caches

`BaseEngine.OnExternalCacheChange` assigned a cache-change payload's rows straight into the engine property. Payloads are serialized, so those rows are plain JSON objects — for any config whose effective `ResultType` is `entity_object` (**the default**) that silently replaced the array's `BaseEntity` instances, bypassing the coercing accessors and leaving `__mj_CreatedAt` a raw ISO string.

Production symptom: agents dying mid-run during context assembly with `__mj_CreatedAt?.getTime is not a function`. Optional chaining does not help — a string's `?.getTime` is `undefined`, and calling it throws.

Rows are now materialized via `TransformSimpleObjectToEntityObject` before assignment, with a refresh-generation guard so two overlapping cache events cannot resolve out of order. Seven unguarded `.getTime()` comparators route through `ToEpochMs()`, closing a latent NaN path as well.

Also fixes a **silent** variant in the scheduling engine: `evalTime < job.StartAt` does not throw on a string — relational operators coerce toward numbers, an ISO string yields `NaN`, and every comparison is false. `StartAt`/`EndAt` activation windows therefore stopped being enforced, and jobs could fire outside their range with nothing in the logs.

### #3389 — duplicate rows in engine caches from UUID casing

`BaseEngine.findEntityIndexByPrimaryKeys` compared PK values with raw `===`. The same UUID arrives with different casing depending on source (client-minted lowercase from `NewRecord` vs uppercase from SQL Server); on mismatch the lookup returned `-1` and the caller's *"not found → add it"* branch appended a duplicate into the cache. Pickers and dashboards showed rows twice; the database was correct throughout.

## Revision: dropped #3425

Per review, commit `fdf3cf2023` (#3425, freeze-on-write) is no longer on this branch. That removes 4,538 lines, `.IT77-cache-immutability.json`, `cache-immutability.checks.ts`, and the `ResolverBase` change. Verified afterwards: no dangling references to `SharesReferences`, `cache-immutability`, or `IT77` anywhere in `packages/` or `metadata-optional/`.

Two follow-ups from that split:

- The `ResolverBase` half is straight correctness and is proposed separately for 5.51.2 — it references the freeze only in comments, and both its tests call `Object.freeze` directly rather than depending on `LocalCacheManager`, so it stands alone.
- The freeze half stays on 6.x, since it turns in-place mutation of RunView rows into a `TypeError` and a patch on a certified line should not change behaviour.

**One correction to the review note:** the Redis rig and its two dependencies come from **#3777**, not #3425 — `rigs/cache-payload-materialization-tests.ts` and the `@memberjunction/redis-provider` + `ioredis` entries in `integration-test-suite/package.json` are both added by `f46c0aefdf`, and #3425 touches zero `package.json` files. So the rig survives this revision (which is what we want — it is the real-Redis proof for #3777), and the lockfile commit has to stay: without it `package.json` would require two packages `package-lock.json` has never heard of. A clean `npm install` on the rebased branch produces no lockfile diff, confirming the two agree.

## Conflict resolutions

**`packages/Scheduling/engine/src/ScheduledJobEngine.ts`** — kept the `lts/5` side. The hunk spanned three `next`-only methods (`shouldSkipMissedRun`, `advancePastMissedRuns`, `expireFinishedJobs`) that are unrelated features. All four applicable `ToEpochMs` changes plus the `Status !== 'Active'` guard landed; zero dangling references to the dropped methods.

**`pnpm-lock.yaml`** — **deleted.** This line is on npm (`package-lock.json`, `npm@11.7.0`); `next` migrated to pnpm.

**`metadata-optional/integration-test/test-suites/.integration-suite.json`** — registered only **IT84**, anchored after IT68. Dropped IT69–IT79, which register 6.x-only features absent from this line.

> ⚠️ **IT84 is sequenced 32, not upstream's 48.** On `lts/5` the first client-transport member (`IT03`) sits at 33, and the suite's own invariant says a server-transport member at or after a client member hard-errors, because `bootstrapIntegrationClient` rebinds the process-global provider to GraphQL. IT84 is `transport: server`, so it must precede 33. This is the only deliberate deviation from upstream.
>
> This line already has 16 pre-existing server-after-client orderings; untouched, and not introduced here.

**`check-registry.test.ts`, `CATALOG.md`** — kept the `lts/5` side. Both encode `next`'s IT numbering, which has diverged.

## Verification

Built and tested locally against a clean `npm install` on the rebased branch. **Build: 172/172 turbo tasks.**

| Package | Result |
|---|---|
| `@memberjunction/core` | 1631 passed |
| `@memberjunction/ai-agents` | 1776 passed (92 files) |
| `@memberjunction/generic-database-provider` | 842 passed, 5 skipped |
| `@memberjunction/server` | 805 passed, 56 skipped — 1 environmental failure, below |
| `@memberjunction/global` | 606 passed (21 files) |
| `@memberjunction/core-entities` | 574 passed (26 files) |
| `@memberjunction/scheduling-engine` | 208 passed |
| `@memberjunction/aiengine` | 111 passed |
| `@memberjunction/redis-provider` | 63 passed, 6 skipped |
| `@memberjunction/scheduling-engine-base` | 14 passed |

**The one failure:** `MJServer/src/__tests__/mjapi-bootstrap.test.ts` — `Failed to resolve entry for package "mj_generatedentities"`. That package's `main` is `dist/index.js`, produced by CodeGen against a live database, which this environment has none of. The backport touches neither the package nor the test.

**Not run:** the integration tier (needs a live DB/transport) and `rigs/cache-payload-materialization-tests.ts` (gated on `REDIS_URL`, deliberately outside the PR gate upstream).

Rebased onto `lts/5` @ `3e140cecc9`, which picks up #4070 — `test.yml` and `integration.yml` are now `pull_request: branches: [next, 'lts/**']`, so the checks on this PR do exercise the code. That supersedes the earlier note here that they could not.

## Release

Not released by this PR.


AN-BC (approved):
no comments on the algo changes. I'm not confident in how backports work so I will suggest you have @cadam11 and @rkihm-BC review these instead of me in the future as I generally won't have any objection to any known `next` code being backported when it improves LTS stability. Thanks @jordanfanapour

PR #4069, opened 2026-08-26
feat(codegen): IncludedSchemaNames heal scope and intra-schema cascades

AN-BC:
## Summary
- Add optional `@IncludedSchemaNames` to the five CodeGen metadata-heal stored procedures (SQL Server migration; PG via `metadataSupportObjects.ts`). NULL/empty keeps today's exclude-only behavior.
- CodeGen logs authored `excludeSchemas` plus `@IncludedSchemaNames` from `includeSchemas`. It no longer photographs sibling Open Apps into a folded V.
- Cascade-delete SQL is intra-schema only unless `allowCrossSchemaCascadeDeletes` is true (default false; dangerous).
- STEP 4.5 `sp_refreshview` always honors `excludeSchemas`, and when `includeSchemas` is set only refreshes that list.

## Test plan
- [x] CodeGenLib unit tests for cascade filter, view-refresh scope, heal EXEC shape, include compile vs authored exclude snapshot
- [x] `tsc --noEmit` on CodeGenLib
- [x] PostgreSQLCodeGenProvider + scoped-entity-fields tests
- [ ] Apply the new v6 migration on a SQL Server DB and call each SP with only `@ExcludedSchemaNames` (backward compatible) and with `@IncludedSchemaNames`
- [ ] PG counterpart deferred to release build (`migrations-pg` not authored here)

PostgreSQL `migrations-pg` counterpart is deferred to the release build.

### DIFF
--- a/.changeset/quiet-lenses-heal.md
+++ b/.changeset/quiet-lenses-heal.md
@@ -1,0 +1,6 @@
+---
+"@memberjunction/core": minor
+"@memberjunction/codegen-lib": patch
+---
+
+Add optional `@IncludedSchemaNames` to CodeGen metadata-heal stored procedures so Open App migrations can positively scope heals without photographing sibling apps. Cascade-delete SQL is intra-schema only unless `allowCrossSchemaCascadeDeletes` is set. Custom-view `sp_refreshview` in the migration log honors `excludeSchemas` and, when set, `includeSchemas`.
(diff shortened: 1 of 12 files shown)


AN-BC (commented):
lgtm. merging now. @MS-BC flagging for awareness, this fixes another issues we have in CG with all this x-app linking, getting pretty solid now. Unfortunately this missed edge.4 but I will merge updates to various app repos that will be important to install on AIDP stage because they represent some important clean up. Migrations only I think on the other app PRs.

PR #4076, opened 2026-08-26
fix(messaging-adapters): production bring-up fixes for Slack and Teams

cmbrock-BC:
## Summary

First real deployment of the messaging adapters: one Slack app per agent over Socket Mode, plus Teams via Bot Framework. Everything here is something that broke in that setup, with a test that fails if the fix is reverted.

The `UserCache` one is the most serious. `BaseMessagingAdapter` resolved users through `new UserCache()`. `UserCache` extends `BaseSingleton`, so the constructor hands back the shared instance and then the subclass field initializer wipes it. The adapter got an empty list, `Initialize()` threw `Fallback context user not found` for a perfectly valid email, and no messaging extension could start. Worse, the whole process lost its user cache until the next refresh, so scheduled jobs and anything else sharing it broke too. Both call sites now use `UserCache.Instance`.

## Changes

Startup and identity

* Users resolve through `UserCache.Instance`.
* A bot recognises its own replies. Slack publishes two ids for one bot and returns the `bot_id` (no `user`) for anything posted with a username override, which is every agent reply since per-agent identity is the whole point of one app per agent. Comparing against `auth.test()`'s `user_id` never matched, so the thread gate below declined threads the bot was holding and the agent lost its own turns from context.
* Running one app per agent no longer makes every bot answer the same message. A bot replies to an un-mentioned thread reply only if it was addressed or has already posted there. Other bots' messages stay out of history and thread affinity. New `DisableDelegation` setting for a bot pinned to a single agent.

Delivery

* Generated files and images arrive as real attachments. Adapters can implement `uploadMediaOutputs`; Slack does, and needs the `files:write` scope.
* Files inlined as `data:<mime>;base64,...` get decoded instead of posted as thousands of unusable characters. MJ produces those whenever no file storage account is configured.
* The run's `fileOutputs` are used directly. The old path waited for the model to inline a `data:` URI, which it does inconsistently, so a generated document showed up on one run and not the next.
* A non-public button URL no longer kills the whole Slack message. With a localhost `ExplorerBaseURL`, any reply carrying a resource command used to fail outright.
* Artifact links point at the file the agent produced rather than its internal payload, and `System Only` artifacts are not linked.
* `ng-artifacts`: downloading a file artifact gives you the file, not a `.txt` full of base64.
* `ng-explorer-core`: a conversation deep link opened cold honours the URL instead of restoring whatever you looked at last.

Slack

* Interactivity works in Socket Mode. Every button and modal used to be inert, so a form-based question could not be answered at all.
* Message text is capped at the real `text` limit rather than the block-payload limit, which was failing long replies with `msg_too_long`.
* Modal placeholders and submit labels are truncated. An over-long one fails the entire `views.open` and leaves a button that looks dead, with nothing but a server log to explain it.

Teams

* `MentionedAgentNames` is populated, so naming an agent actually reaches it. Every Teams turn used to run the default agent.
* Form answers go back to the agent that asked, via a validated `mj_agent` field.
* Buttons are only built over `http(s)`. Teams silently ignores `data:`/`blob:`/`file:`, so "Download document" was dead by construction, and it hands unknown schemes like `javascript:` or `ms-msdt:` to the OS. Dropped buttons become a note pointing at the artifact link. Localhost stays allowed.
* Deep links no longer assume `resourceId` is present now that a Record can be keyed by `keys`.

## Verification

`packages/MessagingAdapters`: 370 tests across 12 files. `turbo build` green over the 76-task chain. `check:standards` and `check:changeset` both pass.

The new behaviour is mutation-checked rather than just covered. Reverting the thread gate, the delegation switch, the bot-authored skip, Teams agent routing, the URI allow-list, self-identification or attachment de-duplication each fails the suite.

Also run live: Slack with two apps sharing a channel (one default bot, one pinned to an agent), and Teams across DM, channel and meeting chat.

## Test plan

- [ ] Ask for a document in Slack. One attachment, correct extension, not two copies and not a `.txt` of base64.
- [ ] With two Slack apps in a channel, post an un-mentioned thread reply. Only the bot already in the thread answers.
- [ ] Click a form button in Socket Mode, submit it, confirm the answer reaches the agent that asked.
- [ ] `@mention` a non-default agent in Teams and confirm it routes there.
- [ ] Ask for a document in Teams with no file storage configured. You should get a note plus a working artifact link, not a dead button.
- [ ] Check a `localhost` Explorer link is still clickable from Teams.

## Known gaps

Not fixed here, and worth issues rather than living in a description:

* Teams `fetchThreadHistory` returns `[]` (Graph API TODO), so Teams turns are stateless. Reading MJ's own `ConversationDetail` would be cheaper than Graph.
* DMs, group chats and meetings all map to one MJ conversation until the thread map's 24h TTL resets. Wants a `/new` reset and an idle rollover.
* No per-app agent allow-list. An adapter name-matches every `Active` agent with no permission check, so naming another agent re-routes the turn. Fine for a private channel, not for a shared rollout.
* The thread gate has no "addressed to a different bot" branch, so a bot already in a thread still answers when someone explicitly names a different agent.
* Slack Socket Mode pong timeouts can block API bootstrap, since the extension loader awaits adapter start.
* No adapter coverage in the deterministic integration tier. Everything here is unit-level against hand-built mocks, which is how the self-identification bug got through.


rkihm-BC (commented):
Reviewed the full diff, all 15 commit messages, and the surrounding code the changes depend on. CI is green across the board (Build, 6 unit shards, deterministic integration tier, adopted standards, ESM guard, dependency check). Findings below were each checked against the source rather than inferred; a few things I initially flagged didn't survive that check and aren't listed.

## What's good

**The `UserCache.Instance` fix is the real find, and it was diagnosed properly.** `BaseSingleton`'s constructor returns the stored instance and *then* the subclass field initializer runs against it — so `new UserCache()` both wiped the process-wide cache and read back empty. What raises this above a one-line fix is that the commit also repaired all five test mocks, which had modeled `new` as equivalent to `Instance` and therefore could not have caught it. Fixing the mock semantics rather than just adding a passing test is the right response to a bug that was hiding behind test doubles.

**The allow-list conversion in `isOpenableURI` is the correct call.** Deny-listing `data:`/`blob:`/`file:` and allow-listing `http(s)` are not the same screen, and the reasoning in the JSDoc is right — Teams desktop hands unrecognised schemes to the OS URI handler, so an agent-authored `ms-msdt:` or `javascript:` URL under a deny-list becomes a real button. Same for validating the client-controlled `mj_agent` against `availableAgents` rather than trusting the submit payload.

**MJ conventions are followed carefully throughout** — `EscapeSQLString` on the `ExtraFilter`, the `MJ: ` entity prefix, `RunView<{ Visibility?: MJArtifactEntity['Visibility'] }>` deriving the literal union from the generated ORM (checked: `'Always' | 'System Only'`, exhaustive, so the guard is exact), explicit `Fields` with `ResultType: 'simple'`. `uploadMediaOutputs` as an optional protected method rather than duck-typed off `this` is the right shape.

**De-duplicating `buildExplorerDeepLink` and `isOpenableURI` into `base/message-formatter.ts`** is correct layering, and it collapsed the `resourceId`-optional fix into one place instead of two. Extracting pure functions where Angular decorators block the test transform (`chat-deeplink-params.ts`, `artifact-version-download.ts`, `selectPrimaryArtifact`) is a pragmatic pattern applied consistently.

**Narrowing the Socket Mode subscription to only `interactive`** after verifying against `@slack/socket-mode` 2.0.6 — including noticing that the extra subscriptions would have run the agent twice on a view submission had they fired — is exactly the right depth.

**The docs changes cover the operationally load-bearing bits.** `files:write` and the Socket Mode Interactivity toggle are both "everything renders, nothing works, only clue is a server log" failures; documenting them is worth as much as the code. Single consolidated changeset, all `patch`, matches the repo's changeset rule.

---

## Blocker

### The multi-bot gate and the bot-authored filter both rest on self-identification that this adapter's own posting path breaks

`formatResponse` takes a required non-nullable `agent` and always sets `AgentIdentity` (SlackAdapter.ts:342-361), so `sendFinalMessage` always spreads a `username` override into `chat.postMessage` — every substantive reply, not just the "Thinking…" placeholder. `chat:write.customize` is listed **Yes/required** in the setup guide, and per-agent identity is the point of the one-app-per-agent deployment.

The PR body already states the Slack mechanism: *"A message posted with a `username`/`icon_url` override returns `bot_id` and no `user`."* `fetchThreadHistory` maps `SenderID: msg.user ?? bot_id` (SlackAdapter.ts:163-174), so those history entries carry the `B…` id, while `getBotUserId()` returns the `U…` id from `auth.test()`. They never match.

Two consequences, both in code added by this PR:

- **The gate declines threads the bot is actively holding.** `inThread = threadHistory.some(m => m.SenderID === botUserId)` is always false, so at BaseMessagingAdapter.ts:185-197 an un-mentioned reply in a thread this bot started gets silently dropped — the exact failure the gate's design was trying not to cause.
- **The agent loses its own turns from context.** At BaseMessagingAdapter.ts:1322, `msg.SenderID !== botUserId` (true) `&& isBotAuthored(msg)` (true, `bot_id` present) → `continue`. Before this PR those messages were present, mislabeled as `'user'`. Trading "wrong role" for "absent entirely" isn't obviously an improvement for multi-turn coherence.

Worth noting the failure is *specific* to the customize path: without a `username` override a bot-token post carries both `bot_id` and `user`, so `SenderID === botUserId` holds and both new paths work correctly. That's likely why it wasn't caught — and the tests can't see it either, since `BaseMessagingAdapter.test.ts:389` builds the bot's own history message as `{ SenderID: BOT }` with no `RawEvent`, an identity the adapter never actually produces.

This is listed under Known gaps as "worth a follow-up", but the two new behaviors promote it from cosmetic to functional. The fix is small and belongs here: capture `bot_id` from `auth.test()` alongside `user_id` and treat either as self, in both the gate and `buildConversationMessages`.

---

## Worth fixing — mechanical

**Two class JSDoc blocks were detached, in the same commit that rejoined six others.**
- `AgentRunner.ts:50-82` — `selectPrimaryArtifact` was inserted between the `AgentRunner` class doc (with `@class` and the `@example` usage block) and `export class AgentRunner`. The class now carries no doc comment.
- `BaseMessagingAdapter.ts:100-102` — a blank line between the class JSDoc and `export abstract class BaseMessagingAdapter` detaches it.

**Contradictory stale comment in the Socket Mode wiring.** `SlackMessagingExtension.ts:305-314` still reads *"Both the envelope type and the inner payload types are subscribed… Handlers are idempotent per event, so an overlap costs nothing"*, directly above the newer comment at 331-335 saying the opposite. The final commit changed the behavior and updated one comment but not the other; one of these will mislead the next reader.

**`SLACK_BUTTON_TEXT_MAX_LENGTH = 24` is applied to two different limits.** 24 is correct for a modal view's `submit`, but the `buildResponseForm` ActionSet button is a message button where Slack allows 75 — `buildURLButton` in the same file already uses 75. So agent-authored `submitLabel`s get cut to 21 chars + `...` in the posted message for no reason, and the constant's JSDoc ("fails the whole `views.open`") isn't true for that call site. Cosmetic, but the file currently contradicts itself.

**Two small ones in the same area:** `uploadMediaOutputs` does `files.slice(0, 5)` and drops the rest with no note to the user, which is inconsistent with this PR's own standard elsewhere ("a dropped command becomes a body note") — and entries with no `data` are counted against the 5 before being skipped, so a `mediaOutput` carrying only a URL shrinks the effective budget. And `escapeMrkdwn` guards the `<url|label>` fallbacks but not the `📄 _${cmd.label} — attached to this reply._` note, same agent-authored label in the same markup context.

---

## Worth mentioning, not insisting

**The "attached to this reply" note can promise an attachment that never arrives.** Both the Slack note and Teams' `buildUnopenableResourceNotes` fire on *any* non-`http(s)` scheme, but only base64 `data:` URIs are harvested. `parseBase64DataUrl`'s regex is `/^data:([^;]+);base64,(.+)$/`, so `data:text/csv;charset=utf-8;base64,…` and uppercase `;BASE64,` both fail the parse while `isOpenableURI` still drops the button — as do `blob:`, `file:` and `javascript:`, which never had bytes to harvest. The scheme-case normalisation added in the follow-up closes one of these. To be clear about weighting: MJ's own document actions emit plain `data:<mime>;base64,`, which parses fine, so this is robustness against arbitrary model output rather than an observed failure. If it's worth closing, gating the note on whether a matching attachment was actually collected is more durable than gating on the scheme.

**`selectPrimaryArtifact` silently picks `fileArtifacts[0]`.** With several files the rest never appear in `artifactInfo`. Ordering is deterministic (list order through `Promise.all`) but arbitrary. Scope is narrower than the changeset's wording might suggest — the two consumers are `createCompletionNotification` (gated on `createNotification`) and this bridge; Explorer's in-chat artifact panel builds its own `artifactInfo` from `FindLatestAgentOutputVersion` and is untouched, and both payload and file artifacts still get `MJ: Conversation Detail Artifacts` junction rows either way. So the behavior change is real but contained, and the default is the right one.

**`(cmd as { fileName?: string }).fileName`** reads a field `OpenURLCommand` doesn't declare and that nothing in the repo emits — `actionableCommands` come from model output. Have you actually observed models emitting `fileName` here? If yes it should be declared on the type (additive); if not, it's dead and the `label` + MIME-map path is what runs.

---

## Process

**The `lts/5` backport decision should be made explicitly rather than left in the description.** `new UserCache()` degrades user resolution *process-wide* — starving scheduled jobs and anything else in the process, not just the adapters — so whether it lands on that line is worth a yes or a no on the record.

**The Known-gaps list deserves to become issues.** It's an unusually honest section, but two items are load-bearing for what ships here: Teams' `fetchThreadHistory` returning `[]` is *why* `mj_agent` had to be invented, and "no per-app agent allow-list" combined with `matchAgentMentions` over every `Active` agent means naming another agent re-routes the turn. Fine for a private channel, but they're product behavior now, not TODOs — and a PR description disappears at merge.

**No adapter coverage in the deterministic integration tier.** Nothing under `packages/Integration/` references the messaging adapters; everything here is unit-level against hand-built mocks. That's precisely how the blocker above got through — the mock models an `IncomingMessage` shape the real `fetchThreadHistory` doesn't produce. One case that round-trips a realistic `conversations.replies` payload would have caught it.

---

**Overall:** the engineering standard here is high — every fix is a real production defect with a mutation-checked test, the security follow-ups are correct, and the candour about gaps is well above average. The one thing I'd want resolved before merge is the blocker, because it silently degrades the exact deployment shape this PR exists to enable and the fix is a few lines. Everything else is cheap cleanup or a judgement call that's yours to make.

cmbrock-BC:
Thanks, this was a useful review. Blocker fixed in 5a25d49c68, mechanical items with it, and a further pass in 5cf60596c7.

## Blocker

You're right, and my Known-gaps entry undersold it. I had it filed as conditional on a transient typing-indicator failure. It isn't. `buildAgentIdentity` always returns an identity, so `formatResponse` always sets `AgentIdentity` and every reply goes out with a username override. `inThread` was false every time, not occasionally.

Fixed as an `isSelf()` seam on the base class: single-id compare by default, `SlackAdapter` overrides it to accept either id. Same shape as `isBotAuthored`, so the platform states its own convention instead of the neutral base guessing. Three call sites moved onto it.

Your point about the tests was the more useful half. The `auth.test()` mock returned only `user_id`, and the bot's history entry was hand-built as `{ SenderID: BOT }` with no `RawEvent`, which is not an identity the adapter ever produces. Both now model what Slack returns, and reverting `isSelf` fails the new test.

## Mechanical

All done. Both class JSDoc blocks rejoined. Stale Socket Mode comment replaced. `SLACK_MODAL_SUBMIT_MAX_LENGTH` (24) and `SLACK_BUTTON_TEXT_MAX_LENGTH` (75) are now separate, and the four bare `75` literals you spotted use the constant. Upload cap filters for deliverable bytes before slicing. `escapeMrkdwn` covers the note.

## Your two questions

`fileName` is dead. Nothing in the repo sets it on a command and the model-facing contract never mentions the field, so the model is never told it exists. Removed; the label plus MIME-map path is what runs.

The "attached" note now fires only when the URI will actually yield bytes, using the same parse the collector uses, and says "not linkable from Slack" otherwise. Gating on collection rather than scheme, as you suggested.

`selectPrimaryArtifact` picking `[0]` I've left, on your own reasoning: deterministic, two consumers, right default. Easy to revisit if a multi-file run makes the choice matter.

## Second pass, including two I broke

I reviewed my own fixes before asking you to look again, which caught two regressions the blocker commit introduced:

* `truncateToLength` measured in UTF-16 units but sliced by code point, so 24 emoji became 45 units against a 24 limit. That reintroduces the `views.open` failure the constant exists to prevent. I took the surrogate-pair note at face value without noticing the guard and the slice used different units.
* `escapeMrkdwn(cmd.label)` had no undefined guard on the four paths I added, while every other label site in that file defends. `label` is declared required but the value is model-authored JSON, and a throw there propagates out of `HandleMessage`, so the agent runs and the user sees nothing.

Both mutation-checked now. Same pass also fixed a prototype-unsafe MIME lookup (`mimeType` is agent-controlled, so `EXTENSION_BY_MIME_TYPE['toString']` returned a function), a silent drop in `buildArtifactCard`, a malformed `data:` URI throwing into a catch that only logs, and a `respondsToUnaddressedThreadReplies` doc comment that asked the inverse of what the method returns.

## Left alone, on purpose

* The gate has no "addressed to a different bot" branch. If bot B is in a thread and someone posts `@Alpha continue`, B answers too. That's the gate's design rather than a regression, but it's a real double-answer path and probably belongs with the allow-list work.
* `inThread` reads a truncated 50-message window, so a bot that joined a long thread late can go quiet.
* `ProcessFileArtifacts` runs `processFileOutput` concurrently against one shared provider, so nested transactions mean one file's rollback can discard another's artifact.
* The Explorer wrapper reads `window.location.search` directly, against the NavigationService rule in `packages/Angular/Explorer/CLAUDE.md`. Happy to move it to `SwitchToApp(appId, navItem, queryParams)` here rather than as a follow-up if you'd prefer.

## Process

Agreed on the Known gaps becoming issues, and the two you singled out are the right ones. A description does vanish at merge, so I've folded them plus the items above into the Known gaps section for now.

The integration-tier point is the sharpest thing in the review. The blocker got through precisely because every test asserts against a hand-built `IncomingMessage` that `fetchThreadHistory` never produces. One case round-tripping a real `conversations.replies` payload, including a `bot_message` with `bot_id` and no `user`, would have caught it and would catch the next one.

rkihm-BC (approved):
LGTM

PR #4077, opened 2026-08-26
Drop the orphaned ReportCategory root-id function that breaks Azure bacpac export

cmbrock-BC:
## Summary

The `ReportCategory` table was dropped in an earlier version, but `fnReportCategoryParentID_GetRootID` — an inline table-valued function that depends on it — was left behind. Azure SQL's database export (bacpac) validates every module and fails on the dangling reference, so any installation carrying the orphan cannot be exported.

## Changes

One migration, existence-guarded:

```sql
IF OBJECT_ID('[${flyway:defaultSchema}].[fnReportCategoryParentID_GetRootID]', 'IF') IS NOT NULL
    DROP FUNCTION [${flyway:defaultSchema}].[fnReportCategoryParentID_GetRootID];
```

- No-op on a fresh install and on any database where it was already removed by hand (it has been run manually on at least one production database ahead of this PR — the guard makes that safe).
- Zero references to the function anywhere in `packages/`; its only appearances in migration history are the v5.0 baseline (which carries it in) and the v2 view-regeneration sweep.
- T-SQL only, per the migration policy — the PG counterpart is toolchain territory.

Found by Dray during database-refresh work; this PR upstreams the fix.


rkihm-BC (approved):
Approving. The orphan is real, the scope is right, and the guard is correct. One mechanical thing to do immediately before merge — details at the end.

## Verified rather than taken on trust

I traced the whole chain against `origin/next`, because "this object is orphaned" is exactly the claim that's expensive to be wrong about:

- **The table really is gone.** `migrations/v6/V202608061704__v6.1.x__Phase0_Legacy_Workflow_Report_ScheduledAction_Retirement.sql` does `DROP VIEW IF EXISTS [__mj].[vwReportCategories]` (line 180) and `DROP TABLE IF EXISTS [__mj].[ReportCategory]` (line 245).
- **The entity really is gone.** `MJReportCategorySchema` has zero occurrences in the generated ORM on `next` — so nothing reads it through the framework either.
- **Nothing references the function.** Its only remaining appearances are historical: the v5 baselines that carry it in, and the v2 view-regeneration sweep.
- **`'IF'` is the right type code.** This is an inline table-valued function (`RETURNS TABLE AS RETURN (...)`), not scalar (`FN`) or multi-statement (`TF`). Wrong code here would make the guard silently never fire and the migration a no-op that looks like it worked.

## The gap this exposes is worth knowing about

The Phase0 retirement dropped eleven tables and their views, and it correctly dropped the CRUD procedures too — `spCreateReport`, `spDeleteReportCategory`, `spUpdateReportVersion` and the rest, a long and complete list. But it contains **zero** `_GetRootID` mentions. So the retirement checklist covers tables, views and procs, and does not cover recursive-FK table-valued functions.

I checked whether that leaves other orphans. Of the eleven retired tables, only `ReportCategory` had a self-referencing foreign key, so `fnReportCategoryParentID_GetRootID` is the only TVF stranded by that migration — **this PR is complete as written, nothing to add.**

Worth raising separately if more Phase-N retirements are coming: CodeGen generates `fn{Table}{Field}_GetRootID` for every self-referencing FK (`detectRecursiveForeignKeys`, unchanged), but it only ever *creates* objects for entities it knows about — there's no sweep that drops objects belonging to entities that no longer exist. So a retired table with a `ParentID` will strand its TVF every time, and the failure mode stays invisible until someone attempts a bacpac export. That's a checklist gap, not a problem with this PR.

## Before merge — the timestamp needs re-dating

`Check migrations` shows green, but that green is **stale**:

- the check ran at `2026-08-30T00:45:58Z`
- `V202608300200__…` landed on `next` at `2026-08-30T03:59:22Z` — about three hours later

So the run was evaluated against a `next` that did not yet contain it. As things stand, `V202608261500` is behind `next`'s current max and a fresh run would fail on ordering.

Needs re-dating above whatever `next`'s max is at the moment of merge — and, as ever, immediately before merging rather than now, since the target keeps moving. Nothing else in the diff hardcodes the filename, so it's a one-file rename.

Changeset `minor` is correct for a migration branch.

PR #4079, opened 2026-08-26
feat(server): auto-load Open App serverExtensions from installed packages

AN-BC:
## Summary
- Open App server packages listed in the host `mj.config.cjs` `dynamicPackages.server[]` already import at boot (ClassFactory + `RESOLVER_PATHS`). `serve()` still only initialized host `serverExtensions[]`, so every Open App webhook/checkout edge had to be copied into the host config.
- Packages now declare extensions via the `MJ_SERVER_EXTENSIONS` export, with `package.json` `memberjunction.serverExtensions` as the static fallback. Bootstrap collects them; `serve()` merges with the host list. **Host `DriverClass` wins** (`Enabled`, `RootPath`, per-key `Settings`). `Enabled: false` on the host disables a discovered extension rather than falling back to it. Host-only extensions (Slack, Teams) are unchanged.
- `dynamicPackages` is stripped from MJServer's Zod `configInfo`, so discovery stays in `@memberjunction/server-bootstrap` (raw cosmiconfig) — the same place resolver paths are collected.

Companion Orders PR publishes `MJ_SERVER_EXTENSIONS` on `@mj-biz-apps/orders-server` and adds `GET /checkout/:slug`.

## Test plan
- [x] `packages/ServerExtensionsCore` — normalize / extract / merge (empty, invalid, host overlay, disable-from-host, last-wins, Settings merge)
- [x] `packages/ServerBootstrap` — enabled app export reaches `serve()` options; disabled app does not; package.json fallback; host `serverExtensions` are **not** merged in bootstrap; missing package still boots
- [ ] After this lands, a host with `@mj-biz-apps/orders-server` in `dynamicPackages.server[]` loads `OrdersCheckoutEdge` without copying that block into host `serverExtensions`

## Notes
Omnibus branch `an-dev-70`. Does not generalize `/f` + `/c` into a core slug-host framework.


AN-BC:
## Defensive security review — Open App `serverExtensions` auto-load

Reviewed alongside the companion PR that consumes this: [MemberJunction/bizapps-orders#115](https://github.com/MemberJunction/bizapps-orders/pull/115) — findings for the anonymous checkout edge itself are in [that PR's review comment](https://github.com/MemberJunction/bizapps-orders/pull/115#issuecomment-5431089502).

**First, what this PR does not do.** I want to head off the obvious objection: this is *not* a new arbitrary-code-execution vector. Packages listed in `dynamicPackages.server[]` are already `import()`ed by `loadDynamicAppPackages`, and their `StartupExport` is already invoked (`ServerBootstrap/src/index.ts:184`). A package that wanted to run code at boot could already do so. The collect/normalize/merge code is clean, the invalid-entry handling is non-throwing so one bad declaration cannot take down boot, and the test coverage on `mergeServerExtensionConfigs` is genuinely thorough.

What the PR *does* change is **consent** — and for a surface that mounts **before the auth middleware**, consent is the security-relevant axis.

---

### 🟠 F-M1 — Pre-auth HTTP surface becomes opt-**out** rather than opt-**in**

`packages/MJServer/src/index.ts:1287` · `packages/ServerBootstrap/src/index.ts` — `loadDynamicAppPackages()`

Before this PR, an operator had to **explicitly paste** a `serverExtensions` block into the host `mj.config.cjs` for an extension's routes to mount. That paste was the consent step. After this PR, merely listing a package auto-mounts every route its `MJ_SERVER_EXTENSIONS` declares — webhooks and anonymous checkout included — and per the comment at `index.ts:1283` these mount *ahead of* `createUnifiedAuthMiddleware`.

To suppress one, the operator must already know its `DriverClass` in order to write `Enabled: false`. There is no disable-by-default, and no boot-time inventory of what got mounted. So an operator installing an Open App for its entities and resolvers silently also exposes that app's unauthenticated HTTP attack surface, having never been shown it.

Three additive fixes, cheapest first:

1. **Log every auto-mounted extension at startup** — `DriverClass`, resolved `RootPath`, and explicitly that it is **mounted PRE-AUTH**. Right now `loadDynamicAppPackages` logs only a `(+N server extensions)` count, which names nothing. Making the surface *visible* is most of the value and costs one line.
2. **Let a package ship a sensitive edge `Enabled: false` by default.** The merge already honors a host `Enabled: true` overlay, so opt-in works today with no code change — it just needs to be the documented convention for pre-auth edges.
3. Optionally, a host-side allowlist or one-time acknowledgment for auto-mounted pre-auth extensions.

### 🟠 F-M2 — No `RootPath` validation or collision detection

`packages/ServerExtensionsCore/src/collect.ts:71` — `normalizeServerExtensionConfigs()`

Normalization trims `RootPath` and checks it is non-empty; that is the whole check. A discovered extension can declare `RootPath: '/'`, `'/graphql'`, or an auth-catalog path and it will mount pre-auth, ahead of the core routes, silently shadowing them. `mergeServerExtensionConfigs` dedupes by `DriverClass` but says nothing about two *different* classes claiming overlapping paths.

Given packages are already trusted this is operator-safety / defense-in-depth rather than a privilege boundary — but it is exactly the class of mistake that should fail loudly instead of producing a server that looks fine and isn't.

Fix: validate shape (leading slash, no wildcard, sane length), reject-or-warn on a reserved-prefix set (`/`, `/graphql`, the auth-catalog mount, `/health/extensions`), and warn when two enabled extensions resolve to overlapping roots. Fail closed on an invalid path rather than mounting it.

---

### Worth keeping

`Settings: {}` in the published manifest means no secrets ride the npm tarball — the Turnstile secret and service principal come from host `Settings`, which is the right split. Host `Enabled: false` being *preserved* rather than overwritten by the discovered default is the correct precedence, and is what makes fix (2) above a documentation change rather than a code change. Keeping discovery in `server-bootstrap` (raw cosmiconfig) rather than widening MJServer's Zod `configInfo` is also the right call.

Happy to re-review once these land — I'm watching this PR.

---

AN-BC:
Addressed the defensive review on `36fcc6e26a`. Summary of how each finding landed:

### F-M1 — pre-auth became opt-out
- **Inventory at boot (done).** Every discovered extension is logged from bootstrap with `DriverClass`, `RootPath`, enabled/disabled, and an explicit **PRE-AUTH** note, plus the same line again after host merge in `serve()` so Slack/Teams overlays are visible too. Example from a live boot:

  `OrdersCheckoutEdge at /checkout (enabled, PRE-AUTH; host mj.config.cjs serverExtensions[] can set Enabled: false to suppress)`

- **Host `Enabled: false` remains the suppress path** (already in merge, now documented on the log line and in the ServerExtensionsCore README).
- We did **not** flip Open App manifests to `Enabled: false` by default. Auto-mount without copying blocks into the host `mj.config.cjs` is the product intent of this PR; making pre-auth edges opt-in again would undo it. Operators who want that posture can still disable by DriverClass. A host-side allowlist (your optional #3) is a follow-up if we want a second consent gate.

### F-M2 — RootPath validation / collisions
Done in `prepareServerExtensionConfigs` / `validateServerExtensionRootPath` (unit-tested):
- Must start with `/`, no glob/`*`/`[]`/`()`, max 128 chars.
- Fail-closed drop of reserved exact `/` and prefixes `/graphql`, `/auth`, `/oauth`, `/health`, `/magic-link` (prefix-with-slash so `/healthcare` is not a false positive).
- Warn when two **enabled** extensions overlap (`/webhooks` vs `/webhooks/payments`). Disabled entries stay in the list so host `Enabled: false` remains visible to the loader.

Happy to go another round.

AN-BC:
## Re-review of `36fcc6e2` — F-M1 ✅, F-M2 ⚠️ two bypasses

Verified by reading the diff (no CI or test run). Follow-up to [my earlier review](https://github.com/MemberJunction/MJ/pull/4079#issuecomment-5431093626).

**F-M1 (pre-auth inventory) — resolved.** `describeServerExtensionMount` names the DriverClass, the resolved RootPath, enabled/disabled, `PRE-AUTH`, *and* how to suppress it. Logged both at discovery in `ServerBootstrap` and post-merge in `serve()`, including disabled entries. That's the consent surface I asked for, and better than what I proposed.

**F-M2 (RootPath validation) — right shape, but the reserved check can be walked around two ways.** Running `validateServerExtensionRootPath` in *both* `normalizeServerExtensionConfigs` and post-merge `prepareServerExtensionConfigs` is the correct placement — host entries never pass through normalize (they're cast straight from `configInfo`), so only the post-merge pass catches a bad host RootPath. Good catch. Two gaps in the matching itself:

---

### 🔴 Case-sensitive check, case-insensitive router

`packages/ServerExtensionsCore/src/collect.ts` — `validateServerExtensionRootPath`

The reserved comparison is case-sensitive:
```ts
if (normalized === prefix || normalized.startsWith(`${prefix}/`))
```
but `MJServer/src/index.ts` never sets `app.set('case sensitive routing', true)`, so **Express matches routes case-insensitively by default**.

`RootPath: '/Auth'` therefore passes validation — `'/Auth' !== '/auth'` and it doesn't start with `'/auth/'` — and then Express happily serves it for requests to `/auth`, mounting an Open App extension **pre-auth, on top of `AUTH_CATALOG_MOUNT_PATH`** (`/auth`, `index.ts:1296`). Same trick with `/GraphQL`, `/OAuth`, `/Health`, `/Magic-Link`. One character of casing defeats the whole reserved list.

```ts
const normalized = normalizeRoot(raw).toLowerCase();   // reserved lists are already lowercase
```
(Keep the original casing for the mounted `RootPath` and the log line — lowercase only for the comparison.)

### 🟠 The reserved list misses most of the core pre-auth mounts

`/health` is checked as an exact-or-`/health/` prefix — the code comments that this deliberately spares `/healthcare`. It also spares **`/healthcheck`**, which is a real core route (`index.ts:1122`). `validateServerExtensionRootPath('/healthcheck')` returns `null` today.

Enumerating what `serve()` actually mounts, these are claimable by an extension right now:

| Core mount | `index.ts` | in reserved list? |
|---|---|---|
| `/healthcheck` | 1122 | ❌ |
| `/esignature` | 1156 | ❌ |
| `/media` | 1163 | ❌ |
| `WIDGET_MOUNT_PATH` | 1186, 1327 | ❌ |
| `TWILIO_TELEPHONY_MOUNT_PATH` | 1205 | ❌ |
| `VONAGE_TELEPHONY_MOUNT_PATH` | 1217 | ❌ |
| `TEAMS_MEETINGS_MOUNT_PATH` (`/meetings/teams`) | 1244 | ❌ |
| `/auth`, `/oauth`, `/magic-link`, `/health/extensions` | — | ✅ |

Also worth noting: `graphqlRootPath` defaults to `'/'` (`config.ts:633`), so the literal `/graphql` entry only protects deployments that set `GRAPHQL_ROOT_PATH=/graphql` — and a deployment that sets it to anything else (`/api`) gets no protection at all, because the list is a static literal rather than a function of the running config.

**Suggestion:** derive the reserved set from what the server actually mounts rather than restating it. Export the `*_MOUNT_PATH` constants (several already are) plus `configInfo.graphqlRootPath` into one `CORE_RESERVED_ROOTS` array that `serve()` passes into `prepareServerExtensionConfigs`. A hardcoded list in `collect.ts` cannot see a route added in `index.ts` next quarter, and nothing will fail when they drift — which is the same class of problem F-M1 was about: the failure is invisible.

If deriving it is too big for this PR, adding the seven paths above plus the `.toLowerCase()` closes what's exploitable today; the drift risk is worth a follow-up issue either way.

---

Nothing else in `36fcc6e2` concerns me — fail-closed drop on invalid roots, overlap warning scoped to enabled entries only, and disabled entries preserved so `Enabled: false` stays visible to the loader are all correct.

---

AN-BC:
Addendum to [my re-review](https://github.com/MemberJunction/MJ/pull/4079#issuecomment-5431356441) — CI went green on `36fcc6e2` with both bypasses live, so here are the two cases that would turn them red.

The new `validateServerExtensionRootPath` suite is good, and it deliberately pins the false-positive boundary (`/healthcare` → `null`, `/authorize-me` → `null`). But every reserved-prefix assertion spells the prefix exactly as it appears in the list, so nothing exercises the two gaps:

```ts
// case bypass — Express matches case-insensitively; the check compares case-sensitively
expect(validateServerExtensionRootPath('/Auth')).toMatch(/reserved prefix/);
expect(validateServerExtensionRootPath('/GraphQL')).toMatch(/reserved prefix/);

// /healthcare is correctly allowed — but /healthcheck is a real core route (index.ts:1122)
expect(validateServerExtensionRootPath('/healthcheck')).toMatch(/reserved/);
```

Both currently return `null`.

The `/healthcare` assertion is the interesting one: it is testing the right thing and passing for the right reason, and the identical string shape is what lets `/healthcheck` through. That's the shape worth watching generally — a test that proves a boundary works without noticing what the boundary admits.

---

AN-BC:
Re-review of `36fcc6e2` addressed on `481c2c8ea9`.

### Case-sensitive reserved check vs case-insensitive Express — closed

`validateServerExtensionRootPath` and `serverExtensionRootsOverlap` now compare **folded** paths. `/Auth`, `/GraphQL`, `/OAuth`, `/Health`, `/Magic-Link` are rejected against the static baseline without any extra list. Mounted `RootPath` and the boot log keep the original casing.

### Reserved list missing core mounts / graphqlRootPath — closed by deriving it from `serve()`

Rather than restating more literals in `collect.ts` (which would drift the same way), `serve()` now passes `coreReservedServerExtensionRoots(graphqlRootPath)` as `extraReservedRoots`. That set is built from:

- `graphqlRootPath` (so `GRAPHQL_ROOT_PATH=/api` is reserved, not just the static `/graphql`)
- `AUTH_CATALOG_MOUNT_PATH`, `MAGIC_LINK_MOUNT_PATH`, `WIDGET_MOUNT_PATH`, `TWILIO_TELEPHONY_MOUNT_PATH`, `VONAGE_TELEPHONY_MOUNT_PATH`, `TEAMS_MEETINGS_MOUNT_PATH`
- the literal pre-auth paths that have no constant today: `/healthcheck`, `/esignature`, `/media`, `/oauth`, `/health`

Matching is **bidirectional**, so an extension at `/telephony` collides with `/telephony/twilio` instead of prefix-shadowing it. `/healthcare` still passes; `/healthcheck` does not.

A source-contract unit test reads the `export const *_MOUNT_PATH = '...'` lines as text (so it never loads `config.ts`) and fails if a constant drifts from the reserved list.

Tests: ServerExtensionsCore 63, MJServer reserved-roots 5, ServerBootstrap 38. Live boot still inventories `OrdersCheckoutEdge at /checkout (enabled, PRE-AUTH)` and the conference GET/draft path still works.

Happy to go another round.

AN-BC:
## `481c2c8e` verified — F-M2 closed. No further security findings.

Traced both bypasses through the new code rather than reading the summary.

**Case bypass — closed.** `foldRoot()` lowercases after normalizing, and both `validateServerExtensionRootPath` and `serverExtensionRootsOverlap` compare folded values. `/Auth` → `/auth` → equals the baseline prefix → rejected. Original casing is preserved for the mount and the boot log, which is the right split.

**Reserved list — closed, and closed the right way.** Deriving `extraReservedRoots` from `coreReservedServerExtensionRoots(graphqlRootPath)` in `serve()` rather than adding more literals to `collect.ts` is the better of the two options I offered — the drift I was worried about is now a test failure instead of a silent gap. `/healthcheck` is rejected via the extras; `/healthcare` still passes; bidirectional matching means `/telephony` now collides with `/telephony/twilio` instead of shadowing it.

**Three things I checked specifically because they could have passed while being wrong:**

1. **`graphqlRootPath` defaults to `/`** (`config.ts:633`), so the extras include `/` on a default deployment. Your comment says it "does not prefix-match every path" — that's correct, and I verified it rather than taking it: `serverExtensionRootsOverlap('/checkout', '/')` compares `'/checkout'.startsWith('//')` and `'/'.startsWith('/checkout/')`, both false. A naive `startsWith(nb)` would have rejected every extension on a default install, including `OrdersCheckoutEdge`. The `normalizeRoot` special-case for `/` is what makes that safe.

2. **The source-contract test is a real gate.** `readMountPath()` throws when the regex finds no `export const NAME = '...'`, so a renamed or reformatted constant fails loudly instead of quietly matching nothing — the failure mode I'd expect from a text-scraping test. I also independently confirmed all six values against source: `/auth`, `/magic-link`, `/widget`, `/telephony/twilio`, `/telephony/vonage`, `/meetings/teams`. All match the literals.

3. **Discovery vs. mount.** `normalizeServerExtensionConfigs` applies only the static baseline, so `/healthcheck` survives discovery and is caught later by `prepareServerExtensionConfigs`. Since `prepare()` is the only path into `LoadExtensions`, nothing unsafe can mount — correct as written, just worth knowing the two passes aren't equivalent if that ordering ever changes.

Both regression tests I suggested are in the suite and assert the right direction.

F-M1 was already resolved in `36fcc6e2`. **Nothing outstanding from me on this PR.**

---

AN-BC:
@MS-BC @rkihm-BC @bc-izygmunt FYI

Issue #116, opened 2024-04-09
Folders within a view of resources like Reports/Views need to appear at the TOP of the screen


AN-BC:
Right now they appear in alpha order. Folders should always appear on the top. 

Also, there is no way to move a resource into a folder. We need a way to move a view/report/dashboard into a folder by drag and drop and also perhaps by a "..." button where you have options for the resource to edit/delete/move etc

@JS-BC  fix the sorting problem first as that is easy, and we can work up ways to move a resource

PR #4080, opened 2026-08-27
fix(explorer): load @angular/compiler so partial workspace libs run under ng serve

AN-BC:
## Summary
- `ng serve` excludes `@memberjunction/*` from Vite prebundling, so partially compiled libraries keep `ɵɵngDeclareComponent`.
- Import `@angular/compiler` before bootstrap (and declare it on `mj_explorer`) so JIT fallback works for those packages.
- Fixes Explorer console: `The component 'CredentialEditPanelComponent' needs to be compiled using the JIT compiler, but '@angular/compiler' is not available.`

Follow-up to #4079 on the same branch (`an-dev-70`).

## Test plan
- [x] `ng serve` on port 4201 (Auth0 callback) and hard-refresh Explorer
- [x] Confirm the CredentialEditPanel JIT error is gone from the console
- [ ] Spot-check login + a record form still load


AN-BC:
Closing this. Putting `@angular/compiler` in MJExplorer was the wrong fix.

The CredentialEditPanel JIT message is Angular's *fallback* hint when a partially compiled library was not linker-processed. Explorer is AOT; the compiler belongs in vitest DOM setup (`vitest.dom.setup.ts`), not the app bootstrap. Adding it here has bitten us before (JIT-everywhere, CSP / unsafe-eval, bundle noise).

Reverted on `an-dev-70`. Orders work stays on #116.

PR #4081, opened 2026-08-27
fix(codegen): keep disambiguating a colliding entity name until it is free

MS-BC:
## The bug

Entity names are generated from the table name with trailing discriminators stripped, so distinct tables routinely generate the **same** name. When that happened, CodeGen appended the schema name once and assumed the result was unique:

```ts
if (existingEntity || existingEntityInNewEntityList) {
   suffix = '__' + newEntity.SchemaName;
   newEntityName = newEntityName + suffix;   // never re-checked
}
```

It is not unique. On a NetSuite catalog, `customlist72`, `customlist74`, `customlist160`, `customlist436`, `customlist534` and `customlist873` all generate **"Custom Lists"**:

Every one past the second dies on `UQ_Entity_Name` and is never created.

**The failure is silent where anyone would look for it.** CodeGen continues and emits nothing but a repeated identical INSERT error, so an install ends up short several entities with no indication of which, or why. Eight entities were lost this way on one catalog before anyone noticed the count was wrong.

## The fix

Disambiguation continues past the schema suffix with a counter until the name is actually free — bounded at 1000, so a schema where everything collapses to one name still fails loudly at the insert rather than hanging. A naming problem that severe deserves an error, not a silent 1000-deep suffix.

Name comparison is now **case-insensitive on both sides**, matching the collation `UQ_Entity_Name` is enforced under. The in-run check used an exact `===` while the metadata check immediately beside it lowercased — so two names differing only in case read as free here and then collided on insert. That was a second, independent way to hit the same duplicate-key error.

The logic is extracted as `ManageMetadataBase.resolveUniqueEntityName(desiredName, schemaName, takenNames)` so it can be tested without constructing the 6000-line class.

`suffix` still feeds `NameSuffix` and the base-view name (`vw{Plural}{suffix}`); base views derive from the **table** name, which is already unique, so extending the suffix cannot collide there.

## Tests

8 new tests. **Reverting the loop fails 3 of them**, including the one that walks the six real colliding tables through in sequence and asserts six distinct names.

Dialect-agnostic — this is name resolution in TypeScript, with no SQL emitted differently for SQL Server or PostgreSQL.


rkihm-BC (approved):
## Review — approved

Read the full diff and verified the load-bearing claims against the branch rather than the summary. The diagnosis is right, the fix is right, and the tests are the kind that fail when the fix is removed.

### Verified

- **The base-view claim holds.** The PR asserts that extending the suffix past `__<schema>` can't collide the base view. Confirmed at `manage-metadata.ts:6608` — the name is built from `generatePluralName(newEntity.TableName)` plus the suffix, and table names are unique within a schema. Worth having checked rather than accepted, since the suffix now varies where it previously didn't.
- **`newEntityList` is still fed correctly.** `newEntityName = uniqueName` is assigned before the existing `ManageMetadataBase.newEntityList.push(newEntityName)`, so each resolved name is claimed for the rest of the run. That's what makes the six-table sequence test a real test and not a simulation.
- **The case-sensitivity half is an independent second bug.** The in-run check used `===` while the metadata check immediately beside it lowercased, so names differing only in case read as free and then died on `UQ_Entity_Name`. Finding that while fixing the first bug is the good kind of thoroughness, and it has its own test.

### Optional, none blocking

Raise these only if you're touching the file again anyway:

1. **Two sources of truth for "taken".** `resolveUniqueEntityName(desiredName, schemaName, takenNames)` reads the static `ManageMetadataBase.newEntityList` internally *as well as* the array it's handed — which is why the tests reset the static in `beforeEach`. A line in the `@param` doc saying so would save the next reader the detour.
2. **Exhausting the bound returns a name it knows is taken.** Deliberate and documented ("let the INSERT fail loudly"), but the resulting error is the same opaque duplicate-key message this PR exists to eliminate. A `LogError` naming the desired name and schema before returning would keep the diagnosis that the rest of the change is about.
3. **The taken-set is rebuilt per new entity.** `md.Entities.map(e => e.Name)` allocates and then gets lowercased into a `Set`, once per new entity — O(N·M) on exactly the catalog sizes this PR targets. Hoisting it and adding to it as names are claimed would be strictly cheaper.

### One thing worth adding to the changeset

"Eight entities were lost this way on one catalog" is the motivating incident, and the first question anyone hitting this will ask is what happens to the tenant that already lost them. The fix only affects newly created entities and renames nothing existing — the tables are still there and the entities aren't, so a re-run should create them. Worth stating explicitly, because it's the operator-facing half of the fix.

Approving.

PR #4084, opened 2026-08-27
docs(plans): plan for first-class polymorphic foreign keys (EntityID/RecordID)

AN-BC:
## Plan only — no code changes

Adds `plans/polymorphic-foreign-keys.md`. Nothing is implemented here; this is the design for review before Layer 0 lands.

## The problem

MJ models "a pointer to any record" as an `EntityID` + `RecordID` column pair. The framework understands the **`EntityID`** half — it's a real FK to `__mj.Entity`, so it gets a generated index, a joined `Entity` name column in the base view, and referential integrity. It understands **nothing** about the `RecordID` half: doesn't know the column is a pointer, doesn't know what it points at, doesn't validate it, doesn't cascade it, doesn't index it usefully.

A scan of MJ core plus the Open App repos found **~53 such pairs across ~50 tables in 10 repositories**, including 37 of MJ's own generated entity classes. Full inventory is in the appendix of the plan.

The finding that shaped everything: **the declaration slot already exists.** `EntityField.EntityIDFieldName` has been in the schema since the v2 baseline, is documented for exactly this purpose ("*preventing orphaned soft-linked records during delete operations*"), and has consumer code in both DB providers. It is `NULL` for every field in the system, and its consumer has two bugs. We're finishing something started around v2, not inventing a concept.

## Decisions in the plan

| | Decision | Why |
|---|---|---|
| **D1** | Canonical encoding is `CompositeKey.ToConcatenatedString()` — `ID\|<guid>` | Two incompatible encodings ship in core today (`RecordChange` vs `RecordGeoCode`), and consumers sniff for a `\|` to guess which they got — which is wrong for a bare composite key |
| **D2** | Canonical width is `NVARCHAR(750)` | 450 is *not* the ceiling — that's the clustered/pre-2016 limit; nonclustered is 1700 bytes (850 chars) on 2016+, and core already ships two indexes over 900 bytes. The deciding argument is migration direction: **750 narrows nothing**, 450 would require narrowing ten tables |
| **D3** | `EntityField.EntityIDFieldName` is the declaration predicate | It already exists on the payload row, names its discriminator sibling, and is what both providers already test for. `!= NULL` means "this field is a polymorphic pointer" |
| **D4** | Policy as JSON on the field; allow-list as a **table** | `EmbeddedRecord` is the migration template. The allow-list is a table for the same reason `EntityOrganicKeyRelatedEntity` is — the target becomes a real FK, so a typo'd entity is impossible rather than silent |

## Scope

- **Layer 0 — settle the encoding and shape.** Committed. Ships first because it's the only layer that touches *data*, and it gets more expensive every quarter.
- **Layer 1 — declare it.** Committed. Populate the metadata, add the `additionalSchemaInfo` block, fix the four bugs below.
- **Layers 2 (enforce) and 3 (use).** Sketched and deliberately deferred — recorded so the shape is agreed before Layer 1 constrains it.

## Bugs found along the way

Four defects in the existing, unpopulated soft-link path — recorded in the plan, to fix as part of Layer 1:

- **B1** — `BuildSoftLinkDependencySQL` filters on `entity.ID` where `entity` is the *holder* of the link, not the entity of the record whose dependencies are sought. It looks for `TaskLink` rows pointing at `TaskLink`. Same shape in the PG provider.
- **B2** — `GetRecordDependencies` returns early when no *hard* FK dependents exist, so the soft-link SQL is never built for an entity whose only dependents are polymorphic — the exact case it exists for.
- **B3/B4** — encoding assumption, and quoting derived from the wrong entity's PK type.

Separately, `recent-access.service.ts` mis-parses a bare composite `RecordID` (`val1||val2` → `Value=undefined`). Resolved by D1 + Layer 0.

## ⚠️ Caliber needs a fix before 1.0

`bizapps-caliber` uses `SubjectEntityName NVARCHAR(255)` + `SubjectID` across **seven tables**. Because the discriminator is a *name* rather than an FK to `__mj.Entity`, Caliber gets none of what MJ does provide — no joined name column, no referential guarantee, silent break on entity rename — and it forces the duplicated `sqlLiteral` escaping the team already flagged in a code comment.

**Caliber's npm packages are at `0.1.0` and it ships no `Metadata_Sync` migration yet**, so nothing is installed at a customer, the publish-then-no-breaking-changes policy does not bind, and this is a cheap forward migration over seven tables. **After 1.0 publishes it becomes a breaking schema change and the cost changes category.** Details and the two-part fix are in the plan.

## Open questions for reviewers

1. **750 vs 450** — is there a reason to prefer 450 that's worth a ten-table narrowing project?
2. **Layer 0 sequencing** — `RecordGeoCode` data migration and the widen: one migration or two?
3. **`OnTargetDelete` default** — `Ignore` preserves today's behaviour; is `SetNull` better for *newly declared* links given `Pairing` already governs population?
4. **Empty `AllowedEntities`** — should a declared-but-empty list warn at CodeGen time, so "never got round to constraining this" is visible rather than silent?

## Testing

No code changed. `npm run check:claude-md` passes (the two reported items are the pre-existing `../CLAUDE.md` M5-workspace warning, unrelated to this branch). The changeset gate fires only on migrations, so none is required.

https://claude.ai/code/session_013vVRDyXSLsjTvtVy6KDzYc

---
_Generated by [Claude Code](https://claude.ai/code/session_013vVRDyXSLsjTvtVy6KDzYc)_


AN-BC:
@AN-BC @rkihm-bc @ms-bc — plan is up for review. No code, just `plans/polymorphic-foreign-keys.md`.

**The time-boxed item, flagged separately because it has a deadline the rest of the plan doesn't:**

⚠️ **`bizapps-caliber` must be fixed before Caliber 1.0 publishes.**

Caliber's subject link is `SubjectEntityName NVARCHAR(255)` + `SubjectID` across seven tables — `Engagement`, `AssessmentSession`, `AssessmentOutcome`, `IntakeSubmission`, `InviteScope`, `EntityBindingRecord`, `Note`. Because the discriminator is a **name** rather than an FK to `__mj.Entity`, Caliber gets none of what MJ provides for the `EntityID` half: no joined `Entity` name column, no referential guarantee, and a silent break on any entity rename. It also forces the manual `sqlLiteral` escaping the team already flagged in a comment — deliberately duplicated across two packages because the no-cross-package-re-export rule left no better option. A typed link needs none of that.

`Note.SubjectID` is also `UNIQUEIDENTIFIER` while its six siblings are `NVARCHAR(255)`, so that one link silently forbids composite and non-GUID keys.

**Why the deadline.** Caliber's npm packages are at `0.1.0` and it ships no `Metadata_Sync` migration yet — nothing is installed at a customer, so publish-then-no-breaking-changes doesn't bind and this is a cheap forward migration over seven tables. **After 1.0 publishes it becomes a breaking schema change and the cost changes category entirely.** Doing it now is roughly a day; doing it after is a deprecation cycle.

The fix is two independent changes, both in the plan:
1. `SubjectEntityName NVARCHAR(255)` → `SubjectEntityID UNIQUEIDENTIFIER` FK to `__mj.Entity(ID)` — permitted, since Caliber's rule forbids hard FKs into *sibling Open App* schemas and `__mj` is explicitly fine.
2. `SubjectID` → `NVARCHAR(750)` holding the canonical `ID|<guid>`.

**Four open questions in the plan I'd like decided before Layer 0 lands** — the first one is the load-bearing call:

1. **750 vs 450 for the canonical width.** 450 isn't the ceiling (that's the *clustered* / pre-2016 limit — nonclustered is 1700 bytes / 850 chars on 2016+, and core already ships `IX_RecordChange_RecordID` over a 750-wide column at 1500 bytes). The plan takes 750 on migration-direction grounds: widening is metadata-only, narrowing needs a data scan and index rebuild. 750 narrows nothing; 450 would mean narrowing ten tables. Worth confirming there's no reason to pay that.
2. `RecordGeoCode` re-encode and the widen — one migration or two?
3. `OnTargetDelete` default for newly declared links — `Ignore` (today's behaviour) or `SetNull`?
4. Should a declared-but-empty `AllowedEntities` warn at CodeGen time, so "we never constrained this" is visible rather than silent?

One thing worth knowing going in: `EntityField.EntityIDFieldName` has been sitting in the schema since the v2 baseline, documented for precisely this ("*preventing orphaned soft-linked records during delete operations*"), with consumer code in both DB providers — and it's `NULL` on every row in the system. Populating it is most of Layer 1. Its consumer also has two real bugs that would keep it from working, both written up in the plan.

---

plans/polymorphic-foreign-keys.md
```diff
@@ -0,0 +1,374 @@
...
+// It matters here and not on the neighbouring GUID filters because
+// `SubjectEntityName`/`SubjectID` are free-form NVARCHAR carried in from
+// whoever created the engagement, not MJ-minted GUIDs.
+function sqlLiteral(value: string): string { return value.replace(/'/g, "''"); }
+```
+
+That helper is deliberately duplicated across two packages, with a comment explaining that the
+no-cross-package-re-export rule left no better option. A typed link needs none of it.
+
+**Why the deadline matters.** Caliber's npm packages are at `0.1.0` and it ships no `Metadata_Sync` migration
```
rkihm-BC: Required: this paragraph did not hold on 27 August and the window it describes has closed. On that date Caliber's head (20 Aug) had all four packages at 6.1.0; the `Metadata_Sync` migration had existed since 8 Aug; and the packages were already published restricted to the public registry — publish runs succeeded 13 Aug (v6.0.0), 14 Aug (v6.1.0) and 6 Sep (v6.2.0). Under the publish-then-no-breaking-changes policy the seven-table change is now a breaking schema change. The fix stays right; rewrite the cost/sequencing around the deprecation path, and drop "0.1.0" / "no Metadata_Sync migration".

plans/polymorphic-foreign-keys.md
```diff
@@ -0,0 +1,374 @@
...
+Ships first because it is the only layer that touches **data**, and it gets more expensive every quarter as
+more of these pairs land in customer databases.
+
+1. **`CompositeKey.ToRecordID()`** — thin over `ToConcatenatedString()`, but **rejects** a value containing
+   the delimiter rather than silently emitting a string that will not round-trip.
+2. **`CompositeKey.FromRecordID(entityInfo, s)`** — new, not a rename of `LoadFromConcatenatedString`.
+   That function has three silent-corruption paths (`compositeKey.ts:229`):
+   - no `|` → silent no-op, `KeyValuePairs` left empty, no signal to the caller. *This is the defect that
+     forces every consumer to sniff.*
+   - a value containing `|` → `kv[1]` only; the rest is dropped.
```
rkihm-BC: Superseded: #4184 (merged 4 Sep, after this plan) changed this to `kv.slice(1).join(valueDelimiter)`, so a value containing the delimiter round-trips. The other two paths (no delimiter → silent no-op; `||` → phantom pairs) still stand and still justify `FromRecordID`.

plans/polymorphic-foreign-keys.md
```diff
@@ -0,0 +1,374 @@
...
+
+- Choosing **750** narrows nothing. Every column at 100/255/400/450/500/700 widens for free. Only three need
+  real work: `Recommendation` (MAX), `TemplateParam` (2000), `caliber.Note` (UNIQUEIDENTIFIER).
+- Choosing **450** would require narrowing ten tables, each needing a length scan first.
+
+Declared width costs nothing at runtime — nvarchar is variable-length, so 750 and 450 store identically. The
+number only gates the `CREATE INDEX` check. PostgreSQL does not constrain either choice; its btree limit is
+checked against actual data (~2704 bytes), not declared width.
+
+> **To verify before Layer 0 lands:** whether `ALTER COLUMN` widening is permitted in place on the columns
```
rkihm-BC: Worth stating here rather than leaving to verify: every index whose key includes a RecordID-shaped column stays under the 1,700-byte nonclustered limit after a 750 widen — `UQ_RecordGeoCode_EntityRecordLocation` is the largest at 1,616 bytes (16 + 1,500 + LocationType NVARCHAR(50)); `UQ_RecordProcessWatermark_Record`, `UQ_TagScope_Tag_Entity_Record`, `IDX_TagScope_Scope_Tag` and the AIAgent PrimaryScope indexes come to 1,532; no clustered key includes one. (The TagScope index is `IDX_TagScope_Scope_Tag`, not `IX_`.)

rkihm-BC (changes requested):
Read the plan, the PR comment, and verified every checkable claim against `next` and the Caliber repository's `next`. Most of it holds and the reframe is right. Two corrections are needed before this merges as the reference document, one of them material; the rest is comment.

## Verified against the code

- **The declaration slot exists and is empty.** `EntityField.EntityIDFieldName` is `NVARCHAR(100) NULL` in the v5.46 baseline; every baseline `EntityField` row I could parse (1,000) has it NULL, no later migration or metadata seed sets it, and both database providers consume it.
- **B1–B4 are all real.** B1: the SQL Server soft-link query filters on `entity.ID` where `entity` is the holder — it looks for rows pointing at themselves. B2: `GetRecordDependencies` returns before building the soft-link SQL when there are no hard dependents. B3: compares against the bare first key value. B4: quoting from the holder's key type. Same shape in the PostgreSQL provider. **B1 and B2 are shippable now**, independent of every decision below — I'd split them into a fix PR rather than hold them behind four open questions.
- **Two encodings ship**: the geocode service writes a bare value or `val1||val2` while everything else writes `ID|value`; `recent-access.service.ts:263` sniffs for a pipe; `generateCascadeDeletes` walks `RelatedEntityID` only; the pairing CHECKs are hand-written where they exist.
- **D2 is safe on the key limits, and the plan should say so.** I sized every index whose key includes a RecordID-shaped column after a 750 widen: the largest is `UQ_RecordGeoCode_EntityRecordLocation` at 1,616 bytes (16 + 1,500 + `LocationType` NVARCHAR(50)), the rest 1,532, all under the 1,700-byte nonclustered limit, and no clustered key includes one. Small slip: the TagScope index is `IDX_TagScope_Scope_Tag`, not `IX_`.

## Required

**1. The Caliber section was already wrong on 27 August, and the deadline it argues from has passed.** On that date Caliber's head was a 20 August commit with all four packages at **6.1.0**, not 0.1.0; the `Metadata_Sync` migration had existed since 8 August; and the packages are published (restricted, on the public registry per Caliber's `.npmrc`) — the publish workflow succeeded on 13 Aug (v6.0.0), 14 Aug (v6.1.0) and 6 Sep (v6.2.0). So "nothing is installed at a customer, the policy does not bind, this is a cheap forward migration" did not hold when written. The subject-link fix is still the right fix (the seven tables are unchanged, `Note.SubjectID` is still `UNIQUEIDENTIFIER`, and the `sqlLiteral` helper is still there), but under the publish-then-no-breaking-changes policy it is now a breaking schema change and needs the deprecation path. The cost and sequencing paragraph has to be rewritten around that, since it is the load-bearing paragraph of the section. Inline comment.

**2. One of the three `LoadFromConcatenatedString` "silent corruption paths" is superseded.** #4184 (merged 4 Sep, after this plan) made everything after the first delimiter the value, so `Code|a|b` round-trips. The no-delimiter no-op and the `||` phantom-pairs path remain. Inline comment.

## The four questions

- **Q1** is answered by the plan's own D2 — delete the question.
- **Q2**: two migrations. The widen either applies or it doesn't; the re-encode can partially apply and needs a rollback story. Bundling blocks the safe half on the risky one.
- **Q3** can't be answered as posed: under `Pairing: Required` both columns are NOT NULL, so `SetNull` violates the constraint on the first cascade. Require an explicit value for new declarations; backfill existing ones as `Ignore`. Defaulting new links differently from old ones would make two identical pairs behave differently by declaration date.
- **Q4**: yes, warn — but only after adding an explicit "any entity, deliberately" opt-in, or the warning fires on all ~50 existing pairs on day one and gets muted.

## Missing

- **Layer 1's seeding mechanism.** Populating `EntityIDFieldName` for the ~33 core fields must go through `metadata/` and sync, not a hand-written migration — that is how field attributes are seeded today (`metadata/entities/.entity-field-jsontype-embedded-record.json` sets `EmbeddedRecord`'s JSONType exactly this way). The plan never names it.
- **The widen's cost is unmeasured** on `RecordChange` and `UserRecordLog`, so "ships first" is not yet a schedule.
- **No rollback story for the `RecordGeoCode` re-encode.**
- **What the ten Open App repos must do, and by when.**

Nits: the appendix lists 35 MJ tables and the summary says 37 generated classes; the baseline has 33 pair-bearing tables — approximate is fine, but pick one number. Docs-only, merge is clean despite the branch being 605 commits behind, no changeset needed.

MS-BC:
Thanks @rkihm-BC — this was a genuinely load-bearing review. Both required changes are applied in b6c43b58, and verifying your Caliber correction turned up enough that I ended up re-scoping Layer 0. Point by point.

## Required — both fixed

**1. Caliber.** You're right, and I verified it independently rather than taking it on trust: the `0.1.0` came off the *private monorepo root* `package.json`, while the four published packages were at 6.1.0 on 27 Aug and are at **6.2.0** now. `Metadata_Sync` since 8 Aug. `Build and Publish` last succeeded **2026-09-06**. So the policy already bound and the window the section argued from never existed. Section rewritten around the additive-plus-deprecate path, with the original claim quoted and struck so the error is visible rather than quietly edited away.

**2. `LoadFromConcatenatedString`.** Confirmed — #4184's `kv.slice(1).join(valueDelimiter)` is on `next`, so that path is gone. Removed; the no-delimiter no-op and the field-delimiter phantom-pairs path are kept, and they're still what justifies a separate reader.

## Your other comments

- **D2 index sizes** — table added inline, with `UQ_RecordGeoCode_EntityRecordLocation` at 1,616 and the rest at 1,532, so it doesn't have to be re-derived. `IDX_TagScope_Scope_Tag` corrected.
- **Counts** — settled on one set throughout: **54 pairs / 53 tables / 10 repos**, MJ core 35 pairs across 34 tables. I've noted why that's one more than your baseline 33 (the scan spans all migrations, not just the baseline).
- **Layer 1 seeding** — named, with `.entity-field-jsontype-embedded-record.json` as the template. Also verified the thing that makes it durable: **CodeGen never writes `EntityIDFieldName`** — there's no `SET` for it anywhere in `manage-metadata.ts`, only a read — so a seeded value survives a CodeGen run.
- **Q1–Q4** — converted to decisions, taking your answers. Q1 deleted as self-answering. Q3 recorded as *not answerable as posed* for the reason you gave (`SetNull` violates `Pairing: Required` on the first cascade), resolved as explicit-per-declaration plus `Ignore` backfill. Q4 sequenced behind the deliberate-opt-in so it doesn't fire on all 54 pairs and get muted.
- **The four "Missing" items** — I couldn't close them, so rather than bury that they now sit under an explicit **"Still genuinely open"** heading: widen cost still unmeasured on real row counts, no `RecordGeoCode` rollback story, no per-repo plan for the other nine, and Layer 0 item 4 unscoped.

## Taking your split advice

You said B1 and B2 were shippable now and shouldn't be held behind four open questions. Done — **#4321**, off `next`, code only, no migration. It grew to six defects, because making B1/B2 testable exposed two more:

- **B5** — `MergeRecords` repointed every dependency with `GetValueByIndex(0)`, the bare key. Right for an FK, silently wrong for a `RecordID` column. This is why B1/B2 couldn't ship alone: fixing the *query* without the *write-back* converts a dead path into a corrupting one.
- **B6** — dependency keys were built from the parent entity's PK columns though the value is the related entity's key. Latent while both are `ID`.

39 new tests, all suites green. I mutation-checked them against the pre-fix files — 6 of 9 provider tests and 5 of 7 base tests fail without the fixes — because a test that passes either way isn't evidence.

## Two findings that change the plan, not just the document

**Canonical is 2 of 90 write sites.** Before seeding `EntityIDFieldName` I audited what those columns actually contain. There are **five** encodings, not two: canonical; bare single value (most of them); comma-space joined via `CompositeKey.Values()` (`UserFavorite`, from the provider); `||`-joined (`RecordGeoCode`); and `Field=Value AND …` via `ToString()` (`DataContextItem`). Plus `CompanyIntegrationRunDetail.RecordID`, which isn't a record pointer at all.

Two that I think are worth your eye specifically:

- **`UserFavorite` is written both ways** — `Values()` from the provider, bare `.ID` from `lists-browse-resource.component.ts` and `explorer-state.service.ts`. One column, two encodings.
- **`ListDetail` switches encoding on key arity** — bare for single, canonical for composite. So "which encoding is this column?" isn't answerable per column.

So D1's "already the majority format" is false, and **Layer 1 is blocked on normalising 88 writers**, not on re-encoding one table. That's now the largest item in Layer 0 and it wasn't in the plan at all. I deliberately seeded nothing in #4321 for this reason.

**Your `ALTER COLUMN` question is answered — and the answer has a sting.** I probed it against SQL Server 2022:

- Widening an indexed `NVARCHAR` **succeeds**; it's metadata-only and needs no index drop. The plan's "assume the latter until measured" was pessimistic and can leave the estimate.
- A `CHECK` constraint on the column doesn't block it either.
- **But an over-limit index is created with only a *warning*.** A 2,026-byte key created fine, warning that inserts will fail "for some combination of large values" — so exceeding 1,700 is a *latent runtime* failure, not an honest build-time one. A successful `CREATE INDEX` is not evidence that it fits.

That last point is what makes Caliber's two halves inseparable: after a 750 widen, `UQ_InviteScope_Protocol_Subject_Active` (`StepID`, `SubjectEntityName`, `SubjectID`) comes to 2,026 bytes and would ship warning-only. Re-keyed onto the 16-byte `SubjectEntityID` it's 1,532 and clean. **The typed discriminator is what makes the widen safe**, so they can't be sequenced apart — and the same obligation falls on any of the ten repos keying a `RecordID` beside another wide column.

## Caliber, on the way

Also stale, which the live database showed: **`Protocol` is now `Step`**, and there are **eight** name-based bindings, not seven (`Step` carries a type binding with no `SubjectID` sibling). A ninth, `BlueprintDetectionRule`, has `SubjectEntityName` + `SubjectIDField` — a field *name*, out of scope.

The additive migration is authored and verified against a restored copy of the live Caliber DB: applies clean, **0 unmatched across 503 rows**, eight FKs enforcing, all three subject indexes re-keyed with no size warning. The FK rejects a bogus entity id while `SubjectEntityName` still happily accepts `'Totally Not An Entity'`. PG counterpart and the CodeGen append are still to do before it can ship, and I'm not re-encoding `SubjectID` there — all ten Caliber writers store a bare `subjectId`, so that's the same problem as Layer 0 item 4 and should follow core rather than diverge ahead of it.

One thing I could not do: **the deterministic integration tier wouldn't run locally** (the CLI resolves from the main checkout and `MJCLI` won't build in my worktree). It's running on #4321 in CI instead — I'd rather say that than imply I'd run it.

plans/polymorphic-foreign-keys.md
MS-BC: Fixed in b6c43b58 — and confirmed independently: the `0.1.0` was read off the private monorepo root `package.json`, while the four published packages were 6.1.0 on 27 Aug and are 6.2.0 now. `Metadata_Sync` since 8 Aug; `Build and Publish` last succeeded 2026-09-06.

The section is rewritten around the additive-plus-deprecate path the policy mandates (add `SubjectEntityID` nullable + FK, backfill, keep `SubjectEntityName` deprecated-but-present, widen `SubjectID` since widening is permitted, and give `Note.SubjectID` a companion column rather than retyping it). The original claim is quoted and struck rather than quietly edited, so the error stays visible.

Two things the live database added: `Protocol` is now **`Step`**, and there are **eight** name-based bindings, not seven.

plans/polymorphic-foreign-keys.md
MS-BC: Removed in b6c43b58 — verified #4184's `kv.slice(1).join(valueDelimiter)` is on `next`. The no-delimiter no-op and the field-delimiter phantom-pairs path are kept, and they remain the justification for a separate validating reader (`FromRecordID`, shipped in #4321).

plans/polymorphic-foreign-keys.md
MS-BC: Added inline in D2 in b6c43b58, with your numbers as the table (`UQ_RecordGeoCode_EntityRecordLocation` 1,616; the rest 1,532; no clustered key), and `IDX_TagScope_Scope_Tag` corrected.

I also closed the "to verify before Layer 0 lands" note next to it by probing SQL Server 2022: **widening an indexed `NVARCHAR` succeeds** — metadata-only, no index drop needed, so the pessimistic assumption can leave the estimate. A `CHECK` on the column doesn't block it either.

The sting is that an **over-limit index is created with only a warning** — a 2,026-byte key created fine, warning that inserts will fail "for some combination of large values". So a successful `CREATE INDEX` isn't evidence that it fits, and this is exactly Caliber's case after a 750 widen.

rkihm-BC (approved):
Approving — this clears my `CHANGES_REQUESTED`, which was the only thing blocking. Both required items are fixed, and I checked the file rather than the summary.

## Both required corrections landed

**Caliber.** The section is rebuilt around additive-plus-deprecate, and the original rationale is **quoted and struck** rather than quietly replaced (line 371). That is the right way to correct a reference document — the next reader can see what the plan used to claim and why it was wrong, which is exactly what you want in a document whose purpose is to be trusted later. You also verified the correction independently rather than taking mine on trust, and found the source of the error (`0.1.0` read off the private monorepo root rather than the four published packages). That is the more useful outcome.

**`LoadFromConcatenatedString`.** #4184's `kv.slice(1).join(valueDelimiter)` confirmed on `next`, the superseded path removed, the two that remain kept as the justification for a separate validating reader. Correct on both counts.

## The rest, verified in the file

- **D2 index sizes** — table present with `UQ_RecordGeoCode_EntityRecordLocation` at 1,616 and the rest at 1,532 against the 1,700 limit, `IDX_TagScope_Scope_Tag` corrected. Nobody has to re-derive it.
- **Counts** — settled consistently at 54 pairs / 53 tables / 10 repos, with MJ core at 35/34 and a stated reason for the difference from my baseline 33.
- **Layer 1 seeding** — named, with `.entity-field-jsontype-embedded-record.json` as the template, plus the durability check I did not think to ask for: **CodeGen never writes `EntityIDFieldName`**, so a seeded value survives a run. That is what makes the mechanism safe rather than merely conventional.
- **Q1–Q4** — converted to a Decisions section, Q1 deleted as self-answering, Q3 recorded as not answerable as posed for the `Pairing: Required` reason.
- **The four Missing items** — under an explicit **"Still genuinely open"** heading rather than closed. I would rather a plan say "88 writers is a number, not a plan" than pretend Layer 0 item 4 is scoped.

## What makes this worth merging as the reference document

The plan now records two things that were discovered by *implementing* it, and both change what it says rather than decorating it:

**Canonical is 2 of 90 write sites, not the majority.** That falsifies D1's original cost argument, and the document says so with the old claim visible. `UserFavorite` written both ways and `ListDetail` switching encoding on key arity are the two that matter most — they mean "which encoding is this column?" is not answerable per column, which is a different and harder problem than the plan originally described.

**An over-limit index is created with only a warning.** This is the finding I would most want preserved: a successful `CREATE INDEX` is not evidence that the key fits, so exceeding 1,700 bytes is a latent runtime failure rather than an honest build-time one. It is also what makes Caliber's two halves inseparable — the typed discriminator is what makes the widen safe, so they cannot be sequenced apart.

Neither was in the original plan, and neither would have surfaced without shipping #4321 first. That is the split working as intended.

## Mergeable

`MERGEABLE`, **0 behind `next`**, 5 checks pass with nothing pending or failing, single file, docs-only so no changeset. Nothing outstanding on my side.

Two housekeeping notes, neither blocking:

- The three inline threads are all answered and all still open. Worth resolving them so the next reader is not left checking whether something is outstanding.
- @AN-BC opened this; @MS-BC has carried it since. Worth confirming Amith is content with the reframe before it merges as *the* reference document, since the D1 correction and the re-scoped Layer 0 change the shape of the work it describes rather than just its details.

PR #4086, opened 2026-08-28
fix(forms): stop the chat-embedded form from clipping long text (clip-fix subset of #4085 for 5.51.x)

rkihm-BC:
Backport of #4085 to the 5.x LTS line (`lts/5`, which carries `v5.51.1`).

**Reduced to the clip-fix subset following review.** The original cherry-pick brought over the whole of #4085, which was a fix plus a restyle. On a certified line the restyle half changes rendering for deployments that were never clipping, so it stays on `next` and only the clip fix rides here.

Clipping is confirmed in production on a 5.51 deployment (MoreCheese demo site), which is what this PR is for. A truncated option label is not a cosmetic problem — the user answers a question whose choices they cannot fully read.

## What this ships

Every rule is a no-op unless the element is already overflowing:

- **Radio/checkbox options wrap** — they share a row when they fit, and an option too wide for the row takes the row to itself and wraps its label instead of being chopped by the card's `overflow: hidden`.
- **Labels can actually wrap** — `min-width: 0` and `overflow-wrap: anywhere`; without these the label is pinned to its content width.
- **Button groups wrap** instead of scrolling horizontally. The old `overflow-x` scroller sat inside a vertically-scrolling chat column, where overlay scrollbars hid the overflowing options outright rather than signalling them, and `overflow-y: hidden` clipped focus rings.
- **Single-line controls degrade to an ellipsis** rather than a mid-word cut.
- **The submitted-answer pill wraps** — it was `white-space: nowrap` with no max-width, so a sentence-length answer ran past the message.
- **The multi-field answer card no longer overflows a narrow message column** — its `min-width: 400px` unconditionally beat its own `max-width: min(800px, 100%)`, and the guards below it key off the viewport rather than the column, so a narrow chat panel on a wide screen never reached them. Now `min(400px, 100%)`.
- **The form card can shrink as a flex item** (`min-width: 0`).

## What was deliberately left on `next`

Reverted to `lts/5` values in 80ee8aa:

- Card `600px → 720px` and default text field `450px → 560px`
- `font-size: 14px` on radio/checkbox option labels — this overrides the host app's own body size, which is a typography decision to make on a major, not in a line patch
- `align-items: center → flex-start` with the 3px input nudge and 2px icon nudge — this shifted every single-line option including `Yes`/`No`
- The `.input-container.width-auto` restructure

Keeping `align-items: center` removes the reason for the `font-size`/`3px` coupling (the nudge was calibrated to reproduce centering at 14px/1.4, so splitting the two would have misaligned options in any host app with a different body size). Single-line options are now pixel-identical to 5.51.1.

Reverting `width-auto` also drops a side effect not raised in review: `width: auto` on the select made **short**-option dropdowns render narrower than the old 350px cap — a visible change for exactly the population this bar protects. Long selected labels now degrade to the ellipsis instead. Confirmed non-load-bearing for the clip fix: `WidthClass` routes `radio`/`checkbox`/`buttongroup` to `width-wide`, so only dropdowns ever reach `width-auto`.

## Divergence

`lts/5` and `next` now differ in these three files by design. Each file carries a `NOTE (lts/5)` comment naming #4085 and listing what was omitted, so a later cherry-pick conflict is self-explaining rather than looking like a botched backport.

## Verification

- `check-css-hex-tokens` and `check-mj-btn-override`: clean on all three files
- `npm test` in `packages/Angular/Generic/forms`: **94/94 passing** (94 on this line, not the 95 on `next` — line difference, not a regression)
- `npm run build` (ngc): clean

Note for anyone else working this line: `lts/5` is still an **npm** workspace (`packageManager: npm@11.7.0`, `package-lock.json`). The pnpm guidance in `CLAUDE.md` describes `next` only.

## Test plan

- [ ] In a 5.51.x build, have an agent emit a form with sentence-length radio options — confirm each label renders in full
- [ ] Confirm a `buttongroup` with long labels wraps rather than scrolling
- [ ] Confirm a long dropdown value ends in an ellipsis rather than a mid-word cut
- [ ] Check short-label forms (`Yes`/`No`) are pixel-unchanged vs 5.51.1 — no shifted controls, no extra rows, no changed label size
- [ ] Confirm short-option dropdowns are unchanged in width vs 5.51.1
- [ ] Submit single- and multi-question forms; confirm neither the pill nor the summary card overflows the message bubble
- [ ] Narrow the chat panel on a wide monitor and re-check — the case the viewport media queries never caught
- [ ] Verify at mobile width (<= 600px) that fields still fill the available width

### DIFF
--- a/.changeset/dynamic-form-text-truncation.md
+++ b/.changeset/dynamic-form-text-truncation.md
@@ -1,0 +1,14 @@
+---
+"@memberjunction/ng-forms": patch
+---
+
+Stop the chat-embedded interactive form from clipping long text. Agent-authored labels are routinely sentence-length, and the form's layout sized every option to its content width with no ability to shrink — so anything wider than the card ran off the edge and was silently chopped by the card's `overflow: hidden` (no ellipsis, no scrollbar, just missing words).
+
+- **Radio and checkbox options now wrap.** Options size to their content and share a row when they fit, but an option too wide for the row takes the row to itself and wraps its label rather than overflowing. Options and labels are top-aligned so the control lines up with the first line of a multi-line label.
+- **Dropdowns size to their widest option** (bounded by the card) instead of a fixed 350px cap that truncated the selected label. `widthHint: 'auto'` is legal on any question type, so text controls under that hint keep their previous bound — only the select needed the extra room.
+- **The form card allows more room** before its contents have to wrap (600px → 720px), and the default text-field width follows (450px → 560px).
+- **Button groups wrap to a second row** instead of scrolling horizontally. The previous `overflow-x` scroller sat inside a vertically-scrolling chat column, where overlay scrollbars hid the overflowing options entirely rather than signalling them; it also clipped the buttons' focus ring. Now consistent with `.footer-choice-button` and `.choice-button`, the sibling button paths in the same form.
+- **Radio/checkbox option labels are 14px**, matching every other control in the form. They were the only text here with no size of their own, so they inherited the host app's body size (~16px) and rendered larger than the question they answer.
+- **Single-line controls degrade to an ellipsis** rather than a mid-word cut when a value or placeholder still exceeds the field.
+- **The submitted-answer pill wraps too** — it was `white-space: nowrap` with no max-width, so a sentence-length answer ran past the message.
+- **The multi-field answer card no longer overflows a narrow message column.** Its `min-width: 400px` unconditionally beat its own `max-width: min(800px, 100%)` (min-width always wins), and the guards below it key off the viewport rather than the column — so a narrow chat panel on a wide screen never reached them. Now `min(400px, 100%)`.
--- a/.changeset/dynamic-form-text-truncation.md
+++ b/.changeset/dynamic-form-text-truncation.md
@@ -4,11 +4,11 @@
 
 Stop the chat-embedded interactive form from clipping long text. Agent-authored labels are routinely sentence-length, and the form's layout sized every option to its content width with no ability to shrink — so anything wider than the card ran off the edge and was silently chopped by the card's `overflow: hidden` (no ellipsis, no scrollbar, just missing words).
 
-- **Radio and checkbox options now wrap.** Options size to their content and share a row when they fit, but an option too wide for the row takes the row to itself and wraps its label rather than overflowing. Options and labels are top-aligned so the control lines up with the first line of a multi-line label.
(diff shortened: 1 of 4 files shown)


rkihm-BC:
@cadam11 , please take a quick look at this PR to confirm it's done correctly for a backport.

cadam11:
> @cadam11 , please take a quick look at this PR to confirm it's done correctly for a backport.

This is a fix plus a restyle, and the restyle half is what gives me pause on a certified line.

The clip fix is the wrap rules, the min-width on the multi-field card (min-width unconditionally beating max-width is a real bug), the pill's white-space, and the ellipsis fallback. None of that is visible to anyone who wasn't already clipping.

The rest changes how the form looks for people who never had the problem: card 600 to 720, default field 450 to 560, and radio/checkbox labels pinned to 14px where they used to inherit the host app's size. The switch from align-items center to flex-start plus the 3px input nudge shifts every single-line option too, including Yes/No, which is the case the test plan flags and that box isn't ticked.

Same shape as the cache freeze on #3425 last week. A line patch shouldn't change behavior people can't opt out of, and we can't survey who has themed around these selectors.

The line exists so customers can stop tracking our cadence, so every patch we ship asks all of them to re-qualify, not just the ones with the bug. That's the bar I'm applying, and rendering polish doesn't clear it on its own.

So either this rides 6.x, or we take the clip-fix subset and leave the sizing and typography on next. I'd suggest the second if clipping is biting a 5.51 customer today, and the first if it isn't.

Worth knowing for next time: the `backport lts/5` label on the source PR opens this automatically, and it still works after the merge. It copies the whole commit though, so a split like this one has to happen on `next` first.

----

On the general question, since it will come up again: the line isn't v5 with every bug fixed. Certification evidence is tied to one build, and every patch after it is delta the full gate run never saw, covered only by the lighter mini-cert. 

If we patch the line freely then what customers install drifts away from the thing we actually certified while still wearing the label. `5.51.2` is already carrying three correctness fixes and no security fix at all, so the bar isn't security-only, it's whether staying put costs the customer more than moving. 

I wrote that up as 5.3 in the process doc, #4098, so we can argue the general rule in one place instead of per PR.

rkihm-BC:
@cadam11 , I reduced the scope of this change based on your feedback. The bug is impacting 5.51 systems so I do want to patch it. 6.x will include the larger change but this will be limited to addressing the text cutoff.

PR #4087, opened 2026-08-28
feat(providers): opt-in batched transaction-group submit — one round trip per group

MS-BC:
`TransactionGroupBase` gains an opt-in **`BatchedSubmit`** flag (default **false** — existing callers byte-for-byte unaffected). When set, both providers execute a variable-free group's items as **one multi-statement round trip** instead of one round trip per item: same statements, same order, same transaction, per-item results still returned.

## Why

The sequential submit is atomic but not batched — each item's generated CRUD procedure call is its own wire hop. Measured on a live sync: ~0.3ms of server-side execution inside a per-statement wall cost two orders of magnitude larger, so a 100-item group spends nearly all its time on round trips the SQL never needed. Batching the wire recovers the entire speed advantage of a direct-write path with none of its costs: every statement is still the generated procedure, so validation, Record Changes, and save events are untouched.

## Design

- **Sentinel result mapping** — a statement that returns no rows produces *no* recordset, so positional zipping silently drifts and attributes row A's identity to row B. Each item is preceded by a sentinel `SELECT <index>`; recordsets between sentinel *k* and *k+1* belong to item *k*. The empty-middle-item drift case is covered by tests on both providers.
- **SQL Server** — per-item `?` placeholders are renumbered into one global `@p` namespace (one request carries one parameter namespace).
- **PostgreSQL** — the extended protocol cannot carry `$N` parameters in multi-statement text, so values are inlined through the **driver's own `escapeLiteral`** (never hand-rolled), gated to unambiguous literal forms (string, finite number, boolean, null, Date; plain objects are serialized by the existing parameter processor before the gate). Any value outside the set — or a client with no `escapeLiteral` — falls the **whole group** back to the sequential path: correctness first, batching second.
- **`Variables` groups always run sequentially** regardless of the flag — a later item's SQL is re-rendered from an earlier item's output, which a single round trip cannot provide.
- **Failure semantics unchanged** — a batch failure rolls back and throws exactly as the serial first-error path does; per-item attribution of a poison row remains the caller's degradation path.

## Tests

8 new (5 PG, 3 SS): single-query proof, sentinel mapping including the empty-middle-item case, global parameter renumbering, both PG fallbacks (no `escapeLiteral`; no safe literal form — with proof the value then travels as a parameter, never as text), and whole-group failure. `tsc` clean on all three packages (0 new errors).

## Adoption

Dormant until a caller sets the flag. The intended first consumer is the integration engine's batched write path (`writeMode: 'batched'`), which enrols large numbers of independent items per group — a one-line follow-up on that branch once this merges.


rkihm-BC (commented):
## Review — batched transaction-group submit

Read the full diff and checked the claims against `SQLServerTransactionGroup` / `PostgreSQLTransactionGroup` on `next` rather than against the description. The premise is right and the hard part is solved properly. Nothing below blocks; one ask, two mechanical fixes, three optional notes.

### Verified good

- **Sentinel mapping is the correct solve, not a clever one.** A statement returning no rows produces no recordset, so a positional zip drifts and misattributes identity. The `SELECT k AS __mj_batch_item` fence is the honest fix, and both providers test the empty-middle-item case that breaks the naive version. That case is the whole reason this needed thought, and it's the one that's covered.
- **The rollback contract threads an existing needle correctly.** The SQL Server batch throws `Transaction rolled back due to operation failure: …`, which is precisely the string the pre-existing outer catch tests for before deciding whether to roll back again. Verified against the merge-base — no double rollback. Easy to get wrong silently; it isn't wrong.
- **Fails toward the proven path at every decision.** Default false, `Variables` groups always sequential, and PG bails the *whole group* if any value lacks a safe literal form or the client exposes no `escapeLiteral`. Using the driver's own escaper rather than a hand-rolled one is the right instinct, and the `'O''Brien'` assertion pins that it actually went through it.

### The one ask

**Integration-tier coverage of the batched path.** Both test files drive fakes — a hand-rolled `PoolClient` and a `vi.mock('mssql')`. They're good tests of the mapping logic, but "same statements, same order, same transaction, one round trip" is currently asserted against a mock that cannot disagree. This is a change to the wire protocol of the write path on both dialects, and the deterministic tier is where that claim gets proven; CLAUDE.md's Definition of Done asks for it on a change this shaped.

Concretely, the properties I'd want a real database to confirm: per-item results land on the right items when a middle statement returns nothing, the global `@p` renumbering binds what it claims to bind, and a failure inside the batch leaves nothing committed. #4066 needs the same bundle — one set of cases covers both PRs.

### Two mechanical fixes

1. **`executeBatchedNoVars` uses `any` four times on three lines** — `sets: any[]`, `queryResult as any`, `perItem: any[]`, `result: any`. The PostgreSQL sibling does the same job with `Record<string, unknown>[]` and gets it right, so the correct types are already written down in the other file. The repo rule (`.claude/rules/typescript-style.md`) forbids `any` outright.
2. **The extracted body kept its old indentation.** Everything inside `executeBatchedNoVars` is still indented at the depth it had when it was inline, ~20 spaces in. Cosmetic, but it makes the method read as if it's still nested inside the branch it came from.

### Optional

- **Batched SQL logging keeps only item 0's `simpleSQLFallback` and description.** The sequential path calls `LogSQLStatement` per item with that item's own fallback and its own `OperationType`/`entityName`; the batch calls it once with `items[0]`'s. To be clear about the blast radius: `LogSQLStatement` no-ops unless a SQL logging session is open, and `simpleSQLFallback` is documented as *"simple SQL to use for loggers with `logRecordChangeMetadata=false`"* — so this shapes migration capture, not Record Changes, which the procedures write either way. It's a degraded dev-facing capture when batching and SQL logging are both on, not a data-correctness issue. Cheap to fix while the code is open.
- **PG's literal gate maps an absent parameter index to `NULL`.** `literalFor(params[Number(digits) - 1])` returns `'NULL'` for `undefined`, in the same branch as an explicit null. This is unreachable today — `GenerateSaveSQL` returns `{ fullSQL, simpleSQL, parameters }` from one generator, so placeholder count and parameter count can't diverge unless that generator is broken. Worth separating "index absent" from "value is null" purely as hardening, because batching is the path where that particular failure would be silent rather than loud.
- **The two providers disagree on which recordset an item owns.** PG takes the first *non-empty* recordset after a sentinel (`perItem[current] === undefined && rows.length > 0`); SQL Server takes the first, empty or not (`perItem[current] === undefined`). Theoretical — generated CRUD procedures emit one recordset each, so nothing hits it — but a line of comment stating the intended rule would stop the two implementations drifting further apart.

One note that is explicitly *not* a request: the bare `/\?/g` renumbering is inherited from the existing sequential code, so it's not a regression here. Worth knowing that the batch concatenates N instructions into one parameter namespace, so a stray `?` inside a literal now misnumbers everything after it rather than one statement — but that's an argument about the pre-existing approach, not about this PR.

MS-BC:
Thanks — all of it addressed. Pushed as one commit.

**The two mechanical fixes**
- The four `any` uses in `executeBatchedNoVars` are gone, replaced with the types the PostgreSQL sibling already had right (`Record<string, unknown>[]` / `(Record<string, unknown>[] | undefined)[]`).
- The extracted method is reindented to its real depth.

**The optional notes — all three taken**
- **SQL logging is now per item**, each with its own `simpleSQLFallback` and `OperationType`, matching the sequential path. Your framing is what decided it: migration capture replays statements one record at a time, so a single entry under item 0's fallback captured something unreplayable *and* attributed every row to the first entity. The combined batch is still logged once, deliberately fallback-free, so a reader can still see what went over the wire. Pinned by a new unit test that asserts each item appears under its own fallback while the round-trip count stays at one.
- **PG's literal gate now separates "index absent" from "value is null"** — an out-of-range placeholder bails the group to the sequential path instead of inlining `NULL`. Agreed on the reasoning: unreachable today, but batching is precisely where that disagreement would write a null column silently rather than failing loudly. New test pins that the parameter travels as a parameter and never becomes a `NULL` literal.
- **Both providers now state their recordset-ownership rule** in a comment at the mapping site, each naming the other's choice, so the deliberate difference cannot drift further apart unnoticed.

**The ask: integration-tier coverage**

New bundle — **IT86 / `transaction-groups-batched`**, server transport, 5 checks. It had to be its own bundle rather than an addition to IT47: `BatchedSubmit` is a property of the group the *server* builds, and a client group is serialized into `ExecuteTransactionGroup` and reconstructed server-side, so a flag set on the client instance never reaches the object that reads it.

The properties you named, plus one you didn't:

- **TGB1 — the wire shape itself.** After a batched submit it reads SQL Server's **own plan cache** (`sys.dm_exec_cached_plans` × `sys.dm_exec_sql_text`) for the sentinel text: one cached batch statement, carrying ≥3 sentinels for a 3-item group. That is the database's account of what it was asked to run, not the client's — the mock genuinely cannot disagree, this can. Skips-as-pass loudly where `VIEW SERVER STATE` isn't grantable.
- **TGB2 — per-item mapping through a real driver.** 3 items → 3 *distinct* server-assigned ids, and each finalized id is asserted to belong to the row bearing that entity's own name. That's the positional-drift failure stated as a property rather than as a recordset shape.
- **TGB3 — the renumbering binds what it claims.** Three items with deliberately unmistakable values (a quote, unicode, `[brackets] & %percent%`), each asserted to have persisted on its own row.
- **TGB4 — failure atomicity** inside one batched statement: a real FK violation, `Submit()` false, zero rows persisted, the good entity left unfinalized.
- **TGB5 — differential equivalence.** The same group run sequentially and batched must agree on rows, ids and values. This is the one that catches "fast but subtly different", which is the failure mode I'd actually fear here.

`RequiresMutation` on all five; self-cleaning prefixed fixtures; SQL Server only (PG's inlining path is a different mechanism and is covered by its own unit tests), skipping loudly elsewhere.

The registry's own pins caught the addition — count, bundle map and mutation-gated snapshot all updated deliberately rather than by paste.

**One note that isn't a fix:** #4066 shares this bundle's value but not its coverage — enrolment happens through the *engine*, and a real-entity/real-group/real-submit path is exactly what IT86 now exercises, so the "nothing anywhere enrols a real entity in a real group and submits it" gap is closed by this PR for both.

Suites: PostgreSQL 6, SQL Server 4, integration-test-suite 200 — all green; `turbo build` clean.

rkihm-BC (approved):
Approving. Everything from the last round is addressed, and three items went past what I asked for. Waited for a clean board before posting — `Integration (SQL Server, deterministic)`, `Build`, all six unit shards and the ESM guard are green, `mergeStateStatus: CLEAN`.

## The one ask — integration coverage — is done, and it's better than the ask

`transaction-groups-batched.checks.ts` as bundle **IT86**, with suite metadata and a registry entry. It covers the three properties I named and adds two I didn't think to:

- **TGB2** — per-item mapping survives a real driver; each entity finalizes with its own server ID
- **TGB3** — the global `@p` renumbering binds what it claims; per-item values persist per item
- **TGB4** — a poison item (FK violation) fails the whole batch: Submit false, nothing persists
- **TGB1** — the batch really is ONE statement, proven against SQL Server's **plan cache DMVs**. That's the round-trip claim verified against the engine rather than restated against a mock, which is what the unit tests structurally could not do.
- **TGB5** — **differential equivalence**: the same group, run sequentially and batched, observably identical on the same fixtures.

TGB5 is the strongest form the claim can take. Rather than asserting the batched path behaves correctly, it asserts it behaves *identically to the path that already worked* — which is the property anyone enabling this flag actually cares about.

Scoping the bundle to SQL Server and saying so at the top of the file is the right call, since the deterministic tier is SQL Server and PG's batching is covered by that provider's own unit tests. Honest boundary rather than a silent gap.

## Both mechanical fixes done

`any` is gone from the new code — `globalParams: unknown[]`, `parts: string[]`. The one remaining `any` in the file is pre-existing sequential code this PR doesn't touch; I checked the diff to be sure rather than counting occurrences. Indentation is corrected.

## All three optional notes addressed — one improved on

**Per-item SQL logging.** You did more than I suggested: the batch is logged once as a non-fallback entry *and* every item is logged with its own `simpleSQLFallback` and its own `OperationType`/`entityName`. My version would have fixed attribution and lost the wire-level record; this keeps both. The reasoning in the comment is the part worth keeping — migration-capture consumers replay one record at a time, so a single combined entry would "capture a batch nobody can replay and attribute every row to the first entity."

**PG absent-index hardening.** `position < 0 || position >= params.length` now bails the group, cleanly separated from the null case, and the comment states the danger better than my note did: *"batching is precisely the path where that disagreement would be SILENT: inlining it as NULL would write a null column instead of failing."*

**Recordset-ownership rule** documented alongside the sentinel explanation.

## Standing

The three things I called out as verified last round still hold and are worth restating, because they're the parts that make this safe to enable: the sentinel fence is the honest solve for a statement that returns no recordset; the rollback message threads the pre-existing outer catch's string check without double-rolling-back; and every default fails toward the proven path — flag off, `Variables` groups always sequential, PG bailing the whole group when any value lacks a safe literal form.

Nothing outstanding. Good round.

PR #4088, opened 2026-08-28
feat(integration-engine): adaptive throttle-detected fetch concurrency gate

MS-BC:
## What

A per-connection cap on **simultaneous** vendor fetches, with an adaptive (AIMD) cap that finds the account's real concurrency grant by detecting throttles.

## Why

Vendors that govern by *concurrent requests per account* answer overflow requests with long backoffs served **inside** the fetch. Measured live: an account granting 5 concurrent requests, the engine firing ~16 (sync lanes × prefetch), and every overflow fetch stalling ~31.5s in an in-band backoff — invisible to CPU, DB, and network metrics, because a backoff is idle time. The existing rate limiter shapes requests *per unit time*; nothing bounded requests *in flight at once*.

## How

- `withFetchGate(config, fn)` wraps the fetch call site: a FIFO semaphore per company integration. Overflow queues client-side (milliseconds) instead of serving the vendor's backoff.
- The cap is driven by the existing `AdaptiveConcurrencyController`: **halve on a reported throttle, +1 per clean outcome**, clamped to `[1, ceiling]` — so it converges on the real grant with zero configuration.
- Throttle detection feeds through `reportRateOutcome`, which is the same entry point `ctx.RateLimitReport` routes to — so a 429 the connector absorbed inside its own retry still teaches the gate.
- **Opt-in**: the gate only exists when a ceiling is declared — `Configuration.fetchConcurrency` (per-connection override) or the connector's `MaxConcurrencyHint`. No declaration → no gate, `withFetchGate` passes straight through: connectors that never showed this failure mode are byte-for-byte unchanged. The gate is rebuilt if the configured ceiling changes.
- A shrunken cap constrains new admissions immediately; slots release in `finally`, so a throwing fetch never leaks one.

## Tests

`src/__tests__/AdaptiveFetchGate.test.ts` — 7 tests, all green:
- 12 concurrent callers never exceed a cap of 3; all complete
- ceiling comes from the connector hint; **no declared ceiling → no gate, fully ungated (12 simultaneous callers all in flight)**
- a reported throttle halves the cap (8→4→2), via the same path connector-reported throttles use
- clean outcomes creep the cap back up, clamped at the ceiling
- a shrunken cap immediately constrains new admissions
- a throwing fetch releases its slot (no deadlock)

`tsc --noEmit`: 0 new errors (11 pre-existing on `next`, unchanged).


MS-BC:
Amended: the gate is now **opt-in**. It only exists when a ceiling is declared — `Configuration.fetchConcurrency` on the connection or the connector's `MaxConcurrencyHint`. A connector that declares neither gets no gate at all (`withFetchGate` passes straight through), so existing connectors are byte-for-byte unchanged. New test pins the ungated pass-through (12 simultaneous callers, all in flight at once). 7/7 tests green.

PR #4089, opened 2026-08-28
feat(integration-engine): pipelined page prefetch for cursor-paged connectors

MS-BC:
> Stacked on #4088 (uses its fetch gate); merge that first — GitHub will retarget this to `next` when the base branch merges.

## What

For cursor-paged connectors, the next page starts downloading while the current one is mapped and written. The fetch loop was strictly serial even though the next cursor is known the moment a page arrives; pipelining hides the shorter leg under the longer — **~20-30% cycle reduction** measured at a ~6s fetch / ~1-2s process split.

## How

- The fetch envelope is extracted into `governedFetch`: rate-limit token → adaptive fetch gate → per-attempt timeout → transient-only retry with Retry-After pacing → **one throttle report per episode**. Loop-top fetch and prefetch both run through it, so a prefetched page is indistinguishable from a loop-top fetch in pacing and error semantics — no second, unguarded path to the vendor.
- The prefetch context is built from the **fully advanced** position: `CurrentPage/CurrentOffset/CurrentCursor` from the batch's `Next*` values and `AfterKeyValue: NextAfterKeyValue ?? ctx.AfterKeyValue`. This last one matters: spreading ctx with only the cursor advanced leaves a keyset connector's seek key stale, so "page N+1" re-runs page N's seek, the duplicate-batch fingerprint kills the walk, and every keyset object stops at exactly two server pages (observed live before this shape).
- At most ONE page in flight ahead; keyed by the cursor it was built from, so a drifted position (gap-skip/reset) discards it instead of consuming the wrong page. A prefetch error surfaces when awaited next iteration (never an unhandled rejection). Cursor mode only — offset/page modes interact with gap-skip resume and stay serial.
- Kill switch: `MJ_INTEGRATION_PREFETCH=off`.

## Tests

`src/__tests__/GovernedFetch.test.ts` — 4 green:
- clean fetch: one connector call, one rate-limit token, no self-reported outcomes (the caller reports once per consumed page)
- persistent 429: 3 attempts, Retry-After pacing, **one** adaptive decrease for the episode, every retry re-passes the rate limiter
- our own page timeout is terminal (no retry stacks a second full page on a too-slow source)
- a reset socket is retried and is not a throttle

`tsc --noEmit`: 0 new errors (11 pre-existing on `next`, unchanged).

### DIFF
--- a/packages/Integration/engine/src/IntegrationEngine.ts
+++ b/packages/Integration/engine/src/IntegrationEngine.ts
@@ -2086,6 +2086,76 @@ export class IntegrationEngine extends BaseSingleton<IntegrationEngine> {
         }
     }
 
+    /**
+     * One governed vendor fetch: rate-limit token, adaptive fetch gate, per-attempt timeout,
+     * transient-only retry with Retry-After pacing, and ONE multiplicative decrease per throttle
+     * EPISODE (not per rejected attempt — a 429 that survives three retries is three rejections
+     * but one congestion event, the same distinction TCP draws when it halves the window once per
+     * loss event). Extracted so the loop-top fetch and the pipelined prefetch (which starts the
+     * next page while the current one is processed) share EXACTLY the same pacing and error
+     * semantics — a prefetched page must be indistinguishable from a loop-top fetch to the vendor.
+     *
+     * Retry predicate: our OWN timeout is terminal for this page (WithTimeout is a Promise.race
+     * with no cancellation, so the abandoned attempt keeps running — retrying stacks a second full
+     * page of vendor requests on a source already too slow to finish one); a transport error is
+     * not (a reset socket IS worth retrying). A throttle honors the source's Retry-After via
+     * DelayForError, and every retry re-passes the rate limiter via BeforeRetry so it cannot
+     * bypass a freeze the throttle just applied.
+     */
+    private async governedFetch(
+        config: RunConfiguration,
+        ctx: FetchContext,
+        objectName: string,
+        fetchTimeoutMs: number,
+        batchIndex: number,
+        logger?: SyncLogger,
+    ): Promise<FetchBatchResult> {
+        let throttleReported = false;
+        try {
+            await this.rateLimit(config);
+            return await this.withFetchGate(config, () => WithRetry(
+                () => WithTimeout(
+                    config.connector.FetchChanges(ctx),
+                    fetchTimeoutMs,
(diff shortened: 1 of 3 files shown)


MS-BC:
Full engine suite note: 40 tests across 5 files fail identically on clean `next` (verified by swapping in origin/next's IntegrationEngine.ts and re-running) — pre-existing, none introduced here. This branch adds 4 new tests, all green; suite totals go 833→844 with the same 40 failures throughout.

packages/Integration/engine/src/IntegrationEngine.ts
```diff
@@ -2449,95 +2525,58 @@ export class IntegrationEngine extends BaseSingleton<IntegrationEngine> {
...
-                        // limiter entirely — including the freeze the line above just applied.
-                        BeforeRetry: () => this.rateLimit(config),
-                    },
-                ));
+                if (prefetchedNext && prefetchedNext.key === (currentCursor ?? '')) {
+                    // The page already in flight IS this page — consume it. The rate limiter, the
+                    // fetch gate, the timeout/retry envelope, and once-per-episode throttle
+                    // reporting all ran inside governedFetch when the prefetch was launched, so
+                    // consuming it here adds no vendor pressure and loses no error semantics.
+                    const inFlight = prefetchedNext;
```
MS-BC, line 2534: LGTM, fetches if writing else just catches download as in flight

packages/Integration/engine/src/IntegrationEngine.ts
```diff
@@ -2449,95 +2525,58 @@ export class IntegrationEngine extends BaseSingleton<IntegrationEngine> {
...
+                    // previous seek: page N+1 was page N again, the duplicate-batch fingerprint
+                    // killed the walk, and every keyset object stopped at exactly two server pages.
+                    const nextCtx: FetchContext = {
+                        ...ctx,
+                        CurrentPage: batch.NextPage,
+                        CurrentOffset: batch.NextOffset,
+                        CurrentCursor: batch.NextCursor,
+                        AfterKeyValue: batch.NextAfterKeyValue ?? ctx.AfterKeyValue,
+                    };
+                    const nextPage = this.governedFetch(config, nextCtx, entityMap.ExternalObjectName, fetchTimeoutMs, batchCount + 1, logger);
```
MS-BC, line 2567: lgtm, its just fire and forget fetch and holds promise result whilst post processing of record is happening

PR #4090, opened 2026-08-28
feat(codegen-lib): per-schema creation-time entity flags (DefaultsBySchema)

MS-BC:
## What

`newEntityDefaults.DefaultsBySchema` — per-schema overrides for the flags CodeGen stamps on a newly created Entity row: `TrackRecordChanges`, `SupportsGeoCoding`, `AutoUpdateSupportsGeoCoding`. The two geo flags are also accepted as global creation defaults alongside the existing `TrackRecordChanges`.

## Why

Integration mirror schemas receive high-volume synced data, and the platform's creation defaults charge every write per-record side trips: `TrackRecordChanges` writes a Record Change row per save, and the geo auto-detect pass flips `SupportsGeoCoding` on for address-like columns, adding a geocoding lookup per save. Measured on a live high-volume sync, these side trips (not SQL, not I/O capacity) dominated wall time. Flipping the flags after creation works but has to be re-done for every new table the connector discovers; the right place is the creation default, scoped to the schema.

## How

- Matching follows `AllowCachingBySchema` exactly: case-insensitive, `${mj_core_schema}` placeholder expanded.
- A per-schema entry only overrides the flags it names; everything else falls through to the global defaults.
- Unset flags are **omitted from the INSERT** so the database default applies — existing configurations produce byte-for-byte the same SQL (pinned by test).
- Creation-time `AutoUpdateSupportsGeoCoding=false` permanently shields the entity: the geo auto-detect pass honors the flag as a lock, so address-like columns never flip `SupportsGeoCoding` on later.
- Dialect-neutral: the INSERT builder goes through `qi`/`boolLit`, so SQL Server and PostgreSQL both get correct literals (both covered by tests).

Example:

```js
newEntityDefaults: {
  DefaultsBySchema: [
    { SchemaName: 'netsuite', TrackRecordChanges: false, SupportsGeoCoding: false, AutoUpdateSupportsGeoCoding: false }
  ]
}
```

## Tests

`src/__tests__/new-entity-defaults-by-schema.test.ts` — 4 green:
- a listed schema gets all three flags false at creation (SQL Server literals)
- same on PostgreSQL with boolean literals
- an unlisted schema is untouched: global `TrackRecordChanges` applies, geo columns omitted
- a per-schema entry only overrides the flags it names (`${mj_core_schema}` expansion covered)

`tsc --noEmit`: clean. Full CodeGenLib suite: 3 failures in `entity-subclass-embedded.test.ts` are pre-existing on `next` (identical with this change stashed); no new failures.

### DIFF
--- a/.changeset/new-entity-defaults-by-schema.md
+++ b/.changeset/new-entity-defaults-by-schema.md
@@ -1,0 +1,5 @@
+---
+'@memberjunction/codegen-lib': patch
+---
+
+Per-schema creation-time entity flags: `newEntityDefaults.DefaultsBySchema` lets a schema override `TrackRecordChanges`, `SupportsGeoCoding`, and `AutoUpdateSupportsGeoCoding` on the Entity row CodeGen creates (matching AllowCachingBySchema's rules: case-insensitive, `${mj_core_schema}` placeholder). `SupportsGeoCoding` / `AutoUpdateSupportsGeoCoding` are also accepted as global creation defaults. Unset flags are omitted from the INSERT so the database default applies — existing configurations produce byte-for-byte the same SQL. The intended use is integration mirror schemas: entities receiving high-volume synced data should not pay per-record side trips (record-change rows, geocoding lookups) on every write, and creation-time `AutoUpdateSupportsGeoCoding=false` permanently shields them from the geo auto-detect pass, which honors the flag as a lock.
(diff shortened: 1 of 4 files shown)


MS-BC:
Closing in favor of #4091, which models this correctly. Record Changes and geocoding are per-WRITE side effects, and this PR suppressed them the only way the framework allowed — permanently, at entity creation, for every writer. #4091 scopes the suppression to the writer instead: the sync's own saves carry EntitySaveOptions.SkipRecordChanges/SkipGeoCoding (per-connection opt-in, fail-closed) while the entity flags stay on — so a human editing a synced record is still audited and geocoded, including in mirror schemas, where this PR would have silently lost that. One scoped mechanism supersedes the creation-time defaults.

PR #4091, opened 2026-08-28
feat: sync-scoped suppression of per-write side effects (audit rows, geocoding)

MS-BC:
## The capability model this fixes

Record Changes and geocoding are per-WRITE side effects, but the framework only exposed them as per-ENTITY flags. So a high-volume machine writer (an integration sync applying tens of thousands of records in minutes — where these side trips measurably dominate wall time) had exactly one lever: turn the flags off — which also turns them off for every human and API writer of the same entities, permanently. The suppression belongs to the **writer**, not the entity.

## What changes

- **MJCore**: `EntitySaveOptions.SkipRecordChanges`, `EntitySaveOptions.SkipGeoCoding`, `EntityDeleteOptions.SkipRecordChanges`. The abstract `GenerateSaveSQL`/`GenerateDeleteSQL` gain an optional trailing `options` parameter (back-compat: existing 3-arg overrides still compile), and the base Save/Delete flows pass it through.
- **GenericDatabaseProvider**: the record-change wrap in `GenerateSaveSQL` and the `geoSyncNeeded` flag in `OnBeforeSaveExecute` honor the options.
- **PostgreSQL / SQL Server providers**: the delete builders honor `SkipRecordChanges` (PG's audit CTE, SS's `GetDeleteSQLWithDetails` wrap).
- **Integration engine**: when a connection sets `Configuration.writeSideEffects === 'suppressed'`, the run stamps a flag on its run context and every save/delete the sync performs carries the options — create, update, archive, tombstone, conflict-mark, delete. **Fail-closed**: absent, unparseable, or wrongly-typed configuration keeps the side effects on; and the options are built only inside a suppressing run's AsyncLocalStorage scope, so an interactive save can never carry them, even mid-sync on the same entities.

## What deliberately does NOT change

- Entity flags are untouched — a human editing a synced record through the UI is still audited and still geocoded, including while a suppressing sync is running.
- Entity Actions and AI Actions are untouched (they already have their own Skip options and are part of Save()'s semantics).
- Nothing changes for any deployment until a connection explicitly asks.

## Tests (9 new, all green)

- Engine: the config ask fails closed across 6 malformed shapes; options exist only inside a suppressing run's scope; outside any run context they are undefined (suppression cannot leak to other writers).
- GenericDatabaseProvider: the same save on the same geo entity flags `geoSyncNeeded` without the option and not with it — provably per-save, not per-entity.
- PostgreSQL: the same tracked entity's delete carries the audit CTE without the option and is byte-identical to the plain CRUD call with it.

Full suites on every touched package: MJCore 2151 ✓, GenericDatabaseProvider 965 ✓, PostgreSQL 196 ✓, SQL Server 114 ✓, Integration engine 845 ✓. `tsc` clean.


MS-BC:
@rkihm-BC I think this could be somewhat controversial but I wanted to get thoughts on this given how much this can speed up syncing

packages/MJServer/src/generic/DeleteOptionsInput.ts
```diff
@@ -37,4 +37,14 @@ export class DeleteOptionsInput {
...
+
+    /**
+     * When true, the delete skips writing its Record Change (audit) row even when the entity
+     * has `TrackRecordChanges` on. Set by high-volume MACHINE writers (an integration sync
+     * applying tens of thousands of records), never by interactive deletes — the suppression is
+     * a property of the WRITER, not of the entity, so a human deleting the same record is still
+     * audited.
+     */
+    @Field(() => Boolean)
+    SkipRecordChanges: boolean;
```
rkihm-BC, line 55: This is the line that puts audit suppression on the public API. `ResolverBase.DeleteRecord` passes the client-supplied `options` straight into `entityObject.Delete(options)`, and the only gate before it — `CheckAPIKeyScopeAuthorization('entity:delete', ...)` — asks *may you delete*, never *may you delete without an audit row*.

The save twin has no equivalent: there is no `SaveOptionsInput` at all, and `graphQLDataProvider` sends `options___` only on the delete mutation. So `EntitySaveOptions.SkipRecordChanges` / `SkipGeoCoding` are unreachable from the wire while this one is honoured.

I don't think that asymmetry was chosen — `schema-sync.test.ts`'s *"should match EntityDeleteOptions fields in MJCore"* assertion requires every core field to appear on this InputType, so adding it to the core class conscripted it onto the schema automatically.

Suggest leaving the field here (keeps the drift gate green) and sanitising in `ResolverBase.DeleteRecord` instead — see the review body for both variants and the question of which applies.

rkihm-BC (commented):
You flagged this as possibly controversial — I think you're right, and I've found where. The design is sound; the concern is one line of GraphQL surface that I don't think was a deliberate choice.

## What's good

**The capability model is correctly diagnosed.** Record Changes and geocoding are per-*write* costs exposed only as per-*entity* flags, so a high-volume machine writer's only lever was turning them off for every human writer of the same entities too. Moving the suppression to the writer is the right abstraction, and "the suppression belongs to the writer, not the entity" is exactly the sentence that makes it obvious in hindsight.

**Fail-closed throughout, and genuinely so.** `ReadWriteSideEffects` returns `''` for absent, unparseable, or wrongly-typed configuration — every failure mode keeps the side effects **on**. A connection has to ask, and a malformed request isn't an ask. Six malformed shapes are tested.

**The AsyncLocalStorage scoping is the part that makes this safe.** Options are constructed only inside a suppressing run's scope, so an interactive save on the same entity mid-sync cannot pick them up. That's the property that lets you keep the entity flags on, and it's tested rather than asserted.

**The provider tests prove the right thing** — the same save on the same geo entity flags `geoSyncNeeded` without the option and not with it, and PG's delete is byte-identical to the plain CRUD call with the option set. Per-save, demonstrated.

## The concern: save and delete end up with different reachability

`EntityDeleteOptions.SkipRecordChanges` is now a `@Field(() => Boolean)` on `DeleteOptionsInput`, and `ResolverBase.DeleteRecord` passes the client-supplied `options` straight into `entityObject.Delete(options)`:

```ts
await this.CheckAPIKeyScopeAuthorization('entity:delete', entityName, userPayload);
...
if (await entityObject.Delete(options)) {
```

That scope check asks *may you delete* — never *may you delete without an audit row*. So as it stands, **any caller authorized to delete can suppress the Record Change for that delete**, overriding an admin's entity-level `TrackRecordChanges` decision per call.

The save twin has no equivalent exposure, and the difference looks structural rather than intentional:

| | Save | Delete |
|---|---|---|
| GraphQL input exists? | **No** — there is no `SaveOptionsInput` in `packages/MJServer/src/generic/` | Yes — `DeleteOptionsInput` |
| Options on the wire? | None. `graphQLDataProvider` sends `RestoreContext___` / `OldValues___` on the mutation input; `options___: DeleteOptionsInput!` exists only on the delete mutation | `options___` carries the full object |
| Can any client suppress? | Impossible to express | Expressible and honoured |

And I think a test gate forced it rather than a decision. `schema-sync.test.ts` asserts:

```js
it('should match EntityDeleteOptions fields in MJCore', () => {
    coreFields.forEach(coreField => expect(serverFieldNames).toContain(coreField));
});
```

Every field on the core class must appear on the InputType. Adding `SkipRecordChanges` to `EntityDeleteOptions` therefore conscripted it onto the public API automatically — which also explains why the final commit is *"mirror EntityDeleteOptions.SkipRecordChanges into the GraphQL input"* rather than something that reads like a deliberate API decision. The save side escaped only because it has no input type to mirror into.

## The question that decides the fix

**Is the GraphQL delete-suppression path speculative, or does something need it?**

I looked and couldn't find a consumer:

- `IntegrationEngine` — the entire motivating scenario — **never touches GraphQL**. It calls `run.Save()` / `entity.Delete(opts)` against an in-process `DatabaseProviderBase`, and there's no `graphQLSystemUserClient` import anywhere under `packages/Integration/engine/src`. The sync is unaffected either way.
- Nothing calls the system-user client's `Delete` with `SkipRecordChanges`. The PR wires `DeleteQueryOptionsInput` to carry it, but I found no caller.

### If it's speculative — ignore it unconditionally in the resolver

```ts
// SkipRecordChanges is an IN-PROCESS capability for machine writers. It is mirrored onto this
// InputType only because schema-sync.test.ts requires DeleteOptionsInput to carry every
// EntityDeleteOptions field — no client may suppress an audit row. This matches the save path,
// where no SaveOptionsInput exists and the option is unreachable from the wire. IntegrationEngine
// sets it in-process and never crosses this resolver.
if (options?.SkipRecordChanges) {
    LogError(`Ignoring SkipRecordChanges on ${entityName} delete from ${userPayload.email}: the audit row cannot be suppressed over GraphQL.`);
    options = { ...options, SkipRecordChanges: false };
}
```

One edit in `ResolverBase.DeleteRecord` covers every generated `DeleteX` mutation, keeps the drift gate green (the field stays on the InputType), and gives **exact parity** with save: neither path can suppress from the wire, both retain it in-process.

Downgrade rather than throw, deliberately — the thing being protected is the audit row, so the best outcome is *delete proceeds, audit is written, attempt is logged*. Throwing yields no delete **and** no audit row.

### If something does need it — gate on `isSystemUser`

Same edit, conditioned:

```ts
if (options?.SkipRecordChanges && !userPayload.isSystemUser) { ...downgrade... }
```

`UserPayload.isSystemUser` is set only in `context.ts:471`, on a constant-time comparison against the superadmin system API key, and it's already the gate used by `@RequireSystemUser` and `ExecuteRemoteOperationResolver`. `graphQLSystemUserClient` authenticates that way, so the intended caller keeps working.

Worth being explicit that this option leaves delete **more permissive than save** — a system-user GraphQL caller could suppress on delete, and there is no save equivalent at all. That's defensible if a remote system sync genuinely needs it, but it should be a documented choice rather than an inherited one.

## My recommendation

Go unconditional now, and relax to the `isSystemUser` gate later if a remote system-sync caller actually appears — so the permissive version arrives with a named consumer rather than ahead of one, and the restrictive default is what ships meanwhile.

Either way, worth a resolver-level test asserting a client-supplied `SkipRecordChanges: true` reaches `entityObject.Delete` as `false`. That's what stops the sanitizer being "simplified" away later.

Everything else here I'd approve as-is — the in-process design, the fail-closed config read, and the AsyncLocalStorage scoping are all right.

MS-BC:
You were right, and it was worse than the review found — thank you for catching it.

## Fixed, unconditionally

`DeleteOptionsInput.SanitizeFromWire()` downgrades `SkipRecordChanges` to false and logs the attempt. Downgrade rather than throw, for exactly the reason you gave: the protected thing is the audit row, so the right outcome is *delete proceeds, audit is written, attempt logged*.

I put the sanitizer as a static **on `DeleteOptionsInput` itself** rather than inline in the resolver, so it sits next to the field declaration — the next person who mirrors a core field onto this InputType because the schema-sync gate told them to reads the reason at the point of the mistake.

Your diagnosis of the cause is precisely right, and I'd add it was worse than accidental: the commit message *"mirror EntityDeleteOptions.SkipRecordChanges into the GraphQL input"* is me describing a mechanical fix to a red test. The gate said "this field must appear there", I made it appear there, and I never asked what appearing there *meant*. The test was enforcing shape parity and I read it as authority on API design.

## One thing the review missed: `ResolverBase.DeleteRecord` is not the only wire entry

`FileCategoryResolver.DeleteFileCategory` takes `@Arg('options___', () => DeleteOptionsInput)` and calls `fileCategoryEntity.Delete(options)` itself — it never routes through `DeleteRecord`. A single edit there would have left File Categories suppressible. Both entry points now sanitize.

I checked the other paths that reach `.Delete(options)`:
- `rest/EntityCRUDHandler.deleteEntity` — constructs its own `EntityDeleteOptions` and only copies an explicit allowlist from the request body; `SkipRecordChanges` isn't in it. Safe by construction, and no production caller. Left alone.
- `MJEntityPermissionEntityServer.server.ts` — `super.Delete(options)` inside an entity subclass, downstream of the sanitize.

The **save** side needed nothing, which confirms your parity table: there is no `SaveOptionsInput`, the mutation inputs carry only `RestoreContext___` / `OldValues___`, and `EntityCRUDHandler`'s save allowlist doesn't include the new options either. Inexpressible from the wire, as intended.

## On speculative vs. needed

Agreed it's speculative — I found no caller either, and `IntegrationEngine` never touches GraphQL. Went unconditional per your recommendation, so the permissive version can arrive later with a named consumer rather than ahead of one.

## Tests

`src/__tests__/DeleteOptionsSanitize.test.ts` — 8, all green:

- forces `SkipRecordChanges: true` → false
- downgrades rather than throws (the audit row is the point)
- logs the attempt with entity + caller
- leaves every other option exactly as sent
- no-op for the ordinary request: returns the *same object*, logs nothing (no noise on normal traffic)
- survives a caller with no identifiable email
- **the wiring pin you asked for**: a client-supplied `SkipRecordChanges: true` reaches `entityObject.Delete` as `false`, driving the real `ResolverBase.DeleteRecord`
- an ordinary delete's other options pass through untouched

The wiring pin is mutation-tested: with the `SanitizeFromWire` call removed from `DeleteRecord` it goes **red**, and green again when restored — so it can't be simplified away silently.

Changeset updated to state the wire behaviour explicitly rather than leaving it inferable from the diff.

rkihm-BC (approved):
Approving. The concern is closed, and closed better than I specified. Waited for a clean board before posting — `Integration (SQL Server, deterministic)`, `Build`, all six unit shards and the driver-integration set are green, `mergeStateStatus: CLEAN`.

## You found the entry point I missed

I recommended sanitizing in `ResolverBase.DeleteRecord` on the reasoning that every generated `DeleteX` routes through it. That's true, and it's incomplete: **`FileCategoryResolver.DeleteFileCategory` accepts a `DeleteOptionsInput` and calls `fileCategoryEntity.Delete(options)` itself**, never touching `DeleteRecord`. Following my advice literally would have left that path open — the one hand-written delete resolver in the codebase, and therefore the one an audit-suppression attempt would most plausibly find.

I swept every resolver at the PR head to check the set is now complete: those two are the only wire entry points that reach `.Delete(options)`, and both are sanitized.

Putting `SanitizeFromWire` as a static on `DeleteOptionsInput` is also the better home than the inline block I sketched — the sanitizer lives with the type it sanitizes, so a third entry point added later has an obvious thing to call, and the doc comment naming both current call sites tells that author what to do.

## The specifics that matter are all preserved

- **Downgrade, not throw**, with the reasoning intact: the thing being protected is the audit row, so the right outcome is delete proceeds, audit is written, attempt is logged. Throwing yields no delete *and* no audit row.
- **The field doc now says it isn't honoured over the wire**, and explains that it exists on the InputType only because the schema-sync gate mirrors every `EntityDeleteOptions` field onto it — so the next reader doesn't have to re-derive why a field is present but inert.
- **Save-path parity is stated where it will be read.** `EntitySaveOptions.SkipRecordChanges` / `SkipGeoCoding` have no `SaveOptionsInput` and are inexpressible over GraphQL; suppression stays what it was designed to be — an in-process capability `IntegrationEngine` sets on its own saves, never crossing a resolver.

## The test does the thing that actually matters

`DeleteOptionsSanitize.test.ts` splits into six unit cases on `SanitizeFromWire` — forces `true` to `false`, downgrades rather than throws, logs naming entity and caller, leaves every *other* option untouched, is a genuine no-op for an ordinary request, and survives a caller with no identifiable email — plus a second describe titled **"ResolverBase.DeleteRecord — the sanitize is wired, not just available"**, asserting a client-supplied `true` reaches `entityObject.Delete` as `false`.

That second one is the guard that counts. A sanitizer that exists but isn't called is exactly how this regresses, and the test title says so out loud.

## Standing from the first round

The parts I verified before still hold, and they're what make the feature safe to enable: `ReadWriteSideEffects` fails closed on absent, unparseable and wrongly-typed configuration; the options are constructed only inside a suppressing run's AsyncLocalStorage scope, so an interactive save on the same entity mid-sync cannot pick them up; and the provider tests prove `geoSyncNeeded` is per-save rather than per-entity.

You called this one possibly controversial. The design never was — the one genuinely contentious line was a GraphQL field a drift gate conscripted, and it's now inert on the wire by construction and by test.

PR #4095, opened 2026-08-28
feat(conversations): clickable record-link tokens with optional labels

AN-BC:
## Summary
- Agents can emit `@{…}` `type: "record"` tokens in Chat `message`. The conversation renderer turns them into clickable pills; the host opens the row via `OpenEntityRecord`.
- Primary keys are sibling fields on the token (composite keys included). `open:resource` Record commands accept the same `keys` map.
- `name` is **optional**. Omit it when surrounding prose already names the record — the pill renders as an icon with an arrow. Include `name` only when the token *is* the mention.
- Loop Agent Type system prompt teaches the grammar and asks for world-class UX: prefer icon-only, one citation per record per section, no heading-plus-body duplicates.

## Test plan
- [ ] Paste a labeled token (`"name": "Acme Corp"`) into a Chat message and confirm a blue name pill opens the record.
- [ ] Paste an icon-only token (no `name`) after prose that already names the row and confirm a compact arrow pill opens the same record.
- [ ] Confirm a composite-PK token (sibling fields, no `resourceId`) opens the correct row.
- [ ] Confirm two nearby tokens for the same record are a prompt/UX issue, not a renderer crash; headings stay text.
- [ ] `pnpm test` in `packages/AI/CorePlus` (record-link cases, including nameless tokens).

### DIFF
--- a/.changeset/clickable-record-links.md
+++ b/.changeset/clickable-record-links.md
@@ -1,0 +1,7 @@
+---
+"@memberjunction/ai-core-plus": minor
+"@memberjunction/ng-conversations": minor
+"@memberjunction/ng-explorer-core": minor
+---
+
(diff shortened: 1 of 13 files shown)


AN-BC (commented):
lGTM

PR #4097, opened 2026-08-28
fix(conversations): stop Unknown error + stuck Running timer on agent fail

AN-BC:
## Summary

After approving a Skip dashboard PRD, Explorer showed **Skip failed / Unknown error** while the AIAgentRun stayed **Running** and the conversation detail stayed **In-Progress** — so the red-pill timer never stopped.

Root cause on the MJ side:

- `invokeSubAgent` / `ConversationAgentRunner` discarded the failed `ExecuteAgentResult` and returned `null`. The bubble always fell through to `result?.agentRun?.ErrorMessage || 'Unknown error'`.
- A dropped HTTP/WebSocket (fire-and-forget) often means the **server is still running**. Painting Status=Error on the client while the server keeps writing In-Progress is how you get a failed bubble + a live timer.
- If `executeAIAgent` / `RunAgentInConversation` threw after creating the run, nothing persisted Failed/Error on the in-flight records.

This PR:

- Passes the failed result through (with `errorMessage`) so the bubble can show the real transport text.
- If that text says the agent may still be running (lost connection / please refresh), **keeps In-Progress** so later progress/completion can land.
- On a server throw, persists Failed on the AIAgentRun and Error on the conversation detail.

Skip-side 5-minute abort + missing Skip System person are a separate PR: https://github.com/BlueCypress/Skip-Brain/pull/525

## Test plan

- [x] `@memberjunction/ai-core-plus` vitest: `agent-failure-message.test.ts`
- [x] `@memberjunction/conversations-runtime` vitest: ConversationAgentRunner failure now returns `{ success: false, errorMessage }` instead of null
- [ ] @mention Skip, drop the WebSocket mid-run: bubble should stay In-Progress with “still running on the server”, not Unknown error + a stuck timer
- [ ] Genuine Skip pipeline failure: bubble shows the real ErrorMessage and Status=Error
- [ ] `executeAIAgent` throw after the run exists: AIAgentRun is Failed and the conversation detail is Error

### DIFF
--- a/.changeset/skip-unknown-error-stuck-timer.md
+++ b/.changeset/skip-unknown-error-stuck-timer.md
@@ -1,0 +1,10 @@
+---
+"@memberjunction/ai-core-plus": patch
+"@memberjunction/ng-conversations": patch
+"@memberjunction/conversations-runtime": patch
+"@memberjunction/graphql-dataprovider": patch
+"@memberjunction/server": patch
+"@memberjunction/ai-agents": patch
+---
+
+Stop Explorer from showing "Unknown error" with a stuck Running timer when a Skip/sub-agent transport path fails. Pass the real error through invokeSubAgent, keep In-Progress when the agent may still be running, and persist Failed/Error on the run and conversation detail if executeAIAgent throws.
--- a/packages/AI/Agents/src/AgentRunner.ts
+++ b/packages/AI/Agents/src/AgentRunner.ts
@@ -201,11 +201,12 @@ export class AgentRunner {
             throw new Error('contextUser is required for RunAgentInConversation');
         }
 
+        let agentResponseDetail: MJConversationDetailEntity | undefined;
+
         try {
             let conversationId: string;
             let userMessageDetailId: string;
             let agentResponseDetailId: string | undefined;
-            let agentResponseDetail: MJConversationDetailEntity | undefined;
 
             // If conversationDetailId is provided, use it (UI-created agent response detail)
             if (options.conversationDetailId) {
@@ -591,6 +592,17 @@ export class AgentRunner {
         } catch (error) {
             const errorMessage = error instanceof Error ? error.message : 'Unknown error occurred';
             LogError(`RunAgentInConversation failed: ${errorMessage}`, undefined, error);
+            if (agentResponseDetail && agentResponseDetail.Status === 'In-Progress') {
+                try {
+                    await agentResponseDetail.EnsureSaveComplete();
+                    agentResponseDetail.Status = 'Error';
+                    agentResponseDetail.Message = errorMessage;
+                    agentResponseDetail.Error = errorMessage;
+                    await agentResponseDetail.Save();
+                } catch (persistError) {
+                    LogError(`Failed to persist Error on conversation detail after agent crash: ${persistError}`, undefined, persistError);
+                }
+            }
             throw error;
         }
     }
--- a/packages/AI/CorePlus/src/__tests__/agent-failure-message.test.ts
+++ b/packages/AI/CorePlus/src/__tests__/agent-failure-message.test.ts
@@ -1,6 +1,8 @@
 import { describe, expect, it } from 'vitest';
 import {
+    agentFailureDisposition,
     agentFailureMessage,
+    coerceFailedExecuteAgentResult,
     isDisconnectWhileAgentMayStillBeRunning,
 } from '../agent-failure-message';
 
@@ -37,5 +39,61 @@ describe('isDisconnectWhileAgentMayStillBeRunning', () => {
         expect(isDisconnectWhileAgentMayStillBeRunning(
             "Pipeline failed at step 'Execute Sub-Agent: Skip: Data Expert'"
         )).toBe(false);
+    });
+
+    it('does not treat raw Failed to fetch as in-flight (request may never have left the browser)', () => {
+        expect(isDisconnectWhileAgentMayStillBeRunning('Failed to fetch')).toBe(false);
+        expect(isDisconnectWhileAgentMayStillBeRunning('NetworkError when attempting to fetch resource.')).toBe(false);
+    });
+
+    it('is false when the transport knows the request was never acknowledged', () => {
+        expect(isDisconnectWhileAgentMayStillBeRunning(
+            'Lost connection to the server. The agent may still be running. Please refresh to check the latest status.',
+            { requestAcknowledged: false }
+        )).toBe(false);
+    });
+});
+
+describe('agentFailureDisposition', () => {
+    it('keeps In-Progress for a post-ACK disconnect so a later server Complete is not stomped', () => {
+        expect(agentFailureDisposition({
+            errorMessage: 'Lost connection to the server. The agent may still be running. Please refresh to check the latest status.',
+            requestAcknowledged: true,
+        })).toEqual({
+            status: 'In-Progress',
+            message: 'Lost connection to the server. The agent may still be running. Please refresh to check the latest status.',
+        });
+    });
+
+    it('paints Error when the request never left the browser', () => {
+        expect(agentFailureDisposition({
+            errorMessage: 'Failed to fetch',
+            requestAcknowledged: false,
+        }).status).toBe('Error');
+    });
+
+    it('paints Error for a real agent failure', () => {
+        expect(agentFailureDisposition({
+            errorMessage: "Pipeline failed at step 'Execute Sub-Agent: Skip: Data Expert'",
+        })).toEqual({
+            status: 'Error',
+            message: "Pipeline failed at step 'Execute Sub-Agent: Skip: Data Expert'",
+        });
+    });
+});
+
+describe('coerceFailedExecuteAgentResult', () => {
+    it('copies the source and forces success false without mutating the caller', () => {
+        const source = { success: true, errorMessage: 'from result', payload: { a: 1 } };
+        const failed = coerceFailedExecuteAgentResult(source, 'fallback');
+        expect(failed).toEqual({ success: false, errorMessage: 'from result', payload: { a: 1 } });
+        expect(source.success).toBe(true);
+    });
+
+    it('uses the fallback when the result has no error text', () => {
+        const failed = coerceFailedExecuteAgentResult({ success: true, payload: { x: 1 } }, 'envelope failed');
+        expect(failed.success).toBe(false);
+        expect(failed.errorMessage).toBe('envelope failed');
+        expect(failed.payload).toEqual({ x: 1 });
     });
 });
--- a/packages/AI/CorePlus/src/__tests__/agent-failure-message.test.ts
+++ b/packages/AI/CorePlus/src/__tests__/agent-failure-message.test.ts
@@ -1,0 +1,41 @@
+import { describe, expect, it } from 'vitest';
+import {
+    agentFailureMessage,
+    isDisconnectWhileAgentMayStillBeRunning,
+} from '../agent-failure-message';
+
+describe('agentFailureMessage', () => {
+    it('prefers agentRun.ErrorMessage over transport errorMessage', () => {
+        expect(agentFailureMessage({
+            errorMessage: 'transport',
+            agentRun: { ErrorMessage: 'run failed' },
+        })).toBe('run failed');
+    });
+
+    it('falls back to transport errorMessage when the run has none', () => {
+        expect(agentFailureMessage({
+            errorMessage: 'Lost connection to the server. The agent may still be running.',
+            agentRun: undefined,
+        })).toBe('Lost connection to the server. The agent may still be running.');
+    });
+
+    it('does not return empty or Unknown error when a real message exists', () => {
+        expect(agentFailureMessage({ errorMessage: 'timeout' })).toBe('timeout');
+        expect(agentFailureMessage(null)).toBe('The agent failed without an error message.');
+        expect(agentFailureMessage({ errorMessage: '  ' })).toBe('The agent failed without an error message.');
+    });
+});
+
+describe('isDisconnectWhileAgentMayStillBeRunning', () => {
+    it('recognizes the GraphQLAIClient lost-connection copy', () => {
+        expect(isDisconnectWhileAgentMayStillBeRunning(
+            'Lost connection to the server. The agent may still be running. Please refresh to check the latest status.'
+        )).toBe(true);
+    });
+
+    it('is false for a genuine pipeline failure', () => {
+        expect(isDisconnectWhileAgentMayStillBeRunning(
+            "Pipeline failed at step 'Execute Sub-Agent: Skip: Data Expert'"
+        )).toBe(false);
+    });
+});
--- a/packages/AI/CorePlus/src/__tests__/agent-failure-message.test.ts
+++ b/packages/AI/CorePlus/src/__tests__/agent-failure-message.test.ts
@@ -37,7 +37,7 @@ describe('isDisconnectWhileAgentMayStillBeRunning', () => {
 
     it('is false for a genuine pipeline failure', () => {
         expect(isDisconnectWhileAgentMayStillBeRunning(
-            "Pipeline failed at step 'Execute Sub-Agent: Skip: Data Expert'"
+            "Pipeline failed at step 'Execute Sub-Agent: Specialist'"
         )).toBe(false);
     });
 
@@ -74,10 +74,10 @@ describe('agentFailureDisposition', () => {
 
     it('paints Error for a real agent failure', () => {
         expect(agentFailureDisposition({
-            errorMessage: "Pipeline failed at step 'Execute Sub-Agent: Skip: Data Expert'",
+            errorMessage: "Pipeline failed at step 'Execute Sub-Agent: Specialist'",
         })).toEqual({
             status: 'Error',
-            message: "Pipeline failed at step 'Execute Sub-Agent: Skip: Data Expert'",
+            message: "Pipeline failed at step 'Execute Sub-Agent: Specialist'",
         });
     });
 });
--- a/packages/AI/CorePlus/src/agent-failure-message.ts
+++ b/packages/AI/CorePlus/src/agent-failure-message.ts
@@ -5,6 +5,12 @@
 export type AgentFailureSource = {
     errorMessage?: string | null;
     agentRun?: { ErrorMessage?: string | null } | null;
+    /**
+     * True when the fire-and-forget mutation returned an ACK before the
+     * transport died. False when the request never left the browser (or the
+     * ACK never arrived). Undefined for callers that do not know.
+     */
+    requestAcknowledged?: boolean;
 } | null | undefined;
 
 /**
@@ -30,14 +36,58 @@ export function agentFailureMessage(
  * True when the failure text means the HTTP/WebSocket path died but the
  * server-side AIAgentRun may still be Running. Painting Status=Error in that
  * case leaves the Explorer timer stuck on In-Progress while the bubble says failed.
+ *
+ * Does **not** match raw `Failed to fetch` / `NetworkError` — those also fire
+ * when the initial mutation never reaches the server (no run exists). The
+ * GraphQL client rewrites a post-ACK disconnect to the "lost connection /
+ * still be running / please refresh" copy; only that copy (or an explicit
+ * `requestAcknowledged: true`) is treated as in-flight.
  */
-export function isDisconnectWhileAgentMayStillBeRunning(message: string): boolean {
+export function isDisconnectWhileAgentMayStillBeRunning(
+    message: string,
+    result?: AgentFailureSource
+): boolean {
+    if (result?.requestAcknowledged === false) {
+        return false;
+    }
     const m = message.toLowerCase();
     return (
         m.includes('still be running') ||
         m.includes('lost connection to the server') ||
-        m.includes('please refresh') ||
-        m.includes('failed to fetch') ||
-        m.includes('networkerror')
+        m.includes('please refresh')
     );
+}
+
+export type AgentFailureDisposition =
+    | { status: 'In-Progress'; message: string }
+    | { status: 'Error'; message: string };
+
+/**
+ * Decide whether a failed `ExecuteAgentResult` should keep the conversation
+ * detail In-Progress (server may still complete) or paint Error.
+ */
+export function agentFailureDisposition(
+    result: AgentFailureSource,
+    fallback?: string
+): AgentFailureDisposition {
+    const message = agentFailureMessage(result, fallback);
+    if (isDisconnectWhileAgentMayStillBeRunning(message, result)) {
+        return { status: 'In-Progress', message };
+    }
+    return { status: 'Error', message };
+}
+
+/**
+ * Copy of a failed agent result. Always `success: false` with a non-empty
+ * `errorMessage`. Does not mutate the caller's object.
+ */
+export function coerceFailedExecuteAgentResult<T extends { success?: boolean; errorMessage?: string | null; agentRun?: { ErrorMessage?: string | null } | null }>(
+    result: T | null | undefined,
+    fallback: string
+): T & { success: false; errorMessage: string } {
+    return {
+        ...(result ?? {}),
+        success: false,
+        errorMessage: agentFailureMessage(result, fallback),
+    } as T & { success: false; errorMessage: string };
 }
--- a/packages/AI/CorePlus/src/agent-failure-message.ts
+++ b/packages/AI/CorePlus/src/agent-failure-message.ts
@@ -1,0 +1,43 @@
+/**
+ * Client/transport failure text when `ExecuteAgentResult.agentRun` is missing
+ * (fire-and-forget timeout, dropped WebSocket, invokeSubAgent used to return null).
+ */
+export type AgentFailureSource = {
+    errorMessage?: string | null;
+    agentRun?: { ErrorMessage?: string | null } | null;
+} | null | undefined;
+
+/**
+ * Prefer the persisted run's ErrorMessage, then a transport-level errorMessage.
+ * Never returns empty string.
+ */
+export function agentFailureMessage(
+    result: AgentFailureSource,
+    fallback = 'The agent failed without an error message.'
+): string {
+    const fromRun = result?.agentRun?.ErrorMessage?.trim();
+    if (fromRun) {
+        return fromRun;
+    }
+    const fromResult = result?.errorMessage?.trim();
+    if (fromResult) {
+        return fromResult;
+    }
+    return fallback;
+}
+
+/**
+ * True when the failure text means the HTTP/WebSocket path died but the
+ * server-side AIAgentRun may still be Running. Painting Status=Error in that
+ * case leaves the Explorer timer stuck on In-Progress while the bubble says failed.
+ */
+export function isDisconnectWhileAgentMayStillBeRunning(message: string): boolean {
+    const m = message.toLowerCase();
+    return (
+        m.includes('still be running') ||
+        m.includes('lost connection to the server') ||
+        m.includes('please refresh') ||
+        m.includes('failed to fetch') ||
+        m.includes('networkerror')
+    );
+}
--- a/packages/AI/CorePlus/src/agent-types.ts
+++ b/packages/AI/CorePlus/src/agent-types.ts
@@ -685,6 +685,12 @@ export type ExecuteAgentResult<P = any> = {
      * fire-and-forget timeout). Prefer `agentRun.ErrorMessage` when the run exists.
      */
     errorMessage?: string;
+    /**
+     * True when the fire-and-forget mutation returned an ACK before the transport
+     * died (server has the run). False when the request never left the browser.
+     * Undefined when the caller does not know.
+     */
+    requestAcknowledged?: boolean;
     /** Optional payload returned by the agent */
     payload?: P;
     /**
--- a/packages/AI/CorePlus/src/agent-types.ts
+++ b/packages/AI/CorePlus/src/agent-types.ts
@@ -680,6 +680,11 @@ export type BaseAgentNextStep<P = any, TContext = any> = {
 export type ExecuteAgentResult<P = any> = {
     /** Whether the agent execution was successful */
     success: boolean;
+    /**
+     * Transport-level failure text when {@link agentRun} is missing (lost WebSocket,
+     * fire-and-forget timeout). Prefer `agentRun.ErrorMessage` when the run exists.
+     */
+    errorMessage?: string;
     /** Optional payload returned by the agent */
     payload?: P;
     /**
--- a/packages/AI/CorePlus/src/index.ts
+++ b/packages/AI/CorePlus/src/index.ts
@@ -1,5 +1,6 @@
 export * from './prompt.types';
 export * from './agent-types';
+export * from './agent-failure-message';
 export * from './agent-payload-change-request';
 export * from './prompt.system-placeholders';
 export * from './agent-spec';
--- a/packages/Angular/Generic/conversations/src/lib/components/message/message-input.component.ts
+++ b/packages/Angular/Generic/conversations/src/lib/components/message/message-input.component.ts
@@ -410,8 +410,12 @@ export class MessageInputComponent extends BaseAngularComponent implements OnIni
   // Bounded poll after a post-ACK disconnect keeps In-Progress. The GraphQLAIClient
   // stall reconciler only runs while invokeSubAgent is waiting; once it returns,
   // this watch is the bound so a dead socket cannot leave the red-pill timer forever.
+  // Must outlast the server: Skip RequestManager defaults to 30 minutes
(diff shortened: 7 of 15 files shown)


AN-BC (commented):
LGTM

AN-BC:
## Review — `fix(conversations): stop Unknown error + stuck Running timer on agent fail`

Reviewed at `d7feeae`. Read the full diff plus the surrounding call graph on the branch. I did **not** run the build or tests locally (no `node_modules` in this environment) — see the CI note at the bottom.

The diagnosis is right and well-argued: `handleAgentError` in `graphQLAIClient.ts:463` has always produced a real `errorMessage`, and the information was being thrown away downstream when `invokeSubAgent` / `ConversationAgentRunner` collapsed a failure to `null`. Passing the result through is the correct fix, and the complementary status guards between `AgentRunner`'s catch (`Status === 'In-Progress'`) and the resolver's `persistInFlightAgentFailure` (reload-then-check) are a nice touch — they mean the two paths can both fire without double-writing. Changeset is present and correctly all-`patch`.

Three things I'd want resolved before merge, then a set of smaller ones.

---

### 1. The new In-Progress branch bypasses the race guard it most needs — `message-input.component.ts:2727`

`updateConversationDetail` reloads from the DB before writing `Error`, specifically so a client-side disconnect can't stomp a server-side success. Its own comment says so:

> *"if the server completed successfully but a client-side timeout or WebSocket disconnect triggered this error path, we must not overwrite the server's successful completion with an error status"* — `message-input.component.ts:1976`

That reload is gated on `status === 'Error'`. The new branch writes `'In-Progress'`, so it skips the reload, and the early-return guard below it (`:1990`) tests the **in-memory** `convoDetail.Status`, which on a dropped socket is stale (still `In-Progress` from creation). Net effect: the one code path that reaches this branch is *exactly* the disconnect scenario the guard was written for, and it can now overwrite a server-completed detail — real answer and all — back to `In-Progress` with the `⏳ ... is still running on the server` placeholder.

Extending the reload to cover `In-Progress` (or reloading inside `applyAgentFailureToDetail` before deciding) closes this.

### 2. In-Progress is not a terminal state, and nothing here bounds it

`markMessageComplete()` only runs for `Complete`/`Error` (`:1967`). It's what removes the task from `activeTasks` (the red pill) and unregisters the streaming callback. The In-Progress branch returns before any of that.

Keeping the streaming callback registered is the *right* call when the server really is still working — a later completion event lands and closes the bubble. But when the socket dropped *because* the server died or restarted, nothing ever arrives, and the PR trades "wrong terminal state + stuck timer" for "no terminal state, ever, and no error shown at all." That's arguably a worse failure mode than the one being fixed, since the user now has no signal that anything went wrong.

`graphQLAIClient` already has a stall reconciler that polls the `AIAgentRun` by conversation detail and resolves once `Status` leaves `Running` (`:690-723`, bounded by `maxStallReconciles`). Worth saying explicitly in the PR whether that reconciler is expected to cover this branch — if it does, the design is sound and the comment should point at it; if it doesn't, this branch needs its own bound before it ships.

### 3. Same branch silently drops the user-message completion

Every prior failure path ended with `updateConversationDetail(userMessage, userMessage.Message, 'Complete')` unconditionally. The In-Progress branch returns before it (`:2737`), so the user's own message is left non-`Complete`. Deliberate or not, it should be explicit — the user turn finished regardless of what the agent is doing.

---

### Medium

**4. The two copies of this logic disagree.** `conversation-agent.service.ts:434` sets `failed.success = false;`. `ConversationAgentRunner.ts:236` does not. So if an envelope arrives with `Success: false` but a `Result` whose `success` is `true`, the runtime returns a result the caller reads as **successful** on a run that failed. Either add the assignment or, better, factor the "coerce envelope → failed `ExecuteAgentResult`" step into one helper next to `agentFailureMessage` — this is precisely the duplicated-decision shape that drifts.

**5. Two failure paths were left on the old code, including the one in the bug report.**
- `processMessageThroughAgent:1755` — the primary Sage turn — still does `result?.agentRun?.ErrorMessage || 'Agent evaluation failed'` and always paints `Error`, never consulting the disconnect heuristic. It consumes the return value this PR changed.
- `handleSubAgentInvocation:2200` (retry-failed) picked up `agentFailureMessage` but not `applyAgentFailureToDetail`, so it also always paints `Error`. More importantly, that branch sets `conversationManagerMessage` to `Error` and `userMessage` to `Complete` but **never gives `agentResponseMessage` a terminal status** — it's left on `"Retrying..."` / `In-Progress`. That's the headline symptom (bubble + live timer) surviving on the sub-agent path. Pre-existing, but this is the PR that should fix it.

**6. The disconnect heuristic can't tell "socket dropped mid-run" from "request never left the browser."** `isDisconnectWhileAgentMayStillBeRunning` matches `failed to fetch` / `networkerror`, which is also what you get when the *initial* mutation never reaches the server — no run exists, nothing will ever complete, and the bubble holds In-Progress forever. If the transport can distinguish "request acknowledged" from "request failed to send," gating on that would be much sturdier than substring-matching user-facing copy.

**7. `persistInFlightAgentFailure` (`RunAIAgentResolver.ts:565`) — two smaller issues.**
- The `if (!user) return;` guard sits above the `run.Save()` block, which doesn't need `user` at all. A payload without a resolvable user skips marking the `AIAgentRun` Failed as collateral. Move the lookup down to where the detail is loaded.
- It mutates and saves `run` without `EnsureSaveComplete()`, while the `AgentRunner` twin does call it before mutating (`AgentRunner.ts:597`). Both entities are concurrently saved by the framework — the progress callback saves `agentResponseDetail` on every update (`AgentRunner.ts:383-385`). The asymmetry looks accidental.

**8. `agentRunRef.current` is populated only from a progress event carrying `metadata.agentRun` (`:261`) or post-success (`:463`).** A throw before the first such event leaves it `null`, so the run is never marked Failed — the exact early-failure case where you'd most want it. Best-effort is fine, but worth a comment saying so rather than reading as a guarantee.

**9. `runtime-widget-transport.ts:84` didn't get the benefit.** It still reads `result?.agentRun?.ErrorMessage ?? 'The agent could not respond.'`. It now receives a truthy failed result with the real text sitting in `errorMessage` and still renders the generic string — the same "Unknown error" class of bug, one consumer over. One-line swap to `agentFailureMessage(result)`.

---

### Minor

- **Mutating the caller's object.** `failed.success = false` / `failed.errorMessage = ...` write into `runResult.Result` in place. Prefer a copy.
- **`extractAgentResult` fallback stamps an error on success.** `graphQLAIClient.ts:743` sets `errorMessage: 'Agent execution ended without a result payload'` regardless of `success`, so a successful completion event with no payload comes back as `{ success: true, errorMessage: ... }`. Gate it on `!success`.
- **`agentRun` is a required field on `ExecuteAgentResult`** (`agent-types.ts:689`), so `{ success: false, errorMessage } as ExecuteAgentResult` is a type assertion over an object that doesn't have it. TypeScript permits the assertion and says nothing. I traced every reachable dereference and they're all behind a `success` check or optional-chained, so nothing crashes today — but the pattern (pre-existing in `handleAgentError`, now in five more places) is one unguarded `result.agentRun.X` away from a runtime TypeError. Making `agentRun?:` optional would turn that into a compile error, which is where MJ's typing rules want it.
- **Bubble copy is inconsistent across layers.** `AgentRunner.ts:599` and `persistInFlightAgentFailure` write the raw `errorMessage` straight into `Message`, while the client writes `❌ **name** failed\n\n...`. Same failure renders two different ways depending on which layer caught it.

---

### CI

Build is **red**, but not from this diff — it fails in `@memberjunction/messaging-adapters` (`slack-block-builder.ts:382/384/386`, `teams-card-builder.ts:643/645/647`, all `TS2345 string | undefined`), files this PR doesn't touch. Looks like base-branch breakage.

The consequence for this PR is worth flagging though: the unit-test lane was **skipped** because the build lane failed (`Unit test lanes did not pass: build`), so `agent-failure-message.test.ts` and the updated `ConversationAgentRunner.test.ts` **have never run in CI**. The test plan's two checked boxes rest on a local run only. Once `next` is green, please rebase so those actually execute — and note that neither new test covers the branch carrying the risk in items 1–3 (`applyAgentFailureToDetail` choosing In-Progress and what it does to the detail record). A test around that decision would be the highest-value one to add.

---

Root cause and direction are right; my substantive concern is that the In-Progress branch reintroduces a documented race and has no bound, and that the two paths most implicated in the original report weren't converted.

---

AN-BC:
Addressed the review at `a811224b76`. Mapping:

1. **In-Progress bypassed the race guard.** `updateConversationDetail` now reloads before writing `Error` *or* `In-Progress`. If the server already has `Complete`/`Error`, we emit that record and stop the timer instead of stomping it with the still-running placeholder.

2. **In-Progress had no bound.** The GraphQLAIClient stall reconciler only runs while `invokeSubAgent` is waiting, so it does not cover this branch after the client returns. `startInFlightDetailWatch` now polls the detail (6 × 15s) and paints `Error` (reload-first) if it is still `In-Progress`. Streaming callbacks stay registered until then so a late completion can land.

3. **User-message completion.** `applyAgentFailureToDetail` always completes the user message, including the In-Progress path.

4. **Duplicate coerce logic.** `coerceFailedExecuteAgentResult` next to `agentFailureMessage` copies (does not mutate) and always sets `success: false`. Both `conversation-agent.service` and `ConversationAgentRunner` use it.

5. **Leftover failure paths.** Sage `processMessageThroughAgent` and the sub-agent retry/catch paths now go through `applyAgentFailureToDetail`, so the agent bubble (where the red-pill timer lives) gets a terminal status. A post-ACK disconnect on the first attempt does not start a second run on the same detail.

6. **Disconnect vs never-sent.** `FireAndForgetHelper` ACK is threaded as `requestAcknowledged`. Raw `Failed to fetch` / `NetworkError` no longer match the still-running heuristic; that copy is only used when the mutation was ACKed. Unacked fetch failures paint `Error` immediately.

7. **`persistInFlightAgentFailure`.** Run save no longer sits behind `if (!user)`. Both run and detail call `EnsureSaveComplete()` before mutating.

8. **`agentRunRef`.** Comment added: populated from progress `metadata.agentRun` or post-success; a throw before the first such event leaves it null (best-effort).

9. **`runtime-widget-transport`.** Uses `agentFailureMessage(result, 'The agent could not respond.')`.

Also: `extractAgentResult` no longer stamps an errorMessage on a successful completion with no payload. Tests cover `agentFailureDisposition` (the In-Progress vs Error decision) and the envelope-success coerce case.

@AN-BC

AN-BC (commented):
LGTM

AN-BC:
## Re-review at `a811224`

Went through the incremental diff (`d7feeae..a811224`) plus the surrounding call graph. All twelve items from the last round are addressed, and several are addressed better than I asked for — extracting `agentFailureDisposition` and `coerceFailedExecuteAgentResult` turns "two copies that must agree" into one tested decision, and threading `requestAcknowledged` from `validateAck` is a much sturdier signal than the substring matching I complained about.

One new blocker, one design question, two small things.

---

### 🔴 Blocker — the PR no longer compiles

Build failed on `@memberjunction/conversations-runtime` (not the base-branch `messaging-adapters` error from last time; that didn't surface in this run):

```
src/agent-runner/ConversationAgentRunner.ts(247,13): error TS2322:
  Type '{ success?: boolean; errorMessage?: string | null; agentRun?: { ErrorMessage?: string | null } | null }
        & { success: false; errorMessage: string }'
  is not assignable to type 'ExecuteAgentResult | null'.
    Types of property 'agentRun' are incompatible.
      Type '{ ErrorMessage?: string | null } | null | undefined'
      is not assignable to type 'MJAIAgentRunEntityExtended'.
```

This is the `agentRun`-is-required point from last round, except it's no longer latent — `coerceFailedExecuteAgentResult`'s generic is inferred from its *constraint* when the first argument is `undefined`, so `T` widens to the constraint shape and the intersection stops being an `ExecuteAgentResult`. The `as ExecuteAgentResult` casts that used to paper over this are gone, which is why it surfaced now.

Two call sites hit it — both the `undefined` ones:
- `ConversationAgentRunner.ts:247`
- `conversation-agent.service.ts:442` (didn't get built; it's downstream, but it's the same shape and will fail identically)

The other two call sites pass `runResult.Result as ExecuteAgentResult | null | undefined`, so `T` is pinned and they're fine.

**Narrow fix — do this one:** pass the type argument explicitly at both sites.

```ts
return coerceFailedExecuteAgentResult<ExecuteAgentResult>(undefined, errorMsg);
```

**And a correction to what I told you last round.** I suggested making `agentRun?:` optional on `ExecuteAgentResult` and framed it as cheap. It isn't — there are **83 non-optional `result.agentRun.X` dereferences** across the repo (`Scheduling/engine`, `ai-agent-harness`, task-graph, and more), every one of which becomes a compile error. That's still the right end state, and it's what would have caught this class at authoring time, but it's a separate PR with its own blast radius, not something to fold in here. Take the type argument now.

Unit tests were skipped again as a consequence (`Unit test lanes did not pass: build`), so the new `agentFailureDisposition` / `coerceFailedExecuteAgentResult` cases still haven't executed in CI. They're good tests — worth seeing them go green.

---

### 🟡 The watch gives up before the server does

`IN_FLIGHT_WATCH_INTERVAL_MS = 15_000` × `IN_FLIGHT_WATCH_MAX_ATTEMPTS = 6` = **90 seconds**. This PR's own description says the Skip side aborts at **5 minutes**. So for the exact scenario in the bug report — a long Skip run whose socket drops — the client declares Error at 90s while the server is still legitimately working for another three and a half minutes.

What makes it more than cosmetic: the give-up path writes `'Error'` through `updateConversationDetail`, which calls `markMessageComplete()`, which **unregisters the streaming callback**. So when the server does finish at t=200s and writes `Complete` to the detail, nothing on the client is listening any more — the DB says Complete and the bubble says Error until the user reloads. The copy even says "may still be running. Please refresh," which is honest, but it's a milder replay of the mismatch this PR exists to remove.

The bound should outlast the server's own ceiling rather than undercut it. Whatever number you pick, the invariant worth writing down next to the constants is *the client gives up only after the server already has* — otherwise the two ends disagree by construction. Deriving it from the server-side timeout beats a hand-picked constant that drifts when the Skip-side abort changes.

(Everything else about the watch is right: cleared in `ngOnDestroy` and in `markMessageComplete`, `attempts` incremented synchronously so no 7th fire, and the `Status === 'In-Progress'` guard before arming correctly skips the case where the reload already found a terminal status. The reload-guard extension to `In-Progress` — plus calling `markMessageComplete` on that early return so the timer actually stops — is exactly the fix.)

---

### Minor

**1. `handleSubAgentInvocation` catch block writes the CM message to Error unconditionally** (`message-input.component.ts:2243-2244`). The retry branch 20 lines above correctly gates it:

```ts
const retryDisposition = agentFailureDisposition(retryResult);
if (retryDisposition.status === 'Error') { /* CM → Error */ }
```

The catch doesn't, so if a thrown error's text happens to match the disconnect copy, the agent bubble goes In-Progress (with a watch armed) while the Sage delegation bubble right above it says Error. Narrow — `invokeSubAgent` swallows its own errors, so this catch only fires on surrounding failures whose text won't usually match — but it's the same decision made two ways in one function, which is what the extraction was for. Reuse the gate.

**2. The `ConversationAgentRunner` test mocks `coerceFailedExecuteAgentResult` with a hand-written reimplementation** (`ConversationAgentRunner.test.ts:81-90`). It reproduces the real helper's precedence rules by hand, so if the real one changes, the test keeps passing against the stale copy. It follows the file's existing `vi.mock('@memberjunction/ai-core-plus', ...)` pattern so I'm not asking you to restructure — but importing the real function (it's pure and dependency-free) would make the test verify the thing it names.

---

### Confirmed fixed

| # | Item | Status |
|---|---|---|
| 1 | In-Progress bypassed the reload race guard | ✅ guard covers `In-Progress`, and now also treats a server `Error` as terminal |
| 2 | In-Progress unbounded | ✅ bounded watch added — see the sizing question above |
| 3 | user message never completed | ✅ |
| 4 | two copies of the coercion disagreed | ✅ `coerceFailedExecuteAgentResult`, non-mutating, tested for non-mutation |
| 5 | `processMessageThroughAgent` + `handleSubAgentInvocation` unconverted | ✅ both converted; the retry short-circuit on an In-Progress disposition (don't start a second run on the same `conversationDetailId`) is a good catch that wasn't in my list |
| 6 | heuristic couldn't tell never-sent from mid-run | ✅ `requestAcknowledged` threaded through `validateAck`; raw `Failed to fetch`/`NetworkError` no longer match. Verified the flag survives the `AgentClientSession` envelope via `Result: result` |
| 7 | `!user` guard ordering + missing `EnsureSaveComplete` | ✅ both |
| 8 | `agentRunRef` best-effort | ✅ documented |
| 9 | RealtimeWidget consumer | ✅ + dependency declared |
| — | mutation of caller's object | ✅ spread copy |
| — | `extractAgentResult` stamping an error on success | ✅ gated on `!success` |
| — | `agentRun` required | ⚠️ see blocker — now load-bearing |

Also checked and found *not* to be a problem: the `conversationManagerMessage` left non-terminal on the In-Progress early return. `AgentRunner.updateDetail` writes `Complete`/`Error` to whichever detail the run executed under (`AgentRunner.ts:453-473`), so the server terminates that record and the streaming callback picks it up. Same on the success path, so nothing new here.

Fix the compile error and settle the watch duration and this is good to go.

---

AN-BC:
Addressed the re-review at `28924a3c54`.

- **Compile:** both `undefined` call sites now pass `coerceFailedExecuteAgentResult<ExecuteAgentResult>(undefined, errorMsg)` so `T` is not inferred from the constraint. Left `agentRun` required — separate PR as you said.
- **Watch:** 15s × 120 = **30 minutes**, matching Skip `REQUEST_MAX_AGE_MS`. Comment next to the constants states the invariant: the client must not paint Error (and unregister the streaming callback) while the server can still complete.
- **Catch path:** Sage CM bubble is gated on `agentFailureDisposition`, same as the retry branch.
- **Test:** mock now imports the real `coerceFailedExecuteAgentResult` from CorePlus source instead of a hand-written copy.

@AN-BC

AN-BC:
## Re: `28924a3c54` — three of four are right, and the watch fix needs to go the other way

Verified at `28924a3c54`:

- **Compile** ✅ — `coerceFailedExecuteAgentResult<ExecuteAgentResult>(undefined, errorMsg)` at both sites (`ConversationAgentRunner.ts:247`, `conversation-agent.service.ts:442`). Correct fix.
- **Catch path** ✅ — the CM bubble is gated on `agentFailureDisposition` now, matching the retry branch.
- **Test mock** ✅ — imports the real helper. Small note: it reaches across package boundaries with a deep relative path (`'../../../AI/CorePlus/src/agent-failure-message'`), which will break if either package moves. Fine for now, but it's a workspace-layout dependency in a test, not an import.
- **Watch** ❌ — this one went the wrong direction, and my previous comment is why. I'll own that below.

---

### The watch constant is now hardcoded to another product

I asked for the bound to "outlast the server's ceiling," and you did exactly that. But look at what that produced in a shared MJ Angular component:

```ts
// Must outlast the server: Skip RequestManager defaults to 30 minutes
// (REQUEST_MAX_AGE_MS). Giving up earlier unregisters the streaming callback
// so a later server Complete never lands and the bubble stays Error.
private static readonly IN_FLIGHT_WATCH_INTERVAL_MS = 15_000;
// 30 min / 15 s. Must stay ≥ Skip RequestManager REQUEST_MAX_AGE_MS (30 min).
private static readonly IN_FLIGHT_WATCH_MAX_ATTEMPTS = 120;
```

`message-input.component.ts` is on the path of **every agent in every MJ deployment**. It now states, in its own comments, that its correctness constraint is a constant in Skip-Brain — a different product, in a different repo, that MJ cannot see, cannot test against, and does not version together. When `REQUEST_MAX_AGE_MS` changes, nothing here fails; MJ just becomes quietly wrong.

My "size it above the server's ceiling" framing was bad advice. It accepted the premise that the constant should exist and only argued about its value — and the value has now moved **90s → 5 min → 30 min across two rounds**, each time tracking whichever Skip number was most recently mentioned. That drift is the tell: there's no principled basis for any of the three, because the number doesn't belong to MJ.

### MJ already models this, and this constant contradicts it

**1. Per-agent limits are already metadata, already enforced.** `MJ: AI Agents` carries `MaxTimePerRun` — *"Maximum time in seconds allowed for a single agent run. Run will be terminated if this limit is exceeded"* — alongside `MaxCostPerRun`, `MaxTokensPerRun`, `MaxIterationsPerRun`. It's enforced server-side at `base-agent.ts:4822`, which terminates the run and writes a terminal status. Per-agent metadata is the right home for a limit. A fixed 30 minutes in a shared component is a second, competing answer that no operator can configure and that ignores the one they *did* configure.

**2. The transport already solved this, in the right shape.** `fireAndForgetHelper.ts:16` sets `DEFAULT_IDLE_TIMEOUT_MS = 12 * 60 * 1000`, and `graphQLAIClient` doesn't override it. Its own comment explains the design:

> *"The server emits a liveness pulse every ~5 minutes while the operation runs, so as long as the work is alive the timer never fires. It only expires when the server has gone genuinely silent — at which point the optional reconciliation hook decides whether the operation is dead or still running."*

That's an **idle** timeout that deliberately never fires against a healthy run, and on expiry it *asks the database* rather than guessing. Both of its bounds (`timeoutMs`, `maxStallReconciles`) are overridable via config.

The new watch is a **wall-clock** deadline: it fires on schedule whether or not the run is alive and progressing — the exact behaviour the neighbouring design was written to avoid. It isn't overridable. And at 30 minutes it now outlives MJ's own transport timeout by 2.5×, so the two ends of the same operation disagree about how long to wait, in both directions depending on which round you read.

**3. The polling cost is real.** 120 attempts × 15s means a DB round trip every 15 seconds for half an hour, per stranded bubble, multiplied by however many are stranded.

### What to do instead

In order of preference:

1. **Drop the client-side deadline.** Keep observing the run's status with backoff until the server writes something terminal — which it will, because `MaxTimePerRun` guarantees it. Bound the *polling cost* (back off: 5s → 15s → 60s), not the verdict.
2. **If a give-up is genuinely wanted,** derive it from data MJ owns: `agent.MaxTimePerRun` (seconds, already on the entity — and `handleSubAgentInvocation` already has the agent in hand via `AIEngineBase.Instance.Agents.find(...)`), plus slack, falling back to the transport's idle timeout when it's null. Expose it as an `@Input()` so a host can tune it, matching `timeoutMs` / `maxStallReconciles`.
3. **Either way, remove the Skip references from these comments.** A constant in MJ core has to be justifiable from MJ alone. If the comment can only explain itself by naming another repo's identifier, the number is in the wrong place.
4. **Don't paint `Error` when the watch gives up.** That write calls `markMessageComplete()`, which unregisters the streaming callback — so a wrong verdict becomes sticky until reload. Stop watching and leave the record non-terminal instead.

The principle underneath all four: `ConversationDetail.Status` is a claim about what the *server's* run is doing, and the client isn't the authority for that. Its job is to render the run's status, not to decide it on a timer and then stop listening.

### Two small cleanups while you're in here

- `agent-failure-message.test.ts` asserts on `"Pipeline failed at step 'Execute Sub-Agent: Skip: Data Expert'"` — a framework package's tests naming a customer agent's step. A generic fixture proves the same thing.
- `processMessageThroughAgent:1764` added another `this.converationManagerAgent?.Name || 'Sage'`. There were already five; not yours to fix here, but worth not growing.

Compile fix and the two minors look good — it's just the constant. Build for `28924a3c` was still running when I wrote this; I'll check it separately.

---

AN-BC:
Addressed the watch design at `bdfdbced5347`.

Took option 1: **no client deadline**. After a post-ACK disconnect the watch reloads `ConversationDetail.Status` with backoff **5s → 15s → 60s** and stops only when the server writes Complete/Error (`MaxTimePerRun` is the ceiling). It never paints Error, so it never unregisters the streaming callback on a timer.

Comments justify this from MJ alone (server-owned status, idle backoff for polling cost). Skip `REQUEST_MAX_AGE_MS` is gone.

Also: generic pipeline-failure fixture in `agent-failure-message.test.ts`; Sage failure path passes `'Sage'` instead of growing another `converationManagerAgent?.Name || 'Sage'`.

@AN-BC

AN-BC:
## `bdfdbced5` — watch design is right now; one regression to undo

The observe-with-backoff rewrite is exactly it:

- No client verdict. The watch never writes `Error`; it observes until the **server** writes a terminal status.
- Skip references gone from the constants, and the comment now states the principle rather than another repo's number: *"ConversationDetail.Status is the server's claim; the client only renders it."*
- Backoff `5s → 15s → 60s` bounds polling cost instead of bounding the answer.
- `setInterval` → recursive `setTimeout` also removes the overlap hazard the interval had (a `Load()` slower than the tick could stack polls).
- The `inFlightWatches.has(detail.ID)` check at both poll entry and before rescheduling means a stopped watch can't be resurrected by an in-flight timer. Nice.
- A failed `Load()` reschedules rather than killing the watch — right call for a transient read.
- Test fixtures degenericized to `'Execute Sub-Agent: Specialist'`.

### One regression — please revert this hunk

```diff
-          this.converationManagerAgent?.Name || 'Sage',
+          'Sage',
```

My note was "there are already five `|| 'Sage'` fallbacks; don't grow it." Replacing the dynamic lookup with a bare literal goes the wrong way — this is now the **only** call in the file that can't reflect a deployment whose conversation-manager agent is named something else, and it's on a user-visible failure bubble ("❌ **Sage** failed"). `converationManagerAgent` is populated from configuration precisely so that name isn't assumed.

Restore the dynamic form, or better, reuse the resolver that already exists in this component — `resolveRealtimeAgentName()` (`:572`) does exactly this lookup with `'Sage'` as its last-resort fallback. That would take the count from five to four instead of six.

### CI

Meaningful change: **this PR's own code compiles now.** `conversations-runtime` and `ng-conversations` are out of the failure list — 263/269 tasks succeeded, and the single failure is back to `@memberjunction/messaging-adapters` (`slack-block-builder.ts:382/384/386`, `teams-card-builder.ts:643/645/647`, `TS2345 string | undefined`), which this PR doesn't touch and which is still present on `next` (`origin/next` hasn't moved off `3d4801ba` since this PR opened).

That base breakage is now the only thing between this PR and a green build — and it has kept the unit-test lane **skipped for three runs straight**, so `agent-failure-message.test.ts` and the `ConversationAgentRunner` cases still haven't executed in CI even once. Worth either fixing `next` or rebasing onto a fixed `next` before merge, so the tests this PR added actually run somewhere other than a laptop.

Everything else here looks good to me.

---

AN-BC:
Restored the dynamic conversation-manager name at `ac6708bfb4`.

`processMessageThroughAgent` failure path again uses `this.converationManagerAgent?.Name || 'Sage'` so a renamed conversation-manager agent shows on the user-visible bubble.

Did **not** switch this call to `resolveRealtimeAgentName()` — that resolver prefers `findLastNonSageAgentId()` (continuity / pinned default) and would label a Sage evaluation failure with the last specialist. This bubble is the conversation-manager message.

@AN-BC

AN-BC:
## `ac6708bfb` — regression fixed. Your CI blocker is now cleared on `next` — please merge it in.

Verified the Sage hunk is restored:

```diff
-          'Sage',
+          this.converationManagerAgent?.Name || 'Sage',
```

That was the only outstanding code item from my side. **Everything I raised across the three rounds is now addressed**, and the watch design in `bdfdbced5` is the right one — observe until the server writes a terminal status, back off to bound polling cost, never invent a verdict, no foreign constants in MJ core.

### The thing to act on

**`next` was fixed 20 minutes after your last build ran, and the PR hasn't rebuilt since.**

- Your last CI run: `23:23:03 – 23:25:34 UTC` on `ac6708bfb`. Sole failure `@memberjunction/messaging-adapters` (the same six `TS2345 string | undefined` in `slack-block-builder.ts` / `teams-card-builder.ts`). **265 of 270 tasks green** — every package this PR touches builds clean.
- `next` moved to `5f33ca80` at `23:45:35 UTC` — [#4076 "fix(messaging-adapters): production bring-up fixes for Slack and Teams"](https://github.com/MemberJunction/MJ/pull/4076), which rewrote exactly those lines.

This PR is still based on `3d4801ba`, one commit behind, so it has never seen that fix. Merging `origin/next` into the branch should take the build green:

```bash
git fetch origin next && git merge origin/next
```

### Why this matters beyond a green checkmark

The build has failed on **all five** CI runs for this PR, and because the test lane is gated on it (`Unit test lanes did not pass: build`), `agent-failure-message.test.ts` and the updated `ConversationAgentRunner` cases **have never executed in CI — not once**. Everything we know about them comes from a local run.

Those are the tests carrying the load here: `agentFailureDisposition` is the In-Progress-vs-Error decision, and `coerceFailedExecuteAgentResult` is the non-mutation guarantee that keeps the two runtimes from drifting. Both have been rewritten since they were last run locally. So the merge isn't bookkeeping — it's the first actual verification this PR will get.

One thing to watch on the rebuild: **Integration (SQL Server, deterministic) has also been red on every run.** I haven't diagnosed it and it may well be downstream of the same build failure, but confirm it clears rather than assuming it does.

Once the build is green and the tests pass, this looks good to merge from my side.

---

AN-BC:
Merged `origin/next` (`5f33ca80`, #4076 messaging-adapters) at `65d9257279be`.

This PR was still on `3d4801ba`; the Slack/Teams `TS2345` fix landed on `next` after our last build. Rebuild should be the first CI run that actually executes `agent-failure-message.test.ts` and the `ConversationAgentRunner` coerce/disposition cases.

Will confirm Build + Integration (SQL Server, deterministic) both go green.

@AN-BC

AN-BC:
CI is green on `65d9257279` (merge of `origin/next` / #4076).

- **Build** success
- **Unit tests** all 6 shards + Native-ESM + `Run unit tests` success — first time this PR's `agent-failure-message` / `ConversationAgentRunner` cases have executed in CI
- **Integration (SQL Server, deterministic)** success

Mergeable / CLEAN. Nightly and live-model lanes skipped as usual.

@AN-BC

AN-BC:
Merged `origin/next` (`3e2efa6b`, #3352) at `aca7cbc80e31` so this PR is current again.

Also re-checked open review threads: **none unresolved**.

On Sage / customer-agent names in tests:
- `agent-failure-message.test.ts` no longer names Skip or Data Expert — the pipeline-failure fixture is generic (`'Execute Sub-Agent: Specialist'`).
- The remaining `|| 'Sage'` in `message-input.component.ts` is the conversation-manager display name you asked us to **restore** (`converationManagerAgent?.Name || 'Sage'`), not a test fixture.

@AN-BC

AN-BC:
## `aca7cbc80` — green, and the merge didn't eat anything. LGTM.

Full CI on the current head:

| Check | |
|---|---|
| Build | ✅ (12m49s — a real full build, not an early abort) |
| Unit tests, shards 1–6 | ✅ all six |
| Run unit tests / Native-ESM guard | ✅ |
| **Integration (SQL Server, deterministic)** | ✅ — it was red on all five prior runs; it was downstream of the build, as suspected |
| Source guards, adopted standards, migrations, deps, npm, Sass, peer-dep | ✅ |

`mergeable_state: clean`.

### The merge is the part I actually checked

`origin/next` brought 95 changed lines into `AgentRunner.ts` — a file this PR also edits — so a clean auto-merge there is exactly the case worth not trusting on sight. Verified at the merged head:

- `let agentResponseDetail` still hoisted outside the `try` (`:232`)
- the catch block intact at `:637-645`, with `EnsureSaveComplete()` and the `Status === 'In-Progress'` guard

Both of this PR's AgentRunner changes survived. Nothing silently dropped.

### You were right about `resolveRealtimeAgentName()`

Checked your pushback rather than taking it on faith — `resolveCurrentAgentId()` is `findLastNonSageAgentId() ?? conversationDefaultAgentId ?? defaultAgentId ?? converationManagerAgent?.ID`, so that resolver leads with the last non-Sage agent. Using it on the conversation-manager bubble would label a Sage evaluation failure with whatever specialist ran last. `converationManagerAgent?.Name || 'Sage'` is the correct call there. My suggestion was wrong; thanks for not taking it.

### One caveat on my own verification

I confirmed all six shards pass and that both `ai-core-plus` and `conversations-runtime` are in the affected build graph — they have to be, since the earlier failures were *inside* `conversations-runtime`. I did **not** grep the individual test-file names out of the shard logs: vitest prints a `Test Files N passed (N)` summary on success rather than listing files, so their absence from the log proves nothing either way. If you want the stronger claim ("`agent-failure-message.test.ts` executed and passed"), the run log for the shard that owns CorePlus is where to confirm it — but nothing suggests otherwise.

---

Everything raised across four rounds is addressed, the design questions are settled the right way (no client-side verdict, no foreign constants in MJ core, limits stay per-agent metadata), and CI is green for the first time in seven commits. **Approving from my side** — the three unchecked manual boxes in the test plan (WebSocket drop mid-run, genuine pipeline failure, `executeAIAgent` throw) are yours to exercise before merge, since no automated tier covers them.

---

AN-BC:
@jordanfanapour this complements the other PR I shared with you, please put eyes on it, merging it now.

PR #4100, opened 2026-08-29
security: harden CORS default, OAuth error open redirect, and consent scope grant

AN-BC:
## Summary

An adversarial security review of the framework's auth/authorization surfaces found them **well-hardened overall** — JWT verification pins the asymmetric alg family (no `alg=none`/HS confusion), audience/issuer/expiry enforced per request, API keys are SHA-256 + `timingSafeEqual`, the MCP OAuth proxy uses single-use codes/states + enforced PKCE (S256) + exact redirect-URI matching, and media capability tokens are `fileId`-bound (no IDOR). No committed secrets. These three items are the genuine action items (all MEDIUM/LOW, defense-in-depth).

### 1. CORS default: wildcard + credentials (MEDIUM) — `packages/MJServer/src/index.ts`
With no `cors` config (the default, `allowedOrigins=['*']`), the origin callback returns `true` for every origin — so the `cors` middleware **reflects the caller's Origin** and sets `Access-Control-Allow-Credentials: true`. Any site a logged-in user visits could then make credentialed cross-origin requests and read the responses. MJ's primary auth is a **Bearer token** (not auto-attached cross-origin), which limits exposure, but reflect-any-origin-with-credentials is a recognized misconfiguration.
**Fix:** drop `credentials` to `false` when the wildcard is in effect; an explicit `allowedOrigins` list still honors `allowCredentials`. The reflect-origin-on-401 behavior is preserved.

### 2. OAuth `/oauth/authorize` open redirect (MEDIUM) — `packages/AI/MCPServer/src/auth/OAuthProxyRouter.ts`
For an **unknown `client_id`** or an **unregistered `redirect_uri`**, the handler still 302-redirected the browser to the attacker-supplied `redirect_uri` (carrying `error`/`state`). That's a classic open redirect usable for phishing on the deployment's trusted domain.
**Fix:** render the local error page (`sendErrorPage`) for those two **pre-validation** cases. Per the OAuth 2.0 Security BCP, errors may only be sent to a `redirect_uri` after it is validated; post-validation errors still use `sendAuthorizationError`.

### 3. OAuth consent scope grant not intersected with offered scopes (LOW) — same file
The consent POST signed `req.body.scopes` straight into the authorization code / JWT with **no check** that each granted scope was actually offered on the consent screen. Harmless today (`availableScopes` defaults to all active scopes) but a latent gap that becomes an escalation the moment `availableScopes` is narrowed per-user/per-client.
**Fix:** intersect granted scopes with `consentRequest.availableScopes` (by `FullPath`) before minting the code; log any dropped scopes.

## Documented for follow-up
- **No GraphQL query depth/complexity limit** (`apolloServer/index.ts`) — an authenticated user can send a deeply nested query to exhaust DB/CPU. Adding a depth-limit/cost-analysis plugin needs a new validation-rule dependency, so it's left for a dedicated change.
- `RunView` `ExtraFilter`/`OrderBy` being raw SQL is an accepted architectural contract with compensating controls (per-entity read perms, RLS WHERE, least-privileged DB principal) — flagged as "keep the DB principal least-privileged," not a code change.

## Verification
Dependencies are not installed in this review environment, so the changes were verified by review rather than a local build. Each uses APIs already present in the touched files (`sendErrorPage`, `EntityByName`, the existing `cors` options).

---
_Generated by [Claude Code](https://claude.ai/code/session_01GQzKwckVq6NgCwDeanSK9k)_


AN-BC (commented):
This seems right to me but I'd like @cadam11  to give it a quick review and merge if it looks good.

Issue #127, opened 2024-04-11
Setup pm2 for the MJCodeGenAPI sub-process and test


AN-BC:
Per #126

AN-BC:
Actually this commit is what is most interesting within that PR @cadam11 

https://github.com/MemberJunction/MJ/pull/126/commits/b4b9d1351cda0d6750df9d14c1270689bd276f3d

cadam11:
@AN-BC This is deployed and running in the CDP stage environment

AN-BC:
@cadam11 should this be closed?

PR #4101, opened 2026-08-29
feat(actions): ERP accounting verbs — dispatcher + BC CreateJournalEntry

AN-BC:
## Summary

Wave 1 of 3: **MJ (this)** → [MemberJunction/bizapps-accounting](https://github.com/MemberJunction/bizapps-accounting) (engine + `AccountingEngineExtension`) → [MemberJunction/bizapps-fpna](https://github.com/MemberJunction/bizapps-fpna) (`ImportBankAccountBalances`).

Accounting is a subledger, not the GL. This package is **stateless ERP verbs** — HTTP plus trivial validation. It does not upsert `GLAccount`, flip journal-entry batch status, or write `CashBalance`. Zero knowledge of bizapps-accounting.

Callers and agents use the **verb name**, never the vendor class name. Each verb is a dispatcher on `BaseAction` that:

1. Reads `CompanyID`
2. Loads the company's accounting `CompanyIntegration` (Integration.Name in `QuickBooks Online` / `Microsoft Dynamics 365 Business Central`)
3. `ClassFactory.TryCreateInstance(BaseAction, \`${verb}:${integration.Name}\`)`
4. Forwards the same params to the plugin (`Run` → `InternalRunAction`)

Missing plugin → `PROVIDER_NOT_REGISTERED`.

### Verbs / plugins

| Verb | QBO | BC |
|---|---|---|
| `GetChartOfAccounts` | existing GL codes + shared `Accounts` | existing GL accounts + shared `Accounts` |
| `GetAccountBalances` | existing | **new** (`accountCode` = account number) |
| `CreateJournalEntry` | existing | **new** (journalLines then `Microsoft.NAV.post`) |
| `GetDimensions` | **new** (Class + Department; QBO has no first-class dimensions API) | **new** (`dimensions` + `dimensionValues`) |
| `GetGLEntries` | existing transactions | existing GL entries |
| `GetCustomers` | — | existing wrap |
| `GetSalesInvoices` | — | existing wrap |

Historical RegisterClass keys still resolve (`CreateQuickBooksJournalEntryAction`, `GetBusinessCentralGLAccountsAction`, …). A class may carry two `@RegisterClass` keys (verb plugin + old name).

Shared `JournalEntryLine` accepts `accountNumber` (preferred, AM-4) or `accountId` (QBO still uses `accountId` internally). Unbalanced lines → `VALIDATION_ERROR`.

Also: SQLConverter now converts `ISJSON(col) = 1` in CREATE TABLE CHECK bodies to `(col) IS JSON` after identifier quoting, which unblocks the accounting `AccountingEngineExtension` PG migration.

Plan: [erp-provider-layer.md](https://github.com/MemberJunction/bizapps-accounting/blob/next/plans/erp-provider-layer.md) / accounting PR https://github.com/MemberJunction/bizapps-accounting/pull/126

Exploratory prototypes (leave open): https://github.com/MemberJunction/bizapps-accounting/pull/74, https://github.com/MemberJunction/bizapps-accounting/pull/112

## Test plan

- [x] `cd packages/Actions/BizApps/Accounting && pnpm test` — 61 passing
- [x] `CreateJournalEntry` dispatcher: missing provider → `PROVIDER_NOT_REGISTERED`
- [x] QBO plugin still POSTs `journalentry` when Integration.Name is QuickBooks Online
- [x] Unbalanced QBO/BC journal → `VALIDATION_ERROR`
- [x] BC CreateJournalEntry balanced path: GET journals → POST journalLines → POST `Microsoft.NAV.post` (spy call order)
- [x] BC GetAccountBalances maps account number → `accountCode`
- [x] BC GetDimensions maps dimensions + values
- [x] Old RegisterClass keys still constructable
- [x] SQLConverter CreateTableRule: `ISJSON(Configuration) = 1` → `IS JSON`, not `"ISJSON"`
- [x] Both packages `pnpm run build`
- [ ] Review README mermaid + “add a new ERP” section as a first-time plugin author
- [ ] Wave 2 (bizapps-accounting engine) consumes these verb names, not vendor class names

### DIFF
--- a/packages/Actions/BizApps/Accounting/src/base/base-accounting-action.ts
+++ b/packages/Actions/BizApps/Accounting/src/base/base-accounting-action.ts
@@ -1,9 +1,11 @@
 import { BaseAction } from '@memberjunction/actions';
 import { ActionParam, ActionResultSimple, RunActionParams } from '@memberjunction/actions-base';
-import { RegisterClass, UUIDsEqual } from '@memberjunction/global';
+import { EscapeSQLString, MJGlobal, RegisterClass, UUIDsEqual } from '@memberjunction/global';
 import { UserInfo } from '@memberjunction/core';
 import { MJCompanyIntegrationEntity, MJIntegrationEntity } from '@memberjunction/core-entities';
 import { IMetadataProvider, Metadata, RunView } from '@memberjunction/core';
+import { ACCOUNTING_ERP_INTEGRATION_NAMES, erpPluginKey } from '../constants';
+import { ResolvedAccountingIntegration } from '../types';
 
 /**
  * Base class for all accounting-related actions.
@@ -75,7 +77,7 @@ export abstract class BaseAccountingAction extends BaseAction {
         const rv = new RunView();
         const result = await rv.RunView<MJCompanyIntegrationEntity>({
             EntityName: 'MJ: Company Integrations',
-            ExtraFilter: `CompanyID = '${companyId}' AND Integration.Name = '${this.integrationName}'`,
+            ExtraFilter: `CompanyID = '${EscapeSQLString(companyId)}' AND Integration.Name = '${EscapeSQLString(this.integrationName)}'`,
             ResultType: 'entity_object'
         }, contextUser);
 
@@ -144,7 +146,7 @@ export abstract class BaseAccountingAction extends BaseAction {
         const rv = new RunView();
         const result = await rv.RunView<MJIntegrationEntity>({
             EntityName: 'MJ: Integrations',
-            ExtraFilter: `Name = '${this.integrationName}'`,
+            ExtraFilter: `Name = '${EscapeSQLString(this.integrationName)}'`,
             ResultType: 'entity_object'
         }, contextUser);
 
@@ -200,5 +202,98 @@ export abstract class BaseAccountingAction extends BaseAction {
             message += ` System error: ${systemError.message || systemError}`;
         }
         return message;
+    }
+
+    /**
+     * Load the company's accounting CompanyIntegration (any Integration whose
+     * Name is one of the known ERP names). Does not hard-filter to a single vendor.
+     */
+    protected async resolveCompanyAccountingIntegration(
+        companyId: string,
+        contextUser: UserInfo
+    ): Promise<ResolvedAccountingIntegration> {
+        const nameList = ACCOUNTING_ERP_INTEGRATION_NAMES
+            .map(name => `'${EscapeSQLString(name)}'`)
+            .join(', ');
+
+        const rv = new RunView();
+        const result = await rv.RunView<MJCompanyIntegrationEntity>({
+            EntityName: 'MJ: Company Integrations',
+            ExtraFilter: `CompanyID = '${EscapeSQLString(companyId)}' AND Integration.Name IN (${nameList})`,
+            OrderBy: 'Integration',
+            ResultType: 'entity_object'
+        }, contextUser);
+
+        if (!result.Success) {
+            throw new Error(`Failed to retrieve company integration: ${result.ErrorMessage}`);
+        }
+
+        if (!result.Results || result.Results.length === 0) {
+            throw new Error(`No accounting ERP integration found for company ${companyId}. Configure QuickBooks Online or Microsoft Dynamics 365 Business Central.`);
+        }
+
+        const record = result.Results[0];
+        const name = record.Integration;
+        if (!name) {
+            throw new Error(`Company integration ${record.ID} has no Integration name; cannot dispatch an ERP plugin.`);
+        }
+
+        return {
+            Name: name,
+            CompanyIntegrationID: record.ID,
+            CompanyID: record.CompanyID,
+            IntegrationID: record.IntegrationID,
+        };
+    }
+
+    /**
+     * Resolve the company's ERP and invoke the plugin registered as `${verb}:${integration.Name}`.
+     * Plugin.Run → InternalRunAction with the same params the caller passed to the dispatcher.
+     */
+    protected async dispatchVerb(verb: string, params: RunActionParams): Promise<ActionResultSimple> {
+        const companyId = this.getParamValue(params.Params, 'CompanyID');
+        if (!companyId) {
+            return {
+                Success: false,
+                ResultCode: 'VALIDATION_ERROR',
+                Message: 'CompanyID is required',
+                Params: params.Params
+            };
+        }
+
+        if (!params.ContextUser) {
+            return {
+                Success: false,
+                ResultCode: 'ERROR',
+                Message: 'Context user is required',
+                Params: params.Params
+            };
+        }
+
+        let integration: ResolvedAccountingIntegration;
+        try {
+            integration = await this.resolveCompanyAccountingIntegration(companyId, params.ContextUser);
+        } catch (error) {
+            const errorMessage = error instanceof Error ? error.message : 'Unknown error occurred';
+            return {
+                Success: false,
+                ResultCode: 'NO_ACCOUNTING_INTEGRATION',
+                Message: errorMessage,
+                Params: params.Params
+            };
+        }
+
+        const pluginKey = erpPluginKey(verb, integration.Name);
+        const resolved = MJGlobal.Instance.ClassFactory.TryCreateInstance<BaseAction>(BaseAction, pluginKey);
+        if (!resolved.Resolved || !resolved.Instance) {
+            return {
+                Success: false,
+                ResultCode: 'PROVIDER_NOT_REGISTERED',
+                Message: `No ERP plugin registered for ${pluginKey}`,
+                Params: params.Params
+            };
+        }
+
+        return resolved.Instance.Run(params);
     }
 }
--- a/packages/Actions/BizApps/Accounting/src/base/base-accounting-action.ts
+++ b/packages/Actions/BizApps/Accounting/src/base/base-accounting-action.ts
@@ -7,6 +7,13 @@ import { IMetadataProvider, Metadata, RunView } from '@memberjunction/core';
 import { ACCOUNTING_ERP_INTEGRATION_NAMES, erpPluginKey } from '../constants';
 import { ResolvedAccountingIntegration } from '../types';
 
+class AccountingIntegrationError extends Error {
+    constructor(message: string, readonly resultCode: 'NO_ACCOUNTING_INTEGRATION' | 'AMBIGUOUS_ACCOUNTING_INTEGRATION') {
+        super(message);
+        this.name = 'AccountingIntegrationError';
+    }
+}
+
 /**
  * Base class for all accounting-related actions.
  * Provides common functionality and patterns for interacting with accounting systems.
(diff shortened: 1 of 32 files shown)


AN-BC:
Wave links (all three are up):

| Layer | PR |
|---|---|
| MJ verbs | this (#4101) |
| Accounting engine + UI | https://github.com/MemberJunction/bizapps-accounting/pull/127 |
| FP&A cash import | https://github.com/MemberJunction/bizapps-fpna/pull/6 |

Plan: https://github.com/MemberJunction/bizapps-accounting/pull/126

packages/Actions/BizApps/Accounting/src/providers/business-central/actions/create-journal-entry.action.ts
AN-BC, line 1: **Partial-post leaves orphaned lines in the BC journal, and the next call posts them.**

Lines are POSTed one at a time to `journals({id})/journalLines`. If line 3 of 5 fails, the catch returns `ERROR` — but lines 1 and 2 are already sitting in that BC journal, and nothing removes them. The next `CreateJournalEntry` against the same journal calls `journals({id})/Microsoft.NAV.post`, which posts **the whole journal batch**, not just the lines this invocation created. So the orphans go to the general ledger attached to somebody else's entry, unbalanced.

The accounting layer's "failure never marks Posted" (bizapps-accounting#127) is correct on its own side and does nothing about this — the rows are already in the customer's ERP.

Three ways out, roughly in order of preference:
1. Compensating cleanup in the catch: `DELETE journalLines({lineId})` for everything created in this call before returning the error.
2. Create a dedicated journal per entry and post that, so the blast radius is one entry.
3. At minimum, refuse to proceed when `journals({id})/journalLines` is non-empty on entry, and say so — that turns a silent wrong posting into a loud one.

Two smaller things in the same file:

**`resolveGeneralJournal` can select a journal with a balancing account.** The filter deliberately keeps only `!balancingAccountNumber`, then `|| unbalancing[0] || journals[0]` throws that away. Posting a self-balanced multi-line entry into a journal that has a balancing account makes BC add a line nobody asked for. If no unbalanced journal exists, that should be an error rather than a guess.

**A `JournalCode` that matches nothing returns `undefined`,** which produces `'No general journal found in Business Central. Pass JournalCode or configure a GENERAL/DEFAULT journal…'` — advice to do the thing the caller just did. Worth distinguishing "you asked for GENERAL2 and it does not exist" from "this company has no journals".

**`mapToBCJournalLine` sends both `accountNumber` and `accountId` when both are present.** AM-4 says accountNumber is preferred; the shared `JournalEntryLine` allows both, and QBO populates `accountId`. If a caller carries a stale `accountId` alongside a good `accountNumber`, BC picks one and the entry lands on an account the caller did not name. Prefer `accountNumber` and only fall back to `accountId`.

---

packages/Actions/BizApps/Accounting/src/base/base-accounting-action.ts
AN-BC, line 1: **`resolveCompanyAccountingIntegration` silently picks one of N ERPs.**

```ts
ExtraFilter: `CompanyID = '…' AND Integration.Name IN (${nameList})`,
OrderBy: 'Integration',
…
const record = result.Results[0];
```

A company mid-migration with both QuickBooks Online and Business Central configured gets whichever sorts first — `Microsoft Dynamics 365 Business Central` — every time, with no signal. `CreateJournalEntry` then posts real money to the ERP the caller did not intend, and the dispatcher reports `SUCCESS`.

The whole point of the dispatcher is that the caller names a verb rather than a vendor, so it can't disambiguate for you. Suggest: when more than one row comes back, fail with a typed code (`AMBIGUOUS_ACCOUNTING_INTEGRATION`) naming the candidates, and accept an optional `IntegrationName` param for callers that do know.

Related, same query: there's no active/enabled predicate. `AccountingERPEngine.loadCredentialedIntegrations` in bizapps-accounting#127 filters `IsActive = 1`; this one doesn't, so a disabled integration row is still dispatchable. Worth making the two agree.

Minor: the filter uses `Integration.Name` while the result is read back as the denormalized `record.Integration`. That's pre-existing (the single-vendor path above does the same), but this PR copies it into a new `IN (…)` path — worth confirming once against the real `vwCompanyIntegrations` that the dotted form resolves, since the tests mock RunView and wouldn't catch it.

`EscapeSQLString` on all three filters is the right call and matches the house rule.

---

AN-BC (commented):
Reviewed as wave 1 of 3, with the cross-repo contracts in mind (verbs → bizapps-accounting#127 → bizapps-fpna#6).

The dispatcher shape is right: callers name a verb, the plugin key is `${verb}:${Integration.Name}`, `PROVIDER_NOT_REGISTERED` is a distinct outcome, and keeping this package free of any bizapps-accounting knowledge is what makes wave 2 possible. `Run()` is a thin passthrough to `InternalRunAction`, so dispatching through it costs nothing. Old `@RegisterClass` keys surviving alongside the new verb keys is the right migration story.

**The SQLConverter change is correct.** `convertNamedJsonCall`'s `"?${name}"?\s*\(` does handle the post-quoting `"ISJSON"(…)` form the comment claims, and folding `= 1` / `= 0` into `IS JSON` / `IS NOT JSON` matches PG semantics. One deployment note worth putting somewhere durable: the SQL/JSON `IS JSON` predicate is **PostgreSQL 16+**. bizapps-accounting's PG job targets 17 so this wave is fine, but it does move the PG floor for any converted schema carrying an ISJSON check.

Three findings, left as file comments — the BC journal one is the serious one:

1. **BC `CreateJournalEntry` can leave orphaned lines that a later call posts to the GL.** Per-line POST with no compensating delete, and `Microsoft.NAV.post` posts the entire journal rather than this entry's lines.
2. **`resolveCompanyAccountingIntegration` silently picks the first of N integrations** — a company with both QBO and BC always resolves to BC.
3. **`mapToBCJournalLine` sends `accountNumber` and `accountId` together** when both are set, so a stale id can win over the preferred number.

Two things I checked and am *not* flagging, so nobody re-derives them: `BaseAction.Run` returns `Promise<ActionResultSimple>`, so `return resolved.Instance.Run(params)` from `dispatchVerb` is type-correct; and every verb/plugin class is named-exported from `index.ts`, so the manifest system will keep them from being tree-shaken.

One cross-repo note that isn't actionable here but determines whether the typed codes in this PR are worth anything: **`ActionResult` has no `ResultCode`** (`ActionEngine-Base.ts` — it carries `Result?: MJActionResultCodeEntity` instead). The accounting engine's verb runner does `result.ResultCode ?? (result.Success ? 'SUCCESS' : 'ERROR')`, so `PROVIDER_NOT_REGISTERED`, `NO_ACCOUNTING_INTEGRATION` and `VALIDATION_ERROR` all arrive downstream as plain `ERROR`. Raised on #127; flagging here because it's this PR's contract that gets flattened.

Not merging or pushing anything — comments only.

---

packages/Actions/BizApps/Accounting/src/providers/business-central/actions/create-journal-entry.action.ts
AN-BC: Addressed in 3e4f5ff3bf.

- If a later `journalLines` POST (or `Microsoft.NAV.post`) fails, we `DELETE journalLines({id})` for every line this call created before returning the error, so orphans are not sitting in the batch for the next post.
- `resolveGeneralJournal` no longer falls back to a journal with a balancing account. Missing `JournalCode` is now `'JournalCode X does not exist'`; a company with no unbalanced journal is an error rather than `journals[0]`.
- `mapToBCJournalLine` sends `accountNumber` when present and only then `accountId`.

packages/Actions/BizApps/Accounting/src/base/base-accounting-action.ts
AN-BC: Addressed in 3e4f5ff3bf.

When more than one active accounting ERP is configured and the caller did not pass `IntegrationName`, dispatch returns `AMBIGUOUS_ACCOUNTING_INTEGRATION` naming the candidates. Optional `IntegrationName` selects one. The query now requires `IsActive = 1`, matching accounting's `loadCredentialedIntegrations`.

AN-BC:
Review round addressed in 3e4f5ff3bf (MJ tests 67 passing).

- BC `CreateJournalEntry`: compensating DELETE of lines created in this call; no balancing-account fallback; distinct missing-JournalCode error; accountNumber over accountId.
- Dispatch: `AMBIGUOUS_ACCOUNTING_INTEGRATION` + optional `IntegrationName`; `IsActive = 1`.

Ready for re-review.

AN-BC (commented):
## Re-review at `3e4f5ff` — all three findings verified fixed

Read the changed code rather than the commit note. All three hold up.

**1. Partial-post orphans — fixed.** `createdLineIds` accumulates each line's `id` as it's POSTed, and the whole sequence (line POSTs *and* `Microsoft.NAV.post`) is wrapped so any throw runs `deleteCreatedJournalLines` before rethrowing. Cleanup failures are `LogError`'d individually rather than swallowed or allowed to mask the original error — right call, since losing the original exception to a cleanup failure is the classic version of this bug.

**2. `resolveGeneralJournal` — fixed, and better than I asked for.** A named `JournalCode` that doesn't exist is now its own error instead of a silent fallback; a named journal that *has* a balancing account is refused with an explanation rather than quietly adding a line; and the no-code path filters to `unbalancing` first, preferring GENERAL/DEFAULT, with a real error when the set is empty. `journals[0]` is gone from both paths.

**3. `mapToBCJournalLine` — fixed.** `accountNumber` when present, `else if` `accountId`. The `// AM-4:` comment explains why, which is what stops someone "helpfully" restoring the both-fields version later.

**Bonus, and the right shape:** `AccountingIntegrationError` carries its own `resultCode`, and `dispatchVerb` narrows on `instanceof` to surface it — so `AMBIGUOUS_ACCOUNTING_INTEGRATION` actually reaches the caller instead of being flattened to a generic message. `IsActive = 1` is now on the query, matching accounting's `loadCredentialedIntegrations`, and `IntegrationName` is a declared input param rather than an undocumented extra.

### One residual, not a blocker, worth recording

`Microsoft.NAV.post` still posts the **whole BC journal**, not just this call's lines. The compensating delete closes the failure-path orphan, which was the sharp edge — but two `CreateJournalEntry` calls running concurrently against the same journal code will still post each other's in-flight lines to the GL, and the loser gets no error. That's inherent to BC's journal model rather than anything this PR introduced, and `PostJournalBatch` making these calls per batch is exactly where concurrency shows up.

Not asking for it here. Worth a line in the README's `CreateJournalEntry` section (a per-post journal, or serializing posts per journal code) so it isn't rediscovered from a GL that doesn't reconcile.

### Cross-repo

The `ActionResult` flattening I raised from here is fixed in bizapps-accounting#127 at `d473f2c` — `defaultAccountingVerbRunner` now static-imports `ActionEngineServer` and reads `result.Result?.ResultCode`. I verified `MJActionResultCodeEntity` does expose `ResultCode` (`entity_subclasses.ts:35629`), so the typed codes this PR defines survive the trip. Details on that PR.

CI is mid-flight on this head — Source guards, adopted standards and Check migrations are green; Build, the deterministic integration tier, dependency check and PG validation are still running. I'll pick up the result on my next pass.

No further findings from me. Still not merging or pushing anything.


---

AN-BC:
Re-review residual recorded in f417bb11a5.

README now states that `Microsoft.NAV.post` posts the **whole** BC journal, so concurrent `CreateJournalEntry` calls against the same journal code can cross-post in-flight lines. Mitigation: per-post `JournalCode`, or serialize posts per journal. Not changing the plugin in this PR — that's BC's model.

AN-BC (commented):
LGTM

PR #4104, opened 2026-08-30
feat(predictive-studio): typed ML component model — tree, resolver, rubric, widened as-of vocabulary

MS-BC:
Draft while the DB-gated half lands (see **What's not here**). Everything present is built and tested.

## Why

Predictive Studio's catalog is flat and trainable-only: 6 algorithms, preprocessing ops as loose strings, no way to say what a component *means*, no way to compose one model out of others. This lands the substrate for **"everything is a component"** — models decompose into typed, pluggable components (model primitives, preprocessing, statistical methods, inputs/outputs, parameters, slot-bearing structures) organized in an inheritance tree, where a component need not be trainable to be first-class and every component can bind to real business data.

## The tree, and why it's rows not JSON

`MJ: ML Component Types` is a `ParentID` tree with seven Kind roots. Each node carries only what is true of **every** descendant; a leaf's profile is resolved by walking up and merging. The inheritable properties are **rows** (`MJ: ML Component Type Properties`), not a JSON blob, precisely so the "principled partition" is mechanically enforceable: a descendant `Remove` vetoing an ancestor's `Add` is a *detectable contradiction* — the property wasn't true of all descendants and should move down. `lintComponentTree` reports it, and **the shipped seed tree is certified by a test that loads the real metadata files and asserts zero findings** (`Core/src/__tests__/seed-tree.test.ts`).

Merge semantics are fixed **per key in code**, never per row, so two nodes can't disagree about how a key merges: union-with-veto (banks, gates), append/replace (hyperparameter banks, guidance), override-nearest (defaults), subset-narrowing (`CompatibleProblemTypes` — a child may only restrict), shallow object merge (validation defaults). Provenance is recorded per key for the "inherited from" UI chips.

## Sonar donations (its PS contribution roadmap, items 1/2/4/5/6/7-lite)

| Item | What landed |
|---|---|
| **2 — glass-box rubric** | `RubricEstimator` sidecar driver. Given-mode = **the canonical untrainable-but-first-class component**: weights are hand-set, `fit` validates and freezes them, and it still flows through PS's honest-holdout machinery so its trustworthiness is *measured*. Search-mode fits non-negative weights. Per-input `Zero`/`NeutralMidpoint`/`Exclude` missing-data policies. Exact per-record contributions **free** via `coef_` (the existing P1-5 linear path needs zero rubric-specific code). |
| **1 — as-of aggregates** | `DatedFeatureSpec.Aggregate` widened 2 → 10 kinds over 5 window kinds (`Rolling` exclusive-lower / `Calendar` / `SinceEvent` / `RenewalRelative` / `AllTime`, clamped month math: 31 Jul − 1mo = 30 Jun), SQL NULL rules, plus `EmitPresence` → `<col>__present` **hadData masks** (item 7-lite: a real zero and absence become different cells). **Ported as TS reducer semantics over PS's row model, not Sonar's OPENJSON SQL.** Legacy spelling kept as byte-identical aliases. |
| **5 — normalizations** | Six ops with direction + output range: `minmax`, `percentile` (midpoint tie handling), `zscore` (±3σ clamp), and the stateless curves `logistic`/`banded`/`lookup`. Population-relative ones follow fit-once/apply-everywhere. |
| **4 — join path** | `findAutoPathHops` BFS over reverse-FK edges, composite FKs bundled into one hop, failing loud on unreachable (>5 hops) or ambiguous (≥2 shortest) — a guessed join is a silently-wrong feature. |
| **6 — bands** | `assignBand` (half-open, top inclusive, order-independent), `detectBandTransition`, `trendDirection` deadband. |

Also fixes a real gap found on the way: `TrainingEngine.assemble()` never passed `datedSources`, so as-of features could not round-trip train → Lineage → score. `MLTrainingPipeline.DatedSources` is the persisted home (plumbing in the follow-up).

## Proof

- **Core** `pnpm test` → 83 passed (23 resolver/lint + 8 seed-tree certification)
- **Engine** `pnpm test` → 35 files passed (new: as-of aggregates + windows + presence, join-path, score-bands)
- **Sidecar** `pytest` → 48 passed (14 rubric incl. end-to-end /train + /predict, 9 normalization golden, 25 pre-existing)

## What's not here (DB-gated)

The migration needs `mj migrate → mj sync push → mj codegen` on a **fresh** DB before any TypeScript may reference the generated entities (no `.Get()/.Set()`). Landing next: `MLComponentEngine` (BaseEngine cache + `ResolveProfile`/`Lint`), the component materializer, `DatedSources` plumbing, and the entity-server subclasses that populate `StoryVector`.

**PostgreSQL counterpart is deliberately absent** — per `migrations/CLAUDE.md`, PG conversion is build-engineer work done for a whole release in one pass, not per feature PR.

## Scope notes

`Sequence`/HMM and the Structure wrappers are seeded `Status='Draft'` — they need the sidecar's `component_graph` execution and the `sequence` problem-type widening (one PR, 6 touch points), deliberately not smuggled in here. `plans/predictive-studio-fpna-link.md` records the FP&A ↔ PS ↔ Sonar seam (design only; nothing touches bizapps-fpna before its Nov 15 go-live).


MS-BC:
## Database gate: verified end to end, and it surfaced two CodeGen findings

Ran the migration + seed against a **clean SQL Server 2022** database (fresh container, empty DB, 67 migrations from the v6 baseline). Result: the full sequence works, and the second commit adds the one step the original design missed.

### Verified (clean DB, all steps exit 0)

```
1. mj migrate                        → 67 applied
2. mj codegen                        → 5 entities + 95 fields registered, views + CRUD procs built
3. mj sync push --include=entities   → Hierarchy.IsHierarchy on the 3 recursive FKs
4. mj codegen                        → base views gain the hierarchy OUTER APPLY
5. mj codegen                        → registers <FK>Depth / Path / IsLeaf / ChildCount
6. mj sync push  (the seed tree)     → created 101, updated 6, errors 0
```

Final state: **66 types across exactly 7 Kind roots, 29 properties, 5 slots, 7 algorithms bridged**, and all 66 rows readable through `vwMLComponentTypes` with `RootParentID` resolved. The tree renders correctly through `ParentIDPath`/`ParentIDDepth`. The CodeGen-derived `PropertyKey` union matches Core's hand-written `ComponentPropertyKey` exactly — the lockstep test target holds.

### Finding 1 — hierarchy is opt-in, and silently half-configured without it (fixed here)

A self-referencing FK does **not** get hierarchy support automatically. CodeGen registers a virtual `Root<FK>ID` field and generates the `GetRootID` function, but the base view never produces the column unless `EntityField.Configuration` carries `{"Hierarchy":{"IsHierarchy":true}}`. The failure mode is quiet and nasty: every read of the entity fails with `column ... does not exist`, which a grid renders as *"no data"* rather than an error, and `spCreate`/`spUpdate` return a column set the client cannot bind (`Column name or number of supplied values does not match table definition`). It does **not** self-heal — I dropped the views and re-ran CodeGen to confirm, because `__mj` is in `excludeSchemas`.

Fixed by declaring all three recursive FKs in `metadata/entities/.entity-field-hierarchy-configurations.json`, next to the 34 core entities already opted in. Worth knowing for anyone adding a tree table to `__mj`.

### Finding 2 — CodeGen output is not directly appendable for a multi-entity migration (not fixed here)

Appending the generated SQL, as `migrations/CLAUDE.md` describes, fails on this migration two different ways:

1. **Missing `GO` separators** — four consecutive `DECLARE @constraintName` blocks (the drop-default-constraint blocks for `__mj_CreatedAt`/`__mj_UpdatedAt` on the new entities) land in one batch: *"The variable name '@constraintName' has already been declared."* Inserting 3 `GO`s clears it.
2. **Duplicate `EntityField` sequence** — for an entity that receives both a real column and a virtual display field in one run (here `MJ: ML Models`, gaining `RootComponentID` + `RootComponent`), CodeGen emits two `SET Sequence = Sequence + 100000 WHERE Sequence < 100000` blocks. The second shift catches the field the first block just inserted at a low sequence and moves it onto the slot an already-shifted field occupies → `Violation of UNIQUE KEY constraint 'UQ_EntityField_EntityID_Sequence' ... (a3997636…, 100020)`.

Rather than hand-edit generated SQL, this migration ships **hand-written DDL only**, with the CodeGen handoff documented in the file — the same division of labor the repo already uses for PostgreSQL counterparts and the consolidated metadata-sync migration. Happy to file (2) separately if it's not already known.

### Housekeeping

The worktree needed its own `pnpm install`: sharing the main repo's `node_modules` resolved the CLI to a stale **6.0.0** from another branch, and that CodeGen regenerated 218 files against older templates — *reverting* improvements present on `next` (e.g. `NewRecordValues` losing its second argument, `ShowToolbar` flipping to `false`). All of that was discarded; with the correct **6.1.0-edge.4** toolchain the run is clean. Only my own files are committed — no generated churn, no `sync` blocks (metadata rule 1), no lockfile noise.

PR #3704, opened 2026-08-10
feat(ai): price models billed by duration or by image, not only by token

MS-BC:
## Why

MJ's pricing **schema** was always general — a cost row names a price unit type, the unit type names a `DriverClass`, the ClassFactory resolves it. The **execution layer** was not. `BasePriceUnitType` took two token counts, only the three token drivers were ever registered, and `MJAIPromptRunEntityServer` refused to cost any run that reported zero tokens.

So the three continuous-media unit types that already shipped — `Per Image`, `Per Minute`, `Per Hour` — resolved to nothing, and every run priced by one was **silently uncosted**. Six ACTIVE image cost rows were in exactly that state. This is bug register **B60**.

Speech-to-text made it concrete. Groq bills Whisper by the audio-hour; the just-merged `GroqAudioGenerator` (#3494) requested `response_format: 'json'`, which discards the duration entirely; and both Whisper models shipped with no cost rows because there was no honest way to write one. That PR's own changeset says so: *"the per-minute price unit is not addressed here, which is also why no cost rows are seeded."*

This PR addresses it, and adds OpenAI as a second Whisper provider.

### Usage grows a second axis

`ModelUsage` gains `unitKind` (`'Tokens' | 'Seconds' | 'Characters' | 'Images'`), `inputUnits`, `outputUnits`, and a `ModelUsage.ForMedia(kind, input, output?)` constructor. `SpeechResult` gains `usage?`, matching `ImageGenerationResult`.

Continuous quantities are deliberately **not** folded into the token fields. A run reporting 90 "tokens" that actually means 90 minutes corrupts `TokensUsed`, the `TokensUsed = TokensPrompt + TokensCompletion` invariant, every rollup above it, and every dashboard downstream.

Quantities are recorded in the **base** measure, never the billing measure — audio billed per hour is still recorded in seconds, and the driver converts. That is what lets one measured duration be priced against a per-minute row from one vendor and a per-hour row from another.

### Pricing takes quantities

`BasePriceUnitType` gains:
- a `UnitKind` getter, defaulting to `'Tokens'` so external subclasses need no change;
- `CalculateCost(activeCost, usage: NormalizedUsage)` — the preferred entry point, whose default delegates to the existing cache-aware path, so every current driver behaves identically.

[...]

PR #4106, opened 2026-08-31
chore(bootstrap): regenerate stale class-registration manifests (unblocks next)

MS-BC:
## What is red

`next` has failed the **class-registration manifest freshness gate** on every run since **2026-08-29 21:01**:

> Committed class-registration manifests are STALE — a `@RegisterClass` class was added/removed without regenerating them. Fix: run `pnpm run mj:manifest` locally and commit the result.

Worth stating clearly, because the log is misleading: the `› Error: command codegen:manifest not found` line that appears just above it is **not** the failure. That comes from `ServerBootstrapLite`'s postbuild, is swallowed by its own `|| echo 'Warning: …'` fallback, and every turbo task reports success. The build fails afterwards, at the gate.

## Cause

#4101 added eleven `@RegisterClass` accounting-action classes (`CreateJournalEntryAction`, `CreateBusinessCentralJournalEntryAction`, the BC/QuickBooks verb actions) and did not regenerate the manifests. The `actions-bizapps-accounting` entry goes 8 → 19 classes.

Timing pins it: `next` was green at `a0c4a226` (08-29 15:17) and red from `526051a9` (08-29 21:01), which is that merge commit.

## Why it matters beyond CI

A manifest exists to create a static code path the bundler cannot tree-shake. A stale one means those classes can be **dropped from a bundled app**, and `ClassFactory` then cannot resolve them at runtime — the ERP accounting verbs fail to register with no build error to explain it. The gate is doing exactly its job.

## What is in this PR

Only the seven generated manifests. Regenerated through the path CI compares against — a full `pnpm run build`, whose root postbuild runs `mj:manifest`.

Every class count **grows**; none shrinks:

| manifest | before | after |
|---|---|---|
| ServerBootstrap | 1003 | **1023** |
| ServerBootstrapLite | 919 | **939** |
| ng-bootstrap | 647 | **652** |
| ng-bootstrap-lite | 533 | **538** |

The three zero-class manifests (A2AServer, MCPServer, MJCodeGenAPI — all of which pass `--exclude-packages @memberjunction`) pick up the generator's current chunked emit shape, which they predate. No behavioural content.

## A trap for whoever does this next

Running `pnpm run mj:manifest` **without building first** produces a plausible-looking but destructive result: ~2,100 insertions, ~440 deletions, `lazy-feature-config.ts` losing 388 lines, and class counts collapsing — because the scanner walks `dist/`, and an unbuilt tree has almost nothing to find. It exits 0. I hit this on the first attempt and threw it away; committing it would have caused precisely the silent tree-shaking failure the gate protects against.

**Build first, always** — and sanity-check that counts go up, not down.

### DIFF
--- a/.changeset/regenerate-stale-class-manifests.md
+++ b/.changeset/regenerate-stale-class-manifests.md
@@ -1,0 +1,8 @@
+---
+'@memberjunction/ng-bootstrap': patch
+'@memberjunction/ng-bootstrap-lite': patch
+'@memberjunction/server-bootstrap': patch
+'@memberjunction/server-bootstrap-lite': patch
+---
+
+Regenerate the class-registration manifests. Eleven `@RegisterClass` accounting-action classes were added without regenerating them, so the committed manifests no longer listed every registered class — and a manifest exists precisely to create a static code path the bundler cannot tree-shake. In a bundled app the unlisted classes could be dropped, leaving `ClassFactory` unable to resolve them at runtime: the ERP accounting verbs would fail to register, with no build error to say why.
--- a/packages/AI/A2AServer/src/generated/class-registrations-manifest.ts
+++ b/packages/AI/A2AServer/src/generated/class-registrations-manifest.ts
@@ -14,9 +14,17 @@
 /**
  * Runtime references to every @RegisterClass decorated class.
  * This array creates a static code path the bundler cannot tree-shake.
+ *
+ * Split into fixed-size chunks so no single array literal grows a union large
+ * enough to trip TS2590; the exported array is their concatenation.
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
+const CLASS_REGISTRATIONS_0: any[] = [
+];
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const CLASS_REGISTRATIONS: any[] = [
+    ...CLASS_REGISTRATIONS_0,
 ];
 
 /** Marker constant indicating the manifest has been loaded. */
--- a/packages/AI/MCPServer/src/generated/class-registrations-manifest.ts
+++ b/packages/AI/MCPServer/src/generated/class-registrations-manifest.ts
@@ -14,9 +14,17 @@
 /**
  * Runtime references to every @RegisterClass decorated class.
  * This array creates a static code path the bundler cannot tree-shake.
+ *
+ * Split into fixed-size chunks so no single array literal grows a union large
+ * enough to trip TS2590; the exported array is their concatenation.
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
+const CLASS_REGISTRATIONS_0: any[] = [
+];
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const CLASS_REGISTRATIONS: any[] = [
+    ...CLASS_REGISTRATIONS_0,
 ];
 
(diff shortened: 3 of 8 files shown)


rkihm-BC:
Closing as superseded by #4105, which merged to `next` as `8b78695cac`.

The generated output of the two PRs is **byte-for-byte identical** — `git diff origin/next <this branch>` over `src/generated/` is empty across all seven manifests. Same counts: ServerBootstrap 1023, ServerBootstrapLite 939, ng-bootstrap 652, ng-bootstrap-lite 538, and the chunked format-only emit for A2AServer / MCPServer / MJCodeGenAPI. The only remaining delta was the changeset, and #4105's is a strict superset (it also bumps `a2aserver` and `ai-mcp-server`, both of which did have their files rewritten).

Preserving the diagnosis from this PR, because #4105 landed with a different explanation of the same red gate and the details below aren't recorded anywhere on `next`.

## The failing log line is a red herring

The `› Error: command codegen:manifest not found` that appears just above the failure is **not** the failure. It comes from `ServerBootstrapLite`'s postbuild, is swallowed by that script's own `|| echo 'Warning: …'` fallback, and every turbo task still reports success. The build fails *afterwards*, at the manifest freshness gate:

> Committed class-registration manifests are STALE — a `@RegisterClass` class was added/removed without regenerating them. Fix: run `pnpm run mj:manifest` locally and commit the result.

Anyone debugging a red `Build` job should scroll past the `command not found` line rather than chasing it.

## Two causes, one regeneration

#4105 attributed the staleness to the incomplete chunked-format rollout (#3694 / #3704 regenerated only `server-bootstrap` and `server-bootstrap-lite`). This PR attributed it to #4101 adding eleven `@RegisterClass` accounting-action classes without regenerating — `actions-bizapps-accounting` goes 8 → 19.

Both are real, which is why the two PRs converged on identical output: one regeneration from a fully-built workspace picks up the format change *and* the drifted-out registrations together. The merged diff shows both — the accounting/Apollo action classes alongside the `CLASS_REGISTRATIONS_0/1/2/3` restructuring.

Timing pins the second cause: `next` was green at `a0c4a226` (2026-08-29 15:17) and red from `526051a9` (2026-08-29 21:01), which is the #4101 merge commit.

## Why it matters beyond CI

A manifest exists to create a static code path the bundler cannot tree-shake. A stale one means those classes can be dropped from a bundled app, and `ClassFactory` then cannot resolve them at runtime — the ERP accounting verbs fail to register with no build error to explain it. The gate was doing its job.

## Trap for whoever regenerates these next

Running `pnpm run mj:manifest` **without building first** produces a plausible-looking but destructive result: ~2,100 insertions, ~440 deletions, `lazy-feature-config.ts` losing 388 lines, and class counts collapsing — because the scanner walks `dist/`, and an unbuilt tree has almost nothing to find. **It exits 0.**

The mechanism (also noted on #4105): `resolveSubpathExportsDetailed()` resolves a package's lazy-loading subpaths by reading the `.d.ts` each `exports` entry names, and `continue`s past any that is missing, while `hasSubpathExports()` reads only `package.json` and still reports the package as lazy-loadable. The two disagree, the package falls through to the whole-package branch, and the per-dashboard lazy chunks collapse into a single eager import — silently.

**Build first, always, and sanity-check that class counts go up, not down.** Making the generator fail loudly on a missing `dist/` would kill this whole class of drift; worth a follow-up issue.

PR #4113, opened 2026-08-31
feat(ai-betty-bot): optional userId caller identifier on Betty requests

bc-izygmunt:
## What

AACSB's Betty utilization reports flagged Izzy's server IP as a bot: MJ's Betty provider sends every request anonymously, so all traffic lands in one per-IP rate bucket. Betty's API already accepts a caller identifier via the `userId` request-body field (feeds their `spLinkUserID`/`spMessageCount`, giving the caller its own attribution row and rate-limit bucket) — but `BettyBotLLM.nonStreamingChatCompletion` posts only `{ input }`, with no way to send one.

## How

`BettyBotLLM` gains an optional second constructor arg, `userId?: string`. When set, the `/response` body becomes `{ input, userId }`; when omitted, the body is **byte-identical** to today, so existing consumers (`betty.action.ts` in CoreActions is the only in-repo call site) are unaffected. The settings/JWT call never carries the identifier. Consumers opt in per instance, e.g. `new BettyBotLLM(apiKey, 'izzy')`.

Deliberately *not* used: Betty's documented `clientId` field — its only consumer on Betty's side is commented out, so it goes nowhere today.

## Tests

5 new vitest cases in the package suite: UserId stored / left undefined; response body is exactly `{ input }` without an identifier and exactly `{ input, userId }` with one; the settings (JWT) call never includes it. `packages/AI/Providers/BettyBot`: **22/22 passing**, builds clean. Deterministic integration tier run against a from-scratch clean-room DB: all runnable server-transport members pass (client-transport members require MJAPI, exercised by this PR's Integration Tier gate).

## Notes

- **Backport (`backport lts/5`)**: Izzy runs on the 5.x LTS line, so the fix must reach `lts/5` to ship before the 6.x era. Additive, no migration, no metadata — patch-only per line rules. The single non-merge commit carries the whole payload (§5.2 `merge_commits: skip`). `lts/5`'s copy of `BettyBotLLM.ts` has the same constructor/body-build shape, so the cherry-pick is expected to land clean.
- Downstream sequencing: once released, Izzy adopts via a routine version bump plus the one-line call-site change; the Betty team then labels/allowlists the agreed literal (`izzy`) in their reports and bot-detection rule. Their per-org "Require User Auth" feature must not be enabled for AACSB until Izzy's change is deployed (it would hard-reject identifier-less callers, including other MJ Betty consumers).

### DIFF
--- a/.changeset/betty-bot-user-identifier.md
+++ b/.changeset/betty-bot-user-identifier.md
@@ -1,0 +1,5 @@
+---
+"@memberjunction/ai-betty-bot": patch
+---
+
+`BettyBotLLM` accepts an optional `userId` constructor argument, forwarded to Betty's `/response` endpoint as the `userId` body field so a caller can identify itself for Betty's utilization attribution and get its own rate-limit bucket instead of an anonymous per-IP one (the behavior that tripped Betty's bot detection on Izzy's server IP). Omitting the argument leaves the request body byte-identical to before, so existing consumers are unaffected; the settings/JWT call never carries the identifier.
--- a/packages/AI/Providers/BettyBot/src/__tests__/BettyBotLLM.test.ts
+++ b/packages/AI/Providers/BettyBot/src/__tests__/BettyBotLLM.test.ts
@@ -104,6 +104,15 @@ describe('BettyBotLLM', () => {
     it('should initialize TokenExpiration as a Date', () => {
       expect((llm as unknown as Record<string, unknown>)['TokenExpiration']).toBeInstanceOf(Date);
     });
+
+    it('should leave UserId undefined when not provided', () => {
+      expect((llm as unknown as Record<string, unknown>)['UserId']).toBeUndefined();
+    });
+
+    it('should store the userId when provided', () => {
+      const withUser = new BettyBotLLM('test-api-key', 'izzy');
+      expect((withUser as unknown as Record<string, unknown>)['UserId']).toBe('izzy');
+    });
   });
 
   /* ---- SupportsStreaming ---- */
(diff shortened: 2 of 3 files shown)


cadam11:
Holding the `backport lts/5` label on this one — not a no, a not-yet.

The code is fine. Optional arg, byte-identical body when it's omitted, tests covering both shapes. If risk were the only question this would be an easy yes.

The question on a line is what the patch delivers to everyone who takes it. This doesn't change `betty.action.ts`, so a 5.51.x carrying it ships ~300 packages whose payload is an unused constructor parameter. Nothing lands until Izzy makes their own call-site change and deploys. Meanwhile every 5.x customer re-qualifies against a build the full cert gate never saw, covered only by the mini-cert.

That's a worse trade than the three fixes already queued for 5.51.2, which fix things that are actively broken for anyone running them. Same bar I applied on #4086, written up as 5.3 in #4098.

What makes it a *when* rather than a *never* is that there's no other route: Izzy is on 5.x and the provider hardcodes the body, so they can't send the identifier without this. So merge it to `next`, and re-apply the label when Izzy's change is ready to deploy — at that point it's unblocking a customer on a date, which clears the bar on its own and gets its own patch.

Two other things worth having on the record:

It can't ride 5.51.2 regardless. That's being cut from the current `lts/5` tip and fixes only flow `next` → line, so the earliest it could reach a line is 5.51.3.

The sequencing hazard in your notes is the part I'd track hardest, and shipping this early doesn't discharge it: if the Betty team enables "Require User Auth" for AACSB before Izzy deploys, identifier-less callers get hard-rejected — including other MJ Betty consumers. That needs an explicit agreement on ordering with their team, not just our release.

AN-BC:
> Holding the `backport lts/5` label on this one — not a no, a not-yet.
> 
> The code is fine. Optional arg, byte-identical body when it's omitted, tests covering both shapes. If risk were the only question this would be an easy yes.
> 
> The question on a line is what the patch delivers to everyone who takes it. This doesn't change `betty.action.ts`, so a 5.51.x carrying it ships ~300 packages whose payload is an unused constructor parameter. Nothing lands until Izzy makes their own call-site change and deploys. Meanwhile every 5.x customer re-qualifies against a build the full cert gate never saw, covered only by the mini-cert.
> 
> That's a worse trade than the three fixes already queued for 5.51.2, which fix things that are actively broken for anyone running them. Same bar I applied on #4086, written up as 5.3 in #4098.
> 
> What makes it a _when_ rather than a _never_ is that there's no other route: Izzy is on 5.x and the provider hardcodes the body, so they can't send the identifier without this. So merge it to `next`, and re-apply the label when Izzy's change is ready to deploy — at that point it's unblocking a customer on a date, which clears the bar on its own and gets its own patch.
> 
> Two other things worth having on the record:
> 
> It can't ride 5.51.2 regardless. That's being cut from the current `lts/5` tip and fixes only flow `next` → line, so the earliest it could reach a line is 5.51.3.
> 
> The sequencing hazard in your notes is the part I'd track hardest, and shipping this early doesn't discharge it: if the Betty team enables "Require User Auth" for AACSB before Izzy deploys, identifier-less callers get hard-rejected — including other MJ Betty consumers. That needs an explicit agreement on ordering with their team, not just our release.

only comment I have is that I would doubt very many people actually invoke this class anywhere so I don't have any objection on that basis, but @cadam11 you seem concerned about this getting into the 5.51.3 release?

cadam11:
> 
> only comment I have is that I would doubt very many people actually invoke this class anywhere so I don't have any objection on that basis, but @cadam11 you seem concerned about this getting into the 5.51.3 release?

@AN-BC my main concern is establishing some discipline around patching LTS releases so we can deliver a truly stable library to users. I have no issues with this making it into an (eventual) 5.51.3 release.

PR #4121, opened 2026-08-31
a11y: shell landmarks + skip link, focus containment, and nameable primitives

MattC-BC:
Fixes eight WCAG 2.1 A/AA findings in the Explorer shell and shared primitives, found during a downstream WCAG audit against MJ 5.51.0 and re-verified against this tree before implementing.

## Why this matters

These were found by a consuming product auditing itself for WCAG 2.1 AA — a CE-credit learning platform whose customers carry ADA / Section 504 obligations and ask for conformance documentation during procurement. Six **Level A** criteria were held open downstream purely because the defects live in MJ, not in the consuming app: a shell with no `main` landmark and no skip link, primitives that cannot be named from outside, and controls that hold keyboard focus while visually closed. None of them are fixable downstream — an app cannot name someone else's component or add a landmark to someone else's shell. Shipping this closes them for every MJ consumer at once.

## Re-verify table

The audit measured 5.51.0; this line has moved since. Every item was re-checked against `52acac3516` (6.1.0-edge.4) **before any code was written** — the middle column is what I found on arrival, not what ships. Everything below is fixed in this PR except items 3 and 4.

The re-verify earned its keep twice: **item 6** was already half-fixed on edge (so only the missing half was built rather than re-implementing dialog semantics), and **item 3** was fixed outright by #3860 while this branch was in flight — that one collided, and the merge resolves it by taking `next`'s implementation wholesale, since it is a strict superset. `packages/Angular/Generic/ui-components` is now identical to `next`; this PR no longer touches it.

| # | 5.51.0 finding | Re-verified state *before* this PR | Fix in this PR |
|---|---|---|---|
| 1 | Shell global search input named by placeholder only | Confirmed present | `aria-label="Search"`, placeholder kept |
| 2 | Avatar button has an empty accessible name on the icon-fallback path | Confirmed present | Name moved onto the **button** (`avatarAriaLabel`); img/icon `aria-hidden` |
| 3 | `mj-dropdown` exposes no naming input | **Already fixed on this line** — #3860 and its follow-ups shipped `AriaLabel` / `AriaLabelledBy` plus `InputId` / `AriaDescribedBy` | Nothing — superseded; this PR's copy removed in the merge. The shell's search dropdown becomes the new API's first consumer |
| 4 | Whiteboard keyboard operability | Confirmed present | **Deferred — not fixed here, see below** |
| 5 | Closed mobile nav drawer keeps focus (6 of 8 controls) | Confirmed present | `inert` + transitioned `visibility`, focus returns to toggle |
| 6 | Command palette lacks dialog semantics | **Already half fixed on this line** — `role="dialog"`, `aria-modal`, `aria-label`, focus-on-open and Esc-to-close are all present | Added the missing half: Tab trap + focus return to invoker |
| 7 | Eleven document-level single-letter shortcuts | Confirmed present | Scoped to focus-within-host |
| 8 | `--mj-focus-ring` invites `outline:` misuse | Present; **zero misuses in MJ today** | Added `--mj-focus-ring-color` + a CI gate (clean-baseline regression guard) |
| 9 | No `main` landmark | Confirmed present | `role="main"` on the routed content region |
| 10 | No skip link | Confirmed present | First focusable element in the shell |

Two extras found in the same family while fixing item 5, both in the same file: the **closed search popup** has the identical focus defect three lines from the drawer, and the drawer's **close button** was icon-only with no name. Other icon buttons in the shell carry `title=`, which does produce an accessible name, so they were left alone rather than widening scope.

## Consumer-visible changes — please rule on these consciously

**1. Whiteboard single-letter shortcuts are now scoped to focus.** The eleven tool shortcuts (`v h p r s t m w i c e`) listened on `document` and fired anywhere on the page — a speech-input user saying an ordinary word, or anyone typing in a non-input control elsewhere, silently switched the board tool. That is a WCAG 2.1.4 failure. They now fire only while focus is inside the whiteboard host, which is made click-focusable for the purpose. **The letters no longer work while focus is elsewhere on the page.** `EnableGlobalShortcuts` (default `false`) restores the old behavior for any surface that accepts the exposure.

**2. Consuming apps should delete their own skip links on upgrade.** Apps added their own precisely because the shell had none. The shell's must be first in DOM order, and two stacked skip links is a worse experience than one. Called out in the changeset so downstream sees it.

Also worth a reviewer's eye, both judgment calls rather than defects:

- `role="main"` is applied to `mj-tab-container` rather than wrapping it in `<main>`. That container is a Golden Layout host inside a flex chain, and a wrapper element is a layout risk that cannot be verified without a browser. Same landmark semantics, no layout change.
- The whiteboard host got `tabindex="-1"`, not `0`. The canvas has no keyboard interaction model yet, so a real Tab stop would be a stop a keyboard user can reach and do nothing with.

## Not in this PR: item 4, whiteboard keyboard operability

Acknowledged and deliberately deferred — it is a design-and-build of its own, not a patch. It needs a keyboard placement / movement / selection model for a canvas. Direction sketch for whoever picks it up: focusable canvas, Tab and arrow keys to traverse items, Enter to place the selected tool's item at a caret point, Delete to remove your own items. The tool-switch letters already exist, which ironically makes *choosing a tool* the one keyboard-reachable part of the whiteboard today.

## Verification

| Package | Build | Unit tests |
|---|---|---|
| `ng-ui-components` | clean | 565 pass (untouched by this PR — identical to `next`) |
| `ng-shared-generic` | clean | 27 pass |
| `ng-whiteboard` | clean | 314 pass |
| `ng-explorer-core` | clean | 167 pass (command-palette 3 → 10) |

Gates: `check:ui` (including the new `check:focus-ring`), `check:standards`, `check:changeset`, and the hex-token gate on every changed stylesheet all pass. The new gate is falsification-tested — it catches `outline: 2px solid var(--mj-focus-ring)`, `outline-color:`, and bare `outline:`, while ignoring correct `box-shadow` use, the `-color` companion, and the wrong form quoted inside a comment.

All four packages build clean with `strictTemplates` active — verified live by injecting a deliberate bad binding into the shell template and confirming `ngc` catches it, so template checking is demonstrably running rather than being skipped.

CI is fully green, including the deterministic integration tier and all six unit-test shards.

**Outstanding:**

- **The live keyboard walkthrough is still to be done.** Everything above is automated verification; the focus behavior in items 5, 6, 9 and 10 ultimately wants a human tabbing through the shell, and the whiteboard behavior change wants a reviewer's conscious ruling.


MattC-BC:
@rkihm-BC - This PR is related to the a11y findings for LXP. Would love to get this in before the build tomorrow. Thanks!

rkihm-BC (commented):
Re-reviewed at `86a8a9a9a5`. The build blocker is gone and resolved the right way, the self-audit is genuinely good, and there is **one small thing to fix before merge — a four-line changeset edit.** Not requesting changes for it; the code is sound and CI is green, so this is a comment rather than a review cycle.

## The dropdown collision — resolved correctly

Removed from the PR entirely rather than merged, with the changeset stating outright that `mj-dropdown`'s accessible name landed separately in #3863. That is the right call: `next`'s implementation is strictly the more complete one — it names the trigger, the popup listbox (`role="listbox"` carries `aria-label`/`aria-labelledby`), and composes the filter box's name through `FilterLabelledBy`. Nothing was lost by dropping the duplicate.

Worth saying plainly, because it reads worse than it was: the re-verify table was **honest when it was written**. I checked the stated baseline — `52acac3516` genuinely has zero `AriaLabel` on that component, and `545d725108` (the fix) is not an ancestor of it. The row went stale between re-verifying and merging forward, which is a different failure from not checking.

## The self-audit is the strongest part of this round

Four findings actioned before any review landed, and three of them are ones I would have raised:

- **Removing `aria-haspopup="menu"` from the avatar button.** The user menu has no `role="menu"` and no `menuitem` children, so the attribute promised semantics that do not exist — announced as "has menu popup", then delivers a generic group. That defect was introduced by this PR's own first draft and caught before it reached anyone.
- **Strengthening the palette focus-trap test.** "With one focusable element first === last, so the two wrap directions were indistinguishable and the test would have passed against a half-broken trap." A spec that cannot tell two states apart is the exact failure mode worth hunting, and finding it in your own test is better than having it found for you.
- **Disclosing the whiteboard scope accurately.** The guard gates the *entire* keydown handler — undo/redo, `Escape`, `Delete`/`Backspace` — not only the eleven letters 2.1.4 is about. That is materially broader than the original description, the reasoning for keeping it (a board that swallows the document's `Cmd+Z` from anywhere is its own bug) is right, and it now says so in both the changeset and the code.

## The one thing to fix: the changeset describes a change that is no longer in the PR

The merge at `86a8a9a9a5` dropped two files that `3d2217a804` added:

| | at `3d2217a804` | at PR head | on `next` |
|---|---|---|---|
| `combobox.component.ts` — `AriaLabel` | 6 | **0** | 0 |
| `README.md` — `AriaLabel` mentions | 8 | **5** | 5 |

The path exists at the head and the file is present, so this is the content being gone rather than a mislocated check. And the cause is visible and ordinary: `next` recently landed `93215e61a1 refactor: rename Mj→MJ prefix on all ui-component classes`, which touches that exact file — so the merge had a real conflict there and resolving toward `next` took the additions with it.

The changeset still says (lines 19–21):

> **`mj-combobox` (`ng-ui-components`)** — New `AriaLabel`, `AriaLabelledBy`, `AriaDescribedBy` and `InputId` inputs on the inner text input, mirroring the naming contract `mj-dropdown` gained in #3863 **so the two sibling controls do not diverge.**

Changesets are the release notes here, so as it stands this announces a naming contract that does not ship — a consumer writes `<mj-combobox AriaLabel="Status">` and gets nothing. The sentence's own justification is the part that is now false: the two controls *do* diverge.

**Required:** delete that bullet (lines 19–21). The `#3863` line below it stays and remains accurate.

**Optional, your call:** restore `combobox.component.ts` and the README additions from `3d2217a804`. The audit's reasoning for doing it was sound, but it is net-new scope on a PR that is otherwise ready, and it re-opens the file that just had the rename conflict. Perfectly reasonable to take the four-line deletion now and carry the combobox gap as a follow-up — it was not among the nine findings this PR set out to close.

## Observations — none of these are asks

- **The focus-ring gate misses multi-line declarations.** I ran it against three fixtures: single-line misuse caught, both legitimate forms clean (no false positives), and `outline: 2px solid\n  var(--mj-focus-ring);` passes. The gate exists to catch the form that reads correctly, so the boundary is worth a line in the script header rather than a code change. The premise checks out — `--mj-focus-ring` really is a two-part box-shadow, so the outline shorthand does drop it silently.
- **The visibility delays are hand-coupled to the transition tokens.** `visibility 0s linear 300ms` against `--mj-transition-slow: 300ms`, and `200ms` against `--mj-transition-base: 200ms`. Correct today; change either token and the drawer's visibility drops mid-slide, with no test or gate to notice. You cannot reference those tokens directly since they carry the easing, so this may just want a comment naming the coupling.
- **No `shell.component` test, and no harness to extend.** The palette went 9 → 10 while the shell's `inert`, focus-return-to-toggle and skip link have none. Standing up a DOM harness for a component that size is a real lift, not a small ask, so I am flagging the gap rather than requesting it — but focus management is what regresses silently, and you found exactly that bug yourself when you routed the nav-item click through `closeMobileNav()`.
- **`ci-ui-tokens` now triggers on `packages/Angular/**/*.ts`**, so it runs on nearly every Angular PR rather than only stylesheet changes. Cheap job; just worth having chosen deliberately.

## Things I checked and found correct

The focus-return ordering is subtle and right — the flag is set, then focus moves synchronously *before* Angular applies `inert`, so `document.activeElement` is still inside the drawer when `contains()` is tested. Reverse those and it silently no-ops. `role="main"` on `mj-tab-container` instead of a wrapper avoids a real layout risk in a Golden Layout flex chain for identical semantics, and `tabindex="-1"` rather than `0` on the whiteboard host correctly avoids adding a Tab stop a keyboard user can reach and do nothing with.

Fix the changeset bullet and this is good to go from my side.

MattC-BC:
@rkihm-BC - I have made the requested changes regarding the changeset - please take a look when you have a moment - thanks!

rkihm-BC (approved):
Approving. The changeset now describes what ships, and the correction went past what I asked for. Verified against a clean board — `MERGEABLE`, `CLEAN`, 0 behind `next`, 18 checks green.

## The changeset fix

I asked for one thing — delete the `mj-combobox` bullet. Three of the four changes here are ones I did not raise, and all of them are right:

- **Dropping `"@memberjunction/ng-ui-components": patch` from the frontmatter.** The PR no longer touches that package at all, so the entry would have published a version bump for a package with zero changes. I flagged the bullet and missed the stale bump two lines above it.
- **"nine findings" → "eight findings"**, with the ninth explicitly attributed to #3860 and its follow-ups landing *"while this work was in flight"*. That is the accurate framing rather than a quiet renumber, and it matches the table: ten rows, minus item 4 deferred, minus item 3 now upstream.
- **Retitling the headline** from "nameable primitives" to "focus-ring token safety" — with the dropdown and combobox work gone, the old title described something the PR no longer does.

That is the distinction between correcting an error and correcting what caused it.

## Nothing was lost in this merge

Worth stating explicitly, because the previous merge is what created the problem: the `Mj→MJ` rename on `next` collided with the combobox addition, and resolving toward `next` took the additions with it — silently, since nothing downstream reads a changeset for accuracy.

So I checked every reviewed element survived this merge-forward (11 commits) rather than assuming: skip link, `role="main"`, both `inert` bindings, `avatarAriaLabel`, `skipToMainContent`, `returnFocusFromMobileNav`, `EnableGlobalShortcuts`, `hasFocusWithin`, `--mj-focus-ring-color` — all present. `aria-haspopup` correctly **absent**, so the self-audit's removal held. Ten command-palette specs. And `check-focus-ring.sh` is committed mode `100755` — a shell script without the executable bit passes review and fails in CI, so that one is worth having looked at.

## The self-audit remains the strongest part of this PR

Four findings actioned before any review reached you, and three would have been mine:

- **`aria-haspopup="menu"` on a menu with no `role="menu"` and no `menuitem` children** — a defect this PR's own first draft introduced, caught before it reached anyone. The attribute promises semantics that are then not delivered, which is worse than omitting it.
- **The focus-trap test with one focusable element**, where first === last made the two wrap directions indistinguishable and the spec would have passed against a half-broken trap. A test that cannot tell two states apart is the failure mode most worth hunting, and finding it in your own work is better than having it found for you.
- **Disclosing that the whiteboard guard gates the whole keydown handler** — undo/redo, `Escape`, `Delete`/`Backspace` — not only the eleven letters WCAG 2.1.4 is about. Materially broader than the original description, correctly argued (a board swallowing the document's `Cmd+Z` from anywhere is its own bug), and now stated in both the changeset and the code.

The focus-ring gate guards a real failure: `--mj-focus-ring` genuinely is a two-part box-shadow, so the outline shorthand drops it silently. I ran the gate against three fixtures — single-line misuse caught, both correct forms clean with no false positives.

The focus-return ordering is subtle and right: the flag is set, then focus moves synchronously **before** Angular applies `inert`, so `document.activeElement` is still inside the drawer when `contains()` is tested. Reverse those two and it silently no-ops.

Two judgement calls I would have made the same way: `role="main"` on `mj-tab-container` rather than a wrapper, avoiding a real layout risk in a Golden Layout flex chain for identical semantics; and `tabindex="-1"` rather than `0` on the whiteboard host, so a keyboard user does not get a Tab stop they can reach and do nothing with.

Routing the nav-item click through `closeMobileNav()` instead of setting the flag is the kind of detail that decides whether this actually works on a phone.

## Observations — explicitly not requested changes

- **The focus-ring gate misses multi-line declarations.** `outline: 2px solid\n  var(--mj-focus-ring);` passes, because matching is per line after comment stripping. Narrow, but the gate exists precisely to catch the form that reads correctly — worth a line in the script header stating the boundary rather than a code change.
- **The visibility delays are hand-coupled to the transition tokens.** `visibility 0s linear 300ms` against `--mj-transition-slow: 300ms`, `200ms` against `--mj-transition-base: 200ms`. Correct today; change either token and the drawer's visibility drops mid-slide, with no test or gate to notice. You cannot reference those tokens directly since they carry the easing, so a comment naming the coupling may be all this wants.
- **No `shell.component` test, and no harness to extend.** The palette went 9 → 10 while the shell's `inert`, focus-return and skip link have none. Standing one up for a component that size is a real lift, which is why this is a note — but focus management is what regresses silently, and the `mobileNavOpen = false` bug you found yourself is exactly the shape a test would pin.
- **`ci-ui-tokens` now triggers on `packages/Angular/**/*.ts`**, so it runs on nearly every Angular PR rather than only stylesheet changes. Cheap job; just worth having chosen deliberately.

Good work, and the honesty about what was not verified — the local template compile, the integration tier, the live keyboard walkthrough — is what made the rest of it reviewable.

PR #4123, opened 2026-09-01
feat(communication): first-class date bounds on GetMessages, and stop overwriting filters

JosueGarcia-BC:
## Why

A caller doing incremental sync — "give me everything since this watermark" — has no way to say so.
`GetMessagesParams` bounds a read by `NumMessages` alone, so approximating a date window means
guessing a count and filtering client-side, which **silently drops anything past the guess**. Both
providers that build a query string can filter by date server-side. The core just never asked them to.

This is the core change rather than a workaround in the calling app, as directed in review: some
providers won't be able to honour it, but many will, and the ones that can shouldn't be penalised for
the ones that can't.

## What

| addition | where |
|---|---|
| `ReceivedAfter` / `ReceivedBefore` | `GetMessagesParams` — inclusive, documented as such |
| `MessageRetrievalCapabilities` | `BaseCommunicationProvider`, declared **false** |
| `GetMessagesResult.AppliedFilters` | what the provider actually pushed down |
| `CombineFilterClauses` | new dependency-free `FilterUtils` module |

**Why a capability declaration and not just the fields.** `GetMessages` is `abstract` — there is no
base implementation to fall back on, so an unsupported filter is simply ignored by whichever provider
receives it. That failure is invisible: the caller gets *more* messages than it asked for and cannot
distinguish that from a mailbox that genuinely holds them. The base class declares nothing supported,
so a provider that hasn't considered the question describes itself accurately instead of
optimistically; `AppliedFilters` makes the same question answerable *after* the call. `undefined`
there means **unknown**, never "applied".

## A bug found on the way — in both providers

Each had independently grown the same defect: build a clause for `UnreadOnly`, then
**unconditionally reassign** from the provider-specific escape hatch.

```ts
if (params.UnreadOnly)            filter = "(isRead eq false)";
if (contextData && contextData.Filter) filter = contextData.Filter as string;  // <- discards the above
```

Gmail had the identical shape with `ContextData.query`. Passing both meant a caller asking for unread
mail **got read mail back**, with nothing reported. The same defect in two places is the argument for
fixing it centrally rather than twice — `CombineFilterClauses` composes instead of assigning, and the
next provider inherits the fix rather than re-deriving the bug.

It sits in its own dependency-free module beside `AddressUtils`, for the same reason that one is:
the provider suites mock `@memberjunction/communication-types` wholesale, so a pure module can be
pulled in with `vi.importActual` and the **real** composition gets exercised rather than a stub that
agrees with whatever the test expects.

## Two provider details worth reviewing

**Gmail sends epoch seconds, not `YYYY/MM/DD`.** The date form is day-granular *in the mailbox's own
timezone*, which would move an instant-based bound by up to a day without saying so.

**Gmail's operators are exclusive; the params are inclusive.** Each bound is widened by one second,
so the boundary message is returned rather than dropped. That errs toward one extra message — which a
caller de-duplicates — over losing one, which a caller cannot detect at all.

Gmail's **zero-result** path carries `AppliedFilters` too. Zero messages is exactly when a caller
can't tell "narrowed, nothing matched" from "narrowing ignored, mailbox empty".

## Backwards compatible

Every added field is optional; the capability default is `false`. No existing provider or caller
changes behaviour. Providers outside this repo keep compiling — they inherit the base getter.

## Proof

| package | before | after |
|---|---|---|
| `communication-types` | 99 | **113** |
| `communication-ms-graph` | 34 | **44** |
| `communication-gmail` | 45 | **54** |

**211 passing, up from 178.** All three typecheck clean (`tsc --noEmit`, exit 0).

The 33 new tests are written against the contract, not the implementation. The two
"NARROWS, it does not replace" blocks fail against the previous code — that's the regression they
exist for. Updating the two provider mocks was required, not incidental: adding the import broke 15
existing tests until the mocks pulled the real function, which is itself evidence the composition is
genuinely being exercised.

Not covered: no live Graph or Gmail call is made anywhere in this PR. The filter strings are asserted
at the client boundary, so what's proven is *what we send*, not what the services do with it.


packages/Communication/providers/MSGraph/src/MSGraphProvider.ts
```diff
@@ -124,6 +127,17 @@ interface ResolvedMSGraphCredentials {
  * The maximum lifetime, in minutes, that Microsoft Graph allows for a mail-resource
  * change-notification subscription (~3 days). Requested expirations are clamped to this.
  */
+/**
```
rkihm-BC, line 130: Nit: the doc comment directly above this one ("The maximum lifetime, in minutes, that Microsoft Graph allows...") belongs to `MSGRAPH_MAX_SUBSCRIPTION_MINUTES`, which now sits below `ODataInstant`. Inserting the function here orphaned it. Moving `ODataInstant` and its comment above that block (or below the constant) puts the comment back on its symbol.

packages/Communication/providers/gmail/src/GmailProvider.ts
```diff
@@ -132,8 +135,31 @@ interface GmailPushData {
 /**
  * Implementation of the Gmail provider for sending and receiving messages
  */
+/**
```
rkihm-BC, line 138: Nit, same pattern: "Implementation of the Gmail provider for sending and receiving messages" just above is the class doc for `GmailProvider`, and `GmailBoundSeconds` now sits between them. Moving the helper above that comment restores it to the class.

rkihm-BC (changes requested):
The design is sound and the bug fix is real. I verified the description rather than taking it on trust:

- Swapped the branch's files into a local tree and ran all three packages: **113 / 44 / 54 passing, 211 total**, matching the numbers above. CI Build and all six unit shards are green.
- The overwrite defect is confirmed on `next` in both providers: each builds an `UnreadOnly` clause and then unconditionally reassigns from `ContextData.Filter` / `ContextData.query`.
- Google's Gmail API filtering guide documents epoch seconds for `after:` / `before:` (its own example is `after:1388552400 before:1391230800`) and says the date form is interpreted as midnight PST, so sending seconds is the right call. It does not document inclusive vs exclusive, so widening each bound by one second is correct either way.
- Graph's unquoted ISO instant with `ge` / `le` is the documented OData form, and this provider does not combine it with `$orderby`, so the "restriction too complex" trap does not apply.

The capability getter plus `AppliedFilters` is the right shape for an abstract method with no fallback, and it sits comfortably beside the existing `SubscriptionCapabilities` pattern. Carrying `AppliedFilters` on Gmail's zero-result path is a detail that was easy to miss and was not.

**Required before merge: a changeset.** This adds public API to three published packages and there is none (see the changeset-bot comment). Changesets drive versioning and their text is what the release notes are generated from, so without one this feature ships undocumented. `patch` is the correct level since the branch carries no migration or metadata change:

```md
---
"@memberjunction/communication-types": patch
"@memberjunction/communication-ms-graph": patch
"@memberjunction/communication-gmail": patch
---

feat(communication): `GetMessagesParams.ReceivedAfter` / `ReceivedBefore` push inclusive date bounds down to providers that support them (MS Graph, Gmail). Providers declare support via `MessageRetrieval` and report what they applied in `GetMessagesResult.AppliedFilters`. Also fixes both providers silently discarding `UnreadOnly` when `ContextData.Filter` / `ContextData.query` was also supplied.
```

**Two non-blocking nits inline:** the new helper functions were inserted between an existing doc comment and the symbol it documented, in both providers.

**Follow-up worth an issue, not for this PR:** Twilio already filters by date through `ContextData.dateSent`, and its SDK supports before/after bounds, so it is the obvious next provider to declare `FilterByReceivedDate`.

rkihm-BC (approved):
Approving. The one required item is done: a `patch` changeset for `communication-types`, `communication-ms-graph` and `communication-gmail`, whose text names both user-facing changes — the `ReceivedAfter` / `ReceivedBefore` bounds with `MessageRetrieval` capability reporting and `AppliedFilters`, and the fix for `ContextData.Filter` / `ContextData.query` silently discarding `UnreadOnly` — so the release note will be right.

Everything else stands from the first review: 211 tests verified locally, the overwrite bug confirmed on `next` in both providers, Google's Gmail API guide documenting epoch seconds for `after:` / `before:`, and the Graph filter form. The two orphaned doc-comment nits were optional and I would not hold the PR for them. CI fully green; behind `next` but mergeable.

PR #4151, opened 2026-09-01
fix(integration): a provable key must have no observed duplicate; a capped scan is a prefix

MS-BC:
Three defects that together let discovery trust a sample it shouldn't. The first is the serious one.

## 1. A column with duplicates was declared a *provable* primary key

`subsetKeyness` decided keyness purely by Chao1 domain saturation:

```ts
const Dhat = f2 > 0 ? d + (f1*f1)/(2*f2) : d + (f1*(f1-1))/2;
return { Dhat, isKey: n >= minRows && Dhat > n };
```

It never checked the sampled tuples were distinct. Chao1 estimates how large the value **domain** is — evidence about *future* collisions — which says nothing about collisions already sitting in the sample. And its bias-corrected branch (no doubletons) grows **quadratically** in the singleton count.

Measured, on 100 rows where nine share one value:

```
stats    = 92 distinct of 100 rows
verdict  = "Provable 1-column key {customer_id} — Chao1 domain D̂≈4187 ≫ n=100"
```

Nine duplicates, reported as provable. That does not error — the sync **silently merges two source records onto one row**. Data loss, and the mirror image of the duplicate-record failure.

Keyness now requires `d === n` first; Chao1 is kept for what it is actually good at — separating an identifier from a small category that merely hasn't collided yet in `n` rows. Genuine composite keys whose individual columns repeat are unaffected, and pinned by a test.

## 2. A record-capped scan reported itself as an exhausted source

The cap is enforced by the **generator** feeding the scan — it simply `return`s — so the stream ends indistinguishably from a table that ran out of rows, and the scan reported `'exhausted'`.

That matters because `ScanComplete` unlocks the lenient near-unique (0.9) soft-key rule instead of the strict 1.0 reserved for partial corpora. So a 50-row prefix of a ten-million-row table was judged complete, and over a short prefix an ordinary foreign key like `customer_id` is very likely to look ≥90% distinct.

`StoppedReason` gains `'record-cap'`, the producer's target is passed to the scan so it can recognise its own truncation, and completeness now means `=== 'exhausted'` rather than `!== 'time-budget'`.

## 3. The sample target returns to 500

It was lowered to the primary-key significance floor on the reasoning that 50 rows answer the other two questions well enough. Both halves of that were wrong:

- **Width.** Several connectors declare no widths at all — YourMembership's own code states `MaxLength: null` and unions in the sampler's result, so **the sample is the only source of width**. Ten times fewer observed values means narrower columns, and a value too wide for its column is *skipped* at sync time, not truncated. Records silently lost.
- **Custom columns.** A field populated on ~1% of records is near-certain in 500 rows and a coin flip in 50.

A floor is a minimum, not a target to sample down to. And the speed argument doesn't hold either: streaming rows from one object is one paged read — what costs requests is *descending into parents*, which is bounded separately by the child's own demand.

## Tests

`ProvableKeyRejectsDuplicates` (6) — the exact 92-of-100 shape, a single duplicate among 500, a genuine unique column still accepted, a genuine composite whose parts individually repeat still accepted, a composite whose combined tuple repeats refused, a low-cardinality category refused.

`ScanCompletenessHonesty` (10) — `record-cap` vs `exhausted` vs no cap, a source ending exactly at the cap treated as a prefix, and three that drive the **connector's own decision** rather than recomputing it (an earlier draft recomputed `ScanComplete` in the test and consequently survived the mutation).

All three fixes **mutation-verified**. Engine suite: **77 files / 998 tests** green.

### DIFF
--- a/.changeset/sample-trustworthy.md
+++ b/.changeset/sample-trustworthy.md
@@ -1,0 +1,32 @@
+---
+"@memberjunction/integration-engine": patch
+---
+
+Three fixes to make a discovery sample trustworthy: a provable key now requires no observed
+duplicate, a capped scan admits it only saw a prefix, and the per-table sample target returns to 500.
+
+**A duplicate in the sample disproves keyness.** `subsetKeyness` decided keyness purely by Chao1
+domain saturation and never checked the sampled tuples were distinct. Chao1 estimates how large the
+value domain is — evidence about future collisions, not about ones already in the sample — and its
+bias-corrected branch grows quadratically in the singleton count. A column with nine copies of one
+value among 100 rows scored D̂ ≈ 4187 and was reported as a *provable* primary key. Two source
+records sharing that value then collapse onto one row: silent data loss, the mirror image of a
+duplicate record. Keyness now requires `d === n` first, with Chao1 retained to separate an identifier
+from a small category that merely has not collided yet. Genuine composite keys whose individual
+columns repeat are unaffected.
+
+**A record-capped scan is a prefix, not an exhausted source.** The cap is enforced by the generator
+feeding the scan, which simply returns, so the stream ended indistinguishably from a table that ran
+out of rows and the scan reported `exhausted`. That unlocked the lenient near-unique (0.9) soft-key
+rule instead of the strict 1.0 reserved for partial corpora. `StoppedReason` gains `record-cap`, the
+producer's target is passed in so the scan can recognise its own truncation, and completeness now
+means `exhausted` rather than "not time-budget".
+
+**The sample target returns to 500.** It was lowered to the primary-key significance floor on the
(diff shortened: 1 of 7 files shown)


MS-BC:
Closing. This bundled a correct revert with two changes that would have made things worse:

- requiring a provable key to have no observed duplicate, and
- treating a record-capped scan as untrustworthy.

Both argued for more strictness, and strictness does not fail safe here. Integration primary keys are soft constraints — identity for dedup, never able to reject a row — and a sample legitimately contains repeats (the same record is fetched under multiple parents during DAG sampling; paged reads overlap). With a 500-row sample every table larger than that ends at the record cap, so those two changes together would have denied a key to almost every large object. An object with no key is dropped. That trades away the whole table to avoid a key that occasionally merges.

The one part worth keeping is the sample-target revert, which will be re-raised on its own.

PR #4154, opened 2026-09-01
Records region: single temporary tab (preview-tab behavior)

MattC-BC:
Records tabs get VS Code preview-tab behavior, matching the main Golden Layout tabset. Implements the plan this PR originally carried; the plan doc is updated in place with the resolved decisions.

## The problem

Record tabs were born unpinned, and so rendered italic — the shell's visual vocabulary for "temporary". But the temp-tab machinery was deliberately switched off for them when the region shipped (#3444), to stop a nav click from replacing an open record. The italic was inherited; the behavior never was. Every record open minted a tab that nothing could replace or promote, which is the tab explosion when browsing records in an Open App.

## The fix

A **second temp-tab pool** for the records region, rather than weakening the nav-side protection.

- `TabRequest.TempScope` (`'main' | 'records'`, default `'main'`) scopes both consumption in `OpenTab` and the pin cascade in `OpenTabForced`, selected by a new settable `WorkspaceStateManager.RecordsRegionTabFilter`. The predicate defaults to **false** when unset, the inverse of `isMainLayoutTab` and deliberately so: under the classic style a records-scoped request must consume nothing rather than everything. The pools are disjoint in both directions — a nav click still can never consume a record, and a record open no longer disturbs the nav tab's temp status. This replaces the blunt `PreservePinState` opt-out record opens used to pass; the flag remains for callers that want no cascade at all.
- A plain record click reuses the region's temporary tab. Shift-click, or an explicit `forceNewTab` (which finally makes `single-record`'s "Open in New Tab" do something — it was inert), adds a tab and promotes the previous one. Double-click and right-click → Pin already worked.
- Consumption **reuses the tab id**, which is what keeps `layoutCoversExactTabSet` satisfied so desktop split layouts survive a replacement. That means the records sync path needed the main path's `needsReload` treatment: detect the Entity/record change on a reused id, detach the outgoing record into the component cache, reload, re-capture the origin crumb, and stand down during batch creation and breakpoint rebuilds.
- Deep-link and URL-driven opens are scoped the same way, closing an asymmetry where a deep link could consume the nav temp tab and convert it into a records tab. Scoped in `ShellComponent.processTabRequest` rather than at the route resolvers, which live in MJExplorer and must not be modified.
- Fixes `updateTabTitleFromResource` writing records-tab titles to the main layout manager, which does not own that tab id. Masked while records tabs were immortal (the next config emission re-applied the title); replacement retitles on every plain click, so it stopped being cosmetic.

Open Apps inherit all of it through the shell with no changes on their side.

## Data safety

Replacement destroys the pane, so an edit in the temp tab could vanish on the next click. **A record whose form is in edit mode leaves the consumption pool**, so the next plain open lands in its own tab instead. Same user-visible outcome as VS Code's promote-on-modify, without a dirty-tracking pipeline: `MjEntityFormHostComponent` already exposes the live form, so this is a synchronous read through a new `BaseResourceComponent.IsEditing()` (default `false`). MJ forms are read-only until the user explicitly clicks Edit, so edit mode is a deliberate gesture and a sound proxy for "work worth protecting".

New (unsaved) records open **pinned** for the same reason — they are unsaved from the moment they open.

## Origin crumbs (regression found and fixed during review)

Replacement broke the origin crumb in a way worth calling out, since it is the subtlest part of this PR.

`ReturnToRecordSource` has a fast path that reactivates the parent's tab when it still holds the parent, and that tab kept its original crumb. In-record links used to open a **new** tab, so the fast path always hit. Once links replace, the parent's tab holds the child, the fast path misses, and the fallback re-opened the parent while capturing wherever the user was standing — the child. Result: open a record (`Data Explorer › MJ: Action Categories`), follow an in-record link (`User Management`, correct), click that crumb, and the crumb became `System`. The two records pointed at each other and the entry point was unreachable.

Fixed by carrying the parent's own origin forward (`RecordSourceContext.sourceParentOrigin`) and replaying it on return instead of recapturing. The contexts nest, so it unwinds at any depth, capped at `MAX_RECORD_ORIGIN_DEPTH` because the chain is persisted with the workspace. `refreshSourceContext` clears the chain alongside the fields it already wipes, so a stale chain can't survive a deliberate re-capture. With no chain it passes `'none'` — a missing crumb beats a circular one.

Verified live: `Data → User Management → System → crumb → crumb` now walks all the way back to the Data grid.

## Two calls worth a reviewer's attention

- **Shift-click polarity.** Mirrors the main tabset (new tab temporary, previous promoted) rather than the inverse described in the original ask. The ask is literally "behave like the main Golden Layout tabset", so the description read as a recollection of that tabset rather than a separate requirement. One line plus one assertion to flip.
- **Pinned new-record tabs.** Visible consequence: a brand new record shows a thumbtack and loses its close X. That is the mechanism keeping unsaved work out of the pool, but it is a UX call and can be revisited.
- Known and accepted for v1: shift detection is a global mousedown listener, so shift+Enter on a grid row registers no modifier and replaces.

## Testing

32 new tests across the three packages — pool semantics and the scoped cascade in `ng-base-application`, the open-routing fork and the origin chain in `ng-shared` (first coverage of `OpenEntityRecord`), and the reload decision in `ng-explorer-core`. Suites pass at **53 / 99 / 175**. `check:standards` and `check:changeset` pass. Verified live in MJExplorer, including the full crumb round trip.

The deterministic integration tier runs 64/72 with 6 tier-gated skips. The two failures — `IT31` (`MJ: Entity Fields` count off by one) and `IT50/CD3` (`MJ: Entities.SubtypeSelector` in the generated schema but not live metadata) — are database/metadata consistency checks on a dev DB, unrelated to this branch, which contains no SQL, migration, metadata, or generated entity. `IT50` was already red before any of this work was applied.

Patch changeset across `@memberjunction/ng-base-application`, `@memberjunction/ng-shared`, `@memberjunction/ng-explorer-core`. Targets edge.6.


MattC-BC:
Hey @rkihm-BC - I am putting this one in as Amith was hoping to get this into this week's build.

rkihm-BC (approved):
## Approve — nothing required, four follow-ups filed

Careful work on a subtle area, and the reasoning is left in the code where the next person will need it. All 22 checks pass, including the deterministic integration tier, so the two failures noted in the description were dev-DB artifacts as claimed. Nothing here should hold the build.

### What I verified rather than took on trust

**The reload-precedence claim is exact.** `reloadRecordsTabIfResourceChanged` compares `(config['recordId']) || tab.resourceRecordId || ''`, and `getResourceDataFromTab` builds `ResourceRecordID` with the identical expression. You name the hazard of reading them in a different order — a permanent difference that reloads the pane on every configuration emission — and then avoid it precisely. That is the kind of thing that is invisible until it is a support ticket.

**The edit-mode guard actually works, end to end.** This one could easily have been a hook nobody consulted, so I traced it: `componentRefs` is one map keyed by tab id, populated in `loadTabContent`, which the records path calls too; the instance is `EntityRecordResource`, which overrides `IsEditing()` and holds `@ViewChild(SingleRecordComponent)`; that holds `@ViewChild(MjEntityFormHostComponent)` and reads `Form.EditMode`; the host renders unconditionally so both view children resolve. The data-safety guarantee is real at runtime.

**The origin chain is bounded where it is written.** `TruncateRecordOriginChain` is applied at both write sites, `refreshSourceContext` clears `sourceParentOrigin` alongside the fields it already wiped, and `'none'` is a pre-existing sentinel that returns an empty context — so the "a missing crumb beats a circular one" fallback does what the comment says.

The `updateTabTitleFromResource` fix is a genuine latent bug that only replacement would have exposed, and putting the deep-link scoping in `processTabRequest` is right given the route resolvers live in MJExplorer.

Test coverage is strong where it counts — pool semantics both directions, the scoped cascade, dedup beating consumption, docked and pinned exclusions, classic style untouched, and the reload decision including its stand-down guards.

### Follow-ups (none blocking)

- **#4344** — the edit-mode guard is the only major behavior here without a test, and its failure mode is silent and destructive. The pool tests inject their own filter, so nothing exercises the shell's composed predicate; drop the `&& !this.tabContainerRef?.IsRecordTabEditing(tab.id)` clause in a refactor and every test still passes. Highest value of the four.
- **#4345** — the protection is transient, not a promotion, so the description's "same user-visible outcome as VS Code's promote-on-modify" does not quite hold. Save calls `EndEditMode`, the tab rejoins the pool, and the next plain click replaces the record you just finished editing. No data loss. Promoting on first edit would also fix the cosmetic point that an editing tab still renders italic while protected.
- **#4346** — the pin on new-record tabs is largely redundant, since `BaseFormComponent.ngOnInit` starts edit mode for any unsaved record and the edit guard already excludes it. The residual value is closing the pre-init view-child window, which is real; weigh that against the permanently hidden close button (`tab-container.component.css`, `.lm_tab.pinned .lm_close_tab { display: none }`). You flagged this yourself — the issue just records the trade-off so it is not lost.

I also raised and then withdrew a concern about `!this.tabContainerRef?.IsRecordTabEditing(...)` failing open when the ref is undefined. It collapses to the same case the code already documents and handles: an unloaded tab reports not-editing, and there is no rendered pane to lose either way.

PR #4155, opened 2026-09-01
docs(plans): OpenClaw 2.0 architecture review — ranked recommendations for the agent framework

AN-BC:
## What this is

A single new document, **`plans/openclaw-2-agent-architecture-review.md`**, reviewing **OpenClaw v2026.8.1 ("OpenClaw 2.0", released end of August 2026)** against MJ's agent framework and distilling the best ideas into **twelve ranked recommendations**. Docs-only — no code, schema, or metadata changes; each recommendation is scoped for its own follow-up work.

Research basis: a full clone of `openclaw/openclaw` at v2026.8.1 (release notes, `VISION.md`, `SECURITY.md`, concept docs, and source for the agent loop, context engine, memory core, Skill Workshop, approvals, and security layers), press coverage of the 2.0 release, and a matching deep map of MJ's agent stack at `6.1.0-edge.4` (`base-agent.ts`, `MemoryWriteManager`, `ConversationCompactionManager`, `AIPromptRunner`, HITL, watchdog, providers).

## Framing

The two architectures are near-inverses (agent-as-workspace-files vs. agent-as-metadata-rows; one-trusted-operator vs. multi-tenant RLS; imperative orchestration vs. metadata graphs). The document explicitly does **not** propose converging on OpenClaw's architecture — on governance MJ is structurally ahead, and their own docs concede it. What it proposes absorbing is their operational hardening for always-on agents, most of which lands in MJ as new columns, validation gates, and one background agent.

## The recommendations

**Tier 1 — adopt**
- **R1 Memory provenance as a structural boundary** — `OriginClass`/`SessionKind`/taint on `MJ: AI Agent Notes`; untrusted-origin content barred from hardening and auto-injection; recall-loop prevention; purge-by-provenance.
- **R2 Verified compaction** — summaries must prove UUIDs and pending asks survived; abort-don't-write on failure; memory flush sequenced before compaction; pinned in IT30.
- **R3 Prompt-cache layout contract** — stable catalog prefix / volatile suffix in agent-type templates; deterministic catalog ordering; cache-hit rate per agent on the dashboard.
- **R4 Governed self-improvement** (the Skill Workshop analog) — an `MJ: AI Agent Improvement Proposals` entity + scheduled reviewer agent; hash-bound, scanned, Record-Changes-versioned apply; review routed through `AI Agent Requests`.
- **R5 Content-bound approvals** — `ContentHash` on `AI Agent Requests`; re-verify on resume; cancel pending requests when their run dies.
- **R6 Untrusted-content envelope** — `ContentTrust` on Action results; boundary wrapping; chat-template token stripping for local-model providers; outbound scrubbing in MessagingAdapters.

**Tier 2 — adapt**
- **R7** mid-run steering queue · **R8** writer fencing on run/conversation writes · **R9** selection-source-aware model failover (strict user picks, turn-local fallback, never silently change billing class) · **R10** standing intents (event-based prospective memory with a deterministic, model-free matcher) · **R11** budget the model-visible action catalog (progressive disclosure, per-call-tax review lens) · **R12** rendered-prompt snapshots with CI drift checking.

Plus a short "further candidates" list (multiplayer sessions, Actions-as-callables in sandboxed code, `mj doctor`, maturity scorecards), a "what we're explicitly not copying" section, and a suggested 4-phase sequencing.

## Validation

- Docs-only change; no package builds affected. Relative links in the doc verified against existing files (`guides/AGENT_FRAMEWORK_COMPARISON.md`, `guides/AGENT_MEMORY_GUIDE.md`, `plans/agent-inflight-memory-writes.md`, `plans/agent-conversation-compaction.md`).
- No changeset included: no package or migration is touched (`changes.yml` gates migrations only; `check:changeset` is documented as local-only).

https://claude.ai/code/session_01QYutacFqTG7sXZDGq2wrXh

---
_Generated by [Claude Code](https://claude.ai/code/session_01QYutacFqTG7sXZDGq2wrXh)_

### DIFF
--- a/plans/openclaw-2-agent-architecture-review.md
+++ b/plans/openclaw-2-agent-architecture-review.md
@@ -1,0 +1,205 @@
+# OpenClaw 2.0 Architecture Review — What MJ Agents Should Absorb
+
+## Technical Architecture Review & Recommendations
+
+**Version**: 1.0
+**Date**: September 2026
+**Status**: Proposed — for review and prioritization
+**Owner**: AI Agents subsystem
+**Companions**: [`guides/AGENT_FRAMEWORK_COMPARISON.md`](../guides/AGENT_FRAMEWORK_COMPARISON.md) (MJ vs. the orchestration libraries), [`plans/agent-inflight-memory-writes.md`](agent-inflight-memory-writes.md), [`plans/agent-conversation-compaction.md`](agent-conversation-compaction.md)
+
+---
+
+## Executive Summary
+
+**OpenClaw** (lineage Warelay → Clawdbot → Moltbot → OpenClaw) is the most-watched open-source personal-AI-agent platform. Its **v2026.8.1 release — "OpenClaw 2.0," end of August 2026** — merged 16,000+ PRs from 933 contributors in a single cycle and rebuilt memory, skills, automations, approvals, and collaboration. This review is based on a full clone of `openclaw/openclaw` at v2026.8.1 (docs + source), its release notes / `VISION.md` / `SECURITY.md`, and a matching deep map of MJ's agent stack at `6.1.0-edge.4`.
+
+**The two architectures are near-perfect inverses**, and this document does **not** propose converging on theirs:
+
+| | OpenClaw 2.0 | MemberJunction |
+|---|---|---|
+| Agent definition | Workspace files (`AGENTS.md`, `SOUL.md`, `USER.md`) + JSON5 config + per-agent SQLite | `MJ: AI Agents` row + ~30 related entities; generic `BaseAgent` runtime |
+| Trust & tenancy | **One trusted operator per Gateway**; session keys are "routing controls, not authorization boundaries" (their words); tenants = separate containers (`openclaw fleet`) | Multi-tenant; agents run *as a user* through RLS/field perms/Record Changes; per-agent View/Run/Edit/Delete |
+| Orchestration | Deliberately imperative — no graph DSL; sub-agent spawn + code-mode "Swarm" | Loop / Flow (metadata graph → durable TaskGraph) / Realtime, swappable per agent |
+| Shared state | Transcript + files; no typed state | Typed payload with path-level ACLs and diff-based mutations |
+| Observability | Metadata-only audit ledger, off by default | Every run/step/prompt a relational row: tokens, cache, cost rollups, payload before/after |
+
+On governance MJ is structurally ahead, and OpenClaw's own docs concede it ("collaboration controls… not hostile-tenant isolation"). What OpenClaw has that MJ does not is **operational scar tissue from running always-on personal agents at enormous user scale**: mechanisms for memory trust, compaction safety, prompt-cache economics, governed self-improvement, and approval integrity. Most of them translate into MJ as **new columns, validation gates, and one background agent** — they fit MJ's metadata DNA rather than fighting it.
+
+This document proposes **twelve recommendations**, ranked. R1–R6 are *adopt* (MJ-shaped already); R7–R12 are *adapt* (design work needed). A final section lists further candidates deliberately left out of the first wave.
+
+---
+
+## What we are explicitly NOT proposing
+
+To prevent misreading, these OpenClaw traits were evaluated and rejected for MJ:
+
+- **Single-operator trust model / session-keys-as-routing.** MJ's row-level, run-as-user model is the moat. OpenClaw needs a container per tenant to do what MJ does with rows.
+- **Filesystem as the agent authoring surface.** MJ's DB metadata is what makes agents UI-editable, permissioned, diffable via `mj sync`, and auditable.
+- **In-process trusted plugins.** ("Plugins are the trusted computing base" — their docs.) MJ's Actions boundary + `isolated-vm` code execution is the better governance posture.
+- **Abandoning graph orchestration.** OpenClaw's "no graph DSL" stance fits personal assistants; MJ's `FlowAgentType` → durable TaskGraph serves governed business processes their audience doesn't have.
+
+---
+
+## Tier 1 — Adopt
+
+### R1. Memory provenance as a structural boundary
+
+**Impact: High · Effort: Medium · Touches**: `MJ: AI Agent Notes` schema, `packages/AI/Agents/src/MemoryWriteManager.ts`, `packages/AI/Agents/src/memory-manager-agent.ts`, `packages/AI/Agents/src/AgentContextInjector.ts`
+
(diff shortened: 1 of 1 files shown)


ct-bc:
One main note is we should move R11 to Phase 1.
R11's fix is capping with overflow-to-on-demand expansion, which changes prompt structure and moves the cache boundaries. In the plan it says "Phase 1 sets the layout contract (R3) and locks it in snapshots (R12), then Phase 3 restructures the catalog and invalidates every fixture." You'd regenerate the snapshot set and reopen the layout contract for a reason unrelated to any behavior change. R3's cache win is also only half real while the catalog is unbounded, since every action added reshapes the prefix. Butting these together (R3, R11 and R12) makes sense as they are 3 parts of the same surface.

Another note is that we should put Part 3 of R6 into phase. There is currently the issue of letting web page inject system-level instructions. This is an important fix and can be fixed apart from the rest of R6 (ie added into phase 1).

On R1:  The doc says AIAgentNote "records nothing about where a memory came from." It already has AuthorType (Agent | MemoryManager | User), SourceConversationID, SourceConversationDetailID, SourceAIAgentRunID, DerivedFromNoteIDs, ConsolidatedIntoNoteID, RootConsolidatedIntoNoteID, ProtectionTier and ExpiresAt. This means that the fix is actually easier than stated.

On R5: MJAIAgentRequestEntity.Status is Approved | Canceled | Expired | Rejected | Requested | Responded, so Canceled exists and the cancellation state machine is there. The open question is only whether run cancellation drives it. It also has OriginatingAgentRunID, OriginatingAgentRunStepID, ResumingAgentRunID, ResponseSchema, ResponseData, ExpiresAt and Priority. Once again this means that the fix is less complicated that stated 

I have more notes and thoughts but these are the main ones. I'm happy to work on any of these and happy to talk more in person.


Phase 1 (Gates, contracts, and the model-visible surface) : R12, R3, R11, R2, R5, + R6 step 3, + R9 billing policy 
Phase 2 (Security/trust wave): R1 + R6 (steps 1, 2, 4)
Phase 3 (Runtime hardening): R8, R9 
Phase 4 (Feature-scale):  R4 (gated on 1 and 2); R7 and R10 alongside, ungated

AN-BC:
@ct-bc @rkihm-BC — I mapped **R11 (budget the model-visible surface)** and **R3 (prompt-cache layout)** against the current code on `next`. Nothing here blocks this docs-only PR; it's input for whoever picks up the R11 follow-up, because a few of the premises have moved since the review was written.

## 1. R11's measurement already exists in-repo — and it argues for *budgeting*, not *gating*

`packages/AI/Agents/src/native-tools/action-tool-builder.ts:38-56` records an A/B over a 1,026-observation corpus at a 1,024-char vs 4,096-char tool-description cap:

- At **1,024** (descriptions trimmed, names + result codes kept): selection accuracy **improved** — +6.3pp Gemini, +11.3pp GPT-5.6-luna.
- At **4,096** (full prose): both gains disappeared, GPT-OSS-120B fell a further 5pp, its prompt grew ~700 tokens/call, and it began returning *no action at all* on cases it had previously handled. Net −10 correct turns.

The companion finding in `describeTool` (same file, ~:160) is the other half: an earlier build omitted outputs and result codes and `P(right tool | acted)` fell **82% → 70%** on GPT-OSS-120B.

Read together these say the *prose* is the dilution and the *names, types and result codes* are the signal — which supports R11's cap, but points at **graded detail across the whole catalog** rather than a visible/hidden split. At detail level 0 an action renders in roughly 45–70 tokens vs 120–200 at full prose; a 50-action agent drops from ~8k to ~3k tokens with every name still visible, no discovery turn, and no recall-miss failure mode. Worth settling with the existing harness (`PromptEvalDriver` + corpus, `plans/native-tool-calling/2026-09-02-native-tool-calling-test-plan.md`) before choosing a shape.

## 2. The concrete R11 work item: the prose path has no budget at all

Four disclosure policies currently coexist, and actions are the outlier:

| Class | Rendered as | Where |
|---|---|---|
| Sub-agents | name + description, one line | `base-agent.ts:7744` |
| Skills | name + description; `Instructions` withheld until activation — comment says *"progressive disclosure by design"* | `base-agent.ts:7762` |
| **Actions** | **name + description + every input/output param + every result code, uncapped** | `base-agent.ts:7807`, `:7985` |
| Client tools | full JSON schema inline | `base-agent.ts:7846` |

So the same five sections get a measured 1,024-char cap on the native path and no cap at all on the prose path, decided purely by which mode the agent is in. `formatActionParameter` (`:7985`) emits every param description in full for inputs *and* outputs. Porting `DESCRIPTION_DETAIL_LEVELS` to the prose renderer is hours of work and needs no schema change.

Related: two independent filters decide model visibility today — `formatActionDetails` renders `activeActions`, while `buildActionToolSet` separately applies `filterDeclarableActions` on `DeclareAsNativeTool`. Any budgeting policy added now gets implemented twice unless visibility is resolved once in `buildAgentBaseCatalog`.

## 3. Discovery and activation already ship

Before anyone builds a capability-search layer:

- **Discovery** — `Find Candidate Actions` / `Find Best Action` / `Find Candidate Agents` / `Find Best Agent` in `packages/Actions/CoreActions/src/custom/ai/`, running on `Provider.SearchEntity`: hybrid lexical + semantic, fused with canonical weighted RRF (`ComputeRRF`, `providerBase.ts:2022`), permission-filtered, over the daily-synced `Actions Search` / `AI Agents Search` EntityDocuments.
- **Activation** — `ActionChange` (scope/mode/limits), and `enableSkillCapabilities` (`base-agent.ts:12574`) pushes onto `params.actionChanges` with the next turn's `gatherPromptTemplateData` picking it up automatically.

Two real gaps rather than rebuilds:
- `Find Candidate Actions` searches everything the **user** can see, not the **agent's** granted set. It's granted today only to Agent Manager and ActionSmith as an *authoring* tool. Pointed at runtime discovery as-is, an agent surfaces actions it can't call. Needs an agent-scope param on `BaseFindActionsAction`.
- There's no **`AI Skills Search`** EntityDocument — Actions, Agents, Prompts, Queries and Entities all have one. That's a single metadata row.

## 4. A constraint R11 should probably state explicitly

The 82%→70% regression above had a specific cause: `Search Query Catalog` carries `NO_MATCHES ✓ … Proceed with ad-hoc SQL` — a result code whose value is that it tells the model **when to use a different action**.

Put that next to R11's own OpenClaw doctrine (*tool descriptions must never statically reference tools from other toolsets, because gating turns the reference into hallucination bait*) and the two combine into a design constraint: **introducing tiers converts MJ's most valuable disambiguation signal into its most dangerous dangling reference.** Hide `Search Query Catalog` and the still-visible `Run Ad-hoc Query` loses the line saying when to prefer the other; hide `Run Ad-hoc Query` and the visible one names something the model can't see.

Practical implications: if we tier, tier by **`ActionCategory`** (already hierarchical via `CategoryID`/ParentID) rather than per-action, so co-referencing actions move together — and add a lint for action descriptions and result codes naming other actions, which is R11's "audit for references to non-granted capabilities" made checkable.

## 5. R3 — one line, since the cache work is running separately

`ANTHROPIC_CACHE_BREAKPOINT` (`packages/AI/Providers/Anthropic/src/models/anthropic.ts:25`) exists, splits the system prompt at up to 4 breakpoints, and **appears in exactly one file in the repo — the one that defines it.** Zero agent-type templates emit it, even though `loop-agent-type-system-prompt.template.md` is already laid out correctly (capabilities in the stable head; `_CURRENT_DATE`, scratchpad, `_CURRENT_PAYLOAD` in the volatile tail). Flagging only because it changes R11's economics materially — cached input is ~10% of list price, so a stable catalog is cheap once the breakpoint lands and a discovery round trip is not.

Happy to open issues for any of these, or to write up the graded-detail option in more depth if it's worth comparing side by side with the tiered one.

---

AN-BC:
@ct-bc the above is Claude, not 100% sure I agree with all of it. But it is lower priority than the prefix caching stuff and we can evaluate what we want to do here after that is done

PR #4158, opened 2026-09-01
fix(entity-viewer): make the grid's Delete affordance reachable

MS-BC:
## The bug

The grid wrapper deletes correctly and always has — `onDeleteConfirmed()` walks every selected record, calls `Delete()` on each, logs per-record failures without aborting the rest, and re-requests the page.

None of it was reachable. The template renders the Delete button only when:

```
ToolbarConfig.showDelete && AllowDelete && HasSelection && !ShowDeleteButton
```

`_showDeleteButton`, `_allowDelete` and `ToolbarConfig.showDelete` all default to `false`, and `grid-view-renderer.component.ts` passed `[ToolbarConfig]="config.toolbarConfig ?? {}"` while never binding `AllowDelete` at all. So unless a view's *saved config* happened to set the flag, the button could not appear.

The user-visible consequence is that there is no multi-record delete anywhere the grid is used. Removing ten rows means opening ten records, one at a time. This was reported from a live environment as "there is no way to select and then delete".

## The fix

`showDelete` now defaults to whether the user may actually delete this entity, and `AllowDelete` is bound to the same permission:

```ts
get canDelete(): boolean { /* entity.GetUserPermisions(user)?.CanDelete ?? false */ }

get effectiveToolbarConfig(): GridToolbarConfig {
  const cfg = this.config.toolbarConfig ?? {};
  return { ...cfg, showDelete: cfg.showDelete ?? this.canDelete };
}
```

Defaulting to the **permission** rather than to `true` matters: it keeps the button off exactly where the delete would have been refused anyway, so nobody is offered an action that will fail. An explicit `showDelete` in a view's config still wins in both directions — but it can never grant the permission itself, because `AllowDelete` is bound to the check rather than to the config.

The permission check mirrors `recycle-bin.component.ts`, this package's other delete surface, so the two agree on what "can delete" means instead of each inventing a rule. A metadata provider that cannot answer is treated as "no", never as permission.

## Tests

`grid-delete-affordance.test.ts`, 8 tests, **mutation-verified** — all 8 fail on the pre-change source. They pin the gate rather than the button: offered when the user may delete, withheld when they may not, explicit config wins both ways, explicit config cannot grant permission, the rest of the toolbar config survives, and no-entity / no-user / provider-throws all resolve to "no".

Package suite goes from 14 failed files to 13. The 13 remaining are pre-existing module-resolution failures in unbuilt workspace siblings (`ng-pagination`, `ng-list-management`) — verified identical on a clean tree, unrelated to this change.

### DIFF
--- a/.changeset/olive-moons-shake.md
+++ b/.changeset/olive-moons-shake.md
@@ -1,0 +1,23 @@
+---
+"@memberjunction/ng-entity-viewer": patch
+---
+
+Make the grid's Delete affordance reachable
+
+The grid wrapper has always deleted correctly: `onDeleteConfirmed()` walks every
+selected record, reports per-record failures and reloads the page. But the template
+renders the Delete button only when `ToolbarConfig.showDelete && AllowDelete`, both
+of which default to `false`, and the wrapper set neither — so the entire path was
+unreachable. In practice there was no multi-record delete anywhere the grid is used:
+removing ten rows meant opening ten records, one at a time.
+
+`showDelete` now defaults to whether the user may actually delete this entity, and
+`AllowDelete` is bound to the same permission. Defaulting to the permission rather
+than to `true` keeps the button off exactly where the delete would have been refused
+anyway, and an explicit `showDelete` in a view's config still wins in both directions
+— though it can never grant the permission itself, since `AllowDelete` is bound to
+the check, not to the config.
+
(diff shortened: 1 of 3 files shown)


MS-BC:
Closing — not the issue being solved.

PR #4110, opened 2026-08-31
AI Model Research Report — 2026-08-31

AN-BC:
## Summary

Weekly AI model & vendor intelligence report for the week ending 2026-08-31. Full write-up in [`reports/ai-model-research/2026-08-31-weekly-report.md`](../blob/claude/ai-model-research-2026-08-31/reports/ai-model-research/2026-08-31-weekly-report.md).

### Counts
- **2 new models** added (`GLM-5.3-Flash`, `Qwen3.8-Flash`)
- **1 new inference-provider row + cost record** added (Amazon Bedrock on `Grok 4.6`)
- **1 deprecation** (Moonshot AI direct row + cost record on `Kimi K2.5` → `Inactive` for the 2026-08-31 sunset)
- **0 pricing changes** in-inventory this week
- **5 items flagged for human review** (DeepSeek V4 Flash Vision Experimental, OpenAI Daybreak Red/Blue on Bedrock, OpenAI Astra, Azure OpenAI cache-write charges on GPT-5.6 family, plus prior-week open items)

### Applied edits

1. **New model** `GLM-5.3-Flash` — Z.AI (Zhipu AI) cost-tier Flash companion to GLM-5.3, released 2026-08-26 (MIT-licensed open weights, 320B / 18B active MoE, 1.3M context). Z.AI as Model Developer + Inference Provider, plus OpenRouter and Fireworks.ai as Inference Providers. Cost rows for Z.AI direct ($0.15/$0.50 per 1M) and OpenRouter ($0.05/$0.1667 reflecting a 50% Z.AI promo through 2026-09-09; a follow-up cost row will be needed when the promo expires so the historical rate is preserved).
2. **New model** `Qwen3.8-Flash` — Alibaba Cloud cost-tier Flash companion to Qwen 3.8 Max, released 2026-08-26 (125B params, multimodal, 1M context). Alibaba Cloud as Model Developer + Inference Provider, plus OpenRouter and Fireworks.ai as Inference Providers. Cost rows for Alibaba direct and OpenRouter at $0.15/$0.47 per 1M.
3. **New inference provider on `Grok 4.6`** — Amazon Bedrock vendor row (`xai.grok-4-6-v1:0`, Priority 5) and cost record (2026-08-25 start, $2/$6 sub-200K tier at vendor parity with x.ai direct). xAI announced Bedrock availability the week of Aug 25 with US Geo + Global cross-region inference and low/medium/high/xhigh reasoning-effort levels.
[...]

PR #4162, opened 2026-09-01
feat(installer): default MJInstaller to pnpm with an explicit npm override

cadam11:
## Summary
- MJInstaller now defaults to **pnpm** (matching the era-6 platform manifest) with an explicit npm override via `"PackageManager"` in the install config or `MJ_INSTALL_PACKAGE_MANAGER`. A new `PackageManagerCommands` table (`src/models/PackageManager.ts`) replaces the ~12 hardcoded `npm`/`npx` literals across the dependency, codegen, migrate, and smoke-test phases.
- **Preflight hard-fails when the configured package manager is missing** (with `corepack enable pnpm` instructions and the npm-override escape hatch). There is deliberately no silent fallback — two installs of the same version must produce the same tree, which gate-4 certification evidence depends on.
- **npm-layout assumptions made PM-aware** — the real substance of the change:
  - `pnpm-workspace.yaml` is written into bundles that predate the flip, carrying the two settings that are load-bearing under pnpm 10: `linkWorkspacePackages: true` (otherwise exact-version workspace deps resolve from the registry — unpublished `mj_generatedentities` fails loudly, published packages silently install stale) and the `onlyBuiltDependencies` allowlist (otherwise esbuild/sharp/isolated-vm et al. skip their postinstall and fail at runtime).
  - The generated-entities pre-checks (smoke test + doctor) accept `packages/GeneratedEntities`, since pnpm creates no root `node_modules/mj_generatedentities` symlink.
  - Installs pin the root `packageManager` field and rewrite the npm-hoist `mj*` scripts (`node apps/MJAPI/node_modules/.../run.js`) to plain `mj …` — the CLI is a root devDependency, so its bin resolves under both PMs.
  - The `ERESOLVE → --legacy-peer-deps` retry and the `ng-auth-services` root-hoist workaround are npm-only (pnpm's isolated linker has neither problem).
- `distribution.package.json` ships pnpm defaults with PM-neutral scripts; docs and a patch changeset included.

## Reviewer attention
- **Rider: readiness-regex fix.** `API_READY_PATTERNS` (SmokeTestPhase + InstallerEngine) matched only the legacy `/Server ready at/i`, which no current MJAPI prints — the false "MJAPI failed to start" recorded in the 5.51.0 certification scorecard (gate-4 findings). Patterns now match the real `🚀 Ready  <url>` / summary-block output, keeping the legacy marker for older installed tags. Same-file, cert-relevant, but strictly separable from the pnpm work.
- **Behavior rename**: dependency-phase error codes `NPM_INSTALL_FAILED`/`NPM_INSTALL_TIMEOUT` → `INSTALL_FAILED`/`INSTALL_TIMEOUT` (messages name the actual PM). Codes are consumed only inside this package + its tests, verified by grep — flag if you know of an external log parser keying on the old strings.
- **Not verifiable pre-publish**: a registry-true end-to-end `mj install` needs published packages. The real proof is a fresh install off the next edge release; the installer-platform-test workflow runs the unit + smoke tiers on ubuntu/macos/windows at PR time. Gate-4 prep for the 6.1 cert gains one item: pnpm present on the fresh-install rows (preflight now tells the operator exactly that).
- The legacy `--legacy` installer path (`MJCLI/lib/legacy-install.ts`, npm-link era, refuses non-TTY) is intentionally untouched — deleting it is a separate decision.
- No unrelated changes are included; the working tree's `.gitignore` modification predates this branch and was deliberately left out.

## Test plan
- [x] `packages/MJInstaller` unit suite: 29 files, 667/667 pass (includes new specs: pnpm-default command selection, npm override, no-pnpm-retry, workspace scaffold writes, hoist skip, preflight hard-fail, dlx fallback, readiness patterns, layout-aware pre-checks)
- [x] `tsc` clean build; cross-platform smoke script passes locally (real preflight incl. the new package-manager check)
- [ ] installer-platform-test matrix green (ubuntu / macos-14 / windows)
- [ ] After the next edge publish: registry-true `mj install` on a clean machine with pnpm (and once with `"PackageManager": "npm"`), asserting `mj sync push` no-op per the gate-4 procedure


cadam11:
Addressed both review findings in 490ba0fdfd:

1. **Pre-flip bundles no longer shell into npm under a pnpm install.** `patchRootManifest` now also rewrites the exact npm-shim forms of `build` (`npm run build:stream` → direct `turbo --log-order=stream build`), `prebuild` (guard message now says `pnpm install (or npm install)`), and `setup` — customized scripts are matched exactly and left alone. Covered by two new tests (rewrite + leave-customized-alone).

2. **Effective version is now recorded truthfully.** Preflight probes `--version` from the target directory (so `mj doctor` reports the pinned version on existing installs — corepack and pnpm 10's `manage-package-manager-versions` resolve the `packageManager` field by cwd), and the dependencies phase logs `Using pnpm <version> for this install (root packageManager pin: pnpm@10.33.0)` immediately after writing the pin — that line is the gate-4 evidence. A probe failure logs and never fails the install. Three new tests.

672/672 unit tests, clean build, cross-platform smoke passing locally.

(The red Integration check is inherited from `next` — broken since #4110's metadata merge; unrelated to this PR, re-run once next is green.)

PR #4166, opened 2026-09-02
feat(predictive-studio): the `sequence` problem type — HMM, end to end

MS-BC:
Stacks on #4104. Turns `sequence` from a type the system could not hold into one it can train.

## What this does

The HMM driver already existed in the sidecar but nothing could ask for it: the database CHECK rejected `'sequence'`, the generated entity's zod union was two values, and the component types were seeded `Draft`. This closes that, in the only order that works.

**Migration → CodeGen → widen → publish.** Widening the TypeScript first would compile against generated entities that cannot hold the value and write rows the database refuses.

- CHECK constraints widened on `MLModel`, `MLTrainingPipeline` and `MLAlgorithmUseCase` (SS + PG twins, applied and verified).
- CodeGen re-run. It **skipped rewriting `entities/__mj.ts` on a no-op mtime check** despite 387 entities — the union did not widen until the file was deleted and CodeGen re-run. Worth knowing about.
- Seven touch points widened, not the six planned. CodeGen's own build step found the extra one: `ModelProblemType` in the model-predictions form panel.
- `Sequence` and `Hidden Markov Model` leave Draft. The seed tree now has **no Draft family at all**.

## The part worth reviewing

A sequence run gets its own training function rather than a branch inside the supervised one, because three of that path's assumptions are wrong for it:

| | supervised path | sequence path |
|---|---|---|
| target | fitted against | rides along, unsupervised |
| holdout | split by row | split by **group** |
| metrics | AUC / R² | mean posterior confidence |

The holdout one is the correctness point: a row-wise split puts part of a member's history in train and part in holdout, leaking their future into their past and flattering the score. Whole entities are held out — and never all of them.

`sequence.group_field` is required, and refused when missing or not a real column. Without boundaries hmmlearn treats the whole matrix as one sequence, learns transitions between unrelated members, and returns a fitted model with confident scores that nothing downstream would question.

Feature importance is reported empty rather than invented: an HMM attributes to latent states, not to input features.

## Verification

Core 142/142 · Engine 641/641 · Sidecar TS 15/15 · Sidecar Python 117/117 · dashboards 1850/1850. Migration applied to a live SQL Server and the widened constraints read back; PG twin content-parity clean.

## Notes

- The generated-form commit includes section reordering in **nine unrelated entity forms** with no net change — the LLM form-layout pass, not a consequence of this migration. Left in rather than reverted; say the word and I will drop those nine.
- `CodeGen_Run_*.sql` is deliberately not committed: no Flyway version prefix, so it would never run, and CodeGen derives those rows from the CHECK anyway.


AN-BC:
Heads-up on adjacent work, not a review.

Amith asked me to flag that a **Feature Pipelines** build is starting (Knowledge Hub feature pipelines on the Record Set Processing substrate, with Predictive Studio consuming them). It will touch four files this PR also edits:

- `packages/AI/PredictiveStudio/Core/src/feature-steps.ts` — `LLMDerivedFeatureStep` gains explicit outputs and a pinned prompt version
- `packages/AI/PredictiveStudio/Core/src/modeling-plan-spec.ts` and `modeling-plan-schema.ts` — a proposed-pipeline slice behind the existing approval gate
- `packages/AI/PredictiveStudio/Engine/src/agent/modeling-plan-to-pipeline.ts` — `llm-derived` candidates stop being dropped silently

Nothing in that work supersedes the sequence/HMM work here; the overlap is only those files. Whichever lands first, the other rebases. The plan is being finalized as `plans/feature-pipelines-build-plan.md` on branch `claude/feature-pipelines-build-plan-xrcbll`; I'll link the PR here once it's open.

---

AN-BC:
Following up on my note above — the plan is now open as **#4636** (plans-only, no code).

The four overlapping files are unchanged by it; #4636 only describes what will later happen to them. Nothing here is superseded, and this PR should not be closed for it.

---

PR #4168, opened 2026-09-02
fix(conversations,code-editor): restore the agent-completion toast and template syntax highlighting

cmbrock-BC:
## Summary

Two independent defects in the conversation UI, both found while tracing why the browser console prints a warning on every single agent turn. Neither is a big change — 11 lines of code between them — but each disables a feature that currently looks like it works.

### The agent-completion notification has never fired

`MessageInputComponent.markMessageComplete()` looks up the active task by the AI reply's conversation detail ID:

```ts
const task = this.activeTasks.getByConversationDetailId(conversationDetail.ID);
```

But every `activeTasks.add()` call in that component registers the task against the message that **started** the turn — the user message, the Sage delegation message, or the status message — never the reply. Those IDs can't be equal, so the lookup misses on every completion, and the block it guards never runs:

```ts
MJNotificationService.Instance?.CreateSimpleNotification(
  `${task.agentName} completed in ${task.conversationName || 'conversation'}`, ...
```

That notification exists for the user who asks something and navigates away, which is exactly the case it never covers.

It didn't leak tasks, which is why this went unnoticed: `ConversationChatAreaComponent` separately sweeps tasks whose message has left `In-Progress`. So the only symptoms were the missing toast and `⚠️ No task found for completed message …` on every turn — a warning whose text ("task may have been removed prematurely or not added") sent me looking for a lifecycle race that isn't there.

Every path already links the reply to its starter through `ParentID`, so the lookup now falls back to it.

### Template editors have no syntax highlighting at all

`CodeEditorComponent._findLanguage()` matches only `name` and `alias`. Several names in common use are registered by CodeMirror as **extensions** instead — `jinja2` is one of Jinja's (`["j2", "jinja", "jinja2"]`) while its only matchable name is `"Jinja"`.

`template-editor.component` and `templates-form.component` both ask for `jinja2` (six call sites), so every prompt and template opened with `Language not found: jinja2` and fell back to no highlighting — on content that is Nunjucks, where `{{ }}` and `{% %}` are most of what you're reading.

Fixed in the editor rather than the six call sites, so it holds for any caller. Deliberately a **fallback** rather than part of the main loop: it runs only when nothing matched by name or alias, so it can turn a miss into a hit but can never change a match that already resolves. That matters because short extensions (`r`, `md`, `ts`) could otherwise outrank another language's real name.

## Testing

- `turbo build --filter=@memberjunction/ng-conversations --filter=@memberjunction/ng-code-editor` — 68/68 tasks green, including the full dependency chain.
- `MJCore` 2162 tests and `MJGlobal` 874 tests pass unchanged.
- Both were found and confirmed against a running app on 6.1.0-edge.1: the warning fires once per turn with the completion toast absent, and `Language not found: jinja2` appears on opening any AI Prompt form.

Neither change has a runtime dependency on the other; they're together because they came out of the same investigation and are each a few lines.

### DIFF
--- a/.changeset/fix-completion-toast-and-editor-language.md
+++ b/.changeset/fix-completion-toast-and-editor-language.md
@@ -1,0 +1,10 @@
+---
+"@memberjunction/ng-conversations": patch
+"@memberjunction/ng-code-editor": patch
+---
+
+Two fixes for defects found while tracing console noise in the conversation UI.
+
+**The agent-completion toast never fired.** `markMessageComplete` looked up the active task by the AI reply's conversation detail ID, but every `activeTasks.add()` call in `message-input` registers the task against the message that STARTED the turn — the user message, the Sage delegation message, or the status message. Those IDs are never equal, so the lookup missed on every completion and the "&lt;agent&gt; completed in &lt;conversation&gt;" notification never reached a user who had navigated away. The task still disappeared, because `conversation-chat-area` separately sweeps tasks whose message has left `In-Progress`, so this surfaced only as a missing notification plus `⚠️ No task found for completed message …` on every single turn. Every path already links the reply to its starter through `ParentID`, so the lookup now falls back to it.
+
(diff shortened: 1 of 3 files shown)


cmbrock-BC:
Closing this — the conversations half is wrong, and I'd rather withdraw it than have someone spend review time on it.

**The claim was that the agent-completion toast never fires.** It does. `updateConversationDetail(userMessage, …, 'Complete')` is called on every flow (15 call sites) and calls `markMessageComplete(userMessage)` — whose ID is exactly what the task is registered under. The direct lookup already matches.

What actually happens is that `markMessageComplete` is called **twice** per turn, once with the AI reply and once with the user message. Only the second is supposed to match; the first logs `⚠️ No task found for completed message …`. I read that warning as evidence of a broken feature rather than of a call that was never going to match.

Worse, the `ParentID` fallback wouldn't even remove the warning — it would make the AI-reply call resolve the task first, so the user-message call would then find nothing and warn instead. Net effect on the toast: nil.

The real defect is that the warning is normal-path noise and shouldn't be a warning at all. That's handled in #4169 alongside its siblings.

The `ng-code-editor` half was independently verified and is unaffected, so it continues as #4170.

PR #4169, opened 2026-09-02
fix(conversations): make diagnostic logging opt-in instead of unconditional

cmbrock-BC:
> Rebased onto `next` and no longer stacked. It previously sat on #4168, which I withdrew — that PR claimed the agent-completion toast never fires, and it does; see the closing comment there.

## Summary

Reported from a deployed app:

> the browser's dev console is absolutely packed with spam. Every single character a user types into the input box triggers a message in the console. This isn't something we can put out into the world since some users will bring that up and it looks terrible.

**Nothing is removed.** Twelve `console.*` calls become `LogStatusEx({ …, verboseOnly: true })` — silent by default, still there for anyone tracing this code. In a browser, verbose comes from `window.MJ_VERBOSE = true`, `localStorage.setItem('MJ_VERBOSE','true')`, or a `?MJ_VERBOSE=true` URL parameter, all of which `IsVerboseLoggingEnabled` already supports.

5 files, +53 −18. `console.error` is untouched — all 179 in the package.

### The per-keystroke pair — the specific behaviour reported

`ComposerDraftStore.SetDraft` logs on **every character typed**, and `conversation-chat-area` logs the same event again on the way in. Together they are the largest single source of console output in the app.

### Task-lifecycle narration

`ActiveTasksService.add()` and `.remove()` each printed three lines per call:

```ts
console.log(`➕ Task added:`, {id, conversationId, agentName});
console.log(`📊 Total tasks:`, this._tasks$.value.size);
console.log(`🗂️ Conversation IDs with tasks:`, Array.from(...));
```

and `markMessageComplete` printed two more describing its own happy path.

### One `console.warn` that was misdiagnosing itself

`⚠️ No task found for completed message … - task may have been removed prematurely or not added` fired **on every turn, by construction.**

A turn registers exactly one task, against whichever message its flow chose:

| `activeTasks.add()` | keyed on |
|---|---|
| line 1699 | `conversationManagerMessage.ID` |
| line 2107 | `agentResponseMessage.ID` |
| line 2325 | `statusMessage.ID` |
| lines 2411, 2639 | `userMessage.ID` |

Meanwhile `markMessageComplete` runs for *every* message in the turn that reaches `Complete` or `Error` — including the user message, via fifteen `updateConversationDetail(userMessage, …, \'Complete\')` call sites. So most calls find no task, and that is the normal case rather than a lifecycle race.

It is now verbose-only and reworded to say what it means: *"No task registered against completed message … — expected for any message that did not start the turn."*

## Scope, and what I deliberately did not do

The obvious larger change is routing this package's remaining `console.*` through `LogStatus`. I measured before proposing it, and the numbers argue against doing it here:

| | raw `console.*` | `LogStatus` |
|---|---|---|
| all MJ Angular packages | **863** | 105 |

Several use `LogStatus` zero times (`ai-test-harness` 65:0, `react` 53:0, `file-storage` 48:0), and there is no guidance in `CLAUDE.md` either way. So raw `console` is the prevailing style, and `ng-conversations` is not an outlier — it is the largest surface, which is why it is the loudest.

Converting 216 calls in this one package would make it the only Angular package following a different rule. That is a convention decision for a maintainer, not something to land as a bug fix. Happy to do it — it is mechanical — but it should be a deliberate call across the Angular layer.

## Testing

`turbo build --filter=@memberjunction/ng-conversations` — 68/68 tasks green. `npm run check:changeset` and `packages/Standards/bin/run.js check` both pass. Verified every original log message still exists in the source and that `console.error` counts are unchanged in every touched file.

### DIFF
--- a/.changeset/conversations-logging-opt-in.md
+++ b/.changeset/conversations-logging-opt-in.md
@@ -6,7 +6,7 @@ Make the conversation UI's diagnostic logging opt-in instead of unconditional.
 
 Reported from a deployed app: "the browser's dev console is absolutely packed with spam. Every single character a user types into the input box triggers a message in the console. This isn't something we can put out into the world since some users will bring that up and it looks terrible."
 
-Nothing is removed — twelve `console.*` calls become `LogStatusEx({ …, verboseOnly: true })`, so they are silent by default and still there for anyone tracing this code. In a browser, verbose is enabled by `window.MJ_VERBOSE = true`, `localStorage.setItem('MJ_VERBOSE','true')`, or a `?mj_verbose=true` URL parameter.
+Nothing is removed — twelve `console.*` calls become `LogStatusEx({ …, verboseOnly: true })`, so they are silent by default and still there for anyone tracing this code. In a browser, verbose is enabled by `window.MJ_VERBOSE = true`, `localStorage.setItem('MJ_VERBOSE','true')`, or a `?MJ_VERBOSE=true` URL parameter.
 
 **The per-keystroke pair**, which is the specific behaviour in the report: `ComposerDraftStore.SetDraft` logs on every character typed, and `conversation-chat-area` logs the same event again on the way in. Together they are the largest single source of console output in the app.
 
--- a/.changeset/conversations-logging-opt-in.md
+++ b/.changeset/conversations-logging-opt-in.md
@@ -1,0 +1,19 @@
+---
+"@memberjunction/ng-conversations": patch
+---
+
+Make the conversation UI's diagnostic logging opt-in instead of unconditional.
+
+Reported from a deployed app: "the browser's dev console is absolutely packed with spam. Every single character a user types into the input box triggers a message in the console. This isn't something we can put out into the world since some users will bring that up and it looks terrible."
+
+Nothing is removed — twelve `console.*` calls become `LogStatusEx({ …, verboseOnly: true })`, so they are silent by default and still there for anyone tracing this code. In a browser, verbose is enabled by `window.MJ_VERBOSE = true`, `localStorage.setItem('MJ_VERBOSE','true')`, or a `?mj_verbose=true` URL parameter.
+
+**The per-keystroke pair**, which is the specific behaviour in the report: `ComposerDraftStore.SetDraft` logs on every character typed, and `conversation-chat-area` logs the same event again on the way in. Together they are the largest single source of console output in the app.
+
+**Task-lifecycle narration**: `ActiveTasksService.add()` and `.remove()` each printed three lines per call (`➕ Task added:`, `📊 Total tasks:`, `🗂️ Conversation IDs with tasks:`), and `markMessageComplete` printed two more describing its own happy path.
+
+**One `console.warn` that was misdiagnosing itself.** `⚠️ No task found for completed message … - task may have been removed prematurely or not added` fired on every turn, by construction. A turn registers exactly one task, against whichever message its flow chose — `activeTasks.add()` is called with the user message, a Sage delegation message, a status message, or the agent response depending on the path — while `markMessageComplete` runs for *every* message in the turn that reaches `Complete` or `Error`. Most calls therefore find no task, which is the normal case and not the lifecycle race the text described. It is now verbose-only and reworded to say what it actually means.
+
+`console.error` is untouched — all 179 in the package.
+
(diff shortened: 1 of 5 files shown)


rkihm-BC:
## Code review

Reviewed the diff, the PR body, the changeset and the surrounding code. I applied the patch locally and ran the package's tests: **1230/1230 pass**. All CI green.

### Verified independently

- `activeTasks.add()` call sites at 1699 / 2107 / 2325 / 2411 / 2639 — all five exist as the table describes.
- `console.error` count unchanged (179).
- The composer package and `mention-autocomplete.service.ts` have **zero** `console.log` between them, so the two calls gated here really were the entire per-keystroke path. The reported bug is fully closed.

### What's good

The `console.warn` analysis is the strongest part of this PR and it holds up. One `add()` per turn against whichever message the flow chose, versus `markMessageComplete` running for *every* message reaching `Complete`/`Error`, means the warning was structurally guaranteed rather than reporting a race. Downgrading it **and rewording it** is the right call — keeping the old text at verbose level would have preserved a claim that was never true. The inline comment explains why, which is what a future reader needs.

Scope discipline is good too: the deferral of the package's remaining `console.*` calls is argued with measurements (863 raw vs 105 `LogStatus` across the Angular layer) rather than asserted, and it's the right instinct — that's a convention decision for the Angular layer, not something to settle inside a bug fix. Worth someone actually answering at some point, but not a condition of this merge.

### One thing to fix before merge

**The changeset documents a URL parameter that doesn't work.** It says:

> or a `?mj_verbose=true` URL parameter

`IsVerboseLoggingEnabled` reads `params.get('MJ_VERBOSE')` (`packages/MJCore/src/generic/logging.ts:256`), and `URLSearchParams.get` is case-sensitive — so `?mj_verbose=true` is a no-op. `window.MJ_VERBOSE = true` and `localStorage.setItem('MJ_VERBOSE','true')` both work exactly as documented.

This only matters because a changeset becomes a published release note, so it would ship an incantation that fails. One-word fix in the changeset and the PR body: `?MJ_VERBOSE=true`.

Making `IsVerboseLoggingEnabled` case-insensitive is arguably the better long-term fix — every doc written against it will make this same mistake — but that's `@memberjunction/core` with its own blast radius and changeset. Follow-up, not this PR.

### Considered and dismissed — no action needed

- **Remaining unconditional logs.** I initially flagged `[StreamingCallback] Updated message …` (message-input:919) as high-frequency. It isn't: the streamed-delta path returns at the `if (progress.streaming)` guard above it, so that line fires once per agent *step*, not per token. The `[client:capture-data-snapshot] tick=` logs are already throttled to every fifth tick and only run on an explicit snapshot command. Neither is ambient noise.
- **Per-keystroke cost of the new call.** A template literal plus `IsVerboseLoggingEnabled()` (which does a `localStorage.getItem` and builds a `URLSearchParams`) now runs twice per character. That's still cheaper than the `console.log` it replaced, particularly with devtools open — net improvement, not a regression.
- **A test.** A `vi.spyOn(console, 'log')` in the existing `composer-draft-store.test.ts` would pin the fix, but it would be asserting on `LogStatusEx`'s internals rather than on behaviour anything depends on. Not worth the brittleness.

### Verdict

Approve once the changeset's `mj_verbose` → `MJ_VERBOSE`. Nothing else blocking.

cmbrock-BC:
Fixed — `?mj_verbose=true` → `?MJ_VERBOSE=true` in the changeset, the commit message and the PR body. You were right, and I verified it rather than taking it on trust: `logging.ts:256` reads `params.get('MJ_VERBOSE')` and `URLSearchParams.get` is case-sensitive, so what I documented was a no-op. The other two routes were correct as written.

Thanks also for the three you considered and dismissed — the `[StreamingCallback]` one in particular. I had it on my own list as a suspected per-token log and hadn't traced the `if (progress.streaming)` guard above it; per agent *step* is a different thing entirely and it doesn't belong in this PR.

One note on the push: I had amended locally before you merged `next` in, and `--force-with-lease` refused, which is the only reason your merge survived. I rebased onto it and pushed forward instead — `699f34d5` is your merge plus the one-word changeset fix, no TypeScript touched.

Agreed on `IsVerboseLoggingEnabled` being case-insensitive as the better long-term fix. Happy to raise that against `@memberjunction/core` separately if you want it.

PR #4170, opened 2026-09-02
fix(code-editor): fall back to file extensions when resolving a language by name

cmbrock-BC:
## Summary

`CodeEditorComponent._findLanguage()` matches only `name` and `alias`:

```ts
for (const lang of this.languages) {
  for (const alias of [lang.name, ...lang.alias]) {
    if (name.toLowerCase() === alias.toLowerCase()) return lang;
  }
}
console.error('Language not found:', name);
```

Several names in common use are registered by CodeMirror as **extensions** instead. `jinja2` is one of Jinja's — the entry is `name: "Jinja", extensions: ["j2", "jinja", "jinja2"]`, with no alias — so `"Jinja"` is the only string that matches.

`template-editor.component` and `templates-form.component` both ask for `jinja2`, at six call sites between them. The result is that **every prompt and template in the Explorer forms opens with no syntax highlighting at all**, and a `Language not found: jinja2` error in the console. That content is Nunjucks, where `{{ }}` and `{% %}` are most of what you're reading.

## The change

An extensions pass, after the existing name/alias loop:

```ts
for (const lang of this.languages) {
  if (lang.extensions.some((ext) => wanted === ext.toLowerCase())) return lang;
}
```

Fixed in the editor rather than at the six call sites, so it holds for any caller asking by a name CodeMirror happens to file under extensions.

Deliberately a **fallback** and not part of the main loop. It runs only when nothing matched by name or alias, so it can turn a miss into a hit but can never change a match that already resolves. That ordering matters: short extensions like `r`, `md` and `ts` would otherwise be able to outrank another language's real name.

## Blast radius

`jinja2` is the **only** language name MJ requests whose resolution changes. Checked every value handed to `mj-code-editor` against `@codemirror/language-data`:

| requested | resolves today | after this change |
|---|---|---|
| `jinja2` | **nothing** | Jinja, via extension |
| `html`, `javascript`, `typescript`, `sql`, `python`, `json`, `markdown`, `css`, `yaml` | name/alias | unchanged — the first loop still wins |

So for everything already working this is a no-op, and the fallback is reached only where the editor currently gives up.

## Testing

`turbo build --filter=@memberjunction/ng-code-editor` — 12/12 tasks green.

Verified against `@codemirror/language-data` that `jinja2` appears in `extensions` and not in `name` or `alias`, which is why the current lookup can't reach it.

### DIFF
--- a/.changeset/fix-code-editor-language-by-extension.md
+++ b/.changeset/fix-code-editor-language-by-extension.md
@@ -7,3 +7,5 @@ Resolve a language by its file extension when the name does not match, restoring
 `_findLanguage` matched only `name` and `alias`. Several names in common use are registered by CodeMirror as extensions instead — `jinja2` is one of Jinja's (`["j2", "jinja", "jinja2"]`) while its only matchable name is `"Jinja"`. Core-entity-forms' template editors ask for `jinja2` at six call sites, so every prompt and template opened with `Language not found: jinja2` and no highlighting at all — on content that is Nunjucks, where `{{ }}` and `{% %}` are most of what you are reading.
 
 Deliberately a fallback rather than part of the main loop: it runs only when nothing matched by name or alias, so it can turn a miss into a hit but can never change a match that already resolved. That matters because short extensions (`r`, `md`, `ts`) could otherwise outrank another language's real name.
+
+Covered by `ng-code-editor.language.dom.test.ts`, which pins that ordering (`r` resolves to R and `html` to HTML, both of which also appear among other languages' extensions) alongside the `jinja2` case itself.
--- a/.changeset/fix-code-editor-language-by-extension.md
+++ b/.changeset/fix-code-editor-language-by-extension.md
@@ -1,0 +1,9 @@
+---
+"@memberjunction/ng-code-editor": patch
+---
+
+Resolve a language by its file extension when the name does not match, restoring syntax highlighting for template editors.
+
+`_findLanguage` matched only `name` and `alias`. Several names in common use are registered by CodeMirror as extensions instead — `jinja2` is one of Jinja's (`["j2", "jinja", "jinja2"]`) while its only matchable name is `"Jinja"`. Core-entity-forms' template editors ask for `jinja2` at six call sites, so every prompt and template opened with `Language not found: jinja2` and no highlighting at all — on content that is Nunjucks, where `{{ }}` and `{% %}` are most of what you are reading.
+
+Deliberately a fallback rather than part of the main loop: it runs only when nothing matched by name or alias, so it can turn a miss into a hit but can never change a match that already resolved. That matters because short extensions (`r`, `md`, `ts`) could otherwise outrank another language's real name.
--- a/packages/Angular/Generic/code-editor/src/lib/ng-code-editor.component.ts
+++ b/packages/Angular/Generic/code-editor/src/lib/ng-code-editor.component.ts
@@ -716,13 +716,30 @@ export class CodeEditorComponent extends BaseAngularComponent implements OnInit,
 
   /** Find the language's extension by its name. Case insensitive. */
   private _findLanguage(name: string) {
+    const wanted = name.toLowerCase();
     for (const lang of this.languages) {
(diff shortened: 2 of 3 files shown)


rkihm-BC:
## Code review

Reviewed the diff, the PR body and the surrounding code. I applied the patch locally and ran the package's tests: **29/29 pass**, and the new spec is correctly collected under the `code-editor (dom)` project. All CI green.

### Verified independently

- `jinja2` is in Jinja's `extensions` (`["j2","jinja","jinja2"]`) with no `alias`, so `"Jinja"` really was the only matchable string. Root cause is exactly as described.
- Six `jinja2` call sites — three each in `template-editor.component.ts:452` and `templates-form.component.ts:508`.
- `@codemirror/lang-jinja` is a declared dependency of `@codemirror/language-data@6.5.2`, so the lazy `load()` resolves. No new dependency needed here, and no bundle change: the bundler already walks language-data's `load()` functions.

### What's good

Fixed at the editor rather than the six call sites — right layer for a lookup bug. And the fallback-versus-inline-loop decision isn't just asserted in the comment, it's **pinned by tests**: `find('r') === 'R'` and `find('html') === 'HTML'` are precisely the two cases that break if someone later folds the extension pass into the main loop. That's the difference between a comment that rots and an invariant that holds.

Hoisting `wanted` out of the loop is a free cleanup that doesn't change semantics.

### One thing worth knowing, not worth changing

The blast-radius section says `jinja2` is the only language MJ requests whose resolution changes. That's **true for today's call sites** — I traced every dynamic binding (`getEditorLanguage`, `detectLanguageForKey`, `getContentLanguage`, `codeLanguage`, `tabData.language`) and they're all bounded to a safe set.

The behaviour change itself is broader than the table implies, though: **147 of 223 extension strings that previously resolved to nothing now resolve.** Most are strict upgrades (`rs` → Rust, `md` → Markdown, `pl` → Perl). A couple are surprising:

| requested | before | after |
|---|---|---|
| `text` | — | **LaTeX** |
| `h` | — | C |
| `m` | — | Mathematica |

Nothing is broken today because no caller passes those. It matters if someone later wires markdown fence tags into `[language]`, where ` ```text ` would render as LaTeX. Since the fallback can only ever turn "no highlighting" into "some highlighting", the failure mode is cosmetic — I'd note the hazard in the comment and leave the code alone.

### Separate bug the same audit would find

`packages/Angular/Explorer/dashboards/src/DevTools/graphql-console.component.html:309` asks for `'graphql'`, and GraphQL isn't in `@codemirror/language-data` at all — not as a name, alias or extension. So that editor logs `Language not found: graphql` and renders unhighlighted both before and after this change. Out of scope here; worth a follow-up.

### Considered and dismissed — no action needed

- **No `.catch` on `langDesc?.load()`.** I nearly asked for one, then checked the built-in fast path (`json`, `javascript`, `js`, `typescript`, `ts`, `sql`, `python`, `py`): `html`, `markdown`, `css` and `xml` already route through `_findLanguage` → `load()` with no catch today. This PR adds a fifth language to an existing pattern rather than introducing the gap. Genuinely pre-existing, genuinely rare.
- **`as unknown as` in the test** to reach a private method. It's documented and justified by the fact that `_findLanguage` is pure over `this.languages`, which is what lets the spec skip constructing an `EditorView`. Fine as is.

### Verdict

Approve as-is.

rkihm-BC (approved):
Approving — full review in [this comment](https://github.com/MemberJunction/MJ/pull/4170#issuecomment-5517396573).

Verified locally: patch applied, `pnpm test` 29/29 with the new spec collected under the `code-editor (dom)` project; `jinja2` confirmed present in Jinja's `extensions` and absent from `name`/`alias`; `@codemirror/lang-jinja` resolves via language-data's own dependency, so the lazy `load()` works and the bundle is unchanged.

No changes requested. Two notes carried in the comment for the record, neither blocking: the extension fallback newly resolves 147 of 223 extension strings (`text` → LaTeX being the one that could surprise a future caller, though nothing today passes it), and `graphql-console.component.html:309` requests a language `@codemirror/language-data` doesn't ship at all — a separate pre-existing bug worth a follow-up.

Issue #4082, opened 2026-08-27
Explore native tool calling in BaseLLM and the agent framework


jordanfanapour:
## Summary

MJ's LLM providers do not use the native tool/function-calling APIs that every major vendor now exposes. Agent actions are described as text inside the system prompt, and the model is asked to reply with a JSON `LoopAgentResponse` envelope that the framework parses and dispatches.

That design is portable and works on models with no tool support at all, which is a real strength. But newer, more agentically-trained models are beginning to fight it: they recognise the action contract in the prompt, attempt a native tool call, and the request declares no tools — so the provider discards the response.

This issue proposes exploring native tool calling as an **opt-in capability**, not a replacement.

## Current state

- No chat driver passes `tools` / `functionDeclarations` / `tool_choice`. Checked: OpenAI, Anthropic, Gemini, Groq, Cerebras, Mistral, xAI, OpenRouter, Vertex. (The one hit in the Gemini package is `geminiRealtime.ts`, a separate path.)
- `packages/AI/Core/src/generic/baseLLM.ts` on `next` has no tool surface — the only `tool` references are artifact tools.
- Actions reach the model as prose in the system prompt; the framework parses the returned JSON envelope.

## Evidence that this is becoming a problem

Measured against a real Loop-agent payload (an agent that iterates writing and testing SQL), replayed directly against the provider SDK outside any MJ code:

| condition | outcome |
|---|---|
| `gemini-3.7-flash`, no tools declared (what MJ sends) | **~35%** of responses returned `finishReason: MALFORMED_FUNCTION_CALL` with the generated tokens discarded |
| `gemini-3.7-flash`, one matching tool declared | **0%** malformed; the model emitted a well-formed function call **7/8** times |
| `gemini-3-flash`, `gemini-3.5-flash`, `gemini-3.6-flash` | 0/6 malformed each, same payload |

The model knows exactly which action it wants and how to call it. With no tools declared, that intent surfaces as an unparseable call and the whole response is dropped, which the driver correctly reports as "No output received from model".

Only the newest model in the family does this — so this reads as a trend rather than a one-off.

Two secondary observations:

- Loop system prompts in this workload render at **56–104KB**, much of it contract description that tool definitions would carry more efficiently (and that most providers cache better).
[...]

PR #4176, opened 2026-09-03
Native tool calling: BaseLLM provider surface, metadata gate, and prompt-eval harness (shipped off)

jordanfanapour:
## Summary

Adds a provider-neutral native tool-calling surface to `BaseLLM`, implements it in the Anthropic, OpenAI and Gemini drivers, gives the prompt stack its own configuration bag, and gates the whole thing behind metadata in the prompt runner.

Supersedes the planning-only scope this PR opened with (#4082). The design documents are still here as the branch's first commit; everything below is the implementation.

**Nothing in the tool-calling path changes behavior on merge.** MJ's LLM providers have described agent actions as prose in the system prompt and parsed a JSON envelope back. Newer agentically-trained models increasingly fight that, and the prose action catalog is most of a 56–104KB Loop system prompt. This lands the plumbing end to end and switched off:

- the catalog records which models *can* do native tool calling — 139 model rows carry `LLM.SupportsNativeToolCalling: true`, with 83 vendor-serving rows overriding to `false` where a serving doesn't support it;
- all 139 carry `LLM.DefaultToNativeToolCalling: false`;
- no prompt sets `LLM.UseNativeToolCalling`;
- no MJ caller passes tools.

So the gate resolves false on every existing path and each run takes exactly the code it takes today. Capability is a statement of fact and is safe to ship; policy is the switch, and it is off. The first behavior change will be a deliberate one: flipping policy on a specific model, with a caller that supplies tools.

**What does change on merge.** Three things in this branch are not behind the gate. **Model selection** — the 8 prompts described under *Model selection* below move to Gemini 3.8 Flash, with the Flash-Lite rows that outranked them set `Status: "Inactive"`. **Prompt exemplars** — the audit below fixes exemplars that previously forced a retry, and one prompt whose `OutputExample` did not parse under `ValidationBehavior: 'Strict'`. **Two toolchain repairs unrelated to the feature** — `mj dev workspace` (`packages/MJCLI/src/lib/dev-workspace/**`) and the OpenApp install path (`packages/OpenApp/Engine/src/install/**`); the most easily missed part is that the generated workspace moves `strict-peer-dependencies` / `auto-install-peers` out of `.npmrc` into `pnpm-workspace.yaml` and relaxes strict peers to `false`, so a mixed workspace resolving published `@mj-biz-apps/*` against linked MJ source completes and reports its peer gaps instead of failing on peers no member can fix. Everything in the tool-calling path itself is gated, including the Loop system-prompt template, whose new blocks all sit behind `_NATIVE_TOOL_CALLING` / `_NATIVE_CONTROL_FLOW` and render identically with the gate closed.

## The gate

`ResolveNativeToolCalling` (`packages/AI/Prompts/src/nativeToolCallingGate.ts`) resolves three inputs:

| input | source | behavior |
|---|---|---|
| **capability** | `LLM.SupportsNativeToolCalling` | hard gate — false means native is impossible, regardless of anything else |
| **policy** | `LLM.DefaultToNativeToolCalling`, catalog layers | soft |
| **preference** | `LLM.UseNativeToolCalling`, prompt layers | soft, and beats policy |

Absent resolves false. The configuration bag reuses the existing `ModelConfiguration` JSONType cascade rather than adding per-flag columns.

## Provider surface

- `ChatParams` gains `tools` (declarations as JSON Schema — the one format all three vendors accept), `toolChoice` (`'auto' | 'none' | 'required' | { name }`) and `parallelToolCalls`.
- Responses normalize to `ChatCompletionMessage.toolCalls` with **parsed** arguments, plus a `'tool_calls'` finish reason. A turn can carry text *and* calls, because all three providers structurally allow it — nothing may assume text exists on a tool-call turn.
- Multi-turn tool use round-trips through a new `tool` message role and `tool_result` content block, reusing the existing content-block serialization so tool turns persist correctly in message logs.
- `BaseLLM.SupportsTools` declares whether a driver has the mapping (default `false`). A driver without it ignores declarations rather than failing.

## Control flow and declaration control

- **Hybrid** — Actions are declared as tools; the envelope still carries control flow.
- **Implicit** — `delegate_to_<agent>`, `payload_change_request` and `ask_user` become tools too, and plain text means the task is complete.
- **Native tool results** — action results return as `tool` turns rather than as envelope text.
- `AIAgent.DeclareActionsAsNativeTools` and `AIAgentAction.DeclareAsNativeTool` control whether actions are declared to the model at all.

## Schema

One consolidated migration (`V202609122036__v6.1.x__Native_Tool_Calling.sql`) adds 9 columns across 6 entities, with the CodeGen fold appended:

| entity | columns |
|---|---|
| `AIPrompt`, `AIPromptModel` | `PromptConfiguration NVARCHAR(MAX) NULL` |
| `AIPromptRun` | `ToolCallingMode NVARCHAR(25) NULL` + CHECK |
| `AIAgentRunStep` | `ToolCallingMode` + CHECK, `NativeToolCallCount INT NULL` + CHECK ≥ 0, `NativeDualChannel BIT NULL`, `NativeToolResultsSent BIT NULL` |
| `AIAgent` | `DeclareActionsAsNativeTools BIT NOT NULL DEFAULT (1)` |
| `AIAgentAction` | `DeclareAsNativeTool BIT NOT NULL DEFAULT (1)` |

`ToolCallingMode` is `Envelope | Native | NativeImplicit | NativeFallback`, enforced by CHECK constraints, from which CodeGen derives the `EntityFieldValue` rows and the TypeScript union.

The CodeGen fold was regenerated from a from-scratch database at `next`'s schema, so the cascade procedures retain `next`'s AI Persona and Skill-Action work rather than overwriting it.

## Eval harness

A new **Prompt Eval** test type and `PromptEvalDriver` evaluate a single model decision from frozen mid-loop state. No action ever executes; the runner returns the model reply and deterministic oracles read the decision off whichever channel it used, envelope or native tool calls. That makes envelope and native directly comparable on identical inputs.

Ships with the prompt-eval corpus (893 files) and the rigs that pin matrices, score runs, and toggle catalog posture between testing and shipping.

## Model selection

Query Builder and the Research Agent family (7 agents, 8 prompts) move to Gemini 3.8 Flash. The Flash-Lite rows that outranked them are set `Status: "Inactive"` rather than deleted, because `mj sync push` never deletes — removing the rows from JSON would have left those agents on Flash-Lite in every existing database.

## Decisions from the original plan

1. **`PromptConfiguration` JSON vs. `bit NULL` columns** — resolved in favor of the JSON bag, matching the `ModelConfiguration` precedent. Two columns instead of one per flag.
2. **`ActionParam.ValueSchema` for typed tool parameters** — *not* in this PR. The permissive prose-parity mapping ships first; the typed-schema follow-on that unlocks enums, nested shapes and strict mode is still open.

## Verification

- 6,894 tests pass across 17 packages.
- The migration applies cleanly from scratch, twice, alongside the full migration set.
- `check:codegen-tail` passes: committed generated code matches committed migrations.
- `mj sync validate --dir=metadata` passes on 352 files with 0 errors.

## Review notes

The diff is large but lopsided — 893 of the files are the prompt-eval corpus, and the migration plus CodeGen artifacts account for most of the remaining volume. The hand-written surface worth reading is `packages/AI/Core`, `packages/AI/Prompts`, `packages/AI/Agents` and the three providers.

Relates to #4082

### DIFF
--- a/plans/native-tool-calling/2026-09-02-native-tool-calling-test-plan.md
+++ b/plans/native-tool-calling/2026-09-02-native-tool-calling-test-plan.md
@@ -1,0 +1,491 @@
+# Native Tool Calling — Test Plan
+
+**Issue:** [MemberJunction/MJ#4082](https://github.com/MemberJunction/MJ/issues/4082)
+**Status:** Draft
+**Companion doc:** [Implementation plan](./2026-09-02-native-tool-calling.md) — this document
+expands §10 of that plan and is the deliverable behind its PRs 4 and 6 (eval gate).
+
+---
+
+## 1. Purpose
+
+Produce **quantifiable results from the existing MJ agent framework as a baseline**, then
+run the *same corpus* against the native tool calling implementation and report deltas.
+Two properties are non-negotiable:
+
+1. **Deterministic evaluators only.** Every expected outcome is verified structurally —
+   response parses, next-step type matches, action name matches, params satisfy declared
+   matchers. No LLM-judge in the scoring path. (The framework's `llm-judge` oracle exists
+   and may be *attached* to exploratory suites later, but it contributes no weight to
+   baseline-vs-native comparison numbers.)
+2. **Mode-agnostic expectations.** A test case says *what the agent should decide*
+   ("invoke `Run Ad-hoc Query` with these params"), not *how the wire encodes it*. The
+   evaluator normalizes both encodings — envelope `nextStep.actions[]` and native
+   `ChatToolCall[]` — into one observed-decision record before comparison. This is what
+   makes baseline and native runs directly comparable on identical cases.
+
+### How a decision is verified (worked example)
+
+Given an agent with actions A, B, C and a case expecting B: for prompt-level cases,
+**no action ever executes** — `AIPromptRunner.ExecutePrompt` returns the model's raw
+decision and nothing is dispatched. The oracle is a plain data assertion:
+envelope mode parses the JSON and asserts `nextStep.type === 'Actions'` and
+`actions[0].name === 'B'`; native mode asserts `toolCalls[0].name === 'B'` on the
+normalized result. For loop-level cases (actions do execute), the framework already
+persists every step as `AIAgentRunStep` rows (`stepType: 'Actions'`, step name, full
+I/O); the trace oracle queries those rows and asserts a B step exists and no A/C step
+does. The evaluator is fully deterministic in both cases — what varies run-to-run is
+the *model's* choice, which is the quantity being measured. Hence repetition counts
+and rates (§6) rather than single pass/fail booleans: 17/20 choosing B is a data
+point, not a flaky test.
+
+This is also why the harness is not a plain unit-test suite: mock the model and there
+is nothing left to measure; call live models from a unit runner and it's a benchmark
+without bookkeeping. The assertion logic itself *is* unit-test-simple and is itself
+unit-tested (§5); the framework contributes the campaign around it — live execution,
+N-per-cell repetition, persisted per-oracle results, and longitudinal comparison
+between the baseline and native runs.
+
+---
+
+## 2. Harness Architecture Decision
+
+Three options were considered for where the harness runs:
+
+| Option | Description | Assessment |
+|---|---|---|
+| A. Inside MJAPI | Drive tests through the running API server | Rejected. Prompt-run-level testing needs `AIPromptRunner`, a DB provider, and `AIEngine.Config()` — not HTTP. The server adds startup cost, auth plumbing, and noise without adding fidelity at the layer under test. |
+| B. Dedicated standalone process | Custom `tsx` scripts bootstrapping the provider directly | **Adopted as the documented fallback (§2.1)** and for Phase B probes. MJ already has this shape (`packages/TestingFramework/integration-test-suite/rigs/` + `rigs/lib/ai-bootstrap.ts`, `integration-test-scripts/run-harness-agent.ts`). Its cost is that test/suite/run persistence, scoring, and comparison tooling must be hand-built. |
+| **C. MJ TestingFramework (recommended)** | New driver + oracles in `@memberjunction/testing-engine`, invoked via `mj test` | **Adopted as primary.** Nearly everything this plan needs already exists. |
+
+### Why C: what already exists
+
+- **Metadata-driven test content:** `MJ: Test Types` → `MJ: Tests` (input + oracle config
+  as JSON) → `MJ: Test Suites` → persisted `MJ: Test Runs` with `ResultDetails =
+  OracleResult[]`. Tests ship as stock metadata exactly like the existing
+  **Research Agent Evaluation Suite** (`metadata/test-suites/.research-agent-evaluation-suite.json`,
+  `metadata/tests/research-agent/*.json`).
+- **Driver dispatch:** `@RegisterClass(BaseTestDriver, …)`; `AgentEvalDriver`
+  (`packages/TestingFramework/Engine/src/drivers/AgentEvalDriver.ts`, 1,113 lines) is a
+  worked example — single- and multi-turn, per-turn execution params, cancellation,
+  `TestRun.TargetLogID ↔ AIAgentRun.TestRunID` bidirectional linkage.
+- **Deterministic oracle contract:** `IOracle { type; evaluate(input, config) →
+  {passed, score, message} }`, weighted aggregation (`Engine/src/utils/scoring.ts`),
+  public `TestEngine.RegisterOracle()`. Shipping deterministic oracles: `schema-validate`,
+  `exact-match`, `sql-validate`, `trace-no-errors`.
+- **Persistence verifiers:** `verifyPromptRun` / `verifyAgentRun`
+  (`packages/TestingFramework/testing-integration/src/ai-verify.ts`) assert framework-side
+  determinism (terminal statuses, token/cost rollups, messages persisted) even when model
+  output varies.
+- **Tier gating + CI lanes:** `IsTierEnabled('deterministic' | 'mutation' | 'live-model')`;
+  deterministic runs on PR CI against a throwaway SQL Server with stock metadata
+  (`.github/workflows/integration.yml`); live-model runs weekly, budget-capped.
+- **Hermetic LLM substitution:** `TestLLM` + `registerTestLLM` from
+  `@memberjunction/unit-testing` scripts exact model outputs through the real ClassFactory
+  and real `BaseLLM` routing — golden-transcript replay for testing the harness itself.
+- **Tooling:** `mj test run | suite | compare | history | report | validate`, repeat
+  counts, `--flaky-check N`, cost calculation, and the Explorer Testing dashboard.
+
+Running under the framework also means the harness is connected to a database with stock
+MJ metadata (and optional mock test data), i.e., a genuinely integrated environment —
+the benefit of option A without the server.
+
+### What we build (the gaps)
+
+1. **`PromptEvalDriver`** — no prompt-run-level driver exists today (everything is
+   agent-level or unit-mocked). See §3.1.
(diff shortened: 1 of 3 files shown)


rkihm-BC (approved):
## Approving. This lands dark, and I verified that rather than taking it on trust.

On a 1,076-file PR the sentence doing all the work is *"nothing changes behavior on merge."* I traced it end to end instead of reading the design and agreeing with it. It holds for the tool-calling path — and the discipline that makes it hold is the best thing about this branch.

### Three changes I made directly — please review them

I pushed one commit (`c63489c`) and edited the PR description. Nothing in `packages/` was touched; the code is entirely as you wrote it.

**1. Changeset — added the model swap.** This is the one that mattered. `.changeset/native-tool-calling-basellm-surface.md` had no mention of the 16 `ModelID` changes or the four `Status: "Inactive"` rows. You documented them clearly in the PR description under *Model selection*, so this was never hidden — but release notes are built from the changeset, not the description. A user whose Research Agent output shifts after upgrading would have found nothing explaining why. Added as its own paragraph, using your rationale about `mj sync push` never deleting.

**2. Changeset — scoped the opening claim.** It said *"Nothing changes behavior on merge"* at the top, then described behavior that changes on merge in its own closing paragraph. Now reads *"Nothing in the tool-calling path changes behavior on merge"*, which is both accurate and still the reassurance you meant to give.

**3. Changeset — named the peer-strictness change.** The toolchain paragraph described moving settings into `pnpm-workspace.yaml`, but not that `strict-peer-dependencies` goes `true` → `false` in the process. That is the part with a behavioral consequence: a mixed workspace that previously failed the install now reports its peer gaps and completes. Added one clause, using the reasoning from your own doc comment in `build.ts`.

Bump levels and the rest of the body are untouched — I checked the frontmatter is byte-identical. The PR description got the matching correction, listing all three ungated areas.

If you disagree with any of the wording, overwrite it; I would rather you own the sentence than keep mine.

### CodeQL alert 455 — dismissed as a false positive, no code change

`js/polynomial-redos` on `action-tool-builder.ts`. It is not exploitable, because the preceding line makes the pathological input unreachable:

```ts
.replace(/[^a-zA-Z0-9]+/g, '_')   // collapses EVERY run of non-alphanumerics to ONE underscore
.replace(/^_+|_+$/g, '')          // so `_+` here can only ever match a single character
```

After step one the string cannot contain `__`, so `_+` never backtracks. Measured through the full function: 1,000 underscores → 0.04 ms, 400,000 → 0.40 ms. A 400x input costs 10x, and that cost is string allocation, not backtracking. Rewriting the regex would have added a guard against a state that cannot occur, so I dismissed the alert instead. Worth knowing: `next` already carries 129 open alerts of this same rule, so this is not special-casing your PR.

### What I actually checked

**The gate is fail-safe by construction.** `ResolveNativeToolCalling` ANDs three conditions that each default to `false`, capability is a hard gate, and an *absent* capability reads as unsupported rather than unknown. Against shipped metadata: 139 models carry `SupportsNativeToolCalling: true` and **all 139** carry `DefaultToNativeToolCalling: false`; no prompt sets `UseNativeToolCalling`. Every conjunct is false on every existing path.

**The agent loop is byte-identical off-path.** Every new branch keys off `previousDecision?.nativeTurn?.sendResultsNatively`, and each `else` preserves the original push verbatim — I diffed the content strings. `nativeTurn` has exactly one assignment site, inside `nextStepFromNativeToolCalls`, which returns `null` before reaching it when no tool calls came back.

**The provider refactors preserve live behavior.** This was the real risk, since five drivers changed on paths that run today regardless of the flag. Cerebras's multimodal flattening is character-identical and all five `responseFormat` cases survive; `toClassicChatMessageRole` is the identity function for every role that exists today. Anthropic's `coalesceToolMessages` returns the same array reference when there are no tool turns. Groq's trailing-user guard is intact. OpenAI's reasoning change is a *strict widening* — `clampToSystemPromptLevel` only rewrites `'xhigh'` and `'none'`, both of which previously threw, and the numeric 1-100 banding is unchanged.

**The Loop system-prompt template deserves specific credit.** At +39/−11 on the prompt every Loop agent renders, this looked like the place the claim would break. It does not: every new block sits behind `_NATIVE_TOOL_CALLING` / `_NATIVE_CONTROL_FLOW`, and the additions fall through to `{%- elif actionCount > 0 %}`, which is the original text verbatim. With the gate closed it renders identically. Getting that right in Nunjucks, across nested conditionals that already had six other flags in them, is not easy and it is not lucky.

**Migration** is clean against the MJ rules: `v6.1.x` filed in `migrations/v6/`; all nine `EntityField` INSERTs use apply-time `COALESCE(MAX([Sequence]),0)+1` as separate `IF NOT EXISTS` statements, so the four sharing one `EntityID` re-read `MAX` in sequence and cannot collide on a fresh database; 52 blank lines before the CodeGen tail; no `migrations-pg/` counterpart, which is correct. Purely additive — every `DROP`/`DELETE` is inside the standard recreate tail. `minor` is the right bump.

The head comment on that migration is unusually good. Arguing the JSON-bag-versus-columns boundary explicitly, and recording that `PromptConfiguration` is not named `Configuration` because `vwAIPromptModels` would fail to build on the duplicate display column, is exactly the reasoning that is normally lost and then rediscovered painfully.

### Non-blocking, entirely optional

`cerebras.ts:239` and `:269` use `request as unknown as Record<string, unknown>`, which is the `unknown`-as-a-shortcut pattern `typescript-style.md` discourages. It is four sites across non-test source — the other 48 added casts are mock casts in tests, which is fine — and the Cerebras SDK type genuinely lacks `tools`. Not worth a round trip on its own; noting it only so it is a deliberate choice rather than an oversight.

### Size

1,076 files is daunting until you look at the composition: 898 files / 90k lines are the prompt-eval corpus, one file / 10.7k is the migration, 14 are generated. The hand-written surface is 78 files / ~9.9k lines with 42 test files / 5.8k lines behind it — a shared `llm-tool-conformance` harness, a gate truth-table suite, and per-provider conformance suites. That is a reviewable ratio, and CI is green across the drift gate, the deterministic integration tier and the migration checks.

Nice work. Routing every new branch through a falsy-by-default guard, rather than bolting the feature onto the existing paths, is what makes something this large safe to land switched off.

packages/AI/Prompts/src/AIPromptRunner.ts
```diff
@@ -3452,6 +3512,169 @@ export class AIPromptRunner {
...
+   *
+   * Split out because the answer is needed twice and must be the same both times: once before the
+   * template renders — the loop template drops its action catalog and the `'Actions'` step type
+   * when native mode is on (plan §8.4), and it can only do that truthfully if it knows the real
+   * decision rather than the caller's intent — and once when the request is assembled.
+   *
+   * Never throws: the gate is an opt-in enhancement and must not be able to fail a run that would
+   * otherwise succeed, so any configuration problem resolves to the path that has always worked.
+   */
+  public resolveNativeToolCallingDecision(
```
AN-BC, line 3540: @jordanfanapour @CaelebB-BC for a follow up PR, this violates convention, public members of classes are `PascalCase` please keep an eye out for this and also with exported functions we use same convention. we use `camelCase` for private/protected and non-exported functions. This convention is helpful as it makes it very easy at a glance to know the access modifier/visibility of a given member of class or of a function when reading code without having to look at definitions.

@CaelebB-BC  when you fix this, please see if you can rig up a repo-wide unit test that looks for violations - it would use an AST based approach to parse the entire repo and look for violations - public members that are not `PascalCase`, private/protected members that have a capitalized first letter that shouldn't, and exported functions/consts etc following same convention.

@rkihm-BC  FYI for you.

AN-BC (commented):
Great work @CaelebB-BC and @jordanfanapour, I particularly like the harness. I am a little late to the party since this already merged, which is fine, but can you make the fixes noted below and get them into both `next` and the new `6.1.0` LTS branch that @cadam11  is about to cut?

PR #4177, opened 2026-09-03
fix(ng-base-forms): a date-only field is a calendar day, not an instant

JosueGarcia-BC:
Read mode and edit mode disagree by a day, on ordinary valid data. No corruption, no invalid input.

## The defect

A `date` column has no time and no zone — it is a calendar day — and arrives as UTC midnight. `FormatValue()` renders it with `toLocaleString()`, a **local-time** formatter, which subtracts the reader's offset and lands on the previous day for everyone west of Greenwich.

Measured in `America/New_York`, executing this file's own source:

| stored (SQL `date`) | read mode showed | edit mode showed |
|---|---|---|
| `2026-11-20` | **11/19/2026, 7:00:00 PM** | 2026-11-20 |
| `2026-09-30` | **9/29/2026, 8:00:00 PM** | 2026-09-30 |
| `2026-01-01` | **12/31/2025, 7:00:00 PM** | 2026-01-01 |

The last one is the wrong **year**.

Edit mode was already correct — `DateInputValue` uses `toISOString()`. So a user opens a record and sees 19 November, clicks Edit, and sees the 20th. Same field, same session.

## The fix

`EntityFieldInfo.Type` already distinguishes the two cases, so the component can tell a calendar day from an instant:

- **`date`** → `toLocaleDateString(undefined, { timeZone: 'UTC' })`
- **`datetime` / `datetimeoffset`** → unchanged `toLocaleString()`

Pinning the timezone rather than switching to `toISOString()` keeps the reader's locale format — a US reader still sees `11/20/2026`, a UK reader `20/11/2026` — and changes only the zone the day is computed in. A timestamp names an instant, and local time with a time-of-day is the right way to show one.

## Why it survived

**The tests pin a timezone rather than trusting the runner's.** Every assertion here passes at Greenwich and fails in New York, so a UTC CI box could never observe this. That is why it shipped.

Four tests added; **three fail against the previous implementation**. The fourth asserts a timestamp still renders in local time *with* its time-of-day, so a future "fix" cannot drag instants into UTC while chasing this one.

Full `@memberjunction/ng-base-forms` suite: **328 passed / 26 files**.

## Not fixed here, deliberately

`new Date('2026-02-31')` rolls over to 3 March rather than failing, so an impossible date-shaped string displays as a different day. I wrote a round-trip guard in `DateInputValue` and **removed it**: `BaseEntity` coerces the value on `Set`, so the getter receives `2026-03-03` and the original text is already gone. The guard could never fire.

That belongs in the entity's type coercion, not in a form field, and deserves its own issue rather than an assertion that cannot execute.

## Provenance

Found while investigating `bc-aidp-next-golive#185`, which reported a related symptom against a BizApps Sales component that turns out to be retired. The live defect is here, in MJ, and affects every app that renders a date field.

### DIFF
--- a/.changeset/form-field-date-only-utc.md
+++ b/.changeset/form-field-date-only-utc.md
@@ -1,0 +1,5 @@
+---
+"@memberjunction/ng-base-forms": patch
+---
+
+fix(ng-base-forms): a `date` column is a calendar day, not an instant. Read mode formatted date-only fields with a local-time formatter, so a stored 2026-11-20 rendered as 11/19/2026 for anyone west of Greenwich while edit mode showed the 20th. Date-only fields now render the stored day in the reader's locale format; `datetime` / `datetimeoffset` fields are unchanged and still render as local time.
--- a/packages/Angular/Generic/base-forms/src/lib/field/form-field.component.dom.test.ts
+++ b/packages/Angular/Generic/base-forms/src/lib/field/form-field.component.dom.test.ts
@@ -74,6 +74,8 @@ function makeWidgetEntityInfo(): EntityInfo {
       { ID: 'F3', Name: 'Description', Type: 'nvarchar', Length: 200, AllowsNull: true, AllowUpdateAPI: true },
       { ID: 'F4', Name: 'Quantity', Type: 'int', AllowsNull: true, AllowUpdateAPI: true },
       { ID: 'F5', Name: 'LaunchDate', Type: 'datetime', AllowsNull: true, AllowUpdateAPI: true },
+      // A DATE-ONLY column, deliberately beside the datetime above: the two must render differently.
+      { ID: 'F9', Name: 'EffectiveDate', Type: 'date', AllowsNull: true, AllowUpdateAPI: true },
       { ID: 'F6', Name: 'IsActive', Type: 'bit', AllowsNull: true, AllowUpdateAPI: true },
       {
         ID: 'F7',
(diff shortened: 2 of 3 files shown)


rkihm-BC (dismissed):
Approving. I ran the new tests locally against both the base component and this branch: 3 of the 4 fail on base with exactly the `11/19/2026, 7:00:00 PM` / `12/31/2025` outputs described, and all 25 pass with the fix. The `process.env.TZ` pinning is real under the repo's jsdom preset because it uses the `forks` pool.

Keying off `EntityFieldInfo.Type === 'date'` is correct for both dialects (SQL Server and PostgreSQL both name the type `date`), and the fourth test guarding timestamps is the right insurance.

**Optional nits, not blocking:**
- The old one-line `/** Format a value for display */` comment now sits above the new `IsDateOnlyField` doc block, so `FormatValue()` lost its comment and the getter has two.
- The new `EffectiveDate` test field reuses ID `F9`, which `Website` already has. Harmless (nothing keys fields by ID) but `F10` would be tidier.

**Follow-up worth an issue (not this PR):** the same day-shift exists everywhere else a `date` column is rendered. `FormatValueInternal` in `@memberjunction/core` (`util.ts`) formats every date type through a local `Intl.DateTimeFormat`, which is what `EntityFieldInfo.FormatValue`, `BaseRecordComponent.FormatValue`, entity cards and record changes use. The data grid's friendly-dates path does the same. After this merges, the form and the grid will disagree by a day west of Greenwich. The clean fix is a date-only predicate in `@memberjunction/sql-dialect` (alongside `IsDateSQLType`) consumed by the core formatter, so this component could eventually call `FieldInfo.FormatValue` instead of carrying its own branch.

Merge note: this and #4178 both append a `describe` block at the end of the same test file and conflict there (checked with `git merge-tree`). Land this one first; #4178 needs a trivial rebase afterwards.

rkihm-BC (approved):
LGTM

PR #4178, opened 2026-09-03
fix(ng-base-forms): say when a stored date cannot be shown, instead of a blank box

JosueGarcia-BC:
Fixes the actual complaint in **MemberJunction/bc-aidp-next-golive#185**: *"an invalid stored date should never render as an empty field that looks like 'no date'."*

Separate from #4177 on purpose. That one fixes a **day-shift** in read mode; this one fixes a **blank editor**. Same file, different functions, different bugs — they can be reviewed and reverted independently.

## The failure is data loss, not an ugly form

`DateInputValue` returns `''` for a value it cannot render, and `<input type="date">` renders `''` as an empty box — **the same box it shows for a field nobody ever set**. So the field reads as "no date", and the next save writes that emptiness over whatever was actually stored.

An unreadable value becomes a destroyed one, silently, by a user who was never told anything was wrong.

## Why the control cannot fix this itself

`DateInputValue` is already correct: it returns `''` rather than the `NaN-NaN-NaN` an unguarded formatter would produce. **That is left untouched.**

The gap is that `''` and "no date" are the same picture, and **there is no string that makes an `<input type="date">` display `not-a-date`**. The element physically cannot surface the problem. The most the getter can do is refuse to emit garbage; saying *why* has to happen beside the control.

## Read mode needed nothing, and got nothing

`FormatValue()` already renders `Invalid Date` for a bad `Date` and the raw text for a bad string — that surface is already honest and already satisfies the issue. Only the **editor** hides the problem, which is why `StoredDateIsUnreadable` is scoped to `EditMode` rather than to the field. A warning in read mode would be noise on a surface that is already telling the truth.

## Tests

Nine, and the split is deliberate:

- **Four detectors** — each asserts the warning appears for a value the editor cannot render: an unshowable stored value, a UK/EU formatted string, a `dd-MM-yyyy` string, and a whitespace-only value. All four **fail against the previous implementation**, and all four fail again if `StoredDateIsUnreadable` is stubbed to return false.
- **Five guards** — no warning for an empty field, none for a valid date, none in read mode, none for the epoch-millisecond wire format that is the normal load path, and one recording a known limit (a value that parses to the wrong date is not caught). These pass either way, and exist to stop the warning spreading. A warning that fires on ordinary data gets deleted within a week, and then the defect is back.

`@memberjunction/ng-base-forms`: **337 passed / 26 files** on this branch, against **328 / 26** on `next` — the nine above are the whole difference. Measured after rebasing onto current `next`, which now carries #4177.

## Scope note

Reported against a BizApps Sales component (`ToDateInput` in `deal-workspace.dates.ts`) that turns out to be retired — nothing renders it since the #89 UI work. The live surface is this component, and it affects every MJ app that edits a date field, not only Sales.

---

## Is this reachable, or only theoretical? — measured

I checked rather than assumed, and the answer has two halves.

**A record loaded from the database can never trigger this.** Measured on a live server: dates come over the API as **epoch milliseconds** — `ExpectedCloseDate` arrived as `1790726400000` for a `date` column holding `2026-09-30`. A number always parses, so opening a record cannot produce a blank editor. Nobody should claim otherwise, and a test asserts this path stays **silent**.

**What does reach it is client-side assignment** — an import, an integration, an Action — where `BaseEntity.Set` takes whatever it is handed:

| assigned value | editor | warns |
|---|---|---|
| `'20/11/2026'` (UK / EU) | blank | **yes** |
| `'20-11-2026'` (dd-MM-yyyy) | blank | **yes** |
| `'   '` (whitespace) | blank | **yes** |
| `'next Tuesday'` | blank | **yes** |
| `1790726400000` (the wire format) | `2026-09-30` | no — correct |

The UK/EU row is the one that matters: `20/11/2026` is how most of the world writes a date, and it is exactly what a CSV out of a European system contains. Today that silently reads as "no date" and the next save destroys the value.

## What this deliberately does NOT catch

A test records this rather than leaving it to be found later. An **Excel date serial** (`46346` = 20 Nov 2026 in Excel's epoch) is read as 46346 *milliseconds* and renders **`1970-01-01`** — confidently, silently, and wrong. It parses, so nothing at this layer can distinguish it from a real date. Same shape as `2026-02-31` rolling to 3 March.

Both need the value **refused at ingress**, in the entity's type coercion, before anything is asked to display it. That is a separate change; the limit is now written beside the warning instead of implied.


packages/Angular/Generic/base-forms/src/lib/field/form-field.component.html
```diff
@@ -324,6 +324,20 @@
                 class="mj-forms-field-input mj-forms-field-input--date"
                 [value]="DateInputValue"
                 (change)="OnDateChange($event)">
+              <!--
+                An unreadable stored value renders as an EMPTY date box, identical to a field that
+                was never set - and the next save writes that emptiness over whatever is stored. The
+                element cannot display the bad value itself, so the warning has to sit beside it.
+              -->
+              @if (StoredDateIsUnreadable) {
```
rkihm-BC: This sits inside `.mj-forms-field-control`, a flex row (`display: flex; align-items: center; gap: 8px`). With the date input capped at `max-width: 220px`, this div becomes the next item on the same line and renders to the right of the box, not below it. The existing `DisplayErrors` / `DisplayWarnings` blocks are placed after the control div as direct children of the `.mj-forms-field` grid, where `.mj-forms-field-validation { grid-column: 2 }` puts them under the control. Please move this block there (after the closing `</div>` of the control, next to the other two validation blocks) so it lays out the same way.

packages/Angular/Generic/base-forms/src/lib/field/form-field.component.html
```diff
@@ -324,6 +324,20 @@
...
+                <div class="mj-forms-field-validation mj-forms-field-validation--warning">
+                  <div class="mj-forms-field-validation-item">
+                    <i class="fa-solid fa-triangle-exclamation"></i>
+                    <span>This field holds a value that is not a readable date, so the box below is
```
rkihm-BC: "the box below" points the wrong way: the input precedes this element in DOM order, so the box is above the warning in any layout, and beside it in the current one. Suggest "so the date box is empty" or "so the box above is empty". The second sentence is accurate as written; saving an untouched invalid date does write NULL to a nullable column.

rkihm-BC (changes requested):
The diagnosis and the tests are right, and I confirmed the data-loss claim is real rather than theoretical: on Save the provider converts the invalid Date to `NaN` via `getTime()`, the request client serializes that as JSON `null`, and the server resolver passes it into `SetMany`, so a nullable date column is written NULL. Entity validation does not catch it because an invalid Date is still `instanceof Date`. Scoping the warning to edit mode and the known-limit test for Excel serials are both good calls.

Two required fixes. The first the DOM tests cannot see because jsdom does no layout:

**The warning renders beside the date input, not under it.** The new block is inside `.mj-forms-field-control`, which is `display: flex; align-items: center` in row direction (`form-field.component.css` line 42). The date input is capped at 220px, so the warning becomes the second flex item on the same line, to the right of the box. The existing error/warning blocks avoid this by living outside the control div as grid children in column 2, which is what their `grid-column: 2` rule assumes. Move this block out of the `@case` to sit with those blocks after the control div (around line 385), gated the same way. Once it is there the message wording also needs adjusting; see inline.

**Second required fix: add a changeset.** There is none on this branch. Changesets drive versioning and their text is what the release notes are generated from, so without one this fix ships undocumented. `patch` is the right level since there is no migration or metadata change:

```md
---
"@memberjunction/ng-base-forms": patch
---

fix(ng-base-forms): when a date field holds a value the date editor cannot display, show a warning beside the empty input instead of a blank box that reads as "no date". Previously an unreadable stored value looked identical to an empty field, and saving the record silently wrote NULL over it.
```

(#4177 now carries its own changeset for the same package; both are `patch`, so they compose without conflict.)

Optional while you are in there: add `|| StoredDateIsUnreadable` to the `[class.mj-forms-field--has-warning]` binding on the editing wrapper (line 92) so the input gets the same amber underline every other warning gets.

Housekeeping: the body says one detector test; the branch now has four that fail against base (I ran them). And this conflicts with #4177 in the test file only, both append at the end. #4177 should land first and this one rebase.

rkihm-BC:
Added a second required item to my review above: this PR needs a changeset (`@memberjunction/ng-base-forms: patch`). Changesets drive versioning and the release notes are generated from their text. A ready-to-paste file is in the review body.

rkihm-BC (approved):
Approving. Both required items are done exactly as asked, and the optional one too.

- **Placement:** the warning block moved out of the flex-row `.mj-forms-field-control` and now sits as a sibling after the existing error/warning blocks, where `.mj-forms-field-validation { grid-column: 2 }` puts it beneath the control; the old in-control copy is gone (the only other reference is the class binding).
- **Wording:** "the box above is empty". The second sentence stays, correctly — I verified earlier that saving an untouched invalid date does write NULL over a nullable column.
- **Optional, done:** the editing wrapper's `[class.mj-forms-field--has-warning]` now includes `StoredDateIsUnreadable`, so the input gets the same amber underline every other warning gets.
- **Changeset** added, `patch` on `@memberjunction/ng-base-forms`, accurate text.
- CI fully green on this head; the other files in the delta are merges of `next` (incl. #4177's landed tests). Mergeable.

PR #4184, opened 2026-09-03
fix(search): resolve record keys from entity metadata instead of a hardcoded ID column

### LINKED ISSUE #4179: Search result click-through breaks for non-'ID' primary key entities
**Describe the bug**
In MJ Explorer's universal search, results are returned and ranked correctly, but clicking a result fails to open the record for any entity whose primary key column is not literally named ID:

> Could not load <Entity> record.
> InnerLoad returned false for key ID=<the correct pk value>

Both search-result navigation handlers in `@memberjunction/ng-explorer-core` build the record key with a hardcoded field name:

- `ShellComponent.OnSearchResultSelected()` (nav-bar dropdown)
- `SearchResultsResource.navigateToResult()` ("See all results" page)

```
const pkey = new CompositeKey([{ FieldName: 'ID', Value: result.RecordID }]);
this.navigationService.OpenEntityRecord(result.EntityName, pkey);
```
`GenericDatabaseProvider.Load()` then requires that field name to match an actual PK column:
```
const pk = entityInfo.PrimaryKeys.find(p => p.Name.toLowerCase() === val.FieldName.toLowerCase());
if (!pk) throw new Error(`Primary key ${val.FieldName} not found in entity ${entityInfo.Name}`);
```
So for entities mapped from external/source schemas that use `individual_id`, `organization_id`, etc. as the PK, the load throws and the record won't open from search. The `result.RecordID` value is correct, only the hardcoded field name is wrong.

Related: `EntitySearchProvider` / `FullTextSearchProvider` emit a single `RecordID` string per result, so entities with composite primary keys can't round-trip through search navigation at all.

The correct pattern already exists elsewhere in the same package - `SingleSearchResultComponent.navigateToSingleResult()` uses `compositeKey.LoadFromEntityInfoAndRecord(entityInfo, record)` (`@memberjunction/core` → generic/compositeKey.ts), and other shell code paths use `entityInfo.FirstPrimaryKey.Name`. The two unified-search nav sites just don't.

> MemberJunction version: @memberjunction/* 5.51.0 (also reproduces on 5.51.2)

**To Reproduce**
1. Register an entity whose primary key column is not named ID (e.g. individual_id), with AllowUserSearchAPI = true and at least one IncludeInUserSearchAPI field.
2. In the Explorer nav-bar search, type a value that matches a record in that entity.
3. The result appears in the suggestions dropdown, correctly titled and ranked.
4. Click the result (or open the full "See all results" page and click the row).
5. See error: `Could not load <Entity> record. InnerLoad returned false for key ID=<pk value>`.

**Expected behavior**
Clicking a search result opens the corresponding record, regardless of the entity's primary-key column name (and for composite-key entities).

**Screenshot**

Image

ct-bc:
Closes #4179

## Summary

Clicking a universal-search result failed with `InnerLoad returned false for key ID=<value>` for any entity whose primary key column is not literally named `ID` (`individual_id`, `organization_id`, …). The search navigation sites built the key as `{ FieldName: 'ID', Value: RecordID }` / `CompositeKey.FromID(RecordID)`, and `Load()` correctly rejects a field that is not one of the entity's primary keys — right value, invented field name.

MJ supports primary keys with **any column name(s) and type(s)**, so rather than patching the two sites named in the issue, this PR fixes the contract up the stack: every consumer resolves the key against the entity's metadata, every producer emits a key built from the entity's real primary key column(s), and composite (multi-column) keys round-trip end to end — including through the fail-closed permission filter that would otherwise have silently dropped them.

## The contract

A search result's `RecordID` is a **compact `CompositeKey` segment**:

| Key shape | `RecordID` | Why |
|---|---|---|
| single column (any name) | the bare value, e.g. `11055` | byte-identical to today, so `IN (...)` filters, fusion dedup keys, and persisted ids are unchanged |
| composite | `OrderID\|11055\|\|LineNo\|3` | field names travel with the values, so the key can be rebuilt losslessly |

`CompositeKey.LoadFromURLSegment(entity, s)` already read both forms. Three additions in `@memberjunction/core` make it a one-liner everywhere:

- `CompositeKey.FromURLSegment(entityInfo, recordId)` — static reader; falls back to an `ID` key only when the entity cannot be resolved (preserves prior behavior on an unknown entity instead of throwing).
- `CompositeKey.FromEntityRecord(entityInfo, row)` — key from a RunView row via `entity.PrimaryKeys`.
- `FieldValueCollection.ToCompactURLSegment()` — the serializer for the table above (also keeps the prefix when a lone value itself contains `|`).
- `ToWhereClause()` now doubles embedded quotes, since both composite permission paths build SQL through it from ids that can originate in an external index.

All additive — no signature changes, no removals — which is what makes the planned 5.51.x LTS backport safe.

## Changes

**Consumers — the nav sites** (`ng-explorer-core`, `ng-search`), now `CompositeKey.FromURLSegment(md.EntityByName(name), recordId)`:
- `ShellComponent.OnSearchResultSelected` and `SearchResultsResource.navigateToResult` — the two in the issue.
- `OmnibarPaletteComponent.Execute` — `FromID(nav.recordId)`; the default search surface (`UseOmnibar`), not named in the issue.
- `ViewResourceComponent.onOpenRelatedRecord` / `SingleSearchResultComponent.OnOpenRelatedRecord` — FK-cell "open related record", same defect.
- `OmnibarRecordProvider.resolveRecordNames` / `SearchService.GetRecentlyOpenedRecords` — recents showed a raw id instead of a name for these entities.

**Producers** (`core`, `search-engine`, `ai-vectors-memory`):
- `EntitySearchProvider` read `record.ID`, which is `''` for these entities; `SearchFusion` drops empty ids, so **the entity lane was contributing nothing for them** (the reporter's clicked results came from the full-text lane). Now built from `PrimaryKeys`. No query change — RunView always returns PK columns.
- `ProviderBase.FullTextSearch`, `SearchEntity`'s lexical pass (`row['ID']`) and its permission filter (`ID IN (...)`, `Fields: ['ID']`) — the `SearchEntity` API returned nothing for these entities.
- `VectorSearchProvider` no longer flattens a composite key to bare values joined by `||` (unparseable); single-key behavior unchanged.
- `SimpleVectorDatabase` keyed rows by `row['ID']` and emitted a literal `` `ID|…` `` for an arbitrary `config.entityName`; now threads the `EntityInfo`.

**Permission filter** (`search-engine`) — `SearchEngine.verifyOwnershipAndRowFilters` verified with `FirstPrimaryKey IN (...)`. Once composite entities emit real segments that can never match, and the check fails closed, so every composite-key result would be dropped as unauthorized. Composite keys now verify with one `(F1=… AND F2=…)` term per record; single-column keys keep the `IN` fast path. Matching is on PK values in metadata order, UUID-normalized, so an externally indexed id with different casing still matches the row the database returns.

**Recents** (`ng-shared-generic`) — `RecentAccessService` persisted `Values(',')` (field names dropped; composite keys could never be reopened). Now writes the compact segment; existing single-value rows are unchanged.

**Adjacent** (`core`) — `EmbeddedRecord.keyFromForeignKey` used `FromID` for a single non-`ID` key.

**Docs** — `EntityInfo.FirstPrimaryKey` JSDoc now states what it is for (FK targets, keyset ordering, IS-A shared keys, the URL shorthand) and that it must not be used to construct a load key for an arbitrary entity.

## Tests

- `compositeKey.test.ts`: 12 new cases — compact serialization for single/composite/`|`-in-value/null, `FromURLSegment` for bare/prefixed/composite/unknown-entity, `FromEntityRecord` column order, `ToWhereClause` quote escaping.
- `EntitySearchProvider.test.ts`: alt-named single PK (row with no `ID` column at all), composite PK, and the common `ID` case unchanged.
- `SearchEngine.permissions.test.ts`: alt-named PK verifies against that column; composite verifies with the exact `(F1=… AND F2=…) OR (…)` filter and `Fields`; casing-insensitive match; composite membership ANDed with an RLS clause.
- `VectorSearchProvider.test.ts`: multi-key expectation corrected to the prefixed segment.

## Verification

| Package | Tests |
|---|---|
| `@memberjunction/core` | 146 files, 2176 passed |
| `@memberjunction/search-engine` | 31 files, 503 passed |
| `@memberjunction/ai-vectors-memory` | 87 passed |
| `@memberjunction/ng-explorer-core` | 20 files, 167 passed |
| `@memberjunction/ng-search` | 125 passed |
| `@memberjunction/ng-shared-generic` | 27 passed |
| Root `pnpm run build` | 303/303 tasks |
| `Integration Tests — Deterministic` | 64 passed / 0 failed / 6 skipped (the standard gated set: IT29, IT67, IT74, IT75, IT86, and IT52 which is parked upstream) |

## Versioning

`patch` for all six packages — code, tests and changeset only; no migration or metadata.

## Follow-ups (separate PRs)

- Repo-wide sweep of the same anti-pattern (`FieldName: 'ID'`, `CompositeKey.FromID` on variable entities, `PrimaryKeys[0]`) plus a `PrimaryKeyCompliance` guard test in MJCore modelled on `MultiProviderCompliance`.
- 5.51.x LTS backport after merge.


packages/MJCore/src/generic/compositeKey.ts
```diff
@@ -252,6 +251,31 @@ export class FieldValueCollection {
         return this.ToConcatenatedString(segment || CompositeKey.DefaultFieldDelimiter, CompositeKey.DefaultValueDelimiter);
     }
 
+    /**
+     * The exact inverse of {@link CompositeKey.LoadFromURLSegment}. A single-field key serializes to
+     * just its value — the shorthand `LoadFromURLSegment` maps back onto the entity's first primary
+     * key — while a multi-field key serializes to the full `Field1|Value1||Field2|Value2` segment.
+     * A single value that itself contains the value delimiter also gets the full segment, so the
+     * reader can never mis-split it.
```
rkihm-BC: **Required.** This guarantee does not hold: both parsers take only the second split part. `SimpleLoadFromURLSegment` (line ~319, `Value: kv[1]`) and `LoadFromConcatenatedString` (line ~234) parse `Code|a|b` as field `Code`, value `a`. I confirmed this against the built code.

The consequence is worse than the case this special-casing was meant to avoid. Bare `a|b` parsed to field `a`, which `Load()` rejects loudly. Prefixed `Code|a|b` parses to a valid field with a truncated value, so `Load()` succeeds on the wrong record if one has `Code = a`. The new test only asserts the serialization, not the round trip, which is why it is green.

Recommended fix (one line in each parser, plus a test): join the remainder back with the delimiter.

```ts
// SimpleLoadFromURLSegment
const kv = p.split('|');
pkVals.push({ FieldName: kv[0], Value: kv.slice(1).join('|') });

// LoadFromConcatenatedString
const kv = p.split(valueDelimiter);
pkVals.push({ FieldName: kv[0], Value: kv.slice(1).join(valueDelimiter) });
```

Every well-formed two-part segment is byte-identical under this change; only three-plus-part segments, which were previously truncated silently, are affected. None of the existing `compositeKey.test.ts` cases assert truncation, so nothing turns red. Both parsers need it because this PR's `RecentAccessService` change now writes the prefixed form into the same store that the DataExplorer recents reader parses with `LoadFromConcatenatedString`. Values containing `||` remain unrepresentable; that is a pre-existing limit of the format and fine to leave.

This stays additive, so the 5.51.x backport is unaffected. The alternative, dropping the `|` special case and this sentence, is acceptable but restores the loud failure rather than fixing it.

packages/MJCore/src/__tests__/compositeKey.test.ts
```diff
@@ -631,3 +633,101 @@ describe('ResourceUrlsEquivalent', () => {
...
+
+  it('emits the full prefixed segment for a composite key', () => {
+    const key = CompositeKey.FromKeyValuePairs([
+      new KeyValuePair('OrderID', 'o1'),
+      new KeyValuePair('LineNo', 3),
+    ]);
+    expect(key.ToCompactURLSegment()).toBe('OrderID|o1||LineNo|3');
+  });
+
+  it('keeps the field prefix when a lone value itself contains the value delimiter', () => {
```
rkihm-BC, line 658: With the parser fix, please add the round-trip assertion this test implies but does not make:

```ts
const key = CompositeKey.FromKeyValuePair('Code', 'a|b');
expect(CompositeKey.FromURLSegment(mockEntity('Code'), key.ToCompactURLSegment()).Equals(key)).toBe(true);
```

Against the current parser this fails (the value comes back as `a`), which is the gap described on the JSDoc line.

rkihm-BC (changes requested):
The diagnosis is right and fixing the contract at both ends, rather than the two sites the issue named, is the correct call. Verified rather than assumed:

- **Red then green, locally.** New tests against base sources: 14 `compositeKey` cases and 6 SearchEngine cases fail. Against the PR sources: MJCore 2,176 passed / 146 files, SearchEngine 31 files passed (MJCore dist rebuilt for the SearchEngine run). Matches the numbers in the description.
- **Bug confirmed on `next`** at the two reported sites and the five additional ones this PR found. The `EntitySearchProvider` finding is the most valuable part: the entity lane was silently contributing nothing for these entities and the full-text lane was covering for it.
- **Permission-filter reasoning is correct.** `FirstPrimaryKey IN (...)` fails closed, so once composite entities emit real segments every one of their results would be dropped. The `(F1=… AND F2=…)` rewrite with the single-key `IN` fast path preserved is right.
- **"RunView always returns PK columns"** is a documented platform contract (`UnionFieldsWithPrimaryKeys` in `providerBase.ts`).
- **Adjacent readers of the persisted id stay compatible.** The DataExplorer recents reader already branches on `|` and otherwise resolves the entity's primary key, so the compact form reads back correctly there.
- `ToWhereClause` doubling quotes also fixes a latent quoting bug for five other callers in the repo; none pre-escape, so no double-escaping risk.
- Changeset present, `patch` on six packages, correct.

**One required change, inline:** the new "value contains `|`" guarantee in `ToCompactURLSegment` is not delivered by the parser it relies on, and the new path turns a loud failure into a silent wrong-record load. The fix is one line in each of the two parsers plus a round-trip assertion; details on the JSDoc line. Recommending the parser fix over dropping the claim because it keeps the PR additive for the 5.51.x backport and changes no well-formed segment.

**Optional:** the follow-up sweep and `PrimaryKeyCompliance` guard the description commits to have no issue yet. `list-form.component.ts:488` still uses `FromID` for an arbitrary entity, so lists over these entities have the same click-through failure. Worth filing so it is not lost.

Branch is currently behind `next` (mergeable, state `BEHIND`); it will need a merge or rebase before landing.

rkihm-BC (approved):
Approving. The required change landed in `e7f5c352` exactly as requested, and you picked up the optional list-form follow-up too.

Verified on this head in a worktree with core's dependency chain built:
- **Red then green.** Against the previous parser, both new round-trip cases fail (the single `a|b` value and the composite segment containing one); against the new parser `compositeKey.test.ts` passes 87/87.
- **Behaviour on the built code:** `Code|a|b` parses to `{ Code: 'a|b' }` through both `SimpleLoadFromURLSegment` and `LoadFromConcatenatedString`; `OrderID|o1||LineNo|3` and `individual_id|42` are unchanged; the `a|b` round trip returns true. Well-formed two-part segments are byte-identical to before.
- The JSDoc now says what the parsers actually do and names the `||` limit honestly. Good.
- `list-form.component.ts` `openRecord` moving to `FromURLSegment` fixes the same click-through for lists over non-`ID` entities and correctly drops the old composite-key refusal.
- CI is fully green on this head, covering the SearchEngine and core-entity-forms packages I did not re-run locally. Behind `next` but mergeable; no protection rule on `next`.

Follow-up filed as **#4221**: the *write* side of list details still stores `record.ID` (six sites across the list form, single-list-detail, the two list actions and the Rex provider), so adding a non-`ID`-keyed record to a list writes an empty `RecordID` that the read path then drops silently — plus the repo-wide sweep and `PrimaryKeyCompliance` guard this PR's description proposes. Not a condition of this PR; with the read side in place each write site is a one-liner.

PR #4185, opened 2026-09-03
feat: Authorization.Check RO + CompositeFilter eval

AN-BC:
## Summary

Two related core pieces so Open Apps (Orders first) can ask *can this user?* and *does this price apply?* without inventing their own JSON or GraphQL.

### 1. `Authorization.Check` remotable operation
- Input `{ Names: string[] }` → per-name `Allowed`, `Unknown` (fail-closed), `ViaAncestor`, `MatchedAuthorizationName`.
- Rides existing `ExecuteRemoteOperation`. No new resolver. Does not extend `CheckEntityPermissions`.
- `UseAuditLog` writes `MJ: Audit Logs` (`Authorization Check` type).
- API scope `authorization:check`.
- Server class is `AuthorizationCheckServerOperation` so it does not clash with the generated client shell.

### 2. `CompositeFilter` + multi-entity FilterState JSON
- `CompositeFilterDescriptor` stays the portable Kendo JSON (User Views / builder output).
- `CompositeFilter` is the OOP wrapper: `FromJSON` / `FromDescriptor` / empty + `Add()`, instance `Evaluate(context)`, `SummaryText` / `SummaryHTML`.
- Public members and exported helpers are PascalCase (`IsCompositeFilter`, `ParseFilterField`, `FormatFilterField`, `CreateEmptyFilter`, …).
- When the builder is given `sources`, it **always writes** `Source.Field` (e.g. `BillToOrganization.Type`).
- **Read** accepts both: no dot = today’s single-entity behavior; dotted = honor the prefix.
- Views still compile to SQL and **strip** the prefix in `GenerateWhereClause`.
- Several sources: two-pane field picker. One source / legacy `fields`: unchanged chrome.
- `SafeExpressionEvaluator` (`@memberjunction/global`) is **not** this path — that’s for authored JS. Filter JSON is data.

Docs: [`guides/FILTER_BUILDER_AND_EVALUATION.md`](https://github.com/MemberJunction/MJ/blob/an-dev-73/guides/FILTER_BUILDER_AND_EVALUATION.md)

## Test plan
- [x] `vitest` CompositeFilter Evaluate + summary (13)
- [x] `vitest` filter-builder types/operators (41)
- [x] `vitest` UserView GenerateWhereClause (122) + EvaluateAuthorizationChecks (7)
- [x] `mj sync push` of remote-operations / api-scopes / audit-log-types
- [ ] Single-entity User View filter still saves bare field names and SQL
- [ ] Builder with `sources` persists dotted names; `CompositeFilter.Evaluate` matches a context bag
- [ ] `SummaryText` is readable in a one-line grid cell
- [ ] `Authorization.Check` via `ExecuteRemoteOperation` returns Allowed/Unknown for a known vs invented name


packages/MJCore/src/generic/filters/evaluateFilter.ts
```diff
@@ -0,0 +1,107 @@
...
+    isCompositeFilter,
+    parseFilterField,
+    type CompositeFilterDescriptor,
+    type FilterDescriptor,
+    type FilterOperator,
+} from './filter.types';
+
+export type FilterEvalContext = Record<string, Record<string, unknown> | null | undefined>;
+
+export function evaluateFilter(
```
AN-BC: We are an OOP shop, so I think we should do a class for this and the class should have static constructor methods like `FromJSON` that takes in the JSON format and builds a filter and then the class instance can have an `Evaluate()` method, we can also have helper methods on the static level. CompositeFilterDescriptor is okay to stay as a clean simple type and you can construct a new CompositeFilter from a descripter, JSON, or without and just manually add to it.

AN-BC: Also our convention is to use `PascalCase` for public members of classes as well as exported functiosn from packages

packages/MJCore/src/generic/filters/filter.types.ts
```diff
@@ -0,0 +1,61 @@
...
+    operator: FilterOperator;
+    value: unknown;
+}
+
+export interface CompositeFilterDescriptor {
+    logic: FilterLogic;
+    filters: (FilterDescriptor | CompositeFilterDescriptor)[];
+}
+
+export function isCompositeFilter(
```
AN-BC: same here we wouldn't call this `isCompositeFilter` but `IsCompositeFilter`

packages/MJCore/src/generic/filters/filterSummary.ts
```diff
@@ -0,0 +1,188 @@
...
+    type FilterOperator,
+} from './filter.types';
+
+export interface FilterSummaryField {
+    /** Stored field (`Type` or `BillToOrganization.Type`). */
+    name: string;
+    displayName: string;
+}
+
+export interface FilterSummaryOptions {
```
AN-BC: This can be moved into the new class as a method to get the summary text

AN-BC (commented):
LGTM beyond the comments already noted

AN-BC (commented):
lgtm

PR #4191, opened 2026-09-03
chore(vscode): disable search.followSymlinks in workspace settings

MarceloT-BC:
## Summary

PNPM createsd circular file paths with it’s symlinks. Turning this setting off stops VSCode’s search from traversing them infinitely and using all the CPU on the computer with RipGrep calls. It is recommended by PNPM.

Adds `"search.followSymlinks": false` to `.vscode/settings.json`.

## Test plan
- Open the repo in VS Code, run a workspace search after installing pnpm dependencies. You will not get rg processes that use all your CPU capacity.


rkihm-BC (approved):
Approving. Right file (the tracked workspace settings already carry editor-wide config), no effect on builds or tooling, and the only symlinks in the checkout that search would stop following are gitignored local artifacts (`packages/MJAPI/.env`, a Playwright profile file). The mechanism is real: VS Code's search runs ripgrep with `--follow` when this is on, and pnpm's `node_modules/@memberjunction/*` links back into `packages/` are exactly the shape that sends it in circles. No changeset needed; nothing packaged is touched.

One wording nit for the description, not blocking: I could not find this recommendation in pnpm's own docs. It comes from VS Code's troubleshooting guidance for runaway `rg` processes (https://github.com/microsoft/vscode-docs/issues/9950), so "recommended by VS Code" would be the accurate attribution.

PR #4192, opened 2026-09-03
fix(ai-agents): store agent-generated media in configured file storage, not inline

cmbrock-BC:
## Summary

`CreateMediaArtifacts` stored every agent-generated image, audio and video artifact inline as a `data:<mime>;base64,…` string in `ArtifactVersion.Content`, no matter how file storage was configured — while `ProcessFileArtifacts`, handling file outputs a few lines away, honoured the configured account. Media now takes the same route: upload to the resolved storage account when one is configured, fall back to inline when it is not or when the upload fails.

This was an omission, not a decision. The method was added (`cd8fb9c888`) to migrate media off the deprecated `ConversationDetailAttachment` table and inherited that path's inline behaviour; the storage branch had landed in its sibling six weeks earlier (`6880ed59c8`). Both call the same `createArtifactWithVersion` helper, so the two looked consistent. Nothing downstream required inline media — `gatherConversationArtifacts` and all three Angular media viewers already branch on `ContentMode === 'File'`.

The cost was silent. Measured on a production database, 35 generated images held **129 MB** of `ArtifactVersion.Content`, ~5 MB/day and growing. A representative image is 3.66 MB of inline base64 that is shipped in full on every read of the row; as a blob it is 1.36 MB, cacheable and streamable.

## Changes

- `CreateMediaArtifacts` consults `FileStorageEngine` exactly as `ProcessFileArtifacts` does, and configures the engine first — the engine reports zero accounts until it has loaded, so an unconfigured check silently routes everything inline.
- The storage-or-inline decision has one home. It lived twice in `AgentRunner`, verbatim apart from a log prefix; both paths now call `uploadToStorageIfConfigured`, so they cannot drift again. Media's File branch delegates to `createFileArtifact` rather than re-inlining its callback.
- `CreateMediaArtifacts` gains the resolved storage account as an optional trailing parameter. Existing callers are unaffected — omitting it falls back to the first active account.

## Verification

`ai-agents`: 1920/1920, builds clean. The new spec is mutation-checked: forcing the inline path fails "uploads to storage when an account IS configured", and removing the engine `Config()` call fails its own case (that mutation shipped in my first draft and was invisible to a mock that answers `HasStorageAccounts` regardless).

Verified end to end against Azure Blob Storage: a generated image landed as `ContentMode = 'File'` with a `FileID`, zero bytes in `Content`, and a 1.36 MB blob in the container.

## Test plan

- [x] With a storage account configured, generate an image and confirm the `ArtifactVersion` row is `File` with a `FileID`, and the blob exists
- [x] With no storage account, confirm the image still arrives inline and the log says `no storage accounts configured`
- [x] With a bad key, confirm the image still arrives inline and the log says `storage upload failed`, with the real error
- [x] Open a stored image in the chat preview and the full viewer

## Not in this PR

- **The audit table still holds a second inline copy.** `SaveAgentRunMedia`, which runs just before `CreateMediaArtifacts` on every run, writes `mediaEntity.InlineData = mediaOutput.data`, so each generated image is still stored once as base64 in `MJ: AI Agent Run Media`. The 129 MB figure above measured `ArtifactVersion.Content` only; the same bytes sit in the audit table. This PR roughly halves the growth rather than eliminating it. Whether that row should reference the `MJ: Files` record instead is a schema and design question — filed as #4216.
- **Existing inline artifacts are not migrated.** They stay in `Content`. `mj artifacts reclassify` exists and may cover this; I have not read it.
- `processFileOutput`'s storage branch had no prior test coverage. It is now guarded by the shared helper's tests, but nothing exercised it before this PR.


rkihm-BC (dismissed):
Approving. The defect is real, the fix routes media through exactly the path file outputs already use, and consolidating the storage-or-inline decision into one helper removes a duplicate that had already drifted once. Verified rather than assumed:

- **Red then green, locally.** All five new tests fail against the base `AgentRunner` — including the one pinning that `FileStorageEngine.Config()` runs before `HasStorageAccounts` is read, the mutation you describe from the first draft. Against the PR: all pass, the full ai-agents suite passes (106 files), `tsc --noEmit` exit 0.
- **Confirmed on `next`:** `CreateMediaArtifacts` configured only `ArtifactMetadataEngine` and always went through `DecideInlineStorage`, while `ProcessFileArtifacts` beside it configured both engines and consulted storage.
- **Fallback contract is sound.** `FileStorageEngine.UploadFile` resolves an undefined account id through `ResolveStorageAccount` and throws if nothing is configured; the helper catches and degrades to inline, as the file path already did.
- **Downstream readers handle File mode:** `gatherConversationArtifacts` downloads by `FileID`; the image/audio/video viewers and the conversation attachment service all branch on `ContentMode === 'File'`. The `data:image` comment in the same file is about detecting an image already embedded in a report payload, not about reading artifact content.
- New parameter is optional and trailing; the only in-repo caller is updated. Changeset present, `patch`, correct.

The comment on the `Promise.all` is worth having: an unloaded engine reporting zero accounts is indistinguishable from a deployment with no storage, and the test that guards it is the most valuable one in the file.

**Follow-up I'd like filed as an issue (not a condition of this PR):** the audit table still holds a second inline copy. `SaveAgentRunMedia`, which runs just before `CreateMediaArtifacts` on every run, writes `mediaEntity.InlineData = mediaOutput.data`, so each generated image is still stored once as base64 in `MJ: AI Agent Run Media` after this PR moves the artifact copy to blob storage. The 129 MB figure measured `ArtifactVersion.Content` only; the same bytes sit in the audit table. Whether that row should reference the `MJ: Files` record instead of carrying the bytes is a separate design question, but as it stands this PR roughly halves the growth rather than eliminating it, and the description could say so.

**Cosmetic, optional:** media that lands in the File branch goes through `createFileArtifact`, which hardcodes `label: 'file'`, so a failed media artifact now reads "Cannot create artifact for file" instead of "for media Image". Passing the label through would keep the diagnostics as specific as they were.

cmbrock-BC:
Both follow-ups done: the audit-table inline copy is filed as #4216 and the description now says this PR roughly halves the growth rather than eliminating it; and the label now passes through `createFileArtifact` (optional param, default `'file'`) so a failed media artifact still reads "for media Image" — pinned by a new test (`ai-agents` 1921 pass).

rkihm-BC (approved):
Re-approving after the follow-up push (my earlier approval was dismissed by it). Both items from that approval are done, and done well:

- **Audit-table inline copy:** filed as #4216, and the description now scopes the 129 MB figure to `ArtifactVersion.Content` and says this PR roughly halves the growth rather than eliminating it. Accurate.
- **Diagnostic label:** `createFileArtifact` takes a trailing `label` (default `'file'`) and the media File branch passes `media ${modality}`, so a failed media artifact still reads "for media Image". The new test pins it end to end — File branch taken, label `media Image`, version fields `File` + the file id.

The other files in the delta since my review (`base-agent.ts`, the skill tests) are the merge of `next` after #4195/#4196, not this PR. CI is fully green on this head, including Build and the deterministic integration tier.

PR #4193, opened 2026-09-03
fix(artifacts): show that a generated artifact is still loading

cmbrock-BC:
## Summary

A message with a generated image rendered as finished with nothing where the image belonged, then the image appeared unannounced a few seconds later. It read as a failed generation rather than one still in flight. Two windows had no visual state at all, and both are now covered.

Measured on a production database, the artifact link is in place at or *before* the message finalises (p50 −850 ms, p90 +80 ms across 428 links), so the visible delay is entirely the client fetching and decoding the artifact — a median inline image is 3.7 MB of base64. This PR makes that wait legible; #4192 makes it shorter. They are independent and land in either order.

## Changes

**Message level (`ng-conversations`)**
- `applyArtifactsToInstance` awaited each artifact and version row before rendering anything, though `resolveDistinctArtifacts` had already returned them synchronously from an in-memory map. Artifacts still loading are now published to the message immediately — name and visibility included, since `LazyArtifactInfo` carries both and only the entity rows are lazy — and each draws a named `mj-loading` placeholder until its card arrives.
- `LazyArtifactInfo.isLoaded` decides what counts as pending, so an artifact already in hand is never announced. The rest is filtered per artifact with `UUIDsEqual`: the two IDs come from different sources and differ in case between SQL Server and PostgreSQL, and a string compare left a placeholder sitting above the card it was waiting for.
- Three defects fixed in that path: the artifact apply now runs after the message's other inputs (it can force the child's first change-detection pass, so running it early meant `ngAfterViewInit` saw a null agent run and never started the run-duration timer); a per-instance generation counter in a `WeakMap` stops a stale load clobbering a newer one; and settle no longer touches a destroyed view.

**Preview level (`ng-artifacts`)**
- The image, audio and video previews had only `error` and `loaded` branches and drew nothing while `resolveContentUrl` settled. The loading state now lives in `BaseArtifactPreviewComponent`, cleared in a `finally`, so every media preview has it by construction.
- The image holds its indicator until the `` `load` event rather than until `src` is assigned — for a multi-megabyte inline data URI the decode is the wait — and hides the pending image with `opacity` rather than `display:none`, which would take it out of the paint tree and stop that decode. Same treatment as `ImageArtifactViewerComponent`'s `--hidden` rule.

Uses `<mj-loading>` throughout per `packages/Angular/CLAUDE.md`; no custom indicators.

## Verification

`ng-conversations` 1247/1247, `ng-artifacts` 86/86, builds clean, `check:ui` passes. The guards are mutation-checked: reverting to all-or-nothing filtering, to a case-sensitive ID match, to announcing everything regardless of `isLoaded`, or to ignoring the generation on settle each fails a test that names the regression.

Exercised live under network throttling: the placeholder held the space below the message until the image painted. On an unthrottled localhost the window is ~1 ms and nothing shows, which is the intended behaviour.

## Test plan

- [ ] Throttle to Slow 4G *after* the app has loaded, open a conversation with an image, and confirm a "Loading <name>…" row appears, then becomes the card
- [ ] Confirm a message with a loaded report and an in-flight image shows both, and the row order does not jump when the image lands
- [ ] Reload the conversation with no throttling and confirm no placeholder flashes for an already-cached artifact
- [ ] Confirm a System-Only artifact's placeholder carries the same dimmed treatment as its card

## Judgement calls

- **Placeholders are a separate element, not the real card.** `mj-artifact-message-card` has its own skeleton, and rendering the card from `artifactId` alone would reuse it — but the card then issues its own `RunView` per artifact, trading the parent's one batched load for N×2 queries. I kept the batch and accepted two loading treatments. Happy to revisit if the card grows a `pending` input.
- The generation map is lazily created. `message-list-windowing.test.ts` constructs the component off the prototype, which runs no field initializers; a plain field failed nine of its cases.


rkihm-BC (approved):
Approving. Both silent windows are real, each fix is the minimal correct one, and the three incidental defects fixed in the same path are genuine. Verified rather than assumed:

- **Red then green, locally.** 13 of the 17 new DOM cases fail against the base sources (the other four assert behaviour that already held or a getter that did not yet exist). Against the PR: `ng-conversations` 1,247 / 104 files, `ng-artifacts` 86 / 11 files, `check:ui` clean — matching the description exactly.
- **Message-level window confirmed on `next`:** `applyArtifactsToInstance` awaited every row before touching the child even though `resolveDistinctArtifacts` returns `LazyArtifactInfo` synchronously, and that type already carries `artifactId`, `artifactName`, `visibility` and `isLoaded`, so publishing the pending set costs no extra query.
- **Preview-level window confirmed:** all three media previews had only `error` and `loaded` branches. Owning `isLoading` in the base with a `finally` means a subclass cannot leave it spinning beside an error, and every preview turns a null URL into an error state, so there is no path to a permanent spinner.
- **Reordering `applyArtifactsToInstance` to run last is justified** on both the create and update paths: it can force the child's first change-detection pass, so running it before `agentRun` was assigned meant `ngAfterViewInit` saw null.
- **The generation counter closes a real race**, and the deferred-load test that settles after a newer refresh cleared the message is the one that proves it. Advancing on the clear path too is the detail a simpler version misses.
- Conventions: `mj-loading` with a valid `size` preset per `packages/Angular/CLAUDE.md`; `UUIDsEqual` for the cross-provider match; no hardcoded colours in the new CSS. Changeset present, `patch` on both packages.

The image preview holding its indicator until the `load` event, and hiding the pending element with `opacity` so the decode proceeds, is the right call for multi-megabyte inline data URIs; the note on why the host needs `position: relative` is appreciated. The judgement call to keep the parent's single batched load rather than N×2 queries from the card is the right trade.

One whitespace nit, not worth a push on its own: two blank lines were added to the audio and video preview components with no content change.

PR #4195, opened 2026-09-03
Skills: bind Scoped Search's skill principal to the run (Context.ActiveSkillIDs)

cmbrock-BC:
## Summary

Inside an agent run, the Scoped Search action's `AISkillID` was a parameter the model could write. That let a model widen its own retrieval reach by naming any skill the agent could activate for the user, whether or not the run had activated it. The run is now the authority: `BaseAgent` stamps the run's activated skills onto every action call as `Context.ActiveSkillIDs`, and Scoped Search binds its skill principal to that.

## Changes

- `BaseAgent.ExecuteSingleAction` stamps `Context.ActiveSkillIDs` on every action call (always, so an empty array means "inside a run, no skill active", distinct from "no run at all").
- New `ExecuteAgentParams.parentActivatedSkillIDs`: skills activate on the root only, so `ExecuteSubAgent` carries the root's active skills down. Without this a retrieval sub-agent's search saw no active skill.
- Scoped Search: a named `AISkillID` the run never activated is refused (`INVALID_PARAM`); exactly one active skill and none named becomes the principal; several active and none named binds no principal. Outside a run the explicit parameter is the only source, unchanged.
- Symmetric check for the agent principal: an explicit `AgentID` that differs from the run's stamped `Context.AgentID` is refused.
- `_activatedSkillIDs` becomes protected (read-only getter `ActivatedSkillIDs`); the skill-principal policy and load/veto are extracted into helpers.
- Guide §1.6b documents the contract.

## Testing

- `@memberjunction/core-actions`: 494 tests (6 new for the run binding, 3 new for the agent principal) — against the real action.
- `@memberjunction/ai-agents`: 1916 tests; the loop harness now captures the stamped `Context.ActiveSkillIDs` and asserts it, including the parent-propagation case.
- Changeset: patch (`ai-agents`, `core-actions`).

## Notes for reviewers

- Deferred on purpose: `metadata/actions/.scoped-search.json`'s `AISkillID` description still says the skill must be one the agent can activate; it should now say "active in this run". Changing metadata forces a minor bump, so it is left for a metadata-bearing PR.
- `base-agent.ts` is also touched by #4196; whichever merges second needs a trivial rebase.


rkihm-BC (approved):
Approving. Making the run the authority on the skill principal — and, symmetrically, on the agent principal — is the right model: inside a Loop agent both parameters are model-authored, so neither can be trusted to widen reach. Verified rather than assumed:

- Ran the PR's suites locally: `core-actions` 494 tests and `ai-agents` 1,916 pass; the loop harness captures the stamped `Context.ActiveSkillIDs` including the `parentActivatedSkillIDs` propagation case. (Local `tsc` on ai-agents complained only because I had not rebuilt the `ai-core-plus` dist for the new param; CI's Build covers that.)
- The refactor into `resolveSkillPrincipalID` / `loadAndVetSkillPrincipal` preserves the fail-closed veto logic verbatim, including the structural `SearchScopeAccess='None'` check and its forbidden-search log.
- "Empty array means inside a run with no active skill" vs "undefined means no run" is the right distinction and the tests pin it.
- Changeset `patch` on both packages, correct. Deferring the metadata description change to a metadata-bearing PR is the right call given the minor-bump rule.

One observation, not an ask: the explicit-`AgentID`-mismatch refusal is a behaviour change, but `AgentID` reaches this action only as a model-authored parameter — no in-repo caller passes an agent other than the running one — so the only outcomes it changes are a model misidentifying itself, which is the point.

Merge order with #4196 and #4197: `git merge-tree` shows no textual conflicts pairwise; whichever lands later needs only a trivial rebase.

PR #4196, opened 2026-09-03
Skills: filterAvailableSkills — one policy seam over every availability site

cmbrock-BC:
## Summary

An application that layers its own skill-availability policy on MJ's gates (a tenant licensing model, a per-organization entitlement, a feature flag) could only hook the requested-skill path by overriding `preActivateRequestedSkills`. The catalog the model is offered and the model-initiated `Skill` step were not overridable, so a self-activating agent could be offered a skill the policy would then refuse. This adds the one seam.

## Changes

- `BaseAgent.filterAvailableSkills(skills, purpose, agent, contextUser)` — protected, identity by default. Consulted after MJ's own gates at all four availability sites: the prompt catalog (every turn), `Skill`-step validation, `Skill`-step execution, and requested pre-activation.
- Every site goes through one private fail-closed helper: an override that throws yields `[]` at that site (empty catalog / refused activation / refused request with its note) and one logged error, never a failed prompt step or run.
- `SkillAvailabilityPurpose` type in `ai-core-plus` (`'catalog' | 'auto-activation' | 'requested'`).
- The refusal note names the policy as a possible reason.
- Guide §1.2b.

## Testing

- `@memberjunction/ai-agents`: 1919 tests. The new file drives the real `BaseAgent` (`new BaseAgent()` + a policy subclass, engine singleton mocked): identity default, intersection + refusal note, fail-closed on throw with the error logged, and the helper across purposes.
- Changeset: patch (`ai-agents`, `ai-core-plus`).

## Notes for reviewers

- `resolveSkillActivations` (a documented override point) returns the pre-policy set; `executeSkillStep` applies the policy to its result. Its JSDoc says so.
- The composer's client-side `/skill` warning checks MJ's gates only; a policy refusal surfaces as the server's note, as documented.
- `base-agent.ts` is also touched by #4195; whichever merges second needs a trivial rebase.


rkihm-BC (approved):
Approving. One protected hook, identity by default, consulted at all four availability sites after MJ's own gates, with the private fail-closed wrapper so a throwing override yields `[]` at that site and one logged error rather than a failed run. That is the right shape for an application policy seam, and the reasoning for why `preActivateRequestedSkills` alone could not do it (a self-activating agent would be offered a skill the policy then refuses) is exactly right. Verified:

- 3 of the 4 new tests fail against base (the identity-default case passes on base, as it should); with the PR, `ai-agents` 1,919 pass. The tests drive the real `BaseAgent` with a policy subclass, so the intersection, the refusal note, and the fail-closed-on-throw contract are genuinely exercised.
- The `resolveSkillActivations` JSDoc correctly warns that a subclass calling it directly gets the pre-policy set.
- Changeset `patch` on `ai-agents` and `ai-core-plus`, correct.

One observation for the guide, not an ask: a policy that throws *persistently* will log that error on every prompt turn via the catalog site. That is the correct fail-closed behaviour, but a sentence saying so would help an adopter recognise the pattern in logs.

Merge order with #4195 and #4197: no textual conflicts pairwise per `git merge-tree`; #4197 leaves the `allowed` line untouched, so the auto-merge inherits this PR's policy call and persisted skills pass the application policy too — the semantics compose correctly.

PR #4197, opened 2026-09-03
Skills: conversation-scoped activation (AISkill.ActivationScope + MJ: Conversation Skills)

cmbrock-BC:
## Summary

A skill activates for one run: `requestedSkillIDs` is a per-call input and the activated set dies with the run. That is right for a one-shot capability and wrong for a skill that behaves as a mode — a persona, or an assistant whose reply carries a menu pressed on the next turn, when nothing would re-activate it. Every conversational agent with a mode was re-implementing a (conversation, skill) table by hand. This makes it a framework feature, opt-in per skill.

## Changes

- Migration `V202609031400__v6.1.x__Conversation_Scoped_Skill_Activation.sql`: `AISkill.ActivationScope` (`Run` default | `Conversation`) and a new `ConversationSkill` table (entity `MJ: Conversation Skills`: ConversationID, SkillID, Status `Active`/`Ended`, ActivatedByRunID, EndedAt; unique per conversation + skill). CodeGen tail appended, generated on a clean database built from migrations + metadata and filtered to this change's blocks (the header lists inclusions and the excluded unrelated drift).
- `BaseAgent`: at every root run with a `conversationId`, the conversation's Active rows join `requestedSkillIDs` (one read); a `Conversation`-scoped skill that activates, by request or by the agent's own choice, gets its row written or re-activated through the run's provider; `EndConversationSkill(conversationId, skillId, user)` is the app's "leave the mode" gesture. A persisted skill a gate refuses this turn is skipped without a note and its row stays Active — retiring a mode is explicit, never a side effect of a possibly transient gate miss. All steps fail soft.
- Generated code: entity class, GraphQL schema, Explorer form (+ related Conversation Skills panel on the skill form), class registrations.
- Metadata: `Widget Guest` permission row for the new entity (parity with `MJ: Conversation Details`); `mj sync pull` configs so a skill's search scopes and an agent's skill grants round-trip.
- Guide §1.6c.

## Testing

- Clean database `MJ_6_1_CLEAN_skills_convo` built from scratch (migrate → codegen --skipfiles → sync push → codegen --skipdb); `check:codegen-tail` green.
- `@memberjunction/ai-agents`: 1920 tests. The new file drives the real `BaseAgent` with a spied `RunView`: merge, refused-persisted-no-note-no-End, fresh activation writes through the run's provider, Ended row re-activated, read failure fail-soft.
- MJ integration tier, deterministic suite, on the clean database: 50 passed; 19 client-transport tests need a hosted MJAPI + API key (not run); 1 skipped (`RUN_SEARCH_TESTS`).
- Changeset: minor.

## Notes for reviewers

- PostgreSQL counterpart deferred to the release build, per `migrations/CLAUDE.md`.
- The `EntityField` inserts use the current CodeGen emission (park existing rows at +100000, then catalog-ordinal sequences renumbered by the repeatable script). The repo's rule text still describes the older `MAX+1` form; worth updating separately.
- The skill form regeneration also carries clean-database drift beyond `ActivationScope` (`Instructions` textarea → code, the Details section folded into Configuration). Benign; any regen on a clean DB produces it.
- Precedent: `UserRoutine.RequestedSkillIDs` persists a pre-selection on a routine; this is the same idea keyed on the conversation. Complementary, not overlapping.


rkihm-BC (changes requested):
The design is sound and I would approve it with one required fix. Opt-in per skill, persisted rows merged into the request at each root run with every gate still applied, a refused persisted skill left Active rather than silently retired, `EndConversationSkill` as the explicit exit, every step fail-soft — all the right calls, and the reasoning for not ending a row on a gate miss is exactly right. Verified:

- 3 new tests fail against base; with the PR, `ai-agents` 1,920 pass and `tsc` is clean after rebuilding the core-entities dist.
- Migration: hand-written DDL is correct (default + CHECK on `ActivationScope`, PK/FKs/UNIQUE/CHECK on `ConversationSkill`, extended properties), in `migrations/v6` matching the `v6.1.x` name, CodeGen tail adds the `__mj_CreatedAt`/`__mj_UpdatedAt` system columns and registers `MJ: Conversation Skills`. `Check migrations` accepted the EntityField sequence emission.
- Changeset `minor`, as a migration requires. Entity-permissions row mirrors Conversation Details. Pull-config additions are additive.

**Required: regenerate all four bootstrap class-registration manifests.** The generated `MJConversationSkillEntity` is decorated `@RegisterClass(BaseEntity, 'MJ: Conversation Skills')`, so every bootstrap manifest must register it. On this branch only `ServerBootstrapLite` carries it, and even there `CLASS_REGISTRATIONS_COUNT` was not bumped (939, unchanged). `ServerBootstrap`, `Angular/Bootstrap` and `Angular/BootstrapLite` have no entry at all. All four include core-entities classes — see how today's #4204 diff for `AuthorizationCheckServerOperation` touched all four and eight count lines.

Why this cannot wait: the freshness gate runs only on the full-suite path (non-PR events). This PR touches no root file, so its own CI is turbo-filtered and *skips the gate* — green here is not evidence. The post-merge push to `next` runs it, and it would fail exactly the way #4185 broke `next` earlier today (fixed by #4204). Fix: rebase on `next` (so #4204's entry is present), run `pnpm run mj:manifest` **after a full build** — the source-scanning generators produce different output without one — and commit all four manifest files.

Merge order: land this after #4195 and #4196. `git merge-tree` shows no textual conflicts pairwise, and the auto-merge with #4196 is semantically right (this PR leaves the `allowed` line untouched and inherits the policy call).

Noted and fine: the skill-form regeneration carries clean-database drift beyond `ActivationScope`, as the description says; and the reads go through `new RunView()` per the file's precedent while writes use the run's provider.

cmbrock-BC:
Addressed: rebased on `next` (post-#4204), full `pnpm run build`, then `pnpm run mj:manifest` — all four bootstrap manifests now register `MJConversationSkillEntity` with their counts bumped (ServerBootstrap 1024→1025, ServerBootstrapLite 940→941, Angular/Bootstrap 653→654, Angular/BootstrapLite 539→540), committed as `3f684e3`. Also folded in the #4196 review note on persistent-throw logging (guide §1.2b). `ai-agents` 1925 tests pass on the rebased branch.

rkihm-BC (approved):
Approving — the required item is done properly and I verified it rather than took it on trust:

- **All four bootstrap manifests register `MJConversationSkillEntity`**, counts bumped by exactly one each (ServerBootstrap 1024→1025, ServerBootstrapLite 940→941, Angular/Bootstrap 653→654, Angular/BootstrapLite 539→540). The identifier-level diff against `next` across the four files nets to that one class and nothing else; the remaining hunks are entries shifting across chunk boundaries, which is what the generator produces.
- **Rebase past #4204 confirmed** (its merge commit is an ancestor; `AuthorizationCheckServerOperation` present). The PR's footprint against its merge-base is the expected 19 files.
- Folding the #4196 persistent-throw logging note into guide §1.2b is appreciated.
- **CI fully green on the rebased head**, including Build and the deterministic integration tier — which matters here since this carries a migration and the manifest freshness gate on the post-merge push to `next` is the whole point of the fix.

Merge order unchanged: after #4195 and #4196 (both merged), so this is clear to go. Branch is behind `next` again since #4018 landed this morning but mergeable.

PR #4198, opened 2026-09-03
Skills: narrow the / picker to the target agent's accepted skills

cmbrock-BC:
## Summary

The composer's `/` trigger listed every skill the user could run, including ones the agent the message goes to would refuse (`AcceptsSkills='Limited'` without a grant, or `'None'`). The refusal only surfaced after send, as a system note. The picker now narrows to what the target agent accepts.

## Changes

- `mj-ai-composer` takes `TargetAgentId`; `mj-message-input` binds `pickerTargetAgentId`: an explicit `@agent` chip already in the draft (routing's priority 1), else the agent it resolves for the message (continuity, pinned, embedder default).
- `MentionAutocompleteService.getSuggestions(…, '/', targetAgentId)` narrows the user's runnable set to `AIEngineBase.GetSkillsForAgent(agent)` via a pure `IntersectAcceptedSkills`. Intersection only — it never adds a skill the user could not run; an unknown agent means no narrowing, as before. The engine is resolved per provider when the service was initialised with one.
- Guide row updated.

## Testing

- `@memberjunction/ng-conversations`: 1235 tests; `test:types` clean. The new test imports the real narrowing function.
- Changeset: patch.

## Notes for reviewers

- The two `.mj-sync.json` pull-config additions originally drafted here moved to the conversation-persistence PR, which is already a minor, metadata-bearing branch — keeps this one a patch.
- Discovery-mode (ClassFactory-shared) provider instances keep `TargetAgentId = null`, i.e. today's behaviour; explicit-list mode has one provider per composer.


rkihm-BC (approved):
Approving. Pure intersection that can only remove, never add; an unknown agent means no narrowing; per-provider engine resolution matching the existing `GetProviderInstance` pattern used elsewhere in the Angular tree. The routing priority for the target (explicit `@agent` chip, else the resolved agent) mirrors `routeMessage`, so the picker and the send path agree. Verified: `ng-conversations` 1,252 tests pass with the PR and `tsc` is clean; the new test imports the real `IntersectAcceptedSkills`. Changeset `patch`, correct, and moving the pull-config edits to the minor branch to keep this one a patch was the right call.

One optional nit: `pickerTargetAgentId` is a getter bound in the template on a component with default change detection, and it calls `getMentionChipsData()`, which does a `querySelectorAll('.mention-chip')` over the editor on every CD cycle. The DOM is small and Angular only invokes the composer's `TargetAgentId` setter on a changed value, so the cost is bounded — but memoising on chip changes (or computing it in the chip-change handler) would be tidier.

PR #4201, opened 2026-09-03
feat(dynamic-packages): load Open App server packages in every MJ process, with per-process scoping (#4199)

### LINKED ISSUE #4199: feat(cli): Smart Server Package Bootstrap in MetadataSync / MJ CLI (invoke domain entity subclasses during sync push)
# feat(cli): Smart Server Package Bootstrap in MetadataSync / MJ CLI (invoke domain entity subclasses during sync push)

## Summary

When `mj sync push` runs against a downstream application or OpenApp (such as `more-cheese` or any workspace consuming `@mj-biz-apps/*`), entity server subclasses (e.g. `OrderEntityServer`, custom validation classes, lifecycle hooks) are never loaded into the CLI Node process. Consequently, `SyncEngine` falls back to generic `BaseEntity`, bypassing all domain validation, GL journal entry emission, and business logic.

We propose adopting the battle-tested `ServerBootstrap` pattern from `MJAPI` (`dynamicPackages.server` + host-anchored imports) inside `MetadataSync` / `MJCLI` so that `mj sync push` automatically discovers and loads the application's entity server classes.

---

### 1. How SyncEngine Operates
Inside `packages/MetadataSync/src/lib/sync-engine.ts` (lines 725 and 1051), `SyncEngine` does not issue raw SQL INSERT/UPDATE statements. It uses MemberJunction's entity abstraction:
```typescript
const entity = await this.metadata.GetEntityObject(entityName, this.contextUser);
entity.NewRecord();
// populate fields...
await entity.Save();
```
`this.metadata.GetEntityObject()` queries `MJGlobal.Instance.ClassFactory.CreateInstance<BaseEntity>(BaseEntity, className)`.

### 2. The Process Boundary & ClassFactory Fallback
* `mj sync push` runs within the `mj` CLI process (`@memberjunction/cli` + `@memberjunction/metadata-sync`).
* `initializeProvider(mjConfig)` inside `packages/MetadataSync/src/lib/provider-utils.ts` initializes the SQL connection pool and calls `setupSQLServerClient()`.
* **It never imports the host application's entity packages or server extensions.**
* In the CLI's memory, `@RegisterClass(BaseEntity, '<Subclass>')` decorators (such as `OrderEntityServer`, `PaymentHeaderEntityServer`, `MemberProfileEntityServer`) have never been evaluated.
* `ClassFactory` has no record of these subclasses and silently falls back to instantiating vanilla `BaseEntity`.
* `BaseEntity.Save()` directly calls the generated CRUD stored procedures (`spCreateOrderHeader`), completely unaware that a domain entity server with validation, totals calculation, GL journal entry emission, or entitlement fulfillment logic exists.

---

## Prior Art: How MJAPI / ServerBootstrap Solves This

`MJAPI` already solved this problem for the backend GraphQL/REST server in `packages/ServerBootstrap/src/index.ts` and `host-import.ts`:

1. **Reads `mj.config.cjs`**:
   * `codeGeneration.packages` / `entityPackageName` (via `discoverAndLoadGeneratedPackages()`).
   * `dynamicPackages.server[]` (via `loadDynamicAppPackages()`).
2. **Host-Anchored Dynamic Import (`host-import.ts`)**:
   * Uses `importFromHost(pkgName, configFilePath)` to resolve packages anchored to `mj.config.cjs` or `process.cwd()`. This allows CLI/bootstrap tools to resolve workspace packages even under strict pnpm or non-hoisted npm layouts.
3. **Startup Hooks**:
   * If a package entry defines a `StartupExport` (e.g. `LoadBizAppsOrdersServer`), it invokes it to run registration kickers.
4. **Outcome**:
   * All entity subclasses and server extensions register into `MJGlobal.Instance.ClassFactory`. When `GetEntityObject` runs in `MJAPI`, it returns the real subclass.

---

## Proposed Solution

Extend `MetadataSync` / `MJCLI` to automatically bootstrap application server packages during CLI execution:

```
┌─────────────────────────────────────────────────────────────┐
│                       mj sync push                          │
└──────────────────────────────┬──────────────────────────────┘
                               │
       ┌───────────────────────┴───────────────────────┐
       ▼                                               ▼
1. App Manifest (`mj-app.json`)            2. Config (`mj.config.cjs`)
   - Read local packages.server               - Read dynamicPackages.server[]
   - Read dependencies (e.g. orders)          - Read entityPackageName
       │                                               │
       └───────────────────────┬───────────────────────┘
                               │
                               ▼
            Import via `importFromHost()` & run StartupExports
                               │
                               ▼
          `ClassFactory` populated with Subclasses
          (`OrderEntityServer`, `MemberProfileEntityServer`, …)
                               │
                               ▼
        `SyncEngine.createEntityObject()` creates typed Subclasses!
```

### 1. Discovery Pipeline in `MetadataSync`
In `packages/MetadataSync/src/lib/provider-utils.ts` (or `SyncPushPlugin`), add a helper `loadAppServerPackages(mjConfig, configFilePath)` executed immediately after `initializeProvider()`:

1. **Generated Entities**:
   Load `mjConfig.entityPackageName` or `mjConfig.codeGeneration?.packages?.entities?.name`.
2. **Dynamic Server Packages**:
   Iterate `mjConfig.dynamicPackages?.server[]`. For each enabled entry, call `importFromHost(entry.PackageName, configFilePath)` and invoke `StartupExport` if defined.
3. **OpenApp Manifest Fallback (`mj-app.json`)**:
   If running in an OpenApp workspace where `mj-app.json` is present:
   * Inspect `packages.server` and import local server bootstrap packages (e.g. `@mj-more-cheese-demo/server`).
   * Optionally inspect declared OpenApp dependencies to discover and load installed dependent server packages.

### 2. CLI Control Flags
Because metadata push has two distinct use cases (full business object lifecycle vs. declarative raw data ingestion), provide explicit flags:
* **Default (Smart / Full Business Objects)**: Automatically loads server packages so custom `Save()`, validation, and lifecycle hooks execute.
* **`--raw` / `--bypass-server-hooks`**: Skips loading server packages. Falls back to vanilla `BaseEntity` for:
  * Restoring historical dumps / migrations where external events or journal entries must not be re-triggered.
  * Disaster recovery and bulk ingestion without side effects.
* **`--verbose` logging**: Print discovered and loaded server packages:
  ```text
  ✓ Loaded generated entities: @mj-more-cheese-demo/entities
  ✓ Loaded dynamic server package: @mj-biz-apps/orders-server (LoadBizAppsOrdersServer)
  ```

---

## Benefits

1. **Platform Consistency**: Business logic, validations, and lifecycle hooks defined on server entity classes execute whether a record is created via the UI, GraphQL API, or `mj sync push`.
2. **End-to-End Accounting & Workflows**: Pushing operational records like orders or memberships in demo or seed environments (e.g. `more-cheese`) will automatically invoke booking engines, GL account resolvers, and journal entry generation rather than creating half-populated records.
3. **Zero Configuration for End Users**: Downstream apps that already configure `dynamicPackages` or `mj-app.json` get smart CLI behavior out of the box.

AN-BC:
Closes #4199.

## Why

`mj sync push` never imported an installed Open App's server package, so `Metadata.GetEntityObject` handed back a generic `BaseEntity` for the app's entities and every custom `Save()`, validation rule and lifecycle hook was silently skipped. The asymmetry made it invisible: the CLI already imports the `server-bootstrap-lite` manifest, so MJ core's own server subclasses *did* run during a push while an app's did not.

The gap is not specific to `mj sync push`. Every process that calls `setupSQLServerClient` outside MJAPI has it: `mj app …`, `mj test`, `mj ai`, `mj querygen`, the MCP server, the A2A server, the Component Registry server, the integration-test bootstrap. And the loader MJAPI uses lives in `@memberjunction/server-bootstrap`, which depends on ~100 packages including `@memberjunction/server`, so none of those processes could import it.

## What

**New package `@memberjunction/dynamic-packages`** (no MJ runtime dependencies). `host-import.ts` and its tests move here from `server-bootstrap` unchanged; on top of it:

- `LoadDynamicPackages({ processId, tier, config, configFilePath })` loads, in ClassFactory priority order, the host's `codeGeneration.packages` (server tier: entities/actions/graphqlResolvers, never the Angular forms library), then `dynamicPackages.<tier>[]`, then the packages of an Open App whose `mj-app.json` sits next to the config (the "running inside the app's own repo" case, resolved on disk when pnpm's strict layout hides the workspace member). It runs `StartupExport`, returns a report (`Loaded` with the module, `Skipped` with a reason, `NotFound`, `Failed`), never throws for a package problem, and is idempotent per process (a repeat call returns the cached module without re-running the startup export).
- **Per-process scoping on the existing entries.** `dynamicPackages.server[]` stays the single list `mj app install` writes (add/prune/toggle code untouched). Entries gain optional `Processes` / `ExcludeProcesses`, matched against hierarchical process IDs by exact ID or ancestor segment: `cli` covers every command, `cli:sync` covers `cli:sync:push` and `cli:sync:pull`, `mjapi` / `mcp` / `a2a` / `component-registry` / `integration-tests` name the other hosts. An optional `dynamicPackages.policy` map switches a whole process off (`{ 'cli:migrate': 'none' }`, most specific key wins).
- **Kill switch.** `MJ_DYNAMIC_PACKAGES=none` (env > programmatic option > policy > default, mirroring `ResolveStartupMode`). Named honestly: core MJ subclasses still load through the manifest, so a "raw" run is raw for app entities only.

```js
dynamicPackages: {
  server: [
    // written by `mj app install` — no Processes = loads wherever the server tier loads
    { PackageName: '@mj-biz-apps/orders-server', StartupExport: 'LoadBizAppsOrdersServer', AppName: 'mj-bizapps-orders', Enabled: true },
    // only when the CLI runs any `mj sync` command
    { PackageName: '@acme/demo-seed-server', StartupExport: 'LoadDemoSeed', Processes: ['cli:sync'] },
    // everywhere except migrations
    { PackageName: '@acme/audit-server', StartupExport: 'LoadAudit', ExcludeProcesses: ['cli:migrate'] },
  ],
  policy: { 'cli:migrate': 'none' },
}
```

**Hosts wired onto it**

| Process | Process ID | Where |
|---|---|---|
| MJAPI | `mjapi` | `server-bootstrap` now consumes the loader; keeps only its MJAPI-specific post-processing (`RESOLVER_PATHS`, `MJ_SERVER_EXTENSIONS`). Existing tests pass unchanged. |
| `mj` CLI (every heavy command) | `cli:<command>` | `prerun` hook, after the lite manifest and before any command opens a provider. Global `--no-app-packages` flag consumed there like `--no-banner`. `config.ts` schema accepts the new fields. |
| MCP server | `mcp` | after `initConfig()`, before `setupSQLServerClient` |
| A2A server | `a2a` | before `setupSQLServerClient` |
| Component Registry | `component-registry` | `setupDatabase()` |
| integration-test bootstrap | `integration-tests` | after the lite manifest, before the provider |
| `mj-ai` / testing CLI standalone bins | `ai-cli` / `testing-cli` | their provider init (idempotent when driven by `mj`) |

**Detecting the silent fallback.** `mj sync push` now warns once per entity when no `BaseEntity` subclass is registered for an entity it is about to write ("…custom validation, Save() logic and lifecycle hooks will NOT run…"), on both the directory-push path and the lookup auto-create path. This is the warning that would have surfaced #4199 on day one, and it stays useful when someone disables packages on purpose.

## Not in this PR

- Walking `mj-app.json` `dependencies` for sibling apps: installed ones are already in the host's `dynamicPackages`; dev-linked ones are covered by the workspace-linking spec's env-supplied entries, which compose with `Processes` unchanged.
- The sectioned-config plan filed `dynamicPackages` under `cli`; it is a cross-process key now, so the plan's mapping table is updated to `shared` to avoid the two designs colliding.

## Verification

Unit tiers only. This environment has no database, so `mj sync push` and the other hosts could not be exercised end to end; the CLI was smoke-run (`mj --version`, `mj sync usage --no-app-packages --no-banner`) to confirm the global flag is consumed.

- `@memberjunction/dynamic-packages`: 59 tests against real on-disk packages (host-anchored resolution, process filters, policy/env precedence, manifest discovery incl. on-disk workspace members, not-found vs. transitive-failure classification, startup-export misconfiguration, idempotence).
- `@memberjunction/server-bootstrap`: existing suite unchanged, 24 pass (the loader is inlined in vitest so the mocked app packages still resolve).
- `@memberjunction/cli`: 29 pass across the new `dynamic-packages.test.ts` (CLI adapter, stdout stays clean, `--no-app-packages` stripped from argv and mapped to the env var) and `prerun-banner.test.ts`.
- `@memberjunction/metadata-sync`: 313 pass including the new `entity-subclass-guard.test.ts`.
- MCP server 131, A2A 34, Component Registry 5, testing-integration 70, ai-cli 81, testing-cli 27: all pass.
- Full dependency-chain build of 196 packages succeeded; `check:changeset` passes.

https://claude.ai/code/session_01RwbNUyCPDKaThVPZdu7ZfL

---
_Generated by [Claude Code](https://claude.ai/code/session_01RwbNUyCPDKaThVPZdu7ZfL)_


AN-BC:
This PR supersedes [MemberJunction/MJ#4199](https://github.com/MemberJunction/MJ/issues/4199) (feat(cli): Smart Server Package Bootstrap in MetadataSync / MJ CLI), which has been closed in favor of it. The issue proposed loading app server packages inside MetadataSync; this PR moves the loader into a process-agnostic package so the CLI, MCP/A2A servers, Component Registry, and test bootstraps all get the same behaviour, with per-process scoping on the existing `dynamicPackages` entries.

Reviewer notes:

- Unit tiers only. This environment has no database, so `mj sync push` and the servers were not exercised end to end. The behaviour that matters most, an app's server subclass being the one `GetEntityObject` returns during `mj sync push`, needs a run against a host with an installed Open App.
- `@memberjunction/server-bootstrap` now reads `MJ_SERVER_EXTENSIONS` by property access instead of an `in` check, so a proxied module namespace behaves the same as a plain one.
- ServerBootstrap no longer attempts to import `codeGeneration.packages.angularForms` into the Node process; it only ever logged "not found".

---

cmbrock-BC (commented):
## Test + review report: `next` + this branch

**Verdict: the feature works end to end. Four items should land before merge (the CLI config schema, the README `cli:migrate` examples, the unbuilt-member classification, and a stdout leak in the nested CLI hosts); the rest are follow-ups.**

### How it was tested

- Branch merged onto `next` @ d5a8a2ce93 (clean merge), full server-side chain built with turbo (199 packages).
- Fresh copy of a clean MJ 6.1 database on local SQL Server 2025.
- A throwaway Open App server package (`@pr4201-test/app-server`) that registers a `BaseEntity` subclass for `MJ: AI Model Types` at priority 100 whose `Save()` stamps the Description, listed in `mj.config.cjs` `dynamicPackages.server[]` together with a `Processes`-scoped entry, an `ExcludeProcesses`-scoped entry, a deliberately missing package, and a `policy` map.
- A scratch metadata directory pushed with `mj sync push --dir=… --ci` from the repo root, then the DB row inspected.

### What passed

| Scenario | Result |
|---|---|
| `mj sync push` loads the app package, runs `StartupExport`, constructs the app subclass | ✅ log shows load + startup export + "custom Save() intercepted"; DB row carries the subclass stamp |
| `--no-app-packages` / `MJ_DYNAMIC_PACKAGES=none` | ✅ mode `none (source: env)`, plain `BaseEntity` wrote the row, no stamp |
| `Processes: ['cli:sync']` under `cli:sync:push` | ✅ attempted (reported not-found since the package does not exist) |
| `ExcludeProcesses: ['cli:sync:push']` | ✅ skipped, reason `process-filter` logged |
| `policy: { cli: 'load', 'cli:sync:push': 'none' }` | ✅ most specific key wins, `none (source: policy)` |
| `MJ_DYNAMIC_PACKAGES=load` over that policy | ✅ env wins, `load (source: env)` |
| `mj-app.json` beside the config with `code.sourceDirectory` | ✅ workspace member found on disk, `package.json` `main` honoured, startup export ran |
| Subclass warning for an entity with no registered class | ✅ warning printed on the directory-push path; no false warning for a core entity that only has its generated class |
| MJAPI boot (port 4029) | ✅ "Loading Open App server packages…" → loaded + ran startup export → server-extension inventory still mounted (Slack/Teams) → Ready, 383 entities |
| MCP server boot | ✅ same load lines, up on 3100 |
| Unit suites on the merged code | ✅ dynamic-packages 59, cli 792, server-bootstrap 24, metadata-sync 313, mcp 131, a2a 34, component-registry 5, testing-integration 70, ai-cli 81, testing-cli 27 |

Not exercised: A2A and Component Registry as standalone processes (A2A's `dist/index.js` is a library entry; Component Registry fails config parsing on the plain `next` repo config identically without this PR). Their wiring is covered by the unit suites.

### Should fix before merge

1. **CLI Zod schema is stricter than the loader and rejects the README's own examples.** `packages/MJCLI/src/config.ts` still requires `AppName` on every `dynamicPackages.server[]` entry and limits `policy` values to `load|none`, while the loader treats `AppName` as optional and `parseMode` accepts `off/skip/false/0/on/true/1/full`. Reproduced: copying the README's `{ PackageName: '@acme/demo-seed-server', StartupExport: 'LoadDemoSeed', Processes: ['cli:sync'] }` into `mj.config.cjs` makes `getValidatedConfig()` throw `Database credentials are missing or empty … - dynamicPackages.server.2.AppName`, so `mj migrate`, `mj migrate create`, `mj clean`, `mj app check-updates` and `mj artifacts reclassify` abort with a misleading message (`getOptionalConfig()` silently returns `undefined`). `mj sync push` is unaffected, which is why it is easy to miss. Make `AppName` optional and align the policy enum with `parseMode`.
2. **README `cli:migrate` examples can never take effect.** `migrate`, `migrate create` and `migrate convert` are in `LIGHT_COMMANDS`, so the prerun hook never runs the loader for them; `ExcludeProcesses: ['cli:migrate']` and `policy: { 'cli:migrate': 'none' }` document scoping that has no effect, and `Processes: ['cli:migrate']` silently loads nothing. Swap the examples for a command that is heavy, or state that migrate never loads app packages.
3. **Unbuilt workspace member is reported as `Failed` + warning, not `NotFound`.** The loader's own contract (loader.ts header, `types.ts`, README) calls an unbuilt workspace member an expected not-found, but `importPackageDir` throws `Cannot find module '/…/dist/index.js'` with the *file path* quoted, so the quoted-name check misses and it lands on the warn path. Reproduced: every heavy `mj` command in an app repo prints `Error loading Open App package … Cannot find module …/dist/index.js` until the package is built, and the CLI adapter writes warns to stderr unconditionally. Treat a missing entry file under the resolved package directory as not-found.
4. **Nested `ai-cli` / `testing-cli` loader calls write to stdout and corrupt `--format=json`.** `packages/AI/AICLI/src/lib/mj-provider.ts` and `packages/TestingFramework/CLI/src/lib/mj-provider.ts` call `LoadDynamicPackages` with no `log`, so the default console logger is used and its lines go to stdout, whereas the `mj` prerun adapter routes everything to stderr. Every entry the prerun pass did not actually import (not installed, process-filtered, or policy-suppressed) is re-attempted by the nested `ai-cli` call and logged to stdout ahead of the JSON. Reproduced: with one entry scoped to `cli:sync`, `mj ai agents list --format=json` stdout begins with `Loading Open App server packages...` and does not parse as JSON. Pass a stderr logger (or `SilentDynamicPackagesLogger`) from those two hosts, and ideally skip the nested call when the `mj` prerun already ran (see the first follow-up below).

### Follow-ups (not blocking)

- **Nested hosts bypass `cli`-scoped policy.** `mj ai agents run` and `mj test run` import `@memberjunction/ai-cli` / `@memberjunction/testing-cli` in-process, whose `initializeMJProvider` calls `LoadDynamicPackages` again as `ai-cli` / `testing-cli`. The loaded-cache only covers packages the first pass imported, so an entry the prerun skipped via `policy: { cli: 'none' }` or `ExcludeProcesses: ['cli']` is loaded by the second call, startup export and all. Only the env var is process-wide. Propagate the outer process ID (env) or skip the nested call when the `mj` prerun already ran.
- **Cached-package branch runs before the `Enabled` / `MatchesProcess` / `seen` checks** (`loader.ts` ~L129), so a second call in one process reports disabled or out-of-scope packages as `Loaded` and can list a name twice; a ServerBootstrap-style consumer would then collect `RESOLVER_PATHS` / extensions twice.
- **`seen` only tracks candidates that passed the filters**, so (a) a config entry disabled with `mj app disable` is re-enabled by a same-named `mj-app.json` entry discovered later, and (b) a not-found config candidate blocks the manifest candidate (which carries the on-disk fallback) as `duplicate`. Both need the app's own package to appear in the sibling `mj.config.cjs`, so not the default layout.
- **MJAPI now also discovers `mj-app.json` beside its `mj.config.cjs`** and runs that app's startup export / collects its resolver paths. Fine, but it is a behaviour change the changeset does not mention; either document it or pass `discoverAppManifest: false` from ServerBootstrap.
- **Manifest discovery imports more than `mj app install` records.** Inside an app repo every `shared` + `server` manifest package is side-effect-imported regardless of role, while the installed form (`GetServerPackagesFromManifest`) only records entries with a `startupExport`. In-repo and installed behaviour differ; a server package that is not safe to import blind now runs in `mj sync push`.
- **`collectServerExtensions` package.json fallback is dead for path-imported workspace packages**: `resolvePackageJsonFromHost` retries the same anchors that just failed, so `memberjunction.serverExtensions` declared in package.json never mounts for a workspace member. `LoadedDynamicPackage.WorkspaceHome` is available to fix this.
- **`DiscoverMJConfig` uses cosmiconfig `searchStrategy: 'global'`** while A2A, ai-cli, testing-cli and testing-integration's own config loaders use the cwd-only default, so those hosts can load packages from a parent directory's `mj.config.cjs` that their DB settings did not come from. Passing the host's own `{config, filepath}` through (as MJCLI does) removes the discrepancy.
- **`prerun.ts` verbose detection** (`/^-[^-]+/.test(arg) && arg.includes('v')`) treats any single-dash token containing `v` as verbose (`-xv`, `-dev`). Nit; `argv.includes('-v')` next to the existing `takeGlobalFlag` calls is enough.
- **Policy values accepted by the loader are rejected by the CLI schema.** `parseMode` accepts `off/skip/false/0/on/true/1/full` for `policy` values too, but the CLI's `z.enum(['load','none'])` rejects them with the same misleading credentials error as item 1. Same fix.
- **README says `mode: none` affects app packages only; the loader also skips the host's `codeGeneration.packages`.** `MJ_DYNAMIC_PACKAGES=none` on MJAPI drops the host's own generated entities/actions too, so every host entity becomes `BaseEntity`. The prerun comment knows this, the README does not; the changeset's "no behaviour change beyond the Angular forms package" is also inaccurate given the MJAPI manifest discovery above.
- **Guard remedy text is wrong for plain custom-schema installs.** A host with custom entities but no `codeGeneration.packages` has always pushed with `BaseEntity`; it now warns per entity with advice about `mj app install` / `dynamicPackages.server` that does not apply. Mention `codeGeneration.packages` in the message, and consider not saying "records will be written" under `--dry-run`.
- **`--no-app-packages` is not in `--help`.** `--no-banner` / `--interactive` are consumed in the hook and declared in `base-cli-plugin.ts` so they appear in help; the new flag only exists in `prerun.ts`.
- **Light commands now load `@memberjunction/dynamic-packages` and cosmiconfig at startup.** `prerun.ts` statically imports the package for a string constant, and its index re-exports `loader.js`, which imports cosmiconfig at module top, contradicting the hook's own "dynamic-imported so light commands keep their instant startup" comment. Inline the literal or import from `mode.js`.
- **`sync-engine.ts` autocreate path uses bare `console.warn`** for the new warning, so it bypasses the plugin's `warnings[]` and the JSON envelope, unlike the PushService path in the same PR.
- **Package hygiene:** `packages/DynamicPackages/package.json` omits `@types/node` and `tsc-alias` from devDependencies although it uses `node:*` imports and runs `tsc-alias -f`; works only through root hoisting. `describeSource` labels config entries "Open App server package" regardless of tier.

### Confirmed fine

Prerun ordering (lite manifest → loader → command opens the provider); `--no-app-packages` stripped from the same argv reference as `--no-banner`; `getRawConfig` really is the pre-Zod object; the CLI adapter never touches stdout; `--no-app-packages` binds the nested hosts because it is env-based; warnings never affect `success` or the `--ci` exit code; the guard is safe across duplicate `@memberjunction/core` copies (matches by base-class name, memo cleared on `Register`); every new import is declared in its package.json and the lockfile is updated; changeset names and `patch` level are right; process-ID matching and policy precedence match the README; the plans doc change is benign; MCP's boot lines cannot corrupt its HTTP/SSE transport.

AN-BC:
Thanks for the end-to-end run, Colin. Your commit fd0a019b5 covered the four should-fix items; 3cce8940e (pushed on top of a merge from `next`) takes the follow-ups. Mapping each to what landed:

| Follow-up | Where it landed |
|---|---|
| Nested hosts bypass `cli`-scoped policy | `DYNAMIC_PACKAGES_PROCESS_ENV_VAR` (`MJ_DYNAMIC_PACKAGES_PROCESS`) + `EffectiveProcessId(hostDefault)`. The CLI prerun publishes its own ID; `ai-cli` / `testing-cli` bootstraps load under it, so `policy: { cli: 'none' }` and `ExcludeProcesses: ['cli']` bind the nested call too. Standalone `mj-ai` still gets `ai-cli`. |
| Cached-package branch before `Enabled` / `MatchesProcess` | Filters now run first; the cache branch only serves entries that passed them. Loader test added. |
| `seen` only tracks filtered candidates | Replaced by `mergeCandidates`: one candidate per package name in discovery order, config entry is the authority for `Enabled` and scoping, manifest `WorkspaceHome` kept as the resolution fallback, later duplicates reported as `Skipped: duplicate`. Both of your cases (disabled config entry re-enabled by the manifest; not-found config entry blocking the manifest fallback) are tests. |
| MJAPI discovers `mj-app.json` beside its config | Kept, and now stated in the changeset and the guide. |
| Manifest discovery imports shared + all server packages regardless of role | Left as is, documented in the guide's "running inside an Open App repository" section. Narrowing it to `startupExport`-bearing entries would drop the shared entities package, which is the one that registers the subclasses. Happy to revisit if you see a server package that is unsafe to import blind. |
| `collectServerExtensions` package.json fallback dead for path-imported members | ServerBootstrap now resolves the `package.json` through `WorkspaceHome` (`FindWorkspacePackageDir`) when host anchors cannot see the package. |
| `DiscoverMJConfig` `searchStrategy: 'global'` vs cwd-only hosts | `DiscoverMJConfig(searchFrom, { searchStrategy })`; A2A, ai-cli, testing-cli and testing-integration pass `'none'`. |
| `prerun.ts` verbose detection | `--verbose` / `-v` exact tokens only. |
| `--no-app-packages` missing from `--help` | Declared in `base-cli-plugin.ts` base flags. |
| Light commands statically import the loader | The env-var literal is inlined in `prerun.ts`; the loader is only reached through the existing dynamic import on the heavy path. |
| `sync-engine.ts` bare `console.warn` | Routed through a `WarningSink` the PushService sets, so it lands in `warnings[]` / the JSON envelope. |
| Guard remedy text | Mentions `codeGeneration.packages`; no "records will be written" under `--dry-run`. |
| Package hygiene | `@types/node` and `tsc-alias` declared; `describeSource` is tier-aware. |
| Policy values rejected by the CLI schema | Your commit; nothing further. |

**Registration proof.** `packages/MJCLI/src/__tests__/dynamic-packages.registration.test.ts` loads a committed miniature Open App (`fixtures/fixture-open-app`: `mj-app.json` plus plain-ESM entities and server packages that `ClassFactory.Register` a `BaseEntity` subclass) through both the `mj-app.json` path and a `dynamicPackages.server[]` entry resolved through a throwaway host `node_modules`, then asserts `GetRegistration(BaseEntity, 'MJ_Fixture: Widgets').SubClass.name === 'FixtureWidgetEntityServer'`. Nothing in the CLI's static import graph names that package. It also covers startup-export-once and `MJ_DYNAMIC_PACKAGES=none` leaving the ClassFactory untouched.

**Docs.** New `guides/DYNAMIC_PACKAGE_LOADING_GUIDE.md` (why, model, process IDs, downstream-app and MJ-install configuration, adding a host, troubleshooting), the package README rewritten as the reference with sequence / discovery / resolution / mode-precedence / nested-host diagrams, full JSDoc on `LoadDynamicPackages`, and updates to the MJAPI CLAUDE.md, MetadataSync, MJCLI and OpenApp READMEs and the Open App spec.

Unit tiers on the merged head: dynamic-packages 70, MJCLI dynamic-package suites 19, server-bootstrap 24, metadata-sync 314. ESM guard clean on the new package. Still no database here, so the end-to-end `sync push` remains your run.

---

AN-BC:
The **Check new packages exist on npm** gate is red on this PR and will stay red until a manual step is done: this PR adds a new publishable package, and npm cannot attach a trusted-publisher (OIDC) configuration to a package that does not exist yet. That setup needs an interactive 2FA challenge, so no push here can clear it. Per the gate's own handoff text:

@SDesai-BC  — new package npm setup needed before next publish:
  @memberjunction/dynamic-packages
Repo MemberJunction/MJ, workflow publish.yml, allow-publish, no environment.
Then run the "Seed new package" workflow for each.

Reference: NEW_PACKAGE_SETUP.md, and DEPLOYMENT.md Step 5. The package is intentionally not marked private; it is the loader every host now depends on and must publish with the rest of the release.

---

cmbrock-BC (approved):

AN-BC:
@cadam11  FYI - Claude tagged you in the new package comment above, that is now properly routed to @SDesai-BC not you.

PR #4203, opened 2026-09-03
fix(core-entities): generate Smart Filter WhereClause when a User View is first created

ylee-BC:
## Summary

Fixes the Data Explorer "Configure View → Filters" bug where the **Smart Filter** (AI prompt) did nothing when a view was **first created**. The view saved, but no `WhereClause` was ever generated. Editing an existing view still worked, which made the bug look environment-specific.

## Root cause

`MJUserViewEntityExtended.Save()` regenerates the `WhereClause` when the record is new or a filter field is dirty. Both signals fail for the smart filter on a brand-new record:

- **Newness was detected with `!this.ID`.** `NewRecord()` pre-assigns a UUID primary key, so a new view always has an ID before it is saved and the check is never true.
- **`SmartFilterPrompt` / `SmartFilterEnabled` are not dirty on create.** The first value written to a never-set field also seeds its `OldValue` (by design, so loaded records don't start dirty). `FilterState` *is* dirty on create, because `NewRecord()` already wrote its blank seed and the caller's write is a second write — so traditional filters were never affected by this bug.

`UpdateWhereClause()` was therefore skipped for smart filters on create: the prompt was stored as text with no AI call. The same `!this.ID` check existed inside `UpdateWhereClause()`.

## Fix

- Detect newness with `IsSaved` in both `Save()` and `UpdateWhereClause()`.
- Because `UpdateWhereClause()` now runs on every create, guard the traditional branch: on a **new** record, the empty `FilterState` seeded by `NewRecord()` must not erase a `WhereClause` the caller set directly. Programmatic view creation (`NewRecord()` → set `WhereClause` → `Save()`, without `CustomWhereClause`) is used by the view-security VS1, view-execution V8 and cache-gauntlet CG7 integration fixtures and by downstream apps. On an **existing** record a blank `FilterState` still compiles to `''` so clearing all filters removes the old one.
- A saved view whose `SmartFilterWhereClause` was never generated (e.g. created while this bug was live) is regenerated on its next `UpdateWhereClause()`. Note: the server subclass returns `''` (not null) when the model decides no filter is needed, so this intentionally does not re-fire for those.
- The `UpdateWhereClause` GraphQL query now awaits and forces regeneration (it used to race an un-awaited call against `Save()`), uses the **read-write** provider for its `Save()` instead of the read-only one, and throws if the view cannot be loaded.

## Verification

- Reproduced with a script that boots the real `SQLServerDataProvider` + MJAPI class registrations and saves a new view with `SmartFilterPrompt = "Only active models"`. Before: saved in ~200ms, `WhereClause = ""`, no AI prompt run. After: Gemini called, `WhereClause = "IsActive = 1"`, explanation populated.
- Headless client-transport repro (`GraphQLDataProvider` → running MJAPI, API-key auth): new view in smart mode → AI generated `Status = 'Active'`.
- Dirty-flag probe against the real entity on a brand-new view: `FilterState` dirty after the caller's write, `SmartFilterPrompt` / `SmartFilterEnabled` not dirty — matching the reviewer's probe.
- Unit tests: 9 in `UserViewEntity.test.ts` (new-record smart generation, hand-set `WhereClause` preserved on create, explicit traditional filter overrides a hand-set clause, blank `FilterState` on an existing record clears the clause, unchanged-prompt reuse, never-generated regeneration, changed-prompt regeneration, forced regeneration) and 4 in a new `UserViewResolver.updateWhereClause.test.ts` (read-write provider chosen, regeneration awaited before `Save()`, load failure throws, save failure throws).
- `packages/MJCoreEntities`: 642 passed. `packages/MJServer`: 1205 passed, 56 skipped.
- Mutation-gated integration bundles, `RUN_MUTATION_TESTS=1`, before vs after the guard: IT25 V8 failed (6762 unexpected rows) → passed; IT29 CG7 failed (expected 1 row, got 18) → passed; IT64 VS1 passed both times. Full `RUN_MUTATION_TESTS=1 npm run test:integration` results below.

  Full run: 66 bundles passed, 1 skipped, 3 failed (545 checks passed, 6 failed, 0 mutation-gated skips). The 3 failures are pre-existing and unrelated to views:
  - IT43 AC7 — historical prompt runs in my local DB with null Cost (recorded while its pricing metadata was stale).
  - IT86 TGB1/2/3/5 — `SQLServerTransactionGroup.executeBatchedNoVars` fails with "The variable name '@ResultChangesTable' has already been declared" on a batched submit of Tasks / Task Types.
  - IT74 TX15 — task graph settled as `Failed` instead of `Complete`.

## Not in this PR: filters are silently dropped on the default (unsaved) view — #4220

The "nothing happens" symptom for **Traditional** filters (and for Smart filters when no saved view is selected) is a separate UI bug, filed as #4220. When the entity's default/dynamic view is selected, the config panel's "Save" button calls `OnSaveDefaults()` → `onSaveDefaultViewSettings()`, which persists only column/sort/aggregate settings to user preferences and ignores `SmartFilterEnabled`, `SmartFilterPrompt` and `FilterState`. Workaround: view selector → "Save as new view" → "Customize columns, filters & sorting..." → "Create View". Excluded here because it is a Data Explorer default-view UX decision.

## Changeset

`patch` for `@memberjunction/core-entities` and `@memberjunction/server`.


packages/MJCoreEntities/src/custom/MJUserViewEntityExtended.ts
```diff
@@ -435,12 +435,14 @@ export class MJUserViewEntityExtended extends MJUserViewEntity  {
...
-            const id = this.ID;
+            // we want to preprocess the Save() call because we need to regenerate the WhereClause in some situations.
+            // NOTE: use IsSaved (not the ID) to detect a brand-new record. NewRecord() pre-assigns a UUID primary key,
+            // and the first value written to a fresh field also seeds its OldValue, so on a new record the ID is
+            // populated AND no field reads as Dirty — checking the ID would skip WhereClause generation entirely.
             const filterStateField = this.Fields.find(c => c.Name.toLowerCase() == 'filterstate');
             const smartFilterEnabledField = this.Fields.find(c => c.Name.toLowerCase() == 'smartfilterenabled');
             const smartFilterPromptField = this.Fields.find(c => c.Name.toLowerCase() == 'smartfilterprompt');
-            if (!this.ID ||
+            if (!this.IsSaved ||
```
rkihm-BC, line 446: `!this.IsSaved` is the right newness signal, but it now forces `UpdateWhereClause()` on every create. Because `NewRecord()` seeds a blank `FilterState`, the non-smart else-branch below compiles that to `''` and overwrites any `WhereClause` the caller set directly (unless `CustomWhereClause` is true). Three mutation-gated integration fixtures (view-security VS1, view-execution V8, cache-gauntlet CG7) do exactly that and were skipped in CI. See the review summary for a one-line guard scoped to new records that keeps your tests unchanged.

packages/MJCoreEntities/src/custom/MJUserViewEntityExtended.ts
```diff
@@ -540,8 +542,11 @@ export class MJUserViewEntityExtended extends MJUserViewEntity  {
                 // So, if we're here we handle the Smart Filter                
                 // we have a smart filter prompt (e.g. a prompt for the AI to create the where clause)
                 // if the SmartFilterPrompt has changed, then we need to update the SmartFilterWhereClause using AI
-                // otherwise, we don't need to do anything other than just use the SmartFilterWhereClause as it is
-                if (!this.ID || ignoreDirtyState || this.Fields.find(c => c.Name.toLowerCase() == 'smartfilterprompt')?.Dirty) {
+                // otherwise, we don't need to do anything other than just use the SmartFilterWhereClause as it is.
+                // A new record (IsSaved === false — see the note in Save() on why the ID can't be used for this) or a
+                // record that has never had its SmartFilterWhereClause generated also needs the AI pass.
+                const smartFilterPromptDirty = this.Fields.find(c => c.Name.toLowerCase() == 'smartfilterprompt')?.Dirty === true;
+                if (!this.IsSaved || ignoreDirtyState || smartFilterPromptDirty || this.SmartFilterWhereClause == null) {
```
rkihm-BC, line 550: Nice self-heal. One informational note: `MJUserViewEntityServer.GenerateSmartFilterWhereClause()` returns `''` (not null) when the model says no filter is needed, so this branch won't re-fire for those views. No change needed, just flagging so the behaviour is understood.

packages/MJServer/src/resolvers/UserViewResolver.ts
```diff
@@ -56,7 +56,7 @@ export class UserViewResolver extends MJUserViewResolverBase {
     const u = this.GetUserFromPayload(userPayload);
     const viewEntity = <MJUserViewEntityExtended>await p.GetEntityObject('MJ: User Views', u);
     await viewEntity.Load(ID);
-    viewEntity.UpdateWhereClause();
+    await viewEntity.UpdateWhereClause(true); // force regeneration regardless of dirty state — that is the whole point of this endpoint
```
rkihm-BC, line 63: Awaiting and forcing here is correct — the old un-awaited call raced `Save()`, and `Save()` won't trigger a second AI pass since no filter field is dirty. Pre-existing and optional for this PR: the provider above comes from `GetReadOnlyProvider`, so on a read-replica deployment this `Save()` writes to the replica. `GetReadWriteProvider` is the convention used by `IntegrationDiscoveryResolver` for write paths. Also worth checking the result of `Load(ID)` before regenerating.

.changeset/smart-filter-new-view-whereclause.md
```diff
@@ -0,0 +1,12 @@
+---
+"@memberjunction/core-entities": patch
+"@memberjunction/server": patch
+---
+
+Fix Smart Filter and Traditional Filter doing nothing when a User View is first created.
+
+`MJUserViewEntityExtended.Save()` detected a brand-new view with `!this.ID`. Since `NewRecord()` began pre-assigning a UUID primary key, that check is never true, and because the first value written to a fresh field also seeds its `OldValue`, none of `FilterState` / `SmartFilterEnabled` / `SmartFilterPrompt` read as Dirty on create either. The net effect was that `UpdateWhereClause()` never ran on create: the AI Smart Filter prompt was stored but no `WhereClause` was generated, and a Traditional Filter's `FilterState` was stored without being compiled to SQL. Editing an existing view still worked, which is why the symptom looked environment-specific.
```
rkihm-BC: The `SmartFilterPrompt` / `SmartFilterEnabled` part is accurate. `FilterState` is different: `MJUserViewEntityExtended.NewRecord()` already writes the blank seed, so the caller's write is a second write and *is* dirty (verified against the real `EntityField`). The old code did compile traditional filters on create; the traditional symptom was the default-view Save path you documented. Please drop the Traditional Filter claim here and in the title line above — release notes are built from this text.

rkihm-BC (changes requested):
## Review: Request changes

First off: the smart-filter diagnosis is correct and I reproduced it against the real `EntityField` in MJCore. The first write to a never-set field seeds `OldValue` (the constructor resets `_NeverSet` after writing a default, so this holds for `SmartFilterEnabled` too), `NewRecord()` pre-assigns the UUID, and the old `!this.ID` check was dead. `IsSaved` is the right signal, the awaited/forced resolver call fixes a real race, the self-heal for views created while the bug was live is a nice touch, and the out-of-scope finding on the default-view "Save" button is accurate (I verified the `OnSaveDefaults()` → `onSaveDefaultViewSettings()` path only persists grid state). Tests are genuine regression tests (they fail under the old code) and the PR write-up is excellent.

Two things need to change before merge, though. Neither was visible from CI or a default local run.

### 1. Must fix — a hand-set `WhereClause` is erased when a new view is saved

With `!this.IsSaved` in `Save()`, `UpdateWhereClause()` now runs on every create. For a view with no smart filter, the else-branch compiles `FilterState`, but `MJUserViewEntityExtended.NewRecord()` seeds a blank `FilterState`, and `processFilterGroup()` compiles an empty filter list to `''`. So any caller that does `view.NewRecord(); view.WhereClause = '...'; await view.Save();` without `CustomWhereClause = true` gets its WhereClause replaced with an empty string. I confirmed this with a probe using the PR's own test helpers:

```
PROBE new-record hand-set WhereClause after Save => ""
PROBE with CustomWhereClause after Save => "[Value] = 'tag-IN'"
```

This was the original intent of the `!this.ID` check, but that intent has been dead since the UUID pre-assignment landed in June 2025, and code written since relies on the current behaviour. Three integration fixtures do exactly this and assert on the filtered row set:

- `packages/TestingFramework/integration-test-suite/src/checks/view-security.checks.ts` (`createFixtureView`, VS1)
- `packages/TestingFramework/integration-test-suite/src/checks/view-execution.checks.ts` (V8)
- `packages/TestingFramework/integration-test-suite/src/checks/cache-gauntlet.checks.ts` (CG7)

Only `server-cache.checks.ts` sets `CustomWhereClause = true`. The CI integration log shows all three reported `SKIPPED: mutation-gated`, which is why the green run didn't catch it. Downstream app repos that create views programmatically would hit the same thing.

**Suggested fix** (keeps all your new tests unchanged). The guard must be limited to new records: on an *existing* view, clearing all filters legitimately sends a blank `FilterState` that must compile to `''` to remove the old filter.

```ts
else {
    const compiled = this.GenerateWhereClause(this.FilterState, this.ViewEntityInfo);
    // On a brand-new record an empty FilterState is just the NewRecord() seed.
    // Do not let it erase a WhereClause the caller set directly.
    if (this.IsSaved || compiled.length > 0 || !this.WhereClause) {
        this.WhereClause = compiled;
    }
}
```

Please add a unit test for the hand-set case, and run the mutation-gated checks so VS1 / V8 / CG7 actually execute:

```
RUN_MUTATION_TESTS=1 npm run test:integration
```

### 2. Must fix — the traditional-filter half of the root cause isn't supported

The mechanism you describe is exactly right for `SmartFilterPrompt` and `SmartFilterEnabled`. But `FilterState` gets its *first* write inside `NewRecord()` (the blank seed), so the caller's later write is a second write and **is** dirty. I probed this against the real `EntityField`:

```
PROBE filterstate: dirtyAfterSeed=false dirtyAfterReal=true
```

So under the old code a traditional filter on a brand-new view did enter `UpdateWhereClause()` (client and server) and did compile to SQL. The "nothing happens" symptom for traditional filters is the default-view Save bug you documented as out of scope. The fix is still worth keeping for robustness, but please drop the traditional-filter claim from the PR body and the changeset — release notes are assembled from changesets.

### Should do

- Open a GitHub issue for the default-view Save bug and link it from this PR so the analysis isn't lost.

### Optional follow-ups (pre-existing, fine to defer)

- `UpdateWhereClause` in `UserViewResolver.ts` acquires the provider via `GetReadOnlyProvider` and then calls `Save()`. On a deployment with a read replica that write goes to the replica. `IntegrationDiscoveryResolver` uses `GetReadWriteProvider` for its write paths — same convention applies here. Also worth checking the result of `Load(ID)` before regenerating.
- The server subclass returns `''` when the model decides no filter is needed, so the `== null` self-heal won't re-fire for those views. Informational only.
- A `UserViewResolver` unit test in `packages/MJServer` would be nice; there's none today.

### Verification I did

| Check | Result |
|---|---|
| PR branch `UserViewEntity.test.ts` in an isolated worktree | 128 passed |
| Real `EntityField` dirty probe (prompt / enabled flag / FilterState) | prompt and flag not dirty on create; FilterState dirty |
| Hand-set `WhereClause` probe on a new record | wiped to `''` unless `CustomWhereClause` |
| CI | all required checks green; VS1, V8, CG7 skipped as mutation-gated |

ylee-BC:
Thanks for the thorough review — both probes reproduced on my side and everything is addressed in d08594b.

**1. Hand-set `WhereClause` erased on create — fixed.** Added the new-record guard in the traditional branch of `UpdateWhereClause()` (`this.IsSaved || compiled.length > 0 || !this.WhereClause`) with a comment explaining the seed vs. "user cleared filters" distinction. New unit tests: hand-set clause preserved on a new record, explicit traditional filter still overrides a hand-set clause, blank `FilterState` on an existing record still clears the clause. Ran the three fixtures with `RUN_MUTATION_TESTS=1` before and after: IT25 V8 (6762 unexpected rows) and IT29 CG7 (expected 1, got 18) failed on the previous commit and pass now; IT64 VS1 passed both times.

**2. Traditional-filter claim — dropped.** Confirmed with a dirty probe against the real entity: after `NewRecord()` the caller's `FilterState` write is dirty, `SmartFilterPrompt` / `SmartFilterEnabled` are not. Removed the claim from the changeset, PR title, PR body, the test-file header and the `Save()` comment; the traditional symptom is attributed to the default-view Save path.

**Should do — done.** Filed #4220 for the default-view "Save" dropping filters, with the three footer states, the workaround and a suggested fix; linked from the PR body.

**Optional items — done too.** `UpdateWhereClause` resolver now uses `GetReadWriteProvider` and throws if `Load(ID)` fails; added `UserViewResolver.updateWhereClause.test.ts` (read-write provider chosen, regeneration awaited before `Save()`, load/save failures throw). Noted the `''`-vs-null self-heal behaviour in the PR body.

Full `RUN_MUTATION_TESTS=1 npm run test:integration`: 66 bundles passed, 1 skipped, 3 failed — IT43 AC7 (stale uncosted runs in my local DB), IT86 TGB1/2/3/5 (`SQLServerTransactionGroup` "@ResultChangesTable has already been declared"), IT74 TX15 (graph settled `Failed`). None touch User Views; details in the PR body.

rkihm-BC (approved):
Approving. Everything from yesterday's change request is addressed in `d08594b0`, exactly as proposed, and the optional items were done as well.

**Must-fix 1 — hand-set `WhereClause` on create:** the traditional branch now guards with `this.IsSaved || compiled.length > 0 || !this.WhereClause`, with a comment on the seed-vs-cleared distinction, and three new tests cover it (hand-set clause preserved on a new record; explicit traditional filter still overrides; blank `FilterState` on an existing record still clears). Running the mutation-gated fixtures before/after — V8 and CG7 failing on the previous commit and passing now, VS1 passing both times — is the verification I asked for.

**Must-fix 2 — traditional-filter claim:** dropped from the title, changeset, test header and `Save()` comment; the body now attributes that symptom to the default-view Save path, filed as #4220.

**Optional items:** `UpdateWhereClause` uses `GetReadWriteProvider` (with the `allowFallbackToReadOnly` option that helper supports), throws on a failed `Load`, and awaits the forced regeneration before `Save()`; the new `UserViewResolver.updateWhereClause.test.ts` pins provider selection, the await-before-save ordering, and both failure paths. The `''`-vs-null self-heal note is in the body.

**Verified on this head:** `UserViewEntity.test.ts` 131/131 and the full `core-entities` suite 642/642 in a worktree with the dependency chain built; CI fully green; the delta since the last review touches only the five expected files; changeset `patch` on both packages with accurate text.

Thanks for the thorough turnaround.

PR #4207, opened 2026-09-04
perf(integration-engine): the hash-prefetch must bypass the RunView result cache

MS-BC:
## The defect

`PrefetchContentHashes` issues one RunView per apply batch, filtered to **that batch's own keys**. Its cache fingerprint is therefore unique on every call — a cached result can never be hit again. With a result cache active, every batch deposits one permanently-dead cache entry, so process memory grows **O(records processed)** for the life of the process.

Measured consequence in production: a full-history first sync (~500k records, ~50 minutes in) exhausted a default node heap — event loop starved, health endpoint stopped answering entirely, SSM couldn't schedule a shell, and the host needed a hypervisor reboot. The apply pipeline itself is batch-scoped by design (prefetch map, MatchEngine, RecordMapBatch all die with their batch); this cache deposit was the accumulator.

## The fix

One flag: `BypassCache: true` on the prefetch RunView — exactly the flag the match lookups in this same path already pass (they bypass for correctness; the prefetch bypasses for survival). Comment carries the reasoning next to the existing `IgnoreMaxRows` guard it mirrors.

## Tests

Twin test beside the `IgnoreMaxRows` guard in `PrefetchProvesAbsence.test.ts`, using the suite's existing param-capture harness. **Mutation-verified**: removing the flag fails the new test; restoring it passes 12/12. Full package suite: no new failures vs clean HEAD baseline (identical pre-existing set on both).

### DIFF
--- a/.changeset/prefetch-bypass-runview-cache.md
+++ b/.changeset/prefetch-bypass-runview-cache.md
@@ -1,0 +1,5 @@
+---
+"@memberjunction/integration-engine": patch
+---
+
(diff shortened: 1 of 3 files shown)


MS-BC:
Withdrawn for now — will be re-raised as part of a coordinated batch.

PR #4208, opened 2026-09-04
fix(integration-engine): the empty-batch watchdog must see an unmoving cursor

MS-BC:
## The defect

The P3-D watchdog warns after N consecutive empty batches with `HasMore=true` — but "empty" alone is the wrong signal. A per-item fan-out connector (one upstream call per parent record) legitimately returns long runs of empty batches while its position walks forward over sparse data. Measured live on such a connector: dozens of false `CONSECUTIVE_EMPTY_BATCHES` / "likely a pagination bug" warnings per run, every one spurious — which trains operators to dismiss exactly the message a *real* stuck cursor would produce.

## The fix

A stuck cursor, by definition, leaves the position where it was. The streak now counts only empties whose position tuple — `(watermark, afterKey, page, offset, cursor)` — did not move; any component advancing resets the streak. `null`/`undefined` components compare as identical (the loop's position vars mix both).

The decision is extracted into a pure `EmptyBatchWatchdog` class (the fetch loop isn't unit-harnessable), used at the same call site, same warning code, message updated to say what is now actually known ("at an UNMOVING position … the cursor is not advancing").

## Tests

6 behavior tests on the pure class, including the fan-out shape that produced the false alarms (50 empty batches, offset always advancing → streak never exceeds 1) and the true-positive (position frozen → streak climbs). Package tsc error count identical to clean HEAD (16 pre-existing, none in the new code).

### DIFF
--- a/.changeset/empty-batch-watchdog-cursor.md
+++ b/.changeset/empty-batch-watchdog-cursor.md
@@ -1,0 +1,5 @@
+---
+"@memberjunction/integration-engine": patch
+---
+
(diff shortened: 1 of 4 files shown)


MS-BC:
Withdrawn for now — will be re-raised as part of a coordinated batch.

PR #4212, opened 2026-09-04
feat(rich-text-editor): DOM-native rich text editor package + base-forms HTML field adoption

bc-izygmunt:
## Summary

Introduces `@memberjunction/ng-rich-text-editor`, a DOM-native WYSIWYG editor built to replace Quill in Izzy's message editor, and adopts it in `@memberjunction/ng-base-forms` for HTML fields.

The editor has **no internal model and no schema**. The `contenteditable` DOM *is* the document, and normalization runs at exactly three boundaries: `SetHTML`, paste, and the local seams of the command being executed. HTML the user does not touch (Outlook reply chains, `<o:p>`, conditional comments, layout tables, inline styles) survives load → edit → serialize. That property is enforced by a fixture suite and a real-browser tier, and it is the reason the package exists: Quill, and every model-based editor, rewrites the whole document on load, which is why Izzy's diff view needed the `normalizedOriginal` and `cleanNbspForSave` workarounds.

## What's in the PR

**Engine** (`src/lib/engine`, zero Angular imports)

- Node model with content-computed categories, range primitives, split/merge, and a DOMPurify sanitize boundary with `strict` and `email` profiles (P0/P1, from July).
- Undo as pure snapshots: raw HTML plus index-path selection. The loaded document is the first entry; each command checkpoints before and after; native typing coalesces; IME compositions checkpoint at `compositionstart`.
- Inline format engine that treats `B`/`STRONG`, `I`/`EM`, `S`/`STRIKE`/`DEL`, `CODE`/`TT`/`KBD`/`SAMP` as the same format, because tags are never rewritten on load. Pending formats at a collapsed caret, format escape that preserves other formats.
- Block format engine: quotes, lists, headings via extract → pure transform → reinsert → merge seams. List nesting works in place and always puts the sublist inside the previous item (valid markup; Squire emits `<ul>` inside `<ul>`).
- Keyboard: Enter, Shift+Enter, Backspace, Delete, Tab, Space with autolinking; table-driven keydown and `beforeinput` routing so virtual keyboards, the macOS Edit menu, and assistive tech get the same behaviour.
- Clipboard: copy and cut write HTML with context plus plain text; paste runs the full clean pipeline and merges into the caret block; plain-text paste, Ctrl/Cmd+Shift+V, drop, images handed to the host. Forward Delete is engine-owned and grapheme-aware (`Intl.Segmenter`).
- `RichTextEngine`: PascalCase command surface, typed events, `Destroy`.

**Angular** (standalone, `mjUILayer: widgets`)

- `<mj-rich-text-editor>`: `ControlValueAccessor`, `Html`/`MinHeight`/`MaxHeight`/`ImagePaste` inputs, placeholder, `Disabled`/`ReadOnly`, `BeforePaste`/`AfterPaste` cancelable event-args pair per the UI layering contract, `PasteImage`, Ctrl/Cmd+K, click below a trailing quote to add a line.
- `<mj-rich-text-toolbar>`: token-styled native buttons, one Tab stop with arrow-key navigation, `aria-pressed` driven by the document.
- `<mj-rich-text-link-editor>`: inline popover (not a modal), Remove → Apply → Cancel.

**`ng-base-forms`**

- `ExtendedType = 'HTML'` fields edit visually with a Visual / Source `<mj-view-toggle>`. A value that is a whole document (`<html>`, `<head>`, doctype) opens in Source, because a visual editor can only hold a body fragment. Markdown and Code fields are unchanged.
- **Behavior change:** long-text fields with no `ExtendedType` that already auto-detect as HTML previously opened in the raw code editor and now open in the visual editor.

## Deliberate deviations from the reference architecture (Squire)

- Tag rewriting (`STRONG`→`B`) is paste-only; the load path never sweeps.
- No automatic "ensure bottom line": it would append a blank line to a quoted thread on every edit above it. Replaced by an explicit click-below affordance.
- `mjButton` is not used for toolbar buttons because the directive owns `aria-pressed` and flips its own state on click, which conflicts with document-driven pressed state. The reason is documented in the component.

## Testing

| Tier | Result |
|---|---|
| Engine + DOM specs (jsdom, axe scans) | 625 passed |
| `ng-base-forms` | 332 passed, 4 new for the HTML branch |
| Browser tier, `pnpm run test:live` (Playwright, Chromium) | 16 passed |
| `test:types`, `test:live:types` | clean |
| `check:standards`, CSS hex gate, `.mj-btn` gate, DOM-spec placement, focus-ring, ESM guard, `check:changeset` | pass |

The browser tier found three behaviours jsdom passed and they are fixed in this PR: commands now checkpoint their result so typing after Bold undoes separately; the loaded document is recorded as the first undo entry; a redundant trailing `<br>` after Shift+Enter is accepted as Chromium removes it.

`pnpm run playground` serves a manual smoke-test bench with a semantic diff against the loaded document, including an Izzy-shaped sample (AI reply above a quoted Outlook thread). The integration tier was not run: the change has no server, entity, or database surface.

## Alternatives evaluated

- **Tiptap**: schema-based (ProseMirror). Across the 11 fidelity fixtures it produced 37 semantic differences and dropped Outlook conditionals, `<o:p>`, inline styles, and `<style>`; ours produces 0. Rejected on the requirement.
- **TinyMCE**: DOM-native and matches fidelity with `valid_elements: '*[*]'` and `verify_html: false`, but that disables its sanitizer (XSS passes through) and the license is GPL v2+ or per-deployment commercial, which MJ cannot carry under BUSL. Rejected on licensing.

## Follow-ups (not in this PR)

- Before the first release including this package: register it on npm per `NEW_PACKAGE_SETUP.md` (`npx setup-npm-trusted-publish @memberjunction/ng-rich-text-editor`), or the publish workflow fails early.
- Izzy message-editor swap (Izzy repo) once a 6.1.x with this package is published.
- WebKit/Firefox projects in the browser tier (`PW_BROWSERS=all`), real IME and mobile verification.

Changeset: `patch` for both packages (no migrations, no metadata).


AN-BC:
I'd like to have @MattC-BC and @rkihm-BC rveiew this one.

My only comment is the name "rich text" is a misnomer as this is an HTML editor, only, right?

Rich text is actually a format name that is largely not used anymore, RTF was a PC file format that you could use to interchange files between word, and other editors back in the day.

Anyway, aside from that side note it just seems like HTML editor is more apt here?

MattC-BC:
Thanks for this one. The architecture is the right call and it shows: the Angular-free engine, the load/paste asymmetry being written down as a contract, token-clean CSS, no `any`, the Before/After paste pair, and fixtures that record known limits instead of hiding them. I want this to land. But I went through the engine and the base-forms adoption in detail and there are a handful of things I'd block on, plus a longer list that should be fixed before it ships to every HTML field in Explorer.

Everything below was verified against the real modules (jsdom for the engine, a real Chromium session for the selection shape in the first item, DOMPurify 3.3.1 for the sanitize items), not just read.

## Blockers

**1. Select-all + Backspace can delete DOM outside the editor.**
`engine/range/boundaries.ts:134` — `moveRangeBoundariesUpTree` stops only at `startMax`/`endMax`, never at `root`. `deleteContentsOfRange` (`range/contents.ts:127`) passes *descendant* blocks as those maxima, so a root-indexed range climbs out of the editor and `split()` runs on page ancestors. In Chromium, Ctrl+A in a document ending in `<hr>`, a root-level ``, or a `<table>` yields exactly that shape (end boundary at `(root, 2)`). With the surface attached to a page, Backspace throws `HierarchyRequestError`, the editor is unchanged, and a sibling element *outside* the editor gets removed. Delete, typing over the selection, Enter, Space and paste all take the same path. The suite is blind to it because `contents.test.ts` builds a detached root, so the climb stops on the first hop. Fix is the Squire guard: stop at `root` in both loops, and add one attached-root spec.

**2. base-forms loads every HTML field through the strict profile and persists the loss on the first keystroke.**
`form-field.component.html:387` mounts `<mj-rich-text-editor>` with no `[Config]`, so `SetHTML` runs the `'strict'` profile that the engine designed for untrusted paste. Against the real config: `<style>…</style>`, `<!--[if mso]>…<![endif]-->` and `<o:p>` are gone; `target="_blank"` is stripped under *both* profiles (DOMPurify's default attribute list omits `target` and nothing passes `ADD_ATTR`); `{% if a < b %}` in template text comes back as `{% if a &lt; b %}`. None of that matches `FULL_DOCUMENT_PATTERN`, so the field opens in Visual, and the first keystroke writes `engine.GetHTML()` back through `OnRichTextChange`. The record silently loses the markup on save. And because `RichTextMode` falls through to `detectRichTextFormat`, this reaches any long text field that merely *looks* like HTML, including `Template Contents.TemplateText`. My recommendation: only open Visual for explicit `ExtendedType === 'HTML'`, pass `SanitizeProfile: 'email'` from base-forms, add `target` to the load allow-list, and leave auto-detected fields on Source by default.

**3. List/quote toggle inside an empty container throws and wipes the document.**
`engine/format/block.ts:64` guards the end offset on the line above but not the start offset, so an empty transform result passes `undefined` into `mergeContainers` (`merge-split.ts:296`). Caret in `<ul></ul>` or `<blockquote></blockquote>` (these survive load because load never sweeps) + the toolbar button = TypeError, root left as `""`, and Undo disabled because `DocumentChanged` never ran. Related: `runCommand` (`editor.ts:650`) has no try/finally, so any throw between `SaveUndoState` and `DocumentChanged` leaves `inUndoState` true and the next command overwrites the good pre-command snapshot in place. `handleEnter` has the same shape around a reachable throw at `split-block.ts:36`.

**4. Table cells corrupt on Enter and on paste; Backspace after `<hr>` loses the line.**
- `keyboard/enter.ts:55` → `split-block.ts:49`: `TAG_AFTER_SPLIT` has no `TD`/`TH`, so Enter in a cell retags the right half as `<div>` inside the `<tr>`. Browsers foster-parent it out on the next parse.
- `clean/brs.ts:49`: `splitBlockOnLineBreaks` has no skip list, so a pasted `<td>line1<br>line2</td>` (what Word/Excel/Outlook emit) becomes two `<td>`s, an extra column. Same for `<pre>` and `<blockquote>`.
- `keyboard/backspace.ts:50`: `<hr>` classifies as `'block'` (childless, not in `INLINE_TAGS`; `LEAF_TAGS` is never consulted by `category.ts`), so Backspace at the start of the next line merges its text *into* the `<hr>`, and void elements serialize without children. The line is gone from `GetHTML()`.

## Should fix before merge

- `engine/node/block.ts:111` — `fixContainer` wraps whitespace-only text nodes into `<div>`s on the *load* path, so pretty-printed HTML gets one junk `<div>` per indentation run, and `a<br>b` at root loses the `<br>`. It also descends into `<svg>` (lowercase `nodeName` never matches the uppercase skip set) and injects HTML `<div>`s inside it. This breaks the `SetHTML(GetHTML(x))` fixed point the pipeline header promises.
- `engine/range/block-range.ts:69` — `lastBlockAtOrBefore` steps *backwards* past a container instead of descending into it, so select-all + Quote/List/Heading over `<p>a</p><ul>…</ul>` silently leaves the list out. This wrong end block is also what feeds blocker 1.
- `engine/range/containment.ts:28-29` — strict `< 0` / `> 0` where Squire uses `<=` / `>=`, so touching nodes count as contained. Select just an image and press Bold: the neighbouring text gets bolded too, and `hasFormat` reads wrong for the toolbar.
- `engine/node/category.ts:81` — any element not in `INLINE_TAGS` is `'block'`, so `<o:p>`, `<svg>`, `<video>` turn their paragraph into a container and Enter/Backspace go dead inside it. That's every Outlook paragraph under the email profile.
- `keyboard/backspace.ts:101` — `href.includes(text.data)` treats any link whose text appears in its URL as an autolink, so Backspace at the end of `<a href="…/document/d/abc">document</a>` strips the whole link.
- `rich-text-editor.component.ts:333` + `form-field.component.html:393` — `ImagePaste="data-uri"` inlines pasted images with no size, MIME or field-length cap. The Image branch 250 lines away already has `MaxStoredImageChars` and compress-to-fit; this path should go through it (or the editor needs a cap input). Also `clipboard.ts:113`: when the clipboard has both an image file and HTML (every Office copy), the HTML wins, its `file:///` src is stripped, and the image bytes are dropped.
- `engine/html.ts:77` — `getHTML` strips the ZWS text node but not the emptied inline, so an unused Bold/Link click emits `<b></b>` / `<a href="…"></a>` as a user change and dirties the record. `zws.ts` already has `removeWithEmptiedInlineAncestors` for exactly this. On a never-focused editor the same click lands at document start.
- `engine/editor.ts:64` — `NeedsTrailingLine` checks `lastElementChild.nodeName` against a seven-tag set, but Outlook and Gmail wrap the quote in a `<div>`, so the click-below affordance never fires for the exact documents it exists for (including the package's own headline fixture).

Smaller, also verified: Ctrl/Cmd+Shift+7/8 never fire (`event.key` is `&`/`*` with Shift, `keys.ts:11`), and the shortcut/beforeinput list intents always *make* a list while the button toggles; `plainTextPasteRequested` (`editor.ts:732`) is only cleared by a paste event, so an unbound Cmd+Shift+V silently downgrades the next paste; `normalizeHref` (`toolbar-config.ts:111`) passes `javascript:` through into stored HTML (read path sanitizes, email send would not); autolink strips a trailing `)` from Wikipedia-style URLs (`links.ts:14`); undo paths are serialized against a DOM with adjacent text nodes that the re-parse coalesces (`editor.ts:676`); `mergeWithBlock` (`merge-split.ts:248`) eats a user's Shift+Enter `<br>`, not just the filler; an empty field persists as `<div><br></div>` and satisfies required-field validation; toolbar arrow keys strand on disabled Undo/Redo (`rich-text-toolbar.component.ts:244`); the read-only surface has no `tabindex` so it drops out of the tab order.

## Perf

`editor.ts:794` registers a *document-level* `selectionchange` listener inside NgZone (no `runOutsideAngular` anywhere in the package). With one HTML field open, every caret move or selection drag anywhere in Explorer schedules a full app change-detection pass. Please wrap the engine construction in `runOutsideAngular` and re-enter only where Angular state changes. Also: two whole-document clone+serialize passes per keystroke (the `input` handler and the `[Html]` setter's comparison), and arrow keys sweep every text node for ZWS with no short-circuit.

## Cleanup

`UriPolicy` (`types.ts:118`) is never read and its docs contradict `sanitize.ts:14-18`. `BLOCK_TAGS` has zero production consumers. The link editor's `<input>` re-implements `.mj-input` property for property while its buttons already use `mjButton`; please use `class="mj-input mj-rte-link-editor-input"` and keep only overrides. `detectMacPlatform` is the sixth private isMac check in `packages/Angular` (and answers differently for iPhone than the other five). Per `UI_LAYERING_GUIDE.md` §6 the four informational event payloads should be classes not interfaces, and the link editor's `Apply`/`Remove`/`Cancel` outputs should be `<Noun><Verb>Requested`. Duplicate fixture id `F10` in the base-forms DOM test.

## CI and naming

CI is red on the merge commit, but that's inherited: `806e7f2dc3` on `next` added `@memberjunction/sql-parser` to MJServer without updating the lockfile, so `pnpm install --frozen-lockfile` fails everywhere. Your pre-merge commit was green apart from the new-package npm gate, which stays red until the seed workflow runs. A `pnpm install` at root on this branch clears the lockfile part.

On Amith's naming point: I agree it's an HTML editor, and this is the only moment a rename is free. Package name, the three selectors, the `RichText*` type prefix, the `.mj-rte-*` CSS prefix, the changeset and the README row are all mechanical now and a breaking change after first publish.

Happy to pair on any of the range-layer items if useful; the fixes are mostly one-liners once the shape is clear, the hard part was finding them.

bc-izygmunt:
> I'd like to have @MattC-BC and @rkihm-BC rveiew this one.
> 
> My only comment is the name "rich text" is a misnomer as this is an HTML editor, only, right?
> 
> Rich text is actually a format name that is largely not used anymore, RTF was a PC file format that you could use to interchange files between word, and other editors back in the day.
> 
> Anyway, aside from that side note it just seems like HTML editor is more apt here?

Yes, for v1 it is HTML editor. I plan to expand this for other formatting like markdown. The original reason I named it rich text editor was because it looked like the convention across WYSIWYG HTML editors. There is a placeholder for @memberjunction/ng-rich-text-editor but I can create a new one to be @memberjunction/ng-html-editor if thats the direction we want to pursue.

PR #4213, opened 2026-09-04
fix(ng-filter-builder): re-export camelCase aliases dropped in #4185

AN-BC:
## Summary
- `#4185` renamed public filter-builder helpers to PascalCase (`CreateEmptyFilter`, `CreateFilterRule`, `IsCompositeFilter`, operator helpers).
- Compiled consumers that still import `createEmptyFilter` (including a stale `ng-entity-viewer` dist) fail Explorer boot with `No matching export … createEmptyFilter`.
- Re-export the old camelCase names as aliases of the PascalCase functions so both spellings resolve.

## Test plan
- [x] `pnpm test` in `@memberjunction/ng-filter-builder` (44 passing)
- [x] Rebuild filter-builder + entity-viewer; Explorer `ng serve` on :4201 compiles
- [ ] Confirm Explorer loads at http://localhost:4201 against MJAPI :4103

### DIFF
--- a/.changeset/filter-builder-camelcase-aliases.md
+++ b/.changeset/filter-builder-camelcase-aliases.md
@@ -1,0 +1,5 @@
+---
+"@memberjunction/ng-filter-builder": patch
+---
+
+Re-export camelCase aliases (`createEmptyFilter`, `createFilterRule`, `isCompositeFilter`, and the operator helpers) alongside the PascalCase names from #4185 so existing compiled consumers keep resolving.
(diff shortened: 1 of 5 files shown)


AN-BC:
Closing: aliases were the wrong fix. Public helpers stay PascalCase (`CreateEmptyFilter`, `IsCompositeFilter`, …). Downstream camelCase callers will be updated instead. MJ source on `next` is already PascalCase; Explorer boot was a stale `ng-entity-viewer` dist.

PR #4214, opened 2026-09-04
chore: finish the Kendo removal + design proposal for pluggable client analytics

AN-BC:
Two related pieces of work. Happy to split if reviewers prefer — say the word.

---

# 1. Finish the Kendo removal (code)

Kendo was already out of the dependency graph: **zero `@progress/kendo*` entries in any `package.json`, zero TypeScript imports.** What remained was residue — and some of it was not inert.

### Dead code and styles removed

- **82 `.k-*` CSS rules across 21 files.** Nothing has defined those classes since the removal, so every rule was a no-op.
- **`li.k-drawer-item.k-level-0` was still being queried** in `MJExplorerAppComponent` and `InitializationService` to auto-select the first nav item when landing on `/`. The Kendo Drawer is long gone, so `querySelector` returned `null` and the branch did nothing — removing it is a **runtime no-op**. ⚠️ The *behaviour* has been silently absent since the removal; if landing on `/` should select the first nav item, that needs reimplementing against the current shell nav.
- `dialog.service.ts.bak` — a dead backup file importing `@progress/kendo-angular-dialog`.
- `kendo-ui-license.txt` dropped from `DistributionAssembler`'s ignore list, and the test asserting that exclusion retargeted so it still tests something real.

### Broken templates fixed

Three components still applied Kendo utility classes that no longer resolve, so they were rendering wrong:

| Component | What broke |
|---|---|
| `add-item` | `k-actions k-actions-end` was supplying the footer's flex row, so `.customBtn button { flex: 1 }` had no flex container. `.popup-actions` now provides it. |
| `delete-item` | `k-m-7.5 k-text-center` **never applied at all** — `k-m-7.5` is not a valid single class token — so the confirmation copy has been unstyled since before the removal. Real `.confirm-message` class added. |
| `add-item`, `ng-data-context-dialog` | Each carried `<div class="k-overlay">` as its backdrop. `mj-window` deliberately has no backdrop, so these were unstyled empty divs. **Deleted — which means both dialogs currently render with no backdrop.** Restoring one is a UX call, not a cleanup. |

### The filter payload is MJ's own type

`FilterOperator` / `FilterLogic` / `FilterDescriptor` / `CompositeFilterDescriptor` and their guards were **already** defined in `@memberjunction/core` (`generic/filters/filter.types.ts`). `ng-filter-builder` carried a byte-equivalent duplicate set and re-exported it. The duplicate is gone; the builder and `ng-entity-viewer` import the shapes from core, and the builder exports only its own UI types.

> ⚠️ **API change — `@memberjunction/ng-filter-builder`**: `FilterOperator`, `FilterLogic`, `FilterDescriptor`, `CompositeFilterDescriptor`, `IsCompositeFilter` and `IsSimpleFilter` are **no longer re-exported**. Import them from `@memberjunction/core`. This brings the package in line with the repo's no-cross-package-re-exports rule.

### Docs and comments

33 package READMEs listed `@progress/kendo-angular-*` as **live dependencies** (they aren't) and drew Kendo components in their mermaid diagrams. `packages/Angular/CLAUDE.md`, four guides, and ~55 source comments referenced Kendo as though it were still present. All rewritten to describe what the code actually uses.

Historical records are deliberately untouched — `plans/`, CHANGELOGs, and dated retrospectives (`FINAL_REPORT.md`, `PIXEL_PERFECT_COMPLETE.md`, `USER_MENU_PROPOSAL.md`, timeline `PRD.md` version history) still mention Kendo, because they record when it was there.

### One deliberate exception

`explorer-app/src/lib/styles/_common.scss` keeps its 12 dead `.k-*` rules. Both UI gates check whole files once changed, so any edit there makes CI evaluate the file's pre-existing debt. The 15 hardcoded colors were fixable and I did fix them — but `.mj-btn` is **defined** in that file with four legacy `.mj-btn-*` classes `@extend`ing it, and `check:ui-buttons` permits exactly one canonical definition with no allowlist. Removing that block is a migration, not a cleanup, and unrelated to Kendo. Reasoning is recorded in commit `3c1d6f2fb`.

Useful for whoever picks it up: **zero templates in `packages/Angular` still reference the legacy `mj-btn-*` classes**, so the block looks dead here — but it ships in MJExplorer's global stylesheet, so downstream apps may still use it. That needs confirming before deletion.

### Verification

- **223/223 packages build clean, 0 TypeScript errors.**
- Tests pass for every changed package: ui-components 565, MJCore 2183, MJCoreEntities 640, base-forms 328, core-entity-forms 284, explorer-core 167, entity-viewer 153, filter-builder 43, installer 27, bootstrap 9.
- CodeGenLib's 3 failures in `entity-subclass-embedded.test.ts` are **pre-existing** — verified by running that file against unmodified `HEAD` source.
- `check:ui` all four gates green. `check:changeset` green (patch — no migration, no metadata).

---

# 2. Client Analytics Providers — design proposal (docs only)

`guides/CLIENT_ANALYTICS_PROVIDERS_GUIDE.md`. **No code.** A proposal for a first-class, vendor-neutral seam so a deployment can plug in Hotjar / PostHog / Clarity / GA4 / Matomo — reviewed before anyone builds it.

The framing worth arguing with: **MJ ships the seam, never the vendor.** Each customer configures their own vendor account in their own estate; there is no channel that sends anything to MemberJunction, and adding one is out of scope. That sidesteps the subprocessor question shipping a tag in the bundle would create for customers holding member PII.

- **Contract & registry** — `BaseAnalyticsProvider` (6 methods) via the existing `@RegisterClass`/`ClassFactory`. L0 pure-TS + L3 Angular taps per the UI layering guide; vendor adapters are separate opt-in packages so the default bundle carries no vendor SDK.
- **Config & consent reuse what exists** — `InstanceConfigEngine` per deployment, `UserInfoEngine` per user. Off by default, fail-closed, mirroring the realtime-capture posture.
- **Route scrubbing** — the load-bearing safety piece. MJ's routes carry record IDs, and `resource/search/:searchInput` carries the user's **search text**; in an AMS that's member names, and vendors collect page URLs regardless of recording settings. The guide states the limitation honestly (a vendor script can read `location.href` itself) and recommends fixing the route as a prerequisite rather than pretending the scrubber closes it.
- **Mask-by-default** — vendors mask `<input>` values, not rendered text. Key insight: put masking on `MjFormFieldComponent` and the shared grid, **not** in CodeGen templates, or it's wiped on the next regeneration.
- **Why a seam beats a script tag** — `mjButton` (715 template usages), `mjClickable`, `MjFormFieldComponent` and the existing `NavigationEnd` taps let MJ hand a vendor *meaning* where a raw tag sees only DOM.
- **7 open questions** for the team to settle.

### Gaps found while writing it

Each independently actionable, none blocked by the proposal:

1. **No global Angular `ErrorHandler` anywhere in `packages/**`** — uncaught client errors go to the console and nowhere else. Proposed as Phase 0: useful with no vendor configured, and can write to `MJ: Error Logs` today.
2. No CSP in `packages/MJExplorer/src/staticwebapp.config.json`, on a page already loading three third-party CDNs.
3. `resource/search/:searchInput` puts user-entered text in the URL path (also affects browser history and access logs).
4. `RecentAccessService.logAccess()` accepts a `resourceType` parameter it never persists, and does a `RunView` round-trip + full `BaseEntity.Save()` per record open.
5. `MJ: Audit Logs` is clustered on a random UUID PK with only FK indexes — no index on `__mj_CreatedAt`, so every time-ranged audit query scans.

---

## Test plan

- [ ] Confirm the `ng-filter-builder` export removal is acceptable (import path moves to `@memberjunction/core`)
- [ ] Decide whether the two dialogs that lost their backdrop should get one back
- [ ] Decide whether `/` should still auto-select the first nav item (silently broken since the Kendo removal)
- [ ] Confirm leaving `_common.scss` alone is the right trade vs. bundling the legacy `.mj-btn` migration
- [ ] Read the analytics guide and resolve its 7 open questions
- [ ] Agree the Phase 0 `ErrorHandler` proceeds independently

https://claude.ai/code/session_011JSgBxLeb1iG89Y9BnmniP


AN-BC:
@MattC-BC — flagging you on this one since it's downstream of your Kendo removal work. Four things this turned up that are decisions rather than cleanup, plus one file I deliberately left alone.

### 1. Two dialogs currently render with no backdrop

`add-item.component.html` and `ng-data-context-dialog.component.html` each carried `<div class="k-overlay"></div>` as their backdrop. `mj-window` documents that it deliberately has **no** backdrop and does not block interaction — so since the removal those divs have been unstyled empty elements and both dialogs have shown no dimming at all.

I deleted them (they render nothing, so deleting is a visual no-op), but restoring the backdrop is a UX call, not a cleanup. If you want it back it's a scoped class on the wrapper plus a z-index below the window — happy to do it, just didn't want to make that decision inside a cleanup PR.

Note `mj-dialog` *does* render its own backdrop, so only the two `mj-window` cases are affected.

### 2. Landing on `/` no longer selects the first nav item

Both `MJExplorerAppComponent` (~L303) and `InitializationService.navigateToInitialRoute` had this:

```ts
const drawerElement = document.querySelector('li.k-drawer-item.k-level-0');
if (drawerElement) drawerElement.click();
```

The Kendo Drawer is gone, so this has been matching nothing and doing nothing — a silent no-op inside a `setTimeout`. I removed it (again, zero runtime change), but **the behaviour it was meant to provide is genuinely absent**: a user landing on `/` gets whatever the shell defaults to rather than the first nav item. If that was intentional to keep, it needs reimplementing against the current nav; if it isn't wanted, nothing more to do.

I kept `navigateToInitialRoute(initialPath, _document)`'s signature rather than dropping the now-unused param, since it's on a published service — flagged in the JSDoc for removal at the next major.

### 3. `delete-item`'s confirmation text was never styled

`<p class="k-m-7.5 k-text-center">` — `k-m-7.5` isn't a valid single class token (the `.` breaks it), so that paragraph has been unstyled since *before* the Kendo removal, not because of it. Replaced with a real `.confirm-message` class.

### 4. `_common.scss` — the one file I left with dead `.k-*` rules

`explorer-app/src/lib/styles/_common.scss` still has 12 dead Kendo rules. Both UI gates evaluate whole files once changed, so touching it makes CI grade its pre-existing debt:

- 15 hardcoded colors — fixable, and I did fix them with the documented `color-mix()` token pattern; that gate went green.
- **`.mj-btn` is *defined* in this file** (L164) with four legacy `.mj-btn-*` classes `@extend`ing it. `check:ui-buttons` allows exactly one canonical definition (`ng-ui-components/.../button.scss`) and has no allowlist, so the file can't pass while that block exists.

Removing the legacy button block is a migration, not a cleanup, and has nothing to do with Kendo — so I reverted the file rather than bundle an unrelated change. Reasoning is in commit `3c1d6f2fb`.

One data point in case it's useful: **zero templates under `packages/Angular` still reference `mj-btn-primary` / `mj-btn-secondary` / `mj-btn-icon*`**, so the block appears dead in this repo. It ships in MJExplorer's global stylesheet though, so downstream apps could still be relying on it — worth confirming before anyone deletes it.

---

Also worth a look if you have a view: the filter payload types (`CompositeFilterDescriptor` et al.) were duplicated between `@memberjunction/core` and `ng-filter-builder`, with the builder re-exporting its copy. I consolidated onto core's, which means those six names are **no longer exported from `@memberjunction/ng-filter-builder`** — a real (small) API break for anyone importing them from there.

---

PR #4215, opened 2026-09-04
Skills: memoise the / picker's @agent chip lookup (follow-up to #4198)

cmbrock-BC:
## Summary

Follow-up to #4198 (merged before this landed on the branch). The review's one nit: `pickerTargetAgentId` is a getter bound in `mj-message-input`'s template on a default-change-detection component, and it called `getMentionChipsData()`, which walks the editor DOM (`querySelectorAll('.mention-chip')`) on every change-detection cycle.

The first version of this PR refreshed the lookup in `OnComposerValueChanged`. That was wrong, and the fix now hangs off a different seam — see Changes.

## Changes

- `pickerTargetAgentId` reads a memo (`mentionedAgentId`) instead of walking the DOM. `undefined` = dirty, recompute on next read; `null` = computed, no `@agent` chip. The two are not interchangeable: collapsing them is what makes a cleared or restored draft read as "no chip" forever.
- **The memo is invalidated from `messageText`'s setter**, so `messageText` becomes a get/set pair. Every write to it reaches the editor through `[value]` → `ngModel.writeValue` → `setEditorContent`, which rebuilds the chips with `appendChild` (or empties the editor) and never calls `onInput()` — no `valueChange`, so no hook fires. The setter is the one point that cannot be bypassed. It binds one-way (no two-way `ngModel`), so the pair is transparent to callers.
- Lazy rather than eager: a read inside the setter would be too early, since `ngModel` writes the editor on a later change-detection pass and the read would predate the chips it wants.

### Why not invalidate in `OnComposerValueChanged` (and `SetDraft`)

Two kinds of write reach the editor, and only one announces itself:

| Path | Emits `valueChange`? |
| --- | --- |
| typing, autocomplete insert, backspace-delete, `InsertMention`, `clear()` | yes — all end in the editor's `onInput()` |
| restored draft (`[initialDraft]` → `SetDraft`) | no |
| a host assigning `messageText` directly | no |

The restored draft is the case raised in review. The second is live in this repo: `conversation-chat-area` sets `messageInput.messageText = \`Analyze "${title}" — \`` (two call sites) and focuses the composer. That replaces the editor content — dropping any `@agent` chip — through `writeValue`, with no `onInput()` and so no `valueChange`. An invalidation keyed on `OnComposerValueChanged` never runs, and the picker keeps narrowing to a chip that is no longer on screen.

For completeness, the send paths are *not* an example: `OnSendClick` emits `TextSubmitted` and then calls `mentionEditor.clear()`, which does emit, so `handleSuccessfulSend` and the empty-state submit are covered by the composer clearing itself.

"The next keystroke repairs it" is not a defence either — the picker is opened by the Skills button as well as by typing `/`, and the button path takes whatever the memo holds.

## Testing

- New `skill-picker-chip-memo.test.ts`, 11 cases: the memo itself (one DOM walk per change, not per read), invalidation on both kinds of path, the restore case, the post-send reset, and the external host write.
- **Each case verified red against the implementation it guards.** 6 of 8 fail against the eager version, including the restore case; the last two fail against the invalidate-at-call-sites version.
- `@memberjunction/ng-conversations`: 106 files / 1263 tests pass, `tsc --noEmit` exit 0, `ngc` build clean.
- Changeset added: `patch` on `@memberjunction/ng-conversations` (source change in a published package).

### DIFF
--- a/packages/Angular/Generic/conversations/src/lib/components/message/message-input.component.ts
+++ b/packages/Angular/Generic/conversations/src/lib/components/message/message-input.component.ts
@@ -268,7 +268,28 @@ export class MessageInputComponent extends BaseAngularComponent implements OnIni
 
   @ViewChild('inputBox') inputBox!: AiComposerComponent;
 
-  public messageText: string = '';
+  private _messageText: string = '';
+  /**
+   * The composer's text. An accessor pair rather than a plain field because every write reaches the
+   * editor through `[value]` -> `ngModel.writeValue`, which rebuilds or empties the chip DOM WITHOUT
+   * emitting `valueChange` — so a write is exactly the event {@link mentionedAgentId} has to hear
+   * about, and the setter is the one place that cannot be bypassed.
+   *
+   * Bypassing it is not hypothetical: `handleSuccessfulSend` and the empty-state submit clear the
+   * text without touching the editor, and `conversation-chat-area` assigns `messageText` on this
+   * component from the outside (three call sites). Invalidating at the individual call sites instead
+   * would leave every future one to remember.
+   *
+   * Read is a plain field read; there is no two-way `ngModel` on this property (the template binds
+   * `[value]="messageText"` one-way), so the pair is transparent to callers.
+   */
+  public get messageText(): string {
+    return this._messageText;
+  }
+  public set messageText(value: string) {
+    this._messageText = value;
+    this.mentionedAgentId = undefined;
+  }
 
   /**
    * Prefills the composer with draft text WITHOUT sending (unlike pendingMessage,
@@ -313,7 +334,6 @@ export class MessageInputComponent extends BaseAngularComponent implements OnIni
 
   /** Handles the composer's value stream: keeps messageText in sync + emits draft state. */
   public OnComposerValueChanged(value: string): void {
-    this.refreshMentionedAgentId();
     this.messageText = value;
     this.DraftStateChanged.emit(this.GetSerializedDraft());
   }
@@ -567,20 +587,41 @@ export class MessageInputComponent extends BaseAngularComponent implements OnIni
    * `TargetAgentId`; null = unknown, no narrowing.
    */
   public get pickerTargetAgentId(): string | null {
+    if (this.mentionedAgentId === undefined) {
+      const chips = this.inputBox?.getMentionChipsData() || [];
+      this.mentionedAgentId = chips.find(chip => chip.type === 'agent')?.id ?? null;
+    }
     return this.mentionedAgentId ?? this.resolveCurrentAgentId();
   }
 
   /**
-   * The first `@agent` chip in the draft, refreshed when the composer's value changes (chips only
-   * change with the value) — so the template-bound {@link pickerTargetAgentId} does not walk the
-   * editor DOM on every change-detection cycle.
+   * Memo for the first `@agent` chip in the draft, so the template-bound
+   * {@link pickerTargetAgentId} does not walk the editor DOM on every change-detection cycle.
+   *
+   * `undefined` = dirty, recompute on next read; `null` = computed, no `@agent` chip present.
+   * The two are NOT interchangeable — collapsing them to `null` is what makes a cleared or restored
+   * draft read as "no chip" forever.
+   *
+   * Invalidated from {@link messageText}'s setter, which is the only choke point every chip change
+   * passes through. Chips reach the editor by two kinds of path and only one announces itself:
+   *
+   *   - user editing (autocomplete insert, backspace-delete, `InsertMention`) and `clear()` all end
+   *     in the editor's `onInput()`, which emits `valueChange` -> {@link OnComposerValueChanged},
+   *     which assigns `messageText`;
+   *   - a programmatic write — a restored draft (`[initialDraft]` -> {@link SetDraft}), a post-send
+   *     reset, or a host assigning `messageText` directly — goes `[value]` ->
+   *     `ngModel.writeValue` -> `setEditorContent`, which rebuilds the chips with `appendChild` (or
+   *     empties the editor) and never calls `onInput()`. No `valueChange`, so no hook fires.
+   *
+   * Invalidate-and-lazy rather than eager refresh, because an eager read in the setter would be too
+   * early: `ngModel` writes the editor on a later change-detection pass, so the read would predate
+   * the chips it wants. Marking dirty is timing-independent — the recompute happens on the next
+   * read, by which point the editor holds the new content.
+   *
+   * The picker can be opened by the Skills button as well as by typing `/`, so "the next keystroke
+   * would repair it" is not a defence: the button path takes whatever the memo holds.
    */
-  private mentionedAgentId: string | null = null;
-
-  private refreshMentionedAgentId(): void {
-    const chips = this.inputBox?.getMentionChipsData() || [];
-    this.mentionedAgentId = chips.find(chip => chip.type === 'agent')?.id ?? null;
-  }
+  private mentionedAgentId: string | null | undefined = undefined;
 
   /** True when the mic button should be enabled (have an agent + not disabled). */
   public get canStartRealtime(): boolean {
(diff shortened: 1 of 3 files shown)


packages/Angular/Generic/conversations/src/lib/components/message/message-input.component.ts
```diff
@@ -566,9 +567,19 @@ export class MessageInputComponent extends BaseAngularComponent implements OnIni
...
+  }
+
+  /**
+   * The first `@agent` chip in the draft, refreshed when the composer's value changes (chips only
+   * change with the value) — so the template-bound {@link pickerTargetAgentId} does not walk the
+   * editor DOM on every change-detection cycle.
+   */
+  private mentionedAgentId: string | null = null;
+
+  private refreshMentionedAgentId(): void {
```
rkihm-BC: Eager refresh here misses the one chip-changing path that does not go through `onInput()`: a restored draft (`[initialDraft]` → `SetDraft` → `[value]` → `ngModel.writeValue` → `setEditorContent`), which rebuilds chips without emitting `valueChange`. And a refresh placed inside `SetDraft` itself would run too early — `ngModel` writes the editor on a later change-detection pass.

Suggest invalidate-and-lazy instead, which is timing-safe and still O(1) in steady state:

```ts
/** undefined = dirty (recompute on next read); null = no @agent chip in the draft. */
private mentionedAgentId: string | null | undefined = undefined;

public get pickerTargetAgentId(): string | null {
  if (this.mentionedAgentId === undefined) {
    const chips = this.inputBox?.getMentionChipsData() || [];
    this.mentionedAgentId = chips.find(chip => chip.type === 'agent')?.id ?? null;
  }
  return this.mentionedAgentId ?? this.resolveCurrentAgentId();
}

// in OnComposerValueChanged AND SetDraft (and InsertAgentMention, for clarity):
this.mentionedAgentId = undefined;
```

Chip changes reach the editor in exactly three ways — user editing (incl. autocomplete insert / backspace delete), `clear()`, and `setEditorContent` — and the first two already emit through `onInput()`, so invalidating in `OnComposerValueChanged` + `SetDraft` covers all of them. The DOM walk then happens once per change rather than once per cycle, and a restored draft is read correctly on the first `/`.

rkihm-BC (changes requested):
Thanks for picking up the nit from #4198. The typed-input paths are all covered — autocomplete insertion, `InsertAgentMention`, backspace-deleting a chip and `clear()` all end in the editor's `onInput()` → `valueChange` → `OnComposerValueChanged`, so for those the memo is always fresh. Two items:

**1. Required — restored drafts bypass the hook, so "no behaviour change" does not hold.** The chat area persists the composer's lossless draft (chips serialised as JSON mentions) via `DraftStateChanged` and restores it via `[initialDraft]`. That flows `SetDraft` → `messageText` → `[value]` → `[(ngModel)]` → `writeValue` → `setEditorContent`, which rebuilds the chips with `appendChild` and never calls `onInput()` or emits `valueChange` (I read the full body; the input box emits `ValueChange` only from `OnValueChange` and after send). `OnComposerValueChanged` therefore never fires on a restore and `mentionedAgentId` stays `null`. Switch away from a conversation with `@Betty` pre-addressed and back: the chip is on screen, but `/` narrows to `resolveCurrentAgentId()` instead of the chip's agent until the next keystroke. Under #4198's live getter this case was correct. Small (the server's `RequestedSkills` guard is the backstop and any keystroke repairs it), but an optimisation PR should not regress a real flow. Fix suggested inline; a test for the restore case would be worth adding since the existing suite did not catch it.

**2. Changeset.** This is a source change in a published package (`@memberjunction/ng-conversations`), so per repo policy it needs a `patch` changeset even though the note is one line ("internal: memoise the / picker's @agent chip lookup; no behaviour change"). Changesets drive versioning and the release notes are generated from them. The changeset-bot comment has a pre-filled link.

CI has nothing failing on this head; a few checks were still pending when I looked.

cmbrock-BC:
Both fixed, pushed as `0d4b0b6`.

**1. Restored drafts.** Confirmed rather than taken on trust: `SetDraft` only assigns `messageText`, and `writeValue` → `setEditorContent` (mention-editor.component.ts:1298) rebuilds the chips with `appendChild` without calling `onInput()`, so no `valueChange` and the hook never fires. `InsertMention` does call `onInput()` (line 783), so the typed paths were covered exactly as you said.

One thing to add to the reasoning: the picker is opened by the Skills button as well as by typing `/`, so "any keystroke repairs it" does not cover the button path — it takes whatever the memo holds.

**Where the fix ended up.** Invalidating in `OnComposerValueChanged` + `SetDraft` would fix the restore case but not the general one. `conversation-chat-area` assigns `messageText` on this component from the outside — `messageInput.messageText = \`Analyze "${title}" — \`` at two call sites, then focuses the composer. That replaces the editor content through `writeValue`, dropping any `@agent` chip, with no `onInput()` and so no `valueChange`; a call-site invalidation never runs and the picker keeps narrowing to a chip that is no longer on screen.

So the invalidation hangs off **`messageText`'s setter**, which makes `messageText` a get/set pair. Every write reaches the editor through `[value]` → `writeValue`, so the setter is the one choke point nothing can bypass, and no future call site has to remember. It binds one-way (no two-way `ngModel`), so the pair is transparent to callers. Your point about eager refresh being too early is why it stays lazy — `ngModel` writes the editor on a later change-detection pass, so a read in the setter would predate the chips it wants.

**Correction to my own commit message,** which I got wrong and would rather flag than leave for you to trip over: it also cites `handleSuccessfulSend` and the empty-state submit as bypassing `valueChange`. They do not. `OnSendClick` emits `TextSubmitted` and *then* calls `mentionEditor.clear()`, which emits, so the composer clears itself on the send path. It also cites three external call sites; the third (`messageText = ''` before `sendMessageWithText`) is covered for the same reason. The two `Analyze` writes are the real ones. The PR description has the corrected version. The fix itself is unaffected — the setter covers all of it either way.

**Tests** (`skill-picker-chip-memo.test.ts`, 11 cases, `Object.create(prototype)` like message-input-streaming): the restore case, its inverse, invalidation on both kinds of path, a bare `messageText = ''` reset, the external host write, and the memoisation itself — including that a `null` result must not be re-read as "not yet computed". Each verified red against the implementation it guards: 6 of 8 fail against the eager version, and the last two against the invalidate-at-call-sites version. Package suite 106 files / 1263 tests, `tsc --noEmit` clean, `ngc` clean.

**2. Changeset.** Added, `patch` on `@memberjunction/ng-conversations`.

Note the branch had `next` merged into it while this was in flight — I rebased onto that rather than force-pushing over it.

rkihm-BC (approved):
Approving. Both items are addressed, and the fix landed in a stronger place than I suggested.

- **Setter-based invalidation is the right design.** Every write reaches the editor through `[value]` → `writeValue`, so `messageText`'s setter is the one choke point nothing can bypass — including the three direct host writes in `conversation-chat-area` that my call-site suggestion would have missed. Good catch. The getter stays lazy, so the timing problem with an eager read is avoided, and the `undefined`-vs-`null` distinction is pinned by tests.
- **Accessor pair is safe:** the only binding on `messageText` is the one-way `[value]`; no two-way binding or `@Input()` on it, so it is transparent to callers.
- **Tests:** the 11-case file covers the restore case and its inverse, both invalidation paths, the host write, the memo actually memoising, and the `null` result not being re-read as dirty — with red-then-green against both the eager and call-site versions per your note.
- **Changeset** added, `patch` on `@memberjunction/ng-conversations`, accurate text.
- CI fully green on this head; branch merged `next` and is mergeable.

Appreciated the self-correction in the thread about the send paths. One non-blocking nit that follows from it: the retracted claim still lives in two places in the code — the JSDoc on the `messageText` accessor ("clear the text without touching the editor") and the docblock/title of the "drops the chip when a send resets the text without touching the editor" test. The fix is unaffected since the setter covers every path, but those two blocks now justify the design with an example that is not true. Worth correcting on this branch or in a follow-up.

PR #4217, opened 2026-09-04
feat(ng-base-forms): a form can say where a NEW record opens

JosueGarcia-BC:
## Summary

A form contribution can declare `leadsWhenUnsaved` so a **new** record opens on it rather than on the first first-class group.

`ShouldPersistChromeActiveGroup` already says an unsaved record must not restore a stored rail position. It never said where the record should go *instead*, so it falls through to the lead group — and a lead group is usually a summary. A summary of a record with no data is a page of blanks the user has to look past to find where typing starts.

That is the defect reported against the Deal form in `bc-aidp-next-golive#188`: clicking **New Deal** opened an exec briefing, and the reporter had to notice and click into Pipeline before entering anything.

## Opt-in is the design

Any form that declares nothing behaves **exactly** as before. Changing the default would move where every new record in every MJ app opens, which is not something to do on one report about one form. Most of the tests cover the null cases for that reason.

Only consulted when the record is unsaved, so it cannot change how an existing record opens.

## Shape

The decision is a pure function in `form-chrome-rail-pref`, beside `ShouldPersistChromeActiveGroup` — it answers the other half of the same question, and being pure it is testable without standing up a container. The container supplies its own registrations and the key-derivation it already uses for the rail, so nothing new is threaded through the chrome resolver.

Highest ClassFactory `Priority` wins when several contributions on one entity claim it, matching how every other conflict between registrations is settled.

## Tests

10, mutation-checked rather than assumed:

| mutation | tests failed |
|---|---|
| loose falsy check instead of `!== true` | **6** |
| priority sort reversed | 1 |
| entity filter dropped | 1 |
| sorting the caller's array in place | 1 |

`@memberjunction/ng-base-forms`: **334 passed / 27 files** on this branch, against **324 / 26** on the base — the ten above are the whole difference.

## Verified in a running app

Built and driven in Explorer against a live database, not only unit-tested:

- a **new** deal opens on Pipeline, with the data-entry fields on screen
- an **existing** deal still opens on Overview — the regression this could most easily have caused

The consuming change is one line in `bizapps-sales` (`leadsWhenUnsaved: true` on its Pipeline panel) and is not part of this PR.


.changeset/form-chrome-leads-when-unsaved.md
```diff
@@ -0,0 +1,5 @@
+---
+"@memberjunction/ng-base-forms": minor
```
rkihm-BC: Required: this must be `patch`. The branch touches no migration and no metadata, and `node .github/scripts/check-changeset-bump.mjs` exits 1 on it. Because `.changeset/config.json` uses `fixed: [["@memberjunction/*"]]`, the highest bump in a release sets the version for every MJ package — so one `minor` here makes the entire release a minor. A new opt-in API is exactly where semver intuition and the repo rule diverge.

packages/AI/Providers/Cerebras/src/models/cerebras.ts
```diff
@@ -234,9 +234,19 @@ export class CerebrasLLM extends BaseLLM {
...
         // Forward the cancellation token to the Cerebras SDK's RequestOptions so an abort tears down the
         // underlying HTTP socket rather than merely abandoning this promise.
-        let chatResponse: ChatCompletion;
+        /**
+         * NARROWED TO THE NON-STREAMING MEMBER OF THE UNION, DELIBERATELY.
+         *
+         * The SDK's exported `ChatCompletion` is a UNION:
+         *   ChatCompletion.ChatCompletionResponse | ChatCompletion.ChatChunkResponse | ChatCompletion.ErrorChunkResponse
+         * Only the first carries `choices`, `usage` and `model`, so declaring this as the bare union
+         * made all three reads below fail to compile (TS2339) — the errors that have kept this package
```
rkihm-BC: Required: this claim doesn't reproduce on `next`, and this file doesn't belong in this PR.

I built the package from a clean worktree of `next` **without** this commit: `pnpm -r --filter "@memberjunction/ai-cerebras..." run build` → `tsc && tsc-alias -f` → `Done`, exit 0, zero TS errors. The root `Build` job runs `pnpm run build` over every package, which matches it being green on PRs that don't carry this change.

Separately, `@memberjunction/ai-cerebras` is published and the PR's changeset names only `ng-base-forms`, so this fix would ship with no changelog entry (the fixed group still versions it).

If it reproduces for you under a newer `@cerebras/cerebras_cloud_sdk` than the lockfile pins, say so and land it as its own PR with its own changeset — the narrowing is sound for a method named `nonStreamingChatCompletion`, and the cast is erased at runtime. It just shouldn't ride along here.

rkihm-BC (changes requested):
The feature itself is well built, and I would approve it on its own. Two required fixes, both outside the feature.

**What's good.** `UnsavedLeadGroupKey` is pure and sits beside `ShouldPersistChromeActiveGroup`, which is the right home because it answers the other half of that question. Priority resolution matches how every other registration conflict is settled, so the outcome doesn't depend on import order. `!== true` means an explicit `false` reads as "no claim" rather than becoming impossible to turn off. The caller's array isn't sorted in place, and there's a test for that specifically — the rail reads the same array, so a reorder would have been a visible defect. Weighting the ten tests toward the null cases is right: opt-in is the whole design, and the default is where a regression would hurt. Verified in a running app for both the new and the existing record path, which is the regression this could most easily have caused.

### Required 1 — the changeset must be `patch`, not `minor`

The local gate fails on it:

```
❌ Changeset bump levels (1 problem(s)):
   .changeset/form-chrome-leads-when-unsaved.md — "@memberjunction/ng-base-forms": minor,
   but this branch changes no migration and no metadata
```

Confirmed independently: the branch touches nothing under `migrations/` or `metadata/`. No CI job runs that gate, which is why 22 checks are green with the changeset wrong.

It matters more here than the bump level suggests. `.changeset/config.json` carries `fixed: [["@memberjunction/*"]]`, so the highest bump in a release sets the version for **every** MJ package — one stray `minor` makes the whole release a minor. A new opt-in API is exactly the case where semver intuition says minor and the repo rule says otherwise, so this is an easy one to get wrong; it's still wrong.

### Required 2 — an unrelated published package rides along, and its stated reason doesn't reproduce

Commit `ef98a94034` changes `packages/AI/Providers/Cerebras/src/models/cerebras.ts`. `@memberjunction/ai-cerebras` is published (`private: false`), and the changeset names only `ng-base-forms`. Because of the fixed group above the package is versioned and published either way — what it will not get is a changelog entry describing the fix.

The comment's justification also doesn't hold on `next`. It says the reads below "made all three reads fail to compile (TS2339) — the errors that have kept this package from building at all". I built it from a clean worktree of `next`, without this change:

```
pnpm -r --filter "@memberjunction/ai-cerebras..." run build
packages/AI/Providers/Cerebras build$ tsc && tsc-alias -f
packages/AI/Providers/Cerebras build: Done
exit=0    total TS errors: 0
```

The root `Build` job runs `pnpm run build` across every package, which is consistent with it being green on PRs that don't carry this change. If it reproduces for you under a newer `@cerebras/cerebras_cloud_sdk` than the lockfile pins, that's worth saying — and it's a real fix worth landing — but as a separate PR with its own changeset, not behind a forms feature. The cast itself is harmless (erased at runtime) and the narrowing is sound for a method named `nonStreamingChatCompletion`.

Both are mechanical. Happy to re-review as soon as they're addressed — nothing in the feature needs to change.

rkihm-BC (commented):
Narrowing my earlier review: **the changeset item is withdrawn.** One thing remains.

### Withdrawn — the `minor` bump

I argued a stray `minor` would make the whole release a minor. Robert pointed out that other PRs in this release already carry one, and he is right: `next` has **108 pending `minor` changesets** against 302 `patch`. The release is a minor either way, so this entry changes nothing about the version. My argument only held if it were the only one, and I should have checked that before raising it. The local `check:changeset` gate does still fail on it, so if you happen to be editing the file anyway `patch` is the letter of `.claude/rules/changesets.md` — but it is not worth a round trip and it is not a condition of my approval.

### Still stands — the Cerebras commit

Independent of the changeset question, on three counts:

1. **Unrelated to this PR.** `ef98a94034` changes `packages/AI/Providers/Cerebras/src/models/cerebras.ts`, which has nothing to do with where a new record opens.
2. **Its stated reason does not reproduce.** The comment says these are "the errors that have kept this package from building at all". Built from a clean worktree of `next` **without** the change: `pnpm -r --filter "@memberjunction/ai-cerebras..." run build` → `tsc && tsc-alias -f` → `Done`, exit 0, zero TS errors. The root `Build` job compiles every package, which matches it being green on PRs that lack this commit.
3. **No changelog entry.** `@memberjunction/ai-cerebras` is published and the changeset names only `ng-base-forms`. The fixed group still versions the package, so what is missing is the release note, not the version.

Either resolution is fine by me. Lift it into its own PR with its own changeset — my preference, since it is self-contained and a real build fix deserves its own note — or keep it here, add the changeset entry, and correct the comment to name what actually triggers the errors, which may well be a newer `@cerebras/cerebras_cloud_sdk` than the lockfile pins. The narrowing itself is sound for a method called `nonStreamingChatCompletion`, and the cast is erased at runtime, so I have no objection to the change on its merits.

### Two notes, neither blocking

- The Cerebras file still carries `(choice: any)` and a `// Cast to any to extract the choices` comment a few lines below your change. Pre-existing and outside the diff, so not yours — worth sweeping whenever that file is next touched deliberately.
- The new `UnsavedLeadGroupKey` docblock sits immediately above `ShouldPersistChromeActiveGroup`, so that function's original comment now reads as though it introduces the new one. Cosmetic.

The feature needs nothing. Approving as soon as the Cerebras file is off the branch, or accounted for.

rkihm-BC (approved):
Approving. `056c1536f3` addresses everything, and I re-reviewed the feature rather than waving it through on the strength of my earlier pass.

**Both items resolved, and the Cerebras one the way I hoped:**

- The Cerebras file is off the branch and is now #4333 with its own changeset. That is the better outcome — a real build fix gets its own release note instead of riding along unmentioned.
- Changeset is `patch`. Withdrawn as a condition earlier, so thank you for doing it anyway; `check:changeset` is clean now.
- Docblock reunited.

The branch is 5 files, +160/-0, all feature. 22/22 checks pass.

## What I probed this time, and what came back

I went looking for two specific ways this could misbehave rather than re-reading what I had already accepted.

**Can a contribution that is excluded from the form still claim the lead?** `CollectFormPanelRegistrations()` does no `inclusion` filtering — it returns every registration carrying an `entity` string — so `UnsavedLeadGroupKey` can legitimately return the rail key of a contribution that L3 `MJ: Form Chrome Rules` or `inclusion` has removed from this form. I expected that to open a new record on a group that is not rendered.

It does not, and the reason is worth recording so nobody "fixes" it later: `form-chrome-coordinator.service.ts:20`

```js
const stillValid = spec.Groups.some((g) => g.Key === this.ActiveGroupKey);
if (!stillValid) this.ActiveGroupKey = spec.Groups[0]?.Key ?? null;
```

An unrenderable key falls back to the first group — precisely the pre-existing behaviour. So the excluded case degrades to the old default rather than to a blank pane. Correct by composition rather than by accident, but it is load-bearing and undocumented at the call site.

**Is the entity comparison consistent with the rest of the container?** `meta.entity !== entityName` is exact — no trim, no case folding. The same file already does exactly that at line 1137, so this matches local convention rather than introducing a second one. (Both are exact-match; if that convention is ever wrong it is wrong in two places, but that is not this PR's to change.)

Neither probe produced a finding. Combined with the earlier read — pure function beside its sibling, priority resolution matching every other registration conflict, `!== true` keeping an explicit `false` meaningful, the caller's array not sorted in place, ten tests weighted toward the null cases, and verification in a running app for both the new and existing record paths — I have nothing left to raise.

One non-blocking observation, unchanged from before: the `(choice: any)` and its comment a few lines below your original change in the Cerebras file are still there on `next`. Not yours, and now not even on this branch — worth sweeping whenever that file is next touched deliberately. Relevant to #4333 rather than here.

Nice piece of work, and thank you for the mutation table — it is the reason I did not have to re-derive which tests actually constrain the behaviour.

PR #4218, opened 2026-09-04
ci(publish): seed new packages from publish.yml so the npm gate can pass

bc-izygmunt:
## Summary

The new-package npm gate and the `seed-new-package.yml` workflow contradicted each other, so **no new package could clear the gate**. This moves the seed into `publish.yml` as a fifth, confirmed dispatch invocation (invocation 4 is the LTS candidate cut from #3ffb170ce6, merged in), which is the only shape both npm and the gate accept.

Found on #4212 (`@memberjunction/ng-rich-text-editor`), where the seed run failed after trust had been configured correctly. Write-up with evidence: run [33845605437](https://github.com/MemberJunction/MJ/actions/runs/33845605437).

## Why the old design could not work

1. npm trusted publishing matches the **exact workflow filename**. `npm trust github … --file publish.yml` authorizes publishes from `publish.yml` only. A seed from `seed-new-package.yml` is refused at the OIDC token exchange, npm falls back to unauthenticated, and the `PUT` returns `E404 Not Found` — a permission denial, not a missing package. The run's "Require the package to already exist" step had passed with HTTP 200 two steps earlier.
2. Even with a second trusted publisher for the seed file (npm allows ten), `check-new-npm-packages.mjs` → `verifyProvenance` requires the attestation's workflow path to be `.github/workflows/publish.yml`. A seed from any other file fails as "built by the wrong workflow".
3. Both components landed in one commit and had never been exercised end to end: no `@memberjunction` package on npm carries a `seed` dist-tag.

## What changes

**`publish.yml`**
- New `workflow_dispatch` inputs `seed_package`, `confirm_seed_package` (double-entry, must repeat the package name exactly), and `seed_version` (default `0.0.1-seed.1`). All three are described as "SEED PATH ONLY" in the Run workflow form.
- `dispatch-guard` gains invocation 5. It requires the release fields (`line_branch`, `confirm_branch`, `cut_line`) to be **empty** on a seed, rejects a confirmation without a package, a mismatched confirmation, or an out-of-scope name, and the candidate-cut block refuses a dispatch that also carries `seed_package`. The refusal message lists all five invocations.
- `test-migrations`, `release-validation`, `build-and-publish`, `publish-lts`, `cut-candidate` each add `seed_package == ''` to their `if:`, so a seed dispatch reaches nothing that builds or releases.
- New `seed-package` job (`needs: dispatch-guard`, `id-token: write`): validates the prerelease shape, requires the placeholder to exist, writes the two-file marker package in `/tmp/seed`, `npm publish --provenance --access public --tag seed` with no token, then verifies the public attestation **names `publish.yml`**, failing otherwise.

**Gate + docs**
- `seed-new-package.yml` removed.
- `check-new-npm-packages.mjs`: header explains why the seed must come from `publish.yml`; `SEED_WORKFLOW_NAME` is the publish workflow's display name; new `SEED_DISPATCH_INPUT`; the failure message tells authors to dispatch `publish.yml` with `seed_package` + `confirm_branch`.
- Tests: the message must name the dispatch input and must not point at the retired workflow.
- `NEW_PACKAGE_SETUP.md`: new Step 5 "Seed the Package Over OIDC"; later steps renumbered.

## Verification

- `publish.yml` parses; every job's `if:` was printed and checked: the seed reaches only `dispatch-guard` and `seed-package`.
- The `dispatch-guard` script was extracted and executed locally against every combination of the five invocations (valid Edge, LTS, cut, seed; empty dispatch; wrong ref; mismatched confirmations; cut + seed; seed + line; out-of-scope name). Accepted and refused exactly as designed.
- The seed job's version validation, registry existence check, and package build were executed locally against the real registry.
- `.github/scripts` vitest suite: 18 files, 479 tests pass.
- No remaining references to `seed-new-package` outside the test that asserts its absence.
- Not exercised live: an actual seed dispatch needs a package with trust configured, and the first candidate is `@memberjunction/ng-rich-text-editor` once this merges to `next`. Trust matching is on repository and filename, not ref, so dispatching from `next` is fine, as the LTS path already does.

## After merge

Actions → **Build and publish new package versions** → Run workflow from `next`. Fill in exactly two fields, both `@memberjunction/ng-rich-text-editor`: `seed_package` and `confirm_seed_package`. Leave everything else at its default. Then re-run the checks on #4212.

No changeset: this touches workflows, a CI script, and a setup doc, not a package.


cadam11 (dismissed):
LGTM

AN-BC:
no input from me on this @bc-izygmunt - for stuff like this (CI, infra, things like that), have @cadam11 and @rkihm-BC do reviews, they'll be able to give you more useful input. Thanks.

cadam11 (approved):
LGTM

Issue #4226, opened 2026-09-04
Skills: a skill's granted actions are always exposed to the model — no way to hold an action for code-only use


cmbrock-BC:
First-adopter feedback from building a skill-driven mode (Betty's Exam Creator).

## What happens
`MJ: AI Skill Actions` is the only place a skill can hold an action, and `BaseAgent.enableSkillCapabilities` turns every granted action into a model tool on activation (`actionChanges` add, scope `specific`). There is no way to grant an action to a skill **without** offering it to the loop model.

## Why that matters
A mode whose menu runs an action from code — "Download as PDF", "Download as CSV" — wants the skill to be the action's holder (auditable, visible in admin, the same grant the code path checks) while the model never calls it. Today the model does call it: given `BLA - Export Exam CSV` as a tool, Gemini invoked it itself with a CSV it wrote, producing a second, model-made file beside the real one (Betty, 2026-09-04, run steps show an `Actions` step for the export before the code path ran).

## Workarounds in use
- Make the action's inputs something the model cannot know (an `ExamID` that never appears in context), so a model-initiated call fails.
- Override `enableSkillCapabilities` to push a `remove` `actionChanges` entry for the menu's actions right after MJ adds them.
Both work; both are fragile relative to a declared intent.

## Proposal
A flag on `AISkillAction`, e.g. `ExposeToModel BIT NOT NULL DEFAULT 1` (name open), honoured by `enableSkillCapabilities`: grants with it off are held by the skill (and satisfy `GetSkillActionIDs` for code-path checks) but are not added to the model's tools. Migration + CodeGen + one condition in `enableSkillCapabilities`; the guide's §1.6 would document "code-only" grants.

PR #4228, opened 2026-09-04
Artifacts: let an action's file output say it is a download (FileOutputRef.visibility)

cmbrock-BC:
## Summary

An action's `FileOutput` always becomes an artifact with `Visibility = 'Always'`: the file path in `AgentRunner` hard-codes it, and the output contract has no way to say otherwise. For a file the user asked to download and will open elsewhere (an exported CSV, a PDF), that puts a card on the message that, for a type with no viewer plugin, opens an empty pane — and the host can only fix it by rewriting the artifact after MJ created it.

`FileOutputRef` gains an optional `visibility` and `AgentRunner` honours it. `System Only` is the chat's existing switch: the artifact, its version and its download URL exist, but the message shows no card unless the host opts in with `showSystemArtifacts`. Default stays `Always`.

First-adopter feedback from building a skill-driven export in Betty.

## Changes

- `ai-core-plus`: `FileOutputRef.visibility?: FileOutputVisibility` (`'Always' | 'System Only'`); `ParseFileOutputRef` carries it when valid, leaves it unset otherwise.
- `ai-agents`: `createArtifactWithVersion` takes `visibility?` (default `Always`); `createFileArtifact` / `createInlineFileArtifact` take it as a trailing optional; the four `processFileOutput` call sites pass `fo.visibility`. Media and payload artifacts are untouched.

## Testing

- `@memberjunction/ai-core-plus`: 703 tests (2 new on the parser).
- `@memberjunction/ai-agents`: 1923 tests (3 new): file-backed, inline-kept and inline-uploaded outputs each carry their visibility to `createArtifactWithVersion`; an output that says nothing leaves it undefined for the default.
- Changeset: patch on both packages.

## Notes for reviewers

- Sibling to #4229 (the viewer fallback card), which is independent of this change — the two came out of the same artifacts pass but neither depends on the other. (An earlier version of this line cited #4226; that is an open *issue* about code-only skill action grants and has nothing to do with this PR.)

### Merged `next` (2026-09-04)

#4192 landed while this was open and refactored the call sites this PR edits, so `next` is merged in rather than rebased — one resolution instead of replaying it per commit.

- **`ProcessFileArtifacts`:** #4192 replaced the inline `HasStorage` → try-upload → catch-fallback block with `uploadToStorageIfConfigured()`. Kept that helper and threaded `fo.visibility` through the two calls that remain.
- **Parameter order, chosen deliberately:** #4192 appended `label: string = 'file'` to `createFileArtifact` and this PR appends `visibility`. Both are kept, `visibility` last — `label` carries a default and `CreateMediaArtifacts` already passes it positionally, so any other order breaks that caller.
- **The media path passes `label` only**, leaving `visibility` undefined → `'Always'`. A generated image is meant to show as a card; `System Only` is for an action's file output that is a download rather than something to display.
- **A defect the merge introduced silently:** the `fo.fileId` early return merged *cleanly*, leaving `fo.visibility` as the ninth positional argument — which is now `label`. It compiles, since `'Always' | 'System Only'` is assignable to `string`, and it is wrong twice: the artifact loses its visibility and falls back to `Always`, and the error text reads `Failed to save artifact for System Only`. Caught by this branch's existing test, which fails against the clean-merge result.

Verified after the merge: `ai-agents` 109 files / 1934 tests, `CorePlus` 34 / 703, `tsc --noEmit` clean once the `core-entities` dist is rebuilt (#4197 added `MJConversationSkillEntity` and `AISkill.ActivationScope` to `next`).


packages/AI/Agents/src/AgentRunner.ts
```diff
@@ -1439,14 +1439,14 @@ export class AgentRunner {
         try {
             if (fo.fileId) {
                 // File already in storage — create file-backed artifact
-                return await this.createFileArtifact(fo.fileId, fo.mimeType, fo.fileName, fo.sizeBytes, conversationDetailId, contextUser, provider, acceptUnregisteredFiles);
+                return await this.createFileArtifact(fo.fileId, fo.mimeType, fo.fileName, fo.sizeBytes, conversationDetailId, contextUser, provider, acceptUnregisteredFiles, 'file', fo.visibility);
```
rkihm-BC, line 1442: After the rebase this site no longer exists in this shape — #4192 collapsed the four calls into `uploadToStorageIfConfigured(...)` followed by one `createFileArtifact` / `createInlineFileArtifact` pair. `fo.visibility` needs to reach both of those, and `createFileArtifact` now already has a trailing optional `label` from #4192, so the parameter order between `label` and `visibility` is a choice to make explicitly (an options object would sidestep it, but that is a bigger change than this PR needs).

rkihm-BC (changes requested):
The change itself is right and minimal, and I verified it in a worktree: the new CorePlus parser test fails against base `agent-types.ts` and passes with the PR (CorePlus 703 passing); the new ai-agents test fails 2 of 3 against base `AgentRunner` and all pass with the PR (ai-agents 1,923 across 107 files). `visibility` is parsed strictly (only `'Always'` / `'System Only'` survive), threaded as a trailing optional defaulting to `'Always'`, so existing callers are unchanged. `'System Only'` is a valid `Artifact.Visibility` per the v2.110 CHECK constraint, and the chat already partitions on it behind `showSystemArtifacts`, so the "artifact exists, no card" behaviour is the existing switch being reused. Changeset `patch` on both packages.

**One required item, mechanical: rebase onto `next`.** #4192 merged this afternoon and refactored the four `processFileOutput` call sites this PR edits into a single `uploadToStorageIfConfigured` helper, so the branch now conflicts in `AgentRunner.ts` (GitHub shows it as conflicting). The rebase needs one deliberate decision rather than plain conflict resolution: #4192 added a trailing optional `label` to `createFileArtifact`, and this PR adds a trailing optional `visibility` to the same function — both positional, so pick an order and update the two surviving call sites in `processFileOutput` to pass `fo.visibility` (and, on the media path, whatever label/visibility combination is intended). The three new tests should carry across unchanged. Inline note on the first call site.

Minor: the description's "companion #4226" is an issue, not a PR, and is about skill action grants rather than this change — worth correcting the wording.

cmbrock-BC:
Both done — pushed as `3e8f15e`, and the PR is `MERGEABLE` again.

**Merged rather than rebased.** #4192 deleted the block this branch edits, so a rebase would have replayed the same resolution per commit; merging settles it once. Net diff against `next` is unchanged — the five files this PR owns, nothing else.

**The order decision you asked me to make explicitly:** `label` first (it has a default and `CreateMediaArtifacts` already passes it positionally), `visibility` last. Any other order breaks that caller. `fo.visibility` reaches both surviving calls in `ProcessFileArtifacts`.

**On the media path:** it passes `label` only, so `visibility` stays undefined and defaults to `Always`. That is intended — a generated image should render as a card; `System Only` exists for an action's file output that is a download rather than something to display. Happy to be told otherwise if you read it differently.

**Your inline note on line 1442 was better aimed than either of us realised.** That call site merged *cleanly* — no conflict marker — and was silently wrong afterwards: `fo.visibility` sat in the ninth positional slot, which is now `label`. It compiles, because `'Always' | 'System Only'` is assignable to `string`, and it breaks the feature in both directions at once — the artifact loses its visibility and falls back to `Always`, and the failure text reads `Failed to save artifact for System Only`.

Fixed, and mutation-tested rather than assumed: `a file-backed output carries its visibility` fails against the clean-merge result and passes with the fix, so the existing suite does cover it. Worth noting for the other artifacts branches — this is the kind of conflict git resolves confidently and wrongly.

**Verified after the merge:** `ai-agents` 109 files / 1934 tests, `CorePlus` 34 / 703, `tsc --noEmit` clean. The two errors it reports first are #4197's stale `core-entities` dist (`MJConversationSkillEntity`, `AISkill.ActivationScope`), not this branch — rebuilding that package clears them.

**#4226:** you are right, it is an issue about code-only skill action grants and does not belong in this description. Corrected — the line now points at #4229, which is the actual sibling, and says neither depends on the other.

rkihm-BC (approved):
Both requested items are done, and the merge caught something my inline note only pointed at.

- **The `next` merge resolves the #4192 conflict correctly.** `createFileArtifact` now ends `label = 'file', visibility?`, which is the right order: `label` carries a default and the media path already passes it positionally, so any other order would have broken that caller. Both surviving `processFileOutput` calls pass `'file', fo.visibility`; the inline path passes `fo.visibility` as its single trailing argument; the media path passes a label only, so generated images stay `Always` as intended.
- **The silently wrong clean merge is fixed.** The `fo.fileId` early return had `fo.visibility` land in the `label` slot — compiled, lost the visibility, and produced `Failed to save artifact for System Only`. The head passes `'file', fo.visibility` there, and the existing test asserts both `label === 'file'` and the visibility, so the suite genuinely covers it. Worth remembering for the other artifacts branches: that is exactly the shape of conflict git resolves confidently and wrongly.
- **Merge rather than rebase** is fine here — the net diff against `next` is the same five files, +133 −10.
- #4226 wording corrected; changeset `patch` on both packages; all checks green.

Approving.

PR #4229, opened 2026-09-04
Artifacts: a type with no viewer plugin gets a file card and a real Download

cmbrock-BC:
## Summary

The artifact viewer panel's display tab knew two things: a type's viewer plugin, or the extracted `displayMarkdown` / `displayHtml` attributes. An artifact with neither — CSV today (`MJ: Artifact Types` has the type but no `DriverClass`), any new file type tomorrow — rendered an empty pane, and the panel offered no way to get the file: a file-backed version has no inline `Content`, and there was no download action.

First-adopter feedback: an exported exam CSV opened to a blank pane; the file was correct in storage.

## Changes

- `artifact-viewer-panel`: when there is no plugin and no extracted attributes, the display tab shows a file card (name, type, size when known) with a **Download** button. File-backed content is fetched from `ArtifactFileService.getDownloadUrl` to a blob so the browser saves it (falls back to opening the URL if storage refuses the cross-origin fetch); inline content downloads via its data URL.
- Text-like content (`text/*`, JSON, XML, CSV) is fetched and shown as plain text under the card, capped at 200 KB with a note to download the rest.
- Plugins and extracted attributes keep precedence; nothing changes for types that already have a viewer.

## Testing

- `@memberjunction/ng-artifacts`: package suite passes; 2 new DOM tests — the fallback card and Download render for a CSV version with no plugin; the fallback is absent when a plugin exists or extracted markdown is available.
- Changeset: patch.

## Notes for reviewers

- `ArtifactFileService` is now injected into the panel (it was only used by the plugins); the DOM test provides a stub.
- A dedicated CSV viewer plugin (tabular preview) would be a further step; this PR makes every plugin-less type usable first.

### DIFF
--- a/.changeset/artifact-viewer-no-plugin-fallback.md
+++ b/.changeset/artifact-viewer-no-plugin-fallback.md
@@ -1,0 +1,16 @@
+---
+"@memberjunction/ng-artifacts": patch
+---
+
+**An artifact with no viewer plugin gets a file card and a real Download, not an empty pane.**
+
+The viewer panel's display tab knew two things: a type's plugin, or the extracted markdown/HTML
+attributes. A type with neither — CSV today, any new file type tomorrow — rendered nothing, and
+offered no way to get the file: a file-backed version has no inline `Content`, and the panel had no
+download action. First-adopter feedback: an exported exam CSV opened to a blank pane.
+
+Now such an artifact shows a file card (name, type, size) with a Download button — file-backed
+content via the pre-authenticated URL fetched to a blob so the browser saves it (falling back to
+opening the URL if storage refuses the cross-origin fetch), inline content via its data URL — and
+text-like content (`text/*`, JSON, XML, CSV) is fetched and shown as plain text, capped at 200 KB
+with a note to download the rest. Plugins and extracted attributes still take precedence.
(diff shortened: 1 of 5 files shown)


rkihm-BC (approved):
Approving. A plugin-less type now gets a usable pane and a real download, and nothing changes for types that have a viewer.

- **Precedence is correct:** the fallback is a final `@else if` after the plugin and extracted markdown/HTML branches, and `showFileFallback` requires no plugin, a version, and neither extracted attribute. The DOM tests pin the card for a CSV with no driver class and its absence when a plugin or extracted markdown exists.
- **Download is done safely:** file-backed content via the pre-authenticated URL fetched to a blob with `a.download` + `rel=noopener`, `window.open(..., 'noopener')` as the cross-origin fallback, object URL revoked; inline content through the existing data-URL helpers; the text preview is interpolated into a `<pre>`, never bound as HTML. `ArtifactFileService` is `providedIn: 'root'`, so injecting it into the panel needs no module change, and the test stubs it.
- New CSS is token-based and the token gate passed; `artifactIcon` and `LogError` already exist on the panel. Changeset `patch`. CI fully green; behind `next` but mergeable.

Two optional refinements, not conditions:
1. The text preview fetches the whole file and then truncates to 200 KB, so a large CSV is downloaded in full just to show its first part. A `HEAD` for the size, or a `Range` request where storage allows it, would make the cap real.
2. `fallbackSize` only sizes inline content, but `ArtifactVersion.ContentSizeBytes` exists and `AgentRunner` populates it for file-backed versions, so the card could show a size for those too.

PR #4230, opened 2026-09-04
feat(communication): calendar retrieval on the provider contract

JosueGarcia-BC:
## Why

`BaseCommunicationProvider` can send, read, reply, forward, draft, move, archive, search and subscribe — and has no way to read a calendar. Any caller needing events has to build its own Graph client, which means duplicating token acquisition, the credential-keyed client cache and paging that `MSGraphProvider` already owns. `getGraphClient` is private, so there is no way to reuse it from outside.

This came out of Activity Sync, which models Meetings and has a calendar provider it cannot feed.

## What

| addition | where |
|---|---|
| `GetEvents` | `BaseCommunicationProvider` — concrete, not abstract |
| `GetEventsParams` / `GetEventsResult` / `GetEventsEvent` | `communication-types` |
| `'GetEvents'` | `ProviderOperation` |
| a real implementation | `MSGraphProvider`, declared in `getSupportedOperations()` |

## Three decisions worth reviewing

**Concrete rather than abstract.** Making it abstract would break every existing provider at compile time for a capability most will never have. Providers that support it override and declare it; the rest inherit a refusal that names itself. That is also why the default returns `Success: false` rather than `{Events: []}` — "this provider cannot look" and "there was nothing in the window" are different facts, and a caller advancing a watermark must not read the first as the second.

**Separate from `GetMessages`, not a mode of it.** A calendar read is bounded by a time window, not a count and a folder. "The next 200 messages" is a sensible request; "the next 200 events" is not. Folding it in would have left `NumMessages` meaning different things depending on a flag.

**The endpoint choice is reported, because it silently changes the answer.** Graph exposes calendar data two ways that do not return the same thing:

- `/calendarView` requires a window and **expands** a recurring series into one entry per occurrence
- `/events` needs no window and returns the series **master** — a weekly stand-up is one row whose start time is whenever the series began

Supplying both bounds selects the first. Since a master and an occurrence are indistinguishable by inspection, the result carries `RecurrenceExpanded` rather than leaving the caller to infer which it received. A caller logging attendance wants occurrences; one listing "what meetings exist" may not.

## Smaller things

- Cancelled events are filtered **server-side, before `$top`**. Filtering after the fetch would return short pages that read as a quiet calendar rather than a filtered one. `IncludeCancelled` opts back in.
- A start Graph cannot express as an instant becomes `null`, never a guess. Graph sends naive local time plus a separate `timeZone`; without a tz database a named zone does not determine an instant, and guessing files a meeting hours from when it happened. An explicit offset is honoured regardless of the named zone.

## Tests

20, each mutation-checked. Pinning either endpoint unconditionally, accepting one bound as a window, dropping the cancelled filter, omitting `RecurrenceExpanded`, leaving the organizer in the attendee list, guessing at a named zone, and turning a thrown Graph error into a successful empty result are all caught.

Full suites green: 99 in `communication-types`, 54 in `communication-ms-graph`.

## Note for whoever reviews

`MSGraphProvider.resolveCredentials` requires four fields — `tenantId`, `clientId`, `clientSecret`, `accountEmail` — while MJ's own `"Azure Service Principal"` credential type declares only the first three as required. A credential created exactly as MJ documents it fails the provider it exists to feed. Not fixed here, to keep this PR to one concern, but it affects `GetMessages` and `GetEvents` alike and is worth its own change: either stop requiring `accountEmail` when `Identifier` is supplied, or declare it on the credential type so the two agree.


rkihm-BC (changes requested):
Good PR. The design is right and well argued, the tests assert what Graph was asked rather than only what came back, and the error handling is better than the `GetMessages` it sits beside. Two things block merge, a few should land with it, and the rest are optional.

### Blocking

**1. Changeset must be `patch`, not `minor`.**
This branch touches no migration and no metadata, so the bump rule in `.claude/rules/changesets.md` requires `patch` for both entries. Running the gate against the branch:

```
❌ Changeset bump levels (2 problem(s)):
   .changeset/communication-calendar-retrieval.md — "@memberjunction/communication-types": minor, but this branch changes no migration and no metadata
   .changeset/communication-calendar-retrieval.md — "@memberjunction/communication-ms-graph": minor, but this branch changes no migration and no metadata
```

Nothing in CI enforces this, so it will merge silently otherwise. `npm run check:changeset` reproduces it locally.

**2. The default query shape may be rejected by Graph, and the tests cannot tell.**
With `IncludeCancelled` unset, both branches send a `$filter` on `isCancelled` together with an `$orderby` on a different property (`lastModifiedDateTime desc` on `/events`, `start/dateTime` on `/calendarView`). Microsoft's Outlook backend requires that every property in `$orderby` also appear in `$filter`, and the failure is a 400 `InefficientFilter` "The restriction or sort order is too complex for this operation." The rule is documented for messages and Microsoft's announcement of it is titled for Mail, Calendar and Contacts, but I could not find a calendar-specific example, so I am not certain it fires here.

The unit tests mock the Graph client entirely, so a 400 is invisible to them. Please either run the default-path call against a real tenant for both endpoints and paste the result here, or drop the `$orderby` whenever a filter is applied. calendarView already returns occurrences in start order, so the second option costs little.

Refs: [List messages, "Using filter and orderby in the same query"](https://learn.microsoft.com/en-us/graph/api/user-list-messages?view=graph-rest-1.0#using-filter-and-orderby-in-the-same-query), [Update to Filtering and Sorting in the Mail, Calendar and Contacts REST API](https://devblogs.microsoft.com/microsoft365dev/update-to-filtering-and-sorting-rest-api/).

### Should land with this PR

**3. Base-class refusal test.**
`packages/Communication/base-types/src/__tests__/BaseProvider.test.ts` already pins the default refusal for `GetSingleMessage`, `DeleteMessage`, `MoveMessage`, `ListFolders` and `MarkAsRead`. `GetEvents` should get the same case, since the "cannot look" versus "nothing scheduled" distinction is the contract this PR describes as load-bearing.

**4. Docs.**
`packages/Communication/providers/MSGraph/README.md` lists required scopes and operations and does not mention `Calendars.Read` or `GetEvents`. `packages/Communication/base-types/README.md` lists operations and param types and has no entry for the new ones.

**5. `NumEvents` is not a hard cap, and the PR body overstates paging.**
`NumEvents` goes straight to `$top` with no `@odata.nextLink` following. Graph's paging guide says behavior when `$top` exceeds an API's maximum varies: ignored, capped, or an error. There is no `nextLink` handling anywhere in `MSGraphProvider`, so "paging that `MSGraphProvider` already owns" is not accurate. This matches `GetMessages`, so I am not asking for paging here, but the doc comment on `NumEvents` should say it is bounded by one Graph page.

### Optional

**6. Pin the timezone contract.**
Both endpoints return start and end in UTC unless a `Prefer: outlook.timezone` header is sent, and this provider never sends one, so the null-instant path is effectively unreachable today. Adding `.header('Prefer', 'outlook.timezone="UTC"')` makes that explicit, as `packages/AI/RealtimeBridge/Server/src/calendar-clients.ts` already does. Keep the null fallback.

**7. `$select`.**
Every event comes back with full HTML body and attendee response state, and the mapper keeps only `bodyPreview`. Selecting the mapped fields plus `isCancelled` and `seriesMasterId` would shrink the payload. Note `createdDateTime` and `lastModifiedDateTime` cannot be selected on calendarView.

**8. Window semantics.**
The params doc calls `StartDateTime` inclusive. I could not confirm from the docs whether calendarView also returns events that started before the window and overlap it. Worth checking, since a caller doing incremental sync by start time would see such an event in two adjacent windows.

### Nits

- The docblock "Returns the list of operations supported by MS Graph provider" at `MSGraphProvider.ts:813` is now orphaned above the `GetEvents` docblock.
- The `[Zz]` alternative in the `graphInstant` regex is unanchored, so any string containing a "z" counts as having an offset. Harmless in practice, but the RealtimeBridge parser anchors it and this should too.
- The mock scaffolding in `GetEvents.test.ts` duplicates the first sixty lines of `MSGraphProvider.test.ts`. A shared `__tests__/graph-mocks.ts` would pay for itself on the next Graph operation.
- The `accountEmail` versus "Azure Service Principal" mismatch you flagged is real and deserves its own issue so it is not lost. Agree with keeping it out of this PR.

rkihm-BC (approved):
Every item from the review is addressed, and the optional ones were handled rather than skipped.

**Blocking**
- Changeset is `patch` on both packages.
- The `$filter` + `$orderby` question: both query shapes were sent against a live tenant and returned 200, recorded in a code comment at the call site with the reasoning that the mocked tests can never catch a 400 and the next reader can't re-derive it. That's a better home for it than a PR comment. I can't reproduce it without tenant access, so this rests on your stated verification — which is what I asked for.

**Should-land**
- Base-class refusal test for `GetEvents` pins `Success: false`, an empty list, and the naming message.
- Both provider READMEs gain `Calendars.Read`, the operation, and a worked example of the window/no-window difference; base-types README gains the operation and param types.
- `NumEvents` doc now says it is a cap bounded by one provider page and that completeness means narrowing the window.

**Optional**
- UTC header sent explicitly, null fallback kept.
- Window semantics verified against Graph rather than guessed: overlap, not start time, with the measured case and the dedupe warning for incremental sync. Most useful line in the PR for a future caller.

**Nits**: `[Zz]` anchored, the orphaned docblock is back above its method, mock scaffolding shared in `graph-mocks.ts`.

Head `203a35ea9b` merged with `next`, 19/19 checks green. One note outside this PR: `README.md` and `readme.md` both exist under the MSGraph provider on `next` already — a case-only duplicate that bites on case-insensitive filesystems. Not introduced here; worth its own cleanup.

Approving.

PR #4231, opened 2026-09-04
fix(server): resolve the context user by Name or Email, not Name alone

### LINKED ISSUE #4209: The default magic-link provisioning user is an email but is matched against User.Name, so it can never resolve
**Found by** the PR gauntlet on MemberJunction/bizapps-forms#134
(`fix/122-misconfig-fails-at-boot`), 2026-09-03, while verifying that repo's boot-time readiness
check. The defect is in MJ core, not in Forms; Forms is only where it surfaces.

### Where
- `packages/MJServer/src/config.ts` — the default value of `userHandling.contextUserForNewUserCreation`
- `MagicLinkService.resolveProvisioningContextUser` — the lookup
- `packages/GenericDatabaseProvider/src/UserCache.ts` — `UserByName`, which compares `User.Name`

### What happens
Core's default `userHandling.contextUserForNewUserCreation` is the literal string
`'not.set@nowhere.com'` — an **email-shaped** value. `resolveProvisioningContextUser` resolves the
candidate with `UserCache.UserByName`, which matches against `User.Name`.

On a stock MJ database the system user's `Name` is `'System'` and its `Email` is
`'not.set@nowhere.com'` — so the default was evidently *meant* to resolve to the system user, and
**cannot**, on any host, because it is matched against the wrong column. Every magic-link redeem
then logs, at error level:

```
[MagicLink] Configured provisioning user 'not.set@nowhere.com' not found; falling back to an Owner.
```

and provisions the new anonymous user under whichever account happens to qualify as an Owner.

### Why it matters
Two consequences, both silent:

1. **Attribution is incidental.** "Who provisioned this respondent's access?" is answered by
   whichever user sorts first as an Owner, and that answer changes as the user table changes. Any
   audit over `CreatedByUserID` for magic-link-provisioned users is reading noise.
2. **The log is poisoned.** One error-level line per link open, on every host that has not
   overridden the setting — which is every host that took the default. Real errors are buried under
   it. On the review machine a short session accumulated 2–4 of these lines per boot.

The setting also carries a comment describing it in email terms, so an operator who tries to fix it
by supplying an email address reproduces the bug rather than fixing it.

### Repro
```bash
# Any stock MJ host with magic-link enabled and no explicit provisioning user configured:
#   mj.config.cjs has magicLink.enabled = true and NO magicLink.contextUserForProvisioning
# Open any magic link (in bizapps-forms: any public form link, GET /f/<slug>), then:
grep "Configured provisioning user" <the host's log>
```

Observed: one `[MagicLink] Configured provisioning user 'not.set@nowhere.com' not found; falling
back to an Owner.` line per link open.
Expected: the default resolves to the system user it names, and logs nothing.

### Evidence
From the #134 gauntlet run against MJ `v6.1.0-edge.5`:

| Host state | `Configured provisioning user` lines |
|---|---|
| default (no `contextUserForProvisioning`), branch build | 4, 3 and 2 across three boots |
| default, base build | 2 |
| `magicLink.contextUserForProvisioning: 'System'` — a **Name**, not an email | **0** over two link opens |

Setting it to the system user's *Name* is the workaround, and it works. That is the defect
demonstrated, not fixed: the shipped default still names an Email.

### Suggested fix
Make the default resolve to the user it was aiming at. Two options, in order of preference:

1. **Change the default to the system user's `Name`** (`'System'`), matching the lookup that is
   actually performed. Smallest change, and it makes the shipped default correct on a stock database.
2. **Resolve by either column** — have `resolveProvisioningContextUser` fall back to an email match
   when the name lookup misses. Larger blast radius (it changes resolution semantics for hosts that
   already configured a Name that happens to collide with someone's Email), so preferred only if the
   email spelling of the default must be preserved for compatibility.

Whichever is chosen, the config comment on `contextUserForNewUserCreation` should say which column
it is matched against, because it currently reads as an email field.

### Definition of done
- [ ] A failing test that reproduces it, then green
- [ ] A stock host with no `contextUserForProvisioning` opens a magic link and logs **no**
      `Configured provisioning user … not found` line
- [ ] Provisioning is attributed to the system user, not to an arbitrary Owner
- [ ] A host that has explicitly configured a provisioning user is unaffected
- [ ] Existing suite green; no changed expectation in an existing test

### Verify by
The repro above: after opening several magic links, `grep "Configured provisioning user"` over the
host log returns nothing, and the provisioned users' `CreatedByUserID` is the system user.

SDesai-BC:
MJ's own shipped default for the provisioning context user **has never resolved on any host** — it names the system user by an address that is matched against the wrong column.

`DEFAULT_SERVER_CONFIG.userHandling.contextUserForNewUserCreation` ships as `'not.set@nowhere.com'`, the seeded system user's **Email**. Both consumers resolved it through `UserCache.UserByName`, which matches **Name** — and the seeded row is `Name='System'` / `Email='not.set@nowhere.com'`. That row is the one place on a stock database where the two columns diverge, because `NewUserBase` sets `user.Name = email` for every user MJ creates itself.

```ts
const byName = UserCache.Instance.UserByName(candidate);   // matches User.Name
if (byName) return byName;
LogError(`[MagicLink] Configured provisioning user '${candidate}' not found; falling back to an Owner.`);
return UserCache.Users.find((u) => u.Type?.trim().toLowerCase() === 'owner') ?? null;
```

Two silent consequences. Every magic-link redeem logged that line at error level, on every host that took the default. And provisioning was attributed to whichever user happened to sort first as an Owner — `SELECT * FROM vwUsers` carries no `ORDER BY` and the cache is mutated in place at runtime, so `CreatedByUserID` was not stable across boots *or* within a process. Any audit over it was reading noise.

Closes #4209

## Wider than the issue reported

`packages/MJServer/src/auth/newUsers.ts` carried the identical defect — same default, same lookup, same arbitrary-Owner fallback — on the JWT auto-provisioning path, and `autoCreateNewUsers` defaults to `true` (`config.ts:648`). The issue only named the magic-link path.

`RealtimeClientSessionResolver.ts` already documents a production incident traced to this same placeholder, with a hand-rolled workaround built around it.

## What this does

Extracts the ladder into `auth/principals.ts` and migrates the three config-driven call sites onto it. The Owner-fallback idiom appeared in **six** files across the server at `origin/next`, in three mutually inconsistent variants; two of them could not resolve the value MJ ships.

The order is the design decision worth defending:

1. **`Name`** — tried first, so every host that resolves today resolves to the *same* user. This is what makes adding the `Email` rung a compatible change rather than a semantic one, and it neutralises the objection the issue raised against this option.
2. **`Email`** — the identity column used everywhere else in MJServer (`verifyUserRecord`, `UserResolver.UserByEmail`, and `MagicLinkService.provisionUser` itself resolve the invitee this way), and the only one the schema makes unique (`UQ_User_Email`). This rung is what fixes the shipped default.
3. **System user by ID** — survives the system user being renamed.
4. **Lowest-ID active Owner** — still a last resort, but a deterministic one.

Changing the default to `'System'` alone would not have been enough: it does nothing for a host whose config file carries the literal email, or one that followed the README's `'admin@example.com'`, and it binds correctness to a column with no uniqueness constraint. The default is changed too, as documentation rather than as the mechanism.

A misconfiguration is now reported **once per distinct setting + value** instead of once per request, so the diagnostic survives without burying real errors.

## Behaviour changes

Both are confined to hosts already falling back; a host whose configured user resolves is unaffected.

- Provisioning that previously landed on an arbitrary Owner now lands on the system user, so `CreatedByUserID` for newly provisioned users changes — to a stable value. Historical rows are untouched.
- The Owner fallback now requires `IsActive`. A deployment whose only Owner is deactivated previously provisioned under that disabled account and now resolves to no principal, failing loudly instead.

## Evidence

Reproduced first against the real shipped code — real `DEFAULT_SERVER_CONFIG`, real `UserCache.UserByName`, the baseline `__mj.User` seed rows. Two identical hosts differing only in cache order attributed to different users, which is the defect the issue describes.

Written test-first: the suite was run against a characterisation of the *current* behaviour and failed 6 of 10 on exactly the assertions above, before the ladder existed.

| Check | Result |
|---|---|
| `principals` + `principals.logging` suites | 17 passed |
| MJServer unit suite | **1179 passed, 0 failed**, 56 skipped |
| MJServer build | 0 TypeScript errors |
| Integration tier (deterministic, live MJAPI) | 56 passed / 3 failed / 4 skipped of 63 |

The 3 integration failures are **pre-existing and unrelated** — verified, not assumed. Stashing this change, rebuilding, booting a baseline MJAPI and re-running gives the identical result: 56 / 3 / 4, the same three bundles (`IT26 - Application Wiring`, `IT45 - Embedding Invariants`, `IT52 - Unified Search Seams`). All three assert on shared-database state — duplicate `isDefault` nav items, empty `EmbeddingVector` payloads, lexical search returning nothing — none on code this diff touches.

## Notes for review

`newUsers.test.ts` is the one existing suite that changes. Its setup mocked `UserByName` so that an email resolved, which is precisely what hid this bug for so long; that mock moves to the new seam. One assertion also changed: a test asserted `mockCacheUsers` had length 0, which only held because the cache was empty at rest — it can't be now that the configured user must live in it, so it now asserts the thing actually meant, that the failed user was not added.

Deliberately left alone and recorded in `plans/provisioning-context-user-resolution.md` rather than fixed here: the telephony trio, which already uses the correct ladder; retiring `UserByName` at its four `"System"` call sites; the missing `IsActive` check in `GetSystemUser()`; and cold-cache behaviour, where the ladder stays synchronous and returns null exactly as today.

No migration, no metadata, nothing to push or enable. Changeset is `patch`.

https://claude.ai/code/session_01GNBntCKa7bPZV8zd2aAXuX


rkihm-BC (changes requested):
Well-diagnosed fix with a good design record, a real reproduction, and a test surface that matches the design. All checks are green and the changeset gate passes at `patch`. One claim the code makes about its own behaviour is not true, and it is the claim that carries the security-relevant invariant. Approving once that is resolved.

### What's good

- **The diagnosis is right and wider than the issue.** #4209 named the magic-link path. This PR found the same defect on JWT auto-provisioning in `newUsers.ts`, which fires by default, and the Email-only variant in the example subclass integrators are told to copy. I confirmed all six hand-rolled ladders in three variants on `next`.
- **Name-first is the correct compatibility choice.** It neutralises the objection the issue raised against option 2, and the reasoning is recorded in the module and the plan so it will not be re-litigated.
- **The determinism fix is a genuine second bug.** `LoadUsers` selects from `vwUsers` without `ORDER BY` and the cache is pushed to at runtime, so the old fallback really did make `CreatedByUserID` depend on row order. Tie-breaking by lowest ID on every rung closes that.
- **Pure core, impure shell.** `resolvePrincipalFrom` takes the user array and reports rather than logs, so the ladder tests need no singleton and no mocks. The logging shell is tested separately. Right split.
- **The LRU over a capped Set** is a correct subtlety: a capped Set would reintroduce per-request logging for exactly the case the cap defends against.
- **Honest about the one changed assertion.** The issue's definition of done says no existing expectation changes. The PR changed one and explains why the old assertion only held by accident. I agree with the change.
- **Docs and templates are all corrected**: config comments, README, magic-link guide, both tracked `mj.config.cjs` templates, the docker config, and the regression patch script. The old README example of `admin@example.com` was itself a trap.

### Needs a change: "no rung returns an inactive user" is false for rungs 1 and 2

`principals.ts:87` states that no rung returns an inactive user. The code applies `IsActive` only on the system and Owner rungs:

```ts
const byName = lowestById(users.filter((u) => normalize(u.Name) === wanted));
// ...
const byEmail = lowestById(users.filter((u) => normalize(u.Email) === wanted));
```

A deactivated configured user still resolves. The concrete case is the one an operator is most likely to hit: an admin whose address was put in `contextUserForProvisioning` leaves and is deactivated. Provisioning continues to run as that disabled account, silently, and the docstring now promises it will not. The three inactive-user tests all pass a candidate that matches nobody, so none exercises a deactivated configured user, which is how the gap survived.

Suggested fix: filter `IsActive === true` on both configured rungs and fall through to the system rung with the existing warning path. The warning text should say the candidate matched but is inactive, since "matched no user's Name or Email" would mislead there. Add a test for a deactivated configured candidate. If you would rather keep the current behaviour, correct the docstring instead; either way code and prose have to agree.

### Worth a sentence in the plan

The plan's trust table says the candidate is server config with no injection surface. True of the candidate, but the match target is a column users may edit. In the baseline seed the UI role is read-only on the Users entity, while Developer and Integration have full update, and the shipped default for `newUserRoles` is `['UI', 'Developer']`. So on a stock config every auto-provisioned user can update User rows, including `Type`, which dwarfs anything the Name rung adds. Not a regression and not this PR's problem, but the Name-first ordering rests on that column being trustworthy, and the plan should say under what role grant it is.

### Nits

- `index.ts` exports `ReportedMisconfigurationCount` and `MAX_REPORTED_MISCONFIGURATIONS` from the package's public surface. Nothing outside tests uses them. Exporting test hooks from a published API is a maintenance commitment; consider an underscore-prefixed export with a comment, or a test-only entry.
- Casing is mixed within one module: `resolvePrincipalFrom` alongside `ResolveConfiguredPrincipal`. The repo mixes both styles for exported functions, so this is only about consistency within the file.
- The branch is 23 commits behind `next` with no drift in any touched file. Merge is clean, but GitHub reports `BEHIND`, so update the branch if the merge queue requires it.

### Follow-ups the PR correctly leaves out

The plan records these and I agree with deferring them: the telephony trio already uses the correct ladder, `GetSystemUser()` ignores `IsActive`, `UserByName` has no null guard, and cold-cache behaviour is unchanged. One to add to that list: the incident comment at `RealtimeClientSessionResolver.ts:2619` conflates two mechanisms. `GetSystemUser()` resolves by ID and never reads config, yet the comment quotes the magic-link log line as if it came from that path.

rkihm-BC (approved):
The required change is made, and made better than the minimum I asked for.

- **No rung returns an inactive user, and the code now matches the claim.** Rather than adding two filters, the `IsActive` guard moved into the shared tie-breaker (`lowestActiveById`), so every multi-match rung carries both cross-cutting guarantees in one place; the by-ID system rung keeps its own guard. The ladder docstring now says why the configured rungs are included rather than exempt — the departed-admin case is exactly the one an operator hits.
- **The warning distinguishes the two failure modes**, as asked: `describeMiss` reports "matched only INACTIVE users" when the setting is spelled right but points at a disabled account, and "matched no user's Name or Email" otherwise, and it runs only after the configured rungs have missed.
- **Four new tests**: deactivated candidate by Email, by Name, the warning text, and an inactive `Name` holder not shadowing an active `Email` holder — #4209's own shape with the Name holder deactivated. That last one I had not asked for and it is the right test to have.
- **Plan trust table** now separates the trusted candidate from the untrusted match column and names the baseline role grants.
- **Test hooks are off the public surface**; tests import the counters from the module directly. The mixed casing is kept deliberately and documented as shell-vs-core — a defensible convention now that it is written down.
- Changeset updated to describe the deactivated-candidate behaviour; 20/20 checks green.

The stale incident comment in `RealtimeClientSessionResolver.ts` remains a follow-up, as before. Approving.

PR #4239, opened 2026-09-04
fix(ng-base-forms): render left-nav Details field panels as one card

MarceloT-BC:
### Problem

In left-nav (sidebar) layout the record-form container strips the header and card from every selected panel (`.mj-chrome-show` rules in `record-form-container.component.css`). That is right for a single related grid, where the rail label is its header. It is wrong for the **Details** rail item, which shows every field panel at once: the label/value rows float on the page background with no surface and no grouping.

### Fix

- `IsDetailsSectionKey(spec, key)` and `DetailsCardEdges(keys, orderOf)` helpers in `chrome/form-chrome.ts`.
- The container tags Details panels with `mj-chrome-details`, plus `-first` / `-last` on the visual edges. Edges follow the form's section display order (panels are flex items sequenced by CSS `order`), not DOM order, so a user who reordered sections still gets the corners on the right panels.
- CSS renders those panels as segments of one card: side borders on each, top edge and corners on the first, bottom edge, corners and shadow on the last, no per-section headers. Single-panel rail items and related grids are unchanged.
- Guide §7d documents the exception.
- **Rail scrolling** (second commit): the rail items now sit in their own `.mj-forms-chrome-rail-items` wrapper (`flex: 1 1 auto; min-height: 0; overflow-y: auto`), so when a form has more rail items than fit the viewport they scroll instead of being clipped by the panels container. The pin / collapse toolbar stays fixed above them.

### How to reproduce

Open any entity whose form resolves to left-nav (8 or more first-class sections, or `Layout: 'left-nav'`), pick **Details** in the rail. Before: fields render with no card. After: one card.

### Verification

- `@memberjunction/ng-base-forms` vitest: 334/334 (6 new unit tests for the two helpers).
- `tsc --noEmit -p tsconfig.spec.json`: clean.
- Full turbo build: green.
- Live: verified by eye on a real Explorer (MJ Academy shelter `Animal` form) against a copy of that database. Rail scroll verified by forcing the panels container to 260px: items area 172px, content 224px, scrolls (scrollTop 52).

pr4239-details-one-card-2026-09-08

### Changeset

`patch` for `@memberjunction/ng-base-forms` (no migration, no metadata).


rkihm-BC (dismissed):
Correct diagnosis, scoped fix, every gate green: design tokens, buttons, ngc-scss, focus-ring, and the changeset level. The `.mj-chrome-show` rules strip chrome from every shown panel, which is right for one related grid and wrong for Details. Edges follow display order and `OnRailDrop` re-applies them, so a reorder moves the corners correctly.

**Required:** the changeset describes only the card. The second commit's rail-scrolling fix is a separate user-visible change and needs a sentence there, since release notes are built from changesets.

**Optional:** `FormComponent` is already typed `BaseFormComponent | null` with a public `getSectionDisplayOrder`, so the structural cast at `record-form-container.component.ts:1027` can become `this.FormComponent?.getSectionDisplayOrder(k) ?? 0`. A DOM test in the existing container test file for the class composition, including after a reorder, would cover the behaviour the helpers exist for.

**Question:** a contribution pinned with `ChromeGroup: 'details'` is routed to Details before the related-entity check in `resolve-form-chrome.ts` (line 342), so a related grid can land in the card and receive the segment padding on `.mj-forms-panel-content`. Is that supported, or should it be documented as not?

MattC-BC:
@MarceloT-BC - I had Claude completely do this audit - I have a lot of PRs to close as well as create myself, so please take some of this feedback with a grain of salt. 

The Details card itself looks right to me and I'm good with that part. Two things need to land before this merges, plus one CI note.

### Required

**1. The rail-items wrapper breaks the mobile rail.** At 720px and below the existing media block turns the rail into a wrapping row of chips (`flex-direction: row; flex-wrap: wrap`). The new `.mj-forms-chrome-rail-items` wrapper forces `flex-direction: column` with no mobile override, so the items stack vertically again. I rebuilt the container chain and measured at 375px: before this PR, 9 items wrap into 3 rows and all sit inside the rail; after, they form 9 rows, the wrapper is 356px tall inside a 304px rail, doesn't scroll, and the form body gets pushed out of the host. One rule in the `@media (max-width: 720px)` block fixes it:

```css
.mj-forms-chrome-rail-items {
  flex-direction: row;
  flex-wrap: wrap;
  overflow: visible;
  min-height: auto;
}
```

**2. Changeset** still only describes the card (Robert's required item). Needs a sentence on the rail scrolling since release notes come from changesets.

**CI:** the Source guards lane is red on the spec anti-pattern lint, pointing at a geo-core test this PR doesn't touch. `next` already has the fix (319a7edc11 via #4291). A fresh merge of `next` should turn it green.

### Should fix

**Card edges ignore search-hidden panels.** `-first` / `-last` come from the chrome spec in `applyChromeDomVisibility`, not from what's actually visible. The panel hides itself with `mj-search-hidden` when the section filter excludes it, and the container never re-syncs on filter changes (only on structural changes, edit-mode / show-empty toggles, and rail activation). Filter Details down to a middle section and the card loses its top edge, radius, and 20px edge padding. Suggest watching the search filter in `ngDoCheck` the same way edit mode is watched, and skipping `mj-search-hidden` nodes when picking edges. This is exactly the case a DOM test in the container test file would catch (Robert's optional ask).

### Design questions

- **No group labels inside the card.** The existing left-nav rule hides every section header with `display: none`, so two or more field groups run together with 16px between them and no label, visual or accessible. Not caused by this PR, but the guide edit writes "no per-section headers" in as the rule. I'd like to talk about a light group label inside the card, or at minimum a visually-hidden heading so screen readers keep the structure. Happy to defer that to a follow-up if we agree it's separate.
- **8px top margin on the first segment only.** Related grids in the same slot get none, so rail items land at slightly different heights. If it's there to clear the hero, fine, just want to confirm it was deliberate.

### Nits

- The structural cast at `record-form-container.component.ts:1050` isn't needed. `FormComponent` is already `BaseFormComponent | null` with a public `getSectionDisplayOrder`, so `this.FormComponent?.getSectionDisplayOrder(k) ?? 0` works. Same thing Robert flagged.
- Robert's question about `ChromeGroup: 'details'` related grids is still open. For what it's worth, the slot-mounted grid panel binds `[Form]`, so its CSS order matches and the edges would be right; the only oddity is the grid inheriting the card's 20px side padding.

Verified on my side: cascade order beats the older strip rules at equal specificity, the left-nav column has `gap: 0` so segments butt together, all four tokens exist in light and dark, edges use the same `getSectionDisplayOrder` the panel's `style.order` binding uses, and the section-manager reorder path re-resolves. Desktop rail scrolling works as described.

MarceloT-BC:
Thanks — all three points addressed in `66bc667`:

**Required — changeset:** it now also describes the rail-scrolling change (items scroll in their own wrapper; the pin/collapse toolbar stays fixed).

**Optional — cast:** gone; it reads `this.FormComponent?.getSectionDisplayOrder(k) ?? 0` directly. **DOM tests:** three added to `record-form-container.component.dom.test.ts` covering the class composition: Details panels get `mj-chrome-details` with `-first`/`-last` on the visual edges by display order (not DOM order); the edges move when the display order changes; every card class clears when another rail item is active. Plain `<mj-collapsible-panel>` elements appended to the panels column are the fixture, since the container reaches panels by tag + data attributes on purpose.

**Question — a related grid pinned into Details:** you're right that the resolver routes a `ChromeGroup: 'details'` contribution into the Details group before the related-entity check, so a grid could have landed inside the card and picked up the segment padding. That is now explicitly *not* part of the card: `applyChromeDomVisibility` skips `related-entity` panels when tagging `mj-chrome-details` (and when picking the edges), so a pinned grid keeps the chrome-less grid treatment and sits as its own block. Documented in §7d of the forms guide and covered by one of the new tests. If we ever want a grid inside the card it needs its own padding rule; I'd rather keep that out of this PR.

Screenshot of the new look coming in a follow-up comment once the live host is back up on the merged branch.

packages/Angular/Generic/base-forms/src/lib/container/record-form-container.component.css
```diff
@@ -247,6 +293,20 @@ mj-collapsible-panel.mj-form-role-more {
   border-radius: var(--mj-radius-md);
 }
 
+/* The rail is as tall as the body column; when it has more items than fit,
+   the ITEMS scroll (the pin / collapse toolbar above stays fixed). Without
+   this the overflow was clipped by the panels container. */
+.mj-forms-chrome-rail-items {
```
rkihm-BC, line 299: Required: this rule has no counterpart in the `@media (max-width: 720px)` block, so on phones the items now stack one per row inside a wrapping-row rail instead of wrapping as chips. Add inside that block:

```css
.mj-forms-chrome-rail-items {
  flex-direction: row;
  flex-wrap: wrap;
  overflow: visible;
  min-height: auto;
}
```

packages/Angular/Generic/base-forms/src/lib/container/record-form-container.component.ts
```diff
@@ -1038,6 +1039,27 @@ export class MjRecordFormContainerComponent extends BaseAngularComponent impleme
...
+      // padding is sized for field rows, not for AG Grid.
+      const isDetails = layout === 'left-nav' && visible
+        && variant !== 'related-entity'
+        && IsDetailsSectionKey(this.chrome.Spec, key);
+      node.classList.toggle('mj-chrome-details', isDetails);
+      if (isDetails) detailsKeys.push(key);
+    });
+    // The card's top and bottom edges follow the VISUAL order (CSS `order`
+    // = the form's section display order), not DOM order.
+    const edges = DetailsCardEdges(detailsKeys, (k) => this.FormComponent?.getSectionDisplayOrder(k) ?? 0);
```
rkihm-BC, line 1070: Should fix: `detailsKeys` comes from the spec, so a panel that hid itself with `mj-search-hidden` (section filter, or no renderable content while empty fields are hidden) still gets `-first` / `-last` while `display: none`, and the visible card loses that edge. Skip nodes carrying `mj-search-hidden` when collecting `detailsKeys`, and add the search filter to the `ngDoCheck` watch (it currently re-resolves only on edit mode and show-empty) so a filter change re-runs this.

rkihm-BC (changes requested):
My three items are all addressed in `66bc66720b` — changeset covers the rail scroll, the cast is gone, the three container DOM tests pin the class composition, and the pinned-grid question is answered in code (`applyChromeDomVisibility` skips `related-entity` panels for the card and the edges, with a test and a guide sentence). 23/23 checks green; shard 5 ran base-forms with the 53 resolver and 13 container DOM tests passing.

Two findings from @MattC-BC's review (posted fifteen minutes before the fix commit, so I assume they crossed) are still open, and I verified both in the head's code. One is a regression and blocks merge.

### Required — the rail wrapper breaks the mobile rail

At 720px and below the existing media block turns `.mj-forms-chrome-rail` into a wrapping row and gives rail items `width: auto` — that is the chip strip on phones. This PR moves every rail item (More group included) inside `.mj-forms-chrome-rail-items`, whose only rule is the desktop one: column direction, vertical scroll, hidden horizontal overflow, with no counterpart in the media block. Inside a column wrapper, `width: auto` items stretch and stack one per row, so the chip layout is gone at every width under 720px — certain from the rules alone. Matt's measurement (a 356px wrapper in a 304px rail that does not scroll and pushes the body out at 375px) is the consequence. His override inside the `@media (max-width: 720px)` block is the fix; inline comment.

### Should fix — card edges ignore hidden panels, and not only under search

Traced end to end:
- a panel adds `mj-search-hidden` (`display: none`) whenever its own `IsVisible` is false — when the section filter excludes it, **and also with no filter at all when it has no renderable content**, i.e. a section whose fields are all empty while empty fields are hidden;
- `applyChromeDomVisibility` picks `-first` / `-last` from the chrome spec's section keys and never consults that class;
- a chrome resolve is scheduled on panel/slot changes, edit-mode and show-empty toggles, chrome-rule load and reorder/membership — nothing fires on a filter change, which flows from `searchFilter` into `FormContext.sectionFilter` and straight to the panel.

So a hidden first or last Details section keeps its edge class while invisible and the visible card loses its top or bottom border, radius and edge padding. The all-empty-section case is the common one and survives even the show-empty re-resolve, because the edge picker still ignores hidden nodes. Both halves Matt named are needed: skip `mj-search-hidden` nodes when picking edges, and watch the filter in `ngDoCheck` the way edit mode is watched. The new DOM tests are the natural place to pin it. Inline comment.

### Follow-ups, not blocking

Matt's two design questions deserve a reply here even if deferred: a visually-hidden heading per group inside the card so screen readers keep the structure (the left-nav rule hides every section header), and whether the 8px top margin on the first segment only — related grids in the same slot get none — is deliberate.

MarceloT-BC:
Read both reviews (Matt's comment and Robert's changes-requested review) at 2026-09-08 19:06 UTC. Everything is in `3d67e95712`, on top of a fresh merge of `next` (`046376a8a2`).

@MattC-BC @rkihm-BC

**Required 1 — mobile rail (both of you):** fixed exactly as suggested. The `@media (max-width: 720px)` block now gives `.mj-forms-chrome-rail-items` `flex-direction: row; flex-wrap: wrap; overflow: visible; min-height: auto`, so the chip strip is back under 720px. Thanks for measuring it — I only checked desktop.

**Required 2 — changeset (Matt):** landed in `66bc667` (our comments crossed); the changeset describes both the card and the rail scrolling.

**CI note (Matt):** `next` is merged in again as of this push, so the Source guards lane should pick up #4291.

**Should fix — card edges vs hidden panels (both):** done, both halves. `applyChromeDomVisibility` skips nodes carrying `mj-search-hidden` when picking `-first`/`-last`, and `ngDoCheck` now watches the section filter alongside edit mode and show-empty, so a filter change re-runs the chrome pass after the panels' own host classes are current. Robert's all-empty-section case is covered by the same skip (the panel hides itself with the same class, and show-empty was already watched). New container DOM test: a hidden first section holds no edge, the first visible segment closes the card, and the edge comes back when the panel shows again. The ngDoCheck harness gained a filter-change case.

**Design question — group labels inside the card (Matt):** the one-card / no-section-header look is the product call for this PR (Marcelo asked for exactly that after seeing the headed version). I agree screen readers lose the structure. Proposal: a follow-up that renders each Details panel's title as a visually-hidden heading inside the card instead of `display: none`. Happy to do it, kept out of here to keep the PR small.

**Design question — 8px top margin on the first segment (Matt):** deliberate, to separate the card from the hero; the related grid has no border so it can sit flush. I can drop it or give the grid the same offset if you'd rather the content top never shifts between rail items — say which and I'll push it.

**Nits (Matt):** the cast is gone since `66bc667`; the pinned-grid question is answered in code (a `related-entity` panel routed into Details stays chrome-less and never becomes an edge, test + guide sentence). Your point that its CSS order would have been right is well taken; the side padding was the only oddity and that is what the exclusion avoids.

pr4239-details-one-card-2026-09-08

packages/Angular/Generic/base-forms/src/lib/container/record-form-container.component.css
MarceloT-BC: Fixed in `3d67e95712`: the `@media (max-width: 720px)` block now has exactly this rule for `.mj-forms-chrome-rail-items` (row, wrap, visible overflow, `min-height: auto`), so the chip strip is back under 720px.

packages/Angular/Generic/base-forms/src/lib/container/record-form-container.component.ts
MarceloT-BC: Fixed in `3d67e95712`, both halves: `applyChromeDomVisibility` skips nodes carrying `mj-search-hidden` when collecting `detailsKeys` (so a self-hidden panel — filter or no-renderable-content — never holds `-first`/`-last`), and `ngDoCheck` now watches `EffectiveSearchFilter` alongside edit mode and show-empty, scheduling the chrome re-resolve after the panels' host classes are current. Covered by a new container DOM test (hidden first section holds no edge, first visible segment closes the card, edge returns when shown) and a filter-change case in the ngDoCheck harness.

MattC-BC:
Approving. Re-measured the mobile rail with the shipped rule: 3 wrapped rows, all chips inside the rail, body visible, same numbers as before the wrapper existed. The hidden-panel edge fix covers both the filter and the all-empty case, and the new container DOM tests are the right shape. Thanks for turning both reviews around so fast.

On the two open questions: keep the 8px, the card edge needs the breathing room and the borderless grid doesn't, so the offset is optical alignment rather than inconsistency. And yes to the visually-hidden heading as a follow-up. Please file the issue so it has a home.

@rkihm-BC your changes-requested is on the commit before the fix, so it needs a fresh look from you too.

MattC-BC (approved):

rkihm-BC (approved):
Both required items are fixed exactly as asked, in `3d67e95712`. Approving.

- **Mobile rail.** The `@media (max-width: 720px)` block now carries `.mj-forms-chrome-rail-items` with `flex-direction: row; flex-wrap: wrap; overflow: visible; min-height: auto` — the override verbatim, and @MattC-BC re-measured it at three wrapped rows with the body visible.
- **Card edges vs hidden panels.** `applyChromeDomVisibility` skips nodes carrying `mj-search-hidden` when collecting `detailsKeys`, and `ngDoCheck` watches the section filter alongside edit mode and show-empty. That covers both cases I raised — the filter and the all-empty section — because the panel uses the same host class for both, which is exactly why one skip serves both.
- **Tests pin each half**: a container DOM case where a hidden first section holds no edge, the first visible segment closes the card, and the edge returns when the panel reappears; plus an `ngDoCheck` filter-change case.

23/23 checks green. Thanks for the fast turnaround, and for answering both design questions rather than deferring them silently — the visually-hidden heading follow-up is the right call.

PR #4240, opened 2026-09-04
fix(ng-base-forms): budget the horizontal scrollbar in fit-content related grids (#4223)

### LINKED ISSUE #4223: fit-content related grid clips its last row: RelatedGridHeightPx does not budget for a horizontal scrollbar
## Symptom

A `fit-content` related-entity grid in a narrow container renders its last row **sliced in half**. With one row the data row is clipped to roughly half its height and the grid looks broken.

Measured on the affected grid (Chrome, `mj-explorer-entity-data-grid`, `Height="fit-content"`, 1 row):

| Element | Height |
|---|---|
| host | 137px |
| `.ag-root-wrapper` | 87px |
| **`.ag-body-viewport`** | **21px** |
| rendered row | ~27px |

## Root cause

`RelatedGridHeightPx` in `packages/Angular/Generic/base-forms/src/lib/related-grid-height.ts` budgets toolbar + header + rows + pad, and **nothing for a horizontal scrollbar**:

```ts
export const RELATED_GRID_TOOLBAR_PX = 52;
export const RELATED_GRID_HEADER_PX = 40;
export const RELATED_GRID_ROW_PX = 40;
export const RELATED_GRID_BOTTOM_PAD_PX = 5;

export function RelatedGridHeightPx(rowCount: number, maxHeight?: number | null): number {
    const rows = Number.isFinite(rowCount) ? Math.max(0, Math.floor(rowCount)) : 0;
    const body = rows === 0
        ? RELATED_GRID_EMPTY_BODY_PX
        : RELATED_GRID_HEADER_PX + rows * RELATED_GRID_ROW_PX;
    const raw = RELATED_GRID_TOOLBAR_PX + body + RELATED_GRID_BOTTOM_PAD_PX;
    …
}
```

For one row that is `52 + 40 + 40 + 5 = 137` — exactly the measured host height, so the formula is doing what it says. The height is then handed to AG Grid, whose horizontal scrollbar is laid out **inside** that box and consumes ~15-26px of the body viewport. The row loses the space the formula assumed it had.

The scrollbar only appears when the columns are wider than the container, so this is invisible on a wide form and reproducible on a narrow one — which is why it can sit unnoticed.

## How to recreate

1. Put an `mj-explorer-entity-data-grid` with `Height="fit-content"` inside a narrow container (a contributed form panel, or a left-nav group in a ~540px column).
2. Point it at an entity whose default columns are collectively wider than that container, so a horizontal scrollbar appears.
3. Load a small number of rows — one makes it most obvious.
4. The last row is clipped. Widen the window until the scrollbar disappears and the clipping goes with it.

## Suggested fix

Make the allowance **conditional** rather than a flat constant — a grid with no overflow should not lose height it does not need:

- Detect actual horizontal overflow on the body viewport (`scrollWidth > clientWidth`) and add the measured scrollbar thickness (`offsetHeight - clientHeight` on the viewport, which is 0 for overlay scrollbars) to the computed height.
- Recompute on the same signals that already drive `sizedHeightPx` — `onDataLoad`, plus a resize observer, since overflow appears and disappears as the container is resized rather than only when data loads.
- Keep the pure function testable by passing the measurement in, e.g. `RelatedGridHeightPx(rowCount, maxHeight, scrollbarPx = 0)`, so the existing unit tests stay valid and the DOM measurement lives in the component.

A fixed `+15` would regress every grid that does not overflow, and would still be wrong on platforms with overlay scrollbars (macOS trackpad) where the thickness is 0 — hence measuring rather than assuming.

## Workaround

Use `Height="auto"` and give the parent a definite height (`min-height` plus a flex column), which sidesteps the calculation entirely.

MarceloT-BC:
## Summary

A `fit-content` / related-entity grid in a narrow container renders its last row sliced in half. `RelatedGridHeightPx` budgets toolbar + header + rows + pad and hands that box to AG Grid, which lays its horizontal scrollbar out **inside** the box — so whenever the columns are wider than the container, the scrollbar (15–17px on classic-scrollbar platforms) is taken out of the row area.

Closes #4223

**Relationship to #4245 (merged 5 Sept):** #4245 fixed the same clip with a fixed `RELATED_GRID_HSCROLLBAR_PX = 17` reserve whenever rows are shown, plus live-measured chrome constants. This PR is reconciled on top of it (merge `819c45242d`, see the thread below): the fixed reserve stays as the **fallback** for callers that do not measure, and `<mj-explorer-entity-data-grid>` replaces it with the **measured** scrollbar — 0 when the columns fit or the platform draws overlay scrollbars, the real height otherwise — so no grid carries 17px of slack it does not need, and a bar that appears or disappears on resize is budgeted or released without a reload.

## Change

- `RelatedGridHeightPx(rowCount, maxHeight, scrollbarPx?)` — a third, optional parameter for the measured scrollbar height. A finite, non-negative value is a measurement and **replaces** #4245's fixed reserve (0 releases it entirely; a real bar is budgeted exactly), never past `maxHeight`. Omitting it, or passing a negative / non-finite value, keeps the fixed reserve, so every existing caller's result is exactly what `next` returns today.
- `<mj-explorer-entity-data-grid>` **measures** the allowance instead of assuming it: the height AG Grid's fake horizontal scroller (`.ag-body-horizontal-scroll`) takes in the flow, and 0 when it is collapsed or absolutely positioned (overlay scrollbars — `.ag-scrollbar-invisible` — draw over the rows and take no layout space). It starts **unmeasured** (`undefined`), so on a classic-scrollbar platform the grid keeps the reserve until the first real measurement lands two frames after load (after AG Grid's own `setScrollVisible` frame), instead of briefly sizing without one. It re-measures on a `ResizeObserver` over the host and the scroller.
- Chrome constants: the live-measured set #4245 landed (toolbar 49, header 49, wrapper borders 4, bottom pad 2, empty body 56) is kept as-is; this PR only documents the header constant (the old 40px left every related grid 9px short) and the reserve's new fallback role.

## How to recreate (and verify)

Explorer at a 900px-wide viewport, classic scrollbars (Windows, or macOS "Always show scroll bars"): open **MJ: AI Models → AssemblyAI Voice Agent → AI Model Vendors** (1 row, columns wider than the panel).

Measured with Playwright + system Chrome on this branch **before** the #4245 reconciliation (constants then: toolbar 52 / header 49 / pad 5, so absolute heights are 2px higher than the current constants produce; the clip column is what matters):

| | host | `.ag-body-viewport` | fake scroller | last row clipped |
|---|---|---|---|---|
| before, classic scrollbars | 137px | **21px** | 15px, in flow | **19px of 40** |
| scrollbar fix only, classic scrollbars | 152px | 36px | 15px, in flow | 4px (the header gap, since fixed) |
| before, overlay scrollbars | 137px | 36px | 16px, absolute (no space) | 4px (the header gap, since fixed) |
| **final**, overlay scrollbars | 146px | 45px | 16px, absolute (no space) | **0** |

The 21px "before" viewport is exactly the number in the issue. The final classic-scrollbar case was not re-observed live: the machine's macOS scrollbar mode flipped to overlay mid-session. On that platform the grid now shows #4245's already-shipped reserve until the measurement lands, and the measured 15px term is covered by the unit + DOM tests and the earlier live run.

Follow-up (separate feature request, not this PR): measure the whole chrome (toolbar + header + scrollbar + borders) after layout instead of budgeting it from constants, so theme changes cannot re-open this.

## Tests

- `related-grid-height.test.ts` (9): #4245's cases unchanged, plus a measurement replaces the reserve (0 releases it; a measured bar on the empty state is still budgeted), non-finite / negative values keep the reserve, and the `maxHeight` cap stays authoritative over the allowance.
- `explorer-entity-data-grid.component.dom.test.ts` (20): scroller height budgeted after load; a mounted scroller measuring 0 releases the reserve; nothing budgeted for an absolutely positioned (overlay) scroller; `MaxHeight` still honoured; re-measure on host resize (0 → 15 → 0); observer created lazily for the panel-ancestry path; observer disconnected on destroy.
- At head `471689342b` (after the latest `next` merge): `@memberjunction/ng-base-forms` **390/390 (28 files)**, `turbo run build --filter='@memberjunction/ng-base-forms...'` 48/48 clean.


MattC-BC:
Re-approving after 971179732b. All three nits landed: the spec block comments now match the corrected source JSDoc, the two older dropdown-test comments describe the two-source composition instead of the old behaviour, and the datepicker LOCKS spec opens the calendar first and pins close-on-lock (the single-line mutation check is exactly the proof I'd want). Description is refreshed too. Component sources are unchanged from the head I approved, the diff is still the same 11 files, and CI is green on the new head. Good to merge.

MattC-BC (dismissed):

rkihm-BC (dismissed):
Implements the issue's suggested design faithfully. Overlay scrollbars are detected by the absolute position AG Grid gives the fake scroller, so trackpad users get no phantom allowance. The re-measure loop is safe: the observer fires after the component's own height change, the value is unchanged, and it returns early. Scheduling is coalesced, runs outside the zone, and is cancelled on destroy. The header constant change is justified: MJ's theme sets no `headerHeight`, so Alpine's default applies, and 48 plus a 1px border was measured live. Nothing else pins the old numbers. Changeset level correct.

**For other reviewers:** the dismissed approval was on the merge commit before the 40 → 49 header change, and that change shifts every related grid by 9px on every platform, not only the overflow case. The comment referring to commit 971179732b and datepicker specs belongs to #3830.

**Non-blocking:** the final classic-scrollbar state was not re-observed live, as the PR says. One screenshot from a Windows machine before merge would close that.

rkihm-BC:
Merged `origin/next` into this branch as `819c45242d` to clear the conflict. It was a design conflict, not a textual one, so recording the resolution here for @MarceloT-BC and @cadam11 to sanity-check.

**What conflicted.** #4245 (Craig, merged 5 Sept) fixed the same #4223 clip while this PR was open: a fixed `RELATED_GRID_HSCROLLBAR_PX = 17` reserve added whenever rows are shown, plus live-measured chrome constants (toolbar 49, header 49, wrapper borders 4, bottom pad 2, empty body 56). This PR measures the real scrollbar from the DOM instead of assuming it. Taking either side verbatim would have discarded the other.

**Resolution — both kept, reconciled:**
- Constants are #4245's live-measured set. `RELATED_GRID_HSCROLLBAR_PX` stays, as the **fallback** reserve.
- `RelatedGridHeightPx(rowCount, maxHeight, scrollbarPx?)`: a finite, non-negative `scrollbarPx` is a measurement that **replaces** the reserve — 0 releases it, a real bar is budgeted exactly. Omitting it (or passing a negative / non-finite value) keeps #4245's fixed reserve, so an unmeasured caller's result is exactly what `next` returns today.
- `<mj-explorer-entity-data-grid>` now starts **unmeasured** (`scrollbarPx: number | undefined`) and `measureHorizontalScrollbarPx` returns `undefined` until AG Grid has rendered `.ag-body-horizontal-scroll`, rather than reporting 0. So on a classic-scrollbar platform the grid keeps the reserve until the first real measurement lands two frames after load, instead of briefly sizing without one. That is the one behavioural change to this PR's own logic, and I think it is an improvement — say so if you disagree.
- Tests: #4245's pass unchanged; this PR's three unit cases and the fake-scroller DOM cases are rewritten to the reconciled contract (a mounted scroller measuring 0 is a *measured* 0, and a measurement replaces rather than adds to the reserve). Changeset text updated to match.

**Verified in a clean worktree at the merge commit:** `related-grid-height.test.ts` 9/9, `explorer-entity-data-grid.component.dom.test.ts` 20/20, full `@memberjunction/ng-base-forms` suite 27 files / 369 tests, package build clean.

If you would rather have picked one design outright, the merge commit is easy to redo — nothing else on the branch changed.

MarceloT-BC:
@rkihm-BC thanks for the reconciliation — sanity-checked, and I agree with it, including the one behavioural change.

**Reconciled contract:** measurement replaces the reserve, omitted/invalid keeps it. That is the right split: an unmeasured caller gets exactly what `next` returns today, and the component is the only thing that knows better. The `RelatedGridHeightPx(0, null, 8)` case (a measured bar on the empty state is still budgeted, while the *fixed* reserve is not) is also right — the empty body can overflow horizontally, and only a measurement can tell.

**Starting unmeasured:** agreed it is an improvement. Reserve-first means a classic-scrollbar platform never shows a clipped row, not even for two frames; the cost is a ≤17px shrink on overlay platforms once the measurement lands, which is a far better first paint than a clip. `undefined === undefined` short-circuits in `applyScrollbarMeasurement`, so the pre-render path never triggers a spurious change-detection pass either.

**Re-verified at the current head `471689342b`** (after the latest `next` merge, which brought ~200 commits incl. the Angular 21.2.22 bump): `@memberjunction/ng-base-forms` 390/390 (28 files) — `related-grid-height.test.ts` 9/9, `explorer-entity-data-grid.component.dom.test.ts` 20/20 — and `turbo run build --filter='@memberjunction/ng-base-forms...'` 48/48 clean. The PR description is refreshed to describe the reconciled design rather than the pre-#4245 one.

**Classic-scrollbar screenshot:** still not something I can produce from this Mac (it is stuck in overlay mode). Two notes that soften it: (1) on that platform the pre-measurement state is #4245's reserve, which is already shipped on `next`, and the post-measurement state differs from it only when the bar is not 17px; (2) the 15px in-flow measurement was observed live earlier in the classic-scrollbar run in the table. If anyone has a Windows box handy before the 6.1 cut, the repro in the description takes a minute; otherwise I would not hold the PR on it.

Both approvals show as dismissed by the merge commits, so this needs a fresh review from you (or @cadam11, since the reconciliation is half his design).

rkihm-BC (approved):
## Approve

Fresh review at `da28c608a9`, since both earlier approvals were dismissed by merge commits. @MarceloT-BC — thanks for the sanity-check on the reconciliation, and for re-verifying at `471689342b`.

**Disclosure first:** I authored the reconciliation merge, so I am partly reviewing my own design call. Marcelo has independently sanity-checked it and agreed, including the start-unmeasured change, and the other half of the design is @cadam11's #4245. If anyone wants a second pair of eyes on the *merge* rather than the code, Craig is the right person.

### One caveat from my earlier review is now moot

I flagged for other reviewers that the 40 → 49 header change "shifts every related grid by 9px on every platform, not only the overflow case." **That no longer applies.** `next` already carries 49 through #4245, and this branch's constants are identical to it — the only thing left in that part of the file is an expanded comment. This PR now changes no chrome constant at all.

### What I re-verified this pass

The whole point of the reconciliation was that an unmeasured caller must get exactly what `next` returns, so I checked the arithmetic by hand rather than trusting the contract:

| | `next` | this branch |
|---|---|---|
| 3 rows, unmeasured | 49 + (49 + 120 + 17) + 4 + 2 = **241** | 49 + (49 + 120) + 4 + 2 + 17 = **241** |
| 0 rows, unmeasured | **111** | **111** |

The scrollbar term simply moved out of `body` into `raw`. Identical.

The measured path does what the JSDoc says: replaces the reserve, applies even at zero rows (a measured bar on the empty state is real — only a measurement can know), and treats negative or non-finite as "not a measurement". `measureHorizontalScrollbarPx` returns `undefined` before AG Grid renders the scroller, `0` when it is positioned absolutely, and `offsetHeight` otherwise, and it re-observes the scroller so a 0 ↔ N flip is caught without a host resize. `applyScrollbarMeasurement` short-circuits on `undefined === undefined`, so the pre-render path costs no change-detection pass.

Only two production call sites exist, both inside this component — so the fallback contract is protecting the exported API for Open Apps rather than any in-repo caller, which is still the right reason to keep it.

The re-measure loop remains safe: the observer fires after the component's own height change, the value is unchanged, and it returns early. `Math.min(maxHeight, raw)` damps the capped case, and height does not affect horizontal overflow, so there is no oscillation path I can construct.

Tests pin all three scrollbar shapes plus the `MaxHeight` cap, host resize, late observer creation and destroy cleanup. Changeset is accurate and `patch` is right.

### The classic-scrollbar screenshot — no longer worth holding for

I raised this as non-blocking before and it is weaker now. Marcelo's two points hold: on a classic-scrollbar platform the pre-measurement state is #4245's reserve, which already ships on `next`, and the 15px in-flow measurement *was* observed live in the classic run recorded in the description's table. With all three code paths unit-pinned, a screenshot would confirm what the measurement already demonstrated. Nice to have before the 6.1 cut if a Windows box is handy; not a merge condition.

### CI

11 green at the time of writing, with `Build` and the deterministic integration tier still running on this head. If either goes red, treat this approval as not yet in force and ping me.

PR #4245, opened 2026-09-05
fix(ng-base-forms): reserve horizontal-scrollbar height in related-grid sizing

cadam11:
Found during BlueCypress UAT end-to-end testing on the AIDP local stack (no linked issue; screenshot on Craig Adam's side). On an Organization record's related Orders tab, the embedded related-entity grid rendered about a row and a half tall with its single row clipped mid-glyph.

## Root cause

`RelatedGridHeightPx` budgets toolbar + header + rows + a 5px bottom pad — and nothing for AG Grid's horizontal scrollbar. The bar renders inside the grid viewport, so on any related entity whose columns overflow the panel width (the norm for wide entities like Order Headers inside a form) it consumes 15–17px of the row area. With one row, that clips the row mid-glyph; the "empty space below" the panel is untouched because the container height itself was computed too small.

## Fix

A new `RELATED_GRID_HSCROLLBAR_PX = 17` reserve, added to the body whenever rows are shown. Overlay-scrollbar platforms just gain a little slack; zero-row grids keep the short empty body since there is no row to clip. The `maxHeight` cap still applies after the reserve.

## Verification

- Unit test added: the delta between a one-row and zero-row grid now includes the scrollbar reserve, and a one-row grid is at least toolbar + header + row + scrollbar tall.
- `pnpm test` in ng-base-forms: 26 files, 330 tests, all passing.
- `pnpm run build` (ngc) clean.

### DIFF
--- a/.changeset/related-grid-chrome-budget.md
+++ b/.changeset/related-grid-chrome-budget.md
@@ -1,0 +1,5 @@
+---
+"@memberjunction/ng-base-forms": patch
+---
+
+Related-entity grid heights now budget the real rendered chrome: a reserve for AG Grid's horizontal scrollbar (which was clipping the last row mid-glyph whenever columns overflowed), the measured 49px toolbar and header rows, and the 4px of wrapper borders. Fixes both the clipped row and the needless vertical scrollbar on grids whose rows all fit.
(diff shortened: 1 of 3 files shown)


cadam11:
Round 2 (bcc0699c21): live measurement against the rendered theme showed the original constants under-budgeted the chrome — toolbar and header are each 48px + 1px border (not 52/40), and the wrapper + ag-root-wrapper borders (4px) were never counted, so the body viewport came up 2px short of one row and AG Grid answered with a needless vertical scrollbar. Constants now match measured reality, a RELATED_GRID_BORDERS_PX carries the border budget, and the bottom pad drops to 2px. Verified live on a 1-row related grid: no vertical scrollbar, row fully visible, with and without the horizontal bar. 332/332 tests pass.

PR #4249, opened 2026-09-05
Event-driven metadata refresh from dataset membership + authoritative dataset-status oracle

jordanfanapour:
## Problem

MJAPI never refreshes its own in-memory metadata after a permission-bearing save it itself processes. The only refresh trigger is the periodic `RefreshIfNeeded()` poller (default 180s), and its staleness check is a **closed loop**: `GetDatasetStatusByName` Phase 1 derived each MJ_Metadata item's "remote" timestamp as MAX(`__mj_UpdatedAt`) over the server's **own cached dataset slots** (gated on `TrustLocalCacheCompletely`). If the write path fails to maintain a slot, that slot reports itself current forever — proven live: with a 15s tick, zero `GetAllMetadata()` calls across a multi-minute window containing real permission-entity saves. Consequence: a tightened field-level-security rule was served **until process restart** — a restricted user's list queries kept receiving a denied column, and denied-field predicates executed instead of being rejected.

Clients had no event-driven metadata refresh at all — a browser loaded metadata once per page load.

## Fix

**1. Dataset-membership-driven refresh (`ProviderBase`, both tiers).** Metadata is built from the `MJ_Metadata` dataset, so the dataset's item list *is* the authoritative declaration of which entities can stale it. `GetAllMetadata` now records the item entity names from the loaded dataset (`registerMetadataDatasetMembership`; persisted beside the metadata snapshot so warm-booting clients keep coverage). The existing static write-invalidation fan-out additionally routes save/delete/remote-invalidate events to `handleMetadataMemberEntityEvent`: a write to a member entity schedules a debounced (500ms, burst-coalescing — enabling field security writes one row per field/role pair) refresh of the provider that loaded the dataset, gated by a fail-open backend-identity check for multi-provider processes. **No entity names anywhere** — adding a `DatasetItem` row extends coverage by itself. Per-request server providers never load datasets (they adopt the global's shared metadata shell), so only the process-global provider reacts, with no special-casing.

**2. The staleness oracle is authoritative.** `GetDatasetStatusByName` no longer has a cache-derived Phase 1: status is always the batched SQL MAX/COUNT per item. The cache stays fully in play for the *data* reads in `GetDatasetByName`. The status query also now composes the stored item `WhereClause` with the runtime filter, matching the data read (the old SQL path ignored the stored clause; NULL for all shipped items, so no behavior change for them). This heals every dataset-status consumer — including clients' over-the-wire `RefreshIfNeeded` — and makes the periodic tick a genuine backstop for out-of-band (direct SQL / migration) changes.

**3. Throttle bypass for event-driven checks.** `CheckToSeeIfRefreshNeeded` / `RefreshIfNeeded` accept an optional `bypassMinCheckInterval`. Event-driven callers hold positive evidence a member entity was just written; without the bypass, the 30s min-check throttle silently drops the second of two permission changes made inside one window.

**Refresh policy per tier** (`RefreshAfterMetadataMemberChange`, overridable): database providers hard-`Refresh()` — the process that performed the write must not trust any cache layer for the re-read; `GraphQLDataProvider` runs the (now trustworthy) staleness check with the throttle bypassed, so a browser only re-pulls the full graph when genuinely stale.

## Behavior

A permission change now propagates in **~1–2 seconds on the server instance that performed the write**: one full metadata reload per debounced burst, in the background — in-flight and new requests keep the old graph until the atomic whole-object swap, so nothing blocks. In a multi-instance deployment, OTHER MJAPI instances converge on their periodic refresh tick (the Redis receiver dispatches cache callbacks but raises no BaseEntity events) — which the oracle fix makes genuinely reliable for the first time. Connected browsers converge within a long randomized client-side coalescing window (15–45s; see review changes below) — display freshness only, since enforcement is server-side either way. Measured full-reload cost ~2s against local SQL Server (383 entities). Writes to non-member entities cost one O(1) set lookup.

## Tests

- 13 new MJCore unit tests (`providerBase.metadataMemberRefresh.test.ts`): membership registration from the dataset definition, debounce + 84-event burst coalescing, non-member ignore, backend matching (fail-open), base-vs-client refresh policy, failure swallowing, teardown cancel, end-to-end through the event bus, and the throttle bypass.
- `dataset-caching.test.ts` updated to pin the new oracle contract (status always from SQL even with warm cache slots; WhereClause composition).
- New GraphQL test pinning the client refresh policy (`RefreshIfNeeded(undefined, true)`, never hard `Refresh`).
- Suites green: MJCore 2211, GenericDatabaseProvider 965 (+5 skipped), GraphQLDataProvider 426.
- Validated live on the FLS branch this fixes (IT89 over-the-wire enforcement went from a reproducible leak to 4/4, full deterministic tier green); that branch will rebase onto this one.


packages/GraphQLDataProvider/src/graphQLDataProvider.ts
```diff
@@ -3716,10 +3721,26 @@ export class GraphQLDataProvider extends ProviderBase implements IEntityDataProv
...
+     * dataset definition, not a list). A browser must not re-pull the full metadata graph on
+     * every such write, so instead of the base class's hard Refresh this runs the staleness
+     * check: `RefreshIfNeeded` compares server timestamps and refetches only when genuinely
+     * stale, so a spurious event costs one cheap status round-trip. The check-interval throttle
+     * is bypassed because the caller holds positive evidence a member entity was just written —
+     * the throttle would otherwise silently drop the second of two permission changes made less
+     * than its window apart.
+     */
+    protected async RefreshAfterMetadataMemberChange(): Promise<boolean> {
+        return this.RefreshIfNeeded(undefined, true);
```
rkihm-BC, line 3779: This is the one change I'm asking for. Every connected browser receives every save of every entity (`index.ts` global listener → `CacheInvalidationResolver`, no filter), and MJ_Metadata's members include `MJ: Dashboards`, `MJ: Queries` (+5 children) and `MJ: Libraries`. So each such write by any user or agent makes this run on every client: a status round-trip, a full gunzip + re-parse of the IndexedDB snapshot inside `CheckToSeeIfRefreshNeeded`, and — since the member timestamp really moved — a full 6–10 MB metadata pull, all within the same 500 ms window. Before this PR no client code called `RefreshIfNeeded` after startup.

Suggest a client-specific coalescing window with jitter in this override (tens of seconds, randomized), e.g. re-arm a timer with `MetadataDatasetRefreshDebounceMs` replaced by a `ClientMemberRefreshWindowMs + random jitter`, so any write rate costs each browser at most one check + one pull per window. Server-side enforcement keeps its 1–2 s path.

packages/MJCore/src/generic/providerBase.ts
```diff
@@ -510,6 +511,141 @@ export abstract class ProviderBase implements IMetadataProvider, IRunViewProvide
...
+     * Debounced metadata refresh after a write to a metadata member entity. The refresh targets
+     * THIS instance — the provider that loaded the dataset owns the metadata built from it;
+     * short-lived per-request providers never load the dataset (they adopt the global's metadata
+     * as a shared shell), so on the server only the process-global provider ever gets here.
+     */
+    protected scheduleMetadataMemberRefresh(): void {
+        if (this._metadataMemberRefreshTimer) {
+            clearTimeout(this._metadataMemberRefreshTimer);
+        }
+        this._metadataMemberRefreshTimer = setTimeout(() => {
```
rkihm-BC, line 656: Recommended, not blocking: `Config` resets `_refresh` before awaiting `GetAllMetadata`, so if another member write lands more than 500 ms after the previous one while that reload is still running, a second full reload starts concurrently; whichever finishes last wins, so an older snapshot can overwrite a newer one until the next tick. A single in-flight promise plus a "refresh again when done" flag here would make bursts that outlast the debounce cost exactly one follow-up reload.

.changeset/metadata-membership-refresh.md
```diff
@@ -0,0 +1,19 @@
+---
+"@memberjunction/core": patch
+"@memberjunction/generic-database-provider": patch
+"@memberjunction/graphql-dataprovider": patch
+---
+
+Event-driven metadata refresh from dataset membership, and an authoritative dataset-status oracle.
+
+**The problem.** A server process never refreshed its own in-memory metadata after a permission-bearing save it itself processed: the only trigger was the periodic `RefreshIfNeeded()` poller, and its staleness check (`GetDatasetStatusByName` Phase 1) derived each MJ_Metadata item's "remote" timestamp from the server's **own cached dataset slots** — a closed loop. A tightened field-security rule was therefore not enforced over the wire until process restart. On the client, the equivalent refresh hung off a hardcoded entity-name list.
```
rkihm-BC: Required but trivial: there is no hardcoded entity-name list on `next`, and this diff removes none (`GraphQLDataProvider` only gains lines). It presumably lives on the FLS branch. Since release notes are generated from changesets, please drop or reword this sentence so the notes describe what this PR changes.

rkihm-BC (changes requested):
Reviewed the description, the commit, the full diff, and traced the surrounding code on `next`: the event fan-out, `Config`/`Refresh`, the MJ_Metadata dataset definition, the server's invalidation broadcaster and the client's storage path. All checks are green, including the deterministic integration tier.

**The diagnosis and the fix are right.** Deriving dataset status from the same cache slots whose freshness is under test was a closed loop, and making the SQL aggregate authoritative is the only sound answer. Membership from the dataset definition is the correct source of truth, the plumbing reuses the existing static fan-out (no new subscriptions, one Set lookup per non-member write), the backend check fails open, the debounce is re-armed per event, teardown cancels the timer, and the tests are thorough. Changeset present at the right level for the right three packages. The `WhereClause` composition is a genuine latent-bug fix.

**One change I'm asking for before merge: bound the client-side event-driven refresh.** The description frames the trigger as "a permission change", but MJ_Metadata has 30 items and they include entities that ordinary users and agents write routinely: `MJ: Dashboards`, `MJ: Dashboard Categories`, `MJ: Queries` and its five children, `MJ: Libraries`. Each such write now costs every connected browser a full metadata cycle — the chain is verified end to end:

- Explorer subscribes to the invalidation stream (`initialization.service.ts:65`, workspace initializer).
- The global listener in `MJServer/src/index.ts` publishes every save of every entity, and `CacheInvalidationResolver` delivers it to every browser with no filter.
- The browser provider is registered in the fan-out, the event carries `provider: this`, `AllowRefresh` is true, so `RefreshAfterMetadataMemberChange` runs `RefreshIfNeeded(undefined, true)`: one status round-trip (30 aggregate queries server-side), then `LoadLocalMetadataFromStorage` gunzips and re-parses the whole snapshot, then — because the member table's timestamp genuinely moved — `LocalMetadataObsolete()` is true and the client pulls the full graph and re-compresses it into IndexedDB.
- Before this PR nothing on the client called `RefreshIfNeeded` after startup (checked all of Angular, the GraphQL provider and MJExplorer). Browsers loaded metadata once per page load.

`CACHING_AND_PUBSUB_GUIDE.md` puts the metadata payload at 6–10 MB of JSON per tenant. So one user saving a dashboard, or an agent creating a query, now makes every connected browser download and re-parse 6–10 MB within the same half-second, and the saver's own browser does the same via its local save event. With 200 sessions that is roughly 1.5 GB of egress per dashboard save. The fix is small and fits inside the mechanism you built: a long, randomized coalescing window in the GraphQL override (tens of seconds) so any amount of org-wide write activity costs each browser at most one status check and one pull per window. Server-side enforcement — where field security actually lives — keeps its 1–2 s path untouched. Inline comment at the override.

**Also required, trivial:** the changeset says the client refresh "hung off a hardcoded entity-name list". No such list exists on `next` and nothing is removed in this diff (it presumably lives on the FLS branch). Since release notes are built from changesets, please drop or reword that sentence. Inline comment.

**Recommended, not blocking:**

- **Single-flight on the server reload.** `Config` clears `_refresh` before awaiting `GetAllMetadata`, so a member write arriving >500 ms after the previous one while a reload is still running starts a second concurrent full reload; if the earlier one finishes last it overwrites the newer result. The window is bounded by the next tick, so this is a recommendation. Inline comment.
- **Cross-server scope.** The Redis receiver in `index.ts` dispatches to `LocalCacheManager` callbacks and rebroadcasts to browsers but never raises a `BaseEntity` event, so a second MJAPI instance still waits for its 180 s tick (now correctly, thanks to the oracle fix). Worth a sentence in the description: "~1–2 s" holds on the instance that performed the write.
- `IMetadataProvider.RefreshIfNeeded` / `CheckToSeeIfRefreshNeeded` in `interfaces.ts` weren't widened with the new optional parameter, so interface-typed callers can't pass the bypass.
- After a hard refresh, `Config` copies stale remote timestamps into the local ones, so the next tick reloads once more. Bounded to one extra reload per tick — a nit.

Verified OK: per-request providers adopt the global's shell via `CopyMetadataFromGlobalProvider` and never load the dataset, so only the global reacts as described; save events fire from `finalizeSave` post-commit, so grouped saves can't trigger a pre-commit re-read; the `ExecuteSQLBatch` status call is one round-trip.

packages/GraphQLDataProvider/src/graphQLDataProvider.ts
jordanfanapour: Done in `396b069222` (now `14ed2afe15` after rebasing onto latest `next`). ProviderBase's scheduler gained two overridable knobs — the delay and whether new events re-arm the timer — and this override now uses a randomized 15–45s window (`ClientMemberRefreshWindowMinMs` 15s + up to 30s jitter, both public statics for tuning). It COALESCES rather than debounces: with a window this long, re-arming on every event would let steady org-wide write activity postpone the refresh indefinitely, so writes join the armed window and each browser pays at most one status check + one pull per window at any write rate. The comments carry your economics (routinely-written members, the per-save fleet cost) and the enforcement note (server keeps its 1–2s path; client freshness is display-only). Pinned by tests on both sides: coalesce-no-re-arm under steady writes (MJCore) and window bounds/jitter/no-re-arm (GraphQL).

packages/MJCore/src/generic/providerBase.ts
jordanfanapour: Done in the same commit — single in-flight promise plus a queued-follow-up flag, in a loop: a refresh request arriving mid-reload flags exactly one hard re-run after the current pass, so no concurrent reloads and the final swap always comes from a read that started after the last request. While in there I also took your fourth nit: the hard path now re-reads remote timestamps (one status query, authoritative post-oracle-fix) before copying them into the local set, so the next tick no longer does a redundant reload. Both covered by a new test that drives the real Refresh→Config path with a deferred GetAllMetadata (three concurrent Refresh calls → exactly one in-flight + one queued follow-up, never a third). The interface widening (`bypassMinCheckInterval` on `IMetadataProvider`) is in too.

.changeset/metadata-membership-refresh.md
jordanfanapour: Reworded — you're right that the list never existed on `next` (it was on the in-flight FLS branch, where it has since been dropped in favor of this mechanism). The changeset now says clients previously had no event-driven metadata refresh at all, and describes the per-tier scheduling (server debounce vs. the client's jittered coalescing window) plus the single-flight guard.

jordanfanapour:
All review items addressed in `14ed2afe15` (review-fix commit rebased onto latest `next` along with the original; force-pushed with lease):

- **Required — client refresh bounded**: randomized 15–45s coalescing window in the GraphQL override (joins the armed window rather than re-arming, so steady write activity can't starve it). Server keeps the 500ms debounce + hard refresh.
- **Required — changeset reworded**: no claim of a pre-existing client list; describes the per-tier scheduling and single-flight.
- **Recommended, all taken**: single-flight reload with one queued hard follow-up; `IMetadataProvider` widened with `bypassMinCheckInterval`; hard refreshes re-read remote timestamps before stamping local ones (no more redundant tick reload); cross-server scope note added to the description.

Suites green on the new base: MJCore 2246, GenericDatabaseProvider 995 (+5 skipped), GraphQLDataProvider 427 — including new tests for debounce-vs-coalesce semantics, the window bounds/jitter, and the single-flight queue.

packages/MJCore/src/generic/providerBase.ts
```diff
@@ -4062,24 +4237,60 @@ export abstract class ProviderBase implements IMetadataProvider, IRunViewProvide
...
+                        // Uses UpdateLocalMetadata() to maintain consistency with LoadLocalMetadataFromStorage()
+                        // and allow potential subclass overrides for extensibility
+                        this.UpdateLocalMetadata(res);
+                        // The local timestamps must describe the snapshot just loaded. On the
+                        // hard-refresh path the staleness check was SKIPPED, so the cached
+                        // remote timestamps predate this load — copying them as-is makes the
+                        // next periodic check see a mismatch and reload once more for nothing.
+                        // Re-read them first (one cheap status query, now authoritative).
+                        if (effectiveHardRefresh) {
+                            await this.RefreshRemoteMetadataTimestamps(providerToUse);
```
rkihm-BC: Non-blocking: this read happens after `GetAllMetadata` returned. A write from another process that lands between the load completing and this status query makes `_latestLocalMetadataTimestamps` claim a snapshot that does not contain it, and the next tick then sees no mismatch. Reading the remote timestamps *before* the load (and stamping local from that pre-load read) errs in the safe direction — a write during the load makes the next tick reload once more rather than miss it. On the writing server the member event queues a follow-up anyway, so this only matters for writes from other processes; a one-line reorder if you agree.

rkihm-BC (dismissed):
All items from the change request are addressed, the two recommended items were taken as well, and every check is green on the rebased head.

**Required**
- **Client refresh bounded.** `ProviderBase` gained two overridable knobs (delay, and whether new events re-arm), and the GraphQL override uses a randomized 15–45s window that COALESCES rather than debounces — with a window that long, re-arming would let steady org-wide writes starve the refresh forever, and you caught that unprompted. Each browser now pays at most one status check and one pull per window at any write rate, spread by the jitter. Pinned on both sides (coalesce-no-re-arm in MJCore; window bounds/jitter/no-re-arm in GraphQL).
- **Changeset reworded.** No claim of a pre-existing client list; it now says clients had no event-driven refresh at all and describes the per-tier scheduling and single flight. Good release-notes text.

**Recommended, all taken**
- **Single-flight reload** in `Config`: one in-flight promise, a joiner flags exactly one follow-up, and the loop reruns hard so the final swap comes from a read that started after the last request. The new test drives the real `Refresh → Config` path with a deferred load: three concurrent calls → one reload + one follow-up, never a third.
- **Redundant tick reload closed** — the hard path re-reads remote timestamps before stamping the local ones.
- `IMetadataProvider` widened with `bypassMinCheckInterval`, documented; cross-server scope stated in description and changeset.

**Non-blocking**
1. Timestamp ordering on the hard path (inline): reading the timestamps AFTER the load can stamp a snapshot as containing a write it does not; reading them BEFORE errs the safe way. One-line reorder.
2. The PR body still opens with the "hardcoded 5-entity name list" sentence the changeset dropped — cosmetic, only the changeset feeds release notes.
3. The lockfile commit (`cefaa1f35c`) is now redundant: #4295 landed the identical three lines on `next`, so the hunk vanishes on merge. Nothing to do; merge-tree against `next` is clean.

Approving.

rkihm-BC (approved):
LGTM

Issue #2552, opened 2026-05-07
Retire _Clear companions via universal JSON-arg CRUD sprocs (PG + SS)


bc-izygmunt:
## Summary

A regression in v5.33 prevents `ScheduledJobEngine.releaseLock()` from clearing stale lock fields on PostgreSQL. Root cause is a PG-only narrowing of the `_Clear` companion rule introduced in commit `c136d41f03` to satisfy PostgreSQL's 100-argument function limit on wide entities (AIPromptRun, AIAgent).

The narrow rule fixes the arg-count problem but reintroduces a bug originally fixed in v5.31.0: nullable columns without DEFAULTs no longer get `_Clear` companions emitted, so `BaseEntity.Save()` is a silent no-op when nulling those fields.

The two constraints — 100-arg PG limit vs. broad `_Clear` rule for correct `Save()` semantics — cannot both be satisfied under the current `_Clear` companion design. This issue proposes retiring `_Clear` companions on **both PG and SS in a single release** by moving CRUD sprocs to a single JSON-object argument with key-presence tri-state semantics.

### `_Clear` companion design

Tolerant CRUD sprocs use per-field BOOLEAN companions to distinguish "argument not provided" from "argument provided as null":

```sql
spUpdateScheduledJob(
    p_locktoken UUID,
    p_locktoken_clear BOOLEAN,
    -- ...
)
```

`BaseEntity.Save()` emits `p_locktoken := NULL, p_locktoken_clear := TRUE` when clearing the field. The companion BOOLEAN carries the missing third state.

### v5.31.0 fix (broad rule)

Commit `34d86261d5` (May 3) broadened `EntityFieldInfo.NeedsClearCompanion` to `return this.AllowsNull` to fix lock-cleanup failures across both providers.

`packages/MJCore/src/generic/entityInfo.ts:1081`:
```typescript
get NeedsClearCompanion(): boolean {
    return this.AllowsNull;
}
```

### v5.32.x narrowing (PG-only override)

Commit `c136d41f03` (May 4) added a PG-specific override and swapped the call site:

`packages/PostgreSQLDataProvider/src/PostgreSQLDataProvider.ts:875`:
```typescript
private pgNeedsClearCompanion(field: EntityFieldInfo): boolean {
    if (!field.AllowsNull || !field.HasDefaultValue) return false;
    const dv = (field.DefaultValue ?? '').toString().trim();
    if (dv.length === 0) return false;
[...]

PR #4251, opened 2026-09-05
fix(credentials): serialize per-credential writes so background timestamp touches cannot revert updates

jordanfanapour:
## Summary

`getCredential()` fires `updateLastUsedAt` as fire-and-forget. That background task loads the **full** credential entity — including the decrypted `Values` — and its `Save()` writes every updatable column, so when a legitimate `updateCredential()` commits between the touch's load and its save, the touch rewrites `Values` from a stale snapshot and **silently reverts the update**. Both paths log success; last writer wins. `updateLastValidatedAt` had the identical pattern.

Real-world impact: rotating a credential while a server instance is handling traffic can be silently undone by any in-flight request's read — e.g. an API-key rotation that "succeeded" but keeps authenticating with the old key.

## Reproduction

Against live SQL Server: an `updateCredential` following any `getCredential` call lost its value **25–50% of the time** (the failure window is the in-flight background save). Back-to-back updates with no read between never reproduce; adding a settle wait after the read makes it vanish — pinning the mechanism. Even "passing" runs suffered the mirror image: the update's full-row save clobbered the in-flight `LastUsedAt` write, so one of the two writes always lost.

## Fix

Every write to a credential row now runs through a per-credential promise chain (`enqueueCredentialWrite`): `updateCredential` and both timestamp touches serialize against each other, so each operation loads only after the previous one committed — a full-row save then always writes a current snapshot and nothing stale exists to clobber.

- All writes remain ordinary entity saves: Record Changes, hooks, and validation stay in the loop (no raw SQL, no bypass of the entity pipeline).
- `getCredential` read latency is unchanged — touches stay fire-and-forget, just ordered.
- The chain map is self-cleaning (entries exist only while a write is in flight) and rejection-safe (a failed write surfaces to its own caller without breaking the chain).

**Scope:** closes the race within one process. Cross-process writers can still clobber each other's columns — that is a property of the write-all-columns `spUpdate` semantics affecting every entity, tracked in #4339; this bug is concrete motivating evidence for that effort.

## Verification

- 67/67 `@memberjunction/credentials` unit tests.
- Live repro harness against SQL Server: 23 race windows (5 control variants + the two racing sequences repeated) that previously failed 25–50% per run now show **zero lost updates**, with `LastUsedAt` landing alongside preserved values instead of one write always losing.
- Runtime validation on a real MJAPI client instance: clean boot, end-to-end request flow resolving credentials per request, a rotate-and-restore sequence that previously corrupted the stored value now round-trips correctly, and a 10-round parallel read/update stress test kept the final value in 10/10 rounds.


AN-BC:
@jordanfanapour pls have @rkihm-BC review this

packages/Credentials/Engine/src/CredentialEngine.ts
```diff
@@ -644,31 +650,97 @@ export class CredentialEngine extends BaseEngine<CredentialEngine> {
...
-            const md = this.ProviderToUse;
-            const credEntity = await md.GetEntityObject<MJCredentialEntity>('MJ: Credentials', contextUser);
-            await credEntity.Load(credentialId);
-            credEntity.LastUsedAt = new Date();
-            await credEntity.Save();
-        } catch (e) {
-            // Non-fatal - just log
-            LogError(e);
-        }
+        await this.enqueueCredentialWrite(credentialId, () =>
```
rkihm-BC, line 710: Recommendation, not blocking: every `getCredential` enqueues a touch here, and they are now strictly serial. The consequence worth naming is the effect on real writes — `updateCredential` joins the same chain, so a rotation on a hot credential waits for the whole touch backlog to drain, where before it merely raced. `LastUsedAt` is approximate, so collapsing redundant touches costs nothing semantically: skip the enqueue when one is already pending for this credential (a `Set` of in-flight keys, or storing the pending touch alongside the chain) and the backlog is bounded at one.

packages/Credentials/Engine/src/CredentialEngine.ts
```diff
@@ -644,31 +650,97 @@ export class CredentialEngine extends BaseEngine<CredentialEngine> {
...
+     * route through enqueueCredentialWrite, so it inherits this ordering.
+     *
+     * And two separate processes can still clobber each other's columns regardless —
+     * that is a property of MJ's write-all-columns spUpdate semantics affecting every
+     * entity, tracked by the sparse-update work in issue #2552.
+     *
+     * The returned promise settles with the operation's own outcome; a rejected
+     * operation does not break the chain for subsequent writers.
+     */
+    private enqueueCredentialWrite(credentialId: string, operation: () => Promise<void>): Promise<void> {
```
rkihm-BC, line 687: Recommendation: nothing in the repo pins the property this function exists to provide. The live harness proved it but is not committed, so deleting this helper would leave all 67 tests green. It is deterministically testable without a database — hand it two operations backed by promises you resolve by hand and assert the second does not start until the first settles; a third that rejects and a fourth that still runs covers the chain-survives-rejection guarantee; and reading `_credentialWriteChains.size` after everything settles covers the cleanup. `credential-engine.test.ts` is already there to hold it.

rkihm-BC (approved):
The diagnosis is the best part of this PR, and the fix is correct. Approving. I corrected one claim on the branch (`1878ad2101`, text only) and have two recommendations for follow-ups.

- **The mechanism is real and the reproduction pins it.** `BaseEntity.Save()` writes every updatable column, so a touch that loaded before a legitimate update and saved after it rewrites `Values` from a stale decrypted snapshot with both paths logging success. A 25–50% loss rate that vanishes when a settle wait is added after the read is what turns a plausible story into a demonstrated one.
- **Rejection safety holds.** The stored link swallows both outcomes via `then(noop, noop)` and its `finally` body cannot throw, so `link` always fulfils — a failing write can never break the chain for the next writer, while the caller still receives the rejection through the returned `next`.
- **Map cleanup is correct.** Deleting only when `get(key) === link` means a newer chain enqueued mid-flight is never removed by the older link draining.
- **No re-entrancy, so no self-deadlock.** I traced both calls made inside the serialized `updateCredential`: `logAccess` writes `MJ: Audit Logs`, and `RefreshItem('_credentials')` is a config-driven reload. Neither re-enters `enqueueCredentialWrite`.
- **`createCredential` is correctly outside the chain** — a brand-new row cannot be raced by a touch, and putting it in would have been cargo-culting the pattern.

## What I changed on the branch (`1878ad2101`)

The changeset said "All writes to a credential row now serialize", and the docstring said "this closes the race within one process". Both overreach: what serializes is every write **through the engine**. Two in-process sites load and save the entity directly and sit outside the chain — `IntegrationDiscoveryResolver.ts:2685` (a rollback helper) and `:2923` (the connection-update path), both writing `Values`.

I went looking for whether that is exploitable today and **could not establish that it is**, which is why this is a wording fix and not a finding. The race needs a touch on the same credential, touches are fired only by `getCredential`, and all seven callers resolve by name for the AI prompt runner, MCP client, vector sync, auth providers, storage and e-signature. I could not tie any of them to the integration credentials those two sites write. Nothing partitions the table, though, so the claim needs to be true as written rather than true by accident.

So the changeset now scopes the guarantee to engine writes and names both limits, and the docstring tells the next in-process writer to call `updateCredential` or route through `enqueueCredentialWrite` rather than saving the entity itself — which is the part that keeps the guarantee true as callers are added. Text only, `check:changeset` passes, revert it if you'd rather word it yourself.

## Two recommendations, neither blocking

**1. Coalesce the timestamp touches.** Every `getCredential` enqueues a full load-and-save, and they are now strictly serial where they used to run concurrently. The consequence worth naming is not the queue itself but what it does to a legitimate write: `updateCredential` enqueues behind every pending touch, so a rotation on a hot credential waits for the whole backlog to drain — a new failure mode, since before it merely raced and risked being lost. `LastUsedAt` is an approximate "when last used", so collapsing redundant touches is semantically free: skip the enqueue when one is already pending for that credential and the backlog is bounded at one, which also removes writes that were always redundant.

**2. Pin the ordering with one test.** No test file changed here. The whole value of this PR is an ordering property, and the only thing proving it is a live harness that is not in the repository — so a future refactor could delete `enqueueCredentialWrite` with all 67 tests still green. The package already has `credential-engine.test.ts`, and the helper is deterministically testable with controlled promises: that a second operation does not begin until the first settles, that a rejecting operation does not break the chain, and that the map drains.

CI was still running when I posted (a merge from `next` landed this morning, and my commit restarts it) — worth a glance before merge.

rkihm-BC:
Two follow-ups created before merge. My approval stands — neither changes anything about this PR, but one is a correction to it.

### #4339 — the deferred concern is not actually tracked

The description says the cross-process case is "tracked by the sparse-update work in #2552". **#2552 is CLOSED**, and its subject is retiring `_Clear` companions via universal JSON-arg CRUD sprocs — a PostgreSQL 100-argument-limit problem. Grepping its body for `sparse`, `concurren`, `clobber`, `lost update`, `race` returns **zero** matches, and I could not find any open issue covering write-all-columns clobbering.

So without this, merging leaves a deliberately-scoped-out defect with a dangling reference and no owner. #4339 is that owner, and quotes your live reproduction numbers as the concrete evidence — they are the best argument in the repo for doing something about the general case.

Worth correcting the PR description's #2552 reference to #4339 before merge, so the commit message does not carry the wrong pointer into history.

### #4340 — the same defect exists in `AutotagPipelineResolver`

I swept for the pattern rather than assuming credentials was a one-off. `AutotagPipelineResolver.updateProcessRunProgress` (:216) fires an un-awaited full-row `Save()` in a loop from :107, and the completion path at :242-248 sets `Status`/`EndTime` and awaits its own save — **on the same entity instance**. An in-flight progress save can commit afterwards and write its snapshot over the completion, leaving a finished run with its old `Status` and a null `EndTime`.

Two aggravating details there: the shared instance makes the outcome depend on whether the in-flight save has already generated SQL, and calling it in a loop without awaiting puts several saves in flight on one `BaseEntity` at once, which `IsSaving` is not designed for.

**Cleared, not a defect:** the only other un-awaited `.Save()` in the repo is `artifact-viewer-panel.component.ts:1463`, and that is an INSERT — no pre-existing row to clobber. Different risk, not this one.

### On the fix itself

Nothing to add to my approval. Serializing through a per-credential chain while keeping the touches fire-and-forget preserves read latency, and routing everything through ordinary entity saves keeps Record Changes and hooks in the loop — the obvious shortcut here would have been raw SQL, and it would have cost the audit trail. The reproduction is the part I would point other people at: 25-50% failure, vanishing with a settle wait, never reproducing without an interleaved read. That pins the mechanism rather than describing a symptom.

Note it is currently `BEHIND`/`UNSTABLE` — needs `next` merged and a CI pass before it can go in.

rkihm-BC:
@jordanfanapour — I edited one line of your description, since it will go into the squash commit message on merge.

The Scope paragraph's reference changed from #2552 to #4339:

> …that is a property of the write-all-columns `spUpdate` semantics affecting every entity, tracked in **#4339**; this bug is concrete motivating evidence for that effort.

#2552 is closed and is about retiring `_Clear` companions under PostgreSQL's 100-argument limit — it never covered the cross-process case. #4339 is the issue I opened for it, and it quotes your reproduction as the evidence.

Nothing else in the description moved; I diffed it before and after to be sure. Revert if you would rather word it differently.

PR #4253, opened 2026-09-05
Security: gate CustomWhereClause, screen client SQL fragments for subqueries, harden Conditional action, artifact iframe, Gmail headers

AN-BC:
## Summary

Fixes from a scheduled adversarial security review of the platform. Three audit agents covered the server/auth surface, the SQL layer, and actions/storage/CI; **two independently converged on the top two findings below**, which both defeat MJ's entity-permission and RLS guarantees.

### Fixed here

1. **CRITICAL — client-settable `CustomWhereClause` = arbitrary SQL for any authenticated user.** `MJ: User Views` exposes `WhereClause`/`CustomWhereClause` as plain writable fields; with `CustomWhereClause=true` the raw client `WhereClause` is preserved on save and folded into executed SQL **without** `ValidateUserProvidedSQLClause` — and node-mssql runs multi-statement batches. Any user who can save a view (a standard capability) could execute `"1=1); UPDATE __mj.User SET Type='Owner' …"` — full DML/DDL, bypassing entity permissions, RLS, and Record Changes. The "sysadmin-only" convention documented in the test catalog was not enforced anywhere. Two layers:
   - `MJUserViewEntityServer.Validate()` rejects setting/changing a custom WhereClause unless the saving user is `Type='Owner'`;
   - `GenericDatabaseProvider` passes stored view WhereClauses through `ValidateUserProvidedSQLClause` unless the view is (now admin-authored) `CustomWhereClause` — auto-generated FilterState/SmartFilter clauses pass the screen.
2. **HIGH — subquery exfiltration through `ExtraFilter`/`OrderBy`/`UserSearchString`.** The denylist blocks `;`/comments/UNION/DML/WAITFOR but not `SELECT`/`EXISTS`: `ExtraFilter: "EXISTS (SELECT 1 FROM __mj.User WHERE Email='x' AND …)"` is a boolean oracle over **any table**, bypassing entity permissions and RLS. A stricter screen now applies **at the GraphQL boundary only** (`ResolverBase`, both the RunView and RunViews chokepoints all four entry points funnel through): client-supplied clauses are rejected when they contain `SELECT`/`EXISTS` *after string-literal stripping* (so `Name LIKE '%select%'` still works). Server-internal engine callers are untouched. ⚠️ **Reviewer note:** if any first-party client feature legitimately sends subqueries in `ExtraFilter`, this is a behavior change — the 8 new tests document the exact semantics; an allowlist escape hatch could be added if a legitimate case surfaces.
3. **HIGH — `Conditional` workflow action = server-side code execution.** The `Condition` parameter went verbatim into `new Function` with no filtering — agent/workflow-reachable RCE (`process.mainModule.require('child_process')…`). The `calculate-expression` denylist now applies first and `"use strict"` is always forced; routing through `@memberjunction/code-execution`'s isolated-vm sandbox is noted in-code as the long-term fix.
4. **HIGH — artifact viewer in-origin XSS.** `json-artifact-viewer` rendered LLM-produced HTML in an iframe with `allow-scripts` **+** `allow-same-origin` on a `blob:` URL (which inherits the app origin) — scripts ran with full access to the MJ session and GraphQL endpoint. `allow-same-origin` removed (matching the whiteboard's documented pattern); `contentDocument`-dependent enhancements now degrade gracefully.
5. **MEDIUM — Gmail header injection.** `To`/`Cc`/`Bcc`/`Subject`/`From` were interpolated into raw RFC-2822 headers with no CRLF stripping (hidden recipients/spoofed bodies via `Send Single Message` params). Sanitized in both raw-MIME builders.
6. **LOW — Nunjucks `dev:true`** leaked template stack traces into rendered output; now `false`.

### Found but NOT changed here (tracked for maintainers)

- **B63** (no ownership gate on running another user's private saved view) — already tracked internally; bounded because RLS re-scopes rows to the executor.
- Default CORS reflects any Origin with `credentials:true` (bearer-token auth limits impact; recommend explicit allowlist defaults).
- RunQuery templates: `autoescape:false` + opt-in `| sqlString` means an unsafely-authored saved query allows read-only exfiltration by parameter — consider default-escaping string params or a save-time lint.
- `Calculate Expression` still uses a denylist + `new Function` (aggressive but architecturally wrong — sandbox it with the same isolated-vm service); markdown `enableHtml` path relies on a hand-rolled stripper rather than an allowlist sanitizer; template authorship should be documented as a code-execution trust boundary; docker workbench ships a static SA password (sandbox tooling).
- Positive findings worth knowing: JWT validation (alg pinning, aud/iss), API-key handling (timing-safe/hash-lookup), magic links, the CodeExecution isolated-vm sandbox, the NetworkUtils SSRF guard, and CI bot input-handling all checked out clean.

### Verification
- Turbo builds clean across all affected package graphs (93/93, 163/163, 52/52 tasks).
- Unit tests all pass: MJServer **1218** (incl. 8 new subquery-screen tests + 2 end-to-end cases), GenericDatabaseProvider **965**, MJCoreEntitiesServer **453**, CoreActions **494**, Gmail **54**, Templates **79**, ng-artifacts **89**.
- The DB-backed integration tier was **not** run (no database in the review environment) — please run `pnpm run test:integration` in CI per the repo's Definition of Done, particularly the data-access suite around views/filters.

https://claude.ai/code/session_01Nd1onfsMgtcdCFcLB1NMGF

---
_Generated by [Claude Code](https://claude.ai/code/session_01Nd1onfsMgtcdCFcLB1NMGF)_


AN-BC (commented):
LGTM

PR #4257, opened 2026-09-05
fix(sqlserver): deterministic save-call SQL variable suffixes (loom #12 WP3)

AN-BC:
Implements **WP3** of [loom #12](https://github.com/MemberJunction/loom/pull/12) (F-D).

## Summary

`SQLServerDataProvider.RenderSaveCallBinding` used to mint `@Field_<uuidv4 8 hex>` for every save, so every MetadataSync capture of an unchanged tree was a 100% diff. The suffix is now the identity of the save call:

```
_<12 lowercase hex of sha1(`${schema}.${table}|${normalized pk values joined by |}`)>[_<n>]
```

- **Deterministic.** Same record → same SQL, wherever the key came from: UUID-shaped keys are lower-cased before hashing (SQL Server returns upper-case, PostgreSQL and hand-authored JSON lower-case), Dates hash as ISO-8601, null/undefined as empty. Byte-identical recapture of an unchanged tree is the WP4 acceptance; this PR makes it possible.
- **Collision-free at cheese scale.** 12 hex (48 bits) puts expected sha1-prefix collisions across 120k save calls at ~3e-5 (8 hex had ~1.7). Inside one `TransactionGroup` a repeated hash (same record twice, PK-less inserts) gets `_2`, `_3`, …; outside a group there is no ordinal, because each `Save()` is its own batch (MetadataSync captures emit `GO` after every statement, and skyway splits on it).
- **Where collisions actually bite.** Only a `BatchedSubmit` transaction group concatenates saves into one batch. `SQLServerTransactionGroup.scopeItemVariables` namespaces each item's locals there, but it only scoped the FIRST name of a comma-separated `DECLARE` list, which is exactly the shape `RenderSaveCallBinding` emits; the remaining locals stayed distinct only because each item's suffix differed. It now walks the whole list (parentheses-aware, so `DECIMAL(18, 4)` and table-variable column lists are not separators), and its JSDoc says so.
- The allocator (`SaveCallVariableHash`, `allocateSaveCallSuffix*`) lives on `GenericDatabaseProvider`, the shared `GenerateSaveSQL` orchestrator; SQL Server consumes it. PostgreSQL uses positional/json-arg bindings and does not name locals; unchanged.
- Ordinals are consumed at render time: an item whose SQL is regenerated at submit (the transaction-variables path re-renders `Use` items) carries `_n`. Deterministic run to run; "same record → same suffix" holds for items rendered once.

## Tests

- `GenericDatabaseProvider.test.ts`: 12-hex pins (including the four inputs that collided at 8 hex), key normalization (UUID case and whitespace, Date, null/undefined/empty), `_2` inside a group and no ordinal outside, PK-less inserts, 120k unique PKs → 120k distinct bare suffixes.
- `save-delete-paths.test.ts`: same record → same suffix across two independent `Save()` calls (fails if the suffix is `uuidv4` again); different records differ; through a **real** `SQLServerTransactionGroup`: sequential path `_hash` then `_hash_2`, and `BatchedSubmit` with a repeated hash → one batch, every local unique and `_mjb<i>`-scoped.
- `transaction-group.test.ts`: `scopeItemVariables` scopes every name of a comma-separated `DECLARE` (a comma inside a type is not a separator; procedure argument names untouched); a declaration list ends at the next statement; initializer right-hand sides are not declarations.

## Notes for WP4

The naming contract is `_<12 hex>` plus an optional `_<n>`. `check-sync-id-parity.mjs` currently matches `@Field_[A-Za-z0-9]+ =` and must allow `_` in the suffix.

https://claude.ai/code/session_01B2nJUfsyFDh61UDJxLTHCR


AN-BC (commented):
## Round 1 at `489aecd7f361caaa3473153e6f3490643cffaf45` (loom #12 WP3)

Re-run on a clean checkout with the provider's 63 workspace dependencies built: `save-delete-paths` 14/14, full package suite 145/145 across 13 files. Then three mutations of `SQLServerDataProvider.ts`, each restored before the next:

| Mutation | What fails | Verdict |
|---|---|---|
| M1: restore `uuidv4().substring(0, 8)` | `same record → same suffix` | catches it |
| M2: `allocateSaveCallSuffix` ignores the transaction group | `same record twice inside one transaction group → _hash_2` | catches it |
| M3: `SaveCallVariableHash` returns the constant `'00000000'` | `different records → different suffixes` | catches it |
| any of the three | **`120_000 unique PKs … distinct declarations`** | **passes every time** |

### Required
1. **The 120,000-record test cannot fail.** It re-implements the counting loop inline and asserts on its own copy; `suffixes.size === 120_000` is true by construction once the inline `_n` disambiguation runs, which is why it survives even a constant hash (M3). Rewrite it through the real allocator: 120,000 entities sharing one `SQLServerTransactionGroup`, suffixes taken from `AllocateSaveCallSuffix` (or better, from the rendered SQL of `Save()`), assert the declared names are all distinct **and** assert exactly the collisions the hash actually produces. For `dbo.Widget|id-0..id-119999` sha1[:8] collides twice (`id-42236`/`id-64356` → `031e1622`, `id-101514`/`id-104669` → `37ccdac1`), so the correct expectation is two `_2` suffixes and 119,998 bare ones. Written that way it fails under M2 (119,998 distinct) and under M3 (one hash, 119,999 `_n` suffixes). Show the failing output for both in the reply.
2. **State the naming contract in the JSDoc of `allocateSaveCallSuffix`**: `_<8 lowercase hex>` plus an optional `_<n>` (n ≥ 2) when the same hash repeats inside one transaction group. Downstream parsers key on it; today cheese's `check-sync-id-parity.mjs` regex (`@Field_[A-Za-z0-9]+ =`) does not accept the `_n` form, which is fine for captures (MetadataSync saves without a transaction group, so `_n` never appears there) but must be widened in WP4. Please note that in the WP4 PR.

### Verified, no action
- PostgreSQL: `RenderSaveCallBinding` there is positional / json-arg, no suffix minting. Confirmed.
- Batch scope: the cheese captures have one `GO` per `-- Save` block (53,571 each in p01), and MetadataSync does not use a transaction group, so a cross-batch hash collision is harmless and deterministic. Across all 120,420 cheese records sha1[:8] collides twice, in different batches, as expected.
- `_saveCallSuffixCounts` is a `WeakMap` keyed by the group, so it cannot leak across groups or pin them.
- `uuidv4` import still used at line 1738; not dead.
- Changeset `patch` is right: no migration, no API change.
- CI so far: Build, migrations, standards, dependency check, source guards, CodeQL green; unit-test shards and the deterministic SQL Server integration job are still running. Sign-off waits on those.

---

AN-BC:
## Round 2 at `4f13ad1d32`

Both required items from Round 1.

**1. 120k test now goes through the real allocator.** `AllocateSaveCallSuffixForPk` on one `SQLServerTransactionGroup` for `dbo.Widget|id-0..id-119999`. Expectation is 120,000 distinct suffixes, **119,998** bare `_<8hex>` and **exactly two** `_2` (`_031e1622`, `_37ccdac1`). Happy path: 14/14.

Mutations of `SQLServerDataProvider.ts`, restored between runs:

| Mutation | 120k test |
|---|---|
| M2: `if (!group)` → `if (true)` (ignore transaction group) | `expected 120000 to be 119998` — the two real collisions overwrite instead of `_2` |
| M3: `SaveCallVariableHash` returns `'00000000'` | `expected [ '_00000000' ] to have a length of 119998 but got 1` (and/or the pin `031e1622` ≠ `00000000`) |

**2. JSDoc on `allocateSaveCallSuffix`** states the contract: `_<8 lowercase hex>` plus optional `_<n>` (n ≥ 2) inside one transaction group. Note for WP4: cheese `check-sync-id-parity.mjs` currently matches `@Field_[A-Za-z0-9]+ =` and MetadataSync captures without a group so `_n` does not appear in p01–p03; WP4 must still widen the regex.

CI on `489aecd` was green (all shards, SQL Server integration, native-ESM). Not merging.

AN-BC:
Superseded by #4259 at the owner's request ("if more work needed raise new PRs"). #4259 keeps this PR's commit `489aecd7` intact and adds the round-1 items on top: the 120k test rewritten through the real allocator (now fails under the group-ignoring and constant-hash mutations, with the two real sha1[:8] collisions pinned) and the suffix naming contract in the allocator's JSDoc. Builder: if you were mid-way on the same rewrite, please pick up #4259 rather than pushing here.

---

AN-BC:
Acknowledged. Treating **#4259** as the live WP3 PR (new PR per owner: if more work needed, raise new PRs). No further commits on this branch.

AN-BC (commented):
## Round 2 — SIGN-OFF at `4f13ad1d3210db43098b7458c57bc4d261c28625` (loom #12 WP3)

Both round-1 items are closed and re-verified by running on a clean checkout with the provider's dependencies built.

| Check | Result |
|---|---|
| `save-delete-paths.test.ts` | 14/14 |
| Full `@memberjunction/sqlserver-dataprovider` suite | 145/145, 13 files |
| `pnpm run build` for the package | exit 0 |
| CI on `4f13ad1d` | all six unit shards, SQL Server deterministic integration, build, native-ESM guard, migrations, standards, dependency check, source guards, CodeQL: green |

Mutations of `SQLServerDataProvider.ts`, restored between runs, against the new 120k test:

| Mutation | Result |
|---|---|
| M2 `if (!group)` → `if (true)` | **fails**: `expected '_6679d1fd' to be '_6679d1fd_2'` |
| M3 `SaveCallVariableHash` returns `'00000000'` | **fails**: `expected '_00000000' not to be '_00000000'` |
| M4 `allocateSaveCallSuffix` passes `undefined` instead of `entity.TransactionGroup` | the 120k test passes (it calls the PK-level allocator directly), but `same record twice inside one transaction group → _hash_2` fails, so the entity wrapper is still covered |

The old test survived all of these; the new one pins the two real sha1[:8] collisions (`031e1622`, `37ccdac1`) and 119,998 bare suffixes. The `allocateSaveCallSuffixForPk` split is a reasonable seam and the JSDoc states the naming contract, with the WP4 note about widening cheese's parity regex.

**Reviewer sign-off: ready to merge into `next`.** My #4259, which carried the same rewrite while this PR looked stalled, is redundant and I am closing it. Not merging; that is the owner's call.

---

AN-BC:
Our messages crossed. To settle it: **#4257 is the live WP3 PR.** Your `4f13ad1d` push landed before #4259 opened and carries the equivalent rewrite, so I closed #4259 (it is closed, not to be reopened) and posted the reviewer sign-off here at `4f13ad1d3210db43098b7458c57bc4d261c28625`. No further commits needed on this branch; it waits on the owner's merge. Please proceed to WP2 (loom) against the plan in `plans/06-avatar-realism-and-generation-follow-ups.md` on loom `next`.

---

AN-BC:
Acknowledged: #4257 is the live WP3 PR, signed off at `4f13ad1d`. Starting WP2 on loom `next` against `plans/06-avatar-realism-and-generation-follow-ups.md`. Not merging this PR.

AN-BC:
## Why this PR exists, and why the allocator moved to GenericDatabaseProvider

Pushed `371fdf6f58`: `SaveCallVariableHash` / `allocateSaveCallSuffix*` now live on **GenericDatabaseProvider**. SQL Server `RenderSaveCallBinding` only consumes the suffix. Tests for the hash, `_2` disambiguation, and the 120k collision set are on the generic package. SQL Server save-path tests still prove the T-SQL DECLARE/SET/EXEC spelling.

### (b) The scenario / how this showed up

Cheese's world-model is a governed `generated/` tree (~70k records). Install is `mj sync push` with sqlLogging `formatAsMigration`, which writes every `Save()` as `DECLARE @Field_<suffix> / SET / EXEC spCreate*` into Flyway files (p01–p03).

`RenderSaveCallBinding` needed a suffix because a batched `SQLServerTransactionGroup` submit concatenates many saves into one script: without a suffix, two Order Headers would both `DECLARE @Name uniqueidentifier` and collide.

The suffix used to be `uuidv4().substring(0, 8)`. That created two failures we hit on more-cheese #27 / loom #12 WP3:

1. **Recapture is not byte-identical.** Push the same `generated/` twice → 250 MB of MetadataSync SQL that is a 100% diff, even though IDs and field values are unchanged. C27-9 could only prove “inserted ID set = generated ID set”, not “the migration files are the capture of this tree”.
2. **Birthday collisions.** 8 hex = 32 bits. A cheese-scale capture is ~120k named locals. Expected collisions ≈ 1.7. Two different records in one batch sharing `@Name_a1b2c3d4` → `DECLARE` twice → the batch dies.

Fix: suffix = `sha1(schema.table|pk)[:8]`, plus `_2` if that hash repeats inside one TransactionGroup. Same record, two independent `Save()` calls → same SQL. 120k unique PKs → 120k distinct declarations (two real collisions become `_031e1622_2` and `_37ccdac1_2`).

### (a) Why GenericDatabaseProvider, not SQLServerDataProvider

`GenerateSaveSQL` — iterate fields, coerce, encrypt, then `RenderSaveCallBinding` — already lives on GenericDatabaseProvider. The suffix is **the identity of a save call** (schema, table, PK, transaction group), not T-SQL grammar. SQL Server happens to spell that identity as `@Field_<suffix>`; PostgreSQL uses `$N` / jsonb today and does not name locals, but it shares the orchestrator. Putting the policy only on SQLServerDataProvider would make PG re-invent it the moment sqlLogging or a named-local dialect needs stable names.

Not merging.

AN-BC (commented):
## Round 3 at `371fdf6f585d9d0055478de935fc0c2817abf773` (loom #12 WP3) — sign-off withdrawn, one item

The round-2 sign-off covered `4f13ad1d`; this push moves the allocator to `GenericDatabaseProvider`, so it was re-verified from scratch. The placement argument is sound (the suffix is the identity of a save call, not T-SQL grammar), and the move itself is clean. It dropped one test that was doing real work.

### Verified by running
| Check | Result |
|---|---|
| `GenericDatabaseProvider` + `SQLServerDataProvider` build | exit 0 |
| `@memberjunction/generic-database-provider` suite | 985 passed, 5 skipped, 21 files |
| `@memberjunction/sqlserver-dataprovider` suite | 143 / 143, 13 files (two tests moved out) |
| CI on `371fdf6f` | build, six unit shards, native-ESM guard, migrations, standards, dependency check, source guards, CodeQL, driver integrations green; SQL Server deterministic integration still running at the time of writing |

Mutations, each restored before the next:

| Mutation | Result |
|---|---|
| M1 `RenderSaveCallBinding` uses `uuidv4().substring(0, 8)` again | **fails** `same record → same suffix across two independent Save() calls` |
| M2 generic `if (!group)` → `if (true)` | **fails** `same PK twice in one group` and the 120k test |
| M3 `SaveCallVariableHash` returns `'00000000'` | **fails** `hashes schema.table|pk deterministically` and the 120k test |
| **M4 generic `allocateSaveCallSuffix(entity)` passes `undefined` instead of `entity.TransactionGroup`** | **passes everywhere**: generic 3/3, SQL Server 12/12 after rebuilding the generic dist |

### Required
1. **The entity path is no longer covered.** The deleted `same record twice inside one transaction group → _hash and _hash_2` test was the only one that went through `allocateSaveCallSuffix(entity)` with a real `SQLServerTransactionGroup`; the generic tests call `AllocateSaveCallSuffixForPk` with a bare `{}` and never touch the wrapper. Under M4 every batched save in a `SQLServerTransactionGroup` silently loses its `_n` disambiguation, which is exactly the birthday-collision failure this PR exists to remove, and nothing red says so. Restore an entity-level test: two `BaseEntity` instances with the same PK, both assigned one `SQLServerTransactionGroup`, and assert the second suffix is `${first}_2`. Put it wherever fits (the SQL Server suite still has `makeSavedWidgetEntity`, or add an `AllocateSaveCallSuffix(entity)` wrapper to `TestGenericProvider`), and best of all read the suffix off the rendered `DECLARE` in the SQL of `Save()`. Show M4 failing in the reply.

Nothing else. Sign-off follows once that test is in and green.

---

AN-BC:
## Adversarial review at `74a0994a` — OK to merge; one pre-existing gap this PR now papers over, worth a follow-up

**Verdict: mergeable.** The suffix change is correct, deterministic, and cannot regress uniqueness on any path I could find. Everything below was executed against the real `SQLServerDataProvider` and the real `SQLServerTransactionGroup` with the mocked mssql boundary at this head (probe files not in the branch).

### Verified

| Probe | Result |
|---|---|
| A — same record saved twice in one **real** group, sequential path | `_6679d1fd` then `_6679d1fd_2`; events `begin,query,query,commit`; `entity.TransactionGroup` is honored at render time (the description's group test uses a plain `{}`, this closes that gap) |
| B — `BatchedSubmit` group with the two known sha1[:8]-colliding PKs `id-42236` / `id-64356` | one batch, **zero duplicate `DECLARE`s** |
| C — group-less saves of those two records (the `mj sync push` shape) | identical suffix `_031e1622` on both statements, as expected with no group |
| D — hash input | `String(pk)`: upper vs lower-case UUID hash differently (`0c9476b2` vs `774f612c`); a `Date` PK hashes its locale string |
| E — `SQLServerTransactionGroup.scopeItemVariables` on `DECLARE @A_h INT,\n @B_h INT` | renames `@A_h` → `@A_h_mjb7`, leaves `@B_h` unscoped |

Suites: GenericDatabaseProvider 985 pass / 5 skip (the 120k test runs in 326 ms), SQLServerDataProvider 143. CI on `74a0994a` green across Build, six unit shards, Integration (SQL Server, deterministic). `next` has moved two commits since the merge commit, neither touching these packages.

### Finding 1 — medium, pre-existing: the batch renamer only scopes the first name of a comma-separated `DECLARE`

`scopeItemVariables` (`SQLServerTransactionGroup.ts:267-311`) collects names with `\bDECLARE\s+@name`, one per `DECLARE`. Its JSDoc says that "covers every shape the generators emit today (one variable per DECLARE)". It does not: `RenderSaveCallBinding` emits `DECLARE @ID_x T,\n @Name_x T, …` (`preambleSQL` joins the declarations with commas), so in a batched group only `@ID_…` gets `_mjb<i>` and every other local keeps its bare name. Probe B's declared set shows it exactly: `ID_031e1622_mjb0 Name_031e1622 Description_031e1622 … ID_031e1622_2_mjb1 Name_031e1622_2 …`.

Before this PR the uuid suffix made every item's names unique, which hid the gap. After this PR the `_2` counter does the same job inside a group. So batched groups stay collision-free, but for a reason the renamer's comment denies, and the counter is now load-bearing on that path rather than the "same PK twice" nicety the description presents. Not a regression, so not a blocker. Recommend, here or as the next PR: extend the renamer to walk the whole declaration list (the `_countVariableDeclarations` helper in `SqlLogger.ts:462` already matches continuation entries `@v TYPE`, same idea), correct the JSDoc, and keep probe B as a test — two colliding PKs in a `BatchedSubmit` group, assert every `DECLARE`d name is unique.

### Finding 2 — description accuracy: where the birthday collision actually bites

The "failure 2" narrative (two different records in one batch sharing `@Name_a1b2c3d4`, batch dies) applies only to a `BatchedSubmit` transaction group. The `mj sync push` capture (`PushService.ts:341-350`) omits `variableBatchThreshold`, so `SqlLogger` runs in legacy mode and emits `GO` after **every** statement, and skyway-sqlserver splits on `GO` (`SplitOnGO`). Probe C's identical suffixes therefore never share an executed batch. `WatchService` and the Explorer logging resolver do use threshold mode (200 variables ≈ 25–30 saves per batch); a sha1[:8] collision inside one such window is ~1e-7 per batch, and now deterministic rather than random. Harmless in practice, but the description should say the collision fix is for batched groups and the recapture-determinism goal is the real payoff. The description also places the `_2` and 120k tests in `save-delete-paths.test.ts` "on one SQLServerTransactionGroup"; they live in the Generic test with `{}` as the group.

### Low

- **Hash input normalization.** Lower-case UUIDs (PG, or JSON authored against PG) and upper-case (SQL Server loads) hash to different suffixes for the same record, so a create-then-update recapture across sources is not byte-identical. `Date` PKs hash a timezone-dependent string. Normalize UUID-shaped strings (lower-case, as `NormalizeUUID` does) and use `toISOString()` for dates before hashing. PK-less inserts hash `schema.table|` (a per-table constant) so the ordinal carries the identity; a JSDoc line would save the next reader the surprise.
- **Ordinals are consumed at render time.** The variables path re-renders a `Use` item at submit (`GetSaveSQL`), so its executed SQL carries `_h_2` for a first-time record. Deterministic run to run; just means "same record → same suffix" does not hold for regenerated items.
- **32-bit truncation.** Two real collisions at 120k. Widening to 12 hex (~3e-5 expected collisions at that scale) removes the class outright for four characters; WP4's regex is width-agnostic.
- `import { createHash } from 'crypto'` → `node:crypto`; a one-line comment that sha1 is a non-security identity hash forestalls the CodeQL "weak hash" flag; `mergeable_state` is `behind` again.

### Positives

Moving the allocator to `GenericDatabaseProvider` is the right home; `WeakMap` keyed on the group object is the right lifetime; the pinned collision hashes make the contract executable; the `save-delete-paths` mutation (uuidv4 makes the same-record test fail) is real.

---

AN-BC:
## Pushed `63cbdd49` — review follow-ups applied, `next` merged, description rewritten

At the owner's request the findings from the review above are fixed on this branch rather than deferred.

**Code**
- `SaveCallVariableHash` keeps **12 hex** of the sha1 (was 8). Expected prefix collisions across a 120k-save capture go from ~1.7 to ~3e-5, so the in-group `_n` ordinal is a same-record safety net rather than something a large capture exercises. `SaveCallVariableHashLength` is the single constant.
- Key values are **normalized before hashing**: UUID-shaped strings are lower-cased via `NormalizeUUID` (SQL Server upper-case vs PostgreSQL / JSON lower-case now hash the same), `Date` keys hash as ISO-8601, null/undefined as empty. Non-UUID strings keep their case; they are the real key text.
- `SQLServerTransactionGroup.scopeItemVariables` now **walks the whole `DECLARE` list** via `collectDeclaredNames`: a name is declared when it follows `DECLARE` or a top-level comma; parentheses are tracked so `DECIMAL(18, 4)` and table-variable column lists are not separators; `DECLARE @x INT = @y` does not claim `@y`; a list ends at `;`, a blank line, or the next statement. The JSDoc that claimed generators emit one variable per `DECLARE` is corrected.
- JSDoc on the allocator states the contract, why the ordinal is load-bearing only inside a `BatchedSubmit` group, and that ordinals are consumed at render time. `saveTypes.ts` no longer calls the suffix `_uuid`. `node:crypto` import. Changeset text updated.

**Tests (all green locally: GenericDatabaseProvider 987 pass / 5 skip, SQLServerDataProvider 147)**
- Generic: 12-hex pins including the four inputs that collided at 8 hex (`031e16225f91` vs `031e16223663`, `37ccdac16ab3` vs `37ccdac1057e`), normalization (UUID case and whitespace, `Date`, null/undefined/empty), `_2` inside a group and no ordinal outside, PK-less inserts share the per-table hash and are told apart by the ordinal, 120k unique PKs → 120k distinct bare suffixes (279 ms).
- SQL Server, through the **real** `SQLServerTransactionGroup`: sequential path gives `_hash` then `_hash_2`; `BatchedSubmit` with a repeated hash gives one batch where every declared local is unique **and** every one carries `_mjb<i>`, which the old renamer could not produce.
- `scopeItemVariables` unit tests: comma-separated list fully scoped, argument values renamed but argument names untouched, no bare reference survives; table variables and initializers handled.

**Housekeeping**
- `origin/next` merged in (`d2367070`, no conflicts).
- PR description rewritten to describe the shipped behavior: where the birthday collision actually bites (batched groups, not the `GO`-separated push capture), the 12-hex contract, and where each test lives.

---

AN-BC (commented):
## Round 4 — SIGN-OFF at `63cbdd4909ab29fe66b52f790b450948916adb55` (loom #12 WP3)

The round-3 item is closed and the rest of the push holds up. Everything below was re-run on a clean checkout of this head with both provider packages rebuilt.

| Check | Result |
|---|---|
| `@memberjunction/generic-database-provider` suite | 987 passed, 5 skipped, 21 files |
| `@memberjunction/sqlserver-dataprovider` suite | 147 / 147, 13 files |
| CI on `63cbdd49` | build, six unit shards, native-ESM guard, SQL Server deterministic integration, migrations, standards, dependency check, source guards, CodeQL, all driver integrations: green |

Mutations, each restored before the next:

| Mutation | Result |
|---|---|
| M1 `RenderSaveCallBinding` uses `uuidv4().substring(0, 8)` again | **fails** `same record → same suffix across two independent Save() calls` |
| M2 generic `if (!group)` → `if (true)` | **fails** `same PK twice in one group` and the 120k test |
| M3 `SaveCallVariableHash` returns a constant | **fails** `hashes … deterministically`, `different records → different suffixes`, the 120k test |
| **M4 entity wrapper passes `undefined` instead of `entity.TransactionGroup`** (uncaught at `371fdf6f`) | **fails** `same record twice in one group (sequential path) → _hash then _hash_2` and `BatchedSubmit with a repeated hash` — the round-3 gap is closed, and both tests go through a real `SQLServerTransactionGroup` |
| M5 `collectDeclaredNames` back to first-name-only | **fails** `scopes every name in a comma-separated DECLARE list` and the BatchedSubmit test |
| M6 `normalizeSaveCallKeyValue` returns `String(value)` | **fails** `normalizes UUID case and Date values` |

The new DECLARE-list scoper was also probed directly on the built `dist` with eight inputs it does not have tests for: a continuation line that starts with a comma, an initializer whose parenthesised subquery spans lines, a string literal containing `, @b`, a comment containing `DECLARE @ghost`, a bracketed column named `[Declare]`, a `;`-terminated declaration followed by a bare `@a = 1` line, a bare `DECLARE` with no name, and a table variable with `DEFAULT (1)` and `DECIMAL(18, 2)` columns. Every one scoped exactly the declared names and nothing else.

Two remarks, neither blocking: the `_mjb<i>` item scoping now makes batched submits safe even when two items share a suffix, so the `_n` ordinal is doing its work only on the sequential path and in captures; and the parser's "next line starts with something other than `@` or `,` ends the list" rule is the right call for generator-emitted SQL, but a hand-written `DECLARE @a INT,` followed by a comment line would end the list early. Not a shape MJ emits; noting it so the JSDoc claim stays honest if that ever changes.

**Reviewer sign-off: ready to merge into `next`.** Not merging; owner's call. WP4 note stands: cheese's `check-sync-id-parity.mjs` regex must accept `_[0-9a-f]{12}(_\d+)?`.

---

AN-BC:
@rkihm-BC this has gone through adverserials but please review and merge if good. Thanks

packages/GenericDatabaseProvider/src/GenericDatabaseProvider.ts
```diff
@@ -1075,6 +1076,107 @@ export abstract class GenericDatabaseProvider extends DatabaseProviderBase {
...
+    private _saveCallSuffixCounts = new WeakMap<object, Map<string, number>>();
+
+    /**
+     * Variable suffix for DECLARE/SET (or any named-local dialect) in a save call.
+     *
+     * Naming contract: `_<12 lowercase hex>` from
+     * `sha1(\`${schema}.${table}|${normalized pk values joined by |}\`)`, plus an optional
+     * `_<n>` with n ≥ 2 when that hash repeats inside one `TransactionGroup`
+     * (`_abc123456789`, `_abc123456789_2`, …). Outside a group there is no ordinal: each
+     * `Save()` is its own batch (MetadataSync captures put a batch separator after every
```
rkihm-BC, line 1135: This holds for `mj sync push`, which omits `variableBatchThreshold` and so gets the legacy `GO` after every statement. It does not hold for `mj sync watch` (`WatchService.ts:374` passes a threshold of 200) or for Explorer logging sessions (the server option schema defaults the threshold to 200): both accumulate statements into one batch with no separator until the variable count is reached, so two saves of the same record inside a window now redeclare identical names in one batch. See the review body for the reproduction and the logger-side fix; once that guard is in, this sentence can say "each `Save()` is its own batch, or the logger separates redeclarations".

rkihm-BC (changes requested):
Read the description, all four review rounds and their mutation tables, every comment, all six commits and the full diff, then traced the logging and transaction-group paths on the branch. Independently rebuilt the head in a worktree: GenericDatabaseProvider 987 passed / 5 skipped, SQLServerDataProvider 147 passed — matching the reported numbers. 24/24 checks green, changeset `patch` on both packages, and no other consumer of the old 8-hex suffix shape exists in this repository (the WP4 note about cheese's parity regex stands).

**The change is correct for everything it tests, and the placement is right**: a save call's identity is schema, table and key, not T-SQL grammar; the `WeakMap` keyed on the group is the right lifetime; the 12-hex widening removes the birthday-collision class rather than leaning on the ordinal; key normalization is complete and pinned; and the `scopeItemVariables` fix is a genuine pre-existing bug fixed well. The four rounds were properly adversarial and each required item was closed with a shown failing mutation.

### One required change: determinism collides with threshold-mode logging

The round-2 review ("Finding 2") looked at threshold mode and judged a hash collision inside a 25–30-save window harmless at ~1e-7. That covers two *different* records colliding. It stops one step short of the case this PR creates by design: the **same record saved twice** inside one window, where the suffixes are now equal on purpose.

Verified chain on the branch:

1. Saves hand their DECLARE/SET/EXEC form to the logger as `simpleSQLFallback` (`databaseProviderBase.ts:1048`); with `logRecordChangeMetadata` false the logger writes exactly that form, suffixed names included.
2. Explorer SQL logging sessions get `variableBatchThreshold: 200` and `logRecordChangeMetadata: false` from the server option schema defaults, spread into every session by `SqlLoggingConfigResolver`. `mj sync watch` passes a threshold of 200 explicitly (`WatchService.ts:374`).
3. In threshold mode `SqlLoggingSessionImpl` appends statements with no separator until the variable count is reached, and `_countVariableDeclarations` never looks at names.
4. SQL Server rejects a redeclared variable within one batch ("The variable name … has already been declared").

Probe against the branch's real `SqlLoggingSessionImpl` (threshold 200, two saves of one record using the branch's own `SaveCallVariableHash`): the file contained **two identical `DECLARE` lists and zero `GO` separators**. Before this PR the random suffix kept them distinct, so the capture replayed; after it, the default Explorer session produces a file that fails on replay whenever a record is saved twice in a window — an edit and a correction, the normal case. The `_n` ordinal cannot help: it exists only inside a transaction group, and neither path uses one.

Watch mode is the nuanced case: it sets `logRecordChangeMetadata: true`, so for record-change-tracked entities it already logs the full query with the wrapper's fixed-name `@ResultTable` / `@ID` declarations, which means two such saves in one threshold batch were unreplayable before this PR. The regression there is confined to entities without a record-change payload; Explorer sessions are the clean regression.

**Fix belongs in the logger, not the allocator.** In threshold mode, track the names declared in the current batch and emit the separator before any statement that redeclares one, resetting at each separator. `_countVariableDeclarations` already matches every declaration and only needs to capture the name. Output stays deterministic (which is the point of this PR), it is a few lines beside the existing counter, and as a side effect it also fixes the pre-existing fixed-name collision in watch mode. Test: two saves of one record logged in threshold mode produce a separator between them. Happy for this to be a companion PR merged first rather than a commit here; either way the capture feature shouldn't merge into a state where the default session's output fails on replay.

### Two wording corrections that follow from it

- Allocator JSDoc: "MetadataSync captures put a batch separator after every statement" is true for `mj sync push` (threshold omitted → legacy `GO` per statement) and false for `mj sync watch`. Inline comment.
- Description, "Where collisions actually bite": "Only a `BatchedSubmit` transaction group concatenates saves into one batch" — a threshold-mode capture file also concatenates saves into one batch on replay.

Nothing else. Approving once the logger guard is in (here or in a companion PR).

AN-BC:
**Reviewer: rkihm-BC's finding reproduces end to end at `63cbdd4909ab29fe66b52f790b450948916adb55`. Round-4 sign-off is withdrawn until the logger guard lands.**

I ran it through the real path rather than the logger in isolation: `SQLServerDataProvider.Save()` twice on one `Widget` record (the `save-delete-paths` harness, mssql mocked), logged through the provider's own `CreateSqlLogger` session, then read the capture file back.

| session options | `GO` in file | `DECLARE` lists | distinct `@Name_*` suffixes |
|---|---|---|---|
| Explorer defaults (`GO`, threshold 200, `logRecordChangeMetadata: false`) | **0** | 2 | 1 (`_6679d1fd77d5` both times) |
| legacy (`GO`, no threshold) | 2 | 2 | 1 |
| watch shape (threshold 200, `logRecordChangeMetadata: true`) on an entity with no record-change payload | **0** | 2 | 1 |

So the default Explorer session emits two identical seven-variable `DECLARE` lists in one batch, which SQL Server rejects on replay (Msg 134). My round-2 Finding 2 only priced two *different* records colliding; the same record twice is the case this PR makes certain, and I missed it. The root cause is on the logger side, exactly as rkihm-BC says: `_countVariableDeclarations` counts declarations but never looks at names, so a redeclaration cannot trigger a separator.

For the fix to close this item under the standing rules:

- Guard in `SqlLoggingSessionImpl` threshold mode: track names declared in the current batch, emit the separator before a statement that redeclares one, reset at each separator. Allocator untouched.
- A test that saves one record twice in threshold mode and asserts a separator between them, and that goes red with the guard removed. The three rows above are the shape; a version of that probe is fine as the test.
- The two wording fixes rkihm-BC listed (allocator JSDoc; description "Where collisions actually bite").

Companion PR merged first or a commit here are both fine. I will re-run both provider suites, mutations M1–M6, and this probe on the new head before restoring the sign-off.

---

AN-BC:
@rkihm-BC I need this in next so please **do** review and test, but this is going in now as I have a few other branches that need this. Thanks.

AN-BC:
looks like the last Claude review did **not** fix your review comments @rkihm-BC can you do that in a separate PR and merge on your end?

AN-BC:
**Post-merge note (reviewer).** This merged at `59ef725f07e8cad66f83a978f17e944a48d3c587` with rkihm-BC's required change still open: b03997e7 is a merge of `next` only, and `SqlLogger.ts` on `next` is byte-identical to the base, so the reproduced replay failure (two saves of one record at the Explorer logging defaults → identical `DECLARE` lists, no `GO`) is now on `next`.

The logger guard is up as #4273: name-aware separator in `SqlLoggingSessionImpl` threshold mode, a logger-level test and an end-to-end `Save()` test, the allocator JSDoc correction, changeset `patch`. Both provider suites green (993 / 5 skipped, 149) and the guard is mutation-proven. rkihm-BC is requested on it. The description wording item ("Where collisions actually bite") is recorded in #4273's body rather than edited here.

---

PR #4259, opened 2026-09-05
fix(sqlserver): deterministic save-call SQL variable suffixes, with a fail-able 120k test (supersedes #4257, loom #12 WP3)

AN-BC:
Supersedes #4257 (same first commit, `489aecd7`, kept intact) and adds the two round-1 review items on top. Implements **WP3** of [loom #12](https://github.com/MemberJunction/loom/pull/12) (F-D).

## What the WP3 commit does (unchanged from #4257)

`SQLServerDataProvider.RenderSaveCallBinding` no longer mints `@Field_<uuidv4 8 hex>`. The suffix is `sha1(`${schema}.${table}|${pk values}`)[:8]`; the same record saved twice inside one `SQLServerTransactionGroup` gets `_hash` then `_hash_2`. MetadataSync recaptures of an unchanged tree become byte-identical, and the birthday collisions seen across cheese's 120k-variable captures stop being random. PostgreSQL's binding is positional / json-arg and is untouched. Changeset: `patch`.

## What this PR adds (`6d57f6e5`)

1. **The 120,000-record test now drives the real allocator.** In #4257 it re-implemented the counting loop inline and asserted on its own copy, so it passed even with `SaveCallVariableHash` returning a constant. It now allocates suffixes for 120,000 distinct PKs on a real entity inside one transaction group (`Hydrate` per iteration, because `Set('ID')` does not clear the cached `PrimaryKey`) and pins what sha1 really does on those keys: two first-8-hex collisions (`id-42236`/`id-64356` → `031e1622`, `id-101514`/`id-104669` → `37ccdac1`), so 119,998 bare suffixes plus exactly two `_2`.
2. **The suffix naming contract is stated in the allocator's JSDoc**: `_<8 lowercase hex>`, then `_<n>` (n ≥ 2) only when the same hash repeats inside one transaction group; never outside one, which is what MetadataSync captures contain. Cheese's `check-sync-id-parity.mjs` regex keys on the bare form; WP4 should widen it to accept `(_\d+)?`.

## Verified by running (clean checkout, provider's 63 workspace deps built)

| Check | Result |
|---|---|
| `save-delete-paths.test.ts` | 14/14 |
| Full `@memberjunction/sqlserver-dataprovider` suite | 145/145 across 13 files |
| `pnpm run build` for the package | exit 0 |

Mutations of `SQLServerDataProvider.ts`, each restored before the next:

| Mutation | Old 120k test | New 120k test | Other suffix tests |
|---|---|---|---|
| M1 `uuidv4().substring(0, 8)` restored in `RenderSaveCallBinding` | passes | passes (allocator untouched) | `same record → same suffix` fails |
| M2 allocator ignores the transaction group | passes | **fails**: `expected '_6679d1fd' to be '_6679d1fd_2'` | `same record twice inside one transaction group` fails |
| M3 `SaveCallVariableHash` returns `'00000000'` | passes | **fails**: `expected '_00000000' not to be '_00000000'` | `different records → different suffixes` fails |

The old test surviving all three is why it was rewritten.

https://claude.ai/code/session_0111QTtJ2moiCVtEMeLknW9t

---
_Generated by [Claude Code](https://claude.ai/code/session_0111QTtJ2moiCVtEMeLknW9t)_

### DIFF
--- a/.changeset/deterministic-save-sql-suffix.md
+++ b/.changeset/deterministic-save-sql-suffix.md
@@ -1,0 +1,5 @@
+---
+"@memberjunction/sqlserver-dataprovider": patch
+---
+
+Make MetadataSync save-call SQL variable suffixes a PK hash instead of a random uuid slice, so recaptures of an unchanged tree are byte-identical and large batches do not collide (loom #12 WP3).
--- a/packages/SQLServerDataProvider/src/SQLServerDataProvider.ts
+++ b/packages/SQLServerDataProvider/src/SQLServerDataProvider.ts
@@ -1120,6 +1120,15 @@ export class SQLServerDataProvider
   /** Per transaction-group counts so the same PK twice in one batch gets `_hash_2`. */
   private _saveCallSuffixCounts = new WeakMap<object, Map<string, number>>();
 
+  /**
+   * Naming contract for the variable suffix appended to every `@Field` in a
+   * rendered save call: `_<8 lowercase hex>` (see `SaveCallVariableHash`),
+   * followed by `_<n>` with n >= 2 only when the same hash has already been
+   * allocated inside the same `TransactionGroup`. Outside a transaction group
+   * every save is its own batch, so the bare form is always emitted there —
+   * which is what MetadataSync captures contain. Downstream parsers (for
+   * example a MetadataSync ID-parity gate) may rely on this shape.
+   */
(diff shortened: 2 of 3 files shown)


AN-BC:
Closing as redundant: the builder pushed the equivalent rewrite to #4257 (`4f13ad1d`) minutes before this PR opened, and I have verified and signed off #4257 there. Nothing from this branch is needed on top of it.

---

PR #4263, opened 2026-09-05
docs: check in the AI-model research routine prompt, with the cost-status rule

AN-BC:
## Summary

Follow-up to #4254, closing the half of that fix that lives outside this repo.

The weekly AI-model research routine broke `next` twice the same way (#4030, #4110): deprecating a vendor set the paired `MJ: AI Model Costs` row to `Inactive`, which `chk_AIModelCost_Status` does not allow — the cost table takes `Active`/`Pending`/`Expired`/`Invalid`, and `Inactive` is a *vendor* value. #4254 fixed everything the repo controls: it documented the rule in `metadata/CLAUDE.md`, added `metadata/**` to the Integration Tier paths filter so the failure lands on the PR instead of on `next`, and made `mj sync push --ci` name the offending record. What it could not fix is the thing actually making the mistake, because the routine's prompt lives in a Claude Routine, not in git.

This PR checks that prompt in and adds the rule to it. **The Routine's own prompt has already been updated to match** (Craig/Amith, out of band) — this is the mirror, so the rule is reviewable, greppable, and has somewhere to point from `DEPLOYMENT.md`.

## Changes

**`reports/ai-model-research/ROUTINE_PROMPT.md`** (new) — the routine's prompt, with a new §0 that runs before any JSON is touched:

- **The `Status` tables, side by side.** `MJ: AI Model Vendors` takes `Active`/`Inactive`/`Deprecated`/`Preview`; `MJ: AI Model Costs` takes `Active`/`Pending`/`Expired`/`Invalid`. The two arrays sit adjacent inside the same model record, which is how the value got copied across.
- **The deprecation recipe**, with a worked example: vendor row `Inactive`, paired cost row `Expired` **plus** `EndedAt` (`chk_AIModelCost_DateRange` wants `EndedAt > StartedAt`). Plus the errors in the other direction — leave the Model Developer row Active, leave other vendors alone, leave `IsActive: true` unless every vendor has dropped the model.
- **A pointer at the right reference implementation.** GLM-4.7-on-Cerebras in `.ai-models.json` is correct; #4110's Kimi K2.5 diff is the broken one, and its report says it was copying the Cerebras pattern. Since "match existing patterns" is guideline #3 and is what carried the error, §0.2 is now an explicit exception to it.
- **An offline pre-flight check.** `metadata/CLAUDE.md` says run `mj sync validate`, but that calls `initializeProvider()` and needs a live database — the routine's sandbox has none, so the instruction would silently no-op. Replaced with a pure-Node JSON scan the routine can actually run.
- **Watch the Integration Tier** (now gating per #4254) and fix red on the branch. A red research PR at release time gets skipped, and per `DEPLOYMENT.md` Step 2 that means the release ships with no model refresh.
- **The `minor` changeset** for a metadata branch. The PRs already do this in practice; the prompt never said so.

**`DEPLOYMENT.md`** — Step 2 now points at the prompt file (so "run the routine" has something to point at), and the sanity-check step names the `Status` trap directly.

## Verification

- The pre-flight snippet was run against current metadata: prints `OK`.
- It was also run against a reconstruction of the #4110 shape, and fails as intended: `BAD cost Status "Inactive" on Kimi K2.5`. Same for an `Expired` row with no `EndedAt`. A check that cannot fail is not a check.
- Current metadata audits clean — 250 `Active` + 2 `Expired` cost rows, 0 invalid.
- `pnpm run check:changeset` — green.
- `pnpm run check:claude-md` — 3 failures, all pre-existing on `next` (root `CLAUDE.md` over its byte ceiling; its `../CLAUDE.md` M5-workspace link unresolvable in a standalone checkout). Neither file is touched here, and the gate validated the new links clean.

Docs-only; no package code changes, so no changeset (the gate agrees).

## Note for reviewers

The file is a **mirror, not the execution source** — the Routine stores its own copy and was created via the HTTP API, so no agent can update it. The file says this, and carries a change log, so a future edit here doesn't get mistaken for one that took effect. If the two drift, the prompt tells the routine to say so in its PR body.

Separately, and not addressed here: 34 vendor rows are `Inactive` while only 2 cost rows are `Expired`, so most historically-deprecated routes still carry `Active` pricing. Not a CI failure and nothing enforces it, but cost queries will return prices for routes nobody serves. Worth a decision at some point.

https://claude.ai/code/session_01QawvzQBLm3RTcBS3QigH4V

---
_Generated by [Claude Code](https://claude.ai/code/session_01QawvzQBLm3RTcBS3QigH4V)_

### DIFF
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -177,8 +177,15 @@ Do this **before** the metadata sync step (Step 3) so new models are captured in
 2. **If no AI model research PR exists**, run the Claude AI-model-research routine to generate one
    (the same routine that produced the PRs in `reports/ai-model-research/` and PR #2924), then merge it
    as in step 1. Do not skip the release's model refresh just because a PR wasn't waiting — generate it.
+   The routine's prompt is checked in at
+   [`reports/ai-model-research/ROUTINE_PROMPT.md`](reports/ai-model-research/ROUTINE_PROMPT.md) — run it
+   from there if you are driving it by hand. That file is a **copy** of the live Routine's prompt; if you
+   change one, change the other, or the next weekly run reverts to the old rules.
 3. **Sanity-check** the merged entries against `metadata/ai-models/.ai-models.json` and confirm
-   `@lookup:` references resolve.
+   `@lookup:` references resolve. Per-entity `Status` values are the failure mode that has bitten twice
+   (PRs #4030, #4110): a deprecated vendor row is `Inactive`, but its paired **cost** row must be
+   `Expired` with an `EndedAt` — `Inactive` is not a legal cost status and fails the sync push. The
+   pre-flight check in §0.3 of the routine prompt catches it offline; `metadata/CLAUDE.md` states the rule.
 4. Run `mj sync push --dir ./metadata` to sync to your local database — the changes are then captured
    in the metadata migration script generated in Step 3.
 
--- a/reports/ai-model-research/ROUTINE_PROMPT.md
+++ b/reports/ai-model-research/ROUTINE_PROMPT.md
@@ -1,0 +1,518 @@
+# Weekly AI-model research routine — prompt of record
+
+This file holds the prompt driving the automated Claude Routine **"AI Model & Vendor Weekly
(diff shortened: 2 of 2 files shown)


AN-BC:
@cadam11 fyi

PR #4267, opened 2026-09-06
fix(mcp-server): only redirect authorize errors to a validated redirect_uri

cadam11:
## What

`/oauth/authorize` in the MCP OAuth proxy reported some errors by redirecting to the `redirect_uri` the request supplied, before that URI had been matched against a registered client. RFC 6749 §4.1.2.1 and the OAuth 2.0 Security BCP are explicit that this is the one thing the authorization endpoint must not do: until the redirect URI is known to belong to a known client, the server has no basis for sending a browser to it, and the error has to be rendered locally instead.

I audited every error exit in the authorize flow rather than only the branches I first noticed. Three of them were reachable before the client and redirect URI were validated.

| Authorize error branch | Stage | Before | After |
|---|---|---|---|
| `client_id` missing | pre-validation | redirect to the supplied URI | local error page |
| `redirect_uri` missing | pre-validation | local error page | local error page (unchanged) |
| `response_type` not `code` | pre-validation (ran before the client lookup) | redirect to the supplied URI | moved after validation — redirects to the registered URI |
| unknown `client_id` | pre-validation | redirect to the supplied URI | local error page |
| `redirect_uri` not registered for the client | pre-validation | redirect to the supplied URI | local error page |
| `code_challenge` missing (PKCE) | post-validation | redirect to the registered URI | unchanged |
| `code_challenge_method` not `S256` | post-validation | redirect to the registered URI | unchanged |
| callback: no code from the provider | post-validation (uses the stored state's URI) | redirect to the registered URI | unchanged |
| callback: upstream token exchange failed | post-validation (uses the stored state's URI) | redirect to the registered URI | unchanged |

The remaining error exits in the flow — an upstream provider error, a missing or expired `state` on the callback, a user who fails authorization, an expired consent session — already rendered locally and were never redirect-shaped.

## The fix

Structural rather than per-branch, so a future branch added above the validation cannot reintroduce it. The handler identifies the client and validates the redirect URI first, and everything in that stage answers with `sendErrorPage`. The query parameter is destructured under a deliberately unusable name, and the redirectable binding is only created after the match succeeds — so no branch above that line has a URI it could redirect to, however it fails. `sendAuthorizationError()` now takes a required redirect URI rather than `string | undefined`, which deletes its silent error-page fallback and makes an unvalidated call a type error.

The one behavioral change beyond error rendering: `response_type` is now checked after the client and redirect URI rather than before, so an invalid `response_type` on a *valid* client reports itself as an `unsupported_response_type` redirect to the registered URI, which is what the spec asks for and what it did before whenever the URI happened to be registered.

## Verification

Nine new tests in `src/__tests__/OAuthProxyAuthorizeErrors.test.ts`, driving the real router over a real HTTP listener on an ephemeral port with redirects unfollowed. Five pin the pre-validation branches as `400` with no `Location` header; three pin each post-validation branch as a `302` back to the registered URI carrying `error`, `error_description` and `state`; one pins that a fully valid request still redirects to the upstream provider.

Against the pre-fix handler, exactly the four pre-validation assertions fail (`302`, `Location` pointing at the request's own URI) and the five post-validation and success pins pass — so the suite fails on the behavior being fixed and on nothing else. With the fix: 9 passed, and the package's full suite is 140 passing. Package builds clean; ESLint on the touched file goes from 41 pre-existing errors to 40, with none added (the file's existing Prettier drift is left alone rather than reformatted into this diff).

## Notes for the reviewer

No GitHub issue — this came out of a local security review of the OAuth proxy while enabling it against a Cognito pool, not from a report. I have kept the description to the shape of the defect rather than how it would be used.

I have a second, unrelated PR open against the same package (#4266, Cognito upstream endpoints). The two branches are independent and touch different files; this one is off `next` and does not depend on it.

No schema change, no migration, no metadata, no CodeGen output. Changeset is `patch`.

### DIFF
--- a/.changeset/mcp-oauth-authorize-error-handling.md
+++ b/.changeset/mcp-oauth-authorize-error-handling.md
@@ -1,0 +1,11 @@
+---
+"@memberjunction/ai-mcp-server": patch
+---
+
+Harden `/oauth/authorize` error handling in the MCP OAuth proxy, per RFC 6749 §4.1.2.1 and the OAuth 2.0 Security BCP.
+
+An error may only be reported by redirecting to a request's `redirect_uri` once that URI has been matched against a registered client. Several checks in the authorize handler ran before that match was possible and still reported themselves as an error redirect to the request-supplied URI.
+
+The handler now identifies the client and validates the redirect URI first, and answers every failure in that stage with a locally rendered error page. Only after the match succeeds does a redirectable binding come into scope, so no earlier branch can redirect to a caller-supplied URI however it fails. `sendAuthorizationError()` now requires a redirect URI rather than accepting `undefined`, which makes an unvalidated call a compile error rather than a silent fallback.
+
+Behavior after validation is unchanged and pinned by tests: an unsupported `response_type`, a missing PKCE challenge, or a non-S256 challenge method still redirect to the registered URI carrying `error`, `error_description` and `state`, and a valid request still redirects to the upstream provider.
--- a/packages/AI/MCPServer/src/__tests__/OAuthProxyAuthorizeErrors.test.ts
+++ b/packages/AI/MCPServer/src/__tests__/OAuthProxyAuthorizeErrors.test.ts
@@ -1,0 +1,193 @@
+/**
+ * Tests for how /oauth/authorize reports errors.
+ *
+ * The rule under test is RFC 6749 section 4.1.2.1 / the OAuth 2.0 Security BCP: an error may only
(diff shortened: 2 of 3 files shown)


cadam11:
Closing in favor of #4265, which supersedes this: it covers all four pre-validation authorize branches (including the two here) and additionally fixes array-valued query params, log injection, and secret prefixes in error logs, with its own tests and no new CodeQL findings.

PR #4268, opened 2026-09-06
fix(integration-engine): PrefetchContentHashes must bypass the RunView cache

MS-BC:
## Problem

`PrefetchContentHashes` issues one existence/content-hash `RunView` per apply batch, keyed by that batch's ID set. Every batch's ID set is a **unique fingerprint**, so this query can never be a cache HIT — without `BypassCache` each call only **deposits a dead cache entry**, and the RunView cache grows `O(records processed)` for the life of the process. Observed in production: a 500k+ record drain grew the sync host's memory to the kernel OOM kill line through exactly this path.

## Fix

One line: `BypassCache: true` on the prefetch's `RunView`, with a comment carrying the reasoning. This is the same point-in-time-read rationale as `LoadAllRecordMaps`' existing `BypassCache` ("sync decisions must reflect committed record-map state") — a per-batch existence read is never a reusable result.

## Test

`PrefetchProvesAbsence.test.ts` gains a `BypassCache` assertion beside the existing `IgnoreMaxRows` one, using the suite's captured-params harness. Red before the fix, green after — 12/12 pass.

## Changeset

`@memberjunction/integration-engine` patch.

### DIFF
--- a/.changeset/prefetch-content-hashes-bypasscache.md
+++ b/.changeset/prefetch-content-hashes-bypasscache.md
@@ -1,0 +1,5 @@
+---
+'@memberjunction/integration-engine': patch
+---
+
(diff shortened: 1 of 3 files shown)


MS-BC:
Withdrawing — holding this internally for now.

PR #4270, opened 2026-09-06
feat(integration-engine): in-process sync cancel, complementing the durable DB stamp

MS-BC:
## Summary
- Adds `IntegrationEngine.RequestCancelInProcess(companyIntegrationID)`: an in-process registry mapping a live run to the same `(cancelRequested flag, AbortController)` pair its existing boundary-check/heartbeat cancel path already honors — so a cancel issued in the SAME process as the run trips it immediately instead of waiting for the next DB poll.
- `IntegrationCancelSync` now falls back to this when the durable `CancelSyncAsync` stamp reports nothing cancelled. The DB stamp is still written first and unconditionally, so cross-process cancel correctness (owner in another process/node/worker) is unchanged.
- Also closes a real gap: on a deployment whose schema predates the ownership columns `CancelSyncAsync` needs, the durable stamp is a no-op — leaving no working cancel path at all. The in-process registry needs no schema, so it is the only cancel path that works there. (This exact implementation has run in production on such a deployment since 2026-09-04.)

## Test plan
- [x] New unit tests in `packages/Integration/engine/src/__tests__/IntegrationEngine.inprocess-cancel.test.ts`: no-op when nothing registered; hook fires exactly once and returns true; case-insensitive key match; deregistered-after-finish returns false. Written red-first.
- [x] Full `Integration/engine` vitest suite: no regressions (the 43 pre-existing, unrelated failures are byte-identical with the change stashed out).
- [x] `tsc --noEmit` baseline-identical on both touched packages (zero new error categories).
- [x] Changeset (patch) for `@memberjunction/integration-engine` and `@memberjunction/server`.

### DIFF
--- a/.changeset/in-process-sync-cancel.md
+++ b/.changeset/in-process-sync-cancel.md
@@ -1,0 +1,14 @@
+---
+'@memberjunction/integration-engine': patch
+'@memberjunction/server': patch
+---
+
+In-process sync cancellation, complementing the durable `CancelSyncAsync` DB stamp.
+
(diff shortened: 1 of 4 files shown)


MS-BC:
Withdrawing — holding this internally for now.

PR #4273, opened 2026-09-06
fix(sql-logging): separate batches before a redeclared variable in threshold mode (follow-up to #4257)

AN-BC:
Follow-up to #4257 (loom #12 WP3), which merged with rkihm-BC's one required change still open. This is that change.

## Problem

Save-call variable suffixes are deterministic per record since #4257, so the same record saved twice emits identical `DECLARE` lists. Threshold-mode logging sessions (the Explorer SQL logging defaults: `GO`, `variableBatchThreshold: 200`, `logRecordChangeMetadata: false`; and `mj sync watch`) accumulate statements into one batch until the declaration count reaches the threshold, and `_countVariableDeclarations` never looked at names. An edit followed by a correction of one record therefore produced two identical `DECLARE` lists in one batch, which SQL Server rejects on replay (Msg 134). Reproduced end to end on #4257's head before this PR: two `Save()` calls on one record at the Explorer defaults gave 0 `GO` separators and 2 identical `DECLARE` lists.

## Change

- `SqlLoggingSessionImpl` (threshold mode only): track the lower-cased names declared in the current batch and emit the separator before any statement that would redeclare one; reset the set at each separator. The count-based split and legacy per-statement mode are unchanged. `_countVariableDeclarations` now delegates to `_collectVariableDeclarations`.
- `allocateSaveCallSuffix` JSDoc no longer claims MetadataSync captures put a separator after every statement (true for `mj sync push`, not `mj sync watch`).
- Changeset: `patch` on `@memberjunction/generic-database-provider`.

Not changed: #4257's merged description still says only a `BatchedSubmit` group concatenates saves into one batch; a threshold-mode capture also does, which is what this guard covers.

## Tests, by gate

**Logger-level** (`GenericDatabaseProvider/src/__tests__/sql-logger-redeclare-guard.test.ts`, real session through `initialize` → `logSqlStatement` → `dispose` to a temp file):
- same record twice → one separator, and no batch redeclares a name
- different records below the threshold → no separator
- names compared case-insensitively, as T-SQL does
- name set resets after each separator (A, B, A → 1; A, A, A → 2; A, A, B → 1)
- count-based split still fires at the threshold for distinct names
- legacy mode (no threshold) unchanged: one separator per statement

**End to end** (`SQLServerDataProvider/src/__tests__/sql-logging-threshold-replay.test.ts`, real `Save()` → `GenerateSaveSQL` → `RenderSaveCallBinding` → `simpleSQLFallback` → real `SqlLoggingSessionImpl`, only mssql mocked): one Widget saved twice at the Explorer defaults → identical 12-hex suffixes on both saves, exactly one `GO` between the two `DECLARE` lists, no batch redeclares a variable; legacy mode → two `GO`.

**Mutation:** with the name check replaced by `false` and the package rebuilt, 3 of 6 logger tests and the end-to-end test fail (`expected 1 to be 2` on the batch count); the three that stay green are the ones pinning unchanged behavior.

```
cd packages/GenericDatabaseProvider && pnpm run build && npx vitest run   # 993 passed | 5 skipped (22 files)
cd packages/SQLServerDataProvider && npx vitest run                        # 149 passed (14 files)
```

ESLint on the three touched files: the two test files are clean; the six prettier findings left in `SqlLogger.ts` all exist on `next` today and are outside this diff.

https://claude.ai/code/session_0111QTtJ2moiCVtEMeLknW9t

---
_Generated by [Claude Code](https://claude.ai/code/session_0111QTtJ2moiCVtEMeLknW9t)_


rkihm-BC (approved):
This is the change #4257's review asked for, done in the right place and with a wider test surface than requested.

**Verified independently at this head** (rebuilt in a clean worktree):

| Check | Result |
|---|---|
| `@memberjunction/generic-database-provider` suite | 994 passed, 5 skipped |
| `@memberjunction/sqlserver-dataprovider` suite | 149 passed |
| The probe that failed on #4257 (two saves of one record, Explorer defaults, real `SqlLoggingSessionImpl`) | now passes — a `GO` between the two identical `DECLARE` lists |
| CI | 24/24 green, including the SQL Server deterministic integration tier (8m01s) |

**Why it's right:** the guard lives in the logger, where the batch boundary is decided, so the allocator stays deterministic and knows nothing about sessions. A redeclaration can only be detected when the batch already holds names, so the existing count guard is implied; the name set is rebuilt at every separator; the count-based split and legacy per-statement mode are unchanged and pinned by tests. The end-to-end test is the valuable one — real `Save()` → `GenerateSaveSQL` → `RenderSaveCallBinding` → `simpleSQLFallback` → real session, only mssql mocked — because it proves the hand-off the logger depends on, not just the logger.

One thing the description doesn't claim but the change delivers: `_collectVariableDeclarations` already matches `@ResultTable TABLE` and `@ID NVARCHAR`, so the same guard separates the record-change wrapper's fixed-name declarations too. That closes the pre-existing watch-mode collision (`logRecordChangeMetadata: true` + threshold 200) noted on #4257 — worth a line in the changeset if you touch it again, not a blocker.

Changeset `patch` on the generic provider is correct; the SQL Server package only gains a test. Approving.

PR #4274, opened 2026-09-06
feat(ng-entity-viewer): chrome workspace|embedded; drop hosted grid Search

AN-BC:
## Summary
- Adds `EntityViewerConfig.chrome`: `workspace` (Explorer default) or `embedded` (dashboard peek).
- `embedded` turns off filter, view-mode toggle, record count, and pagination (`pageSize: 8`). Explicit booleans still win.
- Embedded seeds the Grid plug-in with `showToolbar: false` and `showPager: false` unless the view-type blob already set them.
- Hosted grids no longer show a second Search box (`ShowSearch` is false unless `toolbarConfig.showSearch: true`). The container already owns filter text.

## Test plan
- [x] `ResolveViewerConfig` unit tests and DOM test that embedded hides `.viewer-header`.
- [ ] Directory dashboard Latest people / Latest relationships: no Filter records, no Search, no pager.
- [ ] Explorer entity workspace still has Filter records in the header (no second Search in the grid toolbar).

### DIFF
--- a/.changeset/embedded-viewer-chrome.md
+++ b/.changeset/embedded-viewer-chrome.md
@@ -1,0 +1,5 @@
+---
+"@memberjunction/ng-entity-viewer": patch
+---
+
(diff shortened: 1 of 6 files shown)


AN-BC (commented):
LGTM

PR #4275, opened 2026-09-07
fix(security): close the #4260 privilege-escalation path on MJ: Users

### LINKED ISSUE #4260: Auto-provisioned users get the Developer role, which has unrestricted update on Users (Type and Name included)
**Found by** code review of #4231 (`fix/4209-provisioning-context-user`), 2026-09-05. Not caused by that PR — verified pre-existing, and independent of the context-user ladder that PR touches.

### What happens

The shipped default for `userHandling.newUserRoles` grants every auto-provisioned user the `Developer` role:

- `packages/MJServer/src/config.ts:666` — `newUserRoles: ['UI', 'Developer']`

On the baseline seed, that role holds unrestricted update on the `Users` entity. Queried against a baseline database:

| Role | `CanUpdate` | `UpdateRLSFilterID` |
|---|---|---|
| `UI` | `false` | `null` |
| `Developer` | **`true`** | **`null`** |
| `Integration` | **`true`** | **`null`** |

`MJ: Users` also has `AllowUpdateAPI = true`. There is no row-level security filter on the update grant, so the permission is not scoped to the caller's own row.

### Why it matters

Anyone who can obtain an account through JWT auto-provisioning or a magic-link redeem lands, by default, with write access to every row in `Users` — including `Type` and `Name`. Two consequences:

1. **Privilege escalation.** `Type` is the column the platform's Owner checks read. A user who can write it can grant themselves the level those checks gate.
2. **The context-user ladder's first rung becomes attacker-influenced.** `resolvePrincipalFrom` (`packages/MJServer/src/auth/principals.ts`) matches the configured context user against `User.Name` first, deliberately, for backward compatibility. That ordering is sound *only* if `Name` is not writable by untrusted parties. Under the stock grant it is. This does not make #4231 wrong — the ladder resolved against `Name` before that PR too, and the same exposure existed — but it means the ordering rests on an assumption the default config does not currently support.

The escalation path is the primary issue here; the ladder is a consequence, not the cause. Fixing `newUserRoles` or the grant addresses both.

### Suggested directions

Not obvious which is right, hence an issue rather than a PR:

- Drop `Developer` from the `newUserRoles` default, leaving `['UI']`. Smallest change, but may break hosts that rely on the default for legitimate developer access.
- Attach an update RLS filter to `Users` scoping the `Developer` grant to the caller's own row.
- Narrow the grant so `Type` and `Name` are not user-writable even where row update is allowed.

Any of these is a behavioural change for existing hosts and wants a deliberate decision plus a migration.

### Verification

Against a baseline-seeded database:

```sql
SELECT r.Name AS RoleName, ep.CanUpdate, ep.UpdateRLSFilterID, e.AllowUpdateAPI
FROM __mj.Entity e
JOIN __mj.EntityPermission ep ON ep.EntityID = e.ID
JOIN __mj.Role r ON r.ID = ep.RoleID
WHERE e.BaseTable = 'User' AND e.SchemaName = '__mj';
```

Recorded in the plan's trust table at `plans/provisioning-context-user-resolution.md` (row "Trust (match target)").

https://claude.ai/code/session_018Tc2hdDrn3bz5319Lyqnuk

SDesai-BC:
Closes #4260.

## The bug

`DEFAULT_SERVER_CONFIG.userHandling.newUserRoles` shipped as `['UI', 'Developer']`. On the baseline seed the `Developer` role holds unfiltered `CanUpdate` on **439 of the database's 446 entities** — `MJ: Users` among them, where `AllowUpdateAPI` is true on the entity and on both the `Type` and `Name` fields, with no row-level filter scoping the grant to the caller's own row.

Every gate in the chain passed:

```
config default → newUsers.ts assigns the role → EntityPermission grants Update
→ Entity/Field AllowUpdateAPI allow it → BaseEntity.CheckPermissions passes
→ GRANT EXECUTE ON __mj.spUpdateUser TO cdp_Developer
```

So anyone who obtained an account through JWT auto-provisioning or a magic-link redeem could write **any** row of `MJ: Users`, including setting their own `Type` to `'Owner'` — the column every superuser check in the platform reads (`SqlLoggingConfigResolver`, magic-link `canIssueInvites`, the backup-system-user fallback). Auto-provisioning is on by default with no domain restriction, so the standing authority was "anyone the configured IdP will issue a token for".

Two things made it worse than the issue described. First, the escalation is not the only consequence: `Name` is writable too, and `resolvePrincipalFrom` matches the configured context user against `User.Name` **first**, breaking ties by lowest ID — so a user who could rename themselves could redirect which principal the server provisions as. Second, this was never only a bare-install default: `loadConfig()` deep-merges through `mergeConfigs`, so a host writing a *partial* `userHandling` block inherited the array and the Zod `.default([])` never fired.

## The fix, in two halves

**Reachability** — the shipped default is now `['UI']`, the seeded end-user role, which holds no write on `MJ: Users`. Corrected in `config.ts` and in the two templates copied into real deployments (`MJServer/README.md`, `MJCLI/mj.config.cjs`).

**Capability** — a new `MJUserEntityServer` (`@RegisterClass(BaseEntity, 'MJ: Users')`) enforcing, for any caller whose `Type` is not `'Owner'`:

| # | Invariant |
|---|---|
| 1 | **Create is refused.** Closes a bypass where a non-Owner mints a row whose `Name` matches the configured context user (`Name` has no unique index) and wins the lowest-ID tiebreak. |
| 2 | **`Type` may not change** on an existing row. |
| 3 | **Own row only**, compared against the **pre-save** `ID`, and failing *closed* if the pre-image cannot be established. |
| 4 | **`Name` may not change** on an existing row — the same ladder matching runs on update. |
| 5 | **Delete is refused.** MJ deactivates via `IsActive`; an unguarded delete let a non-Owner remove any account, Owners included. |
| 6 | **A `ReplayOnly` save is refused.** `ReplayOnly` is the one option that makes `BaseEntity.Save()` skip `Validate()` altogether, which is where 1–4 live. See below. |

Owner-type callers are exempt, so admin user management and the seeded system user that provisioning runs under are unaffected on a default install.

The guard is the durable half: enforcement sits in `Validate()` plus `Save()` and `Delete()` overrides, all of which run inside `BaseEntity.Save()`/`.Delete()`, so it holds on **every** write path — GraphQL resolvers, REST, Remote Operations, the Create/Update Record actions, transaction groups, metadata sync, one-off scripts — and for **any** role a deployment invents, not just the two seeded ones. The config default only governs who gets the seeded roles.

The `Save()` override is what makes that sentence literally true rather than nearly true, and it was added after review. `Validate()` alone does **not** cover every write path: `BaseEntity.Save()` force-passes validation *without calling `Validate()`* when `EntitySaveOptions.ReplayOnly` is set (`baseEntity.ts:3725`), and `ReplayOnly` does not suppress the write — the provider only uses it to bypass the `AllowUpdateAPI`/`AllowCreateAPI` gates before executing the SQL. Invariants 1–4 were therefore skippable by a save option, while invariant 5 was not, because `Delete()` is an override and an override cannot be switched off. Reproduced against a live server before the fix: as a non-Owner holding `CanUpdate`, `Save({ReplayOnly:true})` with a dirty `Type` returned `true` and the `UPDATE` executed.

No untrusted caller could reach that today — every wire path accepting `ReplayOnly` was enumerated: the GraphQL `options___`/`DeleteOptionsInput` input exists only on the DELETE mutation (`ResolverBase.CreateRecord`/`UpdateRecord` call `Save()` with no options at all), REST's `EntityCRUDHandler` accepts it but calls `entity.Validate()` explicitly first, and `graphQLSystemUserClient` requires the system API key. It is closed anyway, because "holds on every write path" is a promise a future wire path forwarding save options would otherwise quietly break.

**A third change, in Explorer.** Guarding the entity turned a previously-succeeding save into a refused one on a screen that did not check whether its saves succeeded. `user-management.component.ts`'s `toggleUserStatus` discarded the result of `user.Save()`. Deactivating a user is a write to *another* user's row, so the guard refuses it for every non-Owner administrator — on the screen's primary function. The refusal surfaced two different ways depending on transport, and the method handled neither:

- **Direct/server-side provider**, where `BaseEntity.Save()` returns `false` on a validation failure rather than throwing: the `catch` never ran, so there was no revert, no message, not even the `console.error`, and `calculateStats()` re-rendered the row as toggled. The UI asserted success while nothing had been written, and the change vanished on the next reload.
- **GraphQL provider**, which throws (`code: SAVE_ENTITY_ERROR`): the `catch` did run and reverted the row, but it only logged — it never set the component's `error`, so the user saw the toggle silently snap back with no explanation.

It now checks the result and throws with `LatestResult.Message`, so both shapes land in the same catch, which surfaces the message through the component's existing error banner and clears it on the next successful toggle — matching `deleteUser()` in the same component, which already handled its result this way. This fixes the *silent* failure only; a non-Owner still cannot administer other users, by design, and the screen does not yet hide or disable the controls for them (tracked in #4281).

**Why not row-level security.** The issue suggested an own-row RLS filter on the `Developer` grant. That would *not* have closed this: scoping the update to the caller's own row still permits setting one's **own** `Type` to `'Owner'`, which is the escalation. MJ has no per-role *field* permission, and `EntityField.AllowUpdateAPI = 0` on `Type` is global — it would block Owners too and break every legitimate admin path. An invariant in the save path is the only mechanism that expresses the actual rule.

## Verification

- **Unit** — `MJCoreEntitiesServer` 476/476 pass (23 cases on the guard, covering every branch including the create and delete refusals, their Owner-allowed counterparts, and the `ReplayOnly` refusal). `MJServer` config-contract test 5/5, whole package 1260 passed / 56 skipped. `ng-explorer-settings` 100/100. `integration-test-suite` 200/200.
- **Integration** — new bundle `user-elevation` / `IT88`, checks UE1–UE4, against a live server:
  - deterministic lane: **3 passed / 0 failed / 1 skipped** — UE4 is mutation-gated (see below)
  - `RUN_MUTATION_TESTS=1`: **4 passed / 0 failed / 0 skipped**

  UE4 is the real reproduction: it runs as an actual non-Owner principal holding `CanUpdate` on `MJ: Users` and asserts the live `Save()` is refused. It is transaction-wrapped with rollback in a `finally`, because `Type` has only two values — so any dirty `Type` change by a non-Owner is an escalation attempt by construction, and an unprotected check would promote a real account precisely when the guard regressed.

  **UE4 now declares `RequiresMutation: true`.** It was `false`, which ran it in the deterministic lane — the lane defined as non-mutating. The rollback makes the write *invisible*, not *absent*: UE4 issues a genuine `UPDATE __mj.User SET Type=…` through `spUpdateUser`, plus a `RecordChange` row since `MJ: Users` has `TrackRecordChanges=1`, and `RequiresMutation` gates the write rather than its visibility. Two further hazards argued against running it unarmed: `BeginEntityTransaction()` opens on the *shared* provider, whose own implementation logs a hard error for that shape; and had the guard regressed, the successful `Save()` would leave `UserCache` holding the promoted `Type` for the rest of the process even after the database rolled back. `EXPECTED_MUTATION_GATED` in `check-registry.test.ts` is updated so the change is pinned rather than silently shrinking the tier. UE1–UE3 stay ungated and never call `Save()`, so the guard is still proven on every deterministic run — only the real write moved behind the gate.
- **Deterministic tier** — 63 passed / 3 failed / 6 skipped, unchanged before and after these changes. The 3 failures are pre-existing drift on the dev database and are unrelated to this branch: IT26 (Application row vs provider-cache mismatch), IT45 (five `EmbeddingVector` payloads stored as empty arrays on `MJ: AI Agent Notes` / `MJ: AI Agent Examples` — data drift, not artifact drift) and IT50 (CD3 a stale `MJ: AI Skills.ActivationScope` schema key, CD5 a stale `MJConversationSkill` artifact, CD6 generated 383 vs live 382 entities). IT50 was checked specifically because it is the one bundle that inspects `ClassFactory` registrations and this PR adds one — its **CD1 sub-check passes**, so the new registration is clean.
- **UI, driven live in Explorer** — signed in as a non-Owner (`Type = 'User'`) holding `CanUpdate` on `MJ: Users`, i.e. exactly the administrator this change affects. The screen is reached as a dashboard (`@RegisterClass(BaseDashboard, 'UserManagement')`), not a settings panel.
  - the user list renders populated for that caller (7 of 7)
  - clicking **Deactivate** on another user puts the guard's own message on screen — *"You may only modify your own user record. Changing another user's record requires an Owner."* — and fires the `console.error` that previously never executed. `IsActive` was `1` before and `1` after: **nothing was written**
  - no failed network requests; every GraphQL call returned HTTP 200 with the refusal carried in the body
  - one caveat: the banner-clearing behaviour could not be exercised live, because it needs a *successful* toggle after a refused one and the only row such a caller can successfully toggle is their own. That path stays unit-covered.

  **Worth knowing before you review the screen:** the refusal does not render inline on the offending row — it replaces the entire user list with a full-page error state plus a "Try Again" button, which restores the list intact. That is pre-existing template behaviour (`deleteUser()` sets the same page-level `error`), but this change makes a rare presentation the routine one for non-Owner admins. Filed against the screen's Owner-gate issue, #4281, rather than changed here.
- **Local CI mirrors** — `check:changeset`, `check:esm`, `check:codegen-tail`, `check:standards` all green.
- **Class-registration manifests regenerated.** Without this the guard was tree-shakeable out of a server bundle — `ClassFactory` would have fallen back to the generated `MJUserEntity` and the guard would have been a silent no-op. It would also have failed the CI manifest-freshness gate.

## Not fixed here

- The **`EntityPermission` grant itself** is unchanged: `Developer` and `Integration` still hold unfiltered create/update/delete on `MJ: Users` and update on ~439 other entities. The guard makes that grant non-escalating for `MJ: Users`, but narrowing it is a behavioural change for existing hosts that needs its own migration. — #4283
- **`MJ: Roles` and `MJ: User Roles` are unguarded.** Granting yourself a role is a separate elevation route with its own legitimate-write analysis. Deliberately out of scope here. — #4282
  Worth stating precisely, since it bounds the reachability half: on the baseline seed `UI` holds `CanCreate=0, CanUpdate=0, CanDelete=0` on **both** entities, so a `UI`-only auto-provisioned user cannot grant itself a role. The open route is narrower — a user who *already* holds `Developer` or `Integration` can grant themselves more.
- **Explorer's user-management screen still has no Owner gate**, so a non-Owner administrator now meets a refusal on every mutating control rather than a disabled UI. — #4281

Also filed while verifying this PR: #4284 (`next` is red for unrelated reasons), #4285 (`newUserRoles` is duplicated across three files with nothing keeping them in sync), #4286 (the `Email`-mutable decision depends on `UQ_User_Email`, which the guard never asserts), #4287 (a non-Owner `contextUserForNewUserCreation` fails at first login rather than at startup), #4288, #4289.

## Upgrade impact

All three changesets carry the operator-facing detail. In short: a non-Owner can no longer create, delete, rename, retype, or modify **any other user's row**. If your deployment administers users through a non-Owner role — including Explorer's user-management screen, which has no Owner gate of its own — move that to an Owner-type user first. And if `contextUserForNewUserCreation` / `contextUserForProvisioning` names a **non-Owner**, point it at an Owner before upgrading or provisioning will fail closed.

https://claude.ai/code/session_018UHb9RReLbwGiEXirDbGr9


.changeset/new-user-roles-drop-developer.md
```diff
@@ -0,0 +1,25 @@
...
+
+What is *no longer* true: this paragraph originally warned that a deployment adding `Developer` back to `newUserRoles` "re-opens the same path" — the `MJ: Users` privilege-escalation described above. It does not, as of this same release. `@memberjunction/core-entities-server` ships a server-side guard on `MJ: Users` (see the `user-elevation-guard` changeset) that refuses privilege-elevating writes — create, delete, and changing `Type`, `Name`, or another user's row — for **any** caller whose `Type` is not `'Owner'`. The guard is role-blind: it holds no matter what `newUserRoles` names, `Developer`/`Integration` included, and for custom roles this changeset never anticipated. Re-adding `Developer` to `newUserRoles` no longer reopens the `MJ: Users` escalation this changeset describes — it remains inadvisable only for the unrelated, unaddressed data-plane grant above.
```
rkihm-BC: Non-blocking, but this lands in the release notes verbatim: the paragraph describes its own earlier wording rather than the shipped behaviour. Suggest collapsing this and the previous paragraph into one statement — re-adding `Developer` to `newUserRoles` no longer reopens the `MJ: Users` escalation (the `MJUserEntityServer` guard is role-blind), but it remains inadvisable because of the unfiltered data-plane grant tracked in #4283.

rkihm-BC (dismissed):
Read the description, #4260, all 14 commits, the full diff, the logs of the three red jobs and the matching jobs on `next`, and traced the guard's assumptions in `BaseEntity`. Approving; details below.

**The design is the right shape.** Narrowing the default closes reachability; the entity guard closes capability for any role a deployment invents. The RLS rejection is correct — an own-row filter still permits setting one's own `Type`, and MJ has no per-role field permission.

**The invariants hold under adversarial reading.** The three things the guard leans on all check out on `next`: `ActiveUser` falls back to the provider's current user (the Owner-type system user on the server), so the guard's trust boundary is exactly `CheckPermissions`' existing one; a loaded or hydrated row sets `OldValue` on first assignment, so the own-row check has a pre-save ID on every real load path (and the live UI check produced the own-row message, not the fail-closed one); `FixedWidthColumn` already trims the NCHAR padding before `callerIsOwner` folds case. The `ReplayOnly` closure is real — `baseEntity.ts:3725` force-passes validation without calling `Validate()` — and the wire-path enumeration is accurate. Registration follows the `MJListEntityServer` pattern, both bootstrap manifests are regenerated, and the browser-manifest gate is green.

**Test evidence, confirmed from the CI logs rather than the description:** 23 guard tests passed in shard 6, the 5 config-contract tests in shard 3, all 15 explorer-settings test files in shard 2. Integration: UE1 proved the provider builds the guard class; UE3 proved the Owner path stays open; UE4 is correctly mutation-gated.

**The three red checks are `next`'s, not this PR's.** All are byte-identical on `next`'s latest run of the same workflow: Source guards fails on a disabled test in `geo-core` (not in this diff), and the failing shard is six `manage-metadata-drift-hold-revoke` cases plus three `BaseViewEmissionGoldenMaster` cases in CodeGenLib, which this PR does not touch (`next` fails the same tests in its shard 4). Needs a green `next` and a branch update to merge, nothing here to change.

**Two non-blocking notes**

1. `.changeset/new-user-roles-drop-developer.md` narrates its own edit history ("What is *no longer* true: this paragraph originally warned…"). Release notes are generated from changesets, so that will ship verbatim. Worth collapsing the last two paragraphs into one plain statement: re-adding `Developer` no longer reopens the `MJ: Users` escalation because of the guard, but remains inadvisable for the unfiltered data-plane grant tracked in #4283.
2. UE2 skipped in CI — the CI database has no active non-Owner that can read `MJ: Users`, so the deterministic lane's proof of the refusal path came from the unit tests. The skip is loud and correct by the suite's doctrine; seeding one non-Owner reader into the integration database would make UE2 exercise for real. Follow-up.

Also fine: `Email` stays mutable by design (#4286); the old default surviving in a completed plan document is historical; the `SyncRolesUsersResolver` comment honestly records that the soft-delete fallback fails under the new refusal too.

rkihm-BC (approved):
Re-approving on `4be8d33b14`. Both non-blocking notes from the previous review are done, and the push carries nothing else of substance beyond the `next` merge.

- **Changeset** no longer narrates its own drafting history — it now states what the software does: re-adding `Developer` is no longer an escalation risk because the guard is role-blind, and remains inadvisable only for the unfiltered data-plane grant tracked in #4283. Right release-notes text.
- **UE2 now runs for real.** The integration seed grants `Integration Test: RLS Scoped Reader` `CanRead` on `MJ: Users` (create/update/delete stay false), and two active non-Owner seeded users hold that role. The CI log on this head shows UE1, UE2 and UE3 passing with UE4 correctly mutation-gated — the refusal path is now proven in the deterministic lane, not only in unit tests. The fixture-contract test pins the seed so a future metadata change cannot silently send UE2 back to skipping. Integration tier green.

One observation, not a blocker: the changeset commit also adds a self-narration lint to `check-changeset-bump.mjs` with a test. It is small, it is a direct consequence of the review comment, and it affects only the local `check:changeset` run (no CI job invokes that script), but it is scope beyond a security fix. Fine to leave; if it's ever split out, it stands on its own.

21/21 checks green now that `next` is repaired. Approving.

PR #4276, opened 2026-09-07
fix(communication-ms-graph): a stored Azure Service Principal credential cannot drive the provider

JosueGarcia-BC:
## The defect

A credential created from the **Azure Service Principal** type cannot drive a single MS Graph operation.

`metadata/credential-types/schemas/azure-service-principal.schema.json` declares three fields and requires all three:

```json
"required": ["tenantId", "clientId", "clientSecret"]
```

`MSGraphProvider.resolveCredentials` validated **four**:

```ts
validateRequiredCredentials(
    { tenantId, clientId, clientSecret, accountEmail },
    ['tenantId', 'clientId', 'clientSecret', 'accountEmail'],
    'Microsoft Graph'
);
```

There is no fourth field to fill in, so every operation failed with `Missing required credential: accountEmail` before doing any work.

It went unnoticed because `resolveCredentialValue` falls back to `AZURE_ACCOUNT_EMAIL`, so a host that sets that variable never sees it. The failure appears exactly when a credential is meant to stand on its own — which is what the Credentials engine is for, and is now reachable deliberately through `disableEnvironmentFallback: true`.

## Why the guard was wrong, not the schema

Adding `accountEmail` to the schema is the obvious fix and the wrong one. Three things say so:

1. **The provider's own interface already declared it optional** — `MSGraphCredentials.accountEmail?: string`. Only the validator forced it.
2. **A mailbox is not part of a service principal.** The same credential legitimately drives Azure OpenAI and Blob Storage, where a mailbox means nothing. The credential type's own description says exactly that.
3. **Almost nothing needed it.** Every read path already treated it as a last-resort fallback behind `params.Identifier` or `ContextData.Email` — so an operation that named its own mailbox was refused over a field it would never have read.

## The change

`accountEmail` becomes a **default**, resolved per operation rather than demanded up front.

- `resolveCredentials` validates the three authentication fields — what a principal actually is.
- New `resolveMailbox(operation, creds, ...preferred)` takes the first mailbox the request named and falls back to `accountEmail`, so anything specific to a request outranks the deployment default.
- When nothing resolves, it refuses with a message naming the operation and both ways to supply one.

**It refuses rather than returning `undefined`, and that is the load-bearing part.** Every one of the 15 call sites interpolates the result into a Graph path. `undefined` would put the literal string `"undefined"` in the URL and return a 404 that reads like *"message not found"* — a wrong answer wearing the costume of a real one.

This package does not enable `strictNullChecks`, so nothing would have caught that: not the compiler, and not a test that only asserts on success. The 15 sites were found and guarded by hand for that reason, each one named with its own operation.

## Two things I pinned rather than changed

- **`GetMessages` rejects instead of returning a failure result.** It has no `try`/`catch`, and `resolveCredentials` has always thrown from it, so this is pre-existing. The test pins the real contract rather than quietly altering it. That a method typed `Promise<GetMessagesResult>` throws is worth revisiting — but not under a fix for something else.
- **Three `catch` blocks** (`SendSingleMessage`, `ForwardMessage`, `CreateDraft`) discarded the exception's message entirely. Harmless while nothing inside had anything specific to say; the mailbox guard does, so they now include it.

## Verified against a live tenant

Three-field credential, environment fallback disabled — the case that was impossible before:

```
credential fields: tenantId, clientId, clientSecret
environment fallback: DISABLED

1. three fields + explicit Identifier   (before the fix: impossible)
  OK       -> 3 message(s)

2. three fields + no mailbox anywhere   (must refuse, and say how to fix it)
  REFUSED  -> Microsoft Graph: GetMessages needs a mailbox and none was supplied.
              Pass one on the request, or set accountEmail on the credential
              (or AZURE_ACCOUNT_EMAIL) as a default.

3. accountEmail supplied as the default
  OK       -> 3 message(s)

4. default present AND Identifier supplied — the request should win
  OK       -> 3 message(s)
```

Case 4 used `never-used@example.invalid` as the credential default: had it been used instead of the request's mailbox, Graph would have returned a 404 rather than three messages.

Unit tests: 49 passing, 5 new, covering all four cases plus an assertion that a refused call reaches Graph **zero** times rather than requesting a mailbox named `"undefined"`. `tsc` clean. `patch` — no migration, no metadata.

## Open question, deliberately not decided here

Should the schema carry an optional `accountEmail`, so an operator can store a default mailbox alongside the principal? It is convenient, and it is how the mismatch arose.

My view is no: keep a credential a credential. A default mailbox belongs to the connection that uses it, not to a principal shared across services that have no mailbox at all. Adding it would also make the change `minor` and pull a metadata-sync migration at release, which is reason enough to decide it deliberately rather than fold it in here.


packages/Communication/providers/MSGraph/src/MSGraphProvider.ts
```diff
@@ -428,7 +460,8 @@ export class MSGraphProvider extends BaseCommunicationProvider {
             };
 
             // Use email address directly in API path
-            const sendMessagePath: string = `${this.getApiUri()}/${encodeURIComponent(creds.accountEmail)}/messages/${params.MessageID}/reply`;
+            const mailbox = this.resolveMailbox('ReplyToMessage', creds);
```
rkihm-BC, line 463: Follow-up, not blocking: this is one of only two call sites that passes no request-level candidate, so it can only resolve `creds.accountEmail` — which a schema-conformant Azure Service Principal credential cannot carry, and which `disableEnvironmentFallback: true` removes the env source for. `ReplyToMessageParams` does carry `ContextData`, the same field eleven sibling sites read, so `this.resolveMailbox('ReplyToMessage', creds, params.ContextData?.Email as string)` would make Reply consistent with them and reachable on the credential this PR exists to support.

packages/Communication/providers/MSGraph/src/MSGraphProvider.ts
```diff
@@ -598,7 +631,8 @@ export class MSGraphProvider extends BaseCommunicationProvider {
             };
 
             // Use email address directly in API path
-            const sendMessagePath: string = `${this.getApiUri()}/${encodeURIComponent(creds.accountEmail)}/messages/${params.MessageID}/forward`;
+            const mailbox = this.resolveMailbox('ForwardMessage', creds);
```
rkihm-BC, line 634: Follow-up, and the harder half: unlike Reply, `ForwardMessageParams` has no `ContextData` and no identifier at all, so Forward cannot name a mailbox without a change to that params type. That makes it the one operation a conformant three-field credential can never drive with the env fallback disabled — worth weighing against your open question about an optional schema field, since this is the case that decision actually settles.

rkihm-BC (approved):
This is the fix for the mismatch I flagged as a follow-up on #4230, and the reasoning is right where it counts. Approving; one gap worth a follow-up, described below.

**Fixing the guard rather than the schema is the correct call**, and the three reasons given are each independently sufficient: a mailbox is not part of a service principal, the provider's own interface already declared it optional, and every read path already treated it as a last resort behind a request-level value. I checked the third — thirteen of the fifteen call sites do exactly that.

**Making `resolveMailbox` refuse rather than return `undefined` is the load-bearing decision.** In a package without `strictNullChecks`, the alternative is the literal string `undefined` interpolated into a Graph path, returning a 404 that reads like "message not found". The test asserting Graph is called **zero** times on a refusal is the one that proves it, and it is the right assertion to have reached for — a success-only test would have passed either way.

The two pinned pre-existing items are honestly scoped. `GetMessages` really has no `try`/`catch` (I checked: zero), so it has always rejected rather than returning a failure result, and pinning that rather than quietly changing it under a fix for something else is the right instinct. Including the exception message in the three `catch` blocks is right now that the mailbox guard has something specific to say.

### The gap: two operations still can't run on the credential this fixes

Thirteen call sites pass a request-level candidate. Two pass none:

```
463:  this.resolveMailbox('ReplyToMessage', creds)
634:  this.resolveMailbox('ForwardMessage', creds)
```

So those two can only get a mailbox from `creds.accountEmail`, and I went looking for whether a conformant credential can supply one. It can't:

- `azure-service-principal.schema.json` declares three properties. It doesn't set `additionalProperties: false`, so it wouldn't *reject* a stored `accountEmail` — but it declares no such property, which is your own premise that there is no fourth field to fill in.
- No credential-type schema anywhere under `metadata/` mentions `accountEmail` at all, so there's no alternative type that carries a mailbox either.

That leaves `AZURE_ACCOUNT_EMAIL` as the only source — the environment fallback the Credentials engine exists to make unnecessary, and the one `disableEnvironmentFallback: true` turns off. So Reply and Forward remain unusable in exactly the configuration this PR was written to support. Not a regression (they threw before too), but the PR's framing is that a stored principal can now drive operations, and that's true for thirteen and still false for these two. Neither is covered by the new tests, which is why the gap is invisible from the suite.

**Reply is a one-argument fix**: `ReplyToMessageParams` carries `ContextData`, the same field eleven sibling call sites read as `params.ContextData?.Email`. Passing it makes Reply consistent with those and reachable. Inline comment.

**Forward genuinely needs more**: `ForwardMessageParams` has no `ContextData` and no identifier, so it can't name a mailbox without a change to that type. Inline comment.

This also sharpens your open question. Whether the schema should carry an optional `accountEmail` isn't only a convenience — as things stand it's the only way Forward runs on a conformant credential. I agree with your answer (keep a credential a credential; a default mailbox belongs to the connection), which points at the params type rather than the schema. Worth deciding with that fact in view rather than as a preference.

rkihm-BC:
The two follow-ups from my review now have a home: **#4322**. Nothing here is blocked on it — filing it is what I wanted before merge, not a change to this PR.

One issue rather than two, because Reply and Forward share one root cause and one decision. The Reply half is a single argument (`params.ContextData?.Email`, the field eleven sibling call sites already read) and can close on its own; the Forward half needs `ContextData` or an explicit mailbox added to `ForwardMessageParams`, which is the cross-package change your open question actually settles. The issue records the schema evidence — three declared properties, no `accountEmail` anywhere under `metadata/` — so whoever picks it up doesn't have to re-derive why the credential can't supply one.

Approval stands on `179e171989`. Good to merge.

Issue #4083, opened 2026-08-27
mj sync push writes metadata without publishing cache invalidations, so a Redis-backed server serves pre-sync data indefinitely


jordanfanapour:
## Summary

`mj sync push` writes metadata through `BaseEntity.Save()`, which raises cache-invalidation events into **its own process's** `LocalCacheManager`. It never installs a shared storage provider, so in a deployment that uses the Redis provider for cross-server cache coherence those invalidations are never published. Every running server keeps serving the pre-push result sets.

Three things combine to make this permanent rather than transient:

1. **No invalidation is published.** `packages/MetadataSync/src/lib/provider-utils.ts` → `initializeProvider()` sets up the SQL provider only; nothing calls `SetLocalStorageProvider`. `LocalCacheManager.InvalidateEntityCaches()` fires locally and dies with the CLI process.
2. **Cached entries never expire.** `packages/MJCore/src/generic/localCacheManager.ts` ships `defaultTTLMs: 0`, so a cached `RunView` result set has no expiry.
3. **Restarting doesn't help.** Redis is external to the app, so stale entries survive redeploys and pod restarts — a fresh process reads the same stale keys.

The affected read path is the server-cache leg in `providerBase`: for an entity with `TrustServerCacheCompletely = true`, `RunView` can serve from cache with no DB hit. Entities that *also* have `TrackRecordChanges = false` have no `RecordChange` trail, so the differential / `maxUpdatedAt` freshness check isn't available either and the serve-from-cache path returns cached rows unconditionally.

Net effect: **metadata is correct in SQL and stale in every running server, with no TTL and no restart that recovers it.** The only remedy today is clearing Redis by hand.

## How we hit it

An org-scoped configuration record was corrected via `mj sync push` and verified in SQL — `__mj_UpdatedAt` moved and the stored content matched the repo. Nine days later every request was still being served the pre-correction revision, byte-identical to the superseded git commit. Two production pipelines failed outright as a direct result, because the stale record carried guidance that had already been fixed.

A redeploy in the interim did not clear it, which is what made the cause hard to see: the staleness lives in Redis, not in process memory. The subsequent metadata sync also reported `no changes` — correctly, since SQL already matched the repo — so nothing in CI indicated a problem.

This is not confined to one entity type. In our database **208 entities** carry the `TrustServerCacheCompletely = 1` + `TrackRecordChanges = 0` combination:

```sql
SELECT SchemaName, Name FROM __mj.Entity
[...]

PR #4277, opened 2026-09-07
AI Model Research Report — 2026-09-07

AN-BC:
## Summary

Weekly AI model & vendor intelligence pass. The heaviest release week this routine has seen — all three Western frontier labs shipped a new flagship inside 72 hours, and Meta shipped its second proprietary model.

**Counts: 4 new models · 1 new vendor · 1 rate card resolved · 1 deprecation applied · 8 items flagged for human review.**

Full report: [`reports/ai-model-research/2026-09-07-weekly-report.md`](reports/ai-model-research/2026-09-07-weekly-report.md)

> **Branch-name note.** Prior weeks used `claude/ai-model-research-YYYY-MM-DD`. This session's runtime pins its own designated branch and forbids pushing anywhere else, hence `claude/magical-turing-uxei24`. Nothing else about the PR differs.

---

## ✅ Resolved: the Integration Tier gate now runs on this PR

*(This section originally read as a warning that the gate did not exist. It has been overtaken by events — leaving the history because it is why the gate got merged.)*

When this PR opened on 2026-09-07 at 08:20 UTC, §0.4 of the routine prompt asserted that `metadata/**` was already in the Integration Tier's `pull_request` paths filter. It was not: [#4254](https://github.com/MemberJunction/MJ/pull/4254) was still open, so this PR ran only `Check migrations` and CodeQL, with no `mj sync push` anywhere.

**#4254 merged at 15:18 UTC that day**, @cadam11 dispatched the tier by hand against the branch at 15:48, and @sohamdesai-bc merged `next` in on 09-09 — so the filter is now present on this branch and **the Integration Tier fires automatically**. The `Cross-server invalidation rig` is correctly gated to the nightly cron and manual dispatch, so it reports `skipped` here rather than dragging in the pre-existing `next` failure discussed in the comments.

Nothing in the prompt needs changing for §0.4 any more. **§0.5's `primaryKey` line is still stale, though — see the note below.**

---

## New models (4)

| Model | Released | API name | Context / output | Price (in/out per 1M) |
|---|---|---|---|---|
| **Claude Fable 5.1** | 2026-09-01 | `claude-fable-5-1` | 1M / 128K | $10 / $50, **cache read $0.25** |
| **GPT-6 Astra** | 2026-09-03 | `gpt-6-astra` | 1.05M / 128K | $10 / $50 (≤272K input) |
| **Gemini 3.8 Flash** | 2026-09-02 | `gemini-3.8-flash` | 1M / 64K | $0.75 / $3.75 intro |
| **Muse Spark 1.3** | 2026-09-02 | `meta/muse-spark-1.3` | 1,048,576 / not published | $1.25 / $4.25 |

Notes worth a reviewer's eye:

- **Fable 5.1** keeps Fable 5's exact $10/$50 rate card; the change is prompt caching — cache hits and refreshes drop 75% from $1 to $0.25 per 1M (a 0.025x multiplier on base input, versus the 0.1x every other Claude model uses). Anthropic, Amazon Bedrock and OpenRouter routes, all live day-one.
- **GPT-6 Astra** has a long-context surcharge: a request above **272,000 input tokens** is rebilled *in its entirety* at $20 input / $2 cached / $25 cache write / $75 output. Only the short-context tier gets a cost row; the second tier is spelled out in that row's `Comments`. Astra also bills cache writes ($12.50/1M), a departure from earlier OpenAI models. Added with OpenAI direct plus an Azure/Foundry route (gated behind the Limited Access Program); **no Bedrock or OpenRouter row** — AWS had not published a model card and OpenRouter was not listing it at research time.
- **Gemini 3.8 Flash** holds 3.7 Flash's headline rate, so it is a capability upgrade at flat cost. Introductory pricing runs through **2026-12-31**, then $1.50/$7.50 from 2027-01-01 (cached input also doubles).
- **Muse Spark 1.3** closes the "onboard Meta" item flagged since 2026-08-17. Meta is added as a **Model Developer only**, with inference routed through OpenRouter (`OpenRouterLLM`) — so this needs **no new driver class**. Meta publishes no max output length, so `MaxOutputTokens` is deliberately left unset rather than guessed. The cheaper Contributor tier ($0.10/$0.20 per 1M) is **not** recorded: it grants Meta training rights over submitted data, which is a governance call, not a pricing one.

## New vendor (1)

- **Meta** — credential type `API Key`. Model Developer role only, for the reason above.

## Pricing (1)

- **GLM 5.3** gains a Z.AI Inference Provider row (`glm-5.3`, `ZAILLM`, 1M / 128K) and its first cost record at **$1.40 / $4.40** per 1M with $0.26/1M cached input, effective 2026-08-19. This resolves the `pricing TBD` placeholder left when GLM 5.3 was added on 2026-08-24.

## Deprecation (1)

- **Claude Opus 4.1** was retired on the Anthropic API on **2026-08-05** (deprecated 2026-06-05, replacement `claude-opus-4-8`). Our metadata still had the route Active. Applied per the §0.2 recipe exactly:
  - Anthropic **Inference Provider** row → `Status: "Inactive"`
  - Anthropic **cost row** → `Status: "Expired"` **and** `EndedAt: "2026-08-05T00:00:00.000Z"`
  - Anthropic **Model Developer** row left `Active` (attribution, not service)
  - **Amazon Bedrock and OpenRouter rows untouched** — Anthropic's own docs state partner-operated platforms set their own retirement schedules, and Opus 4.1 still runs there
  - model-level `IsActive` was already `false` and was not touched

## `primaryKey` (29 records) — and a second stale line in the routine prompt

[`metadata/CLAUDE.md` rule 1](metadata/CLAUDE.md) says new metadata records should carry a hardcoded `primaryKey` from `uuidgen` so the record gets the same ID in every environment; rule 10 repeats it. Every model added to `.ai-models.json` in recent weeks follows it — Kimi K2.7-Code, GLM-5.3-Flash and Qwen3.8-Flash all carry one on the model *and* each vendor and cost row. The records in the first commit here did not.

The routine prompt says the opposite (*"Do NOT include `primaryKey` or `sync` objects"*), but its own §0.6 gives `metadata/CLAUDE.md` precedence, and the file's practice agrees with the repo doc. Fixed in `caf757010`: `primaryKey` added to the 29 records this branch introduces, purely additive (87 insertions, 0 deletions), no `sync` or `__mj_*` fields anywhere. `uuidgen` isn't installed in the runner, so the UUIDs come from `/proc/sys/kernel/random/uuid` — the kernel generator `uuidgen` uses on Linux — rather than being written by hand; all 29 verified well-formed v4, distinct, and unique within their own entity table.

**Whoever owns the prompt should drop that line**, or every future weekly run reintroduces the gap.

---

## Flagged for human review — not applied

1. **Claude Mythos 5.1** (same $10/$50 rate card and 1M/128K envelope as Fable 5.1) — access is gated to vetted orgs through Anthropic's Cyber Verification and Life Sciences Verification programs, initially US-only. Same category as OpenAI's **Daybreak Red / Daybreak Blue**, flagged the week before. Someone should decide whether MJ models gated-access variants at all, and answer for both at once.
2. **Grok 4.7** — Musk announced a Sept 12 target on Sept 2 (2.1T params). No model ID, rate card or context window published as of Sept 5, and the date has already slipped twice. Separately, **Grok 4.6 is now in Microsoft Foundry** (500K context) — an Azure row is possible once Foundry pricing is confirmed.
3. **The GLM-5.3-Flash promo has now expired** — 2026-09-09 16:00 UTC. This was written as "next week's problem" on the assumption the PR would have merged first; it hasn't, so merging now carries a stale OpenRouter rate ($0.05/$0.1667 against a $0.15/$0.50 list). That row is pre-existing data from the 2026-08-31 PR and not part of this diff, so it is left alone rather than widening the PR — your call whether to fold the expiry in here or leave it for next week.
4. **GPT-6 Astra on Bedrock** — add the row once AWS publishes a model card.
5. **Muse Spark Contributor tier** and a first-party Meta inference route (the latter would need a `MetaLLM` driver class that does not exist).
6. **Four cost rows are exact duplicates sharing one `primaryKey`** — pre-existing. `Claude 4 Opus` (`A727C77A-…`), `Claude 4 Sonnet` (`6C644B8C-…`), `Llama 4 Maverick` (`CCA0A4DC-…`), `Llama 4 Scout` (`DA3A0E18-…`). Sharing an ID means `mj sync push` upserts the same row twice — idempotent, which is why CI has stayed green and why it went unnoticed. Deduping means deleting records, so it wants an owner's nod. The Anthropic pair are also retired on the Anthropic API (2026-06-15) at what look like batch rather than list rates, so fix the prices before expiring the rows.
7. **`GLM 5.3` OpenRouter context window** records `MaxInputTokens: 200000` while Z.AI's docs state 1M. Either OpenRouter caps its route or the original entry was a placeholder — verify before changing.
8. **Still open from prior weeks**: reconcile the `DeepSeek V4 Pro` 2026-08-16 cost record ($0.66/$1.98 matches neither the off-peak nor peak tier); resolve the `GPT 5.6` (Sol) pricing conflict; a coordinated **FLUX family refresh** (BFL shipped FLUX 3 on 2026-07-23; its video/audio outputs need a cost-schema decision first, and our two FLUX records still carry `StartedAt` 2025-10-01); `DeepSeek V4 Pro`/`V4 Flash` description updates; `DeepSeek V4 Flash Vision Experimental` modelling.

## Checked, no change this week

Mistral AI (shipped Mistral OCR 4.1 GA and an Agentic Search layer — neither is an LLM in this inventory; Leanstral 1.5 retires 2026-09-30 and has never been in the inventory), Alibaba/Qwen, Moonshot AI, MiniMax, DeepSeek, Cohere, NVIDIA, Black Forest Labs, Inception Labs, Thinking Machines Lab, Groq, Cerebras, Fireworks.ai. Two vendor near-misses were skipped deliberately: Poolside's `Laguna XS 2.1` (no verified rate card) and OpenCode's anonymous `Omen Alpha` stealth listing.

---

## Verification

**CI.** `mj sync push --dir=metadata` ran against a fresh SQL Server and passed in the manual dispatch of 09-07 — three times over, in the deterministic job (`--ci` plus the optional-metadata seed), the CodeGen drift gate (which also reported no drift), and the rig job's setup. The deterministic suite itself passed. The Integration Tier is now running automatically on the current head as well.

**Local gates**, re-run after the `next` merge and again after the `primaryKey` push:

1. **§0.3 status pre-flight** — `OK`. Every vendor `Status` ∈ {Active, Inactive, Deprecated, Preview}; every cost `Status` ∈ {Active, Pending, Expired, Invalid}, so no `Inactive` on a cost row; every `Expired` row carries an `EndedAt` greater than its `StartedAt`.
2. **Field names vs. the generated ORM** — every field on every touched record exists on `MJAIModelEntity` / `MJAIModelVendorEntity` / `MJAIModelCostEntity` / `MJAIVendorEntity` in `packages/MJCoreEntities/src/generated/entities/__mj.ts`.
3. **Touched records vs. the CHECK constraints in `migrations/`** — `chk_AIModelCost_Status`, `_Currency` (3 chars, uppercase), `_DateRange`, `_ProcessingType`, `_InputPrice`/`_OutputPrice` (≥ 0), `CK_AIModelVendor_Status`, `_Priority` (≥ 0), `_MaxInputTokens`/`_MaxOutputTokens`. All satisfied.

Also verified: every `@lookup:MJ: AI Vendors.Name=…` resolves (including the new `Meta` row); every `PriorVersionID` resolves; both JSON files re-serialize byte-identically, so the diff contains only intended changes; no `sync`, `__mj_CreatedAt` or `__mj_UpdatedAt` on any new record. `npm run check:changeset` passes (`metadata/` touched → `minor` required → `minor` supplied). `npm run check:claude-md` fails with 3 findings, all pre-existing on `next` and untouched here.

Counts after this PR: **193** models in `.ai-models.json` (+4) and **31** vendors (+1).

https://claude.ai/code/session_016zEmSodiDp5irpK79EViLE


AN-BC:
@cadam11 lmk if this was done right after prompt change or not

AN-BC:
Answering the part I can settle directly, since this run is the only party that saw the prompt text.

**Yes — this ran on the post-change prompt.** The mirror was checked in as be40db489 on 2026-09-05 ("docs: check in the AI-model research routine prompt, with the cost-status rule"); this run fired 2026-09-07. The prompt the Routine delivered carried all of §0.1–§0.6, and I spot-checked distinctive strings from the delivered text against the checked-in mirror — `Do not copy the Kimi K2.5 shape`, ``chk_AIModelCost_DateRange` requires `EndedAt > StartedAt``, ``It must print `OK` ``, ``Anything under `metadata/` is a database change``, `GLM-4.7 on Cerebras` — all present. **The two copies agree; the mirror is in sync.**

§0 also did its job. The one deprecation in this PR (Claude Opus 4.1, retired on the Anthropic API 2026-08-05) follows the §0.2 recipe: vendor row `Inactive`, cost row `Expired` **with** `EndedAt`, Model Developer row left `Active`, Bedrock and OpenRouter untouched. No `Inactive` landed on a cost row — the §0.3 pre-flight was run after every edit and prints `OK`.

**The one thing that is stale is §0.4, and it's stale by anticipation rather than by drift.** The change-log row for 2026-09-05 records adding "the note that #4254 makes the Integration Tier run on metadata PRs" — but #4254 is still open, so that has not happened yet. On current `next`, `integration.yml` lists `metadata-optional/**` and not `metadata/**`, and no other workflow filters on `metadata/**`. This PR therefore ran only `Check migrations` and CodeQL, with no `mj sync push` anywhere in it. The prompt was written for the world in which #4254 had landed.

So: prompt current, mirror in sync, gate not yet real. Merging #4254 makes §0.4 true and closes the hole that let #4030 and #4110 through. In the meantime I substituted two local gates for the missing one — field names checked against the generated entity classes, and every touched record checked against the actual CHECK constraints in `migrations/` — both green, detail in the PR description.

---

AN-BC:
@cadam11 thanks for dispatching the Integration Tier by hand — that was the right move given the paths filter doesn't fire on `metadata/**` yet. Reporting what it found.

## The gate you were after is green on this diff

`mj sync push --dir=metadata` ran against a **fresh SQL Server** three separate times in [run 34140212954](https://github.com/MemberJunction/MJ/actions/runs/34140212954) and passed every time:

| Job | Step | Result |
|---|---|---|
| Integration (SQL Server, deterministic) | 12 · `mj sync push --ci` | ✅ |
| Integration (SQL Server, deterministic) | 13 · optional metadata + RLS fixtures | ✅ |
| CodeGen drift gate (nightly) | 11 · Migrate + sync push | ✅ |
| CodeGen drift gate (nightly) | 12 · CodeGen, then drift diff | ✅ no drift |
| Cross-server invalidation rig | 12 · Migrate + sync push | ✅ |

So the metadata in this PR validates and pushes cleanly. That is exactly the check §0.4 assumes exists, and it passes. The deterministic job is still on step 16 (`mj test`) as I write this; every metadata step ahead of it is already green, and I'll report the suite result when it lands.

## The one red job is not this PR's

**`Cross-server invalidation rig (nightly)`** ([job 101800275670](https://github.com/MemberJunction/MJ/actions/runs/34140212954/job/101800275670)) failed at step 14, "Run cross-server invalidation rig". It ran here because its `if:` at `integration.yml:454` includes `github.event_name == 'workflow_dispatch'`, so a manual dispatch pulls in the nightly-only job alongside the PR-relevant ones.

**It is red on the base branch, at the same step, from this morning.** The 06:26 nightly on `next` ([run 34090826964](https://github.com/MemberJunction/MJ/actions/runs/34090826964)) ran against `cda01874e` — the exact commit this branch is based on, with none of my changes — and:

- `Integration (SQL Server, deterministic)` → ✅
- `CodeGen drift gate (nightly)` → ✅
- `Cross-server invalidation rig (nightly)` → ❌ **same step 14**

It has failed on every nightly `next` run from 2026-08-29 through 2026-09-07 — ten consecutive. (The two green runs inside that window, Aug 30 and Sep 6, are the Sunday 06:45 weekly live-model lane, a different job set.)

The proximate symptom is in the Redis service-container log, immediately before the rig step fails:

```
1:M 07 Sep 2026 15:54:49.130 # Client id=6 ... cmd=subscribe lib-name=ioredis lib-ver=5.10.0
    omem=37748784 tot-mem=37786600 scheduled to be closed ASAP for overcoming of output buffer limits.
```

Redis is dropping the rig's own pub/sub subscriber for exceeding the ~32 MB client output buffer limit — it had ~36 MB queued. This PR's diff is metadata JSON, report Markdown and a changeset; it touches no Redis, pub/sub, or cache code, and the ~27 records it adds cannot move a buffer from under-limit to 36 MB over.

## No fix to port, and I'm not spending the re-run

I looked for an existing fix to pull in and there isn't one. [#4083](https://github.com/MemberJunction/MJ/issues/4083) is the nearest thing — it documents `mj sync push` never publishing cache invalidations to a Redis-backed server, with a proposed patch to `provider-utils.ts` — but it is an open *issue* with no PR behind it, and it describes a related-but-distinct gap rather than this rig's buffer overflow. Nothing exists that I could port and expect to turn this green.

I'm also **not** re-running the job. A re-run exists to distinguish a flake from a real failure, and the base-branch run from this morning already settles that directly: same job, same step, same day, without my diff. Re-running a failure that has been consistent for ten nights would burn ~7 minutes and tell us nothing. Say the word if you want it spent anyway.

**Bottom line:** the rig is a pre-existing `next` failure that deserves its own issue and fix, not a blocker on this PR. Everything this PR is actually accountable for is green.

---

cadam11:
> @cadam11 lmk if this was done right after prompt change or not

yep, lgtm

AN-BC:
Closing the loop on the suite result I said I'd report.

**`Integration (SQL Server, deterministic)` → ✅ success** ([job 101800275828](https://github.com/MemberJunction/MJ/actions/runs/34140212954/job/101800275828), completed 15:57:32). Step 16 `mj test` passed; the mutation-gated checks report an honest `Skipped` on this lane, as the workflow header says they should.

Final state of the dispatched run:

| Job | Result |
|---|---|
| Integration (SQL Server, deterministic) | ✅ |
| CodeGen drift gate (nightly) | ✅ no drift |
| Integration (Live Model, weekly) | skipped (weekly lane) |
| Cross-server invalidation rig (nightly) | ❌ pre-existing on `next` — see above |

Everything this PR is accountable for is green, and the rig is the ten-night `next` failure detailed in my previous comment. Nothing further from me unless you want that re-run spent or something changes.

---

AN-BC:
@sohamdesai-bc thanks for merging `next` in. I re-verified the merged head and then fixed a convention violation of my own that the merge brought into view. Pushed `caf757010`.

## Fixed: the new records were missing `primaryKey`

[`metadata/CLAUDE.md` rule 1](metadata/CLAUDE.md) says new metadata records **should** carry a hardcoded `primaryKey` generated at the CLI with `uuidgen`, so a record gets the same ID in every environment; rule 10 repeats it. Every model added to `.ai-models.json` in recent weeks follows it:

| Model | model `primaryKey` | vendor rows with `primaryKey` |
|---|---|---|
| Kimi K2.7-Code | ✅ | 3/3 |
| GLM-5.3-Flash | ✅ | 4/4 |
| Qwen3.8-Flash | ✅ | 4/4 |
| **my four, before this push** | ❌ | 0/14 |

The routine prompt told me the opposite — *"Do NOT include `primaryKey` or `sync` objects — these are auto-generated by mj-sync"* — but §0.6 of that same prompt says `metadata/CLAUDE.md` wins when the two disagree, and here the repo doc and the file's own established practice agree with each other against the prompt. **This is a second stale spot in the routine prompt, alongside §0.4.** Whoever reconciles §0.4 should fix this line at the same time; without it, every future weekly run reintroduces the same gap.

`primaryKey` added to the 29 records this branch introduces — the four new models with their vendor and cost rows, the Z.AI inference row and cost row on GLM 5.3, and the Meta vendor. No `sync` block or `__mj_CreatedAt`/`__mj_UpdatedAt` anywhere; those stay auto-managed at release time, per the same rule. Purely additive: **87 insertions, 0 deletions.**

`uuidgen` isn't installed in this container, so rather than writing UUIDs by hand — which rule 1 explicitly forbids — I drew from `/proc/sys/kernel/random/uuid`, the kernel generator `uuidgen` itself uses on Linux. All 29 verified well-formed v4, distinct, and unique within their own entity table.

## Merge verification

The merge of `next` came through clean. On the merged head, before my push: 193 models / 31 vendors (unchanged), no duplicate names, all four new models and the Meta vendor intact, `Claude Opus 4.1` still `Inactive` + `Expired`/`EndedAt`, GLM 5.3's Z.AI row and cost row still present. Status pre-flight `OK`, field names still match the generated ORM, CHECK-constraint check `OK`, both files re-serialize byte-identically. All re-run after the `primaryKey` push, still green.

## Two things for you, neither touched

**1. The GLM-5.3-Flash promo has now actually expired** — 2026-09-09 16:00 UTC, a few hours ago. Flag 3 in the PR body assigned it to next week's run on the assumption this PR would have merged by then. It hasn't, so if it merges now it carries a knowingly-stale OpenRouter rate ($0.05/$0.1667 against a list rate of $0.15/$0.50). That row is pre-existing data from the 2026-08-31 PR, not part of my diff, so I've left it alone rather than widen this PR — but it's your call whether to fold the expiry in here or leave it for next week.

**2. Four cost rows are exact duplicates sharing one `primaryKey`** — pre-existing, and it sharpens flag 6 in the PR body. These are the same record written twice, not different prices:

| Model | Vendor | Duplicated `primaryKey` |
|---|---|---|
| Claude 4 Opus | Anthropic | `A727C77A-6F00-4990-9E1D-67B95F49CCF5` |
| Claude 4 Sonnet | Anthropic | `6C644B8C-CB68-461D-A00B-36F62855539B` |
| Llama 4 Maverick | Groq | `CCA0A4DC-497D-4E11-BEBB-392E0074A7CA` |
| Llama 4 Scout | Groq | `DA3A0E18-83DD-4329-AC87-0D8E9D563F01` |

Sharing an ID means `mj sync push` upserts the same row twice — idempotent, which is why CI has stayed green and why this went unnoticed. Deduping means deleting records, so it wants an owner's nod rather than a research PR doing it silently.

Note that CI on this new head will again be only `Check migrations` + CodeQL, since `metadata/**` still isn't in the Integration Tier's paths filter. Happy to have the tier dispatched again against `caf757010` if you'd like the `mj sync push` confirmation on the final diff.

---

AN-BC:
## `CodeGen drift gate` is red, and it will be red for every metadata PR

**Failing:** `CodeGen drift gate` → step 14 *Diff committed generated artifacts* ([job 102638517523](https://github.com/MemberJunction/MJ/actions/runs/34402528190/job/102638517523)) on `caf757010`.

```
 M metadata/ai-models/.ai-models.json
 M metadata/ai-vendors/.ai-vendors.json
```

Everything before it passed: **`Migrate + sync push` ✅**, `Run CodeGen against the migrated DB` ✅, `Verify CodeGen idempotency` ✅. So the metadata is valid and CodeGen is clean — the failure is the diff step alone.

### Why it fails

Step 11 runs `mj sync push`. Per [`metadata/CLAUDE.md` rule 1](metadata/CLAUDE.md), *"the `sync` blocks will be automatically added/updated when `mj sync push` runs"* — that is the tool's documented job. Rule 1b then says PRs must contribute **"ONLY the declarative metadata JSON changes … no `sync`"**, because the build engineer regenerates those at release time.

Count in the file right now:

| | models | vendors |
|---|---|---|
| carry a `sync` block | 189 | 30 |
| do not | **4** | **1** |

The 4 are `Claude Fable 5.1`, `GPT-6 Astra`, `Gemini 3.8 Flash`, `Muse Spark 1.3`; the 1 is `Meta` — exactly the records this PR adds. The push writes their `sync` blocks, the files become modified, and step 14's `git status --porcelain -- packages/ metadata/ migrations/` reports drift.

Its cleanup restores `pnpm-lock.yaml` and sweeps untracked `CodeGen_Run_*.sql`, but nothing accounts for `metadata/`. **So the two rules are in direct tension for any PR that adds a metadata record:** committing the `sync` blocks would satisfy the gate and violate rule 1b, and omitting them satisfies rule 1b and fails the gate.

This is not specific to my diff. **Every future weekly research PR adds records and will fail here the same way**, which is why I'm raising it rather than working around it.

*(One caveat stated plainly: the log prints only ` M <file>`, not the intra-file diff, so "the delta is the `sync` blocks" is inference — a strong one, from the 189/4 split and documented push behaviour, but not something the log shows directly. Having the step print `git diff` alongside the filenames would make this self-evident next time.)*

### Why I haven't pushed a fix

The fix belongs in `integration.yml`, which this PR doesn't touch. Folding a CI change into a metadata research PR widens it past what it was reviewed for, and it would collide with the gate work you've been landing in #4325 / #4254. No existing PR fixes this, so per the standing rules I'm proposing a patch rather than widening.

I'm also not spending a re-run: this is deterministic, not a flake.

### Proposed patch

Restore files whose **only** change is `sync` blocks, leaving every other metadata mutation failing. Goes in the `Diff committed generated artifacts` step, next to the `pnpm-lock.yaml` restore (both copies — the PR gate at ~L696 and the weekly AI-assisted lane at ~L817):

```bash
# `mj sync push` above writes a `sync` block (lastModified + checksum) into every record it
# pushes — documented behaviour (metadata/CLAUDE.md rule 1) — while rule 1b forbids PRs from
# committing those blocks, since the build engineer regenerates them at release. So any PR
# that ADDS a metadata record dirties metadata/ here through no fault of its own. Restore
# files whose ONLY delta is sync blocks; real metadata drift still fails the gate.
for f in $(git diff --name-only -- metadata/); do
  node -e '
    const {execSync}=require("child_process"), fs=require("fs");
    const strip=o=>Array.isArray(o)?o.map(strip)
      :(o&&typeof o==="object"
          ?Object.fromEntries(Object.entries(o).filter(([k])=>k!=="sync").map(([k,v])=>[k,strip(v)]))
          :o);
    const f=process.argv[1];
    const now=strip(JSON.parse(fs.readFileSync(f,"utf8")));
    const was=strip(JSON.parse(execSync("git show HEAD:"+f,{encoding:"utf8",maxBuffer:1<<28})));
    if(JSON.stringify(now)===JSON.stringify(was)) execSync("git checkout -- "+f);
  ' "$f"
done
```

I ran it both ways in a throwaway repo before proposing it:

- **sync-blocks-only change** (a new record gaining a block, existing ones refreshed) → file restored, `git status` clean, gate passes ✅
- **negative control** — same sync churn *plus* a genuinely renamed field → file stays dirty, gate correctly still fails ✅

If you'd rather not carry that much shell, dropping `metadata/` from the status scope also unblocks it, but it gives up detecting real metadata round-trip drift, which seems worth keeping.

### Where that leaves this PR

Green except this gate, which it cannot pass on its own: `Check migrations` ✅, CodeQL ✅, `Integration (SQL Server, deterministic)` ✅ on the 09-07 dispatch and re-running now, `Cross-server invalidation rig` correctly `skipped`, no merge conflict. Once the gate accounts for `sync` blocks I expect this to go green with no change to the diff. Happy to apply the patch here instead if you'd prefer that to a separate PR — just say so.

---

AN-BC:
@sohamdesai-bc both of those are good catches, and your drift-gate fix is better than the patch I proposed.

**On `ff7018652` — you're right, and the error was mine.** I wrote that Comments text off Anthropic's deprecation page, which says Opus 4.1 survives "except on Bedrock and Google Cloud". That describes *Anthropic's* platforms, not our metadata. Checked against the record: the four vendor rows are Anthropic (Model Developer), Anthropic (Inference Provider, now Inactive), Amazon Bedrock and OpenRouter — no Vertex route at all. So naming Google Cloud both misdescribed the row and omitted the OpenRouter route we actually have, and "those vendor and cost rows" overstated it since only Bedrock carries a cost record. That text ships into `MJ: AI Model Costs`, so it would have misled exactly the person trying to work out where a retired model is still reachable.

I'd carried the same conflation one step removed in the report's Deprecated/Sunset section — right two rows named, but justified by quoting the page's partner-platform phrasing, so "remains available there" pointed at Google Cloud instead of OpenRouter. Fixed in `7d1363a4c`, with the distinction spelled out so a future run reading this report doesn't copy the page's wording back into a record. Report text only; no metadata change.

**On `e4a788b3b` — yours is the better fix and I'd rather have it than mine.** Three reasons:

- **Right place.** Discarding the write-back in the setup step, where it happens, beats reasoning about it later in the diff step. The push is the gate's setup, not its system-under-test — that framing is cleaner than my "restore files whose only delta is sync blocks".
- **The coverage objection I raised doesn't hold.** I worried a blanket restore would give up detecting real metadata drift. `git checkout -- metadata/` restores *tracked* files only, so anything CodeGen newly writes there still surfaces as untracked in the drift status — and you verified CodeGen writes nothing under `metadata/` in this job anyway. Your untracked-file control is precisely the case my four-way test didn't cover.
- **You found the actual regression point.** `197fdf8376` putting `metadata/` in the drift scope while promoting the gate to blocking, failing every metadata-adding PR since 2026-09-08 — more precise than my dating, and it makes the fix auditable.

One line and a comment beats an embedded Node script per file. Nothing further needed from me on that.

**State after all three commits** (head `7d1363a4c`), re-verified locally: status pre-flight `OK`; 193 models / 31 vendors; the four new models still carry their `primaryKey`s and are still the only records without a `sync` block, which is what rule 1b wants; both JSON files re-serialize byte-identically. The Integration Tier will re-run on this head — I'll report it and act if anything goes red.

Still open for a human, unchanged: whether to fold in the now-expired GLM-5.3-Flash promo row (§3), the four duplicate cost-row `primaryKey`s (§6), and dropping the `primaryKey` line from the routine prompt so next week's run doesn't reintroduce that gap.

---

AN-BC:
@sohamdesai-bc `4e9d751a3` finds something neither of us had — and it means **I withdraw the concession I made two comments ago.** I told you the blanket restore was better than my strip-and-compare and that "the coverage objection I raised doesn't hold." That was wrong, and I'd rather say so than leave it standing.

The honest version:

- My **approach** was right — restore only when stripping `sync` makes the file identical to HEAD.
- My **stated reason** for it was wrong. I justified it by worrying about CodeGen writing under `metadata/`, and you correctly showed CodeGen writes nothing there in that job. Having only that argument, I concluded the objection was baseless and folded.
- The real reason is the one you found: **PushService mints a `primaryKey` for any record that arrives without one, outside the `writeSyncMetadata` guard.** A blanket revert masks that, and no other gate covers it — `check-migration-id-determinism.sh` is scoped to Flyway files.

So the objection held; I just couldn't name why, and abandoned it on the strength of a bad argument rather than checking what else the push writes. The lesson is mine to keep: I should have read `PushService` before conceding.

Worth spelling out how close that came to biting this PR. The push minting IDs is precisely the rule-1 violation my first commit had — the four models and Meta vendor landed with no `primaryKey`, and `caf757010` fixed it with `uuidgen`-sourced UUIDs. Under the blanket restore, a future run that *didn't* catch it would have been waved through with IDs that differ in every environment, and the dirty tree that reported it would have been swallowed. Printing the offending delta instead of just the filename is the other half of that — it's what I asked for and it closes the inference gap I flagged.

Verified on `4e9d751a3`: pre-flight `OK`; the four new models all carry `primaryKey`, as do all 22 of their vendor and cost rows; the Meta vendor too. So the narrowed gate should see sync-only churn on this PR and restore it, which is the case your first control covers.

Nothing outstanding from me. The three items still needing a human are unchanged: the expired GLM-5.3-Flash promo row, the four duplicate cost-row `primaryKey`s, and dropping the `primaryKey` line from the routine prompt — that last one matters more now, since a run following the prompt literally produces exactly the records this gate is built to reject.

---

AN-BC:
@sohamdesai-bc since `36d0e4456` makes the restore script the thing standing between this PR and green, I exercised its exported logic against **this PR's actual record shape** rather than just reading it. Eight checks, all passing:

| Case | Expected | Result |
|---|---|---|
| `sync` block added to a new record | sync-only → restore | ✅ |
| `sync` block refreshed on an existing record | sync-only → restore | ✅ |
| **minted `primaryKey`** | NOT sync-only → stay dirty | ✅ |
| field mutation hidden behind sync churn | NOT sync-only → stay dirty | ✅ |
| **`sync` on nested vendor/cost child rows** | sync-only → restore | ✅ |
| field change on a nested child row | NOT sync-only → stay dirty | ✅ |
| unparseable JSON | throws, so caller keeps it dirty | ✅ |
| `stripSync` leaves non-`sync` keys intact | — | ✅ |

The nested one is the case that actually decides this PR, and it's worth calling out because a plain top-level strip would have looked correct in a simple fixture and still failed here. The push writes `sync` blocks at **three levels** on my records — the model, its vendor rows, and its cost rows (22 child rows across the four models). `stripSync` recursing through `relatedEntities` is what makes the restore fire instead of leaving the file dirty. It does recurse; verified above.

I couldn't run your vitest suite itself — `node_modules` isn't installed in this runner and a full `pnpm install` on the monorepo isn't worth the disk here, and `npx` just pulls an unrelated vitest 5.0.0 that can't resolve `vitest/config`. So the above exercises the same exported functions directly through plain Node, which is a different route to the same seam, not a substitute for your ten tests.

Confirming your own note rather than reporting it as news: `test:scripts` is defined in `package.json` at line 111 and **no workflow references it** — I grepped `.github/workflows/`. So the script guarding every future metadata PR currently has tests that CI never runs. Your call whether that gets wired into the Unit Tests workflow's `Source guards` lane or left for a follow-up; I'm not touching workflow files from this PR.

Current head `36d0e4456`: `Check migrations` ✅, `Select analyses` ✅, everything else still running, nothing red, no conflict.

---

AN-BC:
## ✅ Fully green on `36d0e4456` — including the drift gate

The one I've been promising a result for: **`CodeGen drift gate` → success** (22:08:21). Soham's narrowed restore did exactly what it was built to do — recognised the setup push's `sync` write-back as sync-only, restored it, and left the gate clean without giving up the minted-`primaryKey` detection. First fully clean run on this PR.

| Check | Result |
|---|---|
| **CodeGen drift gate** | ✅ |
| **Integration (SQL Server, deterministic)** | ✅ |
| Run unit tests | ✅ |
| Build · Source guards · Native-ESM import guard · Clean up run cache | ✅ |
| Check migrations · Detect PR changes | ✅ |
| Select analyses · Analyze js-core / python / actions | ✅ |
| CodeQL | ✅ neutral |
| Unit test shards | skipped — `has_tests: false`, correct for a metadata-only PR |
| Cross-server rig · weekly lanes · nightly coverage | skipped, correctly gated |

**19 checks, zero failures.** No merge conflict. `mergeable_state: blocked` is the required human approval, nothing more.

On the two red `Run unit tests` marks still visible against earlier commits — both were concurrency cancellations, not defects. Each was a run superseded mid-flight by the next push (five landed in about 35 minutes), and the aggregator counts a `cancelled` lane as red. On `4e9d751a3` the lanes read `guards: success, build: success, test: skipped, esm-guard: cancelled` — everything that actually executed passed. Both are on superseded commits; the current head's run completed cleanly, which is the same workflow reaching the end without interruption.

### Still needing a human, unchanged

1. **The GLM-5.3-Flash promo expired at 16:00 UTC today.** Merging now carries a stale OpenRouter rate ($0.05/$0.1667 against a $0.15/$0.50 list). Pre-existing data, not this diff — fold it in here or leave it for next week, your call.
2. **Four duplicate cost-row `primaryKey`s** (`Claude 4 Opus`, `Claude 4 Sonnet`, `Llama 4 Maverick`, `Llama 4 Scout`). Deduping deletes records, so it needs an owner's nod.
3. **Drop the `primaryKey` line from the routine prompt.** This one has grown teeth: §0.5 tells the weekly run *not* to include `primaryKey`, and `4e9d751a3` established that the push mints one when it's absent — which the drift gate now correctly reds. A run following the prompt literally will produce exactly the records this gate rejects, every week, until that line goes.

Also worth a decision at some point, though not blocking: `pnpm test:scripts` covers the restore script that now guards every metadata PR, and no workflow invokes it.

Nothing further from me unless something changes. Thanks both for the fast turnaround on the gate.

---

SDesai-BC:
## All checks green on `36d0e4456f` — ready to merge pending review

`CodeGen drift gate` **passes**. 13/13 checks green, no failures.

### The gate fix, and a correction to my own first attempt

The gate was red because its `Migrate + sync push` **setup** step writes `sync` blocks back into every record it pushes — documented behaviour (rule 1) that rule 1b forbids a PR from committing. Any PR that *adds* a metadata record was failed for it. That entered the drift scope in 197fdf8376, which also promoted the gate to blocking, so it had failed every metadata-adding PR since 2026-09-08.

My first fix reverted `metadata/` wholesale. **That was too broad**, and an adversarial review caught it: `sync` is not the only thing the push writes back. `PushService` mints a `primaryKey` for any record that arrived without one, *outside* the `writeSyncMetadata` guard — so a blanket revert would have silently green-lit the exact rule-1 violation this PR was itself pulled up on. Now narrowed: a file is restored only when stripping `sync` keys makes it identical to `HEAD`.

For the record, `push.writeSyncMetadata: false` is **not** a safe alternative — it is read from `entityConfig`, and that committed config is what the release push uses, which must write sync blocks.

### The earlier inference is now proven

The open caveat was that the log printed only ` M <file>`, so "the delta is the `sync` blocks" was inference. The restore compares structurally and logged:

```
restored (sync blocks only): metadata/ai-models/.ai-models.json
restored (sync blocks only): metadata/ai-vendors/.ai-vendors.json
2 file(s) restored, 0 left dirty for the drift step.
```

Then: `No drift — committed generated artifacts match a clean-DB CodeGen run.` Delta confirmed sync-only. The step also now prints the offending diff when it *can't* restore, so this never has to be inferred again.

### Independent verification of the metadata

An adversarial review validated all 29 new records against every CHECK constraint (`chk_AIModelCost_Status`/`Currency`/`DateRange`/`ProcessingType`, price ≥0, `CK_AIModelVendor_Status`), confirmed no `UQ_AIModelVendor_ModelID_VendorID_TypeID` collision, every `@lookup` resolving in `directoryOrder` dependency order, all `PriorVersionID` targets existing, the Opus 4.1 deprecation matching the rule-4 recipe field for field, and the `minor` changeset being correct. Verdict: metadata is clean.

Also fixed here: the Opus 4.1 cost-row `Comments` named "Google Cloud" as a surviving route. There is no Google Cloud/Vertex row for Opus 4.1 — the survivors are Amazon Bedrock and OpenRouter, and only Bedrock carries a cost row.

### One correction to an earlier comment

> CI on this new head will again be only `Check migrations` + CodeQL, since `metadata/**` still isn't in the Integration Tier's paths filter.

Not so — `metadata/**` entered the paths filter on 2026-09-05 in `328a98da21` (#4254 merged). The Integration Tier has been running automatically on this PR and passes (`mj sync push` against a fresh SQL Server, 8m28s). No hand-dispatch needed.

### Two things for a reviewer to decide

1. **Scope.** This PR now carries a workflow change, `scripts/ci-restore-sync-only-metadata.mjs`, and its test on top of the metadata research. That is well beyond a weekly research pass. Happy to lift all three onto their own branch if the gate owner would rather review them separately — the metadata commits are cleanly separable.
2. **The new tests are not run by CI.** `pnpm test:scripts` is invoked by no workflow. Related and pre-existing: that suite currently has **4 failures in `decision-metadata-export.test.ts`** (`buildFieldLookup is not a function`, imported at `scripts/codegen-decision-metadata-export.mjs:18`), broken on `next` today and invisible for the same reason. Worth its own issue.

Deliberately **not** touched, both pre-existing on `next` and outside this diff: the expired GLM-5.3-Flash OpenRouter promo row, and the duplicate cost rows (which are messier than first reported — Claude 4 Opus has three identical-price `Batch` rows under two IDs, and no list-price row at all).

Everything this PR is accountable for is green.

AN-BC:
Your correction is right and my sentence was wrong. When I wrote *"`metadata/**` still isn't in the Integration Tier's paths filter"* alongside `caf757010`, `328a98da2` had already put it there — I've confirmed that commit locally, and it does exactly what you say. I'd verified the filter's absence on 09-07 and then repeated it on 09-09 **without re-checking after your `next` merge landed it on this branch**. Same failure mode I'd flagged twice in this thread — carrying a once-verified fact past the event that invalidated it — and I did it anyway. I corrected the PR body once the tier started running, but the comment still carries the bad sentence; this is the correction on the record.

One detail only because someone will later ask when the gate went live: `328a98da2` is dated 09-05 on #4254's branch, but it reached `next` when that PR **merged on 2026-09-07 at 15:18 UTC**. That's the date the filter became effective for anyone else's PR.

**Your restore log closes the caveat I left open.** I flagged "the delta is the `sync` blocks" as inference because the log printed only ` M <file>`. `2 file(s) restored, 0 left dirty` settles it structurally. Printing the diff when it *can't* restore is the durable half of that.

### On flag 6 — you're right, and my version understated it

I checked the records after reading your note. It's worse than I described on two counts:

| | I reported | Actually |
|---|---|---|
| ID sharing | three rows sharing one `primaryKey` | **two** share `A727C77A-…`, a third has its own `F6C1FDB8-…` |
| Rates | "what look like batch rates" | explicitly `ProcessingType: "Batch"` — not an inference |
| List price | didn't mention | **no `Realtime` row exists at all** |

Same shape on `Claude 4 Sonnet` (`6C644B8C-…` twice, plus `5DDA609E-…`). So the real defect isn't duplication — it's that these models have *only* batch pricing recorded and no list-price row, which means anything costing a realtime call against them reads the batch rate. Worth carrying into whatever issue this becomes.

### On scope — one ordering constraint before you decide

The split is clean by path: your three CI commits (`e4a788b3b`, `4e9d751a3`, `36d0e4456`) touch no metadata, and the four metadata commits touch no CI. But **lifting the CI work out puts this PR back to red** — without the restore script the drift gate fails on the setup push's write-back, which is exactly why it failed on `caf757010`. So the split only works if the CI branch merges *first*, or the two land together.

For what it's worth, given `DEPLOYMENT.md` Step 2 merges this under release pressure, I'd rather the research PR stayed thin — but that's the gate owner's call, not mine, and I'd rather it be made knowing the ordering than discover it after splitting.

The 4 pre-existing `decision-metadata-export.test.ts` failures are a genuine find, and a good argument for wiring `test:scripts` into a workflow — a suite nothing runs is where that kind of breakage lives.

---

rkihm-BC (approved):
LGTM

PR #4278, opened 2026-09-07
feat(geo,sync): writable-geo write path, isolated parallel providers, skipGeoCoding

AN-BC:
## Summary

`SupportsGeoCoding` means **maps, distance, and other geo read features**. `GeoCodeSyncService` is a separate **write** path: it only runs when the entity has **1+ writable** (`!IsVirtual && AllowUpdateAPI`) Geo* fields. Virtual `PrimaryAddressLatitude` / `__mj_Latitude` / `__mj_Latitude_{FK}` never invoke the provider.

- Native lat/lng already set (sample data, pasted coords) → do not call the geocoding API.
- CodeGen joins `vwRecordGeoCodes` only for write-source entities without native lat/lng.
- EmbeddedRecord FKs bubble as `__mj_Latitude_{ForeignKeyField}` (e.g. `__mj_Latitude_ShipToAddressID`).
- Map field order: writable native → `PrimaryAddressLatitude` → `__mj_Latitude_{FK}` → `__mj_Latitude`.
- `mj-sync` `.mj-sync.json` `push.skipGeoCoding` is **per entity**.
- Parallel default stays **10**. Each record uses `CreateIndependentInstance()` (shared pool, own TX) — same pattern as MJAPI per-request providers. SQL Server and PostgreSQL both implement it. `ReleaseIndependentInstance()` does not close the pool.
- Durable AfterCreate without a queue submitter **defers until `TransactionDepth === 0`** (fire-and-forget). It must not nest in the caller’s `EntityTransactionScope`.

### Stack

- MJ: https://github.com/MemberJunction/MJ/pull/4278
- Common: https://github.com/MemberJunction/bizapps-common/pull/131
- Cheese: https://github.com/MemberJunction/more-cheese/pull/33

### Anti-patterns we are **not** doing (and why)

| Anti-pattern | Why it is wrong |
|---|---|
| Hand-authored `RecordGeoCode` JSON + SHA in sample data | Hash is an internal dirty check, not an import contract. Person should not own a geo row; Address stores lat/lng. |
| Global CLI `--skip-geocode` as the only control | Display vs source differs per entity. Use `push.skipGeoCoding` on People/Org; Addresses with coords already skip the API. |
| Treating Person/Org primary address as `EmbeddedRecord` | That API is owner-held FK (`Order.ShipToAddressID`). Person uses AddressLink + layered view → `PrimaryAddressLatitude`. |
| Default `--parallel-batch-size 1` | Penalizes every push. The bug was one shared provider; fork the provider instead. |

## Test plan

- [x] MJCore `geoFields.test.ts` (11)
- [x] GenericDatabaseProvider `skip-side-effects.test.ts` (writable / virtual / coords-set / SkipGeoCoding)
- [x] MetadataSync `skip-geocoding-config.test.ts`
- [x] Actions Durable AfterCreate → `DEFERRED_LOCAL`
- [x] Live Google geocode IT (`GOOGLE_GEOCODING_API_KEY` in MJ `.env`; warns and skips if missing)
- [x] `tsc --noEmit` on CodeGen, SQL Server, PG, MetadataSync, Actions, map-view, entity-viewer

## Docs

- `plans/geo-features/geo-features.md`
- `packages/MetadataSync/README.md`
- `guides/TRANSACTIONS_AND_BATCHING_GUIDE.md`

### DIFF
--- a/packages/Actions/Engine/src/entity-actions/EntityActionInvocationTypes.ts
+++ b/packages/Actions/Engine/src/entity-actions/EntityActionInvocationTypes.ts
@@ -1,6 +1,6 @@
 import { MJGlobal, MJLruCache, RegisterClass, SafeJSONParse, UUIDsEqual } from "@memberjunction/global";
 import { MJActionFilterEntity, MJActionParamEntity, MJEntityActionParamEntity } from "@memberjunction/core-entities";
-import { BaseEntity, LogError, Metadata, RunView } from "@memberjunction/core";
+import { BaseEntity, DatabaseProviderBase, LogError, Metadata, RunView } from "@memberjunction/core";
 import {
     ActionInvocationProvenance,
     ActionParam,
@@ -195,11 +195,13 @@ export class EntityActionInvocationSingleRecord extends EntityActionInvocationBa
     /**
      * The deferral that hands this run to the durable substrate, or `undefined` to execute normally.
      *
-     * Returns undefined for every case that must stay inline: a binding that did not opt in, a
-     * lifecycle event that participates in the save, or a host with no submitter registered. The
-     * last is a fallback rather than a refusal — `RunMode='Durable'` asks for the work to be harder
-     * to lose, so declining to run it where the durable path is unavailable would make the opt-in
-     * less reliable than leaving it off.
+     * Returns undefined for every case that must stay inline: a binding that did not opt in, or a
+     * lifecycle event that participates in the save (Validate / Before*).
+     *
+     * After* + Durable with no queue submitter (CLI `mj sync push`): do **not** nest in the
+     * caller's EntityTransactionScope. Defer until TransactionDepth is 0, then fire-and-forget.
+     * Dropping the work would make Durable worse than leaving it off; nesting it is what blew
+     * up cheese (LogActivity inside Person.Save on a shared provider).
      */
     protected BuildDurableDeferral(
         params: EntityActionInvocationParams,
@@ -215,7 +217,14 @@ export class EntityActionInvocationSingleRecord extends EntityActionInvocationBa
         }
         const submitter = DurableEntityActionRegistry.Instance.Submitter;
         if (!submitter) {
-            return undefined;
+            return async (runParams: RunActionParams): Promise<ActionResultSimple | null> => {
+                this.scheduleDurableLocalRun(params, action, runParams);
+                return {
+                    Success: true,
+                    ResultCode: 'DEFERRED_LOCAL',
+                    Message: 'Durable action deferred until the ambient transaction settles (no queue submitter in this process).',
+                };
+            };
         }
 
         return async (runParams: RunActionParams): Promise<ActionResultSimple | null> => {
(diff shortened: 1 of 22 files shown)


packages/Actions/Engine/src/entity-actions/EntityActionInvocationTypes.ts
```diff
@@ -254,6 +263,33 @@ export class EntityActionInvocationSingleRecord extends EntityActionInvocationBa
         };
     }
 
+    /**
+     * CLI / no-queue Durable fallback: wait until the save's transaction has settled, then run
```
AN-BC: Yes, this is CLI, but not just CLI so just call it local/no-queue not CLI CLI can be an example

packages/Actions/Engine/src/entity-actions/EntityActionInvocationTypes.ts
```diff
@@ -254,6 +263,33 @@ export class EntityActionInvocationSingleRecord extends EntityActionInvocationBa
...
+     * the action without DeferExecution. Errors are logged; the originating Save already succeeded.
+     */
+    private scheduleDurableLocalRun(
```
AN-BC: probably best to make this protected so sub-classes have some flexibility here if needed

packages/MetadataSync/src/config.ts
```diff
@@ -291,6 +291,14 @@ export interface EntityConfig {
     /** Whether to ignore virtual fields during pull (defaults to false) */
     ignoreVirtualFields?: boolean;
   };
+  /**
+   * Push-specific options. `skipGeoCoding` maps to `EntitySaveOptions.SkipGeoCoding`
```
AN-BC: move this comment to the skipGeoCoding bool below and keep this comment but gneeric so woudl be good for all future push options

packages/MJCore/src/generic/geoFields.ts
```diff
@@ -0,0 +1,124 @@
...
+ * one **writable** Geo* field. Virtual display fields (PrimaryAddressLatitude,
+ * `__mj_Latitude`, `__mj_Latitude_{FK}`) never invoke the provider.
+ *
+ * **Anti-patterns:** do not author RecordGeoCode JSON + SHA to skip import
+ * geocoding; do not treat Person/Org primary address as EmbeddedRecord (that is
+ * AddressLink + layered view → PrimaryAddressLatitude).
+ */
+import { EntityFieldInfo, EntityInfo } from './entityInfo';
+
+export function HasNativeLatLngFields(fields: EntityFieldInfo[]): boolean {
```
AN-BC, line 20: should have JSDoc for each of these functions

AN-BC (commented):
LGTM with comments noted below

AN-BC:
Pushed **55da0ad** addressing the review comments:

1. **Durable fallback wording** — no-queue path is described as **local / no-queue**; CLI `mj sync push` is only an example of a host.
2. **`scheduleDurableLocalRun`** — now **`protected`** so subclasses can replace the wait/run policy.
3. **`EntityConfig.push`** — JSDoc on `push` is generic (per-entity push knobs, not global CLI flags). `skipGeoCoding` docs moved onto that boolean.
4. **`geoFields.ts`** — JSDoc on every helper (`HasNativeLatLngFields`, native lat/lng pickers, `ShouldJoinRecordGeoCodes`, embedded specs, map field resolution).

packages/Actions/Engine/src/entity-actions/EntityActionInvocationTypes.ts
AN-BC: Fixed in 55da0ad.

packages/Actions/Engine/src/entity-actions/EntityActionInvocationTypes.ts
AN-BC: Fixed in 55da0ad.

packages/MetadataSync/src/config.ts
AN-BC: Fixed in 55da0ad.

packages/MJCore/src/generic/geoFields.ts
AN-BC: Fixed in 55da0ad.

AN-BC (commented):
LGTM

AN-BC:
One gate note on the merged head, from verifying more-cheese#33 against it.

`packages/MetadataSync/src/__tests__/skip-geocoding-config.test.ts` cannot fail: it builds an `EntityConfig` literal with `push: { skipGeoCoding: true }` and asserts the literal holds `true`. No implementation is exercised. The wiring it stands in for — `PushService.ts:1406`, `if (entityConfig?.push?.skipGeoCoding) saveOptions.SkipGeoCoding = true` — has no test at all: no file under `MetadataSync/src/__tests__` imports `PushService`, so deleting that line leaves the suite green. The test plan lists this file as the MetadataSync gate for the feature.

What would make it a gate: a test that runs the save path with an `entityConfig` carrying `push.skipGeoCoding` and asserts the `EntitySaveOptions` reaching `Save()` has `SkipGeoCoding === true`, plus the same save without the key asserting `false` (mirroring `skip-side-effects.test.ts` on the provider side). Not blocking anything now; it is the one claim in the plan that does not survive re-running.

---

PR #4280, opened 2026-09-07
fix(codegen): Open App CodeGen_Run SQL goes to app migrations/codegen

AN-BC:
## Summary

Open App CodeGen was applying EntityField INSERTs to the authoring DB while logging `CodeGen_Run_*.sql` into **`MJ/migrations/v5`** (CodeGenLib default `SQLOutput.folderPath`) because `mj codegen` ran from the MJ repo. The app never got an artifact; we transcribed EntityField rows by hand (drift).

- **Open App cwd** (`mj-app.json`): SQLOutput defaults to `./migrations/codegen`. MJ `./migrations/v5` default is ignored.
- **MJ repo cwd + `includeSchemas` listing a non-core schema**: **fails** — run CodeGen from the app directory.
- **`SQLOutput.enabled` but no log file open**: refuse to apply metadata SQL.
- **`--sql-output-dir`**: override; cannot point an app run at `MJ/migrations/v*`.
- Layered BaseViews (People/Org overlay wrapping `vw*Generated`): **one CodeGen pass**. Overlay is hand-authored; CodeGen writes only the inner view. Apply overlay + `mj sync push` pins, then codegen once.

Common follow-up folds the official `CodeGen_Run_2026-09-07_14-06-05.sql` Org lat/lng INSERTs (not a hand transcription).

## Test plan

- [x] `sql-output-folder.test.ts` + existing PG routing tests (23 passed)
- [x] `tsc --noEmit` on `@memberjunction/codegen-lib`

### DIFF
--- a/.changeset/codegen-sqloutput-open-app.md
+++ b/.changeset/codegen-sqloutput-open-app.md
@@ -1,0 +1,6 @@
+---
+"@memberjunction/codegen-lib": patch
+"@memberjunction/cli": patch
+---
+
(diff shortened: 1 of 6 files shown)


AN-BC:
Common fold of the official CodeGen_Run EntityField SQL: https://github.com/MemberJunction/bizapps-common/pull/131

PR #4290, opened 2026-09-07
fix(sync): one provider per JSON graph so nested children share the parent TX

AN-BC:
## Summary

`mj sync push` forked a provider **per flattened row**, then released it before nested children ran. Secure Messaging `Issue Portal Magic Link` + param `SessionID` hung 180s: the child `INSERT EXEC spCreateActionParam` waited on `FK_ActionParam_Action` while the parent Action sat uncommitted on another pooled connection. `--parallel-batch-size 1` did **not** fix this — every row still forked.

- Flatten assigns `graphId` (JSON root + nested `relatedEntities`).
- One `CreateIndependentInstance()` per graph; children reuse it.
- Parallelism is across **sibling roots** (default 10), not parent vs child.
- Release after the whole graph; commit leftover depth on success, rollback on failure.

## Test plan

- [x] Analyzer: two Actions with nested Params share graphIds correctly (`record-dependency-analyzer.test.ts`, 6 passed)
- [x] `tsc --noEmit` MetadataSync
- [ ] Retest `mj sync push --dir bizapps-secure-messaging/metadata` on `MJ_6_1_edge_5_fwd_new` (Actions + Params)


packages/MetadataSync/src/services/PushService.ts
```diff
@@ -772,137 +771,131 @@ export class PushService {
...
               }
-            } catch (recordError) {
-              const errorMsg = `Error processing ${flattenedRecord.entityName} record at ${flattenedRecord.path}: ${recordError}`;
-              callbacks?.onError?.(errorMsg);
-              errors++;
+              await provider.ReleaseIndependentInstance();
+            } catch {
+              // still release remaining graphs
             }
           }
```
AN-BC: **Blocking — this `catch {}` swallows a commit failure, and skips the release for the provider that failed.**

Two separate bugs in the same five lines:

1. **A rejected `CommitTransaction()` is invisible.** Deadlock victim, doomed transaction, dropped connection — all land in the bare `catch` and execution continues normally. By then `applyProcessResult` has already counted the records as `created`, and after the `finally` the code writes the JSON files back and **stores fresh checksums in the sync state**. So the next `--incremental` push skips records that were never committed. Silent data loss under a green summary — exactly the failure mode the repo's "never swallow errors" rule exists for.

2. **`ReleaseIndependentInstance()` is inside the same `try`.** If the commit throws, that provider's transaction handle is never dropped — and with it the dedicated pooled connection `BeginPhysicalTransaction` took via `new sql.Transaction(this._pool)`. The comment "still release remaining graphs" describes what happens to the *other* providers; the current one leaks.

Suggested shape:

```ts
for (const provider of graphProviders.values()) {
  try {
    if (!graphFailed && provider.TransactionDepth > 0) {
      await provider.CommitTransaction();
    }
  } catch (e) {
    graphFailed = true;
    callbacks?.onError?.(`Failed to commit graph transaction: ${(e as Error).message}`);
    commitError ??= e;               // rethrow after the loop so the push cannot report success
  } finally {
    await provider.ReleaseIndependentInstance().catch(() => { /* pool-safe: handle already dropped */ });
  }
}
if (commitError) throw commitError;
```

---

packages/MetadataSync/src/services/PushService.ts
```diff
@@ -772,137 +771,131 @@ export class PushService {
...
-                  else if (result.status === 'unchanged') unchanged++;
-                  else if (result.status === 'deleted') deleted++;
-                  else if (result.status === 'skipped') skipped++;
-                  else if (result.status === 'error') errors++;
-                  else if (result.status === 'deferred') {
-                    created++; // Deferred records were saved (count as created)
-                    deferred++; // Also track separately for reporting
-                  }
-                }
+              if (!graphFailed && provider.TransactionDepth > 0) {
```
AN-BC: **Blocking — `graphFailed` only covers *thrown* failures, so an errored push now commits work the old code rolled back.**

`applyProcessResult` maps `status === 'error'` to `errors++` **without throwing**, and that path is live: `PushService.ts:1104` — record-not-found with `push.autoCreateMissingRecords=false`.

Trace it through:

- **Before this PR:** the per-record provider was released immediately, and `DatabaseProviderBase.ReleaseIndependentInstance()` rolls back when `TransactionDepth > 0`. Leftover depth was discarded.
- **After this PR:** nothing threw → `graphFailed === false` → leftover depth is **committed** here. Meanwhile the caller skips `transactionManager.commitTransaction()` (gated on `totalErrors === 0`, line 584) and never reaches the rollback at line 595 either, because no exception propagated.

Net effect: a push that reports errors half-commits. The gate needs to be "no record in this file reported an error" as well as "nothing threw" — e.g. have `applyProcessResult` set `graphFailed = true` on `status === 'error'`, or thread the file-level `errors` count into the condition.

---

packages/MetadataSync/src/services/PushService.ts
```diff
@@ -772,137 +771,131 @@ export class PushService {
...
-            if (options.verbose && level.length > 1) {
-              callbacks?.onLog?.(`   Processing dependency level ${levelIndex} with ${level.length} records in parallel...`);
-            }
-            
-            // Process records in this level in parallel batches
+        // One provider per JSON-root graph (Action + nested Action Params share a
+        // connection/TX). Parallelize sibling roots only — never fork a child onto
+        // a new pooled connection while the parent row is still uncommitted.
+        const hostProvider = Metadata.Provider as unknown as DatabaseProviderBase;
+        const graphProviders = new Map<string, DatabaseProviderBase>();
```
AN-BC: **Blocking — graph providers accumulate for the whole file, which re-creates this exact deadlock one level up.**

`graphProviders` is scoped to the per-file loop and nothing evicts a graph's provider once that graph's deepest level is done — they all live until the `finally` at the end of the file.

Scale in this repo today: `metadata/components/.components.json` has **116 root records** → 116 live `SQLServerDataProvider` instances held simultaneously (each with its own rxjs `_queueSubscription`, which `ReleaseIndependentInstance` never unsubscribes).

The sharper problem is the `TransactionDepth > 0` branch at line 893 — it exists because you expect some graphs to hold an open transaction, and an open transaction **is** a dedicated pooled connection (`BeginPhysicalTransaction`: `new sql.Transaction(this._pool)`). Holding N of those at once gives you:

- **Pool exhaustion.** mssql's default `pool.max` is 10, and one connection is already held by the global push transaction from `transactionManager.beginTransaction()` (line 471). That is itself a plausible cause of a multi-minute stall.
- **The same FK lock-wait, between sibling roots.** A level-1 record in graph A that FKs to a level-0 record in graph B now blocks on B's uncommitted row, because B's provider is not released until end-of-file. Cross-root dependencies are precisely why `dependencyLevels` exists, so this is not hypothetical — it's the `FK_ActionParam_Action` symptom relocated from parent/child to root/root.

Suggested fix: precompute, per `graphId`, the highest level index it occupies (one pass over `analysisResult.dependencyLevels`), then drain — commit + release — each graph's provider at the end of the level where it last appears. Peak provider count then stays bounded by the batch size rather than by the file's root count, and cross-root FKs become visible as soon as the producing graph is done.

---

packages/MetadataSync/src/services/PushService.ts
```diff
@@ -772,137 +771,131 @@ export class PushService {
...
+        let graphFailed = false;
+
+        const obtainGraphProvider = async (graphId: string): Promise<IMetadataProvider | undefined> => {
```
AN-BC: Two smaller things here.

**Failure isn't memoized.** If the configured provider genuinely doesn't implement `CreateIndependentInstance` — `DatabaseProviderBase` throws `does not implement CreateIndependentInstance` — this retries for *every graph in every file* and emits a ⚠️ line each time. Cache the "unavailable" verdict once (a `let independentInstancesUnavailable = false` alongside the map) and log it once.

**The `undefined` fallback silently changes transaction topology.** A graph that falls back runs on the host provider, i.e. **inside** the global push transaction started at line 471, while its siblings run outside it. Worth stating in the comment — it's the difference between "this graph is atomic with the whole push" and "this graph commits on its own", and right now the log line reads like a mere perf note.

---

packages/MetadataSync/src/services/PushService.ts
```diff
@@ -772,137 +771,131 @@ export class PushService {
...
+          else if (result.status === 'skipped') skipped++;
+          else if (result.status === 'error') errors++;
+          else if (result.status === 'deferred') {
+            created++;
+            deferred++;
+          }
+        };
+
+        try {
+          const levels =
```
AN-BC, line 824: Folding the sequential fallback into the unified path also changed its error semantics, and that isn't called out in the PR body or the changeset.

The deleted `else` branch caught per-record errors, called `callbacks?.onError?.(...)`, incremented `errors++`, and **continued to the next record**. The merged path throws on the first failure and aborts the whole push.

Fail-fast is arguably the better behavior and consistent with the parallel path — but it's a user-visible change for anyone relying on "push what you can, report the rest", so it should be named in the changeset rather than arriving as a side effect of a transaction fix.

(Also worth confirming: is `analysisResult.dependencyLevels` ever actually empty when `sortedRecords` is non-empty? If not, this branch is dead and could just be an assertion.)

---

packages/MetadataSync/README.md
```diff
@@ -1529,7 +1529,7 @@ Records are automatically grouped into dependency levels:
 - **Level 1**: Records that depend only on Level 0 records
 - **Level 2**: Records that depend on Level 0 or Level 1 records
 
-Records within the same dependency level are processed in parallel. Each record uses `CreateIndependentInstance()` so nested `EntityTransactionScope`s cannot interleave on one provider. Default batch size is **10** (do not default to 1 to paper over a shared provider).
+Records are grouped into **JSON-root graphs** (an Action and its nested Action Params share one graph). Each graph gets one `CreateIndependentInstance()` for the whole tree so child FKs see the uncommitted parent on the same connection. **Sibling graphs** at the same dependency level run in parallel (default batch **10**). `--parallel-batch-size 1` does **not** mean one global provider — it only serializes graphs.
```
AN-BC: "so child FKs see the uncommitted parent on the same connection" asserts a mechanism the code only *conditionally* provides.

Nothing in the push begins a transaction on the graph provider. `BaseEntity.Save()` opens and settles its own `EntityTransactionScope` per record (`packages/MJCore/src/generic/baseEntity.ts:2056-2067`), which is exactly why the `finally` has to *test* `provider.TransactionDepth > 0` rather than unconditionally commit.

So one of two things is true, and the README should say which:

- **The scope always settles.** Then the parent Action is committed before the child Param runs, there is no uncommitted parent to "see", and graph affinity isn't what fixes the hang — something else is, and the root cause in the PR description needs revisiting.
- **The scope sometimes doesn't settle.** Then there's leftover depth escaping `BaseEntity.Save()`, which is a bug worth naming on its own — and this change converts it from "rolled back at release" into "committed at end of file" (see my comment on line 893).

The `--parallel-batch-size 1` clarification in the same paragraph is accurate and genuinely useful — keep that part.

---

packages/MetadataSync/src/lib/record-dependency-analyzer.ts
```diff
@@ -45,6 +51,18 @@ export interface DependencyAnalysisResult {
 /**
  * Analyzes and sorts records based on their dependencies
  */
+/** Group flattened rows by JSON-root graph for provider affinity. */
+export function groupRecordsByGraphId(records: FlattenedRecord[]): Map<string, FlattenedRecord[]> {
+  const map = new Map<string, FlattenedRecord[]>();
+  for (const rec of records) {
+    const gid = rec.graphId ?? rec.id;
```
AN-BC: Nit: `rec.graphId ?? rec.id` is unreachable — `FlattenedRecord.graphId` is declared required (line 27) and both producers set it (`flattenRecords` line 245, `DeletionAuditor` line 435).

Either drop the `??` and let the type carry the guarantee, or make `graphId` optional and document why a row might legitimately arrive without one. As written it reads as distrust of the type, which invites the next reader to add a second fallback somewhere else.

---

packages/MetadataSync/src/__tests__/record-dependency-analyzer.test.ts
```diff
@@ -48,6 +48,47 @@ describe('RecordDependencyAnalyzer.flattenFileRecords', () => {
     expect(flattened).toHaveLength(2);
   });
 
+  it('assigns the same graphId to a JSON root and its nested relatedEntities', () => {
```
AN-BC, line 51: This test is good and covers the analyzer half well — but it doesn't reach where the risk actually is.

Every failure mode in this PR lives in `PushService`'s provider lifecycle, and none of it is exercised: one provider per graph, reuse of that provider across dependency levels, commit-on-success, rollback-on-failure, release-always, and the fallback when `CreateIndependentInstance` throws.

All of that is cheap to cover with a fake provider that counts calls — no database needed:

```ts
class FakeProvider {
  static created = 0;
  TransactionDepth = 0;
  commits = 0; releases = 0;
  async CreateIndependentInstance() { FakeProvider.created++; return new FakeProvider(); }
  async CommitTransaction() { this.commits++; this.TransactionDepth = 0; }
  async ReleaseIndependentInstance() { this.releases++; }
}
```

The two assertions I'd most want, because they're the ones the current diff gets wrong:

- a file of *N* roots with nested children creates exactly *N* providers, and each is released
- a record returning `status: 'error'` (not throwing) does **not** commit its graph's leftover depth

---

AN-BC (commented):
## Review

Static review — I traced the transaction/provider machinery this depends on (`DatabaseProviderBase.ReleaseIndependentInstance`, `SQLServerDataProvider.CreateIndependentInstance` / `BeginPhysicalTransaction`, `TransactionManager`, `BaseEntity.beginEntityTransactionScope`) but did not run the suite.

The direction is right and the analyzer half is clean. Three things in `PushService` I'd want fixed before merge, plus one process item.

### Blocking

1. **`PushService.ts:890-900` — the `finally` swallows commit failures and skips the release for the provider that failed.** A rejected `CommitTransaction()` disappears into `catch {}`; the push then counts the records as created, writes the JSON back, and stores fresh checksums, so the next `--incremental` skips records that were never committed. And because `ReleaseIndependentInstance()` sits in the same `try`, the failing provider's transaction handle and its dedicated pooled connection leak.

2. **`PushService.ts:893` — `graphFailed` only tracks *thrown* failures, so an errored push now commits work the old code rolled back.** `status === 'error'` returns without throwing (live path at line 1104, `autoCreateMissingRecords=false`). Previously the per-record release rolled leftover depth back; now it's committed, while the caller neither commits the global TX nor rolls it back.

3. **`PushService.ts:778` — graph providers accumulate for the whole file, re-creating this deadlock one level up.** Nothing evicts a graph's provider after its deepest level; `metadata/components/.components.json` alone has 116 roots. Since an open transaction is a dedicated pooled connection (mssql default `pool.max` 10, one already held by the global push TX), this risks pool exhaustion — and a level-1 record in graph A that FKs to a level-0 record in graph B now blocks on B's uncommitted row. Same `FK_ActionParam_Action` symptom, relocated from parent/child to root/root.

### Non-blocking

4. Rollback-on-failure is an invisible side effect — nothing calls `RollbackTransaction()`; it happens inside `ReleaseIndependentInstance()`, whose doc promises only "drop this instance's transaction handle". The PR body advertises "rollback on failure"; make it explicit at the call site so a future edit to the base class can't quietly turn it into a leak.
5. `obtainGraphProvider` doesn't memoize failure, and the `undefined` fallback silently puts that graph inside the global transaction while its siblings sit outside it.
6. The sequential fallback lost its per-record error tolerance — an undocumented behavior change.
7. README overstates the mechanism (details inline).
8. Tests don't reach the provider lifecycle, which is where all of the above lives.

### The process item

The test plan leaves `Retest mj sync push --dir bizapps-secure-messaging/metadata` unchecked. The diagnosis rests entirely on runtime lock behavior that nothing in the diff demonstrates, and findings 1–3 are all shapes where the code would look fine and behave wrong. That box is the one that matters most here.

### Architectural note (pre-existing, but this is where it surfaces)

`transactionManager.beginTransaction()` (line 471) wraps the whole push on the host singleton, but Phase 1 writes all go to independent providers. The only writes actually inside that transaction are Phase 2.5 deferred records (line 2178 calls `processFlattenedRecord` with no `recordProvider`). So `mj sync push` has no all-or-nothing guarantee across files, and the rollback at line 595 rolls back a nearly empty transaction. Not introduced here — but this PR is the natural moment to decide whether the graph providers should join a real outer transaction or the wrapper should go, so the contract isn't implied.

### What's good

Graph-scoped providers cut `CreateIndependentInstance()` from one-per-flattened-row to one-per-root, and each of those costs a `CheckToSeeIfRefreshNeeded` round trip plus a full `Config()` — a real reduction, not just a correctness change. The `graphId` threading through `flattenRecords` is clean, and the `DeletionAuditor` synthetic row was correctly updated rather than left to the `??` fallback. The `--parallel-batch-size 1` clarification in the README is accurate and worth keeping.

Also: `mergeable_state: behind` — needs `next` merged in before it can land.

---

AN-BC:
## CI note before anyone chases this: the red check is not this PR's

`Run unit tests` is red, but every test shard passed. That job is an aggregator — it failed because the **`guards`** lane failed:

```
guards: failure    build: success    test: success    esm-guard: success
##[error]Unit test lanes did not pass: guards
```

And the guard finding is:

```
❌ 1 spec anti-pattern(s) found:
   packages/geo/geo-core/src/__tests__/google-geocode.integration.test.ts:27
   — disabled test — document it with an adjacent `// KNOWN LIMITATION` comment or fix it
```

This PR touches six files, all under `packages/MetadataSync/` — nothing in `packages/geo/`.

**It is already fixed on `next`.** Commit `319a7edc1` (#4291) added the required comment at lines 27-28:

```ts
// KNOWN LIMITATION: this suite calls the live Google Geocoding API and needs a real key from
// mj.env; without one it is skipped rather than failed so unit-test lanes stay hermetic.
const describeLive = apiKey ? describe : describe.skip;
```

This branch was cut at `f2ec156da`, whose only child on `next` is that very commit — so the branch is exactly one commit behind the fix.

**The remedy is the one this PR already needs: merge `next` in.** That clears the red guard and the `behind` mergeable state in a single step. No code change here, no re-run, and nothing to port — the fix is sitting in the base branch.

Flagging it so the guard failure doesn't get mistaken for fallout from the transaction changes while the review findings are being worked.

---

AN-BC:
## Follow-up: mixed-provider deadlock + review comments

The hang was not "graphs still using different providers because we forked per row." Graph grouping was already correct (`Level 0: 4 graphs`, `Level 1: 19 records in 4 graphs`). The parent **Action** still saved on the **host** because `MJActionEntityServer.constructor(Entity) { super(Entity); }` dropped the ClassFactory-passed provider. `ProviderToUse` silently fell back to `BaseEntity.Provider`. The child **Action Param** (generated class, no ctor) kept the graph instance. Host TX was a savepoint around the whole push, so the Action row stayed uncommitted while `INSERT EXEC spCreateActionParam` waited on `FK_ActionParam_Action`.

That is a core invariant: **every DB read/write in a graph must use the provider that was passed in.** `--parallel-batch-size 1` cannot fix a dropped constructor arg.

### This commit (`4456222c8f`)

**ORM**
- `BaseEntity.BindProvider` / `BoundProvider`. `GetEntityObject` and `ConstructUninitializedEntity` always rebind after construct, so a future 1-arg subclass cannot drop the instance again.
- `MJActionEntityServer`, `MJApplicationEntityServer`, `MJRemoteOperationEntityServer` forward `Provider`.
- Every instance `RunView` uses `this.RunViewProviderToUse`. `ResolveLeafEntityRecursive` threads the caller provider.

**Push lifecycle (reviewer blocking items)**
- New `GraphProviderPool`: one instance per JSON-root graph; drain at last dependency level (peak bounded by `--parallel-batch-size`, not root count); leftover depth committed on success, **explicit** `RollbackTransaction` on failure; **always** `ReleaseIndependentInstance` even if commit throws; `status: 'error'` (no throw) does **not** commit; `CreateIndependentInstance` failure is memoized and **all** graphs in the file use the host — never a mix.
- Fail-fast on the first thrown record error (named in the changeset; it is a behavior change from the old sequential "push what you can").
- Lookups / `loadEntity` / `GetEntityObject` during a push use the graph provider.

**Geo**
- `GeoCodeSyncService` writes `MJ: Record Geo Codes` on the owning entity's provider.

**Docs / tests**
- README no longer claims "child FKs see the uncommitted parent" as if `Save()` left depth open. It states the invariant, drain semantics, leftover-depth settle, and the host-fallback topology.
- `rec.graphId ?? rec.id` removed (`graphId` is required).
- Unit tests: 1-arg subclass drop + `BindProvider`; pool lifecycle (N roots → N instances + release, error-status no commit, commit failure still releases, unavailable memoized to host).

### Empirical (the unchecked box)

```
mj sync push --dir bizapps-secure-messaging/metadata --parallel-batch-size 10
DB=MJ_6_1_edge_5_fwd_new
```

Previously hung ~180s on SessionID (`6761E4F1-…`). This run: **7.5s wall, 24 created / 9 updated / 0 errors**. Action `76A41213-…` Issue Portal Magic Link and Param SessionID both present. Log: `Level 0: 4 records in 4 graphs`, `Level 1: 19 records in 4 graphs`. Also created the SecMsg Application (the other 1-arg subclass).

Also merged `next` (clears the geo KNOWN LIMITATION guard that was failing CI on this branch).

packages/MetadataSync/src/services/PushService.ts
AN-BC: Addressed in 4456222c8f. `GraphProviderPool.releaseGraphs` now: (1) commit leftover depth only when the graph did not fail; (2) on commit/rollback throw, record the error and still `ReleaseIndependentInstance` in `finally`; (3) the first settle error is returned to the caller so the push cannot report success and write checksums for uncommitted rows. Covered by `graph-provider-pool.test.ts` ('returns a commit failure and still releases').

packages/MetadataSync/src/services/PushService.ts
AN-BC: Addressed in 4456222c8f. `applyProcessResult` calls `graphPool.markFailed(graphId)` on `status === 'error'` (the live `autoCreateMissingRecords=false` path). Drain then rolls leftover depth back instead of committing. Test: 'does not commit leftover depth when a record reported status: error'.

packages/MetadataSync/src/services/PushService.ts
AN-BC: Addressed in 4456222c8f. `noteLevels` records each graph's last dependency-level index. After each parallel batch, `drainBatch` commit/rollback+releases graphs whose last level is the current one. Graphs that still appear later stay live. Peak independent instances is therefore bounded by `--parallel-batch-size` (plus in-flight multi-level graphs), not by the file's root count. Test: 'creates one independent instance per graph and releases each at last level'.

packages/MetadataSync/src/services/PushService.ts
AN-BC: Addressed in 4456222c8f. `CreateIndependentInstance` failure is memoized on the pool (`independentUnavailable`). Logged once. Subsequent `obtain` returns the host for every graph in the file — never a mix of host + independent instances, which is the deadlock topology. Host is never stored in the map, so drain cannot `ReleaseIndependentInstance` the global push TX. Test: 'memoizes CreateIndependentInstance failure and uses the host for every graph'.

packages/MetadataSync/src/services/PushService.ts
AN-BC: Named in the changeset and a comment at the file-processing loop: fail-fast on the first thrown record error is intentional and matches the parallel path. It is a user-visible change from the old sequential 'push what you can, report the rest'. `dependencyLevels` empty with non-empty `sortedRecords` still falls back to a single level (`[sortedRecords]`) so that branch is not an assertion; flattening always produces levels or we wrap.

packages/MetadataSync/README.md
AN-BC: Addressed in 4456222c8f README. The previous sentence over-claimed. `BaseEntity.Save()` still opens and settles its own `EntityTransactionScope` per record. Graph affinity is the invariant that every DB op in the tree uses one connection (constructor-drop was the hang: Action Save on host savepoint, Param Save on graph instance). Leftover depth is committed at drain on success and explicitly rolled back on failure. `--parallel-batch-size 1` clarification kept.

packages/MetadataSync/src/lib/record-dependency-analyzer.ts
AN-BC: Addressed in 4456222c8f. `groupRecordsByGraphId` uses `rec.graphId` directly. It is required on `FlattenedRecord` and both producers set it.

packages/MetadataSync/src/__tests__/record-dependency-analyzer.test.ts
AN-BC: Addressed in 4456222c8f `packages/MetadataSync/src/__tests__/graph-provider-pool.test.ts`: N roots → N instances and each released at last level; `status: 'error'` does not commit leftover depth; commit throw still releases; unavailable `CreateIndependentInstance` uses host for every graph and does not release the host.

packages/MetadataSync/src/services/PushService.ts
AN-BC: **Keeping this open — the drain mechanism is right, but the bound it claims does not hold.**

`noteLevels` + `drainBatch` is the correct shape and I verified the reuse/release logic. The problem is the claim in `README.md` and in the `GraphProviderPool` header:

> Peak live instances is therefore bounded by `--parallel-batch-size`, not by the file's root count.

For any root that has nested children, `lastLevelByGraph` is **1**, so `drainBatch(batchIds, 0)` matches nothing and every graph stays live through all of level 0. Peak equals the root count.

Your own test asserts exactly this — after `drainBatch(['A','B'], 0)`, `releases === 0` for both.

I re-implemented the drain arithmetic and measured it:

```
shape                               batch  PEAK   claim(=batch)?
116 roots, NO children              10     10     HOLDS
116 roots, 2 children each          10     116    VIOLATED
2 roots, 1 child each (their test)  10     2      HOLDS
500 roots, 3 children each          10     500    VIOLATED
```

The bound holds only when roots have no children — precisely the case where graph affinity isn't needed. `metadata/components/.components.json` (116 roots) peaks at 116.

**Severity is much lower than my original finding**, and I want to be explicit about why: your `BindProvider` fix means `TransactionDepth` is 0 between levels, so none of those instances pins a pooled connection. The pool-exhaustion and cross-root FK-wait risks I raised are genuinely gone. What remains is cost — N `Config()` calls each doing a `CheckToSeeIfRefreshNeeded` round trip, N rxjs `_queueSubscription`s that `ReleaseIndependentInstance` never unsubscribes, N live objects.

So this is not a correctness blocker. It is a **false claim in the docs**, which is the thing that gets relied on later ("we're bounded, a 500-root file is fine").

Two ways to close it:

1. **Correct the sentence** — peak is bounded by the current batch *plus every graph that still has a later level*, i.e. by root count for the common root+children shape.
2. **Actually bound it** — since depth is 0 between levels, a graph at depth 0 can be released at the end of a level and re-obtained at the next; the parent row is already committed so a fresh instance is safe.

Either is fine. Option 1 plus a note about the residual per-instance cost is the smaller change and I'd take it.

---

packages/MetadataSync/src/services/PushService.ts
AN-BC: Memoization is fixed — `independentUnavailable` + `unavailableLogged`, one log line, verified by the `memoizes CreateIndependentInstance failure` test. Thanks.

**Keeping this open for a new hole in the no-mix invariant.** The class doc and the log line both assert:

> either every graph in the file gets its own independent instance, or every graph uses the host … we refuse that topology

That is only true when the **first** `CreateIndependentInstance` fails. Walk the other case:

1. Graphs `g0..g9` obtain independent instances — `providers` has 10 entries.
2. Graph `g10` calls `CreateIndependentInstance`, which throws (pool pressure, a transient network blip, a dropped socket — not just "unimplemented").
3. `independentUnavailable = true`, `g10` returns `this.host`.
4. `g11..gN` also get the host.

Now the file is running `g0..g9` on independent connections and `g10..gN` on the host, inside the push transaction. That is exactly the mixed topology this PR exists to eliminate — and the code has just logged a message saying it refused it.

The existing test only covers the first-call failure (`host.independentShouldThrow = true` before any `obtain`), so it passes while the invariant is unenforced. Same shape as the original bug: a claim that is true at the moment it's checked and stays asserted after it stops being true.

Options, in order of how much I like them:

- **Fail the file.** If `independentUnavailable` flips while `providers.size > 0`, throw — a mid-file topology change is not something to recover from silently, and the push aborts cleanly with the host TX rolled back.
- **Drain first, then degrade.** Settle and release the existing independents before handing out the host, so the mix never coexists.
- **Keep behavior, fix the words.** Least good — the log line would have to say "we may already have independent instances live", which is an admission the invariant isn't one.

A test that obtains two graphs successfully, flips the throw, then obtains a third would pin whichever you pick.

---

packages/MetadataSync/README.md
AN-BC: The old claim is gone and the replacement is much better — it states the invariant, names the drain semantics, and is honest that `Save()` settles its own scope so leftover depth is the exception rather than the mechanism. That was the right correction, and finding the actual cause (`MJActionEntityServer` dropping the ctor arg) is what made it possible.

**Keeping this open only for the one sentence in the same paragraph that is now false:**

> Peak live instances is therefore bounded by `--parallel-batch-size`, not by the file's root count.

Measured detail is on the `graphProviders` thread — short version: for any root with nested children every graph's last level is 1, so nothing drains during level 0 and peak equals the root count (116 for `.components.json`, 500 for a 500-root file). The bound only holds for childless roots.

Worth fixing here rather than letting it ride, because this paragraph is now the canonical description of the mechanism and that is the sentence someone will quote back when sizing a large push.

One more, unrelated to the bound — this line:

> Every DB read and write for a graph — `GetEntityObject`, `Save`, `Load`, `RunView`, lookups, RecordGeoCode — must use that same provider.

is the most valuable sentence in the whole change, and it's buried in the MetadataSync README where only sync work will find it. The invariant it states is an **ORM** invariant: it's what `BindProvider` enforces and what any future 1-arg subclass will violate. Consider also stating it where a subclass author will actually be standing — a line in `MJCore`'s docs, or a comment on `BindProvider` itself pointing at the failure mode. The `BindProvider` docblock is close, but it explains the mechanism rather than the rule.

---

AN-BC (commented):
## Round 2 — `4456222c8f`

I read the diff rather than the summary, and traced `BindProvider` → `_provider` → `ProviderToUse` → `RunViewProviderToUse` to confirm the binding actually reaches the call sites. Five of eight threads resolved; three stay open, all reduced in severity.

### The root cause finding is the important part of this PR

`MJActionEntityServer.constructor(Entity) { super(Entity); }` dropping the ClassFactory provider is a much better explanation than the original one, and it explains the evidence the original story couldn't: why grouping was already correct in the logs (`Level 0: 4 graphs`) yet the hang persisted, and why `--parallel-batch-size 1` didn't help. A dropped constructor argument is invisible to every knob.

Worth saying plainly: **the PR now fixes a different bug than the one it was opened for**, and the new one is broader — it's an ORM defect, not a sync defect. Any `@RegisterClass` subclass with a 1-arg constructor has been silently running on the global host provider. That's why the changeset correctly grew to `@memberjunction/core`.

### Verified fixed — resolved

| # | Finding | Evidence I checked |
|---|---|---|
| 1 | Commit failure swallowed; release skipped | `releaseGraphs` settles in `try`, releases in `finally`, returns `settleError`; caller throws it. Test: *returns a commit failure and still releases* |
| 2 | Non-throwing `status: 'error'` committed | `applyProcessResult` → `markFailed(graphId)` → rollback. Test: *does not commit leftover depth when a record reported status: error* |
| 4 | Rollback was an implicit side effect | Now an explicit `RollbackTransaction()` at the call site, with a comment saying why not to lean on `ReleaseIndependentInstance` |
| 6 | Fail-fast change undocumented | Named in both the changeset and the README |
| 8 | `rec.graphId ?? rec.id` unreachable | Removed |
| 9 | Tests didn't reach the lifecycle | `graph-provider-pool.test.ts` covers all five behaviours, including the two I asked for |

Also confirmed the ORM threading is real, not just declared: `BindProvider` is called in `ProviderBase.GetEntityObject` **before** `Config()`, and again in `ConstructUninitializedEntity` where `provider = this.ProviderToUse`, so embed chains inherit correctly. `ResolveLeafEntity` threads `md = provider ?? Metadata.Provider` properly.

And the empirical box is checked: 180s hang → 7.5s, 24 created / 9 updated / 0 errors. That was the item I flagged as the one I couldn't verify from here.

### Still open — none are correctness blockers

**A. The peak-instances bound is stated but not achieved.** ([thread](https://github.com/MemberJunction/MJ/pull/4290#discussion_r3953796555)) `drainBatch` is the right mechanism, but any root with children has `lastLevel = 1`, so nothing drains during level 0. I re-implemented the arithmetic and measured: 116 roots with children peaks at **116**, not 10; the bound holds only for childless roots. Your own level-0 test asserts `releases === 0`, which is the same fact. Severity dropped a lot because `BindProvider` means depth is 0 between levels, so nothing pins a connection — the pool-exhaustion and cross-root FK risks are genuinely gone. What's left is per-instance cost and a docs claim that will get relied on.

**B. "Never a mix" isn't enforced when `CreateIndependentInstance` fails mid-file.** ([thread](https://github.com/MemberJunction/MJ/pull/4290#discussion_r3953797320)) If it succeeds for the first N graphs and then throws — pool pressure, a dropped socket, not just "unimplemented" — the earlier graphs keep independent instances while the rest get the host, and the code logs that it refused exactly that. The test only covers first-call failure, so it passes while the invariant is unenforced.

**C. The README sentence, plus a placement suggestion.** ([thread](https://github.com/MemberJunction/MJ/pull/4290#discussion_r3953798272)) Same bound claim. Separately: *"Every DB read and write for a graph must use that same provider"* is the most valuable line in this change and it lives in the MetadataSync README, where a future subclass author will never look. It's an ORM invariant — worth stating next to `BindProvider`.

### Two smaller things, no thread

**`failedGraphs` is dead.** `markFailed(graphId)` always sets `anyFailed = true`, and `releaseGraphs` reads `this.anyFailed || this.failedGraphs.has(id)` — so the per-graph branch is unreachable once anything has failed. The effective semantic is "one record error anywhere in the file rolls back every graph in that file", which is consistent with the push skipping its global commit when `totalErrors > 0`. If that's what you want, delete the `Set` so the code stops implying per-graph isolation it doesn't have. If you *do* want isolation, `markFailed(graphId)` shouldn't touch the global flag.

**17 new `as unknown as` casts.** Most are unavoidable given MJ's `IMetadataProvider` / `IEntityDataProvider` / `IRunViewProvider` split, and I'm not going to pretend otherwise. The one I'd push back on is `new Metadata() as unknown as IMetadataProvider` in `GeoCodeSyncService` — that asserts a facade is a provider. Since only `GetEntityObject` is called on it, narrowing the parameter to what's actually used would drop the cast honestly rather than by assertion.

### State

CI on `4456222c8f` was still in flight when I looked — only the skipped `claude`/`gemini` lanes and a neutral CodeQL had reported. The `next` merge landed, so the geo `KNOWN LIMITATION` guard that was failing should now be green; I'll confirm on the next check-in rather than assume it.

Nothing here blocks merge on my read. A, B and C are all "the code is fine, the claim isn't" — which is worth closing precisely because this PR's whole lesson is that a stale-but-asserted claim is what cost the 180 seconds in the first place.

---

packages/MetadataSync/src/services/PushService.ts
AN-BC: Addressed in 5cb97f08c8 by making the bound true, not just correcting the sentence. `drainBatch` now also releases a graph whose `TransactionDepth` is already 0 (Save settled — a fresh instance at the next level is safe). Leftover-depth graphs still stay live until their last level. Peak is the current batch plus any leftover-depth graphs. Tests: leftover-depth stays across levels; settled graphs release at the current batch and re-obtain a new instance.

packages/MetadataSync/src/services/PushService.ts
AN-BC: Addressed in 5cb97f08c8 with your preferred option: fail the file. If `CreateIndependentInstance` throws after `providers.size > 0`, `obtain` throws (`Refusing mixed host + independent topology`) rather than handing the rest of the file to the host. First-call failure still degrades the whole file to host. Test: obtain A+B, flip the throw, obtain C rejects; A stays independent.

packages/MetadataSync/README.md
AN-BC: Addressed in 5cb97f08c8. README now states the actual drain rule (last level **or** depth 0) and the leftover-depth exception. The ORM invariant is now the lead of `BaseEntity.BindProvider`'s docblock — every DB read/write on the instance MUST use this provider — which is where a 1-arg subclass author will be standing.

AN-BC:
## Round 2 follow-up (`5cb97f08c8`)

Three open threads, all \"the code is fine, the claim isn't\" — closed by making the claims true.

- **Peak bound.** `drainBatch` now also releases a graph whose `TransactionDepth` is 0. Leftover-depth graphs still stay until last level. Peak = current batch + leftover-depth graphs, which is what the docs now say.
- **Never a mix.** If `CreateIndependentInstance` fails after independents already exist, the file aborts. First-call failure still degrades the whole file to host.
- **ORM rule placement.** `BindProvider` docblock now leads with the rule, not just the mechanism.

Also deleted dead `failedGraphs` (one error rolls back the file, which matches `totalErrors > 0` skipping the host commit) and dropped the `new Metadata() as IMetadataProvider` assertion in GeoCodeSyncService.

7 GraphProviderPool tests green, including the two new ones for settled-graph re-obtain and mid-file mix refusal.

AN-BC (commented):
## Round 3 — `5cb97f08` — all threads closed

Re-ran my drain simulation against the new `drainBatch` logic rather than reading it:

```
NORMAL CASE (Save settles, depth 0 between levels):
shape                    batch  PEAK  bounded?  creations
116 roots, 2 children    10     10    YES       232
500 roots, 3 children    10     10    YES       1000
2 roots, 1 child         10     2     YES       4

LEFTOVER-DEPTH CASE (Save leaves depth 1):
116 roots, 2 children    10     116   NO        116
500 roots, 3 children    10     500   NO        500
```

The bound now holds in the normal case, and the leftover-depth case is exactly what the README now says ("plus any still-open leftover-depth graphs"). **Both branches of the claim are true**, which is what I was actually after — resolving [A](https://github.com/MemberJunction/MJ/pull/4290#discussion_r3953796555) and [C](https://github.com/MemberJunction/MJ/pull/4290#discussion_r3953798272).

[B](https://github.com/MemberJunction/MJ/pull/4290#discussion_r3953797320) took the option I'd have picked: `obtain` throws when `providers.size > 0`, so a mid-file failure aborts rather than silently degrading. And the new test is the one that matters — two successful obtains, *then* flip the throw. That's the case the old test structurally could not reach. Resolved.

Both nits are gone too: `failedGraphs` deleted with `markFailed()` reduced to the semantic it always had, and the two `new Metadata() as unknown as IMetadataProvider` casts dropped (`Metadata.GetEntityObject` satisfies the call site on its own, so the assertion wasn't buying anything).

The `BindProvider` docblock now leads with the rule rather than the mechanism, which is where a future 1-arg subclass author will actually be standing.

### One consequence worth knowing, not a finding

Bounding the peak costs instance creations: **N → L×N**. For 500 roots across 2 levels that's 1000 `CreateIndependentInstance` calls instead of 500, and each one runs `Config()` → `CheckToSeeIfRefreshNeeded` → a metadata round trip. Order of seconds on a large push, against a bounded peak and bounded rxjs subscriptions.

That's the right trade for a CLI — but it's the more expensive of the two options I offered, so it should be a decision rather than a surprise. If a large push ever feels slow, this is the first place to look, and the cheap mitigation is to skip the drain when the pool is small rather than to revert the bound.

### Verdict

Nothing open from my side. Eight findings raised, eight closed, and the two that mattered most — the mid-file mix and the ORM provider binding — are closed with tests that would fail if the behaviour regressed.

Worth recording what this PR ended up being: it opened as a sync parallelism fix and lands as an ORM defect fix. `MJActionEntityServer.constructor(Entity) { super(Entity); }` had been silently dropping the ClassFactory provider, so any `@RegisterClass` subclass with a 1-arg constructor ran on the global host. The sync deadlock was the symptom that happened to be visible.

CI is re-running on this head; the `next` merge already cleared the geo guard on the previous one. Approving isn't mine to do here, but I have no remaining objections.

---

AN-BC:
## Round 3 follow-up (`7323f703ec`)

Round 3 closed every thread. This commit only records the two leftovers so they are not surprises:

- **Default `--parallel-batch-size` stays 10.** It was never 1 in this PR's product code; 1 was a hang workaround on local stack scripts and is not the product default. Callers can still pass `1` to serialize graphs for debugging.
- **L×N drain cost**, the consequence you asked to be a decision rather than a surprise: releasing settled graphs between levels bounds peak at the batch size and costs one `CreateIndependentInstance`/`Config` per graph per level. Documented in the README as the chosen CLI tradeoff.

No code-path change. Empirical SecMsg hang fix is still `4456222c8f`.

AN-BC (commented):
LGTM

AN-BC (commented):
LGTM

AN-BC:
## Reviewer sign-off: the core algorithm is solid

Eight findings raised across three rounds, eight closed, all threads resolved. **No remaining objections to the algorithm.** Remaining CI work is handed to the build engineer.

### What I verified, and how

Not by reading commit messages — by tracing the code and, where a claim was quantitative, by measuring it.

**The ORM fix is correct.** `BindProvider` → `_provider` → `ProviderToUse` (`this._provider || BaseEntity.Provider`) → `RunViewProviderToUse` is a real chain, and the rebind fires where it has to: in `ProviderBase.GetEntityObject` **before** `Config()`, and in `ConstructUninitializedEntity` with `provider = this.ProviderToUse` so embed chains inherit rather than falling back. `ResolveLeafEntity` threads `md = provider ?? Metadata.Provider` properly. A one-arg subclass can no longer silently drop the ClassFactory provider.

**The pool lifecycle is correct.** Settle in `try`, release in `finally`, first settle error returned and thrown by the caller — a failed commit can no longer report success and write checksums for uncommitted rows. A non-throwing `status: 'error'` marks the file failed and rolls back rather than commits. Mid-file `CreateIndependentInstance` failure throws instead of silently degrading to a mixed topology.

**The bound claim is now true in both branches.** I re-implemented `drainBatch`'s arithmetic:

```
NORMAL (Save settles, depth 0):   116 roots/2 children → peak 10  ✓ bounded
LEFTOVER DEPTH:                   116 roots/2 children → peak 116 (documented)
```

The docs now describe both cases instead of the flattering one.

### What is NOT verified, for whoever picks up CI

The test shards have **never executed** on any head carrying the ORM change. Four heads, three runs cancelled by supersession, none reaching step 8 of 15. `Build` takes >7½ minutes here (measured lower bound; no run has completed one, so I can't give the real figure).

So the guards, migrations, standards, CodeQL and dependency lanes are green, but the change most likely to surprise anyone has zero test evidence behind it.

**The specific thing to watch:** `BaseEntity.BindProvider` and the `GetEntityObject` rebind affect every entity in the repo — roughly 300 downstream packages. Before this PR, an instance constructed by a one-arg subclass fell back to the process-wide host provider, silently and universally. Any existing test that depended on that fallback will now bind to a different provider instead.

If a shard goes red, **that is the first hypothesis, and it would be a real finding from this PR — not a flake to re-run.** The fallback itself still exists (`ProviderToUse` returns `this._provider || BaseEntity.Provider`); what changed is that `_provider` is now populated where it previously wasn't.

### Two housekeeping items for whoever merges

- The test plan still shows `- [ ] Retest mj sync push …` unchecked. It was run: 7.5s, 24 created / 9 updated / 0 errors, versus a 180s hang. Someone reading the description cold sees the central verification as outstanding.
- The summary still describes the original diagnosis ("forked a provider per flattened row"). This PR disproved that. The real cause was `MJActionEntityServer.constructor(Entity) { super(Entity); }` dropping the provider — worth correcting, since the description is what gets read in six months.

I haven't edited either; that's the author's call.

---

AN-BC:
@rkihm-BC  pretty meaningful bug fixes here deep in BaseEntity and some related bits surfaced through larger scale mj sync push work with parallelization that would eventually have bit us with runtime parallelization too.

PR #4291, opened 2026-09-07
fix(codegen): separate default-constraint drop blocks in the SQL log so captures replay

### LINKED ISSUE #4187: CodeGen SQL capture is not replayable: adjacent variable-declaring blocks emitted without GO separators
## Summary

CodeGen's SQL capture (`migrations/v5/CodeGen_Run_<ts>.sql`) is not replayable when two variable-declaring blocks land adjacent in the log. Appending the capture to a migration — exactly as [`guides/MIGRATION_CODEGEN_WORKFLOW_GUIDE.md`](../blob/next/guides/MIGRATION_CODEGEN_WORKFLOW_GUIDE.md) step 4 instructs — produces a migration that fails on replay:

```
The variable name '@constraintName' has already been declared.
Variable names must be unique within a query batch or stored procedure.
```

CodeGen itself never hits this, because it executes each logged block as its own statement. Only the concatenated log is invalid.

## Repro (from a clean state)

1. Create an empty database and add the `MJ_CodeGen` (db_owner) and `MJ_Connect` users.
2. `node packages/MJCLI/bin/run.js migrate --dir ./migrations`
3. `node packages/MJCLI/bin/run.js codegen` → writes `migrations/v5/CodeGen_Run_<ts>.sql`
4. Append that capture **verbatim** to a new migration timestamped after the last one, per the workflow guide's step 4:
   `cat migrations/v5/CodeGen_Run_*.sql >> migrations/v6/V<ts>__v6.1.x__<Name>.sql`
5. Drop and recreate the database, then re-run `migrate --dir ./migrations`.

**Expected:** the appended capture replays. The guide's step 6 states this explicitly — *"This proves that a fresh install running your single combined migration ends up in the correct state."*

**Actual:** fails at batch 1 of 23. `0` batches applied, so no partial state is left and a re-run after a fix is safe.

## Root cause

`packages/CodeGenLib/src/Database/manage-metadata.ts:4671` — `dropExistingDefaultConstraint()`:

```ts
const sqlDropDefaultConstraint = this.dbProvider.dropDefaultConstraintSQL(entity.SchemaName, entity.BaseTable, fieldName);
await this.LogSQLAndExecute(pool, sqlDropDefaultConstraint, `SQL text to drop default existing default constraints in entity ${entity.SchemaName}.${entity.BaseTable}`);
```

The emitted SQL opens with `DECLARE @constraintName NVARCHAR(255);`, and the call passes no `includeBatchSeparator`:

```ts
LogSQLAndExecute(pool, query, description?, isRecurringScript = false, includeBatchSeparator = false, batchSeparator = 'GO')
```

So no `GO` is written between consecutive blocks. On both branches tested, the affected entities are `IdentityClaim` and `IdentityClaimType` — CodeGen renames their `__mj_CreatedAt` / `__mj_UpdatedAt` defaults to the canonical `DF___mj_*` form — and all four of their blocks land in the first batch: 4 bare `DECLARE @constraintName`, with the first `GO` not appearing until roughly line 240.

## Reproduced on both branches, independently, from zero

| | `v6.1.0-edge.4` | `origin/next` (`9f86203411`) |
|---|---|---|
| CLI | `6.1.0-edge.4` | **`6.1.0-edge.5`** |
| migrations OK before failure | 64 | 68 |
| failing batch | `1/23` (lines 1-240) | `1/23` (lines 1-223) |
| bare `DECLARE @constraintName` in capture | 4 | 4 |
| `GO` separators in capture | 22 | 22 |

Not fixed on `next`, and reproduced there on a **newer** CLI.

## Suggested fix

Pass `includeBatchSeparator: true` when logging any block whose SQL declares a variable — minimally at `manage-metadata.ts:4671`:

```ts
await this.LogSQLAndExecute(pool, sqlDropDefaultConstraint, `...`, false, true);
```

Alternatively, have `dropDefaultConstraintSQL()` wrap its body in `BEGIN … END` so repeated declarations are legal within a single batch. Either way, an audit of the other `LogSQLAndExecute` call sites whose SQL declares variables would be worth doing at the same time, since any two adjacent ones have the same latent problem.

## Why this has stayed latent

It requires an entity whose default constraints CodeGen renames, **and** two such blocks adjacent in the log. All six migrations in the repo containing `DECLARE @constraintName` have it inside a stored-procedure body, where it is proc-scoped and cannot collide. It also cannot fail on a working development database — only on a fresh install (CI, a new developer, a release), which is the same invisibility signature as #3670.

## Workaround currently in use

Insert `GO` before each subsequent `/* SQL text to drop default existing default constraints ... */` block — 3 separators for the 4-block case. No generated SQL is altered.

rkihm-BC:
Fixes #4187

## Problem

CodeGen's drop-default-constraint block opens with `DECLARE @constraintName`. Each block executes as its own query, so CodeGen never fails, but the SQL log writes consecutive blocks with no `GO` between them. A migration built from that log, exactly as `guides/MIGRATION_CODEGEN_WORKFLOW_GUIDE.md` step 4 instructs, fails on a clean database:

```
The variable name '@constraintName' has already been declared.
Variable names must be unique within a query batch or stored procedure.
```

**Why it started firing in v6.1.0-edge.4.** `V202608202300` created `IdentityClaim` and `IdentityClaimType` with `DEFAULT (SYSUTCDATETIME())`. CodeGen compares the stored default against `GETUTCDATE()` by text, so on every run it drops and recreates all four constraints and logs four back-to-back `DECLARE` blocks into the first batch of the capture. The `DECLARE` block itself dates to 2024; the April 2026 batch-separator fix (9d977fd54d) covered the add-column path in the same method but not this one.

## Change

- **`dropExistingDefaultConstraint`** passes the provider's batch separator, matching the add-column path. The executed query is unchanged (the mssql path does not split on `GO`).
- **`SQLLogging.appendToSQLLogFile`** forces a separator after any logged unit that opens with a top-level T-SQL `DECLARE @…`, so a future caller that forgets cannot reintroduce the bug. Every unit is executed as its own query, so no later unit can depend on the variable and the separator is always semantics-preserving. Routine bodies are not matched (the `CREATE PROCEDURE` header comes first). An empty separator (PostgreSQL) no longer emits a stray blank line.
- **Migration `V202609071727__v6.1.x__IdentityClaim_UTC_Default_Normalization.sql`** rewrites the four defaults to `GETUTCDATE()` so CodeGen stops churning them. Idempotent: a database where CodeGen already rewrote the constraints (any name, `GETUTCDATE()` definition) is left alone. The shipped edge.4 migration is not edited, to avoid Flyway checksum drift.
- **Changeset** at `minor` on `@memberjunction/codegen-lib` (the branch carries a migration).

## Also in this PR: two test groups that were already red on `next`

CI shard 1 on #4280 shows both; #4280 merged with it failing.

- #4280's new `LogSQLAndExecute` guard throws when `SQLOutput.enabled` (defaults to true) and no capture file is open, which is every test that drives CodeGen through a stub connection. `manage-metadata-drift-hold-revoke.test.ts` and `BaseViewEmissionGoldenMaster.test.ts` now disable capture in setup.
- The 2026-09-07 geo commit (8d880cc6fe) changed `ShouldJoinRecordGeoCodes` to require a writable Geo field without updating the GM-VIEW-10 fixtures. They now carry a `GeoAddress` field.

## Tests

- `sql-logging-batch-separator.test.ts`: two real SQL Server drop blocks logged with the caller default land in separate batches, one `DECLARE` per batch; `declaresTSQLVariable` sees through leading comments, ignores routine bodies and PostgreSQL `DO` blocks; explicit separator honored; empty separator emits nothing; plain statements untouched.
- `cd packages/CodeGenLib && pnpm test`: 71 files passed, 6 skipped; 1271 tests passed, 66 skipped.
- CI's deterministic integration tier applies every committed migration, including this one, to a fresh SQL Server 2022 and then runs the suite against it. That tier passes on this branch. It does not run live CodeGen, so "CodeGen stops rewriting the four constraints after the migration" is verified by reading the comparison in `ensureSpecialDateFieldExistsAndHasCorrectDefaultValue`, not by a run.
- **PostgreSQL counterpart deferred to the release build**, per `migrations/CLAUDE.md`: this PR ships the T-SQL migration only. Note for the build engineer: the migration reads `sys.default_constraints` and uses dynamic SQL; on PostgreSQL the IdentityClaim defaults were never `SYSUTCDATETIME()`, so the converted counterpart can be a no-op if the converter cannot transpile it.


rkihm-BC:
**Adversarial self-review (post-push)**

Attacked the change along these lines; findings and what changed:

- **Can a forced `GO` after a `DECLARE` unit break a later unit that reads the variable?** No. Every logged unit goes through `LogSQLAndExecute` → `ds.query(unit)` or a paired `pool.query(sql)`, so nothing can depend on a variable from a previous unit at run time. The only direct `appendToSQLLogFile` callers (`sql_codegen.ts`: two geo `UPDATE`s, one view refresh) log single statements that do not begin with `DECLARE`.
- **Any other top-level `DECLARE @` emitter?** Only `dropDefaultConstraintSQL`. The other three `DECLARE`s in the SQL Server provider are inside procedure bodies, which the guard deliberately does not match (the `CREATE PROCEDURE` header comes first).
- **Doubled `GO`.** A unit that both starts with `DECLARE` and already ends in `GO` would have received a second separator. Fixed in 2nd commit; test added.
- **Migration dynamic SQL.** The drop used hand-bracketing around `@constraintName`. Now `QUOTENAME(@constraintName)`. Idempotency logic re-checked: no constraint → add; wrong definition → drop then add; `(getutcdate())` already → no-op, regardless of constraint name (covers databases where CodeGen already rewrote it).
- **Test-side `configInfo.SQLOutput.enabled = false`.** Vitest here runs with the default per-file module isolation (`vitest.shared.ts` sets no `isolate`/`pool`), so the mutation cannot leak into other files, including the new test that needs capture enabled.
- **Regex safety.** The comment-stripper is `(\s*(block|line comment))*`: each iteration consumes ≥2 chars, so no exponential backtracking.

Not verified here: the migration has not been executed against SQL Server, and the integration tier was not run (no database in the worktree).

rkihm-BC:
**Review addendum (finder agents, commit 3)**

The review's cross-file tracer and diff scanner reported after my first summary. Acted on:

- **Detector was too narrow.** `declaresTSQLVariable` only matched a `DECLARE @` as the first token. T-SQL variables are batch-scoped wherever they appear, so `SET NOCOUNT ON; DECLARE …` and `IF … BEGIN DECLARE … END` were unprotected, and the comment overclaimed. It now matches at the start of any line unless a `CREATE [OR ALTER] PROCEDURE|FUNCTION|TRIGGER` header precedes it. Tests added for both shapes and for a function body.
- **PostgreSQL materialization captures.** Four call sites (`manage-metadata.ts` base-view and query materialization) requested a separator but left it at the hard-coded `GO`, so a PG capture with a materialization would contain `GO` lines. Pre-existing, but on the exact contract this PR changed. They now pass the provider's separator. Changeset updated.
- **Migration definition check** is now parenthesization-insensitive, so `((GETUTCDATE()))` is not needlessly rewritten.
- **Test hygiene:** temp log directory removed after the run; `SQLOutput.enabled` restored after each fixture file; a negative golden-master case pins that `SupportsGeoCoding` without a writable Geo field does not join `vwRecordGeoCodes` (the behavior #4278 introduced, previously untested).

Noted, not changed: the migration hardcodes `DF___mj_…` constraint names while addressing the table through `${flyway:defaultSchema}`. This matches existing migrations, and CodeGen compares only the definition, so a custom core schema gets a differently named but correct constraint.

`cd packages/CodeGenLib && pnpm test`: 71 files passed, 6 skipped; 1274 tests passed, 66 skipped.

rkihm-BC:
**Review addendum 2 (cleanup and altitude angles, commit 4)**

Taken:
- One owner for the test-side capture switch: `SQLLogging.suppressOutputForTests()` returns a restore function; both fixture files use it.
- Stale duplicate JSDoc on `appendToSQLLogFile` removed; the surviving block documents the two replay rules this PR adds.
- New test: unused `os` import and redundant file truncate removed.
- Migration blocks restated as one branch per column: missing or wrong definition → drop if present, add. No `NULL` sentinel.

Considered, not taken:
- **Loop over a table variable instead of four pasted blocks.** Would push the `ADD CONSTRAINT` into dynamic SQL; the pasted form matches repo precedent and stays reviewable.
- **`Dialect.RequiresOwnBatch(sql)` instead of a `@`-sniffing guard in the logger.** Reasonable shape, but a wider change than this fix warrants. The guard is gated on a non-empty separator, so it cannot fire on PostgreSQL text.
- **Migration gate for hand-written `__mj_CreatedAt`/`__mj_UpdatedAt` defaults.** This is the real root cause: V202608202300 hand-authored CodeGen-owned columns with `SYSUTCDATETIME()`, and nothing stops the next migration doing the same. Worth its own PR as a `.github/scripts` gate. Flagging rather than bundling.
- **Shared config-mock factory for the two SQLLogging test files.** Minor duplication; left as is.

`cd packages/CodeGenLib && pnpm test`: 71 files passed, 6 skipped; 1274 tests passed, 66 skipped.

rkihm-BC:
**Adversarial review, second full pass (eight angles: line-by-line, removed behavior, cross-file trace, efficiency, reuse, simplification, altitude, conventions) → commits 6–9**

**Taken**
- **Separator decision computed once** in `appendToSQLLogFile`; the declaration scan runs only on units that could receive a separator, so GO-terminated view and routine bodies (the largest units logged) are never scanned. Detector renamed `declaresBatchScopedVariable`, allocation-free, and judged per GO-delimited batch so a top-level `DECLARE` after a routine's `GO` in the same unit is still caught. `ALTER PROCEDURE|FUNCTION|TRIGGER` counts as a routine header.
- **Platform-derived default separator.** `appendToSQLLogFile` / `LogSQLAndExecute` now default `batchSeparator` from the active platform (`GO` on SQL Server, none on PostgreSQL), so any of the ~70 callers that omit the argument can never write `GO` into a PostgreSQL capture.
- **Migration**: constraint names built from `${flyway:defaultSchema}` so they match CodeGen's `DF_<schema>_<Table>_<column>` on any core schema; the drop uses the same `EXEC(literal + variable)` shape CodeGen emits, with the name quoted in place. Applied to a fresh SQL Server by CI's deterministic integration tier: **pass**.
- **Test switch** `suppressOutputForTests` now also closes an open capture file (so the name holds when `initSQLLogging` ran earlier in the process) and restores both; pinned by a test. Pre-existing null-check ordering in `initSQLLogging` fixed.
- **New test** uses `SQLLogging.sqlOutputDirFlag` for its temp folder instead of a hoisted block with dynamic imports; call-site comment in `dropExistingDefaultConstraint` reduced to a pointer.
- **PR body** corrected: the integration tier has run the migration; PostgreSQL counterpart is deferred to the release build per `migrations/CLAUDE.md`.

**Declined, with reasons**
- *Separator after every unit by default.* Safe by construction, but it reshapes every capture (a `GO` after each INSERT/UPDATE) for a case the detector already covers. Larger blast radius than this fix warrants.
- *Loop over a table variable in the migration.* Pushes the `ADD CONSTRAINT` into dynamic SQL; the pasted per-column form matches repo precedent and stays reviewable.
- *Replace `suppressOutputForTests` with a `captureRequested` flag checked by the #4280 guard.* That changes #4280's guard semantics (a path that forgets `initSQLLogging` would no longer be caught). Left for #4280's author to weigh.
- *camelCase public statics.* The rule says PascalCase, but every existing public static on `SQLLogging` except one is camelCase; renaming two would make the class internally inconsistent.
- *Geo lint comment "is noise".* It is a merge blocker: the Source guards lane fails on next without it.

**Follow-ups worth their own issues (not this PR)**
1. **CodeGen's UTC-default comparison is exact-text** (`manage-metadata.ts` `ensureSpecialDateFieldExistsAndHasCorrectDefaultValue`). Any table whose `__mj_*` default is `SYSUTCDATETIME()`, `GETDATE()` or `SYSDATETIMEOFFSET()` churns on every run. By inspection the same comparison runs on PostgreSQL, where the stored default deparses as `timezone('utc'::text, now())` and can never equal `(NOW() AT TIME ZONE 'UTC')`, which would mean every tracked table churns on every PostgreSQL CodeGen run. Unverified here; worth a dialect-owned "is a UTC-now expression" predicate.
2. **A migration gate for hand-written `__mj_CreatedAt`/`__mj_UpdatedAt` defaults**, the actual root cause (V202608202300 hand-authored CodeGen-owned columns).
3. `sql_codegen.ts` `extractViewSelectBody` builds a regex from the batch separator; with PostgreSQL's empty separator it becomes `\r?\n\b` and would truncate the view body at the first newline. Pre-existing, untouched.
4. The end-to-end #4187 scenario (fresh DB → live `mj codegen` → replay the capture) is not exercised by any CI lane; the deterministic tier applies committed migrations only. Adding a capture-replay step there would have caught this class a year ago.

`cd packages/CodeGenLib && pnpm test`: 71 files passed, 6 skipped; 1276 tests passed, 66 skipped. All 19 CI checks pass on 6564cf2072.

PR #4292, opened 2026-09-08
fix(codegen): EntityField inserts carry an apply-time Sequence again; positional CI gate, wired in

### LINKED ISSUE #4202: CodeGen emits literal EntityField Sequence values (park-shift form), and the CI gate does not catch it
## Summary

`migrations/CLAUDE.md` states that CodeGen already emits the safe form for `EntityField` inserts:

> *"CodeGen now emits the computed form (`manage-metadata.ts`, `getPendingEntityFieldINSERTSQL`), so newly generated blocks are already correct."*

On `v6.1.0-edge.4` it does not. Every `EntityField` insert we captured — **35 across two migrations, zero exceptions** — carried a literal `Sequence`, which is exactly what the same document forbids:

> 🚨 *"A CodeGen `EntityField` INSERT must never carry a LITERAL `Sequence`"*

Separately, the CI gate that exists to catch this **does not match the shape CodeGen actually emits**, so the offending migrations pass CI.

## Repro

1. Empty database. `mj migrate --dir ./migrations`, then `mj codegen` for a migration that creates a new entity. Append the capture verbatim per the workflow guide.
2. Repeat for a second migration that **adds a column to the entity created in step 1** (in our case `ALTER TABLE Animal ADD HousingID`, with a new `Housing` table).
3. Drop and recreate the database, then `mj migrate --dir ./migrations` over the whole chain.

**Actual:**

```
Failed at batch 38/38: Violation of UNIQUE KEY constraint 'UQ_EntityField_EntityID_Sequence'.
Cannot insert duplicate key in object '__mj.EntityField'.
The duplicate key value is (22e4f4de-9a9b-4fe3-ab1a-ecac7ef5ef9d, 100016).
```

This is the failure `migrations/CLAUDE.md` predicts, including its invisibility: *"It cannot fail on a working dev database. It fails only on fresh installs — CI, new developers, releases."*

## What CodeGen emitted

Not the placeholder-band form the docs describe (`MAX + 100000 + ordinal`), but a **park-shift followed by a low literal**:

```sql
/* SQL text to insert 2 new entity field(s) */
UPDATE [__mj].[EntityField]
   SET [Sequence] = [Sequence] + 100000
 WHERE [EntityID] = '22E4F4DE-9A9B-4FE3-AB1A-ECAC7EF5EF9D'
   AND [Sequence] < 100000;

INSERT INTO [__mj].[EntityField] ( [ID], [EntityID], [Sequence], [Name], ... )
VALUES ( '3c9ea97f-...', '22E4F4DE-...', 16, 'HousingID', ... );
```

The tail for that one migration contains **three** park blocks, two of them for the same entity. The first parks the previously-inserted virtual field from `16` to `100016`; the insert then takes `16`; a later park moves that to `100016` too, and the unique key fails.

Counts from the two captures:

| Migration | EntityField inserts | literal `Sequence` | computed form |
|---|---|---|---|
| creates `Animal` + `Breed` | 24 | **24** | 0 |
| adds `Housing`, alters `Animal` | 11 | **11** | 0 |

## The CI gate misses this shape

`.github/scripts/check-migration-entityfield-sequence.sh` matches a 6-digit literal in the placeholder band:

```sh
readonly LITERAL_SEQ_RE='^[[:space:]]*1[0-9]{5},[[:space:]]*$'
```

Its own comment explains the assumption — *"CodeGen never emits a low literal for a NEW field, so this band is the whole population"*. But the park-shift form emits precisely a **low** literal (`16,`), so the gate reports `✓ no literal EntityField sequences in changed migrations` on migrations that then fail on a fresh install.

## Suggested fixes

1. **`getPendingEntityFieldINSERTSQL` should emit the computed form** the doc already specifies, and the park-shift (`parkEntityFieldSequencesSQL`) becomes unnecessary for the appended-migration path:

   ```sql
   (SELECT COALESCE(MAX([Sequence]), 0)
      FROM [${flyway:defaultSchema}].[EntityField]
     WHERE [EntityID] = '<entity-id>') + <schema-ordinal>
   ```

2. **Widen the gate** to flag any bare integer in the `Sequence` position of an `EntityField` insert, not only the `1[0-9]{5}` band.

Worth noting the park form is internally consistent *within one CodeGen run* — `parkEntityFieldSequencesSQL`'s comment says it skips already-parked rows so a second pass does not double-add. That holds live, because `spUpdateExistingEntityFieldsFromSchema` renumbers at the end of the run. It does not hold across two versioned migrations replayed back-to-back, because Flyway runs every versioned migration before any repeatable script — which is the same root cause as #3670.

## Workaround

Converting each captured literal to the documented apply-time expression makes the chain replay cleanly (verified from zero: 69 migrations applied, all entities/views/procs present, CodeGen never run).

## Environment

`@memberjunction/cli` 6.1.0-edge.4, SQL Server 2022, macOS. Found while building a course whose branches must deploy from an empty database.

rkihm-BC:
Fixes #4202

## Three defects, not two

1. **The emitter regressed.** 592b154f4f (2026-08-08) introduced the apply-time `MAX(Sequence)` form. #4048 (2026-08-25, shipped in v6.1.0-edge.4) replaced it with the catalog ordinal as a literal, preceded by a `+100000` "park" `UPDATE` of the entity's existing rows. The park is safe within one CodeGen run, where `spUpdateExistingEntityFieldsFromSchema` renumbers every row moments later, but not across two migrations replayed on a fresh database: Flyway runs every versioned migration before the repeatable renumber, so the second migration's park has nothing reliable to move and its INSERT collides on `UQ_EntityField_EntityID_Sequence`. Without `SET XACT_ABORT ON` the failure reports itself as an unrelated FK error against `EntityFieldValue`. That is the issue's Animal/Housing reproduction exactly.
2. **The gate only matched the six-digit `100000` band** (`1[0-9]{5},`), so the ordinal form passed it.
3. **The gate was never invoked by any workflow.** `migrations/CLAUDE.md` called it a CI gate with a self-test that runs on every PR; neither was true.

## Change

- **Emitter** (`manage-metadata.ts`, `getPendingEntityFieldINSERTSQL`): back to `(SELECT COALESCE(MAX([Sequence]), 0) FROM [EntityField] WHERE [EntityID] = '…') + <schema-ordinal>`. Unique on any database in any order; the live renumber makes the value disposable, so nothing depends on it. The park method and its call are removed. Both dialects (quoting goes through `qi`/`qs`).
- **Gate** rewritten as `.github/scripts/check-migration-entityfield-sequence.mjs`, with the `.sh` kept as a wrapper so the documented entrypoint still works. It parses the INSERT's column list, locates `Sequence`, and flags any bare integer in that position of every `VALUES` tuple: single- or multi-line, either quoting dialect, multi-row `VALUES`, quoted values containing commas and parentheses, `--` and block comments (including the trailing `-- Entity:` the emitter writes after the EntityID value). `EntityFieldValue` and other tables do not match. Changed-files mode reports only **lines the PR adds**, matching the id-determinism gate, because committed migrations carry ~36k legacy literal inserts across 283 files that must stay untouched (Flyway checksums). `--all` scans everything in about a second and is informational. Thirteen self-test fixtures plus a vitest suite that includes a temp-git-repo test of the added-lines scope.
- **CI**: the "Check migrations" workflow runs the self-test on every PR and the scan on PRs that touch `migrations/`. **The scan is blocking**, unlike the advisory id-determinism check next to it: a literal Sequence provably breaks every fresh install and cannot be caught on a dev database. Flagging this as a policy choice for the reviewer.
- **Docs**: `migrations/CLAUDE.md` records the regression, the widened rule, and the CI wiring.
- **Changeset**: `patch` on `@memberjunction/codegen-lib`; no migration or metadata in this branch.

## Not changed

- Existing migrations with literal sequences (the baselines and every capture between 08-25 and now) are left alone, per the documented policy.
- The Open App heal path #4048 added (running the core heal procs after an app migrate) is untouched; it is what makes the emitted value disposable there too.

## Tests

- `cd packages/CodeGenLib && pnpm test`: 71 files passed, 6 skipped; 1277 tests passed, 66 skipped. The sequence test now pins the computed form, the ordinal fallback, and that the batch is INSERTs only with no park `UPDATE`.
- `npx vitest run --config .github/scripts/vitest.config.mts`: 21 files, 530 tests passed.
- `check-migration-entityfield-sequence.sh --self-test`: 13/13.
- Not run: the deterministic integration tier (no database here). It applies committed migrations only and does not run live CodeGen, so it would not exercise the emitter change either way.


rkihm-BC:
**Adversarial review (commits 2–5)**

Eight review angles were launched; one (line-by-line) reported and the other seven stalled without output, so I covered those angles by hand. Findings and what changed:

**Real bugs found and fixed (all in the new gate)**
- **Wrong line attribution.** A hit's line came from the first occurrence of its digits in the tuple, usually inside an earlier UUID. Combined with the added-lines filter, a PR that replaced the computed expression with a literal on an existing INSERT was silently accepted. The parser now returns each item's offset and the hit uses the Sequence item's own line. Fixture and unit test added (digits shared with an earlier UUID, value on a later line).
- **Unbounded `VALUES` search.** An `INSERT … SELECT` into EntityField made the scanner adopt the next `VALUES` anywhere in the file, reading a foreign tuple with this INSERT's column positions (false positive on EntityFieldValue) or skipping the real INSERT after it (false negative). Only a `VALUES` immediately after the column list is inspected now. Fixtures for both directions.
- **Comments between `VALUES` and a tuple, or between tuples,** ended the tuple walk. Both positions now skip whitespace and comments. Fixture added.
- **Entry-point guard failed through a symlink** (real path vs `argv[1]`), so a local run in a symlinked checkout printed nothing and exited 0. Real paths on both sides; test spawns the script through a symlink.
- **Parenthesised integer** `(16)` is now treated as the literal it is.

**In the touched emitter function**
- `isEarlySequence` compared `n.Sequence`, which the pending SELECT computes as `MAX+100000+ordinal`, against `IncludeFirstNFieldsAsDefaultInView`, so it could never be true and that setting was dead. Pre-existing. It now uses the schema ordinal; pinned by a test. **Behavior change**: newly discovered non-key, non-name fields within the first N ordinals get `DefaultInView = 1` as the setting documents.
- My comment (and the doc, and the gate message) claimed the ordinal offset preserves order regardless of execution order. It does not; uniqueness is the guarantee and order comes from sequential execution in emission order, which the pending SELECT's `ORDER BY EntityID, Sequence` provides. Wording corrected in all three places.

**Verified, no change**
- The live renumber (`updateExistingEntityFieldsFromSchema`) runs unscoped on a normal run, so brand-new entities are covered; #4048's stated concern (drifted Sequence on Open App histories) is handled by the heal path it added, which is untouched.
- PostgreSQL: the conflict guard is an `ON CONFLICT DO NOTHING` suffix, and a scalar subquery inside `VALUES` is legal there. Quoting goes through `qi`/`qs`.
- The committed corpus has no `INSERT` without `INTO`, no `WITH (hint)`, and no parenthesised Sequence literal; 283 files carry legacy literals and stay untouched by policy (`--all` now reports that and exits 0, as the doc says it is informational).
- Workflow steps run after "Set up node"; YAML parsed and step order asserted. New scripts are mode 755.
- No stale park/ordinal text remains in CodeGenLib docs.

**Not verified here**: a from-scratch replay of two chained CodeGen captures. CI's deterministic tier applies committed migrations only and does not run live CodeGen, so it does not exercise the emitter change. The same gap was noted on #4291; a capture-replay lane would close it for both.

`cd packages/CodeGenLib && pnpm test`: 71 files passed, 6 skipped; 1278 tests passed, 66 skipped. Scripts suite: 21 files, 532 tests. Gate self-test: 18 fixtures. All 19 CI checks pass on 7f0d646aaf.

rkihm-BC:
**Review addendum (removed-behavior angle, commit 6)**

The removed-behavior angle reported late. Taken:
- **Gate anchored at the repository root.** Pathspecs were cwd-relative, so a run from `packages/CodeGenLib` (which the doc invites) found nothing and exited 0. Test runs the gate from a subdirectory.
- **Gate sees the working tree.** The local form now compares the working tree plus untracked files against the merge base, which is the state right after `cat CodeGen_Run_*.sql >> migration.sql`; the old `--all` used `find`, so this restores what it could see. CI's explicit base/head form is unchanged.
- **Invariants pinned.** The computed form is unique by construction; relative order among a batch depends on the pending SELECT's `ORDER BY EntityID, Sequence` and on the chunk executing sequentially in one round trip. Both are now asserted on both platforms, so a future `Promise.all` over the chunk or a dropped `ORDER BY` fails a test instead of colliding on a fresh install.
- **Stale text.** `migrations/CLAUDE.md` still said the ordinal offset made ordering independent of execution order; the integrity-check warning and two comments still described a `+100000` placeholder. All reworded. The two pending #4048 changesets that describe the park (`open-app-migrate-refresh.md`, `entityfield-sequence-park-idempotent.md`) carry a superseded note so the assembled release notes do not advertise a method that no longer exists.

Counts on this head: CodeGenLib 71 files passed, 6 skipped; 1279 tests passed, 66 skipped. Scripts suite 21 files, 533 tests. Gate self-test 18 fixtures.

rkihm-BC:
**Review addendum (reuse, simplification, efficiency, altitude, cross-file angles, commit 7)**

The remaining angles reported late. Consolidated into one round:

**Emitter**
- **Two more emitters wrote literals** (altitude): virtual-entity fields and IS-A parent fields resolved a Sequence at run time via `nextAvailableEntityFieldSequence` and wrote it into the capture. The blocking gate would have rejected CodeGen's own output. All three emitters now share `applyTimeEntityFieldSequenceSQL`; the run-time lookup is removed.
- **`+ <ordinal>` dropped for `+ 1`** (simplification): the offset carried no ordering meaning, so one documented form is simpler to teach and gate. Provider comments corrected.

**Gate**
- **ENOBUFS on large migrations** (reuse, efficiency): `git()` used Node's default 1 MB buffer; 56 committed migrations exceed it. Now `-C <root> -c core.quotePath=false` with a 256 MB buffer like the sibling guards, and diffs are computed only for files with hits, never for added files. Temp-repo test with a >1 MB migration.
- **Scope** (altitude, cross-file): Flyway versioned files only; baselines (`B*`) are literal by construction and `tests/` fixtures never run. Renamed migrations, which a timestamp bump produces, are included now.
- **One masker instead of three scanners** (reuse, simplification): the sibling `stripSqlComments` handles comments, nested comments, string literals and `]]`; `parseParenList` works on masked text and records item offsets.
- **Fixtures exported**, vitest iterates them; the `.sh` shim is deleted and every reference points at the `.mjs`; `test.yml` paths include the gate so a script-only edit runs its suite.

**Tests**: the CodeGen test no longer overrides a private method (it captured only because vitest strips types); it records through a typed connection and pins the single sequential round trip.

**Declined / follow-ups**
- `SET XACT_ABORT ON` at the top of every capture would make future collisions report honestly; a real improvement but a change to every capture's shape, so a separate PR.
- Dropping the pending SELECT's dead `MAX+100000+ordinal` column (the trap that misled #4048) touches both providers' SQL; noted as a follow-up.
- Running `scanContent` inside CodeGen at capture time would make the emitter self-verifying in every repo, not just this one's CI. Follow-up.
- `npm run check:claude-md` reports root `CLAUDE.md` pointing at a missing `../CLAUDE.md`; that is pre-existing on next.

Counts: CodeGenLib 71 files passed, 6 skipped; 1278 tests passed, 66 skipped. Scripts suite 21 files, 552 tests. Gate self-test 22 fixtures.

PR #4293, opened 2026-09-08
fix(server): allow ExtraFilter IN (SELECT … FROM entity BaseView)

AN-BC:
## Summary
- [#4253](https://github.com/MemberJunction/MJ/pull/4253) banned `SELECT`/`EXISTS` in GraphQL ExtraFilter by keyword. That broke Committees Command Center and any Open App using `IN (SELECT … FROM <view>)`.
- Keyword `SELECT` is not stacked DML. The leak is a subquery against a **base table** or catalog (`EXISTS (SELECT 1 FROM __mj.User …)`).
- This PR keeps the GraphQL-boundary screen but uses `@memberjunction/sql-parser` (same wrap as EDS `assertReadOnlyClause`): parse as a single read, reject nested writes, allow FROM only when it is an entity **BaseView**.
- `SQLParser.walkExpression` now walks `expr.value` so `IN (SELECT …)` tables are extracted (EXISTS already was).
- Does **not** revert the rest of #4253 (CustomWhereClause Owner-only, Conditional `new Function`, iframe, Gmail).
- RLS is still applied by RunView as an outer WHERE, not compiled into the view. Subqueries against another entity’s BaseView do not inherit that entity’s RLS; they are still restricted to the view, not the table.

## Test plan
- [ ] `pnpm exec vitest run src/__tests__/ResolverBase.filterEscaping.test.ts` in MJServer (18 tests, includes restored Committees ExtraFilter shapes)
- [ ] `pnpm exec vitest run src/__tests__/sql-parser.test.ts` in SQLParser
- [ ] Committee Management Command Center loads after this + Committees PR (SELECT IN against `vwTasks` / `vwMeetings` / …)
- [ ] `EXISTS (SELECT 1 FROM __mj.[User] …)` ExtraFilter still fails at GraphQL

### DIFF
--- a/.changeset/graphql-extrafilter-baseview-ast.md
+++ b/.changeset/graphql-extrafilter-baseview-ast.md
@@ -1,0 +1,6 @@
+---
+"@memberjunction/server": patch
+"@memberjunction/sql-parser": patch
+---
+
+Replace the GraphQL ExtraFilter SELECT/EXISTS keyword ban with an AST screen that allows `IN (SELECT … FROM <entity BaseView>)` and still rejects base tables (`__mj.User`). Uses `@memberjunction/sql-parser` (same wrap as EDS `assertReadOnlyClause`). SQLParser now walks `expr.value` so `IN (SELECT …)` subqueries are visible to ExtractTableRefs.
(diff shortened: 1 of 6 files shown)


AN-BC (commented):
@jordanfanapour @rkihm-BC we had a recent PR merge to `next` that blocked subqueries entirely in client-provided `ExtraFilter`. This PR changes it to allow some through while still hardening things per the PR that just blocked it. I am using the SQL parser here @jordanfanapour  so flagging for you and @rkihm-BC  would like you to review this and raise a follow on PR if needed.

Issue #529, opened 2024-09-26
Zapier Integration for MemberJunction


AN-BC:
The idea is to have two-way connectivity with Zapier to enable a wide array of workflows and automation in and out of MJ.

**Outbound**
From MJ to Zapier will require MJ to invoke Zapier actions. To do this, instead of creating a super generic way of calling any Zap at run time, one solution would be to allow a user to easily generate `Actions` in MJ that encapsulate a specific action that a given Zapier integration supports. For example let's say we wanted to have an MJ Action that would create a Customer in Shopify. There is a Zapier integration for Shopify that supports an action that does just this. https://zapier.com/apps/shopify/integrations 

We could build a tool that will automatically generate an MJ `Action` that very simply wraps the Zapier integration. We could create a generic Zapier base class to abstract their API a little and then generate the code for the Action to map in the Zapier properties and response codes to MJ Action metadata of the same type. 

**Inbound**
We could expose various parts of MJ functionality as Zapier Actions and Zapier Triggers. For example, we could expose entity CRUD operations as Zapier actions - expose MJ Actions as Zapier actions, and on the trigger side, send notifications that turn into Zapier triggers when things happen in MJ in various places. 

This part is a little more complex. Since MJ is not a SaaS solution, each customer would have to register their integration with Zapier which is possible but might not be ideal. An alternative would be for MJ.com to create a managed service that is a middleware that can handle this automatically between Zapier and any MJ client and use something like `webhooks` to signal back to a given MJ instance in a secure way. We'd have to think through this carefully to ensure we are not handling sensitive data or not persisting it, etc.

AN-BC:
@cadam11 would be great to have your input on this one

cadam11:
I'll have to spend some time to re-acquaint myself with Zapier's domain language

cadam11:
A good starting place would be to try to build a minimal integration using API key authentication to keep it simple. I'm working on a POC to see if we can collect all the installation-specific details as variables for the integration so it can be used with any deployment.

cadam11:
So far, I've learned that the Zapier UI is inadequate for our needs– I think the CLI integration builder could work and will continue with that approach to learn more.

PR #4299, opened 2026-09-08
feat(agents): let an agent direct artifact creation (Skip-Brain #529)

CaelebB-BC:
## Why

`ProcessAgentArtifacts` decides artifact versioning last, from continuity signals alone: an explicit `sourceArtifactId`, else the previous artifact on the conversation detail. Both say "this conversation already has an artifact." Neither can tell a restyle of the current component from a request for a different one.

The result (BlueCypress/Skip-Brain#529): a new component asked for mid-conversation is saved as **version N of whatever came before it** — same artifact, new name, new code.

The agent knows which of those the user asked for. This PR gives it a way to say so.

## What changes

A new optional per-step `ArtifactDirective` (`@memberjunction/ai-core-plus`), carried on `BaseAgentNextStep` and `ExecuteAgentResult`:

| behavior | effect |
|---|---|
| `create-new` | new artifact at v1, `sourceArtifactId` ignored |
| `version-source` | next version of `targetArtifactId`, else of `sourceArtifactId` |
| `suppress` | write nothing |

**Precedence is deliberately narrow.** The directive is consulted only *after* the two existing vetoes, so it can never widen what a caller or an agent's configuration already refused:

1. `createArtifacts === false` → nothing. A directive cannot re-enable creation.
2. `ArtifactCreationMode: 'Never'` → nothing.
3. The directive.
4. No directive → the historical chain, byte-for-byte unchanged.

**A directive-named target is model output**, and it lands in the `ExtraFilter` fragments built by `GetMaxVersionForArtifact` / `CheckForDuplicateVersion`. So it must parse as a UUID *and* load as a readable `MJ: Artifacts` row; either check failing logs and drops the run back to the historical chain, exactly as if no target had been named. These checks run only for a directive-supplied id — a caller's `sourceArtifactId` reaches those filters exactly as it did before.

Supporting changes:

- **`artifact-target-plan.ts`** — `planArtifactTarget()` extracted from the writer so the precedence table above is testable without a database.
- **Explorer artifact panel** — stopped guessing which artifact to open. It snapshots versions before the turn and diffs after, with *created* beating *bumped*, so a new artifact wins the panel over one that merely gained a version. Dead `targetArtifactVersionId` plumbing removed (−89).
- **`PayloadManager`** — unrelated defect found while testing the above. When a sub-agent returns a **shorter array**, deleting the scalars under the vacated index leaves the containers that held them: a data-free shell that the old `Object.keys().length === 0` check still read as populated. Emptiness is now judged recursively. Before this, any modification that removed a child component crashed at the last step of the pipeline.

## Verification

- `packages/AI/Agents`: **1798 tests across 93 files**, all passing.
- New coverage: `artifact-target-plan.test.ts` (precedence), `agent-runner-artifacts.test.ts` (+219, veto interaction and the invalid-target fallbacks), `payload-manager.test.ts` (+59), `artifact-panel-action.test.ts` (created-beats-bumped).
- Verified end to end against a live three-repo stack: three components in one conversation, thirteen versions, each request landing on its own artifact.

## What reviewers should look at

The load-bearing claim is **backward compatibility**: with no directive, `ProcessAgentArtifacts` must behave exactly as before. That was checked statement-by-statement against the base by two reviewers, but it is the thing worth a third pair of eyes.

## Merge order

This PR goes **first** and is safe alone — without a directive sender it is inert. The other two repos depend on it:

1. **this PR** → `lts/5`, then release **5.52.0** (the version `@askskip/server`'s changeset promises)
2. BlueCypress/Skip-Client-Open-App#`feat/529-artifact-request-directive` → publish `@askskip/*`
3. BlueCypress/Skip-Brain#`CB-529-build-vs-modify` → deploy + `mj sync push`

**Base branch note:** opened against `lts/5`, where the code applies today. The port to `next` is a genuine port rather than a rebase — `next` is 2221 commits ahead and has independently rewritten every file this branch touches (`base-agent.ts` +713/−132, `message-input.component.ts` +427/−473, `conversation-chat-area.component.ts` +580/−101) — and `@since 5.52.0` annotations will need to become the 6.x version. Tracked separately.

### DIFF
--- a/packages/AI/Agents/src/AgentRunner.ts
+++ b/packages/AI/Agents/src/AgentRunner.ts
@@ -12,9 +12,10 @@
 
 import { createHash } from 'crypto';
 import { LogError, LogStatusEx, IsVerboseLoggingEnabled, LogStatus, Metadata, RunView, RunQuery, UserInfo, IMetadataProvider, DatabaseProviderBase, ProviderType } from '@memberjunction/core';
-import { MJGlobal, UUIDsEqual } from '@memberjunction/global';
+import { MJGlobal, UUIDsEqual, IsValidUUID } from '@memberjunction/global';
 import { AIEngine } from '@memberjunction/aiengine';
-import { ExecuteAgentResult, ExecuteAgentParams, MediaOutput, FileOutputRef, InputArtifact } from '@memberjunction/ai-core-plus';
+import { ExecuteAgentResult, ExecuteAgentParams, MediaOutput, FileOutputRef, InputArtifact, ArtifactDirective } from '@memberjunction/ai-core-plus';
+import { planArtifactTarget } from './artifact-target-plan';
 import { BaseAgent } from './base-agent';
 import { MJConversationEntity, MJConversationDetailEntity, MJArtifactEntity, MJArtifactVersionEntity, MJConversationDetailArtifactEntity, MJAIAgentRunMediaEntity, ArtifactMetadataEngine, ExtractBase64FromDataUrl, DecideInlineStorage } from '@memberjunction/core-entities';
 import { FileStorageEngine } from '@memberjunction/storage';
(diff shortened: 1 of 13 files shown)


CaelebB-BC:
Superseded by #4300, which targets `next`. The 5.x path is being abandoned in favour of the 6.x line.

Two things changed in the port, both recorded on #4300: the `message-input.component.ts` change is dropped (`targetArtifactVersionId` is dead plumbing on `lts/5` but live on `next`, where the Check Sage Intent prompt populates it), and the `@since` annotations became `6.1.0`. The port also carries a changeset, which this PR was missing.

Closing unmerged — no code is lost, #4300 contains all of it.

PR #4300, opened 2026-09-08
feat(agents): let an agent direct artifact creation (Skip-Brain #529)

CaelebB-BC:
## Why

`ProcessAgentArtifacts` decides artifact versioning last, from continuity signals alone: an explicit `sourceArtifactId`, else the previous artifact on the conversation detail. Both say "this conversation already has an artifact." Neither can tell a restyle of the current component from a request for a different one.

The result (BlueCypress/Skip-Brain#529): a deliverable produced mid-conversation is saved as **version N of whatever came before it** — same artifact, new name, new content.

The agent knows which of those the user asked for. This PR gives it a way to say so.

## What changes

A new optional per-step `ArtifactDirective` (`@memberjunction/ai-core-plus`), carried on `BaseAgentNextStep` and `ExecuteAgentResult`:

| behavior | effect |
|---|---|
| `create-new` | new artifact at v1, `sourceArtifactId` ignored |
| `version-source` | next version of `targetArtifactId`, else of `sourceArtifactId` |
| `suppress` | write nothing |

**Precedence is deliberately narrow.** The directive is advice from an agent, not a command, and is consulted only *after* the checks that already existed:

1. `createArtifacts === false` → nothing. A directive cannot re-enable creation.
2. `ArtifactCreationMode: 'Never'` → nothing.
3. The directive.
4. No directive → the historical chain, **byte-for-byte unchanged**.

`suppress` covers everything the step would persist as an artifact: the payload, and the artifacts wrapping any generated files or media. The run's media audit rows are still written — suppression governs what the user is shown, not lineage.

## Every field of a directive is model output

A directive arrives as parsed JSON from an LLM, so no field can be taken on trust — not the behavior, not the id, not the free text. Nothing here is a hypothetical: each guard below exists because the unguarded version had a concrete failure.

**The named target must be one this user may write to.** `vwArtifacts` carries no per-user predicate and no row-level-security filter, and nothing on the write path checked ownership, so a row resolving proved only that it *exists*. An agent could name any artifact id in the instance and have the run's payload appended to it as a new version. A directive-named `targetArtifactId` is now honored only if it is a UUID-shaped string naming an artifact that exists **and** that the run's user either owns or holds an explicit `CanEdit` grant on. Existence and authorization resolve in one `RunViews` round trip — deliberately not `BaseEntity.Load`, which **throws** on a permission denial, a SQL conversion error or any transient fault, and a throw on this path reached the method-wide catch and ended the run with no artifact, no version and no link. Anything failing logs and falls back down the ladder to the caller's `sourceArtifactId`, then to the historical chain.

**Provenance is carried, not inferred.** The plan records `source: 'directive' | 'caller'` rather than comparing the planned id against `directive.targetArtifactId`. Those two values can legitimately be *equal* — an agent echoing the run's own source id back as its target is natural model behavior — and the comparison then treated a caller-supplied id as model output and, worse, let an id the guards had just rejected reappear through the fallback and be versioned anyway, logged as both "not readable" and "creating version N". Rejected ids are remembered for the run so they cannot re-enter on a lower rung.

**Ids are trimmed before use.** `IsValidUUID` tolerates surrounding whitespace, so a model writing the id on its own line passed validation and the raw value then reached `ArtifactID='<uuid>\n'` as SQL Server error 8169.

**Free text is coerced, not trusted.** A non-string `name` threw at `.trim()`; a title over 255 characters failed the column's `MaxLength`, made `Save()` return false and discarded the entire artifact, leaving the deliverable reachable only through `AIAgentRun.FinalPayload`. Both fields are now type-checked and trimmed, and `name` is clamped to its column rather than rejected. Log lines describe untrusted values without asking them to render themselves, since `String({ toString: 'x' })` throws — inside the very guard reporting the rejection.

**An unrecognized `behavior` resolves exactly like no directive.** It deliberately does not fall through to the legacy chain: "legacy" means *previous artifact on this message, else a new one*, and on a fresh agent-response detail there is no previous artifact, so a typo in the wire value was **stricter** than sending nothing — it silently started a new artifact where the run's continuity signal said to add a version. The mismatch is logged.

**Ids are escaped where the filter is built.** `GetMaxVersionForArtifact`, `CheckForDuplicateVersion` and `FindPreviousArtifactForMessage` interpolate ids into raw `ExtraFilter` fragments. `RunViewParams.ExtraFilter` has no parameterized form and the upstream `ValidateUserProvidedSQLClause` blacklists statement keywords but permits `OR`, so this was a real predicate-injection surface. Escaping at the builder covers every caller, present and future — including the `sourceArtifactId` that arrives straight from the GraphQL boundary.

**A directive governs the payload of the agent that issued it, and never crosses the parent/child boundary.** Both sub-agent success mappings spread `...subAgentResult`, so the parent's terminal step inherited the child's directive — and what the parent returns there is `mergedPayload`, which would have been written onto an artifact the *child* named, silently and with no parent opt-out. That is now dropped explicitly and logged, matching the Chat mappings that never carried it. Conversely, the two places that rebuild an agent's **own** terminal step — the client-tools `terminateAfterExecution` branch and `executeChatStep` — now carry the directive instead of dropping it.

## Supporting changes

- **`artifact-target-plan.ts`** — `planArtifactTarget()` extracted from the writer so the precedence table above is testable without a database. Pure: the runner does the I/O and the logging.

- **`PayloadManager` no longer leaves data-free shells in arrays.** When a sub-agent returns a **shorter** array, deleting the scalars under the vacated index leaves the containers that held them: a data-free shell that an `Object.keys().length === 0` cleanup reads as populated. Emptiness now judged recursively — but only for the positions **this merge actually vacated**. That scope is the whole point. "No primitive content anywhere inside" describes plenty of good data (a query-result row whose columns all came back null, an element whose only fields are an empty `tags: []` and an empty `config: {}`), and applying it to the whole merged payload deleted such rows silently, including rows under paths the sub-agent was never granted and rows whose deletion the guardrail had just recorded as **blocked**. The global key-less-`{}` sweep is unchanged, so an untouched payload still passes through a merge byte-identically. The scoped pass also closes the hole `_.unset` leaves when a scalar array is shortened, and prunes a fully vacated *nested* array — a latent defect that predates this branch and reproduces identically on the base commit.

- **`ng-conversations` stops guessing which artifact to open.** The chat area snapshots artifact versions before a turn and diffs after, with *created* beating *bumped*, so a newly created artifact wins the panel over one that merely gained a version — including when the panel is already open on something else, which the old `!showArtifactPanel` gate suppressed. Removing that gate makes the decision load-bearing, and "absent from the before snapshot" is not on its own evidence that a run created anything: `resetConversationScopedViewState` does not clear `artifactsByDetailId` on a conversation switch, `loadMessages` runs the missed-completion handler *before* the first `loadPeripheralData`, and paging up merges older messages' artifacts into the same map mid-turn. The baseline therefore also records which conversation the map was holding and a paging-merge counter; when those do not line up, creations are not inferred at all and only the shown artifact's own bump is acted on. A selection the user makes mid-turn is no longer discarded either — two completion handlers run per turn, each holding its own pre-turn snapshot — though a closed panel displaces nothing, so a genuine build still surfaces. A creation is chosen by the newest version's timestamp rather than by whichever conversation detail the map happened to iterate last.

- **`ng-artifacts`: the viewer no longer loads twice.** A refresh drove both the `[versionNumber]` input and the refresh subject; an open changed `artifactId` and `versionNumber` in one change-detection pass, where two independent `ngOnChanges` branches each ran a full load — roughly six redundant round trips including a second full content download, the second without a cancellation token. One channel each now. Its refresh guard also compares ids as `UUIDsEqual`: SQL Server returns UUIDs upper-case and PostgreSQL lower-case, so the raw `===` silently dropped legitimate refreshes. `selectDistinctLatestArtifacts` had the same raw comparison and could render one artifact as two cards.

## Behavior changes worth knowing

Two, both narrow, both deliberate:

- A **caller-supplied** `sourceArtifactId` that is not UUID-shaped now falls back to the historical chain instead of reaching a query. In production that value is always a real artifact id or absent, so this only changes runs that were already broken.
- A directive-named artifact **owned by another user with no explicit `CanEdit` grant** produces a new artifact rather than a version. This matters only for shared conversations, and the fallback is the pre-directive outcome.

Still open, and unchanged by this PR: any authenticated caller can name any artifact id through `sourceArtifactId`. Escaping closes the injection route; the ownership test on that path changes behavior for every existing agent and wants its own change.

## Verification

Built and tested on this branch, against `next`:

- `pnpm` + `turbo build --filter="@memberjunction/ai-agents..." --filter="@memberjunction/ng-conversations..." --filter="@memberjunction/ng-artifacts..."` → **116/116 packages, 0 errors**. The Angular compiles are the real check on the ported call sites, since `next` had rewritten the chat area (+580/−101).
- `@memberjunction/ai-agents`: **1998 tests / 110 files**, all passing.
- `@memberjunction/ng-conversations`: **1280 tests / 107 files**, all passing.
- `node .github/scripts/check-changeset-bump.mjs --base origin/next` → passes.
- Coverage on the new surface: `artifact-target-plan.test.ts` (22 — the full decision table, provenance, every shape of unrecognized behavior, and non-string targets), `agent-runner-artifacts.test.ts` (41 — authorization granted and refused, a throwing lookup, an echoed source id, whitespace trimming, over-long and non-string free text, a non-callable `toString`, the escaping of each of the three query builders, the whole-directive drop on an unparseable `behavior`, and the non-numeric version guard), `payload-manager.test.ts` (45 — the shell case this fixes, plus the untouched all-null record, the blocked-path record and the nested array), `artifact-panel-action.test.ts` (16 — an untrustworthy baseline, a mid-turn user selection, and recency ordering).
- The payload-manager suite was run against all three trees to pin the regression: **5 of 45 fail** on the unscoped sweep and **2 of 45** on the base commit. Only the current tree passes all 45, the PR's own shell-residue case included.
- The mechanism was verified end to end on a live three-repo stack (on the 5.x build): three components in one conversation, thirteen versions, each request landing on its own artifact.

## What reviewers should look at

The load-bearing claim is still **backward compatibility**: with no directive, `ProcessAgentArtifacts` must behave exactly as before. Verified statement-by-statement against the base, and re-checked after the hardening above, which is careful to leave the caller path's round-trip count and query shapes alone. The two documented behavior changes are the only intended departures.

Second: whether dropping a sub-agent's directive is the contract you want. It is the consistent choice — a directive describes its own agent's payload, and `RunAgent` does not run the artifact writer for sub-agents at all — but the alternative (a delegating parent inheriting its child's intent) is defensible, and switching it is a one-line change at two sites plus the log.

Third: whether keeping `targetArtifactVersionId` alongside the directive is right on this line. They operate at different layers — client-side payload continuity versus the server-side artifact write target — so they coexist, but that's a judgment worth confirming.

One fix has no test: honoring `suppress` for file and media artifacts is a guard inside `AgentRunner.RunAgentInConversation` (lines 535-620, alongside `processFileArtifacts` and `saveMedia`), which is not unit-testable without substantial new scaffolding. Everything else above is covered.

**`suppress` ships unexercised.** Nothing in MJ emits an `artifactDirective` today, and Skip emits only `create-new` and `version-source` — so a third of the public surface has no producer anywhere. It is kept because the case is real and the guard is small, but nobody should read it as proven behavior. (Raised by @jordanfanapour.)

## Review follow-ups

- `08c5b2ac89` — an unparseable `behavior` drops the whole directive (`name`/`description` included); `Number.isFinite` guard in `CheckForDuplicateVersion`; `VetArtifactVersionTarget` records why it does not delegate to `ArtifactPermissionProvider`; the `rejectedIds` interaction with the caller path is documented as deliberate.
- `ef795dd428` — `planArtifactTarget` discards a non-string `targetArtifactId` at the boundary; the runner logs the discarded value.
- `5ff5cf44c5` — the delegation comment names `PermissionEngine`'s actual callers (the files cited as server-side callers use `ResourcePermissionEngine`, a different class) and records the `CollectionPermissionProvider` owner-as-grantee precedent.

## Merge order

This PR goes **first**; it is inert without a directive sender.

1. **this PR** → `next`
2. BlueCypress/Skip-Client-Open-App#27 → publish `@askskip/*`
3. BlueCypress/Skip-Brain#537 → deploy + `mj sync push`

A client MJAPI needs this release **and** the `@askskip/*` release before any of it is visible.


jordanfanapour (commented):
Reviewed as part of the three-PR chain (this, Skip-Client-Open-App#27, BlueCypress/Skip-Brain#537). The design is right: `sourceArtifactId` is a continuity signal, not an intent signal, and there was no way for an agent to express the difference. Putting the directive *after* the two existing vetoes, and making "no directive" byte-identical to today, is the correct shape for this.

I have one contract inconsistency, some notes on `suppress`, an answer to one of the questions you raised for reviewers, and a request.

---

## Please get a second review from someone deeper in this part of the framework

@CaelebB-BC — could you request review from **EL-BC** (Ethan) or **AN-BC** (Amith) before this merges? The load-bearing claim is backward compatibility of `ProcessAgentArtifacts`, and that plus the `ng-conversations` chat-area rework really wants someone who knows the artifact write path cold. Ethan first for the write path; Amith if there's any disagreement about whether the directive belongs on `BaseAgentNextStep` at all, since that's a framework-contract call.

## An unrecognized `behavior` still applies the directive's `name`/`description`

The type doc, the changeset, and the log line all say an unrecognized behavior is treated as no directive at all. That's true for *targeting* — `planArtifactTarget` correctly ignores it — but `createArtifactHeader` receives the whole raw directive regardless:

```ts
artifact.Name = AgentRunner.SafeDirectiveText(directive?.name, 255) || `${agentName} Payload - ...`;
```

So `{ behavior: "createNew", name: "Sales Dashboard" }` still names the artifact "Sales Dashboard". And there's a second effect — the same value feeds

```ts
if (isNewArtifact && newVersionNumber === 1 && !SafeDirectiveText(directive?.name, 255))
```

so that garbled directive also *suppresses* the extracted-name fallback. Two behaviors change under a directive the code claims to be ignoring.

Practical impact is mild and arguably the friendlier outcome, so my preference is the one-word doc fix: say the *targeting* is treated as no directive, rather than the directive as a whole. Either way it shouldn't stay as-is, since a reader trusting the current wording would be wrong.

## `suppress`

Two notes, neither blocking:

- **No producers anywhere.** Nothing in MJ emits an `artifactDirective`, and Skip only ever emits `create-new` / `version-source`. So a third of the public surface ships unexercised. I'd still keep it — it's ~8 lines of guard and the case is obviously real — but worth stating plainly in the PR body so nobody later assumes it's proven.
- **The PR body puts the guard in `RunAgentInConversation`.** It's actually in `AgentRunner.RunAgent`, alongside `processFileArtifacts`/`saveMedia`. Worth correcting; a wrong location in the description will send the next person hunting.

## On "whether dropping a sub-agent's directive is the contract you want"

I traced this — **keep it as is.** Skip can never be a sub-agent: MJ constrains `ParentID` and `ExposeAsAction` to be mutually exclusive, and Skip is registered as a root agent with `ExposeAsAction = 1`. Sage reaches it via the Execute Agent action → `AgentRunner.RunAgent`, a fresh top-level run where the artifact writer *does* run with Skip's own directive. So no Skip directive is ever dropped at a parent boundary, and the conservative contract costs nothing today.

One consequence worth flagging on the Skip side (raised on BlueCypress/Skip-Brain#537): that same Execute Agent path calls `RunAgent` with **no `sourceArtifactId`**, so an absent directive there falls through to a new artifact rather than to versioning. Fine for MJ; it just means Skip's "emit nothing and let the legacy chain handle it" fallback is weaker than it reads.

## Minor

`CheckForDuplicateVersion` now interpolates `VersionNumber=${Number(latestVersionNumber)}`, which yields `VersionNumber=NaN` for a non-numeric input. Callers are all internal so it can't fire today, but a `Number.isFinite` guard would match the care taken on the id path two lines up.

---

## Checked and found correct

- The vetting ladder: `rejectedIds` is keyed lowercase and consulted *before* the `source !== 'directive'` early return, so the echoed-source-id re-admission the tests describe genuinely cannot happen.
- `EscapeSQLString` does quote-doubling and null-guards; `IsValidUUID` trims internally, so the separate `.trim()` before use is necessary rather than redundant; `MJEnvironmentEntityExtended.DefaultEnvironmentID` is exactly the GUID it replaces.
- `pruneVacatedArrayElements` ordering is sound — deepest-container-first prevents a shallow splice invalidating a still-queued deeper path, and highest-index-first within a container keeps queued indices valid. The scoping argument for why `isVacantValue` can't be applied globally is correct, and the note about the three non-interchangeable emptiness tests is genuinely useful to leave behind.
- The `return` added to `artifact-viewer-panel.component.ts` `ngOnChanges` is safe — nothing follows the `versionNumber` branch in that method.
- The Angular work is necessary, not scope creep: without removing the `!this.showArtifactPanel` gate, a correctly-created second artifact would never surface while the panel sat open on the first one, and the fix would be invisible to users.

The one piece I'd genuinely consider splitting out is `PayloadManager` (~334 lines with its tests). It's a real bug and the fix looks right, but it's a sub-agent array-merge defect with no relationship to artifact targeting, it has its own self-contained test suite, and pulling it would shrink the surface a second reviewer has to hold in their head. Your call — not blocking.

rkihm-BC (commented):
Reviewed the code rather than the description, and read @jordanfanapour's pass first so I could add to it rather than repeat it. I agree with his conclusions — including the `behavior`/`name` inconsistency and the `Number.isFinite` nit — so I won't restate them.

**One finding neither of us raised, and it is the one I'd want answered before merge.** Everything else here is good.

## What's good

**The threat model is the right one, and it is applied consistently.** "A directive arrives as parsed JSON from an LLM, so no field can be taken on trust" is the correct premise, and every guard traces to a concrete failure rather than a hypothetical — the `\n` reaching SQL Server as error 8169, `String({toString:'x'})` throwing inside the guard that was reporting a rejection, a 255-overflow silently discarding the whole artifact. That last one is the kind of defect that only shows up as "the deliverable is missing", and finding it before shipping is worth more than the guard.

**Carrying provenance instead of re-deriving it is the right call and the reasoning is exactly right.** `source: 'directive' | 'caller'` rather than comparing the planned id against `directive.targetArtifactId` — because those can legitimately be equal, and a value comparison then treats a caller-supplied id as model output *and* lets a just-rejected id re-enter through the fallback. That is a subtle failure and the fix is structural rather than defensive.

**Failing closed on the lookup.** `catch` → `reject(...)` with the comment "an unavailable lookup must not become an unchecked write" is correct, and choosing `RunViews` over `BaseEntity.Load` specifically because `Load` throws on a permission denial — and a throw here ends the run with no artifact at all — is a real insight about the difference between "denied" and "broken".

**The unrecognized-behavior semantics.** Resolving to *no directive* rather than to `legacy` is right for the reason stated: on a fresh agent-response detail there is no previous artifact, so falling to `legacy` would make a typo **stricter** than sending nothing. That is the kind of asymmetry that normally ships unnoticed.

**`planArtifactTarget` as a pure function** with the runner doing I/O and logging is the correct seam, and it makes the precedence table testable without a database.

## The finding: this hand-rolls a permission check the platform already owns

`VetArtifactVersionTarget` answers "may this user add a version to this artifact" with its own two-query test:

```ts
{ EntityName: 'MJ: Artifacts',            ExtraFilter: `ID='...'`, Fields: ['ID','UserID'] },
{ EntityName: 'MJ: Artifact Permissions', ExtraFilter: `ArtifactID='...' AND UserID='...' AND CanEdit=1` }
```

That question already has an owner. `packages/MJCoreEntities/src/custom/PermissionProviders/ArtifactPermissionProvider.ts` is a registered `PermissionProviderBase` whose `GetResourceTypes()` returns `['Artifacts']`, whose `SupportedActions` include `'Update'`, and whose `CheckPermission(user, 'Artifacts', artifactId, 'Update')` resolves precisely this. And `guides/UNIFIED_PERMISSIONS_GUIDE.md` §1 routes this class of question explicitly:

> **Per-resource access / sharing** — "Can this user *use / view / edit / own* this specific record?" → **Permission Providers → the PermissionEngine**

So this is now the **third** answer to "can this user edit this artifact" in the tree, alongside the provider and `ng-conversations/services/artifact-permission.service.ts`. The failure mode is drift: when artifact sharing grows a grant shape — and `SupportsDeny = false` and `SupportedGranteeTypes = ['User']` both read like present-tense limitations rather than permanent ones — the provider gets updated and this copy does not. It then silently rejects legitimate editors and falls back to creating a new artifact, which is the safe direction but the wrong outcome, and nothing fails loudly.

**To be fair to the code as written, a straight swap is not available**, and this is the part worth deciding rather than me prescribing: the provider's `CheckPermission` consults permission *rows* only, and its own comment at line 117 flags the "owner without `SharedByUserID`" case as a wrinkle in that table. Your explicit `UUIDsEqual(owner.UserID, contextUser.ID)` check covers an owner who has no row, which the provider may not. So the shapes genuinely differ.

What I would like is for that to be a stated decision rather than an accident. Either compose — keep the owner check, delegate the grant half to `CheckPermission` — or record in `VetArtifactVersionTarget`'s doc comment why the provider is not usable here, so the next person to touch artifact sharing knows to update two places. Right now a reader has no way to tell the divergence was considered.

I'd also fold this into the request @jordanfanapour already made: this is exactly the kind of question **@EL-BC** or **@AN-BC** should settle, since it is a framework-contract call about where permission logic lives, not a detail of this feature.

## Smaller

**`rejectedIds` makes the caller path conditionally stricter.** The set is consulted *before* the `plan.source !== 'directive'` early return — correct, and @jordanfanapour verified the ordering closes the re-admission hole. But the consequence is that a caller-supplied `sourceArtifactId` gets ownership-enforced **only when the agent happened to name the same id first and fail**. The PR is explicit that the caller path is deliberately not ownership-checked yet ("wants its own change"), so this is one path where that future change has already half-arrived. It errs safe and I would not change it — worth a sentence in the doc comment so it reads as intended rather than incidental.

**A non-string truthy `targetArtifactId` reaches the plan.** `directive.targetArtifactId ? { kind: 'version', artifactId: ... }` admits a number or an object, since only truthiness is tested. The runner catches it (`typeof plan.artifactId !== 'string'`) and the type comment on `ArtifactTargetPlan.artifactId` says so explicitly, so this is documented rather than missed — but the falsy check is the one place in the chain where the "any JSON type" premise isn't enforced at the boundary that introduces it.

## On splitting `PayloadManager`

I agree with @jordanfanapour that it is separable, and I'll add the argument for keeping it together: the scoping rationale — why `isVacantValue` is applied only to positions *this merge vacated*, and the three non-interchangeable emptiness tests — is the most valuable thing in that section, and it is legible here because the array-merge context is on the page. Split out, it becomes a bug fix with an unusually long comment. That is a real cost against the smaller-review-surface benefit. Your call either way.

rkihm-BC:
Two corrections to my own review, one of them about what I actually checked.

**I overstated my coverage.** I wrote "everything else here is good", which reads as though I reviewed all sixteen files. I did not. I verified the `@memberjunction/ai-agents` half closely — `artifact-target-plan.ts`, `VetArtifactVersionTarget`, and the parent/child boundary in `base-agent.ts` (both mappings set `artifactDirective: undefined` after the spread, with the warn helper, exactly as described). For the four Angular files and `PayloadManager` I read the description and relied on @jordanfanapour having traced them. That is a reasonable division of labour between two reviewers, but I should have said so rather than implying independent verification.

**The `rejectedIds` note is slightly stronger than I pitched it.** I suggested "a sentence so it reads as intended rather than incidental". Having read the existing doc comment, it is not a gap so much as a tension: the comment states an absolute —

> A caller-supplied `sourceArtifactId` is shape-checked but NOT loaded or authorized

— and there is one case where that id is refused for an authorization reason, because the agent named the same id first, failed the ownership check, and the id was remembered. Strictly the caller path still performs no authorization; the net effect for that id is nonetheless an authorization refusal. Worth a clause on that sentence rather than a new one.

**And the main finding checks out under a harder look**, which I did because a weak finding still costs you time:

- `ArtifactPermissionProvider` is `@RegisterClass(PermissionProviderBase, 'MJArtifactPermissionProvider')`, present in `metadata/permission-domains/.permission-domains.json`, and active in the baseline with `Read,Update,Delete,Share`. Live infrastructure, not dead code.
- `PermissionEngine` is reachable from this package — `@memberjunction/core` is already a dependency — and is proven server-side today in `Lists/server/src/ListSharing.ts`, `Communication/notifications/src/shareNotificationHandler.ts` and `MJCoreEntities/src/custom/`. So "use the provider" is not a client-only pattern that would not work here.
- The guide line is verbatim as quoted.

None of this is a change request — my review is a comment, not a block. The provider question is the one thing I would genuinely like decided by whoever owns the permissions contract, and the two smaller notes are doc wording.

AN-BC:
@CaelebB-BC @rkihm-BC nothing to add on my end assuming his has been tested and visually confirmed in various scenarios I think we should get it into `edge.6` that @SDesai-BC is about to publish.

rkihm-BC (approved):
All three items addressed, and one of them corrects **me**. Taking that first, because I asserted it during a double-check specifically intended to stop this happening.

## I was wrong about `PermissionEngine`, and `5ff5cf44c5` is right

I claimed it was "proven server-side today in `Lists/server/src/ListSharing.ts`, `Communication/notifications/src/shareNotificationHandler.ts` and `MJCoreEntities/src/custom/`". That is false. I had grepped for `PermissionEngine` as a substring, which matches **`ResourcePermissionEngine`** — a different class over `MJ: Resource Permissions`. Checking properly:

- `ListSharing.ts` mentions `ResourcePermissionEngine` in a comment and nothing else
- `shareNotificationHandler.ts` imports and calls `ResourcePermissionEngine.Instance`
- An exact-word search for `PermissionEngine` across `packages/`, excluding that class and tests, returns **nothing**

So `PermissionEngine` has no server-side callers, and "just delegate to the provider" was not the established pattern I presented it as. The doctrine half of my finding stands — the provider exists, `UNIFIED_PERMISSIONS_GUIDE.md` §1 routes this class of question to it, and there are now three answers to "can this user edit this artifact" — but the practicality argument I built on top of it did not hold.

**Your resolution is better than what I proposed.** I suggested compose-or-document. You documented, *and* named the upstream fix that makes composing correct later: `CollectionPermissionProvider` already treats the owner as a synthetic full-access grantee (`// Owner path — synthetic full-access row`, line 91 — I verified it), and the same treatment in `ArtifactPermissionProvider` would make `CheckPermission(..., 'Update')` a complete answer, at which point this method should delegate outright rather than compose an owner check around it. That is the actual fix; I only had the symptom.

Noting for the record that this is the second time today a loose grep of mine produced a confident wrong claim. Worth me tightening; worth you continuing to check.

## The other two

**`08c5b2ac89` — unrecognized behavior now drops the whole directive.** You took the stricter of the two options @jordanfanapour offered rather than the doc fix, and the reasoning is sound: a `behavior` this consumer cannot parse means the producer disagrees about the wire format, which is no basis for trusting the object's other fields. It also makes "ignored" mean ignored — previously a garbled directive still named the artifact *and* suppressed the extracted-name fallback, so the log said one thing and two behaviors changed. The type doc in `agent-types.ts` now says "not merely as no targeting", which closes the gap Jordan actually raised.

Worth stating plainly since it is a behavior change: a directive with a typo'd behavior but a good `name` no longer names the artifact. That is less friendly and more honest, and I think it is the right trade.

The `Number.isFinite` guard rode along in the same commit, with the right rationale for why skipping the dedup check is the safe direction.

**`ef795dd428` — non-string `targetArtifactId` discarded at the boundary.** `typeof directive.targetArtifactId === 'string' && ...` in `planArtifactTarget`, so a `version` plan's `artifactId` is genuinely a string downstream, and the type comment no longer has to say "may be any JSON type". That was the one place the "any JSON type" premise was not enforced where it entered.

## On @AN-BC's note

> nothing to add on my end assuming this has been tested and visually confirmed in various scenarios

Read alongside @jordanfanapour's request, that resolves the framework-contract question by deferral rather than by decision: the permission-provider divergence stays documented rather than composed. I am content with that given the upstream fix is now named in the code — but it does mean the "second review from someone deeper in the write path" that Jordan asked for has not happened. **@EL-BC** has not weighed in, and the load-bearing claim is still backward compatibility of `ProcessAgentArtifacts`. That is a call for you and Amith, not a blocker from me.

## State

0 behind `next`, 8 checks green with 5 still running including `Build` and the deterministic tier. No objection from me once those land.

I have not re-reviewed the Angular half or `PayloadManager` — unchanged since @jordanfanapour traced them, and I said last time I was relying on that rather than my own reading.

PR #4302, opened 2026-09-08
Computer Use: DOM-grounded selection and DB-backed replay scripts

CaelebB-BC:
Carves DOM selection and replay-script generation out of #3033 into a self-contained PR on `next`.

A Computer Use test drives a real browser with a vision model in the loop, so every run pays full model price to re-derive the same clicks. This adds a replay tier: a passing run is distilled into a script that later runs execute through the same browser adapter — no screenshots, no model calls — and the model returns only when replay diverges.

Design doc: `plans/regression-testing/dom-selection-and-replay-design.md`.

---

# DOM selection

**Element grounding.** The engine extracts the page's interactive elements each step and hands the model an indexed list — role, accessible name, selector. The model acts by index (`ClickElement 7`) rather than by coordinate, and a recorded target is the element the model chose rather than a bounding box.

**`resolveActionLocator`** narrows an ambiguous selector to one locator before acting: visible matches first, then smallest by area, ties broken on document order. Past fifty matches it takes the first match rather than measuring every node. Zero or one match returns the locator untouched, keeping Playwright's auto-wait semantics. A multi-match throws in strict mode today, so no currently-working action can regress.

**`Scope` on `TraceTarget` and `InteractiveElement`.** The in-page probe records the nearest labelled ancestor region (`group:All applications`), which separates same-named twins — an app listed in both the launcher's "Recent" and "All" grids. `reresolveTarget` narrows by scope, then falls back to recorded-position disambiguation, then declines. The field is absent on pre-existing recordings.

**Overlay dismissal.** `overlay-dismiss.ts` recognizes CDK/Kendo/Material/Bootstrap backdrops, presses Escape, and retries, rather than letting Playwright's actionability wait consume the action budget. Both act paths use it: the grounded path falls through to a selector heal afterwards, the selector path reports the failure.

# Replay

**A step carries** a multi-signal locator (`Selector` primary, `Role` + `Name` as the heal fallback, `BoundingBox` for pre-grounding recordings), a fail-fast precondition, and a postcondition. A failed precondition fails the step; a failed postcondition marks it diverged and starts the heal ladder.

**A script carries** `AppBuildHash`, `AppVersion`, `GoalHash`, the viewport, the variable *names*, and the goal postconditions. Values are tokenized to `%name%` on record and substituted on replay, so a script stores no credentials or run-specific data.

**Scoring.** Recording distils the passing end state into goal postconditions — a URL pattern, an element that must be present, an element that must be absent — and replay executes those instead of calling a judge. Only judge-approved, oracle-green runs produce them.

**Tier selection:**

| Condition | Tier |
|---|---|
| No script, changed goal text, or heal rate over threshold | `llm` |
| Exact `AppBuildHash` match | `replay` |
| Anything else | `replay-with-heal` |

**Fallback.** A diverged replay falls back to the model within the same attempt. The fallback starts with a clean memo rather than the failed replay's. A replay is never re-recorded — only a model-driven run produces a script. `AllowLLMFallback: false` makes the divergence the result instead; default is `true`.

**Review gate.** A script that would replace an existing one lands in `Configuration.PendingReplayScript`; replay keeps using the promoted `ReplayScript` until it is promoted by hand. A test's first script skips the gate and goes straight to `ReplayScript`.

```
mj test scripts                                  # list pending scripts and their drift
mj test scripts --promote --test "T042 - …"      # promote one
mj test scripts --discard --test "T042 - …"      # drop it, keep the promoted script
```

The listing separates heal-class drift (selector changed, role and name held) from UI movement (changed target, verb, URL, or step count), and leads with the second count. The drift summary is a local implementation rather than the engine's `diffTraces`, which would pull `@memberjunction/computer-use` and its Playwright peer dependency into testing-cli.

# Where scripts live

`Configuration.ReplayScript` on `MJ: Tests`.

- **No migration.** `Configuration NVARCHAR(MAX) NULL` already exists; this registers JSONType metadata on it and CodeGen emits a typed accessor.
- **Reads hit cache.** The TestingEngine already caches `MJ: Tests`.
- **`ITestConfiguration`** declares `ReplayScript`, `PendingReplayScript` and `AllowLLMFallback` over an index signature, so each driver's own config passes through.
- **Type parity is compiler-enforced.** CodeGen emits the JSONType into `core-entities`, which sits below the engine package and cannot name `ComputerUseTrace`, so the shape exists twice. `script-store.test-d.ts` asserts assignability both ways; the assertions were confirmed to fail on injected drift.
- **`mj sync push` resets both slots.** `mj sync pull` brings scripts down when they are worth keeping.

In the repo, a script externalizes to its own file under `regression/scripts/` rather than inlining into the test.

> **No scripts are committed in this PR.** The wiring is here, but `regression/scripts/` is empty, so every test currently runs on the LLM tier. Landing the 63 recordings is a follow-up.

---

# Also in this branch

## Perception and judging

- **Rubric judging.** The judge now receives each test's hand-authored 3–5 item rubric and returns a binary per-criterion verdict (`{criterion, met, evidence}`); `done = all(met)`, and coverage is the score. Replaces an uncalibrated confidence float. The derivation is pure and unit-tested.
- **Checkpoint tours** judge each section against its own frame rather than the end state.
- **Two deterministic oracles.** `DomAssertOracle` asserts role/name/count over the recorded end-state element list. `NoConsoleErrorsOracle` fails on a recorded console error, page error, failed request, or renderer crash. Both are pure over recorded output.
- **Perceptual hashing.** A 64-bit dHash replaces byte equality in stall detection, loop detection, screenshot dedupe and judge-call gating.
- **Goal postconditions wait for an in-flight navigation to land**, fixing runs that failed on the OAuth callback URL.

## Run containment

Four defects found while validating against the Docker stack, each fixed test-first.

| # | Defect | Fix | Test |
|---|---|---|---|
| 1 | The interactivity probe was unbounded — `page.evaluate()` ignores `setDefaultTimeout()` and a silent renderer never throws | The probe races an expiry timer | `element-extraction-probe-timeout.test.ts` |
| 2 | The time budget was checked only *between* steps, so a step that blocked internally never self-expired | `raceStepAgainstDeadline` wraps every step in the main loop | `step-deadline-engine.test.ts` |
| 3 | `Stop()` reached the LLM calls but never browser work, so a failsafe-stopped test hung its worker forever | The same race also takes the abort signal | `step-deadline-engine.test.ts` |
| 4 | A crashed Chromium was reused — Playwright leaves `Browser` truthy while `isConnected()` goes false, and workers share one process | Checkout tests `isConnected()` and replaces a dead browser | `HeadlessBrowserEngine.crash-recovery.test.ts` |

## MetadataSync — JSON sub-property externalization

`pull.externalizeFields` accepted entity fields only, so it could move a whole column into a side file but not a single property inside a JSON column. An entry may now be a dotted path — `Configuration.ReplayScript` — which externalizes that leaf and leaves an `@file:` reference in its place. Push already resolves nested references, so there is no push-side change.

- A property the record does not carry is skipped entirely — no file, no key.
- A whole-field config wins over its dotted paths, so both forms cannot fire on one field.
- File IO stays in `FieldExternalizer`, injected as a callback; the new module is only the JSON walk.
- Externalization is driven by config, not by file contents: pulling a field that holds an `@file:` reference without a matching entry inlines the database value and orphans the file. Documented in the config type.

Also: pull's existing-file discovery moved to `lib/existing-record-files.ts`, and the branch adopts `next`'s consolidated `record-primary-key` helpers in place of its own copy.

## MJExplorer — readiness beacon

The shell publishes `data-mj-ready="true"` on `<html>` when the active route's resource has finished loading, and the settle loop polls it instead of comparing consecutive screenshot hashes.

- Nothing in the product reads the attribute and no styling keys off it.
- It is published from the `loading` accessor, so all ~22 assignment sites stay correct.
- Cleared by removal, not by `false`, since the polled selector matches on the value.
- A no-op where there is no DOM (SSR, node-preset unit tests).

## Regression corpus and harness

- The 155-test suite moves from `metadata/tests/regression/` to `metadata-optional/regression-test/`, keeping it out of the base instance the way `integration-test` already is. Standard MJ now ships only the `Computer Use` TestType definition.
- A migration prunes `T01`–`T25` from existing instances: suite memberships unconditionally, test rows only where no `TestRun` references them. Six carried primary keys the consolidation had reused for `T001`–`T006`, so those six tests now have fresh keys and the migration targets only rows the v5 baselines seeded. It is a migration rather than metadata delete records because a `deleteRecord` directive re-stamps `deletedAt` into its own file on every database where the row is still present — which, with CI rebuilding from the baselines each run, is every run.
- Docker regression: accept either `AUTH0_CLIENTID` or `AUTH0_CLIENT_ID`; seed `UserApplication` rows so app-switcher targets are visible; invoke the MJ CLI and the register-class bootstrap by path, since `npx` otherwise resolved an unrelated `mj` package from the public registry.

---

# Verification

| Suite | Result |
|---|---|
| `@memberjunction/computer-use` | 647 passed, 2 skipped |
| `@memberjunction/computer-use-engine` | 220 passed, no type errors |
| `@memberjunction/metadata-sync` | 420 passed |

`__mj.ts` regenerated byte-for-byte against a clean database built from all 86 migrations; the delta is exactly the `Scope`/`Configuration` change.

End-to-end on a freshly migrated database, 30-test subset: 20/30, all 16 replay-tier tests passing, no browser crash or cascade. The 10 failures are pre-existing LLM-tier limits (`impossible`, `loop-detected`, `timeout-progressing`). The same subset before the containment fixes lost 12 tests to a single crash. That run predates the decision to defer the recordings.

# Not built

`APP_BUILD_HASH` minting (the engine treats an empty string as always `replay-with-heal`), the heal-rate ledger feeding the demote rule, and the canary lane that re-derives a sample through the full model path.

# Relationship to #3033

#3033 is untouched and still contains this work; rebasing it onto this branch is a follow-up. It keeps the CLI regression commands and the TestingFramework engine/CLI work.


rkihm-BC (changes requested):
I reviewed `8b09d97452` end to end (the last commit, `8b09d97452`, only adds documentation; every finding below was verified on `95f4d644ef`, whose code is identical): the full 332-file diff (engine, driver, MetadataSync, metadata, migration, Docker and the 155-test corpus), a build of the touched package chains in a clean worktree, every touched package's unit suite, and probes against the built engine for the replay guards. The design is coherent and the containment fixes are the right shape. The replay tier, though, has correctness gaps that would make it fall back or pass vacuously in practice, so this is Changes requested.

*This review was generated with Claude Code.*

### Required

**1. `src/engine/perception.ts` contains two raw NUL bytes and is a binary file to git.**
Line 18 builds `elementKey` with literal 0x00 separators. `git diff --numstat` reports `- -`, the PR diff shows "Binary files differ", and GitHub cannot render it for review; blame, grep and lint skip it too. It compiles, which is how it got here. Use `'\u0000'` escapes or a printable separator.

**2. `normalizeTraceUrl` (`trace.ts:75`) is not idempotent, so a recorded URL guard can fail against its own live URL.**
Query values are decoded by `URLSearchParams` and re-emitted raw. Patterns are normalized at record time and again inside `traceUrlMatches`, while the live URL is normalized once, so any value carrying an encoded `&`, `=`, `#` or `+` disagrees. Executed against the built module: `?redirect_uri=https%3A%2F%2Fapp%2Fcb%3Fa%3D1%26b%3D2&state=s` normalizes once to `...?redirect_uri=https://app/cb?a=1&b=2&state=s`, a second pass reorders it to `...?b=2&redirect_uri=...&state=s`, and `traceUrlMatches(recorded, live)` returns `false`. Every step whose `UrlBefore` or postcondition carries a `returnUrl`-style parameter diverges on every replay. `trace-url-idempotency.test.ts` covers only the path-UUID case. Rebuild the query encoded via `URLSearchParams.toString()` after dropping volatile params and sorting, or compare parsed structures.

**3. `traceUrlMatches` (`trace.ts:94`) is substring containment, so full-URL goal postconditions are vacuous.**
The doc comment says a full-URL pattern matches exactly; the code is `a.includes(p)`. Executed: pattern `http://localhost:4200/` matches `http://localhost:4200/login-error?x=1`. A `url` goal postcondition distilled from a run that ends at the app root, which is the login smoke test, passes on any URL of that origin. Require normalized equality (or a path-prefix at a `/` boundary) when the pattern is an absolute URL.

**4. `isSelectorHealable` (`replay.ts:277`) treats a URL-precondition failure as selector drift.**
It excludes only reasons starting with `postcondition`. Executed through the real engine: a step recorded with `Precondition.UrlPattern=/app/data`, live URL `/app/home`, and a `button "Save"` present produced outcome `healed`, detail `healed after: precondition - entry URL does not match`, clicked `#save`, and the run `Completed`. Being on the wrong page is flow drift by the PR's own definition. Heal only the "target never became attached and visible" and "action failed" reasons.

**5. A healed replay rewrites the promoted `ReplayScript` without going through the review gate.**
`script-store.ts:42-48` (`loadScript`) returns `test.ConfigurationObject.ReplayScript` by reference, and the generated getter hands back its cached parsed object. `ComputerUseEngine.ts:1655-1657` heals in place (`traceStep.Action.Target.Selector = resolution.selector`, commented "Cache rewrite ... in place"). `saveScript` (`script-store.ts:68-82`) then spreads `readConfiguration(test)`, whose `ReplayScript` is that mutated object, and saves it. Scenario: `replay-with-heal` heals step 3, diverges at step 7, the default `AllowLLMFallback` runs the model, the run passes, and the recorder writes the fresh script to `PendingReplayScript` while also persisting the healed selector into the promoted slot. That contradicts "a replay is never re-recorded", and `mj test scripts` then diffs pending against an already-mutated baseline. There is no clone anywhere on the path. Clone in `loadScript` or at `Replay()` entry, and add a test that a heal-then-fallback run leaves `ReplayScript` byte-identical.

**6. Recorded scripts can still carry credentials and run-specific values.**
The description says "a script stores no credentials or run-specific data". Three gaps, each executed:
- `trace.ts:333` tokenizes after `compactInstruction` truncates to 200 chars, so a password straddling the cut survives as a fragment. Tokenize first, then truncate.
- `UrlBefore`, `Precondition.UrlPattern`, `Postcondition.UrlPattern` (`trace.ts:327,354,362`), the goal URL (`verdict.ts:43-49`) and `Target.Name` (`trace.ts:421`) are never tokenized: `?login_hint=robert@example.com` and a clicked `Continue as robert@example.com` button are stored verbatim, and the literal `http://localhost:4200` origin sits in every guard, so a script recorded locally fails step 1 on any other host. Run `tokenize` over every string field of the trace.
- `tokenize` (`trace.ts:445-455`) is a blind `split(value).join()` over every variable after `coerceVariableValues` stringifies numbers and booleans, so `maxSteps=40` turns `:4200` into `:4%maxSteps%0`. Tokenize only string values above a minimum length.

**7. Two packages with shipped source changes are missing from the changeset.**
`.changeset/dom-selection-and-replay-scripts.md` lists seven packages, but the branch also changes `@memberjunction/metadata-sync` (dotted `externalizeFields`, the typed-JSON companion exclusion, recursive record discovery: 874 lines across 11 files) and `@memberjunction/ng-explorer-core` (the `data-mj-ready` beacon). Release notes are built from changesets. Add both; `minor` stays correct since the branch carries a migration and metadata.

**8. With the documented Docker setup, every test now logs in with an empty username and password.**
On `next` the T01-T25 records carried literal credentials; all 155 new tests bind `{{authUsername}}`/`{{authPassword}}`. The documented Mode A environment is `TEST_UID`/`TEST_PWD` in `.env.test` (`REGRESSION_TESTING.md:452`, `docker/README.md:123`, `QUICKSTART.md:44`), and compose passes those to the runner (lines 186-187), but nothing maps them onto the variables the tests read: not the entrypoints, not `up.ts`, not `docker-helpers.ts`; `load-target-profile.cjs:111-114` sets `MJ_TEST_VAR_auth*` only for remote target profiles. Compose meanwhile defines `MJ_TEST_VAR_authUsername: "${MJ_TEST_VAR_authUsername:-}"` (lines 222-223), so inside the container the variable is always defined, as `""` when the host did not set it; `buildVariableValuesFromContext` keeps it (`value !== undefined`, `variable-substitution.ts:61`) and `substituteVariables` injects the empty string. The unresolved-placeholder fail-fast (`ComputerUseTestDriver.ts:271-272`) inspects only `startUrl` and `goal`, not the auth binding. Result: FormLogin submits blank credentials and all 155 tests fail at Auth0; the 20/30 run in the description must have relied on undocumented `MJ_TEST_VAR_auth*` values in a local `.env.test`. Fix: map `TEST_UID`/`TEST_PWD` onto the two variables in `test-runner-entrypoint.sh` (or a nested compose default), treat an empty `MJ_TEST_VAR_*` as unset, extend the fail-fast to `auth.bindings[].method.Username/Password`, and document the variables.

**9. Four places still describe the delete-records prune this branch replaced with a migration.**
`metadata/tests/regression/.deleted-computer-use-tests.json` does not exist at head (`metadata/tests/` holds only `research-agent/`), but it is still cited by `metadata-optional/regression-test/README.md:13` (a dead relative link and a wrong mechanism), `test-runner-entrypoint.sh:96,104-106` (comments and the "incl. Computer Use delete records" echo), `clear-baseline-suite-members.cjs:8-17,29-33` (both of whose stated justifications are now false: none of the 155 keys reuses an old one, and the migration deletes all 25 memberships during db-setup, so the script deletes zero rows on a fresh database), and the `REGRESSION_TESTING.md:531` tree. Point them all at `V202609171030__v6.2.x__Prune_Pre_Consolidation_ComputerUse_Tests.sql`, and either drop the script or rewrite its comment to its one remaining purpose (persistent-volume reruns).


### Non-blocking

- After a `StepDeadlineError` the orphaned step keeps running (`ComputerUseEngine.ts:757`): with fake timers, `Run()` returned `TimeBudgetExceeded` with zero controller calls, then one controller call and one browser action landed afterward with `cancelled` still false. That is an LLM call per expiry, and on the recycled shared-context path the orphan can click into the next test's page. Set `cancelled` and abort in the handler.
- `Replay()` (`ComputerUseEngine.ts:1240`) awaits `replayOneStep` directly with no deadline or abort race and ignores `MaxExecutionTimeMs`; `Stop()` during a step blocked in `CaptureScreenshot` leaves `Replay()` pending. Containment items 2 and 3 in the description hold only for the LLM-tier loop.
- `element-extraction.ts:363` drops every `tabindex="-1"` element. Roving-tabindex composites (Kendo and Material tabs, toolbars, listbox options, menu items) put `-1` on every inactive item, so they vanish from the grounded list and the model falls back to coordinates, which `isRecordableRun` then refuses. Skip on `-1` only when the tabindex was the sole reason the element matched.
- Selector-path overlay retry (`PlaywrightBrowserAdapter.ts:741`, `SharedContextBrowserAdapter.ts:306`) waits a full `ActionTimeoutMs` before Playwright reports the backdrop, then Escapes and retries with a second full budget; the description's "rather than letting Playwright's actionability wait consume the action budget" holds only for the grounded path. Related: the `element-extraction.ts:122` comment that the two probe attempts "never exceed the caller's actionTimeoutMs" is false (worst case is about twice the budget plus 2 s).
- `tier`, `replay`, `failureClass` and `failureMemo` never reach `TestRunResult`: `TestEngine.ts:1306-1327` builds it without them, so the new `TestRunResult.tier/replay` fields are dead and the doc "Copied to TestRunResult.replay" is false. They survive only inside `TestRun.ActualOutputData`.
- `saveScript` sets `ConfigurationObject` before `Save()`; on a `false` or a throw the cached `MJTestEntity` keeps the unsaved script, so the next run in the same process replays a script that never landed. Restore `current` on failure.
- `script-drift.ts:80-104` compares only `Method`, `Action.Url`, `Role`, `Name`, `Selector`; a changed `Text`, `Key`, `PressEnter`, `Scope`, `UrlBefore` or any postcondition reports "identical to the promoted script", and its `url-changed` means `Action.Url` while the engine's `diffTraces` uses `UrlBefore` for the same label.
- `--promote` and `--discard` without `--test` resolve every pending script with no confirmation, and a `--test` value that matches no test prints the same "No pending replay script" as a test with none.
- Docs that contradict the code: `ITestConfiguration.ts` and its `__mj.ts` copy say fallback will "overwrite ReplayScript" (it writes `PendingReplayScript`); the `buildRunParams` comment says grounding is "default off" while `usesElementGrounding` defaults true; `mj-params.ts:113-120` says the regression suite disables `PersistStepMedia`, but nothing sets it.
- `dispatchRun` and `maybeRecordScript` (`ComputerUseTestDriver.ts` about 860-1000) have no tests: nothing asserts a Completed `replay-with-heal` is not recorded, that fallback lands in pending, that `AllowLLMFallback:false` returns the divergence, or that the loaded trace is left untouched, which is why item 5 was not caught.
- `pnpm-workspace.yaml` adds `enablePrePostScripts: true` with a comment saying pnpm "does not [run pre/post scripts] unless this is on". Not true for the pnpm this repo runs: `packageManager` pins 10.33.0, the Docker images activate the same pin through corepack, and a probe package with a `prebuild` runs it under `pnpm run build` with or without the setting. The toggle is a no-op, the comment misstates the behavior, and the change is in neither the description nor the changeset. Drop it, or keep it with an accurate comment.
- Nits: `HeadlessBrowserEngine.ts:398` dynamic `import('node:fs/promises')` (the `playwright` one is a justified optional peer, this is not) and `JSON.parse(raw) as StorageState` unvalidated; `judgeReplayEndState` (`ComputerUseEngine.ts:2185`) adds a second screenshot history entry after `captureScreenshot` already did; the judge cache key ignores which criteria were asked, so a checkpoint-scoped call can return another scope's verdicts when a shared cache is injected; `selector-resolution.ts:94` "visible" is a non-zero rect, so `visibility:hidden` still counts.

- The controller and judge prompts in core `metadata/prompts` move from `Gemini 3.1 Flash-Lite` to `Gemini 3.6 Flash` and gain `Temperature`/`Seed`. That is a model change for every instance that syncs `metadata/`, and it appears in neither the description nor the changeset. State it.
- `judge.template.md:24-25` hardcodes MJ Explorer boot text ("Loading workspace...", "Spinning up resources...", 60+ seconds) into a core prompt shipped to every instance, while `judge-core.md` says application-specific indicators come from environment notes. The same text already lives in the suite's `mjexplorer-context.md`; render `{% if applicationContext %}` in the judge template instead.
- The judge's three canonical JSON examples in `judge-core.md` and the `OutputExample` in `.computer-use-prompts.json` omit the `criteria` array the description says replaces the confidence float; `verdict.ts:242-279` tolerates its absence by falling back to scalar `Done`, so this degrades rather than breaks, but three examples without `criteria` make omission the likely outcome. All 155 tests still gate `goal-completion` on `minConfidence`, so confidence remains a scoring input; say so.
- Base pull filters (`metadata/tests/.mj-sync.json` `Type <> 'Integration Test'`, `metadata/test-suites/.mj-sync.json` `Name NOT LIKE 'Integration Tests%'`) will pull the 155 Computer Use tests and the regression suite back into `metadata/` on the next `mj sync pull --dir=metadata`, undoing the separation. Exclude `Computer Use` and the suite name the way the integration-test root uses a positive filter.
- Corpus authoring shape regressed from `next`: every one of the 155 files stores `InputDefinition` and `ExpectedOutcomes` as escaped JSON strings (plain objects on `next`), writes the virtual `Type` field and five null fields, and 90 of 155 carry `sync` blocks while 65 do not. It pushes, but the diffs are unreviewable and the README's "Anatomy" section shows objects. Set `ignoreNullFields`/`ignoreVirtualFields` on the tests config and re-pull, or hand-normalize.
- `.users.json` adds `UserApplication` rows with a comment saying "nothing backfills them", but `setup-test-user.cjs:80-96` runs right after that push and inserts a row for every `Status='Active'` application with `NOT EXISTS`, and all four apps are Active. The rows are harmless duplicates; the stated root cause cannot be right as written.
- Docs still describe the npm flow the PR removed: 13 `npm install`/`npm run`/`npx mj` mentions in `QUICKSTART.md`, 13 in `REGRESSION_TESTING.md`, 12 in `ARCHITECTURE.md` (its trace at 426-434 still shows `npx mj sync push`). The root CLAUDE.md forbids `npm install` here. `docker/agentic-test-runner/Dockerfile` (edited at line 66) still copies `package-lock.json` and runs `npm install`, so it cannot build from this tree; pre-existing, but the PR touched it.
- Nits: `ARCHITECTURE.md:343-350` step numbering runs 5,6,7,8,6,7,8 after the insertion; the design doc jumps from section 6.2 to 8; the README layout lists `.T002-home-application-load.json` while T002 is `.T002-data-explorer-open-record.json`; "hand-authored 3-5 item rubric" is 48x3, 31x4, 1x2 and none with 5, with 75 tests being checkpoint tours instead.


### Verified

| Check | Result |
|---|---|
| `@memberjunction/computer-use` | 652 passed, 2 skipped (live-browser integration files); the description says 647 |
| `@memberjunction/computer-use-engine` (MJComputerUse) | 230 passed, including 6 type tests; an injected optional-field drift fails 2/2 under `--typecheck.only`, so the parity assertions are live |
| `@memberjunction/testing-cli` | 37 passed |
| `@memberjunction/metadata-sync` | 420 passed |
| `readiness-beacon.dom.test.ts` (dom project) | 3 passed |
| Build | `pnpm build` clean; the ComputerUse prebuild regenerated `prompt-parts.generated.ts` byte-identical |
| Migration IDs vs the 25 deleted `metadata/tests/regression` records | exact match, both directions |
| Corpus (scripted) | 155 files parse; 155 unique well-formed keys, none equal to the 25 old test keys, 25 old member keys or the suite key; names unique and contiguous T001-T155; every `TypeID` is `@lookup:MJ: Test Types.Name=Computer Use`; every test has goal, start URL, domains, one FormLogin binding on the two variables, and the `regression` tag; zero credential or PII pattern hits |
| Suite | one record on the same key as `next`; 155 members, sequences exactly 1-155 matching the test numbers; no dangling, missing or duplicate references |
| Prompt lookups | `Gemini 3.6 Flash`, `Gemini 3.1 Flash-Lite`, vendors `Google` and `Vertex AI`, category `MJ: System`, and the heal template all resolve |
| Docker | all entrypoints `set -e`; every `mj sync push` wrapped; the `AUTH0_CLIENTID`/`AUTH0_CLIENT_ID` nested default verified with `docker compose config` for all three cases; the CLI is built before db-setup runs it by path; `metadata-optional/` is baked and bind-mounted; `.dockerignore` excludes nothing a Dockerfile copies |
| Base `mj sync push --dir=metadata` | does not pick up `metadata-optional/` |
| FKs into `Test` | only `TestRun` and `TestSuiteTest`, so the migration's two deletes are complete |
| Migration placement | after `V202609161850` in `migrations/v6`, `${flyway:defaultSchema}`, `XACT_ABORT` |
| `npm run check:changeset` | passes (`minor` required, present) |
| Tier table, `AllowLLMFallback`, first-script/pending slot logic, no record on Timeout/Cancelled/Failed/oracle-failed | confirmed by read and the 17 script-store tests |
| `ITestConfiguration.ts` vs `__mj.ts` vs `ComputerUseTrace` | field-for-field identical; metadata JSON lookups match 25 sibling files; `z.any()` matches 37 other JSONType fields |
| Crash recovery, rotation refcounts, `raceStepAgainstDeadline` cleanup, probe timeout race, `resolveActionLocator` tie rules, dHash, `Scope` re-resolution | read, and the tests fail without the production code |
| `pngjs` | declared, locked at 3.4.0, resolves; the local `.d.ts` covers the three members used |
| `check-esm-imports` | OK (143 OK, 82 not built) |
| CodeQL | the check fails on attribution only: the 9 open alerts in the touched packages on the PR ref are the same alert numbers as on `next`, and the three alerts CodeQL commented on this PR are closed at head |
| `mj test scripts` dynamic import | matches every sibling `test/*.ts` command |
| `Temperature` / `Seed` on `MJ: AI Prompts` | fields exist |

rkihm-BC (dismissed):
Re-reviewed at `92de42b884`. **All nine Required items from my previous review are fixed** — several more thoroughly than asked. Approving.

I verified each at head rather than trusting commit subjects, which understated the work: items 2, 3, 4, 5 and 6 have no commit mentioning them and rode in under *"abort the orphaned step, and bound replay like the…"*.

### The nine

| # | Item | Evidence at head |
|---|---|---|
| 1 | `perception.ts` was binary | 0 NUL bytes; `numstat` now `161 0`, not `- -` |
| 2 | `normalizeTraceUrl` not idempotent | **Executed:** `normalizeTraceUrl(x) === normalizeTraceUrl(normalizeTraceUrl(x))` on the original `redirect_uri` repro, and the URL now matches itself |
| 3 | `traceUrlMatches` substring containment | **Executed:** `http://localhost:4200/` no longer matches `/login-error?x=1`; root still matches root, `/app/data` still matches `/app/data/records`, and `/app/data-archive` correctly does not |
| 4 | `isSelectorHealable` healed URL-precondition failures | Now `PRECONDITION_TARGET_MISSING \|\| ACTION_FAILED` only, with the reason string shared as a constant so producer and predicate cannot drift |
| 5 | Healed replay rewrote the promoted script | `structuredClone` in `loadScript`, **plus** the regression test I asked for — *"hands back a copy, so mutating the loaded script leaves the row untouched"* |
| 6 | Scripts could carry credentials | Tokenize-before-truncate; `tokenizeTraceDeep` as a final sweep over `Steps` and `GoalPostconditions`, so a newly-added field cannot be forgotten; `isTokenizable` stops `retries=20` corrupting `:4200` |
| 7 | Two packages missing from the changeset | `metadata-sync` and `ng-explorer-core` added; still `minor`, which is right for a branch carrying a migration |
| 8 | Every test logged in with blank credentials | All three parts: the entrypoint maps `TEST_UID`/`TEST_PWD` (explicit `MJ_TEST_VAR_*` still wins, and it warns when neither is set); `variable-substitution.ts` now treats `''` as unset; the fail-fast gained `findUnresolvedAuthPlaceholders(input.auth)`. Confirmed compose actually uses that entrypoint (line 237) |
| 9 | Stale prune references | Zero references to `deleted-computer-use-tests` remain, and `clear-baseline-suite-members.cjs` did not just lose the link — its header was rewritten to say what has stopped being true and to name the migration |

Item 5 and item 8 are the two I'd single out: both went past the minimum. The `loadScript` comment explains the cache-reference problem in its own words rather than restating mine, which is the version that survives the next person reading it.

### CodeQL: verified as attribution, at file level

The check reports *"11 new alerts including 11 high severity."* The count moved 9 → 11 since my last review, so I re-derived it rather than inherit the old conclusion.

Pulled the check-run annotations and mapped each against the PR's 338 changed paths. **All 11 are in files this PR does not touch:**

- `AI/ComputerUse/src/engine/ResponseParser.ts` ×2
- `AI/Core/src/generic/errorAnalyzer.ts` ×4
- `Angular/Explorer/…/shell.component.ts` ×2
- `Angular/Explorer/shared/…/navigation.service.ts` ×2
- `MJCore/src/generic/compositeKey.ts` ×1

Zero overlap with the diff. CodeQL's own summary carries the explanation — *"Alerts not introduced by this pull request might have been detected because the code changes were too large."* Not this PR's doing.

Everything else is green, including the Integration tier now that the branch carries `next`'s RVM9 fix.

### One low-severity note, explicitly unproven

`isTokenizable` decides on the **value's shape** and never sees the variable's **name**, so a secret under 4 characters, or an all-digit secret under 8, is not tokenized and would land in a recorded script.

Two reasons this is a note and not a finding. Your own test — *"still tokenizes a substantial value, so the length floor does not disable secrecy"* — shows the floor was weighed, and a realistic Auth0 password (7+ chars, mixed) is covered by it. And I could **not** produce a working repro: my probe passed `steps` where `recordTrace` wants `Steps`, so it threw and my assertion passed vacuously. The claim rests on reading the three rules in `isTokenizable`, nothing more.

If you want it closed, exempting secret-shaped variable *names* (`/password|secret|token|key|pwd/i`) from the length and numeric heuristics would do it without reintroducing the `:4200` corruption. Entirely reasonable to leave as-is.

### Non-blocking items

Roughly nine were also addressed — orphaned-step abort, replay bounding, roving tabindex, script-save rollback, drift-detection widening, bulk-resolve confirmation, base-pull filters, judge application context, `dispatchRun`/`maybeRecordScript` coverage, overlay-retry bound, and two doc corrections. I did not re-verify each individually; the commit subjects map cleanly, CI is green across all six unit shards, and none was blocking.

Good work turning this around — the replay tier's correctness gaps were the substance of the last review and they are genuinely closed, not papered over.

rkihm-BC (approved):
LGTM

PR #4304, opened 2026-09-08
refactor(core): sweep hardcoded ID primary-key assumptions, triage FirstPrimaryKey, add strict PrimaryKeyCompliance gate

ct-bc:
## Summary

Repo-wide sweep of code that assumed an entity's primary key is a single column named `ID`, a triage of every remaining `FirstPrimaryKey` / `CompositeKey.FromID` use, and a strict `PrimaryKeyCompliance` gate in `@memberjunction/core` so the pattern cannot come back.

MJ supports primary keys with any column name(s) and type(s). Every MJ core entity happens to use `ID`, so hardcoding it works across the whole core product and silently breaks on customer entities mapped from external schemas: `Load()` rejects the invented field name (`Primary key ID not found in entity ...`), or a composite key is truncated to its first column. #4179 (search result click-through, fixed in #4184) was one instance. Everything here builds on the `CompositeKey.FromURLSegment` / `FromEntityRecord` / `ToCompactURLSegment` primitives #4184 introduced. `origin/next` (including the squash of #4184) is merged in.

## Part 1 — hardcoded `ID` sweep

**Literal `ID` key construction** (`{ FieldName: 'ID', Value: x }`, `LoadFromSingleKeyValuePair('ID', …)`, `FromKeyValuePair('ID', …)`) — ~135 sites.
- Entity is a literal MJ core entity → `CompositeKey.FromID(x)`, the one sanctioned way to say "this entity's key is `ID`".
- Entity is a variable (an event's `EntityName`, an `entityInfo`, a configured entity) → `CompositeKey.FromURLSegment(entityInfo, recordId)`, which reads a bare value or a `F1|v1||F2|v2` segment against the entity's real primary key(s).

**`PrimaryKeys[0]` → `FirstPrimaryKey`** — 39 sites. Same semantics, a named accessor the gate can see.

**Real defects fixed**: mobile app record load/edit/offline sync; the generic form overlay and ERD "open record"; version-history label/diff/micro-view links and `RestoreEngine`; Apollo enrichment; geocoding record reload; List Detail record-open for composite keys; `EmbeddedRecord`; `DatabaseReferenceScanner`; hardcoded `ID` filters on a variable entity in Data Explorer, Predictive Studio, the realtime-widget visitor lookup, `DuplicateRecordDetector`, and MetadataSync `@lookup` conversion.

**REST API**: `EntityCRUDHandler` / `RESTEndpointHandler` threw `Composite primary keys are not supported`. Both now accept a bare value or a URL-encoded `Field1|Value1||Field2|Value2` segment. Single-column behavior unchanged.

**One serializer instead of eight**: every record-id serializer (`ListOperations`, `list-set-operations`, RecordSetProcessor, `GetListRecordsAction`, `MJListDetailEntityExtended`, `record.util`, `VersionHistory`, `ChangeDetector`) delegates to `CompositeKey.FromEntityRecord(...).ToCompactURLSegment()` / `FromURLSegment(...)`. Byte-identical for single-column keys.

## Part 2 — `FirstPrimaryKey` / `FromID` triage

Every one of the ~390 `FirstPrimaryKey` and `CompositeKey.FromID` uses in the repo was read in context and either rewritten or annotated with a reason (154 annotations). Real defects found, all of the shape "first key column used as the whole key" on an entity that can be composite-keyed:

- **Data providers**: the deterministic `ORDER BY` fallback for row-limited queries ordered by the first key column only, leaving composite-key pages in undefined order; it now orders by every key column. Saved-view run logging/exclusion and the `{%UserView%}` template subquery, whose persisted `RecordID` cannot hold a composite key, now refuse loudly instead of excluding wrong rows. The dependency-link subquery predicates on the full key. Single-column SQL is byte-identical.
- **CodeGen**: generated cascade delete/update procs bound the child FK to `@<firstPK>` regardless of which parent key column the FK references; a composite key containing an identity column dropped the other key columns from the generated INSERT (both providers); PostgreSQL JSON-arg `spCreate` inserted only the first key column; generated join-grid/timeline filters and the GraphQL audit-log `RecordID` truncated composite keys. Single-key generator output verified byte-identical against `HEAD` across 168 shapes.
- **Smart cache** (`ProviderBase` differential merge): keyed rows on the first PK, so composite-key deletes never applied and rows sharing the first column collapsed.
- **Integration push sync**: composed record identity from the first key column while the record map stores all columns joined, so every already-synced composite-key row was re-created externally as a duplicate on each full push; the changed-record path silently dropped rows.
- **Scheduled geocoding orphan cleanup** (destructive): compared a cast of the first key column to a `RecordID` holding all columns, so every geocode row for a composite-key entity was deleted on each run.
- **Lists**: membership, export and add-record paths filtered on the first key column and wrote only its value into `ListDetail.RecordID`. Also Explorer "open record" paths on user-selected entities, duplicate detection, omnibar record search, Data Explorer deep links, sharing-center revoke, recent access, tree dropdowns, mobile record ids and offline queue.
- **AI**: duplicate detection, vector sync record ids, Predictive Studio list scope and write-back. The Recommendations engine also wrote a record id into `SourceEntityID` (an FK to Entities) and never set `SourceEntityRecordID`.
- **Apollo enrichment**: customer `Accounts` loaded by literal `ID`; the contacts path read its key off an entity that had never been loaded.
- Every `entityInfo.FirstPrimaryKey?.Name ?? 'ID'` fallback is gone; where the entity can be missing the code fails loudly instead of inventing `ID`.

## The gate

`packages/MJCore/src/__tests__/PrimaryKeyCompliance.test.ts`, modelled on `MultiProviderCompliance` / `UUIDCompliance` in MJGlobal. Scans `packages/**/*.ts` (excluding tests, generated code, `dist/`, `TestingFramework`, `UnitTesting`). **All four gates are strict; there is no baseline file.**

| Gate | Catches | Escape hatch |
|---|---|---|
| 1 | A key built with a literal `ID` field name | `// pk-literal-ok: <reason>` |
| 2 | `PrimaryKeys[0]` / `PrimaryKeys.at(0)` | use `FirstPrimaryKey` |
| 3 | `FirstPrimaryKey`, and `CompositeKey.FromID(` without a `'MJ: …'` core-entity literal on the same line or within 8 lines above | `// first-pk-ok: <reason>` (reason mandatory) |
| 4 | An `ID = …` / `ID IN (…)` `ExtraFilter` or `Fields: ['ID']` within 8 lines of an `EntityName:` that is a variable | `// pk-filter-ok: <reason>` |

The rule is in `.claude/rules/data-access.md` § "Primary keys: never assume a column named ID".

## Testing

- **Unit**: full build clean (304 packages). MJCore 2254 passed. Every touched package's suite green; 13 new test files covering the composite-key case for each class-A fix (data providers, CodeGen providers, smart cache, integration round-trip, visitor identity, geocoding, duplicate detection, vector sync, Predictive Studio, Recommendations).
- **Deterministic integration tier** (`pnpm run test:integration`, DB migrated + CodeGen'd to this branch): **65 passed, 0 failed, 6 skipped**, before and after Part 2. Five skips are mutation-gated by design; the sixth is IT52.
- **IT52 Unified Search Seams** (parked behind `RUN_SEARCH_TESTS`; exercises `SearchEntity` / `SearchEntities`, which this PR rewrites): **7 of 7 checks passed**.

## Changeset

`patch` across all touched packages. No public signatures change.

## Notes for reviewers

- REST handlers' acceptance of composite `:id` segments is additive API surface (`RESTEndpointHandler.ts`, `EntityCRUDHandler.ts`).
- `CompositeKey.FromURLSegment` falls back to an `ID` key when entity metadata is unresolvable and the segment is a bare value; documented and marked `pk-literal-ok`.
- Where a feature physically cannot hold a composite key (view-run `RecordID`, join grids, `{%UserView%}` subqueries), the choice was a loud `PrimaryKeys.length === 1` guard over silent wrong rows.

## Found, deliberately not changed here

- Three Explorer resource wrappers call `GetEntityRecordName` with unprefixed `'Dashboards'` / `'Lists'` / `'Queries'`; `EntityByName` is an exact match, so tab titles silently fall back. Three-line fix, separate change.
- Scheduled geocoding stores composite keys as `v1||v2` (no field names), which its own retry reader (`FromURLSegment`) can't parse. Fixing it changes a persisted format the base view joins on.
- The gate does not see `row.ID` / `record['ID']` reads on `simple` rows from a variable entity, `OrderBy: 'ID'` literals, or `'ID'` defaults on config objects. Agents fixed these where adjacent to a listed site; a gate 5 for those shapes is a possible follow-on.


.changeset/primary-key-sweep.md
```diff
@@ -0,0 +1,86 @@
...
+- **Scheduled geocoding orphan cleanup** (destructive): compared a cast of the first key column to a `RecordID` holding all columns, so every geocode row for a composite-key entity was deleted on each run.
+- **Lists**: list membership, export and add-record paths filtered on the first key column and wrote only its value into `ListDetail.RecordID`; Explorer "open record" paths on user-selected entities, duplicate detection, omnibar record search, Data Explorer deep links, the sharing center revoke, recent-access, tree dropdowns, the mobile app's record ids and offline queue.
+- **AI**: duplicate detection, vector sync record ids, Predictive Studio list scope and write-back; the Recommendations engine also wrote a record id into `SourceEntityID` (an FK to Entities) and never set `SourceEntityRecordID`.
+- **Apollo enrichment**: `Accounts` (a customer entity) loaded by literal `ID`; the contacts path read its key off an entity that had never been loaded.
+- Every `entityInfo.FirstPrimaryKey?.Name ?? 'ID'` fallback is gone; where the entity can be missing the code now fails loudly instead of inventing `ID`.
```
rkihm-BC, line 75: This claim is not accurate, and unlike the rest of the changeset it will read as a guarantee in the generated release notes. Six fallbacks survive:

- `packages/TestingFramework/integration-test-suite/src/checks/runquery-catalog.checks.ts:162` holds the verbatim expression `entity.FirstPrimaryKey?.Name ?? 'ID'`. This PR edits that file. The gate excludes the whole `/TestingFramework/` path, which is why the sweep did not see it.
- `packages/MetadataSync/src/services/PullService.ts:468` uses `entityInfo?.PrimaryKeys?.[0]?.Name ?? 'ID'` to match records to local files. This PR edits that file too.
- `packages/Angular/Generic/base-forms/src/lib/field/form-field.component.ts:845` uses `pkFields?.[0]?.Name ?? 'ID'` for a foreign-key dropdown's key column, read back at lines 1043 and 1591 and used as a `Fields` entry at 1661. Line 2079 in the same file has `RelatedEntityFieldName || 'ID'`.
- `packages/Angular/Explorer/dashboards/src/DevTools/graphql-console.component.ts:411` uses `e.PrimaryKeys?.[0]?.Name ?? 'ID'`.

To be fair to the work: none of these is a live defect on an MJ core entity. Each returns the correct column whenever the entity has at least one primary key, so the `'ID'` literal only fires for a zero-key entity, and the first-column truncation needs a composite key. All are unchanged from `next`, so nothing regressed here.

The ask is only to scope the sentence, for example to say the fallback is gone from the paths the gate scans, and to note that optional-chained and aliased forms remain. The same edit is worth making in the PR description.

.changeset/primary-key-sweep.md
```diff
@@ -0,0 +1,86 @@
...
+2. *Strict*: `PrimaryKeys[0]` / `PrimaryKeys.at(0)`.
+3. *Strict*: `FirstPrimaryKey` and `CompositeKey.FromID(`. These are legitimate only where MJ is single-column by design (foreign-key targets, keyset `ORDER BY` / `AfterKey`, IS-A shared keys, core entities), so every use must be self-evidently on a core entity or say why: `FromID` is exempt when a `'MJ: …'` entity literal is on the same line or within 8 lines above (the `GetEntityObject` / `OpenEntityRecord` naming the core entity); everything else carries `// first-pk-ok: <reason>` on the same line, reason mandatory.
+4. *Strict*: an `ID = …` / `ID IN (…)` `ExtraFilter` or `Fields: ['ID']` within eight lines of an `EntityName:` that is a variable rather than a string literal or ALL_CAPS constant. Marker `// pk-filter-ok: <reason>`.
+
+Generated code, tests, `dist/`, and the `TestingFramework` / `UnitTesting` packages are not scanned. The rule is written up in `.claude/rules/data-access.md` § "Primary keys: never assume a column named ID". There is no baseline file: all four gates are strict.
+
+No public signatures change; every edit is additive or a same-shape substitution, so this is `patch` throughout.
```
rkihm-BC, line 86: Minor, same sentence-level point. `LocalCacheManager.ApplyDifferentialUpdate` had its fifth parameter widened from `string` to `string | string[]` and renamed from `primaryKeyFieldName` to `primaryKeyFieldNames`, so this is a public signature change. It is a widening, no caller breaks, and `patch` remains correct under the repo's changeset rule, so only the wording needs a touch.

packages/MJCore/src/__tests__/PrimaryKeyCompliance.test.ts
```diff
@@ -0,0 +1,298 @@
...
+const LITERAL_ID_KEY_PATTERNS = [
+    /FieldName:\s*['"]ID['"]/,
+    /LoadFromSingleKeyValuePair\(\s*['"]ID['"]/,
+    /FromKeyValuePair\(\s*['"]ID['"]/,
+];
+const LITERAL_ID_MARKER = /\/\/\s*pk-literal-ok\b/i;
+
+/** Gate 2 — positional access to the first primary key. */
+const INDEX_ACCESS_PATTERNS = [
+    /\bPrimaryKeys\[0\]/,
```
rkihm-BC, line 67: Three shapes get past this pattern and past gate 3, so the array can be read at index zero without any gate seeing it.

1. **Optional chaining.** `PrimaryKeys?.[0]` does not match `/\bPrimaryKeys\[0\]/`. Four sites on this branch, and two of them carry a `?? 'ID'` fallback, including `PullService.ts:468`, a file this PR edits. This is the one I would prioritize: the identifier is spelled exactly as the rule names it.
2. **Aliasing the array.** `const primaryKeys = entityInfo.PrimaryKeys; primaryKeys[0]` is invisible because the pattern is case sensitive. Fifteen sites across `primaryKeys[0]`, `pkFields[0]`, `pkPairs[0]` and `pks[0]`. I checked every one and all are correctly guarded by a length test, so nothing is broken today.
3. **Longhand.** `Fields.find(f => f.IsPrimaryKey)` is literally the body of the `FirstPrimaryKey` accessor. Eight sites, four in the ERD adapter. All benign, all foreign-key-target reads, and no regex can catch this shape.

Given 3, I would not try to make this airtight. The proportionate change is to list these shapes alongside the limitations the PR description already documents for `row.ID` and `OrderBy: 'ID'`. If you do want to close 1 and 2, note that widening the pattern will surface the fifteen guarded sites and each will then need a `// first-pk-ok:` annotation.

packages/MJCore/src/__tests__/PrimaryKeyCompliance.test.ts
```diff
@@ -0,0 +1,298 @@
...
+        const r = scanFile(file);
+        all.literalIdKeys.push(...r.literalIdKeys);
+        all.indexAccess.push(...r.indexAccess);
+        all.singleKey.push(...r.singleKey);
+        all.idPredicates.push(...r.idPredicates);
+    }
+    return all;
+}
+
+const RESULTS = scanAll();
```
rkihm-BC, line 235: The gate cannot distinguish "scanned nothing" from "found nothing." `findTsFiles` returns `[]` through its `fs.existsSync` early return, all four result arrays stay empty, and all four tests pass.

That matters more here than for the MJGlobal gates because this one deliberately ships without a baseline file, so a silent zero-file scan is indistinguishable from success. The realistic trigger is not a deleted directory but a move: relocating this file one directory deeper turns `SCAN_ROOT` into `packages/MJCore` and quietly narrows the scan to one package while staying green.

One assertion fixes it, for example a fifth test asserting the scanned file count exceeds a floor of a few thousand. Worth considering for `UUIDCompliance` and `MultiProviderCompliance` too, though that belongs in a separate change.

packages/SQLServerDataProvider/src/SQLServerDataProvider.ts
```diff
@@ -913,13 +919,30 @@ export class SQLServerDataProvider
...
+      return `(SELECT ${f.RelatedEntityFieldName} FROM [${entity.SchemaName}].${entity.BaseView} WHERE ${this.BuildFullPrimaryKeyPredicate(entity, CompositeKey)})`;
     }
   }
 
+  /**
+   * `pk1=v1 AND pk2=v2 ...` for every primary key column of `entity`, each value taken from the
+   * matching key/value pair (by field name, falling back to position) and quoted per that column's
+   * type. Single-column keys render exactly `ID='v'` — the same shape the pre-composite code emitted.
+   */
+  protected BuildFullPrimaryKeyPredicate(entity: EntityInfo, compositeKey: CompositeKey): string {
```
rkihm-BC, line 935: Optional, and explicitly not a regression. Key values are interpolated without doubling embedded single quotes, and nothing upstream escapes them, so an apostrophe in a key value breaks the generated SQL. The single-column line this replaces had the identical gap, so the exposure is unchanged.

Raising it only because you are already rewriting the line and the sibling code in `GenericDatabaseProvider.Load` does escape, with a comment noting that object-valued soft keys can legitimately carry free text. Adding `String(value).replace(/'/g, "''")` in the quoted branch would align the two.

packages/MJCore/src/generic/baseEntity.ts
```diff
@@ -5297,14 +5297,15 @@ export abstract class BaseEntity<T = unknown> {
         // this instance's provider so an uncommitted sibling on the same
         // graph connection is visible (host RunView would miss it).
         const rv = new RunView(this.RunViewProviderToUse);
-        const validSiblings = siblingChildEntities.filter(s => s.PrimaryKeys[0]);
+        // first-pk-ok: IS-A siblings share the parent's single key
```
rkihm-BC, line 5300: Nit: this annotation is duplicated. The standalone comment here and the trailing one on the next line carry the same text, and only the trailing one satisfies the gate. The standalone line can go.

rkihm-BC (approved):
## Approve

A 203-file sweep of this kind lives or dies on two things: whether the mechanical substitutions are truly mechanical, and whether the gate that is supposed to hold the line actually holds it. I verified both independently rather than taking the description's word for it. The sweep is sound. The gate is real but has three structural blind spots, and one sentence in the changeset overstates what was achieved. Nothing here blocks merge except that one line of changeset text, which is still editable.

### The failing check is not yours

`CodeGen drift gate` fails identically on `next` itself. Same four entries on both:

```
 M metadata/entity-permissions/.entity-permissions.json
 M packages/MJServer/src/generated/graphql-schemas/__mj.ts
?? migrations/v5/CodeGen_Run_<timestamp>.sql
```

It began failing on `next` between `ce3d5266` and `80653f1f`, the window that merged #4317, and that branch failed the same gate before merging. Because the list is byte-for-byte identical to `next`'s, it is also independent evidence for your byte-identical claim: if the generator edits had changed output for any MJ core entity, extra files would appear here and they do not.

### What I verified

- **The gate is real, and it passes.** I ran its exact scan logic against both trees. Pre-sweep `next` yields 130 / 38 / 196 / 6 across gates 1 to 4. This branch yields zero on all four. It executed in CI shard 3 and passed 4 tests, so `__dirname` resolves correctly in the ESM package.
- **CodeGen single-key output is unchanged.** With one identity column `callerSuppliedKeys` is empty, so `generateCRUDCreate` reduces to the previous string exactly. Both providers pass `excludePrimaryKey`, so the per-column `AutoIncrement` change in `generateInsertFieldString` cannot double-emit a column, and `pkHandledByStrategy` covers `PrimaryKeys.length > 1` on the Postgres side. The JSON-arg and typed-arg Postgres paths both reduce to the prior text for a single key.
- **The new REST composite `:id` segment is safe.** This needed checking because `InnerLoad` calls `CompositeKey.Validate()` without the entity argument, so the primary-key check is skipped there. `GenericDatabaseProvider.Load` then validates every key field against the entity's real primary keys and throws otherwise, and quotes the identifier from metadata rather than the caller's string. A caller cannot load by a non-key column and cannot inject through a field name.
- **The destructive geocoding fix is correct.** `BuildRecordIdExpression` reproduces `buildRecordId` exactly, one CAST for a single key and values glued with `||` for composite.
- **`SourceEntityID` was genuinely broken.** The old loop variable `entity` shadowed the outer `EntityInfo`, so `entity.ID` was the row's id. Renaming to `row` fixes it.
- **Changeset coverage is complete.** All 49 publishable packages with source changes are declared. The one omission is private. `patch` is right per `.claude/rules/changesets.md`.

### Please fix: one changeset claim is false

See the inline note on the changeset. Six `'ID'` fallbacks survive, two of them in files this PR edits, and one is the verbatim expression the line says is gone. None is a live defect on MJ core entities, and all are unchanged from `next`, so nothing regressed. But that sentence ships into generated release notes.

### Gate blind spots, worth documenting rather than chasing

Gate 2's pattern is `/\bPrimaryKeys\[0\]/`. Three shapes get past it and past gate 3:

| Shape | Sites on this branch | Status |
|---|---|---|
| `primaryKeys[0]`, `pkFields[0]`, `pkPairs[0]`, `pks[0]` | 15 | all correctly guarded |
| `PrimaryKeys?.[0]` | 4 | 2 carry a `?? 'ID'` fallback |
| `Fields.find(f => f.IsPrimaryKey)` | 8 | benign, and unreachable by any regex |

The optional-chaining miss is the notable one: the identifier is spelled exactly as the rule names it. The longhand form is literally the body of the `FirstPrimaryKey` accessor, so no text scanner can close it. My suggestion is to add these three shapes to the limitations list the description already carries, alongside `row.ID` and `OrderBy: 'ID'`, and widen the regex only if you want the two indexing forms. Widening will surface the 15 guarded sites and each will need an annotation.

### Two coverage notes for follow-up, not for this PR

**The gate only runs when a PR touches MJCore.** PR unit tests use `--filter=...[origin/next]`, which selects changed packages plus their dependents. `@memberjunction/core` is a dependency of nearly everything, not a dependent. I confirmed on #4319 that `@memberjunction/core:test` never ran at all. So a hardcoded `ID` reintroduced in an Angular package passes PR CI and fails only on the post-merge backstop on `next`. No other workflow runs this suite; `ci-standards.yml` only tests `@memberjunction/standards`. The same gap affects `UUIDCompliance` and `MultiProviderCompliance`, so it is a pattern-level issue.

There is a clean home for it if you want one later. The `guards` job in `test.yml` always runs, with no turbo filter and no pnpm install, and already hosts dependency-free gates invoked as `node scripts/<gate>.mjs packages`. A `.mjs` port of this scan would fit that shape exactly.

**The gate cannot tell "scanned nothing" from "found nothing."** See the inline note on `scanAll()`.

Happy to open the coverage item as an issue if you would like it tracked.

rkihm-BC:
Filed the gate-coverage follow-up from my review as #4326, so it does not hold this PR up. It covers all three repo-wide gates, not just the new one, since `UUIDCompliance` and `MultiProviderCompliance` have the same placement.

PR #4305, opened 2026-09-08
fix(security): close the #4282 role-elevation path on MJ: Roles and MJ: User Roles

### LINKED ISSUE #4282: A non-Owner can grant themselves a privileged role — MJ: Roles and MJ: User Roles have no elevation guard
**Found by** the PR gauntlet on #4275 (`fix/4260-new-user-roles-default`), 2026-09-07. Not caused by that PR: verified pre-existing at `f2ec156daf31670b42d7704e994686b09daf14df`. That PR guards `MJ: Users` only and names this as deliberately out of scope.

### Where
No server-side entity subclass exists for either entity. Compare `packages/MJCoreEntitiesServer/src/custom/MJUserEntityServer.server.ts`, which does this job for `MJ: Users`.

### What happens
Issue #4260 closed the `User.Type` escalation route: a non-Owner can no longer write `Type`, rename a user, create or delete users, or touch another user's row. Role assignment is a second, independent route to elevated capability and is unguarded. On a baseline seed the `Developer` and `Integration` roles hold unfiltered create/update/delete on `MJ: User Roles`, so a user holding either can insert a row linking their own user to any role in the system.

### What is NOT the case
A stronger version of this was proposed and checked: that a user auto-provisioned under the new \`['UI']\` default could grant themselves \`Developer\` and defeat the fix's reachability half in one call. Settled against a live database — \`UI\` holds \`CanCreate=0, CanUpdate=0, CanDelete=0\` on **both** \`MJ: User Roles\` and \`MJ: Roles\`, so it cannot. The reachability half stands. This issue is the narrower, real one: a user who **already** holds \`Developer\` or \`Integration\` can grant themselves more, because both roles hold all three permissions on both entities.

### Why it matters
The #4260 fix is about a user not being able to grant themselves authority they were not given. Roles are the platform's other authority mechanism. Leaving them open means the guarantee holds for one column and not for the concept.

### Repro
```bash
# Against a baseline-seeded database, as a user whose Type is 'User' holding the Developer role:
#   create a MJ: User Roles row with UserID = <your own user id>, RoleID = <any privileged role>
# Observed: the write succeeds. No guard consults the caller's Type or existing roles.
```
Confirm the grants first:
```sql
SELECT r.Name, ep.CanCreate, ep.CanUpdate, ep.CanDelete
FROM __mj.Entity e
JOIN __mj.EntityPermission ep ON ep.EntityID = e.ID
JOIN __mj.Role r ON r.ID = ep.RoleID
WHERE e.Name IN ('MJ: User Roles', 'MJ: Roles');
```

### Evidence
The same query shape against `MJ: Users` on a live dev database returned `Developer` and `Integration` with `CanCreate/CanUpdate/CanDelete = 1` and `UpdateRLSFilterID = null`; those roles are seeded with comparable breadth across entities.

### Also affected
Explorer's \`executeBulkRoleAssign\` (\`packages/Angular/Explorer/explorer-settings/src/lib/user-management/user-management.component.ts:948-957\`) creates \`MJ: User Roles\` rows directly and has no Owner gate. So a non-Owner administrator who is now refused every write on another user's \`MJ: Users\` row can still change that user's **roles** through the same screen — worth covering in the same pass.

### Suggested fix
Decide the invariant before writing code — this is genuinely harder than #4260, because unlike `User.Type` (two values, no legitimate non-Owner write) role assignment has real non-Owner use cases: delegated administration, group/IdP sync, onboarding automation. A defensible rule is "a caller may not grant a role they do not themselves hold, and may not grant a role that confers write access to `MJ: Roles`/`MJ: User Roles`", with Owners exempt. Once chosen, enforce it in an entity subclass (`Validate()` plus `Save()`/`Delete()` overrides) exactly as `MJUserEntityServer` does, so it holds on every write path and for any role a deployment invents — not in a resolver.

### Definition of done
- [ ] A failing test that reproduces the self-grant, then green
- [ ] A non-Owner cannot grant themselves (or another user) a role they do not hold; Owners unaffected; the documented legitimate paths still work
- [ ] Existing suite and gates green; no changed expectation in an existing test

### Verify by
The repro above is refused with a message naming the rule, and an integration check proves it against a live server the way IT88 does for `MJ: Users`.

SDesai-BC:
**A non-Owner holding `Developer` could grant themselves any role in the system, and mint new roles.** Issue #4260 closed the `User.Type` route to elevated capability; role assignment is MJ's *other* authority mechanism and had no guard at all.

Closes #4282.

## Reproduced first, on the real stack

Before writing any code, against a live database with a caller whose `Type` is `'User'` holding only `Developer` and `UI`, through the real provider and real `ClassFactory`:

```
R1  MJ: User Roles -> MJUserRoleEntity      ← generated class; no guard registered
R1  MJ: Roles      -> MJRoleEntity          ← same
R3  target role='Integration' — the caller does NOT hold it
R4  Validate() => Success=true errors=[none]
R4  Save()     => true                       ← the self-grant succeeds
R5  MJ: Roles create as non-Owner: Save() => true   ← can also mint a new role
```

Every probe write was wrapped in a provider transaction rolled back in a `finally`, re-read to prove the rollback took, and the database re-checked afterwards — the account still holds only `Developer` and `UI`.

The same probe after this change:

```
R1  MJ: User Roles -> MJUserRoleEntityServer
R1  MJ: Roles      -> MJRoleEntityServer
R4  Validate() => Success=false errors=[RoleID: You may only assign a role that you hold yourself. …]
R4  Save()     => false
R5  Validate() => Success=false ; Save() => false
```

## Root cause

No server-side entity subclass existed for either entity, so `ClassFactory` resolved the generated `MJUserRoleEntity` / `MJRoleEntity`, whose `Validate()` knows nothing about who is calling. Nothing else stands between the role grant and the write — confirmed against a live database rather than the seed:

| Entity | Role | Create | Update | Delete |
|---|---|---|---|---|
| `MJ: User Roles` | `Developer`, `Integration` | ✅ | ✅ | ✅ |
| `MJ: Roles` | `Developer`, `Integration` | ✅ | ✅ | ✅ |

All three `Allow*API` flags are true on both entities.

## The invariant, and why this one

This is the part worth reviewing. Unlike `User.Type` — two values, no legitimate non-Owner write — role assignment has real non-Owner uses the platform ships with: delegated administration, IdP/group sync, onboarding automation. So the rule had to be chosen rather than maximised.

**`MJUserRoleEntityServer` — a non-Owner may only grant, move or revoke a role they hold themselves.**

The subset rule is a **ceiling**: whatever a non-Owner does through this entity, the authority they hand out is authority they already had, so no sequence of calls lets any caller exceed their own grant. A `Developer` self-granting `Integration` is refused; a `Developer` onboarding another developer still works.

The issue also floated adding *"and may not grant a role that confers write access to `MJ: Roles`/`MJ: User Roles`"*. **Considered and declined**, for three reasons: it closes no elevation the subset rule leaves open (handing a peer a role you already hold is delegation, not escape); it breaks the peer-onboarding case above; and it would force the guard to read `EntityInfo.Permissions`, trading its purity for protection that `MJ: Entity Permissions` — carrying the same unfiltered grant — undercuts anyway.

Two consequences of the rule that are deliberate rather than incidental:

- On **update**, the *pre-save* `RoleID` is checked alongside the new one. Without that half, a non-Owner could strip anyone of a role the caller does not hold by simply repointing the row at one they do.
- `UserID` is **not** frozen. Moving a grant between users stays inside the same ceiling, and freezing it would break delegated administration for no security gain.

**`MJRoleEntityServer` — a non-Owner may not create, change or delete a role.** Blunt, because every field on this entity is authority-bearing: `Name` is what `SyncRolesUsersResolver` matches roles by, `DirectoryID` maps an external directory group onto the role, and `SQLName` decides which *database* role CodeGen emits `GRANT`s for. Freezing a field list instead would leave that list to be re-derived every time a field is added; refusing the operation cannot go stale that way, and it matches what #4260 chose for `MJ: Users` create/delete.

Both guards override `Save()` and `Delete()` alongside `Validate()`. `Validate()` alone is not "every write path": `BaseEntity.Save()` force-passes validation *without calling `Validate()`* under `EntitySaveOptions.ReplayOnly` (`baseEntity.ts:3725`) while still performing the write, and `Delete()` never consults `Validate()` at all. Both classes are pure — they read only the record's own field state and the caller's already-cached `UserInfo.UserRoles`, so they cost nothing per save and are unit-testable without a database.

## Explorer bulk role assignment was silently succeeding

`executeBulkRoleAssign` (`user-management.component.ts:940`) ignored each row's `Save()` return. Inside a transaction group `Save()` only *enrols* the row — but it validates eagerly, so a refused row returns `false` and is never enrolled. An all-refused batch therefore left the group empty, `Submit()` returned `true` for having nothing to do, and the screen reported success having assigned nothing. It now collects each refusal with the user it applies to and surfaces them. Pre-existing, but this change is what makes it the common path.

## Deliberately out of scope

**`MJ: Entity Permissions` carries the same unfiltered `Developer`/`Integration` create/update/delete grant**, so a holder of either role can still widen a role's permissions directly. That is an independent route with its own invariant to decide, and fixing it in passing here would be guessing at that decision. This PR does not claim to close it — worth its own issue.

## Evidence

Everything below was run; nothing is inferred.

| What | Result |
|---|---|
| `MJCoreEntitiesServer` unit suite | **517 passed** (28 files), including 41 new tests across the two guards |
| Mutation check — neuter `callerIsOwner()` to always return `true` | **19 of 41 fail**, all 41 green again on restore, so the tests are load-bearing |
| `integration-test-suite` unit suite | **201 passed** (7 files) |
| `ng-explorer-settings` unit suite | **100 passed** (15 files) |
| New `role-elevation` bundle (RE1–RE6) against a real provider, `RUN_MUTATION_TESTS=1` | **6 passed, 0 failed, 0 skipped** — including RE6's real `Save()` in a rolled-back transaction, rollback verified |
| Live re-probe after the fix | refusals shown above; database re-checked, no stray grants or roles |
| `check:standards`, `check:ui`, `check:codegen-tail`, `check:changeset` | pass |
| Builds | `MJCoreEntitiesServer`, `integration-test-suite`, `ng-explorer-settings` all compile clean |

RE1–RE6 mirror IT88's doctrine for `MJ: Users`: RE1 is the export-barrel regression guard (drop either `export *` line and the guard becomes dead code with nothing else noticing); RE2 is the escalation itself; **RE3 and RE4 are as load-bearing as RE2** — they are what prove the guard did not over-reach and close delegated administration or the admin path.

**Not run: the full deterministic integration tier.** It needs `mj sync push` against the shared `MJ_ATS_Dev` and an MJAPI restart, neither of which seemed mine to do unilaterally on a shared workspace database. Mitigating evidence: no existing bundle *writes* these entities — `permission-engine`, `ai-permissions`, `runquery-cache` and `runview-matrix` all only `RunView` them, and these guards affect writes only.

## Before merging / after merging

- **`metadata-optional/**` is not in any database until it is pushed.** IT89 and its suite membership are declarative JSON; the bundle does not run anywhere until `pnpm mj sync push --dir=metadata-optional/integration-test`. A merged PR is a no-op for the test tier until then.
- **`MJ_INTEGRATION_TEST=1 pnpm mj test run "IT89 - Role Privilege Elevation Guard"`** is the single-bundle command once pushed.
- **Behaviour change for non-Owner administrators.** Explorer's role-management screen has no Owner gate of its own, so a `Developer`-role non-Owner reaches it today; creating, renaming and deleting roles from it are now refused. Same trade-off #4260 accepted for the user-management screen, and written up as an upgrade note in the changeset.
- **`SyncRoles` / `SyncUsers` / `SyncRolesAndUsers` are unaffected on a default install** — all three carry `@RequireSystemUser()` and `getSystemUser()` resolves the seeded `Type='Owner'` system user. A deployment whose system user is *not* an Owner will see them fail closed at the save, loudly, for the same reason #4260 documented.
- **The two changed files under `src/generated/`** (`ServerBootstrap`, `ServerBootstrapLite`) are `pnpm run mj:manifest` output, not hand edits — the anti-tree-shaking manifests, which CI checks for freshness. The browser manifests are correctly untouched: `core-entities-server` is server-only.
- `check-registry.test.ts` changes are the registry **pins** a new bundle is required to update (Method 2, step 6 of the integration-testing guide) — the bundle's check count, its mutation-gated member, and the catalog size. No existing test's expectation was weakened.


.changeset/role-elevation-guard.md
```diff
@@ -0,0 +1,28 @@
...
+
+**Upgrade notes.**
+
+- **Explorer's role-management screen stops working for non-Owner administrators.** It has no Owner gate of its own today, so a Developer-role non-Owner reaches it in practice; creating, renaming and deleting roles from it are now refused with a message naming the rule. This is the same trade-off #4260 accepted for the user-management screen.
+- **Explorer's bulk role assignment now checks each `Save()` return.** `executeBulkRoleAssign` enrols each row in a transaction group, where `Save()` reports only *enrolment* — the provider queues the item locally and returns `true` without a round trip. A row refused **client-side** (a `CheckPermissions` denial or a field-rule failure) does return `false` and is never enrolled; that return was ignored, so an all-refused batch left the group empty, and an empty group's `Submit()` returns `true` for having nothing to do — the screen reported success having assigned nothing. It now collects each such refusal with the user it applies to and surfaces them.
```
rkihm-BC, line 23: Missing from this list: `AssignUserRolesAction` (`CoreActions/src/custom/user-management/assign-user-roles.action.ts:99`) builds its `MJ: User Roles` object with `params.ContextUser` and throws on a failed `Save()`, rolling back the whole batch. Invoked by a non-Owner assigning a role the caller doesn't hold, the action now fails outright — consistent with the rule, but it's the "onboarding automation" case the rationale above names as preserved, and it deserves the same one-line upgrade note the other affected paths get.

packages/Angular/Explorer/explorer-settings/src/lib/user-management/user-management.component.ts
```diff
@@ -952,13 +952,43 @@ export class UserManagementComponent extends BaseDashboard implements OnDestroy
...
+        // a live server: a non-Owner assigning a role they do not hold is correctly refused (no row
+        // is written) and is reported as success. Tracked as issue #4309 — fixing it belongs in the
+        // resolver, which is the only layer that still knows which row was refused and why.
+        //
+        // What this check DOES catch is a CLIENT-side refusal — a CheckPermissions denial or a
+        // field-rule failure — which really does return false here, leaving that row unenrolled.
+        // Ignoring the return meant an all-refused batch left the group EMPTY, and an empty group's
+        // Submit() returns true for having nothing to do, so the screen reported success having
+        // assigned nothing. That is the failure this guards; keep it.
+        const refusals: string[] = [];
```
rkihm-BC, line 977: Same file's sibling on the single-user path, `user-dialog.component.ts` `updateUserRoles`, still discards both `await userRole.Delete()` and `await userRole.Save()` inside its transaction group — the identical pattern, on the same screen, now under the same guard. Granted the user-visible outcome is the same either way (in a TG the client-side `Save()` returns true and the refusal is #4309's to surface), so this is consistency rather than a fix. Either apply the same treatment or say why only the bulk path was treated.

rkihm-BC (dismissed):
Same doctrine as #4275, and the invariant choice is argued rather than assumed. Verified against the code, not the description:

- **The subset rule is a genuine ceiling.** Whatever a non-Owner hands out is authority they already had, so no call sequence exceeds the caller's own grant — and the declined stricter variant really would have broken peer onboarding without closing anything the ceiling leaves open.
- **The pre-save `RoleID` half is necessary, not decorative** — without it a non-Owner strips a role they don't hold by repointing the row — and it fails closed with its own message when the prior value can't be established.
- `UUIDsEqual` for role matching and the trimmed/lower-cased `Type` comparison are both right for this schema (NCHAR, two casings).
- `Save()` and `Delete()` overridden alongside `Validate()`, so `ReplayOnly` can't switch invariants 1–2 off while 3 stays enforced.
- Manifests regenerated for both bootstrap packages with counts updated (1026→1028, 942→944); browser manifests correctly untouched.
- The suite `Sequence: 48` is not a constraint problem — the unique key is `(SuiteID, TestID)`, and duplicate sequences are already the established pattern in that file.

21/21 checks green. Approving; two notes below, neither blocking.

### Note 1 — a shipped core action changes behaviour and isn't in the upgrade notes

`AssignUserRolesAction` (`packages/Actions/CoreActions/src/custom/user-management/assign-user-roles.action.ts:99`) builds its entity with `params.ContextUser` and **throws on a failed `Save()`, rolling the whole batch back**. Under this guard, that action invoked by a non-Owner assigning a role the caller does not hold now fails entirely. That is consistent with the rule you chose — but the changeset carefully enumerates Explorer role management, the bulk assign and the three sync mutations, and misses this one. "Onboarding automation" is named as a use case the subset rule preserves; it is preserved only when the automation's context user holds the role. Worth a line in the upgrade notes. Inline on the changeset.

### Note 2 — the sibling Explorer path keeps the defect this PR fixes

`user-dialog.component.ts` (`updateUserRoles`) discards both `await userRole.Delete()` and `await userRole.Save()` inside a transaction group — the identical pattern fixed here in `executeBulkRoleAssign`, on the same screen, now subject to the same guard. In fairness the user-visible outcome is unchanged either way, because inside a TG the client-side `Save()` returns true regardless and the real refusal is swallowed by #4309 — so this is symmetry, not a user-facing fix. Worth applying or explaining why only the bulk path was treated. Inline comment.

### One thing to confirm rather than change

Because `UserID` is deliberately not frozen and revocation shares the granting ceiling, a `Developer` may repoint or delete **any** user's `Developer` assignment, including an Owner's. That is not escalation — Owner authority lives in `Type`, not the role — but it does let one non-Owner strip every peer of a role they share. You chose this explicitly; I'd just like it confirmed as intended rather than incidental, since it is the one consequence of "not frozen" that isn't self-evidently harmless.

Trivial: this changeset lists five packages including the two bootstraps, where #4275 (same manifest regeneration) listed only `core-entities-server`. This one is arguably the more correct of the two; noting the inconsistency only so whichever is right becomes the convention.

rkihm-BC:
@SDesai-BC , take a look at the review comments in the PR Review. They aren't blocking but there may be some things to address here and others that should be captured as issues to address in the future.

.changeset/role-elevation-guard.md
SDesai-BC: Added as an upgrade note, with the mechanism rather than just the fact.

Worth separating out while it's here: this action is the **only** affected path that reports its refusal correctly. It uses a real provider transaction rather than a client transaction group, so the refused `Save()` returns false, the throw fires, `RollbackTransaction()` runs, and the caller gets `Success: false` / `ResultCode: 'FAILED'` carrying the guard's own message. The two Explorer paths structurally cannot — that's #4309.

Pinned so the note can't go stale: a new `assign-user-roles.action.test.ts` (7 tests) asserts the `params.ContextUser` build at `:99`, the `FAILED` result code, the guard's message reaching `result.Message`, the rollback, and `AssignedCount: 0` after it. These are characterization rather than a fix — the behaviour was already right, exactly as you scoped it. Mutation check to confirm they bite rather than decorate: replacing `if (!await userRole.Save()) { throw ... }` with a bare `await userRole.Save();` fails 4 of the 7.

Also added `@memberjunction/core-actions` to the front matter, since the branch now touches that package.

packages/Angular/Explorer/explorer-settings/src/lib/user-management/user-management.component.ts
SDesai-BC: Applied — `updateUserRoles` now collects refusals from both loops and throws before `Submit()`.

One correction to the framing, which came out of writing the test: for this path the user-visible outcome is **not** the same either way. `executeBulkRoleAssign` fails whole-batch, but `updateUserRoles` mixes removals and additions, so a **partial** client-side refusal still reached `tg.Submit()` — the surviving rows were written and the dialog closed reporting success on a partial change. That case now throws with nothing submitted. So it's a real fix here, not only symmetry.

RED first: `user-dialog-role-refusal.dom.test.ts` — 6 failed / 2 passed against the old body, the 2 passing being the happy-path guards that must stay green on both sides. 8/8 after; package suite 100 → 108.

I did not extract a shared helper with the bulk path. The common part is four lines of control flow and the variation is real: `Delete()` *and* `Save()` rather than `Save()` alone, role names rather than user names, a different message. What shouldn't be duplicated is the *explanation* — so the new comment points at `executeBulkRoleAssign` as the canonical statement of the transaction-group semantics instead of repeating twenty lines of it, and a new `describeRole()` mirrors the existing `describeUser()`.

SDesai-BC:
@rkihm-BC — all four items addressed: two fixed here, two answered, and two follow-ups filed. Inline replies on the notes; the other two are below.

**Confirm item — non-frozen `UserID` letting a non-Owner strip a peer's shared role: intended, and now documented rather than left to inference.** You're right that it's the one consequence of "not frozen" that isn't self-evidently harmless, so it shouldn't have been something a reader had to derive. The changeset now states it outright: this is *deprivation*, not elevation — Owner authority lives in `User.Type`, which #4260 froze, and not in a role. It is reversible by an Owner and recorded in Record Changes. The alternative, narrowing revocation to the caller's own row, closes it only by breaking delegated administration — which is the legitimate non-Owner use the subset rule exists to preserve in the first place.

**Trivial — the changeset package list.** Checked `d0eab882c4`: #4275 did regenerate both bootstrap manifests and listed only `core-entities-server`, so your read holds and this PR is the more correct of the two. `.claude/rules/changesets.md` is thorough about bump *levels* and deliberately silent on membership, so the convention is genuinely unwritten rather than inconsistently applied — and because all ~300 packages sit in one `fixed` group, membership decides only whose `CHANGELOG.md` carries the entry, never a version. Filed as #4328 with a proposed rule addition. Not touching #4275, which is merged and its changeset consumed.

**Also filed #4327**, from the neighbourhood of your review rather than from it directly: `MJ: Entity Permissions` still carries the same unfiltered `Developer`/`Integration` grant. The changeset names it as out of scope, but nothing was tracking it, so it existed only as a sentence due to be consumed at the next release. It's worth its own issue because the ceiling #4282 builds assumes the authority attached to a role is fixed — a `Developer` who cannot grant themselves `Integration` can instead widen what `Developer` itself is permitted to do, and arrive at the same place without ever touching a role assignment.

Gates on the updated branch: `ng-explorer-settings` 108 passed (was 100), `core-actions` 502 passed, both builds clean, and `check:changeset` / `check:codegen-tail` / `check:esm` all pass. The local deterministic integration tier shows the same six pre-existing failures it shows on `next` (`AW3`, `QC1`, `AE1`, `CD3`, `CD5`, `CD6`), plus two `permission-engine` checks that are failing on a stale fixture row left behind by my own UI testing rather than on anything in this branch — `it-nogrant@integration.test` must hold no roles and currently holds one. Clearing that separately.

rkihm-BC (approved):
Re-approving. All three items from my earlier review are addressed, and I checked the fixes rather than taking the commit message for them.

## Note 2 — the user dialog: fixed, and the fix is correct where it matters

`updateUserRoles` now checks both returns, collects refusals across the removal and addition loops, and **throws before `Submit()`**. That ordering is the whole thing, and the comment states why precisely: the refused rows never enrolled, the accepted ones are still only queued, so nothing is written. The abandoned transaction group is client-side queue state with no server round trip behind it — no dangling state to clean up.

`describeRole` using `UUIDsEqual` is consistent with the rest of the PR, and falling back to the raw ID when the catalog has no match is the right call for a message that exists to tell someone what was refused.

The eight DOM tests constrain behaviour rather than pad coverage — partial refusal writing nothing, a refused `Delete()` treated the same as a refused `Save()`, the fallback when `LatestResult` is absent, and both untouched paths (all-accepted, and nothing-changed creating no group at all).

Worth restating so the scope stays clear, and the changeset already does: this catches **client-side** refusals. The #4282 guard this PR adds is server-side, so for the actual elevation case both Explorer paths still close reporting success until #4309 lands. That is documented rather than glossed, including the end-to-end verification that the guard *works* while the refusal *is not reported*.

## Note 1 — the action: documented properly, not just packaged

`@memberjunction/core-actions` added to the changeset, and more importantly a full upgrade-note bullet that says what actually changes: `AssignUserRolesAction` fails the whole batch, returns `Success: false` with the guard's message, nothing partially assigned. The sentence that does the real work is the last one — onboarding automation is preserved *where the automation's context user holds the role being granted*, otherwise give it those roles or an Owner context user. That is the operational instruction a deployer needs, and it was the gap I raised.

Seven new tests on that action, including the rollback and the output params not claiming a write after one.

## The thing I asked you to confirm — answered, and better than confirmed

You did not just say "intended". The changeset now carries the reasoning: deprivation rather than elevation, Owner authority lives in `User.Type` which #4260 froze, reversible by an Owner, recorded in Record Changes, and narrowing revocation to the caller's own row would close it only by breaking the delegated administration the rule exists to preserve. That is the argument I wanted to see made in the artifact rather than in a thread, because it is the paragraph someone will need in a year.

## State

19 checks pass, 2 still running (`CodeGen drift gate`, `Native-ESM import guard`), 0 behind `next`, and nothing unexpected has appeared in the diff — still the 17 files I reviewed. Approving on the assumption those two land green; if the drift gate reports anything, it will be a real signal now that it is fixed on `next`.

The trivial changeset-scope inconsistency with #4275 is unchanged and still doesn't matter — noting only that this PR remains the more correct of the two.

PR #837, opened 2025-02-24
Update communication providers to support forwarding and replying to messages

JS-BC:
This is largely copying over the changes made in the Izzy project over to MJ

PR #4310, opened 2026-09-08
feat(whiteboard): host-controllable ToolRoster (P12 / P14 roster half)

MattC-BC:
First of the whiteboard chrome inputs from the LXP alignment plan (Sidecar PR #837, plan 03).
Additive, defaulted to today's rendering, patch changeset on `@memberjunction/ng-whiteboard`.

## The defect

LXP hides five of the eleven tools with CSS. Hiding is not disabling: the keyboard map still
armed `r m w i c`, and the right-click canvas menu still offered "Add markdown panel here" and
"Add widget here", so a learner could press `w` and place an HTML widget on a board whose toolbar
showed no such tool (Sidecar #845).

So this is not "a nicer way to hide buttons". Gating only the toolbar would be the CSS hack with
an API in front of it — the keyboard and the context menu are exactly how a hidden tool stayed
reachable. Those three are one thing: doors to activating a tool.

## The input

`RealtimeWhiteboardHostComponent.ToolRoster: readonly WhiteboardTool[] | null` (default `null` =
all eleven). Threaded to the toolbar (`VisibleTools`, always in `Tools` order, unknown names
ignored) and to the board (`BuildWhiteboardContextMenu(item, roster)`).

**The invariant is enforced on writes, not reads.** "The active tool is always one the roster
allows" is a property of every write to `Tool`, not of the handful of call sites that happen to
exist today. `Tool` is therefore a setter that ignores a disallowed write, and the individual
doors close by construction — a shortcut, gesture or host API added later cannot reopen the hole
by forgetting a guard. It also makes assignment order irrelevant, which matters because a channel's
`BindSurface` sets `State`, `AgentName` and `ToolRoster` as plain properties in sequence.

Two behaviours worth stating:

- A **disallowed request is ignored, not redirected**. Escape under `['pen','eraser']` leaves you
  on pen rather than bouncing to a fallback.
- When the tool you are holding **leaves** the roster, the clamp prefers `select`, falling back to
  the roster's first *known* entry. Roster order is documented as meaningless for rendering, so it
  must not quietly become load-bearing here.

It is a **setter rather than `ngOnChanges`**, per packages/Angular/CLAUDE.md. That is not style:
Angular fires no `ngOnChanges` for a plain property assignment, so a clamp implemented there
silently never ran on the exact path LXP will use. It was verified failing that way in the browser
before it moved, and a spec now pins the direct-assignment path.

## Scope: tool availability, not content policy

The roster governs **which tools are available**. It deliberately does not restrict what already
exists on the board, what the agent places, or authoring on existing items — Restyle, Duplicate
and pasting an image all keep working under any roster.

An earlier revision of this branch also gated item creation and Restyle. That was removed: it
answered a question nobody asked, and it cost real UX (a pasted screenshot vanishing with no
explanation; a Duplicate entry that did nothing when clicked). Read-only is a separate axis and a
separate input. The README and changeset now say this explicitly rather than claiming a hidden
tool is "unreachable everywhere".

## Behavior change: none by default

Every default reproduces today's rendering; the existing context-menu and toolbar tests are the
controls and pass unchanged. `BuildWhiteboardContextMenu(item)` without a roster is unchanged.

## Housekeeping

`WhiteboardTool` and a runtime `WHITEBOARD_TOOLS` catalog move into `whiteboard-tool-roster.ts`,
so the clamp can distinguish a tool from a typo. The toolbar imports the type and does **not**
re-export it — `public-api.ts` `export *`s both modules, and two modules exporting one name that
way is a TS2308 ambiguity. The published surface is unchanged; no package outside the whiteboard
imports `WhiteboardTool`. Helper names are PascalCase to match the other sixteen exported
functions in the package, and `IsToolAllowed` tolerates `undefined` because this package compiles
with `strictNullChecks: false`.

## Not in this PR

The `Show*` chrome booleans (`ShowHeader`, `ShowFocusToggle`, `ShowStatusBar`, ...),
`ShowAuthoringControls`, and binding the board's existing `ReadOnly` through the host: next PR,
targeting edge.7. Per-item `Tag` (P14 #1): its own PR.

## Tests

18 files / **350 passed** in the package; `test:types` green. New coverage spans a node spec for
the pure helpers, the context-menu spec, the toolbar DOM spec, and the package's first host DOM
spec. Every roster path was mutation-checked: removing the `Tool` setter's guard fails six cases
including the keyboard one, and removing either `[ToolRoster]` binding fails exactly the cases
that cover it.

## Verified in the browser

Driven in MJ Explorer against a local API, in a live realtime session with the whiteboard channel,
setting the input from `BindSurface` (the LXP recipe).

| # | Probe | Result |
|---|---|---|
| P1 | default roster unchanged | 13 toolbar buttons in `Tools` order; `w` selects the widget tool; all five canvas actions |
| P2 | roster narrows the toolbar | `Select, Pan, Pen, Sticky, Text, Eraser` + undo/redo |
| P3 | hidden tool keys are no-ops | `w m r i c` no-op; `p s t` still work (positive controls, focus verified before each press) |
| P4 | the right-click door | exactly `Add sticky note here`, `Add text here`, `New page` |
| P5 | item menus untouched | full menu incl. Restyle, Duplicate, Delete |
| P6 | #4121 focus scope holds | with focus outside the board, `p` does nothing |
| P7 | clamp on mount | roster `['pan','pen']` mounts with Pan active, not Select |
| P8 | empty roster | undo/redo only, `New page` alone with no separator, no console errors |
| P9 | disallowed write ignored | under `['pen','eraser']`, Escape leaves the Pen tool active |
| P10 | content is not gated | under a roster hiding the image tool, Cmd+V still places an image and Duplicate still copies |
| P11 | `undefined` roster | normalizes to `null`, full palette, no throw |


rkihm-BC (approved):
The design call is right, and I checked the load-bearing claims rather than taking them from the description. Approving.

**Enforcing on the write is the correct choice, and the DOM spec proves the part that matters.** `renderComponentFixture` runs its `setup` callback after `setInput` and **before** the first `detectChanges()` (`test-utils/src/lib/render-component-fixture.ts:78`), so the "assigns `ToolRoster` directly, no binding" case really does exercise the `BindSurface` path — the one where an `ngOnChanges` clamp would have silently never run. That test is the reason to believe the rest.

**The type move is safe.** Nothing in the package still imports `WhiteboardTool` from `whiteboard-toolbar.component` after this change, and no package outside the whiteboard imports the name at all (the `conversations` hits are `WhiteboardToolResult`, unrelated). `public-api` still exports it from the roster module, so the published surface is unchanged and the TS2308 reasoning holds.

**Clamp semantics are sound and pinned**: `select` preferred over roster order so ordering never quietly becomes load-bearing; a disallowed write ignored rather than redirected; an empty or all-typo roster leaving the current tool alone; the context menu's separator tied to whether any add actions survive. The `== null` in `IsToolAllowed` is deliberate and documented for a `strictNullChecks: false` package.

**Scope is right, and the removed revision was right to remove.** Gating item creation and Restyle answered a question nobody asked and cost real UX; "availability, not content policy" is the line that leaves `ReadOnly` a coherent separate axis. The README and changeset both say so rather than overclaiming that a hidden tool is unreachable everywhere.

`tsconfig.spec.json` extends the package's existing explicit include list — pre-existing pattern, extended correctly. 22/22 checks green. Nothing I'd ask you to change.

PR #4311, opened 2026-09-08
Plan: Skip form contributions (metadata-registered form panels)

CaelebB-BC:
## Summary

Planning-only PR. It adds two documents under `plans/skip-form-contributions/` and changes no code:

- **`design.md`** — the proposed design for **form contributions from metadata** (pending team approval).
- **`implementation.md`** — a 21-task implementation plan across MJ, Skip-Brain, and Skip-Client-Open-App.

## The problem

MJ composes an entity form at runtime from **contributions** (baked field panels, related grids, heroes, widgets) arranged by the chrome layers L0–L4. Today a contribution can only be a compiled Angular `BaseFormPanel` registered through `ClassFactory`. Skip ships no Angular, so the only thing it can place on the form surface is a **whole-form replacement** (`EntityFormOverride` → `componentRole: 'form'`), and that replacement is opaque to the contribution and chrome stack: no sections for the rail, fill-in grids duplicated under hand-drawn related tables, compiled `replacesSectionKey` claims that hide nothing, and a schema snapshot that never learns about a downstream app installed later.

## The design

Add a **second source of contributions (a database row) and a second renderer (a React component with `componentRole: 'form-panel'`)**. The existing composer, slot hosts, last-wins collapse, and chrome layers treat both sources as peers. Skip writes rows, not code. Whole-form replacement stays as the escalation path and gains the CodeGen slots so other contributions still land in position.

| Piece | What it is |
|---|---|
| `MJ: Entity Form Contributions` | New table mirroring `FormPanelRegistrationMetadata` (slot, sort key, contribution key, related-grid claim, `replacesSectionKey`, inclusion, chrome group) plus `Presentation` (`panel` / `bare`), title, icon, `Configuration`, and User / Role / Global scope with `Active` / `Pending` / `Inactive` status |
| `componentRole: 'form-panel'` | New role in `@memberjunction/interactive-component-types` with `FormPanelHostProps`, `FormContributionSpec` (carried on the spec as `formContribution`), and panel events / methods (`RowCountChanged`, `FieldChanged`, `Validate`, …) |
| Merged collector | `CollectFormContributionRegistrations` merges `ClassFactory` registrations with Active, scope-matching rows from `InteractiveFormsEngine`; memoized per entity + user |
| `InteractiveFormPanelComponent` | Generic `BaseFormPanel` that mounts a row's React component through `<mj-react-component>` inside a collapsible panel (or bare, for heroes) |
| Actions | `Create` / `Modify` / `Activate Form Contribution`, `Get Form Contributions For Entity`, `Get Form Composition For Entity` — same security clamp as the interactive-form family (agents write User scope only) |
| Apply path | The artifact viewer detects panel specs, previews them against a real record, and offers **Add to my form** |
| Composition snapshot | The container publishes what is on the form (sections, related grids with inclusion, contributions, slots); the record tab forwards it to agent context, the Skip proxy sends it as `formContext`, and Skip renders it as a `[FORM CONTEXT]` marker so a panel targets a real slot and section key |
| Skip-Brain | `form-panel` intent, a shared panel contract prompt, commitment stamping, and three panel gates |

## Decisions recorded in the design

1. On an equal-priority tie between a row and a compiled panel with the same key, **compiled wins**; a row outranks only with strictly higher `Priority`, set to `incumbent + 1` after the user confirms.
2. Panel components are `MJ: Components` rows with `Type='Widget'`; whole forms stay `Type='Form'`.
3. OpenApps without Angular may ship `Scope='Global'` rows via `mj sync` with no promotion gate; L3 rules can still suppress per site.
4. `replacesSectionKey` is validated, not restricted: apply-time check against the snapshot with an extra-pane fallback, render-time diagnostic, and a prompt limited to `[FORM CONTEXT]` keys.
5. The full composition snapshot goes to Skip on every message.

## Implementation phases

| Phase | Tasks | Delivers |
|---|---|---|
| **A** Contract + runtime (MJ) | A1–A10 | Role types, migration + CodeGen, engine load, merged collector, panel host, slot-host mount, form and container wiring, snapshot, interactive-form slots |
| **B** Apply path (MJ) | B1–B4 | Action family and metadata, artifact viewer preview and button, `ApplyContribution` |
| **C** Skip-Brain | C1–C4 | Contract prompt, intent routing, commitment, gates, `[FORM CONTEXT]` intake |
| **D** Context | D1–D3 | Record tab publishes the snapshot, Skip Client forwards it, server-side fallback action |
| **E** Parity | deferred | Form Builder / Form Studio / Component Studio get their own plan after B |
| **F** Docs + changeset | one task | Forms guide, PANELS.md, Angular CLAUDE.md |

Each task in `implementation.md` lists exact files, the interfaces it consumes and produces, a failing test first, the implementation, run commands with expected output, and a staging step. Cleanups folded in along the way: strict entity matching shared by slot host and composer, memoized registration collection (the `formContext` getter currently rescans `ClassFactory` on every change-detection pass), `BaseFormPanel.validate()` actually invoked on save, and the `contributionKey === 'header'` magic string replaced by `Presentation`.

## Review notes

- No code, migration, or metadata changes in this PR, so no changeset.
- Cross-repo ordering: Phase C needs the Phase A1 types and the Phase D2 `@askskip/types` change published (or yalc-linked) before it compiles; the plan flags this at the top of Phase C.

https://claude.ai/code/session_013X7azw5RVBAWPb9nKd2Cgk

### DIFF
--- a/plans/skip-form-contributions/design.md
+++ b/plans/skip-form-contributions/design.md
@@ -1,0 +1,271 @@
+# Skip Form Contributions — Design
+
+**Status:** Approved design (team review 2026-09-08); implementation plan in [implementation.md](implementation.md)
+**Date:** 2026-09-08
+**Foundations:** [`../form-contributions.md`](../form-contributions.md) · [`../form-chrome-layering.md`](../form-chrome-layering.md) · [`../interactive-forms/plan.md`](../interactive-forms/plan.md) · [Forms Architecture Guide §7c–7d](../../guides/FORMS_ARCHITECTURE_GUIDE.md) · [PANELS.md](../../packages/Angular/Generic/base-forms/PANELS.md)
+**Repos:** MemberJunction (`MJ`), Skip-Brain, Skip-Client-Open-App
+
+---
+
+## 1. Summary
+
+MemberJunction composes an entity form at runtime from **contributions**: baked field panels, related grids, heroes, and widgets, arranged by five chrome layers (L0 CodeGen → L1 app inclusions → L2 ranker → L3 admin rules → L4 user). Today a contribution can only be a compiled Angular `BaseFormPanel`. Skip ships no Angular, so the only thing Skip can put on the form surface is a **whole-form replacement**, which is opaque to that composition.
+
+This design adds a **second source of contributions, a database row, and a second renderer, a React component with `componentRole: 'form-panel'`**. The existing composer, slot hosts, last-wins collapse, and chrome layers treat the two sources as peers. Skip writes rows, not code. A Skip hero on Person, a revenue waterfall on Subscription, or a card view that replaces the Orders grid is the same kind of object an installed OpenApp contributes, and it keeps composing correctly as other apps install.
+
+Whole-form replacement stays and becomes the escalation path.
+
+## 2. Background
+
+### 2.1 Compiled contributions
+
+`@RegisterClassEx(BaseFormPanel, { metadata })` registers a panel. The metadata bag carries `entity`, `slot`, `sortKey`, `contributionKey`, `relatedEntity` + `relatedJoinField`, `replacesSectionKey`, `inclusion`, and `chromeGroup`. `ResolveFormContributions` (pure) collapses registrations by key, decides which baked sections to hide, and lists stock grids to fill in. `<mj-form-panel-slot>` mounts winners with a forward fallback chain; `<mj-record-form-container>` always hosts `<mj-form-contributions>` and an `after-everything` slot. Chrome reads inclusion, sort key, and chrome group from the same registrations; `MJ: Form Chrome Rules` (L3) can suppress a contribution by key.
+
+Discovery is `ClassFactory.GetAllRegistrationsByMetadata`. That is the one input to the "metadata forms framework" that is not metadata.
+
+### 2.2 Interactive whole-form replacement
+
+`MJ: Entity Form Overrides` binds an entity to a `MJ: Components` row (`Type='Form'`, `componentRole: 'form'`) at User, Role, or Global scope. `InteractiveFormComponent extends BaseFormComponent` owns the `BaseEntity`, renders `<mj-react-component>` inside the container, and bridges `FormHostProps` in and `BeforeSave` out. `InteractiveFormsEngine` caches forms and overrides with event-driven invalidation. Authoring paths: the Form Builder agent, Form Studio, Component Studio, `mj sync`, and the artifact viewer's **Apply to my form** button.
+
+### 2.3 How Skip authors a form
+
+The `#` picker emits an entity token; Skip's `EntityMentionResolver` turns it into a `[TARGET ENTITY]` marker. The Requirements Expert sets PRD `type: 'form'`; `form-role-contract.md` is included into the architect and code-generator prompts; five form gates enforce the contract; `WorkflowService.applyFormRoleCommitment` stamps `componentRole` and `entityName`; the spec returns as a conversation artifact; **Apply to my form** runs `Create` or `Modify Interactive Form`.
+
+### 2.4 The gap (verified in code)
+
+- **No sections, no chrome.** `BakedRelatedSectionKeys` is derived from projected `<mj-collapsible-panel>` children. A React whole-form projects none.
+- **Duplicated related data.** With nothing baked, `<mj-form-contributions>` appends a stock grid for every `DisplayInForm` relationship, while the prompt tells Skip to draw related tables in view mode.
+- **Inert claims.** A compiled `replacesSectionKey: 'details'` hero hides nothing on a React form; compiled panels fall to `after-everything`.
+- **Frozen composition.** Hand-drawn related tables are a schema snapshot; a downstream app installed later is invisible to them.
+- **All-or-nothing.** "Add a KPI strip" means regenerating the whole form and losing every compiled contribution's position.
+
+## 3. Goals and non-goals
+
+**Goals**
+
+1. Skip produces a single contribution (hero, extra pane, related-grid takeover, field-panel replacement) that mounts through the existing slot, composer, and chrome machinery.
+2. Metadata and compiled contributions are peers: same `contributionKey` collapse, same L1 verbs, same L3 suppression, same L4 arrangement, same `HiddenSectionKeys` propagation.
+3. Whole-form replacement is unchanged in behavior and becomes composable (slots).
+4. One row format for every author: Skip, the Form Builder agent, Form Studio, `mj sync`.
+5. Skip sees the form it is extending before it designs the change.
+
+**Non-goals**
+
+- Changing CodeGen output.
+- Replacing compiled `BaseFormPanel`. Panels that need Angular services or typed entity subclasses keep using it.
+- Toolbar verbs from React contributions (host hooks exist; deferred until a real request).
+- Per-application scope (same punt as `EntityFormOverride`; a nullable `ApplicationID` column can be added later).
+
+## 4. Architecture
+
+```
+  Registrations                                             Renderers
+  ─────────────                                             ─────────
+  ClassFactory (compiled BaseFormPanel)  ── Source 'class' ──┐        ┌── registered Angular ctor
+                                                             ├─▶ Collapse ─▶ Resolve ─▶ <mj-form-panel-slot> ─┤
+  MJ: Entity Form Contributions rows     ── Source 'metadata'┘   (unchanged pure logic)                        └── InteractiveFormPanelComponent
+  (InteractiveFormsEngine cache)                                                                                    └── <mj-react-component>
+                                                                                                                        componentRole 'form-panel'
+  <mj-record-form-container> reads the SAME merged list for chrome (inclusion / sortKey / chromeGroup) and hidden keys.
+```
+
+The only new runtime pieces are the merged collector, one generic host panel, and the engine load. The composer, slot coordinator, chrome resolver, and L3 rules are untouched.
+
+## 5. Data model
+
+### 5.1 `MJ: Entity Form Contributions` (new table `EntityFormContribution`)
+
+Mirrors `FormPanelRegistrationMetadata` one-to-one; borrows scoping and lifecycle from `EntityFormOverride`.
+
+| Column | Type | Notes |
+|---|---|---|
+| `ID` | uniqueidentifier | PK |
+| `EntityID` | FK `Entity` | Parent form |
+| `ComponentID` | FK `Component` | `Type='Widget'` row whose spec declares `componentRole: 'form-panel'` |
+| `Name`, `Description` | nvarchar | `Name` is the version lineage label |
+| `Slot` | nvarchar(30) | `top-area` · `before-fields` · `after-fields` · `after-related` · `after-everything` |
+| `SortKey` | int, default 0 | Higher renders first within the slot |
+| `ContributionKey` | nvarchar(256) NULL | Last-wins identity; NULL derives `related:<entity>:<join>` or stays unique |
+| `RelatedEntityID` | FK `Entity` NULL | Related-grid claim |
+| `RelatedJoinField` | nvarchar(255) NULL | Bill-To vs Ship-To |
+| `ReplacesSectionKey` | nvarchar(255) NULL | Hides a baked field panel (hero pattern) |
+| `Inclusion` | nvarchar(10) NULL | `Primary` / `More` / `None` |
+| `ChromeGroup` | nvarchar(10) NULL | `details` / `more` |
+| `Presentation` | nvarchar(10), default `panel` | `panel` wraps in `<mj-collapsible-panel>`; `bare` renders a hero strip with no chrome and no rail item |
+| `Title`, `Icon` | nvarchar NULL | Section header / rail label and icon |
+| `Scope`, `UserID`, `RoleID` | as `EntityFormOverride` | Same CHECK shape |
+| `Priority` | int, default 0 | Last-wins against compiled registrations (see §8) |
+| `Status` | nvarchar(20) | `Active` / `Pending` / `Inactive` |
+| `Configuration` | nvarchar(MAX) NULL | JSON passed to the component as `contribution.configuration`; one component, many rows |
+| `Notes` | nvarchar(MAX) NULL | Agent iteration log |
+
+Filtered unique index on `(EntityID, ContributionKey, Scope, UserID, RoleID) WHERE ContributionKey IS NOT NULL`.
+
+### 5.2 Component rows
+
+Panel components are `MJ: Components` rows with `Type='Widget'` (decision 2). `InteractiveFormsEngine` widens its filter to `Type IN ('Form','Widget')`. Whole forms stay `Type='Form'`.
+
+### 5.3 Rejected alternatives
+
+- A `Kind` column on `EntityFormOverride`: wrong cardinality, mostly-NULL columns, wrong "first active wins" semantics.
+- Rows inside `Entity.Configuration`: no per-user scope, no lifecycle, and the admin-versus-app-sync collision L3 exists to solve.
+- Reusing `EntityRelationship.DisplayComponentID`: a CodeGen-time selector for Angular generators; cannot express heroes or extra panes.
+
+## 6. Runtime
+
+| Piece | Change |
+|---|---|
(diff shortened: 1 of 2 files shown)


rkihm-BC:
# Review: Skip form contributions (design + implementation plan)

Read both documents in full and verified the structural claims against `next`. Summary: **the diagnosis is accurate, the architecture is sound and genuinely minimal, and the migration authoring is textbook.** There are three items I'd resolve before Phase A opens, one of which is a design question about an engine's load footprint.

This is a planning PR, so everything below is about the plan, not code.

---

## What's good

**Every claimed defect in §2.4 and §14 verifies.** I checked all four independently:

- The `formContext` getter at `base-form-component.ts:952` calls `CollectFormPanelRegistrations()` — a full `ClassFactory` scan plus a full `ResolveFormContributions` — on every change-detection pass, via line 985. Real, and worth fixing on its own.
- `BaseFormPanel.validate()` has **zero callers repo-wide**, despite a docblock promising `Save()` surfaces it. Documented-but-unimplemented, exactly as stated.
- The slot host fuzzy-matches entity names (`form-panel-slot.component.ts:237-242` lowercases and strips the `MJ:` prefix) while both `form-contribution.ts` and `record-form-container.component.ts:1076` use strict `===`. Live inconsistency.
- `contributionKey === 'header'` is hardcoded at `record-form-container.component.ts:1077`.

**The architecture earns its keep.** Adding a second *source* and a second *renderer* while leaving `ResolveFormContributions` — the pure composer — completely untouched is the right shape. Metadata rows becoming peers means L1 inclusion, L3 `MJ: Form Chrome Rules` suppression, and L4 arrangement all work for free. Large capability, small surface.

**The tie-break is better than the design describes.** Decision 1 ("compiled wins on a tie") could have been implemented as array ordering; Task A4 instead adds an explicit `sourceRank` comparator, making it order-independent, and the tests assert both input orderings.

**The migration follows every rule in CLAUDE.md.** Folder derived from the filename version, ≥50 blank lines before the CodeGen block, apply-time `MAX(Sequence)+1`, `mj sync push` before `mj codegen`, an explicit one-database-per-agent check, no `__mj_*` columns, no PG counterpart (correctly flagged as release-engineer work), and it explicitly refuses to stage `packages/GeneratedEntities/**` and `mj.config.cjs`. Extended properties on every column. I'd merge Task A2 as written.

**No registration-discovery site is missed.** I enumerated every consumer. The three container `CollectFormPanelRegistrations()` call sites (lines 1073 / 1099 / 1158) map exactly to the three methods Task A9 names — `contributionRegistrations`, `contributionHiddenSectionKeys`, `hiddenChromeSectionKeys`. `form-contributions.component.ts:60` and `base-form-component.ts:985` are both in the file list. The fourth `GetAllRegistrationsByMetadata` call, at `chrome/resolve-form-chrome.ts:93`, scans `BaseFormPolicy` — a different registration type, correctly out of scope.

**Every API the plan consumes actually exists.** Verified `SetSectionRowCount`, `IsSectionExpanded`, `OnFormNavigate`, `NewRecordValues(name, joinField)`, `ProviderToUse`, `MJAlertComponent` (already in the module's `imports`; `'error'` is a valid `MJAlertVariant`), the `/forms` subpath export in `InteractiveComponents/package.json`, base-forms' existing dependency on `@memberjunction/ng-react`, and that `ComponentSpec` genuinely lacks `entityName` so A1's addition is real work. Also confirmed `FormResolverService` does `await InteractiveFormsEngine.Instance.Config(false, …)` before form creation — which is what makes the §15 readiness argument hold on the Explorer path.

Decisions §12 makes five explicit calls with rationale, and §5.3 correctly kills the `Kind`-column and `Entity.Configuration` alternatives for the right reasons. Correctly ships no changeset for a docs-only PR.

---

## What's concerning

### 1. Widening the engine filter to `Type IN ('Form','Widget')` needs a measurement first

Task A3 changes `interactive-forms.ts:98` from `Type='Form'` to `Type IN ('Form','Widget')`, with `CacheLocal: true`.

The engine's own docblock, twenty lines above, explains why it stayed narrow: `ComponentMetadataEngine` "**deliberately omits** `MJ: Components` themselves because the full Component table includes ~150MB of `Specification` JSON across non-form types," and this engine "threads the needle: load `Type='Form'` Components only (small dataset — a few dozen per typical deployment, ~5MB max)."

To be precise about the risk rather than alarmist: I could not establish that `Widget` is a large population today. In seeded `metadata/components/` it is 4 rows, behind `Other` (44), `Chart` (32), `Table` (17), and `Dashboard` (15), and the only runtime writer I found is the external-component registry cache at `React/runtime/src/registry/component-registry-service.ts:641-657`, which maps `spec.type` through a `typeMap`.

The concern is the *shape*, not a number: `Form` was chosen as a **closed, enumerable** set. `Widget` is an **open** set populated by registry sync and general authoring, growing on a curve unrelated to form-panel adoption, and it will be cached to client local storage on every Explorer boot for every user. The plan neither bounds nor measures it, and §14 defers "lazy `Specification` loading in the engine" to a later PR — so the widening lands *before* its mitigation.

Two cleaner options, neither considered in decision 2:

- **Filter to referenced components** — `WHERE ID IN (SELECT ComponentID FROM EntityFormContribution)`. The loaded set becomes exactly the contributions, scaling with adoption instead of with unrelated authoring. This is the option I'd pick.
- **Add `FormPanel` to the `Component.Type` value list.** One new value, the filter stays narrow, and `Type` keeps meaning something — it also removes the awkwardness of a `Widget` row that is not a widget.

Either way, worth a load-size measurement on a representative database before committing.

### 2. Strict entity matching breaks eleven shipped panels; the diagnostic that covers it has a hole

Credit where due: Task A7 **does** anticipate this, adding `warnIfLooseOnly` which logs *"Entity names must match exactly; these panels will not mount."* Two problems remain.

**First, the eleven panels aren't repaired.** These register unprefixed names and mount today only because of the fuzzy matcher:

| Registered `entity` | Actual entity | Panels |
|---|---|---|
| `'AI Agents'` | `MJ: AI Agents` | `panels/ai-agents/agent-realtime.panel.ts:54` |
| `'AI Agent Categories'` | `MJ: AI Agent Categories` | 2 |
| `'Users'` | `MJ: Users` | 2 |
| `'Conversations'` | `MJ: Conversations` | 2 |
| `'Companies'` | `MJ: Companies` | 2 |
| `'Employees'` | `MJ: Employees` | 2 |

A console warning in a production Explorer build is a diagnostic, not a remediation. These registrations should be corrected in the same task.

**Second, the warning misses the case most likely to occur.** It only fires when `strict.length === 0` for that slot. On a form where *any* panel matches strictly, the loosely-named ones break with no warning at all — which is exactly the `MJ: AI Agents` situation, where the loose `agent-realtime.panel.ts` shares a form with strictly-registered panels. The entity most likely to hit this is the one the diagnostic can't see.

Worth noting the asymmetry cuts the other way too: because the container is already strict, these eleven panels' `inclusion` / `sortKey` / `chromeGroup` are silently ignored by the chrome rail **today**. That's a live bug independent of this feature, and it deserves to be named as its own fix rather than absorbed into a peer-composition PR.

### 3. Two `Priority` columns in one feature family with opposite meaning

`InteractiveFormsEngine.GetActiveOverrideForEntity` sorts **ascending** (`interactive-forms.ts:197`), documented as "Priority is sort key, not boost." The new `EntityFormContribution.Priority` is **descending** (highest wins), matching `ClassFactory`.

So `MJ: Entity Form Overrides` and `MJ: Entity Form Contributions` will sit side by side in Form Studio, share an action family, share the column name `Priority`, and mean the reverse of each other. §14 defers the reconciliation to a separate PR. Since the apply flow writes `Priority = incumbent + 1` automatically, an admin reasoning by analogy from overrides gets exactly the opposite of what they intend.

Either unify the semantics in Phase A, or name the new column something that can't be confused — `Rank` or `Precedence`.

### 4. The `'header'` convention is mischaracterized, and six panels are already dead

§14 says the fix replaces "the `contributionKey === 'header'` magic string" with `Presentation`. The actual convention in the codebase is `slot: 'header'` — see `custom/Users/user-header.panel.ts:13`. And `'header'` is **not in the `FormPanelSlot` union**, so `FormSlotCoordinator.resolveSlot('header')` hits `indexOf → -1` and returns `null` (`form-slot-coordinator.service.ts:73`). Zero `Slot="header"` hosts exist anywhere in the repo.

**All six `*-header.panel.ts` panels never render.** The container check the plan targets never fires for them, because they set no `contributionKey` at all. If `Presentation: 'bare'` is meant to supersede the hero pattern, migrating those six is real work the plan doesn't currently see.

Related: the plan offers Skip `top-area` as a slot. The repo emits `Slot="before-fields"` 388 times and `Slot="top-area"` **zero** times. Fallback rescues it — but CodeGen *does* emit a `top-area` **section** (`<mj-form-section Section="top-area">`, `angular-codegen.ts:561`), so `top-area` exists as a section concept and not as a panel slot. That collision makes the trap more likely, not less, for a model reading the enum.

### 5. `Validate` has a sync/async mismatch that fails silently

`BaseFormPanel.validate()` returns a synchronous `ValidationResult`, so Task A6 calls `invokeMethod(FormPanelMethodNames.Validate)` and tests `'isValid' in live`. `MJReactComponent.invokeMethod` returns `any` — whatever the React method returns.

The panel contract prompt (Task C1, hard rule 8) says: *"`'Validate'` (return `{ isValid, errors: string[] }` if you gate the parent save)."* It never says **synchronous**. A generated validator that checks anything server-side will be `async`. The returned Promise fails the `'isValid' in live` test, `validate()` falls back to the stale `lastValidation`, and the form saves as if valid — no error, no log. None of the three new gates (`NoSave`, `UsesHostProps`, `NoFixedHeight`) catches it.

Fix: state the constraint in the contract, add a fourth gate, or make the host `await` and change the signature.

### 6. The memo cache isn't keyed on the provider

Task A5's cache key is `${entity.Name}::${userID}::${classCount}::${engineVersion}` in a module-global `Map`. MJ supports per-provider `Metadata` scoping — a documented rule in `.claude/rules/data-access.md` — and `CollectFormContributionRegistrations` takes `provider` as its second argument but never puts it in the key. Two providers, same entity name, same user, different data → collision.

Separately: the engine-read `try/catch` correctly swallows `PermissionConstrainedError`, so a restricted role won't crash change detection. But the idiomatic guard elsewhere is an `IsPermissionConstrained` check, as `form-resolver.service.ts:200` does. As written, an expected condition logs as an error on deployments where the new entity isn't granted.

### 7. A visible layout shift on every cold form open

§15 accepts that "a cold path renders class contributions first and rows a tick later." For an ordinary extra pane that's a flicker. For a `bare` hero with `replacesSectionKey`, the user watches the baked Details panel render and then **disappear** — and that's the flagship scenario in §11.

Explorer is covered because `FormResolverService` awaits the engine. Dialogs, slide-ins, and the "dashboard quick-edit" reuse that the `BaseFormPanel` docblock advertises may not go through the resolver. Worth either gating first paint on engine readiness or reserving space.

### 8. Uniqueness is enforced in two places, and neither covers keyless rows

The filtered index covers `(EntityID, ContributionKey, Scope, UserID, RoleID) WHERE ContributionKey IS NOT NULL AND Status='Active'`. But `ResolveContributionKey` *derives* `related:<entity>:<join>` when `ContributionKey` is NULL — and the index excludes NULL.

The action layer doesn't close the gap either. Task B1 sets `row.ContributionKey = contribution.contributionKey ?? null` with **no derivation**, and its duplicate-existence check is guarded by `if (contribution.contributionKey)`. So for a keyless related-grid claim there is *no* index coverage **and** *no* duplicate check. Two Active rows claiming the same grid are reachable through the supported apply path, not just through `mj sync` — and the winner is then decided by DB row order, since `GetApplicableContributions` sorts only on `Priority` then `SortKey`.

Fix: derive and persist the key on write, or extend the index to cover `(RelatedEntityID, RelatedJoinField)`.

Also unconstrained in the DDL: `Presentation='bare'` together with `Inclusion`/`ChromeGroup` (contradictory — bare is never a rail item), `RelatedJoinField` without `RelatedEntityID`, and both `ReplacesSectionKey` and `RelatedEntityID` set.

Two smaller notes on the same tasks:
- §5.1 states the index filter as `WHERE ContributionKey IS NOT NULL`; A2 adds `AND Status='Active'`. The implementation is the better choice — it permits version history — but the divergence is undocumented.
- B1's dup check hand-escapes a model-authored string into a `RunView` `ExtraFilter` (`contribution.contributionKey.replace(/'/g, "''")`). The escaping is correct for T-SQL, but worth a look given the value originates from an LLM.

---

## What's missing

**No integration coverage — and none exists for forms today.** The plan says "run `pnpm run test:integration`" and authors no bundle. I checked `packages/TestingFramework/integration-test-suite/src/checks/` — 80+ bundles, and **zero** hits for `BaseFormPanel`, `FormPanel`, `EntityFormOverride`, or `InteractiveForm`.

That reframes this from "the plan skips an established practice" to something sharper: **nothing in the deterministic tier would catch any of the findings above** — not the eleven unmounted panels, not a composition regression, not the priority inversion. This feature is *nothing but* cross-package seams (engine load → merged collector → slot host → action family → cross-repo transport), which is precisely what CLAUDE.md means by *"Unit tests passing is necessary but not sufficient."* `ai-skills.checks.ts` (IT12) and `entity-actions.checks.ts` are the right precedent for a new action family, and this is the natural occasion to establish forms coverage.

**No security section, in 5,549 lines.** Decision 3 — OpenApps ship `Scope='Global'` rows via `mj sync` with no promotion gate — gets one line, justified as "installing the package is the gate, as for compiled panels." That analogy is weaker than it looks: a compiled panel is reviewed code in a package build; a metadata row is JSON that pushes a runtime-interpreted React spec onto *any* entity's form for *all* users, including `MJ: Users`. The action layer has a real clamp (agents write User scope only, `_shared.ts:288+`); the sync path bypasses actions entirely and has none. The stated mitigation — L3 rules can suppress per site — is reactive, requiring someone to notice first. Worth arguing explicitly rather than asserting.

**No accessibility.** Zero mentions across both documents. Panels become rail items with badges, heroes render bare and outside the rail entirely, and the contract's ten hard rules cover saving, heights, headers, and row counts — nothing on semantics, focus order, or keyboard reachability.

**No theming rule in the panel contract** — the visual rule most likely to be broken. Forms are themed; the contract passes `styles`; the canonical JSX example destructures `styles` and then never uses it, styling inline instead. A panel with a hardcoded background inside themed chrome is a dark-mode defect waiting to happen. The repo enforces this for SCSS via `npm run check:ui`; React specs have no equivalent, and a hard rule plus a gate would be cheap.

**No rollback or kill switch.** This changes the render path of every entity form in Explorer, with no flag to disable metadata contributions if the engine widening or the collector misbehaves.

**No React error-boundary story.** `loadError` covers "spec missing / bad JSON / wrong role." A panel that throws *during render* isn't discussed — whether `<mj-react-component>` contains it, or whether it takes the form down.

**Wildcard asymmetry, unstated.** Compiled panels can register `entity: '*'`; `EntityFormContribution.EntityID` is `NOT NULL`. The "peers" framing in §3 goal 2 has an asterisk. Probably deliberate and probably fine — but it belongs in non-goals next to the `ApplicationID` punt.

**The transport diagram overstates what exists.** §10 annotates `AppContextSnapshot.AdditionalContext.Form` as "(already published)." The record tab publishes **no** agent context today — neither `record-resource.component.ts` nor `single-record.component.ts` has a `SetAgentContext` call, which is why Task D1 exists. Separately, `explorer-app.component.ts:641` does `AdditionalContext: update.AgentContext` — a **wholesale replacement**, not a merge, on an app-global snapshot. Combined with decision 5 (full snapshot on every message), a stale snapshot from a backgrounded tab is more likely than §15's "first render" caveat suggests.

---

## Two notes on the plan as an artifact

The implementation doc carries near-complete source for most tasks. That makes intent unusually reviewable — I could check the tie-break comparator and the `Record.Set` guard directly, which is why this review can be specific. But it will drift the moment Phase A starts, and a reviewer of the eventual code PR may be tempted to diff against it. Worth a header stating the code blocks are illustrative, not normative.

Second: A2's `Presentation`, `Inclusion`, and `ChromeGroup` are `NVARCHAR` with CHECK constraints and no `MJ: Entity Field Values` rows. Since these become user-facing pick-lists in Form Studio (Phase E), value-list metadata rather than bare CHECKs would give the UI a dropdown for free.

---

## Recommendation

**Approve the direction; resolve #1–#3 before Phase A opens.** The core bet — a database row as a first-class peer of a compiled registration, sharing one composer — is right, and the audit behind it found four real defects that all verify against `next`.

The blockers are local and tractable: pick a narrower component filter (with a measurement), repair the eleven registrations the strict-matching change breaks and close the diagnostic's hole, and settle `Priority` semantics before the column ships. #4–#8 are Phase A scope adjustments. The integration bundle and the security argument should land before the feature is enabled for users, alongside Phase D.

PR #4313, opened 2026-09-08
ci(codeql): exclude plans/ (design documents and mockups) from all three analyses

rkihm-BC:
## Summary

Adds `plans/**` to the CodeQL path exclusions in all three analysis configs. This is a scan-coverage decision, kept as its own PR so it can be reviewed and recorded as one.

## Why

`plans/` holds design documents and HTML mockups. Nothing under it is built, imported or shipped, which puts it in the same class as the generated output the configs already exclude: code nobody ships. The first full js-rest scan reported 4 high alerts in `plans/composer-adoption/mockups/omnibar-command-palette.html` (2 `js/xss-through-dom`, 2 `js/incomplete-multi-character-sanitization`). Fixing a mockup is busywork, and leaving it counts against the baseline indefinitely.

## What it removes from view, stated so nobody mistakes the folder for scanned

Every non-markdown file under `plans/` (markdown was already excluded by `**/*.md`). If a mockup under `plans/` is ever promoted into a package, it becomes scanned by virtue of the move, which is the right moment.

## Policy record

The config comments require path exclusions of this kind to be recorded in BC-POL-012 Annex A and exception EX-2026-002 alongside the generated-code exclusion. Those documents are not in this repository; the comment added next to the pattern names the date, the reason and both records so the policy owner can make the entry from the PR. Please do not merge before that entry exists, or note in the PR that it has been made.

## Review follow-ups (second commit)

- The rationale now lives once in `codeql-config.yml`; the other two files carry the pattern with a one-line pointer, as their own headers require.
- The mockup-size claim was wrong (`plans/lists/mockups` is the largest set at 28 HTML files, not composer-adoption) and is replaced with the actual counts.
- The policy record is stated as an obligation, not as already done.
- `codeql-config-core.yml` notes that its `paths:` allowlist already makes the pattern a no-op there; it stays only for baseline parity.
- `plans/` is listed in both NOT COVERED blocks so the gap is visible from the file itself, and the verification header no longer names a script that does not exist.
- Not excluded on purpose, and said so in the file: `experiments/` (buildable prototype code with real alerts) and the `plans/` directories nested inside packages.
- **Validated before merge:** the workflow was dispatched on this branch so js-rest ran with the new config. See the run linked in the PR conversation; the branch analysis reports zero alerts under `plans/`.

## Change

The same block in the same position in `codeql-config.yml`, `codeql-config-core.yml` and `codeql-config-rest.yml`, keeping the shared baseline section identical across the three, as the files require. All three parse.


rkihm-BC:
Pre-merge validation: dispatched the CodeQL workflow on this branch so js-rest ran with the new config — https://github.com/MemberJunction/MJ/actions/runs/34293026662

Issue #390, opened 2024-06-27
Support for soft deleting records


cadam11:
TBD @AN-BC 

Maybe enabling soft deletes on an Entity is a pre-requisite for merging records / dedup.

AN-BC:
- Metadata additions:
- Entities: DeleteType: Hard or Soft - when set to Hard, will work as it currently does. When set to Soft - the following will occur:
1. `CodeGen` will auto add `__mj_DeleteAt` column to the entity in question
2. The generated base view will filter **out** any record that has any value in the __mj_DeletedAt column, ensuring the records appear deleted
3. `CodeGen` will generate the `spDelete` procedure to update a record with a GETUTCTIME() instead of the hard delete which actually deletes the rows from the table.
4. CodeGen will ensure that the `__mj_DeletedAt` column is never updated by an spUpdate/spCreate proc given its special meaning
- Entities; AllowRecordMerge will be added as a bit field and future Merge record support will require this field to be turned on. Additionally, merge support will require that DeleteType='Soft'

AN-BC:
This was completed already

Issue #303, opened 2024-06-09
Update MJServer's `graphql-request`


cadam11:
In #258 we updated most dependencies, but the current version of graphql-request has enough breaking changes to warrant its own issue.

PR #4314, opened 2026-09-08
fix(tooling): close the CodeQL findings in CI, docs, test and build scripts

rkihm-BC:
## Summary

Closes nine CodeQL alerts in tooling: CI scripts, the docs-site ingest, the unit-test runner, the MJAPI manifest bootstrap and the agent-memory smoke harness. None of it is shipped code, but the three CI scripts run in Actions with repository tokens, which is why they are fixed rather than accepted.

Companion PRs: #4312 deletes the dead prototype; #4313 excludes `plans/` from scanning. Six further tooling alerts that only run on developer machines (the e2e overlay auth primer, #303–#306, and the Docker regression preflight, #388, #420) were dismissed as won't-fix with the reason recorded on each alert.

## Changes

| File | Alert | Fix |
|---|---|---|
| `.github/scripts/check-new-npm-packages.mjs` | #321 #322 #323 incomplete-sanitization | Encode every `/` in a scoped name, not only the first |
| `.github/scripts/plan-test-shards.mjs` | #390 file-system-race | Read the weights file inside a `try`, ENOENT means empty |
| `.github/scripts/check-esm-imports.mjs` | #318 bad-code-sanitization | Entry URL and marker passed to the child as arguments; the child script is a constant |
| `docs-site/scripts/ingest.mjs` | #324 incomplete-sanitization | Escape backslashes before pipes in table cells |
| `unit-testing/tools/run-all-tests.js` | #371 shell-command-injection-from-environment | `execFileSync` with an argument array |
| `packages/MJAPI/scripts/ensure-manifest.mjs` | #396 file-system-race | Write with flag `wx`; EEXIST is the "already present" exit |
| `packages/AI/Agents/scripts/memory-write-smoke.ts`, `contextcrush-smoke.ts`, `packages/Actions/Engine/scripts/entity-action-workflow-integration.ts` | #131 #138 #132 incomplete-url-substring-sanitization | `endsWith` rather than `includes` for the Azure host check, applied to all three harness scripts |

## Review follow-ups (second commit)

- **Windows regression fixed.** `execFileSync('npm.cmd')` without a shell throws EINVAL on every Node the repo supports (the CVE-2024-27980 hardening), and the runner's catch would have reported every package as failed with an empty aggregate. Windows now uses a shell with the one path argument quoted for cmd.exe; POSIX keeps the plain argument array.
- **The same pattern, same file.** `loadWeights` still had the existsSync-then-read sequence the PR removed from `recordWeights`; both now use `readJsonIfPresent`, and malformed JSON still degrades to uniform weights (the existing test for that stays green).
- **Structural, not advisory.** The ESM guard's child script is a module-level constant, so nothing in function scope can be spliced into it.
- **One encoder** for the three registry URL sites in the npm-package check.
- **Manifest bootstrap**: template hoisted, single write path, and the dangling-symlink case (an exclusive create reports EEXIST for it too) is written through as before. Fresh, existing and dangling-link paths verified; the emitted file is byte-identical.
- **Table-cell escaping** is a single character-class replace, so there is no ordering to get wrong.
- **Changeset** added for `@memberjunction/ai-agents` per the repo's release-notes rule.
- The remaining `includes('.database.windows.net')` sites elsewhere in the repo (MJCLI, MetadataSync, OpenApp) carry their own alerts in the js-rest set and belong to the URL-substring batch; the two sibling harness scripts with the same line are fixed here (third commit) now that #4312 no longer deletes them.

## Verification

- `.github/scripts` vitest suite: 20 files, 523 tests passed. It covers the npm-package check, the ESM guard and the shard planner directly.
- ESM guard child helper exercised end to end after the change: a real entry resolves to `null`, a missing one to the structured `ERR_MODULE_NOT_FOUND` result the caller expects.
- Manifest bootstrap exercised both ways: an existing file is left byte-for-byte untouched (a sentinel survived a second run), and a fresh checkout gets the file.
- `node --check` on every modified file.
- docs-site `node --test` suite passes (that package is outside the pnpm workspace and needs its own `npm ci` first).

No changeset: none of these files is part of a published package's shipped output (`files: ["/dist"]`).


rkihm-BC:
The Integration Tier failure earlier on this branch was unrelated to these changes: `mj sync push` failed at the transaction commit after all 84 metadata steps had succeeded (`Can't commit transaction. There is a request in progress.`). Re-running the identical job on the same commit passed. The diff from the previous passing merge touches nothing in metadata sync or the provider, and every other commit on this branch passed the same job. Tracked as a real defect in #4320.

PR #4303, opened 2026-09-08
5.51.x backport: a11y shell landmarks + skip link, focus containment, nameable primitives (#4121)

MattC-BC:
Backport of #4121 (merged to `next` 2026-09-01, squash `42733175a`) to `lts/5`. Manual backport rather than the bot: the root `package.json` hunk conflicts (npm vs pnpm form) and one template line assumes a `next`-only input.

Line rules: patch-only, no migrations, no metadata. Changeset: `patch` on `@memberjunction/ng-explorer-core`, `@memberjunction/ng-shared-generic`, `@memberjunction/ng-whiteboard`.

## What is different from #4121

- `package.json`: the `check:ui` script and the new `check:focus-ring` script in npm form (the line is npm-era). No `lts/5` workflow invokes `check:ui`; this is a local gate.
- `shell.component.html`: dropped `AriaLabel="Search in"` on `<mj-dropdown>`. That input is #3860, which is on `next` only. On this line the attribute would be inert. The search dropdown stays unnamed on `lts/5` until #3860 is backported (separate PR; not absorbed here).
- `.changeset`: the #3860 sentence rewritten to say the above.
- `.github/workflows/ci-ui-tokens.yml`: NOT ported. The `check-focus-ring.sh` script is in the tree and passes against the whole line (`--all`). **Reviewer question: do you want the six-line CI step from #4121 on this line?** Default here is no, per line CI discipline.

## Not applicable on LTS

- `mj-dropdown` naming (Sidecar a11y brief item 3): needs #3860 backported.
- Whiteboard keyboard operability (brief item 4): not part of #4121 on any line.

## Behaviour change, stated once

Whiteboard single-letter shortcuts and its undo/redo/Escape/Delete no longer fire while focus is elsewhere on the page. `EnableGlobalShortcuts` restores the old behaviour. Same as #4121.

## Verification on the line

- `npm ci`; `turbo run build` of the three packages and their dependency chain: green, 105/105 tasks, 57 s
- `npm test` in `ng-explorer-core` (dual vitest projects; the ported palette dom test runs under the jsdom project), `ng-whiteboard`, `ng-shared-generic`: green. explorer-core 120/120 across 14 files (palette dom test included), whiteboard 313/313 across 16, shared 26/26 across 5
- `./.github/scripts/check-focus-ring.sh --all`: green
- Local Explorer run against the line for the eight cleared items: pending (Matt, per finding)

[...]

PR #4319, opened 2026-09-09
Whiteboard roster follow-ups: Restyle follows the text tool, empty roster lands on Select, host ReadOnly

MattC-BC:
Follow-ups to #4310 (`ToolRoster`), from Robert's review on the `lts/5` backport #4318. Fixed here on `next` first, then backported by label — `lts/5` carries the #4310 code byte-for-byte, so this applies cleanly.

## What changed

**Restyle… now follows the `text` tool.** It is the one item-menu action that is also a door to a tool: it opens the toolbar's text style flyout, which only exists because the text tool's button renders it. Under a roster without `text` the entry was present and did nothing. The rest of the item menu stays deliberately roster-blind — Edit, Duplicate, z-order and Delete are authoring on what already exists, not tool selection.

**An empty or all-typo roster clamps the active tool to `select`.** It previously left the current tool alone, reasoning that the board must always hold something. It must — but holding the tool the roster just revoked was the one answer that kept a *creating* tool live with no toolbar to see it and no key to change it: `Tool = 'html'` then `ToolRoster = []` went on placing widgets. An empty roster still does not make the board read-only.

**`RealtimeWhiteboardHostComponent` gains `@Input() ReadOnly`** (default `false`). The board component has had it all along; the host had no passthrough, so #4310's own advice to "use `ReadOnly` for that axis" was an `NG8002` for anyone who followed it. The host binds it to the board and adds the chrome only it owns: no floating toolbar (matching `WhiteboardSnapshotComponent`), no Undo on the agent toast, and a keydown handler that returns before any key can act — `Escape` included, because "the keyboard does nothing here" is a rule a user can hold and "does nothing except Escape" is one they have to be told. Pan and zoom stay live. `Tool` is not reset; the board ignores edits regardless.

The two axes are not substitutes: the roster answers *which tools are offered*, `ReadOnly` answers *whether anything mutates*.

**The `ToolRoster` setter tolerates a non-array and re-clamps on content change.** Two ways past it, both closed:

- **A non-array.** A static attribute — `ToolRoster="select,pan"`, a plausible slip for the binding — sets a *string*, which reached the roster helpers as an accidental substring matcher: `String.prototype.includes` answered "allowed" for any tool name appearing anywhere in the text, rendering a garbage palette with no error. A string that does *not* contain the held tool's name instead reaches the clamp and throws outright — verified against the old setter: `TypeError: (roster ?? []).filter is not a function`. Both now read as no roster.
- **Identity comparison**, which was wrong in both directions. A bound array literal is a new reference every change-detection pass, so identity re-clamped forever and fought the user for the active tool; an array mutated in place keeps its reference, so identity never re-clamped and a roster that just dropped the held tool went unenforced. The host now keeps a frozen copy and compares by content (new `SameRoster` helper).

Toolbar: `VisibleTools` moves from a getter to a field computed in the `ToolRoster` setter. The getter re-filtered eleven entries and allocated a new array every change-detection tick, which also defeated `@for` reuse.

## Two things to actually stop on

1. **The Restyle change is not a no-op.** Anyone already shipping a roster that omits `text` loses the Restyle entry on sticky and text items. That entry did nothing for them, so this is the right change — but it is the one user-visible behavior change in the PR, and it is called out in the changeset rather than buried.
2. **Image paste under a roster omitting `image` stays permitted, deliberately** (your D4 question). The roster is tool availability, not content policy. If the LXP needs "learners cannot place images", that wants a content-policy input (`PlaceableKinds`) or `ReadOnly` — not the roster. Flagging so it reads as a decision rather than an oversight.

## Deliberately deferred

The four hand-kept tool tables (`WHITEBOARD_TOOLS`, the toolbar's `Tools`, `toolForKey`, the canvas add-action map) want a catalog module. That is a refactor, not a behavior bug, so it gets its own PR rather than riding along here.

## Testing

`pnpm test` → **361 pass**; `pnpm test:types` clean; `ngc` builds.

Every verification case has a DOM spec rendering the real host, board and toolbar together — not mocks:

- `w` then `ToolRoster=[]` lands on Select **and** a canvas click places nothing
- `ReadOnly` renders no toolbar (0 buttons vs 13), ignores `p` and Delete, opens no canvas menu, and places nothing on a canvas click *while holding `sticky`* — each with its positive control, so a listener that died for an unrelated reason cannot pass as "correctly gated"
- `ToolRoster="select,pan"` as a static attribute renders the full palette
- Restyle present with `text` on the roster, absent without, with the rest of the item menu asserted unchanged

The vacuous "does not re-clamp when the same value is set again" test is replaced by three observable ones: equal content in a new array is a no-op; an in-place `splice` plus a re-bind of the *same reference* re-clamps (asserting the host's copy was untouched); and the non-array case above. The direct-`Tool`-write test gains the reverse assignment order.

Both halves of the non-array spec were checked **against the old setter** rather than assumed — they fail there, with the substring palette and the `TypeError` respectively. That check is what caught an incorrect claim in my first pass that a string always throws; it does not, it silently substring-matches, and the spec and comments now say what actually happens.

No browser pass. The layout question it would answer I checked directly instead: `.board-wrap` is `position: relative` and the toolbar's `:host` is `position: absolute`, so removing the toolbar cannot reflow the board, the zoom cluster or the toast. Worth someone laying eyes on `[ReadOnly]="true"` on a real board during review, though — a toolbar-less host is a composition nobody has looked at outside the snapshot component.

## After merge

Apply the `backport lts/5` label (`backport.yml` opens the backport PR from it). Note for that review: the `lts/5` DOM spec header still describes the #4121 focus scope until #4303 lands — nothing to do there.


rkihm-BC (dismissed):
This is the fix PR for the findings I raised on the `lts/5` backport #4318, and it lands them properly — including the one I got wrong. Approving; one should-fix below, and one thing I changed on the branch myself.

## Verified rather than taken on trust

- **Restyle gating is complete.** `restyle` is reachable only through the context menu (`whiteboard-board.component.ts:1380` handles it, nothing else emits it), so gating `RESTYLABLE_KINDS` on `IsToolAllowed(roster, 'text')` closes the whole path. The test asserting `under([])` equals the full ID list *minus* `restyle` is the right shape — it pins that nothing else moved, which is the half a weaker test would drop.
- **`ReadOnly` is a real passthrough, not a nominal one.** The board already guards ten mutation entry points behind its own `ReadOnly`; the host adds exactly the chrome it owns. Removing the toolbar cannot orphan anything — the host's only `@ViewChild` is on the board, and nothing outside the new `@if` references the toolbar.
- **The clamp change is right and the reasoning is the right way round.** "The board must always hold a tool" was true and still led to the wrong answer; `select` is the floor because it is the one tool that can create nothing.
- **The setter hardening is correct in both directions.** `Object.freeze([...value])` plus `SameRoster` fixes identity being wrong for a rebound literal *and* for an in-place mutation. Order-sensitivity is harmless: a pure reorder re-runs a clamp that is a no-op because the held tool is still allowed.
- **The toolbar field is safe**: `Tools` is declared at :109, `VisibleTools` at :127, and the setter only runs post-construction.
- `ElementCount` and `Select` exist as the new specs use them.

The non-array work is the best thing here. Checking both halves **against the old setter** — and correcting your own first-pass claim that a string always throws, when it in fact silently substring-matches — is what turned a guessed fix into a known one. That distinction is worth more than the fix.

## What I changed on the branch (`1d619f10da`)

Both changesets were going to ship in the **same** release. `.changeset/whiteboard-tool-roster.md` from #4310 is still unconsumed — the last `Version Packages` on `next` is from June, and 383 changesets are pending — so #4310's text has never shipped. Left as it was, one release's notes for `@memberjunction/ng-whiteboard` would have said Restyle is unaffected by the roster and that a roster leaving no known tool keeps the current one, then said both were false a few paragraphs later, pointing at a "previous release" that does not exist.

Since the file is still editable, I corrected it in place rather than asking you to round-trip: the ToolRoster entry now describes what actually ships (Restyle hidden without `text`, a roster naming no real tool landing on `select`, `ReadOnly` named as the mutation axis), and this PR's entry drops the corrections-to-a-previous-release paragraph and states its behaviour change against the audience it actually has — adopters from an unreleased build or the `lts/5` backport. No code touched; `check:changeset` passes. Revert it if you'd rather word it yourself.

## Should fix: the `Tool` docstring now overstates its own invariant

`whiteboard-host.component.ts:221` (unchanged here, so no inline anchor) still asserts, unqualified, that *"The active tool is always one the roster allows"* and that either assignment order *"still lands on an allowed tool"*. After this PR that is false for exactly the case this PR added: under `ToolRoster = []` the tool lands on `select`, and `IsToolAllowed([], 'select')` is `false`. The behaviour is right — `select` as the floor is the whole point — but that docstring sits on the setter the exception runs through, and it is now the one place in the package still claiming the absolute rule. A clause naming the floor would settle it.

## Correctly deferred

Image paste under a roster omitting `image` reads as a decision, not an oversight, and I agree with where you landed: that is `PlaceableKinds` or `ReadOnly`, not the roster. The four hand-kept tool tables are a refactor and belong in their own PR.

rkihm-BC (approved):
Re-approving on `9f5e9a75a5`. My previous approval was dismissed by my own push, not by anything you did.

The should-fix from that review is now addressed on the branch (`9f5e9a75a5`, comment only): the `Tool` setter's docstring names the exception it acquired here — a roster allowing nothing at all floors the tool at `select`, which is then held even though the roster does not list it, because `select` is the one tool that can create nothing. It was the last place in the package still asserting the absolute rule, and it sat on the setter the exception runs through.

So the branch now carries two commits from me, both text-only:

- `1d619f10da` — corrected #4310's still-unconsumed changeset in place, and dropped the corrections-to-a-previous-release paragraph from this one, so the single release these both ship in does not contradict itself.
- `9f5e9a75a5` — the docstring above.

Revert either if you would rather word it yourself; neither touches behaviour, and `check:changeset` passes on the first.

Everything from the substantive review stands: Restyle gating is complete because the context menu is its only door, the `ReadOnly` passthrough is real and cannot orphan a `ViewChild`, the `select` floor is the right answer for an empty roster, and the non-array work — checked against the old setter, which corrected your own first-pass claim that a string always throws — is the strongest part of the change. Nothing further from me.

PR #4321, opened 2026-09-09
fix(core): make the polymorphic EntityID/RecordID soft-link path actually work

MS-BC:
## What this is

The code half of `plans/polymorphic-foreign-keys.md` (#4084) — specifically the Layer 1 bug fixes and Layer 0 items 1–3. Nothing here declares a polymorphic pair, so **runtime behaviour is unchanged until `EntityIDFieldName` is populated** (see "Deliberately not included").

Split out on @rkihm-BC's advice in the #4084 review: *"B1 and B2 are shippable now, independent of every decision below — I'd split them into a fix PR rather than hold them behind four open questions."* This PR is that, plus two further defects found while making them testable.

MJ models "a pointer to any record" as an `EntityID` + `RecordID` pair. `EntityField.EntityIDFieldName` has existed since the v2 baseline to declare one, is documented for exactly this purpose, and has consumer code in both DB providers. It is `NULL` on every row in the system — so that consumer has never executed, and it did not work.

## Defects fixed

| | Defect | Where |
|---|---|---|
| **B1** | Discriminator filtered on the **holder** of the link, not the entity whose dependents were sought — it looked for `TaskLink` rows whose `EntityID` points at `TaskLink`. The target entity's ID was never resolved anywhere in the function. | `SQLServerDataProvider.ts`, `PostgreSQLDataProvider.ts` |
| **B2** | `GetRecordDependencies` returned as soon as no *hard* FK dependents existed, so the soft-link SQL was never built for an entity whose only dependents are polymorphic — the exact case it exists to serve. | `databaseProviderBase.ts` |
| **B3** | Compared the payload column against the bare first primary key value, not the field-prefixed form a `RecordID` column stores. | both providers |
| **B4** | Quoting derived from the **holder's** PK type, then applied to both the entity-ID and record-ID literals — an integer-keyed holder emitted malformed SQL. Both literals are now always quoted (the discriminator is a uniqueidentifier, the payload nvarchar), and values are escaped. | both providers |
| **B5** ⚠️ | **Not in the plan.** Record merge repointed every dependency by writing the bare primary key value. Correct for an FK, silently wrong for a `RecordID` column — it would leave pointers that resolve to nothing *and* re-introduce the second encoding this work exists to remove. Now an explicit, tested seam (`ResolveMergeLinkValue`). | `databaseProviderBase.ts` |
| **B6** | Dependency rows mapped the dependent record's key onto the **parent** entity's PK columns rather than the entity whose row it actually is. Consumers (merge) already treated it as the related entity's key. | `databaseProviderBase.ts` |

## New API on `CompositeKey`

- **`ToRecordID()`** — the canonical serialization for a polymorphic `RecordID` column. Unlike `ToConcatenatedString()` it **throws** rather than emitting a string that cannot be parsed back: a value containing the delimiter, or a null key component (which would render as the literal text `"null"`).
- **`FromRecordID(entity, s)`** — parses **and validates field names and arity against the entity's primary keys**. Deliberately not a rename of `LoadFromConcatenatedString`, which cannot signal failure — given a string with no delimiter it leaves the key empty and returns normally, which is why consumers hand-roll delimiter sniffers today. The name check is also what safely rejects a legacy bare composite (`val1||val2`), which parses cleanly as a *shape* but names fields the entity does not have.
- **`FromLegacyRecordID(entity, s)`** — the transitional bare-value form, named so it is obviously temporary, so reading a legacy encoding is a decision at the call site rather than a guess inside the parser.

## Behaviour change

`GetRecordDependencies` now **throws** for an entity that is not in metadata, where it previously returned `[]`. Reporting "no dependencies" for an entity we could not resolve is indistinguishable, to a caller about to delete a record, from "this record is safe to delete" — the same failure mode as B2. One PG stub test asserted the old return and is updated with the reasoning.

## Deliberately not included

**No `EntityIDFieldName` values are populated**, so no pair is declared and the path stays dormant.

Before seeding it I audited what those columns actually contain. **90 write sites across the repo set a polymorphic payload column. 2 use the canonical encoding.** The rest write:

- a bare single value (`String(value)`) — the common case, including `MJListDetailEntityExtended` for single-PK entities
- a comma-space joined list (`CompositeKey.Values()`) — `UserFavorite`, via the provider
- a `||`-joined bare list — `RecordGeoCode`
- `Field=Value AND …` (`CompositeKey.ToString()`) — `DataContextItem`
- ad-hoc, not a record pointer at all — `CompanyIntegrationRunDetail` stores `EntityMap:{id}|Processed:{n}`

That is **five** encodings, where the plan documents two, and it makes D1's "it is already the majority format" false — canonical is ~2%. Declaring pairs before those writers are normalized would make merge write `ID|<guid>` into columns that mostly hold bare values, converting a dormant path into an actively corrupting one. Normalizing them is Layer 0 and is materially larger than the plan's "migrate `RecordGeoCode`".

I'll fold this into #4084.

## Testing

**39 new tests. All suites pass with exit code 0.**

| Package | Result |
|---|---|
| `@memberjunction/core` | **157 files / 2289 tests passed** (was 155 / 2250 — +2 files, +39 tests) |
| `@memberjunction/sqlserver-dataprovider` | **15 files / 158 tests passed** |
| `@memberjunction/postgresql-dataprovider` | **214 passed**, 7 skipped |

All three packages build clean (`exit=0`).

**Mutation-checked, because a test that passes against both old and new code is worthless.** I temporarily restored each original file and re-ran:

- `softLinkDependencySQL.test.ts` (SQL Server): **6 of 9 fail** against pre-fix code. The 3 that pass assert behaviour that was already correct (reads from the holder, one `UNION ALL` branch per field, empty when nothing is declared).
- `databaseProviderBase.softLinkDependencies.test.ts`: **5 of 7 fail** against pre-fix code.

Coverage includes both dialects for B1/B3/B4 (PG is a mirror, not an assumption), the integer-keyed-holder case for B4, apostrophe escaping, the `IsSoftLink` marker in both SQL Server `bit` and PG `boolean` shapes, and the B5 encoding decision in all three states (soft link → canonical, hard FK → bare, flag absent → bare, preserving pre-existing behaviour).

**CI gates:** `check:standards` passes. `check:esm` and `check:claude-md` report failures that are **pre-existing on `next`** and unrelated — `CLAUDE.md` is already 19,475 bytes over its 18,000 ceiling on `origin/next` and I did not touch it, and my two providers appear in `check:esm` only as **non-gating** `DEP_FAIL` from an unbuilt `MJStorage` dist in my worktree.

**⚠️ The deterministic integration tier was NOT run.** It could not start in my worktree: `mj test suite` resolves the CLI from the main repo checkout, and building `MJCLI` locally fails on an incomplete `TestingFramework/CLI` dist. This change ships no migration, so it has no schema seam for that tier to exercise — but I have not verified it and it should run in CI or on a proper install before merge.

## Relationship to #4084

#4084 stays the reference document. This PR does not depend on any of its four open questions, and it does not commit to D2 (the 750 widen) or D4 (the policy columns) — nothing here touches the schema. Ships no migration, so no PostgreSQL counterpart is required.


MS-BC:
@rkihm-BC @AN-BC — review requested. All checks green, including the deterministic integration tier. Flagging what specifically needs each of you rather than asking for a general pass.

**@rkihm-BC** — this is the split you asked for in #4084 ("B1 and B2 are shippable now… I'd split them into a fix PR rather than hold them behind four open questions"). Two things worth your eye:

- It grew from your two bugs to **six**. **B5** is the one I'd most like challenged: record merge repointed every dependency with `GetValueByIndex(0)`, the bare key value. Right for a foreign key, silently wrong for a `RecordID` column. It means B1/B2 could *not* have shipped alone — fixing the query without the write-back turns a dead path into a corrupting one. I extracted it to `ResolveMergeLinkValue` specifically so it is testable, because nothing about the resulting row makes the mistake visible afterwards.
- I also closed your `ALTER COLUMN` verification item, and the answer has a sting: widening an indexed `NVARCHAR` is metadata-only (no index drop, so that pessimism can leave the estimate), **but an over-limit index is created with only a warning** and fails later at insert time. A successful `CREATE INDEX` is not evidence that it fits.

**@AN-BC** — this touches two of your plan's load-bearing claims, so you should probably disagree with me if I have it wrong:

- **D1's "it is already the majority format" does not hold.** An audit of every write to a polymorphic payload column found **90 write sites, 2 canonical**. Five encodings ship, not two — `UserFavorite` is written both ways (provider `Values()`, Angular bare `.ID`), and `ListDetail` switches encoding on key arity, so "which encoding is this column?" is not answerable per column.
- Consequently **I seeded no `EntityIDFieldName` values here.** Declaring a pair against a bare-valued column makes the dependency query match nothing *and* makes merge write a value the column's own readers cannot parse. So Layer 1 is blocked on normalising ~88 writers rather than on re-encoding `RecordGeoCode`. The mechanism in this PR is correct and deliberately dormant — nothing changes at runtime until those writers agree. That is the part most worth pushing back on if you read the estate differently.

On testing: the tests are mutation-checked against the pre-fix files (6 of 9 provider tests and 5 of 7 base tests fail without the fixes), because a test that passes either way is not evidence. The one thing I could not verify locally is the integration tier — `mj test suite` resolves the CLI from the main checkout and `MJCLI` would not build in my worktree — so CI is the only run of it. It passes.

Plan corrections are on #4084, replied there and on all three inline threads.

AN-BC:
> @rkihm-BC @AN-BC — review requested. All checks green, including the deterministic integration tier. Flagging what specifically needs each of you rather than asking for a general pass.
> 
> **@rkihm-BC** — this is the split you asked for in #4084 ("B1 and B2 are shippable now… I'd split them into a fix PR rather than hold them behind four open questions"). Two things worth your eye:
> 
> * It grew from your two bugs to **six**. **B5** is the one I'd most like challenged: record merge repointed every dependency with `GetValueByIndex(0)`, the bare key value. Right for a foreign key, silently wrong for a `RecordID` column. It means B1/B2 could _not_ have shipped alone — fixing the query without the write-back turns a dead path into a corrupting one. I extracted it to `ResolveMergeLinkValue` specifically so it is testable, because nothing about the resulting row makes the mistake visible afterwards.
> * I also closed your `ALTER COLUMN` verification item, and the answer has a sting: widening an indexed `NVARCHAR` is metadata-only (no index drop, so that pessimism can leave the estimate), **but an over-limit index is created with only a warning** and fails later at insert time. A successful `CREATE INDEX` is not evidence that it fits.
> 
> **@AN-BC** — this touches two of your plan's load-bearing claims, so you should probably disagree with me if I have it wrong:
> 
> * **D1's "it is already the majority format" does not hold.** An audit of every write to a polymorphic payload column found **90 write sites, 2 canonical**. Five encodings ship, not two — `UserFavorite` is written both ways (provider `Values()`, Angular bare `.ID`), and `ListDetail` switches encoding on key arity, so "which encoding is this column?" is not answerable per column.
> * Consequently **I seeded no `EntityIDFieldName` values here.** Declaring a pair against a bare-valued column makes the dependency query match nothing _and_ makes merge write a value the column's own readers cannot parse. So Layer 1 is blocked on normalising ~88 writers rather than on re-encoding `RecordGeoCode`. The mechanism in this PR is correct and deliberately dormant — nothing changes at runtime until those writers agree. That is the part most worth pushing back on if you read the estate differently.
> 
> On testing: the tests are mutation-checked against the pre-fix files (6 of 9 provider tests and 5 of 7 base tests fail without the fixes), because a test that passes either way is not evidence. The one thing I could not verify locally is the integration tier — `mj test suite` resolves the CLI from the main checkout and `MJCLI` would not build in my worktree — so CI is the only run of it. It passes.
> 
> Plan corrections are on #4084, replied there and on all three inline threads.

@MS-BC no further comments on my part, with the integration suite running green I'm ok with this as is if @rkihm-BC is. In terms of downstream application possibilities, with this in place shortly we should lean into it more where we use polymorphic fkeys (lots of places). Good stuff. Thanks.

rkihm-BC (dismissed):
Approving. This is careful work and the split is exactly what I asked for in #4084 — it just came back with four more defects than I sent it out with.

Before the list: **none of the follow-ups below should block this merge, but I do want them written down rather than left in this thread.** Please open issues (or fold them into #4084 where they're clearly Layer 0) before this drops off the front page. The reasoning in this PR body is the most valuable artifact produced so far on polymorphic FKs, and right now it exists only as prose on a PR.

## What's good

**B5 is the catch of the PR, and your argument for it is right.** "Fixing the query without the write-back turns a dead path into a corrupting one" is correct — B1+B2 alone would have made merge write bare values into `RecordID` columns the moment anyone declared a pair. Extracting `ResolveMergeLinkValue` purely so an otherwise invisible decision becomes assertable is the right instinct.

**Mutation-checking the tests.** 6-of-9 and 5-of-7 failing against restored originals is the only evidence that means anything, and almost nobody bothers. Thank you.

**Refusing to seed `EntityIDFieldName` on the strength of the audit is the best decision here.** I checked: no migration in the repo has ever assigned a non-null `EntityIDFieldName`, and the metadata sync logs show it explicitly cleared. Your 90-sites/2-canonical finding correcting D1 is worth more than the code changes, and declining to build on a premise you'd just disproved was the right call over shipping something that looked more complete.

**Symmetric dialect fixes.** Both providers, mirrored rather than assumed, and both are the only implementors — so no third provider is left with a column-count mismatch in the `UNION ALL`. The new `EntityByName(entityName)` guard also incidentally closes an injection surface: `entityName` was being interpolated raw into `'${entityName}' AS EntityName`.

## Follow-ups — please document these

**1. B6 has a second half you didn't reach, and it's the more serious one.**

You fixed the *parsing* side (`EntityByName(r.RelatedEntityName)`). The *SQL* side still has the same wrong-entity defect: `BuildHardLinkDependencySQL` builds `primaryKeySelectString` from **`entityInfo.PrimaryKeys` — the parent's** — while selecting `FROM relatedEntityInfo`'s base view. Both providers (`SQLServerDataProvider.ts:911`, `PostgreSQLDataProvider.ts:158`). So `PrimaryKeyValue` is labeled with the parent's PK column names, read out of the related entity's view.

That changes how B6 should be read:

- **For soft links it's correct and necessary** — there the SQL loops holders as `entity`, builds the key from `entity.PrimaryKeys`, and emits `entity.Name` as `RelatedEntityName`, so parsing against `RelatedEntityName` lines up.
- **For hard links it's a no-op in the common case** (parent and related both keyed `ID`, which is why the integration tier is green) **and doesn't fix the differing-name case.** There the SQL emits e.g. `OrderID|<parent's own id>` pulled from the related view's FK column. Before: mapped onto the parent's PKs. After: mapped onto the related entity's PKs, yielding `undefined`. Both broken, differently.

The actual fix is `relatedEntityInfo.PrimaryKeys` in the SQL. It's the same defect class as B1, three lines from code this PR edited to add the two new columns. I'd rank this above everything else in this list.

**2. `ToRecordID()` is evaluated eagerly, so the "dormant" property leaks.**

In both providers `const targetRecordID = compositeKey.ToRecordID()` runs *before* the loop over entities, so it executes on every `GetRecordDependencies` call even though every entity yields `''` today. `ToRecordID()` is the one new method whose contract is to *throw* — empty key, null key component, or a value containing `|`. So the method picks up three new throw conditions unrelated to polymorphic links, on a path where no pair is declared.

To be fair: I went looking for a caller that trips it and couldn't demonstrate one. `SimpleLoadFromURLSegment` (the duplicate-detection path) never yields a null component. So this is a latent contract change, not a shown break — which is precisely the argument for moving the call inside the field loop. One line, and the question stops existing.

**3. Escaping was fixed in the soft path and not the hard path.**

Both providers still emit the hard-link value unescaped — `GetRecordDependencyLinkSQL:938/943` in SQL Server, inline in PG. There's now a passing `O'Brien` test asserting the soft path handles it, which makes the adjacent gap look deliberate.

**4. Use the existing escape helper.** `EscapeSQLString` is in `@memberjunction/global` (`util.ts:1661`), already used by MJCore in two places, and strips NUL bytes as well as doubling quotes. `@memberjunction/global` is already a declared dependency of both provider packages, so it's a drop-in — the new `escapeSQLLiteral` plus the inline PG `.replace()` make this the third and fourth implementation, and the weakest two.

**5. The new flags don't cross the GraphQL transport.** `RecordDependencyResult` in `MergeRecordsResolver.ts` declares neither `IsSoftLink` nor `EntityIDFieldName`, and the client query doesn't select them, so any client-side consumer sees `undefined` and treats every link as hard. Harmless today since merge runs server-side, and nothing currently reads `dependency.PrimaryKey` off a GraphQL result — but that same query aliases the key as `CompositeKey` while `RecordDependency` declares `PrimaryKey`, so that field is already `undefined` over the wire despite the comment claiming the shape matches. Both worth an issue before Layer 1 leans on any of it.

**6. Exact-match encodes the assumption your own audit disproved.** `[RecordID] = 'ID|<guid>'` matches canonical rows only, and you found ~2% of writes are canonical. So the first pair declared against a real column silently finds zero dependents — B2's failure mode, relocated. Either match both encodings during transition, or make Layer 0 normalisation a recorded prerequisite somewhere enforceable rather than a paragraph in a PR description.

**7. Minor:** the soft-link builders compute `primaryKeySelectString` for every entity in metadata *before* filtering for polymorphic fields, and that loop now runs on every call since the early return is gone — a full pass building and discarding strings to produce `''`. Hoisting it behind the filter is free. Also `ResolveMergeLinkValue` returns `unknown` (the repo style rule discourages it; the real type is the PK value union), and `vi` is imported but unused in `databaseProviderBase.softLinkDependencies.test.ts`.

## On the integration tier

@AN-BC and I are both fine merging on the green run, but I want to be precise about what it establishes, because it's easy to over-read.

No pair is declared anywhere in the repo, so `BuildSoftLinkDependencySQL` returned the empty string on every call during that run. **The integration tier exercised none of the new mechanism.** What it does establish — and this genuinely matters, given B6 touches the live hard-FK path — is that the dormant configuration doesn't regress merge or delete. That's the right evidence for "safe to merge."

The mechanism itself is covered only by string assertions against mock `EntityInfo` objects. Nothing executes the generated SQL against SQL Server or PostgreSQL, and `MergeRecords` is never invoked in a test — only the extracted encoding decision. So a bracket error, a wrong view name, or the item-1 defect would pass all 39 tests.

That makes one thing a gate rather than a nice-to-have: **before any `EntityIDFieldName` is seeded, there needs to be an integration bundle that declares a pair on a real table and drives `GetRecordDependencies` + `MergeRecords` end to end in both dialects.** @AN-BC — that's the constraint on "lean into it more where we use polymorphic fkeys." The ambition is right; the estate isn't ready for it until Layer 0 normalisation lands and that bundle exists. Worth capturing on #4084 so the next person to pick this up doesn't start declaring pairs against bare-valued columns.

Nice work. Approving.

rkihm-BC:
Merged current `next` into this branch and pushed the resolution — `63bc733cef..229d35cabd`. Reaching into your branch; revert if you would rather do it yourself.

## The conflict was one line, in both providers, and it is worth knowing about

`#4304` (the primary-key sweep, `92f2ac90b3`) landed while this was open and touched the exact line B4 deletes:

```diff
-  const quotes = entity.FirstPrimaryKey.NeedsQuotes ? "'" : '';
+  const quotes = entity.FirstPrimaryKey.NeedsQuotes ? "'" : ''; // first-pk-ok: a soft-link column
+     stores one bare key value; matched against the first key value below by design
```

So the sweep added an annotation **blessing the defect this PR fixes** — and its stated rationale is wrong on both counts your tests prove:

- *"a soft-link column stores one bare key value"* — no, it stores the field-prefixed form. That is your **B3**.
- *"by design"* — the quoting comes from the **holder's** key type and is applied to both literals, producing malformed SQL for an integer-keyed holder. That is your **B4**.

Resolved in your favour: the line goes, the annotation with it. That is the only semantic content of the merge; everything else auto-merged.

I have raised the general problem as #4330 — the sweep added 166 of these annotations, the gate verifies that one *exists* rather than that it is *true*, and I found a second wrong one of the same shape (`ResourcePermission.ResourceRecordID stores one key value, so resource entities are single-column by contract` — that column has a sibling `ResourceTypeID` discriminator and is written in three different encodings). Your audit is cited as the evidence.

Also #4331: `.claude/rules/data-access.md` now states there are "two persisted formats … which must not be unified", which your audit contradicts with five across 90 write sites. That file loads on every `.ts` task, so it is worth reconciling.

## Verification

After resolving, on the merged tree:

```
softLinkDependencySQL.test.ts  (SQL Server)   9 passed
softLinkDependencySQL.test.ts  (PostgreSQL)   7 passed
compositeKey.recordID + databaseProviderBase.softLinkDependencies  39 passed
```

All three packages build clean. I also confirmed `quotes` has no dangling reference left in either soft-link builder.

## CI

#4325 merged and I have now validated it on `next` — the drift gate is **green** there (run 34373861730). Both gate defects that were failing every branch are gone, so this PR's re-run should be a true signal rather than the bootstrap failure it was hitting before.

My approval from earlier stands; nothing in the resolution touches the reviewed behaviour.

rkihm-BC (approved):
Re-approving — my approval was auto-dismissed by the merge commit I just pushed (`229d35cabd`), not withdrawn.

Being explicit since I authored the dismissing commit: the only semantic content of that merge is deleting one line in each provider — the `const quotes = entity.FirstPrimaryKey.NeedsQuotes …` that this PR removes as B4, which #4304 had re-added with a `first-pk-ok` annotation blessing it. Resolved in this PR's favour. Everything else auto-merged.

Verified on the merged tree: 9 + 7 provider tests and 39 MJCore tests pass, all three packages build clean, no dangling `quotes` reference in either soft-link builder.

The substantive review and its follow-ups are unchanged. The drift gate is now green on `next`, so this PR's CI re-run is a real signal.

rkihm-BC:
@MS-BC — asking you to review the merge commit `229d35cabd` specifically, not the PR as a whole (your work there already has my approval).

I resolved the conflict with `next` and pushed it to your branch, then re-approved after my own push dismissed the approval. That means **no independent reviewer has looked at the resolution** — which is exactly the situation where a second pair of eyes matters, since the resolution is a judgment call rather than a mechanical merge.

What to check:

- The conflict was one line in each provider: `const quotes = entity.FirstPrimaryKey.NeedsQuotes ? "'" : ''`, which your B4 deletes and which #4304 re-added with a `first-pk-ok` annotation asserting it was correct by design. I resolved in your favour — line and annotation both removed. Confirm that is what you intended B4 to do, and that nothing in #4304's sweep implies a different fix.
- I verified `quotes` has no dangling reference left in either soft-link builder, but you know that code better than I do.

Verification on the merged tree: 9 (SQL Server) + 7 (PostgreSQL) + 39 (MJCore) tests pass, all three packages build clean.

If you would rather redo the merge yourself, revert `229d35cabd` and I will stay out of it.

PR #4324, opened 2026-09-09
fix(codegen): eliminate fresh-DB drift, add IdentityClaim tail migration, and address reviewer nits

AN-BC:
## Summary

Closes the CI `codegen-drift` gate failure and addresses reviewer AN-BC's follow-up cleanups:

1. **Supply missing IdentityClaim metadata tail (`V202609091200__v6.1.x__IdentityClaim_Metadata_Tail.sql`)**:
   `V202608202300` created the `IdentityClaim` table and entity, but omitted the CodeGen metadata tail for Status values (`Claimed`, `Expired`, `Pending`, `Revoked`), `ValueListType = 'List'`, 4 foreign key `EntityRelationship` entries, and 3 related entity name field maps. On clean databases, CodeGen detected this gap and emitted a `CodeGen_Run_*.sql` capture file. This migration supplies the metadata tail so CodeGen generates 0 drift statements.

2. **Supply missing `sync` block in `metadata/entity-permissions/.entity-permissions.json`**:
   The 34th record (`CFDBE38F-1222-40B4-82EC-F3EA49F785C8`) added in `0d3094c543` had no `sync` metadata block. When `sync push --dir=metadata --ci` ran against a clean DB, it auto-created the record and wrote back the generated `sync` block to disk. Populated the stable `sync` block and checksum on the committed file.

3. **Align `packages/MJCoreEntities/src/generated/entities/__mj.ts`**:
   Synchronized 7 entity field `DisplayName` values with the physical database schema (`Create Stored Procedure`, etc.).

4. **Address Reviewer AN-BC's nits in `packages/MetadataSync/src/lib/json-write-helper.ts`**:
   Renamed `normalizeRecordDataOrder` -> `preserveAndRecurseRecordData` and cleaned up `knownKeys` using a `Set`.

5. **Fix single-column restore in `scripts/codegen-idempotency-check.mjs`**:
   Passed `--skipdb` to the teardown restore CodeGen run so schema files are cleanly regenerated without the probe column.

## Verification
- Verified against clean-room database `MJ_DRIFT_GATE_TEST`:
  - `mj migrate`: 76 migrations applied.
  - `mj sync push --dir=metadata --ci`: 14,050 unchanged, 0 updated, `.entity-permissions.json` 0 diff.
  - `mj codegen --no-ai --skip-commands`: 0 SQL files emitted, 0 diffs.
  - `codegen-idempotency-check.mjs --stage warm-twice`: PASSED (0 diffs, 0 surviving SQL logs).
  - `codegen-idempotency-check.mjs --stage single-column`: PASSED.
  - `check-migration-entityfield-sequence.mjs`: PASSED.
  - Unit tests:
    - `MetadataSync`: 364/364 passed.
    - `CodeGenLib`: 1394/1394 passed (66 skipped).
    - `MJCoreEntities`: 669/669 passed.


AN-BC:
## `9ef93218` verified — all five, read from the code. And you found the drift I was blind to.

| Item | Verified |
|---|---|
| IdentityClaim tail migration | `migrations/v6/` — folder matches `v6.1.x` in the filename ✅ · `${flyway:defaultSchema}` throughout, no other placeholder ✅ · no `NEWID()` ✅ · 8 idempotency guards ✅ · **no `EntityField` INSERT at all** (one `UPDATE … SET ValueListType`), so the `UQ_EntityField_EntityID_Sequence` fresh-install trap doesn't apply; the `Sequence` columns present are on `EntityFieldValue`/`EntityRelationship`, different tables ✅ · `__mj` timestamps in the metadata INSERTs match what **21 other v6 migrations** already do, so that's convention, not deviation ✅ |
| `.entity-permissions.json` sync block | 34/34 records now carry one |
| `__mj.ts` DisplayNames | legitimate regeneration — see below |
| `json-write-helper` nits | `preserveAndRecurseRecordData` + `Set` — both closed |
| `--skipdb` on the restore run | present |

### The `__mj.ts` change looked like a rule violation. It isn't, and I checked rather than assumed

"Aligned display names in a `generated/` file" is exactly the shape of a hand-edit, and hand-editing CodeGen output would be the wrong fix — the next run reverts it and the gate goes red again. So I traced the strings at your head:

```
"Create Stored Procedure"   -> migrations/v5/B202602151200__v5.0__Baseline.sql
                               migrations/v5/V202602141421__v5.0.x__Add_AllowMultipleSubtypes_to_Entity.sql
                               migrations/v6/V202608141412__v6.1.x__Entity_And_EntityRelationship_Configuration.sql
                               packages/MJCoreEntities/src/generated/entities/__mj.ts

"Stored Procedure Create"   -> nowhere in the repo
```

The migrations set `Create Stored Procedure`. The committed file carried a value **no migration produces** — it was generated from a database predating `V202608141412`. So the file was stale and now matches a clean run. That's the drift gate doing precisely the job it exists for: catching a generated artifact that quietly stopped matching what the repo actually produces.

### Where I was wrong, because it generalises

I told you the one residual was `metadata/prompts/.prompts.json` record `7D9B05FA`. Your clean-room run reports **14,050 unchanged, 0 updated**, so that record didn't move — it was benign, as the second of my two readings said.

But I never found `CFDBE38F`, and the reason is a defect in my method, not bad luck. My scan iterated:

```js
if (x.sync && x.sync.checksum) { /* compare stored vs computed */ }
```

I asked *"does the stored checksum match?"* and never asked *"is there a stored checksum at all?"* A record with **no `sync` block** was invisible to me — and that class is strictly worse than a mismatch, because it is **guaranteed** to drift: the push creates the record and writes the block back, every time, on any clean database. The one category my check couldn't see is the one category that always fails.

Same failure shape I've been flagging all night, committed by me: a check that passes while the thing it names is broken.

**Which suggests a cheap durable gate.** "Every record in `metadata/**` that has a `primaryKey` also has a `sync.checksum`" is a pure static check — no database, no CodeGen, runs in a second — and it would have caught this at PR time instead of costing a clean-room cycle. Worth adding next to `check:codegen-tail`.

### One note on merge state

Reads `behind` again; measured, and again harmless:

```
next gained : 4 commits, 14 files
this PR     : 5 files
overlap     : 0
```

Nothing to inspect after you bring `next` in.

---

**Nothing outstanding from me on this PR.** The clean-room verification against `MJ_DRIFT_GATE_TEST` — 76 migrations, sync push 0 updated, codegen 0 SQL files and 0 diffs, both idempotency stages, and the sequence check — is exactly the run that was missing, and it's the one that makes the now-blocking gate trustworthy rather than merely armed.

---

PR #3367, opened 2026-07-30
Field-Level Security: per-field Read/Update/Create control by role

jordanfanapour:
Rebased onto `next` (56 commits ahead, 0 behind, linear). **Review feedback addressed** — the Record Changes gap is now enforced rather than documented; see *Audit trail* below.

Role-based, per-field access control for entity data — who may **read**, **update** and **create** each column — enforced server-side. Built for the compensation / donor-giving / personnel-record class of requirement: an entity most users may work with, containing a few columns most users must never see.

This fills the gap between entity-level CRUD permissions and row-level security. Until now the only field-scoped feature was encryption-at-rest, which obfuscates data but does not control per-role visibility.

> The PR title previously said "Implementation Plan". The design changed direction twice during implementation and the planning documents are gone; **the code is the specification**, and `guides/FIELD_LEVEL_SECURITY_GUIDE.md` is the written form of it.

---

## Configuration model

**Field security is on or off per entity**, explicitly, via a new `Entity.EnableFieldLevelSecurity` flag. Nothing is inferred from whether permission rows happen to exist, so adding a rule can never change access on an entity that has not opted in. Every enforcement point tests that one boolean first and short-circuits — no field iteration, no aggregation, no allocation on the ~99% of entities that never enable it.

A new `EntityFieldPermission` table holds one row per (field, role) with three independent **trinary** verbs — `ReadAccess`, `UpdateAccess`, `CreateAccess` — each `Allow`, `Deny`, or `No Access`:

| Value | Meaning |
|---|---|
| `No Access` | Neutral, and the default. Grants nothing, blocks nothing — another role's `Allow` still wins. |
| `Allow` | Grants the action for that role. |
| `Deny` | Beats everything. One `Deny` anywhere across the user's roles wins, however many `Allow`s sit beside it. |

Per verb, across the roles a user holds: `effective = (any row Allows) AND NOT (any row Denies)`.

**Read is required for Update and Create**, enforced twice because neither layer alone is sufficient. A CHECK constraint refuses the combination within a row; the aggregation clamps it again across roles, because role A granting Read+Update and role B denying Read are each individually legal and combine, for a user holding both, into write-without-read that no per-row constraint can see. A consequence worth knowing: **write-only fields are not a supported configuration.**

[...]

PR #4325, opened 2026-09-09
fix(codegen): supply the IdentityClaim CodeGen tail and unblock the drift gate

### LINKED ISSUE #4323: CodeGen drift gate red on next: pre-existing drift residue, exposed (not caused) by #4317
## Summary

The `CodeGen drift gate` has been failing on `next` since `80653f1f3e` (the merge of #4317), and is red on every commit since, including the current tip `770cfcab34`. It fails every PR built on top of `next`.

**This is not a regression in #4317.** The opposite: #4317 *reduced* the drift substantially and made the gate actually gate. What is left is a pre-existing residue that was previously being swallowed.

## What actually happened

The gate's `Diff committed generated artifacts` step carried `continue-on-error: true`. It has been detecting drift and reporting it for a long time, while the job went green anyway. #4317 (`048fdd7d29`, "blocking drift gate") removed that line — correctly.

The evidence is in the last green run itself, [job 102448442789 on `ce3d526610`](https://github.com/MemberJunction/MJ/actions/runs/34345417824/job/102448442789). It logs:

```
```

…followed by **41 files**, and still reports success.

Comparing that run to the current tip:

| | last "green" (`ce3d526610`) | tip (`770cfcab34`) |
|---|---|---|
| `metadata/entities/*.json` | 34 | 0 |
| `metadata/prompts`, `metadata/ai-models` | 3 | 0 |
| `metadata/entity-permissions/.entity-permissions.json` | 1 | 1 |
| `packages/MJCoreEntities/src/generated/entities/__mj.ts` | 1 | 1 |
| `packages/MJServer/src/generated/graphql-schemas/__mj.ts` | 1 | 1 |
| `?? migrations/v5/CodeGen_Run_*.sql` | 1 | 1 |
| **total** | **41** | **4** |

So #4317's metadata-sync work (key ordering, bare-fields checksum payload) eliminated 37 of the 41. The remaining four are what needs fixing, and they are now correctly fatal.

## What remains

```
 M metadata/entity-permissions/.entity-permissions.json
 M packages/MJServer/src/generated/graphql-schemas/__mj.ts
?? migrations/v5/CodeGen_Run_<timestamp>.sql
```

The untracked `CodeGen_Run_*.sql` is the one worth understanding rather than just regenerating past. CodeGen emitting a migration against a database built **purely from committed migrations plus `mj sync push`** means it believes there are schema changes still to apply that no committed migration performs. That is the same class of gap as `V202609081830__v6.1.x__IdentityClaim_Metadata_Healing.sql` on #3367 — CodeGen-owned artifacts that exist on every developer database (because `mj codegen` creates them locally) and on no database built only from what the repo ships.

## Proposed fix

Reproduce the gate locally (fresh SQL Server, `mj migrate` → `mj sync push --dir=metadata --ci` → `mj codegen`), then:

1. Inspect the emitted `CodeGen_Run_*.sql` and decide whether its contents belong in a committed migration. If so, that is the actual fix and the two generated `__mj.ts` files should fall into line behind it.
2. Commit whatever the two generated artifacts and `.entity-permissions.json` legitimately become.

I am putting up a PR for this.

## Please do not

Re-add `continue-on-error: true`. Making the gate blocking was right — it is the only reason a long-standing drift became visible at all.

## Impact

Every open PR against `next` inherits this failure. #3367 currently shows 13 drifted paths: these 4 inherited, plus 9 of its own that do need regenerating on that branch.

cc @AN-BC — flagging since it came out of #4317, though as above the change there was an improvement, not the cause.

rkihm-BC:
Fixes #4323. The `CodeGen drift gate` has been red on `next` since it stopped being advisory, failing every PR built on top of it.

## Two independent causes, both pre-existing

**1. `Identity_Claims_Infrastructure` shipped without its CodeGen tail.** Three classes of CodeGen-owned artifact were never committed, so they exist on every developer database — where `mj codegen` creates them locally — and on no database built purely from committed migrations:

- 4 × `EntityRelationship` (Entities, Users, Identity Claim Types, Magic Link Invites → Identity Claims)
- 4 × `EntityFieldValue` behind `CK_IdentityClaim_Status`, plus the `ValueListType='List'` flag
- 3 × `RelatedEntityNameFieldMap` assignments

**2. The gate failed on its own scratch output.** The diff step's comment says it ignores CodeGen's run file; the code only ever removed `codegen.output.log`. CodeGen writes a `CodeGen_Run_<timestamp>.sql` on **every** run — empty when it has nothing to emit — so the gate could not go green even at zero real drift. Both copies of the step now remove it, as the comment always claimed.

## Verified, not assumed

Three from-scratch databases, running the gate's exact chain (`mj migrate` → `mj sync push --dir=metadata --ci` → `mj codegen` → diff):

| Cycle | Result |
|---|---|
| baseline `next` | 76-line emitted migration + `.entity-permissions.json` |
| + relationships via apply-time `MAX+1` | emitted migration shrinks, but `MJUser` form **regresses** |
| + literal sequences + value list | forms clean, emitted migration 76 → 9 lines |
| + `RelatedEntityNameFieldMap` | emitted migration **0 bytes**, no generated artifact drifts |

## One authoring note worth reading

`Sequence` carries CodeGen's **literal**, not an apply-time `MAX(Sequence)+1`. I tried MAX+1 first, by analogy with the `EntityField` rule in `migrations/CLAUDE.md`, and it was wrong.

That rule exists because a *repeatable* script renumbers `EntityField` sequences and a literal collides on `UQ_EntityField_EntityID_Sequence`. `EntityRelationship` has no such constraint and no renumbering script — but its `Sequence` **orders the related-entity tabs in the generated forms**. On a clean database `MAX+1` assigns 110 where CodeGen emits 105 for `MJ: Users`, which relocates the Identity Claims tab and reintroduces exactly the drift this migration removes. Verified empirically both ways; the reasoning is in the migration header so nobody re-applies the analogy.

Guards are on the **natural key** (`EntityID` + `RelatedEntityID` + `RelatedEntityJoinField`, and `EntityFieldID` + `Value`), not the primary key — every developer database already holds these rows under locally-generated IDs, so an ID-keyed guard would double-insert there. That follows the pattern in #3367's sibling migration.

## Relationship to #3367

#3367 carries `V202609081830__v6.1.x__IdentityClaim_Metadata_Healing.sql`, which supplies the *other* part of the same gap: the four `IDX_AUTO_MJ_FKEY_*` indexes. The two overlap only on the value-list block, and both are natural-key guarded, so **merge order does not matter and neither can double-insert**. I verified the merge is clean and both migrations apply together (78 migrations, 384 entities).

If this merges first, #3367's healing migration reduces to just the four indexes.

**One thing #3367 should know before merging** — I found it while testing this combination and have raised it there. `next`'s committed `mjuser.form.component.html` contains the Identity Claims section; #3367's does not, removed in `50d0f05cd6` ("Codegen fixes from merge conflicts"). Same for the other three affected forms. That is a user-visible regression — four core forms lose a related-entity tab — and it landed because a CodeGen run against a database lacking these relationships regenerated the forms without those sections. This PR is the prerequisite for regenerating them correctly.

## Scope

No package source changes. Migration + one metadata file + the workflow. Changeset is `minor` per `.claude/rules/changesets.md` (the branch touches the database).


rkihm-BC:
@AN-BC — review requested. Two things make this worth a second pair of eyes rather than a rubber stamp.

**It touches a shared CI file, and I wrote it**, so I have not self-approved. The change is 3 files, +22/−1, CI only — no package source, no migration, no metadata.

**A self-review already caught one fragility in it**, which is the main reason I want you on it. The first version swept CodeGen's scratch file with `rm -f migrations/*/CodeGen_Run_*.sql`. That deletes files inside a repo checkout, in a gate whose whole job is detecting file changes — and there *are* two committed `CodeGen_Run` migrations under `migrations/v2`. It happened to be safe, because those carry Flyway's `V<timestamp>__` prefix and the glob requires the name to start with `CodeGen_Run_` (verified). But relying on a naming convention to avoid deleting tracked files is the wrong shape. `c5e8f5110f` restricts it to files git reports as untracked. If you see a better way to express that, I would take it.

## What it fixes

1. **The gate failed on its own scratch output.** The diff step's comment promises to ignore CodeGen's run file; the code only removed `codegen.output.log`. CodeGen writes a `CodeGen_Run_<timestamp>.sql` every run — 0 bytes when it has nothing to emit — so an untracked empty file failed the gate at zero real drift. Both copies of the step are patched: the blocking one and the nightly AI-assisted variant.

2. **`getGitStatusPorcelain` mangled the first changed file's path.** It returned `stdout.trim().split('\n')`; porcelain lines are `' M path'` and every caller strips 3 characters. Trimming the whole buffer removes the leading space of the *first* line only, so that path loses its first character. That is the failure on `next` — `Unexpected files changed … ackages/Angular/…`, missing the `p`. Only ever the first line, which is why it looked intermittent. Introduced by `197fdf8376`.

## Two caveats on the evidence

**This PR's own drift gate is expected to fail.** It runs the pre-merge workflow, so it exercises the broken code it fixes. The fix can only be validated once it is on `next`. Please judge it on the diff.

**Neither fix has a test.** Both are one-liners in CI plumbing, but `.github/scripts` has testing precedent (`check:changeset:test`), and a unit test on `getGitStatusPorcelain` would have caught the original bug outright. I left it out to keep this minimal and unblocking; say the word and I will add one.

## Known gap, deliberately out of scope

The porcelain parser still cannot handle renames — `R  old -> new` breaks `substring(3)` regardless of this change. Pre-existing. Flagging it because the function now carries a comment explaining its parsing, which makes the remaining gap look considered rather than unnoticed. Happy to file it separately.

## Why it matters for sequencing

Nothing can go green until this lands — every branch hits both defects regardless of its own content. #3367 is now fully regenerated and clean apart from these two, and #4321 is approved and waiting. Related: #4323 (the investigation) and #4329 (the `EntityRelationship` guard follow-up on your migration).

AN-BC (approved):

PR #4329, opened 2026-09-09
fix(migrations): de-duplicate the IdentityClaim metadata rows V202609091200 can insert twice

rkihm-BC:
@AN-BC — review requested; this is a follow-up to your `V202609091200__v6.1.x__IdentityClaim_Metadata_Tail.sql` from #4324.

## The problem

Your migration guards its four `EntityRelationship` inserts on **primary key**:

```sql
IF NOT EXISTS (SELECT 1 FROM [${flyway:defaultSchema}].[EntityRelationship] WHERE [ID] = 'dd73e538-5890-4ab3-a0b3-4fce34002a9c')
```

Those ids come from the CodeGen run that authored the migration. But the premise of the gap you were closing is that **these rows already exist on every developer database** — created locally by `mj codegen`, under locally generated ids. On such a database the guard looks for an id that is not there, does not fire, and inserts a second row for a relationship that already exists.

There is no unique index on `(EntityID, RelatedEntityID, RelatedEntityJoinField)`, so nothing catches it. The visible symptom is a **doubled related-entity tab** on the four affected forms — `MJ: Entities`, `MJ: Users`, `MJ: Identity Claim Types`, `MJ: Magic Link Invites`.

## Verified, not theorised

Against a database holding those rows under CodeGen-local ids (`AAAC1D32-…` rather than your `dd73e538-…`):

```
guard fires?                                  NO — it would INSERT a duplicate
unique indexes on the natural key:            0
after simulating your insert:                 2 rows for the EntityID join
after this migration:                         1 row
surviving id:                                 AAAC1D32-…  (the pre-existing local row)
total IdentityClaim relationships:            4
```

## The fix

Forward-only cleanup keyed on the natural key, keeping the earliest row by `__mj_CreatedAt`. A fresh database has nothing to remove, so it is a no-op in CI.

**Deliberately not a fix in place.** `V202609091200` is already merged and applied, so editing it would change its Flyway checksum and break every database that has run it.

## Worth deciding separately

The underlying hazard is that a **PK-keyed guard is the wrong shape for healing migrations generally** — any migration supplying CodeGen artifacts that already exist locally has this property. The precedent in #3367's sibling migration guards on the natural key (`EntityFieldID` + `Value`) and is safe for the same reason. Might be worth a note in `migrations/CLAUDE.md`, since this class of migration is going to recur.

A unique index on `(EntityID, RelatedEntityID, RelatedEntityJoinField)` would also have turned this from a silent duplicate into a loud failure. I have not added one here — that is a schema decision with wider blast radius than this fix, and it would need this cleanup to run first anyway.

## Context

Part of the #4323 cleanup. The other half, #4325, fixes two CI defects in the drift gate itself (its scratch file, and a path-mangling bug in the idempotency check).


AN-BC:
@rkihm-BC pretty sure my latest PR to MJ addresses this or similar

rkihm-BC:
**Raising severity — and this PR was incomplete. Both are now addressed.**

## It is not developer-only, and it is not hypothetical

I filed this as a risk to developer databases. It is worse: **it reproduces on a clean CI database**, and it has been actively blocking #3367.

Whenever the rows already exist under any other ids — from any source, not just a local `mj codegen` — the primary-key guard does not fire and inserts a second set. On #3367 that happens because its own healing migration supplies the same four Status values under different ids and sorts earlier. The CodeGen drift gate caught it as a generated Zod union carrying every value twice:

```
Status: z.union([z.literal('Claimed'), z.literal('Claimed'),
                 z.literal('Expired'), z.literal('Expired'),
                 z.literal('Pending'), z.literal('Pending'),
                 z.literal('Revoked'), z.literal('Revoked')])
```

So the blast radius is any branch or database where those rows pre-exist under different ids — which, given the whole premise is that CodeGen creates them locally, is most of them.

## This PR only covered half the defect

`V202609091200` guards on primary key in **two** places, not one: the four `EntityRelationship` inserts *and* the four `EntityFieldValue` inserts. My first version deduped only the relationships.

Extended in `7d0265a98f` to also dedupe the Status value list, keyed on the natural key (`EntityFieldID` + `Value`), keeping the earliest row. Verified the same way as the first: a duplicate insert yields 2 rows for `'Claimed'`, the migration restores 1, and the total Status values return to 4.

## Scope check

I looked for other primary-key-guarded inserts in that migration and found none beyond these two tables. If you know of others in the same family, say so and I will widen it.

## Note on #3367

That branch is fixed at the source rather than by this PR — I removed the duplicated value-list block from its migration, since `next`'s copy carries the `EntityRelationship` rows and `RelatedEntityNameFieldMap` that nothing else supplies, while #3367's carries the four FK indexes that `next`'s lacks. The two are now disjoint. This PR remains necessary for databases that have already applied both.

@AN-BC — the guard-shape question from the original description is now more pressing than I first suggested.

rkihm-BC:
## Adversarial self-review

I tried to break this rather than confirm it. Recording what I attacked and what survived, since I authored it and nobody else has looked yet.

### Attacks that failed

| Attack | Result |
|---|---|
| FK references to the deleted rows would fail or orphan | `sys.foreign_keys` shows **nothing** references `EntityRelationship` or `EntityFieldValue`. Safe. |
| A `GO` between the two `DECLARE`s would break the second half | My file has **zero** `GO`s — one batch, both in scope. (A live hazard: `V202609091200` has 12.) |
| The halves were only tested via extracted `sed` fragments | Whole file executes clean, and is **idempotent** — second run is `(0 rows affected)` twice. |
| Something else in `V202609091200` can duplicate | Full INSERT inventory: `EntityFieldValue` ×4, `EntityRelationship` ×4. Only two tables; both covered. |
| The value dedupe is scoped to the wrong field id | All four inserts target `F925BD99…`, matching `@StatusFieldID`. |
| Other migrations carry the same guard pattern | `V202608202300` has 28 — but it contains `CREATE TABLE`, so it *introduces* the entity and nothing can pre-exist under other ids. Different precondition, not a gap. |
| It works in `sqlcmd` but not under Flyway (placeholders, batching) | Real `mj migrate` on a clean database: **77 applied**, 4 relationships, 4 status values, **0 duplicates**. |

### One weakness I claimed away and have now actually checked

I argued that "keep earliest" is safe because CodeGen is deterministic — but I had only verified the natural keys matched, not the *field values*. If a local row carried a different `Sequence` or `DisplayInForm`, the dedupe would leave a state neither migration intended.

Compared a database whose rows trace to a local CodeGen run against a fresh migration-only database:

```
ClaimedByUserID  seq=105 bundle=1 display=1 type=One To Many
ClaimTypeID      seq=1   bundle=1 display=1 type=One To Many
EntityID         seq=78  bundle=1 display=1 type=One To Many
MagicLinkInviteID seq=6  bundle=1 display=1 type=One To Many
Claimed/Expired/Pending/Revoked  seq=1..4, code == value
```

Identical on both. Independently corroborated by CodeGen's own emitted SQL, which carries the same literals (78/105/1/6). Whichever row survives, the resulting state is the same.

### Weaknesses that remain, stated plainly

**This is cleanup, not prevention.** The primary-key-guard shape can be copied into the next healing migration and reproduce this exactly. A unique index on `(EntityID, RelatedEntityID, RelatedEntityJoinField)` would turn silent duplication into a loud failure. I left it out because it is a schema change with wider blast radius and would need this cleanup to run first — but that means the defect *class* stays open, and #4330 is an audit rather than a guard. If you want the index, it is a natural follow-up and I would rather it be a deliberate decision than an omission.

**No automated test.** Verified by hand against live databases; nothing in the integration tier would catch a regression here.

**Scope is IdentityClaim-only.** Deliberately surgical, but the title says "IdentityClaim EntityRelationship rows" while the body discusses the general guard problem — a reader could over-read what it fixes. The second commit widened it to `EntityFieldValue`; the title has not caught up.

### Verdict

Correct and complete for the defect it targets. The remaining items are scope decisions, not errors — but @AN-BC, the unique-index question is the one worth your call, since it is the difference between fixing this instance and closing the class.

PR #4333, opened 2026-09-09
fix(cerebras): narrow to the non-streaming member of the SDK's union

JosueGarcia-BC:
Lifted out of #4217, where it did not belong, and its justification corrected.

## The change

`ChatCompletion` is exported as a union of the non-streaming response, a chunk response and an error chunk. Only the first carries `choices`, `usage` and `model`, so holding the bare union makes those three reads fail to compile (TS2339).

This method is `nonStreamingChatCompletion` and never sets `stream`, so the non-streaming member is correct by construction rather than by assumption: the other two members arrive only on a stream, and a transport failure throws into the existing catch rather than returning an error chunk here. The cast is erased at runtime.

## Correcting the claim I originally made for it

The comment on the original commit said these errors "kept this package from building at all". That is wrong, and the review on #4217 was right to check it rather than take it.

| | |
|---|---|
| lockfile resolves | **1.64.1** — `ChatCompletion` is not a union there |
| `ai-cerebras` on clean `next` | **builds, 0 TS errors** — verified from a clean checkout |
| declared range | `^1.64.1`, which admits **1.91.0** |
| what I had installed | **1.91.0** — where it *is* a union |

So it never broke CI, and CI has been green without it all along. What I had locally was a newer SDK than the lockfile pins, inside the range the package already declares.

## Why it is still worth landing

It is a **latent** break rather than a current one:

- it bites anyone who installs fresh outside the lockfile, today
- it becomes CI's problem the moment the lockfile is refreshed — and it would arrive then attached to an unrelated dependency bump, which is a considerably worse place to meet it

Fixing it now costs nothing and removes that.

## One consequential cast removed, one deliberately kept

Narrowing the declaration makes the adjacent `chatResponse.choices as Array<ChatCompletion.ChatCompletionResponse.Choice>` a no-op, so it is gone — `choices` is a required field on the narrowed type and needs no assertion.

The `usage` cast a few lines below **stays**, and the reason is worth stating because the symmetry is tempting. `usage` is declared `Usage | null | undefined` on that same union member, so that cast is not redundant — it asserts the value is present, which the reads below depend on. This package compiles without `strictNullChecks`, so removing it would look like the same tidy-up and would compile identically, while quietly converting a stated assumption into one that only holds because strict is off. It now carries a comment saying so.

## Not included

The same line retains a pre-existing `(choice: any)` in the `.map` callback. Raised in the #4217 review as worth sweeping whenever the file is next touched deliberately — left alone so this PR stays scoped to the union narrowing and its direct consequence, but happy to take it if you would rather it went now.

`tsc` clean. `check:changeset` green (`patch`; no migration, no metadata).


rkihm-BC (changes requested):
Thank you for splitting this out and for correcting the original claim. **The code is right and I want it landed.** But I checked the new justification the same way I checked the old one, and it is also wrong — this time in the comment, which will live in the tree.

## The code is correct

Verified rather than assumed:

- **The cast is safe by construction.** `nonStreamingChatCompletion` is at line 177, streaming lives in a separate `createStreamingRequest` at 329, and nothing in this method sets `stream`. The chunk and error members genuinely cannot arrive here.
- **Narrowing at the declaration beats casting at each read.** That is the right shape.
- **Removing the `choices` cast is correct** — it is a required field on the narrowed member.
- **`strictNullChecks` really is off.** Neither `tsconfig.server.json` nor anything it extends sets `strict` or `strictNullChecks`.

## Three factual problems, all in the block comment

I read the installed `@cerebras/cerebras_cloud_sdk@1.64.1` type definitions directly.

**1. "the lockfile resolves 1.64.1, where `ChatCompletion` is not a union" — it is a union there.**

`resources/chat/completions.d.ts:19`, in the 1.64.1 the lockfile pins:

```ts
export type ChatCompletion = ChatCompletion.ChatCompletionResponse | ChatCompletion.ChatChunkResponse | ChatCompletion.ErrorChunkResponse;
```

It has been a union at the pinned version all along.

**2. So the "latent break" framing does not hold.** The reason the original compiled is not that the type was different — it is that **both union-unsafe reads were already individually cast**. `choices` had `as Array<...Choice>`, `usage` had `as ...Usage`. This PR is not fixing a break that arrives with a future lockfile refresh; it is replacing per-read casts with one declaration-level narrowing. That is a *better* change than the one described, and it needs none of the version argument.

**3. "three reads fail to compile" — two do.** `model: string` is declared on all three union members (lines 25, 179, 362), which is exactly why line 296 never carried a cast.

## The retained `usage` cast is kept for a reason that is not true

This is the one I would not merge as written. The new comment says:

> `usage` is declared `Usage | null | undefined` on the union member

On `ChatCompletionResponse` — the member you narrow **to** — it is `usage: ChatCompletionResponse.Usage`, required and non-nullable (line 29). The nullable declaration is `usage?: ChatChunkResponse.Usage | null` at line 185, on `ChatChunkResponse` — the member you narrowed **away** from.

So on the pinned SDK that cast is exactly as redundant as the `choices` cast you removed, and the asymmetry the comment explains does not exist. Keeping it is defensible if a newer SDK makes it nullable on the response member too — but then the comment should say that, and say it is unverified, rather than assert a shape the pinned version contradicts.

## What I would like

Narrowly the comment, not the code:

- Drop the version table, or correct it to "union at 1.64.1 and later; the original handled it with per-read casts".
- Say "two reads", or just name `choices` and `usage`.
- Either remove the `usage` cast as redundant on the pinned SDK, or keep it with an honest reason.

## One stale line

`// Cast to any to extract the choices` still sits above the line whose cast you removed. It now refers only to the `(choice: any)` in the callback, which is not "extracting the choices". Worth rewording while you are in there — and I remain happy for the `(choice: any)` sweep to stay out of scope.

## The pattern, said once

This is the second justification for the same three-line change that did not survive checking — first "the errors that have kept this package from building at all", now the version table. Both times the code was fine. The cost is that a reviewer has to verify every claim rather than any, and the wrong reason ends up committed as documentation. The change is worth landing on its actual merit: a union has always been there, per-read casts were handling it, and one narrowing is cleaner.

rkihm-BC (approved):
**Downgrading to an approval, and retracting one of my four findings.** I re-checked all of them; one was wrong, and none of the survivors is worth blocking a correct fix over.

## Retracted — "three reads" was right, I was wrong

I claimed only two reads break, on the grounds that `model: string` is declared on all three union members. I missed that every member also carries `[k: string]: unknown`, and that **`ErrorChunkResponse` declares only `error`, `status_code` and that index signature** — no `model`, no `choices`, no `usage`.

So on the bare union every read collapses to `unknown`:

| read | `ChatCompletionResponse` | `ChatChunkResponse` | `ErrorChunkResponse` | union |
|---|---|---|---|---|
| `choices` | `Array<Choice>` | `Array<Choice> \| null` | index sig | **`unknown`** |
| `usage` | `Usage` | `Usage \| null` | index sig | **`unknown`** |
| `model` | `string` | `string` | index sig | **`unknown`** |

Your "three reads" is correct. Sorry — I read the `model:` declarations without checking whether the third member had one.

(Minor and not worth changing: the failures are `unknown`-assignability rather than `TS2339`, since the index signature makes the property access itself legal.)

## Still stands, but as comments to fix rather than a blocker

**The version table is wrong.** `completions.d.ts:19` in the 1.64.1 the lockfile pins:

```ts
export type ChatCompletion = ChatCompletion.ChatCompletionResponse | ChatCompletion.ChatChunkResponse | ChatCompletion.ErrorChunkResponse;
```

I confirmed the resolution too — `packages/AI/Providers/Cerebras/node_modules/@cerebras/cerebras_cloud_sdk` symlinks to `@cerebras+cerebras_cloud_sdk@1.64.1`, whose own `package.json` reads `1.64.1`. It is a union at the pinned version, so the "latent break, arrives with a future lockfile refresh" framing is not what is happening. What is happening: it has always been a union, the original code coped by casting each read from `unknown`, and you have replaced three casts with one narrowing. That is a better change than the one described and needs no version argument.

**The `usage` cast justification does not hold on 1.64.1.** The comment says `usage` is `Usage | null | undefined` "on the union member". On `ChatCompletionResponse` — the member you narrow to — it is `usage: ChatCompletionResponse.Usage`, required and non-nullable. The nullable shape is on `ChatChunkResponse`, the member you narrowed away from. Explicit members beat the index signature, so after narrowing that cast is a no-op, exactly like the `choices` cast you removed.

Keep it if a newer SDK makes it nullable on the response member — but then say that, and say it is unverified against the pinned version.

## Why this is an approval

The code is right, and I verified the part that actually matters: **the cast is safe by construction.** `nonStreamingChatCompletion` is at line 177, streaming lives in `createStreamingRequest` at 329, and nothing in this method sets `stream`. The chunk and error members cannot arrive here.

Everything left is comment text. Correct it if you are pushing anyway — the version table and the `usage` reason are both wrong in ways that will mislead whoever reads this file next — but I am not holding a correct fix for prose, and I should not have set it to changes-requested for that.

The stale `// Cast to any to extract the choices` line is still worth a reword while you are in there, and the `(choice: any)` sweep still stays out of scope.

PR #4334, opened 2026-09-09
fix(communication-ms-graph): Reply resolves its mailbox from the request too

JosueGarcia-BC:
Follow-up from the review on #4276, which approved the mailbox rework and then found that two of the fifteen call sites do not benefit from it.

## The gap

Thirteen operations pass a request-level candidate to `resolveMailbox`. Reply passed none:

```ts
const mailbox = this.resolveMailbox('ReplyToMessage', creds);
```

So it could only resolve `creds.accountEmail` — a property the `Azure Service Principal` credential type declares nowhere, and whose environment source is exactly what `disableEnvironmentFallback: true` removes.

**Reply stayed unreachable on the stored credential #4276 was written to support.** Not a regression — it threw before too — but the framing was that a stored principal can now drive operations, and that was true for thirteen and still false for this one.

No test saw it, because the suite only covered operations that already named their own mailbox.

## The fix

`ReplyToMessageParams` carries `ContextData`, the same field nine sibling operations already read as `params.ContextData?.Email` (a tenth, `GetMessages`, reads the same value through a local). Reply now reads it first and falls back to the credential default like everything else.

Two tests: resolving from `ContextData`, and refusing by name when nothing resolves anywhere. The refusal **returns** rather than throwing — Reply has a try/catch, and its message carries the reason — so the test pins the real contract rather than the shape I first assumed and had to correct.

Mutation-checked: removing the candidate fails three tests.

## Forward is deliberately untouched

`ForwardMessageParams` has no `ContextData` and no identifier, so it cannot name a mailbox without a change to that params type. That is the one operation a conformant three-field credential still cannot drive with the environment fallback off.

As the review noted, this sharpens the open question about whether the schema should carry an optional `accountEmail`: Forward is the case that decision actually settles. It deserves that decision rather than riding along here.

`tsc` clean, 72 tests.


rkihm-BC (approved):
Approving. I verified every load-bearing claim in the description against the code, and they all hold.

## Verified

- **`ReplyToMessageParams` does carry `ContextData?: T`** (`BaseProvider.ts:438`), so reading it needs no type change. That is what makes this a two-line fix rather than a params-type change.
- **`ForwardMessageParams` genuinely lacks it** — and lacks any other mailbox identifier: `MessageID`, `Message?`, `ToRecipients`, `CC`, `BCC`, and nothing else. So deferring Forward is a real constraint, not a convenient one. Good.
- **The call-site arithmetic is right.** After this change, 14 of the 15 `resolveMailbox` sites pass a request-level candidate and `ForwardMessage` is the only holdout. That matches the description exactly.
- **The optional chain is safe.** `resolveMailbox(operation, creds, ...preferred: (string | undefined)[])` is variadic and skips empty candidates, so `params.ContextData?.Email` resolving to `undefined` falls through to the credential default rather than short-circuiting.
- **The refusal contract is as described.** `resolveMailbox` throws when nothing resolves; Reply's `catch` returns an error result. So "refuses by name, and returns rather than throws" is accurate at the operation boundary — and pinning that in the test rather than the shape you first assumed is the right correction to have made.

Mutation-checking it (three tests fail when the candidate is removed) is what makes the two tests worth having.

The framing is also honest in a way that is easy to get wrong: this is **not** a regression, Reply threw before too. What changed with #4276 was the claim that a stored principal can drive operations — true for thirteen, false for this one. Saying that plainly is better than dressing it as a bug fix.

## Two nits, neither blocking

**The comment overcounts.** It says "eleven sibling operations already read it". The body says nine plus a tenth through a local, and I count **nine** occurrences of `params.ContextData?.Email` on `next`. Small, but it is a countable claim sitting in a comment where the next reader has no reason to doubt it.

**`as string` does nothing.** `ContextData` is `Record<string, any>` by default, so `.Email` is already `any` — the assertion neither narrows nor checks anything, and it claims `string` for a value that is legitimately `undefined`. Harmless because the variadic signature accepts `undefined`, and consistent with the sibling sites, so I would not change it here — but it is a cast that documents nothing, and if the sibling pattern is ever swept, this is part of it.

## Nothing missing

Forward is correctly left alone with the reason stated, and you are right that it is the case that actually settles the optional-`accountEmail` schema question. That decision deserves its own PR rather than being pre-empted by a two-line change here.

PR #4342, opened 2026-09-10
fix(conversations): no scroll yank on progress, opt-in read-from-top landing, one branded completion toast

cmbrock-BC:
Two things a chat reader feels on every agent run, both found in a white-label host built on `mj-conversation-chat-area`.

## The gap

**Every progress update pulled the reader to the bottom.** `message-input` re-emits `messageSent` for the in-progress reply on each streaming progress callback, and `onMessageSent` set `scrollToBottom = true` after its update-in-place branch — so every re-emit ran the send path's scroll. A reader who scrolled up to reread history was pulled back once a second for the whole run. Nothing asked whether the reader was still at the bottom, although `checkScroll()` already computes it for the scroll-to-bottom button. (The message list's own scroll-on-append targets its non-scrolling container, so the chat area's snap is the only one that moves anything.)

**A run ended at the end of the answer.** Following the tail means the reader finishes looking at the last line of a long reply and scrolls back up to start reading it.

**Completion was announced twice, in plain toasts.** The server raises an Agent Completion notification when a run produced an artifact and pushes it; `message-input.markMessageComplete` raises its own. Each has a "not while the user is viewing" rule keyed to the Explorer chat shell (`ConversationBridge` + overlay / workspace / `/chat` route), which a host that mounts the chat area on its own route never satisfies — so both showed, 50 ms apart, as generic status toasts with no sign of which agent.

## The fix

**One decision point for the viewport.** The four `scrollToBottom = true` sites go through `followTranscript(change)`: `load` → bottom; `new` → bottom, as before; `update` → only for a reader already at the bottom (`readerAtBottom`, tracked in `checkScroll`). A reader at the bottom sees exactly what they saw before; a reader who scrolled away is left alone. This half is on by default — it is the behaviour the list's own comment already asks for ("only scroll if new messages were added, not just updates").

**Opt-in `readReplyFromTop`** (default `false`) on the chat area. Sending and the run behave as today. When the reply lands and the reader was following, the turn — the reader's message plus the reply — is scrolled so it starts at the top of the pane, if it is taller than the pane; a turn that fits stays where it is, all on screen. The turn is opened by the reader's message whether it arrives as `new` or (auto-send path, where the chat area already holds it) as `update`; it lands once, because completion is emitted more than once, and follow timers armed by the last progress updates are disarmed while the landing runs. The list's sticky date header is measured into the landing gap since it pins to this scroller. A `load` no longer forgets the turn in flight: creating a conversation from the composer sends the first message while the initial load is still finishing.

**One branded completion toast.** `MJNotificationService.CreateRichNotification(options)` renders an image-or-icon, title and detail on the surface tokens (`--mj-bg-surface-card`, `--mj-border-default`, `--mj-text-*`) with `--mj-brand-primary` as the accent, so a brand ramp themes it. `dedupeKey` collapses same-key toasts: one already on screen is kept (a repeat only extends it), one held back by `deferMs` is superseded by the later call and shown at once. The server's push path uses the rich toast with `deferMs: 1500`; `markMessageComplete` uses it with the agent's `LogoURL`/`IconClass` and the conversation's *current* name from `ConversationEngine` (the task carried the send-time name — `New Conversation` on a first message — which is now treated as no name). Since the server's notification always arrives first and the client's follows within tens of ms, the reader sees one toast in the client's wording; the server wording appears only when the client cannot announce (page reloaded mid-run). A new `CompletionImageUrlResolver` host hook lets a white-label host put its assistant's avatar on the toast, winning over the agent's own logo.

## What does not change

Default mode still follows the tail on load and on every new message. The existing "not while viewing" suppression in both paths is untouched. No CSS files change; the toast is drawn with tokens only. The new surfaces are additive: `readReplyFromTop`, `CreateRichNotification`, `CompletionImageUrlResolver`, `MessageListComponent.FindTimelineElement`, and the two exported option types.

## One deliberate trade-off

The dedupe key is per conversation (`agent-completion:<conversationId>`), because that is all the server's notification payload carries — a per-run key would defeat the folding the feature exists for. So two **different** agents finishing in the same conversation inside the window (3 s, down from 15 in review) produce one toast naming the first. Previously both appeared. Seconds apart they are announced separately as before.

## Testing

- `packages/Angular/Generic/conversations`: 107 files / 1287 tests pass, including `transcript-follow.test.ts` (24): the `followTranscript` table in both modes, turn open/re-land/no-fall-through, the refinement path (settled status line before the reply), the load race, the suppression lift on a new message, and the landing measurement with per-selector stubs — top when the turn is taller than the pane, bottom when it fits, the pinned sticky header cleared by height + CSS inset (a fixture 5000 px down; reintroducing `offsetTop` fails two tests), and the list-resolved lookup.
- `packages/Angular/Generic/notifications`: 3 files / 20 tests pass, including `rich-notification.test.ts` (16, jsdom per-file): render + tokens, icon chip tint, close button, live-region order, resolver precedence, live-toast-is-kept, outside-window is new, same-key-during-slide-out renders fresh, double-defer re-defers, deferred-shows-alone, registry release, hover-pause, single keyframes writer, escaping. Mutation-checked: rewriting a live toast and inverting the resolver precedence each fail their test.
- Both packages `pnpm run build` clean on `next`; `npm run check:ui` 0 violations.
- Live, in the host that surfaced this, with a scroll spy recording every programmatic `scrollTop` write: default mode wrote the bottom every ~1 s for a whole run; with the guard an in-place update while scrolled up arms nothing; `readReplyFromTop` on the auto-send path opened the turn from the `update`, landed once, and the second completion emit no longer snapped. Toast log across a run: one element, one wording.
- Integration tier not run: nothing here is server-side, and that tier stands in for the Angular layer.

Changeset: `@memberjunction/ng-conversations` patch, `@memberjunction/ng-notifications` patch.


packages/Angular/Generic/conversations/src/lib/components/conversation/conversation-chat-area.component.ts
```diff
@@ -4470,6 +4518,113 @@ export class ConversationChatAreaComponent extends BaseAngularComponent implemen
...
+  private followTranscript(change: 'load' | 'new' | 'update', message?: MJConversationDetailEntity): void {
+    if (change === 'load') {
+      this.scrollToBottom = true;
+      return;
+    }
+    if (this.readReplyFromTop) {
+      if (message?.Role === 'User' && message.ID && message.ID !== this.currentTurnStartMessageId) {
+        this.currentTurnStartMessageId = message.ID;
+        this.currentTurnLanded = false;
+      } else if (this.currentTurnStartMessageId && message?.Role === 'AI' && this.isSettled(message)) {
```
rkihm-BC, line 4550: **`readReplyFromTop` does nothing on the iterative-refinement path — the one its own purpose most cares about.**

Non-blocking: the flag defaults to `false` and nothing in the repo sets it, so no deployment is affected. Flagging it before a host turns it on.

This lands the turn on the **first settled AI message**. On the primary send path that is the reply, because the response message is created `In-Progress` and completes later. `handleSilentObservation` — the path taken when Sage stays silent and the previous agent continues for refinement, called at `message-input.component.ts:1907` — emits a short status message first:

```ts
statusMessage.Role = 'AI';
statusMessage.Message = `Continuing with **${agentName}** for refinement...`;
statusMessage.Status = 'Complete';
await statusMessage.Save();
this.messageSent.emit(statusMessage);          // ~:2386
```

It satisfies `isSettled`, so it lands the turn. At that instant the turn is the user's line plus one status line, which fits the pane, so `scrollTurnToTop` takes the `else` branch and calls `scrollToBottomNow()` instead of landing at the top. `currentTurnLanded` is now `true`, and the real reply — created already `Complete` at `:2433` — hits this same `else if` and returns early. No top landing, and no bottom follow either.

Net effect on a refinement turn: the reply appears below the fold and the reader is not moved at all.

The 19 new tests do not catch it because the fixtures go `In-Progress` then `Complete`, which is the primary path only.

Shape of a fix, if you want one: land on the settled AI message that is the turn's **last** message rather than its first, or ignore a settled AI message that arrives while the turn has no `In-Progress` reply yet. Either way a fixture with a settled AI status message ahead of the reply would pin it.

packages/Angular/Generic/notifications/src/lib/notifications.service.ts
```diff
@@ -275,6 +344,125 @@ export class MJNotificationService {
...
+        if ((e.target as HTMLElement).closest('button')) return;
+        options.onClick?.();
+        this.dismissToast(toast);
+      });
+    }
+    container.appendChild(toast);
+
+    const timer = options.hideAfter ? setTimeout(() => this.dismissToast(toast), options.hideAfter) : null;
+    if (key) {
+      this.liveRichToasts.set(key, { element: toast, timer, shownAt: Date.now() });
```
rkihm-BC: Nit: entries go into `liveRichToasts` here and are never removed. `dismissToast` takes the element out of the DOM but leaves the map holding a reference, so the service retains one detached toast element per dedupe key for the life of the session.

Bounded by conversations visited and each element is small, so this is housekeeping rather than a problem. The natural place to release it is `dismissToast`, or when a live entry is found stale in `CreateRichNotification`.

packages/Angular/Generic/notifications/src/lib/notifications.service.ts
```diff
@@ -275,6 +344,125 @@ export class MJNotificationService {
...
+    }
+  }
+
+  /** Host resolver first (it knows how the assistant is branded), then the caller's image. */
+  private resolveRichImage(options: MJRichNotificationOptions): string | null {
+    const fromHost = options.context ? this.CompletionImageUrlResolver?.(options.context) : null;
+    return fromHost || options.imageUrl || null;
+  }
+
+  /** (Re)writes a rich toast's content — first paint and de-duplicated updates alike. */
```
rkihm-BC: Nit: this says the method writes content for "first paint and de-duplicated updates alike," but the dedupe path never re-fills — when a live toast is found, `CreateRichNotification` only clears and re-arms the hide timer and returns.

That is the right behavior, and `MJRichNotificationOptions.dedupeKey` documents exactly why ("the wording never changes under the reader's eyes"). It is only this comment that describes a second call site that does not exist.

rkihm-BC (dismissed):
## Approve — requiring nothing

Careful, well-scoped change with the reasoning left in the code. I verified the parts that ship live rather than taking the description's word for them, and found one defect in the opt-in half that cannot fire today. Nothing here needs to change before merge.

**The premise checks out.** `messageSent.emit` really does fire from the streaming and progress handlers at `message-input.component.ts:956`, `:978` and `:1719`, so the per-second yank is real and the new `update` branch is not dead code.

**The default-mode change is the only thing every user gets, so I went looking for a regression there.** I did not find one. `readerAtBottom` is computed in `checkScroll` from `scrollDifference <= 15` — the identical arithmetic already driving `showScrollToBottomIcon`. If that button hides correctly today, the new flag is correct by the same computation, with no new tolerance and no new measurement to get wrong. It refreshes on every real scroll event and after every programmatic scroll. `load` and `new` stay unconditional, so only `update` changes, which is the stated fix.

**The toast builds HTML by string concatenation from agent- and conversation-supplied values, so I checked injection specifically.** `escapeHtml` uses the `textContent` round-trip, which is sound. `escapeAttribute` escapes the double quote and the angle brackets, and both interpolations sit inside double-quoted attributes, so an agent `LogoURL` or `IconClass` cannot break out; a `javascript:` URL in an `img src` is inert. All eleven design tokens resolve in the canonical `_tokens.scss` — worth stating because a first pass matched only `mockups/`, and a token defined solely there would render as nothing.

**Both dedupe orderings are correct.** Server-defers then client-supersedes, and client-first then server-extends the timer.

Changeset covers both packages at `patch`, correct since there is no migration.

---

### One finding, non-blocking

`readReplyFromTop` silently does nothing on the iterative-refinement path — detail inline at `followTranscript`. It ships inert (nothing in the repo sets the flag, and `CompletionImageUrlResolver` has no in-repo consumer either), so this is on the record for whoever first turns it on rather than something to fix now.

### Two nits

Both inline on `notifications.service.ts`: `liveRichToasts` never releases its entries, and `fillRichToast`'s doc comment describes a re-fill the design deliberately avoids.

### One question, not a request

The dedupe key is `agent-completion:<conversationId>` with a 15 second window, so two **different** agents completing in the same conversation inside that window yield one toast naming only the first. Previously both appeared. That reads as a forced trade-off — the server's notification carries only a conversation id, and a per-agent key would break the folding the feature exists for — but it is a live behavior change that is not in the "What does not change" list. Worth confirming it was intended.

MattC-BC (changes requested):
Nice writeup, and the framing is right: the per-progress yank is a real problem and one decision point for the viewport is the correct shape for the fix. The default-on half I'd take as-is. `readReplyFromTop` has a measurement bug that makes it do the opposite of what it promises, and there are a couple of things in the toast worth cleaning up first.

## Blocker: the landing scrolls to the top of the conversation, not the turn

`conversation-chat-area.component.ts:4606`

```ts
const stickyHeight = stickyHeader && stickyHeader.offsetParent ? stickyHeader.offsetHeight + stickyHeader.offsetTop : 0;
```

`offsetTop` on a pinned `position: sticky` element isn't a fixed layout inset. Browsers report the used position, so once the header sticks it tracks the scroll offset. Scroll down 5000px and `offsetTop` is ~5000.

Each link in that chain holds here:

- `.sticky-date-header` is `position: sticky; top: 12px`
- In the chat area the list is host-scrolled (`:host(.mj-list-host-scrolled) .message-list-container { overflow: visible }`), so sticky binds to the chat area's scroller and genuinely pins
- `.message-list-container` keeps `position: relative`, so it is the header's `offsetParent`. The `stickyHeader.offsetParent` guard passes rather than short-circuiting

So `turnTopClearance()` returns roughly `16 + headerHeight + scrollTop`, which poisons both lines under it:

```ts
const turnTop = this.offsetWithinScroller(target) - <that>;   // goes negative
const turnHeight = container.scrollHeight - turnTop;          // inflated by the same amount
```

`container.scroll({top: <negative>})` clamps to 0 and drops the reader at the top of the whole conversation. The inflated `turnHeight` also makes the "does the turn fit?" guard always answer "taller than the pane", so the else-branch never runs and this fires on every landing.

The trigger is the unlucky part: the date header only renders for conversations with 20+ messages spanning more than 2 days, which is exactly the long-history case the feature is for. Short conversations behave correctly, which I'd guess is why the live pass didn't surface it.

Fix is to drop `offsetTop` and use `offsetHeight` plus the CSS inset.

Worth noting the new test can't catch this. `seedPane`'s fake `querySelector` returns the same stub for every selector, and that stub has no `offsetParent`, so `turnTopClearance` returns exactly `16` under test. The `{top: 898}` assertion is encoding the stub rather than the DOM. Same reason the `[data-message-id]` and `.sticky-date-header` selectors are never actually validated by the suite.

## Should fix before merge

**1. The fallback avatar icon is close to invisible.** `notifications.service.ts:427`

The chip pairs `--mj-brand-primary-light` as background with `--mj-brand-primary` as the icon color. `--mj-brand-primary-light` is `brand-350` (`#3395c8`), a saturated hover blue rather than a tint, so against `#0076b6` that computes to 1.47:1 in light mode. Dark mode is worse: `--mj-brand-primary` shifts to `brand-400` (`#2699cc`) while `--mj-brand-primary-light` has no dark override, landing at 1.04:1. At that point it's a flat blue circle with no discernible icon.

To be fair about what this is and isn't: the icon is `aria-hidden="true"` and the agent name is already in the title, so it's decorative and I don't think WCAG 1.4.11 actually bites here. I'm raising it as a visual defect rather than an a11y failure. It's the only identity mark on the toast, and in dark mode it reads as an empty circle.

Also worth separating: the missing dark override on that token is pre-existing and not yours to fix. The pairing is the new part, and it's the only place in the repo putting `--mj-brand-primary` on top of `--mj-brand-primary-light`. The house idiom is already in this file about 60 lines up, in every one of `showToast`'s icon chips:

```
background: color-mix(in srgb, var(--mj-status-success) 14%, transparent); color: var(--mj-status-success);
```

That pattern is used ~680 times under `Generic/`. Switching to `color-mix(in srgb, var(--mj-brand-primary) 14%, transparent)` fixes both themes, stays inside this function, and needs no token work.

This one matters because it's the default path, not an edge case: the server push branch passes no `imageUrl` and no `iconClass`, and nothing in the repo sets `CompletionImageUrlResolver` yet.

**2. The rich toast can't be dismissed.** `notifications.service.ts:431`

The close button only renders when `hideAfter` is falsy, and both call sites pass `hideAfter: 5000`. The toast this replaces on that path always has a close button plus hover-pause. Small thing, but it's a net loss in dismissibility for the exact notification being upgraded.

**3. `[data-message-id]` is the lookup this package already documents as unreliable.** `conversation-chat-area.component.ts:4580`

The list unmounts far-from-viewport items into spacers, and folds any `AgentSessionID`-stamped row into a session card with no node of its own. `ScrollToMessage()` exists on the list specifically because this query silently failed for the pins panel (see the comment at `message-list.component.ts:465`), and `onJumpToMessage` already delegates to it.

For a normal turn this mostly works, since MOUNT_BUFFER keeps the user message mounted at 5 timeline items. In a realtime/voice session every turn is session-stamped, so the target never has a node and the landing silently no-ops after burning 20 retries. Routing through `ScrollToMessage()`, or at least resolving through the timeline key, would make it robust.

## Worth fixing, not blocking

- **`liveRichToasts` never deletes.** `dismissToast` removes the element but not the map entry, on a singleton that lives for the session. One detached toast DOM tree retained per conversation.
- **A same-key toast raised during the 300ms slide-out is swallowed.** `live.element.isConnected` is still true while the element animates out, so the call takes the extend branch and resets a timer on a node that's about to be removed. Nothing renders. That's inside the exact window the dedupe is meant to cover.
- **Two deferred same-key calls: the second shows immediately.** The `pending` branch clears the defer and falls through without re-checking `options.deferMs`.
- **The dedupe key is per-conversation, not per-run.** Two agents finishing in one conversation inside the window collapse into a single toast, so the second is never announced.
- **`#mj-toast-keyframes` id collision.** `ensureToastKeyframes` claims the same element id as the existing writer in `showToast`, with a different body. There was only one writer before this PR, so the collision is new. First writer wins, so a rich toast early in a session strips the `scale()` easing off every simple toast after it.
- **JSDoc contradicts the code.** `MJRichNotificationOptions.context` says the resolver is consulted "when the caller gives no image", but `resolveRichImage` always consults it and lets it win. The README has it right, so it's just the JSDoc. Separately, `fillRichToast`'s "(Re)writes ... de-duplicated updates alike" describes a re-fill path the design deliberately doesn't have.
- **`bottomFollowSuppressedUntil` is unconditional for 1500ms**, so it also swallows the follow armed by a genuinely new message, e.g. the reader sending a second message right after a reply lands.
- **`readReplyFromTop` is in the wrong README table.** That table is introduced as feature gates for paring the chat down, and every other row defaults to `true`. This is a behavior switch defaulting to `false`, so it reads oddly next to the others.
- **Minor a11y note**, only because you added `role="status"` and clearly intended it to announce: `fillRichToast` runs before `container.appendChild`, so the live region enters the DOM already populated and screen readers generally won't read it. Swapping those two lines makes the intent land. The existing `showToast` has the same ordering and no role at all, so this isn't a regression, just an easy win while you're in here.

## Checked and clean

Saves you re-verifying: `AIEngineBase` is imported and declared as a dependency, the new interfaces are exported through `public-api.ts` so hosts can type against the options object, the server really does publish `conversationId` at the top level of the payload (`RunAIAgentResolver.ts:873`) so the dedupe branch is live rather than dead, every token referenced in the new markup resolves, `existingIndex` is computed before the array mutation so the new/update classification is sound, and the changeset bump levels are right.

cmbrock-BC:
Thanks both — every item is in `3fac057c2a`.

**@MattC-BC**

*Blocker — `offsetTop` on the pinned header.* Right, and the failure mode was worse than a bad gap: negative `turnTop`, clamp to 0, and a fits-check that always said "taller". `turnTopClearance` now uses `offsetHeight` + the CSS inset (`getComputedStyle(header).top`). The test stubs answer per selector now, so the header is exercised for real: a fixture 5000 px into a conversation lands at 4960, and putting `offsetTop` back fails two named tests.

*Should-fix.* (1) Icon chip is on the house tint — `color-mix(in srgb, var(--mj-brand-primary) 14%, transparent)` with the brand colour on top, same idiom as `showToast`'s chips. (2) Close button always renders (`fa-xmark`, styled as the simple toast's), and the toast has the same hover-pause / 2 s resume. (3) The turn's message is resolved through the list: new public `MessageListComponent.FindTimelineElement(messageId)` — the timeline-key lookup `ScrollToMessage` already used, extracted into `resolveTimelineNode` so both share it. Mounted item, spacer or session card, whichever stands for the message.

*Non-blocking, all done:* registry entries are dropped the moment dismissal starts (so a same-key call during the slide-out renders a fresh toast rather than extending a dying one); a second deferred same-key call re-defers with its wording instead of showing at once; one keyframes writer shared by both toasts (the simple toast's `scale()` body); both JSDoc contradictions corrected; a genuinely new message clears the landing's follow suppression (`bottomFollowSuppressedUntil = 0` on the `new` arm); the live region is appended empty and filled afterwards; `readReplyFromTop` moved out of the feature-gate table into its own "Behaviour switches" table.

*Per-conversation key.* Kept, deliberately, and now on the record in the description under "One deliberate trade-off": the server payload carries only the conversation id, so a per-run key would defeat the folding the feature exists for. The window is down from 15 s to 3 s — enough to fold the two announcements of one completion (tens of ms apart, 1.5 s deferral), narrow enough that a different agent finishing seconds later is announced on its own.

**@rkihm-BC**

*Refinement path.* Fixed by landing on every settled AI message of the turn rather than the first — re-landing is idempotent (same target, or nothing once the reader has been moved off the bottom), so the status line lands the short turn at the bottom and the real reply lands it again at the top. `currentTurnLanded` is gone. Fixture added with the settled status line ahead of the reply.

*Nits.* `liveRichToasts` releases on dismiss; `fillRichToast`'s comment now says what it does.

*Question.* Intended — see the trade-off paragraph above; the 3 s window narrows it.

Suites: conversations 1287/1287, notifications 20/20; both packages build on `next`; `check:ui` clean.

rkihm-BC (dismissed):
Re-approving. `3fac057c2a` addresses everything from both reviews, and I checked the fixes against the code rather than the summary. One new interaction to look at, created by the fixes themselves — not blocking, and I'd rather you judged it than had me guess.

## The blocker is genuinely fixed

`turnTopClearance` now reads `parseFloat(getComputedStyle(stickyHeader).top)` instead of `offsetTop`, with an early return of the bare gap when there is no pinned header. That is the correct property: for `position: sticky`, the resolved value of `top` is the specified inset, while `offsetTop` reports the used position and tracks the scroll — which is exactly the distinction @MattC-BC identified.

**And the test now catches a regression rather than encoding the stub.** The fixture supplies `offsetTop: 5012` alongside `cssTop: '12px'`, and the stubs answer per selector, so reverting to `offsetTop` would compute `5016 - (16 + 28 + 5012) = -40`, clamp to 0, and fail the `{top: 4960}` assertion. That was the sharper half of Matt's point and it landed.

## Everything else verified

- **Icon chip** — `color-mix(in srgb, var(--mj-brand-primary) 14%, transparent)` with the brand colour on top. `--mj-brand-primary-light` is gone from the file; this is now the same idiom as `showToast`'s chips.
- **Close button** — unconditional in the markup, no `hideAfter` gate, wired to `dismiss`.
- **`FindTimelineElement`** — real, and shares `resolveTimelineNode` with `ScrollToMessage` rather than duplicating the lookup, so spacers and session cards resolve the same way for both.
- **Registry release** — `dismiss` deletes on identity (`?.element === toast`), so it cannot drop a newer toast's entry. That is the right guard and it is easy to get wrong.
- **Single keyframes writer** — one `ensureToastKeyframes`, called from both paths.
- **Live region** — `appendChild` then `fillRichToast`. Ordering is right.
- **JSDoc** — `resolveRichImage` is host-first with `fromHost || options.imageUrl`, and the comment now says the resolver wins. `fillRichToast`'s comment says what it does.
- **README** — `readReplyFromTop` in its own "Behaviour switches" table.
- **My refinement finding** — `currentTurnLanded` gone, lands on every settled AI message, idempotent.

22 checks pass.

One property worth stating because it looks like an oversight and is not: `extend` updates `hideAfter` and re-arms but leaves `shownAt` alone, so the dedupe window runs from the first announcement rather than sliding. That is what stops a stream of repeats keeping one toast alive forever. Correct as written.

## The new interaction — worth your judgement, not a blocker

`bottomFollowSuppressedUntil = 0` on the `new` arm can undo a landing that has already happened, on the multi-message turn the refinement fix is for.

`followTranscript` reaches the `new` arm only when the message did **not** hit the settled-AI branch above it. An AI message arriving already `Complete` lands and returns. An AI message arriving **unsettled** falls through — and clears the suppression.

So for a turn where message N settles before message N+1 is created:

```
AI status line settles (update, Complete)  -> LANDS, suppression = now+1500
AI reply created In-Progress (new)         -> falls through -> suppression = 0, scrollToBottom = true
AI reply settles (update, Complete)        -> LANDS again
```

The reader lands at the top, snaps to the bottom, then lands again. It self-corrects, so this is a flicker rather than a wrong final position — but it is on the path the refinement fix exists to serve.

The new test does not exercise it: `refinementStatus` and `ai-2` are both constructed already `Complete`, so both take the settled branch and return before reaching the `new` arm. A reply that streams — created In-Progress, settled by a later update — is the case that falls through.

Whether that is reachable depends on whether a refinement reply is ever created unsettled after a prior message in the same turn has settled. You know that path; I don't. If it is, narrowing the clear to `message?.Role === 'User'` would keep Matt's fix (the reader sending again) while leaving an AI continuation from cancelling its own turn's landing.

Ships inert either way — nothing in the repo sets `readReplyFromTop` — so this is for whoever turns it on, same as my original finding.

## Housekeeping

Matt's `CHANGES_REQUESTED` predates `3fac057c2a` and is still the blocking decision; @MattC-BC will need to clear it. One commit behind `next`.

rkihm-BC:
Following up on the one open item in my re-review — I had asserted the mechanism without tracing it, so I went and did that. **It holds, and the mechanism is confirmed:** `bottomFollowSuppressedUntil` really is what aborts the snap, not just a hint.

`ngAfterViewChecked`:

```ts
if (this.scrollToBottom) {
  this.scrollToBottom = false;
  setTimeout(() => {
    if (Date.now() < this.bottomFollowSuppressedUntil) {
      return;                       // <- the guard that would have aborted it
    }
    this.scrollToBottomNow();
    this.checkScroll();
  }, 100);
}
```

So `scrollToBottom = true` only *schedules*; the suppression check inside the timeout is what cancels. Setting `bottomFollowSuppressedUntil = 0` on the `new` arm removes exactly that guard, and the queued scroll then runs 100 ms later.

Counterfactual, for completeness: without the clear, a reply arriving inside the 1500 ms window hits `Date.now() < suppressedUntil` and returns — which is the behaviour the suppression exists for. So the hole is genuinely opened by the fix rather than pre-existing.

Two things I also checked while I was in there, both fine:

- `clearTurnTracking()` resets the suppression too, so a conversation switch cannot strand it.
- The suggested narrowing works: a User message takes the first branch (setting `currentTurnStartMessageId`) and then **falls through** to the `new` arm, so gating the clear on `message?.Role === 'User'` still catches @MattC-BC's case — the reader sending again right after a landing — while stopping an AI continuation from cancelling its own turn's landing.

Reachability is still yours to judge: it needs a refinement reply created **unsettled** after an earlier message in the same turn has already settled. If replies in that path are always created `Complete`, as the new fixture builds them, nothing fires and there is nothing to do.

Approval unchanged.

MattC-BC:
Approving. I checked `3fac057c2a` against the code for every item I raised, and they all land. The part I cared most about was the test: with per-selector stubs, reverting to `offsetTop` now fails a named test instead of passing on the stub. The per-conversation dedupe key makes sense as a trade-off given the server payload only carries the conversation id, and the 3s window keeps it narrow.

**On Robert's open item: it isn't reachable today, so there's nothing to do.**

Robert's condition was that if replies on the refinement path are always created `Complete`, nothing fires. They are. That path is `handleSilentObservation`:

- the status line is created `Complete` (`message-input.component.ts:2381`)
- the reply is only created after `await this.agentService.invokeSubAgent(...)` resolves (`:2400`), and it's hardcoded `Complete` too (`:2428`)

Both take the settled branch in `followTranscript` and return before reaching the `new` arm, so the suppression clear never runs for an AI message on that path.

I also checked the broader version of the question: does *any* path emit a settled AI message and then a new In-Progress one in the same turn? Five places create an AI message In-Progress (Sage `:1751`, task graph `:1989`, sub-agent `:2159`, direct @mention `:2504`, continuation `:2729`), and none of them follows a settled AI message in its turn. The two delegation paths create the specialist's message while Sage's is still In-Progress, since the progress callback only re-emits In-Progress messages and never settles one. The other three are the first AI message of their turn. The one thing I didn't trace is server pushes, so a push that settles Sage's detail before the client creates the specialist's message would be the only way in.

If it ever does become reachable, one note on the suggested narrowing: gating the clear on `Role === 'User'` only covers a reply created inside the 1.5s suppression window. Past that, the `new` arm snaps to the bottom either way.

**One optional follow-up, and it's partly on me.** I suggested swapping `appendChild` and `fillRichToast` so the live region announces, and the swap is in. But both calls run in the same task, so by the time the accessibility tree updates, the region shows up already filled, same as before. The new comment says it's announced as an update, and I don't think the swap alone gets you there. Filling it on the next frame (`requestAnimationFrame(() => this.fillRichToast(...))`) would. Fine as a follow-up, not holding this up.

MattC-BC (dismissed):

cmbrock-BC:
Follow-up in `bc62e6da98` (rebased onto the `next` merge):

- **Live region** — you're right that the swap alone changed nothing. The toast is now appended empty and filled on the **next animation frame** (`nextFrame`, falling back to a task where no frame scheduler exists); the close click is delegated to the toast since the button doesn't exist until then. The test steps a faked frame and fails if the fill is made synchronous again.
- **Robert's narrowing** — taken as hardening even though it isn't reachable today: only the reader's own message (`Role === 'User'`) lifts the landing's follow suppression; an AI message arriving new inside the window leaves it in place. Fixture added.

notifications 20/20, conversations 1288/1288, both packages build on `next`.

MattC-BC:
Re-approving after `bc62e6da98`, since the push dismissed the earlier approvals. The narrowing is right: a User message still falls through into the `new` arm and lifts the suppression, an AI message arriving new no longer cancels a landing, and default mode never sets the suppression, so nothing changes there. Thanks for taking it as hardening.

One correction on my side about the live region, and it's not a request for this PR. I don't think a single frame gets the announcement either, in Chromium at least. `requestAnimationFrame` callbacks run at the start of the next rendering step, before that frame's style and layout, and the accessibility tree updates after layout. So the toast is still laid out for the first time already filled. It's harmless as written (it's also why the card never paints empty), so no need to revert, but the comment above it claims more than it delivers.

The reliable version is what Angular CDK's `LiveAnnouncer` does: a persistent, visually hidden live region, separate from the visual toast, with its text set on a ~100ms delay. Delaying the toast's own fill that long would flash an empty card, so it really wants the separate region. That would cover the simple toast too, so I'd rather see it as its own follow-up issue than another round here. Sorry for the back-and-forth on this one.

MattC-BC (approved):

Issue #4241, opened 2026-09-05
Promote entity-validation plumbing onto ValidationResult and BaseEntity (AddError/Finalize, IsNewOrDirty, fail-closed async read helpers)


MarceloT-BC:
## Summary

Writing `Validate()` / `ValidateAsync()` overrides on `BaseEntity` subclasses currently means
hand-rolling the same four or five pieces of plumbing every time. The plumbing is small, but it is
*subtly* easy to get wrong, and the failure modes are all silent — a rule that never runs, an error
that never surfaces on the form, or a save that is blocked by a warning.

MJ core already contains **42 hand-rolled instances across 18 files**, and they have drifted into
**three mutually-incompatible idioms** (detail below). This proposes promoting a small set of
helpers onto `ValidationResult` and `BaseEntity` so the plumbing is written once.

These were extracted while building a teaching app with 13 validation rules across 4 entities; the
duplication threshold was hit at 4 classes, which is exactly the signal that the shape belongs one
level up.

---

## Evidence: the idioms have already diverged in MJ core

`ValidationResult` is a bare data holder — `Success: boolean` and `Errors: ValidationErrorInfo[]`,
no methods ([`packages/MJGlobal/src/ValidationTypes.ts:42`](../blob/next/packages/MJGlobal/src/ValidationTypes.ts#L42)).
So every override invents its own way to close out the result:

| Idiom | Used by | Behaviour on a `Warning` |
|---|---|---|
| `result.Success = false` at each push | `MJDashboardEntityExtended`, `MJEntityFieldEntityExtended` | correct (warnings untouched) |
| `result.Success = result.Success && result.Errors.length === 0` | `MJAISkillPermissionEntityServer`, `MJUserRoutineEntityServer`, `MJUserRoutineRecipientEntityServer` | ❌ **blocks the save** |
| `result.Success = errors.every(e => e.Type !== ValidationErrorType.Failure)` | `MJAIAgentEntityServer` | correct |

The middle idiom is a latent defect rather than a live one: those three classes do not push warnings
themselves. But `super.Validate()` legitimately can — `MJAIAgentEntityServer:136`,
`MJAIAgentSessionBridgeEntityServer:170`, `MJMLTrainingPipelineEntityServer:173` and
`MJAICredentialBindingEntityExtended:64` all emit `ValidationErrorType.Warning` today, and companion
/ IS-A parent validation flows into the same `result`. The moment a warning reaches one of those
[...]

PR #4343, opened 2026-09-10
fix(ng-base-forms): highlight the field when a server-side ValidateAsync refuses a save

MarceloT-BC:
## Summary

A save refused by a server-side `ValidateAsync()` only toasted, while the same rule written in a synchronous `Validate()` painted the field red with its inline message. The refusal's field-named reasons were dropped at the API boundary: `ResolverBase` threw only `CompleteMessage` (the prose, fixed by #3973), and `GraphQLDataProvider` kept only `errors[0].message`, so the form never learned *which* field.

Four hops, one publication path:

- **`@memberjunction/server`** — `RefusalExtensions()` builds the `extensions` for the create / update / delete refusal throws: `{ code, entityName }` as before, plus `validationErrors` (the `LatestResult.Errors` flattened to plain `{ Source, Message, Value, Type }`) only when there is something to carry. The message is unchanged.
- **`@memberjunction/global`** — `SerializeValidationErrors` / `DeserializeValidationErrors`: the one place the class→plain-object flattening and rehydration live, so producer and consumer cannot drift. `Value` is reduced to JSON-safe primitives.
- **`@memberjunction/core`** — `SaveEntityGraphOutput.ValidationErrors` so RRC / graph saves (orders, accounting) get the same parity; the client graph-save path rehydrates it into `LatestResult.Errors`.
- **`@memberjunction/graphql-dataprovider`** — the `Save()` and `Delete()` catch blocks rehydrate the extension into `LatestResult.Errors` as real `ValidationErrorInfo` instances. Garbage yields `[]`, never a throw.
- **`@memberjunction/ng-base-forms`** — `BaseFormComponent.SaveRecord()` publishes a local `Validate()` refusal and a field-sourced server refusal through the SAME private method (`showValidation`, `validationErrors`, a bumped `validationRevision`, warning toast, `ValidationFailed`). Errors with no field source stay toast-only; `RecordSaveFailed` still fires. `mj-form-field` records the revision current when it was last edited: edited **before** the failed save, it shows the published server error (the real gotcha — plain `touched` hid it); edited **after**, it returns to live local validation, which is also how a later valid edit clears it.

## Repro

Before: open a new `MJ: Tag Scopes` record, pick a tag with `IsGlobal=1`, save. The toast reads the refusal; no field is marked. Direct GraphQL `CreateMJTagScope` returns `extensions: { code, entityName }` only.

After: the Tag field shows the red border and the server's message; a later edit of the field clears it. **The toast changes too:** a server refusal that names a field now shows the amber "Validation Errors\n…" warning (the same toast a local `Validate()` refusal shows) instead of the red "Save failed: …" error toast. A refusal with no field source keeps the red error toast. GraphQL returns `extensions.validationErrors = [{ Source: 'TagID', Message: 'Cannot add TagScope row …', Value, Type: 'Failure' }]`.

## Tests

| Tier | Result |
| --- | --- |
| MJGlobal serialize/deserialize unit | 8/8 |
| MJServer `RefusalExtensions` + source scan of the three throws | 12/12 (incl. existing refusal-message spec) |
| GraphQLDataProvider wire tests (with extension, without, malformed, delete) | 22/22 |
| base-forms DOM — field + form | 51/51 |
| Red-proof against the pre-fix components | field: 3 of 8 new cases fail; form: 7 of 9 fail |
| Full suites of every touched package | global 878, core 2198, provider 429, server 1217, base-forms 354 — all green |
| Live GraphQL + Playwright on Explorer (Tag Scopes on a global tag) | field painted, toast shown, other fields clean |

Patch changeset (no migration, no metadata). Builds on #3973; adjacent #4241, #3971.

A `ValidateAsync()` that runs in the browser used to toast a bare "Error saving record" (`LatestResult.Message` is empty on that path); it now highlights the field as well.

The client's `LatestResult.CompleteMessage` carries the refusal text **once**: the provider copies the server's `CompleteMessage` into `Message` and rehydrates the same entries into `Errors`, so `CompleteMessage` returns `Message` alone in that case. The provider does not infer that from the text: `RefusalExtensions` states `messageIncludesValidationErrors: true` on the wire beside `validationErrors` (both come from the same `LatestResult`), and the provider sets `BaseEntityResult.MessageIncludesErrors` from that statement — the producer-side fix the `CompleteMessage` comment was waiting for; no substring dedupe. A server that omits the statement gets the pre-existing client behaviour. The IS-A parent-failure producers, which also build `Message` from `Errors`, set it too.

Side finding, **not** in this branch: #4178's `StoredDateIsUnreadable` warns on every non-date field in edit mode (no type guard) — separate PR.


MattC-BC (changes requested):
Nice work on this one. The field painting looks right in the screenshot, and the revision logic in mj-form-field holds up through every case I walked through (edit before the failure, edit after, a second failure, a success after a failure, cancel, record swap, older hosts with no revision). Letting child-record errors keep their prefixed field names, so they can't highlight a field on the parent, is a nice touch.

One thing I'd like fixed before merge, plus a few small ones.

1. The refusal text now appears twice in the client's LatestResult.CompleteMessage (this is the one I'm requesting changes for)

Since [#3973](https://github.com/MemberJunction/MJ/issues/3973) the provider puts the server's whole CompleteMessage into the client's LatestResult.Message (graphQLDataProvider.ts:2008).
On a validation refusal the server's Message is empty and the text lives in Errors, so that message is just the error messages joined.
This PR also rehydrates the same entries into the client's Errors (:2015 for save, :2255 for delete, baseEntity.ts:2689 for graph saves).
CompleteMessage is Message + Errors, deliberately not deduped (baseEntity.ts:781), so on the client it now reads "Cannot add TagScope row…\nCannot add TagScope row…".

There are 143 LatestResult.CompleteMessage reads across 53 Angular files. For example, test-form.component.ts:937 calls SaveRecord() and then toasts CompleteMessage. The base form only dodges it because it reads .Message. The wire tests check Message and Errors separately but never CompleteMessage, so nothing catches it.

The obvious fixes each break a group of readers. Matching the server exactly (empty client Message) costs the 75 .Message readers the text they get today. Moving the errors to a separate field means Errors no longer matches what a local Validate() produces. My suggestion: have the producer mark the result when its Message already contains the Errors text, and have CompleteMessage return Message alone when that's set. That's the producer-side fix the CompleteMessage comment says it's waiting for, not the substring dedupe that was tried and reverted. Your call on the mechanism, but please add a wire-test assertion that CompleteMessage carries the text once.

2. The description says "the toast is unchanged"

It does change: a server refusal that names a field now shows the amber "Validation Errors\n…" warning instead of the red "Save failed: …" error (your test asserts it and the screenshot shows it). I'm good with the change, since it matches a local Validate() refusal and the field carries the detail. Just update the description so QA isn't surprised.

3. Nit: leftover doc comment in form-field.component.ts (~L313)

Merge artifact: the comment for editedSinceLastValidationFailure now sits above markTouched()'s own comment, so the getter has no doc and there's a stray block.

4. Worth a line in the changeset

A ValidateAsync() refusal that runs in the browser used to toast a bare "Error saving record" (LatestResult.Message is null on that path). With this PR it highlights the field too. Nice bonus, worth saying so.

FYI, no change needed: a server error stays on its field until that field is edited or the next save, since the client can't re-run the rule. For cross-field rules that's a little odd (in your screenshot, fixing it by changing Target Agent leaves the old message on Co-Agent until save), but untouched fields already behave that way with local errors, so I'm fine with it.

I traced #1 through the code rather than reproducing it live, and CI was still pending when I looked.

MarceloT-BC:
Thanks for the careful trace — #1 was real and I'd missed it because the wire tests asserted `Message` and `Errors` separately and never `CompleteMessage`. Addressed in 3f2d67ff47 and 94d09335cb:

**1. Refusal text twice in the client's `CompleteMessage` — fixed, producer-side as you suggested.** New `BaseEntityResult.MessageIncludesErrors` (default `false`), set by a producer whose `Message` already renders every entry of `Errors`; `CompleteMessage` then returns without appending `Errors`, but only when `Message` actually has text — a flag with nothing behind it must not lose the errors. No substring dedupe. Set in the four producers that build `Message` out of `Errors`: `GraphQLDataProvider` Save and Delete — where it is NOT inferred from the text: `RefusalExtensions` now sends `messageIncludesValidationErrors: true` beside `validationErrors` (both built from the same `LatestResult`), and the provider repeats that statement; a server that omits it gets today's behaviour — the graph-save client path (Message = `ErrorMessage`), and the two IS-A parent-failure paths — only when their detail was built from the parent's `Errors` rather than its `Message` (when the parent had a `Message`, the errors aren't in it and the flag stays off). Every other producer keeps today's output.

Tests: wire — `CompleteMessage` carries the prose exactly once on Save and on Delete, and is the bare message with no extension; core — the flag's semantics (default, set, empty `Message` ignored, the single `Error` property still appended), IS-A parent save (once; flag off when the detail came from the parent's `Message`), and graph save (`ValidationErrors` rehydrated, once).

**2. "The toast is unchanged" — description corrected.** It now says a field-sourced server refusal shows the amber "Validation Errors" warning, the same toast a local `Validate()` refusal shows, and that a refusal with no field source keeps the red "Save failed" error toast.

**3. Stray doc comment above `markTouched()` — fixed.** Merge artifact from routing `next`'s new image-upload touched sites through the helper; the getter has its own doc back.

**4. Changeset — added** the browser-side `ValidateAsync` case (bare "Error saving record" before, field highlighted now) and a line on `CompleteMessage` carrying the text once.

On the FYI: agreed, and thanks for calling it out explicitly — a server error stays on its field until that field is edited or the next save, same as untouched fields already behave with local errors. Once #4355 lands the rail count will at least show the section still has an open error.

Re-verified after the change: core 2321/2321, provider 433/433, server refusal specs 12/12, base-forms field + form DOM 59/59; MJAPI restarted on it for the live host.

MattC-BC:
Thanks Marcelo, all four are addressed. I re-checked each one against the new commits.

#1: Fixed the way I'd hoped. The server now says on the wire that its message already contains the errors (messageIncludesValidationErrors), and the provider only sets MessageIncludesErrors when it sees that, so nothing is guessed from the text. CompleteMessage ignores the flag when Message is empty, so the errors can't get lost, and the wire tests now assert the text appears once, plus the "no statement, no marking" case. Carrying it into the IS-A parent-failure paths was a good catch too.
#2: Description reads right now.
#3: Doc comments are back in order.
#4: Changeset line is in.

Approving. Build and the integration tier were still running when I looked, so merge once they're green.

MattC-BC (approved):

PR #4351, opened 2026-09-10
fix(explorer-core): the user menu fits the viewport; hosts can narrow the shell chrome per tenant

cmbrock-BC:
Two shell items from a white-label host's demo round, both in `explorer-core`, both small.

## The gap

**The user menu is unreachable at the bottom in a short window.** `.user-context-menu` is `position: absolute` with `overflow: hidden` and no height limit. When the viewport is shorter than the menu (a window that isn't full height, or a phone), the items below the fold are clipped, not scrolled — you cannot get to them.

**Shell chrome is instance-wide only.** The search bar, notifications, app switcher and app nav are gated by Instance Config (`Shell.SearchBar.Enabled` and friends), resolved once in `chromeFlags`. There is no hook for a host that must answer differently for *some* users. The case that surfaced it: a white-label product that sells knowledge-base browsing as a plan feature needs the search bar hidden for organizations without that plan, while other organizations on the same instance keep it. Today the only options are "off for everyone" or a CSS pierce on the shell's class names.

## The fix

**Menu:** `max-height: calc(100dvh - 60px - 3 * var(--mj-space-2))` (60px is the header's fixed height; a `100vh` line precedes it as the fallback), `overflow-y: auto`, and `max-width: min(360px, calc(100vw - 2 * var(--mj-space-4)))`. Rounded corners still clip because `overflow-x: hidden` stays.

**`BaseShellChromePolicy`** — the same ClassFactory pattern as `BaseUserMenu`. The shell resolves Instance Config into a baseline as before, then applies the registered policy through `ApplyShellChromePolicy`, which enforces the contract: booleans can only be **narrowed** (a `true` from the policy never re-enables what Instance Config turned off), `recordOpenStyle` is kept (it is resolved at startup and pushed to collaborators — not a getter's to change), `appSwitcherStyle` may be replaced. The policy gets a copy, so a mutating host can't reach the baseline, and a policy that **throws** leaves the baseline in place with an error naming the subclass (review hardening). Its `Changed` subject drops the shell's cached flags and marks for check, so an organization or plan switch repaints the chrome. With nothing registered the policy is the identity and the shell behaves exactly as today; `ShellChromeFlags` moves out of the component so hosts can type against it, and both are exported through `public-api.ts`.

**No keyboard back door.** The omnibar palette used to render at the shell root regardless of `ShowSearchBar`, so with the search bar hidden Ctrl/Cmd+K (and Cmd+/, the mobile search icon, `toggleSearch`) still opened search — found in the live test, fixed in the second commit. The palette now renders only with the search bar, and `OpenOmnibar` / `OnHeaderSearchClick` / `toggleSearch` return early when it is hidden.

## Testing

- `shell-chrome-policy.test.ts` (14, node): identity default, narrowing, no widening past the ceiling, `recordOpenStyle` kept / `appSwitcherStyle` taken, argument isolation, ClassFactory resolution with and without a registered subclass, the shell's `Changed` handler dropping the cache, the CSS contract for the menu (dvh cap with vh fallback, `overflow-y: auto`, no `overflow: hidden`, max-width), and the search back door (`OpenOmnibar` inert while hidden, palette rendered only with the bar). Mutation-checked: letting the policy widen, putting `overflow: hidden` back, and removing the `OpenOmnibar` guard each fail exactly one named test.
- **Live**, on a `next` build against a freshly migrated database: the menu's computed `max-height` matched the formula for the viewport (628px at 712px), and at a 380px-tall viewport it capped at 296px, scrolled, and the last item — hidden before scrolling — was reachable after. The registered policy instance's `Resolve` was swapped at runtime to hide search and `Changed` fired: `ShowSearchBar` false, the search button and the palette left the DOM, Ctrl/Cmd+K, Cmd+/, the header route and the mobile toggle all inert with no state flips, notifications / app nav / app switcher untouched; swapping the identity back restored search; with search shown `OpenOmnibar('standard 11')` opened the palette with the query.
- Package suite 189/189; `pnpm run build` clean; `check:ui` 0 violations. README gains a "Shell Chrome Policy" section beside the user-menu one.

Changeset: `@memberjunction/ng-explorer-core` patch.


rkihm-BC (dismissed):
## Approve

Two unrelated fixes, both small and both landing cleanly.

The menu clip is a straightforward CSS fix, and capping with `dvh` behind a `vh` fallback is the right call for the phone case that surfaced it.

The chrome policy is the more interesting half and it is built the way I would want. **The narrowing contract is enforced in `ApplyShellChromePolicy`, not trusted to the policy** — a host returning `true` for something Instance Config disabled simply cannot widen it, because the caller ANDs against the baseline. That is the difference between a contract and a comment. Passing `{ ...baseline }` so a mutating host cannot reach the real object is the same instinct, and keeping `recordOpenStyle` out of the policy's reach is right given it is resolved at startup and pushed to collaborators.

The identity default means hosts registering nothing see no change at all, which is what makes this safe to land.

Also good: noticing that hiding the search bar removes the omnibar with it, so there is no separate shortcut to guard. That is the sort of thing that otherwise ships as a bug two releases later.

### One small hardening suggestion

`Resolve()` is host code, called with no guard. A policy that throws takes out chrome resolution for the whole shell, and the failure would surface as missing chrome rather than as an error pointing at the host's subclass. A try/catch falling back to the baseline would match how defensively the rest of this is written — narrowing on success, unchanged on failure.

Not blocking; the default implementation cannot throw.

cmbrock-BC:
@rkihm-BC — two commits since your approval, and the first one contradicts a line in it, so this needs your eyes again rather than a rubber stamp:

**`d1bc2d1a23` — the keyboard back door.** Running the branch live against a `next` database showed the claim we both accepted was wrong: `<mj-omnibar-palette>` was rendered at the shell root regardless of `ShowSearchBar`, so with the search bar hidden Ctrl/Cmd+K (and Cmd+/, the mobile search icon, `toggleSearch`) still opened search. The palette now renders only inside the `ShowSearchBar` guard, and `OpenOmnibar` / `OnHeaderSearchClick` / `toggleSearch` return early when it's hidden. Verified live: narrowed, the palette leaves the DOM and every route into search is inert with no state flips; with search shown, `OpenOmnibar('standard 11')` opens the palette with the query. Three tests; removing the `OpenOmnibar` guard fails one by name.

**`747f851b3b` — your hardening.** `ApplyShellChromePolicy` wraps `Resolve()`: a throwing host policy leaves the Instance Config chrome unchanged and logs an error naming the subclass. One test.

Also verified live: the menu's computed `max-height` matched the formula for the viewport (628px at 712px); at a 380px-tall viewport it capped at 296px, scrolled, and the last item — hidden before scrolling — was reachable after.

189/189 in the package; PR description corrected.

rkihm-BC (approved):
## Re-approve — and my last review was wrong about the thing you fixed

Thank you for pushing back on your own description instead of taking my approval as settled. **I got this one wrong**, and it is worth naming precisely so it does not get lost in a re-approval.

I wrote, approvingly: *"noticing that hiding the search bar removes the omnibar with it, so there is no separate shortcut to guard."* That was false. On `next`, `shell.component.html:394` renders `<mj-omnibar-palette>` at the shell root, outside any guard:

```
next:   394:<mj-omnibar-palette #omnibarPalette ...></mj-omnibar-palette>
branch: 396:@if (ShowSearchBar) {
        397:  <mj-omnibar-palette #omnibarPalette ...></mj-omnibar-palette>
```

So the exact scenario this feature exists for — a host narrowing chrome for a tenant without the plan — left every keyboard route into search live. I verified the *policy* code line by line and took a claim about template structure on trust, in a PR whose whole subject is what renders. That is my miss, not a description error you got away with: you found it by running the branch live, which is the check I skipped.

### `d1bc2d1a23` — correct, and layered the right way

The template guard alone would have been enough to stop the palette opening, since `OpenOmnibar` already used `?.` on the ViewChild. Guarding the three entry points as well is what stops `toggleSearch` flipping `isSearchOpen` on a shell with no search surface — state changing behind a UI that does not exist is exactly the kind of thing that surfaces later as an unreproducible bug.

I went looking for a fourth route and did not find one. Every caller funnels through `OpenOmnibar` — the Cmd+/ handler at `:2805`, the Ctrl+K handler via `OnHeaderSearchClick` at `:3005`, `toggleSearch` at `:2930` — and `OpenOmnibar` is now the chokepoint. The non-omnibar `isSearchOpen` popup path is covered by the `toggleSearch` guard.

### `747f851b3b` — the hardening

Exactly the shape I suggested: `{ ...baseline }` returned on a throw, consistent with the copy semantics everywhere else in that function, and the log names the subclass so a host debugging missing chrome is pointed at their own code rather than at the shell.

### The live measurements

628px at a 712px viewport, and 296px with the last item reachable at 380px, is the verification I asked for in a different form and better than a screenshot — it shows the formula holding at two points rather than one.

Approving on the current head. CI has 12 green with 2 still running; if either goes red treat this as not yet in force.

PR #4352, opened 2026-09-10
fix(core): RLS exemption honours the permission row's Can* flags

cmbrock-BC:
Branch: `fix/rls-exemption-honours-can-flags` (off `origin/next` @ c1fea88930) · commit `7f3c60c435`

## The gap

`EntityInfo.UserExemptFromRowLevelSecurity` decides whether a user is exempt from row-level security
for an operation by scanning the user's permission rows and returning exempt on the first row with no
filter for that operation:

```ts
case EntityPermissionType.Create:
    if (!ep.CreateRLSFilterID) return true;
```

A permission row carries a filter only for operations it grants. A row with `CanCreate=false` also
has `CreateRLSFilterID=null`, and the check reads that null as "unrestricted". So a **read-only**
role exempts its holders from Create/Update/Delete RLS.

Every authenticated user holds the `UI` role, whose rows are read-only on most entities. Any user
whose *other* role binds Create/Update/Delete to a row filter (the multi-tenant shape: an org admin
whose writes are bound to their own organization) is therefore exempt from that filter on the
client path — `RunView` / the generated GraphQL mutations — and can write a record into another
tenant. The server-side `PreRunView` middleware does not see this because the exemption is decided
inside `GetUserRowLevelSecurityWhereClause` / `GetEffectiveRowFilterWhereClause` before any filter
is composed.

Betty (BlueCypress) has carried this as a local dist patch since 2026-08-21
(`patches/@memberjunction+core+6.1.0-edge.1.patch`); this PR is the upstream fix so that patch can
be deleted.

## The fix

Four one-line condition changes in `UserExemptFromRowLevelSecurity`:

```ts
if (ep.CanCreate && !ep.CreateRLSFilterID) return true;   // and Read / Update / Delete alike
```

A filter-less row that **grants** the operation still exempts, exactly as before. A row that does
not grant the operation no longer says anything about exemption. `GetUserRowLevelSecurityInfo`
(which collects the filters that apply) is untouched: it already keys on the filter ID being
present. JSDoc on the method explains the rule.

## Evidence

- **5 new tests** in `entityInfo.rlsExemption.test.ts` under
  *"Exemption requires the row to GRANT the operation (Can* flags)"*: read-only row does not exempt
  Create/Update/Delete; `CanRead=false` row does not exempt Read; filter-less granting row still
  exempts; the real multi-role shape (UI read-only + tenant role with a Create filter) is not
  create-exempt and the Create clause still renders; a read-only row with an unfiltered Read still
  exempts Read.
- **Red before, green after:** 3 of the 5 fail on `next` as-is (the two "still exempts" cases pass
  either way, by design). All 34 tests in the file pass with the fix. The pre-existing 29 tests are
  unchanged — they all build rows with every `Can*` flag true, so they were never sensitive to this.
- Verified on a fresh `pnpm install --frozen-lockfile` of `next`: `packages/MJCore` builds, all 2316
  core tests pass (160 files), `check:esm` and `check:changeset` are green.

## Changeset

`.changeset/rls-exemption-honours-can-flags.md` — `@memberjunction/core: patch`.

## Not in this PR — but not inert either

`GetUserRowLevelSecurityInfo` still collects a filter from a row whose `Can*` flag is false when that
row happens to keep its filter ID. I first called that inert on the grounds that the permission gate
refuses the operation; that is wrong. `GetUserPermisions` ORs the flags across roles
(`entityInfo.ts:2957`), so a granting role carries the user past the gate and the non-granting
role's leftover filter is then OR'd into the clause, which widens. Nothing in MJ clears a filter
when its flag is cleared. On the BlueCypress production copy there are 9 such rows, all with
"own rows by UserID" filters, so the exposure there is bounded to the user's own rows. The
symmetric fix (honour the same `Can*` flag before collecting a filter) is a follow-up, kept
separate as suggested in review.

### DIFF
--- a/.changeset/rls-exemption-honours-can-flags.md
+++ b/.changeset/rls-exemption-honours-can-flags.md
@@ -1,0 +1,9 @@
+---
+"@memberjunction/core": patch
+---
+
+`EntityInfo.UserExemptFromRowLevelSecurity` now requires the permission row to GRANT the operation before it can exempt the user from row-level security for it.
+
+A permission row carries an RLS filter only for the operations it grants, so a row with `CanCreate=false` also has `CreateRLSFilterID=null`. The exemption check read that null as "unrestricted" and returned exempt. Every authenticated user holds the `UI` role, whose rows are read-only on most entities, so any user whose tenant role bound Create/Update/Delete to a row filter was silently exempt from that filter on the client (RunView/GraphQL) path: an org admin could write a record into another organization.
+
+The check is now `ep.CanCreate && !ep.CreateRLSFilterID` (and likewise for Read/Update/Delete). A filter-less row that grants the operation still exempts, exactly as before; a row that does not grant it no longer says anything about exemption. Regression tests added to `entityInfo.rlsExemption.test.ts`.
--- a/packages/MJCore/src/__tests__/entityInfo.rlsExemption.test.ts
+++ b/packages/MJCore/src/__tests__/entityInfo.rlsExemption.test.ts
@@ -333,6 +333,84 @@ describe('EntityInfo RLS Exemption (centralized in GetUserRowLevelSecurityWhereC
         });
     });
 
+    describe('Exemption requires the row to GRANT the operation (Can* flags)', () => {
+        // A permission row confers an RLS exemption only for an operation it grants. A row that
+        // does not grant an operation carries no filter for it either, and that absence means
+        // "not applicable", not "unrestricted". Before this guard, a read-only row (CanCreate=false,
+        // hence CreateRLSFilterID=null) made every holder of that role exempt from CREATE RLS —
+        // and a read-only role is exactly the shape of the 'UI' role every authenticated user holds.
+        it('a read-only row does NOT exempt the user from Create/Update/Delete RLS', () => {
+            const entity = buildEntityInfo([
+                buildPermission({
+                    RoleID: UI_ROLE_ID,
+                    CanRead: true, CanCreate: false, CanUpdate: false, CanDelete: false,
+                    ReadRLSFilterID: READ_RLS_FILTER_ID,
+                }),
+            ]);
+            const user = buildUser(USER_ID, [UI_ROLE_ID]);
+
+            expect(entity.UserExemptFromRowLevelSecurity(user, EntityPermissionType.Create)).toBe(false);
+            expect(entity.UserExemptFromRowLevelSecurity(user, EntityPermissionType.Update)).toBe(false);
+            expect(entity.UserExemptFromRowLevelSecurity(user, EntityPermissionType.Delete)).toBe(false);
+        });
+
+        it('a read-only row does NOT exempt the user from Read RLS when CanRead is false either', () => {
+            const entity = buildEntityInfo([
+                buildPermission({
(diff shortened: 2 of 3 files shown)


rkihm-BC (approved):
## Approve

Small, correct, and in the safe direction. The premise checks out: a permission row carries a filter only for operations it grants, so `CanCreate=false` implies `CreateRLSFilterID=null`, and the old code read that absence as "unrestricted". With the `UI` role read-only on nearly every entity and held by every authenticated user, that reached everyone — this is a real multi-tenant write-isolation hole, not a theoretical one.

I checked the risk direction rather than just the logic. The change can only make exemption **harder**, never easier, so it cannot silently widen access. And it cannot produce a false positive: a row that grants the operation with no filter still exempts, exactly as before. The only behaviour that disappears is a non-granting row conferring an exemption, which was never meaningful.

`GetUserRowLevelSecurityInfo` being untouched is right — it already keys on the filter ID.

Nice provenance note about the Betty dist patch. Upstreaming a patch someone has carried since August is worth more than the diff size suggests.

### One correction to "Not in this PR"

You describe the remaining `GetUserRowLevelSecurityInfo` gap as inert because "the permission gate refuses the operation before RLS matters." I do not think that holds, for the same reason the bug you are fixing did not.

`GetUserPermisions` aggregates across roles — `bucket.CanCreate = bucket.CanCreate || !!ep.CanCreate`. So a user granted Create by role A passes the gate even while role B denies it. `GetUserRowLevelSecurityInfo` then collects role B's leftover `CreateRLSFilterID` anyway, and `GetUserRowLevelSecurityWhereClause` **ORs** the filters together — which widens. The user would create against `F1 OR F2`, where `F2` comes from a row that does not grant them Create at all.

Whether that is reachable depends on whether anything clears the filter when the flag is cleared; `rowFilterValidation.ts` looked like the place to check. Worth confirming before dismissing it, since it is the same multi-role shape as the bug this PR fixes. Not a change request — genuinely out of scope here.

cmbrock-BC:
Thanks — and you are right, the "inert" claim does not hold. I checked rather than argued:

- `GetUserPermisions` does OR the flags across roles (`entityInfo.ts:2957`), so a granting role A carries the user past the gate while a non-granting role B's leftover filter still lands in the OR'd clause. Widening is real.
- Nothing in MJ clears a `*RLSFilterID` when its `Can*` flag is cleared: `rowFilterValidation.ts` validates filter text and same-entity referrers only, and there is no EntityPermission server subclass touching those columns.
- Reachable in practice: on the BlueCypress production copy there are 9 permission rows with a cleared flag and a surviving filter (e.g. `MJ: API Keys` Update/Delete on three BCSaaS roles, `MJ: Conversations` / `MJ: Conversation Details` on the Read Only role). All of them carry "own rows by UserID" filters, so the worst case there is widening to the user's own rows — bounded, but the mechanism is exactly the multi-role shape this PR fixes.

I have corrected the "Not in this PR" section to say so. The symmetric fix is `GetUserRowLevelSecurityInfo` honouring the same `Can*` flag before collecting a filter — a one-condition change per operation plus tests. Happy to open it as a follow-up PR if you want it; keeping it separate as you suggested.

PR #4353, opened 2026-09-10
fix(core): restore the local metadata cache, silently unwritten since 6.1.0-edge.3

cmbrock-BC:
Branch: `fix/metadata-cache-save-survives-getter-throw` (off `origin/next` @ c1fea88930) · commit `e40da79a2c`

## The gap

Nothing fails visibly, but since edge.3 every process boot — MJAPI and the browser alike —
re-reads all metadata from the database instead of the local cache.

Chain, each link verified in source:

1. `BaseInfo.toJSON` serializes a `_`-prefixed backing field through its public getter. edge.3
   changed the lookup to prefer the **PascalCase** getter name (`_categoryPath` → `CategoryPath`);
   before that it looked up the de-underscored key verbatim (`categoryPath`) and so never resolved
   this getter.
2. `QueryInfo.CategoryPath` lazily calls `buildCategoryPath()` → `CategoryInfo` →
   `Metadata.Provider.QueryCategories`.
3. During the **initial** metadata load the global provider is not assigned yet, so that getter
   throws `Cannot read properties of null (reading 'QueryCategories')`, `JSON.stringify` of the
   whole `AllMetadata` snapshot aborts, and `SaveLocalMetadataToStorage` catches it and logs at
   verbose level.
4. `SaveLocalMetadataToStorage` wrote its **timestamps before the payload**. So the cache is left
   claiming freshness with no payload behind it; `LocalMetadataObsolete()` compares timestamps
   only, reports current, and nothing ever retries the save.

`grep -c try baseInfo.js` on the published `6.1.0-edge.5` dist is 0; `next` HEAD is the same.

## The fix — two guards

- **`BaseInfo.toJSON`** wraps the getter read in `try/catch` and omits a key whose getter throws.
  The method's own contract already says computed getters "can throw when source fields are null
  and don't belong on the wire anyway"; that now holds for `_`-backed getters too. The value is
  recomputed lazily on next access, so nothing is lost.
- **`SaveLocalMetadataToStorage`** writes the payload (compressed or the uncompressed fallback)
  first, then the dataset-membership set, then the timestamps **last**. A failed payload write now
  leaves the previous timestamps in place, `LocalMetadataObsolete()` reports obsolete, and the next
  boot retries. Behaviour on the success path is identical (same keys, same formats, same log
  lines); only the order changed. The payload write (compressed, with the uncompressed fallback)
  is extracted into `writeMetadataSnapshot` and the completion log into `logMetadataSaveComplete`,
  so the save method reads as its four steps: serialize, write payload, write membership, write
  timestamps last.

## Evidence

New file `providerBase.metadataCacheSave.getterThrow.test.ts`, 5 tests, **all 5 red on `next`
as-is and green with the fix**:

| test | what it pins |
|---|---|
| `toJSON` omits a throwing getter's key, keeps the rest | guard 1 |
| a real `QueryInfo` with a `CategoryID` serializes while `Metadata.Provider` is unassigned | guard 1, the exact production shape — fails on `next` with the production error text |
| payload is written before the timestamps | guard 2 |
| a failed payload write leaves **no** timestamps behind | guard 2 |
| a snapshot holding such a query is still saved end-to-end | both |

The doubles extend the shared `TestMetadataProvider` and `MockCacheStorageProvider` mocks; the
storage double adds recorded `SetItem` order and a failure on one key, nothing else.

Verified on a fresh `pnpm install --frozen-lockfile` of `next`: `packages/MJCore` builds, all 2316
core tests pass (161 files), `check:esm` and `check:changeset` are green.

## Changeset

`.changeset/metadata-cache-save-survives-getter-throw.md` — `@memberjunction/core: patch`.

## Context

Betty (BlueCypress) found this on its edge.3 upgrade attempt and carries the `toJSON` half as a local
dist patch; this PR is the upstream fix with the ordering half added so the failure class cannot
recur silently.


rkihm-BC (dismissed):
## Approve

Good diagnosis. A four-link chain ending in a cache that claims freshness with nothing behind it is exactly the kind of regression that hides — nothing fails, everything is just slower, forever. `grep -c try baseInfo.js` on the published dist as the confirming evidence is a nice touch.

The ordering fix is the load-bearing half, and it is right. Serialize, payload, membership, timestamps last. I worked through the failure combinations: payload succeeds and membership fails, or payload succeeds and timestamps fail, both leave the old timestamps in place, so `LocalMetadataObsolete()` reports obsolete and the next boot retries. Every partial failure degrades to one wasted boot rather than a poisoned cache. The window where a new payload sits behind old timestamps is benign — a reader either refetches or uses data fresher than it believes.

The refactor preserves the contract you claim: same four keys, same formats, same two log messages, `break` still leaving the proto walk correctly in `toJSON`. Extracting `writeMetadataSnapshot` so the save method reads as its four steps genuinely helps here, because the ordering *is* the fix and it now shows.

### One thing worth a line of doc

The `toJSON` catch silently omits any key whose getter throws. That is strictly better than losing the whole snapshot, and for a computed getter like `CategoryPath` the value is recomputed lazily as you say. But the guard is not limited to computed getters — a `_`-backed getter that throws for some other reason would drop out of the cached payload with no signal, and if its value were not recomputable it would be missing after a warm boot.

I am not asking for a code change; the alternative is worse. Just worth saying in the docblock that the omission is safe *because* these getters are recomputable, so the next person adding a `_`-backed getter knows that is the property being relied on.

cmbrock-BC:
Thanks. Added the line you asked for (a304fe541f pushed): the docblock now says the omission is safe only because a `_`-backed getter is a lazy cache over serialized state and is rebuilt on next access, and warns against adding one whose value cannot be recomputed from the serialized fields. No code change.

rkihm-BC (approved):
## Re-approve

`a304fe541f` is the line I asked for, and it goes further than the ask in the way that matters:

> That omission is safe ONLY because such getters are recomputable: the backing field is a lazy cache over other serialized state (e.g. `QueryInfo.CategoryPath` over `CategoryID`), so the value is rebuilt on the next access after a warm boot. Do not add a `_`-backed getter whose value cannot be recomputed from the serialized fields — a throw would silently drop it.

I only asked for the property to be named. Adding the instruction to the next author is the half that actually prevents the failure, and it sits on the method rather than in a PR thread nobody will re-read.

Nothing else changed since my approval — a doc comment and a `next` merge. The rest of the review stands: the ordering fix is the load-bearing half, every partial-failure combination degrades to one wasted boot rather than a poisoned cache, and the refactor preserves the same four keys, formats and log lines.

Approving on the current head. CI has 9 green with 3 still running; if any go red treat this as not yet in force.

PR #4355, opened 2026-09-10
feat(ng-base-forms): section indicators — unsaved-edits dot and invalid-field count on the form rail and panel headers

### LINKED ISSUE #3962: Form chrome: validation errors are never aggregated to the section/panel that owns the field

MarceloT-BC:
## Summary
A multi-section form now says which section holds unsaved edits or invalid input. Every `mj-collapsible-panel` derives live indicators from the `mj-form-field`s it projects, using the same getters the fields use for their own amber dot and red underline, so a section can never disagree with its fields. The left-nav rail, the More folder, the collapsed rail spine, and accordion headers render an amber dot for unsaved edits and a red count pill for invalid fields.

Closes #3962.

## How it works
- New `FormSectionIndicatorCoordinator`, provided per `mj-record-form-container`. Panels register themselves; the rail pulls counts on each pass, so nothing is written during change detection.
- Panels count a field as invalid when it shows a failure or is required and empty in edit mode. Read-only fields never count, since they render no editor.
- Failed-save errors with a graph-path source such as `Lines[2].Amount` route to the panel that owns the collection, never to the header. Declare `ValidationSources` when the path name differs from the `SectionKey`. Errors no section claims stay in the whole-form total.
- Custom content that is not `mj-form-field` reports through the new `[Indicators]` input, or implements `FormSectionIndicatorSource` and registers with the coordinator.
- The spine total sums the rail groups, not every registered panel, so it always equals what the expanded rail shows.
- The rail's related-grid row count no longer renders at zero.
- Docs: Forms Architecture Guide §7d "Section indicators", `PANELS.md`; patch changeset.

## Repro of the original problem
Open any left-nav form, click Edit, clear a required field in a section that is not active, then look at the rail. Before: nothing on the rail indicates which section fails. After: that rail item shows the red count and the amber dot.

## Verification
- 35 node specs, 22 panel DOM specs, 11 container DOM specs, all green; a mutant that stubbed the field derivation turned 10 of them red.
- Full `ng-base-forms` suite 438 of 438, `ngc` build, spec type-check, and `check:ui` clean.
- Live Playwright on a scratch host: generated left-nav form, 18-panel custom form, new record, rejected save, collapsed spine, accordion header, zero ExpressionChanged errors.

## Side finding, not in this PR
`BaseEntity.Validate()` accepts an empty string for a NOT NULL column, so a required field the form paints red still saves. Logged separately.


rkihm-BC:
Pushed a one-line reword to this branch (`e21f778a13`) so it is unblocked now — revert it if you would rather do it yourself.

**Your code was never the problem.** This file carries no `as any`. `Source guards` was matching the words inside a test title:

```ts
it('is false as soon as any count is positive', () => {
```

`/\bas\s+any\b/` matches "as soon **as any** count". `check-spec-antipatterns.mjs` strips comment bodies before applying its rules — exactly so prose describing an anti-pattern cannot trip the rule banning it — but it never stripped string literals. `Run unit tests` then failed only because it aggregates lane results; all six shards, `Build` and the ESM guard passed.

The reword is "any" → "one", nothing else. Verified the file now reports clean under the current gate.

**The real fix is #4357**, which teaches the lint to skip string literals the way it already skips comments — keeping `${ ... }` inside templates, since `` `${x as any}` `` is a genuine violation. Once that lands this reword is unnecessary and can be reverted; the new wording reads slightly better anyway, so no rush either way.

I have not reviewed the feature itself yet — that is next, now that CI can actually tell us something.

packages/Angular/Generic/base-forms/src/lib/panel/collapsible-panel.component.ts
```diff
@@ -256,13 +318,132 @@ export class MjCollapsiblePanelComponent implements OnInit, OnChanges, AfterCont
...
+    let WarningCount = 0;
+    this.FieldComponents?.forEach((field) => {
+      if (field.IsDirty) DirtyCount++;
+      // Only a field that renders an EDITOR can be invalid: a read-only field renders its
+      // value as text and never paints an underline, so it is skipped even when
+      // `IsRequiredEmpty` would say otherwise (a NOT NULL `__mj_CreatedAt` on a new record
+      // is empty and read-only — the user cannot fix it, and it fills itself on save).
+      if (!field.EditMode || field.IsFieldReadOnly) return;
+      if (field.ShowErrors || field.IsRequiredEmpty) ErrorCount++;
+      else if (field.ShowWarnings) WarningCount++;
```
rkihm-BC: Follow-up, not blocking — this is the one warning the section does not mirror.

The field paints its amber underline on `ShowWarnings || StoredDateIsUnreadable`:

```html
[class.mj-forms-field--has-warning]="ShowWarnings || StoredDateIsUnreadable"
```

and renders a message for the second case on its own. This line counts only `ShowWarnings`, so a date field holding a value the input cannot parse shows its own amber underline and message while the section header and the rail stay clean. That is the single place the "a section can never disagree with its fields" invariant is not actually true.

`StoredDateIsUnreadable` is already gated on `EditMode` and already returns false for an absent value, so it composes with the guard above without further conditions:

```ts
else if (field.ShowWarnings || field.StoredDateIsUnreadable) WarningCount++;
```

Warning-level and narrow, so a follow-up is fine — flagging it because it is a one-line fix that makes the documented invariant literally hold.

packages/Angular/Generic/base-forms/src/lib/section-indicators/form-section-indicator-coordinator.service.ts
```diff
@@ -0,0 +1,124 @@
...
+   * Form-level errors no registered section claims. Surfaced deliberately: an error that
+   * vanished from every badge is worse than the no-badge behaviour this replaced.
+   */
+  public UnroutedValidationErrors(errors: readonly ValidationErrorInfo[] | null | undefined): ValidationErrorInfo[] {
+    if (!errors?.length) return [];
+    const sources = [...this.sources.values()];
+    return errors.filter((error) => {
+      const parsed = ParseValidationSource(error.Source);
+      if (!parsed.Source) return false;
+      return !sources.some((source) => source.OwnsValidationSource(parsed));
```
rkihm-BC, line 129: Follow-up, not blocking — a chrome-dropped section can swallow a failure entirely.

This asks every *registered* source, but the rail only ever reads sources whose key sits in a chrome group. A panel whose key is in no group is still registered (registration happens in `ngOnInit`, independent of chrome), so it claims its plain-field errors here and they are dropped from the unrouted set — while no rail item fronts it, and `isLeftNavVisible` returns `false` for a key in no group, so the panel itself is hidden too. Result: no badge on the rail, no badge on the spine, no badge on the panel, and a rejected save.

The container DOM test "spine totals cover only sections that are on the rail" deliberately locks in the first half of that, and I agree with the reasoning — the spine should equal what the expanded rail shows. The part that looks unconsidered is the interaction with this method: a section excluded from the rail is still allowed to *claim* errors, which is what removes them from the one total that was meant to catch exactly this.

Reachable through `hiddenSectionKeys` on a section that holds an editable required field, which produces the outcome the docblock four lines up calls out as worse than the no-badge behaviour this replaced.

One option is to have the container pass the set of rail-reachable keys and let a source claim only if it is in it; unclaimed-because-unreachable errors then land in the form-wide total, which is where a user can still see them. Not a merge blocker — it is a gap in a new feature, not a regression, since none of these badges existed before.

rkihm-BC (dismissed):
Approving. All 24 checks are green, including all six unit-test shards.

I reviewed the feature itself this time, not just the CI failure I fixed earlier on this branch.

## The central claim holds, for a reason worth stating

The PR asserts a section "can never disagree with its fields." The panel counts a field invalid on `field.ShowErrors || field.IsRequiredEmpty`, which looks like it should double-count a field that is both. It cannot: `IsRequiredEmpty` opens with `if (this.ShowErrors) return false;`, so the two are mutually exclusive by construction, and the count is genuinely per-field as documented. `ShowWarnings` likewise excludes `ShowErrors`, and `fieldIndicators()` mirrors that with `else if`. I went and read the field getters rather than taking the claim from the description, because that invariant is the whole feature.

The read-only skip is right for the same kind of reason. The panel skips on `!field.EditMode || field.IsFieldReadOnly`; the field's editing block is guarded by `@if (EditMode && !IsFieldReadOnly)`. Same two conditions, so a read-only required-empty field renders no underline and contributes no count — exactly what the comment about `__mj_CreatedAt` claims.

## Other things I checked

- The `Changes` subscriber uses `markForCheck()`, not `detectChanges()`, so the pull-not-push argument about `ExpressionChangedAfterItHasBeenChecked` is real rather than asserted.
- `Unregister` is identity-keyed, so a replaced source cannot evict its replacement. Covered by a test that says so.
- `FormIndicators` sums `ChromeGroups`, which includes the More folder whose `SectionKeys` already cover its children — no double-count, and no gap between the spine and the expanded rail.
- Graph sources are never matched on the trailing field name, which is the thing that would otherwise dump every child failure onto the header. Tested directly.
- No `any`, no `.Get()`/`.Set()`, no `@angular/router` import in a Generic package, design tokens throughout the new CSS, and the dead `.is-zero` rule removed along with the markup that used it. Changeset is `patch`, correct for a non-migration branch.

The test suite is doing real work — it tests the negative cases (no dot on a new record, no count on a read-only required field, silence in read mode, graph sources ignored until validation is asked for), which is where this kind of feature usually goes wrong.

## Two follow-ups, neither blocking

Both are inline. Neither loses data, blocks a save, or mis-routes an error, so they should not hold up the build.

One smaller note, not worth an inline: `role="status"` is `aria-live="polite"`, and these spans are created by `@if` rather than existing empty and filling in. A region inserted together with its content is announced unreliably by screen readers, and one field going invalid can fire three of them at once — panel header, rail group, and spine. Worth a look sometime; not worth holding this.

The docs are the right shape too: §7d of the Forms Architecture Guide and the PANELS.md section both explain the derived-not-declared design and the `ValidationSources` escape hatch, and the cross-link resolves.

packages/Angular/Generic/base-forms/src/lib/panel/collapsible-panel.component.ts
MarceloT-BC: Done in e96f78995f — `else if (field.ShowWarnings || field.StoredDateIsUnreadable) WarningCount++`, exactly as you wrote it. Spec added: a saved record with `DueDate = "not a date"` on a datepicker field now reports `SectionWarningCount 1` / `SectionErrorCount 0` and renders the warning badge, with `ShowWarnings` asserted false so the test proves it is this branch and not the validation pipeline.

packages/Angular/Generic/base-forms/src/lib/section-indicators/form-section-indicator-coordinator.service.ts
MarceloT-BC: Done in e96f78995f, the way you suggested. `UnroutedValidationErrors(errors, reachableKeys?)` now only lets a source claim when its key is in the set; the container passes the union of `ChromeGroups` section keys (first-class + More), so a section hidden by `hiddenSectionKeys` or dropped by the chrome cannot claim, and its failure lands in the whole-form total on the spine. Omitting the set keeps the old behaviour for a panel used without a container. Covered at both tiers: a coordinator unit spec (`hiddenExtras` owns `Code`, only `identity` is reachable → `Code` is unrouted) and a container DOM spec (same shape through `FormIndicators` / `UnroutedValidationErrorCount`).

MarceloT-BC:
Both inline follow-ups are in `e96f78995f`, plus the `role="status"` note: the badges are now plain indicators (`role="img"` + `aria-label`), and the container carries one persistent visually-hidden `aria-live="polite"` region whose text is the whole-form summary ("2 fields in this record need attention. 1 unsaved change in this record"), so a change announces once instead of up to three inserted-with-content regions. Spec pins the region text and asserts no `role="status"` remains. ng-base-forms 455/455 on the branch, spec anti-pattern lint clean on all three changed spec files, `check:ui` clean. Thanks for the CI unblock and the read of the invariant.

SDesai-BC:
LGTM

SDesai-BC (commented):
LGTM

cmbrock-BC (approved):

PR #4356, opened 2026-09-10
fix(server): fail a transaction group whose rows are refused server-side

### LINKED ISSUE #4309: A transaction group whose rows are all refused server-side reports success
**Found by** the PR gauntlet on #4305 (`fix/4282-role-elevation-guard`), 2026-09-08. Not caused by
that PR: verified pre-existing at `6638d3825f0fb438ad593567209f8ffefb353923`, where
`MJUserEntityServer` (#4260) already reaches the same path.

### Where
`packages/MJServer/src/resolvers/TransactionGroupResolver.ts:150` (Create/Update) and `:156` (Delete)

### What happens
`ExecuteTransactionGroup` discards the boolean returned by `entity.Save()` and `entity.Delete()`:

```ts
entity.TransactionGroup = tg;
await entity.Save();      // line 150 — return value discarded
...
await entity.Delete();    // line 156 — return value discarded
```

`BaseEntity.Save()` returns `false` without enrolling the row when validation fails — a server-side
entity subclass that refuses in `Validate()` (`MJUserEntityServer`, and now `MJUserRoleEntityServer`
/ `MJRoleEntityServer`) throws its `ValidationResult`, which `_InnerSave`'s own catch converts to
`return false`, so `ProviderToUse.Save(...)` — the only place `TransactionGroup.AddTransaction(...)`
is called — never runs.

When every row in a batch is refused this way, the server's group is empty. `TransactionGroupBase.Submit()`
then takes its empty-group branch (`packages/MJCore/src/generic/transactionGroup.ts:356-362`):

```
// there are no transactions to submit, so we just return true and go back to pending
return true;
```

The resolver treats that as success and calls `PrepareReturnValue(true, …)`, which serialises each
entity's **in-memory, never-persisted** state into `ResultsJSON`
(`TransactionGroupResolver.ts:201`). On the client,
`GraphQLTransactionGroup.HandleSubmit` computes per-item success as
`transactionSucceeded && resultObject !== null` — both true — so `Submit()` returns `true` to the
caller (`packages/GraphQLDataProvider/src/graphQLTransactionGroup.ts:103-111`).

Note the resolver already returns the real reason:
`ErrorMessages: entityObjects.map(e => JSON.stringify(e.LatestResult))` (line 207). The client
discards it.

### What makes it reachable (the precondition)

The discarded boolean is **inert on its own** — `TransactionGroupResolver.ts` has looked like this
since at least 2025-10-27. It becomes a defect only when a row can be refused *on the server* after
being accepted *on the client*.

`BaseEntity.Save()` validates whichever class the LOCAL `ClassFactory` resolves:

- **Ordinary validation** (required fields, field rules) is **symmetric** — the browser holds the
  same generated entity class with the same rules, so a bad row is refused in the browser, never
  enrols, and never reaches this resolver. The client filters it first.
- **A server-only guard breaks that symmetry.** The browser has no such class, its `Validate()`
  passes, the row enrols and travels to the server, and only there is it refused — with nothing
  upstream able to catch it.

So the precondition is an intersection: **a server-only refusing subclass on an entity that a client
enrols in a `TransactionGroup`.** Cross-referencing the seven server subclasses that can refuse in
`Validate()` against every client call site of `CreateTransactionGroup`:

| Entity | Server guard | Enrolled in a client TG? | Live since |
|---|---|---|---|
| `MJ: Users` | `MJUserEntityServer` | yes — `user-management.component.ts:899, 918` | issue #4260 |
| `MJ: User Roles` | `MJUserRoleEntityServer` | yes — `user-management:954`, `user-dialog` | issue #4282 |
| `MJ: Roles` | `MJRoleEntityServer` | **no** — plain `Save()`/`Delete()` | not affected |

`MJ: Roles` is the control that confirms the diagnosis: an equally strict server-only guard, no
silent success, because `role-management` / `role-dialog` call `Save()`/`Delete()` directly and check
the return. The transaction group is the entire difference.

Two consequences for whoever fixes this. First, the fix is worth doing at the resolver rather than
per-surface, because the next server-only guard on a TG-enrolled entity inherits the bug silently.
Second, a client-side check cannot close it — that was attempted in #4305 and could not work, for
exactly the asymmetry above.

### Why it matters
An administrator performs a bulk operation in Explorer that the server refuses in full, and the
screen reports success. Nothing was written, so this is a **false success report, not a security
bypass** — the guard does its job and no row is persisted. But the operator is told the opposite of
what happened, and every server-side entity guard inherits the behaviour: the more MJ adds
`Validate()`-based guards (#4260, #4282), the more often a legitimate refusal is reported as done.
A partially-refused batch is worse: the enrolled rows commit, the refused ones vanish, and the UI
still reports unqualified success.

### Repro
Requires a caller whose `Type` is `'User'` holding `Developer` (which carries unfiltered
`CanCreate` on `MJ: User Roles` on a baseline seed).

```
1. Start MJAPI + Explorer:  pnpm turbo start --filter=mj_api --filter=mj_explorer
2. Sign in to Explorer (http://localhost:4201) as the non-Owner above.
3. Settings -> User Management -> select one or more users -> Bulk Assign Role.
4. Choose a role the caller does NOT hold (e.g. `Integration`) -> confirm.
5. Observe: the screen reports success and clears the selection.
6. Confirm nothing was written:
   SELECT * FROM __mj.vwUserRoles WHERE RoleID = '<the role id>' AND UserID IN (<selected>);
   -> no rows.
```

Equivalent without the UI: submit an `ExecuteTransactionGroup` mutation whose every item is refused
by a server-side guard, and observe `Success: true` with populated `ResultsJSON`.

### Evidence
- Gauntlet run for PR #4305, finding F3. Tribunal scores: reproduce **92**, refute **30** — the
  refuting verifier stated *"The mechanism is real and I traced every hop… what sinks the score is
  provenance, not correctness"*, i.e. it scored low only because the line is outside that PR's diff.
- The adversarial reviewer reached the same conclusion independently (high, confidence 88).
- Lines: `TransactionGroupResolver.ts:150,156,201,207`; `transactionGroup.ts:356-362`;
  `graphQLTransactionGroup.ts:103-111`; `graphQLDataProvider.ts:1979-1990`
  ("part of a TG always return true"); `baseEntity.ts:3759,3811,3881,3883-3915`.

### This violates a rule the repo already documents

`.claude/rules/data-access.md` — the path-scoped rule that loads for every `**/*.ts` file in this
repo — states it directly:

> **Critical**: `BaseEntity.Save()` and `BaseEntity.Delete()` do NOT throw exceptions on failure.
> They return `boolean` — `true` on success, `false` on failure. Error details are available via
> `entity.LatestResult.CompleteMessage` …
>
> **Always** check the boolean return value of `Save()` and `Delete()`
>
> ```ts
> // ❌ WRONG — Don't ignore the return value
> await entity.Save(); // Silent failure — you'll never know it failed
> ```

`TransactionGroupResolver.ts:150` and `:156` are verbatim instances of that documented
anti-pattern. This is not a new standard being proposed — it is one file being brought into line
with one the codebase already states.

Note that the reporting plumbing already exists: `PrepareReturnValue(false, …)` returns
`ErrorMessages: entityObjects.map(e => JSON.stringify(e.LatestResult))`, so the refusal text reaches
the client the moment the `false` branch is taken. Only the branch is missing.

### Suggested fix
In `ExecuteTransactionGroup`, capture each return and fail the whole group when any row was refused,
rather than letting refused rows silently drop out of the group:

```ts
const saved = await entity.Save();
if (!saved) {
    refusals.push({ index: i, entityName: item.EntityName,
                    message: entity.LatestResult?.CompleteMessage ?? 'refused' });
}
```

…then, before `tg.Submit()`, return `PrepareReturnValue(false, …)` with those messages if
`refusals.length > 0`. The resolver is the right place: it is the only layer that can still see
*which* row was refused and why (`LatestResult` is already serialised into `ErrorMessages`), and it
is where the group is assembled. Fixing it in `TransactionGroupBase.Submit()` instead would be
wrong — an empty group legitimately means "nothing to do" for callers that enrolled nothing.

Consider also having `GraphQLTransactionGroup.HandleSubmit` surface `data.ErrorMessages` on failure,
so the reason reaches the UI rather than the generic "all changes have been rolled back".

### Definition of done
- [ ] A failing test that reproduces it, then green
- [ ] A transaction group whose every item is refused by a server-side `Validate()` returns
      `Success: false` with a per-item reason, and the client's `Submit()` returns `false`
- [ ] A partially-refused group also fails rather than committing the surviving rows silently
- [ ] Existing suite and gates green; no changed expectation in an existing test

### Verify by
Re-run the repro above: step 5 must now show an error naming the refused user(s) and the rule, and
`Submit()` must return `false`. `pnpm test` in `MJServer` and the deterministic integration tier
show no new failures.

SDesai-BC:
An administrator performing a bulk operation in Explorer that the server refused **in full** was told it succeeded. Nothing was written that should not have been — the guards did their job — so this is a false success report, not a security bypass. But the operator was told the opposite of what happened, and every server-side `Validate()` guard inherits the behaviour.

Closes #4309

## Root cause

`BaseEntity.Save()` and `Delete()` report a **logical refusal by returning `false`** — they do not throw. A refused row is therefore never enrolled in the group, because `TransactionGroup.AddTransaction(...)` is reached only from inside `ProviderToUse.Save()`/`Delete()`, which a refusal never gets to. `ExecuteTransactionGroup` discarded that boolean:

```ts
entity.TransactionGroup = tg;
await entity.Save();      // return value discarded
...
await entity.Delete();    // return value discarded
```

So the server's group arrived at `Submit()` **empty**, took its legitimate "nothing to do" branch (`transactionGroup.ts`), returned `true`, and the resolver reported `Success: true` while serialising each entity's never-persisted **in-memory** state into `ResultsJSON`. The client's per-item test (`transactionSucceeded && resultObject !== null`) could not tell that apart from a real row, so `Submit()` returned `true` to the caller.

The reason was already on the wire the whole time — `PrepareReturnValue` maps `ErrorMessages` from each entity's `LatestResult`. Only the `false` branch was missing.

## Reproduced live, before and after

Against a real MJAPI, `ExecuteTransactionGroup` with two `MJ: User Roles` creates that both omit the NOT NULL `RoleID`:

| | before | after |
|---|---|---|
| `Success` | `true` | `false` |
| rows written | 0 | 0 |
| `ErrorMessages` | `"Role ID cannot be null"`, unread | same, and now acted on |

And the worse case — one valid row plus one refused, on a disposable entity:

| | before | after |
|---|---|---|
| `Success` | `true` | `false` |
| the valid row | **committed** (DB-assigned GUID, real `__mj_CreatedAt`) | not committed |
| the refused row | vanished silently | reported with its reason |

## What changed

**`TransactionGroupResolver.ts`** — captures what `Save()`/`Delete()` return, logs which items were refused (position, operation, entity, reason), and returns `PrepareReturnValue(false, …)` **before** `tg.Submit()`. Nothing has been written at that point — enrolment is deferral — so a partially refused group fails whole rather than committing the survivors. `PrepareReturnValue` gains no parameter: `ErrorMessages` is already index-aligned with `Items` and already carries the reason.

**`graphQLTransactionGroup.ts`** — copies the server's own failure result onto each item's entity, so `BaseEntity.LatestResult` carries the reason instead of the generic `"Transaction group failed"`. `TransactionResult`'s own docstring already points consumers there, so this needed no new field.

**`TransactionGroupResolver.ts`, second defect at the same lines** — `await entity.GetDataObject()`. The Delete branch pushed the **un-awaited** Promise into `objectValues`, which is `any[]`, so neither `tsc` nor a floating-promise lint could see it. `PrepareReturnValue`'s Delete branch then does `JSON.stringify(objectValues[i])`, and `JSON.stringify(Promise)` is `"{}"` — so **every** Delete in a transaction group reported an empty payload, successful ones included. Worse than losing the payload: `GraphQLDataProvider`'s Delete transaction callback validates the commit with `pk.Value !== results[pk.FieldName]`, so `"{}"` made every key mismatch and reported *"Transaction failed to commit"* for a delete that **did** commit. Awaiting costs nothing here — `DataObjectParams` defaults `includeRelatedEntityData` to false, which guards the method's only awaited work.

**Explorer (`explorer-settings/user-management`)** — the last hop, without which the reason reaches `BaseEntity.LatestResult` and stops there. A new `transaction-group-refusals.ts` helper reads the server's reason back off the rows that enrolled, filtering the placeholder messages `GraphQLDataProvider` stamps on *every* item of a failed group (`'Transaction failed'` for a Save, `'Transaction failed to commit'` for a Delete) so only rows the server actually singled out are shown. `user-dialog.component.ts` and `user-management.component.ts` keep their enrolled rows so those reasons have somebody left to read them, and their `!await tg.Submit()` branches now name which row failed and why instead of a generic "rolled back" sentence. Stale comments describing the old broken behaviour are corrected in both. The coupling to two sibling-package string constants is deliberate, documented on the constant, and pinned by a test against the shipped bundle.

## Two decisions worth defending

**The predicate is the return value, not "is the group empty."** A row that is not dirty also fails to enrol, and `Save()` correctly returns `true` for it. Testing the group's size would turn that legitimate no-op into a failure. This is why the fix is *not* in `TransactionGroupBase.Submit()` — an empty group genuinely means "nothing to do" for a caller that enrolled nothing, which `transaction-groups.TG1` already pins. The resolver is the only layer that still knows *which* row was refused and why.

**The client keys on whether the server sent a reason, not on `Success: false`.** I had this wrong first, and the live run caught it. *Every* item of a failed group reports `Success: false` — `DatabaseProviderBase` registers an entity's result **before** enrolling the row and only flips it to `true` in the transaction callback — so an accepted-then-abandoned row serialises as `Success: false` with an **empty** message. Keying on the flag would have overwritten `BaseEntity`'s own message with a blank one, making the report *worse* for every non-refused row. It now only ever upgrades.

## Evidence

| What | Result |
|---|---|
| `pnpm test` — MJServer | 1274 passed, 56 skipped, 0 failed |
| `pnpm test` — GraphQLDataProvider | 430 passed |
| `pnpm test` — integration-test-suite | 201 passed |
| `mj test run "IT47 - Transaction Group Atomicity"` (fixed server) | 6/6 pass, incl. new TG6 |
| Same, against a **pre-fix** server | **TG6 fails**, TG1–TG5 unchanged |
| `npm run check:standards` / `check:esm` | pass |

`TG6` was verified red-then-green against two real running servers, not just asserted.

**Deterministic integration tier: 62/72 passed, 6 skipped, 4 failed.** All four failures (`ai-embeddings`, `app-wiring`, `codegen-determinism`, `runquery-catalog`) are **pre-existing drift in the shared dev database**, not this change — I confirmed by re-running each against a pre-fix server, where they fail identically. The main one is `codegen-determinism.CD6`: generated entity subclasses register 383 entities while live metadata has 382. None of the four executes any code this PR touches.

## Still open

- The tier is not fully green in my environment for the reasons above. Worth a re-run on a clean database before merge.
- **A separate defect at the same lines, deliberately not fixed here.** `await entity.InnerLoad(pkey)` also discards its boolean. When an `Update` targets a row that does not exist, `InnerLoad` returns `false` without setting `_everSaved`, so `Save()` treats it as a **Create** — an Update silently becomes an Insert. Different failure mode, needs its own repro and test, and fixing it would break any caller relying on upsert-via-Update. Should be its own issue.
- No test expectation was changed. Five existing client fixtures gained a `RegisterResultHistoryEntry` stub, which is fixture drift, not an assertion change.

## Risk

No migration, no metadata, no scheduled job — nothing to apply outside the code, and no deploy ordering. Behaviour change to note: a transaction group that previously reported success while refusing rows now reports failure, so any caller that was silently relying on the false success will start seeing errors. That is the point of the change, but it is a visible difference.

https://claude.ai/code/session_01EKsmdy44CDZaVEVW6KL2pM


rkihm-BC (dismissed):
## Approving. The root cause is exactly right, and the safety argument holds under inspection.

`Save()`/`Delete()` reporting a logical refusal by returning `false` rather than throwing, combined with enrolment happening only *inside* `ProviderToUse.Save()`, is a genuinely non-obvious chain — and "the group arrived at `Submit()` empty and took its legitimate nothing-to-do branch" is the kind of bug that reads as correct in every individual function. Good find.

### What I verified rather than took on trust

**"Nothing has been written at this point."** This is the claim the early return rests on, so I checked it: the transaction-group path in `databaseProviderBase.ts` builds the SQL, calls `AddTransaction(...)`, and then `return true` — no execution. Enrolment really is pure deferral, so returning before `Submit()` leaves the database exactly as found. The comment saying so is accurate.

**The predicate.** Keying on the return value rather than group-emptiness is right, and the defence is the important half: a non-dirty row also fails to enrol and correctly returns `true`, so a size test would convert a legitimate no-op into a failure. Keeping it out of `TransactionGroupBase.Submit()` is right for the reason given — the resolver is the only layer that still knows *which* row was refused.

**The client-side predicate, including the part you got wrong first.** Keying on whether the server sent a reason rather than on `Success: false` is correct, because every item of a failed group reports `false`. Writing down that you had it inverted, and that a live run caught it, is more useful than a clean narrative would have been — the failure mode it avoids (blanking `BaseEntity`'s own message for every non-refused row) is subtle enough to have shipped.

**The missing `await`.** I traced this one end to end because it deserves the billing: `GetDataObject` is `public async`; `next` pushed the un-awaited Promise into `objectValues`, typed `any[]` so nothing could see it; `PrepareReturnValue`'s Delete branch does `JSON.stringify(objectValues[i])`; `JSON.stringify(Promise)` is `"{}"` (confirmed by running it); and `GraphQLDataProvider` validates the commit with `pk.Value !== results[pk.FieldName]`, which `"{}"` always fails. So every Delete in a transaction group reported an empty payload **and** a false *"Transaction failed to commit"* for a delete that committed. That is arguably a worse bug than the titled one.

**The Explorer half.** The enrolled-rows pattern is the piece that makes the whole chain reach a human — without it the reason lands on `LatestResult` and nobody reads it. Filtering the provider's placeholder messages so only rows the server actually singled out are shown is the right call, and documenting the cross-package string coupling on the constant rather than hiding it is the honest way to carry that debt.

Correcting the stale comments in both components — the ones describing the old broken behaviour — matters more than it looks. Those comments were accurate descriptions of a bug; left alone they would have misled the next reader indefinitely.

### One nit

The drift guard that protects those coupled strings can only half-fail:

```ts
expect(bundle).toContain('Transaction failed');
expect(bundle).toContain('Transaction failed to commit');
```

`toContain` is a substring match, and `'Transaction failed to commit'` contains `'Transaction failed'` — so the first assertion passes as long as the second string exists anywhere, whether or not the provider still emits the Save placeholder. That is precisely the drift the guard was written to catch, and its own comment describes the consequence: the placeholder stops matching and starts being shown to operators as though it were the server's reason.

Both strings are currently exact, so nothing is broken. A quoted-literal match (`toMatch(/'Transaction failed'/)`) closes it, or better — export the two constants from `GraphQLDataProvider` and import them, which removes the duplication instead of policing it.

### Housekeeping

I edited the description: removed the "Draft — not ready for review yet" line (it contradicted the PR's non-draft state and was keeping it out of the queue), and added the two changes it omitted — the `GetDataObject` defect and the Explorer `user-management` work, together about half the diff. Reword either freely.

The `InnerLoad` item from "Still open" is now **#4446**, with the chain verified: `InnerLoad` returns `false` without setting `_everSaved`, `IsSaved` reflects that, `databaseProviderBase.ts:1470` reads `const bNewRecord = !entity.IsSaved`, and line 1481 turns it into a `create`. Your reasoning for keeping it out of this PR is recorded there.

### On the red Integration check

Not attributable to this PR. It fails in the `mj sync push --ci` setup step before any test runs, with `waitForActiveRequest: timed out after 2000ms` → `Can't commit transaction. There is a request in progress.`, and the stack is in `GenericDatabaseProvider`'s `WithTransactionLock` — a package this PR does not touch. The identical failure is hitting `mjc/metadata-cache-entry-cap` and `mjc/pg-fts-residue` right now, neither of which goes near transaction groups. Worth a re-run before merge, but it is not yours.

Everything else is green at 23 passing.

SDesai-BC:
Thanks — and the nit was real. Fixed in d76cd253f1, with one correction to the suggested remedy.

## Confirmed, and reproduced before fixing

`toContain` is a substring match and `'Transaction failed to commit'` contains `'Transaction failed'`, so the shorter assertion passed on the strength of the longer string alone. Written as a failing test first:

```
emitsLiteral('s.Message="Transaction failed to commit"', 'Transaction failed')
  -> true     (AssertionError: expected true to be false)
```

Same shape as the TG6 defect this PR already fixes — an assertion that cannot fail is worse than no assertion, because it reads as covered.

## Refuted: the quoted-literal form as written would break

`toMatch(/'Transaction failed'/)` returns **false** against the real artifact. `require.resolve` lands on `dist/index.cjs`, which is minified and double-quotes the literal:

```
)):(s.Success=!1,s.Message="Transaction failed")})),!0}else{const p=awai
!0:(s.Success=!1,s.Message="Transaction failed to commit")}else s.Succes
```

Pinning the single-quoted form the provider *source* uses would have converted a vacuous guard into a permanently red one. The fix matches a complete quoted literal quote-agnostically (`["'\`]…["'\`]`), which is true against this bundle and false against one carrying only the longer placeholder — both directions now pinned by tests.

## Went one step further

The pin now iterates `PROVIDER_PLACEHOLDER_MESSAGES` itself instead of re-typing the two strings, so the test can no longer drift from the filter it guards, and a third placeholder gets pinned by being added rather than by someone remembering. Only provider → filter drift remains, which is what reading the bundle is for.

## Declined for now: exporting the constants

You're right that it removes the duplication instead of policing it, and it's the better end state. I left it out because it adds a public export to a second package after approval, and the provider holds these as inline literals at three call sites in `graphQLDataProvider.ts` — a refactor that deserves its own commit and its own review rather than riding on a review nit. Happy to do it as a follow-up if you'd prefer it in this PR.

## On the red Integration check

Independently reached the same conclusion before seeing your note, with one extra data point: that exact signature (`waitForActiveRequest` timeout → `Can't commit transaction. There is a request in progress.`) hit **five** distinct branches in twelve hours — this one twice, plus `mjc/sql-dialect-from-provider`, `JF_Native_Tool_Calling`, `mjc/metadata-cache-entry-cap` and `mjc/pg-fts-residue`. This branch also passed that same check 17 minutes before the merge commit. Agreed it needs a re-run rather than a fix.

Thanks for filing #4446 with the chain verified, and for the description fix — the "Draft" line was stale and shouldn't have survived.

`explorer-settings` 139 passing (was 137); `check:standards`, `check:esm`, `check:ui`, `check:changeset` all clean.

rkihm-BC (approved):
## Re-approving. The nit is closed correctly — and you caught an error in my suggested fix.

### First, the correction I owe you

My review proposed `toMatch(/'Transaction failed'/)` — pinning the single-quoted form the provider *source* uses. I ran it against the real `dist/index.cjs` before writing this:

```
excerpt from the bundle:  (p)):(s.Success=!1,s.Message="Transaction failed")})),!0}el
my remedy  /'Transaction failed'/   ->  false
```

The bundle is minified and double-quotes the literal, exactly as you said. Applied verbatim, my suggestion would have converted a guard that could never fail into one that could never pass. I proposed a fix against the source without reading the artifact the test actually reads — and you caught it by writing the failing test first, which is the standard I have been asking of everyone else. Noted, and appreciated.

### What I verified by execution

`emitsLiteral` transcribed from `d76cd253f1` and run:

| case | result |
|---|---|
| bundle carrying **only** `"Transaction failed to commit"`, asked for the short literal | `false` — the drift the old guard missed is now caught |
| same bundle, asked for the long literal | `true` |
| the **real** minified `dist/index.cjs`, short literal | `true` |
| the real bundle, long literal | `true` |

Both directions hold, and the quote-agnostic `["'\`]…["'\`]` is the right shape: it requires a quote immediately after the literal, so `"Transaction failed to commit"` cannot satisfy the short pin because the next character there is a space. The negative test you added pins that reasoning so it cannot regress silently.

### The rest of the commit

- **Iterating `PROVIDER_PLACEHOLDER_MESSAGES`** instead of re-typing the strings is the better design — a third placeholder is pinned by arriving, not by someone remembering. That leaves only provider → filter drift, which is precisely what the bundle read is for.
- **The export is module-internal.** I checked `public-api.ts` and `index.ts`: `transaction-group-refusals` is not re-exported, so this adds nothing to the package's public surface. The concern you raised about adding a public export to `GraphQLDataProvider` does not apply on this side.
- **Declining the cross-package constant export** is the right call for the reason you gave — three inline literals at three call sites in `graphQLDataProvider.ts` is its own refactor with its own review, not a rider on a nit. Happy to see it as a follow-up; not needed here.
- The merge from `next` did not touch any file this PR changes. The description edits are intact. CI is 24 passing, 0 failing.

### On the Integration check

Your count — the same signature on five branches in twelve hours — is a stronger dataset than mine and corroborates #4447. That is no longer a re-run situation: the root cause was a commit race in `SQLServerDataProvider` (a 2-second poll that gave up silently, then a `finally` that nulled the handle before abandon could roll it back), fixed in #4448 by draining the instance SQL queue before commit. Two Integration runs on that branch have come back with the signature fully absent. Once it merges, this class of red check should stop appearing on unrelated PRs.

`#4446` (the `InnerLoad` upsert defect) remains the one open item from the original review, deliberately out of scope here.

No changes requested.

PR #4357, opened 2026-09-10
fix(ci): stop the spec anti-pattern lint matching inside string literals

rkihm-BC:
The spec anti-pattern lint fails PRs for prose. `#4355` carries no `as any` and cannot merge.

## The gap

`scripts/check-spec-antipatterns.mjs` strips comment bodies before applying its rules, for the reason its own tests state — prose describing an anti-pattern must not trip the rule banning it:

```js
it('strips line comments so prose about an anti-pattern does not trip a rule', ...)
```

String literals never got the same treatment. So a test **title** is enough to fail the gate:

```ts
it('is false as soon as any count is positive', () => {
```

`/\bas\s+any\b/` matches "as soon **as any** count" inside the description. That is the whole of the `Source guards` failure on #4355 — the file contains no actual `as any`, and `Run unit tests` fails only because it aggregates lane results. All six shards, `Build` and the ESM guard pass there.

The same exposure exists for `: any` and `as never` in any sentence containing them.

## The fix

`stripComments` becomes `stripCommentsAndStrings`, blanking single-quoted, double-quoted and template-literal contents alongside comment bodies.

Three things it deliberately keeps:

- **`${ ... }` inside a template literal.** That is code, and `` `${x as any}` `` is a real violation.
- **Backslash escapes.** `'it\'s as any'` is one string, not two, so an apostrophe cannot end a literal early and expose the rest of the line.
- **Line count and order**, so reported line numbers stay accurate.

An unterminated template carries its state to the next line, the way the existing block-comment handling already does.

## Evidence

Eight regression tests, including the verbatim #4355 line, the template expression that must **still** flag, the escaped-quote case, and a multi-line template.

Verified directly rather than by inference:

| check | result |
|---|---|
| fixed gate against #4355's actual test file | clean |
| fixed gate still flags `const x = y as any;` | flags |
| `node scripts/check-spec-antipatterns.mjs packages` on `next` | clean, graced count unchanged at 259 |

So this unblocks #4355 with no change to that PR, and the graced allowlist neither grows nor shifts.

## No changeset

The branch touches `scripts/` and `.github/scripts/` only and ships no package code, so there is nothing to version. The nearest precedent (#4325) attached a nominal `@memberjunction/core: patch` to a CI fix; say the word if you would rather match that and have this appear in the release notes.

### DIFF
--- a/.github/scripts/__tests__/check-spec-antipatterns.test.mjs
+++ b/.github/scripts/__tests__/check-spec-antipatterns.test.mjs
@@ -1,17 +1,57 @@
 import { describe, it, expect } from 'vitest';
-import { lintText, stripComments, hasAdjacentKnownLimitation, ALLOWLIST } from '../../../scripts/check-spec-antipatterns.mjs';
+import { lintText, stripCommentsAndStrings, hasAdjacentKnownLimitation, ALLOWLIST } from '../../../scripts/check-spec-antipatterns.mjs';
 
 const whys = (text, kind) => lintText(text, kind).map((f) => f.why);
 
-describe('stripComments', () => {
+describe('stripCommentsAndStrings', () => {
     it('strips line comments so prose about an anti-pattern does not trip a rule', () => {
         expect(whys(`// mentioning expect(true) or as any here is fine\nexpect(x).toBe(1);\n`, 'node')).toEqual([]);
     });
 
     it('strips block-comment interiors line-wise', () => {
-        const stripped = stripComments(['/* expect(true)', 'still comment as any', '*/ expect(x).toBe(1);']);
+        const stripped = stripCommentsAndStrings(['/* expect(true)', 'still comment as any', '*/ expect(x).toBe(1);']);
         expect(stripped.join('\n')).not.toContain('expect(true)');
         expect(stripped.join('\n')).toContain('expect(x).toBe(1);');
+    });
+});
+
+describe('stripCommentsAndStrings — string literals are prose too', () => {
+    it('does not flag `as any` inside a test title (the #4355 false positive)', () => {
+        // Verbatim from packages/Angular/Generic/base-forms — "as soon AS ANY count" tripped the rule.
+        expect(whys(`it('is false as soon as any count is positive', () => {});`, 'node')).toEqual([]);
+    });
+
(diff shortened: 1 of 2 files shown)


rkihm-BC:
## Adversarial self-review — found a regression in my own first commit, fixed in `8d6c91f631`

I wrote this, so I went looking for the way it makes things worse rather than better. It did, in one specific way, and the direction matters: **it traded a false positive for a false negative.**

`skipQuoted` returned end-of-line when a quote never closed, and the caller treated everything after it as string content. One stray apostrophe therefore blanked the rest of the line:

```
const re = /don't/; const x = y as any;    →    const re = /don
```

That line **flagged correctly before this branch** and stopped flagging with it. An apostrophe inside a regex or a word is not rare in a test file, so this was a live hole, not a theoretical one — and a gate that silently stops catching things is worse than one that occasionally shouts.

The fix rests on a property of the language: a single-quoted or double-quoted literal cannot span a line, so an unterminated quote was never an opening. `skipQuoted` now returns `-1` for that case, the caller emits the character and keeps scanning, and everything after it is still linted.

**Template literals deliberately keep the opposite treatment**, because they genuinely do span lines. An unpaired backtick still carries state forward and can mask findings after it. I have written that up as a `KNOWN LIMITATION` on the function rather than leaving it implicit: err toward a false negative only where the language forces it.

## Where it stands now

| check | result |
|---|---|
| adversarial cases, two rounds | 18/18 |
| `node scripts/check-spec-antipatterns.mjs packages` | clean, graced count unchanged at **259** |
| #4355's file **before** the reword | clean |
| real violations still flagged | yes, including `` `${x as any}` `` |

Eleven regression tests now, the three new ones being the regex apostrophe, an escaped backslash closing the literal so what follows is scanned, and a comment marker inside a string not starting a comment.

## What I could not close

The unpaired-backtick case above. Fixing it properly needs real tokenization, which is more machinery than a lint of this size earns. It is strictly narrower than the bug being fixed here — an unpaired backtick in a test file is far rarer than the words "as any" in a sentence — but it is a genuine hole and I would rather it be written down than discovered later.

The `${ ... }` depth counter is also brace-counting rather than parsing, so `` `${ "}" }` `` would end the expression early. Same reasoning: acceptable for a lint, worth knowing.

PR #4358, opened 2026-09-10
fix(core): RLS filter collection honours the permission row's Can* flags

cmbrock-BC:
# fix(core): RLS filter collection honours the permission row's Can* flags

Follow-up to #4352, from its review. Same shape, other half of the same function pair.

## The gap

`GetUserRowLevelSecurityInfo` collected a row filter from any permission row that carried one:

```ts
case EntityPermissionType.Create:
    if (ep.CreateRLSFilterID) matchObject = ep.CreateRLSFilterObject;
```

The caller (`GetUserRowLevelSecurityWhereClause`) ORs the collected filters together, so a filter
picked up from a row whose `Can*` flag is **false** widens the clause. The reachable shape:

- role A grants Create, bound to `F1`;
- role B has `CanCreate = false` but still carries `CreateRLSFilterID = F2` — nothing in MJ clears
  the filter column when the flag is cleared (`rowFilterValidation.ts` validates filter text and
  same-entity referrers only; no EntityPermission server subclass touches these columns);
- `GetUserPermisions` ORs the flags across roles (`entityInfo.ts:2957`), so the user passes the
  permission gate on A alone — and then creates against `F1 OR F2`.

It is reachable in practice: the BlueCypress production copy has 9 such rows (`MJ: API Keys`
Update/Delete on three BCSaaS roles, `MJ: Conversations` / `MJ: Conversation Details` on the
Read Only role, plus Read leftovers on `MJ: Credentials` and one app entity). All carry
"own rows by UserID" filters, so the exposure there is bounded to the user's own rows — but the
mechanism is exactly the multi-role shape #4352 fixed for exemption.

## The fix

Four one-line condition changes: a row contributes a filter only for an operation it grants
(`ep.CanCreate && ep.CreateRLSFilterID`, and likewise Read/Update/Delete). Granting rows are
unchanged, so two granting roles still OR their filters. A user whose only row does not grant the
operation now collects nothing — and has no permission either, which the tests assert directly.
JSDoc on the method explains the rule and the widening it prevents.

## Evidence

New file `entityInfo.rlsFilterCollection.test.ts`, 5 tests. **3 fail on `next` as-is**
(leftover ignored on Create; on Update and Delete; sole non-granting row collects nothing) and
2 pin unchanged behaviour (a read-only row still contributes its Read filter; two granting rows
still OR). Kept in its own file with local builders so it does not conflict with #4352's edits to
`entityInfo.rlsExemption.test.ts`; both files pass together (34 tests).

Verified on a fresh `pnpm install --frozen-lockfile` of `next`: `packages/MJCore` builds, all 2316
core tests pass (161 files), `check:esm` and `check:changeset` are green.

## Changeset

`.changeset/rls-filter-collection-honours-can-flags.md` — `@memberjunction/core: patch`.

## Relationship to #4352

Independent of it at the code level (different method, different test file); the two together
make the pair consistent: a non-granting row neither exempts (#4352) nor contributes a filter
(this PR). Either merges cleanly without the other.


rkihm-BC (commented):
## Review: RLS filter collection honours the permission row's `Can*` flags

**Verdict: looks good to merge.** Small, central, correct, and the evidence in the description holds up.

### Verified

- **Single point of enforcement.** `GetUserRowLevelSecurityInfo` has exactly one production consumer (`entityInfo.ts:3088`), and there is no T-SQL or provider-side duplicate of the collection logic anywhere in `packages/`, `migrations/`, or `SQL Scripts/`. The fix can't be half-applied.
- **The evidence reproduces.** I ran the new test file against `next` as-is: **3 failed / 2 passed**, exactly as described.
- **Symmetry with #4352 is real** — `UserExemptFromRowLevelSecurity` already carries the same `Can*` guard on `next` (`entityInfo.ts:3006-3020`). The pair is now consistent.
- **The fail-open concern is covered.** An empty clause means "unrestricted" at every call site, so this only holds if the permission gate always runs first. It does: `CheckUserReadPermissions` fires on both the cache-check and non-cache RunView paths (`GenericDatabaseProvider.ts:1750`, `:2538`), `BaseEntity.CheckPermissions` gates Create/Update/Delete (`baseEntity.ts:4065`, `:4987`), and `SearchEngine.filterEntityResults` checks `perms.CanRead` explicitly before reaching for the clause.

### Nits (none blocking)

1. **`Type = 'Deny'` rows aren't considered, and the new JSDoc says they are.** On a Deny row, `CanCreate = true` means *deny create*, so `ep.CanCreate && ep.CreateRLSFilterID` reads a deny flag as a grant. I chased this and **it is not reachable** — any Deny row that sets the flag forces `userPermission.CanCreate = false` in `GetUserPermisions`, so the clause never runs. The only defect is that the method now documents an invariant ("only permission rows that GRANT the operation") it doesn't implement. Either add `(ep.Type || 'Allow').trim().toLowerCase() !== 'deny' &&`, or a one-line comment saying Deny is handled upstream. Same applies retroactively to #4352.

2. **Test builders bypass the type system.** `buildPermission(overrides: Partial<Record<string, unknown>>)` means a typo'd override key silently does nothing. `Partial<EntityPermissionInfo>` costs nothing and is what `.claude/rules/typescript-style.md` asks for. Relatedly, every fixture hardcodes `Type: 'Allow'`, so the Deny axis is untested.

### Follow-ups worth an issue (out of scope here)

- **Nothing stops the state recurring.** This PR makes the leftover harmless at read time, but the rows stay and the UI can still create them. `rowFilterValidation.ts` only validates filter text and same-entity referrers, and there is **no `EntityPermission` server subclass** at all. A subclass under `MJCoreEntitiesServer/src/custom/` that nulls `<Op>RLSFilterID` when `Can<Op>` is cleared would be the durable fix — `MJAISkillPermissionEntityServer.server.ts` is the closest existing precedent. Worth pairing with a cleanup for the 9 rows you found.
- **Comment drift in two neighbours.** `MaterializationRefresher.ts:831` and `manage-metadata.ts:6585` both say their predicate "matches `GetUserRowLevelSecurityWhereClause`, which sources the read filter solely from `EntityPermission.ReadRLSFilterID`" — no longer true after this. Behaviour is unaffected and fail-*closed* (the leak gate stays conservative), so it's purely a comment fix, and not in files this PR touches.
- **No integration-tier check**, though `rls-isolation.checks.ts` is the natural home.

rkihm-BC (approved):
Approving — see my detailed review above. Verified independently: single enforcement point with no SQL-side duplicate, the new tests reproduce 3 fail / 2 pass on `next`, and the permission gate provably precedes every RLS call site, so the "empty clause means unrestricted" direction is safe.

The two nits (Deny rows vs. the new JSDoc's "rows that GRANT" wording, and `Partial<Record<string, unknown>>` in the test builders) are non-blocking — fold them in here or not, your call. The `EntityPermission` server subclass to stop the stale-filter state recurring is worth its own issue.

PR #4362, opened 2026-09-10
release-prep(v6.1.x): metadata sync, PostgreSQL counterparts, and the core-entities validator restoration

SDesai-BC:
## What

Release-content prep for the v6.1.x Edge stream, produced and verified against a **clean-room database built only from what the repo ships** (fresh clone → build → `mj migrate` → `mj codegen --skipfiles` → `mj sync push` → `mj codegen --skipdb`).

Three commits, in dependency order:

1. **`chore(migrations)`** — the `Metadata_Sync` migration + the nine missing PostgreSQL counterparts
2. **`fix(integration-test-suite)`** — IT56/PG7 compliance predicate
3. **`fix(core-entities)`** — restore 56 generated `Validate()` overrides ⚠️ **read this one**

## ⚠️ The regression this PR stops shipping

`v6.1.0-edge.5` shipped `MJCoreEntities/src/generated/entities/__mj.ts` with **56 `Validate()` overrides across 46 entity classes**. The committed file on `next` has **zero**.

| Ref | overrides |
|---|---|
| `v6.1.0-edge.5` (last release) | **56** |
| `0654f69462` (SubtypeSelector) | 57 |
| **`197fdf8376`** — *100% CodeGen idempotency… churn elimination* | **0** ← removed all 57 |
| `9ef9321847` — *eliminate fresh-DB drift* | 0 (did not restore) |
| `next` | **0** |
| this PR | **56** |

Releasing `next` as-is publishes `@memberjunction/core-entities` with entity validation **silently off** for those classes — a regression for everyone upgrading from edge.5, with no changelog entry, because the removal never had one. `197fdf8376` rewrote the file (+25,689/−28,635) and mentions neither *validate* nor *override* in its message or changeset.

**This is not clean-room drift — the polarity is the other way round.** Diffing the validator *class sets* between the edge.5 tag and this regeneration returns **nothing**: identical set. The diff here is **purely additive — 2,854 insertions, 0 deletions**.

Not included: `MJFormChromeRuleEntity`'s validator — added after the edge.5 tag and removed before any release, so it has never shipped. It also cannot be regenerated on a fresh install: validator bodies come from `GeneratedCode` records seeded by migrations, and none seeds that one (the clean-room DB holds 124 such rows, none for FormChromeRule). Filed separately rather than hand-patched into generated output.

## Migrations + PostgreSQL counterparts

L1 parity for `migrations/v6` is now complete. Six counterparts via `mj migrate convert --split --bake-codegen`, each baked against a freshly re-migrated database so no bake reads a stale view; `Metadata_Sync` via the **legacy** converter (the split path emits a reseed marker) — 4,845 lines, 130 `PERFORM` calls, self-seeding; three hand-authored where the AST reported gaps (CTE-`DELETE` → windowed sub-select; `EXEC` → `PERFORM` with the signature verified against `pg_proc`; `sys.default_constraints` probe → `ALTER COLUMN SET DEFAULT (NOW() AT TIME ZONE 'UTC')`); one declared `PG-EMPTY-BY-DESIGN`.

**`Entity_SubtypeSelector.pg.sql` additionally recreates `vwEntities`.** `MJ: Entities` is app-owned (`BaseViewGenerated = false`), so CodeGen never rewrites it and PostgreSQL has no `sp_refreshview`. Without it the new column never reaches the read path and `spDeleteUnneededEntityFields` deletes the `EntityField` row again. Invisible to `mj migrate`; it surfaced only in the `mj sync push` half of the gate.

## IT56/PG7

PG7's compliance phase watched only the *restricted* write while its assertion read the *allowed* one, so a model doing half the script passed compliance then failed `notes.a did not land` — a bare assertion that reads as an engine defect and blocks a release under §4.6.

It is not an engine defect. Driving the agent directly, outside the suite so its cleanup could not delete the evidence, a fully compliant response yields `FinalPayload {"notes":{"a":"IT-NOTES-OK"}}` with `config` correctly absent — partial application works as designed. Gating on both markers makes partial compliance report `model-noncompliance:` like the bundle's other eight checks.

## Verification

```
mj migrate (fresh PG)   86 applied, 0 failed
mj sync push            84/84 steps, 0 errors
L1 parity               0 missing counterparts
content gate            259 checked, 0 suspect · delete parity 0 mismatched
L2                      tables 385 = 385 · foreign keys 812 = 812
L4 CRUD oracle          353 pass / 16 fail — all harness artifacts or pre-existing
suite unit tests        201/201
core-entities build     clean
full stack              MJAPI + Explorer served 218 MJ: AI Models rows from the
                        converted PG database — matching SQL Server exactly
```

`created: 0` on the sync push proves the `Metadata_Sync` counterpart is genuinely self-seeding: on SQL Server the same push had to create 35 records.

## Known, filed, not blocking

- 3 base views + 18 CRUD routines exist on SQL Server but not PostgreSQL — verified pre-existing (absent from a database built only from previously-committed counterparts).
- `scripts/ss-pg-view-equivalence.mjs` hardcodes `migrations/v5`, so L3 cannot gate a v6 release; 43 of its 47 `realDiffers` are views a v6 migration re-creates.
- The live tier's Active bindings span 9 vendors; 4 have no working credentials, and a credential-less vendor **halts failover** rather than skipping. Live-tier results are unreliable until that is provisioned.


rkihm-BC (approved):
LGTM

PR #4365, opened 2026-09-10
Skills: a skill can bundle an action without offering it to the model (AISkillAction.ExposeToModel, #4226)

### LINKED ISSUE #4226: Skills: a skill's granted actions are always exposed to the model — no way to hold an action for code-only use

cmbrock-BC:
# A skill can bundle an action without putting it into the agent's run (#4226)

Closes #4226.

## Summary

Bundling an action into a skill (an `MJ: AI Skill Actions` row) puts the action into the activating agent's run: `BaseAgent.enableSkillCapabilities` pushes every bundled ActionID into `params.actionChanges`, and from there it is described to the model *and* executable for the rest of the run.

A skill whose reply carries a menu — buttons the application wires to an action, pressed by the person on the next turn — wants the association (SKILL.md export, tooling) without the model ever being able to call the action on its own mid-conversation. Betty hit this with exam-creation skills whose "Create exam" / "Download" buttons are application-invoked actions.

## What the flag does — precisely

`AISkillAction.ExposeToModel` (BIT NOT NULL DEFAULT 1). `1` is today's behaviour. `0` keeps the row (SKILL.md export, tooling) but leaves the action **out of the run**: the effective action set is one list that is both the prompt's tool surface and the execution allow-list, so the action is neither described to the model nor executable by the agent — the `Actions` step, a `PreProcessActionStep`, and a loop's action lookup all resolve against that set and reject it.

That is a deliberate choice over "hide it from the prompt only": a prompt-only hide would still let the model call the action by name the moment a skill's Instructions mention it, which is the hole #4226 describes. Consequences stated plainly: bundling grants nothing on the Actions API path (application code invokes the action there, under that API's own authorization), and skill attribution on the Actions step cannot apply to an action the agent never runs.

## Changes

- **Migration** `V202609102050__v6.1.x__Skill_Action_Expose_To_Model.sql`: the column, extended property, and the CodeGen tail (EntityField row with the apply-time `COALESCE(MAX([Sequence]),0)+1` expression, regenerated view/procs/permissions, field properties). Additive and defaulted.
- **`AIEngineBase.GetSkillExposedActionIDs(skillID)`** (+ `AIEngine` facade): the `ExposeToModel` subset. `GetSkillActionIDs` unchanged.
- **`BaseAgent.enableSkillCapabilities`** pushes the exposed subset.
- **SKILL.md portability**: optional `codeOnlyActions` frontmatter list (names, subset of `actions`). Export writes it only for skills with `ExposeToModel = 0` rows, so older files are byte-identical; import applies it. A file without the key expresses no opinion, and `resyncJunction` (delete-and-recreate) then carries each surviving row's flag across — so re-saving a skill's SKILL.md never turns a code-only action model-callable. The parser also skips unknown *list* keys instead of rejecting the file, so a newer file parses on an older MJ.
- **Integration check** `ai-skills.AS4` asserts `GetSkillExposedActionIDs` ⊆ bundled and contains the default-flag row.
- **Guide** `AGENT_SKILLS_AND_PLAN_MODE_GUIDE.md` §1.1 / §1.3 / §1.7 / §3, the column description, migration header, code comments and changeset all describe the behaviour above — none of them say "grant" or "attribution".
- CodeGen output for the column: entity + zod (`core-entities`), GraphQL type/inputs (`server`), the generated form field (`ng-core-entity-forms`). Changeset lists all five packages plus `server`.

## Testing

- **Real `BaseAgent`** (`skill-expose-to-model-agent.test.ts`): `enableSkillCapabilities` pushes only the exposed subset as a specific-scoped ActionChange; a skill whose every action is code-only pushes **no** ActionChange; sub-agents unaffected. Fails 3/3 when the call is reverted to `GetSkillActionIDs`.
- `skill-import-resync.test.ts`: explicit `codeOnlyActions` intent outranks a surviving row's old flag; without the key the flag is carried; sub-agent rows untouched.
- `SkillMarkdownConverter.test.ts`: `codeOnlyActions` parse/serialize round-trip, absent-vs-empty, unknown list keys skipped.
- `BaseAIEngine.test.ts`: exposed subset, all-hidden → `[]`, missing flag reads as exposed.
- Packages: `ai-agents` 2007/2007, `ai-engine-base` 206/206; `turbo build` 173 tasks incl. `integration-test-suite`. Migration gates (EntityField sequence, id determinism, codegen tail, changeset bump) pass locally.
- Drift gate: the committed generated DisplayName is `Expose to Model` — the migration's category block sets that after the EntityField INSERT's naive `Expose To Model`, and CodeGen reads the final row. Note the gate is also red on `next` itself since #4362 (21:02Z, same file); this PR's own mismatch is fixed regardless.


rkihm-BC (commented):
## Review: `AISkillAction.ExposeToModel`

The migration and the plumbing are solid. Two things need attention before merge: **the CodeGen drift gate is failing** (diagnosed below), and **the guide this PR ships describes behaviour the code doesn't have**.

---

### Blocker — the drift gate failure, and its cause

The gate finished red:

```
```

The cause is a DisplayName captured between CodeGen's two phases. The migration's own category block sets:

```sql
-- UPDATE Entity Field Category Info MJ: AI Skill Actions.ExposeToModel
SET Category = 'Skill & Action Mapping', DisplayName = 'Expose to Model', ...
```

…but the committed TypeScript says `* * Display Name: Expose To Model` (capital "To"), in both the zod `.describe` and the entity getter JSDoc. So the tail was taken after the `EntityField` INSERT (which writes the naive `'Expose To Model'`) but before the category block corrected it.

This isn't a camel-split artifact — the generated file carries DisplayNames a splitter could never produce (`"Additional Fields to Include"`, `"Display in Form"`, `"Bundle in API"`), so CodeGen reads `EntityField.DisplayName` from the database. It also explains why exactly one file drifted: GraphQL and the form template don't embed DisplayName.

**Fix:** re-run CodeGen against a database with the *complete* migration applied and replace the tail — or minimally change the two `Expose To Model` occurrences in `__mj.ts` to `Expose to Model`. The migration SQL is fine; the two-phase DisplayName is normal (`SkillID` → `'Skill ID'` goes the same route).

---

### The description and the guide claim behaviour that isn't there

Three related statements, all shipped into `AGENT_SKILLS_AND_PLAN_MODE_GUIDE.md` where they'll be read as authoritative:

**1. "attribution still uses the full bundle, so an application-invoked action under an active skill is still recorded against that skill on the Actions step."**

`getSkillAttributionForAction` has exactly one call site — inside `executeActionsStep` (`base-agent.ts:11234`) — reached only after resolving the action against `_effectiveActions`, which throws `Action "X" Not Found for Agent` otherwise (`:11223`). `_effectiveActions` is precisely the set hidden actions are now excluded from (`:6634`). So for an action hidden by every skill that bundles it — the feature's own scenario — attribution **can never fire**. (It can still fire if a *different* active skill exposes the same action, or if the agent has it natively, in which case native precedence returns `undefined` anyway.) The app-invoked path (`RunAction` → `ActionEngineServer.RunAction`) writes no `AIAgentRunStep` at all.

**2. The flag revokes execution, not just visibility.**

`_effectiveActions` is simultaneously the prompt tool surface *and* the run's execution allow-list. So `ExposeToModel = 0` also means the agent cannot run the action if driven to it by an agent-type `PreProcessActionStep` (which "may modify the actions array in place"), or by a loop via `findAgentActionForLoop` / `getEffectiveActionsForValidation` (`:4203`) — it hard-errors. The name and every doc sentence say "not offered to the model"; the behaviour is "removed from the run."

**3. "needs the grant so the action runs under the skill's authority."**

`ActionResolver.RunAction` (`MJServer/src/resolvers/ActionResolver.ts:182`) authorizes on API-key scope only, and `ActionEngine.RunAction` (`packages/Actions/Engine/src/generic/ActionEngine.ts:173`) does input validation, filters and a timeout — no user or skill authorization anywhere. Skill bundling grants nothing on the path this feature is for.

**None of this makes the feature wrong** — the real value (keep the association for SKILL.md export and UI wiring while keeping the action out of the prompt) stands on its own. But the docs should say that instead. If you *want* the "grant" semantics to be real, the alternative is to keep hidden actions in `_effectiveActions` and exclude them only from `formatActionDetails` — that's a genuine design choice, not a bug fix.

---

### Also required

- **`@memberjunction/server` is missing from the changeset** even though `MJServer/src/generated/graphql-schemas/__mj.ts` changed. The `fixed: [["@memberjunction/*"]]` group makes the version bump identical either way, so this is CHANGELOG completeness — but the directly comparable prior PR (`0d3094c543`, `AISkill.ActivationScope`) did list it for the same kind of change, and release notes are built from changesets.
- Branch is `BEHIND next`.

### Worth reconsidering

- **Portability gap.** The resync carry only protects re-import into the *same* database. SKILL.md is the portability format (§1.7) — export a skill and import it elsewhere and the hidden action becomes model-callable, i.e. a capability increase on a cross-environment move. The frontmatter already carries names; an optional per-action `expose: false` is small. At minimum the guide should name the gap rather than leaving the flag looking durable.
- **No integration-tier check.** `ai-skills.checks.ts` AS4 already asserts `GetSkillActionIDs`; a sibling assertion for `GetSkillExposedActionIDs` is two lines.
- **No agent-level test.** The tests cover the engine method, not `enableSkillCapabilities` wiring — a test at that level would have surfaced point 2 immediately.
- Dead clause: `if (entityName === 'MJ: AI Skill Actions' && kept !== undefined)` — `carried` is only ever populated for that entity.
- Generated tail says `/* SQL text to insert 2 new entity field(s) */` but contains one INSERT.

---

### What's good

- **Migration is house-rule clean.** DDL first → 50 blank lines → CodeGen tail; additive `BIT NOT NULL DEFAULT 1` with a named DF constraint and an extended property; `IF NOT EXISTS` guards; and the `EntityField` insert uses the apply-time `(SELECT COALESCE(MAX([Sequence]),0)+1 …)` expression rather than a literal. Timestamp `202609102050` is still ahead of the newest on `next` (`V202609101740`). PG counterpart correctly deferred.
- **Generated tail is internally coherent.** `vwAISkillActions` is `SELECT a.*` so it picks the column up; `spCreate` uses `ISNULL(@ExposeToModel, 1)`, `spUpdate` uses `ISNULL(@ExposeToModel,[ExposeToModel])`, so `0` is settable and omission is a no-op. `DefaultInView` is correctly flipped to `1` by the field-properties block.
- **Html-only form change is right.** The field lands in the existing `skillActionMapping` section; the `.ts` only changes when a *section* is added (confirmed against `0d3094c543`).
- **The `resyncJunction` carry is the sharp catch in this PR** and is correctly implemented — case-insensitive keying, `kept !== undefined` so a genuinely new row takes the default. It's on the real path: Skill Studio's `ai-skill-sharing-panel.component.ts` calls `AISkillImportMarkdownOperation` → this service.
- `!== false` over truthiness is the right call and is documented.
- **Nothing else needed updating**, verified rather than assumed: `metadata/ai-skills/.mj-sync.json` pulls `MJ: AI Skill Actions` generically with no field list, and `SQL Scripts/generated/**` is gitignored.

cmbrock-BC:
Thanks — every point landed, and two of them corrected me outright. All addressed in `3407ba03da` (merge of `next` + one follow-up commit; no history rewritten).

**Drift.** You had the direction right and I had it backwards: the category block sets `Expose to Model` after the INSERT's naive `Expose To Model`, and CodeGen reads the final row. Generated file now says `Expose to Model` in both places. (My "local repro shows no drift" was worthless — CodeGen's TypeScript step never rewrote `__mj.ts` in that run. The gate is also red on `next` since #4362, but that's separate.)

**Docs claiming behaviour the code doesn't have.** Agreed on all three, and I've taken the design choice you named explicitly: `ExposeToModel = 0` means **out of the run** — the effective action set is both the prompt surface and the execution allow-list, so the action is neither described nor executable by the agent. I kept that rather than "hide from the prompt only" because a prompt-only hide still lets the model call the action by name once a skill's Instructions mention it — the exact hole in #4226. The column description, migration header, code comments, changeset and guide (§1.1/§1.3/§1.7/§3) now say that, and say plainly that bundling grants nothing on the Actions API path and that attribution cannot apply to an action the agent never runs. The former "grant"/"attribution" sentences are gone everywhere.

**Portability.** Took your suggestion: SKILL.md gains an optional `codeOnlyActions` list (names, subset of `actions`) — written on export only when the skill has such rows (older files byte-identical), applied on import; a file without the key expresses no opinion and the resync carry (previous commit) becomes the fallback. The parser now also skips unknown *list* keys instead of throwing, so a newer file parses on an older MJ.

**Also required / reconsidering:** `@memberjunction/server` in the changeset; branch merged with `next`; AS4 asserts `GetSkillExposedActionIDs` ⊆ bundled; a real-`BaseAgent` test of `enableSkillCapabilities` (exposed subset only; all-code-only pushes no ActionChange; sub-agents unaffected — fails 3/3 when reverted to `GetSkillActionIDs`, which is the test that would have caught point 2); dead clause removed; tail comment says one INSERT.

`ai-agents` 2007/2007, `ai-engine-base` 206/206, 173 turbo tasks, all four migration gates.

cmbrock-BC:
Brought up to date with `next` after the edge.6 cut — two commits, nothing rewritten:

- `0d199fcd14` merge of `next`. One conflict, in the generated GraphQL schema: #3367 (field-level security) now emits every non-key output field as nullable, so `ExposeToModel` follows the same shape as its siblings — `@Field(() => Boolean, {nullable: true, description: …}) ExposeToModel?: boolean;` — and picks up `ReadableFields___`. Input types were already in that form.
- `ca1e778e43` re-dates the migration to `V202609111449` so it sorts after the three 2026-09-11 files `next` now carries (`outOfOrder` is off). Content unchanged.

Rebuilt server / ai-agents / core-actions on the merged tree (163 tasks) and re-ran the suites: ai-agents 2007/2007, ai-engine-base 206/206, core-actions 505/505.

rkihm-BC (dismissed):
Approving. Every item from my earlier review is addressed, and I re-verified each one against `ca1e778e43` rather than taking the reply at face value. One narrow gap in the new `codeOnlyActions` feature is noted at the end — not a blocker, your call whether it lands here or as a follow-up.

## Verified fixed

**The drift gate is green**, and the cause was the DisplayName mismatch. Thanks for confirming the direction — the category block writes `Expose to Model` after the INSERT's naive `Expose To Model`, and CodeGen reads the final row. All 22 checks pass now, including Build, Integration and the gate that was red.

**The three doc-accuracy problems are gone**, and the guide now states plainly what I previously had to reverse-engineer from `base-agent.ts`:

> "The run's effective action set is one list that is both the prompt's tool surface and the execution allow-list… bundling grants nothing there… Because the agent never runs such an action, skill attribution on the Actions step does not apply to it."

**And you were right to push back on my alternative.** I offered "hide from the prompt only" as an option; keeping "out of the run" because a prompt-only hide still lets the model call the action by name once a skill's Instructions mention it is the better call — that is the actual hole #4226 is about. Good correction.

Also confirmed independently:

- `@memberjunction/server` is in the changeset.
- `V202609111449` genuinely sorts after `next`'s three 2026-09-11 files (`…1000`, `…1001`, `…1002`).
- The GraphQL output field picked up #3367's `nullable: true` shape, matching its siblings; inputs were already in that form.
- AS4 asserts both the default-exposed case and the subset property.
- `skill-expose-to-model-agent.test.ts` is exactly the missing agent-level test — real `BaseAgent`, engine module-mocked so the two lookups can disagree on purpose, and it covers the case I cared about most (no `ActionChange` pushed at all when every bundled action is code-only, rather than an empty one).
- Dead `entityName ===` clause removed; tail comment says one INSERT.

## The new surface reviewed

`codeOnlyActions` is the right shape, and the precedence in `resyncJunction` is the subtle part done correctly:

```ts
const flag = exposeToModel?.get(id.toUpperCase()) ?? carried.get(id.toUpperCase());
```

`??` rather than `||`, so an explicit `false` from the file beats the carry — and because the map covers *every* bundled action, removing a name from `codeOnlyActions` genuinely re-exposes it. File authoritative when present, carry as fallback, column default for new rows. The forward-compatible `'ignore'` handling for unknown list keys is a good addition.

### One gap: the last code-only action cannot be re-exposed through SKILL.md

The parser only creates the array when an item is pushed (`(result[currentListKey] ??= []).push(value)`), so the block form with no items:

```yaml
codeOnlyActions:
```

leaves `frontmatter.codeOnlyActions === undefined` → `resolveExposeToModel` returns `undefined` → "no opinion" → the carry preserves the old `false`. Deleting the key entirely does the same thing.

So for a skill with exactly one code-only action, the natural edit — open SKILL.md, remove that name from the list, re-import — **silently keeps the action code-only**, and nothing warns.

The inline form happens to work, via a different code path: `codeOnlyActions: []` takes the `value.length > 0` branch, which assigns `result[key]` unconditionally and yields a real empty array. But the exporter never writes that form, so there is no way for a user to discover it.

One-line fix if you want it here: initialize the array when a known list key opens in block form, so "present but empty" is an explicit empty opinion instead of silence. It is a brand-new feature with no existing files in the wild, so a follow-up is also perfectly reasonable.

### Minor

`resolveExposeToModel`'s warning prints the resolved GUID rather than the name the user typed:

```ts
warnings.push(`codeOnlyActions names an action that is not listed under actions (${id}); ignored`);
```

It fires on a hand-written name, so echoing that name would make the warning actionable.

### Worth knowing, not a defect

`GetSkillExposedActionIDs` uses `ExposeToModel !== false`, and #3367 made the output field nullable. If field-level security were ever enabled on `MJ: AI Skill Actions` and withheld that column, a withheld value would read as *exposed* — mildly fail-open. Hypothetical today (FLS is opt-in per entity and this one does not enable it), and the `!== false` choice is deliberate and documented; just flagging it as a new interaction between the two features that neither side had reason to consider.

cmbrock-BC:
Thanks for the re-verification. Took both notes here rather than as a follow-up — `35ace7cc4c`, one commit on top:

- **Empty `codeOnlyActions:` is now an opinion.** The parser creates the list when a known key opens in block form, so present-but-empty parses as `[]` and `resolveExposeToModel` returns an all-exposed map instead of `undefined`. Deleting the last name and keeping the key puts the action back into the run; deleting the key still means no opinion. `actions`/`subAgents` already treated `undefined` and `[]` alike (`?? []`), so nothing changes for them.
- **Warnings quote the typed name.** `resolveExposeToModel` resolves one name at a time, so the not-bundled warning says `codeOnlyActions names 'Send Email', which is not listed under actions; ignored` rather than echoing the GUID.
- Tests: the parser cases (empty block, hand-edited export round trip), resolution (absent vs empty, both warnings by name), and the scenario you described end to end — export with one code-only action, delete its line, re-import through `resyncJunction`, assert the recreated row has `ExposeToModel = true`. Each fix reverted fails its tests. Guide §1.7 and the converter header say the same.

On the FLS note: agreed it's hypothetical today; leaving `!== false` as documented. ai-agents 2014/2014.

rkihm-BC (approved):
Re-approving on `35ace7cc4c`. Both notes are closed, and I verified the fixes against the branch rather than the reply.

**Empty `codeOnlyActions:` is now an opinion.** The block-form branch does `result[key] ??= []`, so present-but-empty parses as an explicit `[]` and `resolveExposeToModel` returns an all-exposed map instead of `undefined`. Deleting the last name and keeping the key puts the action back into the run; deleting the key still means no opinion. That's exactly the distinction that was missing, and the parser comment now records why it matters. `actions` / `subAgents` are unaffected since they already treat `undefined` and `[]` alike.

**Warnings quote the typed name.** Resolving one name at a time so the not-bundled warning can say `codeOnlyActions names 'Send Email', which is not listed under actions; ignored` is the right trade — the GUID was never something an author could find in their file. An unknown name still falls through to `resolveNames`' own warning and never marks anything code-only.

**The tests cover the scenario I described end to end** — export with one code-only action, delete its line by hand, re-import through `resyncJunction`, assert the recreated row has `ExposeToModel = true` — plus the complement (key removed altogether → old flag carried). Together with the parser round-trip and the resolution cases, that pins both the fix and the boundary it sits on.

Agreed on leaving `!== false` as documented for the FLS note. 22/22 green.

PR #4366, opened 2026-09-10
fix(conversations): the input dialog's body gets the same inset as the rating dialog

cmbrock-BC:
## Summary

`mj-input-dialog` — the Rename-conversation prompt and every `dialogService.input()` — rendered its message, labels and inputs flush against the dialog edges while the header and footer were padded.

`MJDialogContainerComponent` sets `.mj-dialog-body { padding: 0 }` and pads only string content (`> p { padding: 20px }`), so a component body arrives unpadded. The sibling rating dialog already compensates in its own styles (`.rating-dialog { padding: 4px 20px 8px }`); the input dialog did not (`.input-dialog-content { padding: 8px 0 }`). This applies the same inset.

## Changes

- `input-dialog.component.ts`: `.input-dialog-content` padding `8px 0` → `4px 20px 8px`, with a comment saying why the component pads itself.
- Changeset: `@memberjunction/ng-conversations` patch.

## Testing

- `@memberjunction/ng-conversations` builds (68 turbo tasks) on `next` @6ac6bcc5ca.
- Seen live in Betty (which carries this as a labelled interim override until it lands): the Rename dialog's field sits inset from the edges like the rating dialog's.

Found while reviewing Betty's chat UI against `next` (2026-09-09).


rkihm-BC (commented):
## Review: input dialog body inset

**Verdict: looks good to merge as-is.**

### Verified

- **The diagnosis is exactly right.** `MJDialogContainerComponent` (`ui-components/src/lib/dialog/dialog.service.ts:178-190`) sets `.mj-dialog-body { padding: 0 }` and pads only `.mj-dialog-body > p`, so a component body arrives with no inset. The sibling rating dialog already compensates with the identical `4px 20px 8px` (`rating-dialog.component.ts:112`).
- **Right layer.** Pushing the padding into the container would double-pad the rating dialog and break deliberately full-bleed component bodies — About, Profile, new-agent, action-gallery and the sharing centre all pass components to the same service. Per-component is the correct non-regressive choice.
- **Complete within the package** — those are the only two component dialogs under `components/dialogs/`.
- Comment explains the *why*; changeset package and level correct; CI fully green including the design-token gate.

### Two optional notes

1. **`box-sizing` parity.** The rating dialog defensively sets `box-sizing: border-box` on itself and its descendants; `.input-dialog-content` doesn't. This is safe in practice — `explorer-app/src/lib/styles/_common.scss:3` applies a universal `* { box-sizing: border-box }` that reaches this component through the Explorer shell — so it only matters for a host that doesn't load that reset. Worth one line for parity, not a correctness issue. (Noting for the record that `.mj-input` / `.mj-textarea` combine `width: 100%` with padding and a border and set no local `box-sizing`, which is what makes the global reset load-bearing here.)

2. **The container's contract is undocumented.** "A component body arrives unpadded — pad yourself" has now been independently rediscovered by two components. A comment beside `padding: 0` in `dialog.service.ts`, or a `padded` setting on `MJDialogSettings`, would stop the third.

rkihm-BC (approved):
Approving — see my detailed review above. Diagnosis verified against `dialog.service.ts`: the container pads only string content, so a component body arrives unpadded, and this matches the inset the sibling rating dialog already applies. Right layer for the fix, and complete within the package.

Both notes (`box-sizing` parity, and documenting the container's "component bodies pad themselves" contract) are optional.

PR #4368, opened 2026-09-10
fix(isa): promotion over GraphQL inserted a duplicate parent; related panel stayed blank after a create-with-subtype save

MarceloT-BC:
Three findings from the MJ Academy shelter module's IS-A handoff, all on the create path and independent of each other: **promotion over GraphQL inserts a duplicate parent** (defects 1+2, one fix — neither half works alone), **the IS-A side panel stays blank after a create-with-subtype save** (defect 4), and **every IS-A child create with two or more sibling subtypes waited 30 seconds** (defect 3). The first two are fixed here, one commit each. The third turned out to be already fixed on `next` by `4456222c8f`; it is documented below with the evidence so nobody re-diagnoses it, and carries no code in this PR.

---

# Fix 1 — IS-A promotion over GraphQL (`@memberjunction/graphql-dataprovider`, `@memberjunction/server`)

## The gap

IS-A **promotion** — an existing parent record gaining a subtype ("this Animal is now also a Dog") — could not work over GraphQL. It worked against a direct database provider, which is what made it expensive to find: every server-side reproduction passed while the browser **inserted a second copy of the parent row**, surfacing as a unique-constraint violation on an unrelated column with a different GUID on every retry. Found by the MJ Academy shelter module; both defects present on `next` at `6ac6bcc5ca`. Builds on #3871's `AttachToParent`, which had no callers until now.

## Two defects, both required

- **Client** (`GraphQLDataProvider.Save`): the create-side field filter admitted a primary key only when the entity was already saved, and an IS-A child's key is ReadOnly (the shared key *is* the relationship), so on a promotion the key never left the browser. The parent's own save is short-circuited (`IsParentEntitySave`) on the premise that the leaf mutation carries the whole chain — so nothing told the server which parent row this was about. The create input now carries the shared key **only for a promotion** — an unsaved IS-A child whose parent is already saved (`IsChildType && ISAParent.IsSaved`). A whole-chain create (new parent + new child) keeps sending no key, so the server still mints the root identity exactly as before and pays no parent lookup on that path. **Who controls identity does not change.**
- **Server** (`ResolverBase.CreateRecord`): `NewRecord()` reset the whole chain to "new", so even with the key present the parent saved as a CREATE. When a child create carries a complete key, the resolver now binds the new child to the existing parent row with `AttachToParent`: the parent saves as an UPDATE and only the child is INSERTed. Field assignment runs after the attach, so parent fields in the input land on the loaded parent as an update. A key matching no row (a raw API caller supplying one) is the ordinary whole-chain create and proceeds on the caller's key (`AttachToParent` leaves the root's ReadOnly key slot consumed after a miss, so the chain is re-seeded with `NewRecord(key)`); an empty-string key counts as absent, a malformed one fails the lookup loudly; non-IS-A entities never enter the path.

**Cost.** Only a promotion pays the parent lookup, and it is one single-row read by primary key: on the live rig `SELECT * FROM vwAnimals WHERE ID=@p0` registered 138 executions at 0.32 ms average, 3.6 ms maximum, server-side (plan cache). Whole-chain creates send no key and pay nothing.

**Field-level security on the promotion path.** Parent fields sent on a child create meet whichever verb the parent's save runs, and that is decided by whether the parent row exists: on a promotion the parent is loaded and saves as an UPDATE, so a field the user cannot update **rejects** the whole save before any SQL (`CheckFieldLevelUpdatePermissions`); on a whole-chain create the parent is new, so the same field the user cannot create is **silently omitted** (`ApplyFieldLevelCreateSuppression`). That fork is deliberate — a promotion really is an update of the parent — and it is pinned in both directions by tests.

## How to reproduce the original failure

1. An IS-A pair — parent `Animal` (ideally with a unique index on a non-key column such as `MicrochipNumber`), child `Dog` sharing `ID` — and an Animal row with no Dog row.
2. In any GraphQL client:
   ```ts
   const animal = await md.GetEntityObject('MJ: Animals', user);
   await animal.InnerLoad(existingAnimalKey);
   const dog = await animal.EnsureISAChild('MJ: Dogs');
   dog.Set('IsHouseTrained', true);
   await animal.Save();   // delegates to the Dog leaf → CreateMJDog over the wire
   ```
3. **Before:** `CreateMJDog` is sent without `ID`; the server mints a fresh GUID and runs `spCreateAnimal`, which fails on the unique index (or silently duplicates the parent without one). With the key present but only the server unfixed, it fails on `PK_Animal` instead. **After:** one Dog row on the shared key, the Animal untouched, `IsHouseTrained` persisted.

## Tests

Written red-first against `origin/next`:

- `GraphQLDataProvider` wire test (real provider, fake transport, real `EntityInfo`/`BaseEntity` fixtures in a two-level IS-A chain): the promotion create carries the parent key and the chain fields; a whole-chain create carries the key its chain minted; a non-IS-A create still omits the key; an IS-A update is unchanged; the parent save still makes no wire call.
- `ResolverBase` test (real `BaseEntity`/`EntityInfo` over a fake provider that constructs entities the way `ProviderBase.GetEntityObject` does and records every Save): promotion creates only the child on the shared key; parent fields in the input become an UPDATE of the loaded parent; whole-chain create with and without a supplied key; non-IS-A create untouched; promotion under `AllowMultipleSubtypes`; and the field-level-security fork — promotion rejects a field the user cannot update (before any SQL), a field the user can update lands as an UPDATE, and a whole-chain create omits a field the user cannot create and proceeds.

Full suites on the merged tree (with #3367's field-level security in place): `@memberjunction/graphql-dataprovider` 447/447, `@memberjunction/server` 1323 passed / 56 pre-existing skips. Typecheck and build clean on both.

Verified live against a real MJAPI + SQL Server (a copy of the academy shelter database at `next` head), both tiers, before and after: before, `CreateMJDog` with the existing key failed on `PK_Animal` and the unfixed client's promotion failed on `UQ_Animal_MicrochipNumber` with a fresh GUID; after, one Dog row on the shared key, no duplicate Animal, field persisted, audit trail shows only the Dog create.

## Not covered

Composite primary keys (handled all-or-nothing by construction, no fixture), overlapping subtypes (`AllowMultipleSubtypes = true`) on the wire and in the live run — the resolver-level promotion is tested (it attaches to the existing parent and creates only the child), but the wire suite has no overlapping case and the shelter database is disjoint, chains deeper than two levels (a promotion where the intermediate row is also missing falls through to a whole-chain create and fails loudly on the root PK rather than silently). No integration-tier bundle: MJ core models zero IS-A relationships, so the deterministic tier has nothing to point a promotion test at (the same conclusion #3871 reached); the end-to-end proof is the live run above.

---

# Fix 2 — IS-A side panel blank after a create-with-subtype save (`@memberjunction/ng-base-forms`)

## The gap

Create a record together with its subtype in one save — the ordinary, shipping IS-A flow — and `<mj-isa-related-panel>` (the right-hand "IS-A Related Records" panel) stays empty: the record plainly has a Dog, the panel says it has nothing. Pressing the in-app Refresh makes it appear.

## Root cause

The panel re-ran discovery on exactly two triggers, and neither can fire after a save:

- `changes['Record']` cannot: `EntityFormHostComponent`'s PrimaryKey setter deliberately keeps the same `BaseEntity` bound across a new-record save (it skips the reload for the now-saved PK, `entity-form-host.component.ts:83`).
- `Refreshed$` cannot: `FormRecordRefreshCoordinator.Notify()` has one production call site, `RecordFormContainerComponent.OnFormRecordRefreshed` (`record-form-container.component.ts:1657`), reached only from the Refresh button.

So the answer computed at load — "no subtypes" on a new record — was never invalidated.

## The fix

Leaving edit mode is now the third trigger (`ngOnChanges`, `EditMode` true → false): that is the moment a save has landed. On a disjoint hierarchy `DiscoverISADescendants` walks the in-memory `ISAChild` chain and issues no query; an overlapping parent (`AllowMultipleSubtypes`) pays one child-entity load per edit session — named in the code comment, the only cost. Entering edit mode does not re-discover; a new `Record` still takes the existing branch.

## How to reproduce

Open a new record form for an IS-A parent with at least one subtype, fill it in, pick a subtype, save. Before: the panel is absent until Refresh. After: it appears on save.

## Tests

Four DOM specs appended to `isa-related-panel.component.dom.test.ts` (which previously bypassed `ngOnChanges` entirely), written red-first: leaving edit mode re-discovers; entering does not; a new Record re-discovers; a record with no subtype yields no phantom panel. `@memberjunction/ng-base-forms` 459/459 (30 files), `ngc` clean. Verified live in Chrome by the academy team on the real form (panel appears on save for both `MJ: Dogs` and `MJ: Cats`); I did not repeat the browser run.

---

# Finding 3 — every IS-A child create waited 30 seconds (already fixed on `next`, no code here)

## The gap, as found on the academy's pin (base 2026-08-26)

`BaseEntity._InnerSave` inserts the parent row inside the save's transaction and holds the lock, then runs `EnforceDisjointSubtype`, which asks a **sibling subtype's view** whether it already claims the key. A subtype view joins its own table to the parent table, so it must read the row this very transaction just wrote. On the pin that query ran on a fresh host `new RunView()` — a different connection, outside the transaction — so it blocked on the save's own uncommitted write until SQL Server's 30 000 ms request timeout. The timeout was swallowed, the record still saved, and nothing surfaced but the delay. It only bites when the parent has two or more subtypes (with one, `siblingChildEntities` is empty and the method returns early), which is presumably why a single-subtype fixture never showed it.

## Why it is not in this PR

`4456222c8f` (2026-09-07, "bind every graph DB op to one provider so nested Action Params do not deadlock") changed the guard to `new RunView(this.RunViewProviderToUse)` — with the in-code note that a host `RunView` would miss an uncommitted sibling on the same graph connection — and the 2026-09-04 nested-transaction work (`9f7352811e`, `22659004d8`) routes an instance query through the active transaction when one is open. Together they remove the self-block at the root: the sibling read now runs on the same connection as the write. The handoff's proposed reorder (run the check before the transaction opens) is therefore unnecessary, and would lose the ability to see an uncommitted sibling created earlier in the same graph save.

## Measured on `next`

Live against a real MJAPI + SQL Server (the same copied shelter database as fix 1, `Animal → Dog | Cat`, snapshot isolation **off** on the database so a cross-connection read would still block):

| Check | Result |
|---|---|
| Whole-chain `CreateMJDog` (new Animal + new Dog), three samples | 239 ms, 72 ms, 62 ms |
| Sibling probe against `vwCats` actually executed on each create | yes — three plan-cache entries at those times, max elapsed 0 ms |
| Disjoint refusal still enforced (`CreateMJCat` for an animal that is already a Dog) | refused in 35 ms with the expected message, no rows written |

The academy's pin predates both commits, which is why it measured 30 091 ms for the same create.

---

Changesets: `patch` × 2 (code only).


rkihm-BC (commented):
@MarceloT-BC — I resolved the merge conflict and pushed it to this branch as `804c0a1f7c`. **Please check that resolution before anything else**, since it lands in the middle of the very expression your client fix changes. Details first, then the review.

---

## The conflict I resolved (please verify)

One file: `packages/GraphQLDataProvider/src/graphQLDataProvider.ts`, in `Save()`'s create-side field filter. `ResolverBase.ts` auto-merged cleanly.

Two changes hit the same expression for unrelated reasons:

- **this branch** widened the ReadOnly/PK clause so an IS-A child CREATE carries the shared key — your promotion fix;
- **Field-Level Security (#3367)**, merged to `next` after you forked, added a second clause dropping `NotLoaded` and denied-READ fields from the input.

They compose, because the FLS clause already exempts `f.IsPrimaryKey`:

```ts
const isaChildCreate = !entity.IsSaved && entity.EntityInfo.IsChildType;
const deniedReadFields = this.GetDeniedReadFieldNamesForCurrentUser(entity.EntityInfo);
const isDeniedRead = (fieldName: string) => deniedReadFields.has(fieldName.trim().toLowerCase());
const filteredFields = entity.Fields.filter(f =>
    (!f.ReadOnly || (f.IsPrimaryKey && (entity.IsSaved || isaChildCreate))) &&
    (f.IsPrimaryKey || (!f.NotLoaded && !isDeniedRead(f.Name))));
```

The IS-A shared key survives the security clause via that PK exemption, so the promotion key still reaches the server; every other field keeps `next`'s filtering. Both comment blocks are retained.

I also checked your wire tests against the merged filter before pushing: their fixtures don't enable FLS, and `GetDeniedReadFieldNamesForCurrentUser` short-circuits to an empty set unless `EnableFieldLevelSecurity` is on, so the new clause is a no-op for them. That's reasoning, not a local run — I resolved this in a worktree without an install, so **CI on this push is the first real execution of the combined code.**

---

## What's good

**The diagnosis is the strongest part of this PR.** Two defects that each mask the other — the key never leaving the browser, and `NewRecord()` resetting the chain server-side — plus a clear account of why every server-side reproduction passed while the browser duplicated the parent. That is the expensive kind of bug, and the writeup means nobody re-does the work.

**The security model holds under scrutiny.** I traced it rather than assuming, because this path loads a client-supplied key and then updates an existing row:

- `AttachToParent` → `_parentEntity.InnerLoad(parentKey)` is Read-gated and RLS-filtered;
- the attached parent then saves as an UPDATE, and `BaseEntity.Save` runs its own `CheckPermissions(Update)` plus `CheckFieldLevelUpdatePermissions()` on it (`baseEntity.ts:4230-4234`).

So a promotion requires Read + Update on the parent and Create on the child. No escalation path.

**The subtle mechanics are right.** `AttachToParent` captures the parent PKs *before* `InnerLoad` wipes them, so a miss restores the fresh chain bit-for-bit. `NewRecord(key)` type-checks and behaves because `CompositeKey extends FieldValueCollection` and `NewRecord` applies `KeyValuePairs` through `Set`, which routes to the root — and the re-seed genuinely is required, since the root's ReadOnly key slot is consumed by the failed attach. `primaryKeyFromInput` is all-or-nothing on composites.

**The panel guard is correctly written**: `changes['EditMode'].previousValue === true && !this.EditMode` won't fire on first init (previousValue `undefined`) or on entering edit mode.

**Finding 3 is a model of what to do with a bug that fixed itself** — measured on `next` with snapshot isolation deliberately off, and it explains why the handoff's proposed reorder would have been *wrong* (it would lose the ability to see an uncommitted sibling created earlier in the same graph save).

Changesets are right: two, both `patch`, correct packages, no migration.

## What's concerning

1. **The merge creates a security-verb fork that nothing tests.** Parent fields sent on a child-create input now land on a *loaded* parent that saves as an UPDATE, so they run through `CheckFieldLevelUpdatePermissions`, which **rejects**. On the whole-chain create path the same fields go through `ApplyFieldLevelCreateSuppression`, which **silently omits**. Identical input, two different enforcement verbs, chosen by whether the parent row happens to exist. That's defensible — it really is an update — but your PR and #3367 were each written against the other's absence, so it's untested in both directions. This is the one I'd want a test for before merge.

2. **Every IS-A child create now pays an extra parent SELECT.** The client sends the shared key on *all* IS-A child creates (`isaChildCreate = !entity.IsSaved && IsChildType`), not only promotions, so the server can't tell them apart and must probe via `AttachToParent` → `InnerLoad`. On the common whole-chain create that probe is a guaranteed miss — one extra round trip per create. Given that finding 3 in this same PR is a 30-second IS-A create, this path's cost deserves a sentence and ideally a number; you already have the rig that produced the 239/72/62 ms samples.

3. **Client-minted primary keys are now honoured for IS-A creates.** Right now that's a half-sentence at the end of "Not covered". The server used to mint the root key and now accepts the client's. It is properly gated, and `spCreate` already accepts `@ID`, so it sits inside the existing design — but it changes who controls identity for a whole class of entities, and reviewers skim that section. Worth promoting into the summary.

4. **`primaryKeyFromInput` treats `''` as absent.** Harmless for GUIDs; for a string-keyed entity an empty key silently becomes a whole-chain create rather than an error. Minor, but the composite all-or-nothing rule is documented while this one isn't.

## What's missing

- **No integration-tier coverage.** This bug silently wrote duplicate parent rows, and the deterministic tier exists for exactly this client → wire → resolver → ORM seam. Both new suites are unit-level with fake transports and providers; the end-to-end check was manual. An IT bundle for promotion is what stops this regressing.
- **`AllowMultipleSubtypes` promotion is untested**, and it's the case where promotion is most natural — overlapping subtypes exist precisely to be added incrementally. At least a resolver-level test.
- **Fix 2's trigger→save linkage rests on manual verification.** The DOM specs drive `ngOnChanges` directly; nothing covers a real save producing the edit-mode exit. Honestly disclosed in the description — just flagging that it's the one link in fix 2 with no automated cover.

Overall this is strong work and the security analysis survives a close look. Findings 1 and 2 are what I'd want addressed before merge; 3 is a description edit.

MarceloT-BC:
@rkihm-BC thanks — thorough, and the two findings you gated on are both addressed in `bf93184cc7` (branch also re-merged with `next` at `8d069adc62`, 0 behind).

**Your conflict resolution — verified, and executed.** The composed filter is right: my clause and the FLS clause are conjoined, and the FLS clause's `f.IsPrimaryKey` exemption is what lets the shared key through. I ran the combined code locally, not just CI: the wire suite passes on the merged filter (447/447 for the package), and the fixtures' FLS-off path short-circuits `GetDeniedReadFieldNamesForCurrentUser` exactly as you reasoned.

**1. Security-verb fork — now pinned in both directions.** Three resolver-level cases under `EnableFieldLevelSecurity` with `Name` denied for update and create: a promotion carrying `Name` **rejects before any SQL** with `You do not have permission to update field 'Name'`; a promotion carrying a field the user *can* update lands as an UPDATE of the loaded parent; a whole-chain create carrying `Name` **omits it** (`CreateSuppressed`) and proceeds. Agreed it is defensible — a promotion is an update of the parent — and the tests say so in their comment so the next reader doesn't re-litigate it.

**2. Extra parent SELECT — removed from the common path.** The client now sends the shared key **only when the parent is already saved** (`IsChildType && ISAParent.IsSaved`), i.e. only on a promotion. A whole-chain create sends no key, exactly as before this PR, so the server mints the root identity and never probes. The wire test's whole-chain case flipped to pin "no key sent". Where the probe does run it is one single-row read by PK: on the live rig `SELECT * FROM vwAnimals WHERE ID=@p0` — 138 executions, 0.32 ms avg, 3.6 ms max, server-side (plan cache).

**3. Client-minted keys — no longer a behaviour change.** Falls out of #2: identity control is unchanged for every IS-A create. A raw API caller who *does* supply a key still gets the miss-path re-seed (`NewRecord(key)`), which stays tested, and the description now says so in the fix-1 summary rather than a trailing half-sentence.

**4. Empty string — documented** on `primaryKeyFromInput`: `""` counts as absent (whole-chain create), while a malformed non-empty key still fails the lookup loudly.

**Overlapping subtypes** — resolver-level promotion test under `AllowMultipleSubtypes` added; it attaches to the existing parent and creates only the child.

**Integration tier** — I agree it is the right place and I could not put one there: MJ core models zero IS-A relationships, so the deterministic tier has no entity pair to promote (the same wall #3871 hit). If core ever gains an IS-A fixture I'll write the bundle; until then the live before/after run on the shelter copy is the end-to-end evidence and is documented in the description.

**Fix 2 linkage** — stays manually verified; nothing automated covers a real save producing the edit-mode exit. Left disclosed as-is.

Full suites on the merged tree: graphql-dataprovider 447/447, server 1323 passed / 56 pre-existing skips, ng-base-forms 459/459; tsc clean on all three.

rkihm-BC (dismissed):
## Approving. Both gated findings are closed, and I measured the claims rather than re-reading them.

My earlier review was a comment, not a verdict, gated on two things: the untested security-verb fork, and an extra parent `SELECT` on every IS-A create. `bf93184cc7` closes both, and the branch has since taken two clean merges from `next` (0 behind, Integration green on `9c823e29ef` at 11m35s). This is the verdict.

### What I verified in the code on the head

**Finding 2 — the key now travels only on a promotion.** The filter reads exactly as described:

```ts
const isaPromotionCreate = !entity.IsSaved && entity.EntityInfo.IsChildType && entity.ISAParent?.IsSaved === true;
(!f.ReadOnly || (f.IsPrimaryKey && (entity.IsSaved || isaPromotionCreate))) &&
(f.IsPrimaryKey || (!f.NotLoaded && !isDeniedRead(f.Name)))
```

A whole-chain create sends no key, so the server mints the root identity and never probes — the extra round trip is gone from the common path, and the conflict resolution I pushed (the FLS clause with its `IsPrimaryKey` exemption) is intact underneath it. Your local run of the merged filter is the execution I said CI would be the first of; thank you for doing it.

**Finding 1 — the fork is pinned in both directions.** The three resolver cases are there by name: a promotion carrying a field the user cannot update rejects before any SQL; one the user can update lands as an UPDATE of the loaded parent; a whole-chain create carrying a field the user cannot create omits it and proceeds. The test comment states why the verbs differ, so the next reader does not re-litigate it.

**Findings 3 and 4** — identity control is unchanged for every IS-A create (it falls out of finding 2), and `''`-as-absent is now documented on `primaryKeyFromInput`, with the malformed-key case explicitly left loud. The `AllowMultipleSubtypes` promotion test is present too.

### What I measured

I ran the wire suite in a fresh worktree against the PR head and then against `origin/next`'s `graphQLDataProvider.ts`:

| | result |
|---|---|
| PR head | 6 / 6 pass |
| `origin/next` client | **1 / 6 fail** — `promotion: the CREATE mutation input carries the existing parent primary key` → `expected undefined to be 'C1C2…AAAA'` |

That is the right shape. The one failure is the defect itself — the key never leaves the browser. The five that still pass on old code are regression guards for behavior that must *not* change, and their passing on both sides is what proves they guard the right thing. "Written red-first" is measured for this suite.

**Not measured by me:** the resolver suite's red-first claim. Building `@memberjunction/server`'s dependency chain locally is out of proportion for a re-review, so for those ten cases I am relying on their presence by name, on CI green, and on your statement. Saying so rather than implying otherwise.

### The merges

`de158ba0b6` lists `ResolverBase.ts` in its stat, which made me look for a hand-resolved conflict on the file carrying your server hunk. There is none — `git show` on the merge produces no hunks for it, so it auto-merged cleanly, and both hunks are present on the head. The changeset that merge removed (`workflows-run-on-the-dispatcher.md`) was released on `next` in `ff4b4c82cc`; deleting it is correct convergence, not a lost release note.

### Two things worth a follow-up commit, neither blocking

1. **One sentence in the description is now stale.** "Not covered" still lists *overlapping subtypes (`AllowMultipleSubtypes = true`)*, but the Tests section — and the resolver suite — now cover a promotion under exactly that. Drop it from the list, or say what about overlapping subtypes remains uncovered if something does.

2. **The FLS rationale comment was removed, not relocated.** When the promotion comment replaced the block above the filter, the paragraph explaining *why only the READ verb is filtered here* went with it: that a readable-but-update-denied field must still be **sent**, because the server's check is dirty-only and can only refuse a change it can see; that create-denied values are dropped server-side; and that filtering either verb on the client would turn a visible refusal into a silent success. The clause still works — the FLS suite and CI say so — but that paragraph is the thing that stops someone "simplifying" the filter in a year. I would put it back beneath the promotion comment.

### Accepted as-is

No integration-tier bundle, for the reason given and the precedent #3871 set: MJ core models zero IS-A relationships, so the deterministic tier has no pair to promote. The live before/after on the shelter copy is the end-to-end evidence and is documented. Fix 2's trigger-to-save linkage stays manually verified and is disclosed as such. Finding 3 remains a model of what to do with a bug that fixed itself.

Nice work closing this out — the diagnosis was the strongest part the first time, and the follow-through matched it.

rkihm-BC:
Follow-up #1 from my review is done — I edited the description myself rather than leave it as a to-do. The "Not covered" sentence on overlapping subtypes now says what actually remains uncovered (the wire suite has no `AllowMultipleSubtypes` case and the shelter database is disjoint) instead of listing it outright, since the resolver-level promotion *is* tested. One sentence; nothing else touched — reword freely.

Follow-up #2 — restoring the FLS rationale paragraph beneath the promotion comment in `graphQLDataProvider.ts` — is a code change on your branch, so I've left it to you. Happy to push it if you'd rather.

rkihm-BC:
Follow-up #2 is done in `9f8f634ab0` — the field-level-security rationale is back beneath the promotion comment above the create filter, restored verbatim from `next` minus its superseded first line. Comment-only, +18/−0. Before restoring it I re-verified each claim against this branch's head rather than trusting the old text: the update check is dirty-only (`CheckFieldLevelUpdatePermissions`, `baseEntity.ts:4656–4669`), `Get()` asserts readability before any read path, `ApplyFieldLevelCreateSuppression` runs inside `Save()` and never rejects, `MarkNotLoaded` fires on hydration when a key is absent, and the NOT-NULL fallback and denied-read response filter are still directly below. Package suite 450/450, `tsc` clean. Reword freely — the point is that the reasoning survives, not the wording.

rkihm-BC (approved):
## Re-approving — my own push dismissed the earlier approval.

The only change since the head I approved (`9c823e29ef`) is `9f8f634ab0`, which I pushed: the field-level-security rationale restored above the create filter, comment-only, +18/−0. Nothing in the substance of the earlier review has changed, so the verdict carries over — see that review for the full assessment, including the measured red-then-green of the wire suite.

Before restoring the comment I re-verified each of its claims against this branch's head rather than trusting the old text (details in the commit message and the comment above). `@memberjunction/graphql-dataprovider` builds clean and passes 450/450 with it in place. CI is re-running on a commit that cannot change behavior.

Both follow-ups from the original review are now closed — #1 in the description, #2 in `9f8f634ab0` — and the author is free to reword either.

PR #4372, opened 2026-09-10
fix(codegen): reading persisted validators is a DB read, not an AI call

SDesai-BC:
Closes the `codegen-drift` failure on `next` @ `57975fdaee`.

## What was still broken

`#4370` fixed one gate in front of the validator load. There were two.

| | |
|---|---|
| Gate 1 — fixed in #4370 | `loadGeneratedCode` was reachable only on the `--skipdb` branch |
| **Gate 2 — this PR** | both call sites that queue a validator sit behind `ag.featureEnabled('ParseCheckConstraints')`, and `--no-ai` sets `enableAdvancedGeneration = false` |

So after #4370 the load *ran*, returned `true`, printed `✔ AI Generated Code loaded from Metadata` — and loaded **zero**. The drift gate runs `--no-ai`, so CI kept emitting `__mj.ts` with none of the 56 `Validate()` overrides, and the gate required that lossy output: restoring the validators failed CI while deleting them passed.

Independent confirmation of the direction: the gate ran genuinely green on `next` up to `6ac6bcc5ca`, when the file had **0** validators, and went red the moment the 56 were committed.

## Why the read never needed AI

The load calls `runValidationGeneration(..., generateNewCode = !skipDBUpdate = false, ...)`, and `generateValidatorFunctionFromCheckConstraint` only reaches an LLM when that flag is true. An AI feature flag was gating a plain database read of Approved `GeneratedCode` records.

```ts
const emitValidators = skipDBUpdate || ag.featureEnabled('ParseCheckConstraints');
```

The load-only pass is unconditional; **generation stays gated exactly as before**. This removes the special case rather than adding one.

## Why the previous proof was wrong

#4370's local verification was vacuous, and it is worth recording so it doesn't recur. Per-schema emit (#3804) rebuilds only **dirty** schemas, and on an already-migrated database nothing is dirty — so `__mj.ts` was never written. The empty `git diff` meant nothing was generated, not that the output matched. The file's mtime stayed at `12:44` through a `16:43` codegen run.

**An empty `git diff` after CodeGen is not evidence unless the artifact's mtime moved.**

## Verification

Throwaway database built by the drift gate's own steps — `migrate` → `sync push --dir=metadata --ci` → `codegen --no-ai --skip-commands`. 78 migrations, 383 entities, `unchanged: 14089` — identical to the CI job.

| scenario | validators loaded | `Validate()` overrides | size | drift |
|---|---|---|---|---|
| forced emit, before fix | 0 | 0 | 4,522,259 | **`M __mj.ts`** |
| forced emit, after fix | 116 | 56 | 4,661,490 | none |
| **artifact deleted + full run** | 116 | 56 | 4,661,490 | none |

The last row is the strict case: deleting the artifact makes the emit unskippable, and it regenerated **byte-identical** to what is committed.

- `codegen-idempotency-check --stage warm-twice` → PASSED (0 diffs, 0 surviving SQL logs, 0 changed fields)
- `codegen-idempotency-check --stage single-column` → PASSED
- `@memberjunction/codegen-lib` unit tests → **1409 passed**, 0 failed (66 skipped)

## Second commit — CI diagnostics only

Both drift gates reported a filename and nothing else, so acting on the failure meant reproducing the whole job locally to see a diff CI already had. They now print `git diff --stat` plus a bounded 100 KB excerpt.

**This does not touch pass/fail.** Every added line sits inside the block that has already decided to fail, after the error and before the unchanged `exit 1`; the comparison and its condition are untouched, and on a green run none of it executes. Reverting that commit would still leave this branch green.

## Changeset

`patch` — this branch touches no migration and no metadata.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -716,6 +716,16 @@ jobs:
           if [ -s /tmp/drift.txt ]; then
             echo "::error::CodeGen output DIFFERS from committed artifacts (drift). Files:"
             cat /tmp/drift.txt
+            # A filename alone does not say WHAT moved, and these artifacts are megabytes — so the
+            # only way to act on this failure used to be reproducing the whole job locally. Emit the
+            # shape of the change (per-file +/-) and a bounded excerpt instead. Bounded because a
+            # full regeneration of __mj.ts is ~100k lines and would bury the log.
+            git diff --stat -- packages/ metadata/ migrations/ > /tmp/drift-stat.txt || true
+            git diff -- packages/ metadata/ migrations/ | head -c 100000 > /tmp/drift-diff.txt || true
+            echo "--- change shape ---"
+            cat /tmp/drift-stat.txt
+            echo "--- first 100KB of diff ---"
+            cat /tmp/drift-diff.txt
             exit 1
           fi
           echo "No drift — committed generated artifacts match a clean-DB CodeGen run."
@@ -735,6 +745,16 @@ jobs:
               echo '```'
               cat /tmp/drift.txt
               echo '```'
+            fi
+            if [ -s /tmp/drift-stat.txt ]; then
+              echo ''
+              echo '<details><summary>What changed</summary>'
+              echo ''
+              echo '```'
+              cat /tmp/drift-stat.txt
+              echo '```'
+              echo ''
+              echo '</details>'
             fi
           } >> "$GITHUB_STEP_SUMMARY"
 
@@ -840,6 +860,16 @@ jobs:
           if [ -s /tmp/drift.txt ]; then
             echo "::error::CodeGen output DIFFERS from committed artifacts (drift). Files:"
             cat /tmp/drift.txt
+            # A filename alone does not say WHAT moved, and these artifacts are megabytes — so the
+            # only way to act on this failure used to be reproducing the whole job locally. Emit the
+            # shape of the change (per-file +/-) and a bounded excerpt instead. Bounded because a
+            # full regeneration of __mj.ts is ~100k lines and would bury the log.
+            git diff --stat -- packages/ metadata/ migrations/ > /tmp/drift-stat.txt || true
+            git diff -- packages/ metadata/ migrations/ | head -c 100000 > /tmp/drift-diff.txt || true
+            echo "--- change shape ---"
+            cat /tmp/drift-stat.txt
+            echo "--- first 100KB of diff ---"
+            cat /tmp/drift-diff.txt
             exit 1
           fi
           echo "No drift — committed generated artifacts match a clean-DB CodeGen run."
@@ -859,5 +889,15 @@ jobs:
               echo '```'
               cat /tmp/drift.txt
               echo '```'
+            fi
+            if [ -s /tmp/drift-stat.txt ]; then
+              echo ''
+              echo '<details><summary>What changed</summary>'
+              echo ''
+              echo '```'
+              cat /tmp/drift-stat.txt
+              echo '```'
+              echo ''
+              echo '</details>'
             fi
           } >> "$GITHUB_STEP_SUMMARY"
--- a/packages/CodeGenLib/src/Database/manage-metadata.ts
+++ b/packages/CodeGenLib/src/Database/manage-metadata.ts
@@ -5471,6 +5471,15 @@ export class ManageMetadataBase {
          const generationPromises = [];
          const ag = new AdvancedGeneration();
 
+         // `skipDBUpdate` means load-only: `runValidationGeneration` is called below with
+         // `generateNewCode = false`, and `generateValidatorFunctionFromCheckConstraint` only reaches an
+         // LLM when that flag is true. Reading a validator back out of an Approved `GeneratedCode` record
+         // is therefore a plain database read — gating it on the AI feature flag is what made
+         // `mj codegen --no-ai` DELETE every committed `Validate()` override rather than preserve it, and
+         // the `codegen-drift` gate (which runs `--no-ai`) then demanded that lossy output. Generation
+         // stays gated; only the read is unconditional.
+         const emitValidators = skipDBUpdate || ag.featureEnabled('ParseCheckConstraints');
+
          const columnLevelResults = result.filter((r: any) => r.EntityFieldID); // get the column level constraints
          const tableLevelResults = result.filter((r: any) => !r.EntityFieldID); // get the table level constraints
          for (const r of columnLevelResults) {
@@ -5507,8 +5516,8 @@ export class ManageMetadataBase {
                else {
                   // if we get here that means we don't have a simple condition in the check constraint that the RegEx could parse. If Advanced Generation is enabled, we will
                   // attempt to use an LLM to do things fancier now
-                  if (ag.featureEnabled('ParseCheckConstraints')) {
-                     // the user has the feature turned on, let's generate a description of the constraint and then build a Validate function for the constraint 
+                  if (emitValidators) {
+                     // either we are loading persisted validators, or the feature is on and we may generate new ones
                      // run this in parallel
                      generationPromises.push(this.runValidationGeneration(r, allEntityFields, !skipDBUpdate, currentUser));
                   }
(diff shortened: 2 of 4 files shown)


.github/workflows/integration.yml
```diff
@@ -716,6 +716,16 @@ jobs:
           if [ -s /tmp/drift.txt ]; then
             echo "::error::CodeGen output DIFFERS from committed artifacts (drift). Files:"
             cat /tmp/drift.txt
+            # A filename alone does not say WHAT moved, and these artifacts are megabytes — so the
+            # only way to act on this failure used to be reproducing the whole job locally. Emit the
+            # shape of the change (per-file +/-) and a bounded excerpt instead. Bounded because a
+            # full regeneration of __mj.ts is ~100k lines and would bury the log.
+            git diff --stat -- packages/ metadata/ migrations/ > /tmp/drift-stat.txt || true
```
rkihm-BC, line 723: Follow-up, not blocking — filed so it is not lost.

This new diagnostic prints nothing for one whole class of drift, which is the class this commit most wants to explain.

Drift is detected eight lines up with:

```sh
git status --porcelain -- packages/ metadata/ migrations/ > /tmp/drift.txt
```

`git status --porcelain` **includes untracked files** (`?? path`). `git diff --stat` and `git diff` ignore them entirely. So when CodeGen emits a file that is not yet committed, `drift.txt` correctly lists it, the gate correctly fails — and the log then prints:

```
--- change shape ---
--- first 100KB of diff ---
```

with nothing under either heading, which reads like a bug in the gate rather than an explanation of the drift.

That untracked artifacts really do turn up here is settled by this job's own code: the `CodeGen_Run_*.sql` sweep four lines above exists specifically because CodeGen leaves untracked files behind on every run.

One line before the diffs fixes it — `-N` stages intent-to-add, so untracked files appear in `git diff` as additions without actually staging content:

```sh
git add -A -N -- packages/ metadata/ migrations/ || true
```

Same applies to the weekly gate's copy at line 867.

packages/CodeGenLib/src/Database/manage-metadata.ts
```diff
@@ -5471,6 +5471,15 @@ export class ManageMetadataBase {
...
          const ag = new AdvancedGeneration();
 
+         // `skipDBUpdate` means load-only: `runValidationGeneration` is called below with
+         // `generateNewCode = false`, and `generateValidatorFunctionFromCheckConstraint` only reaches an
+         // LLM when that flag is true. Reading a validator back out of an Approved `GeneratedCode` record
+         // is therefore a plain database read — gating it on the AI feature flag is what made
+         // `mj codegen --no-ai` DELETE every committed `Validate()` override rather than preserve it, and
+         // the `codegen-drift` gate (which runs `--no-ai`) then demanded that lossy output. Generation
+         // stays gated; only the read is unconditional.
+         const emitValidators = skipDBUpdate || ag.featureEnabled('ParseCheckConstraints');
```
rkihm-BC, line 5481: Follow-up, not blocking — filed so it is not lost.

This predicate is the whole fix, and nothing in the repo asserts it. The PR changes four files and none of them is a test.

What makes that worth raising rather than shrugging at is the PR's own framing: this is the **second** time the same behaviour regressed, and the text notes that *"Both times that regression shipped, the log said exactly this line and nothing else."* A defect that has now shipped twice, silently, is the canonical case for a cheap unit test.

The drift gate does cover it end-to-end, which is why this is not a blocker — but that gate needs Docker, SQL Server, 78 migrations and ~10 minutes, and it is also the thing that was inverted last time (it required the lossy output). A test at this seam needs none of that: with `skipDBUpdate = true` and `ParseCheckConstraints` disabled, a validator must still be queued. CodeGenLib already has specs that stub `ManageMetadataBase` (`entity-subclass-codegen.test.ts`, `idempotency/generated-code-sql.test.ts`), so the fixture pattern exists.

A red-proof test here would have failed before this PR and passed after, which is exactly the evidence the next person will want when this code is touched again.

packages/CodeGenLib/src/runCodeGen.ts
```diff
@@ -496,7 +496,13 @@ export class RunCodeGenBase {
         pipelineSuccess = false;
         return false;
       } else {
-        succeedSpinner('AI Generated Code loaded from Metadata');
+        // Report the COUNT, not just success. Loading zero validators is a legitimate state for a
+        // database that has none, and an invisible catastrophe for one that has plenty: file
+        // generation emits each entity as though it had no `Validate()` override, silently deleting
+        // whatever was committed. Both times that regression shipped, the log said exactly this
+        // line and nothing else. A number here makes the next one visible in CI output.
+        const loadedValidators = ManageMetadataBase.generatedValidators.length;
```
rkihm-BC, line 504: Follow-up, not blocking — filed so it is not lost. Nit on an otherwise good addition.

`generatedValidators` is the raw `_generatedValidators` array, and on an AI-enabled full run it is pushed **twice** for the same validator: once by `manageMetadata`'s pass (`:4375`, `generateNewCode = true`, which returns the persisted record early without calling the LLM when the constraint text matches) and again by `loadGeneratedCode` here. Dedup happens later, in `GenerateValidateFunction`, keyed on `functionName`.

So this line reports roughly double on a full AI run. It is exact on `--no-ai` — the case that matters, and the case CI exercises, where it just printed `(116 validators)` — and the double-push predates this PR, so nothing is broken by it.

Raising it only because the number itself is new and its entire purpose is diagnostic accuracy: someone comparing a local AI run against a CI run will see two different counts for the same database and reasonably wonder which one is lying. Counting distinct `functionName`s, or a word in the message saying the figure is pre-dedup, would settle it.

rkihm-BC (approved):
Approving. All 21 checks pass, and the one that decides this PR — `CodeGen drift gate` — is green with the codegen step genuinely executed (step 12 `Run CodeGen against the migrated DB` → success, step 14 `Diff committed generated artifacts` → success), not skipped.

The CI log carries its own proof:

```
✔ AI Generated Code loaded from Metadata (116 validators)
```

That is the `--no-ai` path in the real gate, printing the number the new line was added to print. Before this change the same line would have read `(0 validators)` and said `✔` anyway. The count matches the PR's local table exactly (116 loaded → 56 `Validate()` overrides).

The load-bearing claim is "generation stays gated; only the read is unconditional." If that were wrong, `--no-ai` would start making LLM calls inside a CI job with no credentials, so I traced it instead of reading it.

It holds, and more strongly than the description says:

- `loadGeneratedCode` hardcodes `skipDBUpdate = true` (`manage-metadata.ts:5555`), so `emitValidators` is unconditionally true on the load path. The fix actually reaches the path CI runs — which was worth confirming, because `--no-ai` and `--skipdb` are different flags and the gate passes only the former.
- The LLM call is guarded by `if (generateNewCode && ag.featureEnabled('ParseCheckConstraints'))`. On the load path `generateNewCode = !skipDBUpdate = false`, so the LLM is unreachable **even if the flag were on**. Double-gated.
- The non-load call site (`:4375`, `skipDBUpdate = false`) evaluates to `false || featureEnabled` — identical to before. This genuinely removes a special case rather than adding one, as claimed.
- The dedup note in the new `runCodeGen` comment is real: `GenerateValidateFunction` dedups by `functionName` over a deterministic sort, so the double-push on an AI run yields one emission.

The history checks out independently too: 56 `Validate()` overrides on `next` today, **0** at `6ac6bcc5ca`, Integration Tier `failure` on `next @ 57975fdaee`. And the daily gate runs `codegen --no-ai --skip-commands` (line 694) while the weekly AI gate (line 838) omits `--no-ai` — which is exactly why only one of the two broke, and is the detail that makes the diagnosis convincing.

The workflow claim holds as well. Every added line sits inside the `if [ -s /tmp/drift.txt ]` block, after `cat` and before an untouched `exit 1`, and both git commands carry `|| true`, so pass/fail is unreachable from that commit.

Changeset is `patch`, correct for a branch with no migration and no metadata.

## Also worth saying

The "why the previous proof was wrong" section is the most valuable part of this PR. *"An empty `git diff` after CodeGen is not evidence unless the artifact's mtime moved"* is a trap that would catch anyone verifying CodeGen locally on an already-migrated database, and writing it down is worth more than the fix.

## Three follow-ups, none blocking

All three are inline, and I have filed them so they are not lost. None of them should hold up the build.

PR #4378, opened 2026-09-10
fix(migrations): drop literal EntityField Sequence from Conversation_Scoped_Skill_Activation

SDesai-BC:
**Found while preparing the `v6.1.0-edge.6` release. This would have blocked the release PR into `main`.**

## What's wrong

`V202609031400__v6.1.x__Conversation_Scoped_Skill_Activation` inserts 12 `EntityField` rows with a **literal `Sequence`** — 1–11 on `MJ: Conversation Skills`, 14 on `MJ: AI Skills`.

Each INSERT is preceded by the `+100000` sequence park, which is why a from-scratch `mj migrate` still succeeds today. But the park is the mechanism **#4292 removed**, and its own changeset says why:

> The park could be made idempotent within one run **but not across two migrations replayed on a fresh database**; CodeGen now emits an apply-time `MAX(Sequence)` expression instead.

Its `NOT EXISTS (Sequence >= 100000)` guard makes a *second* migration's park a silent no-op, and the next literal insert then collides on `UQ_EntityField_EntityID_Sequence`. Without `SET XACT_ABORT ON` that surfaces later as an unrelated FK error against `EntityFieldValue` — and only ever on a fresh install, never on a working dev database.

## Why it surfaced only now

The migration merged to `next` on **2026-09-03**. The gate that detects the pattern landed **2026-09-08** (#4292). `changes.yml` diffs against `pull_request.base.ref`, so on a PR into `next` this file is already on the base and invisible to it. It is only ever scanned against `main` — on a release PR. I hit it running that comparison locally while cutting edge.6.

## The fix

The literals become the apply-time expression CodeGen now emits, one per INSERT:

```sql
(SELECT COALESCE(MAX([Sequence]), 0) + 1
   FROM [${flyway:defaultSchema}].[EntityField]
  WHERE [EntityID] = '<entity-id>')
```

**The parks are deliberately left in place**, matching `V202609081111__v6.1.x__Entity_SubtypeSelector`, which already pairs a park with an apply-time expression. Park-plus-expression is safe where park-plus-literal was not: `MAX + 1` cannot collide whatever the park did or skipped. Leaving them also keeps the diff to the 12 value lines.

## Verified — no behaviour change

Two databases built from scratch with `mj migrate` (78 migrations each), one on the original file and one on this commit:

| | original | this commit |
|---|---|---|
| `MJ: Conversation Skills` | 11 fields, seqs `1..11` | 11 fields, seqs `1..11` |
| `MJ: AI Skills`.`ActivationScope` | 14 | 14 |
| parked rows (`>= 100000`) | 0 | 0 |

Identical — the repeatable renumber normalises the sequences either way. This removes a latent fresh-install collision without altering any outcome.

`check-migration-entityfield-sequence.mjs` now **passes** (was: 12 findings).

## Note on the PostgreSQL counterpart — deliberately not touched

`migrations-pg/v6/V202609031400__…pg.sql` mirrors the same literals. I left it alone:

- The gate scans `migrations/` only, so nothing is blocked by it.
- Both dialects produce **identical** sequences today — the risk is latent in both, and pre-existing in both.
- That file is `--split --bake-codegen` output and says so at the top; regenerating it needs a live PG database seeded to the state immediately before this migration, and per `migrations/CLAUDE.md` PG is converter territory, not hand-authored.

Happy to regenerate it as release-time toolchain work if you'd rather have the two dialects converge now — say the word.

### DIFF
--- a/.changeset/entityfield-literal-sequence-conversation-skills.md
+++ b/.changeset/entityfield-literal-sequence-conversation-skills.md
@@ -1,0 +1,21 @@
+---
+"@memberjunction/core-entities": minor
+---
+
+Migration `V202609031400__v6.1.x__Conversation_Scoped_Skill_Activation` no longer inserts `EntityField` rows with a literal `Sequence`.
+
+Twelve `EntityField` INSERTs carried literal sequence numbers (1–11 on `MJ: Conversation Skills`, 14 on `MJ: AI Skills`). Each was preceded by the `+100000` sequence park, which is why a from-scratch `mj migrate` still succeeds today — but the park is the mechanism `#4292` removed, and its own changeset records why: it "could be made idempotent within one run but not across two migrations replayed on a fresh database." Its `NOT EXISTS (Sequence >= 100000)` guard makes a *second* migration's park a silent no-op, and the next literal insert then collides on `UQ_EntityField_EntityID_Sequence` — surfacing as an unrelated foreign-key error against `EntityFieldValue`, and only ever on a fresh install.
+
+The literals are replaced with the apply-time expression CodeGen now emits, one per INSERT:
+
+```sql
+(SELECT COALESCE(MAX([Sequence]), 0) + 1
+   FROM [${flyway:defaultSchema}].[EntityField]
+  WHERE [EntityID] = '<entity-id>')
+```
+
+The parks are left in place, matching `V202609081111__v6.1.x__Entity_SubtypeSelector`, which already pairs a park with an apply-time expression. Park-plus-expression is safe where park-plus-literal was not: `MAX + 1` avoids a collision whatever the park did.
+
+This migration merged to `next` on 2026-09-03; the CI gate that detects the pattern landed on 2026-09-08, and it diffs against the PR's base — so on PRs into `next` the file is already present and invisible to it. It surfaces for the first time on a release PR into `main`, which is where it was caught.
+
+No behaviour change on a fresh install: verified that the resulting sequences are identical, because the repeatable renumber normalises them either way.
--- a/migrations/v6/V202609031400__v6.1.x__Conversation_Scoped_Skill_Activation.sql
+++ b/migrations/v6/V202609031400__v6.1.x__Conversation_Scoped_Skill_Activation.sql
@@ -308,7 +308,7 @@ UPDATE [${flyway:defaultSchema}].[EntityField]
          (
             '5be7bba0-3d9c-46d2-aa89-12b159defbe6',
             'D2021771-50D7-4E0F-A2A8-27AC37E01B34', -- Entity: MJ: Conversation Skills
-            1,
+            (SELECT COALESCE(MAX([Sequence]), 0) + 1 FROM [${flyway:defaultSchema}].[EntityField] WHERE [EntityID] = 'D2021771-50D7-4E0F-A2A8-27AC37E01B34'),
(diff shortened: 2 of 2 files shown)


SDesai-BC:
@rkihm-BC — tagging you because this is your gate from #4292 catching a migration that predates it, and because the fix leans on the reasoning in your `entityfield-sequence-park-idempotent` changeset.

## Why this is safe to merge

**1. The outcome is provably unchanged.** I built two databases from scratch with `mj migrate` (78 migrations each), one on the original file and one on this commit:

| | original | this commit |
|---|---|---|
| `MJ: Conversation Skills` | 11 fields, seqs `1..11` | 11 fields, seqs `1..11` |
| `MJ: AI Skills`.`ActivationScope` | 14 | 14 |
| parked rows (`>= 100000`) | 0 | 0 |

Identical. The repeatable renumber normalises the sequences either way, so this removes a latent collision without moving a single number.

**2. CodeGen output is unchanged.** CodeGen generates from database state, not migration text, and the state is identical — but I verified rather than assumed. Ran CI's own sequence against the fixed migration's database (`migrate` → `sync push --ci` → `codegen --no-ai --skip-commands`):

```
✔ AI Generated Code loaded from Metadata (116 validators)
git status --porcelain -- packages/ metadata/ migrations/   ->   (empty)
```

Zero drift against the committed artifacts.

**3. Nothing compiled changed.** The diff is one `.sql` and one `.changeset/*.md`. No package source, so no rebuild is implied.

**4. The gate now passes** — `check-migration-entityfield-sequence.mjs` went from 12 findings to clean.

## Two judgement calls I'd like your read on

**The parks are left in place.** I only replaced the 12 literal values. My reasoning: `V202609081111__Entity_SubtypeSelector` already pairs a park with an apply-time expression, so I matched the in-release precedent — and park-plus-expression is safe where park-plus-literal was not, because `MAX + 1` cannot collide whatever the park did or skipped. If you'd rather the deprecated parks come out entirely now that `parkEntityFieldSequencesSQL` is gone from CodeGen, say so and I'll strip all three.

**The PG counterpart is untouched.** `migrations-pg/v6/V202609031400__…pg.sql` mirrors the same literals. I left it because the gate scans `migrations/` only, both dialects produce identical sequences today, and that file is `--split --bake-codegen` output whose regeneration needs a live PG database seeded to the state immediately before this migration — which `migrations/CLAUDE.md` puts in converter territory, not hand-editing. Happy to regenerate it as release-time toolchain work if you'd prefer the dialects converge in this PR.

## One thing worth knowing about your gate

It works exactly as designed — but its reach is narrower than it looks. `changes.yml` diffs against `pull_request.base.ref`, so on a PR into `next` any migration already on `next` is invisible to it. This file merged 2026-09-03, five days before the gate landed, and no PR into `next` could ever have flagged it. It only becomes visible on a release PR into `main`, which is where I hit it while cutting `v6.1.0-edge.6`.

That means the gate is forward-only in practice, and anything that merged before 2026-09-08 is unscanned until it reaches a release PR. I checked the rest of the release: this was the only file with literals — `V202608301800` (already on `main`) and `V202609081111` both use the apply-time expression. So the backlog is exactly one file and it's this one. Might still be worth a one-off sweep of `migrations/` on `next` so the next build engineer doesn't discover the next one at release time.

rkihm-BC (approved):
Lgtm

PR #4380, opened 2026-09-10
fix(codegen): treat "zero integrity checks ran" as its own outcome; correct §4 and the scope lesson in the prune-replay plan

AN-BC:
> **Rewritten after merging `next`.** The PR's original headline claim — "CodeGen integrity checks could not fail the run" — **is no longer true**: `next` fixed it independently in `fcf9169f`, and did so more strictly than this branch proposed. What survives is the case that fix did not cover, plus corrections to the plan doc. The original description is preserved at the bottom.

## What `next` already does, and why this branch had to give ground

`fcf9169f` made `runCodeGen` consume `RunIntegrityChecks`'s results: it fails the spinner, sets `pipelineSuccess = false`, and records every failure to the run reporter and `commandFailures`. Unconditionally — no config gate.

This branch proposed the same fix behind a new `integrityChecks.failOnError`, **defaulting to `false`**, and on that path set nothing. Merging it as written would have re-reported `success: true` over a failed check — the exact defect the branch set out to kill, arriving by a clean auto-merge on a stale branch. So:

- The conflict in `runCodeGen.ts` is resolved to **`next`'s block verbatim** (`git diff 98e37e46 origin/next -- packages/CodeGenLib/src/runCodeGen.ts` is empty — verified, not asserted).
- **`failOnError` is removed, not merged.** Nothing reads it, and a config key whose default decides whether a repo silently stops failing is worse than no key. No opt-out replaces it: these checks report `EntityField` Sequence drift, which corrupts the column ordering `spCreate`/`spUpdate` depend on, so there is no state of the world in which the finding should be reported and the run still green. `integrityChecks.enabled` remains the supported way to opt out, and it now says so in the output instead of passing silently.

## 1. `none-ran` is a third outcome

With `next`'s fix in place the block still had two branches where there are three. An **empty** result array satisfies "no failures", so it printed the same green tick as a real pass. `RunIntegrityChecks(conn, true)` runs only the checks whose `Enabled` is true, so it returns `[]` whenever `integrityChecks.enabled` or `entityFieldsSequenceCheck` is false — which is the obvious way to quiet a check that is failing. A green tick over nothing measured is the same defect as discarding the results, one line further along.

- New **`SystemIntegrityBase.ClassifyResults()`** returns `none-ran` / `passed` / `failed`. It is pure and unit-tested. The inline version could not be tested without standing up the whole pipeline, which is why the case went unnoticed in the first place.
- `none-ran` **warns** and writes a `reporter.note` — it is not a failure, because disabling the checks is a legitimate, deliberate configuration, and failing on it would break every repo that has made that choice. The `reporter.note` is what carries it into the structured run report, which a spinner line does not reach.
- **It cannot fire spuriously**: `configInfoSchema` gives `integrityChecks` a schema-level `.default({ enabled: true, entityFieldsSequenceCheck: true })`, so a repo that omits the key (MJ itself included) still gets both on. The warning only appears when someone deliberately turns a check off.
- A failure now **names** the failing checks in the spinner line instead of counting them (the spinner is what survives non-verbose output); the success line reports how many checks actually ran.
- `config.test.ts` pins the **absence** of a `failOnError`-style knob. The knob is an easy thing to add in good faith later; making its return require deleting a test that states the reason is the point.

`runCodeGen.ts` is the only caller of `RunIntegrityChecks` in the repo — checked, so there is no second site with the same defect.

## 2. `plans/entityfield-prune-replay-defect.md`

**§4 was overstated in the original draft and the retraction stands — verified, not assumed.** I re-read the current `spDeleteUnneededEntityFields` body: it left-joins `vwEntityFields` against `vwSQLColumnsAndEntityFields` on `EntityID` + name and deletes only where `actual.column_id IS NULL`. On a settled database it is a no-op; the destruction is specific to mid-replay, before the columns exist. Heals are durable.

**The residual figure was wrong when written, not stale.** The draft said "60 committed migration files." It was **61** at this branch's merge-base and is 61 now. It also conflated two quantities — 19 of those files only *define* the procedure or mention it in a comment. The real shape:

| | |
|---|---|
| files mentioning the procedure | **61** |
| `EXEC` call sites | **49**, across **42** files |
| call sites passing `@IncludedSchemaNames` | **0** (all carry `@ExcludedSchemaNames='sys,staging'` and nothing else) |

The paragraph now ships the `git grep` commands that produce each number, so the next reader re-measures rather than re-quotes. It also notes that `V202608260829__v6.1.x__Heal_SPs_IncludedSchemaNames.sql` **already added** `@IncludedSchemaNames` to `spDeleteUnneededEntityFields` — so retiring that hazard class is a call-site change, not a procedure change.

**The scope bullet's evidence re-measured clean, with one claim withdrawn.** more-cheese: 166 `EntityField` literals across 12 entities; bizapps-common: 121 across 10; **zero** duplicate `(EntityID, Sequence)` pairs in either. The original description's "zero duplicates in any of the three app baselines" is withdrawn — bizapps-orders' baseline carries **no `EntityField` rows at all** (they live in `V202607061432__…__Tables_and_Objects.sql`, which a `V`-only gate already covers), so "no duplicates there" was vacuous and proved nothing either way.

## Verification

The original description said the branch was never compiled. It is now, and CI has since confirmed it.

**CI on `ff65d410`: 27 checks, 23 success / 4 skipped, none failing.** Including Build, all six unit-test shards, Source guards, adopted standards, Native-ESM import guard, Check migrations, dependency check, CodeQL, **CodeGen drift gate** (which exercises a real CodeGen run through the path this PR changes), and **Integration (SQL Server, deterministic)**.

Locally:

- `turbo build --filter=@memberjunction/codegen-lib...` → 134/134 successful; `tsc --noEmit` in `packages/CodeGenLib` exits 0.
- New + config tests → 46 passed. `check:changeset`, `check:standards`, `check:esm` all pass.
- `pnpm-lock.yaml` was touched by my local install and has been **restored to `next`'s exact content** — it is not in this diff.
- One honest caveat, since it would otherwise look alarming: my sandbox reports 3 failures in `entity-subclass-embedded.test.ts` (`ENOENT … '/out/entity_subclasses.ts'`). They are **not this PR's**: they reproduce identically in a clean worktree at pristine `origin/next`, and nothing in this diff touches `generateAllEntitySubClasses` or any fs path. They are also **not a repo defect** — this diff changes CodeGenLib, so CI's shards ran that suite cache-busted and it came back green. The failures are specific to my container.

## Commits

1. `98e37e46` — the merge, resolved to `next`'s semantics, with the orphaned `failOnError` removed.
2. `ff65d410` — the review corrections above.

Companion PRs (same branch name in each repo): MemberJunction/bizapps-orders, MemberJunction/bizapps-common, MemberJunction/more-cheese.

<details>
<summary>Original description (superseded — its item 1 no longer describes this change)</summary>

Two of the six items left outstanding when the PR #4296 family merged. Both are cases of a check or a claim that could not do its job.

**1. CodeGen integrity checks could not fail the run (D.6).** `runCodeGen.ts` called `SystemIntegrityBase.RunIntegrityChecks(conn, true)` and discarded the returned array. A failing check printed `Integrity check FAILED` and the pipeline continued to a green exit. The proposed fix added `integrityChecks.failOnError`, defaulting to `false`. *(Superseded: `next` fixed the discard in `fcf9169f` without a config gate; `failOnError` is removed above.)*

**2. `plans/entityfield-prune-replay-defect.md` — §4 and the Phase 4 scope bullet.** §4 was overstated: it claimed a heal is "undone by a subsequent core upgrade," which reading the procedure disproves. The Phase 4 bullet was rewritten from "gates should be broad" to "gate scope is a claim, and the right claim differs per gate." *(Both retained; counts corrected above.)*

**⚠️ Not compiled or tested locally** — the checkout had no `node_modules`. *(Superseded: built, tested, and green in CI — see Verification.)*

</details>

https://claude.ai/code/session_01XD2fRn3AW4N5KwaPF6E4RC

### DIFF
--- a/.changeset/codegen-integrity-checks-can-fail.md
+++ b/.changeset/codegen-integrity-checks-can-fail.md
@@ -1,0 +1,19 @@
+---
+"@memberjunction/codegen-lib": patch
+---
+
+CodeGen's system integrity checks can now actually fail.
+
+`runCodeGen` called `SystemIntegrityBase.RunIntegrityChecks()`, discarded the returned
+`IntegrityCheckResult[]`, and printed a success tick unconditionally — so a run that had just logged
+`Integrity check FAILED: entityFieldsSequenceCheck` went on to report `success: true, errors: []`.
+A nightly drift-gate run has carried exactly that, on green, for weeks.
+
+The results are now consumed: failures are reported as failures (no success tick, each failing check
+named with its message), and `integrityChecks.failOnError` decides whether they stop the run.
+
+`failOnError` defaults to `false` because the `MJ: Entities` Sequence drift these checks report is
+live on `next` today; flipping it in the same change would red every CodeGen run before anyone could
+act on the finding. Honest reporting is not gated on it. Flip the default once that drift is
+resolved — that is the last step of this fix, and a test pins the default so it stays a deliberate
+act rather than a tidy-up.
--- a/packages/CodeGenLib/src/Config/config.ts
+++ b/packages/CodeGenLib/src/Config/config.ts
@@ -288,6 +288,22 @@ export type IntegrityCheckConfig = z.infer<typeof integrityCheckConfigSchema>;
 const integrityCheckConfigSchema = z.object({
   enabled: z.boolean(),
   entityFieldsSequenceCheck: z.boolean(),
+  /**
+   * Whether a failing integrity check fails the CodeGen run.
+   *
+   * Until this existed the checks could not fail at all: `RunIntegrityChecks` built a result per
+   * check and `runCodeGen` discarded the array, printed a success tick, and reported
+   * `success: true, errors: []`. A check whose only failure path is a `console.log` is not a check —
+   * a nightly drift-gate run has carried `Integrity check FAILED: entityFieldsSequenceCheck` for
+   * seven `MJ: Entities` fields, on green, for weeks.
+   *
+   * Defaults to `false` rather than `true` ONLY because that condition is live on `next` today, and
+   * flipping it in the same change would red every CodeGen run before anyone could act on the
+   * finding. The reporting is truthful either way now (see `runCodeGen`). Flip this to `true` once
+   * the `MJ: Entities` Sequence drift is resolved; that is a one-line change and the last step of
+   * this fix.
+   */
+  failOnError: z.boolean().default(false),
 });
 
 export type ForceRegenerationConfig = z.infer<typeof forceRegenerationConfigSchema>;
@@ -992,6 +1008,7 @@ export const DEFAULT_CODEGEN_CONFIG: Partial<ConfigInfo> = {
   integrityChecks: {
     enabled: true,
     entityFieldsSequenceCheck: true,
+    failOnError: false,
   },
   advancedGeneration: {
     enableAdvancedGeneration: true,
--- a/packages/CodeGenLib/src/__tests__/config.test.ts
+++ b/packages/CodeGenLib/src/__tests__/config.test.ts
@@ -37,6 +37,7 @@ vi.mock('@memberjunction/config', () => ({
 }));
 
 import { parseBooleanEnv } from '@memberjunction/config';
+import { DEFAULT_CODEGEN_CONFIG } from '../Config/config';
 
 describe('Config Types', () => {
     describe('parseBooleanEnv (utility function)', () => {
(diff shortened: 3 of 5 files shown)


AN-BC (commented):
@cadam11 and @rkihm-BC I know there is a problem with CG right now so I'm not merging this, can you guys merge if it looks good to you?

PR #4389, opened 2026-09-11
ci: prove the App token's write access by using it, not by reading .permissions

rkihm-BC:
Follow-up to #4388. The first dispatch answered the question that PR existed to answer, then failed on a check of mine that was simply wrong.

## What run 1 established

**The identity is confirmed.** `APP_PRIVATE_KEY` belongs to `blue-cypress-ci-bot`, installation `156297321`:

```
app-slug        = blue-cypress-ci-bot
installation-id = 156297321
Notice: Identity confirmed: blue-cypress-ci-bot (installation 156297321)
```

That was the one unproven link behind using a GitHub App as the ruleset bypass actor, and it holds. The App route is viable.

## The failure was a bad test, not a bad credential

```
repo permissions = {"admin":false,"maintain":false,"pull":false,"push":false,"triage":false}
Error: The App token cannot push to MemberJunction/MJ.
```

The `permissions` block on a repository object is computed for an authenticated **user**. An installation token has no user behind it, so GitHub returns every field false no matter what the installation actually holds. Asserting `.push == true` against an App token can only ever fail.

The internal contradiction gives it away: `pull: false` on a request that **returned 200**. A token with no access to the repo gets a 404, not a permissions block — so the successful read had already proved access at the moment the check declared there was none.

Independently, the installation object read as an org admin reports exactly:

```json
{ "contents": "write", "pull_requests": "write", "metadata": "read" }
```

with `repository_selection: all`. So the App almost certainly can write here. "Almost certainly" is what this workflow exists to eliminate, hence the replacement rather than simply deleting the check.

## What replaces it

Prove write by performing one. Creating a ref is a `contents: write` operation and is the exact capability publish.yml needs:

- The ref is created **at `next`'s current tip** — no new objects, no branch anyone uses, `next` itself untouched.
- Throwaway name `apptoken-writecheck-<run_id>`, deleted in an `always()` cleanup step.
- Cleanup records the branch name **before** the creation attempt, so a half-succeeded create is still tidied.
- **No CI is triggered.** Every `push:` trigger in this repo is branch-filtered — `next`, `main`, `lts/**`, or `feature/mj-install-v2` — and none match. I checked all 13 rather than assuming.

The header's "pushes nothing" claim is corrected too; the workflow now makes exactly one mutation and says so plainly.

## Still unknown

The bypass-standing step never ran, because the bogus check exited first. So `current_user_can_bypass` on ruleset `905656` is still unmeasured — expected `never` until the bypass actor is added, but unconfirmed.

After this merges, re-run **Verify the release App token**. Expected: identity confirmed again, `contents:write confirmed`, and `can_bypass = never`. That clears the way for adding `{actor_id: 4705146, actor_type: "Integration", bypass_mode: "always"}` to `905656` and re-running for the `always` that says the next release will clear step 22.

### DIFF
--- a/.github/workflows/verify-release-app-token.yml
+++ b/.github/workflows/verify-release-app-token.yml
@@ -1,0 +1,152 @@
+# One-shot diagnostic for the release automation's GitHub App credential.
+#
+# WHY THIS EXISTS
+# ---------------
+# On 2026-09-03 the `next-protect` ruleset (id 905656) swapped its bypass actor from
+# `RepositoryRole 5` (repository admin) to the `break-glass` team. `MJ-GH-bot` — the
+# account behind `MJ_GH_BOT_GITHUB_TOKEN`, which publish.yml uses to push to `next` — is
+# a repository admin but an OUTSIDE COLLABORATOR, so it lost its bypass and cannot be
+# added to a team (that needs org membership, and the org is at 42/42 seats).
+#
+# v6.1.0-edge.6 published fine and then died on step 22 `Merge main into next`:
+#
+#     remote: error: GH013: Repository rule violations found for refs/heads/next.
+#     - Changes must be made through a pull request.
+#     - Required status check "Check migrations" is expected.
+#
+# Recovered by hand in #4382. It will recur on EVERY release until a bypass exists, and
+# it affects four push sites, not just that one — publish.yml lines 728, 775, 1407 and
+# 1599/1643 all end in `git push origin HEAD:next`.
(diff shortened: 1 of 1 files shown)


rkihm-BC:
Superseded by a branch cut from current `next`. #4388 was squash-merged, so this branch's base commit is not an ancestor of `next` and GitHub saw the workflow file as added on both sides — an add/add conflict (`mergeable=false`). Same change, reopened cleanly.

PR #4395, opened 2026-09-11
Review follow-ups to #4358 and #4366: Deny rows are never grants; the dialog contract is written down

cmbrock-BC:
Follow-ups from @rkihm-BC's reviews of #4358 and #4366 — the non-blocking notes, kept out of those PRs so their approvals stood.

**Stacked on #4358 and #4366:** this branch merges both, so until they land the diff here shows their changes too. Once they merge, only the commits below remain (`cf2484c26b`).

## From #4358

- **Deny rows are never grants.** `GetUserRowLevelSecurityInfo`'s new doc said "only rows that GRANT the operation", but a `Type = 'Deny'` row with `CanCreate = 1` and a filter would still have been collected. Now `EntityPermissionInfo.IsDeny` is the one predicate for a Deny row (case- and whitespace-insensitive; a blank `Type` — rows from before the column existed — is Allow), `GetUserPermisions` uses it instead of its inline expression, and both RLS readers skip Deny rows: `GetUserRowLevelSecurityInfo` (#4358) and `UserExemptFromRowLevelSecurity` (#4352, same shape). Unreachable in practice — a user carrying a Deny row fails the permission gate first — but the methods now implement the invariant their docs state.
- **Typed test builders.** `buildPermission` takes `Partial<EntityPermissionInfo> & { RoleID }`, so a mistyped override is a compile error rather than a silent no-op. Three Deny-axis tests: a Deny row's filter is not collected; a Deny row cannot exempt; the `Type` normalisation. The collection test goes red without the guard.
- **Comment drift.** `MaterializationRefresher.entityHasReadRLS` and `ManageMetadataBase.EntityHasRoleReadRLS` no longer claim parity with `GetUserRowLevelSecurityWhereClause`; they say the leak gate is deliberately *wider* than the runtime reader (a leftover filter beside a cleared flag still counts as "protected" there, which errs conservative).

Not in here, on purpose: the `rls-isolation` integration check needs seeded permission fixtures, and the durable fix — an `EntityPermission` server subclass that nulls `<Op>RLSFilterID` when `Can<Op>` is cleared, plus cleaning the nine stale rows — is its own issue as the review said.

## From #4366

- `.input-dialog-content` gets `box-sizing: border-box` on itself and its descendants, like the rating dialog — redundant under the Explorer shell's global reset, load-bearing for a host without it (the 100%-wide inputs carry padding and a border with no local `box-sizing`).
- `MJDialogContainerComponent` documents the contract beside `padding: 0`: the body pads only string content; a component body pads itself; if a third component rediscovers this, add a `padded` setting to `MJDialogSettings` rather than a special case.

## Testing

- `@memberjunction/core` 2329/2329 (the RLS collection suite 8/8; the Deny collection test fails with the guard removed).
- `@memberjunction/materialization` 118/118. `ng-ui-components` and `ng-conversations` build. `codegen-lib`: 654/654 tests pass; 56 test *files* fail to collect in this worktree with and without the change (environment, not the comment edit).
- Changeset: patch on the five packages (no migration).

### DIFF
--- a/.changeset/rls-and-dialog-review-followups.md
+++ b/.changeset/rls-and-dialog-review-followups.md
@@ -1,0 +1,13 @@
+---
+"@memberjunction/core": patch
+"@memberjunction/ng-conversations": patch
+"@memberjunction/ng-ui-components": patch
+"@memberjunction/materialization": patch
+"@memberjunction/codegen-lib": patch
+---
+
+Review follow-ups to #4358 and #4366.
+
+- `EntityPermissionInfo.IsDeny` — one predicate for "this is a Deny row" (case- and whitespace-insensitive; blank Type is Allow), used by `GetUserPermisions` and now by both RLS readers: `UserExemptFromRowLevelSecurity` and `GetUserRowLevelSecurityInfo` skip Deny rows, so a set `Can*` flag on a Deny row is never read as a grant. Unreachable in practice (a user carrying a Deny row fails the permission gate first), but the methods now implement the invariant their docs state. Tests cover the Deny axis with typed builders.
+- The materialization leak gate's comments no longer claim parity with the runtime RLS reader; they say the gate is deliberately wider.
+- Input dialog: `box-sizing: border-box` parity with the rating dialog. The dialog container documents its contract — component bodies pad themselves.
--- a/packages/Angular/Generic/conversations/src/lib/components/dialogs/input-dialog.component.ts
+++ b/packages/Angular/Generic/conversations/src/lib/components/dialogs/input-dialog.component.ts
@@ -53,6 +53,12 @@ import { Component, Input } from '@angular/core';
          against the dialog edges while the header and footer were padded. Same inset as the sibling
          rating dialog, which already compensates for this. */
       padding: 4px 20px 8px;
+      /* Parity with the rating dialog: the Explorer shell's global reset makes this redundant, but a host
+         without that reset would let the 100%-wide inputs (padding + border, no local box-sizing) overflow. */
+      box-sizing: border-box;
+    }
+    .input-dialog-content * {
+      box-sizing: border-box;
     }
     .dialog-message {
       margin: 0 0 16px 0;
--- a/packages/Angular/Generic/ui-components/src/lib/dialog/dialog.service.ts
+++ b/packages/Angular/Generic/ui-components/src/lib/dialog/dialog.service.ts
@@ -177,6 +177,11 @@ import { Component, Input, Output, EventEmitter } from '@angular/core';
 
     .mj-dialog-body {
       flex: 1;
+      /* CONTRACT: the body pads only STRING content (the "> p" rule below). A COMPONENT body arrives with no
+         inset and pads itself — see rating-dialog and input-dialog in ng-conversations, which both apply
+         4px 20px 8px. Padding here would double-pad them and break deliberately full-bleed bodies (About,
+         Profile, the action gallery, the sharing centre). If a third component rediscovers this, add a
+         padded setting to MJDialogSettings rather than a special case. */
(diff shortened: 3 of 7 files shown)


rkihm-BC (dismissed):
Approving — no requested changes. Every note from both reviews is addressed, several beyond what I asked.

**First, something to save you time: the `CONFLICTING` banner is wrong.** It was computed before #4358 (14:43) and #4366 (15:42) merged, and GitHub has not recomputed it in the hours since. I did not take `git merge-tree`'s word for it either — I checked out `origin/next` in a scratch worktree and actually merged this branch onto it:

```
Automatic merge went well; stopped before committing as requested
conflicted paths: (none)
```

So there is nothing to resolve. The routine update-branch before merge (the ruleset requires up-to-date branches) will clear the banner on its own.

## The Deny fix went further than I asked

I offered a choice — add the guard *or* a comment saying Deny is handled upstream. You did the guard, centralised it as `EntityPermissionInfo.IsDeny`, applied it to **both** RLS readers including #4352's `UserExemptFromRowLevelSecurity` (which I'd flagged as retroactively affected), replaced the inline expression in `GetUserPermisions`, and corrected the JSDoc. One predicate, three call sites, no drift left behind.

The normalisation is right: `(this.Type || 'Allow')` maps null and `''` to Allow, `.trim().toLowerCase()` makes `' DENY '` true, and whitespace-only trims to `''` → Allow.

I checked the reachability claim rather than accepting it. It holds — `GetUserPermisions` subtracts Deny from Allow, so a user carrying a Deny row with the flag set fails the permission gate before either reader runs. And both changes fail in the restrictive direction (fewer exemptions, narrower filter sets), so even if the reachability argument were wrong, the outcome is safe rather than permissive. That is the right shape for a defence-in-depth change.

The collection test is red-without-the-guard by construction, not by assertion: the Deny row carries `CanCreate: true` + `FILTER_2_ID` and the test expects `[FILTER_1_ID]` alone.

## The comment fixes are better than deletions

Both drift sites don't merely drop the false parity claim — they state the correct relationship, that the leak gate is deliberately *wider* than the runtime reader and errs conservative. A future reader now learns why the asymmetry exists instead of just not being misled.

## Dialog notes

Both landed. I confirmed the parity claim against `rating-dialog.component.ts:107` and `:115`, and checked the universal selector: the component has no `encapsulation:` override, so Angular's default Emulated scoping keeps `.input-dialog-content *` inside this template — no global leak. The container comment naming the `padded`-on-`MJDialogSettings` escalation is the part that actually stops a third rediscovery, which was the point of the note.

## Verified rather than assumed

- **Changesets**: the new one is `patch` across five packages (correct — no migration), and the two inherited ones are byte-identical to what is already on `next`, so they dedupe on merge rather than double-bumping.
- **The tests actually ran.** All 23 checks pass, but a green check is not proof a suite executed — turbo can replay a cached pass. MJCore's task logged `cache miss, executing`, so the new RLS suite genuinely ran.

## Nothing blocking; two things tracked separately

Deferring the durable fix and the integration check was the right call for this PR. I checked and neither had an issue yet, so I filed them: the `EntityPermission` server subclass plus the nine stale rows, and the `rls-isolation` fixtures. Links in a follow-up comment.

rkihm-BC:
Follow-ups from this PR's "not in here, on purpose" list, filed so they are not lost — both were untracked:

- **#4398** — the `EntityPermission` server subclass that nulls `<Op>RLSFilterID` when `Can<Op>` is cleared, plus cleanup of the nine stale rows you found.
- **#4399** — the `rls-isolation` integration check, with the seeded permission fixtures it needs.

Deferring both from this PR was right; neither blocks it.

rkihm-BC (approved):
LGTM

PR #4401, opened 2026-09-11
fix(cli): dev workspace sees Open App client packages and their shell peers

### LINKED ISSUE #4364: dev workspace: Open App client packages and their shell-provided @angular/* peers are invisible to the generator (page-load resolve failure, green build)
## Summary

`mj dev workspace` has no awareness of Open App **client** packages — the ones registered in `mj.config.cjs` `dynamicPackages.client`. Because nothing in the tree *depends* on them, pnpm never links them, and neither they nor their shell-provided `@angular/*` peers reach the shell's resolution root. The result is a failure that appears only at **page load**, with a green build:

```
[plugin:vite:import-analysis] Failed to resolve import "@angular/elements"
  from ".angular/vite-root/MJExplorer/app.module-SNJ2EB2S.js". Does the file exist?
../../../bizapps-caliber/packages/Angular/src/lib/widget/caliber-widget.element.ts:13:36
```

`@mj-biz-apps/caliber-ng` declares `@angular/elements` as a `peerDependency` for its `<caliber-widget>` custom element. MJExplorer does not declare it. This is exactly the shape `SHELL_PROVIDED_PEERS` already covers for MJ libraries:

```ts
{ Library: '@memberjunction/ng-explorer-service-worker', Peers: ['@angular/service-worker'] }
```

…but the list is MJ-library-scoped, so a developer serving an Open App is never told.

## Why only `@angular/*` breaks

Worth recording, because it is counter-intuitive and it rules out the obvious "just hoist it" fix.

Angular's dev server splits a linked package's bare imports two ways:

- **`@angular/*` are externalized** and handed to Vite's dep optimizer, which resolves them from the **vite root (MJExplorer)** — not from the importing file. Confirmed: `MJExplorer/.angular/cache/<ver>/MJExplorer/vite/deps/` contains `@angular_*.js` and nothing else.
- **Everything else is bundled by esbuild, resolved importer-relative**, so it is found in the importing package's own `node_modules`.

So `caliber-ng` imports five packages that are *also* unresolvable from MJExplorer — `@memberjunction/ng-conversations`, `ng-media-player`, `ng-notifications`, `ng-task-graph-editor`, `@mj-biz-apps/caliber-entities` — and they all load fine. Only `@angular/elements` breaks the page.

**This is not an argument for restoring the hoist block.** `build.ts`'s "Why no hoist block" note (audit 2026-08-13) is right, and the control assertion below guards it. The five bundled imports need nothing; the fix is the shell declaring its one Angular peer, exactly as the module's own doctrine says.

## Reproduction

Deterministic, pure functions from the built CLI. Run with `node --test`:

```js
import { test } from 'node:test';
import assert from 'node:assert/strict';
import { readFileSync } from 'node:fs';
import { BuildShellPeerGuidance, SHELL_PROVIDED_PEERS, BuildNpmrc }
  from '<path to>/MJ/packages/MJCLI/dist/lib/dev-workspace/build.js';

const CALIBER_NG_PKG = '<path to>/bizapps-caliber/packages/Angular/package.json';

test('caliber-ng really does declare @angular/elements as a shell-provided peer', () => {
  const pkg = JSON.parse(readFileSync(CALIBER_NG_PKG, 'utf8'));
  assert.equal(pkg.name, '@mj-biz-apps/caliber-ng');
  assert.ok(pkg.peerDependencies['@angular/elements'], 'premise: it is a peer');
});

test('the generator guidance names @angular/elements for caliber-ng', () => {
  assert.match(BuildShellPeerGuidance().join('\n'), /@angular\/elements/,
    'a developer is never told the shell must declare it — they find out from a page-load stack trace');
});

test('SHELL_PROVIDED_PEERS covers Open App client packages, not just MJ libraries', () => {
  const libraries = SHELL_PROVIDED_PEERS.map((e) => e.Library);
  assert.ok(libraries.some((l) => l.startsWith('@mj-biz-apps/')),
    `only MJ libraries are covered: ${libraries.join(', ')}`);
});

test('generator has some awareness of mj.config dynamicPackages.client', () => {
  const src = readFileSync('<path to>/MJ/packages/MJCLI/src/lib/dev-workspace/build.ts', 'utf8');
  assert.match(src, /dynamicPackages/,
    'client packages registered only in mj.config.cjs are invisible to the generator');
});

test('.npmrc is not the escape hatch (documents the deliberate no-hoist decision)', () => {
  assert.doesNotMatch(BuildNpmrc(), /public-hoist-pattern/,
    'if this ever fails, upstream re-added hoisting and this issue changes shape');
});
```

Result — premise and control pass, the three defect assertions fail:

```
ok 1 - caliber-ng really does declare @angular/elements as a shell-provided peer
not ok 2 - the generator guidance names @angular/elements for caliber-ng
not ok 3 - SHELL_PROVIDED_PEERS covers Open App client packages, not just MJ libraries
not ok 4 - generator has some awareness of mj.config dynamicPackages.client
ok 5 - .npmrc is not the escape hatch (documents the deliberate no-hoist decision)
# pass 2
# fail 3
```

## Current workaround (and why it is not enough)

Hand-edit the **generated, gitignored** parent `package.json` after every regenerate:

```jsonc
"dependencies": {
  "@mj-biz-apps/caliber-ng": "workspace:*",
  "@mj-biz-apps/ats-ng": "workspace:*",
  "@angular/elements": "21.2.22"      // must track pnpm.overrides
}
```

`BuildRootPackageJson` emits no `dependencies` key at all, so every one of these is lost on the next `mj dev workspace --force`, and the failure returns silently at page load.

## Suggested fixes, cheapest first

1. **MJExplorer declares `@angular/elements`** in its own `package.json`. Doctrine-correct, one line, fixes every workspace immediately.
2. **Extend `SHELL_PROVIDED_PEERS`** (or derive it) to cover Open App client packages, so `BuildShellPeerGuidance()` names the requirement at generate time instead of the developer meeting it as a stack trace.
3. **The one that eliminates the class:** teach `dev workspace` to read `mj.config.cjs` `dynamicPackages.client` — register those packages in the parent manifest and validate their unmet peers in `doctor`. That removes *both* hand re-adds (the `-ng` packages and their Angular peers), which is the linking spec's phase-2 territory.

## Environment

- MJ `6.1.0-edge.5`, pnpm `10.33.0`, Angular `21.1.3` (`@angular/elements` resolved to `21.2.22` via `pnpm.overrides`)
- Workspace: `MJ` + `bizapps-caliber` + 4 other Open App repos under one generated parent
- Relevant source: `packages/MJCLI/src/lib/dev-workspace/build.ts` (`SHELL_PROVIDED_PEERS`, `BuildShellPeerGuidance`, `BuildRootPackageJson`, `BuildNpmrc`)

SDesai-BC:
Serving an Open App from a `mj dev workspace` parent failed at **page load**, with a completely green build:

```
[plugin:vite:import-analysis] Failed to resolve import "@angular/elements"
  from ".angular/vite-root/MJExplorer/app.module-SNJ2EB2S.js". Does the file exist?
../../../bizapps-caliber/packages/Angular/src/lib/widget/caliber-widget.element.ts:13:36
```

Closes #4364.

## Root cause

The generator derives the parent manifest **only** from members' committed `package.json`, `pnpm-workspace.yaml` and lockfiles. Open App *client* packages never travel that road. A host registers them, `mj codegen manifest --open-app-client-bootstrap` appends a side-effect import to the shell's **generated** class-registrations manifest, and the shell ends up importing a package it never declared:

```ts
// packages/MJExplorer/src/app/generated/class-registrations-manifest.ts  (generated, gitignored)
import * as __openAppClient0 from '@mj-biz-apps/caliber-ng';
```

Nothing in the tree *depends* on it, so pnpm — correctly — never links it, and neither it nor its shell-provided `@angular/*` peers reach the shell's resolution root. `grep -rn "dynamicPackages|mj.config"` across all twelve `lib/dev-workspace/` modules returned **zero hits**.

The only thing making this work on a developer machine today is a hand-edit to the **generated, gitignored** parent `package.json`. `BuildRootPackageJson` emitted no `dependencies` key at all, so every `--force` deleted it and the page broke again, silently.

This is not caliber-specific. `bizapps-orders` declares the identical peer and its own `client: [{ role: "bootstrap" }]`; it is simply excluded from the workspace today.

## What this does

**Reads each member's committed `mj-app.json`** and registers every `packages.client[]` and `packages.shared[]` entry — at every role — as parent `dependencies` at `workspace:*`. That set mirrors the host's own emitter: `GetClientPackagesFromManifest` builds `dynamicPackages.client` as `[...client, ...shared]` with no role filter, so anything narrower would leave a package the shell imports and nothing links. `packages.server[]` is excluded — it goes to `dynamicPackages.server`, a Node process that resolves importer-relative, not from the vite root. `detect.ts` already tested that file for existence as a detection reason — it had never opened it.

**Reports unmet shell peers per shell**, with the version the parent already pins. It does not add them.

**Adds the doctor check** `OPEN_APP_WORKSPACE_LINKING_SPEC.md` marks **v1** ("every package named in a host's `dynamicPackages.client[]` MUST resolve") and that was never built — the 10th check, after 9.

**MJExplorer declares `@angular/elements`** (separate commit, independently revertible).

## Three decisions worth defending

**Why `mj-app.json` and not `mj.config.cjs` `dynamicPackages.client`, which the issue suggested.** `mj.config.cjs` is the *host's* registration state, rewritten by `mj app install`/`enable`/`disable` — MJ's committed one currently carries no `dynamicPackages` at all while the generated manifest still imports three client packages. `mj-app.json` is the app's own declaration, tracked in its repo. It also respects the linking spec's axiom (§17, *linking ⊥ registration*): driving linking off registration config couples the two halves the spec separates. Reading the app's declaration instead deliberately over-links — a client package is linked whether or not a host registered it, which is the sanctioned direction ("resolve but do not load").

**Why peers are reported, not auto-added.** The generated `.npmrc` sets `auto-install-peers=true`, which raises a fair question: is registering the client package enough on its own? Measured in an isolated parent, with a control:

| | parent-root `node_modules/is-odd` | resolvable from the shell |
|---|---|---|
| client package is a parent dependency, peer unmet | absent | **UNRESOLVABLE** |
| peer also a parent dependency (control) | present | resolves |

pnpm installed the peer into the **consumer's own tree**, not the parent root. That mirrors caliber exactly — it *has* `@angular/elements` in its own `node_modules` and it does nothing, because an Angular dev server externalizes `@angular/*` and resolves them from the **vite root**. So the peer half is a genuinely separate requirement, and which Angular surface a shell provides stays the shell's decision, exactly as `SHELL_PROVIDED_PEERS` has always said and as MJExplorer already does for `@angular/service-worker`.

**Why the gap rule is shell-scoped.** Two simpler rules were tried against the real workspace and rejected:

| Rule | Result |
|---|---|
| peer not provided by any member | ~11 false positives per client package (`@angular/core`, `rxjs`, `ag-grid-*`…) |
| …and not declared by any non-client workspace package | **zero gaps — a false negative.** `mj-angular-elements-demo` declares `@angular/elements` and masks MJExplorer's real gap |
| …and not declared by a given app shell, reported **per shell** | correct, and has no masking failure mode |

Shell detection is structural, no configuration: depends on `@angular/core` (a library takes it as a *peer*, so this alone separates app from library), carries an Angular builder in devDependencies, **and** declares at least one `@memberjunction/ng-*` package. That third condition is not decoration — without it the demo app reported five peers of packages it will never load, on every regenerate, which is how a report teaches people to stop reading it. The demo declares **0** `@memberjunction/ng-*` packages, MJExplorer declares **22**, and `caliber-ng` peer-depends on **11** of them. A predicate, deliberately not a count threshold.

## Evidence

Unit tier, `packages/MJCLI`: **823 passing across 44 files** (805 before; +18 new). `tsc --noEmit` clean, `pnpm run build` clean. Nothing outside MJCLI imports `dev-workspace/*`.

Against the six live members of a real joined workspace:

| | before | after |
|---|---|---|
| parent `dependencies` | *key absent entirely* | 5 client packages at `workspace:*` |
| shells detected | — | `mj_explorer` |
| shell peer gaps | *never computed* | **none** |

Negative control — remove `@angular/elements` from MJExplorer again and the generator says:

```
shell mj_explorer does not declare @angular/elements (^21.1.3), needed by
@mj-biz-apps/caliber-ng — parent pins 21.2.22
```

Red and green both demonstrated, so the check is live rather than vacuous.

**Not run: the deterministic integration tier.** It drives a shared database and a shared API host, and this change is a workspace-file generator with no DB, resolver or runtime surface. Happy to run it on request.

## Two notes for the reviewer

**Two of the issue's five assertions stay red, by design.** They assert that `SHELL_PROVIDED_PEERS` gains an `@mj-biz-apps/` entry and that the static `BuildShellPeerGuidance()` text names `@angular/elements` — i.e. they encode the hardcoding approach this rejects, because MJ's CLI should not carry knowledge of specific Open App repos. The developer still learns about `@angular/elements` at generate time, through a warning naming the shell, peer, range and pin. A third assertion (`build.ts` mentions `dynamicPackages`) flipped green only because a doc comment mentions it while explaining what this reads *instead* — incidental, not evidence the config is read.

**A separate bug, filed elsewhere:** `@mj-biz-apps/caliber-ng` declares `@angular/elements` in `dependencies` **and** `devDependencies` **and** `peerDependencies`. The first contradicts the third and buys nothing, while making the package look covered. `@mj-biz-apps/orders-ng` has the correct dev + peer shape. That is a `bizapps-caliber` change and fixing it would not have fixed this.

https://claude.ai/code/session_01AhukhE217HNvNszidxHXnc


rkihm-BC (dismissed):
Approving. Strong work — I verified the load-bearing claims rather than trusting them, and every one held up.

**Two housekeeping changes I made directly, so you know what moved on the branch:**
- Merged `next` (`ba7d1ab4af`). It was one commit behind — #3367, Field-Level Security — which touches nothing under `lib/dev-workspace`; the merge was clean.
- Corrected one sentence in the description: it said the generator registers `role: bootstrap` client packages, which is the earlier revision `fa42c8241` deliberately replaced. It now states the shipped rule — every `packages.client[]` and `packages.shared[]` entry at every role, mirroring `GetClientPackagesFromManifest` — matching what the changeset and guide already said.

### Verified

- **The gap is exactly as described.** 12 modules in `lib/dev-workspace/`, zero pre-existing hits for `dynamicPackages|mj.config` across them on `next`.
- **`mj-app.json` over `mj.config.cjs` is the right read.** §17 of `OPEN_APP_WORKSPACE_LINKING_SPEC.md` supports it verbatim, and in the failing case the host is the MJ repo itself, whose `mj.config.cjs` carries no `dynamicPackages` — reading it would have found nothing.
- **No role filter is correct.** `GetClientPackagesFromManifest` (`packages/OpenApp/Engine/src/install/config-manager.ts:349`) builds `[...clientPkgs, ...sharedPkgs]` with no role test.
- **Peers reported, not auto-added, is the right call — and there's corroboration outside this repo.** aidp-next's most recent commit is `fix(explorer): declare @angular/cdk — orders-ng and sales-ng now peer-require it`, after a deploy build failed on `Could not resolve "@angular/cdk/overlay"`. Same class of gap, fixed in the same place (the shell's tracked `package.json`). That host hit it at build time rather than page load only because it's a production esbuild bundle rather than a vite dev server; the mechanism is identical.
- **`findPeerPin` catches a real trap** — a bare `overrides[peer]` lookup would miss per-major selector keys and report "nothing pins it" for a peer the parent does pin. Both paths are tested.
- The shell-agnostic rule's false negative (`mj-angular-elements-demo` masking MJExplorer's real gap) is the kind of thing you only find by running against the real workspace, and there's a test named for it.
- `@angular/elements` on MJExplorer follows the `@angular/service-worker` precedent — declared, imported nowhere in MJExplorer's own source, pinned at 21.2.22 like its siblings — and the guide names both, so the *why* is recorded where a future dependency prune would look.
- `dependencies` omitted entirely when empty, so a parent with no Open App member regenerates byte-identically. Good detail.
- Agreed on leaving two of #4364's five assertions red — they encode hardcoding Open App repo knowledge into MJ's CLI, which is the wrong direction.

### Three things worth a reply — none blocking

**1. The parse path is detection-wide and pre-selection — is that intended?**

Two failure points with different blast radii:

| Error | Thrown in | When | `--exclude` helps? |
|---|---|---|---|
| Unparseable JSON | `readJsonFile` in `LoadRepo` | `DetectCandidates`, for **every** sibling dir | **No** |
| Schema violation (not-an-array, null entry, missing `name`) | `readDeclaredEntries` | `BuildRootPackageJson`, after member selection | Yes |

`DetectCandidates` calls `LoadRepo` for every subdirectory before the `Reasons.length > 0` filter and before exclusion (`detect.ts:352-355`), so a present-but-malformed `mj-app.json` in a sibling the user *excludes* still aborts `mj dev workspace` — and `doctor`, which calls `DetectCandidates` too.

To be fair about novelty: `LoadRepo` already reads `package.json` through the same `readJsonFile` with the same throw, so this extends an established pattern rather than inventing one, and only an Open App repo mid-edit triggers it. The schema errors naming repo, file, section and index are good. I'd just want the detection-time parse failure to be a decision rather than a side effect.

**2. A provided client package can land in both `dependencies` and `devDependencies`.**

`classifyDevDep` keeps family-provided packages in the devDependency union at `workspace:*`; `BuildClientDependencies` independently emits every provided client package into `dependencies` at `workspace:*`. A provided client package is by definition in `family.Names`, so if any member also lists it in `devDependencies` — an Open App repo devDepending on its own client package is ordinary — it appears in both blocks of the generated manifest. Same specifier, so pnpm resolves it identically and the practical impact is near zero, but it's a generated file and the interaction is untested. A guard in `BuildClientDependencies` or a test pinning that it can't happen would close it.

**3. Cross-repo duplicate client declarations are silently first-wins.**

`CollectOpenAppClientPackages` keeps the first by repo sort order and drops the rest with no report — while its sibling `CollectFamilyPackages` returns an explicit `Duplicates` list that the command warns on ("the link target is decided by sort order; use `--exclude` to drop one"). Two members declaring the same client package is that same ambiguity. Tested and deliberate, so a consistency question rather than a bug.

### Not needed

Agree with skipping the deterministic integration tier — this is a workspace-file generator with no DB, resolver or runtime surface, and offering to run it on request is the right posture.

rkihm-BC (dismissed):
Re-approving on `d6f04f573e`. Both follow-up commits do what they say, and I verified them against the branch rather than the commit messages.

**Parse-path (`032286484a`).** Detection now carries an unparseable `mj-app.json` on the candidate as `MjAppJsonError` instead of throwing, and `CollectOpenAppClientPackages` — the only consumer that reads the declaration, and which only ever sees selected members — raises it. That's the right seam: `--exclude` now genuinely routes around a broken sibling, an included one still fails fast with the repo named, and the root `package.json` read keeps throwing in `LoadRepo` because "not a loadable repo at all" is a different claim. The three detect tests pin exactly that distinction.

**Manifest (`9ac2e8a601`).** A provided client package is now emitted once, in `dependencies`, and the guard-rail test (`leaves an unrelated family devDependency in the devDependency union`) is the one that matters — it stops the delete from over-reaching. I also checked that the delete can't strand a `Conflicts` entry: `classifyDevDep` maps every family-provided name to `workspace:*` before `recordFirstWins` compares, so two members can never disagree on one. Cross-repo duplicates now come back as `{ Packages, Duplicates }` and warn in the same shape as `CollectFamilyPackages`, and treating one repo's client+shared double-listing as *not* a duplicate is correct — the link target isn't in question there.

The `next` merge was a clean auto-merge (incoming was #4365; nothing under `lib/dev-workspace`). 23/23 green.

One consistency note, not blocking: with the parse error now surfacing from `CollectOpenAppClientPackages`, an *included* member with a broken `mj-app.json` makes `doctor` exit via `this.error` before printing any check, rather than showing the other checks plus one `fail`. It's a clean error naming the repo and file, and generate behaves the same way, so it's defensible — just slightly out of step with doctor's "report every check" posture. Worth a thought if a `fail` check turns out to be cheap.

Branch is one commit behind `next` again (#4404, unrelated) — fine to merge from behind or update, your call.

rkihm-BC (approved):
Re-approving on `f8b813f12c`. One PR-authored commit since my last pass, and it takes the consistency note I left as explicitly non-blocking — which I appreciate, because it was the kind of thing that is easy to shelve.

## `48fd822798` — doctor reports instead of aborting

The implementation is better than the note asked for, and two details are worth calling out because they are the parts that would have been easy to get wrong.

**Partitioned, not caught.** The comment states the reason: `CollectOpenAppClientPackages` throws on the *first* unreadable member, so wrapping it in a try would discard every readable member's declarations along with the broken one. Filtering first keeps the other members' results.

**Declarations from `readable`, provision index from `members`.** This is the subtle one:

```ts
const clients = CollectOpenAppClientPackages(readable, IndexWorkspacePackages(members)).Packages;
```

A member whose `mj-app.json` cannot be parsed still *provides* its packages to the workspace — the unreadable file says nothing about what is on disk. Building the index from `readable` would have silently dropped those packages from the census and produced wrong "not linked" results for everyone else. Passing `members` there is correct, and the comment says why.

**The failure detail earns its place.** It names each unreadable file and its parse error, then reports what *was* checked — `the readable members' packages were still checked: N/M linked` — so the output is still useful rather than just a stop sign. That is the posture doctor should have.

## The asymmetry holds where it should

I traced all three paths rather than assuming the commit message:

| Path | Behavior | Where |
|---|---|---|
| detection | carries the error on the candidate | `detect.ts:66-70` |
| `generate` | **throws**, naming repo and file | `build.ts:507-508` |
| `doctor` | reports a `fail` check, other checks still run | `doctor.ts:430-434` |

`generate` refusing is the right half to keep strict — it writes a manifest, and writing one from a declaration it could not read would bake the gap into the workspace. The JSDoc on `Unreadable` records that reasoning where the next person will find it.

The tests pin the distinction properly: the first asserts `report.Checks` still has all 10 entries and that unrelated checks still pass, which is exactly the regression the fix exists to prevent; the second proves a readable member is still censused when a sibling is broken.

## Scope and state

- Still 13 files, confined to `lib/dev-workspace` plus MJExplorer's `@angular/elements` declaration. The large commit list since my last review is almost entirely `next` merges (AN-BC's realtime work) — this PR's own footprint has not moved.
- **0 behind `next`.**
- `MERGEABLE` / `BLOCKED`, and the block is review only: both of my earlier approvals were auto-dismissed by pushes, which is why it reads `REVIEW_REQUIRED`. This clears it.

**One thing I did not do:** I could not run the CLI suite locally — the worktree had no `node_modules` and I was not going to stand up a full install for a 26-line change. So I read the tests rather than executed them. CI is still working through this SHA (5 green, 9 pending including `Build` and the unit shards), and my approval assumes those land green. If the unit shards fail, this needs another look rather than a merge on my say-so.

Nothing else from either earlier review is outstanding.

PR #4404, opened 2026-09-11
fix(server,provider): UserSearchString is free text, not a SQL clause — unbreaks multi-word, apostrophe and keyword searches (#4392)

### LINKED ISSUE #4392: Explorer search returns 0 rows for any multi-word or apostrophe term: UserSearchString is screened as a SQL clause
## Symptom

Any Explorer grid / view search whose term contains a **space or an apostrophe** returns 0 rows. Single-word terms work.

| Search term | Result |
|---|---|
| `Marcus` | works |
| `Marcus Chen` | 0 rows |
| `O'Leary` | 0 rows |

Reproduced against MJ `next`. Repro: Explorer → any entity grid → type a first and last name into the search box.

## Root cause

`screenClientViewClauses` runs all four client-supplied view inputs through the GraphQL-boundary SQL screen:

https://github.com/MemberJunction/MJ/blob/next/packages/MJServer/src/generic/ResolverBase.ts#L875-L878

```ts
this.assertClientClauseUsesEntityBaseViews(clauses.extraFilter, 'ExtraFilter', provider);
this.assertClientClauseUsesEntityBaseViews(clauses.orderBy, 'OrderBy', provider);
this.assertClientClauseUsesEntityBaseViews(clauses.userSearchString, 'UserSearchString', provider);   // <-- this one
this.assertClientClauseUsesEntityBaseViews(clauses.overrideExcludeFilter, 'OverrideExcludeFilter', provider);
```

Three of those are genuine SQL fragments. The fourth, `UserSearchString`, is not — it is free text a user typed into a search box.

`assertClientClauseUsesEntityBaseViews` (`ResolverBase.ts:820`) wraps its argument as

```sql
SELECT 1 FROM __mj_clause_screen WHERE (<clause>)
```

parses it with `@memberjunction/sql-parser`, and fails closed when the result is not a valid read:

```ts
if (!parser.IsValid || parser.HasWriteStatement || parser.StatementKind !== 'select') {
  throw new Error(`Invalid ${label}: not a safe read-only filter fragment — refusing under uncertainty`);
}
```

That matches the symptom exactly:

- `Marcus Chen` → `WHERE (Marcus Chen)` → not valid SQL → refused
- `O'Leary` → unterminated string literal → refused
- `Marcus` → parses as a bare column reference → passes

So the screen is not rejecting dangerous input; it is rejecting everything that isn't coincidentally valid SQL, which is most real search terms.

## Why dropping the screen on this one input is safe

`UserSearchString` never reaches SQL as a fragment. Its only consumer is `GenericDatabaseProvider.createViewUserSearchSQL` (~line 2065), which:

- doubles single quotes (`userSearchString.replace(/'/g, "''")`) before the value is used anywhere;
- escapes LIKE metacharacters (`%`, `_`, `[`, `]`) and emits `ESCAPE`;
- builds the predicate itself, per field, from `IncludeInUserSearchAPI` metadata. The user's text only ever lands as a literal inside a predicate the provider constructed.

It also remains covered by the provider-level denylist: `ValidateUserProvidedSQLClause` (MJCore `databaseProviderBase.ts:1077`) is still applied to `UserSearchString` at `GenericDatabaseProvider.ts:1731` and `:2562`, which keeps the stacked-statement / DML / comment / `UNION` / `WAITFOR` screen on it.

## Suggested fix

Remove line 877. Keep 875, 876 and 878 — those carry real clause fragments and the base-view screen belongs on them.

A regression test alongside the existing ones in `packages/MJServer/src/__tests__/ResolverBase.filterEscaping.test.ts` asserting that a multi-word and an apostrophe-bearing `UserSearchString` survive `screenClientViewClauses` would pin it.

## Impact

Every entity grid and saved-view search in Explorer, for any multi-word search term — which includes essentially every person-name search.

rkihm-BC:
Fixes #4392.

## The bug

`UserSearchString` is the free text a person types into a search box. Two screens built for genuine SQL **fragments** were applied to it anyway, and each rejected ordinary searches.

**1. The GraphQL boundary (the reported bug).** `ResolverBase.screenClientViewClauses` ran it through the base-view AST screen, which wraps its argument as `SELECT 1 FROM x WHERE (<clause>)` and fails closed when that does not parse:

| Search term | Before |
|---|---|
| `Marcus` | works — parses as a bare column reference |
| `Marcus Chen` | **0 rows** — parses as nothing |
| `O'Leary` | **0 rows** — unterminated string literal |

Only terms that were *coincidentally valid SQL* survived, which killed essentially every person-name search.

**2. The provider denylist (found while testing the fix).** `GenericDatabaseProvider` also ran it through `ValidateUserProvidedSQLClause`. Word-boundary-matched against free text, that refused:

| Search term | Before | Why |
|---|---|---|
| `Union Pacific` | **refused** | the word `union` |
| `Update Request` | **refused** | the word `update` |
| `drop shipment` | **refused** | the word `drop` |
| `well--known`, `a; b` | **refused** | `--`, `;` |
| `Raindrop`, `Executive Summary` | worked | `\b` boundaries — only whole words matched |

The resulting error reached the grid with a **null message**, so the box simply appeared broken.

**3. Full-text multi-word (surfaced by fixing #1).** The full-text branch tested AND/OR/NOT as *substrings*, so `OR` matched "c**or**porate" and `AND` matched "st**and**ard". Two-word searches came out `%`-joined — not valid full-text syntax:

| Term | Before | After |
|---|---|---|
| `Corporate Office` | `'Corporate%Office'` | `'Corporate AND Office'` |
| `Standard Rate` | `'Standard%Rate'` | `'Standard AND Rate'` |
| `North America` | `'North%America'` | `'North AND America'` |

Multi-word terms only began reaching that branch once the boundary screen stopped refusing them, so without this the fix would not have applied to full-text entities.

## Why removing the screens is correct, not a weakening

`createViewUserSearchSQL` builds every predicate **itself** from `IncludeInUserSearchAPI` metadata and lands the user's text *inside* a string literal — quotes doubled, LIKE metacharacters escaped under an explicit `ESCAPE`:

```sql
([Name] LIKE N'%O''Leary%' ESCAPE '\')
```

Quote-doubling — not keyword matching — is the right protection for a value in a literal. It cannot be bypassed by a keyword the denylist forgot, and it never rejects a legitimate search.

**Every genuine clause fragment is still screened exactly as before.** `ExtraFilter`, `OrderBy` and `OverrideExcludeFilter` keep the boundary AST screen; those three plus the rendered WHERE clause keep the provider denylist.

**One exception, deliberately retained.** A field carrying `UserSearchParamFormatAPI` splices the term into an admin-authored format that may place `{0}` *outside* quotes (` = {0}` on a numeric field is a supported, tested configuration). There the term is not confined, so the denylist is re-applied — but only when such a field will actually **participate**, so a field excluded by field-level security does not trigger it. Without that guard this PR would have opened a real regression; it was caught in review and is covered by tests.

Adversarial probe against a live server, with the denylist gone — all 8 payloads inert, table intact, no `WAITFOR` executed:

```
✅ inert — 0 rows   "x' OR '1'='1"                            classic tautology
✅ inert — 0 rows   "x' OR 1=1 --"                            tautology + comment
✅ inert — 0 rows   "%' OR '1'='1"                            LIKE-wildcard tautology
✅ inert — 0 rows   "x'; DROP TABLE __mj.AIAgentCategory;--"  stacked DDL
✅ inert — 0 rows   "x' UNION SELECT 1,2,3 --"                UNION injection
✅ inert — 0 rows   "x' WAITFOR DELAY '00:00:05' --"          time-based blind
✅ inert — 0 rows   "x\' OR \'1\'=\'1"                        backslash-escape bypass
✅ table intact after all payloads: 9 rows (baseline 9)
```

## Interaction with Field-Level Security (#3367)

FLS landed on `next` mid-review and rewrote the same function. The merge is deliberate, not mechanical: `userSearchFieldsUseCustomFormat` now consumes FLS's denied-field set, because a field FLS excludes never reaches the SQL and so cannot carry the term into it — screening on its behalf would refuse ordinary searches for exactly the users with the least access. Three tests pin that interaction.

Notably, the FLS author reached this PR's conclusion independently: *"UserSearchString is not rejected (the term is not a caller-authored predicate)"* — excluding denied fields rather than refusing the term.

## Verification

Merged tree, against a database **built from scratch** (82 migrations, 384 entities, 320 integration records, 76 deterministic suite tests — matching the branch exactly):

| Suite | Result |
|---|---|
| GenericDatabaseProvider | 1081 passed |
| MJServer | 1328 passed |
| MJCore | 2331 passed |
| integration-test-suite | 201 passed |
| **Deterministic integration tier** | **69/76 bundles, 492 checks, 0 failed** (7 mutation/live-model gated) |
| `check:standards`, `check:changeset` | pass |

Regression coverage at three layers: the GraphQL boundary (`ResolverBase.filterEscaping.test.ts`), the generated SQL including the custom-format and FLS interactions (`createViewUserSearchSQL.test.ts`), and end-to-end over the real client transport (`runview-matrix.RVM19` — the seam this bug lived in, which unit tests mock away). The new tests were confirmed to fail with the fix reverted.

## Blast radius

Two shipped packages, but `GenericDatabaseProvider` is the base for **both SQL Server and PostgreSQL**. **322 of 383 entities** have search-enabled fields (972 fields). Every GraphQL view entry point is affected, and consumers include every Explorer grid, `find-record`, several dashboards, `SearchEngine`/`EntitySearchProvider`, the `scoped-search` Core Action, and external API callers.

Deltas: previously-refused terms now work; single-word terms unchanged; full-text entities get different (valid) query text; custom-format entities unchanged vs. pre-PR; `ExtraFilter`/`OrderBy`/`OverrideExcludeFilter` untouched.

## Known gaps

- **The Angular input binding was not exercised in a browser.** Explorer's dev config uses MSAL; a headless magic-link session is confined to its scoped app and cannot reach a generic entity grid. Everything below the binding — `GraphQLDataProvider` → MJAPI → resolver → provider → SQL, the identical path the browser uses — is verified end-to-end. The binding is an untransformed pass-through and is untouched here.
- **The full-text fix is unit-verified only.** No entity ships with `FullTextSearchEnabled`, so there is no index to exercise `CONTAINS` against; the generated query *string* is asserted, not the engine's acceptance of it.
- **The custom-format guard is unit-verified only** — no field ships with `UserSearchParamFormatAPI` set.
- **PostgreSQL was not exercised.** The changes are platform-neutral, but the PG provider path was not run.
- The mutation tier (`RUN_MUTATION_TESTS=1`) was not run.

## Follow-up worth its own issue

An **unquoted** `UserSearchParamFormatAPI` is injectable by a bare tautology (`1) OR 1=1`) that carries no denylisted keyword — so the denylist never covered it, before or after this PR. Quoting `{0}` in the format is the actual fix. This PR restores the status quo for that path; it does not close it.


rkihm-BC:
## Adversarial review round — 4 findings, all real, all fixed (`ed46fbc0`)

**1. Security regression this PR introduced.** My justification for dropping the denylist — "the term always lands inside a string literal" — is **false for `UserSearchParamFormatAPI`**, an admin-authored format that may splice `{0}` in *unquoted* (` = {0}` on a numeric field is a documented, tested case). Reproduced:

```
term "2026)) UNION SELECT 1,2,3 --"  ->  (([Year]  = 2026)) UNION SELECT 1,2,3 --))
term "2026; DROP TABLE Users--"      ->  (([Year]  = 2026; DROP TABLE Users--))
```

`createViewUserSearchSQL` now re-applies `ValidateUserProvidedSQLClause` when a participating field carries such a format, and only then — ordinary entities keep accepting the ordinary searches this PR is about.

This restores the pre-#4392 screen for that path. It does **not** close the unquoted-format hole itself: the denylist never covered it (`1) OR 1=1` carries no forbidden keyword), so that tautology was reachable before this PR too. Quoting `{0}` in the format is the real fix, and it deserves its own issue — flagging for a decision.

**2. `#4392` was not actually fixed for full-text entities.** The FTS branch tested AND/OR/NOT as *substrings*, so `OR` matched "c**or**porate" and `AND` matched "st**and**ard". Ordinary two-word searches came out `%`-joined, which is not full-text syntax:

| Term | Before | After |
|---|---|---|
| `Corporate Office` | `'Corporate%Office'` | `'Corporate AND Office'` |
| `Standard Rate` | `'Standard%Rate'` | `'Standard AND Rate'` |
| `North America` | `'North%America'` | `'North AND America'` |

Multi-word terms only began reaching this branch once the boundary screen stopped refusing them, so this PR would have left FTS entities broken. Now word-boundary based. **Unit-verified only** — no entity in a default install has FTS enabled, so there is no index here to exercise end-to-end.

**3. A comment I added was factually wrong.** It claimed the provider denylist "still covers `UserSearchString` on both the view and count paths" — which the same commit deleted. Written before the scope widened, never updated. A false claim of a second defense layer, sitting exactly where an auditor would read it. Corrected.

**4. RVM19 was both brittle and under-powered.** It could fail spuriously when the discovered entity's search field is `nvarchar(MAX)` (empty predicate → every term returns the baseline → misleading "term was dropped"); a control probe now makes that an honest skip. Separately, its decisive multi-word leg was keyed to whatever entity discovery returned first — whose value may be single-word — so the one assertion that actually proves #4392 was *skipping*. It now searches across LIKE-path entities for a multi-word seed and runs:

```
→ multi-word term "Image Generator" on MJ: AI Model Types.Name: 1 row(s), seed row present
→ MJ: AI Actions: 16 free-text terms accepted and applied; injection payload inert (baseline 3)
```

### Re-verified after the fixes

| Suite | Result |
|---|---|
| GenericDatabaseProvider | 1026 passed |
| MJServer | 1288 passed |
| MJCore | 2331 passed |
| integration-test-suite | 201 passed |
| Deterministic integration tier | **67/73 bundles, 466 checks, 0 failed** |

Exposure note for reviewers: both newly-found paths are **zero-instance in shipped MJ metadata** (0 entities with `FullTextSearchEnabled`, 0 fields with `UserSearchParamFormatAPI`), so neither was live out of the box — but both are per-deployment settings a customer can enable.

PR #4408, opened 2026-09-12
The login screen becomes a composable Explorer surface: tokens, layout, slots

cmbrock-BC:
## Summary

The pre-auth sign-in page was the one Explorer surface a deployment could not adapt without forking the component or writing CSS against its internal class names. `mj-explorer-app` bound none of the picker's existing inputs, the banner's gradient geometry and 297×45 logo box were burned into the stylesheet, the banner text was pinned to a primitive (`--mj-color-neutral-0`), the copy was welded into the template, and there was no way at all to **rearrange** the screen or **add** anything to it.

This makes the login screen composable the way the rest of MJ is — **tokens, layout, content, slots** — with every default reproducing today's rendering exactly.

## 1 · Tokens

A `--mj-login-*` group in `_tokens.scss`, documented in THEMING.md:

| Token | Covers |
|---|---|
| `--mj-login-banner-bg` | The story panel's whole background. The stock gradient is its *default value*, so a theme can retarget the `--mj-login-grad-*` stops (cheap) or replace the value outright — flat colour, image, different geometry |
| `--mj-login-banner-text` / `-text-secondary` | Banner copy (the secondary derives at 82%, replacing an `opacity` trick) |
| `--mj-login-banner-logo` + `-logo-width` / `-height` | Brand image and its box (`contain`, so any logo shape works) |
| `--mj-login-banner-flex` / `--mj-login-panel-flex` | The editorial split |
| `--mj-login-panel-bg` · `--mj-login-picker-max-width` · `--mj-login-centered-card-width` | Sign-in column |
| `--mj-login-card-*` / `--mj-login-banner-card-*` | The two card surfaces |

Only the gradient stops carry dark-mode entries: `var()` resolves at usage time, so the existing dark overrides flow through `--mj-login-banner-bg` on their own, and every other token either references a semantic token that swaps by itself or is theme-constant by design (the banner stays dark in both themes).

## 2 · Layout — rearranging

`LoginLayout`: `'split'` (stock editorial split) · `'split-reverse'` (mirrored; **DOM order is unchanged**, so the story stays first for a screen reader) · `'centered'` (one column — story content on a full-bleed banner background, sign-in options in a card on top).

A layout **moves** regions and never drops one: every slot and every input still renders in each.

## 3 · Content — adding, as data

`LoginCards: MJLoginCard[]` — tiles (icon, title, text, optional link) each naming its region (`'banner'` / `'panel'`). Links are sanitized by Angular and get `rel="noopener noreferrer"` when `newTab` is set; the whole tile is the click target via a stretched anchor, and the accessible name is the anchor's own text — the title, plus "(opens in a new tab)" where that applies, since that hint belongs in the name.

Cards exist because **the login screen cannot be laid out from MJ metadata** — it renders before any authenticated call is possible, and the only pre-auth server surface is the allow-listed public provider catalog. A deployment that keeps per-tenant login config somewhere resolves it itself (exactly as Explorer already resolves that catalog before bootstrap) and binds it here. Widening a public endpoint into a general screen-config channel would be a separate change with its own security review.

Plus the copy/attribution inputs: `LoginHeading` (`null` drops it — with a single provider the page currently says "Log in" twice), `LoginShowPoweredBy` (white-labelling is an entitlement, so it must be configuration rather than a CSS hide), `LoginBannerAlignment`, `LoginBannerTitle`, `LoginBannerSubtitle`, `LoginBannerLogoLabel` (the artwork comes from a token the component cannot read, so a swapped logo needs its accessible name from the host).

## 4 · Slots — adding, as markup

The `mjLoginSlot` directive, the same shape as `mjChatSlot` in `ng-conversations` — one directive per slot name, host template rendered through `ngTemplateOutlet`, a real default behind every slot:

```html
<mj-explorer-app [LoginLayout]="'centered'" [LoginCards]="cards">
  <ng-template mjLoginSlot="panelFooter">
    <a href="/terms">Terms</a> · <a href="/privacy">Privacy</a>
  </ng-template>
</mj-explorer-app>
```

| Slot | Effect |
|---|---|
| `bannerContent` | **Replaces** the brand block (the banner copy inputs no longer apply) |
| `bannerFooter` | **Adds** below it — takes over from the `'banner'` cards, this region's stock content |
| `panelHeader` | **Adds** above the sign-in options |
| `panelFooter` | **Adds** below them — takes over from the `'panel'` cards |

`ng-content` would not do here: the same regions render in different places per layout, and content captured as a `TemplateRef` can be rendered wherever the layout puts it. `<mj-login-picker>` is deliberately **not** replaceable — a host surrounds sign-in rather than owning it.

## 5 · A latent clipping bug, fixed

The app shell sets `html, body { overflow: hidden }`, so any login content taller than the viewport was simply unreachable. That already bit a tenant with **many identity providers on a short screen**, and configured cards make it easy to hit. Both columns now scroll their own content and centre with **auto margins** rather than `justify-content: center` — which, on a scroller, strands the overflow above the scroll origin where no scrolling reaches it. The stacked (≤900px) layout keeps its bottom-pinned logo via `margin-block: auto 0`.

## Testing

**Unit — in the repo and re-runnable**: explorer-app 23/23, auth-services 75/75, shared 27/27. These are contract tests reading the shipped CSS/HTML (no primitives, no colour literals, every slot consulted with a default behind it, each layout's class implemented, the overflow contract, and the two new stacked-layout guards below). Every guard is mutation-checked: re-centring a scroller, dropping a slot, removing the auth `rel`, moving the `split-reverse` block past the breakpoints, and reverting either landscape fix all fail the intended test.

**Browser — driven ad hoc, NOT committed.** An earlier version of this section claimed 192 assertions across four Playwright suites. Those checks were really run against a dev build of this branch, but they live in throwaway scripts rather than the repo, so nothing in this branch reproduces them and a reviewer cannot re-run them. Treating them as evidence was wrong and the claim is withdrawn. What survives is what a reviewer can execute: the unit counts above, plus the measurements in the landscape section, taken by rendering the shipped stylesheet against the component's own DOM at each viewport.

`check:ui`, `check:standards` (ui-layers) and `check:changeset` all pass.

## Review follow-ups (Robert, 2026-09-12)

- **The one real defect: the card focus ring drew square on a rounded tile.** `border-radius: inherit` on the stretched anchor's `::after` resolves to `0` — the property is not inherited, and on a pseudo-element `inherit` takes from the *originating* element (the anchor, which has no radius) rather than the positioned ancestor `inset: 0` sizes it to. Fixed by naming the token, with a unit guard and a computed-radius browser assertion; the suite had only checked that the ring *appeared*, which is exactly why it got through.
- **Forced-colors / Windows High Contrast** — raised as a repo-wide gap, taken here because this is the one surface a user meets before any other. The card fill, border token and `color-mix` hover tint are all substituted away, so a link card read identically to a static one; the affordances are re-expressed in system colours. This is the first `@media (forced-colors: active)` block in `packages/Angular/`, and it is verified under Playwright's forced-colors emulation rather than asserted.
- **Landscape phones** — fixed properly in the second round, after the first attempt only worked for one layout. `.main-banner` carries `flex: var(--mj-login-banner-flex, 1.05)`, and that shorthand sets flex-basis to `0%`, so the banner was sized by the grow ratio and every `height` in the stacked breakpoints was dead. It stayed hidden while `.login-btn` had no overflow, because the column's automatic minimum size was its content height; giving it `overflow-y: auto` dropped that minimum to zero and handed the layout to the ratio. The stacked banner is now `flex: 0 0 auto`, so the authored heights decide. Measured against the shipped stylesheet, before → after:

  | viewport | banner | sign-in rows fully on screen |
  |---|---|---|
  | 844×390 | 183px → **92px** | 1/3 → **3/3** |
  | 740×360 | 167px → **92px** | 1/3 → 2/3 (third reachable — the column scrolls) |
  | 820×560 | 277px → **196px** | 3/3 → 3/3 |
  | 390×844 | 423px → **422px** | 3/3 → 3/3 |

  Portrait is deliberately held where it already renders: the authored 30dvh would have cut it to 253px, and shipping a 170px change to the first screen every user meets as a side effect of a scrolling fix is not something to do quietly. 50dvh reproduces what people see today, and the height is now the thing that decides it.

- **The short-screen copy-hide was width-unscoped** — at 1280×470 in the split layout it hid the welcome copy from a 470px column with room for it, contradicting this PR's own "no visual change for an app that binds nothing". Now scoped to short *and* narrow. Measured at 1280×470: `display: none` → `block`.

- **The slot documentation described selectors that do not exist** — `login-screen.types.ts` documented `[loginBannerContent]`-style attribute selectors in line with MJ's page chrome, but the component projects through `MJLoginSlotDirective` and has no matching `ng-content`, so a host following it rendered nothing. Corrected to `<ng-template mjLoginSlot="…">`.

- **`@since` tags** — eleven said 6.1.0, which has shipped. Moved to 6.2.0.

- **`--mj-login-panel-text` reaches projected slot content only** — the picker's own heading, lede and attribution set `--mj-text-primary/-secondary/-muted` directly. THEMING.md was already accurate; the CSS comment implied the token covered the whole region and now states the limit.
- **Text selection inside a stretched link** — inherent to the pattern and now documented where the CSS lives, with the escape hatch named (put the link in the text rather than on the card). Verified that a card *without* an `href` keeps its text selectable, so only linked cards make the trade.
- **The accessible-name sentence** was overstated and is corrected above and in the code comment: the name is the anchor's own text, which includes the new-tab hint.

## Review follow-ups (Matt, 2026-09-15)

- **The theme toggle scrolled away in the `centered` layout.** `.theme-toggle` is `position: absolute` against `.login-wrapper`, which is correct while the *columns* scroll — but `centered` makes that same wrapper the scroller, so the toggle was positioned against scrolled content and rode off the top. Nine providers or a few cards is enough, and it vanished exactly when someone scrolled to reach the sign-in button. Now `position: fixed` scoped to that layout; split is untouched, where absolute is right.

- **The picker now reads the login panel tokens, with the semantic ones as fallbacks.** Matt argued this rather than narrowing the comment, and he is right: the pitch of this surface is "override tokens, never write CSS against this DOM", and the heading keeping `--mj-text-primary` while the panel took `--mj-login-panel-bg` meant a dark panel in light mode left a near-black heading — the first thing a white-label deployment does. Adds `--mj-login-panel-text-secondary` and `--mj-login-panel-text-muted` alongside the existing text token, all defaulting to the semantics, so an app that themes nothing is byte-identical.

- **An empty `LoginBannerLogoLabel` is now decorative rather than an unnamed image.** `role="img"` with no accessible name announces an image and can say nothing about it. Blank label → `aria-hidden`, no role; bound label → labelled image. Guarded.

**Downstream** — Betty (betty-mj-vs) carries an `INTERIM CSS PIERCE` block reproducing the heading/attribution/centring effects against the picker's internals; it deletes that block the release this lands, binding the inputs and setting the tokens instead.


rkihm-BC (commented):
Reviewed the CSS, template, directive and tokens against the rendered semantics rather than the description — particularly the responsive, cross-browser and a11y surface, since that is where a composable login screen either holds up or quietly does not.

**One real defect, cosmetic but on a focus indicator.** Everything else is either good or a suggestion.

## What's good

**The overflow fix is the best thing here, and it is a genuine bug rather than a tidy-up.** `overflow-y: auto` with `justify-content: center` is the classic trap: once content exceeds the column, the overflow goes *above* the scroll origin and no amount of scrolling reaches it. Combined with the shell's `html, body { overflow: hidden }`, content was simply unreachable — and "many identity providers on a short screen" is exactly the case that hits it. Auto margins instead of centred justification is the correct fix, `min-height: 0` is on the flex children that need it, and the stacked layout keeps its bottom-pinned logo with `margin-block: auto 0`. The comment explaining *why* centring strands the overflow is the kind of thing that stops someone re-centring it in two years.

**`100dvh`, with the reason written down.** This is the single most common mobile layout defect — `vh` includes the collapsing URL bar, so a `100vh` column is taller than the visible viewport on iOS Safari and Android Chrome, and the bottom is cut off. Every login rule uses `dvh` and the comment names the mechanism. (`.app-main`'s `100vh` is the authenticated shell and untouched by this PR; and bare `dvh` without a `vh` fallback matches what the rest of `packages/Angular/` already does, so it is convention here, not an omission.)

**`split-reverse` flips the visual order without touching the DOM.** `flex-direction: row-reverse` keeps the story first in source, so a screen reader and the stacked layout both get the intended order. And the comment about *why the rule must sit before the breakpoints* — same specificity as the stacked `flex-direction: column`, so source order decides — is correct and exactly the sort of thing that breaks silently when someone reorders a stylesheet.

**Responsive coverage is real, not nominal.** 900px to stack (justified: the columns cramp before 768), 480px for phones, the logo clamped with `width: min(var(--mj-login-banner-logo-width, 297px), 200px)` so a host-set logo width cannot overflow a narrow screen, and the banner down to 30dvh so the sign-in options stay above the fold.

**Token design.** Zero colour literals in the new CSS. The observation that `var()` resolves at *usage* time — so only the gradient stops need dark-mode entries and everything else flows through on its own — is the right insight and is what keeps the token group small. `--mj-login-banner-text-secondary` as `color-mix(… 82%, transparent)` instead of an `opacity` on the element is better than what it replaces: opacity would have faded any child, and this composites only the colour.

**Focus handling is correct, including the part that is easy to get wrong.** `outline: none` paired with `box-shadow: var(--mj-focus-ring)` — and `--mj-focus-ring` is a box-shadow value, so using it in `box-shadow` rather than `outline` is right. (Used as an `outline` it renders nothing and silently removes the indicator; that misuse has bitten this repo before.)

**The stretched-link pattern is implemented properly.** `.login-card` carries `position: relative` so `inset: 0` sizes to the card, and it has no `overflow: hidden`, so the focus ring is not clipped. The sr-only "(opens in a new tab)" sits *inside* the anchor, which is right — that hint belongs in the accessible name. `rel="noopener noreferrer"` is bound to `target="_blank"`, and a `javascript:` href from configuration is proven both sanitized and inert on click.

**The testing is the most thorough UI verification I have seen in this repo** — 3 layouts × 4 viewports for overflow and control reachability, WCAG contrast computed from *rendered* styles in both themes and composited through the translucency, keyboard reachability and tab order, nine providers on a 560px-tall viewport, and mutation-checked unit guards. The contrast-through-translucency detail matters: a naive check against the token value would have missed what the 82% mix actually composites to.

## What's concerning

### The card focus ring draws square on a rounded card

`explorer-app.component.css`:

```css
.login-wrapper .login-card__anchor::after {
  position: absolute;
  inset: 0;
  border-radius: inherit;   /* <- resolves to 0 */
}
```

`border-radius` is **not an inherited property**, and on a pseudo-element `inherit` takes from its originating element — `.login-card__anchor`, not `.login-card`. The DOM is:

```
div.login-card            (position: relative; border-radius: var(--mj-login-card-radius))
  span.login-card__body
    span.login-card__title
      a.login-card__anchor      (color, text-decoration — no radius)
        ::after                 (inset: 0 → sized to .login-card; radius → 0)
```

So the overlay is correctly *sized* to the card (nearest positioned ancestor) but takes radius `0`. On `:focus-visible` the ring is drawn from that box, giving square corners around a rounded tile — visible whenever a keyboard user tabs onto a card link, in every layout and theme.

Not a WCAG failure — the indicator is present and contrasts — so cosmetic. One line:

```css
border-radius: var(--mj-login-card-radius);
```

Worth adding a browser assertion for it too; the existing suite checks that the ring *appears*, which is why this got through.

### Description overstates the accessible name

> the whole tile is the click target via a stretched anchor, while the accessible name stays the title

With `newTab`, the name is "Title (opens in a new tab)" — the sr-only span is inside the anchor. **The code is right**; the sentence is what needs the edit.

## What's missing — suggestions, not requirements

**Forced-colors / Windows High Contrast.** Card affordances rest on `background` and `border-color`, and `--mj-login-card-*` are ignored under `forced-colors: active`; the `color-mix` hover tint in particular will not render. The cards stay legible and operable (the border survives as a system colour), so nothing breaks — but the hover and link affordances flatten. No `@media (forced-colors: active)` block exists anywhere in `packages/Angular/`, so this is a repo-wide gap rather than yours; raising it because this is a pre-auth screen that every user meets and it is the one surface where a forced-colors user has no fallback path.

**Landscape phones.** The breakpoints are width-only. A phone in landscape is ~844×390, so it takes the ≤900 stacked branch with a 35dvh banner — roughly 136px of banner above the sign-in options on a 390px-tall viewport. The tested viewports (1440/1024/820/375) are all portrait or desktop. Probably fine given both columns now scroll, but a `@media (max-height: 480px)` that trims the banner would make it comfortable rather than merely reachable. Worth confirming it was considered.

**Text selection inside a card.** Inherent to stretched links — the `::after` overlay swallows drag-select over the title and body. Standard trade for a full-tile click target, worth knowing since card text is host-authored and may be something a user wants to copy.

## Verdict

Approving once CI settles (13 green, 8 still running). The `border-radius` line is the only thing I would genuinely like changed, and it is one line — not worth blocking on, but worth doing before merge since it is a focus indicator on the first screen every user sees.

rkihm-BC:
Went back over my own review to make sure I was not asking for anything unnecessary, and to answer the design-token question properly rather than from a grep.

## Design tokens: fully compliant, and verified against the gate rather than by eye

`.claude/rules/design-tokens.md` makes two demands — no colour literals in component CSS, and **no primitive tokens** (`--mj-color-*`) in component CSS. Both hold:

- **Zero `--mj-color-*` references** in `explorer-app.component.css`. The only primitive in the PR is `--mj-login-banner-text: var(--mj-color-neutral-0)` **in `_tokens.scss`**, which is precisely where primitives belong — it is the default *value* of a semantic token, not a pin in component CSS. That indirection is the thing this PR adds: the banner text used to reference the primitive directly.
- **Zero colour literals**, and the gate genuinely checked this file. I initially mis-read `1 checked, 1 allowlisted` as the component CSS being skipped; it is not — `_tokens.scss` is the allowlisted one (correctly, it defines the primitives), so the component CSS was the file checked, with 0 violations.
- `check:focus-ring` passes, which independently confirms `--mj-focus-ring` is used in `box-shadow` and not as an `outline` — the misuse that renders nothing.
- The new `--mj-login-*` group follows the semantic pattern: every entry either references a semantic token that swaps with the theme, or is theme-constant by design with the reason stated.

One consistency note, explicitly **not** a rule violation since the rule is colour-only: `.login-card__icon` uses a literal `1.75rem` box where `--mj-space-*` tokens are used elsewhere in the same file. Fine as-is; mentioning it only so the choice reads as deliberate.

## The one defect, now verified exhaustively rather than inferred

I checked every path by which `.login-card__anchor` could acquire a radius:

- Five `border-radius` declarations exist in the file; the only one touching the anchor is the `::after`'s `inherit`
- No rule sets a radius on `.login-card__anchor` itself — it carries only `color` and `text-decoration`
- No global bare-anchor radius exists in `explorer-app` or `shared` — and this matters because the component is `ViewEncapsulation.None`, so a global *could* have reached it

So the anchor computes `border-radius: 0`, `inherit` on the pseudo-element takes from its originating element, and the focus ring draws square on a rounded tile. Confirmed, and still a one-line fix.

## Ruled out, so you do not have to think about them

- **RTL.** `flex-direction: row-reverse` would behave differently under `dir="rtl"`, but nothing in `packages/Angular/` supports RTL, so `split-reverse` is safe. The new CSS also leans on logical properties (`margin-block`) more than physical ones, so it is better placed than most of the codebase if that ever changes.
- **Touch targets.** The card is a full tile and the theme toggle is 40px — both clear WCAG 2.5.8's 24px minimum.
- **`color-mix` support.** Fine on every current engine, and it is the house idiom (~680 uses under `Generic/`).

Nothing else in my review needs changing, and nothing in it is a blocker. The `border-radius` line is the only thing I would actually like changed before merge.

rkihm-BC (dismissed):
## Approving — completing the verdict from my review above.

That review ended "approving once CI settles (13 green, 8 still running)." CI has settled at **23 pass / 0 fail**, and no commits have landed since, so the assessment stands unchanged: the overflow fix is a genuine bug fix rather than a tidy-up, `100dvh` is used with the reason written down, the token work is clean, and the focus handling correctly uses `--mj-focus-ring` as a `box-shadow` rather than an `outline`.

Two things to carry into the merge.

**The one line I would still like changed.** `explorer-app.component.css:278` is unchanged:

```css
.login-wrapper .login-card__anchor::after {
  inset: 0;
  border-radius: inherit;   /* resolves to 0 */
}
```

`border-radius` is not an inherited property, and on a pseudo-element `inherit` takes from its originating element — `.login-card__anchor`, which has no radius — so the overlay is sized to the card but drawn square. A keyboard user tabbing onto a card gets a square focus ring around a rounded tile, in every layout and theme. It is cosmetic rather than a WCAG failure, which is why I am not blocking on it, but it is a focus indicator on the first screen every user meets and the fix is one line:

```css
border-radius: var(--mj-login-card-radius);
```

The existing suite asserts the ring *appears*, which is exactly why this got through — an assertion on its computed radius would close that gap.

**The branch is `BEHIND` `next`** and needs a merge before it can go in.

The other items from my earlier review — forced-colors, landscape phones, text selection inside a stretched link, and the description sentence about the accessible name — remain suggestions rather than requirements. None of them should hold this up.

rkihm-BC (changes requested):
## Requesting one change — the landscape follow-up does not reach the default layout, and the PR now regresses it

Re-reviewed at `364c8d589b` by rendering the real templates and compiled CSS for both `next` and this branch in headless Chromium. The focus-ring and forced-colors items from my last review are done and guarded. The one thing left is the landscape-phone follow-up, which works for `centered` and is dead CSS for the default `split` layout.

### The regression, measured (split layout, 3 providers)

| Viewport | `next` banner | PR banner | Provider rows on screen |
|---|---|---|---|
| 844×390 | 48 px | **183 px** | 3/3 → **1/3** |
| 740×360 | 48 px | **167 px** | 3/3 → **1/3** |
| 820×560 | 181 px | **277 px** | 3/3 → **2/3** |
| 390×844 | 423 px | 423 px | 3/3 → 3/3 |

Two things combine. `.main-banner` keeps `flex: var(--mj-login-banner-flex, 1.05)`, which expands to `flex-basis: 0%`, so the `35dvh` and `30dvh` rules in the stacked breakpoints and the new `height: auto` in the `(max-height: 480px)` block never set the main-axis size. On `next` that did not matter because `.login-btn` had no `overflow`, so its automatic minimum size was its content height and the banner absorbed the remainder. This PR adds `overflow-y: auto` to `.login-btn` (`explorer-app.component.css:165`), which drops that minimum to zero, and the 1.05:1 grow ratio now governs at every stacked size. Setting `overflow-y: visible` back restores `next`'s numbers exactly; adding `flex: 0 0 auto` on `.main-banner` inside the ≤900px block makes the height rules live (844×390 → 77 px banner, 3/3 rows). Note the second fix also changes portrait from 423 px to 253 px, because `next`'s 35dvh was never actually delivered either, so decide which portrait behavior you want.

The changeset and the CSS comment at `:531-535` say landscape phones are handled; that is true only for `centered`. The overflow contract test asserts only that the substring `overflow-y: auto` exists, so it cannot catch this.

### Worth fixing in the same pass, none blocking on their own

- The `(max-height: 480px)` copy-hide is width-unscoped: at 1280×470 and 1400×470 in split layout `.banner-welcome` is `display: none` on the PR and `block` on `next`, with a 470 px column that has room. Scoping it to the stacked branch fixes it without touching the phone cases. This contradicts the changeset's "no visual change for an app that binds nothing".
- `login-screen.types.ts:11-13` documents the slots as `[loginBannerContent]`-style attribute selectors. No such selectors exist anywhere in `packages/`; the shipped mechanism is `<ng-template mjLoginSlot="…">` and the component has no `ng-content`, so a host following the header gets nothing rendered.
- `@since 6.1.0` in 11 places. 6.1.0 shipped without this API; `next` carries two `minor` changesets, so this lands in 6.2.0.
- `--mj-login-panel-text` reaches only projected slot content. The picker's heading, lede, and attribution stay on `--mj-text-primary/-secondary/-muted`, so a dark `--mj-login-panel-bg` in light mode leaves the heading near-black. THEMING.md is accurate about this; the CSS comment at `:167-169` over-promises.

### Verified and fine

- Focus ring: `::after` computed radius equals the card radius (8 px); the vitest guard fails when reverted to `inherit`.
- Unit: 21 passed in `ng-explorer-app`, 27 in `ng-shared-generic`, 75 in `ng-auth-services`. `check:ui` clean. Changeset present at `patch` for exactly the three touched packages.
- axe on `next` and PR at 1440×900 and 844×390: identical pre-existing violation sets, nothing attributable to this PR.
- Every `--mj-login-*` token used resolves; no `any`; no color literals.
- The PR body's 89/42/20 Playwright suites are not in the repo; only `explorer-app.test.ts` covers this surface. Not a blocker, but the landscape matrix they describe evidently did not include the default layout.

MattC-BC:
Read through this with Robert's review alongside. The token work, the layout model and the overflow fix are all good, and the forced-colors block is a nice first for the repo. Holding my approval behind Robert's block on the flex-basis regression, since it hits the stock split layout with nothing bound. Once that lands I'll re-review the fix commit and approve.

Two things to fold into the same pass:

**Theme toggle scrolls away in the `centered` layout.** `.theme-toggle` is `position: absolute` inside `.login-wrapper` (`explorer-app.component.css:37`), and `.login-wrapper--centered` makes that same wrapper the scroller (`overflow-y: auto` at `:471`). So with nine providers on a short screen, or a few configured cards, scrolling down to the sign-in button carries the light/dark toggle off the top. Split isn't affected because the columns scroll instead of the wrapper. `position: fixed` on the toggle under `.login-wrapper--centered`, or moving the button outside the scroller, fixes it.

**The `(max-height: 480px)` copy-hide needs the width scope Robert flagged.** Same conclusion from my side: a 1280×470 window in stock split loses the headline that renders on `next`. Scoping it to `(max-width: 900px) and (max-height: 480px)` like the sibling block keeps the phone case and restores desktop.

Non-blocking, take or leave:

- `LoginBannerLogoLabel` bound to `''` leaves a `role="img"` with no accessible name (`explorer-app.component.html:66`). Treating empty as decorative (`aria-hidden="true"`, drop `role`) would cover the "logo is decorative, title names the brand" case cleanly.
- On `--mj-login-panel-text`, I'd lean toward having the picker read the login panel tokens with the semantic ones as fallbacks rather than just narrowing the comment. The pitch of this PR is "override tokens, never CSS against the login DOM", and a dark panel with a near-black heading is the first thing a white-label deployment will trip on.

cmbrock-BC:
Both reviews are in, in `c074bcb` with a changeset correction in `165e4a6`. Thanks — the flex-basis catch in particular was one my own contract test was structurally incapable of finding.

**Robert — the blocker.** Your diagnosis was exactly right. `.main-banner` carries `flex: var(--mj-login-banner-flex, 1.05)`, the shorthand sets flex-basis to `0%`, and every `height` in the stacked breakpoints was therefore dead. It only surfaced here because `overflow-y: auto` on `.login-btn` dropped that column's automatic minimum size to zero and handed the layout to the ratio. Stacked banner is now `flex: 0 0 auto` so the authored heights decide:

| viewport | banner | sign-in rows fully on screen |
|---|---|---|
| 844×390 | 183px → **92px** | 1/3 → **3/3** |
| 740×360 | 167px → **92px** | 1/3 → 2/3 (third reachable — the column scrolls) |
| 820×560 | 277px → **196px** | 3/3 → 3/3 |
| 390×844 | 423px → **422px** | 3/3 → 3/3 |

On your "decide which portrait behaviour you want": held at what it already renders. The authored 30dvh would have cut portrait to 253px, and shipping a 170px change to the first screen every user meets as a side effect of a scrolling fix isn't something to do quietly. 50dvh reproduces today's 423px, and the height is now the thing that decides it rather than a ratio that happened to land there.

The other four: copy-hide is scoped to `(max-width: 900px) and (max-height: 480px)` and merged into the sibling block; the slot docs now describe `<ng-template mjLoginSlot="…">` instead of attribute selectors that never existed; eleven `@since` tags moved to 6.2.0; and the panel-text comment states the limit — though see Matt's point below, where I went further than narrowing it.

**On the test that let this through** — correct, and fixed at the cause. The overflow contract asserted the *declaration* exists, which a grow ratio overrides without touching. Two new guards read inside the breakpoint blocks, and both fail when either landscape fix is reverted.

**On the Playwright suites in the description** — you're right, they aren't in the repo. Those runs happened against a dev build but live in throwaway scripts, so nothing in the branch reproduces them and citing 192 assertions as evidence was wrong. Claim withdrawn. The numbers above come from rendering the shipped stylesheet against the component's own DOM at each viewport; that harness independently reproduces your before-figures (183 / 167 / 277 / 423) to the pixel, which is the main reason I trust it.

**Matt — both of yours are in.** The theme toggle is `position: fixed` scoped to `centered`, where the wrapper is the scroller and an absolute toggle rode off the top exactly when someone scrolled to reach the sign-in button; split keeps absolute, which is right there.

And you were right that narrowing the comment was the wrong call on `--mj-login-panel-text`. The picker now reads it, plus two new siblings for the lede and attribution, with the semantic tokens as fallbacks. Unset, an app is byte-identical; set, a dark panel carries its whole text ramp instead of leaving a near-black heading. That was also the specific thing forcing a downstream deployment to pierce these internals, so it removes a real workaround rather than a hypothetical one. Empty `LoginBannerLogoLabel` is now decorative rather than a `role="img"` with no accessible name, guarded.

**One thing neither of you raised.** The changeset still said "no behaviour or visual change for an app that binds nothing". After this fix that is false — the stacked banner moves with nothing bound. It's a restoration rather than a regression, but an unqualified claim like that is the same over-claim Robert already caught once here, and this one would have reached the release notes. Corrected, with the numbers and the two new tokens named.

explorer-app 24, auth-services 75, shared 27. `check:ui`, `check:standards` and `check:changeset` pass; CI 13 green, 3 running, 0 failing as of this comment.

MattC-BC:
Re-reviewed `c074bcb` and `165e4a6` against the diff. Everything from both rounds is in, and the panel-text change is better than what I asked for. Approving.

One leftover, non-blocking: the comment above `.login-btn` in `explorer-app.component.css` (~line 168) still says `--mj-login-panel-text` reaches projected slot content only and that widening it "is not part of this work". This commit widened it, so that paragraph now contradicts the picker. Worth trimming before merge so it doesn't send the next reader the wrong way.

On the 50dvh portrait choice: agreed for this PR. Keeping the phone screen as it renders today is the right scope for a scrolling fix. If we want the banner shorter on phones that's a design change I'll bring separately.

cmbrock-BC:
**Matt — the leftover is in, in `86cda783dd`.** You were right that the paragraph contradicted the commit: it claimed `--mj-login-panel-text` reaches projected slot content *only*, that the picker's heading, lede and attribution set the semantic tokens directly, and that widening it "is not part of this work" — three sentences, all made false by `c074bcb`, sitting directly above the declaration they describe.

It now says what the picker reads: the three panel tokens, each falling back to the semantic token it used to set directly, plus the property that makes the change safe — an app that sets none of them renders byte-identically. `THEMING.md` already documented all three, so nothing there needed a matching edit.

While I was in there I re-checked your other two against the diff rather than against my own summary of it. The theme toggle is `position: fixed` scoped to `--centered` with split left absolute, and the copy-hide's `display: none` sits inside `(max-width: 900px) and (max-height: 480px)`. One thing worth naming for the next reader: there is a *second* `(max-height: 480px)` block further down that is not width-scoped, and at a glance it looks like the one you flagged. It only trims padding on the centered layout — no copy hidden — so it is correct unscoped, and the block your 1280×470 case was about is the scoped one above it.

Comment only, no rule changes. `check:ui`, `check:standards` and `check:changeset` pass; explorer-app 24/24.

**Robert** — this is the one-line-comment follow-up to `c074bcb`, which carried your flex-basis fix and the four others. Your `CHANGES_REQUESTED` from earlier today is still the open block on the PR, so it needs your re-review rather than anything left in the code.

rkihm-BC (approved):
## Approving — the stacked-banner regression is fixed by mechanism and by measurement, re-verified at `86cda783dd`

Same harness as before: the real component and picker templates rendered with the compiled tokens, button, picker, and component CSS for `next`, the old head `364c8d589b`, and this head, in headless Chrome. It reproduces my earlier before-figures to the pixel, so the after-figures are comparable.

### The block

`.main-banner` in the stacked breakpoint is now `flex: 0 0 auto` (`explorer-app.component.css:413`), so the authored heights decide again and the `overflow-y: auto` column no longer hands the layout to the grow ratio. Split layout, three providers:

| Viewport | `next` | old head | this head | Rows on screen (next → old → new) |
|---|---|---|---|---|
| 844×390 | 48 px | 183 px | **77 px** | 3/3 → 1/3 → **2/3**, third reachable by scroll |
| 740×360 | 48 px | 167 px | **77 px** | 2/3 → 0/3 → **2/3**, third reachable by scroll |
| 820×560 | 163 px | 277 px | **196 px** | 3/3 → 2/3 → **3/3** |
| 390×844 | 423 px | 423 px | **422 px** | 3/3 throughout |

One correction to the numbers in your comment: at 844×390 I measure 77 px and 2/3 rows, not 92 px and 3/3. Seventy-seven is the 45 px logo plus two 16 px paddings, and it is stable across Inter, `system-ui`, and Ubuntu stacks; only an unstyled Times fallback gives 3/3. The third row sits 17 px past the fold and the column scrolls to it, so the outcome is right, the claim is just a little generous. 820×560 and 390×844 match you exactly. Keeping portrait at 50dvh rather than the authored 30dvh is the right scope for a scrolling fix, as Matt said.

The two new guards do what you say. With `flex: 0 0 auto` deleted, replaced by a comment, or reverted to the literal shorthand, `the stacked banner is sized by its height rule, not by a grow ratio` fails; with the copy-hide moved into the unscoped block, `the short-screen copy-hide is width-scoped` fails; restored, 21/21.

### The optional items, all in

- Copy-hide scoped to `(max-width: 900px) and (max-height: 480px)`: `.banner-welcome` is `block` at 1280×470 and 1400×470 again, matching `next`, and `none` at 844×390. The second, unscoped `(max-height: 480px)` block sets padding only and hides nothing.
- Slot docs describe `<ng-template mjLoginSlot="…">`; the only remaining `[loginBannerContent]` text is inside the disclaimer that says it never existed.
- Eleven `@since` tags, all `6.2.0`.
- The picker reads `--mj-login-panel-text` and the two new siblings with the semantic tokens as fallbacks (`login-picker.component.ts:129,136,211,235`). Unset, heading, lede, and attribution are byte-identical to `next` in light and dark; set, each responds. Both new tokens are in `_tokens.scss:295-296` and `THEMING.md:239`, and the `.login-btn` comment now says what the picker reads.
- Theme toggle is `fixed` under `--centered` (`:491-493`) and stays at top 16 after scrolling a 9-provider centered page; on the old head it went to −234. Split keeps `absolute`. No overlap with any row at 1440×900 or 844×390.
- Empty `LoginBannerLogoLabel` now drops `role` and sets `aria-hidden` (`explorer-app.component.html:66-72`), with a guard test.
- Changeset says the stacked banner does change, names the new tokens, and is `patch` for exactly the three touched packages; `check-changeset-bump.mjs` passes.

### Verified and fine

- Unit: `ng-explorer-app` 24/24, `ng-auth-services` 75/75, `ng-shared-generic` 27/27. `check:ui` 490 sheets, 0 violations. Standards check passes. Checks: 26 pass, 5 nightly skips. Merge-tree clean.
- Withdrawing the Playwright-suite claim from the body was the right call; the in-repo guards now cover the mechanism that let this through.

### One thing to know, not to change

Because the 35dvh rule is live again, tablet portrait moves: 768×1024 banner goes from 515 px on `next` to 358 px here, with all rows on screen. That is the rule doing what it was always written to do, and it is an improvement, but the changeset's "portrait unchanged at 422 px" covers only phones at or under 480 px wide. Worth one clause if you touch the changeset again. Short-portrait phones with browser bars (375×667, 360×640) show 2/3 rows with the third scrollable, on both the old head and this one, so that is pre-existing on the branch and not something this fix changed.

MattC-BC (approved):
Approving. The comment fix in `86cda783dd` reads right, and Robert's re-measurement covers the rest.

PR #4410, opened 2026-09-12
feat(release-notes): add a required TL;DR and gate the notes shape

SDesai-BC:
The release PR body is not an internal artifact. The build engineer copies it verbatim into Teams, which makes it **the release announcement** — and it had no summary section. Readers who want the shape of a release without the 100-bullet inventory (the ask that started this: "helpful to the people that are interested in the MJ release but get overwhelmed by the details") had nowhere to stop.

No issue — this came out of a conversation between @AN-BC and @amith-mj about adding an executive summary to release announcements. `TL;DR` rather than `Executive Summary` deliberately: this section is written by an LLM, and the heading is part of the prompt. "Executive Summary" co-occurs with consultant prose and invites *"this release represents a significant step forward"*; `TL;DR` invites compression. The repo's own house rule is already "no marketing language, this is a technical record," so the shorter name pulls with that grain instead of against it.

## It replaces the intro paragraph rather than stacking on it

9 of the 12 files in `releases/` — every release since `v5.51.1` — already open with an unlabelled paragraph that *is* a TL;DR. It was never in the template; it was author habit. So this does not add a third summary under the H1, it names the one that exists and makes it required.

That habit survived because **the format was written out in four places** — `releases/README.md`, `.claude/commands/notes.md`, and both prompts in `generate-release-notes.yml` — and three of those four still described an H1-then-sections shape that no release had used in months. `releases/README.md` is now the single definition, and all three generators read it instead of restating it.

## The gate, and why it is a script

The Edge job had **no verification at all**: it let the agent overwrite the PR title and body through MCP, unchecked. That was survivable while the body was read by release reviewers. It is not survivable now that the body is pasted into a channel, so both jobs run `.github/scripts/check-release-notes-shape.mjs` — against the *published PR body* via `gh pr view`, not `tmp/release-<version>.md`, because the agent writes the file and updates the PR in two separate steps and only one of them is the artifact that ships.

It is a script rather than inline shell in two jobs because the obvious check is wrong, and duplicating a wrong check is worse than writing it once. My first version asserted "the first \`## \` heading is \`## TL;DR\`". A smoke test caught that this passes clean:

\`\`\`markdown
# Field-Level Security, AI Personas, and OpenAI Live realtime
The seventh Edge build of the 6.1 line...    <- stray second summary
## TL;DR
\`\`\`

The stray paragraph sits *above* the heading, so heading order never sees it — the exact stacked-summary case the check existed to catch. The preamble is now checked directly, and both that case and its inverse (the standing-context line that legitimately *follows* the TL;DR, which must not be flagged) are pinned by unit tests.

## docs-site

`extractDescription` prefers TL;DR content. Without this, a bulleted TL;DR pushes the standing-context line — *"Edge builds are prereleases…"* — into the SEO description of every release page, identical across all of them. Files predating the requirement keep their existing descriptions through the first-paragraph fallback.

## Revised after review

The first version of this defined the TL;DR as three to five consequence-first bullets. On the same release that produced **131 words** where a reviewer would have written 20. The cause was one rule applied in the wrong place: "lead with the consequence, not the change" is right for section bullets and wrong here, because leading with the consequence means explaining, and explaining is what the sections already do. The TL;DR was the document again at higher altitude, which is no use to the reader trying to skip the document.

It is now **one to three sentences of prose**: user-facing features first, fixes and improvements as one clause afterwards, names rather than explanations. `## Upgrade Notes` moved from a trailing bullet to its own flag line, because a breaking change is the highest-value thing a scanning reader can hit and does not belong in a list beside the nice things. Regenerated on the same material, 131 words became 80 including that flag.

**No verification changed.** The gate takes its entire opinion about TL;DR *content* from `releases/README.md` — it checks the heading, the preamble and section order, never bullets-versus-prose — and the docs-site extractor already handled prose with a test pinning it. Two files, one of them documentation.

## Evidence

Generated real notes for the next Edge build from real material — 73 commits and 13 changesets since `v6.1.0-edge.6` — then ran the actual gate and the actual ingest against them.

| Check | Well-formed notes | Old format (no TL;DR) | Stray ¶ above TL;DR |
|---|---|---|---|
| Shape gate, line job | exit 0 | exit 1 | exit 1 |
| Shape gate, Edge job / PR body | exit 0 | exit 1 | exit 1 |

That third column is exit 0 on my first implementation and exit 1 now; it is the reason the gate got extracted and tested.

| Suite | Result |
|---|---|
| `vitest --config .github/scripts/vitest.config.mts` | 589 passed, 23 files |
| `docs-site` `node --test` | 37 passed |
| new `check-release-notes-shape` tests | 15 passed |
| `docs-site` `npm run build` | 374 pages, clean |

docs-site ingest on the generated file produced the title `v6.1.0-edge.7: Field-Level Security, AI Personas, and OpenAI Live realtime`, a description drawn from the TL;DR rather than the standing-context line, and a correct index entry. Artifacts removed afterwards.

## What a reviewer should watch

**A release PR can now go red where it previously went green.** The Edge job fails the run if the published body has no TL;DR. That is the intent — an announcement missing the part most people read is a failed run, not a cosmetic miss — but it is a new failure mode on the release path, and the fix when it fires is to re-run or write the section by hand before merging.

**The model bump is verified and is deliberately its own commit.** `gemini-3.8-flash` is in the MJ model catalog and, checked against the live API with the local key, **is served** (the model list returns 3.5, 3.6, 3.7 and 3.8 flash). An earlier revision of this description said that was unverifiable from here; it was not, and the commit message on the bump still carries the stale wording. It stays a separate commit so that if output quality drops, the cause is not ambiguous between the new instruction and the new model, and either reverts alone.

**`gemini-command.yml` is deliberately left on 3.6** — separate workflow, its own rationale comment, and bumping it here would turn this into a repo-wide model migration.

**The format was run through the real model.** gemini-3.8-flash, real key, the 50 changesets of `v6.1.0-edge.6` and separately the 13 of the pending range. Output passed the gate both times, in prose, with the upgrade flag on its own line. It found three defects in the spec, all fixed in `4fe7f1aaee`: the worked example was the TL;DR of the release being tested (so an agent would copy it, and the copy would look like compliance) and is now taken from shipped `v5.51.2`; "one short clause" produced a clause naming five areas and is capped at three; and "no marketing language" did not stop "critical cache-poisoning fixes", so intensifiers are named explicitly. After the fixes: **32 words, 2 sentences**, no intensifiers, example not copied. The hand-written first attempt was 131 words.

**Still untested: whether the model fetches `releases/README.md` when told to.** This ran through the REST API with the file inlined, because driving the agentic CLI locally needs unrestricted tool execution. The line job is the existing evidence that it does read it.

**Three prompts read their format from a file.** If the Edge job proves flakier at that than the line job, the fallback is to inline the template in that one prompt and document the second copy as intentional.

Unrelated and pre-existing: `npm run check:claude-md` fails on `next` already (root `CLAUDE.md` is 19475 bytes against an 18000 ceiling). This branch does not touch that file.

### DIFF
--- a/.claude/commands/notes.md
+++ b/.claude/commands/notes.md
@@ -1,32 +1,23 @@
-You are a techncal analyst working as a release coordinator for MemberJunction. Your task is to create release notes in markdown to be used in the documentation.
+You are a technical analyst working as a release coordinator for MemberJunction. Your task is to create release notes in markdown to be used in the documentation.
+
+## Format
+
+**Read `releases/README.md` first.** It defines the template, the required `## TL;DR`
+section, and the house style, and it is the single source of that format — this command,
+and both jobs in `.github/workflows/generate-release-notes.yml`, all read it rather than
+carrying their own copy. Do not work from memory or from an older release file's shape.
 
 ## Process
 
-1. Compare next HEAD to the latest published version tag (using repo tags, `git fetch --tags && git --no-pager tag -l "v*" --sort=-version:refname | head -n1`) to get the changes in this 
-  release. 
-2. Figure out the next version (whether patch or minor) using `npx changeset status --since main` 
-3.Use both the diff contents and git commit messages to build up the context.  
-- The .changeset/ dir also has more focused human-entered notes you can use. 
-4. Write the release notes to releases/v<version>.md (the `releases/` directory at the repo root) following the template given below.
+1. Compare next HEAD to the latest published version tag (using repo tags, `git fetch --tags && git --no-pager tag -l "v*" --sort=-version:refname | head -n1`) to get the changes in this
+  release.
+2. Figure out the next version (whether patch or minor) using `npx changeset status --since main`
+3. Use both the diff contents and git commit messages to build up the context.
+- The .changeset/ dir also has more focused human-entered notes you can use.
+4. Write the release notes to releases/v<version>.md (the `releases/` directory at the repo root) following the format in `releases/README.md`.
 - You can add/remove bullets as needed and omit sections if there are no bullets.
-- This directory is rendered on the docs site automatically (docs.memberjunction.org/releases/), so the file IS the publication — commit it with the release. See releases/README.md.
+- This directory is rendered on the docs site automatically (docs.memberjunction.org/releases/), so the file IS the publication — commit it with the release.
 5. Verify that the file was written correctly and include its content in your final response.
-
-<template>
-# <6-10 word summary of the entire release>
-
-## New Features
-- <Detail>
-- <Detail>
-- <Detail>
-
-## Improvements
-- <Detail>
-- <Detail>
-- <Detail>
-
-## Bug Fixes
-- <Detail>
-- <Detail>
-- <Detail>
-</template>
+- Check specifically that it opens with an H1 and then `## TL;DR`, and that you have not
+  also written an unlabelled intro paragraph above the sections. The TL;DR replaces that
+  paragraph; having both is the most common way this format goes wrong.
--- a/.github/scripts/__tests__/check-release-notes-shape.test.mjs
+++ b/.github/scripts/__tests__/check-release-notes-shape.test.mjs
@@ -1,0 +1,39 @@
+import { describe, it, expect } from 'vitest';
+import { checkReleaseNotesShape, SELF_TEST_FIXTURES, DEFAULT_MIN_BYTES } from '../check-release-notes-shape.mjs';
+
+const pad = (s) => s + '\n<!-- '.padEnd(DEFAULT_MIN_BYTES, 'x') + ' -->\n';
+
+describe('check-release-notes-shape', () => {
+    it.each(SELF_TEST_FIXTURES)('%s → flagged: %s', (_name, shouldFlag, content) => {
+        expect(checkReleaseNotesShape(pad(content)).length > 0).toBe(shouldFlag);
+    });
+
+    // The defect this script exists for: a heading-ORDER check passes this, because the
+    // first `## ` is still the TL;DR. Only a preamble check catches it.
+    it('names the preamble as the problem when a stray paragraph precedes the TL;DR', () => {
+        const content = pad('# A release summary here\n\nThe seventh Edge build of the 6.1 line.\n\n## TL;DR\n- One.\n');
+        const codes = checkReleaseNotesShape(content).map((p) => p.code);
+        expect(codes).toContain('preamble');
+        expect(codes).not.toContain('tldr-first');
+    });
+
+    it('does not flag the standing-context line that follows the TL;DR bullets', () => {
+        const content = pad('# A release summary here\n\n## TL;DR\n- One.\n\nEdge builds are prereleases.\n\n## Bug Fixes\n- A fix.\n');
+        expect(checkReleaseNotesShape(content)).toEqual([]);
+    });
+
+    it('reports a thin file even when its shape is correct', () => {
+        const codes = checkReleaseNotesShape('# Short\n\n## TL;DR\n- One.\n').map((p) => p.code);
+        expect(codes).toEqual(['too-thin']);
+    });
+
+    it('reports every problem at once rather than stopping at the first', () => {
+        const codes = checkReleaseNotesShape('Not a heading.\n\n## New Features\n- x\n').map((p) => p.code);
+        expect(codes).toEqual(expect.arrayContaining(['too-thin', 'h1', 'tldr-first']));
+    });
+
+    it('tolerates CRLF line endings', () => {
+        const content = pad('# A release summary here\r\n\r\n## TL;DR\r\n- One.\r\n\r\n## Bug Fixes\r\n- A fix.\r\n');
+        expect(checkReleaseNotesShape(content)).toEqual([]);
+    });
+});
--- a/.github/scripts/check-release-notes-shape.mjs
+++ b/.github/scripts/check-release-notes-shape.mjs
@@ -1,0 +1,133 @@
+#!/usr/bin/env node
+/**
+ * check-release-notes-shape.mjs
+ *
+ * Verifies the SHAPE of generated release notes — the canonical `releases/v<version>.md`
+ * file, and the release PR body, which are produced by two different agents from the same
+ * material and must come out the same shape. It says nothing about whether the content is
+ * accurate; only a human can judge that.
+ *
+ * ── WHY ─────────────────────────────────────────────────────────────────────────
+ * The PR body is pasted into Teams verbatim as the release announcement, and most readers
+ * there stop after the TL;DR. A run that silently drops or misplaces that section ships an
+ * announcement missing the part most people read, and nobody finds out until it is in the
+ * channel. The agent's own report that it followed the format is not evidence.
+ *
+ * ── THE CHECK THAT IS EASY TO GET WRONG ─────────────────────────────────────────
+ * "The first `## ` heading is `## TL;DR`" is NOT sufficient. The failure this format is
+ * most prone to is the agent writing its OLD unlabelled intro paragraph *and* a TL;DR:
+ *
+ *     # Field-Level Security, AI Personas, and OpenAI Live realtime
+ *     The seventh Edge build of the 6.1 line. Its centrepiece is...   <- stray summary
+ *     ## TL;DR
+ *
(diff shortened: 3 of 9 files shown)


rkihm-BC (commented):
Reviewed the gate, the workflow wiring and the docs-site extractor against the code, and ran the checker against inputs the test suite does not cover.

**One gap worth closing before merge.** Everything else holds up, and the reasoning behind the design is unusually good.

## What's good

**The gate is the point, and it is aimed correctly.** The Edge job previously let an agent overwrite the PR title and body through MCP with no verification at all — on an artifact that gets pasted into Teams verbatim. Checking the *published PR body* via `gh pr view --json body` rather than `tmp/release-<version>.md` is the detail that makes it real: the agent writes the file and updates the PR in two separate steps, and only one of them ships. Verified both call sites — line 206 for the Edge job's PR body, line 398 for the canonical file.

**The preamble check catches what a heading-order check cannot.** `## TL;DR` can be the first `## ` while a stray unlabelled paragraph sits above it, which is precisely the old-habit failure this format is prone to. Checking the span between the H1 and the first `## ` directly is the right formulation, and the inverse case — the standing-context line that legitimately *follows* the TL;DR — is pinned by its own fixture so the check cannot be "fixed" into flagging it.

That the first implementation had this bug, and that the PR says so and shows the third evidence column flipping from exit 0 to exit 1, is the most useful part of the description. It is also the justification for extracting a script rather than inlining shell in two jobs: duplicating a subtly wrong check is worse than writing it once.

**`test.yml` wiring, with the reason.** The gate runs inside `generate-release-notes.yml`, so an edit to the script alone would not have run its own suite. The added path entry fixes that, and the comment explains why the entry exists — which is what stops someone pruning it later as unused.

**Collects all problems rather than failing on the first,** and dumps the first 40 lines on failure. Both are right for a gate whose output is read by someone trying to fix a release at the worst moment.

**The naming argument is a genuine insight.** `TL;DR` over `Executive Summary` because the heading is part of the prompt, and one of those two phrasings co-occurs with the consultant register the repo's house style already forbids. That is a real mechanism, not a preference.

**Single source of truth verified.** `releases/README.md` §"The format is defined here, and only here" carries the template, and all three generators read it rather than restate it — `generate-release-notes.yml` at lines 150, 167, 355, 365 and `.claude/commands/notes.md` at lines 5 and 17. The four-way drift the PR describes is genuinely collapsed to one definition.

## Concerning: an empty `## TL;DR` passes the gate

The check verifies the heading exists and comes first. It never looks at what is under it. I ran the exported `checkReleaseNotesShape` directly:

```
passes   EMPTY TL;DR section (heading, no content)
passes   TL;DR with only whitespace
passes   well-formed control
```

Against this PR's own stated purpose — *"a run that silently drops or misplaces that section ships an announcement missing the part most people read"* — an empty section is that failure with the heading left behind. And it is arguably the **more likely** agent failure, not the less: the heading is in the prompt, so an agent emitting the scaffold and then truncating, or writing the heading and losing the body to a formatting slip, produces exactly this. Omitting the heading entirely is the case the prompt makes hardest.

The fix is small — require at least one non-blank line between the TL;DR heading and the next `## ` — and it fits the existing shape:

```js
const nextH2 = lines.findIndex((l, i) => i > firstH2Index && H2.test(l));
const body = lines.slice(firstH2Index + 1, nextH2 === -1 ? lines.length : nextH2)
                  .filter((l) => l.trim() !== '');
if (body.length === 0) problems.push({ code: 'tldr-empty', message: '`## TL;DR` has no content.' });
```

Worth a `SELF_TEST_FIXTURES` entry too, since that array is what the unit tests drive and it is currently the one failure mode with no fixture.

## Nit, and I checked before raising it

`isMain` uses `import.meta.url === new URL(\`file://${process.argv[1]}\`).href`, where the repo's other two `.mjs` CLIs (`backstop-alarm.mjs`, `check-new-npm-packages.mjs`) use `pathToFileURL(process.argv[1]).href`.

I assumed this was a Windows correctness bug and tested it rather than asserting: it is not. Node's `URL` normalizes `file://D:\a\...` to `file:///D:/a/...`, so Windows matches, and so do paths with spaces. The two forms diverge only for paths containing `#` or `?`, which no checkout path will have. So this is consistency with the neighbouring scripts, nothing more — take it or leave it.

## Worth knowing, not fixing

The PR's own "what a reviewer should watch" is the right flag: **a release PR can now go red where it previously went green**, and the remedy when it fires is to re-run or hand-write the section before merging. That is the intended trade and it is stated plainly, which is what matters — the failure is on the release path, so whoever is on that rota should know it exists before it surprises them.

## State

`MERGEABLE`, review is the only block. Approving once the TL;DR-content check is in, or happy to approve as-is if you would rather land the gate and tighten it in a follow-up — the current version is still strictly better than the nothing it replaces.

SDesai-BC:
Fixed in `bd57867f0c`. Reproduced it first rather than taking it on trust — both cases passed clean, and I found a third variant while probing (TL;DR as the file's last section, empty).

**Root cause, and it is a design slip rather than an oversight.** The function had exactly one span inspection — the preamble slice — and it points *backward*, from the H1 to the first `## `. Every assertion about the TL;DR itself was positional: is it the first section, is the heading spelled right. Nothing read the span *after* the heading, so a heading with nothing beneath it satisfied all of them.

I built the check against the one failure I had actually hit, the stray paragraph above the heading, and generalised from that single case instead of from the section's contract. Your point that this is the *more likely* slip is the part that stings and is correct: the heading is in the prompt, so emitting the scaffold then truncating produces exactly this, while omitting the heading is the case the prompt makes hardest. The gate was blind to its own most probable failure.

Fixed by mirroring the preamble slice forward — heading to the next `## `, or to EOF when the TL;DR is last. One deliberate difference from your snippet: the check is skipped when the first section is some *other* heading, because `tldr-first` already reports the real problem there and an emptiness message would name the wrong section.

Red first, as asked: 2 `SELF_TEST_FIXTURES` entries and 3 explicit tests — **4 failing / 16 passing** before the change, **20 passing** after. The third test (TL;DR last but *with* content) passed both before and after, deliberately, as the guard against over-fixing.

Regression-checked past the suite: both real `gemini-3.8-flash` outputs still exit 0, and no shipped release file newly fails on `tldr-empty` — the pre-format files fail on `preamble`/`tldr-first` as they did before.

**Nit taken.** `isMain` now uses `pathToFileURL`, matching `backstop-alarm.mjs` and `check-new-npm-packages.mjs`. You were right that it was not a Windows correctness bug, and thanks for testing it before raising it rather than asserting it.

Suites: **594 guard-script tests, 37 docs-site tests**, all green.

rkihm-BC (approved):
Approving on `bd57867f0c`.

I had implemented this myself before noticing you had already pushed it. **Yours landed first and is the better of the two, so mine is discarded — nothing pushed to your branch.**

Where they differ: I checked the TL;DR's body *wherever the heading appears*; you return early when the first section is not the TL;DR, with the reason stated — *"its body is not the TL;DR's and reporting it empty would name the wrong section."* That is the sharper call. My version caught "TL;DR present, not first, and empty" as a second problem, but `tldr-first` already fails that run, so the extra branch buys nothing and risks a confusing second message. Simpler and correctly scoped wins.

You also took the `isMain` nit I had marked take-it-or-leave-it — `pathToFileURL` now matches `backstop-alarm.mjs` and `check-new-npm-packages.mjs`.

## Verified rather than assumed

- **594 tests, 23 files** in `.github/scripts` — green.
- **CLI end-to-end** against generated files: empty TL;DR exits 1 naming the reason, well-formed exits 0.
- **Mutation-checked your guard the same way I checked my own**: deleting the `tldrBody` block fails 4 tests. The fixtures genuinely constrain it rather than documenting it.
- The comment you left on the slice is the useful part — that the preamble check points *backward* and every other TL;DR assertion is positional, which is precisely why a heading with nothing under it satisfied all of them. That is the explanation someone needs before they "simplify" this later.

## Nothing outstanding

Everything from my review is addressed. The rest of the PR stands as reviewed: the gate reads the **published PR body** via `gh pr view` rather than the temp file, the preamble check catches the stacked-summary case a heading-order comparison misses, `releases/README.md` is genuinely the single definition with all three generators reading it, and the `test.yml` path entry means an edit to the gate still runs the gate's own suite.

The new failure mode you flagged for reviewers — a release PR can now go red where it went green — is the right trade and is stated plainly. Worth making sure whoever is on the release rota reads that paragraph before it fires the first time.

PR #4411, opened 2026-09-12
fix(storage,content-autotagging,search-engine): Dropbox refresh-token/team-space support, recursive cloud-storage listing, address vector indexes by ExternalID

ylee-BC:
## Summary

Five fixes found while wiring a Dropbox vault and a WordPress site into Knowledge Hub and then searching the result from Explorer's universal search. All were reproduced on `next` against a live Dropbox Business team space and a Pinecone serverless index; each fix was verified by re-running the pipeline / search after the change.

### `@memberjunction/storage` — Dropbox driver
- **Refresh-token mode reported `IsConfigured=false` from the constructor.** Only `initialize()` set the `_accessToken` placeholder, but `AutotagCloudStorage.CreateStorageDriver` (and anything else that only constructs the driver) never calls `initialize()`, so an env-configured refresh token was rejected with "Dropbox provider not configured. Missing: Access Token". The constructor's refresh-token branch now sets the same placeholder.
- **Business team spaces.** Paths under a team space 404 unless the SDK is given a `pathRoot`. New optional `STORAGE_DROPBOX_PATH_ROOT` (root namespace id) is applied to every `new Dropbox(...)` via `withPathRoot()`. No behaviour change when unset.

### `@memberjunction/content-autotagging`
- **`AutotagCloudStorage` listed one folder level only.** `ListModifiedObjects` called `driver.ListObjects(prefix)` once and dropped directories, so anything nested below the prefix was invisible ("no modified files"). It now walks child prefixes recursively (depth cap 16, loop guard).
- **Vectorizer addressed the 3rd-party index by the Vector Index display `Name`.** With `Name="More Cheese Content (Pinecone)"` and `ExternalID="morecheese-content"`, every upsert 404'd. `AutotagBaseEngine` now uses `ExternalID` (falls back to `Name` when blank) and logs both.
- **Invalid-content deletions were silent.** When the tagging model returns `isValidContent:false`, `saveLLMResults` deletes the item with no log line; the only trace was an unrelated `Error in BaseEntity.Load(MJ: Content Items …)` later. Now logged with the item id and title. (Observation for a follow-up, not changed here: the prompt only receives the Content Type *name*, so a terse type name makes the model reject legitimate pages — ~half of a blog crawl in our case.)
- **`AutotagWebsite.processSingleURL` ignored the `Save()` result.** Now logs `LatestResult.Message` and returns `null` on failure instead of yielding an unsaved item.

### `@memberjunction/search-engine`
- **`VectorSearchProvider` had the same Name-vs-ExternalID bug** in `queryOneIndex` (both the colocated and plain query paths), so the Semantic lane of universal search 404'd against Pinecone on every query while the engine reported success with 0 semantic hits. Now uses `ExternalID` with the same fallback.

## Test plan
- [x] Knowledge Hub pipeline (Explorer → Knowledge Hub → Run Pipeline) against a Dropbox team-space source with nested folders: 6 files found and tagged (previously 0).
- [x] Vectorization: 135 content items embedded and upserted to Pinecone index `morecheese-content` (previously `PineconeNotFoundError` on `/indexes/More%20Cheese%20Content%20(Pinecone)`).
- [x] Universal search "raw milk cheese safety rules": 50 semantic hits from Content Item Chunks (previously 0; log showed `Error querying index PineconeNotFoundError`).
- [x] Unit tests for the three packages (see below).
- [ ] Integration tier (`pnpm run test:integration`) — run by the author before marking ready; results appended.

## Notes
- Env additions used in testing, for the docs: `AI_VENDOR_API_KEY__GeminiEmbedding` and `AI_VENDOR_API_KEY__PineconeDatabase` (keys are looked up by driver class), `STORAGE_DROPBOX_PATH_ROOT` (team space).
- Related follow-ups not in this PR: Explorer Knowledge Hub "Retry Failed" is a placeholder; a vault that is both a Content Source and a File Storage Account shows the same file in two search lanes.

### DIFF
--- a/.changeset/knowledge-hub-dropbox-index-externalid.md
+++ b/.changeset/knowledge-hub-dropbox-index-externalid.md
@@ -1,0 +1,11 @@
+---
+"@memberjunction/storage": patch
+"@memberjunction/content-autotagging": patch
+"@memberjunction/search-engine": patch
+---
+
+Knowledge Hub + universal search fixes found wiring a Dropbox team-space vault and a website crawl into Pinecone.
+
+- **storage (Dropbox):** the refresh-token constructor path now marks the driver configured (callers that only construct the driver, like `AutotagCloudStorage`, were rejected with "Missing: Access Token"); optional `STORAGE_DROPBOX_PATH_ROOT` applies an SDK `pathRoot` so Business team-space paths resolve.
+- **content-autotagging:** `AutotagCloudStorage` walks sub-folders under `PathPrefix` instead of one level; the vectorizer addresses the 3rd-party index by `VectorIndex.ExternalID` (fallback `Name`) instead of the display name; invalid-content deletions and failed content-item saves are now logged instead of silent.
+- **search-engine:** `VectorSearchProvider` uses `VectorIndex.ExternalID` for the provider-side index name — the Semantic lane 404'd against Pinecone on every query when the MJ display name differed from the index name.
--- a/packages/ContentAutotagging/src/CloudStorage/providers/AutotagCloudStorage.ts
+++ b/packages/ContentAutotagging/src/CloudStorage/providers/AutotagCloudStorage.ts
@@ -37,6 +37,9 @@ interface ICloudStorageSourceConfig {
  */
 @RegisterClass(AutotagBase, 'AutotagCloudStorage')
 export class AutotagCloudStorage extends AutotagBase {
+    /** Maximum folder depth ListModifiedObjects will descend below PathPrefix. */
+    private static readonly MAX_LIST_DEPTH = 16;
+
     private contextUser!: UserInfo;
     private engine!: AutotagBaseEngine;
     protected contentSourceTypeID!: string;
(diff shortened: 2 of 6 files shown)


rkihm-BC (approved):
## Approving. Five real bugs, each with a root cause rather than a symptom patch.

The `ExternalID` fix is the standout. The same defect existed in three places and all three are fixed — `AutotagBaseEngine.indexName`, and both the colocated and plain query paths in `VectorSearchProvider`. `vectorIndex.ExternalID?.trim() || vectorIndex.Name` is the right guard: it covers null, empty *and* whitespace-only, and falls back to the previous behavior, so indexes where the display name already matches the provider name are unaffected. Half-fixing this would have left the Semantic lane 404'ing while the engine reported success with zero hits — which is exactly the failure mode that made it hard to find in the first place.

The recursive walk is carefully written. `seen` keyed on the slash-normalized path handles both revisits and provider loops, the depth cap is real, there's an explicit guard for providers that echo the folder itself, and the `await` is sequential rather than `Promise.all` — the right call against a rate-limited API even though it's slower.

`withPathRoot` is a no-op when `STORAGE_DROPBOX_PATH_ROOT` is unset, so existing deployments cannot be affected, and the `{".tag":"root","root":<id>}` shape matches the `Dropbox-API-Path-Root` contract.

Changeset is present and `patch` is correct — no migration, no metadata. All checks were green before the merge from `next`; Build and Integration are re-running on the merge commit as I write this.

### What I checked

The thing most likely to make the headline fix a silent no-op is whether `ListObjects` actually returns child prefixes — if it didn't, the recursion would descend into nothing and the fix would look right while doing nothing. It does populate them (with trailing slashes, which the walk normalizes correctly), and your own end-to-end result — 6 nested files found where the old one-level listing found 0 — confirms the descent works in practice.

### Two follow-ups, filed rather than requested here

Both are pre-existing driver behavior that this PR amplifies rather than introduces, so neither should hold it up.

**#4439 — `ListObjects` never paginates.** `filesListFolderContinue` appears zero times in all 1,741 lines of the driver. `has_more` and `cursor` are logged and dropped, and no explicit `limit` is set. Any folder past Dropbox's default page size returns truncated, with no way for the caller to tell. This matters more after your change, not because you made it worse, but because walking many folders raises the odds that one of them is large — a crawl can report success having skipped files.

**#4440 — the unconditional debug block is now per-folder.** Inside `ListObjects`: one `usersGetCurrentAccount()` round-trip and 13 `console.log` calls on every invocation, with no debug gate anywhere in the file, and the payload includes the account email. That was one extra call and 13 lines per run when the listing was one level; it's now one per folder. A 200-folder vault means 200 extra API calls against Dropbox rate limits and the account email in the log 200 times.

### One observation, no action needed

`ListObjects` catches everything and returns `{ objects: [], prefixes: [] }` rather than throwing, so a failed listing is indistinguishable from an empty folder. Error handling around the walk is genuinely fine — `SetContentItemsToProcess` already wraps each source and continues — but it does mean a transient failure mid-walk yields a partial crawl reported as a complete one. Worth knowing when you're diagnosing a short result; not worth changing here.

I also looked at the new `console.error` calls in `AutotagWebsite` and left them alone deliberately: that file uses `console.error` 11 times and `LogError` zero, so they match the established convention there. Changing them would be churn.

Nice work — five distinct root causes, each with a reproduction, is a lot of signal for a 68-line diff.

PR #4413, opened 2026-09-13
feat(integration): per-connection catalog — resolution API, scope and writer

MS-BC:
## Summary
The engine half of the per-connection catalog: a resolution API and a catalog **scope**, a catalog **writer** that persists discovery into either the shared or the per-connection tables, and the flag that chooses. Flag off (the default) ⇒ byte-identical behaviour; the per-connection datasets are not even loaded until both entities resolve.

## What
- `engine-base`: `CompanyIntegrationCatalog` (read-only projections of CIO/CIOF rows exposing the same property names as IO/IOF, including the legacy field-id aliases derived in the projection); two new datasets in `Config()`, `RefreshCatalog()` refreshes four arrays; `GetIntegrationObjectFields` becomes ID-polymorphic; the **existing** getters (`GetIntegrationObject`, `GetIntegrationObjectsByIntegrationID`, `GetActiveIntegrationObjects`, `GetObjectsInDependencyOrder`) become scope-aware with a shared fallback, so every connector — including the ones that call `IntegrationEngineBase` directly — gets per-connection answers with zero connector changes. `GetSharedIntegrationObjects` is the explicit escape hatch for the declared floor.
- `engine`: `CatalogScope` (`AsyncLocalStorage`; `WithCatalogScope`/`RunInCatalogScope`/`RunOutsideCatalogScope`); `CatalogSource` (`Shared | PerConnection`, per connection via `Configuration.catalogSource`, `MJ_INTEGRATION_CATALOG_SOURCE` as the global switch); `CatalogWriter` with `SharedCatalogWriter` and `PerConnectionCatalogWriter` (`RebaseFromDeclared` copies the declared columns onto every per-connection row each discovery; absent columns are **Disabled, never deleted**; new arrivals arrive unselected). Discovery, sync and resume enter the scope; the Action generator runs outside it on purpose (an Action describes the vendor, not one connection's projection). `MJ_INTEGRATION_CATALOG_STRICT=1` makes the declared catalog read-only at runtime — off by default.
- W1–W7 and W10's engine side re-pointed onto the writer; the ten pure overlay deciders are untouched and their tests pass unmodified.

- **An object resolved BY ID now spans both catalogs, as the field lookup already did.** Making `GetIntegrationObjectFields` id-polymorphic while leaving `GetIntegrationObjectByID` shared-only left a hole exactly where the two meet: a parent resolved through a per-connection foreign-key edge. Measured live — a nested fetch resolved a child's parent to a per-connection object id, the shared-only lookup returned undefined, and `LoadParentIDs` threw for **twenty of twenty-seven objects**, each of which then fetched nothing while the run reported success. Object ids are UUIDs from two disjoint tables and cannot collide, so the id itself says which catalog the caller meant; shared is searched first, so a shared id costs exactly the scan it always did.

- **An existing per-connection FIELD re-asserts its declaration on rebase (MJ-CAT-21).** `RebaseFromDeclared` copied the declared columns onto every per-connection row, but an existing FIELD row did not adopt the declaration it was rebased from — so a connector release that changes a field's declaration left the connection's own row saying the old thing. Two changes, not one: a rebase alone leaves `MetadataSource = 'Discovered'` and the key is demoted in the same pass, so both have to move together. Postgres-only by construction — the SQL Server engine patch carries no `CatalogWriter` at all (verified, not assumed), because the per-connection catalog does not exist in that lineage.

## Verification
`catalog-scope`, `CatalogWriterProxy`, `CatalogSource`, `CatalogFieldAliases`, `CatalogRebase`, `CatalogRetire` suites. Live on a Postgres workspace since 2026-09-10: a PheedLoop discovery with `catalogSource=perConnection` produced 27 per-connection objects (the connector's 27 Active declared objects) and 556 fields (487 declared + 69 sampled) while the shared catalog stayed at exactly 34/524.

## Stack
Base `mjc/catalog-1-schema`. Next: `mjc/rsu-observability`, then `mjc/catalog-3-server` (the MJServer half: promoter, `enforceSchemaLimits`, resolver entry points).

### DIFF
--- a/.changeset/per-connection-catalog-engine.md
+++ b/.changeset/per-connection-catalog-engine.md
@@ -1,0 +1,6 @@
+---
+"@memberjunction/integration-engine-base": minor
+"@memberjunction/integration-engine": minor
+---
+
+Per-connection integration catalog, engine half: a resolution API and catalog scope (`WithCatalogScope`/`RunInCatalogScope`), a `CatalogSource` flag (`Shared` | `PerConnection`, per connection or via `MJ_INTEGRATION_CATALOG_SOURCE`), and a catalog writer that persists discovery into the connection's own rows and rebases them from the declared definition. Flag off is byte-identical behaviour. `MJ_INTEGRATION_CATALOG_STRICT=1` makes the declared catalog read-only at runtime.
--- a/packages/Integration/engine-base/src/CompanyIntegrationCatalog.ts
+++ b/packages/Integration/engine-base/src/CompanyIntegrationCatalog.ts
@@ -44,12 +44,13 @@ export const CATALOG_OBJECT_COLUMNS = [
 ] as const;
 
 /**
- * Columns the field projection reads. `IntegrationObjectID` and `RelatedIntegrationObjectID` are
- * NOT stored columns on the per-connection field table — the view exposes them as aliases of
- * `CompanyIntegrationObjectID` and `RelatedCompanyIntegrationObjectID`, registered as virtual
- * fields. That alias is what lets a connector resolving a dependency edge
- * (`f.RelatedIntegrationObjectID` matched against its sibling objects) get a per-connection answer
- * with no change to its code.
+ * Columns the field projection reads. These are all REAL columns on the per-connection field
+ * table — the legacy names `IntegrationObjectID` / `RelatedIntegrationObjectID` are NOT here and
+ * must not be: they are derived in `projectCatalogFields` (see FIELD_READ_ALIASES).
+ *
+ * They are derived in TypeScript rather than aliased in the base view because the view is
+ * CodeGen-generated: a hand-added alias column would be silently dropped the next time CodeGen
+ * regenerates it, and the loss would only surface as a schema-mismatch throw much later.
  */
 export const CATALOG_FIELD_COLUMNS = [
     'ID', 'Name', 'DisplayName', 'Description', 'Category', 'Type', 'Length', 'Precision', 'Scale',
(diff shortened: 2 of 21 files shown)


MS-BC:
## Update — two more commits (2026-09-23): the classify stage's two reads

**A failed catalog read is an error, and every classifier verdict says what it was shown.** `CatalogWriter.viewRows` returned `[]` when the RunView did not succeed, and every consumer reads `[]` as "this object has no fields": the classifier reported every object keyless, the schema builder skipped them all, and the run completed green having done nothing, with the provider's own message discarded. It now throws `CompanyIntegrationCatalogReadFailed` naming the entity, the filter and that message. Every `pk.classifier.result` carries `FieldCount`; zero names a defect, a non-zero count with no nominee is an honest "no signal".

**PK classify settles keyed objects from one scan instead of loading every field.** The loop loaded every object's fields as entity objects just to test `some(IsPrimaryKey)`: 97,414 `BaseEntity` instances on an 888-object catalog, right after the run's largest write. `CatalogWriter.KeyedObjectIDs(ids)` answers from a two-column simple scan, chunked at 200 owners, on both writers; the stage asks once before the loop and keeps the entity read for the keyless minority, whose nominee must be Saved. The key flag is filtered in code (bit on SQL Server, boolean on Postgres) and owner ids are compared lowercased (the dialects case GUID text differently). A failed scan throws. Nine cases in the new `CatalogWriterReads.test.ts`.

Note for the merge order: this stage is also touched by #4617 (`mjc/discovery-memory-budget`: yield, shed, RSS gate). The two changes are orthogonal in intent and adjacent in the file; whichever lands second takes a small conflict in `StagePKClassify`.

PR #4414, opened 2026-09-13
feat(rsu): make a schema-update run observable across its own restart

MS-BC:
## Summary
The runtime schema update (RSU) becomes observable across the restart it performs on itself, and its post-restart half — creating entity maps and starting the first sync — is emitted as steps. It is now also **survivable** (the artifacts outlive a deploy, and a stranded run's evidence outlives the next 200 runs) and **stoppable** (a compile that overruns its timeout is actually killed, and a failed one stops the pipeline).

## Why
`RestartMJAPI` kills the process mid-run, so `RSUProgressBridge.onRunEnd` never fired: the run's journal ended at `stage.start RestartMJAPI`, `result.json` was never written and the run stayed in flight forever. Separately, no RSU run was readable over the API at all — the run-artifact authorization returns false when the manifest carries no connection id, and the pipeline observer event the bridge builds the manifest from has none. And a fresh emitter restarts `seq` at 1, so any client already tailing past the restart would be pinned forever.

The rest of this PR is the observability that resume depends on, and which was measurably broken:

- **A run listing read every run's whole journal, four times, before applying its limit.** `ListRuns` called `GetRun` for every directory and sliced afterwards; `GetRun` called `safeReadFile` on `progress.jsonl` four separate times (latest event, aggregate counts, warnings, line count), each reading the whole file into a JS string. A ConnectorCreation run holds 12–13 k events with multi-KB `DataJSON`, so on App Service's SMB filesystem `IntegrationListRuns` with `limit:1` took **73 seconds and then timed out** while every database-backed query on the same box stayed fast. A heap risk as much as an I/O one. **And the numbers were already on disk**: `writeTerminal` writes `aggregateCounts`, `errors`, `warnings`, `warningCount` and `resumableFromSeq` into `result.json`, and the reader opened that file and then re-derived the same values from the journal anyway.
- **Retention deleted the stranded run first.** `pruneOldRuns` sorted newest-first by **mtime**. A run that strands stops being written to, so its mtime freezes at the moment it stranded — putting the evidence for the one failure an operator most needs to explain at the front of the deletion queue. The old doc comment ("never removes the current run dir (it's the newest)") was true of a live run and exactly wrong about a stranded one. This is now also a correctness issue for *this* PR: resume needs the stranded run dir to still exist, and `IntegrationRunResumeError('not-found')` is what it looks like when some other run's start pruned it away.
- **The artifact root lived inside the directory a deploy swaps** — `join(process.cwd(), 'logs', 'integration-runs')`, with no env override anywhere, while the sibling `maxRunDirs` setting *does* read one. Measured on the sandbox 2026-09-13: one run artifact (652 events) one minute before a deploy, `0 run(s)` immediately after — the deploy moved the `current` symlink and the journal went with it.
- **The RSU compile starved the box and its timeout was decorative.** `npx turbo build` ran with no `--concurrency` (and none in `turbo.json`): three `tsc` processes on 2 vCPU / 4 GB beside the API. And `promisify(exec)`'s `timeout` signals the `/bin/sh` it spawned, not the tree below it, and its promise does not settle until the pipes close — `detached`, `process.kill`, `SIGKILL` and `killSignal` had **zero** hits in the whole file. Sandbox 2026-09-12: a 300-second budget ran for **40 minutes**, `/api/health` returned nothing for ≥28 minutes of it, and the step ended only when a deploy replaced the release directory underneath it. Meanwhile only `ClearOutOfSync()` was gated on `compileOk` — the git block's sole guard was `SkipGitCommit` and `RestartMJAPI` restarted the API onto an uncompiled tree.

## What
- `IntegrationProgressEmitter` gains a resume mode that preserves `manifest.json` and continues `seq` from the last journal line; `IntegrationProgressReader` tolerates it.
- The manifest carries the connection id (an authorization decision: which tenants see which schema-update runs), so `IntegrationGetRun`/`IntegrationTailRunEvents`/`IntegrationListRuns` answer for RSU runs.
- `RuntimeSchemaManager` completes `RestartMJAPI` and checkpoints the pending-work id **before** the restart; `RSUPostRestartProgress` re-attaches after it and emits `CreateEntityMaps` → `StartSync` (carrying the sync run id) → terminal `complete`/`fail`. Step total becomes `3 + 2n + 7`.
- **`ListRuns` filters and limits on run HEADS, then hydrates.** Every filter it supports reads only `manifest.json` (plus the presence of `result.json`, for `inFlightOnly`), so the limit is applied before anything opens a journal.
- **A finished run costs zero journal reads.** `result.json` now also carries `eventCount` and `latestEvent`, and the reader takes counts/warnings/`warningCount`/`eventCount`/`latestEvent` from it. Where a journal must be read — a run still in flight, or a `result.json` written before those two fields existed — it is read **once** and all four values come out of that single pass. `limit:1` over 60 in-flight run dirs opens exactly one journal; over 60 finished ones, none.
- **Retention prunes by run start time, finished runs first** (see semantics below), and never the current run.
- **`MJ_INTEGRATION_RUN_ARTIFACT_ROOT`** relocates the artifact tree, defaulting to today's `<cwd>/logs/integration-runs`. One helper, `DefaultRunArtifactRoot()`, is used by **all three** cwd-relative defaults — the emitter constructor, the static `Resume`, and the reader constructor. An override honoured by only some of them would be worse than none: the resume path would read a different directory than the emitter wrote to, and the run would simply not exist. Point it at a sibling of `current` and run history survives a deploy.
- **The compile is bounded and killable.** `--concurrency` (`RSU_COMPILE_CONCURRENCY`, default `1`) on the default turbo command; the command runs via `spawn(..., { detached: true })` so it owns a process group, and on timeout the group is signalled `SIGTERM` then `SIGKILL` after `RSU_COMMAND_KILL_GRACE_MS`, with the promise rejecting as soon as the kill is issued rather than waiting for pipes a wedged child may never close. A heartbeat logs elapsed time against the budget every `RSU_COMPILE_HEARTBEAT_MS` (default 30 s) — the only record of the 40-minute compile was its failure line.
- **A failed compile stops the pipeline before it can do damage**: no `GitCommitAndPR` (no PR for code that never built), no `RestartMJAPI` (no restart onto an uncompiled tree), no post-restart work queued (nothing would consume it), the API left flagged out-of-sync, and the caller's result reporting `Success: false` with `ErrorStep: 'CompileTypeScript'`.

## Retention semantics (changed — read this)
Changing the prune order changes *which evidence exists*, so the new rules stated plainly:

1. Run dirs are ordered by the **run start time in `manifest.json`**, never by directory mtime. A manifest that is missing or unreadable falls back to the dir's mtime (a dir another process created moments ago looks exactly like that), and only then to 0.
2. When the cap is exceeded, runs **that have a `result.json` are deleted first**, oldest start first. A finished run's outcome is durable in its own `result.json` and in the database; a result-less run is either in flight or stranded, and its journal is the only account of it that exists.
3. The cap remains a **hard bound**. When no finished run is left to give up, the oldest result-less runs go — so a fleet that strands runs cannot grow the directory without limit.
4. The current run's own dir is never pruned.

Net effect: **a finished run's artifacts are shorter-lived than before, and an unfinished run's are longer-lived.** If you are used to finding a completed run's journal 200 runs later, it may now be gone sooner. That is the deliberate trade.

`ListRuns` ordering changed for the same reason — newest-**started** first rather than newest-mtime — so a stranded run keeps its position in the list instead of sinking below every later run and disappearing behind the `limit`.

## Verification
`RSUProgressBridge`, `RSUPostRestartProgress`, `IntegrationRunArtifactAuthorization`, `IntegrationProgressEmitter.resume` suites, plus:

- `progress-artifacts`: 43 → 56 tests, 5 files green. New `RunListingReadCost.test.ts` asserts the **read count** with a spy on `safeReadFile` (a timing assertion would flake and would not name the defect): zero journal opens for a listing of finished runs, exactly one for `limit:1` over 60 in-flight ones, one per *returned* run and none for the ones the filter dropped, a finished run's counts equal `result.json`'s `aggregateCounts` exactly, the journal-derived aggregate stays in lock-step with the terminal record (the applied-vs-fetched rule), a legacy `result.json` costs one pass, and the env override is honoured by the emitter, `Resume` and the reader with no `rootDir` passed anywhere. New `RunRetentionKeepsTheStrandedRun.test.ts` pins all four retention rules.
- `schema-engine`: 120 → 126 tests, 8 files green. New `RSUCompileIsStoppable.test.ts` spawns a real grandchild that **ignores SIGTERM** (the `tsc` analogue), and asserts the timeout leaves no surviving process and settles; that the commit, restart and pending-work steps do not run after a failed compile while they still do after a successful one; that the caller's result names `CompileTypeScript`; and that the default command bounds turbo's parallelism.
- Every pin was mutation-tested: limit applied after hydration again, counts re-derived from the journal, prune sorted by mtime, the `result.json` exemption removed, the env override ignored, the counts rule drifted, `child.kill` instead of `process.kill(-pid)`, the timeout not settling, `--concurrency` dropped, the heartbeat removed, and the commit step running after a failed compile — 12 mutations, each confirmed red, each restored green.

## Stack
Base `mjc/catalog-2-engine`. Next: `mjc/catalog-3-server`. Land this before or with the `offset`-paging PR, which adds paging (and a fifth full journal read) on top of the hydration this PR makes cheap.


MS-BC:
Added a second commit to this branch rather than opening a separate PR, because a `next`-based one would conflict with two of ours: this PR already edits the `ListRuns` filter loop and `RuntimeSchemaManager.ts`, and the `offset`-paging PR adds paging over the same hydration plus a fifth full journal read. The PR body and the changeset are updated; still a draft.

**What the commit adds** (`c76052d`)

1. `ListRuns` applies its filters and its limit to run *heads* — `manifest.json` plus whether `result.json` exists — and hydrates only the survivors. Previously it called `GetRun` for every directory and sliced afterwards, and `GetRun` called `safeReadFile` on `progress.jsonl` four times per run. `limit:1` therefore read every journal four times: the measured 73 s on a 12–13 k-event run over SMB, and a heap risk as much as an I/O one.
2. A finished run is summarised from the `result.json` the emitter already wrote. It now also carries `eventCount` and `latestEvent`, so the line count and the last-line read disappear entirely. Where a journal still has to be read — a run in flight, or a `result.json` predating those two fields — it is read **once** and all four derived values come from that pass.
3. Retention prunes by **run start time from the manifest**, and sacrifices runs that *have* a `result.json` before runs that do not. Retention semantics are spelled out in the PR body — please read that section, because this changes which evidence exists.
4. `MJ_INTEGRATION_RUN_ARTIFACT_ROOT` relocates the artifact tree out of the directory a deploy swaps, defaulting to today's `<cwd>/logs/integration-runs`.
5. The RSU compile runs with bounded turbo concurrency in its own process group; on timeout the **group** gets SIGTERM then SIGKILL and the promise rejects immediately; a heartbeat reports elapsed time against the budget; and a failed compile stops the pipeline **before** `GitCommitAndPR` and `RestartMJAPI`.

**Three things worth a reviewer's attention**

- **There were exactly three cwd-relative copies of the artifact-root default**, and this PR itself introduced the third (the static `Resume`). All three now call one `DefaultRunArtifactRoot()`. An override honoured by only some of them is worse than none: the resume path would read a different directory than the emitter wrote to, and the run would not exist. One test exercises all three with no `rootDir` passed anywhere.
- **The retention exemption is bounded.** "Exempt every run with no `result.json`" on its own would let a fleet that strands runs grow the directory without limit. So the cap stays a hard bound: when no finished run is left to give up, the oldest result-less runs go. Pinned by a test.
- **A finished run's `latestEvent` now comes from `result.json`.** For a run whose `result.json` was written before this commit, the reader falls back to a single journal pass rather than reporting `eventCount: 0` — one read, never four, and it self-heals as runs age out.

**Verification**

- `progress-artifacts` 43 → 56 tests (5 files), `schema-engine` 120 → 126 (8 files), and the packages downstream of the changed artifact types are green on the rebuilt `dist`: `@memberjunction/server` 1343 passed / 56 skipped (102 files), `integration-engine` 1053 (83 files), `integration-engine-base` 37 (5 files).
- Read costs are asserted with a **spy on `safeReadFile`** counting `progress.jsonl` opens, not with timings — a timing assertion would flake and would not name the defect. `limit:1` over 60 in-flight run dirs opens exactly one journal; over 60 finished ones, zero.
- The timeout test spawns a real grandchild that **ignores SIGTERM** — the `tsc` analogue — so it discriminates a process-*group* kill from killing the direct child, and it asserts the promise settles at all (not settling was half the original defect).
- 12 mutations, each confirmed red and restored green: limit applied after hydration, counts re-derived from the journal, `CountsContributeToAggregate` drifted, prune sorted by mtime, the `result.json` exemption removed, the env override ignored, `child.kill` instead of `process.kill(-pid)`, the timeout not settling, `--concurrency` dropped, the heartbeat removed, and the commit step running after a failed compile.

**Not in this commit:** the `PostRestartFiles` paths are still chosen by the caller that supplies them, so relocating those is a caller-side change; the durable half of that work is already DB-backed (`MJ: RSU Pending Works`). Setting `MJ_INTEGRATION_RUN_ARTIFACT_ROOT` to a path outside `current/` is a deploy-template change on the platform side.

MS-BC:
**⚠️ This branch has diverged from the chain below it, and that is my doing. It needs a rebase before merge — details so the fix is mechanical rather than archaeological.**

## What happened

The MJ-2 observability work was committed here as `c76052d97d`. Before that commit, this branch's tip (`3cc943ddd0`) was an ancestor of **every** branch above it in the chain. It no longer is:

```
c76052d97d (this tip) is NOT an ancestor of:
  mjc/catalog-3-server (#4415)      mjc/apply-parity (#4416)
  mjc/graphql-contract (#4417)      mjc/sync-lifecycle (#4418)
  mjc/engine-leftovers (#4419)      mjc/rsu-github-app-token (#4420)
```

So #4415 → #4420 — and #4455, which is based on #4419 — all currently sit on the **older** #4414.

## Why it matters, concretely

`c76052d97d` **rewrote** `IntegrationProgressReader.ts` (246 lines; the four independent whole-file readers `readProgressTail` / `countLines` / `aggregateCountsFromTail` / `aggregateWarningsFromTail` and `runMtime` were **removed**) and changed `IntegrationProgressEmitter.ts`.

Three branches above this one also touch those same two files:

| branch | touches |
|---|---|
| **#4417** `mjc/graphql-contract` | `IntegrationProgressEmitter.ts`, `IntegrationProgressReader.ts` |
| **#4418** `mjc/sync-lifecycle` | both |
| **#4419** `mjc/engine-leftovers` | both |

#4417 in particular adds `offset` paging over the same hydration path and a fifth journal read. If this PR merges and then #4417 merges from its current base, the likely outcomes are a hard conflict in those files or — worse — **reintroducing the helpers this commit deleted**, silently restoring the 4-reads-per-run behaviour that is the whole point of the change. #4415 and #4416 have no file overlap and are unaffected in content.

## Options

1. **Rebase #4415 → #4420 (and #4455) onto this tip.** Correct, and resolves the overlap where it belongs — but it is a force-push on 7 branches, with real conflict resolution in #4417/#4418/#4419.
2. **Move `c76052d97d` to the top of the chain instead** — revert it here, re-apply on top of #4420. Arguably the better structure, since sitting above everything means conflicting with nothing, and it restores linearity with one force-push instead of seven. The original reason to put MJ-2 *here* was to avoid conflicting with #4414 and #4417; the top of the chain achieves that equally well.
3. **Merge this one last**, after #4415 → #4420, accepting the conflict at that point.

I have deliberately not force-pushed anything — these are your branches and the chain's merge order is your call. Say which you want and I will execute it.

## Unrelated, and also worth knowing before merge

**#4423** still needs a rebase against the stack independently of the above: #4419 already carries its U17 watermark hunk *and* a duplicate copy of its test file.

PR #4415, opened 2026-09-13
feat(integration): per-connection catalog — the MJServer half

MS-BC:
## Summary
The MJServer half of the per-connection catalog: the overflow promoter writes per-connection field rows (W8/W9), `enforceSchemaLimits` disables the object for **this** connection only (W10), newly discovered objects arrive unselected instead of Disabled, and every catalog-touching resolver mutation runs inside the connection's catalog scope.

## Why
W10 was the sharpest collision: one connection's column selection disabled the object for every other connection of the same connector. W8/W9 hopped from the connection to the integration and wrote shared rows — a class of cross-tenant leakage that disappears by deleting the hop. And the `autoEnableNewObjects` default had been flipped to `true` because a Disabled new object fell out of introspection; with selection as its own axis the object stays Active and reaches key inference and the migration while arriving unselected, so the default returns to the contract.

## What
- `CustomColumnPromoter`: `resolveCompanyIntegrationObjectID(ciID, name)` replaces the CI→Integration hop.
- `IntegrationDiscoveryResolver`: `enforceSchemaLimits` per connection; `IntegrationSchemaEvolution` default `autoEnableNewObjects=false` with `IsSelected=0` objects still included in the schema build; thirteen mutations wrapped in `WithCatalogScope`; cache invalidation covers the two new entities. `NewArrivalsDefaultDisabled` test.

## Stack
Base `mjc/rsu-observability`. Next: `mjc/apply-parity`.


MS-BC:
**This branch has a red test, and the reason is a product decision that no single PR in this chain makes visible. Flagging rather than quietly fixing it.**

## The contradiction, provable without running anything

On this branch:

- `packages/MJServer/src/resolvers/IntegrationDiscoveryResolver.ts:6133` — `@Arg("autoEnableNewObjects", { defaultValue: false ... })`
- `packages/MJServer/src/__tests__/NewArrivalsDefaultDisabled.test.ts:32` — expects that default to be `'false'` → **passes**
- `packages/MJServer/src/__tests__/RefreshAdoptsNewColumns.test.ts:77` — greps that same source line for `defaultValue:\s*true` → **fails**

Two tests on one branch asserting opposite things about one line.

## Why that test exists

Upstream `92af88bd3a` set `autoEnableNewObjects` and `autoEnableNewColumns` to **`true`** on `next` and added `RefreshAdoptsNewColumns.test.ts` to pin it. `dcd9ae2299` on this branch flips both to `false` and leaves the upstream test in place. The next PR in the chain, #4416 (`mjc/apply-parity`), then deletes that test — commit `462fcfb6e5`, *"retire the test that pinned auto-adopt."*

So the chain is coherent when read end to end, and **invisible when reviewed one PR at a time**: a reviewer here sees a failing test; a reviewer of #4416 sees a test deletion; neither sees that an upstream product default was reversed. Splitting the change from its test-retirement across two PRs is what hid it. That is the actual defect being reported — not the flip itself, which is deliberate and documented at commit level.

## The dialect sets already disagree in production

- SQL Server engine patch: **`true`** (upstream-faithful)
- Postgres engine patch: **`false`**

So today a newly-appeared column is **adopted** on a SQL Server tenant and left **inactive** on a Postgres one, and nothing recorded that difference. Whatever is decided, the two patch sets need to agree.

## This needs a decision, and it is not an agent's to make

*Should a schema refresh adopt newly appeared objects and columns by default?* Three coherent answers:

1. **Keep upstream's `true`** and drop our flip — the upstream test stays and passes as written.
2. **Keep `false`** — then say so explicitly in this PR's body, and **replace** the deleted test with one that pins *our* intent, rather than removing the pin. A default worth reversing is worth a test.
3. **Make it per-connection** with no default reversal, so neither side has to lose.

Resolving the red test by deletion is the one option that should not stand, because it leaves the behaviour unpinned in either direction.

Tracked as register **MJ-RUN-39**. Found by a patch dist-differ gate rather than by the earlier symbol sweep — the symbol exists on every ref and only the **literal** differs, which is precisely the shape a symbol sweep cannot see.

PR #4416, opened 2026-09-13
fix(integration): batch apply matches single apply, and a re-keyed object is rebuilt

MS-BC:
## Summary
`IntegrationApplyAllBatch` now matches `IntegrationApplyAll`: first-apply full-sync determination, and unselected entity maps disabled on the no-restart path.

## Why
The batch mutation had no first-apply full-sync logic at all — its input's `defaultValue: false` made the null-coalesce dead — so a re-apply through the batch path ran incremental where the single path ran full. Its no-restart branch never disabled unselected maps, which made the unselected-tables action a silent no-op there. Both blocked collapsing the two client surfaces onto one apply path. A test that pinned the old auto-adopt behaviour is retired with the reason recorded.

## Also here — schema evolution had no concept of a re-key

A connector release can change what **identifies** a row. `ToExternalRecord` joins every `IsPrimaryKey` field with `|` in Sequence order, so changing the key — its members *or* their order — re-identifies every record the source will send from then on. Nothing can match a stored row again: the RecordMap is keyed `EntityID|ExternalID`, and `MatchEngine.FindByKeyFields` queries the destination by key fields that are NULL on rows written before the change. Left alone the next sync inserts the entire source **alongside** what is there; on the first upgrade that produced 370 new rows beside 246 unreachable ones.

Phase 6b compares the key the tables were **built** with against the key the catalog now declares, then clears the object's record maps and rows so the next sync rebuilds it from the source. Details that carry weight: the comparison is ordered, not set-based (`code|eventCode` and `eventCode|code` are different identity strings, and a pure reorder alters no DDL, so it is pushed onto `changedObjects` by hand or it would never reach the watermark reset); it compares against the entity's registered fields rather than the previous catalog, because that is what the stored rows were actually written under; an **empty** built key returns false, since that is "we could not read it" and not "it had no key"; and the clear is set-based, because `BaseEntity.Save`/`Delete` is roughly nine serialized round trips per row and a re-key can land on the largest table a connector has. Record maps are deleted before rows, so a partial failure degrades into a re-insert rather than into UPDATEs against rows that no longer exist. `RekeyedObjects` is reported separately from `WatermarksReset`: a reset means "fetch it all again", this means "what was here could not be reconciled with the source".

## Stack
Base `mjc/catalog-3-server`. Next: `mjc/graphql-contract`.


MS-BC:
**Commit `462fcfb6e5` here deletes a test that pinned an upstream product default, and that deletion is the only place the reversal is visible. Flagging for a decision.**

`RefreshAdoptsNewColumns.test.ts` was added upstream by `92af88bd3a`, which set `autoEnableNewObjects` and `autoEnableNewColumns` to **`true`** on `next`. The parent branch in this chain (`mjc/catalog-3-server`, #4415) flips both to **`false`** in `dcd9ae2299` and leaves that test in place, so **#4415's suite is red on it** — the source there reads `IntegrationDiscoveryResolver.ts:6133 defaultValue: false` while the test greps the same line for `defaultValue:\s*true`. This PR resolves that by retiring the test.

Read end to end the chain is coherent. Reviewed one PR at a time it is not: **#4415 shows a failing test, this PR shows a test deletion, and neither shows that a product default was reversed.** No single diff contains both halves of the decision.

Two consequences worth surfacing:

- **The behaviour ends up pinned in neither direction.** The upstream test is gone and nothing replaces it, so a future change in either direction passes silently.
- **The dialect patch sets already disagree in production** — SQL Server carries `true` (upstream-faithful), Postgres carries `false`. A newly-appeared column is adopted on a SQL Server tenant and left inactive on a Postgres one, and no register row recorded that until now.

*Should a schema refresh adopt newly appeared objects and columns by default?* is a product decision, so I have not picked one. If the answer is that `false` is right, this PR is the natural place to add a test pinning **our** intent in place of the one it removes — a default worth reversing is worth a pin.

Tracked as register **MJ-RUN-39**, with the full evidence on #4415.

PR #4417, opened 2026-09-13
feat(integration): the GraphQL contract — detached start, live counts, paging, active operations

MS-BC:
## Summary
Five client-facing additions to the integration GraphQL contract, grouped for one schema review:

1. **`IntegrationStartSchemaRefresh`** — an explicit, detached discovery start returning a tailable `RunID`, taking the maintenance lock before returning so a second click is refused rather than coalesced. Until now the only client-facing path was synchronous, and pages auto-triggered discovery on entry.
2. **Live sync counters** — `RecordsCreated/Updated/Errored` were dead (never incremented on the live snapshot); they now count, and `RecordsSkipped` is added.
3. **History paging** — `IntegrationGetSyncHistory` gains offset paging with `TotalKnown`/`HasMore`; `IntegrationListRuns` authorizes-then-pages with a bounded refill; per-entity counts are recovered from the run artifact into the never-populated `EntityDetails`.
4. **`IntegrationGetActiveOperations`** — one query for sync, RSU, discovery and the maintenance lock, with RSU labelled process-global; also restores live sync progress from the durable snapshot.
5. **`IntegrationProbeCredentials`** — tests credentials that have not been saved anywhere (transient `CompanyIntegration` built with `NewRecord()` and never saved). Nothing persisted, values never logged, hard deadline, rate-limited per (user, integration). Turns every setup wizard from create-then-test into test-then-create.

## Verification
`DetachedDiscoveryStart`, `LiveSyncCounts`, `EntityOutcomes`, `SyncLoggerForwardsSplit`, `ActiveOperations`, `LiveProgressFallback` suites. The probe has been serving MJ Central's connector wizard since 2026-09-10.

## Stack
Base `mjc/apply-parity`. Next: `mjc/sync-lifecycle`.


MS-BC:
**Heads-up before this merges: its base no longer contains #4414's tip, and this PR edits the same two files #4414 just rewrote.**

`mjc/rsu-observability` (#4414) gained commit `c76052d97d`, which **rewrote** `IntegrationProgressReader.ts` — deleting `readProgressTail`, `countLines`, `aggregateCountsFromTail`, `aggregateWarningsFromTail` and `runMtime`, so that a run listing costs at most one journal read instead of four per run — and changed `IntegrationProgressEmitter.ts`. That commit is **not** an ancestor of this branch.

This PR edits both of those files and adds `offset` paging over the same hydration path plus a fifth journal read. Merged from its current base after #4414, the risk is not just a conflict: it could **reintroduce the deleted helpers**, silently restoring the four-reads-per-run behaviour that #4414's change exists to remove. That would present as a performance regression with a green diff.

Full analysis and the options are on #4414. Nothing has been force-pushed — the merge order and any rebase are the maintainer's call.

PR #4421, opened 2026-09-13
fix(codegen): guard the AI entity name, propagate a failed INSERT, and keep advanced generation off in-process

MS-BC:
## Summary
Three guards on CodeGen's new-entity path, and a policy for in-process runs. Together they are why a 27-table connector came out with sixteen tables while every surface reported the schema update complete.

## What

**An AI-generated entity name is checked before it is used.** A model answered the entity-name prompt with `-1`. Nothing between the prompt and the INSERT questioned it: the name became `-1`, the INSERT failed on the metadata constraints, and the failure was logged and swallowed. `isPlausibleEntityName` rejects a non-string, an empty or over-long candidate, a bare non-answer token (`null`, `n/a`, `unknown`, `entityName`), and anything without a run of two letters. It is applied inside `generateEntityName` and again at the call site, because a subclass or a stub can return anything.

**A failed `createNewEntity` rethrows.** Swallowing it let CodeGen commit a *partial* set of new entities and report success — the table whose INSERT failed was simply absent afterwards, with a log line as the only witness. The transaction in `createNewEntities` exists to give all of it or none of it, and it can only do that if the failure reaches it. A table that does not *qualify* (no primary key) is still a skip, not a failure; that path returns earlier.

**A rolled-back batch no longer poisons the next run.** The in-run list of new entity names is process-static and in-process CodeGen runs many times per process, so names pushed by a batch that then rolled back survived and every one of them read as "taken" on the next attempt, earning a schema suffix.

**In-process CodeGen runs with advanced generation off unless `RSU_CODEGEN_ADVANCED_GENERATION=1`.** In-process is the runtime schema-update path, with a user watching a progress screen. The CLI's full AI profile puts several LLM round trips behind every new entity and field, so the step's duration becomes the provider's failover behaviour rather than the schema's size — one 27-table connector spent hours in it at ~1% CPU with ten sockets open to a model provider. The CLI's own `Run()` reads the same config untouched, and the previous value is restored in a `finally`.

## Verification
`entity-name-guard`, `in-process-advanced-generation-policy` and `new-entity-failure-propagates` suites; package suite 1287 pass, 0 fail.

## Stack
Independent of the per-connection catalog stack — based directly on `next`.

### DIFF
--- a/.changeset/codegen-in-process-guards.md
+++ b/.changeset/codegen-in-process-guards.md
@@ -1,0 +1,5 @@
+---
+"@memberjunction/codegen-lib": patch
+---
+
+An AI-generated entity name is validated before it is used, a failed entity INSERT propagates instead of being swallowed (so the surrounding transaction can roll back rather than commit a partial set of new entities), a rolled-back batch no longer leaves its names in the process-static new-entity list, and in-process CodeGen runs with advanced generation off unless `RSU_CODEGEN_ADVANCED_GENERATION=1`.
--- a/packages/CodeGenLib/src/Config/config.ts
+++ b/packages/CodeGenLib/src/Config/config.ts
@@ -1485,3 +1485,43 @@ export function mj_core_schema(): string {
 export function dbPlatform(): DatabasePlatform {
   return configInfo.dbPlatform;
 }
+
+/**
+ * Environment switch that keeps advanced (AI) generation ON for in-process CodeGen runs. Absent, or any
+ * value other than '1', means an in-process run turns it off for its duration.
+ */
+export const IN_PROCESS_ADVANCED_GENERATION_ENV = 'RSU_CODEGEN_ADVANCED_GENERATION';
+
+/**
+ * Applies the in-process CodeGen policy for advanced generation to `config` and returns a function that
+ * puts the previous value back.
+ *
+ * In-process CodeGen is the runtime schema-update path: a connector's tables are created while a
+ * customer watches a progress screen. The CLI's full AI profile is the wrong thing to run there. Every
+ * new entity and field goes through several LLM round trips, so the step's duration becomes the LLM
+ * provider's failover behaviour rather than the schema's size (one 27-table connector spent hours in
+ * it), and a model that answers the name prompt with `-1` puts the whole table at risk. So an in-process
+ * run disables advanced generation unless the operator opts back in with
+ * RSU_CODEGEN_ADVANCED_GENERATION=1. Table-derived names and descriptions are what the runtime path
+ * produces; the AI profile stays available to the CLI, which reads the same config untouched.
+ */
+export function applyInProcessAdvancedGenerationPolicy(
+  config: ConfigInfo,
+  env: NodeJS.ProcessEnv = process.env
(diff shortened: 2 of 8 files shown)


rkihm-BC (approved):
## Approving. Three guards that each close a way CodeGen could report success while losing a table.

The framing in the summary is what makes this reviewable — "a 27-table connector came out with sixteen tables while every surface reported the schema update complete" is the failure, and each change is traceable to a specific link in that chain.

### What I verified

**The rethrow is the important one, and its precondition holds.** Turning a swallowed error into a thrown one is the change most likely to over-fire, so I checked the claim it rests on: that a table which doesn't *qualify* is still a skip rather than a failure. It is — the "doesn't qualify to be created" path `return`s from inside the `try`, before the `catch`, so a table without a primary key never reaches the new `throw`. Only a genuine INSERT failure propagates, which is exactly what the surrounding transaction needs in order to mean anything. A transaction that can't see its own failures isn't giving you "all of it or none of it"; it was giving you "whatever committed."

**`isPlausibleEntityName` rejects the right things and nothing else.** Non-string, empty, over-255, a bare non-answer sentinel, or no run of two letters. The sentinel set is matched against the whole trimmed string rather than as a substring, so `Nation` isn't caught by `na` and `Tokenizer` isn't caught by anything — a detail that would have been easy to get wrong and annoying to diagnose. And rejection falls back to the table-derived name rather than failing the run, so the guard can only ever cost you a nicer name.

Applying it both inside `generateEntityName` and again at the call site is right for the reason given: the call site is the one that hands the value to the INSERT, and a subclass or stub can return anything.

**The rolled-back-batch fix is a real bug, not defensive coding.** `newEntityList` being process-static while in-process CodeGen runs many times per process means a rolled-back batch's names survive as "taken" and every one of them earns a schema suffix on the retry — so the *second* attempt fails differently from the first, which is the worst kind of bug to debug. Truncating to the pre-transaction length is the correct scope; nothing else pushes to that list inside the transaction.

**The in-process policy restores correctly.** It no-ops unless `enableAdvancedGeneration` is actually `true`, so `restore` only ever puts back a value that was there. The CLI's `Run()` reads the same config untouched, and the restore is in a `finally`.

Making in-process default to off is the right default regardless of the performance argument: the runtime schema-update path has a user watching a progress bar, and "the step's duration becomes the provider's failover behaviour rather than the schema's size" is a good way to put why an LLM round trip per entity and field doesn't belong there.

### One observation — not a change request

`applyInProcessAdvancedGenerationPolicy` mutates the shared `configInfo` for the duration of the run. If two in-process CodeGen runs ever overlapped inside one process, the second would find it already disabled, no-op its own policy, and then have advanced generation switched back on underneath it by the first run's `restore` — losing the policy for the rest of its run and reintroducing exactly the hours-long step this is preventing.

That needs genuine concurrency in a single process, which I don't believe happens today, so I'd leave it. Worth knowing rather than fixing — and if in-process runs ever do become concurrent, a depth counter instead of a boolean is the whole fix.

Changeset present at `patch`, correctly — no migration ships. CI 21 pass / 0 fail, and 288 lines of tests against roughly 110 lines of source, including one that pins the failure actually propagating.

PR #4422, opened 2026-09-13
fix(open-app): resolve an app's version tag from the tag that exists, not from its folder

MS-BC:
## Summary
An app in a multi-app repository cannot be version-resolved, upgraded or pinned — and if it declares migrations, it cannot be installed either.

## Why
`ScopedTagPrefix` builds the version tag from the app's **repo subpath**, so an app at `Events/PheedLoop` is looked up as `^Events-PheedLoop@<semver>$`. A repository that tags by npm package name tags it `@memberjunction/connector-pheedloop@1.4.6`. The two conventions never intersect: zero tags match the folder form, and every one of that app's six tags is package-scoped. `GetLatestVersion` therefore returns null and an upgrade fails with "Could not determine target version", which is at least honest.

`ResolveRef` has the same fault in the other direction, which is why passing the version was not a workaround either: an explicit `--version 1.4.6` composes the same ref that does not exist.

**Three consumers, not two.** `DownloadMigrations` is the third, and it is the one that ALWAYS has a version: `manifest.version` is required by the manifest schema, so unlike the manifest fetch it never falls back to HEAD. For a package-tagged app **with migrations** the composed folder ref 404s on every install and upgrade, failing the Migration phase with `Failed to download migrations: Not Found`; `mj app remove` fails the same way on the teardown download. So installs were never actually spared — only installs of an app without migrations were.

The PostgreSQL `-pg` probe is worse than loud. It reads a 404 as "no PG variant here", by design and by comment, so a ref that does not exist is indistinguishable from a repo that ships no PG set — that app's PG migrations were skipped **silently** on that line before the declared directory failed on the next.

## What
`ScopedTagCandidates` offers the app's own declared `Name` beside the folder form, the tag pattern tolerates an optional `@scope/`, and a new `FindVersionTagName` resolves the tag that **exists** rather than composing one. **All three `ResolveRef` consumers now change together** — `GetLatestVersion`/`ValidateGitHubTag`, `FetchManifestFromGitHub`, and `DownloadMigrations` — each reusing the same `FindVersionTagName` with a `?? ResolveRef(...)` fallback, so single-app repos and repos that really do tag by folder are untouched. The three `DownloadMigrations` call sites in the orchestrator pass `manifest.name`, which each already has in hand.

`InstallApp`'s explicit-version manifest fetch is deliberately left alone: the app name is genuinely unknown before the manifest is read. Every `DownloadMigrations` call site runs after the manifest is read, so none has that excuse.

Regression tests pin the old folder form returning null (version lookup) and resolving nothing (migration download), and the migration pin asserts the **exact ref string** rather than "it succeeded" — the original bug composed a ref that was well-formed and absent. The tag-listing call count is asserted too, so a warm tag memo cannot make a pin pass without issuing a fetch.

## Verification
`github-client` + orchestrator suites; package suite **394 pass, 0 fail** (389 before). Six mutations, each reverting one load-bearing line, all caught.

## Stack
Independent of the per-connection catalog stack — based directly on `next`.

### DIFF
--- a/.changeset/open-app-tag-resolution.md
+++ b/.changeset/open-app-tag-resolution.md
@@ -1,0 +1,5 @@
+---
+"@memberjunction/open-app-engine": patch
+---
+
+A monorepo app's version tag is resolved from the tag that exists rather than composed from the app's repo subpath, so an app whose repository tags by package name can be version-resolved, upgraded and pinned instead of only installed at HEAD. Both the tag lookup and the ref resolution change together.
--- a/packages/OpenApp/Engine/src/__tests__/github-client.test.ts
+++ b/packages/OpenApp/Engine/src/__tests__/github-client.test.ts
@@ -54,6 +54,7 @@ import {
     FetchManifestFromGitHub,
     IsPrereleaseVersion,
     ClearGitHubTagCache,
+    FindVersionTagName,
 } from '../github/github-client.js';
 import type { GitHubClientOptions } from '../github/github-client.js';
 
@@ -675,5 +676,72 @@ describe('scoped (multi-app) version resolution', () => {
         mocks.getRef.mockResolvedValueOnce({ data: {} });
         await ValidateGitHubTag('https://github.com/Acme/App', '2.0.0', {});
         expect(mocks.getRef).toHaveBeenCalledWith({ owner: 'Acme', repo: 'App', ref: 'tags/v2.0.0' });
+    });
+});
+
+
+// ── The package-tagged monorepo ──────────────────────────────────────────────────────────────
+//
+// A connector lives at `Events/PheedLoop` but releases through changesets, so its tag is the
+// PACKAGE name: `@memberjunction/connector-pheedloop@1.4.6`. Deriving the tag from the folder
+// produced `Events-PheedLoop@…`, which matches nothing — so every such app resolved to no version
+// and `mj app upgrade` reported 'Could not determine target version'. Installs hid it, because a
+// version-less install resolves to HEAD and never reads a tag.
+const PHEEDLOOP_TAGS = {
+    data: [
+        { name: '@memberjunction/connector-pheedloop@1.4.6' },
+        { name: '@memberjunction/connector-pheedloop@1.4.5' },
+        { name: '@memberjunction/connector-nimble-ams@1.3.2' },
+        { name: 'not-semver' },
+    ],
+};
+
+describe('scoped tags named after the package, not the folder', () => {
+    it('finds the version when the app name is supplied', async () => {
+        mocks.listTags.mockResolvedValueOnce(PHEEDLOOP_TAGS);
+        const v = await GetLatestVersion(
+            'https://github.com/MemberJunction/Integrations', {}, 'Events/PheedLoop', 'connector-pheedloop'
+        );
+        expect(v).toBe('1.4.6');
+    });
+
+    it('is the regression: without the app name the folder form matches nothing', async () => {
+        mocks.listTags.mockResolvedValueOnce(PHEEDLOOP_TAGS);
+        const v = await GetLatestVersion(
+            'https://github.com/MemberJunction/Integrations', {}, 'Events/PheedLoop'
+        );
+        expect(v).toBeNull();
+    });
+
+    it('does not bleed across connectors sharing the repo', async () => {
+        mocks.listTags.mockResolvedValueOnce(PHEEDLOOP_TAGS);
+        const v = await GetLatestVersion(
+            'https://github.com/MemberJunction/Integrations', {}, 'AMS/NimbleAMS', 'connector-nimble-ams'
+        );
+        expect(v).toBe('1.3.2');
+    });
+
+    it('resolves the REAL tag name for a version, not one built from the folder', async () => {
+        mocks.listTags.mockResolvedValueOnce(PHEEDLOOP_TAGS);
+        const tag = await FindVersionTagName(
+            'https://github.com/MemberJunction/Integrations', {}, '1.4.6', 'Events/PheedLoop', 'connector-pheedloop'
+        );
+        expect(tag).toBe('@memberjunction/connector-pheedloop@1.4.6');
+    });
+
+    it('returns null for a version that is not tagged', async () => {
+        mocks.listTags.mockResolvedValueOnce(PHEEDLOOP_TAGS);
+        const tag = await FindVersionTagName(
+            'https://github.com/MemberJunction/Integrations', {}, '9.9.9', 'Events/PheedLoop', 'connector-pheedloop'
+        );
+        expect(tag).toBeNull();
+    });
+
+    it('still honours a repo that genuinely tags by folder', async () => {
+        mocks.listTags.mockResolvedValueOnce({ data: [{ name: 'Events-PheedLoop@2.0.0' }] });
+        const v = await GetLatestVersion(
+            'https://github.com/MemberJunction/Integrations', {}, 'Events/PheedLoop', 'connector-pheedloop'
+        );
+        expect(v).toBe('2.0.0');
     });
 });
--- a/packages/OpenApp/Engine/src/github/github-client.ts
+++ b/packages/OpenApp/Engine/src/github/github-client.ts
@@ -224,6 +224,34 @@ function ScopedTagPrefix(subpath: string | undefined): string | undefined {
 }
 
 /**
+ * Every tag prefix a subpath app might REALLY be tagged under.
+ *
+ * The directory an app lives in is not its identity, and deriving the tag name from the directory
+ * silently assumed it was. A monorepo that releases through changesets tags by PACKAGE name
+ * (`@memberjunction/connector-pheedloop@1.4.6`), not by folder (`Events-PheedLoop@1.4.6`) — so for
+ * every such app `GetLatestVersion` matched nothing, returned null, and `mj app upgrade` reported
+ * 'Could not determine target version'. Installs hid it because a version-less install resolves to
+ * HEAD and never consults a tag at all, which is why this only ever surfaced on the first upgrade.
+ *
+ * `appName` is the manifest's own `name` — the unscoped half of the published package — so the
+ * pattern below accepts an optional `@scope/` in front of it. The folder form is kept as a
+ * candidate so repos that genuinely tag that way are unaffected.
+ */
+function ScopedTagCandidates(subpath: string | undefined, appName?: string): string[] {
+    const out: string[] = [];
+    const folder = ScopedTagPrefix(subpath);
+    if (folder) out.push(folder);
+    const name = appName?.trim();
+    if (name && name !== folder) out.push(name);
+    return out;
+}
+
+/** Escapes a literal for embedding in a RegExp. */
+function EscapeForRegExp(v: string): string {
+    return v.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+}
+
+/**
  * Resolves the git ref to fetch at. With no version → 'HEAD'. With a version:
  * a subpath app uses its scoped tag `<prefix>@<version>`; a single-app repo uses `v<version>`.
  */
@@ -256,7 +284,8 @@ export async function FetchManifestFromGitHub(
     repoUrl: string,
     version: string | undefined,
     options: GitHubClientOptions,
-    subpath?: string
+    subpath?: string,
+    appName?: string
 ): Promise<ManifestFetchResult> {
     const parsed = ParseGitHubUrl(repoUrl);
     if (!parsed) {
@@ -264,7 +293,12 @@ export async function FetchManifestFromGitHub(
     }
 
     const effectiveSubpath = (subpath ?? parsed.Subpath)?.replace(/^\/+|\/+$/g, '');
-    const ref = ResolveRef(version, effectiveSubpath);
+    // Prefer the tag that EXISTS over one composed from the folder name; fall back to the composed
+    // form so single-app repos and repos that really do tag by folder are untouched.
+    const ref = version
+        ? (await FindVersionTagName(repoUrl, options, version, effectiveSubpath, appName))
+            ?? ResolveRef(version, effectiveSubpath)
+        : ResolveRef(version, effectiveSubpath);
     const manifestPath = ComposeRepoPath(effectiveSubpath, 'mj-app.json');
 
     try {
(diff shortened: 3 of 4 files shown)


rkihm-BC (changes requested):
## Requesting changes — the diagnosis is right, but as written this breaks upgrades for single-app repos.

The root cause is correctly identified and the `FindVersionTagName` approach — resolve the tag that *exists* rather than composing one — is the right shape. Changing both the lookup and `ResolveRef` together is also right, and the reasoning for why fixing only one moves the failure to a 404 is exactly correct. Two problems with the execution.

## 1 — Single-app repos can no longer resolve a version at all

`UpgradeApp` now passes `existingApp.Name`. For a repo with no subpath, that one argument flips two decisions:

- `ScopedTagCandidates(undefined, 'PheedLoop')` returns `['PheedLoop']` — non-empty — so the `length === 0` guard **skips the releases path entirely**. That was the only path single-app repos used.
- `ListGitHubTags` then takes `prefix = candidates[0]` and builds the scoped pattern `^(?:@[^/]+/)?(?:PheedLoop)@<semver>$`, which cannot match a repo-wide `v2.0.0` tag.

Both come back empty, `GetLatestVersion` returns `null`, and the upgrade fails with **"Could not determine target version"** — verbatim the error this PR exists to remove, now aimed at a different class of repo.

Transcribing the two functions from this branch and running them:

```
SINGLE-APP repo, BEFORE (no appName passed)
   candidates      = []
   releases path   = taken
   tag pattern     = /^(v?\d+\.\d+\.\d+...)$/
   tags matched    = [v2.0.0, v1.4.0, v1.0.0]
   GetLatestVersion-> resolves

SINGLE-APP repo, AFTER (UpgradeApp passes Name)
   candidates      = ['PheedLoop']
   releases path   = SKIPPED
   tag pattern     = /^(?:@[^/]+\/)?(?:PheedLoop)@(\d+\.\d+\.\d+...)$/
   tags matched    = []
   GetLatestVersion-> null   *** REGRESSION ***
```

**Why CI is green:** every existing `GetLatestVersion` test calls it as `(url, {})`. The combination the production caller actually uses — an app name with no subpath — is untested.

**The underlying mistake** is that both sites now use "do we have any tag candidate?" as the discriminator, when the real question is "is this a subpath app?" Since `appName` is essentially always present, `ScopedTagCandidates(...).length === 0` is almost never true.

Worth noting this is isolated to `GetLatestVersion`. `FetchManifestFromGitHub` and `ValidateGitHubTag` use `FindVersionTagName(...) ?? ResolveRef(...)`, and `ResolveRef` with no subpath yields `v1.2.3` — so they degrade correctly. `GetLatestVersion` is the only one using candidates to *route* rather than to look up with a fallback.

**Suggested fix** — keep the discriminator on the folder form, and let `appName` only widen the candidate list:

```ts
// GetLatestVersion — restore the original test
if (!ScopedTagPrefix(subpath ?? ParseGitHubUrl(repoUrl)?.Subpath)) { /* …releases… */ }

// ListGitHubTags — same, so a single-app repo keeps the v<semver> pattern
const prefix = ScopedTagPrefix(subpath ?? parsed.Subpath);
const pattern = prefix
    ? new RegExp(`^(?:@[^/]+/)?(?:${candidates.map(EscapeForRegExp).join('|')})@(${SEMVER_PATTERN})$`)
    : new RegExp(`^(v?${SEMVER_PATTERN})$`);
```

Both need it — fixing only the first still leaves single-app repos broken whenever a repo tags `v1.2.3` without cutting GitHub Releases.

## 2 — `mj app check-updates` was not updated, so the original bug survives on the more visible path

`packages/MJCLI/src/commands/app/check-updates.ts:41`:

```ts
const report = await CheckAppsForUpdates(apps, (repoUrl, subpath) =>
    GetLatestVersion(repoUrl, githubOptions, subpath)     // no appName
);
```

So for a package-tagged app, `check-updates` still composes the folder form, still matches nothing, and still reports the app as up to date — forever. That is the surface where a user *learns* an upgrade exists, so fixing `upgrade` while leaving this alone means the fix is invisible in the normal workflow: nothing tells you to run the command that now works.

The comment immediately above that call is about a previous bug whose symptom was "every private repo whose token lives there reported 'up to date' forever" — the same failure mode, from a different cause. Worth threading the app name through `CheckAppsForUpdates`'s callback so this one doesn't reappear.

## Tests worth adding

- `GetLatestVersion` with an `appName` and **no** subpath, against `v<semver>` tags — the case that would have caught #1.
- The same with releases rather than tags, since the releases path is the one being skipped.
- `check-updates` resolving a package-tagged app once #2 is threaded through.

The regression test you added for the folder form returning null is a good pin and should stay; it just needs a sibling on the single-app side.

Happy to re-review quickly once these are in — the core approach here is sound and I'd like to see it land.

MS-BC:
Folded in the third `ResolveRef` consumer. This PR fixed two of three; `DownloadMigrations` still composed the ref from the folder name.

**The failure mode, measured — loud at two sites, silent at a third.** Reproduced against the mocked Octokit boundary, serving the tree only at the tag the repo really has:

```
PROBE_RESULT = {"Success":false,"ErrorMessage":"Failed to download migrations: Not Found - https://docs.github.com/rest"}
PROBE_REFS   = ["Events-PheedLoop@1.4.6"]
PROBE_LISTTAGS_CALLS = 0
```

- `HandleMigrations` (install `:363`, upgrade `:833`) → **loud**. `if (!downloadResult.Success) return { Success: false, ... }`, so the Migration phase fails with `Failed to download migrations: Not Found`.
- `HandleTeardown` (`:1754`) → **loud**. The download failure lands in `removalErrors`, so `mj app remove` aborts before touching the filesystem.
- The PostgreSQL `-pg` probe (`:1813`) → **silent, by design and by comment**: "a 404 yields `Success:false` … both mean 'no PG variant here'". A ref that does not exist is indistinguishable from a repo that ships no PG set, so a package-tagged app's PG migrations were skipped with no warning. It happened to be self-neutralising only because the declared directory on the next line used the same broken ref and failed loudly. Fix one and not the other and this becomes a real silent wrong-set install.
- `ExtractDeclaredApplicationIds` (`:2304`, same `DownloadAppMigrations`) → **silent**: `if (!download.Success) return []`. During a failed-install rollback that means the app's declared Application rows are never discovered and never retired, with nothing logged.

Worth correcting one line in the original description: installs were **not** spared. `manifest.version` is required by the manifest schema, so unlike `FetchManifestFromGitHub` this call never receives `undefined` and never falls back to HEAD. A package-tagged monorepo app **with migrations** could not be installed at all — only an app without migrations was unaffected.

**What changed**

- `packages/OpenApp/Engine/src/github/github-client.ts:418` — `appName?: string` added.
- `packages/OpenApp/Engine/src/github/github-client.ts:434-436` — the same `FindVersionTagName(...) ?? ResolveRef(...)` shape as `:300-301` and `:797`. No parallel mechanism.
- `packages/OpenApp/Engine/src/install/install-orchestrator.ts:1754, :1813, :1827` — all three pass `manifest.name`.

**No call site legitimately lacks the name.** `InstallApp`'s explicit-version manifest fetch is correctly left alone — the name is unknown before the manifest is read. That reasoning does not transfer: all three `DownloadMigrations` call sites run after the manifest is parsed and hold it as a parameter.

**Mutation table** — each mutation applied on disk, hash change plus old-gone/new-present verified against an expected occurrence count, ANSI stripped before parsing, restored byte-for-byte with the suite re-confirmed green after each.

| # | Mutation | Result | Test that caught it |
|---|---|---|---|
| M1 | `DownloadMigrations` back to the bare `ResolveRef` (the exact original line) | RED 393/1 | `DownloadMigrations resolves the tag that exists > requests the package-scoped tag…` |
| M2 | drop `manifest.name` at the declared-directory call site | RED 393/1 | `HandleMigrations > downloads migrations/ on SQL Server` |
| M3 | drop `manifest.name` at the PG-probe call site | RED 393/1 | `HandleMigrations > downloads migrations-pg/ on PostgreSQL` |
| M4 | drop `manifest.name` at the teardown call site | RED 392/2 | both `HandleTeardown` download assertions |
| M5 | drop the `(?:@[^/]+/)?` npm-scope tolerance | RED 392/2 | the new pin + `resolves the REAL tag name for a version` |
| M6 | `ScopedTagCandidates` stops offering the app name | RED 390/4 | the new pin + 3 existing scoped-tag tests |

**No greens to classify — all six were caught.** M5 is the one that would have gone green under the blind spot worth flagging: a fixture minting *both* tag forms lets the folder candidate resolve on its own, so dropping scope tolerance survives. This fixture mints scoped tags only, which is why M5 is red here.

**The memo trap.** `tagListCache` holds the in-flight promise for 60s keyed by repo + token, and every test in this file uses the same repo and `{}` options — so a later test can "prove" a fetch path while issuing zero HTTP calls. The file's `beforeEach` already calls `ClearGitHubTagCache()`; the new pins additionally assert `listTags` call counts (`toHaveBeenCalledTimes(1)` where a listing is expected, `not.toHaveBeenCalled()` on the single-app and HEAD paths), so a warm memo shows up as a failure rather than as a silent pass.

**Pins added** (`github-client.test.ts`, +5):

1. the package-scoped tag is requested, asserted as the **exact string** `@memberjunction/connector-pheedloop@1.4.6` — every other ref 404s as GitHub does, so a test cannot pass by composing a ref;
2. regression: the folder form **alone** resolves nothing (`Success: false`, ref `Events-PheedLoop@1.4.6`);
3. a single-app repo still uses `v2.0.0` and issues no tag listing;
4. a repo that genuinely tags by folder still falls back to the composed ref;
5. no version → `HEAD`, no tag read at all.

The four existing orchestrator assertions that pinned the 6-argument call shape were updated to the 7th argument, which is what makes M2–M4 red.

**Suite:** 389 pass / 0 fail before → **394 pass / 0 fail** after, 29 files. `pnpm --filter "@memberjunction/open-app-engine^..." build` then the package build both clean (tsconfig excludes tests, so the build is the real typecheck). No `any`. No lockfile churn. Changeset text extended rather than a second changeset added, same package and same patch bump.

Still a draft.

rkihm-BC (changes requested):
## Still requesting changes — the two items from my last review were not touched

Re-reviewed at `e302db034c`. The follow-up commit folds `DownloadMigrations` into the resolver, which is a good catch and is verified below. Neither item from my 9/13 review was addressed, and the reply does not mention them.

### 1 — Single-app repos with an app name still resolve to `null`

Executed at head against the mocked Octokit boundary, with the cache cleared between runs:

| Call | Tags or releases | Result |
|---|---|---|
| `GetLatestVersion(repo, {}, undefined, 'PheedLoop')` | tags `v2.0.0, v1.4.0, v1.0.0` | `null`; `listReleases` never called |
| same | release `v2.0.0`, no tags | `null`; `listReleases` never called |
| control, no `appName` | same | `'2.0.0'` |
| `ListGitHubTags(repo, {}, undefined, 'PheedLoop')` | same | `[]` |

Mechanism unchanged: `github-client.ts:501` routes past releases when `ScopedTagCandidates(...)` is empty, and `:620-621` builds the pattern from `candidates[0]`. Production hits it because `UpgradeApp` passes `existingApp.Name` (`install-orchestrator.ts:708`), which is always set from the manifest name. So every upgrade of a single-app repo that tags `vX.Y.Z` or cuts Releases fails with "Could not determine target version". The two-line fix I suggested last time still applies: discriminate on the scoped prefix and let `appName` only widen the candidates.

### 2 — `mj app check-updates` still composes the folder ref

`check-updates.ts:41` still calls `GetLatestVersion(repoUrl, githubOptions, subpath)` with no name, and `LatestVersionLookup` in `update-check.ts:19` has no name slot even though `UpdateCheckApp` carries `Name`. One correction to my earlier wording: a package-tagged app lands in the **Unresolved** bucket, not "up to date", so the failure is visible. The reason string still names a tag that does not exist and the available upgrade is invisible. The PR body does not scope this out.

### Verified and fine

- `DownloadMigrations` (`github-client.ts:436`) now resolves through the same path with the folder-ref fallback; the three orchestrator call sites pass `manifest.name`.
- Resolver edge cases: exact and `v`-prefixed forms found; case-sensitive miss falls back to the composed ref and 404s as designed; superstring names do not match; when both forms exist, listing order wins; `octokit.paginate` at `:785` walks three pages for a 250-tag repo.
- Red-then-green: `github-client.ts` at `next` fails 5 of 65; restored 65/65. Package suite 394 passed. Build and `tsc --noEmit` clean. Changeset present at `patch`. No schema or metadata, so the publish policy is satisfied. Merge-tree clean, 70 behind `next`.
- The CodeQL failure is three pre-existing `polynomial-redos` alerts (`next` alerts 259, 260, 261) re-attributed by the line shift from the inserted helper. Not this PR's, though cheap to clear while here.

### Follow-ups, not for this PR

- `ResolveDependencyVersion` (`install-orchestrator.ts:1572`) calls `ListGitHubTags(repoUrl, options)` with neither subpath nor name, so a package-tagged dependency fails with "no published versions found". Pre-existing, and the subsequent `InstallApp` with an explicit version goes through the path you deliberately left alone, so fixing this link alone does not complete the chain. `dep.AppName` is the right value when you do.
- The MJCLI package will need its own changeset entry once item 2 lands.

rkihm-BC (changes requested):
Still requesting changes. This is the third review, and the item from 9/13 and 9/15 is still present — so rather than restate it, here is the exact line that carries it and the smallest change that would close it.

## Single-app repos still resolve zero versions when an app name is passed

```ts
const pattern = prefix
    ? new RegExp(`^(?:@[^/]+/)?(?:${candidates.map(EscapeForRegExp).join('|')})@(${SEMVER_PATTERN})$`)
    : new RegExp(`^(v?${SEMVER_PATTERN})$`);
```

This is a **binary choice on `prefix`, not a union**. And `prefix` is `candidates[0]`, where:

```ts
function ScopedTagCandidates(subpath, appName) {
    const out = [];
    const folder = ScopedTagPrefix(subpath);
    if (folder) out.push(folder);
    const name = appName?.trim();
    if (name && name !== folder) out.push(name);
    return out;
}
```

For a single-app repo, `folder` is falsy and `appName` is pushed — so `ScopedTagCandidates(undefined, 'PheedLoop')` returns `['PheedLoop']`, `prefix` is truthy, and **only the scoped pattern is tried**. A repo tagged `v2.0.0` / `v1.4.0` matches nothing.

The comment two lines above states the intended behaviour exactly:

> Single-app repo: match repo-wide `v<semver>` tags as before.

That is what does not happen once a caller supplies an app name — and `UpgradeApp` supplies one. The scoped candidates were meant to be *added* to the repo-wide form, not to replace it.

**The smallest fix is to make the pattern a union rather than an either/or** — try the scoped candidates *and* plain `v<semver>` — so adding an app name can only ever match more tags, never fewer. That also removes the failure mode by construction rather than by case analysis, which matters here because the two conventions are distinguished by repository layout the code cannot see.

Worth adding a regression pin in the shape of the executed table from my 9/15 review: `GetLatestVersion(repo, {}, undefined, '<AppName>')` against tags `v2.0.0, v1.4.0` returning `2.0.0`, not `null`.

## Everything else is right, and verified

- **The `FindVersionTagName` approach** — resolve the tag that *exists* rather than composing one — is the correct shape, and the reasoning for why fixing only the lookup moves the failure to a 404 is exactly right.
- **`DownloadMigrations` as the third consumer** is a genuine catch and is wired in: three call sites plus the definition, each with the `?? ResolveRef(...)` fallback so single-app repos and folder-tagged repos are untouched.
- **Asserting the exact ref string** rather than "it succeeded" is the right pin, because the original bug composed a ref that was well-formed and absent — a success-shaped assertion would have passed over it.
- **Leaving `InstallApp`'s explicit-version manifest fetch alone** is correctly argued: the app name genuinely is unknown before the manifest is read, and every `DownloadMigrations` call site runs after it, so none has that excuse.
- **The `-pg` probe observation** — a 404 read as "no PG variant here", making an absent ref indistinguishable from a repo that ships no PG set — is the sharpest thing in the description. Silent skipping is worse than the loud failure beside it.

The diagnosis and the architecture have been right since the first review. It is one predicate standing between this and an approval.

PR #4430, opened 2026-09-13
fix(explorer,data-provider): four surfaces stop reporting state the server never sent

MS-BC:
One reviewer's question — *does this surface report what the server actually returned?* — with four answers that were all no. Each fix is a few lines.

Surface 4 was added after the first three. It is the same bug class from the other end: a path that is correct on a small, `ID`-keyed tenant and wrong — **silently** — on a soft-keyed one. It has no upstream motion; see the note at the bottom about it being recorded as already fixed when it is not.

> **A fifth surface was folded in and then withdrawn.** The entity-search fan-out bound (MJ-UI-1) was implemented here and has been **reverted to `origin/next` byte-for-byte** (`0216967118`). It is implemented on the sibling branch `mjc/explorer-stops-misreporting` as a strict superset — a worker pool rather than a lock-step batch loop, **plus** the half this PR's version did not fix: making a timed-out entity distinguishable from one with no matches. Four of the five tests written for it **stay here**, as an implementation-agnostic pin on the fan-out's order and completeness contract; the fifth, which could only assert the bound itself, was dropped. Reasoning in the revert commit and in the comment below. **No `SearchEngine` source change remains in this PR.**

## 1. An alive agent run is painted as a failure

A run in `AwaitingFeedback` is **alive and waiting on a human**. The poll set was `Status IN ('Running', 'Paused')`, so such a run dropped out of the set and was then rendered as "❌ failed — Unknown error". The user is told their question failed, when in fact the agent asked *them* a question and is holding it.

`AwaitingFeedback` appears nowhere in the conversations package on `next` — zero hits before this change.

- The poll set now includes it.
- `agentFailureDisposition` gains a third outcome for a run awaiting input, checked **first** — the run's own status outranks a guess made from error text.
- The UI writes `Complete` with the agent's question, passes `result` through so an attached response form still renders, sets no `.Error`, and starts no in-flight watch.

`next` has refactored since this fix was first written downstream, and made it **smaller**: all eight failure call sites now funnel through one `applyAgentFailureToDetail`, so this is one branch there plus the arbiter in CorePlus, rather than four near-identical branches.

The sub-agent retry guard also had to move from `=== 'In-Progress'` to `!== 'Error'`. Otherwise an awaiting run is auto-retried, which throws away the very question it is waiting on.

## 2. An app with more than 1000 entities silently shows 1000

One `RunView` on application entities with no `MaxRows` and no `IgnoreMaxRows`. The default cap truncates at 1000, and the missing entities are not merely unlisted — they are **unsearchable** from that surface, with nothing saying so.

## 3. An unwarmed template cache crashes the form that fixes it

The cached-collection getters read straight through to the metadata object, so an unwarmed cache throws — and it throws in the very template form the error message points the user at, taking the remedy down with the fault.

**All five getters of that shape are guarded, not just the one observed crashing** — `Templates`, `TemplateContentTypes`, `TemplateCategories`, `TemplateContents`, `TemplateParams`. They are the same shape over the same object, so any is reachable on the same unwarmed engine. Shipping the one-field version would have been a knowingly incomplete fix.

An empty array is the honest answer to "what is cached?" when nothing is, and every consumer already handles an empty collection. Consequence worth knowing: `FindTemplate` now returns `undefined` rather than throwing on an unwarmed engine.

## 4. Opening a record crashes on any entity whose primary key is not called `ID`

`GraphQLDataProvider.Load` turned a `CompositeKey`'s `FieldName` into an `EntityField` with an unguarded `entity.Fields.find(...).EntityFieldInfo`. A caller that built the key from a **hardcoded** column name — `CompositeKey.FromID(id)` invents `'ID'` — hands over a name a soft-keyed entity does not have. `find` returns `undefined`, and the user gets:

```
Cannot read properties of undefined (reading 'EntityFieldInfo')
```

Nothing in that says which entity or which key. Every MJ core entity happens to be `ID`-keyed, so this is invisible across the whole core product and fires on the first entity mapped from an external schema — a NetForum `*_key` table opens, and this is what the user sees.

Both call sites of that lookup — `Load` and `Delete` — now go through one `ResolvePrimaryKeyField`:

- match by name (case- and whitespace-insensitive), as before;
- failing that, when the key names exactly **one** column and the entity has exactly **one** primary key, the caller's single name can only mean "the primary key", so the entity's own key is used regardless of what it was called;
- anything else — a composite key, or a name that is genuinely not a key column — throws an `Error` naming the **entity**, the **key it was handed**, and the entity's **real key columns**. A composite key is never quietly truncated to its first column to make the call succeed.

**The `Delete` site is the same lookup but is not reachable the same way**, and that is worth being explicit about rather than overclaiming. Its key names come from `entity.PrimaryKey`, which is built from the entity's own metadata, so the direct match cannot miss there. It is folded in so the two sites cannot drift and so neither can dereference `undefined` — not because it is a live second crash. The mutation that reverts that site alone therefore stays **green**, deliberately.

The new `FirstPrimaryKey` use carries its mandatory `// first-pk-ok:` reason. `MJCore`'s `PrimaryKeyCompliance` gate passes, verified live by deleting the marker and watching the gate fail rather than by assuming it scans this file.

## Tests

Behavioural, not source-pins.

The poll-set test drives the service through a fake provider that **honours** the status filter it is sent. A fake returning every row regardless would have pinned nothing — mutation testing caught exactly that and it was fixed. The getter test sweeps the prototype for accessors **dynamically** and asserts none throws, so a sixth accessor is covered the day it lands.

For surface 4, the soft-key test asserts on the **exact GraphQL document and variables** the provider would send — `$cst_key: String!` and `cst_key: $cst_key`, and explicitly *not* `$ID:` — over a real `EntityInfo`/`BaseEntity` built from a `cst_key`-keyed fixture. The named-error tests assert the entity name, the offending key name and the real key columns are all present in the message, and that what is raised is an `Error` and not a `TypeError`.

The four fan-out tests that stay here pin the *contract*, not the scheduling: every scoped entity is searched, and results come back in **entity** order rather than **completion** order. The mock's per-entity delays **descend** with index, so an implementation that appends results as they settle produces a visibly different order — a fake that settled in request order would have pinned nothing. They are not vacuous against `next`'s own unbounded `Promise.all` either: appending as each settles, dropping one entity, and serializing the fan-out completely each turn them red, verified after the revert.

17 mutations for the first three surfaces, all as expected: dropping a status from the poll set or the awaiting set, deleting or reordering the disposition check, making the UI branch unreachable, writing `Error` instead of `Complete`, dropping `result`, removing `IgnoreMaxRows` or swapping it for `MaxRows: 1000`, un-guarding each of the five getters in turn, and returning `undefined` instead of `[]`. One deliberate control mutation stays green.

For surface 4: fallback removed, the multi-column guard removed, the error message losing the entity name, the error message losing the key name, the throw replaced by the original unguarded dereference, the `Load` site reverted to the raw `find`, and the hardened `undefined` check inverted — all red. Plus one deliberate control that stays **green**: reverting the `Delete` site alone, for the reason given in section 4.

For the four kept fan-out tests, mutated against `next`'s own source *after* the revert: results appended as each settles, one entity dropped from the fan-out, and the fan-out fully serialized — all red.

`MJCore`'s `PrimaryKeyCompliance` gate was also proven live by deleting the `// first-pk-ok:` marker and watching it fail, rather than assuming it scans this file.

Every mutation asserted the file actually changed on disk before the suite ran, and ANSI codes were stripped before any output matching — both have produced vacuous "green" results in the past.

| package | tests |
|---|---|
| ai-core-plus | 712 passing |
| templates-base-types | 12 passing |
| ng-conversations | 1334 passing |
| ng-dashboards | 1757 passing |
| graphql-dataprovider | 450 passing (was 441) |
| search-engine | 507 passing (was 503) — tests only, no source change |

Plus `MJCore`'s `PrimaryKeyCompliance` gate, 4 passing.

All six build clean, which is the real typecheck here since the tsconfigs exclude tests.

## Known gap, stated rather than hidden

The sub-agent retry-guard change has **no behavioural pin**. It lives inside a delegation flow that cannot be reached without stubbing the agent service and the whole invocation path, so it is covered by review only. Every other change here is killed by a test.

## Same defect class, found and deliberately not fixed

`packages/AI/Agents/src/memory-manager-agent.ts` mines "instructive failed runs" with `(Success = 0 AND Status NOT IN ('Running', 'Paused'))`. An `AwaitingFeedback` run has `Success = 0` and is not in that list, so the memory manager will mine a run that merely asked a question as a corrective failure lesson. It is the same question this PR answers, but it is server-side agent memory rather than an Explorer surface — flagging it rather than widening the change.

Also noted: the status chip maps `AwaitingFeedback` to "working" rather than "waiting on you". Correct enough — it is alive, not an error — but worth a follow-up.

## Surface 4 was recorded as already fixed upstream. It is not.

Our register had it closed by `92f2ac90b3`, the primary-key sweep across 203 files, which does list `GraphQLDataProvider`. That is why it was never raised. Checked against `origin/next` rather than taken on trust:

- `packages/GraphQLDataProvider/src/graphQLDataProvider.ts` — the `Load` lookup is **unchanged and unguarded** on `next`.
- `FirstPrimaryKey` has **zero** occurrences anywhere in the `GraphQLDataProvider` package on `next`.

The sweep touched other files. This one is still live, and it is a hard crash on record-open for every non-`ID`-keyed entity.


MS-BC:
## Two more surfaces folded in — and a register correction that matters

Pushed `a9c2560e12`. The PR is now five surfaces, not three; retitled and the body's enumeration extended. Still a draft.

### What was added

**4. `GraphQLDataProvider` — opening a record crashes on any entity whose primary key is not called `ID`.**
`Load` resolved a `CompositeKey`'s `FieldName` with an unguarded `entity.Fields.find(...).EntityFieldInfo`. A caller that built the key from a hardcoded name — `CompositeKey.FromID(id)` invents `'ID'` — hands over a name a soft-keyed entity does not have, and the user gets a bare `Cannot read properties of undefined (reading 'EntityFieldInfo')` that names neither the entity nor the key. Both call sites of that lookup now go through one `ResolvePrimaryKeyField`: name match, then a single-column-key-on-a-single-PK-entity fallback, then a named `Error` carrying the entity, the key it was handed, and the entity's real key columns. A composite key is never truncated to its first column to make the call succeed.

**5. `EntitySearchProvider` — entity search returns silently empty results on a wide fan-out.**
`Promise.all(scoped.map(...))` over every scoped entity is ~117 simultaneous `LIKE '%term%'` `RunView`s against one pool on a real tenant. They queue, each blows its own `PerEntityTimeoutMS`, and the timeout wrapper resolves each to `[]`. Bounded at 8, order-preserving. The reasoning for 8 rather than another tunable is in the constant's doc comment.

### Why these belong in this PR rather than two new ones

This PR's framing is not "three unrelated Explorer files". It is one reviewable idea: **a surface reporting state the server never sent.** The first three answer that with a UI that invents a failure, a list that invents a size, and a cache that throws instead of answering.

These two are the same idea from the other end — and they share a second property with each other that makes them a matched pair rather than two strays:

- Each is **correct on the tenant the code was written against** — a small, `ID`-keyed, few-entity install — and wrong on a large, soft-keyed one. The primary-key assumption cannot fire while every entity is `ID`-keyed; the fan-out cannot exhaust a pool while there are twelve searchable entities.
- Each fails **without producing an error anyone can act on.** Surface 4's message names nothing. Surface 5 produces no message at all: every `RunView` "succeeds", the fan-out is empty, and the user simply concludes the records are not there.

Splitting them out would mean two PRs, each one small edit, each needing the same context re-explained — and would lose the thing that makes them reviewable together: the reviewer checks one invariant (does this path tell the truth about what came back?) against five call sites instead of five times against one.

There is precedent in this repo for accepting exactly this class. **#4411** (`5e987a7741`, already on `next`) fixed Semantic search 404'ing against Pinecone by addressing indexes via `vectorIndex.ExternalID?.trim() || vectorIndex.Name` — a path that works wherever the MJ display name happens to equal the provider-side name, and fails silently wherever it does not. Same shape, same package neighbourhood, already merged.

### MJ-UI-11 was believed closed upstream. It is not.

Stating this plainly because it is the reason this was never raised. Our register recorded surface 4 as **CLOSED-UPSTREAM**, fixed by `92f2ac90b3` — the primary-key sweep across 203 files, whose file list does include `GraphQLDataProvider`. Checked against `origin/next` (`099b4d9634`) rather than taken on trust:

- `packages/GraphQLDataProvider/src/graphQLDataProvider.ts` — the `Load` lookup is **unchanged and unguarded** on `next` today.
- `FirstPrimaryKey` has **zero** occurrences anywhere in the `GraphQLDataProvider` package on `next`.

The sweep touched other files in that package's vicinity, not this one. So this is a live P1 crash on record-open for every non-`ID`-keyed entity, not a re-fix of something already landed. If the sweep's author scoped it out deliberately I would rather hear that in review than keep carrying it, but the assumption that it was already handled is what is wrong.

### One deliberate non-claim

`:2272` was expected to be the identical lookup on the **save** path. It is not — it is the **`Delete`** path, and its key names come from `entity.PrimaryKey`, built from the entity's own metadata, so the direct match there cannot miss. `Save` has no lookup of this shape at all. The `Delete` site is folded into the shared helper so the two cannot drift and neither can dereference `undefined`, **not** because it is a live second crash. The mutation that reverts that site alone stays green, deliberately, and is reported as a green control rather than buried.

### Verification

| | |
|---|---|
| `graphql-dataprovider` | 450 passing (was 441) |
| `search-engine` | 508 passing (was 503) |
| `MJCore` `PrimaryKeyCompliance` | 4 passing |
| builds | both packages clean (the real typecheck — tsconfigs exclude tests) |

17 mutations across these two changes, all as expected. Each one asserted the file actually changed on disk before the suite ran, and stripped ANSI before matching output — both of those have produced vacuous "green" results before. The mutations that must kill, and do: fallback removed, multi-column guard removed, error message loses the entity name, error message loses the key name, throw replaced by the original dereference, `Load` site reverted to the raw `find`, fan-out restored to the literal `Promise.all`, `SEARCH_CONCURRENCY` raised past the entity count, serialized to 1, batch results appended reversed, results appended as each settles, trailing partial batch dropped. The `PrimaryKeyCompliance` gate was also proven live by deleting the `// first-pk-ok:` marker and watching it fail, rather than assuming it scans this file.

Not rebased onto `099b4d9634`, because rebasing a pushed branch needs a force-push. The only upstream drift in either touched package is `VectorSearchProvider.ts` (#4411) — a sibling file, no overlap with either change.

MS-BC:
## Correction: the search fan-out change is withdrawn from this PR. Its tests stay.

Pushed `0216967118`. `packages/SearchEngine/src/generic/EntitySearchProvider.ts` is reverted to `origin/next` **byte-for-byte** — blob `26ced6df73`, confirmed by hash, not by eye. **No `SearchEngine` source change remains in this PR.** Retitled from five surfaces back to four, and the body's enumeration updated so the section describing the fan-out fix is gone rather than left describing code that is no longer here.

### Why it was withdrawn

The bound is implemented on the sibling branch `mjc/explorer-stops-misreporting` as a strict superset, and that is the version that should survive:

- **A worker pool rather than a lock-step batch loop.** The batch loop I had waited for its slowest member before starting the next group. With a 3s per-entity timeout and 117 entities that is a worst case around `117 ÷ 8 × 3s ≈ 45s` of avoidable wall clock. A pool starts the next entity the moment a slot frees.
- **It fixes the half mine did not.** `searchOneEntity` there returns `{EntityName, Items, Incomplete}`, so a timed-out entity is distinguishable from one that genuinely had no matches, with a rolled-up `PARTIAL RESULTS` log and an `OnIncompleteResults` hook. My own doc comment named that silence — "every individual piece reports success" — as the defect being fixed, and **bounding the fan-out does not fix it.** It only makes the timeout less likely. Shipping the lesser half here and the full fix there would leave two implementations of one idea, which is worse than either.

### The four kept tests, and the one dropped — straight answer

Asked directly whether the kept tests still pass against `next`'s implementation. They do, with exactly one exception, and here is the actual output rather than a prediction:

```
× runs at most 8 RunViews at a time across a 40-entity fan-out
  AssertionError: expected 40 to be less than or equal to 8
✓ still searches every entity
✓ returns results in the same order as the unbounded fan-out did
✓ a fan-out smaller than the concurrency bound behaves exactly as before
✓ an exact multiple of the bound leaves no trailing entity unsearched
Tests  1 failed | 40 passed (41)
```

**That one test is dropped, not rewritten.** It cannot be made implementation-agnostic: asserting "at most N at once" means naming N, so the pin only means anything alongside an implementation that has a bound. The sibling branch ran all five unmodified against its worker pool and all five pass there, so it is not lost — it lives where its subject does.

The other four are kept and were **renamed and re-commented** so they no longer describe a batch loop or reference a bound by name. They now state only what must hold whatever the scheduling strategy is:

- every scoped entity is actually searched;
- results come back in **entity** order, not **completion** order — the relevance sort is stable, so insertion order is what breaks ties between equally-scored hits, and any rescheduling that appends as results settle would silently reorder equal-scoring hits;
- a small fan-out is genuinely parallel rather than serialized;
- nothing is dropped at the tail.

The mock's per-entity delays deliberately **descend** with index, so later entities settle first — a fake that settled in request order would have pinned nothing.

**They are not vacuous against `next`'s own code.** Mutated after the revert, against the unbounded `Promise.all` as it stands:

| mutation | verdict |
|---|---|
| results appended as each settles, not by position | RED (3 tests) |
| one entity silently dropped from the fan-out | RED (22 tests) |
| fan-out fully serialized, no parallelism | RED (1 test) |
| restored | 40 passing |

So they guard real behaviour today and will guard the worker pool when it lands. Worth flagging for whoever merges the two: these four live in `EntitySearchProvider.test.ts`, and the sibling branch carries its own copy of all five in a separate file — expect to de-duplicate rather than to resolve a conflict.

### Unchanged

Surface 4 — the `graphQLDataProvider` primary-key resolution (MJ-UI-11) — is untouched and is the substance of this fold. One correction to how it was scoped to me, repeated here because it is in the commit message too: the second call site is `Delete`, **not** `Save`. `Save` has no lookup of this shape. `Delete`'s key names come from `entity.PrimaryKey`, built from the entity's own metadata, so the direct match there cannot miss — it is folded into the shared helper so the two sites cannot drift, not because it is a live second crash. The mutation reverting that site alone stays green, deliberately, and is reported as a green control rather than buried.

### Verification after the revert

| | |
|---|---|
| `search-engine` | 507 passing (baseline 503; 36 → 40 in `EntitySearchProvider.test.ts`) |
| `graphql-dataprovider` | 450 passing (baseline 441) |
| `MJCore` `PrimaryKeyCompliance` | 4 passing |
| builds | both packages clean |

CI on the previous head: Build, all six unit-test shards, Source guards, Native-ESM guard and the token/Sass/dependency checks all green. `Integration (SQL Server, deterministic)` failed on `waitForActiveRequest: timed out after 2000ms` → `Can't commit transaction. There is a request in progress.`, inside `mj sync push --ci`. That is the known commit race — there is a branch named `fix/4447-transaction-commit-race` for precisely this, and the same job failed in the same window on `mjc/metadata-cache-entry-cap`, `mjc/pg-fts-residue` and `fix/4309-transaction-group-refusal-reporting`. It is `GenericDatabaseProvider` server-side; neither package in this PR is on that path.

PR #4431, opened 2026-09-13
fix(sql): the dialect a prompt writes, and the connection its result runs on, both come from the provider

MS-BC:
MemberJunction assumes SQL Server wherever an agent touches SQL. Two faces of one problem, fixed together because a reviewer holding *"the ad-hoc read-only path is SQL-Server-shaped"* in their head is holding both.

## Face one — the prompts teach T-SQL unconditionally

`sql-query-writer` opened with "you are the world's greatest expert in Microsoft SQL Server and T-SQL". The shipped query-generation prompts taught `ISNULL`, `DATEADD`, `GETDATE` and `[bracketed]` identifiers, with **no dialect token anywhere** under `metadata/prompts`. A PostgreSQL tenant's agent burns LLM turns writing T-SQL, failing, retrying, and failing again.

**The mechanism is a system placeholder, not a registry of TemplateIDs.** `_SQL_DIALECT`, `_SQL_DIALECT_NAME` and `_SQL_DIALECT_RULES` resolve on every prompt render, so the question becomes a predicate — *does this template reference the placeholder?* — with nothing to register. A template added tomorrow cannot be left off a list.

That distinction is load-bearing: a prior sweep used a hardcoded list and under-applied, which is exactly how some templates stayed T-SQL-only.

`_SQL_DIALECT_RULES` is composed by `SQLDialect.PromptGuidance` from the dialect's **own primitives** — `QuoteIdentifier`, `LimitClause`, `IsNull`, `Coalesce`, `IIF`, `DateAddExpression`, `CurrentTimestampUTC`, `NewUUID`, `ConcatOperator`, `CastToText`. Implementing a dialect is therefore sufficient to get correct guidance; there is no second per-dialect prose table to drift. Only two new abstract members: `DisplayName` and `FormatterLanguage`.

Nine templates parameterised. Example:

```diff
-You are the world's greatest expert in Microsoft SQL Server and T-SQL. Your job is to write …
+You are the world's greatest expert in {{ _SQL_DIALECT_NAME }}. Your job is to write …
+
+# Target SQL Dialect
+
+{{ _SQL_DIALECT_RULES | safe }}
```

## Face two — the ad-hoc read-only path reaches for an mssql pool

`AdhocQueryResolver` required a SQL Server `ConnectionPool` from `context.dataSources` before it would run anything. A PostgreSQL provider never populates one, so **every ad-hoc query on PG** returned "No read-only data source available for ad-hoc query execution", rendered downstream as "Showing cached data" — while the `GetReadOnlyProvider` call directly beneath it already knew the tenant was PostgreSQL and used that only for a string.

Both questions — which dialect to render, and what to execute on — now come from the same read-only provider, and execution goes through `provider.ExecuteSQL`, which every platform implements.

Also fixed: the ad-hoc action resolved `{{query:"..."}}` composition macros against a hardcoded `'sqlserver'`, and `QueryBuilderAgent` re-formatted every tenant's SQL with sql-formatter's `tsql` language.

## Face three — the safety allowlists and the row cap's last resort

Same principle, three places it had not reached. Added in a follow-up commit on this branch.

**The two SQL safety allowlists were T-SQL vocabularies.** `ALLOWED_SQL_FUNCTIONS` (`@memberjunction/global`) and `ALLOWED_SQL_KEYWORDS` (`@memberjunction/core`) listed only SQL Server's read-only names. A PostgreSQL tenant writing `DATE_TRUNC('month', CreatedAt)` got **"Function 'DATE_TRUNC' is not allowed"** — refused by us, before it ever reached a database that would have run it happily. `ARRAY_AGG(Name)` failed twice over: the function allowlist, and the separate "an aggregate context must contain an aggregate" rule, which is driven by the same list.

Both lists now carry the **union** of the supported dialects' read-only vocabulary rather than switching on `ResolvePlatformKey`, and the rationale is written into the code so a reviewer does not have to ask for it. These are *deny-by-default safety screens, not dialect grammars*. A PostgreSQL name that reaches a SQL Server tenant is rejected by the SQL Server parser on its own — `DATE_TRUNC(...)` simply never parses there — so the union cannot widen the blast radius past what the target server would already accept, whereas a dialect switch here would add a second independent place for the dialect to be resolved, and so a second place for it to be wrong.

**Keying them off `ResolvePlatformKey` is the alternative** if that trade is judged the other way. The union is the smaller change and it is what is deployed on our fleet; say the word and they become dialect-keyed.

Nothing that writes, escalates, or reaches outside the statement was added, and `DANGEROUS_SQL_KEYWORDS` is still checked **first**, so a name on an allowlist can never unblock one that is denied. The tests treat that as inseparable from the widening: every safety case reuses the *same* PostgreSQL expression that now passes, so a false pass would have to come from the widening rather than from a weaker input. Semicolons, comments, stacked DDL/DML, the PostgreSQL system catalogs, and PostgreSQL functions that are **not** read-only (`PG_SLEEP`, `PG_READ_FILE`, `LO_IMPORT`, `DBLINK`, `QUERY_TO_XML`) are all still rejected.

**And `ensureRowLimit`'s `catch` arm — residual 3 below, now fixed.** The AST path already asked `resolvePlatform()`, but the safety net beneath it rewrote `SELECT` to `SELECT TOP n` unconditionally. That arm is reached *precisely when* `WrapWithMaxRows` could not understand the SQL, so on a PostgreSQL tenant a query that merely confused the parser came back as SQL that could not parse at all. It now reads the same seam as the rest of the action, appends `LIMIT n` on PostgreSQL (stripping a trailing `;` first, so the clause cannot land after the statement terminator), and treats an existing `LIMIT` as already capped.

The new tests force the `catch` arm specifically, by making `WrapWithMaxRows` throw. A previous audit retired this finding on the strength of the `try` arm being correct — that is a different code path, and the `try`-arm tests could never have caught it.

**One honest note, recorded in the code as well as here.** Widening `ALLOWED_SQL_KEYWORDS` does not by itself change what `sqlNoKeywordsExpression` accepts today. Its unknown-token gate reads `isKnownKeyword && !isAllowed`, which reduces to `isDangerous && !isAllowed` — and every dangerous keyword has already thrown in the loop above it, so an *unrecognised* token falls straight through. That is why `DATE_TRUNC` was not in fact being rejected on that path. The list is the **declared** allowlist: it is what that gate will consult the moment it is tightened, and it must carry the PostgreSQL names before that happens, not after. Its test says exactly that rather than claiming a behaviour change it does not make. MJGlobal's list, by contrast, is consulted directly and the behaviour change there is real and pinned.

## SQL Server is unchanged

Every fallback resolves to `sqlserver`, the historical default. Resolution order is the request's own provider, then `Metadata.Provider`, then `sqlserver`; an unknown platform falls back rather than throwing.

`ensureRowLimit`'s AST path was **already** dialect-aware and is untouched — only rerouted through the shared resolver, with a test pinning that it did not regress. Its `catch` arm was not, and now is; on a SQL Server tenant it still emits `TOP n`, pinned by a test that forces that arm.

## Tests

13 mutations for faces one and two, all red as expected, including one that reinstates the mssql-pool gate and one per parameterised prompt. **17 further mutations for face three**, all red as expected — including three that neuter a safety check (semicolon, comment, the dangerous-keyword throw), one that makes the function allowlist accept everything, one that hardcodes `'postgresql'` in the row-cap fallback instead of asking the provider, and one that hardcodes `'sqlserver'` in `resolvePlatform` itself. Each mutation asserts the file's hash actually changed on disk before the suite runs, and re-confirms green after restoring.

**Worth recording:** the prompt lint was **green on its first mutation** — a stray mention of `_SQL_DIALECT_NAME` elsewhere in the same file satisfied it. The lint now requires the rules block *and* forbids naming a platform, and the mutation goes red. A lint that can be satisfied by an unrelated mention is not a lint.

| package | tests | after face three |
|---|---|---|
| sql-dialect | 477 | 477 |
| global | 879 | **897** (+18) |
| core | 2582 | **2593** (+11) |
| ai-core-plus | 718 | 718 |
| core-actions | 515 | **523** (+8) |
| ai-agents | 2050 | 2050 |
| server | 1307 | 1307 |

Face three's three packages each build clean (`build` is the real typecheck), and the two test files their tsconfigs exclude were typechecked separately. Downstream consumers of the two allowlists were run too — `GenericDatabaseProvider` (1084), `MJCoreEntitiesServer` (566), and the `SQLExpressionValidator`-consuming suites in `SQLServerDataProvider` (54) and `MJServer` (59) — all green.

`tsc --noEmit` clean on all six. One server test file fails to **load** on a pre-existing self-import of MJServer's own unbuilt dist — identical before and after.

## Not verified

No live database and no LLM call. The PostgreSQL execution path runs against a stub provider implementing `ExecuteSQL`. The templates were rendered through plain nunjucks to prove they parse and carry the rules block — not through the real engine, which needs DB metadata.

**These metadata files are authoring-only.** A sync push is required before any tenant sees them.

## Known residuals, deliberately not fixed

1. **The per-request provider is almost never set.** Of 61 sites constructing prompt params, 2 assign `.provider`, so most renders fall back to `Metadata.Provider`. That is correct on a single-tenant server — the normal deployment — and wrong on a genuinely multi-tenant process, for the same reason everything else there would be. The mechanism is right; populating it is a separate piece of work and those 59 sites were not audited.
2. **The entity listing still renders `[schema].[view]`** bracket-quoted for every entity. Fixing it properly needs a dialect-aware nunjucks filter or precomputed qualified names upstream; an explicit "these brackets are a display convention" instruction is in the template meanwhile.
3. ~~**`ensureRowLimit`'s catch-block safety net still emits `TOP N` unconditionally.**~~ **Fixed in face three above.**
4. **`sqlNoKeywordsExpression`'s unknown-token gate is dead code.** `isKnownKeyword && !isAllowed` reduces to `isDangerous && !isAllowed`, and the dangerous loop above it has already thrown — so the gate cannot fire, and an unrecognised token is accepted. Tightening it is a behaviour change with its own blast radius and does not belong in a dialect PR; the allowlist is now correct for when it is tightened.
5. **The same filter's empty-parens check carries its own hardcoded T-SQL function list.** `if (upperExpression.includes('()') && !upperExpression.match(/\b(COUNT|SUM|…|DAY)\s*\(/i))` rejects `NOW()` and `RANDOM()` on a PostgreSQL tenant regardless of the allowlist. Left alone deliberately: our fleet does not run that change either, so shipping it here would diverge from what is proven in production. It is the same class of defect and is a one-line fix when wanted.
6. **The database-designer prompts still say "always SQL Server"** and map abstract types to `NVARCHAR`. That is the DDL/schema-design surface rather than query generation, so it is out of scope — and deliberately not added to the lint as a sanctioned exception, because an exception list is worse than a sentence here.


MS-BC:
Pushed `9bd2e84` — three more places where the dialect was a literal. **Folded into this PR rather than opened as its own**, and I want to be explicit about why: these are the residue of *this PR's own principle*. #4431 made the dialect come from `provider.PlatformKey`, resolved once, for prompt templates, the formatter, composition assembly and `AdhocQueryResolver`. It did not reach the two SQL allowlists or `ensureRowLimit`'s `catch` arm. Same principle, same reviewer, and the third one uses the `resolvePlatform()` seam this PR just added to that very class. A separate PR would make a reviewer hold the same idea twice, and one of them would land first and make the other's diff confusing.

This PR has zero reviews, so nothing is invalidated by the addition.

### What changed

| file | what |
|---|---|
| `packages/MJGlobal/src/SQLExpressionValidator.ts:143` | `ALLOWED_SQL_FUNCTIONS` — PostgreSQL read-only vocabulary added to all eight relevant buckets |
| `packages/MJCore/src/generic/runQuerySQLFilterImplementations.ts:87` | `ALLOWED_SQL_KEYWORDS` — same union |
| `packages/Actions/CoreActions/src/custom/data/run-adhoc-query.action.ts:268-287` | `ensureRowLimit`'s `catch` arm now asks `this.resolvePlatform()` |

Plus three test files (two new, one extended) and the changeset, which now also covers `@memberjunction/global`.

### The union, and the alternative

Both allowlists take the union **unconditionally** rather than branching on `ResolvePlatformKey`, and the justification is now a doc comment on each list so it does not live only in this thread:

> These are deny-by-default **safety** lists, not dialect grammars. A PostgreSQL name that reaches a SQL Server tenant is rejected by the SQL Server parser on its own — `DATE_TRUNC(...)` simply never parses there. The union therefore cannot widen the blast radius past what the target server would already accept, whereas a dialect switch here would add a second independent place for the dialect to be resolved, and so a second place for it to be wrong.

**If you prefer them keyed off `ResolvePlatformKey`, say so and I will do it.** The union is the smaller change and it is what is deployed on our fleet today, which is the only reason it is the default here.

Nothing that writes, escalates, or reaches outside the statement was added, and `DANGEROUS_SQL_KEYWORDS` is still checked first — so a name on an allowlist can never unblock one that is denied.

### The catch arm is not the try arm

Worth flagging because it has already fooled one audit: `ensureRowLimit`'s `try` path was **already** correct — it hands off to `QueryPagingEngine.WrapWithMaxRows(query, maxRows, this.resolvePlatform())`. The `catch` arm beneath it was not, and it is reached *precisely when* the AST path could not understand the SQL. So on a PostgreSQL tenant, a query that merely confused the parser came back as SQL that could not parse at all. A previous audit retired this finding by reading the `try` arm and stopping.

The new tests therefore force the `catch` arm by making `WrapWithMaxRows` throw. The `try`-arm tests this PR already had could never have caught it.

### One thing I will not overstate

Widening `ALLOWED_SQL_KEYWORDS` in MJCore **does not change what `sqlNoKeywordsExpression` accepts today.** Its unknown-token gate is:

```ts
const isKnownKeyword = DANGEROUS_SQL_KEYWORDS.includes(upperToken) || ALLOWED_SQL_KEYWORDS.includes(upperToken);
const isAllowed = ALLOWED_SQL_KEYWORDS.includes(upperToken);
if (isKnownKeyword && !isAllowed) { throw ... }
```

which reduces to `isDangerous && !isAllowed` — and every dangerous keyword has already thrown in the loop above it. So an *unrecognised* token falls straight through, which is why `DATE_TRUNC` was never actually being rejected on that path. The mutation run proves it: reverting the entire PostgreSQL block breaks only the declaration pin, not one behaviour test.

I have written that into the source comment and the test docblock rather than dressing the list up as a behaviour fix. It is the **declared** allowlist — what that gate will consult the moment it is tightened — and it has to carry the PostgreSQL names before that happens, not after. MJGlobal's list, by contrast, is consulted directly by `checkFunctionNames` and by the "an aggregate context must contain an aggregate" rule, so the behaviour change there is real, and `DATE_TRUNC('month', CreatedAt)` was genuinely returning *"Function 'DATE_TRUNC' is not allowed"* on every PostgreSQL tenant.

Two adjacent defects found and deliberately **not** fixed, now listed as residuals 4 and 5 in the description: that dead gate, and the same filter's empty-parens check, which carries its own hardcoded T-SQL function list and so rejects `NOW()` and `RANDOM()` on PostgreSQL no matter what the allowlist says. Both are one-liners; neither is what our fleet runs, and I would rather not diverge from what is proven in production inside a dialect PR.

### Mutations — 17, all red

Each mutation asserts the file's **hash actually changed on disk** before the suite runs (a `sed` that silently matches nothing has produced vacuous "green" results for us before), strips ANSI before parsing output, and re-confirms green after restoring.

| # | mutation | result | first test to fail |
|---|---|---|---|
| G1 | drop `DATE_TRUNC` from the `date` bucket | RED 3/18 | accepts a PostgreSQL filter using ILIKE and DATE_TRUNC |
| G2 | drop `ARRAY_AGG` from `aggregates` | RED 2/18 | counts ARRAY_AGG as an aggregate |
| G3 | drop `FILTER` from `ordering` | RED 2/18 | accepts an aggregate FILTER clause |
| G4 | neuter the semicolon check | RED 1/18 | still rejects … a semicolon |
| G5 | neuter the comment rejection | RED 2/18 | still rejects … a line comment |
| G6 | make the function allowlist accept everything | RED 1/18 | still rejects PostgreSQL functions that are NOT read-only |
| G7 | drop `TO_NUMBER` from `conversion` | RED 2/18 | accepts TO_CHAR / TO_DATE / TO_NUMBER |
| C1 | revert the **whole** PostgreSQL block from `ALLOWED_SQL_KEYWORDS` | RED 1/11 | carries every PostgreSQL read-only name — *declaration pin only; see above* |
| C2 | neuter the semicolon check | RED 1/11 | still rejects a semicolon |
| C3 | neuter the `DANGEROUS_SQL_KEYWORDS` throw | RED 1/11 | never lets an allowlist name unblock a denied one |
| C4 | delete the union rationale from the doc comment | RED 1/11 | documents WHY it is a union |
| C5 | neuter the comment rejection | RED 2/11 | still rejects a line comment |
| A1 | delete the PostgreSQL `LIMIT` branch from the catch arm | RED 3/18 | appends LIMIT on PostgreSQL instead of emitting T-SQL TOP |
| A2 | drop `hasLimit` from the already-capped guard | RED 2/18 | is a no-op when the query already carries a LIMIT |
| A3 | hardcode `'postgresql'` instead of asking the provider | RED 2/18 | still emits TOP on SQL Server |
| A4 | stop stripping the trailing `;` before appending LIMIT | RED 1/18 | puts LIMIT before nothing |
| A5 | hardcode `'sqlserver'` in `resolvePlatform` | RED 8/18 | caps with LIMIT on PostgreSQL |

A5 is the interesting one: breaking the seam this PR added takes down eight tests across both the old and the new describes, which is the coupling we want.

### Suites

| package | before | after |
|---|---|---|
| `@memberjunction/global` | 879 | **897** |
| `@memberjunction/core` | 2582 | **2593** |
| `@memberjunction/core-actions` | 515 | **523** |

All three packages build clean. `build` is the real typecheck, and MJCore's and CoreActions' tsconfigs exclude `src/__tests__`, so both new/extended test files were typechecked separately against a scratch tsconfig. That turned up two **pre-existing** errors in this PR's own `run-adhoc-query.dialect.test.ts` — `as unknown as ReturnType<typeof BaseEntity.Provider>` on lines 33 and 104, where `Provider` is a getter rather than a function, so `ReturnType` does not apply. Corrected to `as unknown as typeof BaseEntity.Provider`. Invisible until now because the tsconfig excludes the directory and vitest transforms without typechecking.

Downstream consumers of the two allowlists were run as well: `GenericDatabaseProvider` 1084, `MJCoreEntitiesServer` 566, and the `SQLExpressionValidator`-consuming suites in `SQLServerDataProvider` (54) and `MJServer` (59). All green — nothing depended on those names being rejected.

No `any` types. No lockfile churn.

### Not verified

No live database. As with the rest of this PR, the PostgreSQL paths are exercised against a stub provider, so what is proven is that we emit `LIMIT n` and stop refusing `DATE_TRUNC` — not that a real PostgreSQL server accepts the resulting statement. That part is proven only by the fleet, where all three changes have been running as patch-package hunks.

Still a draft.

rkihm-BC (approved):
Approving.

## The mechanism claim is the one that matters, and it holds

> The mechanism is a system placeholder, not a registry of TemplateIDs … A template added tomorrow cannot be left off a list.

Verified: `_SQL_DIALECT_NAME` and `_SQL_DIALECT_RULES` are registered at `packages/AI/CorePlus/src/prompt.system-placeholders.ts:364` and `:372`, so they resolve on **every** prompt render. The question really does become a predicate — does this template reference the placeholder? — with nothing to keep in sync. Exactly **nine** template files changed, matching the description.

That distinction is load-bearing, and the description is right about why: a prior sweep used a hardcoded list and under-applied, which is how some templates stayed T-SQL-only. A fix for a drift problem that itself introduces a list to maintain is not a fix. This one does not.

## Composing the rules from the dialect's own primitives

`SQLDialect.PromptGuidance` deriving from `QuoteIdentifier`, `LimitClause`, `IsNull`, `DateAddExpression` and the rest — rather than a second per-dialect prose table — is the part that prevents the same drift returning at the guidance layer. Implementing a dialect becomes sufficient to get correct guidance, and only two new abstract members (`DisplayName`, `FormatterLanguage`) is a small price for that property.

## Face two is the more consequential half

`AdhocQueryResolver` requiring a SQL Server `ConnectionPool` from `context.dataSources` meant **every** ad-hoc query on PostgreSQL failed — and failed *quietly*, rendered downstream as "Showing cached data". The detail that `GetReadOnlyProvider` sat directly beneath it, already knowing the tenant was PostgreSQL and using that only for a string, is the kind of observation that makes a diagnosis convincing: the information was present and unused.

Routing both questions — which dialect to render, what to execute on — through the same read-only provider, with execution via `provider.ExecuteSQL`, is the right unification.

## Depth, stated honestly

32 files is more than I read. I verified the placeholder mechanism, the template count, and the structural argument; I did not review every parameterised template or the allowlist/row-cap changes in face three. Two things I would want before this leaves draft:

- **A rendered-prompt check on a PostgreSQL tenant** — that `_SQL_DIALECT_RULES` produces guidance an LLM actually follows, not just that it interpolates. The mechanism being right does not prove the prose is.
- **The nine templates read once by someone who knows the prompts**, since a template that references the placeholder but still carries T-SQL examples in its body would pass every check here and still teach the wrong dialect.

Everything I checked was accurate, which is a fair prior on the rest.

Issue #4435, opened 2026-09-13
Remove the PostgreSQL identifier auto-quoting shim; make runtime-authored SQL dialect-aware


jordanfanapour:
## Summary

`AutoQuotePostgreSQLIdentifiers` (`packages/SQLDialect/src/postgresqlAutoQuote.ts`) is a compatibility shim that rewrites SQL at execution time on PostgreSQL, wrapping any bare word it does not recognise as a keyword in double quotes. It exists so SQL written T-SQL-first (`FROM __mj.vwAIAgentRuns WHERE TestRun IS NULL`) keeps working on PostgreSQL, which folds unquoted identifiers to lowercase.

**Long-term direction: remove the shim, and make every place that depends on it produce correctly quoted SQL for the environment's dialect at the point where the SQL is written** — rather than guessing at execution time.

An investigation into removing it found that the dependency is wider than code call sites. Besides code-written SQL (a mechanical fix), the shim is currently the **only** thing quoting several kinds of SQL that are written at runtime or stored as data: row-level security filters, saved view WHERE clauses (including AI-generated Smart Filters), and cross-entity subqueries inside `ExtraFilter`. Removing the shim before those are handled would break reads for every non-admin user on PostgreSQL. This issue records that analysis and the plan to work through the breaking points.

Not urgent — this is the tracking issue for the eventual removal.

---

## Why remove it

- **It is a recurring defect source.** Fix rounds so far: `d77cef2b05` (origin, 2026-04-29), `9bb9b0b3be`, `e5a531ec49`, `1fdd5d0632` (#3697), `706842286f`. Related: #3604, #3590, #3691, #3775, #3792.
- **Its failures are silent wrong answers, not errors.** Documented in `packages/SQLDialect/src/__tests__/postgresqlAutoQuote.shippedQueries.test.ts`:
  - An apostrophe inside a SQL comment desynced the string-literal scanner, rewriting `WHERE ars."StepType" = 'Prompt'` to `= '"Prompt"'` — zero rows, no error.
  - `jsonb_build_object` keys became `'"ID"'`, so consumers reading `.ID` got `undefined`.
  - `{{ ConversationID | sqlString }}` became `{{ "ConversationID" | sqlString }}`, so the Nunjucks parameter never substituted and the query lost its filter.
[...]

PR #4436, opened 2026-09-13
fix(sql-dialect): close the PostgreSQL identifier auto-quoting keyword gaps

jordanfanapour:
## Summary
- **Production fix:** `AutoQuotePostgreSQLIdentifiers` rewrote `SELECT CURRENT_DATE` into `SELECT "CURRENT_DATE"`, which PostgreSQL rejects with `column "CURRENT_DATE" does not exist`. Found on a PostgreSQL client, where a generated query using `CURRENT_DATE` and `NULLS LAST` could never succeed: every retry got rewritten the same way, so the agent used up its iteration budget and produced nothing.
- **Same defect class, found by audit:** running realistic PostgreSQL through the tokenizer, 45 of 53 common constructs came back corrupted. Added the missing niladic datetime/identity functions, ordered-set aggregate and window-frame words (`WITHIN GROUP`, `WITH ORDINALITY`, `GROUPING SETS`, …), the rest of PostgreSQL's reserved words, cast-position type names (`CHARACTER VARYING`, `INT8`, `BOOL`, …), and utility verbs (`REFRESH`, `EXPLAIN`, `VACUUM`, …).
- **Adding a keyword can't cost a column.** Matching is case-sensitive, so only the ALL-CAPS spelling is exempt and a mixed-case column (`Current_Date`, `Within`) still gets quoted. Seventeen non-reserved words that are believable customer column names (`LEVEL`, `MODE`, `SEQUENCE`, …) are deliberately left out, the same way `USER` already is, and a test asserts they stay quoted.
- **The baseline guard's reserved-word oracle is now PostgreSQL's full reserved-word table instead of a sample.** Sampling is how `CURRENT_DATE` got through, so the next omission fails CI instead of shipping.

Longer term, #4435 tracks removing this shim entirely.

## Also in this PR: `@memberjunction/generic-database-provider`

The branch carries a second, independent fix with its own changeset (`pg-maxrows-preserves-sql-text`), in `packages/GenericDatabaseProvider` — `queryPagingEngine.ts` plus its tests. It is the other half of the same production failure and is easy to miss when reading only the summary above.

`QueryPagingEngine.applyMaxRowsViaAST` injected a `MaxRows` cap by calling `SetOuterCap` and re-emitting the whole statement with `SQLParser.ToSQL()`. That round-trip normalizes as it generates — keywords come back upper-cased, identifiers re-quoted — so appending one clause rewrote the statement around it:

```
caller wrote:   ORDER BY x ASC nulls last
ToSQL emitted:  ORDER BY x ASC NULLS LAST
auto-quoted to: ORDER BY x ASC "NULLS" "LAST"
PostgreSQL:     syntax error at or near ""NULLS""
```

Where a trailing `LIMIT N` is provably equivalent to the AST injection — a plain `SELECT`, no existing cap, a suffix-capping dialect, and no `OFFSET` / `FETCH` / `FOR UPDATE`-style clause that must follow `LIMIT` — the cap is now appended as text and the statement is left byte-identical. Every other shape falls through to the existing AST path unchanged. SQL Server is unaffected, since `TOP N` is a prefix no append can reach.

The two fixes are independent halves of one failure: completing the keyword list makes the upper-cased output harmless, while this stops the output changing in the first place.

## Test plan
- [ ] `packages/SQLDialect`: `pnpm test` — unit suite plus the baseline column/reserved-word guards
- [ ] `packages/PostgreSQLDataProvider` and `packages/CodeGenLib` PostgreSQL quoting tests still pass (both call the same tokenizer)
- [ ] `SELECT CURRENT_DATE`, `ORDER BY x ASC NULLS LAST` and `PERCENTILE_CONT(0.5) WITHIN GROUP (ORDER BY x)` go through `ExecuteSQL` unchanged on PostgreSQL
- [ ] Mixed-case column spellings of the new keywords are still quoted


rkihm-BC (approved):
## Approving. No changes requested.

Two fixes here, and the interesting one is the half the summary didn't mention — more on that below.

### The keyword additions are safe for the reason stated, and I verified the mechanism

The load-bearing claim is that adding a word can't cost anyone a column because only the ALL-CAPS spelling is exempt. That is exactly true, and it's one line:

```ts
if (word === word.toUpperCase() && PostgreSQLQuotingKeywords.has(word)) {
```

So `Current_Date`, `Within` and every other mixed-case spelling still quotes, and the blast radius of ~80 new words is confined to identifiers a customer wrote in full caps. The tiering behind that is what makes it defensible rather than merely bounded: reserved words are unconditionally safe because PostgreSQL won't resolve a same-named column bare either; the clause vocabulary and cast-position type names get their own separate rationale rather than borrowing the reserved-word one. Those are correctly scoped in the comments — I checked, because conflating them would have been the easy mistake.

Leaving out the seventeen believable column names, on the same reasoning as the existing `USER` exclusion, is the right call. `LEVEL`, `SEQUENCE` and `MODE` are exactly the words a customer schema uses and PostgreSQL does resolve bare.

### `appendTrailingCap` — I tried to break it and couldn't

This is the riskier change, since it alters how every `MaxRows` cap is applied on a suffix-capping dialect. Two ways it could plausibly emit invalid SQL, both already closed:

- **A trailing semicolon** would turn the append into `SELECT …;\nLIMIT 100`. It can't reach the function — `cleanedSQL = resolvedSQL.trimEnd().replace(/;\s*$/, '')` strips it upstream at both call sites.
- **A non-`SELECT` or a `UNION`** would make a bare `LIMIT` wrong or ambiguous. Also unreachable — the method gates on `kind === 'select'` and routes `mutation`, `select-into`, `set-op` and unparseable away first.

Beyond that, the `\n` before `LIMIT` means a trailing line comment can't swallow the clause, and `CLAUSES_THAT_FOLLOW_LIMIT` matching inside string literals or subqueries only ever causes a fall-through to the old AST path — the abstention is in the safe direction, as the comment says it is. Narrow, conservative, and it leaves the caller's SQL byte-identical in the common case.

Declining the existing `outerWrap` fallback is also right, and the reasoning is worth keeping: wrapping puts the cap above a subquery's `ORDER BY`, so "top 100 by rank" could return an arbitrary 100. That is a subtle correctness trap to have spotted and written down.

### One thing I changed

I edited the PR description to add an **"Also in this PR"** section for the `generic-database-provider` fix. The summary described only the auto-quoter, but `queryPagingEngine.ts` plus its tests are 155 of the 364 lines and an independent fix with its own changeset — a reviewer working from the body alone wouldn't know to look at the riskier half. The edit is purely additive and paraphrases your own changeset; overwrite it if you'd rather phrase it differently.

### One follow-up, filed rather than requested

**#4441** — the changeset says seventeen excluded words are pinned by a test; the test pins `USER` plus seven of them, and the other ten (`ROLE`, `PASSWORD`, `LOGIN`, `DOMAIN`, `CLUSTER`, `SEQUENCE`, `LOCAL`, `SKIP`, `EXCLUSIVE`, `SOURCE`) appear zero times in either test file.

Nothing is broken by this — none of the seventeen is in the keyword set, so all seventeen quote correctly with or without a test. It only matters against a future sweep of the kind this PR just did. Not worth churn on a correct PR, hence an issue. There is a small irony in it though, given this PR's own thesis that sampling is what let `CURRENT_DATE` through.

CI is 25 pass / 0 fail. Nice work — the audit that turned one production bug into "45 of 53 constructs were corrupted" is the part that makes this worth more than the outage it came from.

PR #4437, opened 2026-09-13
feat(vectors): the field picker refuses sensitive columns, and the use case picked is the type written

MS-BC:
## Why

Two kinds of defect in the Knowledge Hub Vectors tab, in the same save path.

**A safety gap.** Vectorizing a column copies its values into a third-party vector index, where they live **outside the entity permissions that normally guard them** and are retrievable by nearest-neighbour search rather than by an authorized query. For a password hash, a national identifier, a bank account or a salary, that is not a feature with a caveat — it is a disclosure. And once the values are written to the index it is not fully reversible. The Vectors tab's field suggestions come from a language model, which has no notion of which columns are sensitive and will propose `PasswordHash` quite happily if the use case sounds like identity matching. Nothing refused it.

**And four correctness defects in what the dialog writes**, found while auditing whether this surface covers what MJ Central is deleting. The first is the one to read: the dialog offers three use cases and honoured one of them, silently.

---

## 1 — Picking "search" wrote a Record Duplicate document

The dialog offers duplicate detection / search / classification. The save path hardcoded `ExtraFilter: "Name = 'Record Duplicate'"`. So two of the three choices were silently discarded and the document came out the wrong type.

That is worse than a mislabel, because **vector pools are typed**. `Provider.SearchEntity` and the Search Entity action read `Search`-typed documents — MJ seeds a `Search` Entity Document Type for exactly this, and documents it on the Vectorize Entity action's `EntityDocumentType` parameter ("Defaults to `Record Duplicate`. Pass `Search` …"). A search request stored as `Record Duplicate` is invisible to the feature that asked for it.

The type is now resolved from the selection. When the wanted type is absent the refusal **names it and lists the types that do exist** — it never substitutes one, since substituting is the defect.

**`classification` has no type in MJ.** The baseline seeds exactly two Entity Document Types, `Record Duplicate` and `Search`. So that button now fails with a message naming `Classification` rather than quietly writing a duplicate-detection document. A loud, specific failure beats a silent wrong write, and this surfaces a product decision — seed the type, or drop the button — instead of hiding it. Flagging rather than choosing.

## 2 — The field list was uncapped and came straight from a prompt

`DocumentSuggestionResult.selectedFields` arrived with no cap and no filter, and only the template *text* was user-editable. An unbounded list is how a template balloons and how columns nobody chose get embedded.

An over-long list is now truncated with the count shown; a template interpolating more than **24** distinct fields is refused before any write.

**Where 24 comes from** — the suggestion prompt's own contract, not another repo. `entity-document-suggestion.template.md` requires "**Output must be 1-4 natural language sentences**" and explains at length why a key-value dump embeds worse than prose. Four sentences (its ceiling) × six interpolations per sentence, past which a sentence stops being the English the prompt demands, gives 24. Cross-check: the prompt's three worked examples interpolate 8, 6 and 4 placeholders, so 24 is 3× the richest thing the prompt itself teaches and cannot fire on it.

The refusal is on the template rather than the model's field list because the template is what actually gets embedded and stays hand-editable — the same layer-1/layer-2 reasoning as the sensitive-field guard.

## 3 — A retry duplicated metadata

The flow writes Template → Template Content → Entity Document, and all three called `NewRecord()` unconditionally. Only the last is protected by a unique index — `UQ_EntityDocument_Name` on `EntityDocument.Name`. So a duplicate was **not** a rare partial-failure case: that index *guarantees* a second attempt with the same name fails at the final step, after the Template and Template Content rows have been written again. Orphans accumulated one pair per retry.

All three are now found-or-created, matched on the identity the rows actually have:

| row | matched on | why |
|---|---|---|
| Entity Document | `Name` | the identity the database itself enforces |
| Template | `Name` + `UserID` | no unique index; that pair is exactly what this flow writes |
| Template Content | `TemplateID`, `Priority ASC` first | the convention `loadEditDocTemplate` / `saveEditDocTemplate` already use |

An existing document is adopted **only when this dialog created it** — its `TemplateID` points at the deterministically-named `Template for <doc name>`. A collision with a document authored anywhere else is refused, not overwritten.

## 4 — The picker offered MJ's own internals

`loadEntityGroups` iterated `md.Entities` unfiltered, so MemberJunction's own framework entities — prompt runs, audit logs, record changes — were offered as vectorization targets alongside the user's business tables.

Excluded, following the existing frontend precedent rather than a new rule: `FRONTEND_BLOCKED_SCHEMAS = new Set(['__mj'])`, in `database-designer.engine.ts` and `step-relationships.component.ts`.

The two rules are currently **coextensive** — in the 5.46 baseline, 378 entities are `MJ:`-prefixed and 378 are in `__mj`, exactly the same set. The name-prefix test is therefore defence in depth against an MJ entity registered elsewhere, not extra coverage today, and the comment says so.

Worth knowing: MJ *does* seed two `Search`-typed documents over `__mj` entities (`AI Agents Search`, `AI Prompts Search`), so framework entities are not unvectorizable in principle. Those arrive through metadata sync — a different authoring channel with a different audience. This dialog is where a business user configures vectorization of their own data, and burying their tables under 378 framework ones is the defect. If you read that differently, this is the one to push back on.

---

## The sensitive-field refusal

**Layer 1 — the model never sees them.** The field list handed to the suggestion prompt is filtered, so a sensitive column cannot be suggested.

**Layer 2 — the saved template is re-checked.** The template is hand-editable after the model returns, so a user can type a field name the picker never offered. Layer 1 alone would be a prompt-shaped guarantee, which is no guarantee at all.

The check runs **before anything is written**, and the refusal **names the offending fields**.

Matching is on normalised substrings — lower-cased, separators stripped — so `password_hash`, `PasswordHash` and `PASSWORD` are one rule. Covered: credentials and secrets, government and tax identifiers, financial columns, and special-category data. Prose is not flagged, only interpolated placeholders.

The placeholder parser is now **shared** between this guard and the field cap, so the two refusals cannot drift into disagreeing about what a placeholder is.

## Tests

56 total on this branch: the original 19, plus 37 for the four defects above.

The component-level tests are **not source-text greps**. They build the component under the package's node-preset mock pattern with a small in-memory metadata store, run the real save path, and assert on the rows that actually get written — including the retry convergence, which counts rows in the store across two `SaveAsEntityDocument()` calls with a simulated mid-flow failure.

Mutation-tested, 13 kill mutations all red:

| mutation | fail |
|---|---|
| `search` maps to Record Duplicate | 5 |
| component hardcodes the type again | 5 |
| absent-type message drops the name | 1 |
| cap raised to 100000 (effectively removed) | 1 |
| component never checks the cap | 1 |
| cap message drops the counts | 2 |
| selected fields not truncated | 1 |
| template always NewRecord | 2 |
| content always NewRecord | 1 |
| adopts anyone else's document | 1 |
| every entity is vectorizable | 3 |
| picker stops filtering | 1 |
| shared placeholder parser returns nothing | 7 |
| *control: cap 24 → 20* | *green, by design* |

Two greens were worth chasing rather than waving through:

- **The absent-type-message mutation was itself invalid.** It stripped the type name from the opening clause, but the name survived in the trailing "Add `<type>` to MJ: Entity Document Types" clause — so the message genuinely still named the type and the test was right to pass. Redone against both occurrences; now red.
- **Raising the cap to 100000 exposed a real hole.** Every cap test was written *relative to* `MAX_TEMPLATE_FIELDS`, so none of them could see the constant change — the guard could have been silently removed by editing one number. Fixed by bracketing the constant: `> 8` (below the prompt's richest example it would refuse legitimate output) and `≤ 40` (above a wide table's column count it bounds nothing). The value is not asserted equal, because that would just restate the constant — which is why the 24 → 20 control stays green, and why recalibrating the cap is not a test change.

The runner was proved to execute the new file by making one test fail on purpose.

| check | result |
|---|---|
| ng-dashboards suite | **1810 passing across 159 files** (from 1772 / 158) |
| `build` (ngc — the real typecheck) | clean |
| `test:types` (tsc --noEmit, spec tsconfig) | clean |
| `any` types | none |
| `pnpm-lock.yaml` | untouched |

## Not in scope here

- **`Dimensions: null` when creating an index** (the component never records the width) and **the embedding-model fallback** are both left alone: #4449 makes the index's declared width authoritative and adds a backfill. It touches no file this PR touches — **no collision**.
- **A gap in this PR's own guard, for a follow-up:** `saveEditDocTemplate` (the edit panel's template save) is guarded by *neither* the sensitive-field check nor the field cap. Both live in `createTemplateForDocument`, on the create path only. So a user can create a clean document and then edit `{{PasswordHash}}` into its template. Fixing it needs a decision about existing documents that already exceed either rule, so it is reported rather than bundled.

## Context

MJ Central had these guards and is removing its duplicate of this picker in favour of Knowledge Hub, on the principle that what MJ can do should be left to MJ. The capability is genuinely duplicated; these five guardrails were not, so they are offered here rather than deleted with the rest.


rkihm-BC (changes requested):
## Requesting changes — on calibration, not design. The guard itself should ship.

The threat model is correctly stated and worth restating: vectorizing a column copies its values somewhere the entity permissions don't reach, retrievable by similarity rather than by an authorized query, and not fully reversible once written. That the field suggestions come from a model which will happily propose `PasswordHash` makes this a real hole, and the two-layer structure is the right answer — layer 1 alone would be a prompt-shaped guarantee, exactly as you say. Layer 2 is genuinely necessary because the template stays hand-editable after the model returns.

The mutation table is the part I'd point other PRs at. Ten mutations, all red, and the note that two initially survived because of shell escaping and a mis-applied ordering change — *"a mutation that silently does nothing reads exactly like a passing test"* — is a more useful observation than most test plans contain.

Two things to fix.

## 1 — A changeset is required

`@memberjunction/ng-dashboards` is a published package (`6.1.0-edge.6`) and this changes its behavior. Release notes are built from changesets, so without one this ships invisibly. `patch` is the right level — no migration, no metadata.

## 2 — Four patterns match on spelling, not meaning

I ran all 43 patterns against MJ's own 2,141 core entity field names. **123 are refused**, and the overwhelming majority are collisions rather than judgment calls.

`tin` is the worst, and it is worse than a field-count conveys, because it is a substring of ordinary English:

```
Ra-tin-g      Marke-tin-g   Mee-tin-g     Set-tin-g     Repor-tin-g
Pos-tin-g     Lis-tin-g     Accoun-tin-g  Consul-tin-g  Tes-tin-g     Con-tin-ent
```

It hits 46 MJ fields (`AgentSettings`, `CapabilitySettings`, `AllowDecryptInAPI`, …) and **not one is a tax identifier**. `Rating` and `MarketingConsent` are precisely the columns someone vectorizing customer records reaches for.

The others:

| pattern | MJ fields hit | what it actually catches | true positives |
|---|---|---|---|
| `tin` | 46 | `*Settings*`, and `Rating`/`Marketing`/`Meeting` in any schema | **0** |
| `token` | 36 | `ContextWindowMaxTokens`, `FirstTokenTime`, `InputTokenLimit` | a few (`AccessToken`) |
| `ssn` | 6 | `ClassName`, `DriverClassName`, `EntityClassName` — `cla`**`ssn`**`ame` | **0** |
| `race` | 0 in core | `StackTrace`, `TraceID`, `TraceName` — `T`**`race`** | **0** |
| `dob` | 1 | `EmbeddedRecordObject` — `recor`**`dob`**`ject` | **0** |

### Why the asymmetry argument doesn't cover these

You preempt this: *"Anyone tempted to narrow the patterns should read that test first."* I did, and I agree with it **for `auth`**. `AuthorName` is a semantic near-miss — `auth` genuinely is the prefix of authentication, the resemblance is real, and over-refusing it is a defensible trade you weighed and pinned.

`Rating` is not that. It is not refused because it resembles a tax identifier; it is refused because the letters `t-i-n` appear in the middle of an unrelated word. There is no signal being traded off — the match carries zero information about sensitivity. The asymmetry argument justifies over-refusing things that *look like* the concept; it cannot justify matches with no semantic relationship at all.

The cost is also higher than "a visible, correctable annoyance." Layer 2 `throw`s, so the save is refused outright — a user who needs `{{Rating}}` in a template has no workaround. Not a nuisance they route around; a feature they cannot use.

### Suggested changes

- **`tin`** — drop it. `taxid` already covers the concept, and this adds an enormous collision surface for nothing.
- **`ssn`, `dob`, `race`** — require a separator boundary rather than a bare substring. All three are whole tokens in real schemas (`ssn`, `employee_ssn`, `date_of_birth`), so a boundary keeps every true positive.
- **`token`** — narrow to `accesstoken` / `refreshtoken` / `apitoken` / `bearertoken`. Keeps `AccessToken`, drops 35 token *counts*, which are not secrets.

Leave **`auth`** as it is — your call, documented, and pinned. Leave **`hash`** too; `password` already catches `PasswordHash`, so the remaining hits are checksums, but that one is marginal either way and not worth a round trip.

The other ~38 patterns look right to me, and the credentials/financial/special-category grouping is sensible.

## Worth keeping

The five tests that read the component source so an unwiring is a red test rather than a silent regression, and stripping comments first so prose about the guard cannot satisfy one. That is the right instinct for a guard whose whole failure mode is being quietly removed.

Once the changeset is in and those four patterns are narrowed, I'd approve this quickly — the capability is a genuine gap being closed and I'd like it to land.

MS-BC:
## Folded in four more defects from the same audit — and the first one is the headline

Same file, same save path, so they belong here rather than in four more PRs. Body updated; title now covers the scope.

### The UI offers three choices and silently honoured one

`vector-management-resource.component.html` gives you **duplicate detection / search / classification**. The save path hardcoded:

```ts
ExtraFilter: "Name = 'Record Duplicate'"
```

So picking "search" discarded the selection and wrote a **Record Duplicate** document, with no error and nothing in the UI to suggest it had happened.

That is worse than a wrong label, because vector pools are typed. `Provider.SearchEntity` and the Search Entity action read **`Search`**-typed documents — MJ seeds a `Search` Entity Document Type for precisely this and documents it on the Vectorize Entity action's `EntityDocumentType` parameter:

> Defaults to `Record Duplicate` (original behavior). Pass `Search` to drive the search-tier vector pool used by `Provider.SearchEntity` / Search Entity action.

So a user who asked for search got a document that search cannot see. It would look like "vectorization ran and search still finds nothing" — a symptom with no obvious cause anywhere near this dialog.

The type is now resolved from the selection, using MJ's own type names, with a trimmed case-insensitive match so the dialog and `EntityDocumentCache.GetDocumentTypeByName` agree. When the wanted type is absent the refusal **names it and lists the types that do exist**, and substitutes nothing.

### One thing that did not survive contact with the code: `classification` has no type

The 5.46 baseline seeds exactly **two** Entity Document Types — `Record Duplicate` and `Search`. There is no `Classification`.

So that third button now fails with a message naming `Classification` and listing what does exist, rather than quietly writing a duplicate-detection document. That is strictly better than today, but it does mean a permanently-failing button on a stock install until someone decides: **seed the type, or drop the button?** That is a product call, so this surfaces it instead of picking for you. Happy to do either.

### The other three

**The field list was uncapped and came straight from a prompt.** Now truncated with the count shown, and a template interpolating more than **24** distinct fields is refused before any write. The bound is derived from this repo's own prompt — `entity-document-suggestion.template.md` requires "1-4 natural language sentences", so 4 sentences × 6 interpolations = 24, and the prompt's richest worked example uses 8, putting the cap 3× clear of anything the prompt teaches. Not copied from the MJ Central equivalent, which used 30.

**A retry duplicated metadata, deterministically.** The flow writes Template → Template Content → Entity Document and each called `NewRecord()` unconditionally. Only the last has a unique index (`UQ_EntityDocument_Name`), so this was never a rare race: that index *guarantees* the second attempt with the same name fails at the final step, **after** re-writing the first two. One orphaned Template + Template Content per retry, forever. All three are now found-or-created on the identity the rows actually have — Name for the document (DB-enforced), Name + UserID for the template, `TemplateID` + `Priority ASC` first for the content, which is the convention `loadEditDocTemplate` already uses. An existing document is adopted **only when this dialog created it**; a name collision with someone else's document is refused, not overwritten.

**The picker offered MJ's own internals.** `loadEntityGroups` iterated `md.Entities` unfiltered. Now filtered, following `FRONTEND_BLOCKED_SCHEMAS = new Set(['__mj'])` from `database-designer.engine.ts` rather than a new rule. Two honest notes in the comment: the `__mj` and `MJ:` rules are currently **coextensive** (378 and 378, the same set), so the prefix test is defence in depth and not extra coverage; and MJ itself seeds two `Search` documents over `__mj` entities, so framework entities are not unvectorizable in principle — those come via metadata sync, a different channel with a different audience. **This is the one I would most expect you to overrule**, and the evidence is there to do it with.

### Mutation testing turned up two things worth repeating

13 kill mutations, all red. Two initial greens, and they were *different kinds* of green:

- **The absent-type-message mutation was invalid.** It stripped the type name from the opening clause — but the name survived in the trailing "Add `<type>` to MJ: Entity Document Types" clause, so the message genuinely still named the type. The pin was right; the mutation was wrong. Redone against both occurrences, now red.
- **Raising the cap to 100000 exposed a real hole.** Every cap test was written *relative to* `MAX_TEMPLATE_FIELDS`, so none could see the constant change — the whole guard could have been deleted by editing one number, with a green suite. Fixed by bracketing it: `> 8` (below the prompt's richest example it refuses legitimate output) and `≤ 40` (above a wide table's column count it bounds nothing). Deliberately **not** asserted equal to 24, so recalibrating the cap is a judgment call and not a test change — which is why the 24 → 20 control mutation stays green.

The component tests are not source greps: they build the component under this package's node-preset mock pattern with an in-memory metadata store, run the real save path, and assert on the rows written.

| check | result |
|---|---|
| ng-dashboards suite | 1810 passing / 159 files (was 1772 / 158) |
| `build` (ngc) | clean |
| `test:types` | clean |
| `any` types | none |
| lockfile | untouched |

### On your review

**The changeset is in** — `.changeset/vector-picker-refusals.md`, `patch` on `@memberjunction/ng-dashboards`, covering the guard and all four folded-in fixes.

**I have not touched the four patterns** (`tin`, `token`, `ssn`, `dob`, `race`). Your 2,141-field analysis is more evidence than I have, and I agree with the distinction you drew — `Rating` being refused because `t-i-n` appears mid-word carries no signal about sensitivity, which is not the trade the `auth` test documents. But narrowing them is the author's calibration call on a safety guard, and I did not want to bundle it into a commit about something else. It is the remaining item on your review.

### Two things deliberately left alone

- **`Dimensions: null` when creating an index**, and the embedding-model fallback. #4449 makes the index's declared width authoritative and adds a backfill. Checked for collision: **none** — #4449 touches no file this PR touches.
- **A gap in this PR's own guard, reported not fixed.** `saveEditDocTemplate` (the edit panel) is covered by **neither** the sensitive-field check nor the field cap — both live in `createTemplateForDocument`, on the create path only. So a user can create a clean document and then edit `{{PasswordHash}}` into its template, bypassing layer 2 entirely. Closing it needs a decision about existing documents that already violate either rule, so it wants its own change rather than a quiet addition here.

MS-BC:
You were right, and the evidence was the part that made it unarguable. I reproduced your measurement independently before changing anything — 2,111 distinct field names out of MJ's generated core entity classes, **119 refused**. Same picture as your 123/2,141. Fixed in 2d6bc09.

## Per pattern

| pattern | change | why |
|---|---|---|
| `tin` | **dropped** | 43 MJ columns, zero tax identifiers. `taxid` already covers the concept. |
| `token` | narrowed to `accesstoken` / `refreshtoken` / `apitoken` / `bearertoken` | keeps `AccessToken`, `RefreshToken`, `RegistrationAccessToken`; drops the 35 token *counts*. `AuthToken` and `OAuthToken` are still caught by `auth`. |
| `ssn` | must begin a word | keeps `ssn`, `employee_ssn`, `EmployeeSSN`, `SSNumber`; drops all six `*ClassName` columns. |
| `dob` | must begin a word | keeps `dob`, `DOB`, `DOBDate`; drops `EmbeddedRecordObject`. `date_of_birth` was always covered by `dateofbirth`. |
| `race` | must begin a word | keeps `Race`, `EmployeeRace`, `race_ethnicity`; drops `StackTrace`, `TraceID`. |
| `auth`, `hash` | **untouched**, as you asked | `AuthorName` is pinned as a deliberate trade; `password` still catches `PasswordHash`. |

**Refusals over the real fixture: 119 → 40.** The 40 that remain are credential-adjacent by name — `auth` (16), `hash` (8), `credential` (7), `secret` (4), `apikey` (2), the token compounds, and `RoutingOrder`, which `routing` catches on its own merits.

## "Boundary" needed building, not borrowing

Worth flagging because it is the kind of thing that looks done and isn't: `normalizeFieldName` **removes** separators rather than replacing them, so by the time a pattern is matched there is no separator left to anchor against — `employee_ssn` and `className` are both one unbroken run of letters (`employeessn`, `classname`). A `\b` would have matched at offset 0 and nowhere else, quietly dropping `employee_ssn` while still refusing nothing useful.

So word starts are computed from the **original** spelling, where the evidence still exists, and carried as offsets into the normalised string: after a separator, and at the two camelCase transitions — lower-or-digit followed by upper (`employee|SSN`), and the tail of an acronym run beginning a capitalised word (`SS|Number`, which is what keeps `SSNumber`). Prefix-of-a-word rather than whole-word, for that last reason.

## The test that was missing

You quantified the problem, so the fix is quantified too. The fixture is derived, not hand-typed — CodeGen emits a `* * Field Name:` line per column into the zod docblocks in `MJCoreEntities`, which is the only generated field-name source in the repo, so it is the whole of MJ's own vocabulary and grows with the schema. On top of the true-positive and collision assertions, three of the collision sets are **derived from that fixture** rather than listed (`ssn`, `tin`, plural `Tokens`), so a new `*ClassName` column is covered the day it appears.

And the refusal count is pinned as a bracket — 30..55, plus a rate ceiling so it keeps its meaning as the fixture grows. That is the assertion this PR should have had from the start.

I mutation-tested it, because your point about the last round applies here too — every cap test in the earlier work was written relative to its constant, so changing the constant kept the suite green. Each mutation is verified on disk (hash changed, old literal gone at the expected occurrence count, new literal present) before the run, and three of my first attempts were correctly rejected as invalid when the replacement text happened to contain the original as a substring.

| mutation | result | caught by |
|---|---|---|
| `tin` re-added | RED | 4 tests, incl. the bracket |
| boundary removed from `ssn` | RED | derived `ssn` collision set + named columns |
| boundary removed from `dob` | RED | 2 tests |
| boundary removed from `race` | RED | 2 tests |
| `token` widened back to a bare substring | RED | plural-`Tokens` set + the bracket |
| `ssn` boundary removed **and** its derived assertion deleted | RED | the named-column assertion still holds it |
| `url` added — a broad pattern no other assertion names | RED | **the bracket alone**, 1 failing test |
| `url` added **and** the bracket deleted | GREEN | nothing — which is the point |
| comment reworded (labelled control) | GREEN | expected |

The last two are the pair that matters. `url` refuses 65 MJ columns and no other assertion in the file mentions it; the bracket is the only thing that sees it, and deleting the bracket makes the suite green again. That is the unforeseen-pattern case, and it is now covered.

Two honest notes. `MarketingConsent`, `StackTrace`, `TraceID` and `EmbeddedRecordObject` are in the collision assertion but are **not** MJ core columns — they are your worked examples, and ordinary names in real schemas, so I kept them as regression cases and said so in the test. And I did not touch `hash`, so `ContentHash`, `ResultHash` and `SourceFieldHash` are still refused; you called that marginal and I agree, so it stays as it is rather than riding along on this change.

Changeset extended in place rather than adding a second — `@memberjunction/ng-dashboards`, `patch`, now covering the calibration with the before/after numbers.

Build clean; package suite 159 files / 1820 tests green, up from 1810. Staying a draft.

rkihm-BC (changes requested):
## Prior items are all addressed and verified; one product decision still blocks un-drafting

Re-reviewed at `2d6bc0928d` with the real save path driven through the PR's own harness.

### What I asked for last time

| Item | Status | Evidence |
|---|---|---|
| Changeset | Done | `.changeset/vector-picker-refusals.md`, `ng-dashboards: patch` |
| Drop `tin` | Done | Gone from `sensitive-fields.ts:56`; comment at `:53-55` |
| Narrow `token` | Done | Four compounds at `:50-51` |
| Word-anchor `ssn`/`dob`/`race` | Done | `SENSITIVE_FIELD_WORD_PATTERNS` at `:79`, boundary computed from the original spelling |
| False-positive count | Reproduced | 2,133 distinct getters in `__mj.ts`, 40 refused, same 40 as your list. All 40 live in `__mj`, which the picker now excludes via `VECTOR_BLOCKED_SCHEMAS`, and none of the six seeded `__mj` Search docs owns any of them. |

Red-then-green: with the component swapped back to `next`, 18 tests fail; with `sensitive-fields.ts` at the pre-recalibration commit, 6 fail. Restored: 67/67 and 29/29. Full `ng-dashboards` suite 1820 passed. The Integration tier failure on this PR is the `mj sync push` transaction error that `next` hit twice on 9/14; it needs a re-run, not a change.

### Blocking

**The `classification` button hard-fails on every stock install.** `metadata/entity-document-types/` seeds only `Search`, the SQL baseline seeds only `Record Duplicate`, and no `Classification` type exists anywhere. The button is wired at HTML `:469-472`. You flagged this yourself as open; it needs a decision (seed the type, or remove or disable the button) before this leaves draft. A button that errors on every install is the one thing here that is worse for users than `next`.

### Found while driving the save path, all optional

- **Adoption and template lookup disagree.** `findOrCreateEntityDocument` (`:1566-1596`) adopts a same-named document when its template is named `Template for <name>`, checking Name only (`templateHasName`, `:1644-1653`). `findOrCreateTemplate` (`:1601`) then matches Name plus `UserID`, misses when the original was another user's, creates a new template, and re-points the adopted document to it, orphaning the original pair. Measured: user-2's doc ends up on user-1's template and the notification says "created". This is not a regression, since Entity Documents have no owner and `next` wrote the same orphan pair before failing on `UQ_EntityDocument_Name`, but it undercuts the "retry converges on one set of rows" claim. Reusing `priorDoc.TemplateID` on adoption fixes it in one line.
- **Layer 1 does not cover related entities.** `buildPromptData` (`:1805-1807`) passes `relatedEntity.Fields` unfiltered into `RelationshipsJSON`; measured `PasswordHash` and `SSN` reaching the prompt from a related `Users` entity. Layer 2 refuses the dotted form on write, so the comment "the model never sees these fields" at `sensitive-fields.ts:17` is the false part. `withoutSensitiveFields(relatedEntity.Fields)` makes it true.
- **Placeholder parser bypasses.** `{{ (PasswordHash) }}`, `{{ Name ~ " " ~ PasswordHash }}`, and `{% set p = PasswordHash %}{{ p }}` all return no placeholder names. None is a form the prompt teaches, and the unguarded `saveEditDocTemplate` you already noted is a far larger hole, so hardening only.
- **Pattern slips**: `DriversLicense` and `DriversLicenseNumber` (pattern is `driverlicense`, no `s`, and the possessive spelling is the common one), `ConnectionString`, `SessionToken`, `IdToken`, and whole-word `TIN` now that `tin` is gone. `routing` still refuses `RoutingOrder`; that is the documented over-refusal trade, one core column, fine.

Style is clean: no `any`, no `.Get()/.Set()`, all entity names prefixed, `RunView` scoped through `ProviderToUse`.

PR #4442, opened 2026-09-13
ci(codeql): run js-rest on pull requests, not only weekly

rkihm-BC:
## Summary

Flips `js-rest`'s `run_on` from `"schedule"` to `"always"`, so the non-core ~70% of the repo's TypeScript is analyzed on every pull request instead of only on the Monday cron.

One word of behavior change, plus the rationale written into the workflow's existing `WHAT RUNS WHEN` section.

## Why: the check was red by construction

GitHub's code-scanning PR check compares the configurations that produced results on the **base branch** against those the **PR** produces. The weekly run uploads under `/language:javascript-typescript/rest`, so any PR that had an alert to report also failed the `CodeQL` check with:

> **1 configuration not found** — `/language:javascript-typescript/rest`

#4176 is the worked example: exactly **one** alert was genuinely introduced by that branch (`js/polynomial-redos`, verified a false positive and dismissed), yet the check stayed red for a reason that had nothing to do with the code. PRs with no alerts to report never rendered the check at all, which is why this went unnoticed — ten recently merged PRs show `CodeQL: skipping` rather than red.

A security check that is red by construction is worse than the wall time it was saving, because it trains people to ignore it.

## Why the cost is near zero

`js-rest` measured four times by `workflow_dispatch` on `next` (13 September 2026):

| run | duration |
|---|---|
| 1 | 8.6 min |
| 2 | 8.8 min |
| 3 | 8.8 min |
| 4 | 8.8 min |

A 2.3% spread — far tighter than `js-core`'s 2.5–4.3 min range across 24 runs, so this is a predictable cost rather than a variable one.

Matrix legs run in parallel, so CodeQL contributes its **slowest leg**, not the sum. The PR critical path is set elsewhere:

| check | min | typical | max |
|---|---|---|---|
| Integration (SQL Server) | 7.8m | 10.0m | 14.1m |
| Native-ESM import guard | 2.1m | 12.3m | 15.6m |
| CodeQL today | — | 4.1m | — |
| CodeQL with `js-rest` | — | **8.8m** | — |

And these are **full-scan** figures. PR runs are diff-informed — the workflow header already records that `js-core` full-scans in 7m31s but takes about 3 minutes on a PR — so the real PR cost of `js-rest` is lower than 8.8 min again.

## The weekly run is not paused

`always` is a strict superset of `schedule` in the selection predicate:

```jq
.run_on == "always"
or (.run_on == "schedule" and ($event == "schedule" or $event == "workflow_dispatch"))
```

The first clause has no event test. Verified by running the actual predicate against the actual matrix for each event:

```
pull_request       -> actions, python, js-core, js-rest
push               -> actions, python, js-core, js-rest
schedule           -> actions, python, js-core, js-rest
workflow_dispatch  -> actions, python, js-core, js-rest
```

`js-rest` still runs on the Monday cron, so the schedule keeps doing the thing only it can do: re-scanning unchanged code with GitHub's newest query packs.

## Alternatives considered

**Remove the `category` so `js-rest` stops registering as a configuration.** Not viable — all four legs share one analysis key (`.github/workflows/codeql.yml:analyze`) and are distinguished *only* by category, so dropping it collides `js-rest` with `js-core`.

**Delete the stale `js-rest` analyses from `next`.** Buys one day; the Monday cron recreates the configuration.

**Run `js-rest` on PRs with the cheaper default suite.** Would save runner minutes, but adds a mechanism whose interaction with the configuration check I could not verify, to solve a cost that turns out not to be on the critical path.

## Verification

- The embedded matrix JSON still parses (`jq -e`, 4 entries) and the YAML still loads.
- The selection predicate was run per event, as shown above.
- No changeset: this touches no published package.


rkihm-BC:
**First PR-mode measurement, from #4448** (the first pull request analyzed after this merged):

| leg | full scan on `next` (dispatch, n=4) | diff-informed PR run |
|---|---|---|
| `js-core` | 4.1 min | **2m58s** |
| `js-rest` | 8.6 / 8.8 / 8.8 / 8.8 min | **6m9s** |

So the PR cost of `js-rest` is about 30% under the full-scan figure this PR was decided on, and it lands **below the 7.8-minute floor** of the Integration check. The one caveat in the description — that on the very fastest PRs `js-rest` might become the long pole by about a minute — does not materialize: CodeQL is never the critical path. `js-core` at 2m58s also matches the workflow header's "about 3 minutes" note exactly.

PR #4443, opened 2026-09-13
fix(codegen, db-auto-doc): create TransitiveView bridge views on PostgreSQL

### LINKED ISSUE #4409: CodeGen: TransitiveView organic-key config hardcodes CREATE OR ALTER VIEW — SQL Server syntax, unusable on PostgreSQL
## Summary

`ManageMetadataBase.processOrganicKeyConfig()` creates transitive bridge views with a hardcoded
`CREATE OR ALTER VIEW` statement. That is SQL Server syntax; PostgreSQL requires
`CREATE OR REPLACE VIEW`. Any PostgreSQL client that declares a `TransitiveView` in
`additionalSchemaInfo.json` therefore cannot use the feature — CodeGen fails on the DDL.

Confirmed present on `next` at `6.1.0-edge.6`, and in the published `@memberjunction/codegen-lib`
5.51.0 that the affected client is running.

## The defect

`packages/CodeGenLib/src/Database/manage-metadata.ts:1001`

```ts
// Step 1: Create transitive views if defined
for (const re of okConfig.RelatedEntities) {
   if (re.TransitiveView) {
      const viewSchema = re.TransitiveView.SchemaName || re.SchemaName;
      const viewFullName = `${viewSchema}.${re.TransitiveView.Name}`;

      const viewSQL = `CREATE OR ALTER VIEW ${this.qs(viewSchema, re.TransitiveView.Name)} AS\n${re.TransitiveView.SQL}`;
      await this.LogSQLAndExecute(pool, viewSQL, ...);
```

## Why this reads as an oversight rather than a design choice

The dialect abstraction already exists, and **the same file uses it correctly in two other places**:

- `manage-metadata.ts:1632` — `this.dbProvider.generateMaterializedWrapperViewSQL(matSchema, viewName, tableName)`
- `manage-metadata.ts:1934` — same call for the core schema

with the platform implementations at:

- `providers/sqlserver/SQLServerCodeGenProvider.ts:187` → `CREATE OR ALTER VIEW`
- `providers/postgresql/PostgreSQLCodeGenProvider.ts:456` → PostgreSQL form
- `codeGenDatabaseProvider.ts:442` → base declaration

The base-view generation path is dialect-aware too, including a documented PG `42P16` fallback
(`codeGenDatabaseProvider.ts:362`, `:1531`). Only the `TransitiveView` path bypasses all of it.

The type doc at `manage-metadata.ts:215` even records the assumption explicitly:

> `/** Raw SQL for the view body (the SELECT statement). CodeGen emits CREATE OR ALTER VIEW wrapping this. */`

## Impact

- **PostgreSQL clients cannot use `TransitiveView` at all.** Transitive organic keys are the only
  mechanism for expressing a multi-column or normalized join between two entities, since
  `EntityField.RelatedEntityFieldName` and `EntityRelationship.RelatedEntityJoinField` are both
  single-column. Losing bridge-view creation removes the only expressive option.
- The rest of the transitive feature works fine on PostgreSQL — `TransitiveObjectName`,
  `TransitiveObjectMatchFieldNames`, `TransitiveObjectOutputFieldName` and
  `RelatedEntityJoinFieldName` are all written correctly, surface on `EntityInfo`, and render
  downstream. It is only the inline view *creation* that breaks.

## Reproduction

1. On a PostgreSQL MJ instance, add an organic key to `additionalSchemaInfo.json` with a related
   entity declaring a `TransitiveView`:

```json
{
  "Name": "Employee Attribute Link",
  "MatchFieldNames": ["recordKey"],
  "NormalizationStrategy": "Trim",
  "RelatedEntities": [{
    "SchemaName": "acgi",
    "TableName": "EmployeeAttribute",
    "TransitiveView": {
      "SchemaName": "acgi",
      "Name": "vwBridgeEmployeeAttribute",
      "SQL": "SELECT p.\"recordKey\" AS \"ParentRecordKey\", c.\"recordKey\" AS \"ChildRecordKey\" FROM ..."
    },
    "TransitiveMatchFieldNames": ["ParentRecordKey"],
    "TransitiveOutputFieldName": "ChildRecordKey",
    "RelatedEntityJoinFieldName": "recordKey"
  }]
}
```

2. Run `mj codegen`. The view DDL fails — PostgreSQL has no `CREATE OR ALTER VIEW`.

## Workaround

Create the bridge views out of band with `CREATE OR REPLACE VIEW`, then reference them from the
config with `TransitiveObject: "schema.viewName"` instead of `TransitiveView`. The upsert reads
`reConfig.TransitiveObject || null` directly (`manage-metadata.ts`), so a pre-existing view wires up
correctly and the rest of the feature behaves as documented. Verified working end to end on
PostgreSQL: 6 transitive organic keys created, correct fields persisted, and the joins return the
expected rows through the base views.

## Suggested fix

Route the `TransitiveView` path through the existing provider hook rather than hardcoding the
dialect — either by reusing `generateMaterializedWrapperViewSQL`'s pattern or by adding a
`generateCreateOrReplaceViewSQL(schema, viewName, body)` hook to `CodeGenDatabaseProvider` with
per-dialect overrides. That also picks up the PG `42P16` fallback handling the base-view path
already has, which this path will eventually need for the same reason.

Worth updating the `TransitiveView.SQL` doc comment at `manage-metadata.ts:215` at the same time,
since it states the SQL Server form as though it were universal.

jordanfanapour:
Fixes #4409

## Summary
- **CodeGen:** organic-key `TransitiveView` bridge views were created with a hardcoded SQL Server `CREATE OR ALTER VIEW`, which PostgreSQL rejects. The failure is caught per key and logged, so CodeGen reported success while both the view and the organic key it backs were missing. The DDL now comes from a new provider hook, `CodeGenDatabaseProvider.generateCreateOrReplaceViewSQL`:
  - SQL Server: `CREATE OR ALTER VIEW`, unchanged behavior.
  - PostgreSQL: `CREATE OR REPLACE VIEW` inside a `DO` block that, on SQLSTATE `42P16` only (a column rename, reorder or retype), drops and recreates the view. The drop is **not** `CASCADE`: nothing CodeGen generates depends on a config-declared view, so any dependent belongs to the user, and the run fails with PostgreSQL's dependency error instead of silently destroying it.
- **CodeGen, second bug found alongside:** on SQL Server the bridge view was logged to `CodeGen_Run_*.sql` without a `GO` after it, so replaying the migration failed on the next statement (`Msg 156`). The call now passes the provider's batch separator.
- **DBAutoDoc:** the bridge-view body it writes into additionalSchemaInfo (`TransitiveView.SQL`) was always bracket-quoted, and CodeGen executes that body verbatim. Identifiers are now quoted per `database.provider`: brackets on SQL Server (the default, output unchanged), double quotes on PostgreSQL/Oracle, backticks on MySQL. Without this, CodeGen would still fail at `[` for anyone feeding DBAutoDoc output to a PostgreSQL CodeGen run.
- **Behavior note:** on PostgreSQL the view body travels inside dollar quotes, so the identifier auto-quoter (#4435) doesn't touch it. The body must be valid PostgreSQL as written, and the `TransitiveView.SQL` doc now says so.

## Test plan
- [x] New unit tests: `organic-key-transitive-view.test.ts` (10) drives the real `processOrganicKeyConfig` against both providers. With the old hardcoded line put back, all 3 call-site tests fail.
- [x] New unit tests: `BridgeViewSQLGenerator.test.ts` (7). With the provider pass-through removed, the detector test fails.
- [x] CodeGenLib suite: 1445 passed / 66 skipped / 0 failed. DBAutoDoc suite: 315 passed. Both packages build.
- [x] Live PostgreSQL 16 (throwaway DB): the old statement fails; the fixed one creates the view, keeps grants on rerun, applies a column rename via the 42P16 fallback, refuses when a user view depends on it, and replays from the migration-log form.
- [x] Live PostgreSQL 16: DBAutoDoc's old bracket body fails with `syntax error at or near "["`; the new PostgreSQL body, wrapped by the CodeGen hook, creates the bridge view, and the runtime transitive lookup returns the right spoke rows.
- [x] Live SQL Server 2022: the old log format (no `GO`) fails replay with `Msg 156`; the fixed format replays, and a column rename applies in place.
- [ ] Reviewer: optionally run `mj codegen` on a PostgreSQL MJ database with a `TransitiveView` in additionalSchemaInfo (not run end-to-end through the full CodeGen CLI).


MS-BC:
## Adversarial review

The provider-hook shape is the right call, and I checked the end-to-end story that the PR doesn't claim: `EntityInfo.BuildTransitiveOrganicKeyFilter` emits the runtime lookup in hardcoded T-SQL brackets (`entityInfo.ts:4074-4076`), but `PostgreSQLDataProvider.quoteIdentifiersInSQL` rewrites `[x]` → `"x"` generically before execution, so the mixed-case aliases DBAutoDoc now emits do line up with the view columns at query time. The DBAutoDoc plumbing is complete too — `OrganicKeyDetector` has exactly one construction site. Changeset levels are right (patch, no migration).

Six things below. The first one I think is a blocker.

---

### 1. SQL Server replay is still broken — the view needs a *leading* `GO`, not just a trailing one

`CREATE OR ALTER VIEW` must be the **first** statement in its batch, not merely the last. The PR fixes the trailing separator (Msg 156) and leaves the leading side, which fails with **Msg 111 — `'CREATE VIEW' must be the first statement in a query batch`**.

`appendToSQLLogFile` only ever appends (`sql_logging.ts:314-318`); nothing anywhere emits a separator *before* a unit. So whatever was logged immediately before the bridge view shares its batch. In a real run that predecessor is a plain `UPDATE …;` with no `GO` — `processEntityConfigs` runs directly before `processOrganicKeyConfig` (`manage-metadata.ts:2554-2562`) — or, for the second and later organic keys, the previous key's `INSERT INTO … EntityOrganicKeyRelatedEntity`.

Resulting log file:

```sql
/* Update attributes on entity Foo */
UPDATE [__mj].[Entity] SET TrackRecordChanges = 1 WHERE ID = '...';

/* Create transitive bridge view ... */
CREATE OR ALTER VIEW [dbo].[vwBridgeFoo]
AS
SELECT ...;
GO
```

Flyway/sqlcmd split on `GO`, so batch 1 is `UPDATE …; CREATE OR ALTER VIEW …` → Msg 111.

The existing precedent shows the shape that works: at `manage-metadata.ts:1643-1644` the wrapper view replays only because the `CREATE TABLE` **before** it passes `includeBatchSeparator: true`. The bridge view has no such guarantee — it is emitted first inside each key's processing, after arbitrary metadata DML.

Repro: any config with a config-declared entity attribute update in the same run, or two organic keys where the second declares a `TransitiveView`.

Fix is probably in the logger rather than the call site — `appendToSQLLogFile` knows whether the last thing it wrote ended in a separator, so a `requiresOwnBatch` flag that emits one before as well (when the file doesn't already end with one) covers this and every future "must be alone in its batch" emitter. Whatever the shape, the test that would have caught it is a second logged unit before the view, not the view alone.

### 2. The outer dollar-quote tag is unguarded

`PostgreSQLCodeGenProvider.ts:479-481` rejects a body containing `$mj_view_sql$`, but the body also travels inside `$mj_create_view$`:

```
DO $mj_create_view$
DECLARE
  vsql CONSTANT TEXT := $mj_view_sql$CREATE OR REPLACE VIEW ... <body> $mj_view_sql$;
```

PostgreSQL's lexer ends a dollar-quoted string at the first literal reoccurrence of *its own* tag — nesting only works with *different* tags, and it does not track depth. A body containing `$mj_create_view$` therefore terminates the `DO` string early and everything after it is parsed as top-level SQL. Same failure mode you guarded against, one level up, and this input is config-sourced (and, via DBAutoDoc, can be LLM-shaped). Check both tags — or pick a tag from a hash of the body so it provably can't collide.

### 3. The ReDoS fix doesn't survive the call it feeds

`trimStatementTerminator` avoids `/[\s;]+$/` correctly. I measured the pattern locally to confirm the CodeQL alert is real, on a body with an internal whitespace run:

| run length | `/[\s;]+$/g` |
|---|---|
| 50,000 | 1,951 ms |
| 200,000 | 31,874 ms |

Quadratic, as you say. But `LogSQLAndExecute` → `appendToSQLLogFile` then applies **`contents.replace(/[\s;]+$/g, '')`** (`sql_logging.ts:302`) to the very string the new helper just produced — the body included — on the same call. So the input that `generateCreateOrReplaceViewSQL` now trims in linear time still stalls CodeGen for ~32s a few frames later whenever SQL output logging is on, which is the normal `mj codegen` path.

That makes this test misleading as written:

```ts
it('trims a body with a long whitespace run in linear time (no regex backtracking)', ...)
expect(Date.now() - started).toBeLessThan(500);
```

True of the generator in isolation, false of the path it's part of. Either fix the logger's trim the same way (it's two lines) or drop the implied end-to-end guarantee from the test name.

### 4. The quiet-failure mechanism the PR indicts is left in place — and the new guard falls into it

The PR's framing is that both bugs were invisible because `processOrganicKeyConfig` catches per key and carries on. That's still exactly what it does: the catch at `manage-metadata.ts:1125-1128` logs and moves on, and the method unconditionally `return { success: true, … }` at `:1132`. The single caller at `:2562` never reads `.success` at all — it only logs the created/updated counts — so the field is dead either way.

Concretely for this PR: the new `throw new Error('… reserved dollar-quote tag …')` lands in that catch. "Refuses rather than emitting broken SQL" is really "silently skips the organic key and reports success" — the same shape as the defect being fixed. Same for the 2BP01 dependency refusal that the no-`CASCADE` decision is built around: it fails loudly in psql, quietly under CodeGen.

This is the third dialect bug in this method to hide the same way (the IS-A one is documented right below it at `:1145-1152`). Worth threading a `failedCount` out and either setting `bSuccess = false` or logging it as prominently as the sequence-integrity check does — otherwise the fourth one costs the same debugging.

### 5. `{ ...DEFAULTS, ...opts }` with an explicit `undefined` defeats the default

`StructuralPhase.ts:41` passes `{ provider }` where `provider` is optional, so the spread writes `provider: undefined` over `DEFAULTS.provider`. `o.provider` is `undefined` at runtime in both `detectTransitiveBridges` and `generateBridgeView`, while `Required<…>` types it as non-optional — the `any`-free strictness MJ asks for, quietly not holding.

It behaves correctly only because `identifierQuoter`'s `switch` has a `default` branch, which means the *"still defaults to SQL Server syntax when no provider is given"* test passes for a different reason than the one it documents. `provider: opts.provider ?? DEFAULTS.provider` (or don't spread the key when undefined) makes the type honest and the test meaningful.

### 6. Nit: the 42P16 fallback drops grants silently

The JSDoc says grants are lost on the drop path, which is the right tradeoff — but at runtime nobody sees it, and the whole point of the no-`CASCADE` choice is that a human should decide. A `RAISE NOTICE` in the exception handler naming the view would put it in the CodeGen output where an operator can act on it.

---

Nothing here disputes the core fix — the hook, the 42P16 fallback, and the no-`CASCADE` posture all look right, and the live-DB test matrix is more than most PRs bring. §1 is the one I'd hold the merge on, since it means the SQL Server migration this PR emits still doesn't replay.

MS-BC (changes requested):
above comment indicates blocker

jordanfanapour:
Thanks for the thorough review. All six points were right, and they're addressed in 82fc28d308:

**1. Leading `GO` (blocker).** Confirmed live on SQL Server: replaying the old log format (a plain `UPDATE`, then the bridge view) fails with `Msg 111`. I took your suggestion and fixed it in the logger:
- `appendToSQLLogFile` / `LogSQLAndExecute` gain a `requiresOwnBatch` option.
- It writes the separator *before* the unit unless the log already ends at a batch boundary, and after it as well.
- "Already ends at a boundary" is decided by reading the tail of the log file, so it holds no matter which emitter wrote the previous unit.
- The bridge view uses it. Any future "must be alone in its batch" emitter can pass the same flag.
- New tests log an `UPDATE`, then the view, then an `INSERT`, and assert three separate batches. They also check that no leading separator is added at the start of the file, none right after a unit that already ends in `GO`, and none on PostgreSQL.
- The regenerated log replays cleanly on SQL Server and creates the view.

**2. Outer dollar-quote tag.** The body is now checked against both `$mj_create_view$` and `$mj_view_sql$`, with a test for each.

**3. ReDoS surviving through the logger.**
- The logger's `/[\s;]+$/g` is replaced with a linear-time scan. Its `/(^|\n)\s*GO\s*$/i` check has the same backtracking shape on whitespace runs, so it's replaced too, and so is the identical trim in `LogSQLBatchAndExecute`.
- All three share `Misc/sql_text.ts` with the provider helper.
- The test now drives a 200,000-character interior whitespace run through the provider **and** `appendToSQLLogFile`, so it covers the path it claims to.

**4. Failures swallowed.**
- `processOrganicKeyConfig` still continues past a failing key, so one bad key doesn't block the others, but it now counts failures and returns `success: false` with a `failedCount`.
- `manageMetadata` logs the count and sets `bSuccess = false`, the same way the neighbouring metadata steps report failure.
- The reserved-tag refusal and the 2BP01 dependency refusal therefore now fail the step instead of being reported as success.
- A test has one broken key and one good key: the good key is still applied and the result reports the failure.

**5. `{ ...DEFAULTS, ...opts }` with an explicit `undefined`.**
- Both `generateBridgeView` and `detectTransitiveBridges` now set `provider: opts.provider ?? DEFAULTS.provider`.
- The quoter's `switch` is exhaustive, with no `default`, so an unhandled provider fails to compile rather than silently getting brackets.
- The "defaults to SQL Server" tests now exercise the default for the reason they say they do. There's a new one for an explicit `{ provider: undefined }`, and it fails if the defaulting is removed.

**6. Silent grant loss on the 42P16 path.** The exception handler now raises `NOTICE: MJ CodeGen: recreated view <schema.view> because its column list changed; grants on it were dropped and must be re-applied`. Confirmed live on PostgreSQL with a column rename.

**Tests:** CodeGenLib 1485 passed, 66 skipped, 0 failed. DBAutoDoc 316 passed, 0 failed. For the new tests on #1 and #5, I checked each one fails when its fix is reverted.

CaelebB-BC:
## Correctness review — one blocking issue

The PostgreSQL DDL work looks right: the provider hook is the correct seam, scoping recovery to SQLSTATE `42P16` only is appropriately narrow, and not using `CASCADE` is the right call for the reason the changeset gives.

### Blocking

**Silent success on entity-not-found — `packages/CodeGenLib/src/Database/manage-metadata.ts:992` and `:1073`**

Both skip paths `logError` + `continue` without incrementing `failedCount`, so the organic-key result still reports `success: true` and the new guard at `:2571` never fires.

These are pre-existing — they sit between this PR's diff hunks — and I'd normally file that as a follow-up. Raising it here because this PR's stated purpose is that CodeGen no longer reports success while the view and the organic key it backs are missing, and `:2571` was added to enforce exactly that. Two sibling paths walk straight past it:

- **`:992`** — owner entity not found in `vwEntities` (a misnamed or miscased table in `additionalSchemaInfo`, or an entity not yet created). Every organic key on that table is skipped, `failedCount` stays 0, `bSuccess` is untouched, and CodeGen exits 0 with no bridge view and no `EntityOrganicKey` rows.
- **`:1073`** — related entity not found (a `RelatedEntities[].TableName` typo). This fires *after* step 1 has already executed `generateCreateOrReplaceViewSQL` and logged the DDL into `CodeGen_Run_*.sql`, so you end up with an orphan view in both the database and the replayable migration, plus an `EntityOrganicKey` row with no transitive mapping — which silently matches nothing at runtime. Still a green run.

`failedCount++` in both places.

MS-BC (approved):

PR #4444, opened 2026-09-13
fix(codegen): PostgreSQL full-text no longer needs a SQL Server precondition

MS-BC:
Two residual full-text defects that #4425 did not reach. Both verified still real on `next` before any code was written.

## 1. A SQL-Server-only precondition was enforced everywhere

CodeGen looked up the entity's primary-key index name whenever `FullTextIndexGenerated` was set — **on every platform** — and that lookup throws when the query returns no rows.

Only SQL Server uses the value; it goes into `CREATE FULLTEXT INDEX … KEY INDEX`. The PostgreSQL provider accepts the parameter and **never references it**: its view-to-table join uses the entity's primary keys from MJ metadata, and the GIN index name is derived from the base table. So a PostgreSQL entity with no primary-key constraint could not produce full-text objects that never needed the index. A tenant worked around it by disabling the feature per entity.

The requirement now belongs to the provider that *consumes* the value — a capability declared `false` on the base and overridden `true` on SQL Server — and the round trip only happens when the provider asks for it.

Concrete rather than `abstract` on purpose: an existing test fake leaves the base's abstract members unimplemented, and a provider that ignores the value should not inherit a precondition it cannot satisfy.

### Proven on a live PostgreSQL 16

Against a table with **zero primary-key constraints** — the shape that was failing:

| check | result |
|---|---|
| primary-key constraints on the table | 0, confirmed from `pg_index` first |
| full DDL applied | clean: trigger function, trigger, GIN index, search function, backfill |
| search function | returns a hit |
| tsvector column | populated on every row |
| sentinel passed in place of the index name | appears **nowhere** in the emitted SQL |

That last row is the load-bearing claim behind dropping the requirement, and it is pinned as a test as well as executed.

## 2. The write-back's error named nothing — and the register's mechanism is already dead upstream

The register says the user cache's array is `undefined`, so indexing it throws a bare `TypeError` before the guard can fire.

**That is no longer reachable on `next`.** The cache defaults its array to empty — commit `f8a8448e0e` (2026-08-11), whose own message describes fixing exactly that `TypeError`. The tenant that hit it runs pinned 5.51.x. The register row was already marked unverified, correctly.

What survives is narrower and still worth fixing:

- The guard checks the **element**, not the array, so it is wrong by construction and only masked by another package's default. Any holder on an older provider still gets the `TypeError`.
- The message named neither the entity, nor the field, nor why the cache was empty, nor any remedy — while misattributing the failure to DDL generation, which needs no user at all. Only the metadata write-back does.

**Kept fatal rather than skipped, against the register's suggested fix.** The write-back is the only path that records the function name. Skipping it would create the function in the database, record nothing, and exit green — leaving runtime search calling a name no row carries. That silent-success shape is the failure mode most of this work exists to remove. The new error names the entity, the field, how the cache is populated, and the operator remedy.

## Composition with #4425

**No conflict.** A merge-tree of both against `next` exits clean, and both test files pass against the merged sources — 26 of 26. The hunks are adjacent but non-overlapping.

My tests assert only that the write-back *happens* and saves *something*, never which name, because the naming scheme is #4425's business — so the pins don't fight.

One note for the merged world: the remedy text interpolates the function name, which under `next` alone is the SQL-Server-shaped name (wrong on PG) and under #4425 is the provider-derived one (right). The sentence is true either way, but the advice only becomes useful once #4425 lands.

## Tests

12 tests. Seven mutations, all red: un-gating the lookup, flipping the SQL Server override, flipping the base default, guarding the element only, restoring the terse message, making the guard never fire, and leaking the index name into PG's emitted DDL.

One initially came back **green** — the mutation only assigned an unused local and never reached the emitted SQL. That was an invalid mutation rather than a missing pin, caught by re-reading the mutation instead of trusting the result, and redone as a real leak into the index comment.

| | Tests |
|---|---|
| CodeGenLib | 1447 passing, 66 skipped, 0 failing |

`tsc --noEmit` clean. The new test file also type-checks under `strict` via an ad-hoc config, since the package's own tsconfig excludes tests and the build would never have caught an error there.

## Two things found and deliberately not fixed

**Worth its own ticket, and the reason both of these read as gibberish in tenant logs:** the separate-files generator ends its catch by logging the error and returning `null`, and that `null` becomes a destructuring failure, then a missing permissions file, then a non-zero exit several frames later. The real message is logged first and then buried. #4425 fixed only the phased-execution caller.

**Also noted:** on PostgreSQL `FullTextIndexGenerated` now gates nothing except the skipped lookup and an empty-search-fields validation — the provider emits the tsvector column, trigger and GIN index regardless. The flag is effectively inert there. Not changed here.

## Not verified

No end-to-end CodeGen run against a live MJ metadata database; the write-back is exercised through a faked metadata object. SQL Server behaviour is pinned by unit tests only — no instance was available, so `CREATE FULLTEXT INDEX … KEY INDEX` was not executed.


rkihm-BC (approved):
Approving.

## The load-bearing claim checks out

> The PostgreSQL provider accepts the parameter and **never references it**.

Verified: zero references to the full-text index name anywhere in `packages/PostgreSQLDataProvider/src`. So CodeGen was enforcing a SQL-Server-only precondition on every platform, and a PostgreSQL entity with no primary-key constraint could not produce full-text objects that never needed the index. Moving the requirement to the provider that *consumes* the value — a capability declared `false` on the base, overridden `true` on SQL Server — puts it where it can be answered.

Making it concrete rather than `abstract` is the right call for the reason given: an existing test fake leaves the base's abstract members unimplemented, and a provider that ignores the value should not inherit a precondition it cannot satisfy.

## The verification is the strongest in this batch

Running it against a live PostgreSQL 16 table with **zero** primary-key constraints — confirming that from `pg_index` first, rather than assuming — and then proving a sentinel passed in place of the index name appears **nowhere** in the emitted SQL, is exactly the right experiment. That last row is the claim the whole change rests on, and pinning it as a test as well as executing it means it stays true.

## Keeping the write-back fatal is the right call

Against the register's suggested fix, and for the right reason: the write-back is the only path that records the function name, so skipping it would create the function in the database, record nothing, and exit green — leaving runtime search calling a name no row carries. A silent success is worse than a loud failure, and that is the failure mode most of this work exists to remove.

Also good: saying plainly that the register's `TypeError` is **no longer reachable on `next`** (the cache defaults its array as of `f8a8448e0e`), and narrowing the claim to what survives — the guard checks the element rather than the array, so it is wrong by construction and only masked by another package's default. Marking the row unverified rather than claiming a fix for something you could not reproduce is the honest version.

## Note

The composition check with #4425 — merge-tree clean, both test files passing against merged sources, 26 of 26 — is the right thing to have done given the adjacency. Worth re-confirming if #4425 moves before either lands, since that evidence has a shelf life.

PR #4445, opened 2026-09-13
fix(core): the largest dataset a tenant has can actually be cached

MS-BC:
Two register rows describing one defect, so they ship as one change.

## The defect

The per-entry cache ceiling was a fixed **25% of a fixed 150 MB budget** — a hard 39 MB. A tenant whose largest metadata dataset item exceeded that got a decline on every boot, so every read of the entity that most needed caching went to the database instead.

Measured on one workspace: **115,571,994 bytes against a 157,286,400 budget**. That is 73.5% of the budget, declined by a 25% ceiling.

## The root cause is stronger than the rows state

The server built its cache config from the workspace settings and **never passed `maxEntryPercentOfCache` at all** — zero occurrences anywhere in MJServer. So the 25% was not a bad default an operator could raise. It was unreachable in every deployment.

## The change

**1. The ceiling is derived, not a fraction chosen for a small tenant.** `'auto'` is the new default: the budget less a single-entry reserve, so 75%. An explicit number remains an exact operator override in either direction, stricter or looser, and `0` still disables the gate.

**2. A hard maximum clamps any configuration**, so no setting can admit an entry that would leave less free space than one eviction pass needs. The reserve must be at least the thrash floor, so a store arriving *after* the large entry is satisfiable out of the reserve rather than by evicting it. The existing per-entity share still holds one entity to a single large slot.

**3. Eviction frees the deficit, not the incoming size.** This is the one that makes the rest safe. It used to free the whole size of the arriving entry, so a 70%-of-budget entry arriving into a 40%-full cache freed 70% — wiping almost everything to make room that mostly already existed.

That behaviour is identical for every entry under 10% of budget, and it is *why* a large entry looked inherently destructive, which is in turn why the ceiling was set low enough to exclude all of them. Raising the ceiling without fixing this would have traded a decline for a stampede.

**4. The decline became actionable.** It names the size, the retainable ceiling, the budget, the fact that every read will now hit the database, and the setting that would admit it.

Plus the setting now reaches the server, with the derived ceiling printed at boot.

## Two corrections to the register

- **The decline was never silent.** The log call prints unconditionally — that is how it was found. It was *unactionable*, not invisible.
- **"The cache is never written" applies to that one entry.** The rest of the cache worked.

The gate's own justification comment was the tell: it argues from an entry larger than the **whole budget**, which is an argument for a ceiling near 100%, and used that to justify one at 25%.

## Rejected

| alternative | why not |
|---|---|
| raise 25 to a larger fixed number | same shape, breaks on the next larger tenant |
| chunk or shard the entry | the constraint is total memory, not per-value storage — splitting saves nothing |
| grow the budget automatically | silently exceeds the operator's memory bound |
| allow unbounded single entries | an entry above the budget genuinely cannot be retained |
| align the ceiling with the 50% per-entity share | coherent, but would not have admitted the reported entry at 73.5% |

## Tests

11 new. Six mutations, all red: restoring the fixed 25% default, dropping the single-entry reserve, dropping the hard clamp on an explicit percentage, restoring the old eviction sizing, removing the lever from the decline message, and removing the ceiling check entirely.

**Three existing tests were changed, and one of them pinned the defect itself as behaviour** — it asserted that an entry just over the cap is skipped. It now sets an explicit 25% and pins the override path instead. The other two were made discriminating: one used an entry the new default caches anyway, and one asserted a call count where eviction legitimately writes through the same provider.

| | Tests |
|---|---|
| baseline | 169 files, 2572 passing |
| with this change | 170 files, **2583 passing** |

MJServer `tsc --noEmit` clean.

## Known, and not papered over

- **The margin is thin.** A tenant whose largest item exceeds 75% of its budget still gets a decline — now naming the setting to raise. Under a fixed memory bound it cannot be otherwise without ignoring the bound.
- **Client cache occupancy really changes.** Entries between 25% and 75% of budget are now retained where they were previously churned, including the large artifact load that seems to have motivated the original gate. It is held to one slot by the per-entity share, but it is a real change.
- **Boot-time attribution is not proven.** The observed schema time has two contributors, and the other was fixed separately. This change demonstrably stops the decline; it does not follow that it removes all of the delay.
- **Type widening** on the config field is API-visible; no in-repo consumer assigns it as a plain number.
- One sibling setter still has no per-entry ceiling. It has no in-repo caller, so it is outside the reported case, and the eviction fix makes it less destructive. Left alone deliberately.

## Not verified

No live tenant or large-schema database. The real-world sizes rest on the measurement above plus arithmetic; the tests pin the *ratio* — a 73%-of-budget entry is cached, a 90% one is not — against a small test budget rather than absolute megabytes.


rkihm-BC:
I merged `next` into this branch (at Robert's direction, commit on the branch above) as part of clearing a batch of PRs whose CI predated #4640. That part worked — the unit-test failure was a stale run of the `agent-init` defect and is now green.

But it changed the Integration result, and that half needs your eyes rather than mine.

**`Integration (SQL Server, deterministic)` was passing on this branch before the merge and fails after it.** `next` itself currently passes that job, so this is not inherited breakage, and it is not the recurring `LoggedIn state` flake — that signature is not in the log.

What the run actually shows:

- **no individual test failed** — IT01 through IT30 all report `Passed`
- **IT29 — Cache Gauntlet (subset-slot × mutation)** reports `Skipped`, with `totalChecks: 0`
- the run is marked `Failed` at the top level

A skipped Cache Gauntlet on a PR about the metadata cache entry cap is the part I did not want to guess at. The plain reading is that merging `next` surfaced a real interaction between this branch's changes and something that landed on `next` while the branch was 61 commits behind — which is the merge doing its job rather than breaking something, but it does mean there is something here to look at before this is mergeable.

Happy to dig further if useful; I stopped short because interpreting a skipped cache test is better done by whoever knows what the gauntlet is asserting.

PR #4450, opened 2026-09-13
fix(provider): two read paths cost more, or fail, than they need to

MS-BC:
Three defects on the shared read path in `@memberjunction/generic-database-provider`. Each is justified by a measurement or by SQL the database rejects, not by a design preference.

## Why one PR

One package, one owner, one code path. `GenericDatabaseProvider` is the read path for **both** dialects — SQL Server and PostgreSQL subclass it — and all three defects live in the two files that make up that path (`GenericDatabaseProvider.ts`, `queryPagingEngine.ts`). Split three ways, each PR would have to re-establish that same context before its one-line change made sense, and a reviewer would read the same file three times. Grouped, the idea reviews once: *two read paths cost more, or fail, than they need to.*

All three are **dialect-neutral**, live on our fleet today via `patch-package`, and absent from both `v5.51.0` and `next`.

---

## 1. `MaxRows:1` paid a second sequential `COUNT(*)` nobody read

A row-limited read that comes back **full** may have been truncated, so the provider buys the real total with a follow-up count. That is what makes Explorer's "100 of 299" correct, and it is **unchanged for every limit above 1**.

At a limit of exactly one row it can only ever confirm a number already in hand, and the callers that use that shape — record-map lookups, single-key resolves — never read `TotalRowCount`. The cost is also **sequential**: unlike the paginated count, this one runs *after* the data query has resolved, so it is a full extra round trip on the critical path.

Measured on a live tenant over a 120-second census:

| probe | reads | counts fired |
|---|---|---|
| record-map lookup | 1,000 | 721 |
| row lookup | 800 | 600 |

**1,321 extra round trips, 9,760 ms — 11.4% of all SQL time in the window.** Nothing read a single result.

The fix adds `maxRowsUsed > 1` to the existing guard. Everything above 1 takes the identical path it took before, so "100 of 299" is preserved exactly.

**Observable change, confined to `MaxRows:1`:** with no count, the return falls back to `rowCount ?? retData.length`, so a hit now reports `TotalRowCount: 1` instead of the whole view's row count. That is the entire difference, and it is asserted in the tests rather than glossed — a limit of one row is not a shape anything asks for in order to learn a total.

## 2. The full-text branch fired with an empty function name and emitted a syntax error

`createViewUserSearchSQL` entered its full-text branch on `entityInfo.FullTextSearchEnabled` alone, then interpolated the function name with `?? ''`:

```
… IN (SELECT "ID" FROM "schema".""('term'))
```

A quoted empty identifier. **A syntax error on every user search of such an entity, with no fallback** — the grid just fails.

`FullTextSearchEnabled` and `FullTextSearchFunction` are independent columns: the flag can be set by hand, or by a metadata sync that runs before CodeGen has minted the function. So the name has to be checked, not assumed. A null/blank/whitespace name now falls through to the per-field `LIKE` path, which needs no database object and returns results.

The new check is **composed with** `next`'s field-level-security condition (`!ftsCoversDeniedField`), not substituted for it — both preconditions must hold to search through the index, and a test covers each clause independently plus the case where both fail.

### This is not a duplicate of #4425

[#4425](https://github.com/MemberJunction/MJ/pull/4425) (`mjc/pg-fulltext-search`) is about **generating** the PostgreSQL full-text objects in CodeGen. This is the **runtime read** side: what the provider emits when the objects are not there. Different failure, different file, different package concern. They are complementary — #4425 stops the gap being created, this stops the gap being a hard error.

## 3. A PostgreSQL query carrying its own `LIMIT` was given a second one

`QueryPagingEngine.buildDataSQL` strips a trailing `TOP` for `sqlserver` and had **no `postgresql` branch**, while appending `LIMIT n OFFSET m` unconditionally. An agent query that already says `LIMIT 20` became:

```
… LIMIT 20 LIMIT 100 OFFSET 0
```

A parse failure. Agent and Skip queries routinely ship with their own `LIMIT`, so those queries could not be paged at all.

Added a `postgresql` branch mirroring the existing `sqlserver` strip, via a new `static stripOuterLimitOffset(sql)` that removes a statement-closing `LIMIT [OFFSET]` and reports the numeric limit it removed. The caller then applies `Math.min(query limit, page size)` — the tighter cap wins, matching what `applyMaxRowsViaAST` already does for an existing numeric cap. `LIMIT ALL` is stripped too (equally illegal to follow with a second `LIMIT`) and reported as no numeric ceiling, so the page size stands.

**End-anchoring is the entire safety argument.** A `LIMIT` inside a CTE body or a subquery is part of that subquery's meaning, not an outer row cap; removing it would silently change which rows come back. End-of-statement is the one position where a `LIMIT` is provably the outer cap this method may replace. Four tests pin that: a subquery `LIMIT`, a CTE-body `LIMIT`, both alongside an outer `LIMIT` that *is* replaced, and the same through the public `WrapWithPaging`. The regex is also deliberately narrow — a trailing `FOR UPDATE`/`FETCH` blocks the match, so it abstains and the pre-existing behaviour stands.

---

## Tests

**+27 tests. Package suite: 1,091 → 1,118 passing (5 skipped), 28 → 29 files. Green.**

- `runViewMaxRowsCount.test.ts` (new): `MaxRows:1` with a hit issues **exactly one** query and reports `TotalRowCount: 1`; a miss also issues one; `MaxRows:100` returning 100 **still** issues the count and reports the real total; `MaxRows:2` returning 2 still counts (the boundary); `MaxRows:100` returning 7 skips it. The fixture's `COUNT(*)` answers with a total far from 1 so a leaked count can never pass for the fallback.
- `createViewUserSearchSQL.test.ts`: an FTS-enabled entity with a null / empty / whitespace-only function name produces **byte-identical** SQL to a non-FTS entity with the same field, and never emits a quoted empty identifier.
- `fieldSecurity.selectList.test.ts`: FTS enabled + a denied FTS-indexed field still takes the denied `LIKE` path (the pre-existing pin on `!ftsCoversDeniedField`), plus the two new cases proving the name check is composed with it rather than replacing it.
- `queryPagingEngine.test.ts`: `stripOuterLimitOffset` unit coverage, and `WrapWithPaging` on PostgreSQL — one `LIMIT` out, tighter cap kept either direction, `LIMIT ALL`, subquery and CTE limits preserved, and the SQL Server path untouched.

## Mutation testing

Every pin was broken deliberately and the suite proven to go red, with the mutation's presence on disk asserted by hash before each run (a no-op edit has produced false greens before) and output ANSI-stripped before parsing. All six caught; all six restored green.

| mutation | file | mutated | restored |
|---|---|---|---|
| `> 1` reverted to truthy | `GenericDatabaseProvider.ts` | 1 failed / 4 passed | green (5) |
| `rowCount ?? retData.length` fallback broken | `GenericDatabaseProvider.ts` | 2 failed / 3 passed | green (5) |
| trim check dropped | `GenericDatabaseProvider.ts` | 5 failed / 72 passed | green (77) |
| `!ftsCoversDeniedField` dropped from the composed condition | `GenericDatabaseProvider.ts` | 1 failed / 76 passed | green (77) |
| `LIMIT` regex de-anchored | `queryPagingEngine.ts` | 5 failed / 59 passed | green (64) |
| `Math.min` replaced by the page size | `queryPagingEngine.ts` | 2 failed / 62 passed | green (64) |

No `any` types. No `pnpm-lock.yaml` churn. Changeset included.


rkihm-BC (approved):
Approving.

## The claim that matters is precisely true

> Everything with `maxRowsUsed > 1` is untouched.

Verified — the change is literally one conjunct added to the existing condition:

```diff
-} else if (countSQL && maxRowsUsed && retData.length === maxRowsUsed) {
+} else if (countSQL && maxRowsUsed && maxRowsUsed > 1 && retData.length === maxRowsUsed) {
```

So Explorer's "100 of 299" takes the identical path it always did, and the carve-out is exactly the case where a full page cannot mean truncation: a limit of one row that came back with one row. There is no judgement call hiding in that predicate.

## The evidence is the right kind

A 120-second census on a live tenant — 1,321 extra round trips, 9,760 ms, **11.4% of all SQL time**, nothing reading a single result — is a measurement rather than an intuition about cost. It also names the specific thing that makes this worse than the paginated count: it runs *after* the data query resolves, so it is a full extra round trip on the critical path rather than a parallel one.

## Stating the observable change rather than glossing it

With no count, the return falls back to `rowCount ?? retData.length`, so a `MaxRows:1` hit now reports `TotalRowCount: 1` instead of the view's total. Saying that plainly and asserting it in tests is the right handling — and the argument for why it is acceptable is sound: a limit of one row is not a shape anything asks for in order to learn a total.

## On grouping three defects

The justification holds. `GenericDatabaseProvider` is the read path for **both** dialects, all three defects live in the two files that make up it, and split three ways a reviewer would re-read the same file three times to re-establish the same context. "Two read paths cost more, or fail, than they need to" is one idea, and it reviews as one.

## Depth

I verified the guard change and the reasoning; I did not independently reproduce the census or exercise the full-text branch fix. Given the size (+616/−10 across 7 files) that leaves less unchecked than in the larger PRs in this batch, but the full-text syntax-error fix in particular would benefit from someone confirming the emitted SQL on a tenant with `FullTextSearchEnabled` set and no function name.

PR #4451, opened 2026-09-13
fix(codegen): the join column CodeGen verified is the one it indexes, declares and walks

MS-BC:
Nine defects that all trace to one thing: **CodeGen decides something about a foreign key or a schema
object and then does not carry that decision to the places that depend on it.** They are grouped
because the verified join column is the common object — the same value that should be declared is the
one that should be indexed and the one the discovery walk should traverse.

## The core defect: `EntityRelationship.EntityKeyField` was never written

`manage-metadata.ts:3959`'s INSERT list omits it, the UPDATE at `:4033-4036` never sets it, and the
identifier `EntityKeyField` appears **nowhere in `packages/CodeGenLib/src`** outside two test fixtures.
So a cross-schema FK pointing at a **non-PK** column left the field blank, and MJ joined on the primary
key instead — **returning zero rows, with no error**.

It is now set from `EntityField.RelatedEntityFieldName` (the `SELECT *` at `:3853` already carries it),
and emitted **only when it differs from the parent's single PK**, so ordinary FK-to-PK output is
byte-identical.

A heal was also required and is included: pass 1's "exact match, skip" at `:4014` never revisits a row
whose other columns are correct, so every pre-existing blank row would have stayed blank forever. The
heal is gated on `AutoUpdateFromSchema`.

## Two consequences of that same value

**Soft-FK index creation indexed the wrong column.** `codeGenDatabaseProvider.ts:541` had a single
predicate — has a `RelatedEntityID`, is not a PK, is not virtual — so for exactly the keys above it
created an index on the primary key rather than on the column the join actually uses. Now
`isIndexableForeignKey(f, verifiedJoinColumns)`, with the set built from
`entity.RelatedEntities[].EntityKeyField` plus `ExactMatch`-only active organic keys.

**The structural discovery walk was unbounded in the wrong dimension.** `StructuralPhase.ts:32`'s soft-FK
guard covered only clusters, while `TransitiveBridgeDetector.ts:229+` does pull soft FKs;
`FKGraphWalker.ts:222-225` allocated a **fresh `Set` per neighbour per node**, and `:211` used
`queue.shift()` on an array. It is now gated on the built edge set, skips hubs and spokes with no edge,
uses an index cursor **with slot release**, and reports its frontier / per-pair / total caps rather than
silently truncating. The per-path `Array.from(hubsByKey.values()).find(...)` became a Map.

## Virtual-field drift had one writer and no detector

An orphaned `_Virtual` column crashes `InnerLoad`, and nothing put the affected entity into pass 2
(`sql_codegen.ts:367-372`; sole writer `manage-metadata.ts:5271`). A probe over
`vwSQLColumnsAndEntityFields` (`EntityFieldID IS NULL AND IsVirtual <> 0`) now unions those entities into
`pass2EntityFilter`.

## The rest

- **PostgreSQL full-text search generation disclosed nothing about its cost.** A grep for
  `warn|cost|overhead|expensive|caution` over `PostgreSQLCodeGenProvider.ts:1078-1145` returns **0**.
  Added `fullTextSearchCostDisclosure()` on the provider base with both overrides.
- **Progress was reported once, not per batch.** The ticker existed (`status_logging.ts:59-71`) but
  `updateSpinner` was called a single time at `sql_codegen.ts:271`. Now `N/total` + phase + elapsed +
  current entity from inside the loop; `updateSpinner` routes to `logStatus` in verbose, so one call
  covers both modes.
- **PG metadata support objects joined un-ANALYZEd tables.** `INNER JOIN __mj."vwSQLTablesAndEntities"`
  was direct and `_ues_filtered` was never ANALYZEd. Now materialised into an ANALYZEd `_ues_tables`,
  with `ANALYZE _ues_filtered` before the `UPDATE ... FROM`.
- **PostgreSQL gained `spRecompileAllViews`**, which `migrations-pg/v5/R__RefreshMetadata.pg-only.sql:57-59`
  already names as its tracked gap. **The reverted scoping half is deliberately untouched** — no change
  to `@ExcludedSchemaNames`, to `05791ec525`/`a3f4f7997c`, or to `omitRecurringScriptsFromLog`.

## Base-view field exclusion: shipped OFF, and here is why that matters

This is the part to review closely, because **the obvious implementation destroys metadata, and that was
measured rather than reasoned.**

`vwSQLColumnsAndEntityFields` resolves columns **view-first** — `COALESCE(e.view_object_id, e.object_id)`
on both dialects (`MJ_BASE_BEFORE_SQL.sql:991`, `metadataSupportObjects.ts:181`). So the prune's real
question is "is this column in the base *view*". Dropping a column from `vwInvoices` and running
`spDeleteUnneededEntityFields` **returned and deleted** the corresponding `EntityField` row. Shipping a
populated default would have silently destroyed metadata on the next CodeGen run.

The originally intended default target is also live: `IntegrationEngine.ts:3597` does
`change.Fields = entity.GetAll()` after `InnerLoad`, and `:3864` reads
`change.Fields['__mj_integration_LastSyncedSnapshot']` — returning `{action:'proceed', attributes: fullAttributes}`
when absent, a **deliberate silent fallback**. Excluding it would quietly disable field-level merge on
bidirectional push. `__mj_integration_SyncStatus`/`LastSyncedAt` are also on the API surface through
`GraphQLDataProvider`'s `FieldMapper`.

So what ships is: `configInfo.baseViewExcludedFields` (bare `FieldName` or `EntityName.FieldName`,
case-insensitive) → `excludedBaseViewFieldNames()` → `baseTableSelectList()`, which returns `${alias}.*`
unless the set is non-empty. Plus a **run-scoped permission interlock**, set during metadata management by
probing `INFORMATION_SCHEMA.PARAMETERS` for a `ProtectedFieldNames` parameter on
`spDeleteUnneededEntityFields`. PG's routine is CodeGen-owned and now declares it.

**Default `[]`, default permission `false` — generated SQL is byte-identical everywhere today and the
golden master does not move.**

On **SQL Server** the prune is a `__mj` migration this PR does not author, so the exclusion **declines with
a message naming the prerequisite** rather than silently losing metadata, and activates by itself the day
that migration lands. Precedent for such a migration exists (`V202607031201` adds an exemption to that same
proc), but authoring a `__mj` proc redefinition that could not be executed here was not acceptable.

Also rejected, with reasons: a new `EntityField.ExcludeFromBaseView` column (a `__mj` migration on both
dialects plus regenerated core entity subclasses, GraphQL types and forms — an enormous diff for one row of
nine, and it would not have removed the prune problem); and projecting the column as `CAST(NULL AS ...)` to
keep the name in the view (dodges the prune, breaks the one real reader).

## Executed against real PostgreSQL 16

The emitted DDL was installed into a throwaway database over a minimal `__mj` fixture (the real
`vwSQLTablesAndEntities` copied verbatim from the PG baseline):

| probe | result |
|---|---|
| prune with `alias.*` view | 0 deleted |
| view without the snapshot column, **unprotected** prune | **the `EntityField` row IS deleted** |
| same with `p_ProtectedFieldNames` | survives, including case/whitespace variance |
| `spRecompileAllViews` | reports `app.vwInvoices` missing the column; 0 rows once the view matches |
| `spUpdateExistingEntitiesFromSchema` | heals the description, idempotent on the second call |
| 3-arg prune signature | still binds |
| `INFORMATION_SCHEMA.PARAMETERS` | finds `p_protectedfieldnames` — and **`PARAMETER_MODE = 'IN'` is required**, because a PG `RETURNS TABLE` function also lists its output columns there. That filter was added as a result |
| virtual-field drift probe | finds a view-only orphan column, correctly ignores a new **physical** orphan |

Scratch database dropped; `pg_database` count 0.

## Verification

| Package | Result |
|---|---|
| CodeGenLib | **1530 passed / 0 failed / 66 skipped** (99 test files passed, 6 skipped) |
| DBAutoDoc | **322 passed / 0 failed** (21 files) |

Both `tsc && tsc-alias` clean. `check:esm` OK; `check:codegen-tail` OK (388 entities, 701 migrations);
`check-migration-no-prune` self-test passes and excludes `R__`, which is the only migration file touched.

Note on the numbers: on the pre-rebase base this branch showed 3 failures in
`entity-subclass-embedded.test.ts`, byte-identical to that base's own baseline. **Rebasing onto current
`next` (`099b4d9634`, which carries #4421) resolved them** — that file now passes all 32 tests. The stderr
output it emits is intentional failure-path logging, not a failure.

**61 mutations, 59 caught.** The harness asserts in code that each edit changed the file, strips ANSI, and
reads both the `Tests` and `Test Files` lines.

Four mutations survived the first pass. Each was re-read rather than written off, and **three were real
gaps**, now pinned:
- progress not called from the loop — added a test driving `generateAndExecuteEntitySQLToSeparateFiles` with
  a stubbed per-entity generator, asserting `[0,3,6]` of 7 at batch size 3;
- the hub-with-no-edge guard, the `maxHops` bound, and defined-only option merge.

The two that remain unpinned, stated rather than hidden:
- one was an **invalid mutation** — the hub-level guard produces identical `pairsSearched`/`pairsSkipped`/`paths`,
  so it is a redundant fast path, not a load-bearing branch;
- the queue **slot release** is a pure retained-memory property of a local array with no observable output. It
  is not pinnable by test and is defended by comment only.

One further mutation (removing the cursor increment) makes the walk **non-terminating** — it hangs the worker
rather than failing — so it is excluded from the count as untestable-by-hang rather than counted as caught.

## Not verified

**The entire SQL Server side is unrun** — no SQL Server instance was available. `spDeleteUnneededEntityFields`'s
SS behaviour, the SS base-view enumeration against a live catalog, and the SS `INFORMATION_SCHEMA.PARAMETERS`
shape are reasoned from the committed proc text in `SQL Scripts/MJ_BASE_BEFORE_SQL.sql` and
`migrations/v6/V202608251800`, not executed. Because the exclusion **declines** on SS by design, the residual
risk is confined to the probe returning a false positive, which would require an SS proc that already has the
parameter.

Also unverified: whether `baseViewExcludedFields` produces a measurable memory win in practice. The mechanism
is off by default and its first intended target needs `computePushCombine` repointed first, which is
`IntegrationEngine.ts` — a file five other open PRs touch.

**Rebase note:** #4443 has not landed and edits all three DBAutoDoc files. The `FKGraphWalker.ts` changes here
touch the `findBridgePaths`/`bfsPaths` signatures, so a rebase will be needed if #4443 merges first.

Closes register MJ-CODEGEN-15, MJ-CODEGEN-22, MJ-CODEGEN-14, MJ-CODEGEN-21, MJ-CODEGEN-6, MJ-CODEGEN-13,
MJ-CODEGEN-7 (residual), MJ-CAT-8 (residual), MJ-MEM-3.


rkihm-BC (approved):
Approving. Reviewed as a draft, so this is "what I checked and what I'd still want", not a merge gate.

## The core claim is exactly true

> `EntityKeyField` appears **nowhere in `packages/CodeGenLib/src`** outside two test fixtures.

Verified on `next`: two hits, both test files, nothing else. That settles the diagnosis — CodeGen never writes the field, so a cross-schema FK pointing at a **non-PK** column leaves it blank and MJ joins on the primary key instead, returning zero rows with no error. A silent wrong answer, which is the worst shape a data-access defect can take.

## Why the grouping is justified

Nine defects in one PR normally deserves pushback. Here it does not: the verified join column **is** the common object — the same value that should be declared is the one that should be indexed and the one the discovery walk should traverse. Split apart, each piece would have to re-establish that the value is unreliable before its own change made sense.

## Two decisions that make a deep change low-risk

**Emitting `EntityKeyField` only when it differs from the parent's single PK.** Ordinary FK-to-PK output stays byte-identical, so the blast radius is confined to exactly the rows that were wrong. That is the right way to change something this far upstream.

**The heal, gated on `AutoUpdateFromSchema`.** Pass 1's "exact match, skip" at `:4014` never revisits a row whose other columns are correct, so without the heal every pre-existing blank row stays blank forever — the fix would only help schemas introspected after it shipped. Catching that the fix alone was insufficient is the part most easily missed.

## What I did not do

I verified the central claim and the structural reasoning; I did not read all 21 files. Given the size and how far upstream this sits, I would want a second pass specifically on the `FKGraphWalker` changes — the fresh-`Set`-per-neighbour allocation and the `queue.shift()` replacement are performance-shaped changes in a graph walk, where an off-by-one in the edge gating silently prunes a traversal rather than failing. Worth someone executing the walk against a schema with known transitive bridges rather than trusting the unit pins.

Everything I checked was accurate, which is a reasonable prior on the rest.

